PR #44, opened 2018-10-04
Fix `recent_block_hashes`

djrtwo:
# Issue
There has been a known bug in which `ActiveState.recent_block_hashes` would not hold the requisite block hashes to validate all potential attestations when it came time during a state recalc. This was due to the list of hashes always being truncated to fit the `2 * CYCLE_LENGTH` requirement even though an attestation could be from longer than that ago when considering large slot skips.

# Proposed solution
I propose that we only append to `recent_block_hashes` until the end of a state recalc, during which we remove CYCLE_LENGTH hashes. In the normal case, the length of `recent_block_hashes` will range between `2 * CYCLE_LENGTH` and `3 * CYCLE_LENGTH`. The length of `recent_block_hashes` might exceed `3 * CYCLE_LENGTH` temporarily during block processing/state recalc, but will always be trimmed down to size at the end.

EDIT: this also allows the `get_block_hash` function/`active_state.recent_block_hashes` to be more usable for during reward calculations. The current spec required using extra DB/local support for tracking hashes.


vbuterin:
>  * Set `active_state.recent_block_hashes = active_state.recent_block_hashes[CYCLE_LENGTH:]` 

Are you sure you don't want  * Set `active_state.recent_block_hashes = active_state.recent_block_hashes[-CYCLE_LENGTH:]` (see the minus before CYCLE_LENGTH) ?

hwwhww:
> attestation could be from longer than that ago when considering large slot skips

Here, do you mean when `block.slot >= parent_block.slot + CYCLE_LENGTH`?
1. Please correct me if I'm wrong, was `active_state.recent_block_hashes` designed for easily getting `get_block_hash`? Alternatively, any chance if we can update `active.recent_block_hashes` **after** attestation validation, so that we can use the `active.recent_block_hashes` in the context of parent block?

2. Increasing the size of `active.recent_block_hashes` might also increase the serialization time of calculating `active_state_root`, although just a little bit I guess. I wonder if it's necessary to be in-protocol? Or we can just allocate a memory to calculate it?

djrtwo:
`recent_block_hashes` is designed to be able to ensure that you have the block hashes required to process the current state transition on hand. The `prev_active_state + prev_crystallized_state + current_block` should be able to compute the state transition for the new active_state and crystallized_State.  The spec as is forgets hashes to quickly from `recent_block_hashes` for this to be the case.

After a state transition in the current proposed spec, the `active_state.recent_block_hashes` is never greater than `3 * CYCLE_LENGTH`

Only sometimes in the middle of a state transition are you holding greater than `3 * CYCLE_LENGTH` recent_block_hashes in memory. The pruning of CYCLE_LENGTH block hashes per state recalc ensures this.

hwwhww:
> After a state transition in the current proposed spec, the `active_state.recent_block_hashes` is never greater than `3 * CYCLE_LENGTH`

> Only sometimes in the middle of a state transition are you holding greater than `3 * CYCLE_LENGTH` recent_block_hashes in memory. The pruning of CYCLE_LENGTH block hashes per state recalc ensures this.

Yep! But my point was, although it’s in memory, it still larger the source of calculating `active_state_root` per block. But I see your point that it’s probably unavoidable.

The exceptional case looks correct to me. 👍

djrtwo:
You only calculate the new `active_state_root` at the end of a state transition, and the current scheme guarantees to trim the size of `recent_block_hashes` down to `< 3 * CYCLE_LENGTH` by the end of the state transition so I don't think it's really an issue.

Waiting on another look from @vbuterin before merge

PR #46, opened 2018-10-05
Fix typo calling AttestationSignedData

JustinDrake:


djrtwo:
cleared up a comment in `AttestationSignedData` too.
merging

PR #48, opened 2018-10-05
Change balances to Gwei

JustinDrake:
See discussion here https://github.com/ethereum/eth2.0-specs/issues/34


djrtwo:
I think this is a good call. I know prysmatic has been seeking a suitable bigint library but this will remove that requirement outside of the context of the WASM implementation

PR #55, opened 2018-10-07
Refactor the sample code and fix #47

### LINKED ISSUE #47: get_active_validator_indices is sometimes called with two arguments
The function `get_active_validator_indices` is defined with one argument but sometimes called with two arguments.

hwwhww:
1. Added type hints, refactored the code, and added some comments for readability
2. Fixed #47

Updated:
Per @paulhauner 's suggestion for optimization, updated `remaining` check in `shuffle` function: break the `position` loop when `remaining == 1`.

### DIFF
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -321,7 +321,6 @@ def shuffle(values: List[Any],
             seed: Hash32) -> List[Any]:
     """
     Returns the shuffled ``values`` with seed as entropy.
-    Mainly for shuffling active validators in-protocol.
     """
     values_count = len(values)
 
@@ -437,7 +436,7 @@ def get_shards_and_committees_for_slot(crystallized_state: CrystallizedState,
 
 def get_block_hash(active_state:ActiveState,
                    current_block: BeaconBlock,
-                   slot: int):
+                   slot: int) -> Hash32:
     earliest_slot_in_array = current_block.slot - CYCLE_LENGTH * 2
     assert earliest_slot_in_array <= slot < earliest_slot_in_array + CYCLE_LENGTH * 2
     return active_state.recent_block_hashes[slot - earliest_slot_in_array]
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -317,61 +317,109 @@ def get_active_validator_indices(validators):
 Now, a function that shuffles this list:
 
 ```python
-def shuffle(lst, seed):
+def shuffle(values: List[Any],
+            seed: Hash32) -> List[Any]:
+    """
+    Returns the shuffled ``values`` with seed as entropy.
+    Mainly for shuffling active validators in-protocol.
+    """
+    values_count = len(values)
+
     # entropy is consumed in 3 byte chunks
     # rand_max is defined to remove the modulo bias from this entropy source
-    rand_max = 2**24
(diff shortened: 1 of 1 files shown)


specs/beacon-chain.md
```diff
@@ -304,61 +304,108 @@ def get_active_validator_indices(validators):
...
+    index = 0
+    while index < values_count:
+        # Re-hash the source
+        source = blake(source)
+        for position in range(0, 30, 3):  # gets indices 3 bytes at a time
+            # Select a 3-byte sampled int
+            sample_from_source = int.from_bytes(source[position:position + 3], 'big')
+            # `remaining` is the size of remaining indices of this round
+            remaining = values_count - index
             if remaining == 0:
```
paulhauner: Optimally this should be `remaining == 1`. If there's only 1 element left the only option is to swap it with itself. It doesn't change the result of the shuffle, it's simply an unnecessary iteration.

mkalinin: And since this is a spec, most part of implementers will just port this algorithm to their languages without thinking much on how it exactly works. Would be great to have it optimal just in case.

hwwhww:
@djrtwo  thanks for the review!
I filled all on_startup setup, but some of them are stub, please take a look again. 🙂 
Also confirm with you, `validator_set_delta_hash_chain` is ZERO_HASH32 at the begining? Or we have to calculate the diff set from empty list `[]` -> `validators` for genesis?

specs/beacon-chain.md
```diff
@@ -317,61 +317,107 @@ def get_active_validator_indices(validators):
...
+    Returns the shuffled ``values`` with seed as entropy.
+    """
+    values_count = len(values)
+
     # entropy is consumed in 3 byte chunks
     # rand_max is defined to remove the modulo bias from this entropy source
-    rand_max = 2**24
-    assert len(lst) <= rand_max
+    rand_max = 2 ** 24
+    assert values_count <= rand_max:
```
paulhauner: I don't think this `:` should be there.

paulhauner:
I had a go at fuzzing the new `shuffle()` algorithm against my reference but it got stuck in a loop. I suspect the issue is with `rand_max` being modified.

There's a bit of an issue with the naming of variables in the shuffle function. They existed before this PR and I list them below:

- The use of the word `MAX` is confusing. It's actually `COUNT` because the maximum number returned from the rng is actually `RAND_MAX - 1`. As it is, is inconsistent with the C++ `cstdlib` [implementation](http://www.cplusplus.com/reference/cstdlib/RAND_MAX/).
- `RAND_MAX` should really be called `sample_max` or something.
- `16777216` should be `RAND_MAX`.

I feel like the easiest way for me to explain how I think this should work is to write it myself. I'd be happy to make a PR with how I think it should work (it'll maintain very much the same form). Alternatively I can list my thoughts here. Let me know what works best for you.

hwwhww:
@paulhauner 
I actually renamed `rand_max` to `SAMPLE_RANGE` in Py-EVM. `SAMPLE_MAX` makes sense to me too. And yes please make a PR for describing your thought! I think that will be great. :)

hwwhww:
@djrtwo  just `validator_set_delta_hash_chain` setting. But I think we can leave that in another PR.

PR #61, opened 2018-10-09
Removed all uses of the word dynasty, and replaced the functionality …

vbuterin:
…of dynasties with other variables.


djrtwo:
This looks good to me.

@hwwhww, @JustinDrake any comments before merge?

specs/beacon-chain.md
```diff
@@ -587,20 +580,20 @@ For each `SpecialRecord` `obj` in `active_state.pending_specials`:
...
-### Dynasty transition
+### Validator set change
 
-A dynasty transition can happen after a state recalculation if all of the following criteria are satisfied:
+A validator set change can happen after a state recalculation if all of the following criteria are satisfied:
 
-* `block.slot - crystallized_state.dynasty_start_slot >= MIN_DYNASTY_LENGTH`
-* `last_finalized_slot > dynasty_start_slot`
-* For every shard number `shard` in `shard_and_committee_for_slots`, `crosslinks[shard].slot > dynasty_start_slot`
+* `block.slot - crystallized_state.validator_set_change_slot >= MIN_DYNASTY_LENGTH`
```
hwwhww: Do you want to also rename `MIN_DYNASTY_LENGTH`?

specs/beacon-chain.md
```diff
@@ -242,8 +236,8 @@ A `CrosslinkRecord` has the following fields:
 
 ```python
 {
-    # Dynasty number
-    'dynasty': 'int64',
+    # Since last validator set change?
+    'recently_changed': 'int8',
```
JustinDrake: Somewhat more semantic/readable as type `bool`

djrtwo: bool is not currently supported in ssz. If we add it here, need to add it to the ssz spec

djrtwo: (which I support)

vbuterin:
Done both edits.

djrtwo:
Cool. just resolved merge conflicts. merging

PR #62, opened 2018-10-09
Remove multiple slots per committee option

vbuterin:
As discussed last week.

Rationale is that the `slots_per_committee > 1` case only appears when the validator count is less than 8192, or ETH deposited is less than 262144; in these cases, having smaller committees is not a big deal as the system is going to not be extremely secure in any case.

### DIFF
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -378,8 +378,6 @@ def clamp(minval: int, maxval: int, x: int) -> int:
         return maxval
     else:
         return x
-def clamp(minval, maxval, x):
-    return minval if x < minval else maxval if x > maxval else x
 ```
 
 Now, our combined helper method:
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -348,25 +348,25 @@ def split(lst, N):
     return [lst[len(lst)*i//N: len(lst)*(i+1)//N] for i in range(N)]
 ```
 
+A helper method for readability:
+
+```python
+def clamp(minval, x, maxval):
+    return minval if x < minval else maxval if x > maxval else x
+```
+
 Now, our combined helper method:
 
 ```python
 def get_new_shuffling(seed, validators, crosslinking_start_shard):
     active_validators = get_active_validator_indices(validators)
-    if len(active_validators) >= CYCLE_LENGTH * MIN_COMMITTEE_SIZE:
-        committees_per_slot = min(len(active_validators) // CYCLE_LENGTH // (MIN_COMMITTEE_SIZE * 2) + 1, SHARD_COUNT // CYCLE_LENGTH)
-        slots_per_committee = 1
-    else:
-        committees_per_slot = 1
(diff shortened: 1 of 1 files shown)


specs/beacon-chain.md
```diff
@@ -348,25 +348,25 @@ def split(lst, N):
...
-    if len(active_validators) >= CYCLE_LENGTH * MIN_COMMITTEE_SIZE:
-        committees_per_slot = min(len(active_validators) // CYCLE_LENGTH // (MIN_COMMITTEE_SIZE * 2) + 1, SHARD_COUNT // CYCLE_LENGTH)
-        slots_per_committee = 1
-    else:
-        committees_per_slot = 1
-        slots_per_committee = 1
-        while len(active_validators) * slots_per_committee < CYCLE_LENGTH * MIN_COMMITTEE_SIZE \
-                and slots_per_committee < CYCLE_LENGTH:
-            slots_per_committee *= 2
+    committees_per_slot = clamp(1,
```
djrtwo: No need for the `clamp` here. The min value of `len(active_validators) // CYCLE_LENGTH // (MIN_COMMITTEE_SIZE * 2) + 1` is `1` due to the `+ 1` term.

hwwhww: @djrtwo do you mean only using `min()`?
```python
committees_per_slot = min(
    len(active_validators) // CYCLE_LENGTH // (MIN_COMMITTEE_SIZE * 2) + 1,
    SHARD_COUNT // CYCLE_LENGTH
)
```
 There is a chance that it will return 0 if we set `SHARD_COUNT < CYCLE_LENGTH`. (e.g., initial test net?)

specs/beacon-chain.md
```diff
@@ -348,25 +348,25 @@ def split(lst, N):
     return [lst[len(lst)*i//N: len(lst)*(i+1)//N] for i in range(N)]
 ```
 
+A helper method for readability:
+
+```python
+def clamp(minval, x, maxval):
+    return minval if x < minval else maxval if x > maxval else x
```
hwwhww: Shoud be `return minval if x <= minval else maxval if x >= maxval else x`. Otherwise, we have to ensure that `minval 
 < x < maxval` in somewhere else.

Easy-to-read version:
```python
def clamp(minval: int, x: int, maxval: int) -> int:
    if x <= minval:
        return minval
    elif x >= maxval:
        return maxval
    else:
        return x
```

specs/beacon-chain.md
djrtwo: You're right.

I think we should not allow `SHARD_COUNT < CYCLE_LENGTH`. It would result in more than one committee (from more than one slot) working on crosslinking a single shard. This is a similar corner case that would increase complexity of the code but for no tangible reason. Especially because these are constants that we define once for the protocol rather than ones that are in flux (like num validators)

djrtwo:
cleaned up and handled conflict. merging

PR #65, opened 2018-10-11
Fix shuffle() function errors

paulhauner:
I updated the `shuffle()` function as discussed in #55.

The following changes were implemented:

1. `sample_max` was made distinct from `rand_max`. 
2. `rand_max` now represents the greatest value returned by the RNG
instead of the length of the range. This is consistent with the [cstdlib RAND_MAX constant](http://www.cplusplus.com/reference/cstdlib/RAND_MAX/).
3. `while` condition fixed to stop infinite loop.
4. Comments updated.

After fixing (2) I noticed that there was an error in generating `sample_max` caused by `rand_max` being too high. **This affects the output of the shuffle.** All clients should update their shuffler if this PR is accepted.

I have used the [shuffling_sandbox](https://github.com/sigp/shuffling_sandbox/) to fuzz this code against my [reference implementation](https://github.com/sigp/shuffling_sandbox/blob/master/src/shufflers/reference.py).

### DIFF
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -324,36 +324,39 @@ def shuffle(values: List[Any],
     """
     values_count = len(values)
 
-    # entropy is consumed in 3 byte chunks
-    # sample_max is defined to remove the modulo bias from this entropy source
-    sample_max = 2 ** 24
-    assert values_count <= sample_max
+    # Entropy is consumed from the seed in 3-byte (24 bit) chunks.
+    rand_max = 2 ** 24 - 1
+    assert values_count < rand_max
 
     output = [x for x in values]
     source = seed
     index = 0
-    while index < values_count:
+    while index < values_count - 1:
         # Re-hash the source
         source = hash(source)
-        for position in range(0, 30, 3):  # gets indices 3 bytes at a time
-            # Select a 3-byte sampled int
-            sample_from_source = int.from_bytes(source[position:position + 3], 'big')
-            # `remaining` is the size of remaining indices of this round
+        for position in range(0, 30, 3):  # Reads indices 3 bytes at a time
+            # Determine the number of indices remaining and exit once the last
+            # index is reached.
             remaining = values_count - index
             if remaining == 1:
                 break
 
-            # Set a random maximum bound of sample_from_source
-            sample_max = sample_max - sample_max % remaining
+            # Read 3-bytes of the seed as a 24-bit big-endian integer.
+            sample_from_source = int.from_bytes(source[position:position + 3], 'big')
 
-            # Select `replacement_position` with the given `sample_from_source` and `remaining`
+            # Sample values greater than or equal to `sample_max` will cause
+            # modulo bias when mapped into the `remaining` range.
+            sample_max = rand_max - rand_max % remaining
+
+            # Perform a swap if the consumed entropy will not cause modulo bias.
             if sample_from_source < sample_max:
-                # Use random number to get `replacement_position`, where it's not `index`
+                # Select a replacement index for the present index.
                 replacement_position = (sample_from_source % remaining) + index
-                # Swap the index-th and replacement_position-th elements
+                # Swap the present index with the replacement index.
                 output[index], output[replacement_position] = output[replacement_position], output[index]
                 index += 1
             else:
+                # The sample causes modulo bias. A new sample should be read.
                 pass
 
     return output
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -325,24 +325,32 @@ def shuffle(values: List[Any],
     values_count = len(values)
 
     # Entropy is consumed from the seed in 3-byte (24 bit) chunks.
-    rand_max = 2 ** 24 - 1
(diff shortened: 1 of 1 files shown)


specs/beacon-chain.md
```diff
@@ -324,36 +324,39 @@ def shuffle(values: List[Any],
...
+    rand_max = 2 ** 24 - 1
+    assert values_count < rand_max
 
     output = [x for x in values]
     source = seed
     index = 0
-    while index < values_count:
+    while index < values_count - 1:
         # Re-hash the source
         source = hash(source)
```
JustinDrake: Out of curiosity, why are we rehashing the source? Isn't it sufficient to use `seed` as is?

specs/beacon-chain.md
```diff
@@ -324,36 +324,39 @@ def shuffle(values: List[Any],
...
-        for position in range(0, 30, 3):  # gets indices 3 bytes at a time
```
JustinDrake: The number `3` is used here and in `position + 3`. Maybe worth having a constant (similar to `rand_max`) for that.

JustinDrake: The number `30` is not immediately obvious. Consider replacing by `31` (the max index) `- 3` (the constant above).

JustinDrake: For consistency consider putting the comment on top (changing "Reads" by "Read"). Also notice there are two spaces before the `#`.

specs/beacon-chain.md
```diff
@@ -324,36 +324,39 @@ def shuffle(values: List[Any],
     """
     values_count = len(values)
 
-    # entropy is consumed in 3 byte chunks
-    # sample_max is defined to remove the modulo bias from this entropy source
-    sample_max = 2 ** 24
-    assert values_count <= sample_max
```
JustinDrake: I'm a bit scared by `assert`s in consensus code. What happens if the `assert` is hit? Does the beacon chain error out and stall?

specs/beacon-chain.md
paulhauner: The seed is re-hashed on two occasions:

1. On the first iteration of the shuffle.
2. Each time the seed is exhausted (i.e., we perform more than `floor(32 / 3)` iterations).

Assuming the seed initially supplied to the function is 32-bytes of entropy, I don't see an issue with removing the first (1) rehash.

specs/beacon-chain.md
paulhauner: Seems like a good idea to me. We could also use this constant to generate `rand_max`. E.g., `rand_max = 2**(x * 8) - 1`.

paulhauner: Agreed

paulhauner: I agree with putting the comment above the line.

FYI, PEP 8 [specifies](https://www.python.org/dev/peps/pep-0008/#inline-comments) that "inline comments should be separated by at least two spaces from the statement."

specs/beacon-chain.md
paulhauner: I'll leave this one for @djrtwo or @hwwhww.

paulhauner:
Thanks for the comments @JustinDrake!

I've added my responses and will await further direction from the repository maintainers.

specs/beacon-chain.md
djrtwo: Confusing either way imo 😬

specs/beacon-chain.md
djrtwo: The shuffling algorithm is independent of what it is shuffling. It just so happens that it will often (always?) be used to shuffle validators. I think it makes sense to restrict the number of validators somewhere else in the code (either implicitly via deposit size or explicitly with a constant)

This assertion in practice should be handled independently for each client. Assertions in python are bad practice because they can trivially be turned off at runtime.

specs/beacon-chain.md
paulhauner: This is interesting.. I suspect it should be `range(0, 32, 3)` and we're throwing away 3 bytes.

Will investigate and report back.

paulhauner: However, that's only sensible when `rand_bytes % seed_bytes != 0`.

Will think on this some more after the call.

djrtwo: that works for me.

paulhauner:
Thanks for the feedback. I will implement the following changes:

1. Introduce a `rand_bytes` constant set to `3` and use it wherever possible.
2. Set the range on the byte-sampling `for` loop to: `range(0, 32 - (32 % rand_bytes), rand_bytes)`
3. Ensure all comments are on the line above to the statement (instead of inline).

I could use additional direction regarding `assert values_count <= sample_max`. I see @hwwhww suggested throwing a `ValueError`, is this how we should handle it?

djrtwo:
I'm cool with an assert in the spec code as long as there is a comment about why it is there and how it is up to a client on how to proceed. Similar to the commented assert in `add_validator` 

If we want to remove asserts entirely (there are many), we can do it in a separate issue.

djrtwo:
lgtm!

PR #69, opened 2018-10-11
Add RANDAO multi-hashing

vbuterin:
Requires more hashes in a RANDAO reveal the more time passes. This greatly reduces the chance that a RANDAO reveal accidentally revealed during an orphaned block will be usable for future blocks.

### DIFF
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -611,7 +611,7 @@ Extend the list of `AttestationRecord` objects in the `active_state` with those 
 
 Let `proposer_index` be the validator index of the `parent.slot % len(get_shards_and_committees_for_slot(crystallized_state, parent.slot)[0].committee)`'th attester in `get_shards_and_committees_for_slot(crystallized_state, parent.slot)[0]`. Verify that an attestation from this validator is part of the first (ie. item 0 in the array) `AttestationRecord` object; this attester can be considered to be the proposer of the parent block. In general, when a block is produced, it is broadcasted at the network layer along with the attestation from its proposer.
 
-Additionally, we need to verify and update the RANDAO reveal. This is done as follows:
+Additionally, verify and update the RANDAO reveal. This is done as follows:
 
 * Let `repeat_hash(x, n) = x if n == 0 else repeat_hash(hash(x), n-1)`.
 * Let `V = crystallized_state.validators[proposer_index]`.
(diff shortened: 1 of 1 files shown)


specs/beacon-chain.md
```diff
@@ -605,7 +611,11 @@ Extend the list of `AttestationRecord` objects in the `active_state` with those
 
 Let `proposer_index` be the validator index of the `parent.slot % len(get_shards_and_committees_for_slot(crystallized_state, parent.slot)[0].committee)`'th attester in `get_shards_and_committees_for_slot(crystallized_state, parent.slot)[0]`. Verify that an attestation from this validator is part of the first (ie. item 0 in the array) `AttestationRecord` object; this attester can be considered to be the proposer of the parent block. In general, when a block is produced, it is broadcasted at the network layer along with the attestation from its proposer.
```
djrtwo: We should define this as `parent_proposer_index` and define `proposer_index` separately below.

specs/beacon-chain.md
```diff
@@ -605,7 +611,11 @@ Extend the list of `AttestationRecord` objects in the `active_state` with those
...
 
-Additionally, verify that `hash(block.randao_reveal) == crystallized_state.validators[proposer_index].randao_commitment`, and set `active_state.randao_mix = xor(active_state.randao_mix, block.randao_reveal)` and append to `ActiveState.pending_specials` a `SpecialObject(kind=RANDAO_CHANGE, data=[bytes8(proposer_index), block.randao_reveal])`.
+Additionally, verify and update the RANDAO reveal. This is done as follows:
+
+* Let `repeat_hash(x, n) = x if n == 0 else repeat_hash(hash(x), n-1)`.
+* Let `V = crystallized_state.validators[proposer_index]`.
```
djrtwo: Define `proposer_index` here.

specs/beacon-chain.md
```diff
@@ -605,7 +611,11 @@ Extend the list of `AttestationRecord` objects in the `active_state` with those
...
+* Verify that `repeat_hash(block.randao_reveal, (current_slot - V.randao_last_reveal) // RANDAO_SLOTS_PER_LAYER + 1) == V.randao_commitment`, and set `active_state.randao_mix = xor(active_state.randao_mix, block.randao_reveal)` and append to `ActiveState.pending_specials` a `SpecialObject(kind=RANDAO_CHANGE, data=[bytes8(proposer_index), block.randao_reveal])`.
```
djrtwo: Use `block.slot` rather than `current_slot`

djrtwo: Consider making the "verify", "set", and "append" from this statement as different bullet points

vbuterin:
Fixed @djrtwo's issue about `proposer` vs `parent_proposer`.

djrtwo:
Thoughts on changing `current_slot` to `block.slot` when passing it in as param?

I'm worried `current_slot` could be interpreted as the current global slot based on the current time rather than the slot of the block that we happen to be processing.

PR #73, opened 2018-10-12
Re-add removed rationale for min. committee size

jamesray1:


JustinDrake:
A separate document will explain the rationale behind various design choices.

djrtwo:
I'm cool with adding it in there until we have such a document so it doesn't get lost in the mix.

jamesray1:
> I'm cool with adding it in there until we have such a document so it doesn't get lost in the mix.

Thanks, I was going to propose the same thing.

PR #75, opened 2018-10-14
Remove remaining usage of dynasty

rawfalafel:
Looks like it was accidentally brought back in after it was removed in an earlier commit.


djrtwo:
nice! added `validator_set_change_slot` to the init params in that call.

PR #77, opened 2018-10-14
Bug in update_ancestor_hashes?

terenc3t:
Not sure if this is a bug. 

By checking `parent_slot_number % 2**i == 0`, all the parent hashes before i gets updated as well.

**For example:**
parent_slot_number: 32, parent_hash: 0xA
ancestor_hashes[0], [1], [2], [3], [4] will all have 0xA


vbuterin:
This is intentional. `ancestor_hashes[i]` is meant to represent the most recent height with `height % 2**i == 0`, so of course at large powers of two it's going to update all lower powers as well.

terenc3t:
Ahh, I see. Thanks for confirming!

PR #81, opened 2018-10-16
Clarify description in ancestor_hashes

rawfalafel:
`i'th ancestor` really means `most recent ancestor that is a multiple of 2**i`, not `go up the chain i times`.

If I'm understanding the code correctly, that is :)


djrtwo:
lgtm

PR #85, opened 2018-10-16
Misc minor cleanups

JustinDrake:


specs/beacon-chain.md
```diff
@@ -300,22 +300,16 @@ Here's an example of its working (green is finalized blocks, yellow is justified
...
 
 ### Helper functions
 
-We start off by defining some helper algorithms. First, the function that selects the active validators:
-
-```python
-def get_active_validator_indices(validators):
-    return [i for i, v in enumerate(validators) if v.status == ACTIVE]
-```
+Below are various helper functions.
```
djrtwo: Why nuke `get_active_validator_indices`?

djrtwo:
I also fixed the comment about `recent_changed`

specs/beacon-chain.md
hwwhww: Support for `get_active_validator_indices` +1

JustinDrake: I don't feel strongly either way. 

1) It's a one-liner
2) It's only used twice
3) It's readable as is, especially with the variable name `active_validators`

djrtwo:
@JustinDrake I want to integrate the other changes. Can I cleanup and remove the removal of `get_active_validator_indices`?

JustinDrake:
@djrtwo Go ahead :)

PR #86, opened 2018-10-16
clarify attestation justified slot validity conditions

### LINKED ISSUE #67: justified_block_hash.slot != justified_slot
With reference to the following:

> Verify that the `justified_slot` and `justified_block_hash` given are in the chain and are equal to or earlier than the `last_justified_slot` in the crystallized state.

Is it intended that the block identified by `justified_block_hash` has a `slot_number` that does not equal `justified_slot`? In other words, can an `AttestationRecord` point to a `justified_slot` which contains a block other than `justified_block_hash`?

With the present wording in the spec, I believe that is possible.

Thanks!

djrtwo:
Resolves #67

### DIFF
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -632,7 +632,8 @@ A block can have 0 or more `AttestationRecord` objects
 For each one of these attestations:
 
 * Verify that `slot <= parent.slot` and `slot >= max(parent.slot - CYCLE_LENGTH + 1, 0)`
-* Verify that the `justified_slot` and `justified_block_hash` given are in the chain and are equal to or earlier than the `last_justified_slot` in the crystallized state.
+* Verify that `justified_slot` is equal to or earlier than `last_justified_slot`
+* Verify that `justified_block_hash` is the hash of the block in the current chain at the slot -- `justified_slot`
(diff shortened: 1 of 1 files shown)


specs/beacon-chain.md
```diff
@@ -632,7 +632,8 @@ A block can have 0 or more `AttestationRecord` objects
 For each one of these attestations:
 
 * Verify that `slot <= parent.slot` and `slot >= max(parent.slot - CYCLE_LENGTH + 1, 0)`
-* Verify that the `justified_slot` and `justified_block_hash` given are in the chain and are equal to or earlier than the `last_justified_slot` in the crystallized state.
+* Verify that `justified_slot` is equal to or earlier than `last_justified_slot`
+* Verify that `justified_block_hash` is the hash of the block in the current chain at the slot -- `justified_slot`
```
hwwhww: nitpick: add period `.` at the end of all these sentences.

specs/beacon-chain.md
```diff
@@ -632,7 +632,8 @@ A block can have 0 or more `AttestationRecord` objects
 For each one of these attestations:
 
 * Verify that `slot <= parent.slot` and `slot >= max(parent.slot - CYCLE_LENGTH + 1, 0)`
```
hwwhww: Not in this PR, but just noticed that `parent` hasn't been defined.

djrtwo: addressed here https://github.com/ethereum/eth2.0-specs/pull/86/commits/9b9cfae117bb1b30975cc90759d3a6c5f3c9fa64

PR #88, opened 2018-10-18
Minor edit for style consistency

ralexstokes:
There is a function given in text where similar utilities are given as python-like code.

This PR adjusts one such instance for style consistency with the rest of the document.

Feel free to close if the previous representation is preferred.


ralexstokes:
Do I need to push another commit that signs this branch or is this not the right venue to suggest these changes?

djrtwo:
Lgtm

PR #89, opened 2018-10-18
Assign validators to persistent committees for shard block production

vbuterin:
This adds a persistent committee assignment that changes once every `SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD` slots (currently set to ~12 days).

### DIFF
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -42,6 +42,7 @@ The primary source of load on the beacon chain are "attestations". Attestations 
 | `RANDAO_SLOTS_PER_LAYER` | 2**12 (= 4096) | slots | ~18 hours |
 | `SQRT_E_DROP_TIME` | 2**16 (= 65,536) | slots | ~12 days |
 | `WITHDRAWAL_PERIOD` | 2**19 (= 524,288) | slots | ~97 days |
+| `SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD` | 2**16 (= 65,536) | slots | ~12 days |
 | `BASE_REWARD_QUOTIENT` | 2**15 (= 32,768) | — |
 | `MAX_VALIDATOR_CHURN_QUOTIENT` | 2**5 (= 32) | — |
 | `LOGOUT_MESSAGE` | `"LOGOUT"` | — |
@@ -205,6 +206,9 @@ The `CrystallizedState` has the following fields:
     'justified_streak': 'uint64',
     # Committee members and their assigned shard, per slot
     'shard_and_committee_for_slots': [[ShardAndCommittee]],
+    # Persistent shard committees
+    'current_persistent_committees': [['uint24']],
+    'next_persistent_committees': [['uint24']],
     # Total deposits penalized in the given withdrawal period
     'deposits_penalized_in_period': ['uint32'],
     # Hash chain of validator set changes (for light clients to easily track deltas)
@@ -441,6 +445,15 @@ def get_new_shuffling(seed: Hash32,
 Here's a diagram of what's going on:
 
 ![](http://vitalik.ca/files/ShuffleAndAssign.png?1)
+
+We also make a function for generating persistent committees:
+
+```python
+def get_persistent_shuffling(validators: List[ValidatorRecord], seed: Hash32):
+    active_validators = get_active_validator_indices(validators)
+    shuffled_active_validator_indices = shuffle(active_validators, seed)
+    return split(shuffled_active_validator_indices, SHARD_COUNT)
+```
 
 We also define two functions for retrieving data from the state:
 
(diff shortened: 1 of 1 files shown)


specs/beacon-chain.md
```diff
@@ -442,6 +446,15 @@ Here's a diagram of what's going on:
 
 ![](http://vitalik.ca/files/ShuffleAndAssign.png?1)
 
+We also make a function for generating persistent committees:
+
+```python
+def get_persistent_shuffling(validators: List[ValidatorRecord], seed: Hash32):
```
hwwhww: fix type hint:
```python
def get_persistent_shuffling(validators: List[ValidatorRecord], seed: Hash32) -> List[List[ValidatorRecord]]:
```

specs/beacon-chain.md
```diff
@@ -791,13 +795,32 @@ def change_validators(validators: List[ValidatorRecord]) -> None:
...
+
+For any validator that was added or removed from the active validator list during this state recalculation:
+
+* If the validator was removed, remove their index from the `current_persistent_committees` and `next_persistent_committees` objects.
+* If the validator was added with index `validator_index`, let `assigned_shard = hash(crystallized_state.randao_mix + bytes8(validator_index)) % SHARD_COUNT`. Add `validator_index` to the end of `next_persistent_committees[assigned_shard]`.
+
+If `block.slot % SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD == 0`, then:
+
+* Set `current_persistent_committees = next_persistent_committees`
+* Set `next_persistent_committees = get_persistent_shuffling(validators, crystallized_state.randao_mix)`
```
djrtwo: I thought we were going to slowly, but consistently, be shuffling small amounts of validators at each cycle (or validator change). So we never had _all_ of the validators changing at once. I see instead of going that route (which would certainly be more complicated algorithmically) you are instead changing all of the validators at once, but giving them a 12 day look ahead.

I suppose this is reasonable :)

vbuterin: Could try to come up with a version that changes more gradually. Perhaps something like saving a list of "future shard change records", and every cycle adding a record for 12 days from now that moves a randomly selected active validator index to a random shard?

I'd totally be ok with that approach.

specs/beacon-chain.md
```diff
@@ -791,13 +795,32 @@ def change_validators(validators: List[ValidatorRecord]) -> None:
...
 * Set `crystallized_state.validator_set_change_slot = crystallized_state.last_state_recalculation_slot`
 * For all `c` in `crystallized_state.crosslinks`, set `c.recently_changed = False`
 * Let `next_start_shard = (shard_and_committee_for_slots[-1][-1].shard + 1) % SHARD_COUNT`
 * Set `shard_and_committee_for_slots[CYCLE_LENGTH:] = get_new_shuffling(active_state.randao_mix, validators, next_start_shard)`
 
+#### Finally...
+
+* For any validator with index `v` with balance less than `MIN_ONLINE_DEPOSIT_SIZE` and status `ACTIVE`, run `exit_validator(v, crystallized_state, penalize=False, current_slot=block.slot)`
+* Set `crystallized_state.last_state_recalculation_slot += CYCLE_LENGTH`
+* Remove all attestation records older than slot `crystallized_state.last_state_recalculation_slot`
```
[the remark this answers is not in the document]
vbuterin: Maybe the right way to figure this out is to agree on this PR or an alternative/edits (eg. shuffling over time instead of all at once) first, and then after that make a full coherent attempt at reordering and making sure the order of operations is fully correct?

specs/beacon-chain.md
```diff
@@ -791,13 +795,46 @@ def change_validators(validators: List[ValidatorRecord]) -> None:
...
+### Possible alternative (run every state recalculation):
+
+active_validator_indices = get_active_validator_indices(validators)
+for i in range(len(active_validator_indices) // SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD):
+    vid = active_validator_indices[hash(active_state.randao_mix + bytes8(i * 2)) % len(active_validator_indices)]
+    new_shard = hash(active_state.randao_mix + bytes8(i * 2 + 1)) % SHARD_COUNT
+    crystallized_state.persistent_shard_reassignments.append(ShardReassignmentRecord(validator_id=vid, shard=new_shard, slot=block.slot + SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD))
+    
+while len(crystallized_state.persistent_shard_reassignments) > 0 and crystallized_state.persistent_shard_reassignments[0].slot <= block.slot:
+    rec = crystallized_state.persistent_shard_reassignments[0]
```
hwwhww: Should this line be `rec = crystallized_state.persistent_shard_reassignments.pop(0)`?

specs/beacon-chain.md
```diff
@@ -791,13 +795,46 @@ def change_validators(validators: List[ValidatorRecord]) -> None:
...
+* Remove all attestation records older than slot `crystallized_state.last_state_recalculation_slot`
+* Empty the `active_state.pending_specials` list
+* Set `active_state.recent_block_hashes = active_state.recent_block_hashes[CYCLE_LENGTH:]`
+* Set `shard_and_committee_for_slots[:CYCLE_LENGTH] = shard_and_committee_for_slots[CYCLE_LENGTH:]`
 * Let `next_start_shard = (shard_and_committee_for_slots[-1][-1].shard + 1) % SHARD_COUNT`
 * Set `shard_and_committee_for_slots[CYCLE_LENGTH:] = get_new_shuffling(active_state.randao_mix, validators, next_start_shard)`
 
+For any validator that was added or removed from the active validator list during this state recalculation:
+
+* If the validator was removed, remove their index from the `current_persistent_committees` and `next_persistent_committees` objects.
```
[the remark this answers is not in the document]
djrtwo: This is in an effort to give validators a lookahead in the version where we only store current and next persistent committees (rather than constantly, slowly reshuffling)

specs/beacon-chain.md
```diff
@@ -791,13 +795,46 @@ def change_validators(validators: List[ValidatorRecord]) -> None:
...
+* If the validator was added with index `validator_index`, let `assigned_shard = hash(active_state.randao_mix + bytes8(validator_index)) % SHARD_COUNT`. Add `validator_index` to the end of `next_persistent_committees[assigned_shard]`.
+
+If `block.slot % SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD == 0`, then:
```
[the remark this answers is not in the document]
djrtwo: With the current setup, you would need to track `last_persistent_committee_change_slot` and use the `>=` logic. Similar to `last_state_recalculation_slot`

specs/beacon-chain.md
```diff
@@ -791,13 +795,46 @@ def change_validators(validators: List[ValidatorRecord]) -> None:
...
+
+* Set `current_persistent_committees = next_persistent_committees`
+* Set `next_persistent_committees = get_persistent_shuffling(validators, active_state.randao_mix)`
+
+```
+### Possible alternative (run every state recalculation):
+
```
djrtwo: +1
I prefer this method to the two lists that change at once.

It also nicely fits into the state recalc model.

specs/beacon-chain.md
djrtwo: yes

djrtwo:
I have a couple of things to fix on this before merge. will do on fliht tomorrow.

PR #93, opened 2018-10-20
Fix `shard_and_committte_for_slots` update process

rawfalafel:
If I understand correctly, `shard_and_committee_for_slots` only changes during a validator set change.


djrtwo:
The first half of the array always represents the previous shuffling. The second half represents the current shuffling.

So we only update the _current shuffling_ when there was a change. We always update the prev shuffling to reflect what the current shuffling was.

Assume the following happens:
- at state transition 0, we get a new shuffling (normal case)
  - assign the first half as the second half to remember the prev shuffling
  - update the second half
- at state transition 1, we don't get a new shuffling
  - the shuffling in the first half needs to always represent the previous shuffling, so we need to update it to the reflect that
  - the shuffling is going to stay the same in the second half
- at state transition 2, we don't get a new shulffling
  - the shuffling of the first half doesn't actually need to change, and _won't_ change, but we do the assignment of the second half to the first half for simplicity.
  - the shuffling is going to stay the same in the second half
- at state transition 3, we get a new shuffling
  - assign the first half as the second half to remember the prev shuffling
  - update the second half

PR #98, opened 2018-10-24
update initial validators to status ACTIVE

### LINKED ISSUE #97: Should all initial validators be set to ACTIVE?
Currently, all initial validators are set to PENDING_ACTIVE during `on_startup()`, this means the 16384 validators will get slowly trickle in instead of all at once. Not sure if this is correctly intended.
Given `MIN_VALIDATOR_SET_CHANGE_INTERVAL` is set to ~3%, best case scenario is all validators become ACTIVE after 36 hours

terenc3t:
This fixes #97

### DIFF
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -503,7 +503,8 @@ def on_startup(initial_validator_entries: List[Any]) -> Tuple[CrystallizedState,
             withdrawal_shard=withdrawal_shard,
             withdrawal_address=withdrawal_address,
             randao_commitment=randao_commitment,
-            current_slot=0
+            current_slot=0,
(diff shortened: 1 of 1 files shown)


specs/beacon-chain.md
```diff
@@ -571,6 +572,7 @@ def add_validator(validators: List[ValidatorRecord],
                   withdrawal_shard: int,
                   withdrawal_address: Address,
                   randao_commitment: Hash32,
+                  status: uint8,
```
hwwhww: In Python, we didn't define `uint8`, `uint16`... .etc for type hints. At runtime, it's just `int`, and the types of fields will be checked during data serialization.

PR #104, opened 2018-10-27
Mention that lists are of elements of homogeneous type

mratsim:
This was implicit in the description but better say it explicitly.


djrtwo:
lgtm!

PR #107, opened 2018-10-31
Chain initialization and main chain block inclusion

vbuterin:
Still TODO: deposit processing


djrtwo:
It this ready for review @vbuterin?

I can handle merge conflict

specs/beacon-chain.md
```diff
@@ -209,6 +211,12 @@ The `CrystallizedState` has the following fields:
     'deposits_penalized_in_period': ['uint32'],
     # Hash chain of validator set changes (for light clients to easily track deltas)
     'validator_set_delta_hash_chain': 'hash32'
+    # Genesis time
+    'genesis_time': 'hash32',
+    # PoW chain reference
+    'known_pow_receipt_root': 'hash32',
+    'candidate_pow_receipt_root': 'hash32',
+    'candidate_pow_receipt_root_votes': 'uint64',
```
djrtwo: This can be a smaller int type as the value should never exceed `POW_HASH_VOTING_PERIOD`

specs/beacon-chain.md
```diff
@@ -729,6 +790,7 @@ For each `SpecialRecord` `obj` in `active_state.pending_specials`:
...
 * **[covers RANDAO updates]**: If `obj.kind == RANDAO_REVEAL`, interpret `data[0]` as an integer and `data[1]` as a hash32. Set `validators[data[0]].randao_commitment = data[1]`.
+* **[covers deposits]**: If `obj.kind == DEPOSIT_PROOF`, interpret `data[0]` as being a Merkle branch from a deposit object to the `known_pow_receipt_root` (hashes in order lower to higher, concatenated), `data[1]` as being the index in the tree, and `data[2]` as being the deposit object itself, having the format `concat(data, as_bytes32(msg.value), as_bytes32(block.timestamp)` where `data` itself is a `DepositParams` object. Run `add_validator(validators, data.pubkey, data.proof_of_possession, data.withdrawal_shard, data.withdrawal_address, data.randao_commitment, PENDING_ACTIVATION, block.slot)`.
```
djrtwo: I think we need a pubkey uniqueness assert in `add_validator` to prevent double registration.

djrtwo: You removed the concept of deposit window and replaced it with the deletion period, right?

mkalinin: I think it would, also, be useful to have such an assert in the contract. It could preserve one's ETH from being locked as a result of several accidental registrations with the same pubkey.

vbuterin: There is a deposit window (it's inside of the code in `add_validator`) that's as long as the deletion period.

specs/beacon-chain.md
```diff
@@ -97,8 +99,8 @@ A `BeaconBlock` has the following fields:
     'slot': 'uint64',
     # Proposer RANDAO reveal
     'randao_reveal': 'hash32',
-    # Recent PoW chain reference (block hash)
-    'pow_chain_reference': 'hash32',
+    # Recent PoW chain reference (receipt root)
```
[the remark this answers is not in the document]
vbuterin: I suppose if we want to do that we could make the data objects themselves provide the block prevhash. Less protocol-level complexity :smile:

specs/beacon-chain.md
vbuterin: > I think we need a pubkey uniqueness assert in add_validator to prevent double registration.

Hmm... I remember adding this line but somehow it's not in the code. Maybe I just forgot to write it down? In any case I'll add it.

djrtwo: I'm not seeing the deposit window in `add_validator`. Maybe it disappeared along with the key uniqueness assert?

djrtwo:
@vbuterin Because we don't have a "genesis block" but instead just time at which people can start fulfilling their proposer duties:
* We need to have an allowance in block validity for a block without a parent
* We need to remove the requirement of parent proposer attestation inclusion in blocks without a parent

specs/core/0_beacon-chain.md
```diff
@@ -741,6 +808,24 @@ For each `SpecialRecord` `obj` in `state.pending_specials`:
...
 * **[covers `RANDAO_CHANGE`]**: If `obj.kind == RANDAO_CHANGE`, interpret `data[0]` as a `uint24`, `data[1]` as a `hash32`, and `data[2]` as a `uint64`,  where `proposer_index = data[0]`, `randao_commitment = data[1]`, and `randao_last_change = data[2]`. Set `validators[proposer_index].randao_commitment = randao_commitment`, and set `validators[proposer_index].randao_last_change = randao_last_change`.
+* **[covers `DEPOSIT_PROOF`]**: If `obj.kind == DEPOSIT_PROOF`, interpret `data[0]` as a Merkle branch from a deposit object to the `known_pow_receipt_root` (hashes in order lower to higher, concatenated), `data[1]` as being the `int128` index in the tree, and `data[2]` as being the deposit object itself, having the format `concat(data, as_bytes32(msg.value), as_bytes32(timestamp)` where `data` itself is a `DepositParams`. Verify that `msg.value == DEPOSIT_SIZE` and `block.slot - (timestamp - state.genesis_time) // SLOT_DURATION < DELETION_PERIOD`. Run `add_validator(validators, data.pubkey, data.proof_of_possession, data.withdrawal_shard, data.withdrawal_address, data.randao_commitment, PENDING_ACTIVATION, block.slot)`.
```
djrtwo: Need to state that the merkle branch needs to be verified

djrtwo: we've avoided bigint requirements in the spec. Can this be reformulated without int128 or is it required?

vbuterin: Done.

Also resolved this point by just adding an explicit genesis block:

> @vbuterin Because we don't have a "genesis block" but instead just time at which people can start fulfilling their proposer duties:

vbuterin: Once #138 gets included, we can rework the description further to be more clear, like those for `LOGOUT` and `CASPER_SLASHING` in the PR.

vbuterin: Done.

specs/core/0_beacon-chain.md
```diff
@@ -562,7 +562,21 @@ If the user wishes to deposit more than 32 ETH, they would need to make multiple
 
 ### On startup
 
-Run the following code:
+A valid block with slot `0` (the "genesis block") has the following values. Other validity rules (eg. requiring a signature) do not apply.
```
djrtwo: I think this is still a bit underspecified.
* The child of this block needs to not require the proposer signature
* Do/can committees at slot 0 attest to this block?

Need to make it clear at what slot (0 or 1) do the committee/proposer duties kick in

specs/core/0_beacon-chain.md
```diff
@@ -503,12 +512,74 @@ def int_sqrt(n: int) -> int:
...
+ChainStart: event({hash_chain_tip: bytes32, time: timestamp})
+
+receipt_tree: bytes32[int128]
+total_deposit_count: int128
+
+@payable
+@public
+def deposit(data: bytes[2048]):
+    log.HashChainValue(self.receipt_tree[1], data, msg.value, block.timestamp, self.total_deposit_count)
+    index:int128 = self.total_deposit_count + 2**32
```
djrtwo: We changed the index to `int64` in the Specials but the index can still exceed `int64` here

vbuterin: It can't, because that would require 2**64 - 2**32 deposits to happen.

specs/core/0_beacon-chain.md
vbuterin: I guess the resolution of this depends on whether or not we separate out the proposer to allow a minimum aggregation delay as we had discussed :)

PR #111, opened 2018-11-05
Changed crosslink mechanics

vbuterin:
Makes the following changes:

* Multiple crosslinks for one shard can happen between validator set transitions.
* Validator reshuffling now only happens at all in three cases: (i) during a validator set transition, (ii) during the 4 cycles after a validator set transition, (iii) after an exact power of 2 of cycles after a validator set transition.

This ensures that (i) no committee that is selected has an infinite amount of time to coordinate on an attack, as eventually the next power of 2 is reached, and (ii) every committee must process the shard chain at most twice as fast as the shard chain's progress, as if it is selected at time T + 2^k then it has the 2^k-sized backlog to process from [T...T+2^k] as well as the new data from [T+2^k....T+2^(k+1)] before its deadline at time T + 2^(k+1).

### DIFF
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -210,6 +210,8 @@ The `CrystallizedState` has the following fields:
     # Persistent shard committees
     'persistent_committees': [['uint24']],
     'persistent_committee_reassignments': [ShardReassignmentRecord],
+    # Randao seed used for next shuffling
+    'next_shuffling_seed': 'hash32',
     # Total deposits penalized in the given withdrawal period
     'deposits_penalized_in_period': ['uint32'],
     # Hash chain of validator set changes (for light clients to easily track deltas)
@@ -541,6 +543,7 @@ def on_startup(initial_validator_entries: List[Any]) -> Tuple[CrystallizedState,
         persistent_shuffling=split(shuffle(validators, bytes([0] * 32)), SHARD_COUNT),
         shard_reassignment_records=[],
         deposits_penalized_in_period=[],
+        next_shuffling_seed=b'\x00'*32,
         validator_set_delta_hash_chain=bytes([0] * 32),  # stub
         pre_fork_version=INITIAL_FORK_VERSION,
         post_fork_version=INITIAL_FORK_VERSION,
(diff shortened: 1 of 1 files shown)


djrtwo:
handled merge conflict.

The repeated reshuffling after a validator set change utilizing the `active_state.randao_mix` gives very little look ahead for the first few slots at each cycle. I think the likelihood of a successful crosslink for these first slots in each cycle is very low.

djrtwo:
The following condition for a validator set change needs to be either removed or reworked -- "For every shard number `shard` in `shard_and_committee_for_slots`, `crosslinks[shard].slot > crystallized_state.validator_set_change_slot`"

There are now a minimum of 4 different shufflings between validator set changes. We need to either have validator set change conditional upon all (or some fraction) of the possible crosslinks, or just not based upon crosslinks at all.

vbuterin:
> The repeated reshuffling after a validator set change utilizing the active_state.randao_mix gives very little look ahead for the first few slots at each cycle. I think the likelihood of a successful crosslink for these first slots in each cycle is very low.

Good point! It seems to me the simplest way to fix this is to store the seed that will be used for shuffling N+1 when you're calculating shuffling N?

> There are now a minimum of 4 different shufflings between validator set changes. We need to either have validator set change conditional upon all (or some fraction) of the possible crosslinks, or just not based upon crosslinks at all.

We do need to make sure all crosslinks  get updated IMO. Otherwise, the time between reshufflings resets when there's still one or more "dangling" shard chains with many blocks that have not been verified...

So seems best to reshuffle the validator slots but not increment the shard ID.

djrtwo:
I think the N vs N+1 seed is a reasonable approach for now.

Agreed on not updating the crosslink shard range. I think that gives us closer to what we want here.

vbuterin:
Done.

djrtwo:
lgtm

overchuk-coder:
Hello , please text me gmail overchuk.www@gmail.com

overchuk-coder (commented):
Hello please text me gmail. overchuk.www@gmail.com

PR #113, opened 2018-11-06
typos from persistent shard committees

terenc3t:
I think the validator index should be uint24, and some of the on_start variables are named wrong


djrtwo:
yup! thanks :)

Issue #50, opened 2018-10-05
Extend SSZ hash type


mkalinin:
### Proposal
~Replace SSZ hash type family with the type that has arbitrary length, something like `hash[32..128]`.~
**EDIT:**
Consider any `hashN` type, where `N` is positive integer, a valid `hash` type if `N` satisfies condition:
 `N % 8 == 0`

### Rationale
Using strict number of hash types has no forward compatibility. If there will be a need in a new e.g. `hash128` type then implementers will have to update their code that runs `hash` type validation or use `bytes` type as a workaround.

vbuterin:
So a dynamic length type? That seems like a bad idea; introducing more dynamic length objects inherently makes it harder to serialize, deserialize, and poke into an object.

mkalinin:
> So a dynamic length type?

I don't think so. It just proposes to expand a family of hash types that we have in SSZ spec so far to `[hash32, hash33, ..., hashN]`. Whatever `N` would be reasonable to use. Each of those types is a static length type.

Currently, spec describes only three `hash32/96/97` and validation is strictly tight to those three. Btw, `hash97` looks odd.

mkalinin:
Hmm, maybe it would be just enough to make a list from all reasonable hash lengths and add them to the spec. what do you think?

djrtwo:
I think it pretty reasonable to define `hashN` where `N % 8 == 0` for the time being and to address an exceptional case like `hash97` if we and when we do in fact need it.

mkalinin:
> I think it pretty reasonable to define `hashN` where `N % 8 == 0`

Looks like the best solution to me. I am pretty confident that for `hash` type we won't need anything else than `N % 8 == 0`

mratsim:
I always thought the current hashes were just examples cc @NatoliChris and it was indeed arbitrary fixed length;

mkalinin:
I have updated original proposal with what we finally came up to. It should read better now.

djrtwo:
checking back on this. Do we need to support  arbitrary byte length (like 97) or just where `N % 8 == 0`?

PR #116, opened 2018-11-06
extend ssz hash type to arbitrary lengths

djrtwo:
addresses #50 
Make the hash type in SSZ allow for arbitrary byte length, `N`, for `hashN`

I left the examples of hash32, hash96, and hash97. I can pull them out if we think it's overkill

### DIFF
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -117,7 +117,7 @@ return value
 |:---------------------------------------|:---------------------|
 | Length in bytes is correct for `hashN` | ``len(value) == N``  |
 
-##### Example for arbitrary hashN
+##### HashN
 
 ```python
 assert(len(value) == N)
@@ -297,7 +297,6 @@ return rawbytes[current_index:current_index+20], new_index
 ```
 
 #### Hash
-
 
 ##### HashN
 
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -298,6 +298,17 @@ return rawbytes[current_index:current_index+20], new_index
 
 #### Hash
 
+
+##### HashN
+
+Return the N bytes.
+
+```python
+assert(len(rawbytes) >= current_index + N)
+new_index = current_index + N
+return rawbytes[current_index:current_index+N], new_index
+```
+
 ##### Hash32
 
 Return the 32 bytes.
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -16,9 +16,6 @@ deserializing objects and data types.
       - [uint: 8/16/24/32/64/256](#uint-816243264256)
       - [Address](#address)
       - [Hash](#hash)
-         * [Hash32](#hash32)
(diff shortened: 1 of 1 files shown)


specs/simple-serialize.md
```diff
@@ -116,26 +110,17 @@ return value
 |  `hash32` | Hash size of ``keccak`` or `blake2b[0.. < 32]`. |
 |  `hash96` | BLS Public Key Size.                            |
 |  `hash97` | BLS Public Key Size with recovery bit.          |
+|  `hashN`  | Hash of arbitrary byte length, N                |
```
hwwhww: It's a little confusing if we keep `hash32`, `hash96`, `hash97`, and `hashN` in the table when they can all be replaced with `hashN`. I think we can just remove the description of `hash32`, `hash96`, and `hash97`.

hwwhww: nitpick: add a period in the end and markdown code mark:

Hash of arbitrary byte length `N`.

specs/simple-serialize.md
```diff
@@ -313,6 +298,16 @@ return rawbytes[current_index:current_index+20], new_index
...
+
+Return the N bytes.
+
+```python
+assert(len(rawbytes) >= current_index + N)
+new_index = current_index + N
+return rawbytes[current_index:current_index+N], new_index
+```
+
 ##### Hash32
```
hwwhww: `##### Hash32`, `##### Hash96`, and `##### Hash97` sections could be removed.

specs/simple-serialize.md
```diff
@@ -116,26 +110,17 @@ return value
...
-| Length is correct (32) if `hash32` | ``len(value) == 32`` |
-| Length is correct (96) if `hash96` | ``len(value) == 96`` |
-| Length is correct (97) if `hash97` | ``len(value) == 97`` |
+| Checks to perform                      | Code                 |
+|:---------------------------------------|:---------------------|
+| Length in bytes is correct for `hashN` | ``len(value) == N``  |
 
-
-**Example all together**
+##### HashN
```
hwwhww: `##### Hash32`, `##### Hash96`, and `##### Hash97` sections could be removed.

specs/simple-serialize.md
```diff
@@ -313,6 +298,16 @@ return rawbytes[current_index:current_index+20], new_index
 
 #### Hash
 
+##### HashN
+
+Return the N bytes.
```
hwwhww: nitpick: Return the `N` bytes.

specs/simple-serialize.md
hwwhww: Since `HashN` is a custom type like `Address`, IMO we can use CapWords notation `HashN`  in this spec.

djrtwo: In the implementation, it is assumed that everything is lowercase. I think we should be clearer and more consistent across the board. Maybe for another pr

PR #119, opened 2018-11-08
Remove hash97

vbuterin:
Because the prime modulus is less than 2**381, we can just use the top bit of the x coordinate for the parity bit. This is what we do for the current alt_bn128 implementation already. So we don't need 97 bytes for anything, 96 works just fine.


djrtwo:
Already handled in #116

PR #120, opened 2018-11-08
Added tree hashing algorithm

vbuterin:
Adds to the SSZ spec an algorithm for tree-hashing SSZ objects.

Experiments suggest that the hashing time required is increased by a factor of ~4-6, partly because Merkle hashing inherently raises the amount of data hashes due to multiple layers of hashing and partially because hash functions have a fixed computation cost in addition to the per-byte cost.

### DIFF
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -449,6 +449,77 @@ assert item_index == start + LENGTH_BYTES + length
 return typ(**values), item_index
 ```
 
+### Tree hash
+
+The below algorithm is defined recursively in the case of lists and containers, and it outputs a value equal to or less than 32 bytes in size. For the final output only (ie. not intermediate outputs), if the output is less than 32 bytes, right-zero-pad it to 32 bytes. The goal is collision resistance *within* each type, not between types.
+
+We define `hash(x)` as `BLAKE2b-512(x)[0:32]`.
+
+#### uint: 8/16/24/32/64/256, bool, address, hash32
+
+Return the serialization of the value.
+
+#### bytes, hash96
+
+Return the hash of the serialization of the value.
+
+#### List/Vectors
+
+First, we define some helpers and then the Merkle tree function. The constant `CHUNK_SIZE` is set to 128.
+
+```python
+# Returns the smallest power of 2 equal to or higher than x
+def next_power_of_2(x):
+    return x if x == 1 else next_power_of_2((x+1) // 2) * 2
+
+# Extends data length to a power of 2 by minimally right-zero-padding
+def extend_to_power_of_2(data):
+    return data + b'\x00' * (next_power_of_2(len(data)) - len(data))
+
+# Concatenate a list of homogeneous objects into data and pad it
+def list_to_glob(lst):
+    if len(lst[0]) != next_power_of_2(len(lst[0])):
+        lst = [extend_to_power_of_2(x) for x in lst]
+    data = b''.join(lst)
+    # Pad to chunksize
+    data += b'\x00' * (CHUNKSIZE - (len(data) % CHUNKSIZE or CHUNKSIZE))
+    return data
+
+# Merkle tree hash of a list of items
+def merkle_hash(lst):
+    # Turn list into padded data
+    data = list_to_glob(lst)
+    # Store length of list (to compensate for non-bijectiveness of padding)
+    datalen = len(lst).to_bytes(32, 'big')
+    # Convert to chunks
+    chunkz = [data[i:i+CHUNKSIZE] for i in range(0, len(data), CHUNKSIZE)]
+    # Tree-hash
+    while len(chunkz) > 1:
+        if len(chunkz) % 2 == 1:
+            chunkz.append(b'\x00' * CHUNKSIZE)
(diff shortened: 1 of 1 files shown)


specs/simple-serialize.md
```diff
@@ -449,6 +449,77 @@ assert item_index == start + LENGTH_BYTES + length
...
+def next_power_of_2(x):
+    return x if x == 1 else next_power_of_2((x+1) // 2) * 2
+
+# Extends data length to a power of 2 by minimally right-zero-padding
+def extend_to_power_of_2(data):
+    return data + b'\x00' * (next_power_of_2(len(data)) - len(data))
+
+# Concatenate a list of homogeneous objects into data and pad it
+def list_to_glob(lst):
+    if len(lst[0]) != next_power_of_2(len(lst[0])):
```
djrtwo: Why are we just checking the 0th item? I don't think there is any guarantee that the data length of these objects are the same considering they can have variable length lists inside of them.

specs/simple-serialize.md
```diff
@@ -449,6 +449,77 @@ assert item_index == start + LENGTH_BYTES + length
 return typ(**values), item_index
 ```
 
+### Tree hash
+
```
djrtwo: Need to be clearer about defining `tree_hash` as a function separate from `merkle_hash`. I think just stating it as `tree_hash` this header will help.

specs/simple-serialize.md
djrtwo: Oh is it because the items have already been `tree_hash`ed to get to this point?

arnetheduck: Why do `extend_to_power_of_2` at all?

specs/simple-serialize.md
```diff
@@ -449,6 +449,77 @@ assert item_index == start + LENGTH_BYTES + length
...
+Where `tree_hash` is a recursive application of the tree-hashing function (returning less than 32 bytes for short single values).
+
+
+#### Container
+
+Recursively tree hash the values in the container in order sorted by key, and return the hash of the concatenation of the results.
+
+```python
+return hash(b''.join([tree_hash(getattr(x, field)) for field in sorted(value.fields)))
```
mkalinin: Recalling @sorpaas's comment https://github.com/ethereum/eth2.0-specs/issues/54#issuecomment-435348622.
What if we just have:
```python 
return hash(SSZ(obj))
``` 
It would save us some hashing rounds and would, also, decrease spec and implementation complexity.

mkalinin:
It would be nice to have an example of `tree_hash` function implementation at the end of `Tree hash` section.

specs/simple-serialize.md
vbuterin: > In my understanding, it guarantees that despite the object's length, chunks will fit boundaries of extended objects. A list of address objects is a good example here.

Yes, exactly this.

> Oh is it because the items have already been tree_hashed to get to this point?

Yep! If the subtype is anything variable-sized, then the tree hash will have 32 bytes.

specs/simple-serialize.md
vbuterin: Because the object's sub-values could have a huge size, and you don't want to re-hash all of that every time you update it. This is particularly true for the beacon chain state.

specs/simple-serialize.md
vbuterin: Done!

vbuterin:
> It would be nice to have an example of tree_hash function implementation at the end of Tree hash section.

@mkalinin you can find this here: https://github.com/ethereum/beacon_chain/blob/vitalik_hash_ssz/hash_ssz.py

I imagine @djrtwo will copy the code or something like it into the python impl once this gets finalized/approved.

specs/simple-serialize.md
```diff
@@ -449,6 +449,77 @@ assert item_index == start + LENGTH_BYTES + length
...
+    datalen = len(lst).to_bytes(32, 'big')
+    # Convert to chunks
+    chunkz = [data[i:i+CHUNKSIZE] for i in range(0, len(data), CHUNKSIZE)]
+    # Tree-hash
+    while len(chunkz) > 1:
+        if len(chunkz) % 2 == 1:
+            chunkz.append(b'\x00' * CHUNKSIZE)
+        chunkz = [hash(chunkz[i] + chunkz[i+1]) for i in range(0, len(chunkz), 2)]
+    # Return hash of root and length data
+    return hash(chunkz[0] + datalen)
```
arnetheduck: nitpick: `lst` / `chunkz` can be empty in the general case

specs/simple-serialize.md
arnetheduck: > chunks will fit boundaries of extended objects

alternative that packs 6 instead of 4 addresses (assuming len < CHUNKSIZE as simplicifcation), by padding only once per chunk:

```python
items_per_chunk = CHUNKSIZE // len(l[0])
padding = b'\00' * (CHUNKSIZE - items_per_chunk * len(l[0]))
chunkz = [b''.join(l[i:i+items_per_chunk])+padding for i in range(0, len(l), items_per_chunk)]
```

PR #122, opened 2018-11-08
Minimal replacement of active/crystallized state with single state

vbuterin:
Depends on https://github.com/ethereum/eth2.0-specs/pull/120.

Does not include deeper simplifications that become possible with this approach; this is a fairly simple search and replace.

### DIFF
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -546,7 +546,7 @@ def on_startup(initial_validator_entries: List[Any]) -> State:
     return state
 ```
 
-The `State()` constructor should initialize all values to zero bytes, an empty value or an empty array depending on context. The `add_validator` routine is defined below.
+The `add_validator` routine is defined below.
 
 ### Routine for adding a validator
 
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -178,7 +178,7 @@ The `BeaconState` has the following fields:
     'validators': [ValidatorRecord],
     # Most recent crosslink for each shard
     'crosslinks': [CrosslinkRecord],
-    # Last crystallized state recalculation
+    # Last cycle-boundary state recalculation
     'last_state_recalculation_slot': 'uint64',
     # Last finalized slot
     'last_finalized_slot': 'uint64',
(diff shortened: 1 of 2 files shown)


prestonvanloon:
Context? Why make this change?

vbuterin:
It's a protocol simplification (two states down to one) that makes way for other protocol simplifications (adjusting RANDAO instantly instead of at the end of the cycle, among other things), which was not viable before because of re-hashing costs, but is viable now, because of the SSZ tree hashing proposal: https://github.com/ethereum/eth2.0-specs/pull/120

specs/beacon-chain.md
```diff
@@ -668,22 +649,22 @@ For each one of these attestations:
...
 
 Additionally, verify and update the RANDAO reveal. This is done as follows:
 
 * Let `repeat_hash(x, n) = x if n == 0 else repeat_hash(hash(x), n-1)`.
-* Let `V = crystallized_state.validators[curblock_proposer_index]`.
-* Verify that `repeat_hash(block.randao_reveal, (block.slot - V.randao_last_reveal) // RANDAO_SLOTS_PER_LAYER + 1) == V.randao_commitment`, and set `active_state.randao_mix = xor(active_state.randao_mix, block.randao_reveal)` and append to `ActiveState.pending_specials` a `SpecialObject(kind=RANDAO_CHANGE, data=[bytes8(curblock_proposer_index), block.randao_reveal])`.
+* Let `V = state.validators[curblock_proposer_index]`.
+* Verify that `repeat_hash(block.randao_reveal, (block.slot - V.randao_last_reveal) // RANDAO_SLOTS_PER_LAYER + 1) == V.randao_commitment`, and set `state.randao_mix = xor(state.randao_mix, block.randao_reveal)` and append to `State.pending_specials` a `SpecialObject(kind=RANDAO_CHANGE, data=[bytes8(curblock_proposer_index), block.randao_reveal])`.
```
djrtwo: Should we consider removing the `RANDAO_CHANGE` special record and just allow the one validator record to be updated?

This kind of optimization can occur in a separate PR

djrtwo: Ah, I just saw your comment above. Deeper changes to follow.

specs/beacon-chain.md
```diff
@@ -170,24 +168,7 @@ A `SpecialRecord` has the following fields:
...
-    'pending_specials': [SpecialRecord]
-    # recent beacon block hashes needed to process attestations, older to newer
-    'recent_block_hashes': ['hash32'],
-    # RANDAO state
-    'randao_mix': 'hash32'
-}
-```
-
-The `CrystallizedState` has the following fields:
+The `state` has the following fields:
```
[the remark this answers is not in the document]
vbuterin: Done.

PR #123, opened 2018-11-09
Add shard blocks, shard data roots and how data is computed into cros…

vbuterin:
…slinks

Includes:

* Shard block structure
* Shard block header verification rule
* Shard block fork choice rule
* Shard block body verification rule
* Crosslink verification rule

Possible simplification: require `calc_block_maxbytes` to always output an exact power of two; if we desire the average maxbytes to be smooth, we can simply make it a pseudorandom chose between powers. This reduces some of the padding complexity.


specs/beacon-chain.md
```diff
@@ -870,6 +905,81 @@ while len(crystallized_state.persistent_committee_reassignments) > 0 and crystal
...
+## Shard block processing
+
+A block on a shard can be processed only if its `parent` has been accepted. To validate a block header on shard `shard_id`, compute as follows:
+
+* Verify that `beacon_chain_ref` is the hash of the `slot`'th block in the beacon chain.
+* Let `state` be the state of the beacon chain block referred to by `beacon_chain_ref`. Let `validators` be `[validators[i] for i in state.current_persistent_committees[shard_id]]`.
+* Assert `len(attester_bitfield) == ceil_div8(len(validators))`
+* Let `curblock_proposer_index = hash(state.randao_mix + bytes8(shard_id)) % len(validators)`. Let `parent_proposer_index` be the same value calculated for the parent block.
+* Make sure that the `parent_proposer_index`'th bit in the `attester_bitfield` is set to 1.
+* Generate the group public key by adding the public keys of all the validators for whom the corresponding position in the bitfield is set to 1. Verify the `aggregate_sig` using this as the pubkey and the `parent_hash` as the message.
```
djrtwo: So the only opportunity for shard block signatures to be included is in the immediate following block?

djrtwo: but fork choice is still based upon signatures seen on the wire rather than just those included in the block?

specs/beacon-chain.md
```diff
@@ -870,6 +905,81 @@ while len(crystallized_state.persistent_committee_reassignments) > 0 and crystal
...
+
+Note that at network layer we expect blocks to be broadcasted along with the signature from the `curblock_proposer_index`'th validator in the validator set for that block.
```
djrtwo: Move this to the top of the section and have it as one of the pre-processing conditions (alongside parent being accepted)

specs/beacon-chain.md
```diff
@@ -870,6 +905,81 @@ while len(crystallized_state.persistent_committee_reassignments) > 0 and crystal
...
+
+### Verifying shard block data
+
+At network layer, we expect a shard block header to be broadcasted along with its `block_body`. First, we define a helper function that takes as input beacon chain state and outputs the max block size in bytes:
```
djrtwo:
```suggestion
At network layer, we expect a shard block header to be broadcast along with its `block_body`. First, we define a helper function that takes as input beacon chain state and outputs the max block size in bytes:
```
"broadcast" is the much more common past tense

specs/beacon-chain.md
```diff
@@ -870,6 +905,81 @@ while len(crystallized_state.persistent_committee_reassignments) > 0 and crystal
...
+
+```python
+def calc_block_maxbytes(state):
+    max_grains = MAX_SHARD_BLOCK_SIZE // CHUNK_SIZE
+    validators_at_max_committees = SHARD_COUNT * TARGET_COMMITTEE_SIZE
```
djrtwo: variable name is misleading. Maybe `validators_at_target_committee_size`

specs/beacon-chain.md
```diff
@@ -870,6 +905,81 @@ while len(crystallized_state.persistent_committee_reassignments) > 0 and crystal
...
+    grains = min(len(get_active_validator_indices(state.validators)) * max_grains // validators_at_max_committees,
```
djrtwo: This could use a clarifying comment. Something like the following
```
# number of grains per block is proportional to the number of validators
# up until `validators_at_target_committee_size`
```

specs/beacon-chain.md
```diff
@@ -870,6 +905,81 @@ while len(crystallized_state.persistent_committee_reassignments) > 0 and crystal
...
+* Generate the group public key by adding the public keys of all the validators for whom the corresponding position in the bitfield is set to 1. Verify the `aggregate_sig` using this as the pubkey and the `parent_hash` as the message.
+
+Note that at network layer we expect blocks to be broadcasted along with the signature from the `curblock_proposer_index`'th validator in the validator set for that block.
+
+### Verifying shard block data
+
+At network layer, we expect a shard block header to be broadcasted along with its `block_body`. First, we define a helper function that takes as input beacon chain state and outputs the max block size in bytes:
+
+```python
+def calc_block_maxbytes(state):
```
djrtwo:
```suggestion
def shard_block_maxbytes(state):
```

rename to be explicit about what type of block we are talking about. Remove `calc` from name as it is obvious that the function does a calculation.

specs/beacon-chain.md
```diff
@@ -870,6 +905,81 @@ while len(crystallized_state.persistent_committee_reassignments) > 0 and crystal
...
+
+First, the conditions must recursively apply to the crosslink referenced in `last_crosslink_hash` for the same shard (unless `last_crosslink_hash` equals zero, in which case we are at the genesis).
+
+Second, we verify the `shard_block_combined_data_root`. Let `B[0]` be the first block _after_ the last crosslink and `B[n-1]` be the block directly referenced by the `shard_block_hash`. Let `bodies[0] .... bodies[n-1]` be their bodies and `roots[0] ... roots[n-1]` the data roots. Define `compute_merkle_root` be a simple Merkle root calculating function that takes as input a list of objects, where the list's length must be an exact power of two. Let `state[0] ... state[n-1]` be the beacon chain states at those times, and `depths[0] ... depths[n-1]` be equal to `log2(next_power_of_2(calc_block_maxbytes(state[i]) // CHUNK_SIZE))` (ie. the expected depth of the i'th data tree).
```
djrtwo: Need to specify what to do with missing slots

specs/beacon-chain.md
```diff
@@ -870,6 +905,81 @@ while len(crystallized_state.persistent_committee_reassignments) > 0 and crystal
...
+    data = [default_value for _ in range(next_power_of_2(len(roots)))]
+    for i, (depth, root) in enumerate(zip(depths, roots)):
+        value = root
+        for j in range(depth, max(depths)):
+            value = hash(value, get_zeroroot_at_depth(depth + j))
+        data[i] = value
+    return compute_merkle_root(data)
+```
+
+Essentially, this outputs the root of a tree of the data roots, with the data roots all adjusted to have the same height if needed. The tree can also be viewed as a tree of all of the underlying data concatenated together, appropriately padded. Here is an equivalent definition that uses bodies instead of roots [TODO: check equivalence]:
```
djrtwo:
```suggestion
This outputs the root of a tree of the data roots, with the data roots all adjusted to have the same height if needed. The tree can also be viewed as a tree of all of the underlying data concatenated together, appropriately padded. Here is an equivalent definition that uses bodies instead of roots [TODO: check equivalence]:
```

No need for "essentially"

specs/beacon-chain.md
```diff
@@ -870,6 +905,81 @@ while len(crystallized_state.persistent_committee_reassignments) > 0 and crystal
...
+                 max_grains)
+    return CHUNK_SIZE * grains
+```
+
+* Verify that `len(block_body) == calc_block_maxbytes(state)`
+* Define `filler_bytes = next_power_of_2(len(block_body)) - len(block_body)`. Compute a simple binary Merkle tree of `block_body + bytes([0] * filler_bytes)` and verify that the root equals the `data_root` in the header.
+
+### Verifying a crosslink
+
+A node should sign a crosslink only if the following conditions hold. **If a node has the capability to perform the required level of verification, it should NOT follow chains on which a crosslink for which these conditions do NOT hold has been included, or a sufficient number of signatures have been included that during the next state recalculation, a crosslink will be registered.**
```
djrtwo: Is this bolded test suggesting that if a bad crosslink (wrt data) is made into the beacon chain (or is to be made), that the local nodes that are fully verifying a shard should ignore the bad crosslink?

specs/beacon-chain.md
```diff
@@ -870,6 +905,81 @@ while len(crystallized_state.persistent_committee_reassignments) > 0 and crystal
...
+    validators_at_max_committees = SHARD_COUNT * TARGET_COMMITTEE_SIZE
+    grains = min(len(get_active_validator_indices(state.validators)) * max_grains // validators_at_max_committees,
+                 max_grains)
+    return CHUNK_SIZE * grains
+```
+
+* Verify that `len(block_body) == calc_block_maxbytes(state)`
+* Define `filler_bytes = next_power_of_2(len(block_body)) - len(block_body)`. Compute a simple binary Merkle tree of `block_body + bytes([0] * filler_bytes)` and verify that the root equals the `data_root` in the header.
+
+### Verifying a crosslink
```
djrtwo: I'm a bit confused about this section.

Who is doing this crosslink data verification and when should they do it? It seems that this is slightly prescriptive to crosslinkers but also for maybe for people who might contest data proofs and also for nodes to override bad crosslinks?

specs/beacon-chain.md
```diff
@@ -870,6 +905,81 @@ while len(crystallized_state.persistent_committee_reassignments) > 0 and crystal
...
+    padded_body_length = max([CHUNK_SIZE * 2**d for d in depths])
+    data = b''
+    for body in bodies:
+        padded_body = body + bytes([0] * (padded_body_length - len(body)))
+        data += padded_body
+    data += bytes([0] * (next_power_of_2(len(data)) - len(data))
+    return compute_merkle_root([data[pos:pos+CHUNK_SIZE] for pos in range(0, len(data), CHUNK_SIZE)])
+```
+
+Verify that the `shard_block_combined_data_root` is the output of these functions.
```
djrtwo: Should this verification be happening at the beacon chain level? or is this independent?

JustinDrake:
From a pragmatic standpoint I'm inclined to avoid specifying phase 1 (this pull request and the data availability proofs) before phase 0. Reasons include:

1. There is a real possibility that we will be "forced" to launch the beacon chain without a STARK-friendly hash function. But this doesn't preclude the shard blocks using a STARK-friendly hash function required for the data availability proofs.
2. We want to limit the complexity of the phase 0 to launch the beacon chain ASAP. Delaying the specification of phase 1 is a significant simplification to the phase 0 spec.
3. We don't want to rush phase 1. (As a side note, I'm not totally sold on variable sized blocks.)
4. We want phase 0 to be maximally forward compatible to easily incorporate new ideas from between the launches of phase 0 and phase 1.

djrtwo:
What if we move the phase 1 specifications into a separate doc so that teams can easily target phase 0 without getting bogged down in phase 1?

I don't want to halt specifying phase 1, but do see the value in having a clean separation from phase 0. I think it is important that the research team begins to attempt to implement phase 1 ahead of the production implementations so that we can iron out the kinks.

specs/core/1_shard-data-chains.md
```diff
@@ -0,0 +1,122 @@
...
+    'attester_bitfield': 'bytes',
+    'aggregate_sig': ['uint256'],
+}
+```
+
+## Shard block processing
+
+A block on a shard can be processed only if its `parent` has been accepted. To validate a block header on shard `shard_id`, compute as follows:
+
+* Verify that `beacon_chain_ref` is the hash of the `slot`'th block in the beacon chain.
```
djrtwo: This needs to be tightened up a bit to account for missing blocks in the beacon chain. It should be the hash of the head of the beacon chain during that slot. This is accounted for in the state.recent_block_hashes but only in a future block after the slot that was skipped.

djrtwo: Or, is the shard block skipped if the beacon block is skipped?

The below curblock_proposer_index scheme seems to imply either it is skipped or it is the same proposer back to back.

specs/core/1_shard-data-chains.md
```diff
@@ -0,0 +1,122 @@
...
+* Let `state` be the state of the beacon chain block referred to by `beacon_chain_ref`. Let `validators` be `[validators[i] for i in state.current_persistent_committees[shard_id]]`.
+* Assert `len(attester_bitfield) == ceil_div8(len(validators))`
+* Let `curblock_proposer_index = hash(state.randao_mix + bytes8(shard_id)) % len(validators)`. Let `parent_proposer_index` be the same value calculated for the parent block.
+* Make sure that the `parent_proposer_index`'th bit in the `attester_bitfield` is set to 1.
+* Generate the group public key by adding the public keys of all the validators for whom the corresponding position in the bitfield is set to 1. Verify the `aggregate_sig` using this as the pubkey and the `parent_hash` as the message.
```
djrtwo: So the only opportunity for shard block signatures to be included is in the immediate following block?

djrtwo: but fork choice is still based upon signatures seen on the wire rather than just those included in the block?

specs/core/1_shard-data-chains.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+## Shard block processing
+
+A block on a shard can be processed only if its `parent` has been accepted. To validate a block header on shard `shard_id`, compute as follows:
+
+* Verify that `beacon_chain_ref` is the hash of the `slot`'th block in the beacon chain.
+* Let `state` be the state of the beacon chain block referred to by `beacon_chain_ref`. Let `validators` be `[validators[i] for i in state.current_persistent_committees[shard_id]]`.
+* Assert `len(attester_bitfield) == ceil_div8(len(validators))`
+* Let `curblock_proposer_index = hash(state.randao_mix + bytes8(shard_id)) % len(validators)`. Let `parent_proposer_index` be the same value calculated for the parent block.
```
djrtwo: A couple of questions:
* a shard block proposer does not know he is the proposer for a given slot until he sees the beacon chain block for that slot. True/False?
  * So the shard block proposer is proposing some fractional time into the slot?
* if there is no beacon block for a given slot, then there is similarly no shard block for that slot. True/False?
  * If this is false and we do expect there to be shard blocks even when the beacon block is skipped, then I suggest that we incorporate the slot into the proposer index as such: `curblock_proposer_index = hash(state.randao_mix + bytes8(shard_id) + slot) % len(validators)`

specs/core/1_shard-data-chains.md
```diff
@@ -0,0 +1,122 @@
...
+* Make sure that the `parent_proposer_index`'th bit in the `attester_bitfield` is set to 1.
+* Generate the group public key by adding the public keys of all the validators for whom the corresponding position in the bitfield is set to 1. Verify the `aggregate_sig` using this as the pubkey and the `parent_hash` as the message.
+
+Note that at network layer we expect blocks to be broadcasted along with the signature from the `curblock_proposer_index`'th validator in the validator set for that block.
```
djrtwo: Move this to the top of the section and have it as one of the pre-processing conditions (alongside parent being accepted)

specs/core/1_shard-data-chains.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Verifying shard block data
+
+At network layer, we expect a shard block header to be broadcasted along with its `block_body`. First, we define a helper function that takes as input beacon chain state and outputs the max block size in bytes:
```
```suggestion
At network layer, we expect a shard block header to be broadcast along with its `block_body`. First, we define a helper function that takes as input beacon chain state and outputs the max block size in bytes:
```
"broadcast" is the much more common past tense

specs/core/1_shard-data-chains.md
```diff
@@ -0,0 +1,122 @@
...
+
+```python
+def calc_block_maxbytes(state):
+    max_grains = MAX_SHARD_BLOCK_SIZE // CHUNK_SIZE
+    validators_at_max_committees = SHARD_COUNT * TARGET_COMMITTEE_SIZE
```
djrtwo: variable name is misleading. Maybe `validators_at_target_committee_size`

specs/core/1_shard-data-chains.md
```diff
@@ -0,0 +1,122 @@
...
+    grains = min(len(get_active_validator_indices(state.validators)) * max_grains // validators_at_max_committees,
```
djrtwo: This could use a clarifying comment. Something like the following
```
# number of grains per block is proportional to the number of validators
# up until `validators_at_target_committee_size`
```

specs/core/1_shard-data-chains.md
```diff
@@ -0,0 +1,122 @@
...
+* Generate the group public key by adding the public keys of all the validators for whom the corresponding position in the bitfield is set to 1. Verify the `aggregate_sig` using this as the pubkey and the `parent_hash` as the message.
+
+Note that at network layer we expect blocks to be broadcasted along with the signature from the `curblock_proposer_index`'th validator in the validator set for that block.
+
+### Verifying shard block data
+
+At network layer, we expect a shard block header to be broadcasted along with its `block_body`. First, we define a helper function that takes as input beacon chain state and outputs the max block size in bytes:
+
+```python
+def calc_block_maxbytes(state):
```
```suggestion
def shard_block_maxbytes(state):
```

rename to be explicit about what type of block we are talking about. Remove `calc` from name as it is obvious that the function does a calculation.

specs/core/1_shard-data-chains.md
```diff
@@ -0,0 +1,122 @@
...
+
+First, the conditions must recursively apply to the crosslink referenced in `last_crosslink_hash` for the same shard (unless `last_crosslink_hash` equals zero, in which case we are at the genesis).
+
+Second, we verify the `shard_block_combined_data_root`. Let `B[0]` be the first block _after_ the last crosslink and `B[n-1]` be the block directly referenced by the `shard_block_hash`. Let `bodies[0] .... bodies[n-1]` be their bodies and `roots[0] ... roots[n-1]` the data roots. Define `compute_merkle_root` be a simple Merkle root calculating function that takes as input a list of objects, where the list's length must be an exact power of two. Let `state[0] ... state[n-1]` be the beacon chain states at those times, and `depths[0] ... depths[n-1]` be equal to `log2(next_power_of_2(calc_block_maxbytes(state[i]) // CHUNK_SIZE))` (ie. the expected depth of the i'th data tree).
```
djrtwo: Need to specify what to do with missing slots

specs/core/1_shard-data-chains.md
```diff
@@ -0,0 +1,122 @@
...
+    data = [default_value for _ in range(next_power_of_2(len(roots)))]
+    for i, (depth, root) in enumerate(zip(depths, roots)):
+        value = root
+        for j in range(depth, max(depths)):
+            value = hash(value, get_zeroroot_at_depth(depth + j))
+        data[i] = value
+    return compute_merkle_root(data)
+```
+
+Essentially, this outputs the root of a tree of the data roots, with the data roots all adjusted to have the same height if needed. The tree can also be viewed as a tree of all of the underlying data concatenated together, appropriately padded. Here is an equivalent definition that uses bodies instead of roots [TODO: check equivalence]:
```
```suggestion
This outputs the root of a tree of the data roots, with the data roots all adjusted to have the same height if needed. The tree can also be viewed as a tree of all of the underlying data concatenated together, appropriately padded. Here is an equivalent definition that uses bodies instead of roots [TODO: check equivalence]:
```

No need for "essentially"

specs/core/1_shard-data-chains.md
```diff
@@ -0,0 +1,122 @@
...
+                 max_grains)
+    return CHUNK_SIZE * grains
+```
+
+* Verify that `len(block_body) == calc_block_maxbytes(state)`
+* Define `filler_bytes = next_power_of_2(len(block_body)) - len(block_body)`. Compute a simple binary Merkle tree of `block_body + bytes([0] * filler_bytes)` and verify that the root equals the `data_root` in the header.
+
+### Verifying a crosslink
+
+A node should sign a crosslink only if the following conditions hold. **If a node has the capability to perform the required level of verification, it should NOT follow chains on which a crosslink for which these conditions do NOT hold has been included, or a sufficient number of signatures have been included that during the next state recalculation, a crosslink will be registered.**
```
djrtwo: Is this bolded test suggesting that if a bad crosslink (wrt data) is made into the beacon chain (or is to be made), that the local nodes that are fully verifying a shard should ignore the bad crosslink?

specs/core/1_shard-data-chains.md
```diff
@@ -0,0 +1,122 @@
...
+    validators_at_max_committees = SHARD_COUNT * TARGET_COMMITTEE_SIZE
+    grains = min(len(get_active_validator_indices(state.validators)) * max_grains // validators_at_max_committees,
+                 max_grains)
+    return CHUNK_SIZE * grains
+```
+
+* Verify that `len(block_body) == calc_block_maxbytes(state)`
+* Define `filler_bytes = next_power_of_2(len(block_body)) - len(block_body)`. Compute a simple binary Merkle tree of `block_body + bytes([0] * filler_bytes)` and verify that the root equals the `data_root` in the header.
+
+### Verifying a crosslink
```
djrtwo: I'm a bit confused about this section.

Who is doing this crosslink data verification and when should they do it? It seems that this is slightly prescriptive to crosslinkers but also for maybe for people who might contest data proofs and also for nodes to override bad crosslinks?

specs/core/1_shard-data-chains.md
```diff
@@ -0,0 +1,122 @@
...
+    padded_body_length = max([CHUNK_SIZE * 2**d for d in depths])
+    data = b''
+    for body in bodies:
+        padded_body = body + bytes([0] * (padded_body_length - len(body)))
+        data += padded_body
+    data += bytes([0] * (next_power_of_2(len(data)) - len(data))
+    return compute_merkle_root([data[pos:pos+CHUNK_SIZE] for pos in range(0, len(data), CHUNK_SIZE)])
+```
+
+Verify that the `shard_block_combined_data_root` is the output of these functions.
```
djrtwo: Should this verification be happening at the beacon chain level? or is this independent?

specs/beacon-chain.md
```diff
@@ -668,12 +702,13 @@ For each one of these attestations:
 * Verify that `slot <= parent.slot` and `slot >= max(parent.slot - CYCLE_LENGTH + 1, 0)`.
 * Verify that `justified_slot` is equal to or earlier than `last_justified_slot`.
 * Verify that `justified_block_hash` is the hash of the block in the current chain at the slot -- `justified_slot`.
+* Verify that `last_crosslink_hash` equals `state.crosslinks[shard].shard_block_hash`.
```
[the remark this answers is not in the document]
vbuterin: Yes, I think that works.

specs/core/1_shard-data-chains.md
vbuterin: > So the only opportunity for shard block signatures to be included is in the immediate following block?

Yes. Given that we don't need them to be included for finalization purposes and we won't quadratic-leak anyone for not signing on shards that seems reasonable to me. We can set up a reward structure so that the reward to the proposer equals the sum of the rewards to the attesters so there is a large incentive to include attestations, if we wish.

> but fork choice is still based upon signatures seen on the wire rather than just those included in the block?

I would say so.

specs/core/1_shard-data-chains.md
vbuterin: > a shard block proposer does not know he is the proposer for a given slot until he sees the beacon chain block for that slot. True/False?

Yes.

> So the shard block proposer is proposing some fractional time into the slot?

I would say the shard block proposer should propose either (i) immediately after they see the proposal, or (ii) halfway through a slot when they see no proposal.

> then I suggest that we incorporate the slot into the proposer index as such: curblock_proposer_index = hash(state.randao_mix + bytes8(shard_id) + slot) % len(validators)

I was just about to suggest the same thing :)

specs/core/1_shard-data-chains.md
vbuterin: The local nodes that are aware of a bad crosslink being included in some beacon chain, then the nodes should ignore that beacon chain.

This does mean that there will be some chaos and splitting in the event of a 51% attack bringing in invalid data, but IMO that's a good thing, the invariant that exactly 100% of crosslinks in the "real" beacon chain are legitimate is worth defending, and it's better than letting the attacker succeed. In the long term, data availability proofs and similar tech should make this easier.

specs/core/1_shard-data-chains.md
vbuterin: It's prescriptive to primarily crosslinkers, but ultimately it's a chain validity rule. A "super-full node" should validate all of this for example.

specs/core/1_shard-data-chains.md
vbuterin: A super-full-node would be verifying all of this. A "regular" node would be a light client with respect to the shards, so they would generally skip this part of the verification, relying on the assumption that the committee did its job.

specs/core/1_shard-data-chains.md
vbuterin: > From a pragmatic standpoint I'm inclined to avoid specifying phase 1 (this pull request and the data availability proofs) before phase 0. Reasons include:
> 
> 1. There is a real possibility that we will be "forced" to launch the beacon chain without a STARK-friendly hash function. But this doesn't preclude the shard blocks using a STARK-friendly hash function required for the data availability proofs.
> 2. We want to limit the complexity of the phase 0 to launch the beacon chain ASAP. Delaying the specification of phase 1 is a significant simplification to the phase 0 spec.
> 3. We don't want to rush phase 1. (As a side note, I'm not totally sold on variable sized blocks.)
> 4. We want phase 0 to be maximally forward compatible to easily incorporate new ideas from between the launches of phase 0 and phase 1.

@JustinDrake splitting phase 1 from phase 0 and putting it in a separate doc seems totally fine to me. All of it is pretty logically separable in any case.

djrtwo:
@vbuterin @JustinDrake I separated out phase 0 and phase 1 into two separate docs. I also did some general cleanup.

I still have a number of outstanding questions. Can you take a look?

specs/core/1_shard-data-chains.md
vbuterin: Fixed.

specs/core/1_shard-data-chains.md
vbuterin: Just fixed now.

Issue #43, opened 2018-10-04
Max length of specials


djrtwo:
# Issue
`Block.specials` currently has no max length. We also don't check the validity of `SpecialRecord`s until doing state recalcs. This can lead to a dos vector in which Blocks are filled with junk specials causing an increased load on block size as well as state recalc time.

# Proposed solution
- Add constant `MAX_SPECIALS_PER_BLOCK` such that block validity requires `len(block.specials) <= MAX_SPECIALS_PER_BLOCK`
- Check `SpecialRecord` validity (signature, format, etc) in block processing after checking attestation validity. Throw out block if any invalid specials
- During state recalc, `ActiveState.pending_specials` are valid so just process the related state transitions rather than the signatures at this point.

djrtwo:
cc: @vbuterin, @JustinDrake

JustinDrake:
Further questions:

1) How are duplicate specials handled?
2) What about specials of unknown type?
3) Is there a maximum number of `data` elements?
4) Is there a maximum byte length per special `data` element?

vbuterin:
> How are duplicate specials handled?

More broadly, it seems like the issue is that specials are missing validity conditions. A slash should only be includable if the validator has not been already slashed, a logout should only be includable if the validator is logged in, etc. So we don't need explicit anti-duplicate rules.

> What about specials of unknown type?

I'm inclined to say ban them.

> Is the length of SpecialRecord.data bounded? Is the byte length of individual SpecialRecord.data members bounded?

I'm inclined to say not explicitly, but we can require the type-specific checkers to require exact byte length and element count.

djrtwo:
> A slash should only be includable if the validator has not been already slashed

If we check special validity during block processing, but then only apply the effects of the special during a state recalc, then someone could easily slip in a duplicate (effect) of a special multiple times in cycle. I suppose this is also true for attestations in the current spec..

djrtwo:
This is addressed at least in it's basic iteration in #124

PR #124, opened 2018-11-09
Increased strictness re specials

vbuterin:
In response to #43

### DIFF
--- a/specs/beacon-chain.md
+++ b/specs/beacon-chain.md
@@ -676,7 +676,7 @@ For each one of these attestations:
 * Let `fork_version = pre_fork_version if slot < fork_slot_number else post_fork_version`.
 * Verify that `aggregate_sig` verifies using the group pubkey generated and the serialized form of `AttestationSignedData(fork_version, slot, shard, parent_hashes, shard_block_hash, justified_slot)` as the message.
 
-Extend the list of `AttestationRecord` objects in the `active_state` with those included in the block, ordering the new additions in the same order as they came in the block. Similarly extend the list of `SpecialRecord` objects in the `active_state` with those included in the block.
+Extend the list of `AttestationRecord` objects in the `active_state` with those included in the block, ordering the new additions in the same order as they came in the block. Similarly extend the list of `SpecialRecord` objects in the `active_state` with those included in the block. Verify that there are at most `MAX_SPECIALS_PER_BLOCK` special records.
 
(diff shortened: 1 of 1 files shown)


specs/beacon-chain.md
```diff
@@ -742,11 +743,15 @@ For every shard number `shard` for which a crosslink committee exists in the cyc
 
 #### Process penalties, logouts and other special objects
 
+Verify that there are at most `MAX_SPECIALS_PER_BLOCK` special records.
```
djrtwo: This should go up in the block validity conditions rather than in the state recalc section

specs/beacon-chain.md
```diff
@@ -742,11 +743,15 @@ For every shard number `shard` for which a crosslink committee exists in the cyc
...
+* **[covers RANDAO updates]**: If `obj.kind == RANDAO_REVEAL`, interpret `data[0]` as an `int64` and `data[1]` as a hash32. Set `validators[data[0]].randao_commitment = data[1]`. Expects exactly 2 data items.
+
+In all cases, each special record must have exactly the number of data items specified. Each special record must be in one of the above categories. Each data item must be exactly the correct length depending on what it represents.
```
djrtwo: This requirement of special records to be "well formatted" should go in the block validity section as well. They should never be added to the state to be processed during a state recalc if malformed.

The conditions _beyond_ just having correct number and length of items plus being in the type categories should remain in state recalc I believe.

djrtwo:
Check out my couple of comments. I can handle merge conflict after you address

specs/beacon-chain.md
JustinDrake: `MAX_SPECIALS_PER_BLOCK` feels suboptimal because specials have different overhead for beacon chain. Here are two extreme scenarios:

1) The block proposers have a large influx of lightweight specials where including more than `MAX_SPECIALS_PER_BLOCK` per block would be preferable.
2) An attacking proposer DoSes the network by filling blocks with `MAX_SPECIALS_PER_BLOCK` maximally-heavy specials.

What about replacing total specials count by total specials size? (E.g. a maximum of 4kB of specials per block.) Total specials size only captures the bandwidth overhead (not the computational overhead), so maybe it's wise to have both total count and total size limits.

djrtwo: Having a max specials per block along with a block size limit might be the right compromise

specs/beacon-chain.md
vbuterin: State recalcs are part of the block validity check, so is there a benefit to separating them out?

specs/beacon-chain.md
vbuterin: I'm ok either way. Maybe better to not touch this until we're done all the specials including proof of custody challenges, and we get numbers on how long each take to process?

djrtwo:
@vbuterin Are you going to rework this PR to just process specials on the fly?

vbuterin:
Replaced with #138

PR #139, opened 2018-11-16
ssz: switch integer encoding to little endian

arnetheduck:
the choice between little and big endian is arbitrary from a functional
point of view, but practially:

* most commodity hardware these days is either little- or biendian
* mechanical sympathy between encoding and hardware allows a wider range
of tricks to be used when encoding and decoding data leading to better
efficiency
* we're developing a format that favors "decoding-free" access to data

### DIFF
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -40,7 +40,7 @@ overhead.
 
 | Term         | Definition                                                                                     |
 |:-------------|:-----------------------------------------------------------------------------------------------|
-| `big`        | Big Endian                                                                                     |
+| `little`     | Little Endian                                                                                  |
 | `byte_order` | Specifies [endianness:](https://en.wikipedia.org/wiki/Endianness) Big Endian or Little Endian. |
 | `len`        | Length/Number of Bytes.                                                                        |
 | `to_bytes`   | Convert to bytes. Should take parameters ``size`` and ``byte_order``.                          |
@@ -63,7 +63,7 @@ overhead.
 
 Convert directly to bytes the size of the int. (e.g. ``uint16 = 2 bytes``)
 
-All integers are serialized as **big endian**.
+All integers are serialized as **little endian**.
 
 | Check to perform       | Code                  |
 |:-----------------------|:----------------------|
@@ -72,7 +72,7 @@ All integers are serialized as **big endian**.
 ```python
 assert(int_size % 8 == 0)
 buffer_size = int_size / 8
-return value.to_bytes(buffer_size, 'big')
+return value.to_bytes(buffer_size, 'little')
 ```
 
 #### bool
@@ -135,7 +135,7 @@ For general `bytes` type:
 
 ```python
 assert(len(value) < 2**32)
-byte_length = (len(value)).to_bytes(LENGTH_BYTES, 'big')
+byte_length = (len(value)).to_bytes(LENGTH_BYTES, 'little')
 return byte_length + value
 ```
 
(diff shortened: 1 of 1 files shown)


JustinDrake:
Pinging @AlexeyAkhunov and @karalabe. Big or little endian?

djrtwo:
The main push-back I heard on this was that eth1.0 uses big-endian so don't introduce a different-endian encoding in eth2.0. The small gain in efficiency is not worth the potential confusion and overhead of having to remember which is which.

arnetheduck:
Well, there's a fairly clean break here, considering SSZ vs RLP - it's also an unfortunate fact of life that you have to remember endianness whenever you deal with binary protocols in general.

The main performance difference will be that adjacent integers can either be bulk-copied or will have to be byte-flipped one-by-one. It affects both network serialization and hashing.

JustinDrake:
> The main performance difference will be that adjacent integers can either be bulk-copied or will have to be byte-flipped one-by-one.

Can this be quantified? What is the performance difference?

> It affects both network serialization and hashing.

Would it negatively affect light clients of Ethereum 2.0 built in Ethereum 1.0 contracts?

mkalinin:
It could be that a win in efficiency gained with this optimization is too low comparing with efficiency of other operations, for example, calculating a hash of validators registry.

Another thing is that all big number implementations in Java that I've seen uses big-endian to encode/decode numbers to/from byte arrays. So does Milagro, even in [C implementation](https://github.com/milagro-crypto/milagro-crypto-c/blob/master/src/big.c.in#L141). What about other languages, btw? And in our case it results in reversing signature bytes on each encoding/decoding since signature has `uint384` type. There is a workaround to it, use `byteN` type for all big numbers starting from `uint72`. Otherwise, reversing `48` bytes could be much less efficient than reversing byte order in several adjacent primitives that I believe could be done with bitwise arithmetic.

arnetheduck:
> Can this be quantified? What is the performance difference?

I'll see if I can pull up some numbers, but we're really not on that stage yet (it's a pretty low-level / final-touch optimization) - the idea itself is mainly taken from other modern serialization formats that state "direct access" as one of their design goals, for example [flatbuffers](https://google.github.io/flatbuffers/md__internals.html). 

> all big number implementations 

yep - though here the machine endianess no longer matters - there's no mechanical sympathy to consider because you can't directly use these numbers anyway, and at this point, it's kind of.. arbitrary.

arnetheduck:
anyway, if there's pushback, we can certainly drop this - it's a drop in the sea, as @mkalinin points out (or one of many paper cuts).

JustinDrake:
> one of many paper cuts

@arnetheduck Is there anything beyond [your current 3 issues and 1 PR](https://github.com/ethereum/eth2.0-specs/issues?q=author%3Aarnetheduck+is%3Aopen)? I'm keen to address as many issues as possible before we declare the spec a release candidate, so now is a good time to flag things. 👍

arnetheduck:
> one of many paper cuts
> Is there anything beyond

ah, I hope that it did not came across wrong - it was intended as a general command and not to say that there are necessarily many in the spec as of now :)

I'll go over my notes and see what is still relevant after the latest refactorings (:+1: good work!), and post ASAP!

sorpaas:
SSZ and RLP are already vastly different, so I think using big-endian because eth1.0 uses big-endian may not be that of a really strong argument.

Besides all the architectures using little-endian, for Parity there's also a really specific reason we would prefer little-endian -- our [parity-codec](https://github.com/paritytech/parity-codec/) uses little-endian, and `parity-codec` and `ssz` are nearly identical in all the basic forms, just except the endianness. By using little-endian, we can unify those two formats!

JustinDrake:
Pros of little-endian:

* Consistent with WASM
* Consistent with commodity hardware
* Consistent with `parity-codec`

Cons of little-endian:

* Inconsistent with RLP (I agree this is a weak argument)
* Inconsistent with big number implementations (can be worked around with `byteN`)

Little-endian feels on net positive :)

PR #140, opened 2018-11-17
simplify merkle tree chunking

arnetheduck:
* pack small items tightly to fit more items in single chunk, decreasing
the number of hash operations needed
* remove chunk padding - hash algorithm will pad to its own block size
anyway
* express data length in number of items instead of binary bytes at leaf
level (equivalent)

### DIFF
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -402,40 +402,32 @@ Return the hash of the serialization of the value.
 First, we define some helpers and then the Merkle tree function. The constant `CHUNK_SIZE` is set to 128.
 
 ```python
-# Returns the smallest power of 2 equal to or higher than x
-def next_power_of_2(x):
-    return x if x == 1 else next_power_of_2((x+1) // 2) * 2
-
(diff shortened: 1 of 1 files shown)


vbuterin:
Seems reasonable to me...

specs/simple-serialize.md
```diff
@@ -402,40 +402,32 @@ Return the hash of the serialization of the value.
...
+
     # Tree-hash
     while len(chunkz) > 1:
         if len(chunkz) % 2 == 1:
             chunkz.append(b'\x00' * CHUNKSIZE)
         chunkz = [hash(chunkz[i] + chunkz[i+1]) for i in range(0, len(chunkz), 2)]
+
     # Return hash of root and length data
-    return hash((chunkz[0] if len(chunks) > 0 else b'\x00' * 32) + datalen)
+    return hash((chunkz[0] + datalen)
```
hwwhww: A redundant `(`

```suggestion
    return hash(chunkz[0] + datalen)
```

PR #145, opened 2018-11-18
Add proofs of custody

vbuterin:
Introduces the following major changes:

* The bitfield now has 2 bits per validator; the second bit is a proof of custody bit
* Three new special objects, one for updating a validator's proof of custody seed, one for making a challenge, one for responding to a challenge
* If a validator fails to respond on time (~3 months) they get kicked out and penalized.

PR #149, opened 2018-11-19
Replace cycles with epochs

vbuterin:
Also adjusted constants to keep interest rates and the quadratic leak period the same.

Also, did some simplifications of the cycle calculation procedure. Make the decision to remove automatic registration of a validator as a proposer when they join, because that can just happen on its own due to the reshuffling procedure.


specs/core/0_beacon-chain.md
```diff
@@ -836,20 +857,19 @@ def update_ancestor_hashes(parent_ancestor_hashes: List[Hash32],
...
-* Verify that `len(attester_bitfield) == ceil_div8(len(attestation_indices))`, where `ceil_div8 = (x + 7) // 8`. Verify that bits `len(attestation_indices)....` and higher, if present (i.e. `len(attestation_indices)` is not a multiple of 8), are all zero.
-* Derive a group public key by adding the public keys of all of the attesters in `attestation_indices` for whom the corresponding bit in `attester_bitfield` (the ith bit is `(attester_bitfield[i // 8] >> (7 - (i %8))) % 2`) equals 1.
+* Compute `full_parent_hashes` = `[get_block_hash(state, block, slot - CYCLE_LENGTH + i) for i in range(1, CYCLE_LENGTH - len(parent_hashes) + 1)] + parent_hashes` (eg, if `CYCLE_LENGTH = 4`, `slot = 5`, the actual block hashes starting from slot 0 are `Z A B C D E F G H I J`, and `parent_hashes = [D', E']` then `parent_hashes = [B, C, D' E']`). Note that when *creating* an attestation for a block, the hash of that block itself won't yet be in the `state`, so you would need to add it explicitly.
```
hwwhww: Should that be "and `parent_hashes = [D', E']` then the `full_parent_hashes = [B, C, D' E']`"?

specs/core/0_beacon-chain.md
```diff
@@ -520,6 +518,25 @@ def get_beacon_proposer(state:BeaconState, slot: int) -> ValidatorRecord:
     return state.validators[index]
 ```
 
+The following is a function that determines the validators that participated in an attestation:
+
+```python
+def get_attestation_participants(state: State,
```
hwwhww: Should this update be part of #145 instead of this PR? If not, could you explain more about why we need two bits for having epochs?

specs/core/0_beacon-chain.md
```diff
@@ -955,60 +935,79 @@ Verify that `deposit_data.msg_value == DEPOSIT_SIZE` and `block.slot - (deposit_
...
 
-* Let `total_balance_participating` be the total balance of validators that voted for the canonical beacon block at slot `s`. In the normal case every validator will be in one of the `CYCLE_LENGTH` slots following slot `s` and so can vote for a block at slot `s`.
-* Let `B` be the balance of any given validator whose balance we are adjusting, not including any balance changes from this round of state recalculation.
-* If `time_since_finality <= 3 * CYCLE_LENGTH` adjust the balance of participating and non-participating validators as follows:
-    * Participating validators gain `B // reward_quotient * (2 * total_balance_participating - total_balance) // total_balance`. (Note that this value may be negative.)
-    * Non-participating validators lose `B // reward_quotient`.
-* Otherwise:
-    * Participating validators gain nothing.
-    * Non-participating validators lose `B // reward_quotient + B * time_since_finality // quadratic_penalty_quotient`.
+* Any validator `V` in `prev_s_attesters` gains `adjust_for_inclusion_distance(balance_at_stake(V) // reward_quotient * (prev_s_attesters - total_balance) // total_balance, inclusion_distance(V))``.
```
JustinDrake: Formatting issue with double tick ``

JustinDrake: Also use lowercase `v` throughout.

specs/core/0_beacon-chain.md
```diff
@@ -955,60 +935,79 @@ Verify that `deposit_data.msg_value == DEPOSIT_SIZE` and `block.slot - (deposit_
...
 
-* Let `total_balance_of_v` be the total balance of `V`.
-* Let `winning_shard_hash` be the hash that the largest total deposits signed for the `shard` during the cycle.
-* Define a "participating validator" as a member of `V` that signed a crosslink of `winning_shard_hash`.
-* Let `total_balance_of_v_participating` be the total balance of the subset of `V` that participated.
-* Let `time_since_last_confirmation = block.slot - crosslinks[shard].slot`.
-* Adjust balances as follows:
-    * Participating validators gain `B // reward_quotient * (2 * total_balance_of_v_participating - total_balance_of_v) // total_balance_of_v`.
-    * Non-participating validators lose `B // reward_quotient`.
+* If `v in attesting_validators(obj)`, `V.balance += adjust_for_inclusion_distance(balance_at_stake(V) // reward_quotient * total_attesting_balance(obj) // total_balance(obj)), inclusion_distance(V))`.
```
JustinDrake: Notation mismatch with `v` and `V`

JustinDrake:
Looking forward to having this PR merged :)

One suggestion is to split `attester_bitfield` into `signature_bitfield` and `custody_bitfield` for cleaner logic and more modularity.

specs/core/0_beacon-chain.md
```diff
@@ -941,60 +976,79 @@ Verify that `deposit_data.msg_value == DEPOSIT_SIZE` and `block.slot - (deposit_
...
+
+_Note: `last_state_recalculation_slot` will always be a multiple of `CYCLE_LENGTH`. In the "happy case", this process will trigger, and loop once, every time `block.slot` passes a new exact multiple of `CYCLE_LENGTH`, but if a chain skips more than an entire cycle then the loop may run multiple times, incrementing `last_state_recalculation_slot` by `CYCLE_LENGTH` with each iteration._
+
+#### Precomputation
+
+* Let `active_validators = [state.validators[i] for i in get_active_validator_indices(state.validators)]`.
+* Let `total_balance = sum([balance_at_stake(v) for v in active_validators])`.
+* Let `this_cycle_attestations = [a for a in state.pending_attestations if s <= a.slot < s + CYCLE_LENGTH]`. (note: this is the set of attestations _of slots in the cycle `s...s+CYCLE_LENGTH-1`_, not attestations _that got included in the chain during the cycle `s...s+CYCLE_LENGTH-1`_)
+* Let `prev_cycle_attestations = [a for a in state.pending_attestations if s - CYCLE_LENGTH <= a.slot < s]`.
+* Let `this_cycle_s_attestations = [a for a in this_cycle_attestations if get_block_hash(state, block, s) in a.parent_hashes and a.justified_slot == state.justification_source]`.
```
djrtwo: It is very unclear what is going on here and in the next line.

What does `this_cycle_s_attestations` mean?
Why are we checking to see if the current block is in `a.parent_hashes`? `a.parent_hashes` is the oblique hashes, right? This seems wrong. The current block should _never_ be in `a.parent_hashes`

I'm not in favor of changing `Attestation.oblique_parent_hashes` to `parent_hashes`. `parent_hashes` is very unclear imo.

specs/core/0_beacon-chain.md
```diff
@@ -941,60 +976,79 @@ Verify that `deposit_data.msg_value == DEPOSIT_SIZE` and `block.slot - (deposit_
...
+* Let `prev_s_attesters` be the union of the validator index sets given by `[get_attestation_participants(state, a) for a in prev_s_attestations]`.
+* Let `total_balance_attesting_at_s = sum([balance_at_stake(v) for v in s_attesters])`.
+* Let `total_balance_attesting_at_prev_s = sum([balance_at_stake(v) for v in prev_s_attesters])`.
+* For every `ShardAndCommittee` object `obj` in `shard_and_committee_for_slots`, let:
+    * `attesting_validators(obj, shard_block_hash)` be the union of the validator index sets given by `[get_attestation_participants(state, a) for a in this_cycle_attestations + prev_cycle_attestations if a.shard == obj.shard and a.shard_block_hash == shard_block_hash]`
+    * `attesting_validators(obj)` be equal to `attesting_validators(obj, shard_block_hash)` for the value of `shard_block_hash` such that `sum([balance_at_stake(v) for v in attesting_validators(obj, shard_block_hash)])` is maximized (ties broken by favoring lower `shard_block_hash` values)
+    * `total_attesting_balance(obj)` be the maximal sum balance
```
djrtwo: This is unclear. Maybe
```suggestion
    * `total_attesting_balance(obj)` be the sum of the `balance_at_stake` of the `attestation_validators`
```

specs/core/0_beacon-chain.md
```diff
@@ -941,60 +976,79 @@ Verify that `deposit_data.msg_value == DEPOSIT_SIZE` and `block.slot - (deposit_
...
+    * `winning_hash(obj)` be the winning `shard_block_hash` value
+    * `total_balance(obj) = sum([balance_at_stake(v) for v in obj.committee])`
+    
+Let `inclusion_slot(a)` equal the slot in which attestation `a` was included, and `inclusion_distance(a) = inclusion_slot(a) - a.slot`. Let `inclusion_slot(v)` equal `inclusion_distance(a)` for the attestation `a` where `v` is in `get_attestation_participants(state, a)`, and define `inclusion_distance(v)` similarly. We define a function `adjust_for_inclusion_distance(magnitude, dist)` which adjusts the reward of an attestation based on how long it took to get included (the longer, the lower the reward). Returns a value between 0 and `magnitude`
```
djrtwo: `inclusion_slot` is not directly known by the state at this point. The client would have to do caching or windback to find it.

Overloading `inclusion_slot` and `inclusion_distance` is confusing. I think we should prefer an alternative name.

specs/core/0_beacon-chain.md
```diff
@@ -941,60 +976,79 @@ Verify that `deposit_data.msg_value == DEPOSIT_SIZE` and `block.slot - (deposit_
...
-* Let `quadratic_penalty_quotient = SQRT_E_DROP_TIME**2`. (The portion lost by offline validators after `D` slots is about `D*D/2/quadratic_penalty_quotient`.)
-* Let `time_since_finality = block.slot - last_finalized_slot`.
+* Let `reward_quotient = BASE_REWARD_QUOTIENT * int_sqrt(total_balance_in_eth)`. (The per-slot maximum interest rate is `2/reward_quotient`.)
+* Let `quadratic_penalty_quotient = SQRT_E_DROP_TIME**2`. (The portion lost by offline validators after `D` cycles is about `D*D/2/quadratic_penalty_quotient`.)
+* Let `time_since_finality = slot - last_finalized_slot`.
+
+Case 1: `time_since_finality <= 4 * CYCLE_LENGTH`:
 
-For every slot `s` in the range `last_state_recalculation_slot - CYCLE_LENGTH ... last_state_recalculation_slot - 1`:
+* Any validator `v` in `prev_s_attesters` gains `adjust_for_inclusion_distance(balance_at_stake(v) // reward_quotient * (prev_s_attesters - total_balance) // total_balance, inclusion_distance(v))`.
```
djrtwo: Define in a line above
```
* Let `base_reward_magnitude = balance_at_stake(v) // reward_quotient * (prev_s_attesters - total_balance) // total_balance`
```

specs/core/0_beacon-chain.md
```diff
@@ -941,60 +976,79 @@ Verify that `deposit_data.msg_value == DEPOSIT_SIZE` and `block.slot - (deposit_
...
+```python
+def adjust_for_inclusion_distance(magnitude: int, dist: int) -> int:
+    return magnitude // 2 + (magnitude // 2) * MIN_ATTESTATION_INCLUSION_DELAY // dist
+```
 
 #### Adjust justified slots and crosslink status
 
-For every slot `s` in the range `last_state_recalculation_slot - CYCLE_LENGTH ... last_state_recalculation_slot - 1`:
+* Set `state.justified_slot_bitfield = (state.justified_slot_bitfield * 2) % 2**64`.
+* If `3 * total_balance_attesting_at_prev_s >= 2 * total_balance` then set `state.justified_slot_bitfield &= 2` (ie. flip the second lowest bit to 1) and `new_justification_source = prev_s`.
```
djrtwo: `prev_s` is undefined.

specs/core/0_beacon-chain.md
```diff
@@ -941,60 +976,79 @@ Verify that `deposit_data.msg_value == DEPOSIT_SIZE` and `block.slot - (deposit_
...
+* If `3 * total_balance_attesting_at_s >= 2 * total_balance` then set `state.justified_slot_bitfield &= 1` (ie. flip the lowest bit to 1) and `new_justification_source = s`.
+* If `justification_source == prev_s and state.justified_slot_bitfield % 4 == 3`, set `last_finalized_slot = justification_source`.
```
djrtwo: `state.` is implicit when referencing `justification_source` and `last_finalized_slot` in this section. Consider making explicit

specs/core/0_beacon-chain.md
```diff
@@ -941,60 +976,79 @@ Verify that `deposit_data.msg_value == DEPOSIT_SIZE` and `block.slot - (deposit_
...
+* If `justification_source == prev_s - 2 * CYCLE_LENGTH and state.justified_slot_bitfield % 16 in (15, 14)`, set `last_finalized_slot = justification_source`.
+* If `justification_source == prev_s - CYCLE_LENGTH and state.justified_slot_bitfield % 8 == 7`, set `last_finalized_slot = justification_source`.
+* Set `prev_cycle_justification_source = justification_source` and `justification_source = new_justification_source`.
```
djrtwo: There's a chance that `new_justification_source` is not defined when reaching this line.

specs/core/0_beacon-chain.md
```diff
@@ -941,60 +976,79 @@ Verify that `deposit_data.msg_value == DEPOSIT_SIZE` and `block.slot - (deposit_
...
 Note: When applying penalties in the following balance recalculations implementers should make sure the `uint64` does not underflow.
 
-* Let `total_balance` be the total balance of active validators.
 * Let `total_balance_in_eth = total_balance // GWEI_PER_ETH`.
-* Let `reward_quotient = BASE_REWARD_QUOTIENT * int_sqrt(total_balance_in_eth)`. (The per-slot maximum interest rate is `1/reward_quotient`.)
-* Let `quadratic_penalty_quotient = SQRT_E_DROP_TIME**2`. (The portion lost by offline validators after `D` slots is about `D*D/2/quadratic_penalty_quotient`.)
-* Let `time_since_finality = block.slot - last_finalized_slot`.
+* Let `reward_quotient = BASE_REWARD_QUOTIENT * int_sqrt(total_balance_in_eth)`. (The per-slot maximum interest rate is `2/reward_quotient`.)
+* Let `quadratic_penalty_quotient = SQRT_E_DROP_TIME**2`. (The portion lost by offline validators after `D` cycles is about `D*D/2/quadratic_penalty_quotient`.)
+* Let `time_since_finality = slot - last_finalized_slot`.
```
djrtwo: considering using `state.last_finalized_slot`

specs/core/0_beacon-chain.md
```diff
@@ -941,60 +976,79 @@ Verify that `deposit_data.msg_value == DEPOSIT_SIZE` and `block.slot - (deposit_
...
+* Any active validator `v` not in `prev_s_attesters`, and any validator with `status == PENALIZED`, loses `balance_at_stake(v) // reward_quotient + balance_at_stake(v) * time_since_finality // quadratic_penalty_quotient`.
+
+For each `v` in `prev_s_attesters`, the validator `proposer = get_beacon_proposer(state, inclusion_slot(v))` gains `balance_at_stake(proposer) // INCLUDER_REWARD_QUOTIENT`.
 
 #### Balance recalculations related to crosslink rewards
 
-For every shard number `shard` for which a crosslink committee exists in the cycle prior to the most recent cycle (`last_state_recalculation_slot - CYCLE_LENGTH ... last_state_recalculation_slot - 1`), let `V` be the corresponding validator set. Let `B` be the balance of any given validator whose balance we are adjusting, not including any balance changes from this round of state recalculation. For each `shard`, `V`:
+For every `ShardAndCommittee` object `obj` in `shard_and_committee_for_slots[:CYCLE_LENGTH]` (ie. the objects corresponding to the slot before the current one), for each `v in obj.committee`, where `v = state.validators[b]` adjust balances as follows:
```
djrtwo: "to the slot before the current one" is unclear

djrtwo: Is it "to the cycle before the current one"?

djrtwo: `b` is unclear and undefined in the line `v = state.validators[b]`

specs/core/0_beacon-chain.md
```diff
@@ -1003,11 +1057,39 @@ If `last_state_recalculation_slot % POW_RECEIPT_ROOT_VOTING_PERIOD == 0`, then:
 * If for any `x` in `state.candidate_pow_receipt_root`,  `x.votes * 2 >= POW_RECEIPT_ROOT_VOTING_PERIOD` set `state.processed_pow_receipt_root = x.receipt_root`.
 * Set `state.candidate_pow_receipt_roots = []`.
 
-### Validator set change
+#### Proposer reshuffling
```
djrtwo: Why did we move this here?

specs/core/0_beacon-chain.md
```diff
@@ -1065,70 +1147,73 @@ def change_validators(validators: List[ValidatorRecord], current_slot: int) -> N
...
+        # STUB: withdraw to shard chain    
 ```
 
-* Set `state.validator_set_change_slot = state.last_state_recalculation_slot`
-* Set `shard_and_committee_for_slots[:CYCLE_LENGTH] = shard_and_committee_for_slots[CYCLE_LENGTH:]`
-* Let `next_start_shard = (shard_and_committee_for_slots[-1][-1].shard + 1) % SHARD_COUNT`
-* Set `shard_and_committee_for_slots[CYCLE_LENGTH:] = get_new_shuffling(state.next_shuffling_seed, validators, next_start_shard)`
+And perform the following updates to the `state`:
+
+* Set `state.validator_set_change_slot = s`
```
djrtwo: This actually needs to be `state.validator_set_change_slot = s + CYCLE_LENGTH`

Because we want to capture that it happened _now_ rather than a cycle ago, and we have't yet updated the `last_state_recalculation_slot`

djrtwo:
Not quite ready @JustinDrake !

JustinDrake:
@djrtwo Yeah, a bunch of cleanup is required but want to avoid merge conflicts :)

specs/core/0_beacon-chain.md
```diff
@@ -836,20 +817,19 @@ def update_ancestor_hashes(parent_ancestor_hashes: List[Hash32],
 
 ### Verify attestations
 
-For each `AttestationRecord` object:
+For each `AttestationRecord` object `obj`:
 
 * Verify that `slot <= block.slot - MIN_ATTESTATION_INCLUSION_DELAY` and `slot >= max(parent.slot - CYCLE_LENGTH + 1, 0)`.
-* Verify that `justified_slot` is equal to or earlier than `last_justified_slot`.
+* Verify that `justified_slot` is equal to `justification_source if slot >= block.slot - (block.slot % CYCLE_LENGTH) else prev_cycle_justification_source`
```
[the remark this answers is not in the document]
vbuterin: Replace >= with > ?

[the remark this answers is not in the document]
djrtwo: Don't think that fixes it.
The problem is with the block that triggers the state recalc. It might not even be at the cycle boundary.

We can check it with `last_state_recalculation_slot`

```suggestion
* Verify that `justified_slot` is equal to `justification_source if slot >= state.last_state_recalculation_slot else prev_cycle_justification_source`
```

djrtwo:
I've been keeping up to date with master :)
Should be good to merge when ready

specs/core/0_beacon-chain.md
```diff
@@ -956,60 +992,92 @@ Verify that `deposit_data.msg_value == DEPOSIT_SIZE` and `block.slot - (deposit_
...
+* `attesting_validators(obj)` be equal to `attesting_validators(obj, shard_block_hash)` for the value of `shard_block_hash` such that `sum([balance_at_stake(v) for v in attesting_validators(obj, shard_block_hash)])` is maximized (ties broken by favoring lower `shard_block_hash` values)
+* `total_attesting_balance(obj)` be the sum of the balances-at-stake of `attesting_validators(obj)`
+* `winning_hash(obj)` be the winning `shard_block_hash` value
+* `total_balance(obj) = sum([balance_at_stake(v) for v in obj.committee])`
+    
+Let `inclusion_slot(v)` equal `a.slot_included` for the attestation `a` where `v` is in `get_attestation_participants(state, a.data, a.attester_bitfield)`, and `inclusion_distance(v) = a.slot_included - a.data.slot` for the same attestation. We define a function `adjust_for_inclusion_distance(magnitude, dist)` which adjusts the reward of an attestation based on how long it took to get included (the longer, the lower the reward). Returns a value between 0 and `magnitude`
```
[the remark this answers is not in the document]
vbuterin: I'd say keep the words to "between 0 and magnitude". We could decide to change the function later on to have a floor lower than magnitude//2; the only guarantee we're committing to is "0 <= answer <= magnitude"

specs/core/0_beacon-chain.md
```diff
@@ -551,8 +557,32 @@ def get_beacon_proposer(state:BeaconState, slot: int) -> ValidatorRecord:
...
+                                 attester_bitfield: bytes) -> List[int]:
+    sncs_for_slot = get_shards_and_committees_for_slot(state, attestation_data.slot)
+    snc = [x for x in sncs_for_slot if x.shard == attestation_data.shard][0]
+    assert len(attester_bitfield) == ceil_div8(len(snc.committee))
+    participants = []
+    for i, vindex in snc.committee:
+        bit = (attester_bitfield[i//8] >> (7 - (i % 8))) % 2
+        if bit == 1:
+            participants.append(vindex)
+    return participants
```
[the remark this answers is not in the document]
vbuterin: One important issue is that we are being a bit sloppy by returning validator entries and then casually mutating them, using that as shorthand for "do that edit in the state". It means that functions that seem like they're returning pure objects are actually returning pointers, which could confuse people.

One option would be to err on the side of returning indices always, and then add a shorthand notation `sv(i) = state.validators[i]` that we use everywhere.

specs/core/0_beacon-chain.md
```diff
@@ -956,60 +992,93 @@ Verify that `deposit_data.msg_value == DEPOSIT_SIZE` and `block.slot - (deposit_
...
+def adjust_for_inclusion_distance(magnitude: int, dist: int) -> int:
+    return magnitude // 2 + (magnitude // 2) * MIN_ATTESTATION_INCLUSION_DELAY // dist
+```
+
+For any validator `v`, `base_reward(v) = balance_at_stake(v) // reward_quotient`
+
+Miscellaneous:
+
+* Let `quadratic_penalty_quotient = SQRT_E_DROP_TIME**2`. (The portion lost by offline validators after `D` cycles is about `D*D/2/quadratic_penalty_quotient`.)
+* Let `time_since_finality = block.slot - state.last_finalized_slot`.
```
djrtwo: `time_since_finality` needs to be defined after the next section because `state.last_finalized_slot` might have been updated

(last nitpick for now. sorry!)

specs/core/0_beacon-chain.md
vbuterin: Aah, I think my goal was to move it out of the then-overloaded "Finally" section. We could move it to happen right before Finally if desired. I'm not sure what order is best.

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
djrtwo: Returning indices is sensible 👍 

Shorthand is fine, but not 100% needed

specs/core/0_beacon-chain.md
djrtwo: The order doesn't particularly matter (I think). 

The only reason I noticed was -- if we have it before the change validator section then it might produce some `ShardReassignmentRecord`s that immediately get wiped out when removing validators in `change_validators`.

Again... not a pressing concern

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
vbuterin: Done.

specs/core/0_beacon-chain.md
```diff
@@ -545,14 +551,38 @@ def get_block_hash(state: BeaconState,
 The following is a function that determines the proposer of a beacon block:
 
 ```python
-def get_beacon_proposer(state:BeaconState, slot: int) -> ValidatorRecord:
+def get_beacon_proposer_index(state:BeaconState, slot: int) -> ValidatorRecord:
```
hwwhww: Type hint:
```suggestion
def get_beacon_proposer_index(state:BeaconState, slot: int) -> int:
```

specs/core/0_beacon-chain.md
```diff
@@ -137,60 +138,49 @@ An `AttestationRecord` has the following fields:
...
-    # Beacon block hashes not part of the current chain, oldest to newest
-    'oblique_parent_hashes': ['hash32'],
-    # Shard block hash being attested to
-    'shard_block_hash': 'hash32',
-    # Last crosslink hash
-    'last_crosslink_hash': 'hash32',
-    # Root of data between last hash and this one
-    'shard_block_combined_data_root': 'hash32',
-    # Attester participation bitfield (1 bit per attester)
+    data: AttestationSignedData
```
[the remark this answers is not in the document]
hwwhww: And 👍 👍 👍  to this!

specs/core/0_beacon-chain.md
```diff
@@ -716,7 +747,7 @@ This routine should be run for every validator that is inducted as part of a log
 First, some helper functions:
 
 ```python
-def min_empty_validator(validators: List[ValidatorRecord], current_slot: int):
+def min_empty_validator_index(validators: List[ValidatorRecord], current_slot: int):
```
hwwhww: Type hint
```suggestion
def min_empty_validator_index(validators: List[ValidatorRecord], current_slot: int) -> int:
```

specs/core/0_beacon-chain.md
```diff
@@ -326,6 +317,21 @@ A `CandidatePoWReceiptRootRecord` object contains the following fields:
...
 ```
 
+A `ProcessedAttestation` object has the following fields:
+
+```python
+{
+    # Signed data
+    'data': AttestationSignedData,
+    # Attester participation bitfield (2 bits per attester)
+    'attester_bitfield': 'bytes',
```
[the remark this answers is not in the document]
vbuterin: There are two bitfields, attester_bitfield and poc_bitfield. More modular that way :smile:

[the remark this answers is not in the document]
hwwhww: Ah I sorry, I mean the comment:

```python
    # Attester participation bitfield (2 bits per attester)
    'attester_bitfield': 'bytes',
    # Proof of custody bitfield
    'poc_bitfield': 'bytes',
```

Maybe just say "Attester participation bitfield" for `attester_bitfield`:
```python
    # Attester participation bitfield
    'attester_bitfield': 'bytes',
    # Proof of custody bitfield
    'poc_bitfield': 'bytes',
```

specs/core/0_beacon-chain.md
```diff
@@ -545,14 +551,38 @@ def get_block_hash(state: BeaconState,
...
     first_committee = get_shards_and_committees_for_slot(state, slot)[0].committee
     index = first_committee[slot % len(first_committee)]
-    return state.validators[index]
+    return index
+```
+
+The following is a function that determines the validators that participated in an attestation:
+
+```python
+def get_attestation_participants(state: State,
```
Bimnbuvala, line 563: Seeds

PR #154, opened 2018-11-22
Typo in Vyper contract

terenc3t:
Quick fixes:

* `s/s_byte/as_byte/` in vyper contract
* We should use `bytesX` casting in `add_validator` not `as_bytesX`

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -561,8 +561,8 @@ total_deposit_count: int128
 @public
 def deposit(deposit_params: bytes[2048]):
     index:int128 = self.total_deposit_count + 2**POW_CONTRACT_MERKLE_TREE_DEPTH
-    msg_gwei_bytes8: bytes[8] = slice(as_bytes32(msg.value / 10**9), 24, 8)
-    timestamp_bytes8: bytes[8] = slice(as_bytes32(block.timestamp), 24, 8)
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -562,7 +562,7 @@ total_deposit_count: int128
 def deposit(deposit_params: bytes[2048]):
     index:int128 = self.total_deposit_count + 2**POW_CONTRACT_MERKLE_TREE_DEPTH
     msg_gwei_bytes8: bytes[8] = slice(as_bytes32(msg.value / 10**9), 24, 8)
-    timestamp_bytes8: bytes[8] = slice(s_bytes32(block.timestamp), 24, 8)
+    timestamp_bytes8: bytes[8] = slice(as_bytes32(block.timestamp), 24, 8)
```
hwwhww: Actually, with current Vyper, we probably should use [`convert` function](https://vyper.readthedocs.io/en/latest/built-in-functions.html?highlight=convert#convert).
I haven't tested yet but it seems to be:
```suggestion
    timestamp_bytes8: bytes[8] = slice(convert(block.timestamp, 'bytes32'), 24, 8)
```

terenc3t: Sounds good! I updated `msg_gwei_bytes8` to use convert as well.

PR #155, opened 2018-11-22
ssz: Add CC0 copyright declaration

hwwhww:
Just to make sure.


NatoliChris:
@hwwhww I have no problems at all :smile:

PR #156, opened 2018-11-22
Fork choice rule

vbuterin:
Added the LMD GHOST fork choice rule.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -337,30 +337,29 @@ Beacon block production is significantly different because of the proof of stake
 
 ### Beacon chain fork choice rule
 
-The beacon chain fork choice rule is a hybrid that combines justification and finality with Latest Message Driven (LMD) GHOST. At any point in time a validator `v` calculates the head as follows.
+The beacon chain fork choice rule is a hybrid that combines justification and finality with Latest Message Driven (LMD) Greediest Heaviest Observed SubTree (GHOST). At any point in time a validator `v` calculates the beacon chain head as follows.
 
-* Let `store` be the set of all attestations and blocks that the validator `v` has verified (in particular, ancestors must be recursively verified), including attestations not included in any chain.
+* Let `store` be the set of attestations and blocks that the validator `v` has observed and verified (in particular, block ancestors must be recursively verified). Attestations not part of any chain are still included in `store`.
 * Let `finalized_head` be the finalized block with the highest slot number. (A block `B` is finalized if there is a descendant of `B` in `store` the processing of which sets `B` as finalized.)
 * Let `justified_head` be the descendant of `finalized_head` with the highest slot number that has been justified for at least `CYCLE_LENGTH` slots. (A block `B` is justified is there is a descendant of `B` in `store` the processing of which sets `B` as justified.) If no such descendant exists set `justified_head` to `finalized_head`.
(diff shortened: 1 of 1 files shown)


JustinDrake:
1) "Set `finalized_head` to the finalized block with the highest slot number." => Can I confirm that `finalized_head` can be a placeholder block if the slot is empty?
2) "Set `justified_head` be the descendant of `finalized_head` that is justified" => What if no such descendant exists?
3) Are `finalized_head` and `justified_head` relative to both onchain and offchain attestations, or only onchain votes?
4) "`store` contains all attestations and blocks that the validator has received" => Can I confirm that `store` contains offchain attestestations?
5) "Let `get_most_recent_attestation_target(store, v)` be a function that returns the block that a validator attested to in the attestation in store that has the highest slot number" => Can I confirm that `get_most_recent_attestation_target` must first filter out blocks that are not connected (via a path of ancestors) to `justified_head`?
6) How is `get_most_recent_attestation_target` defined if a validator made more than one attestation at the same slot?

vbuterin:
Just made an update to address these. For (1), the head should be an actual block (I don't think we use the "placeholder block" terminology in the spec....), and I think it's already the case that the block hash at slot N in a chain is the block hash of the most recent block in the chain with slot <= N.

JustinDrake:
Thanks :)

1) Can you check my edits make sense?
2) "it's already the case that the block hash at slot N in a chain is the block hash of the most recent block in the chain with slot <= N" => My intuition tells me it's not the "right" definition, i.e. that things progress (e.g. slot count increases, inactivity penalty is processed, ...) even if there are no "real" blocks.
3) Should we define the edge case where `get_most_recent_attestation(store, validator)` should return `None`?

specs/core/0_beacon-chain.md
```diff
@@ -337,15 +337,31 @@ Beacon block production is significantly different because of the proof of stake
...
 
-For an implementation with a network simulator see: **https://github.com/ethereum/research/blob/master/clock_disparity/ghost_node.py**
-
-Here's an example of its working (green is finalized blocks, yellow is justified, grey is attestations):
-
-![](https://vitalik.ca/files/RPJ.png)
+```python
+def lmd_ghost(store, start):
+    validators = start.state.validators
+    active_validators = [validators[i] for i in range(get_active_validators(validators, start.slot))]
```
hwwhww: `get_active_validators` has not been defined and we have `get_active_validator_indices(validators)`.

```suggestion
    active_validators = [validators[i] for i in range(get_active_validator_indices(validators))]
```

JustinDrake: Ok will fix 👍

vbuterin:
> "it's already the case that the block hash at slot N in a chain is the block hash of the most recent block in the chain with slot <= N" => My intuition tells me it's not the "right" definition, i.e. that things should progress (e.g. slot number increases, inactivity penalty is processed, ...) even if there are no "real" blocks, and block hashes should reflect that progress.

If we want to make things work this way, we can, though it would entail some deep changes to the state transition function. Essentially, the top-level block processing function would say "if `block.slot - block.parent.slot == N`, then run the underlying block processing function N-1 using a standardized empty block, and then once using `block`.

> How does the fork choice rule work in the edge case where there are more than one justified or finalised block?

Higher slot wins. If more than one with the same slot number (ie. an attack happened), then I would say pick the one that the client learned about earlier.

> "there are implementations that compute the head in logarithmic time" => logarithmic in what?

Logarithmic in slot count. Linear in the number of distinct objects that validators voted for in their most recent votes (ie. usually ~64).

PR #158, opened 2018-11-22
Various fixes regarding PoW receipt roots

JustinDrake:
Questions:

1) What happens if the block proposer at the `POW_RECEIPT_ROOT_VOTING_PERIOD` boundary puts a bad receipt root? Does this give too much power to that block proposer?
2) What if a significant portion of validators are offline ("WW3 scenario"). Does the condition `state.candidate_pow_receipt_root_votes * 3 >= POW_RECEIPT_ROOT_VOTING_PERIOD * 2` prevent new deposits from coming from the Ethereum 1.0 chain?

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -44,17 +44,17 @@ The primary source of load on the beacon chain are "attestations". Attestations 
 | `MIN_VALIDATOR_SET_CHANGE_INTERVAL` | 2**8 (= 256) | slots | ~25 minutes |
 | `SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD` | 2**17 (= 131,072) | slots | ~9 days |
 | `MIN_ATTESTATION_INCLUSION_DELAY` | 2**2 (= 4) | slots | ~24 seconds |
-| `RANDAO_SLOTS_PER_LAYER` | 2**12 (= 4096) | slots | ~7 hours |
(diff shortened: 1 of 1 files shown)


vbuterin:
> What happens if the block proposer at the POW_RECEIPT_ROOT_VOTING_PERIOD boundary puts a bad receipt root? Does this give too much power to that block proposer?

Then everyone else in the voting period would vote against it, so the block proposer can delay publication by 1024 blocks (ie. one hour). Oh well.

> What if a significant portion of validators are offline ("WW3 scenario"). Does the condition state.candidate_pow_receipt_root_votes * 3 >= POW_RECEIPT_ROOT_VOTING_PERIOD * 2 prevent new deposits from coming from the Ethereum 1.0 chain?

Yes. It's definitely consistency-favoring. Though note that if this happens in reality, then the offline validators will get leaked away, and so everntually the online validators will get back to 2/3.

JustinDrake:
> Then everyone else in the voting period would vote against it, so the block proposer can delay publication by 1024 blocks (ie. one hour). Oh well.

I think there's an easy fix. Just maintain a map from candidate PoW receipt roots (included in beacon blocks) to votes counts.

> Yes. It's definitely consistency-favoring.

Nooooo. I would advocate changing it to something more liveness favouring.

(One partial fix is that `state.candidate_pow_receipt_root_votes * 2 >= POW_RECEIPT_ROOT_VOTING_PERIOD` should be sufficient because we are making a 2/3 honesty assumption, so a threshold of 1/2 should be safe when sampling a committee of size `POW_RECEIPT_ROOT_VOTING_PERIOD == 1024`.)

PR #162, opened 2018-11-24
Only increment total_deposit_count for 32 ETH deposits from PoW chain

### LINKED ISSUE #152: The amount of checking performed by the PoW chain contract
The spec says the following:

> The registration contract emits a log with the various arguments for consumption by the beacon chain. It does not do validation, pushing the registration logic to the beacon chain.

This is reflected in the proposed Vyper PoW chain contract that does no checking.

At a minimum it would be good to check that `msg.value==DEPOSIT_SIZE`. This avoids a least a couple of issues:

* If someone mistakenly sends a different value to the contract, their funds will be unrecoverable under the current scheme.
* I could call `deposit()` 16384 times with 0 value and junk data, and yet start the Beacon Chain without a single genuine, committed validator.

vbuterin:
fixes #152

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -34,6 +34,7 @@ The primary source of load on the beacon chain are "attestations". Attestations 
 | --- | --- | :---: | - |
 | `SHARD_COUNT` | 2**10 (= 1,024)| shards |
 | `DEPOSIT_SIZE` | 2**5 (= 32) | ETH |
+| `MIN_TOPUP_SIZE` | 1 | ETH |
 | `MIN_ONLINE_DEPOSIT_SIZE` | 2**4 (= 16) | ETH |
 | `GWEI_PER_ETH` | 10**9 | Gwei/ETH |
 | `DEPOSIT_CONTRACT_ADDRESS` | **TBD** | - |
@@ -743,6 +744,7 @@ Now, to add a validator:
 ```python
 def add_validator(state: State,
                   pubkey: int,
+                  deposit_size: int,
                   proof_of_possession: bytes,
                   withdrawal_credentials: Hash32,
                   randao_commitment: Hash32,
@@ -756,23 +758,32 @@ def add_validator(state: State,
                      sig=proof_of_possession,
                      domain=get_domain(state, current_slot, DOMAIN_DEPOSIT))
     # Pubkey uniqueness
-    assert pubkey not in [v.pubkey for v in state.validators]
-    rec = ValidatorRecord(
-        pubkey=pubkey,
-        withdrawal_credentials=withdrawal_credentials,
-        randao_commitment=randao_commitment,
-        randao_skips=0,
-        balance=DEPOSIT_SIZE * GWEI_PER_ETH,
+    validator_pubkeys = [v.pubkey for v in state.validators]
+    if pubkey not in validator_pubkeys:
+        assert deposit_size == DEPOSIT_SIZE`
+        rec = ValidatorRecord(
+            pubkey=pubkey,
+            withdrawal_credentials=withdrawal_credentials,
+            randao_commitment=randao_commitment,
+            randao_skips=0,
+            balance=DEPOSIT_SIZE * GWEI_PER_ETH,
         status=status,
-        last_status_change_slot=current_slot,
-        exit_seq=0
-    )
-    index = min_empty_validator(state.validators)
-    if index is None:
-        state.validators.append(rec)
-        return len(state.validators) - 1
+            last_status_change_slot=current_slot,
+            exit_seq=0
+        )
+        index = min_empty_validator(state.validators)
+        if index is None:
+            state.validators.append(rec)
+            return len(state.validators) - 1
+        else:
+            state.validators[index] = rec
+            return index
     else:
-        state.validators[index] = rec
+        index = validator_pubkeys.index(pubkey)
+        val = state.validators[index]
+        assert val.withdrawal_credentials == withdrawal_credentials
+        assert deposit_size >= MIN_TOPUP_SIZE
+        val.balance += deposit_size
         return index
 ```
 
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -587,8 +588,9 @@ def deposit(deposit_params: bytes[2048]):
     for i in range(POW_CONTRACT_MERKLE_TREE_DEPTH):
         index //= 2
         self.receipt_tree[index] = sha3(concat(self.receipt_tree[index * 2], self.receipt_tree[index * 2 + 1]))
-    self.total_deposit_count += 1
-    if self.total_deposit_count == 16384:
+    if msg.value >= MIN_DEPOSIT_SIZE:
```
hwwhww: `MIN_DEPOSIT_SIZE` has not been defined. Do you mean `as_wei_value(DEPOSIT_SIZE, "ether")`?
```suggestion
    if msg.value >= as_wei_value(DEPOSIT_SIZE, "ether"):
```

vbuterin:
> Are we leaving this as more generic for top ups?

I _believe_ that was @JustinDrake's goal. Preserve optionality for possible future changes.

JustinDrake:
> Right now, if someone submits a non-32ETH deposit, it will go through in the contract, but then hang as a DEPOSIT into the beacon chain.

I would go ahead and implement sub-32ETH deposits on the beacon chain. It's just a few lines of code and unlocks emergency smaller-than-32ETH top-ups that may come in handy. As suggested [here](https://github.com/ethereum/eth2.0-specs/issues/152#issuecomment-441393333) I'd set the minimum deposit to be 1 ETH to mitigate spam on the beacon chain.

I don't care so much about larger-than-32ETH deposits. Certainly disallowing huge deposits to protect users may be wise (e.g. no deposits larger 64ETH).

djrtwo:
The 1 ETH mitigation doesn't prevent the early chain start spam.
I think we should still do the following regardless. Otherwise our min validator count can be bypassed for the cost of `1 * min_val` ETH

```python
if msg.value >= as_wei_value(DEPOSIT_SIZE, "ether"):
    self.total_deposit_count += 1
```

djrtwo:
updated the contract so that it actually builds in `vyper`

vbuterin:
Don't we want the minimum deposit to be equal to the MIN_ONLINE_DEPOSIT_SIZE ideally plus a safety margin?

Really we have three parameters:

* EVICTION_BALANCE (currently 16 ETH)
* MIN_DEPOSIT_SIZE (currently 32 ETH)
* MAX_EFFECTIVE_DEPOSIT (currently 32 ETH)

MIN_DEPOSIT_SIZE > EVICTION_BALANCE seems sensible. I could see MIN_DEPOSIT_SIZE < MAX_EFFECTIVE_DEPOSIT being reasonable too.

JustinDrake:
I'd like to differentiate between the "balance" (the current amount of ETH associated with a validator) and a "deposit" (an amount by which a balance is increased). For balances I'd suggest three parameters:

* `INDUCTION_BALANCE`—the minimum balance to be inducted (e.g. 32 ETH)
* `EVICTION_BALANCE`—the threshold balance under which a validator is automatically exited (e.g. 16 ETH)
* `MAX_EFFECTIVE_BALANCE`—the maximum balance at risk (e.g. 32 ETH)

For deposits I'd suggest two parameters:

* `MIN_DEPOSIT` (e.g. 1 ETH)
* `MAX_DEPOSIT` (e.g. 32 ETH)

To simplify things I'm happy with `INDUCTION_BALANCE == MAX_EFFECTIVE_BALANCE == MAX_DEPOSIT`.

specs/core/0_beacon-chain.md
```diff
@@ -582,31 +584,36 @@ def int_sqrt(n: int) -> int:
...
 
 @payable
 @public
 def deposit(deposit_params: bytes[2048]):
-    index:int128 = self.total_deposit_count + 2**POW_CONTRACT_MERKLE_TREE_DEPTH
-    msg_gwei_bytes8: bytes[8] = slice(concat("", convert(msg.value / 10**9, bytes32)), start=24, len=8)
+    index: uint256 = self.total_deposit_count + 2**POW_CONTRACT_MERKLE_TREE_DEPTH
+    msg_gwei_bytes8: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
     timestamp_bytes8: bytes[8] = slice(concat("", convert(block.timestamp, bytes32)), start=24, len=8)
     deposit_data: bytes[2064] = concat(deposit_params, msg_gwei_bytes8, timestamp_bytes8)
```
hwwhww: super nitpick: IMO this ordering is clearer when deserializing:
```suggestion
    deposit_data: bytes[2064] = concat(msg_gwei_bytes8, timestamp_bytes8, deposit_params)
```

specs/core/0_beacon-chain.md
```diff
@@ -615,6 +622,7 @@ def deposit(deposit_params: bytes[2048]):
 @constant
 def get_receipt_root() -> bytes32:
     return self.receipt_tree[1]
+
 ```
 
 The contract is at address `DEPOSIT_CONTRACT_ADDRESS`. When a user wishes to become a validator by moving their ETH from the 1.0 chain to the 2.0 chain, they should call the `deposit` function, sending along 32 ETH and providing as `deposit_params` a SimpleSerialize'd `DepositParams` object of the form:
```
hwwhww: 32 -> `DEPOSIT_SIZE`

```suggestion
The contract is at address `DEPOSIT_CONTRACT_ADDRESS`. When a user wishes to become a validator by moving their ETH from the 1.0 chain to the 2.0 chain, they should call the `deposit` function, sending along `DEPOSIT_SIZE` ETH and providing as `deposit_params` a SimpleSerialize'd `DepositParams` object of the form:
```

specs/core/0_beacon-chain.md
```diff
@@ -749,23 +758,32 @@ def add_validator(state: State,
...
+            return len(state.validators) - 1
+        else:
+            state.validators[index] = rec
+            return index
     else:
-        state.validators[index] = rec
+        index = validator_pubkeys.index(pubkey)
+        val = state.validators[index]
+        assert val.withdrawal_credentials == withdrawal_credentials
+        assert deposit_size >= MIN_TOPUP_SIZE
```
JustinDrake: I suggest the deposits contraints (`MIN_DEPOSIT`, `MAX_DEPOSIT`) should go in the Vyper deposit contract.

specs/core/0_beacon-chain.md
```diff
@@ -749,23 +761,32 @@ def add_validator(state: State,
...
+            exit_seq=0
+        )
+        index = min_empty_validator(state.validators)
+        if index is None:
+            state.validators.append(rec)
+            return len(state.validators) - 1
+        else:
+            state.validators[index] = rec
+            return index
     else:
```
djrtwo: Should probably make sure that the validator represented by the pub_key is not `WITHDRAWN`

vbuterin: Just fixed with https://github.com/ethereum/eth2.0-specs/pull/162/commits/9a20ecb9e46817bea17637193ca866d976ff952b

djrtwo: This now bypasses the deletion period.

The `deposit_data.msg_value == DEPOSIT_SIZE` conditional in `DEPOSIT_PROOF` validity also needs to be reworked

vbuterin: > This now bypasses the deletion period.

Aah, actually you're right, that last commit would have made it possible to use the same receipt to deposit twice. Just edited to fix.

There is no longer a `deposit_data.msg_value == DEPOSIT_SIZE` conditional in `DEPOSIT_PROOF`.

PR #166, opened 2018-11-25
remove `GENESIS_TIME` from constants

djrtwo:
This constant has been superseded by `BeaconState.genesis_time`


JustinDrake:
:+1: Also need to change "minimum timestamp as computed by `GENESIS_TIME` + block.slot * SLOT_DURATION"

PR #170, opened 2018-11-25
change SpecialRecord.kind to uint64

djrtwo:


hwwhww:
Is it part of Justin's point 3 (Type homogenisation)? If so, should `status` code also be changed to `uint64`?

djrtwo:
`Validator.status`?
I think `ValidatorRecord` is one of the places we are trying to restrict type sizes where reasonable

JustinDrake:
I'd change `status` to `uint64` 😂

With 1,000,000 validators that's nominally 7MB of extra storage. Implementers can compress this by storing `status` as a `uint8` in their database implementations if they care about saving a few megabytes.

vbuterin:
> I'd change status to uint64 

Seems fine to me; we lose 7 MB but we gain a hard norm that we use uint64 everywhere with the sole exception of validator indices, which seems reasonable enough.

PR #175, opened 2018-11-26
make explicit pending_attestations update

arnetheduck:
https://gitter.im/ethereum/sharding?at=5bf9ca28b6c0701052812041


hwwhww:
I think it will be overlapped by #149

arnetheduck:
right, missed that one.

PR #180, opened 2018-11-27
ssz: proofread + `uintN` notation

hwwhww:
1. Proofread work
2. `uint: 8/16/24/32/64/256` -> `uintN`

### DIFF
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -1,6 +1,6 @@
 # [WIP] SimpleSerialize (SSZ) Spec
 
-This is the **work in progress** document to describe `simpleserialize`, the
+This is the **work in progress** document to describe `SimpleSerialize`, the
 current selected serialization method for Ethereum 2.0 using the Beacon Chain.
 
 This document specifies the general information for serializing and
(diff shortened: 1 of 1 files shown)


hwwhww:
@vbuterin check with you, `CHUNK_SIZE` is set to `128` in SSZ spec while another `CHUNK_SIZE` constant is set to `256` in phase 1 spec, is that correct?

vbuterin:
> @vbuterin check with you, CHUNK_SIZE is set to 128 bytes in SSZ spec while another CHUNK_SIZE constant is set to 256 bytes in phase 1 spec, is that correct?

Yes, maybe rename one to SSZ_CHUNK_SIZE. Though honestly we should consider setting both to the same value; I'd be fine with 128.

specs/simple-serialize.md
```diff
@@ -53,13 +56,19 @@ overhead.
...
 ## Overview
 
 ### Serialize/Encode
 
-#### uint: 8/16/24/32/64/256
+#### uint
+
+| uint Type | Usage                                           |
+|:---------:|:------------------------------------------------|
+|  `uintN`  | Type of arbitrary `N` bits unsigned integer.    |
```
djrtwo: Not arbitrary. Needs to be `% 8`

PR #183, opened 2018-11-28
proofread and add ToC

hwwhww:
😵

generate with https://github.com/ekalinin/github-markdown-toc


specs/core/0_beacon-chain.md
```diff
@@ -1,18 +1,61 @@
 # Ethereum 2.0 Phase 0 -- The Beacon Chain
 
-###### tags: `spec`, `eth2.0`, `casper`, `sharding`, `beacon`
+**tags: `spec`, `eth2.0`, `casper`, `sharding`, `beacon`**
```
ChihChengLiang: The tags were originally written for better sorting experience in Hackmd, they can now be moved to Github repository tags, which would looks like the following example image.
screen_shot_2018-11-28_at_6_47_41_pm

PR #188, opened 2018-11-29
Add notation section

### LINKED ISSUE #187: Add a Notation section
Although the Python reference implementation is not supposed to be considered normative whenever it conflicts with the contents of the phase 0 spec, many of the algorithm definitions in the spec are presented via Python code.  I think it might be helpful for new or novice readers to have a brief explanation that such code is to be interpreted merely as definitions of algorithms rather than as code that must be used verbatim in implementations.

I will create a PR with my proposed text additions.

burnburn:
Fixes #187.

This is a proposal to add a brief Notation section at the top of the doc to help clarify for new/novice readers how to interpret the various code bits in the doc.


burnburn:
FYI, my force-push was merely a squashed rebase in which I signed my commit.  From now on my commits should all appear as Verified.

Issue #203, opened 2018-11-30
Add references section


burnburn:
I'd like to propose the explicit addition of a References section to the end of the document.  In-text links would then point to the specific reference within that section, which would then point to the external resource.

While it does mean that following an in-text reference would then require two clicks, moving all external references to one section has several benefits:
1. It provides a way to make clear, in one place, all references that are Normative, meaning that they must be read in order to implement the specification.
2. It provides a single place to check the external references to ensure that the reference links are not stale, either in terms of the links themselves no longer being valid, or in terms of the most recent / most applicable version being the one referenced.
3. It provides an opportunity to give more details (such as the document name, etc.) than would be appropriate in the main body of the document.

I will follow with a PR creating the References section, then several PRs suggesting how existing links should be moved to and described in the References section.  My goal was to prioritize simplicity of in-text reference over simplicity of the Reference section content while ensuring that full Markdown was usable in the References section.

djrtwo:
Closing as section was added. Feel free to continue to add references into the section on a continual basis

PR #206, opened 2018-11-30
First/sample reference

burnburn:
This PR takes the very first external reference in the doc and moves it to the References section to show how this will work. (See #203)  If people can live with this referencing mechanism/syntax, then I will go through the doc and propose Reference-section moves for each of the other external links in the doc.

We may want to change the expected content for the items in the References section over time to match/follow a particular style, but for now I just want to get something in there that is lightweight enough that people will use it.

Note that there are two blank spaces after the first of the two text lines that make up a reference (not visible in GitHub).  This causes a line break to be inserted between the two lines.  In practice I expect most people will copy/paste to insert new references, which will automatically pick up the two blank spaces.  Welcome to the joys of (any kind of) styling in Markdown.


djrtwo:
This looks good to me for a start

Issue #225, opened 2018-12-04
Validator status transition issues


hwwhww:
### Issues
- [x] 1. **The current spec doesn't describe how `PENDING_EXIT` transits to `EXITED_WITHOUT_PENALTY`**. A draft diagram:
    ![validator_state_transition](https://user-images.githubusercontent.com/9263930/49433564-575fb500-f7ed-11e8-90f5-3a8b9d4d3936.png)
    (Please let me know if anything incorrect or unclear.)

- [x] 2. **Potential `min_empty_validator_index()` issue**: [follow up this comment](https://github.com/ethereum/eth2.0-specs/pull/185#discussion_r237649259). Agreed with @djrtwo. IMO It doesn't make sense to allow the validators who *have sent voluntary exit/withdraw message* to make a top-up deposit.
    **edited**: just noticed the second case: involuntary `ACTIVE` -> `PENDING_EXIT` when `v.balance < MIN_BALANCE`.

- [ ] 3. **`COLLECTIVE_PENALTY_CALCULATION_PERIOD` and "withdrawal period"**: In the codes, we use `COLLECTIVE_PENALTY_CALCULATION_PERIOD` while it is described as "withdrawal period" everywhere.

- [x] 4. [`get_active_validator_indices` returns both `ACTIVE` and `PENDING_EXIT` validators](https://github.com/ethereum/eth2.0-specs/blob/c2227a59de89a8b233974f58d55e8a4401568bce/specs/core/0_beacon-chain.md#get_active_validator_indices).
    ```python
    def get_active_validator_indices(validators: [ValidatorRecord]) -> List[int]:
        """
        Gets indices of active validators from ``validators``.
        """
        return [i for i, v in enumerate(validators) if v.status in [ACTIVE, PENDING_EXIT]]
    ```

    Maybe we should rename `ACTIVE` status to another name?

#### For issue 1
1.  @djrtwo suggested: https://github.com/ethereum/eth2.0-specs/pull/185/files#r237660001
`get_changed_validators()` needs to check `if state == PENDING_EXIT` and then immediately set `state = EXITED_WITHOUT_PENALTY`. 
2. [Add the stub of the actual withdraw](https://github.com/ethereum/eth2.0-specs/pull/185/files#diff-51a43328a58414e132a744f3771f018cL1309). For testing withdraw, maybe we can actually set `v.balance=0` and note that the balance will be transferred to shard chain in phase 1.

---

potentially related to #198?

[...]

Issue #198, opened 2018-11-30
Deposit issues


djrtwo:
# Issue
There are some known issues in the current deposit contract and deposit processing. This issue will probably be resolved via a couple of PRs

Known issues:

- [x]  by removing `assert deposit_size == DEPOSIT_SIZE` for the creation of a validator, a user can now create a validator with as little as 1 ETH. This validator would be inducted and immediately be kicked out of the validator set.
- [x] Beyond that, this allows validators to enter the validator set (and begin participating) with as little as 16 ETH. The deposit mechanism should enforce 32 ETH as initial deposit to prevent validators entering with low and risky deposits.
- [x] The deposit contract is incorrectly using `full_deposit_count` to create and broadcast the leaves of the merkle tree. If a `< 32 ETH` deposit is made, it does not update `full_deposit_count` but update the merkle tree, overwriting the previously written leaf.  We need to maintain deposit_count separately from full_deposit_count. deposit_count should be incremented upon every successful deposit, should be used to construct the index into the merkle tree, and should be broadcast in the HashChainValue instead of full_deposit_count.

vbuterin:
> by removing assert deposit_size == DEPOSIT_SIZE for the creation of a validator, a user can now create a validator with as little as 1 ETH. This validator would be inducted and immediately be kicked out of the validator set.

Is this *that bad*? Their ETH is just stuck and they'll get it out in phase 2

>Beyond that, this allows validators to enter the validator set (and begin participating) with as little as 16 ETH. The deposit mechanism should enforce 32 ETH as initial deposit to prevent validators entering with low and risky deposits.

I'm inclined to agree. I don't think the deposit contract needs to be too paternalistic; it's already (unavoidably) possible to deposit with a bad randao preimage, for example; we could just make deposits fail if they have less than 32 ETH.

> The deposit contract is incorrectly using full_deposit_count to create and broadcast the leaves of the merkle tree. If a < 32 ETH deposit is made, it does not update full_deposit_count but update the merkle tree, overwriting the previously written leaf. We need to maintain deposit_count separately from full_deposit_count. deposit_count should be incremented upon every successful deposit, should be used to construct the index into the merkle tree, and should be broadcast in the HashChainValue instead of full_deposit_count.

Agree!

djrtwo:
> we could just make deposits fail if they have less than 32 ETH.

I'm inclined to go this route.

djrtwo:
addressed in #228

PR #228, opened 2018-12-04
Fix deposit logic

djrtwo:
addresses
* all points in #198 
* 1, 2, and 4 of #225

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -192,7 +192,7 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted 
 | - | - |
 | `PENDING_ACTIVATION` | `0` |
 | `ACTIVE` | `1` |
-| `PENDING_EXIT` | `2` |
+| `ACTIVE_PENDING_EXIT` | `2` |
 | `EXITED_WITHOUT_PENALTY` | `3` |
 | `EXITED_WITH_PENALTY` | `4` |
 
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1023,7 +1027,7 @@ First, some helper functions:
 ```python
 def min_empty_validator_index(validators: List[ValidatorRecord], current_slot: int) -> int:
     for i, v in enumerate(validators):
-        if v.balance == 0 and v.latest_status_change_slot + ZERO_BALANCE_VALIDATOR_TTL <= current_slot:
+        if v.balance == WITHDRAWN and v.latest_status_change_slot + ZERO_BALANCE_VALIDATOR_TTL <= current_slot:
```
JustinDrake: I think you meant `v.status == WITHDRAWN` but:

1) We don't have a `WITHDRAWN` status
2) The `EXITED_WITHOUT_PENALTY` and `EXITED_WITH_PENALTY` are the "withdrawable" statuses, so that combined with `v.balance == 0` is the safest/cleanest IMO. (It's safest because it protects against topups being processed after an exit, which can happen because there's a fundamental race condition.)

djrtwo: hmmm... I'll leave as is for now.
I think that we need to add `WITHDRAWN` back in to remedy some of the Penalty problems, but I'll address that in another PR.

Are you suggesting that the following should be able to occur?
* validator withdraws
* 1 ETH deposit is made to this withdrawn validator
* validator can withdraw the 1 eth balance

djrtwo: I'm thinking we are going to need some sort of distinction between `EXITED_WITH_PENALTY` and actually `EXITED_AND_PENALIZED` once the penalty is incurred. 

Right now penalties are broken

JustinDrake: > hmmm... I'll leave as is for now

If so you probably want to define `WITHDRAWN` in the constants and change `v.balance` to `v.status`.

> Are you suggesting that the following should be able to occur?

Yes. Checking `v.balance == 0` is asynchronously safe, but `v.status == WITHDRAWN` is not (because of the race condition).

PR #229, opened 2018-12-04
change `INITIAL_SLOT_NUMBER` to avoid underflow checks

djrtwo:
Changed `INITIAL_SLOT_NUMBER` to be define as `3 * EPOCH_LENGTH` to avoid a host of underflow checks/errors that needed to be fixed in state recalculations.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -158,8 +158,10 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted 
 | Name | Value |
 | - | - |
 | `INITIAL_FORK_VERSION` | `0` |
-| `INITIAL_SLOT_NUMBER` | `0` |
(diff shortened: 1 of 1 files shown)


JustinDrake:
Paging @paulhauner and @arnetheduck

djrtwo:
The genesis block at `INITIAL_SLOT_NUMBER` is finalized. A client can disregard any blocks that come prior to or are in forks that do not include the last finalized block. 

Because of this, I don't think we need to include that extra validity condition

hwwhww:
> A client can disregard any blocks that come prior to or are in forks that do not include the last finalized block.

Yeah, that works (covered `block.slot >= INITIAL_SLOT_NUMBER` condition).

I think "a client can disregard any blocks that come prior to or are in forks that do not include the last finalized block" is similar to what I mean by "the client side still has to implement the type check upon receiving the data"? It's implicit here.

PR #234, opened 2018-12-04
Move to a per-slot state transition function

JustinDrake:
Migrate to a per-block state transition function:

* Add `slot` to the state
* Introduce skip blocks if there is no block from the proposer
* Deemphasize `latest_state_recalculation_slot` (can probably be removed altogether)
* Per-epoch processing simply happens at exact epoch boundaries, without queuing
* Rely on `state.slot` instead of `current_state` or `block.state` where appropriate
* Initial cleanups in the "Per-slot processing" and "Per-epoch processing"
     * Review the table of content titles
     * Cleaner sub-sections
     * Misc cleanups

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1222,8 +1222,9 @@ If there is a block from the proposer for `state.slot`, we process that incoming
 If there is not a block from the proposer for `state.slot`, a "skip block" is inserted into the beacon chain:
 * Let `block` be a skip block defined by `get_skip_block(state.slot, parent, parent_hash)`.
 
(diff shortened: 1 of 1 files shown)


hwwhww:
Where do we set `state.slot`? I assume is at the beginning of each per-slot processing? In this scheme (different from #241):
1. At the beginning of processing genesis block, Set `state.slot` to `0`. 
2. At the beginning of processing `block_1` where `block.slot = 1`, Set `state.slot` to `1`.
3. No proposed block for slot 2 -> it's a skip block.
	1. At the end of `state.genesis_time + 2 * SLOT_DURATION`, the client processing the skip block `block_2`.
	2. Set `state.slot` to `2`.
4. At the beginning of processing `block_3` where `block.slot = 3`, Set `state.slot` to `5`.

JustinDrake:
Closing in favour of #241

Issue #246, opened 2018-12-06
New beacon block format


JustinDrake:
This is mostly a heads-up for implementers that we are moving to a new `BeaconBlock` format:

```
{
    'slot': 'uint64',
    'parent_root': 'hash32',
    'state_root': 'hash32',
    'deposit_root': 'hash32',
    'randao_reveal': 'hash32',
    'signature': ['uint384'],
    'body': {
        'proposer_slashing': [ProposerSlashingObject],
        'casper_slashing': [CasperSlashingObject],
        'attestations': [AttestationObject],
        'deposits': [DepositObject],
        'exits': [ExitObject],
    },
}
```

Several things are going on here:

- [x] Unify attestations with the other beacon transactions (currently known as "specials")
- [x] Stricter typing for beacon transactions
- [x] Merge the logic around checking maximum number of transactions per kind
- [ ] Remove `ancestor_hashes` from the blocks. A historical block accumulator will go in the state instead.
- [x] Order the beacon transaction kinds so as to minimise edge cases and attacks
- [x] Cleaner naming
- [x] Put all beacon transactions in a `body` for light-client friendliness
- [x] Clean conceptual separation of the "header" (the six fields from `slot` to `signature`) and the `body`

hwwhww:
Suggestion:
1. Add `'body_root': 'hash32'` in `BeaconHeader` fields, where `body_root` is the hash of corresponding `BeaconBody`.
2. Use the hash of `BeaconHeader` instead of the hash of `BeaconBlock` to represent the "Beacon Block Hash".

I feel that will be more light-client friendly and data storing friendly.

JustinDrake:
> I feel that will be more light-client friendly and data storing friendly.

Why? The tree hashing algorithm plus the `header` and `body` separation makes the data structure pretty light-client friendly.

hwwhww:
So that the light-client can only download the `header`s and the certain `body`s when they need?

JustinDrake:
> So that the light-client can only download the headers and the certain bodys when they need?

Light clients can download the `header`s together with `body` Merkle roots. That's enough to calculate `block_hash` and download specific bodies trustlessly.

vbuterin:
> Clean separation of header and body, somewhat more light-client friendly

I actually don't like this :cry: 

[...]

PR #252, opened 2018-12-07
new block format pt2 -- remove specials

djrtwo:
part 2 in a series of PRs addressing #246.

* removes `block.body.specials`
* move all specials to specific block objects contained in `block.body`
* Create section "Block objects" in per-slot processing for all these `body` object types


specs/core/0_beacon-chain.md
```diff
@@ -200,14 +200,15 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
 | `EXITED_WITHOUT_PENALTY` | `3` |
 | `EXITED_WITH_PENALTY` | `4` |
 
-### Special record types
+### Max transactions
```
vbuterin: Do we really want to call them "transactions"? I would be inclined to save that term for user-defined activity, and call them something more neutral, eg. "records".

specs/core/0_beacon-chain.md
```diff
@@ -317,27 +321,106 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
...
+        'deposit_parameters': DepositParametersRecord,
+        # Value in Gwei
+        'value': 'uint64',
+        # Timestamp from deposit contract
+        'timestamp': 'uint64',
+    },
+}
+```
+
+#### `DepositParametersRecord`
```
```suggestion
##### `DepositParametersRecord`
```

specs/core/0_beacon-chain.md
```diff
@@ -317,27 +285,130 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
...
+}
+```
+
+#### `BeaconBlock`
+
+```python
+{
+    'header': BeaconBlockHeader,
+    'body': BeaconBlockBody,
+}
```
vbuterin: Not sure I like this; I'd feel better with `body` being an element of the header.

Remember that at the SSZ tree hashing level, that would just mean that the body gets represented by a nice clean 32 byte hash.

djrtwo: Are you suggesting the following? Just doing away with `BeaconBlockHeader`? I like that.

```
#### BeaconBlock

{
    'slot': 'uint64',
    # Skip list of ancestor beacon block hashes
    # i'th item is the most recent ancestor whose slot is a multiple of 2**i for i = 0, ..., 31
    'ancestor_hashes': ['hash32'],
    'state_root': 'hash32',
    'randao_reveal': 'hash32',
    'candidate_pow_receipt_root': 'hash32',
    'signature': ['uint384'],
    'body': BeaconBlockBody,
}

File: specs/core/0_beacon-chain.md [hunk 4]
```diff
@@ -1262,61 +1265,63 @@ For each `attestation` in `block.body.attestations`:
...
-#### `VOLUNTARY_EXIT`
+For each `proposer_slashing` in `block.body.proposer_slashings`:
 
-* Let `validator = state.validator_registry[validator_index]`.
-* Verify that `BLSVerify(pubkey=validator.pubkey, msg=ZERO_HASH, sig=signature, domain=get_domain(state.fork_data, slot, DOMAIN_EXIT))`.
-* Verify that `validator.status == ACTIVE`.
-* Verify that `state.slot >= slot`.
-* Verify that `state.slot >= validator.latest_status_change_slot + SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD`.
-* Run `exit_validator(validator_index, state, penalize=False, current_slot=state.slot)`.
+* Let `proposer = state.validator_registry[proposer_slashing.proposer_index]`
```
JustinDrake: Missing period at the end

specs/core/0_beacon-chain.md
```diff
@@ -226,54 +231,58 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
...
+    # Second proposal signature
+    'proposal_signature_2': '[uint384]',
 }
 ```
 
-### Beacon chain blocks
+#### Casper slashings
 
-#### `BeaconBlock`
+##### `CasperSlashingRecord`
```
JustinDrake: I would agree with Vitalik that we should keep `Record` for objects stored in the state. What about `CasperSlashingObject` or `CasperSlashingSpecial`?

specs/core/0_beacon-chain.md
```diff
@@ -1262,61 +1265,63 @@ For each `attestation` in `block.body.attestations`:
...
-* Verify that `BLSVerify(pubkey=validator.pubkey, msg=ZERO_HASH, sig=signature, domain=get_domain(state.fork_data, slot, DOMAIN_EXIT))`.
-* Verify that `validator.status == ACTIVE`.
-* Verify that `state.slot >= slot`.
-* Verify that `state.slot >= validator.latest_status_change_slot + SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD`.
-* Run `exit_validator(validator_index, state, penalize=False, current_slot=state.slot)`.
+* Let `proposer = state.validator_registry[proposer_slashing.proposer_index]`
+* Verify that `BLSVerify(pubkey=proposer.pubkey, msg=hash(proposer_slashing.proposal_data_1), sig=proposer_slashing.proposal_signature_1, domain=get_domain(state.fork_data, proposer_slashing.proposal_data_1.slot, DOMAIN_PROPOSAL))`.
+* Verify that `BLSVerify(pubkey=proposer.pubkey, msg=hash(proposer_slashing.proposal_data_2), sig=proposer_slashing.proposal_signature_2, domain=get_domain(state.fork_data, proposer_slashing.proposal_data_2.slot, DOMAIN_PROPOSAL))`.
+* Verify that `proposer_slashing.proposal_data_1 != proposer_slashing.proposal_data_2`.
+* Verify that `proposer_slashing.proposal_data_1.slot == proposer_slashing.proposal_data_2.slot`.
```
JustinDrake: I think we should also check that `proposer_slashing.proposal_data_1.shard == proposer_slashing.proposal_data_2.shard`.

specs/core/0_beacon-chain.md
```diff
@@ -1262,61 +1265,63 @@ For each `attestation` in `block.body.attestations`:
...
-* Let `validator = state.validator_registry[validator_index]`.
-* Verify that `BLSVerify(pubkey=validator.pubkey, msg=ZERO_HASH, sig=signature, domain=get_domain(state.fork_data, slot, DOMAIN_EXIT))`.
-* Verify that `validator.status == ACTIVE`.
-* Verify that `state.slot >= slot`.
-* Verify that `state.slot >= validator.latest_status_change_slot + SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD`.
-* Run `exit_validator(validator_index, state, penalize=False, current_slot=state.slot)`.
+* Let `proposer = state.validator_registry[proposer_slashing.proposer_index]`
+* Verify that `BLSVerify(pubkey=proposer.pubkey, msg=hash(proposer_slashing.proposal_data_1), sig=proposer_slashing.proposal_signature_1, domain=get_domain(state.fork_data, proposer_slashing.proposal_data_1.slot, DOMAIN_PROPOSAL))`.
+* Verify that `BLSVerify(pubkey=proposer.pubkey, msg=hash(proposer_slashing.proposal_data_2), sig=proposer_slashing.proposal_signature_2, domain=get_domain(state.fork_data, proposer_slashing.proposal_data_2.slot, DOMAIN_PROPOSAL))`.
+* Verify that `proposer_slashing.proposal_data_1 != proposer_slashing.proposal_data_2`.
```
JustinDrake: Suggest making this more explicit: `proposer_slashing.proposal_data_1.block_hash != proposer_slashing. proposal_data_2.block_hash`.

specs/core/0_beacon-chain.md
```diff
@@ -1262,61 +1265,63 @@ For each `attestation` in `block.body.attestations`:
 * If `block.candidate_pow_receipt_root` is `x.candidate_pow_receipt_root` for some `x` in `state.candidate_pow_receipt_roots`, set `x.votes += 1`.
 * Otherwise, append to `state.candidate_pow_receipt_roots` a new `CandidatePoWReceiptRootRecord(candidate_pow_receipt_root=block.candidate_pow_receipt_root, votes=1)`.
 
-### Special objects
-
-* Verify that the quantity of each type of object in `block.body.specials` is less than or equal to its maximum (see table at the top).
-* Verify that objects are sorted in order of `kind`. That is, `block.body.specials[i+1].kind >= block.body.specials[i].kind` for `0 <= i < len(block.body.specials-1)`.
+### Operations
```
JustinDrake: "Operations" feels a bit awkward. What about "beacon transaction"?

djrtwo: I'm not totally against `Beacon transaction`, but @vbuterin wanted to reserve that term exclusively for user level activity.

I think operations is better here than "special" or "objects".

The following seems natural to me
- Verify the block header
- Verify and perform block operations
- Transit epoch if necessary

specs/core/0_beacon-chain.md
djrtwo: In general object is implicit on all objects... but due to the conflicting debate on naming these things, I'll go with object for now.

djrtwo: I actually just removed the suffix from the objects. I think it is reasonable.

djrtwo:
* stuck with "Operations" to describe handling the `block.body`
* removed `Record` from the suffix of these data structures
* addressed other PR comments
* handled merge conflict from master

Let me know thoughts/further review

djrtwo:
cc: @JustinDrake @hwwhww for another look

hwwhww:
Highlight @JustinDrake 's changelog:

### Casper cleanups and bug fixes

* (cleanup) `SpecialAttestationData` => `CasperVotes`
* (cleanup) `vote_1`, `vote_2` => `votes_1`, `votes_2`
* (bug fix) Place a bound on the number of votes with `MAX_CASPER_VOTES`
* (bug fix) Supply `state` when calling `verify_casper_votes`

specs/core/0_beacon-chain.md
```diff
@@ -17,19 +17,28 @@
...
         - [Reward and penalty quotients](#reward-and-penalty-quotients)
         - [Status codes](#status-codes)
-        - [Special record types](#special-record-types)
+        - [Max operations per block](#max-operations-per-block)
         - [Validator registry delta flags](#validator-registry-delta-flags)
         - [Signature domains](#signature-domains)
     - [Data structures](#data-structures)
-        - [Deposits](#deposits)
-            - [`DepositParametersRecord`](#depositparametersrecord)
+        - [Beacon chain operations](#beacon-chain-operations)
```
hwwhww: Not sure if the following categories are very accurate.
1. `AttestationData` is also used in `CasperVotes`.
2. If we put all `*Data` class to here, Why `ProposalSignedData` is still in `ProposalSignedData` section?

specs/core/0_beacon-chain.md
```diff
@@ -226,54 +232,58 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
...
-    'body': BeaconBlockBody,
+    # First batch of votes
+    'votes_1': CasperVotes,
+    # Second batch of votes
+    'votes_2': CasperVotes,
 }
 ```
 
-#### `BeaconBlockBody`
+##### `CasperVotes`
```
hwwhww: IMO plural `Votes` is confusing sometimes. Per @djrtwo: `SlashableAttestationData`? Or something like `VoteData`?

specs/core/0_beacon-chain.md
```diff
@@ -311,27 +321,101 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
...
 
 ```python
 {
-    # Slot number
+    # Receipt Merkle branch
+    'merkle_branch': '[hash32]',
+    # Merkle tree index
+    'merkle_tree_index': 'uint64',
+    # Deposit data
+    'deposit_data': {
```
hwwhww: Whoops, I didn't notice this before. Maybe move it to a new structure? (In other PRs)

specs/core/0_beacon-chain.md
```diff
@@ -311,27 +321,101 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
...
+        'deposit_parameters': DepositParameters,
+        # Value in Gwei
+        'value': 'uint64',
+        # Timestamp from deposit contract
+        'timestamp': 'uint64',
+    },
+}
+```
+
+##### `DepositParameters`
```
hwwhww: Maybe `DepositInput`?

specs/core/0_beacon-chain.md
djrtwo: yeah, separate pr

specs/core/0_beacon-chain.md
djrtwo: This mirrors the input arg in the pow contract
```python
def deposit(deposit_parameters: bytes[2048])
```

Going to leave as is. We can change both in a separate PR if you want

PR #253, opened 2018-12-07
Bug fixes and cleanups around exit_validator

JustinDrake:
Cleanups

* Use `state.slot` over `current_slot`.
* Specify `new_status`
* Segregate ejection processing

Bug fixes:

* Put ejections at the end of the per-slot processing (TODO: per-slot processing should come *after* per-epoch processing)
* Immediately exit ejections
* Include `ACTIVE_PENDING_EXIT` as an active state
* Renamed two `State` to `BeaconState`

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -55,6 +55,7 @@
         - [Beacon chain fork choice rule](#beacon-chain-fork-choice-rule)
     - [Beacon chain state transition function](#beacon-chain-state-transition-function)
         - [Helper functions](#helper-functions)
+            - [`is_active_validator`](#is_active_validator)
             - [`get_active_validator_indices`](#get_active_validator_indices)
             - [`shuffle`](#shuffle)
             - [`split`](#split)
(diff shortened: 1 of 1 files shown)


JustinDrake:
I've resolved conflicts with `master`.

JustinDrake:
@djrtwo We also need to:

1) Replace the code under `# Exit validator` with a call to `exit_validator`
2) Only remove validator from persistent committees and update `validator_registry_delta_chain_tip` if there's an actual exit event (i.e. `EXITED_WITH_PENALTY` or `EXITED_WITHOUT_PENALTY`).
3) Abstract away all the grunt work around status changes (e.g. regarding `validator_registry_delta_chain_tip`, `status`, `latest_status_change_slot`) into an `update_validator_status` function.

Feel free to merge this one first to avoid a huge PR.

specs/core/0_beacon-chain.md
```diff
@@ -701,14 +703,23 @@ The per-slot transitions generally focus on verifying aggregate signatures and s
 
 Note: The definitions below are for specification purposes and are not necessarily optimal implementations.
 
+#### `is_active_validator`
+ ```python
+def is_active_validator(validator: ValidatorRecord) -> bool:
```
djrtwo: :)

and to think, we almost got rid of `get_active_validator_indices` long ago!

Issue #239, opened 2018-12-05
initial assignment of `state.persistent_committees`


djrtwo:
# Issue
We are currently assigning `state.persistent_committees` as a shuffling of `ValidatorRecord`s rather than just `validator_indices`

```python
persistent_committees=split(shuffle(initial_validator_registry, ZERO_HASH), SHARD_COUNT)
```

# Proposed solution
Something like the following
```python
persistent_committees=split(
    shuffle(
        [validator_index for validator_index in range(len(initial_validator_registry))],
        ZERO_HASH
    ),
    SHARD_COUNT
)
```

djrtwo:
ha! I'll take that route and only make it verbose if I get any questions/complaints 😼

djrtwo:
that's quite a one-liner
```
persistent_committees=split(shuffle(list(range(len(initial_validator_registry))), ZERO_HASH), SHARD_COUNT)
```

hwwhww:
Also more succinct != more readable. 👻

PR #258, opened 2018-12-07
fix initial initial persistent_committees

djrtwo:
addresses #239

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1013,6 +1013,9 @@ def on_startup(initial_validator_entries: List[Any],
 
     # Setup state
     initial_shuffling = get_new_shuffling(ZERO_HASH, initial_validator_registry, 0)
+    validator_indices = list(range(len(initial_validator_registry)))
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1013,6 +1013,9 @@ def on_startup(initial_validator_entries: List[Any],
 
     # Setup state
     initial_shuffling = get_new_shuffling(ZERO_HASH, initial_validator_registry, 0)
+    validator_indices = list(range(len(initial_validator_registry)))
```
JustinDrake: I'm not sure this quite works because the `initial_validator_registry` can have validators pending activation (because their balance is <32 ETH).

djrtwo: Ah, your right. Need to cherry pick the active indices

vbuterin: I think `get_active_validator_indices(initial_validator_registry)` is what you're looking for?

djrtwo:
Okay, fixed. requesting final review.

PR #263, opened 2018-12-08
Make G1/G2 points non-malleable

JustinDrake:


JustinDrake:
Close in favour of #262.

PR #269, opened 2018-12-10
Moved ancestor hashes to state, changed to batched accumulator

vbuterin:
Added two new objects:

* `ancestor_interval_merkle_roots`: Merkle roots of groups of 2**16 blocks
* `next_interval_merkle_partial`: data used to build up the next Merkle root

Updates these per-slot as needed.


specs/core/0_beacon-chain.md
```diff
@@ -1075,6 +1091,8 @@ def on_startup(initial_validator_entries: List[Any],
         latest_block_hashes=[ZERO_HASH for _ in range(EPOCH_LENGTH * 2)],
         latest_penalized_exit_balances=[],
         latest_attestations=[],
+        next_interval_merkle_partial=[ZERO_HASH for _ in range(ANCESTOR_HASH_DEPTH)],
+        ancestor_interval_merkle_roots=[ZERO_HASH for _ in range(INITIAL_SLOT_NUMBER // 2**ANCESTOR_HASH_DEPTH)]
```
vbuterin: A possible alternative is to use `next_interval_merkle_partial` and `ancestor_interval_merkle_roots` as constructed by this formula:

```python
next_interval_merkle_partial, ancestor_interval_merkle_roots = [ZERO_HASH for _ in range(ANCESTOR_HASH_DEPTH)], []
for i in range(INITIAL_SLOT_NUMBER):
    next_interval_merkle_partial, ancestor_interval_merkle_roots = update_ancestor_accumulator(next_interval_merkle_partial, ancestor_interval_merkle_roots, i, ZERO_HASH)
```

(Note that for values above 2**ANCESTOR_HASH_DEPTH the computation can be short-circuited, but this would go in an implementation not the spec)

vbuterin:
> Suggested changes (most discussed and agreed with @vbuterin):

Done!

Issue #274, opened 2018-12-10
Fix and cleanup validator status updates


JustinDrake:
1) Replace the code under `# Exit validator` with a call to `exit_validator`.
2) Only remove a validator from the persistent committees and update `validator_registry_delta_chain_tip` if there's an actual exit event (i.e. `EXITED_WITH_PENALTY` or `EXITED_WITHOUT_PENALTY`).
3) Abstract away the grunt work around status changes (e.g. regarding `validator_registry_delta_chain_tip`, `status`, `latest_status_change_slot`) into an `update_validator_status` function, and replace the code under `# Activate validator` with that

Issue #268, opened 2018-12-09
Include initial validator registry in validator_registry_delta_chain_tip


JustinDrake:
The `validator_registry_delta_chain_tip` is currently initialised to `ZERO_HASH` but should take into the steps to build `initial_validator_registry`.

djrtwo:
This was addressed in #280

PR #280, opened 2018-12-10
validator status cleanup

djrtwo:
addresses #274 and #268 

* Setup initial validators via `process_deposit` and normal activation to better reuse code and ensure that the `validator_registry_delta_chain` is built
* handle all validator status updates via `update_validator_status`
* `exit_validator` is no longer used for changing `ACTIVE` -> `ACTIVE_PENDING_EXIT`
* `exit_validator` only removes validator from `persistent_committees` and updates `validator_registry_delta_chain` if not previously already exited
    * _note_: This protects against doing repeated actions when slashing a validator that has already exited to status `EXIT_WITHOUT_PENALTY

Not addressed but worth noting here:
* There was language around the definition of `process_deposit` that said deposits from 1.0 should be processed in order as they were submit in the 1.0 chain, but there is no mechanism in "per-slot - deposits" to ensure this validity when processing a block. This needs to be considered more and handled in another PR. the `Deposit` object might need a `source` field (to distinguish 1.0 vs 2.0 deposits) and the `BeaconState` might need to remember the last deposit hash processed from the 1.0 chain.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -71,9 +71,11 @@
             - [`bytes1`, `bytes2`, ...](#bytes1-bytes2-)
             - [`get_effective_balance`](#get_effective_balance)
             - [`get_new_validator_registry_delta_chain_tip`](#get_new_validator_registry_delta_chain_tip)
+            - [`get_fork_version`](#get_fork_version)
             - [`get_domain`](#get_domain)
             - [`verify_casper_votes`](#verify_casper_votes)
             - [`integer_squareroot`](#integer_squareroot)
+            - [`BLSVerify`](#blsverify)
         - [On startup](#on-startup)
         - [Routine for activating a validator](#routine-for-activating-a-validator)
         - [Routine for exiting a validator](#routine-for-exiting-a-validator)
@@ -959,6 +961,17 @@ def get_new_validator_registry_delta_chain_tip(current_validator_registry_delta_
     )
 ```
 
+#### `get_fork_version`
+
+```python
+def get_fork_version(fork_data: ForkData,
+                     slot: int) -> int:
+    if slot < fork_data.fork_slot:
+        return fork_data.pre_fork_version
+    else:
+        return fork_data.post_fork_version
+```
+
 #### `get_domain`
 
 ```python
@@ -997,6 +1010,10 @@ def integer_squareroot(n: int) -> int:
         y = (x + n // x) // 2
     return x
 ```
+
+#### `BLSVerify`
+
+`BLSVerify` is a function for verifying a BLS12-381 signature, defined in the [BLS12-381 spec](https://github.com/ethereum/eth2.0-specs/blob/master/specs/bls_verify.md).
 
 ### On startup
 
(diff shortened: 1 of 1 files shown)


djrtwo:
actually still have a couple of cleanups to do and will handle merge conflict @JustinDrake. I'll ping for review

djrtwo:
okay, now this is ready for review @JustinDrake @hwwhww

specs/core/0_beacon-chain.md
```diff
@@ -1145,85 +1155,89 @@ def get_new_validators(validators: List[ValidatorRecord],
...
+    )
 ```
 
-### Routine for exiting a validator
+```python
+def initiate_validator_exit(index: int,
+                            state: BeaconState) -> None:
+    validator = state.validator_registry[index]
+    validator.status = ACTIVE_PENDING_EXIT
+    validator.latest_status_change_slot = state.slot
```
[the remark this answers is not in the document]
djrtwo: There is already a "Verify that `validator.status == ACTIVE`" in the `EXIT` validity conditions. Still probably wouldn't hurt to add it here.

specs/core/0_beacon-chain.md
```diff
@@ -998,6 +1011,10 @@ def integer_squareroot(n: int) -> int:
     return x
 ```
 
+#### `BLSVerify`
+
+`BLSVerify` is a function for verifying a BLS12-381 signature, defined in the [BLS12-381 spec](https://github.com/ethereum/eth2.0-specs/blob/master/specs/bls_verify.md).
```
hwwhww: Ah, I thought `BLS Verification` spec is "Boneh-Lynn-Shacham signatures verification" because it's not talking about the details of the curve, no?

djrtwo: Well that doc is doing a BLS signature verification for a specific BLS12-381 curve

djrtwo: renamed spec ref to "BLS Verification spec"

specs/core/0_beacon-chain.md
```diff
@@ -1145,85 +1155,101 @@ def get_new_validators(validators: List[ValidatorRecord],
...
+The following are helpers and should only be called via `update_validator_status`:
+
+```python
+def activate_validator(state: BeaconState,
+                       index: int) -> None:
+    """
+    Activate the validator with the given ``index``.
+    Note that this function mutates ``state``.
+    """
+    if validator.status != PENDING_ACTIVATION:
```
arnetheduck: need some proof-reading here - validator does not yet exist ;)

specs/core/0_beacon-chain.md
```diff
@@ -1145,85 +1155,101 @@ def get_new_validators(validators: List[ValidatorRecord],
...
+        return
+
+    validator = state.validator_registry[index]
+    validator.status = ACTIVE
+    validator.latest_status_change_slot = state.slot
+    state.validator_registry_delta_chain_tip = get_new_validator_registry_delta_chain_tip(
+        validator_registry_delta_chain_tip=validator_registry_delta_chain_tip,
```
arnetheduck: likewise - where's the chain tip coming from?

Issue #243, opened 2018-12-06
Problem with previous epoch balance


mkalinin:
### Issue
`previous_epoch_boundary_attesting_balance` calculated for epoch with `N` boundary may not be equal to `this_epoch_boundary_attesting_balance` with `N - EPOCH_LENGTH` boundary. 

The former calculation uses validator set that has been affected by rewards and registry changes which have happened _after_ `this_epoch_boundary_attesting_balance` calculation took place.

As a result there are two different attesting balances for the same epoch, depending on slot.

### Proposed solution
It could probably be solved by storing original balance in the state. Something like
```
this_epoch_boundary_attesting_balance = sum([get_effective_balance(v) for v in this_epoch_boundary_attesters])
previous_epoch_boundary_attesting_balance = state.latest_attesting_balance

...

# At the end of epoch transition
state.latest_attesting_balance = this_epoch_boundary_attesting_balance
```

vbuterin:
The other solution is to mark this as an asterisk and explicitly not worry about it, because there's a pretty low bound on the extent to which validator balances can change epoch-by-epoch so it affects consensus safety by at most a fraction of a percentage point.

mkalinin:
Yep, actually, I was thinking about the same. Probably the biggest difference between balances would be in case of temporal eclipse when there've been only one block proposed during whole epoch and validators got penalized. But even in this case the difference would be negligible.

hwwhww:
@djrtwo @mkalinin 
Is this issue fixed?

PR #288, opened 2018-12-11
per-epoch Balance difference consensus safety note

djrtwo:
addresses #243


djrtwo:
eh, I hear you.
I just want to get it in there so it is not forgotten as we do more formal security analysis.

djrtwo:
How about this @JustinDrake?

JustinDrake:
I meant a cleaner solution, not a cleaner presentation 😂. I think it would simplify formal analysis if we didn't have this wrinkle.

vbuterin:
I think wrinkles are unavoidable. The reason is that it's already inevitably going to be the case that if two competing chains get finalized, the validators will have slightly different balances on them due to various rewards and penalties, so the fault tolerance will not be exactly 1/3. If we end up formally proving security, we'll have to use an argument involving bounding the per-epoch change in validator balances in any case.

PR #298, opened 2018-12-12
deposit contract: reduce on-chain computation

hwwhww:
There's a `2**32` computation in `deposit()` function. Reduce on-chain computation by defining the constant variable `TWO_TO_POWER_OF_TREE_DEPTH: constant(uint256) = 4294967296`.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -608,6 +608,7 @@ MAX_DEPOSIT: constant(uint256) = 32  # ETH
 GWEI_PER_ETH: constant(uint256) = 1000000000  # 10**9
 CHAIN_START_FULL_DEPOSIT_THRESHOLD: constant(uint256) = 16384  # 2**14
 DEPOSIT_CONTRACT_TREE_DEPTH: constant(uint256) = 32
+TWO_TO_POWER_OF_TREE_DEPTH: constant(uint256) = 4294967296  # 2**32
 SECONDS_PER_DAY: constant(uint256) = 86400
 
 Eth1Deposit: event({previous_receipt_root: bytes32, data: bytes[2064], deposit_count: uint256})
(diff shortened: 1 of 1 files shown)


JustinDrake:
If the compiler doesn't do this optimisation, then sure :)

Other possible changes:

* `msg.value / GWEI_PER_ETH` => `msg.value // GWEI_PER_ETH` (technically equivalent but clearer IMO when we work with integers exclusively)
* `# add deposit to merkle tree` => Capitalise first word
* Now that the Ethereum 2.0 chain votes on receipt roots, do we even need `Eth1Deposit` and `ChainStart`? Isn't it sufficient to have "virtual" logs?
* `timestamp_day_boundary` => rename to `genesis_time` (maybe add a comment that it's boundary is at 00:00 UTC)

djrtwo:
* A quick look through some docs and a quick test leads me to believe that `//` is not supported between two ints in vyper.
* Explicit logs are always going to be easier to pull data out of the state. I can light sync eth1.0 chain to follow the hash of the headers and just request the needed receipts and verify against the headers. I'm pretty sure if we don't do logs, we are putting full sync requirements on all validators as opposed to some light variant

JustinDrake:
> as opposed to some light variant

Right, we want light sync. What about something like `log.Eth1Deposit(self.receipt_tree[1], sha3(deposit_data))`? Would save us ~2kB of logging per deposit.

djrtwo:
and deposit data is passed to beacon chain network manually?
I'm not entirely opposed but want to think through how deposits are processed in the beacon chain a bit more (questions around required ordering, etc) before making this change. If we enforce deposit ordering on the beacon chain, then this hashing could cause deposits to hang until someone does a full sync to get the original data! (if the depositor didn't broadcast the data to the beacon chain network themself)

djrtwo:
@fubuloubu says "Compiler will intelligently reduce the constant" in the vyper gitter

so no need to do the manual optimization

JustinDrake:
> and deposit data is passed to beacon chain network manually?

Are we not doing this already with the `Deposit` and `DepositParameters` objects? https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#deposit

> If we enforce deposit ordering on the beacon chain

I think we definitely should enforce deposit ordering on the beacon chain :)

> this hashing could cause deposits to hang until someone does a full sync to get the original data

Hmm, I'm starting to dislike my suggestion to make the deposit data 2048 bytes for futureproofness. Maybe it suffices to create another deposit contract on Ethereum 1.0 if we want to add new deposit logic.

JustinDrake:
> 470,518 gas

That's pretty horrendous. My suggestion is:

* Revert back to having `deposit_parameters` be of minimal size.
* If we need different deposit logic on Ethereum 1.0, we simply instantiate a new contract (and fork the beacon chain, which we'd need to do anyway).
* Because the contract is specific to a deposit method it can do some validation of `withdrawal_credentials` with prefix byte equal to `0x00`.

NIC619:
We currently store the whole receipt tree on-chain and thus updating the receipt tree consumes huge amount of gas. We can instead only store the tree root and provide merkle proof each time we need to update one of the leaf.

PR #303, opened 2018-12-12
clarified language around start_up

GregTheGreek:
Renamed `on_startup` to `get_initial_beacon_state` and clarified language around the `get_initial_beacon_state ` functionality.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -85,8 +85,8 @@
     - [Per-slot processing](#per-slot-processing)
         - [Misc counters](#misc-counters)
         - [Block roots](#block-roots)
-    - [Per-block processing](#per-block-processing)  
-        - [Slot](#slot)  
+    - [Per-block processing](#per-block-processing)
+        - [Slot](#slot)
         - [Proposer signature](#proposer-signature)
         - [RANDAO](#randao)
         - [PoW receipt root](#pow-receipt-root)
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1060,10 +1060,10 @@ A valid block with slot `INITIAL_SLOT_NUMBER` (a "genesis block") has the follow
 }
 ```
 
-`STARTUP_STATE_ROOT` is the root of the initial state, computed by running the following code:
+`STARTUP_STATE_ROOT` (in the above "genesis block") is generated from the `get_initial_beacon_state` function below. When enough full deposits have been made to the deposit contract and the `ChainStart` log has been emitted, `get_initial_beacon_state` will execute to compute the `ssz_tree_hash` of `BeaconState`.
 
 ```python
-def on_startup(initial_validator_deposits: List[Deposit],
+def get_initial_beacon_state(initial_validator_deposits: List[Deposit],
                genesis_time: int,
```
hwwhww: indent
```suggestion
                             genesis_time: int,
```

specs/core/0_beacon-chain.md
```diff
@@ -1060,10 +1060,10 @@ A valid block with slot `INITIAL_SLOT_NUMBER` (a "genesis block") has the follow
...
                processed_pow_receipt_root: Hash32) -> BeaconState:
```
hwwhww: indent
```suggestion
                             processed_pow_receipt_root: Hash32) -> BeaconState:
```

specs/core/0_beacon-chain.md
JustinDrake: Done :)

PR #306, opened 2018-12-13
Fix `verify_slashable_vote_data`: use `get_domain`

hwwhww:


specs/core/0_beacon-chain.md
```diff
@@ -1023,7 +1023,11 @@ def verify_slashable_vote_data(state: BeaconState, vote_data: SlashableVoteData)
         pubkeys=pubs,
         messages=messages,
         signature=vote_data.aggregate_signature,
-        domain=DOMAIN_ATTESTATION,
+        domain=get_domain(
+            state.fork_data,
+            state.slot,
+            DOMAIN_ATTESTATION
+        ),
```
mkalinin: Looks like a redundant comma mark
```suggestion
        )
```

JustinDrake: We like redundant commas :) A comma should be added after `DOMAIN_ATTESTATION` for consistency.

mkalinin: cool 😀

hwwhww: @JustinDrake oh right. 👍 

@mkalinin we think the redundant dangling commas may reduce the size of the unrelated diff when new items are added.

e.g.,
```python
x = my_function(
    a=1,
    b=2,
    c=3
)
```

If we add a new paramaters `d`:
```python
x = my_function(
    a=1,
    b=2,
-   c=3,
+   c=3,
+   d=4
)
```

PR #310, opened 2018-12-13
Deposits: define `DepositData` and rename `DepositParameters` to `DepositInput`

hwwhww:
1. Add `DepositData`.
2. Fix `get_initial_beacon_state` a tiny bug by adding a comma.
3. **Rename** `DepositParameters` to `DepositInput`.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -33,7 +33,7 @@
             - [Deposits](#deposits)
                 - [`Deposit`](#deposit)
                 - [`DepositData`](#depositdata)
-                - [`DepositParameters`](#depositparameters)
+                - [`DepositInput`](#depositinput)
             - [Exits](#exits)
                 - [`Exit`](#exit)
         - [Beacon chain blocks](#beacon-chain-blocks)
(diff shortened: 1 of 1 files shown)


JustinDrake:
On second thought, is there a good reason for not merging `DepositData` and `DepositInput` into a single flat object? cc @djrtwo

djrtwo:
It is marginally beneficial that the log outputs `deposit_data` as a a properly formatted SSZ object corresponding to `DepositData`.  This is done primarily through requiring the input to be an SSZ object itself `DepositInput` and then concatenating it with the additional time and deposit gwei value fields.

This is nice because the deposit contract doesn't need to a priori know anything about the required input fields which could allow us to change them down the line without deploying a new pow deposit contract.

djrtwo:
You could flatten the object, but the depositor would need to know that they need to remove two fields from the specified SSZ  object. 

With the separate objects, it is clearer how to construct the deposit input data

JustinDrake:
> With the separate objects, it is clearer how to construct the deposit input data

Makes sense. Next question: Can `Deposit` and `DepositData` be flattened? That is, we define `Deposit` to be

```
{
    'merkle_branch': '[hash32]',
    'merkle_tree_index': 'uint64',
    'value': 'uint64',
    'timestamp': 'uint64',
    'deposit_input': DepositInput,
}
```

Issue #286, opened 2018-12-11
Use objects/tree-hashing where possible


djrtwo:
Replace the following snippets of code with an object definition and tree hashing:

```
    return hash(
        current_validator_registry_delta_chain_tip +
        bytes1(flag) +
        bytes3(index) +
        bytes48(pubkey)
    )
```

```
[SSZTreeHash(votes)+bytes1(0), SSZTreeHash(votes)+bytes1(1)
```

(Notice the missing `]` above.)

```
hash(bytes32(pubkey) + withdrawal_credentials + randao_commitment)
```

```
SSZTreeHash(attestation.data) + bytes1(0)
```

PR #315, opened 2018-12-13
More ssz objects pt 1

djrtwo:
addresses most of  #286 

* proof of custody data ssz object
* validator delta block ssz object
* _note_: didn't address the custody bit

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -61,6 +61,7 @@
     - [Beacon chain state transition function](#beacon-chain-state-transition-function)
         - [Helper functions](#helper-functions)
             - [`hash`](#hash)
+            - [`hash_tree_root`](#hash_tree_root)
             - [`is_active_validator`](#is_active_validator)
             - [`get_active_validator_indices`](#get_active_validator_indices)
             - [`shuffle`](#shuffle)
@@ -77,7 +78,6 @@
             - [`get_new_validator_registry_delta_chain_tip`](#get_new_validator_registry_delta_chain_tip)
             - [`get_fork_version`](#get_fork_version)
             - [`get_domain`](#get_domain)
-            - [`hash_tree_root`](#hash_tree_root)
             - [`verify_slashable_vote_data`](#verify_slashable_vote_data)
             - [`integer_squareroot`](#integer_squareroot)
             - [`bls_verify`](#bls_verify)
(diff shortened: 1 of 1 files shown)


djrtwo:
wasn't quite sure how I wanted to do the custody bit. putting this up for review and will do the custody bit in another PR

specs/core/0_beacon-chain.md
```diff
@@ -377,6 +379,16 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
 }
 ```
 
+#### `ProofOfPossessionData`
```
JustinDrake: I would simply reuse `DepositInput `with `proof_of_possession` set to `[0, 0]`.

specs/core/0_beacon-chain.md
```diff
@@ -967,17 +990,19 @@ def get_effective_balance(validator: ValidatorRecord) -> int:
 
 ```python
 def get_new_validator_registry_delta_chain_tip(current_validator_registry_delta_chain_tip: Hash32,
-                                               index: int,
+                                               validator_index: int,
                                                pubkey: int,
                                                flag: int) -> Hash32:
     """
-    Compute the next hash in the validator registry delta hash chain.
+    Compute the next root in the validator registry delta hash chain.
```
JustinDrake: delta hash chain => delta root chain

specs/core/0_beacon-chain.md
```diff
@@ -967,17 +990,19 @@ def get_effective_balance(validator: ValidatorRecord) -> int:
...
     """
-    return hash(
-        current_validator_registry_delta_chain_tip +
-        bytes1(flag) +
-        bytes3(index) +
-        bytes48(pubkey)
+    return hash_tree_root(
+        ValidatorRegistryDeltaBlock(
+            current_validator_registry_delta_chain_tip,
```
JustinDrake: `validator_registry_delta_chain_tip= current_validator_registry_delta_chain_tip,`

specs/core/0_beacon-chain.md
JustinDrake: While we're at it, maybe define the constant `EMPTY_SIGNATURE` to be `[0, 0]`. It's used three times.

specs/core/0_beacon-chain.md
```diff
@@ -967,17 +990,19 @@ def get_effective_balance(validator: ValidatorRecord) -> int:
 
 ```python
 def get_new_validator_registry_delta_chain_tip(current_validator_registry_delta_chain_tip: Hash32,
-                                               index: int,
+                                               validator_index: int,
```
hwwhww: Also, need to update the keyword argument where that this function is being called.
(in `def activate_validator()` and in `def exit_validator()`)

specs/core/0_beacon-chain.md
```diff
@@ -34,6 +34,7 @@
                 - [`Deposit`](#deposit)
                 - [`DepositData`](#depositdata)
                 - [`DepositInput`](#depositinput)
+                - [`ProofOfPossessionData`](#proofofpossessiondata)
```
JustinDrake: That line should go

djrtwo: agh!

PR #317, opened 2018-12-14
Put validator balances into a separate list in the state

vbuterin:


vbuterin:
One aesthetic change I think we should consider is switching the `get_effective_balance` function to just take the state as an argument instead of (as in this PR right now) the validator balance entry. I could see future changes that use parts of the validator registry entry as well as the balance entry to calculate the effective balance, so we may as well just make grab all the data it needs from the state.

specs/core/0_beacon-chain.md
```diff
@@ -956,11 +955,11 @@ def get_attestation_participants(state: BeaconState,
 #### `get_effective_balance`
 
 ```python
-def get_effective_balance(validator: ValidatorRecord) -> int:
+def get_effective_balance(balance: int) -> int:
```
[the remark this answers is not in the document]
JustinDrake: What do you think of making `state` a global variable, and encourage the helper functions to fish the relevant fields from the state as opposed to passing the information as inputs. That would remove a bunch of boilerplate.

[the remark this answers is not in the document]
djrtwo: I'm resistant to the suggestion. Not that people are just taking these functions directly and running them, but even implying that a client should keep a singly global `state` var around is just asking for hard to trace bugs in implementations.

I'm pro both of the following depending on the operation
- requesting updated fields
- passing in state and mutating it (but making it _clear_ state is mutated in the comment)

djrtwo:
@vbuterin I'm going to clean up per my recommendations and see if we like it.

vbuterin:
I updated `get_effective_balance` to take state and index and fixed a couple other issues.

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
vbuterin: I'm resistant as well. I like the idea of making data dependencies explicit.

specs/core/0_beacon-chain.md
```diff
@@ -1486,40 +1488,35 @@ All [validators](#dfn-validator):
 
   * Let `this_epoch_boundary_attestations = [a for a in this_epoch_attestations if a.data.epoch_boundary_root == get_block_root(state, state.slot-EPOCH_LENGTH) and a.justified_slot == state.justified_slot]`.
   * Let `this_epoch_boundary_attester_indices` be the union of the [validator](#dfn-validator) index sets given by `[get_attestation_participants(state, a.data, a.participation_bitfield) for a in this_epoch_boundary_attestations]`.
-  * Let `this_epoch_boundary_attesters = [state.validator_registry[i] for indices in this_epoch_boundary_attester_indices for i in indices]`.
-  * Let `this_epoch_boundary_attesting_balance = sum([get_effective_balance(v) for v in this_epoch_boundary_attesters])`.
+  * Let `this_epoch_boundary_attesting_balance = sum([get_effective_balance(state, i) for v in this_epoch_boundary_attester_indices])`.
```
djrtwo: fix bad var -- `v`

```suggestion
  * Let `this_epoch_boundary_attesting_balance = sum([get_effective_balance(state, i) for i in this_epoch_boundary_attester_indices])`.
```

specs/core/0_beacon-chain.md
```diff
@@ -1486,40 +1488,35 @@ All [validators](#dfn-validator):
...
-  * Let `previous_epoch_attesters = [state.validator_registry[i] for indices in previous_epoch_attester_indices for i in indices]`.
 * Validators targeting the previous justified hash:
   * Let `previous_epoch_justified_attestations = [a for a in this_epoch_attestations + previous_epoch_attestations if a.justified_slot == state.previous_justified_slot]`.
   * Let `previous_epoch_justified_attester_indices` be the union of the validator index sets given by `[get_attestation_participants(state, a.data, a.participation_bitfield) for a in previous_epoch_justified_attestations]`.
-  * Let `previous_epoch_justified_attesters = [state.validator_registry[i] for indices in previous_epoch_justified_attester_indices for i in indices]`.
-  * Let `previous_epoch_justified_attesting_balance = sum([get_effective_balance(v) for v in previous_epoch_justified_attesters])`.
+  * Let `previous_epoch_justified_attesting_balance = sum([get_effective_balance(state, i) for v in previous_epoch_justified_attester_indices])`.
```
djrtwo: fix bad var -- `v`
```suggestion
  * Let `previous_epoch_justified_attesting_balance = sum([get_effective_balance(state, i) for i in previous_epoch_justified_attester_indices])`.
```

specs/core/0_beacon-chain.md
```diff
@@ -1486,40 +1488,35 @@ All [validators](#dfn-validator):
...
 * Validators justifying the epoch boundary block at the start of the previous epoch:
   * Let `previous_epoch_boundary_attestations = [a for a in previous_epoch_justified_attestations if a.epoch_boundary_root == get_block_root(state, state.slot - 2 * EPOCH_LENGTH)]`.
   * Let `previous_epoch_boundary_attester_indices` be the union of the validator index sets given by `[get_attestation_participants(state, a.data, a.participation_bitfield) for a in previous_epoch_boundary_attestations]`.
-  * Let `previous_epoch_boundary_attesters = [state.validator_registry[i] for indices in previous_epoch_boundary_attester_indices for i in indices]`.
-  * Let `previous_epoch_boundary_attesting_balance = sum([get_effective_balance(v) for v in previous_epoch_boundary_attesters])`.
+  * Let `previous_epoch_boundary_attesting_balance = sum([get_effective_balance(state, i) for v in previous_epoch_boundary_attester_indices])`.
```
djrtwo: fix bad var -- `v`
```suggestion
  * Let `previous_epoch_boundary_attesting_balance = sum([get_effective_balance(state, i) for i in previous_epoch_boundary_attester_indices])`.
```

specs/core/0_beacon-chain.md
```diff
@@ -1486,40 +1488,35 @@ All [validators](#dfn-validator):
...
 * Validators attesting to the expected beacon chain head during the previous epoch:
   * Let `previous_epoch_head_attestations = [a for a in previous_epoch_attestations if a.beacon_block_root == get_block_root(state, a.slot)]`.
   * Let `previous_epoch_head_attester_indices` be the union of the validator index sets given by `[get_attestation_participants(state, a.data, a.participation_bitfield) for a in previous_epoch_head_attestations]`.
-  * Let `previous_epoch_head_attesters = [state.validator_registry[i] for indices in previous_epoch_head_attester_indices for i in indices]`.
-  * Let `previous_epoch_head_attesting_balance = sum([get_effective_balance(v) for v in previous_epoch_head_attesters])`.
+  * Let `previous_epoch_head_attesting_balance = sum([get_effective_balance(state, i) for v in previous_epoch_head_attester_indices])`.
```
djrtwo: fix bad var -- `v`
```suggestion
  * Let `previous_epoch_head_attesting_balance = sum([get_effective_balance(state, i) for i in previous_epoch_head_attester_indices])`.
```

specs/core/0_beacon-chain.md
djrtwo: This is still not resolved

Trenorenos (approved):

Issue #295, opened 2018-12-12
Keep latest 2^n RANDAO mixes in the state


JustinDrake:
Several reasons:

* Allows to change the RNG lookahead (currently at least 1 epoch) without messing with the `BeaconState` format.
* Allows to expose randomness as an opcode in the EVM 2.0 (similar to the `BLOCKHASH` opcode).
* Useful for the VDF logic. (Note: the plan is to include VDF placeholders in the `BeaconState`.)

djrtwo:
Also ensures that we have the seed of the current shuffling _in state_. Right now, we discard the current shuffling seed immediately after performing the shuffling. Not very friendly for reconstructing the shuffling outside the specific state transition.

JustinDrake:
> Not very friendly for reconstructing the shuffling outside the specific state transition.

To reconstruct the shuffling you also need the validator registry, and that will have moved on right?

djrtwo:
Not if you are two blocks into an epoch. The source of the current shuffling has been discarded. If say, I am a validator or user application querying a node through an RPC interface, it becomes opaque to me if the current shuffling has changed say due to a fork because the seed source of the shuffling is query-able (unless it is just stored locally).

Regardless, your solution keeps it around which I support

vbuterin:
Solved with https://github.com/ethereum/eth2.0-specs/pull/327

PR #327, opened 2018-12-15
Keep randao mixes in the state

vbuterin:
Response to #295 

Also a simplification, as two special cases get replaced: `state.randao_mix` -> `state.latest_randao_mixes[-1]`, and `state.next_seed` -> `state.latest_randao_mixes[-CYCLE_LENGTH-1]`.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1413,7 +1413,7 @@ Below are the processing steps that happen at every `block`.
 * Let `repeat_hash(x, n) = x if n == 0 else repeat_hash(hash(x), n-1)`.
 * Let `proposer = state.validator_registry[get_beacon_proposer_index(state, state.slot)]`.
 * Verify that `repeat_hash(block.randao_reveal, proposer.randao_layers) == proposer.randao_commitment`.
-* Set `state.latest_randao_mixes = state.latest_randao_mixes[1:] + [xor(state.latest_randao_mixes[-1], block.randao_reveal)]
+* Set `state.latest_randao_mixes[block.slot % LATEST_BLOCK_ROOTS_LENGTH] = xor(state.latest_randao_mixes[(block.slot - 1) % LATEST_BLOCK_ROOTS_LENGTH], block.randao_reveal)`
 * Set `proposer.randao_commitment = block.randao_reveal`.
 * Set `proposer.randao_layers = 0`.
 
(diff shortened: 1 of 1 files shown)


JustinDrake:
A couple suggestions:

* Only save the latest "boundary" `randao_mix`, i.e. when `slot % EPOCH_LENGTH == 0`. Indeed, we only use `randao_mix` as a seed (with or without VDFs) when it is at the epoch boundary. (I guess the counter-argument here is that maybe it's useful for the EVM 2.0 to have more RANDAO mixes.)
* Parametrise (with a new constant) the minimum seed lookahead. E.g. `MINIMUM_SEED_LOOKAHEAD = 1` (denominated in epochs).

vbuterin:
>  (I guess the counter-argument here is that maybe it's useful for the EVM 2.0 to have more RANDAO mixes.)

Yeah, I actually kind of want the extra RANDAO values around because later on down the line we can make a low-influence function out of them. 

Also unrelated, but I want to make one small change to both the block hash ancestry list and the randao list: instead of storing them as a queue, store them as a list where in slot `i` the `i % 8192`'th value gets modified. This greatly increases efficiency from a hash calculation perspective at basically no cost.

JustinDrake:
> make a low-influence function out of them

Makes sense. I guess that's an opportunity for the application layer.

> store them as a list where in slot `i` the `i % 8192`'th value gets modified

Oh, clever!

specs/core/0_beacon-chain.md
```diff
@@ -1749,15 +1747,14 @@ Also perform the following updates:
 
 * Set `state.validator_registry_latest_change_slot = state.slot`.
 * Set `state.shard_committees_at_slots[:EPOCH_LENGTH] = state.shard_committees_at_slots[EPOCH_LENGTH:]`.
-* Set `state.shard_committees_at_slots[EPOCH_LENGTH:] = get_new_shuffling(state.next_seed, state.validator_registry, next_start_shard)` where `next_start_shard = (state.shard_committees_at_slots[-1][-1].shard + 1) % SHARD_COUNT`.
-* Set `state.next_seed = state.randao_mix`.
+* Set `state.shard_committees_at_slots[EPOCH_LENGTH:] = get_new_shuffling(state.latest_randao_mixes[(block.slot - CYCLE_LENGTH) % LATEST_BLOCK_ROOTS_LENGTH], state.validator_registry, next_start_shard)` where `next_start_shard = (state.shard_committees_at_slots[-1][-1].shard + 1) % SHARD_COUNT`.
```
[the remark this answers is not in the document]
vbuterin: If we're going to do that, why tie `SEED_LOOK_AHEAD` to being a multiple of `CYCLE_LENGTH` (should be `EPOCH_LENGTH`; I fixed); it could also be half an epoch. So if the variables should be separated, just use `state.slot - SEED_LOOK_AHEAD`.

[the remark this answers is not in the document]
djrtwo: You're right.
Do you see this as a param likely to be tuned or do you want to keep it CYCLE_LENGTH for now??

specs/core/0_beacon-chain.md
```diff
@@ -166,6 +166,7 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
 | `BLS_WITHDRAWAL_PREFIX_BYTE` | `0x00` | - |
 | `MAX_CASPER_VOTES` | `2**10` (= 1,024) | votes |
 | `LATEST_BLOCK_ROOTS_LENGTH` | `2**13` (= 8,192) | block roots |
+| `LATEST_RANDAO_MIXES_LENGTH` | `2**13` (= 8,192) | block roots |
```
djrtwo: Unit is not "block roots". 
Use "randao mixes" or "seeds" if you keep a unit here.

specs/core/0_beacon-chain.md
```diff
@@ -1389,11 +1388,12 @@ Below are the processing steps that happen at every slot.
 
 * Set `state.slot += 1`.
 * Set `state.validator_registry[get_beacon_proposer_index(state, state.slot)].randao_layers += 1`.
+* Set `state.latest_randao_mixes = state.latest_randao_mixes[(block.slot - 1) % LATEST_RANDAO_MIXES_LENGTH]`
```
djrtwo: * This needs to be `state.slot`. You are not guaranteed to have a `block` here.
* missing index into `latest_randao_mixes` on the left side of the assignment

```suggestion
* Set `state.latest_randao_mixes[state.slot % LATEST_RANDAO_MIXES_LENTH] = state.latest_randao_mixes[(state.slot - 1) % LATEST_RANDAO_MIXES_LENGTH]`
```

specs/core/0_beacon-chain.md
```diff
@@ -1389,11 +1388,12 @@ Below are the processing steps that happen at every slot.
...
 
 ### Block roots
 
 * Let `previous_block_root` be the `hash_tree_root` of the previous beacon block processed in the chain.
-* Set `state.latest_block_roots = state.latest_block_roots[1:] + [previous_block_root]`.
+* Set `state.latest_block_roots[(block.slot - 1) % LATEST_BLOCK_ROOTS_LENGTH] = previous_block_root`.
```
djrtwo: This needs to be `state.slot`. You are not guaranteed to have a `block` here.

```suggestion
* Set `state.latest_block_roots[(state.slot - 1) % LATEST_BLOCK_ROOTS_LENGTH] = previous_block_root`.
```

djrtwo: `get_block_root` helper also needs to be fixed to reflect this change in `latest_block_roots`

specs/core/0_beacon-chain.md
```diff
@@ -1415,7 +1415,7 @@ Below are the processing steps that happen at every `block`.
 * Let `repeat_hash(x, n) = x if n == 0 else repeat_hash(hash(x), n-1)`.
 * Let `proposer = state.validator_registry[get_beacon_proposer_index(state, state.slot)]`.
 * Verify that `repeat_hash(block.randao_reveal, proposer.randao_layers) == proposer.randao_commitment`.
-* Set `state.randao_mix = xor(state.randao_mix, block.randao_reveal)`.
+* Set `state.latest_randao_mixes[block.slot % LATEST_RANDAO_MIXES_LENGTH] = xor(state.latest_randao_mixes[block.slot % LATEST_RANDAO_MIXES_LENGTH], block.randao_reveal)`
```
djrtwo: In per-block processing, after "Verify that `block.slot == state.slot`", we are using `state.slot` everywhere rather than `block.slot`.

```suggestion
* Set `state.latest_randao_mixes[state.slot % LATEST_RANDAO_MIXES_LENGTH] = xor(state.latest_randao_mixes[state.slot % LATEST_RANDAO_MIXES_LENGTH], block.randao_reveal)`
```

specs/core/0_beacon-chain.md
vbuterin: Done! (Also done the other changes below).

PR #331, opened 2018-12-17
Edit latest_block_roots in place instead of as a queue

vbuterin:
Faster editing that way; otherwise every block will require completely reconstructing a 8192-sized Merkle tree.


arnetheduck:
nice!

PR #333, opened 2018-12-17
Make Casper ref normative

burnburn:
This PR moves the Casper FFG ref to the Normative References section.

One possible question is whether the reference should specifically state that this is 'v3' of the document referenced.

FYI, the extra spaces added to the end of the python-poc reference are intended; they cause a newline that looks better when rendered.


djrtwo:
I would argue that the FFG paper is informative rather than normative. You can read to paper to help inform you about what is going on when you implement the finality portion of the spec, but you don't have to read it to implement it.

Let me know if my understanding of normative vs informative is correct here.

burnburn:
If the finality descriptions in the spec are sufficient for implementation, then you are correct that the FFG paper is informative only.  I will make that change.

PR #334, opened 2018-12-18
Bugfix around on startup functions

hwwhww:
Fix some typos and parameter name.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -982,7 +982,7 @@ def get_new_validator_registry_delta_chain_tip(current_validator_registry_delta_
     """
     return hash_tree_root(
         ValidatorRegistryDeltaBlock(
-            validator_registry_delta_chain_tip=current_validator_registry_delta_chain_tip,
+            latest_registry_delta_root=current_validator_registry_delta_chain_tip,
             validator_index=validator_index,
             pubkey=pubkey,
             flag=flag,
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1125,8 +1125,7 @@ A valid block with slot `INITIAL_SLOT_NUMBER` (a "genesis block") has the follow
 `STARTUP_STATE_ROOT` (in the above "genesis block") is generated from the `get_initial_beacon_state` function below. When enough full deposits have been made to the deposit contract and the `ChainStart` log has been emitted, `get_initial_beacon_state` will execute to compute the `hash_tree_root` of `BeaconState`.
 
 ```python
-def get_initial_beacon_state(initial_validator_registry: List[ValidatorRecord],
-                             initial_validator_deposits: List[Deposit],
+def get_initial_beacon_state(initial_validator_deposits: List[Deposit],
                              genesis_time: int,
                              processed_pow_receipt_root: Hash32) -> BeaconState:
     state = BeaconState(
@@ -1184,7 +1183,7 @@ def get_initial_beacon_state(initial_validator_registry: List[ValidatorRecord],
             update_validator_status(state, validator_index, ACTIVE)
 
     # set initial committee shuffling
-    initial_shuffling = get_new_shuffling(ZERO_HASH, initial_validator_registry, 0)
+    initial_shuffling = get_new_shuffling(ZERO_HASH, state.validator_registry, 0)
     state.shard_committees_at_slots = initial_shuffling + initial_shuffling
 
     # set initial persistent shuffling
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1125,7 +1125,8 @@ A valid block with slot `INITIAL_SLOT_NUMBER` (a "genesis block") has the follow
 `STARTUP_STATE_ROOT` (in the above "genesis block") is generated from the `get_initial_beacon_state` function below. When enough full deposits have been made to the deposit contract and the `ChainStart` log has been emitted, `get_initial_beacon_state` will execute to compute the `hash_tree_root` of `BeaconState`.
 
 ```python
-def get_initial_beacon_state(initial_validator_deposits: List[Deposit],
+def get_initial_beacon_state(initial_validator_registry: List[ValidatorRecord],
+                             initial_validator_deposits: List[Deposit],
```
djrtwo: We don't need `initial_validator_registry` as an input. The initial registry is built by processing deposits into the initial state.

specs/core/0_beacon-chain.md
```diff
@@ -1179,8 +1180,8 @@ def get_initial_beacon_state(initial_validator_deposits: List[Deposit],
             withdrawal_credentials=deposit.deposit_data.deposit_input.withdrawal_credentials,
             randao_commitment=deposit.deposit_data.deposit_input.randao_commitment
         )
-        if state.validator_registry[index].balance >= MAX_DEPOSIT * GWEI_PER_ETH:
-            update_validator_status(state, index, ACTIVE)
+        if state.validator_registry[validator_index].balance >= MAX_DEPOSIT * GWEI_PER_ETH:
+            update_validator_status(state, validator_index, ACTIVE)
 
     # set initial committee shuffling
     initial_shuffling = get_new_shuffling(ZERO_HASH, initial_validator_registry, 0)
```
djrtwo: use `state.validator_registry` here instead of `initial_validator_registry`

djrtwo: looks like an indentation error as well

specs/core/0_beacon-chain.md
```diff
@@ -1179,8 +1180,8 @@ def get_initial_beacon_state(initial_validator_deposits: List[Deposit],
             withdrawal_credentials=deposit.deposit_data.deposit_input.withdrawal_credentials,
             randao_commitment=deposit.deposit_data.deposit_input.randao_commitment
         )
-        if state.validator_registry[index].balance >= MAX_DEPOSIT * GWEI_PER_ETH:
-            update_validator_status(state, index, ACTIVE)
+        if state.validator_registry[validator_index].balance >= MAX_DEPOSIT * GWEI_PER_ETH:
```
djrtwo: This is going to change a bit after merging #317

specs/core/0_beacon-chain.md
```diff
@@ -1218,7 +1219,7 @@ def process_deposit(state: BeaconState,
     Process a deposit from Ethereum 1.0.
     Note that this function mutates ``state``.
     """
-    proof_of_possession_data = DepositInput(
+    deposit_input = DepositInput(
```
djrtwo: I want the name here to signal that it is the data that was signed for the proof of possession (namely, the DepositInput with an empty signature)

djrtwo: open to name changes but `depost_input` isn't descriptive enough imo

hwwhww: Right, it also took me a couple seconds to understand that `DepositInput.proof_of_possession` is set to `EMPTY_SIGNATURE` instead of the given `proof_of_possession`. What do you think about extracting a  `validate_proof_of_possession` function like: https://github.com/ethereum/py-evm/blob/56792003f676508c58371ebdc58aa0911aede717/eth/beacon/deposit_helpers.py#L43-L70

djrtwo: yeah, I like that

djrtwo: but still think `deposit_input` isn't aptly named in that context

djrtwo: or at least needs an explanatory comment

hwwhww: I changed back to `proof_of_possession_data`. 😅

specs/core/0_beacon-chain.md
hwwhww: sorry, what's the indentation error do you mean?

djrtwo: nevermind, I think I was hallucinating

PR #339, opened 2018-12-19
Don't lexicographically sort in SSZ, use manual ordering

vbuterin:
See https://github.com/ethereum/eth2.0-specs/issues/275


arnetheduck:
lgtm

Issue #323, opened 2018-12-14
Possible alternative numer-theoretic shuffling algorithm


vbuterin:
### Motivation

Construct a shuffling algorithm where you can compute the value in the shuffled list at any specific position relatively cheaply without computing all of the other values at the same time. This could be used to replace the current `shuffle`.

### Helpers

```python
def is_prime(x):
    return [i for i in range(2, int(x**0.5)+1) if x%i == 0] == []
```

### Algorithm

```python
def value_at_position(n, value, seed):
    # We do the shuffling mod p, the lowest prime >= n, but if we actually shuffle into
    # the "forbidden" [n...p-1] slice we just reshuffle until we get out of that slice
    p = n 
    while not is_prime(p):
        p += 1 
    # x -> x**power is a permutation mod p
    power = 3 
    while (p-1) % power == 0 or not is_prime(power):
        power += 2 
    for round in range(40):
        a = int.from_bytes(seed[(round % 8)*4: (round % 8)*4 + 4], 'big')
        value = (pow(value, power, p) + a) % p
        while value >= n:
            value = (pow(value, power, p) + a) % p
        # Update the seed if needed
        if round % 8 == 0:
            seed = hash(seed)
    return value

def shuffle(values, seed):
    return [values[value_at_position(len(values), i, seed)] for i in range(len(values))]
```

Note that the above is a maximally simple definition, not an optimal implementation. An optimal implementation would calculate the prime, the exponent, and the `a` values once, and pass them as inputs to `value_at_position`, which would then need to do much less work per value.

The main weaknesses are: (i) the computation of the total shuffle is slower even with optimizations, and (ii) the security argument is more heuristic ("it repeats `x -> x**3 + k[i]` so it's like MIMC") than the current [Fisher-Yates shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle). However, the strengths may be worth it, especially if we wish to cease storing the shuffling in the state.

mratsim:
What are the security property that are needed?

### Bottlenecks

The main bottleneck in the current `Fisher-Yates` shuffling is the repeated `modulo` in the loop as it's an expensive operation. In the new proposed shuffling there are `gcd`, `pow` and `modulo`.

In my opinion the main advantage of Fisher-Yates is the ability to do it in-place, however I suppose to be able to rollback state, most clients will not implement it in-place but in a temporary location instead.

### Alternative

[...]

PR #340, opened 2018-12-19
Remove MIN_VALIDATOR_REGISTRY_CHANGE_INTERVAL

vbuterin:
Slight simplification. Only substantive change is that if the validator registry stays constant, we don't reshuffle 3 epochs after the last reshuffling (ie. before the reshufflings happened after 1, 2, 3, 4, 8, 16... epochs, now it's just 1, 2, 4, 8, 16...)

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -195,7 +195,6 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted 
 | `SLOT_DURATION` | `6` | seconds | 6 seconds |
 | `MIN_ATTESTATION_INCLUSION_DELAY` | `2**2` (= 4) | slots | 24 seconds |
 | `EPOCH_LENGTH` | `2**6` (= 64) | slots | 6.4 minutes |
-| `MIN_VALIDATOR_REGISTRY_CHANGE_INTERVAL` | `2**8` (= 256) | slots | 25.6 minutes |
 | `POW_RECEIPT_ROOT_VOTING_PERIOD` | `2**10` (= 1,024) | slots | ~1.7 hours |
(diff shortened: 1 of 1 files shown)


djrtwo:
PR looks good.

But on a related note: 
Weren't we originally just utilizing the hash (or repeated hash) of the original see when doing non-validator-change reshufflings? I think the intention there was to have the future shufflings in times of non-finality to be deterministic based upon some agreed upon info at the last point of finality.

With the current continual use of new/recent randao mixes in times of non-finality (attacks, short range forks, etc), I worry we end up making the shuffling extremely subjective and ultimately an attack vector. (for example: easier to construct reasonable looking blocks when the proposer for that slot is not entirely certain)

vbuterin:
> for example: easier to construct reasonable looking blocks when the proposer for that slot is not entirely certain

What would the purpose of this attack be? Just to DoS the client?

Trying to understand better what the threat model is.

djrtwo:
Yeah, I'm worried about DoS vectors here.

If a client is uncertain of the current shuffling (seed not finalized) and receives a new block that would trigger recomputing the epoch boundary that handles the shuffling with a different seed, then the client can't immediately dismiss the block as from a bad proposer without recomputing the shuffling.

This becomes substantially more unlikely because we are grabbing the seed from `EPOCH_LENGTH` slots ago rather than from the head, but as long as it we are not grabbing a finalized seed for the shuffling, then there is some amount of a vector here.

This particular vector of blocks from uncertain proposers triggering reshufflings is reduced if utilize the alternative shuffling algorithm proposed in #323

vbuterin:
This specific PR _reduces_ the number of reshufflings going on, so it should not add any new issues in this regard.

That said, in general I'd agree that if shuffling is a large cost, then requiring clients to store many shufflings if there are many forks of the chain could reduce efficiency quite a bit. Though I'd say switching to number-theoretic shuffle or Feistel shuffle is the way to solve that particular problem.

Issue #321, opened 2018-12-14
Target Committee Size inconsistency


benjaminion:
Regarding `TARGET_COMMITTEE_SIZE`, the spec currently has a note as follows,

> For the safety of crosslinks a minimum committee size of 111 is recommended...  The shuffling algorithm generally ensures (assuming sufficient validators) committee sizes at least `TARGET_COMMITTEE_SIZE // 2`.

The second part seems incorrect: `get_new_shuffling()` produces committees of size at least `TARGET_COMMITTEE_SIZE` (not divided by two).

The current value for `TARGET_COMMITTEE_SIZE ` is 256. In view of the first part of the comment, ought it be 128?

djrtwo:
Nice catch! Making a PR

My ramblings to further explain:

With the current algorithm `TARGET_COMMITTEE_SIZE` ends up being the min and the committee size fluctuates at or above that min as a function of vset size, epoch length, and num shards. (until you get to below `EPOCH_LENGTH * TARGET_COMMITTEE_SIZE` validators at which the committees start to drop below the target.

By making `TARGET_COMMITTEE_SIZE` 256 we are being twice as strict for active validator size requirements to crosslink X shards per epoch than if we went with 128. and 128 still guarantees we are above the 111 from the blog post (until times of very small v set and thus decreased security anyway)

vbuterin:
Agree with cutting the committee size!

It's worth noting that cutting the committee size has another important benefit. If on average a committee has size n, then a validator with > 1/n of the stake will need to process almost the entire chain. Pushing n down reduces the number of validators who will need to do that.

djrtwo:
And by chain, you mean _all_ the chains.

vbuterin:
Yes :)

PR #341, opened 2018-12-19
change TARGET_COMMITTEE_SIZE to 128

djrtwo:
addresses #321

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -169,7 +169,7 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted 
 | `LATEST_RANDAO_MIXES_LENGTH` | `2**13` (= 8,192) | randao mixes |
 | `EMPTY_SIGNATURE` | `[bytes48(0), bytes48(0)]` | - |
 
-* For the safety of crosslinks a minimum committee size of 111 is [recommended](https://vitalik.ca/files/Ithaca201807_Sharding.pdf). (Unbiasable randomness with a Verifiable Delay Function (VDF) will improve committee robustness and lower the safe minimum committee size.) The shuffling algorithm generally ensures (assuming sufficient validators) committee sizes at least `TARGET_COMMITTEE_SIZE`.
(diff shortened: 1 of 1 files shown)


mratsim:
Should we call it again `MIN_COMMITTEE_SIZE` then?

djrtwo:
I considered it, but it's not actually a strict minimum. When `len(active_validators) < EPOCH_LENGTH * TARGET_COMMITTEE_SIZE` (~10k), then the committees begin to be less than `TARGET_COMMITTEE_SIZE`. This is because the least number of committees per slot is 1.

specs/core/0_beacon-chain.md
```diff
@@ -169,7 +169,7 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
 | `LATEST_RANDAO_MIXES_LENGTH` | `2**13` (= 8,192) | randao mixes |
 | `EMPTY_SIGNATURE` | `[bytes48(0), bytes48(0)]` | - |
 
-* For the safety of crosslinks a minimum committee size of 111 is [recommended](https://vitalik.ca/files/Ithaca201807_Sharding.pdf). (Unbiasable randomness with a Verifiable Delay Function (VDF) will improve committee robustness and lower the safe minimum committee size.) The shuffling algorithm generally ensures (assuming sufficient validators) committee sizes at least `TARGET_COMMITTEE_SIZE // 2`.
+* For the safety of crosslinks a minimum committee size of 111 is [recommended](https://vitalik.ca/files/Ithaca201807_Sharding.pdf). (Unbiasable randomness with a Verifiable Delay Function (VDF) will improve committee robustness and lower the safe minimum committee size.) The shuffling algorithm generally ensures (assuming sufficient validators) committee sizes at least `TARGET_COMMITTEE_SIZE`.
```
hwwhww: I found your explanation (https://github.com/ethereum/eth2.0-specs/pull/341#issuecomment-448757497) is easier to understand. Can we adjust this paragraph to be like:


* For the safety of crosslinks, `TARGET_COMMITTEE_SIZE` is the committee size that exceeds [the recommended minimum committee size `111`](https://vitalik.ca/files/Ithaca201807_Sharding.pdf); with sufficient active validators, the shuffling algorithm ensures committee sizes at least `TARGET_COMMITTEE_SIZE`. Unbiasable randomness with a [Verifiable Delay Function (VDF)](https://vdfresearch.org) will improve committee robustness and lower the safe minimum committee size.

PR #342, opened 2018-12-20
Remove line repeat in activate_validator & initiate_validator_exit

akhila-raju:
`    validator = state.validator_registry[index]` is assigned twice in `activate_validator` and `initiate_validator_exit`; the repeat assignments aren't necessary.


djrtwo:
Thanks!

PR #349, opened 2018-12-20
Proof of custody placeholders

vbuterin:
Replacing #145

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -378,6 +378,8 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted 
     'withdrawal_credentials': 'hash32',
     # Initial RANDAO commitment
     'randao_commitment': 'hash32',
+    # Initial proof of custody commitment
+    'poc_commitment': 'hash32',
     # a BLS signature of this ``DepositInput``
     'proof_of_possession': ['uint384'],
 }
@@ -475,11 +477,6 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted 
     # Placeholders for now; ProofOfCustodyChallenge is defined in phase 1, implementers can
     # put a dummy class in for now, as the list will remain empty throughout phase 0
     'poc_challenges': [ProofOfCustodyChallenge],
-    # Proof of custody commitment
-    'poc_commitment': 'hash32',
-    # Slot the proof of custody seed was last changed
-    'last_poc_change_slot': 'uint64',
-    'second_last_poc_change_slot': 'uint64',
 
     # Finality
     'previous_justified_slot': 'uint64',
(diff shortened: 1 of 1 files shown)


hwwhww:
Added commits:
1. Add initial setting of proof of custody placeholders.
2. Fix ToC.
3. Fix grammar:
    > `ProofOfCustodySeedChange`, `ProofOfCustodyChallenge`, `ProofOfCustodyResponse` defined in phase 1; for now put dummy classes as these lists will remain empty throughout phase 0.

    > `ProofOfCustodySeedChange`, `ProofOfCustodyChallenge`, and `ProofOfCustodyResponse` will be defined in phase 1; for now, put dummy classes as these lists will remain empty throughout phase 0.

specs/core/0_beacon-chain.md
```diff
@@ -418,11 +419,16 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
 
 #### `BeaconBlockBody`
 
+`ProofOfCustodySeedChange`, `ProofOfCustodyChallenge`, and `ProofOfCustodyResponse` will be defined in phase 1; for now, put dummy classes as these lists will remain empty throughout phase 0.
```
djrtwo: Notes should follow rather than prepend

specs/core/0_beacon-chain.md
```diff
@@ -465,6 +471,16 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
     'persistent_committees': [['uint24']],
     'persistent_committee_reassignments': [ShardReassignmentRecord],
 
+    # Proof of custody
+    # Placeholders for now; ProofOfCustodyChallenge is defined in phase 1, implementers can
+    # put a dummy class in for now, as the list will remain empty throughout phase 0
+    'poc_challenges': [ProofOfCustodyChallenge],
+    # Proof of custody commitment
```
djrtwo: the following three fields belong in the `ValidatorRecord`

* `poc_commitment`
* `last_poc_change_slot`
* `second_last_poc_change_slot`

specs/core/0_beacon-chain.md
```diff
@@ -1150,6 +1169,12 @@ def get_initial_beacon_state(initial_validator_deposits: List[Deposit],
         persistent_committees=[],
         persistent_committee_reassignments=[],
 
+        # Proof of custody
+        poc_challenges=[],
+        poc_commitment=ZERO_HASH,
```
djrtwo: These 3 fields belong in the `ValidatorRecord` init.

Validators should actually make and remember their commitment in the deposit phase so that all of the validators don't have to submit a commitment at once when we switch to phase 1

specs/core/0_beacon-chain.md
```diff
@@ -1533,6 +1558,10 @@ For each `exit` in `block.body.exits`:
 * Verify that `bls_verify(pubkey=validator.pubkey, message=ZERO_HASH, signature=exit.signature, domain=get_domain(state.fork_data, exit.slot, DOMAIN_EXIT))`.
 * Run `update_validator_status(state, validator_index, new_status=ACTIVE_PENDING_EXIT)`.
 
+#### Miscellaneous
```
djrtwo: I like the brevity, but it might be easier to translate in the phase 1 doc if each of the stubbed out poc operations get's their own header/section. Then in phase 1, it can be a complete replacement of the phase 0 section and order can be preserved

djrtwo: Not 100%. Anything right now would be a good guess. The operation order in the block body looks good except I think it makes sense to be put poc_seed_changes after the challenges and responses.

Leaving this section is fine.

PR #350, opened 2018-12-21
Delay exits with penalty

vbuterin:
Delaying exits with penalty by 1+epsilon epochs ensures that self-slashing single validators does not change the shuffling for the next epoch and so cannot (normally) be used as a way of  manipulating the shuffling.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -219,9 +219,9 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted 
 | `PENDING_ACTIVATION` | `0` |
 | `ACTIVE` | `1` |
 | `ACTIVE_PENDING_NO_PENALTY_EXIT` | `2` |
-| `ACTIVE_PENDING_PENALTY_EXIT` | `2` |
-| `EXITED_WITHOUT_PENALTY` | `3` |
-| `EXITED_WITH_PENALTY` | `4` |
+| `ACTIVE_PENDING_PENALTY_EXIT` | `3` |
+| `EXITED_WITHOUT_PENALTY` | `4` |
+| `EXITED_WITH_PENALTY` | `5` |
 
 ### Max operations per block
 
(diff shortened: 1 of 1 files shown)


hwwhww:
1. A bug of [`process_ejections`](https://github.com/ethereum/eth2.0-specs/blob/d722a74bb0c5886cda8a9448ee1bf3a91bf3df23/specs/core/0_beacon-chain.md#ejections), should have added `active_validator_indices = get_active_validator_indices(state.validator_registry)` 
```python
def process_ejections(state: BeaconState) -> None:
    """
    Iterate through the validator registry
    and eject active validators with balance below ``EJECTION_BALANCE``.
    """
+   active_validator_indices = get_active_validator_indices(state.validator_registry)
    for index in active_validator_indices(state.validator_registry):
        if state.validator_balances[index] < EJECTION_BALANCE:
            update_validator_status(state, index, new_status=EXITED_WITHOUT_PENALTY)
```

2. However, another issue is that `process_ejections` may cause a validator that `validator.status == ACTIVE_PENDING_PENALTY_EXIT` be ejected (`new_status=EXITED_WITHOUT_PENALTY`) in `process_ejections`, right before executung `update_validator_registry`. I think that's not what we want? Should `process_ejections` only processes the validators that `validator.status in {ACTIVE, ACTIVE_PENDING_NO_PENALTY_EXIT}`?

specs/core/0_beacon-chain.md
```diff
@@ -1722,15 +1726,23 @@ def update_validator_registry(state: BeaconState) -> None:
...
 
             # Exit validator
             update_validator_status(state, index, new_status=EXITED_WITHOUT_PENALTY)
-
+            
+    # Exit penalized validators, more than one epoch after they were penalized
+    for index, validator in enumerate(state.validator_registry):
+        is_delaying_exit = (
+            validator.status == ACTIVE_PENDING_PENALTY_EXIT and
+            state.slot > validator.latest_status_change_slot + MAX_SEED_LOOKAHEAD
```
hwwhww: Check: `>` or `>=`?
We currently use `>=` for the boundary `ZERO_BALANCE_VALIDATOR_TTL` and `MIN_VALIDATOR_WITHDRAWAL_TIME`.

```suggestion
            state.slot >= validator.latest_status_change_slot + MAX_SEED_LOOKAHEAD
```
As a side note, probably good to align these constant names.

specs/core/0_beacon-chain.md
```diff
@@ -1340,7 +1344,7 @@ def initiate_validator_exit(state: BeaconState,
     if validator.status != ACTIVE:
         return
 
-    validator.status = ACTIVE_PENDING_EXIT
+    validator.status = exit_type
     validator.latest_status_change_slot = state.slot
 ```
```
djrtwo: We need to move the `whistblower_reward` from `exit_validator` to `initiate_validator_exit`

PR #354, opened 2018-12-22
Added withdrawals with accumulator

vbuterin:
Validators withdraw a minimum of ~27 hours after exiting, with a maximum of 4 per epoch. When a validator withdraws, we add a record of this into a merkle accumulator.

Possible extension/simplification: if we have this data structure, then we can also reuse it to store the list of validator deposits and exits, reducing the need for separate logic for the validator set change hash list. This will slightly decrease efficiency of light-syncing the validator change set (in the long run by ~1.5x data-wise as anyone looking for validator set changes will also have to deal with one withdrawal for each entry+exit, and 3x computation-wise because Merkle trees are harder than hash chains), but it would simplify consensus code so it could be worth it. This would let us get rid of ValidatorRegistryDeltaBlock, get_new_validator_registry_delta_chain_tip, state.validator_registry_delta_chain_tip


specs/core/0_beacon-chain.md
```diff
@@ -168,9 +168,11 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
 | `MAX_CASPER_VOTES` | `2**10` (= 1,024) | votes |
 | `LATEST_BLOCK_ROOTS_LENGTH` | `2**13` (= 8,192) | block roots |
 | `LATEST_RANDAO_MIXES_LENGTH` | `2**13` (= 8,192) | randao mixes |
+| `MAX_WITHDRAWALS_PER_EPOCH` | 4 | [validators](#dfn-validator) |
 | `EMPTY_SIGNATURE` | `[bytes48(0), bytes48(0)]` | - |
 
 * For the safety of crosslinks `TARGET_COMMITTEE_SIZE` exceeds [the recommended minimum committee size of 111](https://vitalik.ca/files/Ithaca201807_Sharding.pdf); with sufficient active validators (at least `EPOCH_LENGTH * TARGET_COMMITTEE_SIZE`), the shuffling algorithm ensures committee sizes at least `TARGET_COMMITTEE_SIZE`. (Unbiasable randomness with a Verifiable Delay Function (VDF) will improve committee robustness and lower the safe minimum committee size.)
+* 4 withdrawals per cycle means that eg. if a 1/3 subset of 10 million ETH equivocate to cause a safety failure and all withdraw immediately, they will be able to leave in `3333333 / 32 / 4 = 26042` epochs, or 10 million seconds ~= 4 months.
```
JustinDrake: cycle => epoch

specs/core/0_beacon-chain.md
```diff
@@ -1418,6 +1456,20 @@ def exit_validator(state: BeaconState,
                 break
 ```
 
+```python
+def withdraw_validator(state: BeaconState,
+                       index: int):
+    new_accumulator_object = WithdrawalData(
+        withdrawal_credentials=validator.withdrawal_credentials,
+        slot_created=state.slot,
+        value=state.validator_balances[index]
```
JustinDrake: missing comma

specs/core/0_beacon-chain.md
```diff
@@ -400,6 +404,19 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
...
+#### `WithdrawalData`
+
+```python
+{
+    # Withdrawal credentials
+    'withdrawal_credentials': 'hash32',
+    # Slot created
+    'slot_created': 'uint64',
+    # ETH to withdraw
+    'value': 'uint64'
```
JustinDrake: missing comma

specs/core/0_beacon-chain.md
```diff
@@ -466,6 +483,9 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
     'validator_registry_latest_change_slot': 'uint64',
     'validator_registry_exit_count': 'uint64',
     'validator_registry_delta_chain_tip': 'hash32',  # For light clients to track deltas
+    'validator_withdrawal_accumulator': 'hash32',
+    'validator_registry_withdrawal_count': 'uint64'
```
JustinDrake: missing comma

specs/core/0_beacon-chain.md
```diff
@@ -1758,22 +1810,26 @@ def update_validator_registry(state: BeaconState) -> None:
...
-
-    return validator_registry, latest_penalized_exit_balances, validator_registry_delta_chain_tip
+    # Withdraw validators
+    all_indices = range(len(state.validator_registry))
+    def eligible(index):
+        return state.validator_registry[x].status in (EXITED_WITHOUT_PENALTY, EXITED_WITH_PENALTY) and \
+            state.slot >= state.validator_registry[x].latest_status_change_slot >= MIN_VALIDATOR_WITHDRAWAL_TIME
+    eligible_indices = filter(eligible, all_indices)
+    sorted_indices = sorted(eligible_indices, filter=lambda index: state.validator_registry[index].exit_count)
+    removed = 0
```
JustinDrake: For consistency: removed => withdrawn

specs/core/0_beacon-chain.md
```diff
@@ -950,6 +970,22 @@ def merkle_root(values):
     return o[1]
 ```
 
+#### `update_merkle_accumulator`
+
+```python
+def update_merkle_accumulator(acc, position, value):
+    h = value
+    j = 0
+    while (acc+1) % 2**(j+1) == 0:
```
JustinDrake: Spacing: `+1` => ` + 1`

specs/core/0_beacon-chain.md
```diff
@@ -950,6 +970,22 @@ def merkle_root(values):
...
+        h = hash(acc[j] + h)
+        j += 1
+    return acc[:j] + [h] + acc[j+1:]
```
JustinDrake: spacing

specs/core/0_beacon-chain.md
```diff
@@ -950,6 +970,22 @@ def merkle_root(values):
     return o[1]
 ```
 
+#### `update_merkle_accumulator`
+
+```python
+def update_merkle_accumulator(acc, position, value):
```
JustinDrake: rename `acc` => `accumulator`

specs/core/0_beacon-chain.md
JustinDrake: Should the throttling of withdrawals be for `EXITED_WITH_PENALTY` only? That is, anyone `EXITED_WITHOUT_PENALTY` can withdraw immediately?

specs/core/0_beacon-chain.md
```diff
@@ -400,6 +404,19 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
 }
 ```
 
+#### `WithdrawalData`
```
JustinDrake: Rename to `Withdrawal`?

specs/core/0_beacon-chain.md
```diff
@@ -466,6 +483,9 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
     'validator_registry_latest_change_slot': 'uint64',
     'validator_registry_exit_count': 'uint64',
     'validator_registry_delta_chain_tip': 'hash32',  # For light clients to track deltas
+    'validator_withdrawal_accumulator': 'hash32',
```
JustinDrake: This should be an array of `'hash32'`

specs/core/0_beacon-chain.md
```diff
@@ -1418,6 +1458,20 @@ def exit_validator(state: BeaconState,
...
 
+```python
+def withdraw_validator(state: BeaconState,
+                       index: int):
+    new_accumulator_object = WithdrawalData(
+        withdrawal_credentials=validator.withdrawal_credentials,
+        slot_created=state.slot,
+        value=state.validator_balances[index]
+    )
+    state.validator_withdrawal_accumulator = update_accumulator(
```
JustinDrake: Should this be `update_merkle_accumulator`?

specs/core/0_beacon-chain.md
JustinDrake: Rename `value` => `balance`

specs/core/0_beacon-chain.md
```diff
@@ -400,6 +404,19 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
...
 ```
 
+#### `WithdrawalData`
+
+```python
+{
+    # Withdrawal credentials
+    'withdrawal_credentials': 'hash32',
+    # Slot created
+    'slot_created': 'uint64',
```
JustinDrake: Rename `slot_created` => `slot`

specs/core/0_beacon-chain.md
```diff
@@ -400,6 +404,19 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
 }
 ```
 
+#### `WithdrawalData`
+
+```python
+{
+    # Withdrawal credentials
+    'withdrawal_credentials': 'hash32',
```
JustinDrake: Consider renaming to `credentials` (the context of a withdrawal is captured in the data structure name).

specs/core/0_beacon-chain.md
vbuterin: If we do that, we should do the same for DepositData. Should we?

JustinDrake: We have three data structures for deposits: `Deposit`, `DepositData` and `DepositInput`. Definitely an opportunity for simplification/cleanup :)

JustinDrake:
Should `MAX_WITHDRAWALS_PER_EPOCH ` be `MAX_PENALISED_WITHDRAWALS_PER_EPOCH`? That is, the withdrawal throttling would apply to `EXITED_WITH_PENALTY` only. Validators with status `EXITED_WITHOUT_PENALTY` can withdraw immediately, and therefore we may not need `EXITED_WITHOUT_PENALTY`.

vbuterin:
> Should MAX_WITHDRAWALS_PER_EPOCH be MAX_PENALISED_WITHDRAWALS_PER_EPOCH? That is, the withdrawal throttling would apply to EXITED_WITH_PENALTY only. Validators with status EXITED_WITHOUT_PENALTY can withdraw immediately, and therefore we may not need EXITED_WITHOUT_PENALTY.

No, because the point of the throttling is to give more time for validators to get penalized if they actually did something wrong.

specs/core/0_beacon-chain.md
vbuterin: Maybe leave that for a separate patch at some point down the line? :smile:

JustinDrake:
Questions regarding withdrawals:

1) What happens if, say, a custody challenge is started right after an exit event? Specifically, do we have logic to delay the withdrawal, by how much can the withdrawal be delayed by, and is this a DoS vector for anyone looking to withdraw? (IDEA: make custody challenges and/or responses non-interactive with SNARKs once we have a SNARK-friendly hash function.)
2) I guess we need to be careful to make sure that the beacon chain has enough bandwidth in the worst of conditions to penalise everyone that needs penalising. A conservative assumption is that every 128 slots (2 epochs) there is at least 1 honest validator (see safety argument [here](https://ethresear.ch/t/minimal-vdf-randomness-beacon/3566)). The smallest [Max operations per block](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#max-operations-per-block) is 16, which seems to suggest that we may be able to increase the `MAX_WITHDRAWALS_PER_EPOCH` closer to 16/2. Does that make sense?

djrtwo:
For (2), it is important to ensure that we can catch the validators faster than they can withdraw, but it is also important to ensure that if some sizable cohort of validators plan on committing a slashable offense in the future that we bound the time it would take for them to fully withdraw (thus the weak subjectivity online time synchrony assumption).

If 1/3 of the validators plan on committing a slashable offense, we want to bound the time in which they could fully withdraw at which point they could freely create a conflicting chain (which couldn't fool online nodes, but could fool nodes that have been offline for at least the time it took for the malicious cohort to withdraw).

At `MAX_WITHDRAWALS_PER_EPOCH == 4` and `validator_count == 300k` (~10M eth deposited), this would take the `100k` cartel ~111 days to fully withdraw which is close to the 4 month withdrawal synchrony assumption originally in the ffg paper. If we double to 8 per epoch then this reduces the time for full withdrawal to ~55 days.

Due to the length of time for a sizable cartel to withdraw being a function of the size of the validator set, we also need to consider what this looks like in times of lower validation say 30k validators (~1M eth), This would take the 1/3 cartel of 10k validators only ~11 days to withdraw which puts an even stricter node online assumption to remain synced.

I think the 4 is reasonable and that increasing by 2x would start to increasingly degrade the node syncing experience for more casual consumer users especially in the early days of low validation participation by moving the frequency of coming online requirement in the less than a week range. 

Note: that to safely remain synced after being offline for greater than the times specified, a user would need to supply their node with some sort of recent information from the network -- recent finalized hash or root of the validator set.

specs/core/0_beacon-chain.md
```diff
@@ -1418,6 +1461,26 @@ def exit_validator(state: BeaconState,
...
+        credentials=validator.withdrawal_credentials,
+        slot=state.slot,
+        value=state.validator_balances[index],
+    )
+    state.validator_withdrawal_accumulator = update_merkle_accumulator(
+        state.validator_withdrawal_accumulator,
+        state.validator_registry_withdrawal_count,
+        new_accumulator_object,
+    )
+    state.validator_registry_withdrawal_count += 1
```
djrtwo: Should zero out the validator balance here

specs/core/0_beacon-chain.md
```diff
@@ -1758,22 +1821,33 @@ def update_validator_registry(state: BeaconState) -> None:
...
-    for index in validators_to_penalize:
-        state.validator_balances[index] -= get_effective_balance(state, index) * min(total_penalties * 3, total_balance) // total_balance
+    # Withdraw validators
+    all_indices = range(len(state.validator_registry))
 
-    return validator_registry, latest_penalized_exit_balances, validator_registry_delta_chain_tip
+    def eligible(index):
+        is_exited = state.validator_registry[x].status in {EXITED_WITHOUT_PENALTY, EXITED_WITH_PENALTY}
+        is_greater_than_or_equal_to_min_validator_withdrawal_time = (
+            state.slot >= state.validator_registry[x].latest_status_change_slot >=
```
djrtwo: This isn't giving us what we want.
should be:
```
state.slot >= validator.latest_status_change_slot + MIN_VALIDATOR_WITHDRAWAL_TIME
```

specs/core/0_beacon-chain.md
```diff
@@ -1758,22 +1821,33 @@ def update_validator_registry(state: BeaconState) -> None:
...
-        return state.validator_registry[index].status == EXITED_WITH_PENALTY
-    validators_to_penalize = filter(to_penalize, range(len(validator_registry)))
-    for index in validators_to_penalize:
-        state.validator_balances[index] -= get_effective_balance(state, index) * min(total_penalties * 3, total_balance) // total_balance
+    # Withdraw validators
+    all_indices = range(len(state.validator_registry))
 
-    return validator_registry, latest_penalized_exit_balances, validator_registry_delta_chain_tip
+    def eligible(index):
+        is_exited = state.validator_registry[x].status in {EXITED_WITHOUT_PENALTY, EXITED_WITH_PENALTY}
```
djrtwo: suggest to define local var `validator` at start of this function to make subsequent calls to its attributes less verbose

```python
validator = state.validator_registry(index)
```

djrtwo: also var `x` is not defined. Make sure we use `index`

specs/core/0_beacon-chain.md
```diff
@@ -1758,22 +1821,33 @@ def update_validator_registry(state: BeaconState) -> None:
...
+        is_greater_than_or_equal_to_min_validator_withdrawal_time = (
+            state.slot >= state.validator_registry[x].latest_status_change_slot >=
+            MIN_VALIDATOR_WITHDRAWAL_TIME
+        )
+        return is_exited and is_greater_than_or_equal_to_min_validator_withdrawal_time
```
djrtwo: naming of `is_greater_than_or_equal_to_min_validator_withdrawal_time` seems a bit verbose without adding much to readability. maybe do the following:

```python
is_eligible_for_withdrawal = (
    is_exited and
    state.slot >= state.validator.latest_status_change_slot + MIN_VALIDATOR_WITHDRAWAL_TIME
)
return is_eligible_for_withdrawal
```

specs/core/0_beacon-chain.md
```diff
@@ -1758,22 +1821,33 @@ def update_validator_registry(state: BeaconState) -> None:
...
+
+    eligible_indices = filter(eligible, all_indices)
+    sorted_indices = sorted(eligible_indices, filter=lambda index: state.validator_registry[index].exit_count)
+    withdrawn_so_far = 0
+    for index in sorted_indices:
+        validator = state.validator_registry[index]
+        if validator.status == EXITED_WITH_PENALTY:
+            # Calculate and apply penalties for slashed validators
+            start_period = max(validator.latest_status_change_slot // COLLECTIVE_PENALTY_CALCULATION_PERIOD - 1, 0)
+            current_period = current_slot // COLLECTIVE_PENALTY_CALCULATION_PERIOD
```
djrtwo: `current_slot` not defined. Use `state.slot`

```suggestion
            current_period = state.slot // COLLECTIVE_PENALTY_CALCULATION_PERIOD
```

specs/core/0_beacon-chain.md
djrtwo: nevermind @hwwhww  just fixed

vbuterin:
Yeah, I think the `MAX_WITHDRAWALS_PER_EPOCH` parameter should be set based on how long we want the weak subjectivity period to be, and the limits should be calculated at the end to make sure people can be slashed quickly enough. 

> What happens if, say, a custody challenge is started right after an exit event? Specifically, do we have logic to delay the withdrawal, by how much can the withdrawal be delayed by, and is this a DoS vector for anyone looking to withdraw? (IDEA: make custody challenges and/or responses non-interactive with SNARKs once we have a SNARK-friendly hash function.)

First of all, we don't necessarily need to decide the final answer right now, as it's a phase 1 matter and we can just add a filter later.

But to actually answer the question, yes we do need to delay withdrawals if they get challenged, in order to facilitate multi-round challenge games such as those needed to discover which specific branch a client computed incorrectly if it computed a bad custody root.

So far the best idea I can come up with is to come up with a way that while there are extant custody challenges on a validator, it does not progress toward withdrawal, and progress resumes as soon as the validator responds. As for how to do this, one solution would be that when a validator has extant custody challenges, we stop storing their minimum exit slot and exit count, and switch to storing their remaining time until exit and exit queue position, which we don't change until all extant custody challenges are resolved, and at that point we switch back to storing their exit slot and exit count. We don't even need to modify phase 0 as any such information could be stored in the custody challenge objects themselves.

hwwhww:
Updated diagram:
spec diagram

PR #366, opened 2018-12-26
Adding Comment for Helper Function

nisdas:
Just stating that the merkle root helper is only valid if 
```py
len(values) % 2**n == 0
```

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -956,11 +956,11 @@ def get_beacon_proposer_index(state: BeaconState,
 ```python
 def merkle_root(values):  
     """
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -954,7 +954,10 @@ def get_beacon_proposer_index(state: BeaconState,
 #### `merkle_root`
 
 ```python
-def merkle_root(values):
+def merkle_root(values):  
+    """
+    This helper is only valid if `len(values) % 2**n == 0` , where `n` is any non-negative integer.
```
JustinDrake: Nitpick: use double backticks `` in docstring

PR #367, opened 2018-12-26
crosslink rewards typo

terenc3t:
One more layer to unpack to loop around all the `index` inside `state.shard_committees_at_slots[:EPOCH_LENGTH]`


JustinDrake:
Thanks. The good news is that `shard_committees_at_slots` (as well as `persistent_committee_reassignments`, `validator_registry_delta_chain_tip`) will likely be removed from `state`.

Issue #352, opened 2018-12-21
`shard_committees_at_slots` missing when processing previous epoch attestations


arnetheduck:
The scenario is the following:

At slot 6, we included an attestation for slot 1. We advance through time all the way to slot 127. The next slot will be an epoch update. We start the slot processing by increasing `state.slot` to 128. Then we proceed through the functions until we pick out `previous_epoch_attestations = [a for a in state.latest_attestations if state.slot - 2 * EPOCH_LENGTH <= a.slot < state.slot - EPOCH_LENGTH]` - our attestation from slot 1 fits this criteria (`0 <= 1 < 64`). 

We call `get_attestation_participants` which calls `get_shard_committees_at_slot(state, 1)` which computes:
`earliest_slot_in_array = state.slot - (state.slot % EPOCH_LENGTH) - EPOCH_LENGTH` or `128 - 0 - 64 = 64`. 

subsequently, `    assert earliest_slot_in_array <= slot < earliest_slot_in_array + EPOCH_LENGTH * 2` fires.

related to #348, #92, #323

c8efdc7

arnetheduck:
I'd love to, but that causes another error with `get_beacon_proposer_index(state, state.slot)` - at the epoch boundary there's no longer coverage for the proposer!

djrtwo:
yeah, yeah, subtle quirks in shuffling and helpers.
on it!

JustinDrake:
Is this addressed @djrtwo?

Issue #92, opened 2018-10-19
`get_shards_and_committees_for_slot` for future slots


djrtwo:
# Issue
`get_shards_and_committees_for_slot` only provides a shuffling for the previous cycle and the next cycle (since the last state recalc). If a cycle's worth of proposers don't show up, and the next block proposed is in the following cycle, then this function cannot be used to get the committees and proposers.

This function needs to provide valid committee/proposer for slots of arbitrary distance into the future.

# Proposed solution
```python
def get_shards_and_committees_for_slot(crystallized_state: CrystallizedState,
                                       slot: int) -> List[ShardAndCommittee]:
    earliest_slot_in_array = crystallized_state.last_state_recalculation - CYCLE_LENGTH
    assert earliest_slot_in_array <= slot
    if slot < crystallized_state.last_state_recalculation:
        index = slot - earliest_slot_in_array
    else:
        index = CYCL_LENGTH + (slot % CYCLE_LENGTH)
    return crystallized_state.shard_and_committee_for_slots[index]
```

The else statement `index` allows for slots arbitrarily in the future to be mapped to the second half of the `crystallized_state.shard_and_committee_for_slots` array.

vbuterin:
Seems reasonable to me. Typo `CYCL_LENGTH` -> `CYCLE_LENGTH`

rawfalafel:
Would this cause weird timing issues? Two nodes could have different proposers for the same slot depending on whether or not state recalculations had happened for that cycle.

hwwhww:
So the `slot >= earliest_slot_in_array + CYCLE_LENGTH * 2` case only happens when there's a large slot gap between blocks, so we need to determine the proposer of the in-coming block?

This sentence in the spec might need to be modified because we can set any future slot, but the consistency is not guaranteed now:
> `get_shards_and_committees_for_slot(_, s)` should not change unless the validator set changes.

rawfalafel:
This is an issue for the first block of every new cycle, not just when there's a large gap. It's a bit of a chicken-and-egg problem where the proposers for a new cycle are determined _after_ the first block gets processed, so the proposer for that first block isn't known when it's time to propose.

@hwwhww brings up a good point, and I think there's two issues with the proposed solution:
* `get_shards_and_committees_for_slot(_, s)` is no longer consistent, but is now dependent on whether or not `s` is the first block in the cycle. 
[...]

PR #369, opened 2018-12-27
Epoch transition during last slot of epoch

djrtwo:
* Addresses @mkalinin's comment here https://github.com/ethereum/eth2.0-specs/issues/267#issuecomment-446139113
* Also solves #352 issue 

There are a number of bugs and undefined behaviors floating around the boundaries of epochs wrt shuffling. Putting this up for review. I need to take a look at the implications to other slot-based helpers before we merge.

Note: Does not solve #92 but is related

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1578,9 +1578,12 @@ For each `exit` in `block.body.exits`:
 
 ## Per-epoch processing
 
-The steps below happen when `state.slot % EPOCH_LENGTH == 0`.
+The steps below happen when `state.slot % EPOCH_LENGTH == EPOCH_LENGTH - 1`.
 
 ### Helpers
+
+* Let `current_epoch_start_slot = state.slot - EPOCH_LENGTH + 1`.
+* Let `previous_epoch_start_slot = state.slot - 2 * EPOCH_LENGTH + 1`.
 
 All [validators](#dfn-validator):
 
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1578,34 +1578,37 @@ For each `exit` in `block.body.exits`:
 
 ## Per-epoch processing
 
-The steps below happen when `state.slot % EPOCH_LENGTH == 0`.
+The steps below happen when `state.slot % EPOCH_LENGTH == EPOCH_LENGTH - 1`.
 
 ### Helpers
 
+* Let `current_epoch_start_slot = state.slot - EPOCH_LENGTH + 1`.
+* Let `previous_epoch_start_slot = state.slot - 2 * EPOCH_LENGTH + 1`.
```
hwwhww: It seems now `state.slot=63` is the first valid slot number to perform per-epoch transtion? When `state.slot=63`, `previous_epoch_start_slot = state.slot - 2 * EPOCH_LENGTH + 1 = 63 - 2 * 64 + 1 = -64`.
1. I think we somehow assume that "slot number is an unsigned 64-bit integer"? In this special case, `-64` may not conform "the previous epoch start slot" context.
2. Might related to: https://github.com/ethereum/eth2.0-specs/issues/192

JustinDrake:
A few thoughts:

* Avoiding unnecessary `+ 1` and `- 1` feels like win to avoid off-by-one bugs and keep things clean.
* To minimise the `+ 1` and `- 1` we have some flexibility regarding the placement of the processing logic, as well as reworking the genesis block (e.g. make the block at slot 0 be a real block, as opposed to the block at slot 1).

specs/core/0_beacon-chain.md
djrtwo: I think putting a min on this value of 0 actually makes the logic work.

djrtwo:
@JustinDrake We can avoid the +1 -1 logic if we have the epoch transition at the _start_ of the next slot. This will generally make the numbers cleaner but does not 100% address #352. We have the necessary data to get the proposer from the start of the next epoch (v-set and seed) but the `get_crosslink_committees_at_slot` and `get_beacon_proposer_index` helpers do not currently allow for retrieving the shufflings until after the state transition is performed.

I think we can modify/add shuffling helpers to address this. 
I'll pick it up in the morning

djrtwo:
Let me know your thoughts on inserting the epoch transition into the _start_ of the `slot % EPOCH_LENGTH == 0` slots.

So pretty much:
* increment state.slot
* if start of epoch, transition
* do per-slot
* do per_block

JustinDrake:
> Let me know your thoughts on inserting the epoch transition into the start of the `slot % EPOCH_LENGTH == 0` slots.

If you feel that's clean then great :) I understand this PR is somewhat of a blocker, so feel free to move fast on it.

> * if start of epoch, transition
> * do per-slot

Having the per-epoch and per-slot right next to each other is nice because per-epoch is a special case of per-slot.

djrtwo:
closing in favor of #480

specs/core/0_beacon-chain.md
```diff
@@ -1656,14 +1659,14 @@ If `state.slot % ETH1_DATA_VOTING_PERIOD == 0`:
...
+* Set `state.justification_bitfield |= 1` and `state.justified_slot = current_epoch_start_slot` if `3 * current_epoch_boundary_attesting_balance >= 2 * total_balance`.
 
 Set `state.finalized_slot = state.previous_justified_slot` if any of the following are true:
 
-* `state.previous_justified_slot == state.slot - 2 * EPOCH_LENGTH and state.justification_bitfield % 4 == 3`
-* `state.previous_justified_slot == state.slot - 3 * EPOCH_LENGTH and state.justification_bitfield % 8 == 7`
-* `state.previous_justified_slot == state.slot - 4 * EPOCH_LENGTH and state.justification_bitfield % 16 in (15, 14)`
+* `state.previous_justified_slot == previous_epoch_start_slot and state.justification_bitfield % 4 == 3`
+* `state.previous_justified_slot == previous_epoch_start_slot - 1 * EPOCH_LENGTH and state.justification_bitfield % 8 == 7`
+* `state.previous_justified_slot == previous_epoch_start_slot - 2 * EPOCH_LENGTH and state.justification_bitfield % 16 in (15, 14)`
```
vbuterin: Weren't we going to fix https://github.com/ethereum/eth2.0-specs/issues/410 at the same time? Or did you want to get this PR merged first and then do a separate one for that?

djrtwo: Closed this PR. I think https://github.com/ethereum/eth2.0-specs/pull/480 is the way to go. I will fix #410 when we decide which of the two approaches to go with.

#480 does the epoch transition at the _start_ of slots (before block processing) at which `slot % EPOCH_LENGTH == 0`. this avoids all the nasty -1/+1 logic found in this PR.

Issue #358, opened 2018-12-23
Misc renamings


hwwhww:
(Work in progress!)

- [x] 1. `poc_` to `custody_` (#381)
    - `poc_seed_changes` => `custody_reseeds`
    - `poc_challenges` => `custody_challenges`
    - `poc_responses` => `custody_responses`
    - `poc_bit` => `custody_bit`
    - `ProofOfCustodySeedChange` => `CustodyReseed`
    - `ProofOfCustodyChallenge` => `CustodyChallenge`
    - `ProofOfCustodyResponse` => `CustodyResponse`
    - `poc_commitment` => `custody_commitment`
    - `last_poc_change_slot`=> `latest_custody_reseed_slot`
    - `second_last_poc_change_slot` => `penultimate_custody_reseed_slot`
    - `aggregate_signature_poc_0_indices` => `custody_bit_0_indices` (notice `'[uint24]'` should be `['uint24']`)
    - `aggregate_signature_poc_1_indices` => `custody_bit_1_indices`
- [x] 2. `INITIAL_` to `GENESIS_` (#374)
    - `INITIAL_FORK_VERSION` => `GENESIS_FORK_VERSION`
    - `INITIAL_SLOT_NUMBER` => `GENESIS_SLOT`
- [x] 3. `pow_receipt_` to `deposit_` (#383)
    - `POW_RECEIPT_ROOT_VOTING_PERIOD` => `DEPOSIT_ROOT_VOTING_PERIOD`
    - `CandidatePoWReceiptRootRecord` => `DepositRootVote`
    - `candidate_pow_receipt_root` => `deposit_root_vote`
    - `processed_pow_receipt_root` => `processed_deposit_root`
- [x] 4. Remove `Record` suffix
    - `ValidatorRecord` => `Validator`
    - `CrosslinkRecord` => `Crosslink`
    - `ShardReassignmentRecord` => `ShardReassignment`
    - `PendingAttestationRecord` => `PendingAttestation`
- [x] ~~5. `min_empty_validator_index` => `get_min_empty_validator_index`~~
- [x] 6. Committees (#443)
    - `shard_committee` => `crosslink_committee`
    - ~~`persistent_committee` => `shard_committee`~~
- [x] 7. Misc shortenings in `state` (#358)
    - `validator_registry_latest_change_slot` => `validator_registry_update_slot`
    - `fork_data` => `fork`
[...]

Issue #357, opened 2018-12-23
Reduce justification/finalization threshold from 2/3 to 5/8


vbuterin:
Reasoning:

* Safety faults are relatively unlikely, because they are accountable and the easiest to extract large penalties for
* At least a small percentage of honest nodes is going to be offline regardless
* Users seem afraid of high inactivity leaks (eg. see https://www.reddit.com/r/ethereum/comments/a8jnrh/future_staking_questions/), this will reduce both the chance and the maximum net losses to inactivity leaks

Expected consequences:

* In the "model world" (everyone online, super-high network latency, attackers are optimal), this will reduce safety fault tolerance from ~30% to ~22%, but increase liveness fault tolerance from ~33% to ~37.5%.
* In the real world, because of unavoidable honest nodes going offline, it may well be the case that de facto safety fault tolerance is currently higher than de facto liveness fault tolerance, and this will correct for this.

JustinDrake:
I'm tempted to stick with 2/3 purely for the sake of conceptual simplicity ("2/3" is a canonical threshold in various designs).

> Users seem afraid of high inactivity leaks

In the January AMA inactivity leaks didn't seem to be a significant worry.

Please reopen if you still think this is a good move :)

PR #374, opened 2018-12-28
Status code / slot reorganization

vbuterin:
Supersedes #350 and #354

* Added `activation_slot`, `exit_slot`, `penalized_slot`, `withdrawal_slot`, use these to determine if a validator is active
* Universal min activation/exit delay of 256 slots
* Min exit time of ~1 day, but penalization delays this to ~18 days
* Penalty calculation period of `[time penalized - 18 days, time penalized + 18 days]`; made the total penalties array fixed size and wraparound to make calculation more fine-grained
* Processes withdrawals in all epochs, not just dynasty-changing epochs
* Change `get_shuffling` function to take slot as argument

Not yet done:

* Removed `shard_committees` from the state
* Removed persistent committees from the state


specs/core/0_beacon-chain.md
```diff
@@ -1304,8 +1317,10 @@ def process_deposit(state: BeaconState,
             withdrawal_credentials=withdrawal_credentials,
             randao_commitment=randao_commitment,
             randao_layers=0,
-            status=PENDING_ACTIVATION,
-            latest_status_change_slot=state.slot,
+            activation_slot=FAR_FUTURE_SLOT,
+            exit_slot=FAR_FUTURE_SLOT,
+            withdrawal_slot=FAR_FUTURE_SLOT,
+            penalized_slot=FAR_FUTURE_SLOT,
             exit_count=0,
```
hwwhww: Missing: `status_flags`

specs/core/0_beacon-chain.md
```diff
@@ -188,6 +189,7 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
 | `INITIAL_FORK_VERSION` | `0` |
 | `INITIAL_SLOT_NUMBER` | `0` |
 | `ZERO_HASH` | `bytes([0] * 32)` |
+| `FAR_FUTURE_SLOT` | `2**63` |
```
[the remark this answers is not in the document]
vbuterin: > Would it be more clear if using 2**64 - 1?

The issue is that some of the protocol may involve adding numbers to the slot number, and I want to keep the results within the 64-bit int range.

> I'd suggest putting _NUMBER in the name to align with INITIAL_SLOT_NUMBER. -> FAR_FUTURE_SLOT_NUMBER.
 
Sounds good to me! Though if we're trying to reduce variable name sizes, then maybe it makes more sense to instead change `INITIAL_SLOT_NUMBER` -> `INITIAL_SLOT`?

specs/core/0_beacon-chain.md
```diff
@@ -213,15 +216,13 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
...
 
 | Name | Value |
 | - | - |
-| `PENDING_ACTIVATION` | `0` |
-| `ACTIVE` | `1` |
-| `ACTIVE_PENDING_EXIT` | `2` |
-| `EXITED_WITHOUT_PENALTY` | `3` |
-| `EXITED_WITH_PENALTY` | `4` |
+| `INITIATED_EXIT` | `2**0` (= 1) |
+| `WITHDRAWABLE` | `2**1` (= 2) |
```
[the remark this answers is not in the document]
vbuterin: I don't think we need a special constant for this; we have lots of values that we initialize to zero. Though it seems like I forgot to initialize it; fixed now.

specs/core/0_beacon-chain.md
```diff
@@ -1017,6 +1029,7 @@ def get_new_validator_registry_delta_chain_tip(current_validator_registry_delta_
             latest_registry_delta_root=current_validator_registry_delta_chain_tip,
             validator_index=validator_index,
             pubkey=pubkey,
+            slot=slot,
             flag=flag,
         )
     )
```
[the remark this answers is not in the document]
vbuterin: What do you mean by removing the wrapper? Just remove the function and say `validator_registry_delta_chain_tip = hash_tree_root(ValidatorRegistryDeltaBlock(...))` directly? I would support that.

specs/core/0_beacon-chain.md
```diff
@@ -1359,15 +1355,15 @@ def activate_validator(state: BeaconState,
     Note that this function mutates ``state``.
     """
     validator = state.validator_registry[index]
-    if validator.status != PENDING_ACTIVATION:
+    if validator.activation_slot <= state.slot:
         return
 
-    validator.status = ACTIVE
-    validator.latest_status_change_slot = state.slot
+    validator.activation_slot = state.slot + ENTRY_EXIT_DELAY
```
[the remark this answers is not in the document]
vbuterin: I'd say the cleaner solution would be to just have a flag in `process_deposit`, so you would call it with `immediate=True` in the initialization.

specs/core/0_beacon-chain.md
```diff
@@ -1380,40 +1376,22 @@ def initiate_validator_exit(state: BeaconState,
     Note that this function mutates ``state``.
     """
     validator = state.validator_registry[index]
-    if validator.status != ACTIVE:
-        return
-
-    validator.status = ACTIVE_PENDING_EXIT
-    validator.latest_status_change_slot = state.slot
+    validator.status_flags |= INITIATED_EXIT
```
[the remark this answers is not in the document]
vbuterin: I'd prefer to rename `set_validator_withdrawable`. Maybe `prepare_validator_for_withdrawal`. The name `initiate_validator_exit` _is_ pretty clear in terms of what it does...

And as for order, I feel like the logical order is the same order as the natural progression of steps in the validator's lifecycle -> activate, exit, [penalize], withdraw

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
hwwhww: > The issue is that some of the protocol may involve adding numbers to the slot number, and I want to keep the results within the 64-bit int range.

Got it 👍 

> `INITIAL_SLOT_NUMBER` -> `INITIAL_SLOT`

Sounds good to me 👍 
Also, per #358:
`INITIAL_FORK_VERSION` => `GENESIS_FORK_VERSION`
`INITIAL_SLOT_NUMBER` => `GENESIS_SLOT`

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
hwwhww: It's alright if we don't define this initial constant in spec. Just FYI, in Python implementation side, we may make `ValdiatorStatusFlags` as an ` Enum` class, so that we can add type hinting when setting `status_flags`:
```python
class ValidatorStatusFlags(IntEnum):
    INITIAL = 0
    INITIATED_EXIT = 1
    WITHDRAWABLE = 2
```

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
hwwhww: Yes! :)

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
hwwhww: Sounds good to me!

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
vbuterin: Is enum the appropriate class to use for flags? These are bit positions, not numbers that only get checked for equality (which is the usual use case of enum), so it feels like a sightly different thing.... though I don't care too much.

specs/core/0_beacon-chain.md
```diff
@@ -1330,90 +1324,42 @@ def process_deposit(state: BeaconState,
...
+_Note that all functions in this section mutate `state`_.
 
 ```python
-def activate_validator(state: BeaconState,
-                       index: int) -> None:
-    """
-    Activate the validator with the given ``index``.
-    Note that this function mutates ``state``.
-    """
+def activate_validator(state: BeaconState, index: int, immediate: bool) -> None:
```
JustinDrake: Consider renaming `immediate` to `genesis` for clarity.

(Side question: could we always do `validator.activation_slot = state.slot + ENTRY_EXIT_DELAY`, including at genesis?)

specs/core/0_beacon-chain.md
```diff
@@ -1217,14 +1208,14 @@ def get_initial_beacon_state(initial_validator_deposits: List[Deposit],
             poc_commitment=deposit.deposit_data.deposit_input.poc_commitment,
         )
         if get_effective_balance(state, validator_index) == MAX_DEPOSIT * GWEI_PER_ETH:
```
JustinDrake: This may be a bug. I think we want to process the deposits before the activations, in two separate loops. And for activations we want `>= MAX_DEPOSIT * GWEI_PER_ETH` (as opposed to `==`).

specs/core/0_beacon-chain.md
```diff
@@ -1217,14 +1208,14 @@ def get_initial_beacon_state(initial_validator_deposits: List[Deposit],
...
-            update_validator_status(state, validator_index, ACTIVE)
+            activate_validator(state, validator_index, True)
 
     # set initial committee shuffling
-    initial_shuffling = get_new_shuffling(ZERO_HASH, state.validator_registry, 0)
+    initial_shuffling = get_shuffling(ZERO_HASH, state.validator_registry, 0, GENESIS_SLOT)
```
JustinDrake: (Maybe a bug.) `state.validator_registry` should be filtered to remove non-activated validators

specs/core/0_beacon-chain.md
```diff
@@ -1238,8 +1229,8 @@ First, two helper functions:
 def min_empty_validator_index(validators: List[ValidatorRecord],
                               validator_balances: List[int],
                               current_slot: int) -> int:
-    for i, (v, vbal) in enumerate(zip(validators, validator_balances)):
-        if vbal == 0 and v.latest_status_change_slot + ZERO_BALANCE_VALIDATOR_TTL <= current_slot:
+    for i, v in enumerate(validators):
+        if v.withdrawal_slot + ZERO_BALANCE_VALIDATOR_TTL <= current_slot:
```
JustinDrake: For phases 0 and 1 I'd simply have `min_empty_validator_index` return `None` all the time. (I.e. remove `min_empty_validator_index` and `ZERO_BALANCE_VALIDATOR_TTL` logic.)

specs/core/0_beacon-chain.md
```diff
@@ -1330,90 +1324,42 @@ def process_deposit(state: BeaconState,
...
-def activate_validator(state: BeaconState,
-                       index: int) -> None:
-    """
-    Activate the validator with the given ``index``.
-    Note that this function mutates ``state``.
-    """
+def activate_validator(state: BeaconState, index: int, immediate: bool) -> None:
     validator = state.validator_registry[index]
-    if validator.status != PENDING_ACTIVATION:
+    if validator.activation_slot <= state.slot:
```
JustinDrake: Do we need these two lines because `activate_validator` is only called when `validator.activation_slot > state.slot`.

specs/core/0_beacon-chain.md
vbuterin: `get_shuffling` already does filtering.

specs/core/0_beacon-chain.md
vbuterin: We need to keep `ZERO_BALANCE_VALIDATOR_TTL` because there's logic constricting deposits that's dependent on that parameter (technically, we could delay that logic until phase 2 as well, but IMO that would cause more harm than good as it would just require us to add more special-case logic).

I'm ok with making `min_empty_validator_index` a stub for now.

specs/core/0_beacon-chain.md
vbuterin: What's the bug? I don't see it (the reason for two loops; I see the rationale for >= instead of ==).

specs/core/0_beacon-chain.md
vbuterin: Renamed to `genesis`.

> (Side question: could we always do validator.activation_slot = state.slot + ENTRY_EXIT_DELAY, including at genesis?)

I suppose we could, but not sure what the value is...

specs/core/0_beacon-chain.md
```diff
@@ -1002,26 +1013,6 @@ def get_effective_balance(state: State, index: int) -> int:
     return min(state.validator_balances[index], MAX_DEPOSIT * GWEI_PER_ETH)
 ```
 
-#### `get_new_validator_registry_delta_chain_tip`
-
-```python
-def get_new_validator_registry_delta_chain_tip(current_validator_registry_delta_chain_tip: Hash32,
```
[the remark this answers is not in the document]
vbuterin: Fixed.

specs/core/0_beacon-chain.md
djrtwo: `get_effective_balance` can never be above `MAX_DEPOSIT * GWEI_PER_ETH` as is, but `==` is fragile if the internals of `get_effective_balance` change

JustinDrake: The bug (which I fixed before reading this) is that there could be multiple calls to `activate_validator`. This would create two entries in the delta hash chain.

specs/core/0_beacon-chain.md
JustinDrake: I made some cleanups here which I'm happy to revert. I removed `min_empty_validator_index` and the associated logic, which removed a decent number of lines of code. I left the explanation message that withdrawn validator indices will be recycled.

I also removed the `ZERO_BALANCE_VALIDATOR_TTL` logic because I felt that that we don't want to unnecessarily commit to somewhat arbitrary parameters and logic. I guess I separate phases 0, 1, 2 quite strictly, with the exception of stubbing data structures. (Note: I personally lean towards @benjaminion (see [here](https://notes.ethereum.org/c/Sk8Zs--CQ/https%3A%2F%2Fbenjaminion.xyz%2Fnewineth2%2F20181228.html)) on not even adding placeholders in phase 0.)

vbuterin: Looks good to me!

specs/core/0_beacon-chain.md
```diff
@@ -859,16 +865,17 @@ def split(values: List[Any], split_count: int) -> List[Any]:
...
-                      crosslinking_start_shard: int) -> List[List[ShardCommittee]]:
+def get_shuffling(seed: Hash32,
+                  validators: List[ValidatorRecord],
+                  crosslinking_start_shard: int,
+                  slot: int) -> List[List[ShardCommittee]]:
     """
     Shuffles ``validators`` into shard committees using ``seed`` as entropy.
     """
-    active_validator_indices = get_active_validator_indices(validators)
+    active_validator_indices = get_active_validator_indices(validators, slot)
```
djrtwo: `get_shuffling` maybe should normalize slot to be the start slot of the epoch it is contained within. There are some operations that can make a validator's `exit_slot` not at a precise epoch boundary which would make `get_shuffling` consistent throughout an epoch (and a less useful helper outside of the state transition function).

Another option is to ensure that the `exit_slot` rounds up to the next cycle boundary. This is specifically an issue because `penalize_validator` can be called in the middle of an epoch.

specs/core/0_beacon-chain.md
```diff
@@ -1563,10 +1515,10 @@ Verify that `len(block.body.exits) <= MAX_EXITS`.
 For each `exit` in `block.body.exits`:
 
 * Let `validator = state.validator_registry[exit.validator_index]`.
-* Verify that `validator.status == ACTIVE`.
+* Verify that `validator.exit_slot > state.slot + ENTRY_EXIT_DELAY`.
```
djrtwo: Should be `== FAR_EXIT_SLOT`, otherwise the validator could initialize multiple exits as long as they do so before `ENTRY_EXIT_DELAY`, or if they got penalized, they could keep pushing back their exit time!

specs/core/0_beacon-chain.md
```diff
@@ -1739,61 +1691,87 @@ def update_validator_registry(state: BeaconState) -> None:
     # Activate validators within the allowable balance churn
     balance_churn = 0
     for index, validator in enumerate(state.validator_registry):
-        if validator.status == PENDING_ACTIVATION and state.validator_balances[index] >= MAX_DEPOSIT * GWEI_PER_ETH:
+        if validator.activation_slot > state.slot and state.validator_balances[index] >= MAX_DEPOSIT * GWEI_PER_ETH:
```
djrtwo: This should be `if validator.activation_slot == FAR_FUTURE_SLOT`. Because `activation_slot` is set for a future slot, this loop could re-activate already pending active validators in a future epoch transition (if a registry changed occurred within `ENTRY_EXIT_SLOTS` from the current transition)

specs/core/0_beacon-chain.md
```diff
@@ -1739,61 +1691,87 @@ def update_validator_registry(state: BeaconState) -> None:
...
 
             # Activate validator
-            update_validator_status(state, index, new_status=ACTIVE)
+            activate_validator(state, index, False)
 
     # Exit validators within the allowable balance churn
     balance_churn = 0
     for index, validator in enumerate(state.validator_registry):
-        if validator.status == ACTIVE_PENDING_EXIT:
+        if validator.exit_slot > state.slot + ENTRY_EXIT_DELAY and validator.status_flags & INITIATED_EXIT:
```
djrtwo: Same. This should be `== FUTURE_SLOT`, otherwise this loop might trigger a double exit in across multiple epoch transitions.

specs/core/0_beacon-chain.md
```diff
@@ -1739,61 +1691,87 @@ def update_validator_registry(state: BeaconState) -> None:
...
+```python
+def process_penalties_and_exits(state: BeaconState) -> None:
+    # The active validators
+    active_validator_indices = get_active_validator_indices(state.validator_registry, state.slot)
+    # The total effective balance of active validators
+    total_balance = sum([get_effective_balance(state, i) for i in active_validator_indices])
 
-    return validator_registry, latest_penalized_exit_balances, validator_registry_delta_chain_tip
+    for i, validator in enumerate(validators):
+        if (state.slot // EPOCH_LENGTH) - (validator.penalized_slot // EPOCH_LENGTH) == LATEST_PENALIZED_EXIT_LENGTH // 2:
```
djrtwo: This is going to underflow for all non-penalized validators.
Maybe we should do a check for penalized first `if validator.penalized == FAR_FUTURE_SLOT: continue`

specs/core/0_beacon-chain.md
```diff
@@ -1739,61 +1691,87 @@ def update_validator_registry(state: BeaconState) -> None:
...
+            return state.slot >= validator.penalized_slot + PENALIZED_WITHDRAWAL_TIME
+        else:
+            return state.slot >= validator.exit_slot + MIN_VALIDATOR_WITHDRAWAL_TIME
+
+    eligible_indices = filter(eligible, all_indices)
+    sorted_indices = sorted(eligible_indices, filter=lambda index: state.validator_registry[index].exit_count)
+    withdrawn_so_far = 0
+    for index in sorted_indices:
+        validator = state.validator_registry[index]
+        if validator.penalized_slot <= state.slot:
```
djrtwo: We can remove this conditional, right? We are already doing penalizations above.

djrtwo:
> total_penalties * 3 => Explain the hardcoded number 3 (maybe create a new constant)

This 3 is related to the justification/finalization threshold (1/3). We might define that in the spec in general so that it can be set depending on the instantiation of the protocol (for instance #357). This number here can then be a function of the safety threshold constant(s)

specs/core/0_beacon-chain.md
```diff
@@ -1301,128 +1283,99 @@ def process_deposit(state: BeaconState,
...
-    if new_status == EXITED_WITH_PENALTY:
-        state.latest_penalized_exit_balances[state.slot // COLLECTIVE_PENALTY_CALCULATION_PERIOD] += get_effective_balance(state, index)
-
-        whistleblower_index = get_beacon_proposer_index(state, state.slot)
-        whistleblower_reward = get_effective_balance(state, index) // WHISTLEBLOWER_REWARD_QUOTIENT
-        state.validator_balances[whistleblower_index] += whistleblower_reward
-        state.validator_balances[index] -= whistleblower_reward
-
-    if prev_status == EXITED_WITHOUT_PENALTY:
+    if validator.exit_slot < state.slot + ENTRY_EXIT_DELAY:
```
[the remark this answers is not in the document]
vbuterin: It's not actually possible to push the exit event back further, because the exit slot is the current slot plus 256, so later exits will have later exit slots, so this line will throw them out.

More philosophically, the principle I'm going after here is "state changes only succeed if they bring the state change slot _earlier_". If hypothetically we add a feature where voluntary exits take 3 weeks but penalized exits happen immediately, then it would work as is as I wrote it, but not if we change it to FAR_FUTURE_SLOT.

Same applies to the other similar cases below.

specs/core/0_beacon-chain.md
```diff
@@ -1301,128 +1283,99 @@ def process_deposit(state: BeaconState,
...
+Note: All functions in this section mutate `state`.
 
 ```python
-def activate_validator(state: BeaconState,
-                       index: int) -> None:
-    """
-    Activate the validator with the given ``index``.
-    Note that this function mutates ``state``.
-    """
+def activate_validator(state: BeaconState, index: int, genesis: bool) -> None:
```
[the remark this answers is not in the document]
vbuterin: I like the idea that the fact that activation is delayed by 256 slots is part of the activation function, and not whatever is calling it; it feels like the right place to have that logic. `activation_slot` would make things less intuitive in that regard imo; it would also break symmetry with `exit_validator` and other similar functions that don't have an activation slot.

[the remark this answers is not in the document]
vbuterin: > Comments regarding `process_penalties_and_exits`:
> 
> * `validators` is used but not defined
> * `all_indices` is used but not defined
> * `MIN_VALIDATOR_WITHDRAWAL_TIME` is used but not defined
> * `total_penalties * 3` => Explain the hardcoded number 3 (maybe create a new constant)
> * "# TODO: calculate and apply penalties for slashed validators" => Is that done above (see `state.validator_balances[index] -= penalty`)?
> * Did we agree on no withdrawal logic in phase 0? That is, can the 20 lines of code starting from `def eligible(index):` be deleted? (That would explain a lot of the above!)

Did we? I thought we agreed that we finish at validators being in the WITHDRAWABLE state?

specs/core/0_beacon-chain.md
vbuterin: I made `get_shuffling` normalize to epoch boundaries.

specs/core/0_beacon-chain.md
vbuterin: Done!

specs/core/0_beacon-chain.md
```diff
@@ -1301,128 +1288,99 @@ def process_deposit(state: BeaconState,
...
     else:
-        # Increase balance by deposit
+        # Increase balance by deposit amount
         index = validator_pubkeys.index(pubkey)
         assert state.validator_registry[index].withdrawal_credentials == withdrawal_credentials
 
-        state.validator_balances[index] += deposit
+        state.validator_balances[index] += amount
 
     return index
```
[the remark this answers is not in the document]
JustinDrake: Done thanks! 👍

specs/core/0_beacon-chain.md
```diff
@@ -1691,7 +1644,7 @@ def process_ejections(state: BeaconState) -> None:
     """
     for index in active_validator_indices(state.validator_registry):
         if state.validator_balances[index] < EJECTION_BALANCE:
-            update_validator_status(state, index, new_status=EXITED_WITHOUT_PENALTY)
+            exit_validator(state, index)
```
terenc3t: `exit_validator` checks `if validator.exit_slot < state.slot + ENTRY_EXIT_DELAY: return`. Will it work here since `validator.exit_slot` was set to `FAR_FUTURE_SLOT`

specs/core/0_beacon-chain.md
```diff
@@ -1723,61 +1676,83 @@ def update_validator_registry(state: BeaconState) -> None:
...
+    for index, validator in enumerate(state.validator_registry):
+        if (state.slot // EPOCH_LENGTH) == (validator.penalized_slot // EPOCH_LENGTH) + LATEST_PENALIZED_EXIT_LENGTH // 2:
+            e = (state.slot // EPOCH_LENGTH) % LATEST_PENALIZED_EXIT_LENGTH
+            total_at_start = state.latest_penalized_exit_balances[(e + 1) % LATEST_PENALIZED_EXIT_LENGTH]
+            total_at_end = state.latest_penalized_exit_balances[e]
+            total_penalties = total_at_end - total_at_start
+            penalty = get_effective_balance(state, index) * min(total_penalties * 3, total_balance) // total_balance
+            state.validator_balances[index] -= penalty
+
+    def eligible(index):
```
terenc3t: we should unindent  `def eligible(index):`

specs/core/0_beacon-chain.md
```diff
@@ -1285,128 +1271,97 @@ def process_deposit(state: BeaconState,
...
-        whistleblower_index = get_beacon_proposer_index(state, state.slot)
-        whistleblower_reward = get_effective_balance(state, index) // WHISTLEBLOWER_REWARD_QUOTIENT
-        state.validator_balances[whistleblower_index] += whistleblower_reward
-        state.validator_balances[index] -= whistleblower_reward
-
-    if prev_status == EXITED_WITHOUT_PENALTY:
-        return
+    validator.exit_slot = state.slot + ENTRY_EXIT_DELAY
 
     # The following updates only occur if not previous exited
```
terenc3t: I don't think this comment(`# The following updates...`) is still relevant, we can remove?

specs/core/0_beacon-chain.md
djrtwo: @vbuterin This is actually a problem. Should be `validator.activation_slot > state.slot + ENTRY_EXIT_DELAY`

Look at the following scenario with `validator`:
* `state.slot` is 256. A validator registry change occurs
    * `validator` is not activated so `activation_slot == FAR_FUTURE_SLOT`
    * in this loop, `validator` has `activation_slot > stat.slot` and gets activated
    * `validator.activation_slot == 512` (`256 + ENTRY_EXIT_DELAY`)
*  Next epoch -- `state.slot` is now 320. A registry change occurs
    * `validator.activation_slot == 512`
    * in this loop, `validator.activation_slot > state.slot` (`512 > 320`) so gets activated again
    * `validator.activation_slot == 576`

specs/core/0_beacon-chain.md
djrtwo: `exit_validator`  does that check as a blocking condition so validators are not exited twice.

In the case that a validator is not exited then `exit_slot` is _not_ less than `state.slot + ENTRY_EXIT_DELAY` so it does not immediately return and instead performs the exit

specs/core/0_beacon-chain.md
djrtwo: `eligible` is a helper function defined only within the scope of `process_penalties_and_exits`. This is valid python and a reasonable approach, but I don't see much issue in pulling it out either

specs/core/0_beacon-chain.md
djrtwo: It is still relevant but could go up above the line above it. Again, that `if validator.exit_slot < state.slot + ENTRY_EXIT_DELAY: return` is a blocker that exits the function early if the validator is already exited.

specs/core/0_beacon-chain.md
terenc3t: Ah I see. Not a python guy, ignore my comment 😅

djrtwo:
> Did we? I thought we agreed that we finish at validators being in the WITHDRAWABLE state, and "no withdrawals" meant we don't try to move beyond that point (as doing that would require for example specifying a withdrawal shard, and in general is more dependent on state execution).

I'm pro having them end in the withdrawable state. I think it's cleaner to in terms of processing penalties.

specs/core/0_beacon-chain.md
vbuterin: Yes, you're right. Will fix.

specs/core/0_beacon-chain.md
vbuterin: I'm in favor of keeping `eligible` in its current position. It really is a function only used in the line right below it, and honestly if python had non-ugly anonymous inline functions we would have just used one of those.

specs/core/0_beacon-chain.md
```diff
@@ -1285,128 +1271,97 @@ def process_deposit(state: BeaconState,
...
 ```
 
+```python
+def penalize_validator(state: BeaconState, index: int) -> None:
+    exit_validator(state, index)
+    validator = state.validator_registry[index]
+    state.latest_penalized_exit_balances[(state.slot // EPOCH_LENGTH) % LATEST_PENALIZED_EXIT_LENGTH] += get_effective_balance(state, index)
+
+    whistleblower_index = get_beacon_proposer_index(state, state.slot)
+    whistleblower_reward = get_effective_balance(state, index) // WHISTLEBLOWER_REWARD_QUOTIENT
```
hwwhww: `get_effective_balance(state, index)` is called twice in this function, could be refactored:
```python
    exit_validator(state, index)

    last_penalized_epoch = (state.slot // EPOCH_LENGTH) % LATEST_PENALIZED_EXIT_LENGTH
    effective_balance = get_effective_balance(state, index)

    state.latest_penalized_exit_balances[last_penalized_epoch] += effective_balance
    whistleblower_index = get_beacon_proposer_index(state, state.slot)
    whistleblower_reward = effective_balance // WHISTLEBLOWER_REWARD_QUOTIENT
```

djrtwo: helper var not needed here in the spec imo

djrtwo:
fixed merge conflict. Ready to merge @vbuterin

terenc3t:
I've also implemented these changes in code, if y'all wanna see what the diff looks like in code, here it is 😄  https://github.com/prysmaticlabs/prysm/pull/1233/files

PR #376, opened 2018-12-29
Remove references to the persistent committees, except the shuffling period

vbuterin:
We keep the shuffling period because it also serves as the minimum time that a validator must stay logged in.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -197,7 +197,6 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted 
 | `MIN_ATTESTATION_INCLUSION_DELAY` | `2**2` (= 4) | slots | 24 seconds |
 | `EPOCH_LENGTH` | `2**6` (= 64) | slots | 6.4 minutes |
 | `POW_RECEIPT_ROOT_VOTING_PERIOD` | `2**10` (= 1,024) | slots | ~1.7 hours |
-| `SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD` | `2**17` (= 131,072) | slots | ~9 days |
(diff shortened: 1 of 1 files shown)


JustinDrake:
👍👍 

Is the `state.slot >= validator.latest_status_change_slot + SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD` check useful in phases 0 and 1? If not I'd advocate removing `SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD` completely until phase 2.

vbuterin:
Persistent committees are a phase 1 thing, not a phase 2 thing.

I guess this depends on what the demarcation line is between phases 0 and 1: is phase 0 things _that are part of_ the basic deposit/validate/exit cycle, or things that are _needed for_ the basic deposit/validate/exit cycle? I've been going with the former so far, though ultimately it's one line of code and the time we could spend arguing about it is longer than the time needed to implement it, so I don't care too much.

JustinDrake:
Sorry I meant the specific check `state.slot >= validator.latest_status_change_slot + SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD` is part of phase 2 thing.

djrtwo:
persistent committees are a phase 1 thing, but we have been considering using the shard subnet for aggregation in phase 0. Would behoove us to have validators organize themselves into the subnets during phase 0 for this purpose

JustinDrake:
@djrtwo A relevant point you made is that it's unlikely we will have many validators before phase 0. If we have, say, 32k online validators and 100 bytes per signature then that's "only" 50kB per epoch, ~10kB/s. Given that we will have no shard data, we can temporarily have a single shard "virtual shard" where the shard data is the signature data.

A couple worries with having a subnet per shard in phase 0:

* The networking stack may not be ready in time.
* When there are few validators each subset is weak (e.g. liable to eclipse attacks).

PR #384, opened 2019-01-02
RANDAO hash chain and no stale shuffles

JustinDrake:
* Make RANDAO into a hash chain (this makes it easy for applications to prove the intermediate RANDAO reveals).
* Include `state.slot` when shuffle to avoid stale shuffles with skip slots

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -880,7 +880,7 @@ def get_shuffling(seed: Hash32,
     )
 
     # Shuffle
-    seed = xor(randao_mix, hash(slot))
+    seed = xor(randao_mix, bytes32(slot))
     shuffled_active_validator_indices = shuffle(active_validator_indices, seed)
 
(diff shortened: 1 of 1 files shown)


vbuterin:
Looks fine to me!

Though keep in mind that `get_new_shuffling` is being renamed to `get_shuffling` in the other PR.

djrtwo:
handling merge conflict as  I review @JustinDrake

specs/core/0_beacon-chain.md
```diff
@@ -880,7 +880,7 @@ def get_shuffling(seed: Hash32,
     )
 
     # Shuffle
-    seed = xor(randao_mix, Hash32(slot))
+    seed = xor(randao_mix, hash(slot))
```
JustinDrake: The intent here was a type cast, not a call to the `hash` function.

djrtwo: ah

PR #392, opened 2019-01-03
Fix a type error with units of wei when determining ejection balance.

ralexstokes:
`EJECTION_BALANCE` is in units of ETH.

`state.validator_balances[index]` is in units of Gwei.

For the ejection computation to work as desired, we need to convert the
`EJECTION_BALANCE` constant from ETH to Gwei.


pipermerriam:
Might I suggest normalizing units and instead converting `EJECTION_BALANCE` to be in `gwei` and only operating on a single *canonical* unit type within the spec.  Forcing all *amounts* to be expressed in a single unit would likely eliminate this entire class of bug.

PR #397, opened 2019-01-04
Fixes bug with Merkle proof branch selection

ralexstokes:
This code determines the order in which the next branch element and the
current value should be hashed to produce the parent node in the Merkle tree.

The existing code fails to verify branches constructed in the standard way.

This patch fixes the spec code so that it works properly by using an appropriate
parity calculation.

Example code here to illustrate it working:
https://gist.github.com/ralexstokes/9d82e188bd3286ff74a1fa1dcb5068e0


JustinDrake:
Keep it up Alex 👍You just joined and are finding lots of bugs. Are you part of the Eth2.0 Implementers call?

ralexstokes:
yep, i've joined the last few -- could not make the most recent one unfortunately but try to attend them all

PR #408, opened 2019-01-08
Remove shard committees from the state

vbuterin:
Removes shard committees from the state and instead in the spec specifies them as needing to be calculated in real time, adding a note that real life implementations should cache/memoize.

Functionality should be equivalent to status quo (I think) with the exception that the chain will be more efficient due to the reduced amount of hashing that will need to happen. Does NOT implement any of the changes proposed in https://github.com/ethereum/eth2.0-specs/issues/403.


specs/core/0_beacon-chain.md
```diff
@@ -854,75 +846,97 @@ def split(values: List[Any], split_count: int) -> List[Any]:
     ]
 ```
 
+#### `get_committees_per_slot`
+
+```python
+def get_committees_per_slot(active_validator_count: int):
```
hwwhww: type hinting:
```suggestion
def get_committees_per_slot(active_validator_count: int) -> int:
```

specs/core/0_beacon-chain.md
```diff
@@ -854,75 +846,97 @@ def split(values: List[Any], split_count: int) -> List[Any]:
...
-                total_validator_count=len(active_validator_indices),
-            )
-            for shard_position, indices in enumerate(shard_indices)
-        ]
-        output.append(shard_committees)
+#### `get_prev_epoch_committees_per_slot`
 
-    return output
+```python
+def get_prev_epoch_committees_per_slot(state: BeaconState):
```
hwwhww: type hinting:
```suggestion
def get_prev_epoch_committees_per_slot(state: BeaconState) -> int:
```

specs/core/0_beacon-chain.md
```diff
@@ -854,75 +846,97 @@ def split(values: List[Any], split_count: int) -> List[Any]:
...
+    cur_active_validators = get_active_validator_indices(validators, state.cur_epoch_calculation_slot)
+    return get_committees_per_slot(len(cur_active_validators))
+```
 
 #### `get_shard_committees_at_slot`
 
 ```python
 def get_shard_committees_at_slot(state: BeaconState,
-                                 slot: int) -> List[ShardCommittee]:
+                                 slot: int) -> (List[List[int]], int):
```
```suggestion
                                 slot: int) -> Tuple(List[List[int]], int):
```

specs/core/0_beacon-chain.md
```diff
@@ -1165,7 +1179,10 @@ def get_initial_beacon_state(initial_validator_deposits: List[Deposit],
         # Randomness and committees
         latest_randao_mixes=[ZERO_HASH for _ in range(LATEST_RANDAO_MIXES_LENGTH)],
         latest_vdf_outputs=[ZERO_HASH for _ in range(LATEST_RANDAO_MIXES_LENGTH // EPOCH_LENGTH)],
-        shard_committees_at_slots=[],
+        prev_epoch_start_shard=0,
+        cur_epoch_start_shard=0,
```
hwwhww: echo https://github.com/ethereum/eth2.0-specs/pull/408/files#r246137690:
```suggestion
        cur_epoch_start_shard=GENESIS_START_SHARD,
```

specs/core/0_beacon-chain.md
```diff
@@ -854,75 +846,97 @@ def split(values: List[Any], split_count: int) -> List[Any]:
...
 ```
 
-**Invariant**: if `get_shuffling(seed, validators, shard, slot)` returns some value `x`, it should return the same value `x` for the same `seed` and `shard` and possible future modifications of `validators` forever in phase 0, and until the ~1 year deletion delay in phase 2 and in the future.
-
-Here's a diagram of what is going on:
+#### `get_cur_epoch_committees_per_slot`
 
-![](http://vitalik.ca/files/ShuffleAndAssign.png?1)
+```python
+def get_cur_epoch_committees_per_slot(state: BeaconState):
```
```suggestion
def get_cur_epoch_committees_per_slot(state: BeaconState) -> int:
```

specs/core/0_beacon-chain.md
```diff
@@ -946,8 +960,8 @@ def get_beacon_proposer_index(state: BeaconState,
     """
     Returns the beacon proposer index for the ``slot``.
     """
-    first_committee = get_shard_committees_at_slot(state, slot)[0].committee
-    return first_committee[slot % len(first_committee)]
+    committees, shard = get_shard_committees_at_slot(state, slot)
```
hwwhww: `shard` is unused.

```suggestion
    committees, _ = get_shard_committees_at_slot(state, slot)
```

specs/core/0_beacon-chain.md
```diff
@@ -478,7 +480,10 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
     # Randomness and committees
     'latest_randao_mixes': ['hash32'],
     'latest_vdf_outputs': ['hash32'],
-    'shard_committees_at_slots': [[ShardCommittee]],
+    'prev_epoch_start_shard': 'uint64',
+    'cur_epoch_start_shard': 'uint64',
```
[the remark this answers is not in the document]
vbuterin: I think we might have to store them. The issue is that we don't remember whether or not there was a validator set change between prev and cur, and that's needed info to determine whether or not the start shard changed between the two.

specs/core/0_beacon-chain.md
```diff
@@ -478,7 +480,10 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
...
+    'prev_epoch_calculation_slot': 'uint64',
```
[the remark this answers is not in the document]
vbuterin: Seems reasonable.

specs/core/0_beacon-chain.md
```diff
@@ -854,75 +846,97 @@ def split(values: List[Any], split_count: int) -> List[Any]:
...
+                                  state.slot - (state.slot % EPOCH_LENGTH) - EPOCH_LENGTH)
+        start_shard = state.prev_epoch_start_shard
+    else:
+        committees_per_slot = get_cur_epoch_committees_per_slot(state)
+        shuffling = get_shuffling(state.latest_randao_mixes[state.cur_epoch_calculation_slot % LATEST_RANDAO_MIXES_LENGTH],
+                                  state.validator_registry,
+                                  state.slot - (state.slot % EPOCH_LENGTH))
+        start_shard = state.cur_epoch_start_shard
+
+    return shuffling[committees_per_slot * offset: committees_per_slot * (offset + 1)], \
```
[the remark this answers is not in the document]
vbuterin: If we do it this way, then doing the zipping inside of this function probably makes the most sense

djrtwo:
Doing another pass on this now. I'll handle the merge conflict from master

specs/core/0_beacon-chain.md
```diff
@@ -975,14 +1006,14 @@ def get_attestation_participants(state: BeaconState,
     Returns the participant indices at for the ``attestation_data`` and ``participation_bitfield``.
     """
 
-    # Find the relevant committee
+    # Find the committee in the list with the desired shard
     shard_committees = get_shard_committees_at_slot(state, attestation_data.slot)
-    shard_committee = [x for x in shard_committees if x.shard == attestation_data.shard][0]
-    assert len(participation_bitfield) == ceil_div8(len(shard_committee.committee))
+    shard_committee = [committee for committee, shard in shard_committees if shard == attestation_data.shard][0]
```
djrtwo: Suggest we add an assertion before this line to be explicit about handling the case in which the attestation is for a shard that is not found at that slots shuffling

```python
assert attestation.shard in [shard for _, shard in shard_committees]
```

specs/core/0_beacon-chain.md
```diff
@@ -1663,7 +1695,7 @@ def process_ejections(state: BeaconState) -> None:
 If the following are satisfied:
 
 * `state.finalized_slot > state.validator_registry_latest_change_slot`
-* `state.latest_crosslinks[shard].slot > state.validator_registry_latest_change_slot` for every shard number `shard` in `shard_committee` from `shard_committee_at_slot` in `state.shard_committees_at_slots`
+* `state.latest_crosslinks[shard].slot > state.validator_registry_latest_change_slot` for every shard number `shard` in `[(state.current_epoch_start_shard + i) % SHARD_COUNT for i in range(get_current_epoch_committees_per_slot(state) * EPOCH_LENGTH)]` (that is, for every shard in the current committees)
```
djrtwo: Should this actually be conditioned upon the previous epoch crosslink shards rather than the current? It is nearly impossible (100% impossible if we move epoch transition to last slot in epoch) that all of the crosslinks from the first epoch of a new shuffling be crosslinked during that epoch.

Neither prev or current are that clean. Both concede that this can't be satisfied within 1 epoch.

specs/core/0_beacon-chain.md
```diff
@@ -855,76 +850,99 @@ def split(values: List[Any], split_count: int) -> List[Any]:
...
 
+#### `get_committees_per_slot`
+
+```python
+def get_committees_per_slot(active_validator_count: int) -> int:
+    return max(
+        1,
+        min(
+            SHARD_COUNT // EPOCH_LENGTH,
+            len(active_validator_indices) // EPOCH_LENGTH // TARGET_COMMITTEE_SIZE,
```
terenc3t: typo, it should just be `active_validator_count // EPOCH_LENGTH // TARGET_COMMITTEE_SIZE`

specs/core/0_beacon-chain.md
```diff
@@ -855,76 +850,99 @@ def split(values: List[Any], split_count: int) -> List[Any]:
...
-        1,
-        min(
-            SHARD_COUNT // EPOCH_LENGTH,
-            len(active_validator_indices) // EPOCH_LENGTH // TARGET_COMMITTEE_SIZE,
-        )
-    )
+    committees_per_slot = get_committees_per_slot(len(active_validator_indices))
 
     # Shuffle
     seed = xor(randao_mix, bytes32(slot))
```
terenc3t: `randao_mix` is undefined, shouldn't it just be `seed = xor(seed, bytes32(slot))`?

specs/core/0_beacon-chain.md
```diff
@@ -855,76 +850,99 @@ def split(values: List[Any], split_count: int) -> List[Any]:
...
-            )
-            for shard_position, indices in enumerate(shard_indices)
-        ]
-        output.append(shard_committees)
+#### `get_previous_epoch_committees_per_slot`
 
-    return output
+```python
+def get_previous_epoch_committees_per_slot(state: BeaconState) -> int:
+    previous_active_validators = get_active_validator_indices(validators, state.previous_epoch_calculation_slot)
```
terenc3t: should be `state.validator_registry`

specs/core/0_beacon-chain.md
```diff
@@ -940,6 +958,19 @@ def get_block_root(state: BeaconState,
...
 
+#### `get_randao_mix`
+
+```python
+def get_randao_mix(state: BeaconState,
+                   slot: int) -> Hash32:
+    """
+    Returns the randao mix at a recent ``slot``.
+    """
+    assert state.slot <= slot + LATEST_RANDAO_MIXES_LENGTH
```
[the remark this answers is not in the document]
vbuterin: Ok, seems reasonable, given that the randao mix is adjusted at the start of block processing.

specs/core/0_beacon-chain.md
```diff
@@ -940,6 +958,19 @@ def get_block_root(state: BeaconState,
...
+
+```python
+def get_randao_mix(state: BeaconState,
+                   slot: int) -> Hash32:
+    """
+    Returns the randao mix at a recent ``slot``.
+    """
+    assert state.slot < slot + LATEST_RANDAO_MIXES_LENGTH
+    assert slot <= state.slot
+    return state.latest_block_roots[slot % LATEST_RANDAO_MIXES_LENGTH]
```
terenc3t: s/latest_block_roots/latest_randao_mixes/

specs/core/0_beacon-chain.md
```diff
@@ -1167,7 +1201,12 @@ def get_initial_beacon_state(initial_validator_deposits: List[Deposit],
         # Randomness and committees
         latest_randao_mixes=[ZERO_HASH for _ in range(LATEST_RANDAO_MIXES_LENGTH)],
         latest_vdf_outputs=[ZERO_HASH for _ in range(LATEST_RANDAO_MIXES_LENGTH // EPOCH_LENGTH)],
-        shard_committees_at_slots=[],
+        previous_epoch_start_shard=0,
```
terenc3t: Use `GENESIS_START_SHARD` instead of 0?

djrtwo:
addressed @terenc3t feedback and handled merge conflict

terenc3t:
Can confirm everything looks good. Implemented this change in Prysm with build passing. Thanks guys

https://github.com/prysmaticlabs/prysm/pull/1300

specs/core/0_beacon-chain.md
djrtwo: I might be missing something here.

Say we are at the epoch transition at slot 128 (the end of epoch 64-127). The current committees are the committees related to slots 64-127. The committees at slot 127 have had 1 (or 0 if we do epoch transition at the end of 127) slots to be included. So the committees at slot 127 will not be able to crosslink their shard by this state transition (especially considering `MIN_ATTESTATION_INCLUSION_DELAY`) and will only have potentially successfully crosslinked by the end of the next epoch.

vbuterin: Aah, I see, yes you're right; I was thinking of justification.

So in this case regardless of how we do it, shards would need to switch over only once every two epochs. I do think it has to be `current` though, because if we used `previous` then we could have a situation where if the previous shards are eg. 0-31 and the current shards are 32-63, and the previous shards are crosslinked but the current shards are not then the current shards increment to 64-95, which means it skips over 32-63 without crosslinking them.

djrtwo: agreed. previous isn't the solution.
This is a nasty snag. I'm not immediately seeing a good solution...

One potential would be to just have committees crosslink in the first 4/5 or something of the epoch with the remaining 1/5th is just attestation inclusion and finality gadget with shard as `2**64-1` and `shard_block_hash` as `0x0` or something.

djrtwo: Leave as is for this PR and make note in an issue?

vbuterin: Actually, it's not technically an efficiency decrease in the current code, because it turns out _both_ of the conditions for validator set rotation + start shard rotation require 2+ epochs to work. The conditions are:

If the following are satisfied:

* `state.finalized_slot > state.validator_registry_latest_change_slot`
* `state.latest_crosslinks[shard].slot > state.validator_registry_latest_change_slot` for every shard number shard in `[(state.current_epoch_start_shard + i) % SHARD_COUNT for i in range(get_current_epoch_committees_per_slot(state) * EPOCH_LENGTH)]` (that is, for every shard in the current committees)

The second rule takes two epochs because of what we discussed above. The first rule also takes two epochs because, well, finality takes two epochs.

djrtwo: You can continually finalized epoch after epoch after epoch when the chain is running smoothly. This crosslink delay does not get better due to the shuffling remaining stagnant

vbuterin: Yes, the finalized block can increase every epoch, but `state.finalized_slot > state.validator_registry_latest_change_slot` is not going to happen more than once every two blocks, because at time T you're finalizing at best T - 2 * EPOCH_LENGTH.

djrtwo: oh, well we don't technically need the `validator_registry_latest_change_slot` to lag like that. We can easily have it be `validator_registry_latest_finalized_change_slot` or something and set it to the `finalized_slot` when the registry change happens. so the condition becomes "have we finalized something new since the last registry change" instead of "have we finalized a slot past the slot when the last registry change happened"

If you do that then the crosslinks become the holdup.

djrtwo: Maybe some weirdness falls out here because the new finality could be based upon attestations included from the previous shuffling, but I don't think this will reduce the fault tolerance by more than `MAX_BALANCE_CHURN_QUOTIENT`

djrtwo: Taking this convo to #436

PR #411, opened 2019-01-08
Only require crosslinks from current epoch

djrtwo:
# Issue
We were requiring crosslinks from both the previous epoch's crosslink_committees and the current for a validator registry change to occur. Depending on the number of validators and the configuration, this could prevent validator registry changes indefinitely.

# solution
only require crosslinks from the latter half of the `crosslink_committees_at_slots` array.


vbuterin:
I think I came upon the same issue :smiley: 

My PR solves it in addition to moving committees out of the state, see line 1674 here: https://github.com/ethereum/eth2.0-specs/pull/408/files#diff-51a43328a58414e132a744f3771f018cR1674

djrtwo:
oh nice! closing

PR #413, opened 2019-01-08
Replace numeric constant with symbolic in Vyper contract

benjaminion:
Vyper v0.1.0-beta.6 supports ranges with symbolic constants.


djrtwo:
@NIC619 Can you make this change in the beacon chain repo? https://github.com/ethereum/beacon_chain/blob/master/contracts/validator_registration.v.py

PR #414, opened 2019-01-08
Add getter to Vyper contract for Merkle branches

benjaminion:
Returns the Merkle branch for the leaf at index `index`. This getter provides an alternative way for beacon chain proposers to access the Merkle tree of deposits, rather than being Ethereum 1.0 light clients. The method can be called on a trusted Ethereum 1.0 archive node at specific past block numbers via RPC/IPC to retrieve the Merkle branch needed to register a validator.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -694,9 +694,9 @@ def get_deposit_root() -> bytes32:
 
 @public
 @constant
-def get_merkle_branch(index: uint256) -> bytes32[32]: # size is DEPOSIT_CONTRACT_TREE_DEPTH (symbolic const not supported)
+def get_merkle_branch(deposit_count: uint256) -> bytes32[32]: # size is DEPOSIT_CONTRACT_TREE_DEPTH (symbolic const not supported)
     branch: bytes32[32] # size is DEPOSIT_CONTRACT_TREE_DEPTH
-    index = index + TWO_TO_POWER_OF_TREE_DEPTH
+    index: uint256 = deposit_count + TWO_TO_POWER_OF_TREE_DEPTH
     for i in range(DEPOSIT_CONTRACT_TREE_DEPTH):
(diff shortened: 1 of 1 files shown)


djrtwo:
@NIC619 Can you test this in a PR in the beacon_chain repo?

benjaminion:
I worked out how to do a bitwise xor in Vyper :-)
This is much more elegant and reduces the final bytecode size by over 100 bytes.

specs/core/0_beacon-chain.md
```diff
@@ -695,15 +695,12 @@ def get_deposit_root() -> bytes32:
 @public
 @constant
 def get_merkle_branch(index: uint256) -> bytes32[32]: # size is DEPOSIT_CONTRACT_TREE_DEPTH (symbolic const not supported)
-    idx: uint256 = index + TWO_TO_POWER_OF_TREE_DEPTH
-    ret: bytes32[32] # size is DEPOSIT_CONTRACT_TREE_DEPTH
+    branch: bytes32[32] # size is DEPOSIT_CONTRACT_TREE_DEPTH
+    index = index + TWO_TO_POWER_OF_TREE_DEPTH
```
benjaminion: This gives me a compilation error:
```
vyper.exceptions.ConstancyViolationException: Cannot modify function argument: index
```

Hence introducing the `idx` local.

benjaminion: Sorry, no :-(

```
vyper.exceptions.FunctionDeclarationException: Variable name duplicated between function arguments and globals: deposit_count
```

While we're on naming, perhaps the method name should be just `get_branch()` - wdyt?

benjaminion: We have compilation! 🎉 

I'll run it through the tests later to double check, but I can't imagine anything broke, so I'm happy for it to be merged as-is

djrtwo:
I think we need to be logging the branch along with the other deposit details for light clients to be able to construct deposits on the beacon chain, right?

Issue #399, opened 2019-01-06
Max Deposit and Max Deposits


terenc3t:
I find it slightly confusing  `MAX_DEPOSIT` and `MAX_DEPOSITS` are so close. Given they define two very different meanings, but only one `s` apart. The developer can easily misconfigure block's max operation from 16 to 32...

hwwhww:
What do people think about:
`MIN_DEPOSIT` -> `MIN_DEPOSIT_IN_ETHER`
`MAX_DEPOSIT` -> `MAX_DEPOSIT_IN_ETHER`
`GWEI_IN_ETH` -> `GWEI_IN_ETHER`

Swader:
`MIN_VALIDATOR_DEPOSIT`? Clearly differentiates from the _number_ of max deposits in the other value.

JustinDrake:
My vote is `MAX_DEPOSIT_AMOUNT` and `MIN_DEPOSIT_AMOUNT`. We no longer use "deposit" to refer to a `uint64`.

terenc3t:
That sounds good. Happy to open a PR for` MAX_DEPOSIT_AMOUNT` and `MIN_DEPOSIT_AMOUNT` if there's no objection

hwwhww:
@JustinDrake 
My point is indicating the denomination unit explicitly. Doesn't `MAX_DEPOSIT_AMOUNT` make it still/more confusing since we have `DepositData.amount` for the amount in gwei?

djrtwo:
I like 
* `MIN_DEPOSIT_IN_ETH`
* `MAX_DEPOSIT_IN_ETH`
* `GWEI_PER_ETH`

I think "eth" is clear enough and very commonly used. Not opposed to "ether" if that's the consensus.
To echo @hwwhww, py-evm has had trouble in the past with denomination ambiguities, and ultimately resolved it by being very explicit in naming and typing. We've already had a number of small bugs around manipulating ether values in the spec up to this point. I think explicit in name will almost entirely eliminate these types of bugs.

hwwhww:
#418 sounds like a great idea to solve the denomination issue.
If #418 is applied, `MAX_DEPOSIT_AMOUNT` sounds correct.

terenc3t:
Sounds good. I'll open PR today. We can just use `MAX_DEPOSIT_AMOUNT` as `32 * 1e9`

JustinDrake:
I believe this is address by #420 :)

PR #420, opened 2019-01-09
Normalize ETH units to Gwei

### LINKED ISSUE #418: Normalize ETH units to Gwei.
Following up on this: https://github.com/ethereum/eth2.0-specs/pull/392#issuecomment-452381326

The proposal is to use a single unit type for ETH/wei amounts across the spec. The idea being that as soon as such an amount enters the scope of the spec, we convert to Gwei and *then* propagate the value. Doing so eliminates an entire class of errors as we don't have to worry about unit conversions in any other computations.

Looking at the Vyper code (ETH entry point) and the validator balances in the state, they already keep amounts in Gwei.  So the only places we would have to change are places where constants are being used (simply because they are in units of ETH).

The fix for this issue converts the ETH-based constants to Gwei units and removes the `_ * GWEI_PER_ETH` modifier used throughout the spec.

Happy to make a PR if this gets enough support.

terenc3t:
This fixes #418 and #399 

* Normalized all ETH unites to Gwei
* Changed *_DEPOSIT to *_DEPOSIT_AMOUNT

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -679,12 +679,12 @@ def deposit(deposit_input: bytes[2048]):
         self.deposit_tree[index] = sha3(concat(self.deposit_tree[index * 2], self.deposit_tree[index * 2 + 1]))
 
     self.deposit_count += 1
-    if msg.value == as_wei_value(MAX_DEPOSIT, "gwei"):
+    if msg.value == as_wei_value(MAX_DEPOSIT_AMOUNT, "gwei"):
         self.full_deposit_count += 1
         if self.full_deposit_count == CHAIN_START_FULL_DEPOSIT_THRESHOLD:
             timestamp_day_boundary: uint256 = as_unitless_number(block.timestamp) - as_unitless_number(block.timestamp) % SECONDS_PER_DAY + SECONDS_PER_DAY
-            timestamp_day_boundary: bytes[8] = slice(concat("", convert(timestamp_day_boundary, bytes32)), start=24, len=8)
-            log.ChainStart(self.deposit_tree[1], timestamp_day_boundary)
+            timestamp_day_boundary_bytes8: bytes[8] = slice(concat("", convert(timestamp_day_boundary, bytes32)), start=24, len=8)
+            log.ChainStart(self.deposit_tree[1], timestamp_day_boundary_bytes8)
 
 @public
 @constant
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -670,7 +670,7 @@ def deposit(deposit_input: bytes[2048]):
     deposit_timestamp: bytes[8] = slice(concat("", convert(block.timestamp, bytes32)), start=24, len=8)
     deposit_data: bytes[2064] = concat(deposit_amount, deposit_timestamp, deposit_input)
     merkle_tree_index: bytes[8] = slice(concat("", convert(index, bytes32)), start=24, len=8)
-    
+
     log.Deposit(self.deposit_tree[1], deposit_data, merkle_tree_index)
 
     # add deposit to merkle tree
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -205,7 +205,7 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted 
 
 | Name | Value |
 | - | - |
-| `BASE_REWARD_QUOTIENT` | `2**10` (= 1,024) |
+| `BASE_REWARD_QUOTIENT` | `2**5` (= 32) |
 | `WHISTLEBLOWER_REWARD_QUOTIENT` | `2**9` (= 512) |
 | `INCLUDER_REWARD_QUOTIENT` | `2**3` (= 8) |
 | `INACTIVITY_PENALTY_QUOTIENT` | `2**24` (= 16,777,216) |
@@ -666,10 +666,11 @@ def deposit(deposit_input: bytes[2048]):
     assert msg.value <= as_wei_value(MAX_DEPOSIT_AMOUNT, "gwei")
 
     index: uint256 = self.deposit_count + TWO_TO_POWER_OF_TREE_DEPTH
-    amount: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
-    timestamp: bytes[8] = slice(concat("", convert(block.timestamp, bytes32)), start=24, len=8)
-    deposit_data: bytes[2064] = concat(amount, timestamp, deposit_input)
-
+    deposit_amount: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
+    deposit_timestamp: bytes[8] = slice(concat("", convert(block.timestamp, bytes32)), start=24, len=8)
+    deposit_data: bytes[2064] = concat(deposit_amount, deposit_timestamp, deposit_input)
+    merkle_tree_index: bytes[8] = slice(concat("", convert(index, bytes32)), start=24, len=8)
+    
     log.Deposit(self.deposit_tree[1], deposit_data, merkle_tree_index)
 
     # add deposit to merkle tree
@@ -1600,7 +1601,7 @@ For every `shard_committee_at_slot` in `state.shard_committees_at_slots` and for
 
 First, we define some additional helpers:
 
-* Let `base_reward_quotient = BASE_REWARD_QUOTIENT * integer_squareroot(total_balance)`.
+* Let `base_reward_quotient = integer_squareroot(total_balance) // BASE_REWARD_QUOTIENT`.
 * Let `base_reward(state, index) = get_effective_balance(state, index) // base_reward_quotient // 5` for any validator with the given `index`.
 * Let `inactivity_penalty(state, index, epochs_since_finality) = base_reward(state, index) + get_effective_balance(state, index) * epochs_since_finality // INACTIVITY_PENALTY_QUOTIENT // 2` for any validator with the given `index`.
 
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -666,9 +666,9 @@ def deposit(deposit_input: bytes[2048]):
     assert msg.value <= MAX_DEPOSIT_AMOUNT
 
     index: uint256 = self.deposit_count + TWO_TO_POWER_OF_TREE_DEPTH
-    msg_gwei_bytes8: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
+    amount_bytes8: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -680,7 +678,7 @@ def deposit(deposit_input: bytes[2048]):
         self.deposit_tree[index] = sha3(concat(self.deposit_tree[index * 2], self.deposit_tree[index * 2 + 1]))
 
     self.deposit_count += 1
-    if msg.value == as_wei_value(MAX_DEPOSIT, "ether"):
```
ralexstokes: this may be more than a one-liner now but we want to render `MAX_DEPOSIT` as a wei value here.

specs/core/0_beacon-chain.md
```diff
@@ -663,11 +661,11 @@ full_deposit_count: uint256
 @payable
 @public
 def deposit(deposit_input: bytes[2048]):
-    assert msg.value >= as_wei_value(MIN_DEPOSIT, "ether")
-    assert msg.value <= as_wei_value(MAX_DEPOSIT, "ether")
+    assert msg.value >= MIN_DEPOSIT_AMOUNT
+    assert msg.value <= MAX_DEPOSIT_AMOUNT
 
     index: uint256 = self.deposit_count + TWO_TO_POWER_OF_TREE_DEPTH
-    msg_gwei_bytes8: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
```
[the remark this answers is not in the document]
terenc3t: Ah! you are right! It's kept now

specs/core/0_beacon-chain.md
terenc3t: I changed it to `as_wei_value(MAX_DEPOSIT, "gwei")`

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
JustinDrake: Suggest renaming `msg_gwei` to `amount` for consistency with `DepositData`. Also tempted to drop the `_bytes8` suffixes.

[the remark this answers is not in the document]
terenc3t: Done and done

specs/core/0_beacon-chain.md
```diff
@@ -678,7 +679,7 @@ def deposit(deposit_input: bytes[2048]):
         self.deposit_tree[index] = sha3(concat(self.deposit_tree[index * 2], self.deposit_tree[index * 2 + 1]))
 
     self.deposit_count += 1
-    if msg.value == MAX_DEPOSIT_AMOUNT:
+    if msg.value == as_wei_value(MAX_DEPOSIT, "gwei"):
```
ralexstokes: this should be `as_wei_value(MAX_DEPOSIT_AMOUNT, "gwei")`.

nice find on the `as_wei_value` functionality! wasn't sure if we could do a Gwei basis in vyper or not. looks like we can :)

specs/core/0_beacon-chain.md
```diff
@@ -663,13 +662,13 @@ full_deposit_count: uint256
 @payable
 @public
 def deposit(deposit_input: bytes[2048]):
-    assert msg.value >= as_wei_value(MIN_DEPOSIT, "ether")
-    assert msg.value <= as_wei_value(MAX_DEPOSIT, "ether")
+    assert msg.value >= MIN_DEPOSIT_AMOUNT
```
JustinDrake: Do you want `as_wei_value(MAX_DEPOSIT_AMOUNT, "gwei")` here?

specs/core/0_beacon-chain.md
```diff
@@ -663,13 +662,13 @@ full_deposit_count: uint256
...
+    assert msg.value <= MAX_DEPOSIT_AMOUNT
 
     index: uint256 = self.deposit_count + TWO_TO_POWER_OF_TREE_DEPTH
-    msg_gwei_bytes8: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
-    timestamp_bytes8: bytes[8] = slice(concat("", convert(block.timestamp, bytes32)), start=24, len=8)
-    deposit_data: bytes[2064] = concat(msg_gwei_bytes8, timestamp_bytes8, deposit_input)
+    amount: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
```
JustinDrake: I'm tempted to abstract away in a helper function the (somewhat horrendous) hack of converting `x` to  `slice(concat("", convert(x, bytes32)), start=24, len=8)`. It's used 3 times in the Vyper contract.

djrtwo: lol

Will come at a cost in gas. I'm not opposed unless we end up needing to shave down gas costs.

specs/core/0_beacon-chain.md
terenc3t: Fixed

specs/core/0_beacon-chain.md
terenc3t: That sounds good. Let me know if you want the abstraction to happen in this PR

specs/core/0_beacon-chain.md
```diff
@@ -158,9 +158,8 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
 | - | - | :-: |
 | `SHARD_COUNT` | `2**10` (= 1,024) | shards |
 | `TARGET_COMMITTEE_SIZE` | `2**7` (= 128) | [validators](#dfn-validator) |
-| `EJECTION_BALANCE` | `2**4` (= 16) | ETH |
+| `EJECTION_BALANCE` | `2**4 * 1e9` (= 16,000,000,000) | Gwei |
```
djrtwo: TIL `1e9` is valid python

specs/core/0_beacon-chain.md
```diff
@@ -1602,7 +1600,7 @@ For every `shard_committee_at_slot` in `state.shard_committees_at_slots` and for
 
 First, we define some additional helpers:
 
-* Let `base_reward_quotient = BASE_REWARD_QUOTIENT * integer_squareroot(total_balance // GWEI_PER_ETH)`.
+* Let `base_reward_quotient = BASE_REWARD_QUOTIENT * integer_squareroot(total_balance)`.
```
djrtwo: the `BASE_REWARD_QUOTIENT` is actually selected to expect the int_sqrt to be of total_balance in eth rather than GWEI. We need to modify `BASE_REWARD_QUOTIENT` if we choose to not normalize `total_balance` to ETH

terenc3t: Oh yeah. It corresponds to the 10 million participating ETH in every epoch. Should we just keep `GWEI_PER_ETH` for `integer_squareroot`?

JustinDrake: I'm tempted to modify `BASE_REWARD_QUOTIENT` for Gwei accordingly.

djrtwo: yeah, I think that's the way to go.

A  comparable  interest rate is when `BASE_REWARD_QUOTIENT` is set to `1 / 32`.
If we do this, we would need to reward this calculation to be `integer_squareroot(total_balance) // BASE_REWARD_QUOTIENT`.

specs/core/0_beacon-chain.md
```diff
@@ -663,14 +662,13 @@ full_deposit_count: uint256
...
+    assert msg.value >= as_wei_value(MIN_DEPOSIT_AMOUNT, "gwei")
+    assert msg.value <= as_wei_value(MAX_DEPOSIT_AMOUNT, "gwei")
 
     index: uint256 = self.deposit_count + TWO_TO_POWER_OF_TREE_DEPTH
-    msg_gwei_bytes8: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
-    timestamp_bytes8: bytes[8] = slice(concat("", convert(block.timestamp, bytes32)), start=24, len=8)
-    deposit_data: bytes[2064] = concat(msg_gwei_bytes8, timestamp_bytes8, deposit_input)
-    merkle_tree_index: bytes[8] = slice(concat("", convert(index, bytes32)), start=24, len=8)
+    amount: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
+    timestamp: bytes[8] = slice(concat("", convert(block.timestamp, bytes32)), start=24, len=8)
```
hwwhww: Since `timestamp` is a keyword for Vyper:
```
vyper.exceptions.VariableDeclarationException: Variable name invalid or reserved: timestamp
```
(Thanks @NIC619!)

Suggest keeping the old name: `timestamp_bytes8`

specs/core/0_beacon-chain.md
```diff
@@ -663,14 +662,13 @@ full_deposit_count: uint256
...
-    assert msg.value >= as_wei_value(MIN_DEPOSIT, "ether")
-    assert msg.value <= as_wei_value(MAX_DEPOSIT, "ether")
+    assert msg.value >= as_wei_value(MIN_DEPOSIT_AMOUNT, "gwei")
+    assert msg.value <= as_wei_value(MAX_DEPOSIT_AMOUNT, "gwei")
 
     index: uint256 = self.deposit_count + TWO_TO_POWER_OF_TREE_DEPTH
-    msg_gwei_bytes8: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
-    timestamp_bytes8: bytes[8] = slice(concat("", convert(block.timestamp, bytes32)), start=24, len=8)
-    deposit_data: bytes[2064] = concat(msg_gwei_bytes8, timestamp_bytes8, deposit_input)
-    merkle_tree_index: bytes[8] = slice(concat("", convert(index, bytes32)), start=24, len=8)
```
hwwhww: Any reason of removing `merkle_tree_index`? `merkle_tree_index` is used in the following line:
```python
log.Deposit(self.deposit_tree[1], deposit_data, merkle_tree_index)
```

specs/core/0_beacon-chain.md
JustinDrake: What about `deposit_amount` and `deposit_timestamp`? Putting the type in the variable name feels wrong.

specs/core/0_beacon-chain.md
```diff
@@ -663,15 +662,15 @@ full_deposit_count: uint256
...
     index: uint256 = self.deposit_count + TWO_TO_POWER_OF_TREE_DEPTH
-    msg_gwei_bytes8: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
-    timestamp_bytes8: bytes[8] = slice(concat("", convert(block.timestamp, bytes32)), start=24, len=8)
-    deposit_data: bytes[2064] = concat(msg_gwei_bytes8, timestamp_bytes8, deposit_input)
+    deposit_amount: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
+    deposit_timestamp: bytes[8] = slice(concat("", convert(block.timestamp, bytes32)), start=24, len=8)
+    deposit_data: bytes[2064] = concat(deposit_amount, deposit_timestamp, deposit_input)
     merkle_tree_index: bytes[8] = slice(concat("", convert(index, bytes32)), start=24, len=8)
-
+
```
JustinDrake: extraneous trailing spaces

PR #422, opened 2019-01-10
Fix #320

### LINKED ISSUE #320: BLS signature verification requires double hashing message.
BLS signature verification requires double hashing message.

In current version of [BLS signature verification](https://github.com/ethereum/eth2.0-specs/blob/master/specs/bls_verify.md#hash_to_g2) present such lines:

```python
  x_re = int.from_bytes(hash(bytes8(domain) + b'\x01' + message), 'big')
  x_im = int.from_bytes(hash(bytes8(domain) + b'\x02' + message), 'big')
```

In such case `message` hashing performs 2x times, this can be ineffective, because
message can be big in size or data gathering process for `message` can be cpu intensive task.

This is pseudo-code for current hashing method:

```nim
var ctx1, ctx2: keccak256
ctx1.init()
ctx1.update(bytes(domain))
ctx2 = ctx1
ctx1.update(bytes(0x01))
ctx1.update(message)
var xre = ctx1.finish()
ctx2.update(bytes(0x02))
ctx2.update(message)
var xi = ctx2.finish()
```

As you can see only `domain` field hashing can be cached and optimized, and `hash(message)` must be performed twice for both `xre` and `xi`.

If you change current hashing method to:

```python
  x_re = int.from_bytes(hash(message + bytes8(domain) + b'\x01'), 'big')
  x_im = int.from_bytes(hash(message + bytes8(domain) + b'\x02'), 'big')
```

Pseudo code will looks like this:

```nim
var ctx1, ctx2: keccak256
ctx1.init()
ctx1.update(message)
ctx1.update(bytes(domain))
ctx2 = ctx1
ctx1.update(bytes(0x01))
var xre = ctx1.finish()
ctx2.update(bytes(0x02))
var xi = ctx2.finish()
```

As you can see not only `domain` argument cached, but `message` too.
So `hash(message)` operation performed only once.

But if you dont like scheme above, here is one more:

```python
  x_re = int.from_bytes(hash(bytes8(domain) + b'\x01' + hash(message)), 'big')
  x_im = int.from_bytes(hash(bytes8(domain) + b'\x02' + hash(message)), 'big')
```

In such way you can perform `hash(message)` only one time too.

Pseudo code will looks like this:

```nim
var ctx1, ctx2: keccak256
ctx1.init()
ctx1.update(message)
var msgdigest = ctx1.finish()
ctx1.init()
ctx1.update(bytes(domain))
ctx2 = ctx1
ctx1.update(bytes(0x01))
ctx1.update(msgdigest)
var xre = ctx1.finish()
ctx2.update(bytes(0x02))
ctx2.update(msgdigest)
var xi = ctx2.finish()
```

JustinDrake:
* Avoids unnecessary hashing of `message`
* Makes the ordering of arguments (`message` followed by `domain`) consistent with usage order

### DIFF
--- a/specs/bls_signature.md
+++ b/specs/bls_signature.md
@@ -71,8 +71,8 @@ q = 4002409555221667393417789825735904156556882819939007885332058136124031650490
 
 def hash_to_G2(message: bytes32, domain: uint64) -> [uint384]:
     # Initial candidate x coordinate
-    x_re = int.from_bytes(hash(bytes8(domain) + b'\x01' + message), 'big')
-    x_im = int.from_bytes(hash(bytes8(domain) + b'\x02' + message), 'big')
+    x_re = int.from_bytes(hash(message + bytes8(domain) + b'\x01'), 'big')
+    x_im = int.from_bytes(hash(message + bytes8(domain) + b'\x02'), 'big')
     x_coordinate = Fq2([x_re, x_im])  # x = x_re + i * x_im
     
     # Test candidate y coordinates until a one is found


JustinDrake:
One way to get the optimisation "for dummies" is to replace `message` by `hash(message)`, at the cost of a bit more hashing.

PR #429, opened 2019-01-11
Potential UInt underflow in get_shard_committees_at_slot

Nashatyrev:
It seems better to operate on unsigned values only. Original `earliest_slot` variable is effectively signed (is < 0 when `state.slot < EPOCH_LENGTH`) which may be error-prone


djrtwo:
The commit history looks out of whack. Can you try rebasing from master?

Nashatyrev:
@djrtwo Oh... Seems to look ok now

PR #438, opened 2019-01-14
Fix one of proposer slashing verifications

mkalinin:
### What's fixed
One point in proposer slashing verification that looked odd to me.


specs/core/0_beacon-chain.md
```diff
@@ -1471,7 +1471,7 @@ For each `proposer_slashing` in `block.body.proposer_slashings`:
 * Verify that `proposer_slashing.proposal_data_1.slot == proposer_slashing.proposal_data_2.slot`.
 * Verify that `proposer_slashing.proposal_data_1.shard == proposer_slashing.proposal_data_2.shard`.
 * Verify that `proposer_slashing.proposal_data_1.block_root != proposer_slashing.proposal_data_2.block_root`.
-* Verify that `validator.penalized_slot > state.slot`.
+* Verify that `proposer.penalized_slot < state.slot`.
```
djrtwo: This actually should be `>`. Because `penalized_slot` defaults to `FAR_FUTURE_SLOT` for all validators, this check makes sure taht the validator has not already been penalized.

If `proposer.penalized_slot <= state.slot` then we know the validator has already been slashed at this or prior slot and do not repeatedly set to penalized.

```suggestion
* Verify that `proposer.penalized_slot > state.slot`.
```

mkalinin: got it! 👍

PR #442, opened 2019-01-14
0_beacon_chain.md: Remove an empty else branch from get_shuffling

drozdziak1:


hwwhww:
I believe this empty `else` is for the explicitness and completeness. Quote the Zen of Python: explicit is better than implicit.

djrtwo:
The `else` also gives an explicit place to put the associated comment. 
Closing in favor of Zen and in favor of the Snake Charmer's (EF python team) general philosophy and design

PR #444, opened 2019-01-15
phase 0 validator doc

djrtwo:
As eth2.0 client development has ramped up, I've been getting more and more questions about what validators are supposed to do. Right now, what a validator is supposed to do is just implicit upon what is valid in the phase 0 beacon chain spec. This document is a first pass on making the expectations of an "honest validator" explicit.

EDIT: This in a good place for initial review and subsequent merge. It is not complete, but the bones are in place such that it can continue to grow over time.

### DIFF
--- a/specs/validator/0_beacon-chain-validator.md
+++ b/specs/validator/0_beacon-chain-validator.md
@@ -310,7 +310,7 @@ Set `attestation.data = attestation_data` where `attestation_data` is the `Attes
 * Let `aggregation_bitfield` be a byte array filled with zeros of length `(len(committee) + 7) // 8`.
 * Let `index_into_committee` be the index into the validator's `committee` at which `validator_index` is located.
 * Set `aggregation_bitfield[index_into_committee // 8] |= 2 ** (index_into_committee % 8)`.
-* Set `attestation.participation_bitfield = participation_bitfield`.
+* Set `attestation.aggregation_bitfield = aggregation_bitfield`.
 
 _Note_: Calling `get_attestation_participants(state, attestation.data, attestation.participation_bitfield)` should return `[validator_index]`.
 
--- a/specs/validator/0_beacon-chain-validator.md
+++ b/specs/validator/0_beacon-chain-validator.md
@@ -10,6 +10,8 @@ __NOTICE__: This document is a work-in-progress for researchers and implementers
     - [Table of Contents](#table-of-contents)
     - [Introduction](#introduction)
     - [Prerequisites](#prerequisites)
+    - [Constants](#constants)
+        - [Misc](#misc)
     - [Becoming a validator](#becoming-a-validator)
         - [Initialization](#initialization)
             - [BLS public key](#bls-public-key)
@@ -71,7 +73,7 @@ All terminology, constants, functions, and protocol mechanics defined in the [Ph
 ### Misc
 
 | Name | Value | Unit | Duration |
-| - | - | :-: |
+| - | - | :-: | :-: |
 | `ETH1_FOLLOW_DISTANCE` | `2**10` (= 1,024) | blocks | ~4 hours |
 
 ## Becoming a validator
--- a/specs/validator/0_beacon-chain-validator.md
+++ b/specs/validator/0_beacon-chain-validator.md
@@ -141,7 +141,7 @@ Deposits cannot be processed into the beacon chain until the eth1.0 block in whi
 
 ### Validator index
 
-Once a validator has been processed and added to the state's `validator_registry`, the validator's `validator_index` is defined by the index into the registry at which the [`ValidatorRecord`](https://github.com/ethereum/eth2.0-specs/blob/bd506e12227850888df167926b52f8fd2db31915/specs/core/0_beacon-chain.md#validatorrecord) contains the `pubkey` specified in the validator's deposit. This `validator_index` is used throughout the specification to dictate validator roles and responsibilities at any point and should be stored locally.
+Once a validator has been processed and added to the state's `validator_registry`, the validator's `validator_index` is defined by the index into the registry at which the [`ValidatorRecord`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#validatorrecord) contains the `pubkey` specified in the validator's deposit. This `validator_index` is used throughout the specification to dictate validator roles and responsibilities at any point and should be stored locally.
 
 ### Activation
 
@@ -374,7 +374,7 @@ If the software crashes at some point within this routine, then when the validat
 
 ### Casper slashing
 
-To avoid "Casper slashings", a validator must not sign two conflicting [`AttestationData`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#attestationdata) objects where conflicting is defined as a set of two attestations that satisfy either [`is_double_vote`](https://github.com/ethereum/eth2.0-specs/blob/bd506e12227850888df167926b52f8fd2db31915/specs/core/0_beacon-chain.md#is_double_vote) or [`is_surround_vote`](https://github.com/ethereum/eth2.0-specs/blob/bd506e12227850888df167926b52f8fd2db31915/specs/core/0_beacon-chain.md#is_surround_vote).
+To avoid "Casper slashings", a validator must not sign two conflicting [`AttestationData`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#attestationdata) objects where conflicting is defined as a set of two attestations that satisfy either [`is_double_vote`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#is_double_vote) or [`is_surround_vote`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#is_surround_vote).
 
 Specifically, when signing an `Attestation`, a validator should perform the following steps in the following order:
 1. Save a record to hard disk that an attestation has been signed for source -- `attestation_data.justified_slot // EPOCH_LENGTH` -- and target -- `attestation_data.slot // EPOCH_LENGTH`.
--- a/specs/validator/0_beacon-chain-validator.md
+++ b/specs/validator/0_beacon-chain-validator.md
@@ -331,27 +331,16 @@ signed_attestation_data = bls_sign(
 
 "Slashing" is the burning of some amount of validator funds and immediate ejection from the active validator set. In Phase 0, there are two ways in which funds can be slashed -- [proposal slashing](#proposal-slashing) and [attestation slashing](#casper-slashing). Although being slashed has serious repercussions, it is simple enough to avoid being slashed all together by remaining _consistent_ with respect to the messages you have previously signed.
 
-_Note_: Signed data must be within the same `Fork` context to conflict. Messages cannot be slashed across forks.
+_Note_: Signed data must be within a sequential `Fork` context to conflict. Messages cannot be slashed across diverging forks. If the previous fork version is 1 and the chain splits into fork 2 and 102, messages from 1 can slashable against messages in forks 1, 2, and 102. Messages in 2 cannot be slashable against messages in 102 and vice versa.
 
 ### Proposal slashing
 
-To avoid "proposal slashings", a validator must not sign two conflicting [`ProposalSignedData`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#proposalsigneddata) (suggest renaming `ProposalData`) where conflicting is defined as having the same `slot` and `shard` but a different `block_root`.
+To avoid "proposal slashings", a validator must not sign two conflicting [`ProposalSignedData`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#proposalsigneddata) where conflicting is defined as having the same `slot` and `shard` but a different `block_root`. In phase 0, proposals are only made for the beacon chain (`shard == BEACON_CHAIN_SHARD_NUMBER`).
 
-The following helper can be run to check if two proposal messages conflict:
-
-```python
-def proposal_data_is_slashable(proposal_data_1: ProposalSignedData,
-                               proposal_data_2: ProposalSignedData) -> bool:
-    if (proposal_data_1.slot != proposal_data_2.slot):
-        return False
-    if (proposal_data_1.shard != proposal_data_2.shard):
-        return False
-        
-    return proposal_data_1.block_root != proposal_data_2.block_root
-```
+In phase 0, as long as the validator does not sign two different beacon chain proposals for the same slot, the validator is safe against proposal slashings.
 
 Specifically, when signing an `BeaconBlock`, a validator should perform the following steps in the following order:
-1. Save a record to hard disk that an beacon block has been signed for the `slot` and `shard`.
+1. Save a record to hard disk that an beacon block has been signed for the `slot=slot` and `shard=BEACON_CHAIN_SHARD_NUMBER`.
 2. Generate and broadcast the block.
 
 If the software crashes at some point within this routine, then when the validator comes back online the hard disk has the record of the _potentially_ signed/broadcast block and can effectively avoid slashing.
--- a/specs/validator/0_beacon-chain-validator.md
+++ b/specs/validator/0_beacon-chain-validator.md
@@ -26,7 +26,7 @@ __NOTICE__: This document is a work-in-progress for researchers and implementers
                 - [Parent root](#parent-root)
                 - [State root](#state-root)
                 - [Randao reveal](#randao-reveal)
-                - [Deposit root](#deposit-root)
+                - [Eth1 Data](#eth1-data)
                 - [Signature](#signature)
             - [Block body](#block-body)
                 - [Proposer slashings](#proposer-slashings)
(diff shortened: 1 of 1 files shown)


specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,338 @@
...
+
+#### RANDAO commitment
+
+A validator's RANDAO commitment is the outermost layer of a 32-byte hash-onion. To create this commitment, perform the following steps:
+
+* Randomly generate a 32-byte `randao_seed`.
+* Store this `randao_seed` in a secure location.
+* Calculate `randao_commitment = repeat_hash(randao_seed, n)` where `n` is large enough such that within the lifetime of the validator, the validator will not propose more than `n` beacon chain blocks.
+
+Assuming `>= 100k validators`, on average a validator will have an opportunity to reveal once every `>= 600k seconds`, so `<= 50 times per year`. At this estimate, `n == 5000` would last a century. To be conservative, we recommend `n >= 100k`.
```
vbuterin: On second thought, I'd remove the last line. There are legitimate "laziness" reasons to set `n` to be smaller (eg. 5000), namely that it allows you to easily recompute the seed you need from the initial hash every time you need it without writing the extra lines of code to do caching (though with an optimal algo you can cache in a way that lets you do log(n) recalculations per reveal AFAIK).

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,338 @@
...
+
+Up to `MAX_ATTESTATIONS` aggregate attestations can be added to the `block`. The attestations added must satify the verification conditions found in [attestation processing](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#attestations-1). To maximize profit, the validator should attempt to add aggregate attestations that include the most available that have not previously been added on chain.
+
+##### Deposits
+
+##### Exits
+
+### Attestations
+
+A validator is expected to create, sign, and broadcast an attestion during each epoch. The slot during which the validator performs this roal is any slot at which `get_shard_committees_at_slot(state, slot)` contains a committee that contains `validator_index`.
```
hwwhww: typo:
- `attestion` -> `attestation`
- `roal` -> `goal`? or `role`?

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,338 @@
...
+Assuming `>= 100k validators`, on average a validator will have an opportunity to reveal once every `>= 600k seconds`, so `<= 50 times per year`. At this estimate, `n == 5000` would last a century. To be conservative, we recommend `n >= 100k`.
+
+_Note_: A validator must be able to reveal the next layer deep from their current commitment at any time. There are many strategies that trade off space and computation to be able to provide this reveal. At one end of this trade-off, a validator might only store their `randao_seed` and repeat the `repeat_hash` calculation on the fly to re-calculate the layer `n-1` for the reveal. On the other end of this trade-off, a validator might store _all_ layers of the hash-onion and not have to perform any calculations to retrieve the layer `n-1`. A more sensible strategy might be to store every `m`th layer as cached references to recalculate the intermittent layers as needed.
+
+#### Custody commitment
+
+A validator's custody commitment is the outermost layer of a 32-byte hash-onion. To create this commitment, perform the following steps:
+
+* Randomly generate a 32-byte `custody_seed`.
+* Store this `custody_seed` in a secutre location.
```
paulhauner: "secutre" -> "secure"

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,338 @@
...
+
+## Prerequisites
+
+All terminology, constants, functions, and protocol mechanics defined in the [Phase 0 -- The Beacon Chain](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md) doc are requisite for this document and used throughout. Please see the Phase 0 doc before continuing and use as a reference throughout.
+
+## Becoming a validator
+
+### Initialization
+
+A validator must initial many parameters locally before submitting a deposit and joining the validator registry.
```
paulhauner: initial -> initialize

djrtwo:
Added section [process deposit](https://github.com/ethereum/eth2.0-specs/pull/444/files#diff-abbdef70b08ada829d740f06c004c154R136) to clarify. Can you review @prestonvanloon?

prestonvanloon:
@djrtwo Much clearer, thank you!

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,381 @@
...
+## Prerequisites
+
+All terminology, constants, functions, and protocol mechanics defined in the [Phase 0 -- The Beacon Chain](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md) doc are requisite for this document and used throughout. Please see the Phase 0 doc before continuing and use as a reference throughout.
+
+## Constants
+
+### Misc
+
+| Name | Value | Unit | Duration |
+| - | - | :-: |
```
prestonvanloon: This table is messed up, I think you are missing a column here

JustinDrake:
What is the precise scope of this document? I ask because in my mind the Eth1 data voting logic for honest validators and the fork choice rule for honest validators go in the same bucket, yet they currently live in different documents (validator doc and spec respectively).

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,383 @@
...
+                - [Aggregate signature](#aggregate-signature)
+    - [How to avoid slashing](#how-to-avoid-slashing)
+        - [Proposal slashing](#proposal-slashing)
+        - [Casper slashing](#casper-slashing)
+
+<!-- /TOC -->
+
+## Introduction
+
+This document represents the expected behavior of an "honest validator" with respect to Phase 0 of the Ethereum 2.0 protocol. This document does not distinguish between a "node" and a "validator client". The separation of concerns between these (potentially) two pieces of software is left as a design decision that is outside of scope.
```
vbuterin: Maybe:

> does not distinguish between a "node" (ie. the functionality of following and reading the beacon chain) and a "validator client" (ie. the functionality of actively participating in consensus)

Otherwise some readers might wonder "huh? aren't those synonyms?"

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,383 @@
...
+
+A validator is an entity that participates in the consensus of the Ethereum 2.0 protocol. This is an optional role for users in which they can post ETH as collateral to seek financial returns in exchange for building and securing the protocol. This is similar to proof of work networks in which a miner provides collateral in the form of hardware/hash-power to seek returns in exchange for building and securing the protocol.
```
vbuterin: "post ETH as collateral" -> "post ETH as collateral and verify and attest to the validity of blocks"

That would help clarify that staking is NOT pure financial passive income.

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,383 @@
...
+To submit a deposit:
+
+* Pack the validator's [initialization parameters](#initialization) into `deposit_input`, a [`DepositInput`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#depositinput) object.
+* Set `deposit_input.proof_of_possession = EMPTY_SIGNATURE`.
+* Let `proof_of_possession` be the result of `bls_sign` of the `hash_tree_root(deposit_input)` with `domain=DOMAIN_DEPOSIT`.
+* Set `deposit_input.proof_of_possession = proof_of_possession`.
+* Let `amount` be the amount of eth to be deposited by the validator where `MIN_DEPOSIT <= amount <= MAX_DEPOSIT`.
+* Send a transaction on the Ethereum 1.0 chain to `DEPOSIT_CONTRACT_ADDRESS` executing `deposit` along with `deposit_input` as the singular `bytes` input along with a deposit `amount` in ETH.
+
+_Note_: Multiple deposits can be made by the same validator (same initialization params). A singular `Validator` will be added to `state.validator_registry` with each deposit amount being added to the validator's balance. A validator can only be activated when total deposits meet or exceed
```
vbuterin: Meet or exceed what?

Actually there is a hard invariant that each pubkey must be unique.

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Process deposit
+
+Deposits cannot be processed into the beacon chain until the eth1.0 block in which they were deposited or any of its ancestors is added to the beacon chain `state.eth1_data`. This takes _a minimum_ of `ETH1_FOLLOW_DISTANCE` eth1.0 blocks (~4 hours) plus `ETH1_DATA_VOTING_PERIOD` slots (~1.7 hours). Once the necessary eth1.0 data is added, the deposit will normally be added to a beacon chain block and processed into the `state.validator_registry` within an epoch or two. The validator is then in a queue to be activated.
+
+### Validator index
+
+Once a validator has been processed and added to the state's `validator_registry`, the validator's `validator_index` is defined by the index into the registry at which the [`ValidatorRecord`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#validatorrecord) contains the `pubkey` specified in the validator's deposit. This `validator_index` is used throughout the specification to dictate validator roles and responsibilities at any point and should be stored locally.
```
vbuterin: Maybe emphasize more that the validator index is permanent and unchanging all the way until you withdraw?

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Activation
+
+In normal operation, the validator is quickly activated at which point the validator is added to the shuffling and begins validation after an additional `ENTRY_EXIT_DELAY` slots.
```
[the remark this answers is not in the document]
vbuterin: I think keeping "an" here looks a bit better; I know it's grammatically weird but it is standard usage, eg. see https://github.com/kubernetes/kubectl/issues/508#issuecomment-449782961 for a random example

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,383 @@
...
+
+_Note_: There is a maximum validator churn per finalized epoch so the delay until activation is variable depending upon finality, total active validator balance, and the number of validators in the queue to be activated.
+
+## Beacon chain responsibilities
+
+A validator has two primary responsibilities to the beacon chain -- [proposing blocks](block-proposal) and [creating attestations](attestations-1). Proposals happen infrequently, whereas attestations should be created once per epoch.
+
+### Block proposal
+
+A validator is expected to propose a [`BeaconBlock`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#beaconblock) at the beginning of any slot during which `get_beacon_proposer_index(state, slot)` returns the validator's `validator_index`. To propose, the validator selects the `BeaconBlock`, `parent`, that in their view of the fork choice is the head of the chain during `slot`. The validator is to create, sign, and broadcast a `block` that is a child of `parent` that creates a valid [beacon chain state transition](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#beacon-chain-state-transition-function).
```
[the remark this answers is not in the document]
vbuterin: BTW should we define "canonical chain" in the Terminology section of the main doc? We use it everywhere, so should make it clear what it means.

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,383 @@
...
+
+##### Casper slashings
+
+Up to `MAX_CASPER_SLASHINGS` [`CasperSlashing`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#casperslashing) objects can be included in the `block`. The casper slashings must satisfy the verification conditions found in [casper slashings processing](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#casper-slashings-1). The validator receives a small "whistleblower" reward for each casper slashing found and included.
+
+##### Attestations
+
+Up to `MAX_ATTESTATIONS` aggregate attestations can be included in the `block`. The attestations added must satisfy the verification conditions found in [attestation processing](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#attestations-1). To maximize profit, the validator should attempt to add aggregate attestations that include the most available that have not previously been added on chain.
```
vbuterin: that include attestations from the largest number of validators whose signatures from the same epoch have not previously been added on chain?

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,383 @@
...
+
+## How to avoid slashing
+
+"Slashing" is the burning of some amount of validator funds and immediate ejection from the active validator set. In Phase 0, there are two ways in which funds can be slashed -- [proposal slashing](#proposal-slashing) and [attestation slashing](#casper-slashing). Although being slashed has serious repercussions, it is simple enough to avoid being slashed all together by remaining _consistent_ with respect to the messages you have previously signed.
+
+_Note_: Signed data must be within the same `Fork` context to conflict. Messages cannot be slashed across forks.
+
+### Proposal slashing
+
+To avoid "proposal slashings", a validator must not sign two conflicting [`ProposalSignedData`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#proposalsigneddata) (suggest renaming `ProposalData`) where conflicting is defined as having the same `slot` and `shard` but a different `block_root`.
```
vbuterin: The `shard` in this case is always `BEACON_CHAIN_SHARD_NUMBER`; probably worth clarifying that.

vbuterin: I feel like this section is too verbose; the impression you should be giving readers is something like "don't propose two different things in the same slot and you'll be fine, dawg"; making it much longer than that risks diluting the message. I'd actually consider just merging this section in with the Casper slashing section, as you gain succinctness by describing the "save to disk then broadcast" pattern once.

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,383 @@
...
+
+##### Eth1 Data
+
+`block.eth1_data` is a mechanism used by block proposers vote on a recent Ethereum 1.0 block hash and an associated deposit root found in the Ethereum 1.0 deposit contract. When consensus is formed, `state.latest_eth1_data` is updated, and validator deposits up to this root can be processed.
+
+* Let `D` be the set of `Eth1DataVote` objects `vote` in `state.eth1_data_votes` where:
+    * `vote.eth1_data.block_hash` is the hash of an eth1.0 block that is (i) part of the canonical chain, (ii) >= `ETH1_FOLLOW_DISTANCE` blocks behind the head, and (iii) newer than `state.latest_eth1_data.block_data`.
+    * `vote.eth1_data.deposit_root` is the deposit root of the eth1.0 deposit contract at the block defined by `vote.eth1_data.block_hash`.
+* If `D` is empty:
+    * Let `block_hash` be the block hash of the `ETH1_FOLLOW_DISTANCE`th ancestory of the head of the canonical eth1.0 chain.
```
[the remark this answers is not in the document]
djrtwo: ancestor!

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,383 @@
...
+
+##### Exits
+
+Up to `MAX_EXITS` [`Exit`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#exit) objects can be included in the `block`. The exits must satisfy the verification conditions found in [exits processing](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#exits-1).
+
+### Attestations
+
+A validator is expected to create, sign, and broadcast an attestation during each epoch. The slot during which the validator performs this role is any slot at which `get_shard_committees_at_slot(state, slot)` contains a committee that contains `validator_index`.
+
+A validator should create and broadcast the attestation halfway through the `slot` during which the validator is assigned -- that is `SLOT_DURATION * 0.5` seconds after the start of `slot`.
```
mkalinin: >- Validators attesting to the expected beacon chain head during the previous epoch:
>    - Let `previous_epoch_head_attestations = [a for a in previous_epoch_attestations if a.data.beacon_block_root == get_block_root(state, a.data.slot)]`.
>    - Let `previous_epoch_head_attester_indices be the union of the validator index sets given by [get_attestation_participants(state, a.data, a.aggregation_bitfield) for a in previous_epoch_head_attestations]`.
>    - Let `previous_epoch_head_attesting_balance = sum([get_effective_balance(state, i) for i in previous_epoch_head_attester_indices])`.
> ...
> - Expected beacon chain head:
>    - Any validator `index` in `previous_epoch_head_attester_indices` gains `base_reward(state, index) * previous_epoch_head_attesting_balance // total_balance)`.
>    - Any active validator `index` not in `previous_epoch_head_attester_indices` loses `base_reward(state, index)`.

Doesn't this reward calculation logic make an intention to attest to a slot when block with attesting slot number is imported or, if block didn't show up, at the end of the slot? Does it make sense to attest to a slot right at the beginning of the next one?

specs/validator/0_beacon-chain-validator.md
```diff
@@ -0,0 +1,383 @@
...
+
+_Note_: There is a maximum validator churn per finalized epoch so the delay until activation is variable depending upon finality, total active validator balance, and the number of validators in the queue to be activated.
+
+## Beacon chain responsibilities
+
+A validator has two primary responsibilities to the beacon chain -- [proposing blocks](block-proposal) and [creating attestations](attestations-1). Proposals happen infrequently, whereas attestations should be created once per epoch.
+
+### Block proposal
+
+A validator is expected to propose a [`BeaconBlock`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#beaconblock) at the beginning of any slot during which `get_beacon_proposer_index(state, slot)` returns the validator's `validator_index`. To propose, the validator selects the `BeaconBlock`, `parent`, that in their view of the fork choice is the head of the chain during `slot`. The validator is to create, sign, and broadcast a `block` that is a child of `parent` that creates a valid [beacon chain state transition](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#beacon-chain-state-transition-function).
```
hwwhww: Note that in here, the `state.slot` has to be set to `slot` when calling `get_beacon_proposer_index(state, slot)` to calculate the future slot.

@vbuterin suggested add something like: "simulate the epoch transition and see who the proposer is".

specs/validator/0_beacon-chain-validator.md
djrtwo: I'm not sure I understand the question. The intention of attesting to the block at the `slot` halfway through that `slot` is so that there is a chance for the most recent block to propagate to the attester and time for hte attester to broadcast the attestation to be included in the next block. (this is the ideal case at least)

specs/validator/0_beacon-chain-validator.md
djrtwo: Right. I am going to leave as is until we resolve the logic to allow for a guaranteed knowledge of shuffling at least 1 epoch ahead. (probably using a `next_` set of info in the state)

specs/validator/0_beacon-chain-validator.md
mkalinin: As I understand from reward calculation logic, attesting at the _end_ of the slot increases a chance to get reward with no negative side effects to validator. Cause it increases a chance to receive a block that eventually becomes a new head.

This condition `if a.data.beacon_block_root == get_block_root(state, a.data.slot)` says that reward will be gained only if attested block was a head of the chain at that particular slot. But the case is that head block could show up in _any_ moment during the slot, not only in its first half. And this incentivizes validator to defer attestation until the end of the slot, doesn't it?

mkalinin: > to broadcast the attestation to be included in the next block. (this is the ideal case at least)

Including attestation right in the next block violates `MIN_ATTESTATION_INCLUSION_DELAY` condition.

PR #448, opened 2019-01-16
Vote for block hash along with deposit root

vbuterin:
Instead of validators just voting for the PoW chain deposit root, vote for the pair of the block hash and the deposit root inside it. This ensures that these votes actually do uniquely identify a chain, (i) making code simpler (no need to handle the case where a deposit root has changed to a new value but multiple chains have this value), and (ii) making the PoS chain actually useful as a finality gadget for the PoW chain.


lightuponlight:
Finalizing the PoW chain (and a corresponding modification to the ETH 1.0 fork choice rule based on that finalization) is crucially important given the large motivation for 51% attacks on Ethereum and the ability to arbitrarily reorganize past block history.

specs/core/0_beacon-chain.md
```diff
@@ -556,17 +557,6 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
 }
 ```
 
-#### `DepositRootVote`
```
hwwhww: - `DepositRootVote` is still being used in https://github.com/ethereum/eth2.0-specs/blob/212062fc4e7d8f7cd0857ceb42a3a2fd75f16d60/specs/core/0_beacon-chain.md#deposit-root
- The whole "Per-block processing - Deposit Root" section needs to be updated to "PoW Chain Root" version.
- Do we need to update `BeaconBlock.deposit_root: Hash32` -> `BeaconBlock.pow_chain_data: PoWChainData` so that the block proposers are voting on the PoW chain block hash?

specs/core/0_beacon-chain.md
```diff
@@ -1249,7 +1261,7 @@ def get_initial_beacon_state(initial_validator_deposits: List[Deposit],
         batched_block_roots=[],
 
         # Deposit root
-        latest_deposit_root=latest_deposit_root,
+        latest_pow_chain_data=latest_pow_chain_data,
         deposit_root_votes=[],
```
```suggestion
        pow_chain_data_votes=[],
```

JustinDrake: I suggest changing `pow_chain_` to `eth1_` because the Ethereum 1.0 chain may not remain a PoW chain in the long term. Also shorter :)

djrtwo: yeah, `eth1_` is more descriptive and succinct imo

JustinDrake:
(i) makes sense.

> (ii) making the PoS chain actually useful as a finality gadget for the PoW chain.

The new fork choice rule finalising the PoW chain would need to be enforced somehow. We probably cannot trust the miners to play nice so the new fork choice rule needs to be enforced by the community. That means making Ethereum 1.0 nodes (e.g. those of exchanges, explorers, etc.) Ethereum 2.0 aware. In particular, Ethereum 1.0 light clients would need to also become Ethereum 2.0 light clients, and the combination may not be so "light".

djrtwo:
We need to sort through some weirdness wrt chain start conditions

If the chain start sets `latest_pow_chain_data` to the eth1 head, then the validators don't necessarily have an eth1 block at depth >= 1000 that is also newer than the initial pow chain data to vote on in initial block proposals. I suggest we have the genesis time in the chain start even set to "round up to midnight + 1 day" to ensure >= block follow distance to start.

vbuterin:
> If the chain start sets latest_pow_chain_data to the eth1 head, then the validators don't necessarily have an eth1 block at depth >= 1000 that is also newer than the initial pow chain data to vote on in initial block proposals. I suggest we have the genesis time in the chain start even set to "round up to midnight + 1 day" to ensure >= block follow distance to start.

I would add a rule to your honest behavior rulebook that if no new block is available then just vote for the same data that's already in the state.

specs/core/0_beacon-chain.md
vbuterin: All done!

specs/core/0_beacon-chain.md
vbuterin: Done!

vbuterin: Done!

specs/core/0_beacon-chain.md
```diff
@@ -106,7 +107,7 @@
             - [Custody](#custody)
     - [Per-epoch processing](#per-epoch-processing)
         - [Helpers](#helpers)
-        - [Deposit roots](#deposit-roots)
+        - [PoW chain data](#pow-chain-data)
```
djrtwo: suggest "Eth1 chain data"

specs/core/0_beacon-chain.md
```diff
@@ -203,7 +204,7 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted
 | `EPOCH_LENGTH` | `2**6` (= 64) | slots | 6.4 minutes |
 | `SEED_LOOKAHEAD` | `2**6` (= 64) | slots | 6.4 minutes |
 | `ENTRY_EXIT_DELAY` | `2**8` (= 256) | slots | 25.6 minutes |
-| `DEPOSIT_ROOT_VOTING_PERIOD` | `2**10` (= 1,024) | slots | ~1.7 hours |
+| `POW_CHAIN_DATA_VOTING_PERIOD` | `2**10` (= 1,024) | slots | ~1.7 hours |
```
djrtwo: suggest `ETH1_CHAIN_DATA_VOTING_PERIOD`

specs/core/0_beacon-chain.md
```diff
@@ -1248,9 +1263,9 @@ def get_initial_beacon_state(initial_validator_deposits: List[Deposit],
         latest_attestations=[],
         batched_block_roots=[],
 
-        # Deposit root
-        latest_deposit_root=latest_deposit_root,
-        deposit_root_votes=[],
+        # PoW chain data
```
djrtwo: suggest "Eth1 chain data"

PR #460, opened 2019-01-17
Clean Up Deposit Contract

nisdas:
This PR cleans up the deposit contract, with changes in declarations, which was suggested by @jacqueswww. 

Also another issue was that the vyper contract was not able to be compiled, however @jacqueswww has been kind enough to put up a quick fix here: https://github.com/ethereum/vyper/pull/1202. After that is merged in the deposit contract should have no issues being compiled.

Changes in Contract
- Update version number to reflect latest vyper release.
- Perform all calculations in terms of gwei
- Change global constant from `Gwei_Per_Eth` to `Wei_Per_Gwei`

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -666,7 +666,7 @@ deposit_tree: map(uint256, bytes32)
 deposit_count: uint256
 full_deposit_count: uint256
 
-@private
+@public
 @constant
 def to_bytes8(value: uint256) -> bytes[8]:
     return slice(concat("", convert(value, bytes32)), start=24, len=8)
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -652,9 +652,8 @@ When sufficiently many full deposits have been made the deposit contract emits t
 ```python
 ## compiled with v0.1.0-beta.7 ##
 
-MIN_DEPOSIT: constant(uint256) = 1000000000 # In terms of gwei
-MAX_DEPOSIT: constant(uint256) = 32000000000 # In terms of gwei
-WEI_PER_GWEI: constant(uint256(wei)) = as_wei_value(1,"gwei") # 10**9
+MIN_DEPOSIT_AMOUNT: constant(uint256) = 1000000000  # In Gwei
+MAX_DEPOSIT_AMOUNT: constant(uint256) = 32000000000  # In Gwei
 CHAIN_START_FULL_DEPOSIT_THRESHOLD: constant(uint256) = 16384  # 2**14
 DEPOSIT_CONTRACT_TREE_DEPTH: constant(uint256) = 32
 TWO_TO_POWER_OF_TREE_DEPTH: constant(uint256) = 4294967296  # 2**32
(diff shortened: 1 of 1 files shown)


hwwhww:
There's a `MAX_DEPOSIT_AMOUNT` thing on L691.

hwwhww:
@nisdas 
I labeled `DO NOT MERGE` since it's pending on the Vyper release.

jacqueswww:
Just tagged beta-7, please give it a shot :)

hwwhww:
Second thought, the bytecodes are exactly the same. So now I don't feel strongly of having this refactoring or not because it brings the "ether" unit back to the spec again. /cc  @terenc3t @ralexstokes @justinDrake

nisdas:
@hwwhww Is there any reason we shouldn't be using the `ether` unit ?

nisdas:
@hwwhww ah , k , fair enough then. Yeah then it would make sense to keep it in gwei

nisdas:
@hwwhww alright changed it to gwei :)

specs/core/0_beacon-chain.md
```diff
@@ -670,8 +670,8 @@ full_deposit_count: uint256
 @payable
 @public
 def deposit(deposit_input: bytes[2048]):
-    assert msg.value >= as_wei_value(MIN_DEPOSIT_AMOUNT, "gwei")
-    assert msg.value <= as_wei_value(MAX_DEPOSIT_AMOUNT, "gwei")
+    assert msg.value >= MIN_DEPOSIT
+    assert msg.value <= MAX_DEPOSIT
 
     index: uint256 = self.deposit_count + TWO_TO_POWER_OF_TREE_DEPTH
     deposit_amount: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
```
JustinDrake: `msg.value / GWEI_PER_ETH` is only "correct" by chance. It should be `msg.value / WEI_PER_GWEI`.

specs/core/0_beacon-chain.md
```diff
@@ -670,8 +670,8 @@ full_deposit_count: uint256
 @payable
 @public
 def deposit(deposit_input: bytes[2048]):
-    assert msg.value >= as_wei_value(MIN_DEPOSIT_AMOUNT, "gwei")
-    assert msg.value <= as_wei_value(MAX_DEPOSIT_AMOUNT, "gwei")
+    assert msg.value >= MIN_DEPOSIT
```
JustinDrake: In line with using Gwei everywhere, my suggestion is to define `deposit_amount` (e.g. `deposit_amount = msg.value / WEI_PER_GWEI`) and then only deal with Gwei amounts (never `msg.value` directly).

nisdas:
@JustinDrake Alright I changed all references to gwei. Also changed min/max deposits to be represented in terms of gwei

nisdas:
@JustinDrake Hey, so I was testing this contract out and a bug came out, which is referenced here: 
https://github.com/ethereum/vyper/issues/1206

We shouldn't call functions when emitting logs, as it leads to undecodable logs

JustinDrake:
@nisdas Can you try making `to_bytes` `@public` as suggested in https://github.com/ethereum/vyper/issues/1206? (BTW, how can I reproduce your tests?)

nisdas:
@JustinDrake The bug applies to both public and private function calls. This is our current test: 
https://github.com/prysmaticlabs/prysm/blob/ec9465483c739d86ffdc4e04d1d1cd357f45562f/contracts/deposit-contract/depositContract_test.go#L101

And it breaks here: 
https://github.com/prysmaticlabs/prysm/blob/ec9465483c739d86ffdc4e04d1d1cd357f45562f/contracts/deposit-contract/depositContract_test.go#L139
https://github.com/prysmaticlabs/prysm/blob/ec9465483c739d86ffdc4e04d1d1cd357f45562f/contracts/deposit-contract/POWlogs.go#L11

The abi decoder will fail to unpack the log data into the separate log arguments

JustinDrake:
I suggest we wait until Vyper gets fixed https://github.com/ethereum/vyper/pull/1213

nisdas:
@JustinDrake Hey https://github.com/ethereum/vyper/pull/1213 is merged in now, are we waiting on #490, or is this good to be merged in ?

JustinDrake:
@nisdas Shall we wait for the 0.1.0-beta.8 release? Also need to merge with https://github.com/ethereum/eth2.0-specs/pull/490

nisdas:
@JustinDrake Sounds good to me :+1:

specs/core/0_beacon-chain.md
```diff
@@ -667,45 +666,48 @@ deposit_tree: map(uint256, bytes32)
 deposit_count: uint256
 full_deposit_count: uint256
 
+@private
+@constant
+def to_bytes(value: uint256) -> bytes[8]:
```
vbuterin: to_bytes -> to_bytes8

jacqueswww:
With regards to when beta8 will be released, I hope sometime next week, this will also include the bytes32 slice() fix, which one can then inline the `to_bytes` function at a much lower  gas cost (I expect, still have to make the change).

djrtwo:
@nisdas Can you sync this with the latest changes in master?

nisdas:
@djrtwo Alright, done

nisdas:
With #562 being merged in, this PR doesnt make sense now. So I am closing it

Issue #440, opened 2019-01-14
Proposer attestation inclusion reward underspecified


djrtwo:
# Issue
We currently do not prevent aggregate attestations that have overlap in indices from being included across multiple blocks. This is handled in finality, crosslinks, and attester reward but just taking the set of the indices included to prevent double counting. The [reward for the proposers](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#attestation-inclusion) that might have double included is currently underspecified. 

> For each index in previous_epoch_attester_indices, we determine the proposer `proposer_index = get_beacon_proposer_index(state, inclusion_slot(state, index))` and set `state.validator_balances[proposer_index] += base_reward(state, index) // INCLUDER_REWARD_QUOTIENT`

This could be interpreted in a number of ways:
* the _first_ proposer to include the index
* all proposers that included the index
* maybe even, the _last_ proposer to include the index

I suggest we clarify the above language to specify the reward go to the _first_ proposer to include a given index. This will eliminate the incentive to double include for additional reward. The proposer might only double include at this point because they have additional attestations to add but do not have the requisite data to un-aggregate repeat indices from the aggregate attestation.

vbuterin:
Proposed fix. Change:

> Let `inclusion_slot(state, index) = a.slot_included` for the attestation `a` where `index` is in `get_attestation_participants(state, a.data, a.participation_bitfield)`.

To:

> Let `inclusion_slot(state, index) = a.slot_included` for the attestation `a` where `index` is in `get_attestation_participants(state, a.data, a.participation_bitfield)`. If multiple attestations are applicable, the attestation with lowest `slot_included` is considered.

hwwhww:
closing in favor of #464

PR #464, opened 2019-01-17
clarify `inclusion_distance`

djrtwo:
addresses #440

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1641,7 +1641,7 @@ For every `slot in range(state.slot - 2 * EPOCH_LENGTH, state.slot)`, let `cross
 
 Define the following helpers to process attestation inclusion rewards and inclusion distance reward/penalty. For every attestation `a` in `previous_epoch_attestations`:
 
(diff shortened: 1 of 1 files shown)


JustinDrake:
> Let `inclusion_slot(state, index) = a.slot_included`

This is confusing to me. I would avoid implicit arguments to `inclusion_slot` (if indeed there is an implicit argument as suggested by "For every attestation `a` in `previous_epoch_attestations`"). Also, does `a` refer to the above implicit argument, or something else (as suggested by "the attestation `a` where `index` is in `get_attestation_participants(state, a.data, a.aggregation_bitfield)`")?

vbuterin:
> (if indeed there is an implicit argument as suggested by "For every attestation a in previous_epoch_attestations").

It's not an implicit argument because it's part of the `state`. Maybe explicitly state `state.previous_epoch_attestations` to make that clearer?

Maybe change "Let `inclusion_slot(state, index) = a.slot_included`...." to "Let `inclusion_slot(state, index)` be the value `a.slot_included`.....". `a` is defined right after ("for the attestation `a` where..."), but I can see how having `inclusion_slot(state, index) = a.slot_included` as a single "unit" in code-formatted text may make readers confused and wonder where `a` is coming from.

djrtwo:
These helpers at the top are currently where I get the most questions about the spec. I think it could use some love outside of this PR. Merging this and addressing as a whole somewehre else.

Issue #459, opened 2019-01-17
Light client proposal


vbuterin:
See for background: https://github.com/ethereum/eth2.0-specs/issues/403

### Protocol changes

* An activation or exit triggered in slot n now happens at slot `(n - n % EPOCH_LENGTH - 1) + EXIT_ENTRY_DELAY ` (so always pegged to an epoch boundary)
* There is a new entry in the `BeaconState`, `latest_active_index_roots: List[uint64]`, a list of `merkle_tree(get_active_validator_indices(state, block.slot))` of the last 8192 epoch boundaries, maintained similarly to the randao roots (if the indices are not a power of two, we zero-pad the list up to a power of two to make it Merkelizable)
* There is a new entry in the `BeaconState`, `latest_active_validator_set_lengths: List[uint64]`, a list of `len(get_active_validator_set_lengths(state, block.slot))` of the last 8192 epochs, maintained similarly to the randao roots
* To compute the `seed` used in `get_shuffling`, use `seed = sha3(get_randao_mix(slot), get_active_index_root(slot))` where `get_active_index_root` is defined similarly to `get_randao_mix`
* We remove all logic related to the `validator_set_delta_hash_chain`

**Version 1b**: instead of `merkle_tree`, use `tree_hash_root`. The tree hash root has hashed-in metadata about the length of the list, so `latest_active_validator_set_lengths` becomes superfluous and can be removed.

### Light client procedure

Suppose that a light client already authenticated the chain up to some `KNOWN_BLOCK`, that it considers final. As per the requirements of persistent committees, this means that the light client has the information that it needs to predict persistent committees up to ~9 days in the future (see https://github.com/ethereum/eth2.0-specs/issues/375). The light client will ask for a `LightClientSkipProof` object for a future block up to ~9 days in the future, which contains the following objects:

* The `header` of the proposed block
* 128 `validator_index_proofs` (Merkle branches + leaves)
* 128 `validator_record_proofs` (Merkle branches + leaves)
* 128 `validator_balance_proofs` (Merkle branches + leaves)
* N-128 proofs of possession
* A `participation_bitfield` 16 bytes long
* An `aggregate_signature`

The light client privately chooses a `shard` in `[0...SHARD_COUNT-1]`. The light client will use the randao mixes, active index roots and active validator set lengths that it can derive from the state root of `KNOWN_BLOCK` to derive enough information to compute:

[...]

PR #476, opened 2019-01-19
Implement #459

vbuterin:
Contents:

* Peg entries and exits to epoch boundaries
* Add a store of historical active index roots
* Mix it into the randomness
* Remove the delta hash chain

Note that the actual light client implementation is beyond the scope of the spec.

[Note to reviewers: verify that the invariant added in the PR is correct]

Question:

* Do we want to also only store epoch-boundary randao values? I don't think we use the epoch-intermediate ones anywhere.....

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -18,7 +18,6 @@
         - [Reward and penalty quotients](#reward-and-penalty-quotients)
         - [Status flags](#status-flags)
         - [Max operations per block](#max-operations-per-block)
-        - [Validator registry delta flags](#validator-registry-delta-flags)
         - [Signature domains](#signature-domains)
     - [Data structures](#data-structures)
         - [Beacon chain operations](#beacon-chain-operations)
@@ -47,7 +46,6 @@
             - [`Crosslink`](#crosslink)
             - [`PendingAttestation`](#pendingattestation)
             - [`Fork`](#fork)
-            - [`ValidatorRegistryDeltaBlock`](#validatorregistrydeltablock)
             - [`Eth1Data`](#eth1data)
             - [`Eth1DataVote`](#eth1datavote)
     - [Ethereum 1.0 deposit contract](#ethereum-10-deposit-contract)
@@ -73,6 +71,7 @@
             - [`get_crosslink_committees_at_slot`](#get_crosslink_committees_at_slot)
             - [`get_block_root`](#get_block_root)
             - [`get_randao_mix`](#get_randao_mix)
+            - [`get_active_index_root`](#get_active_index_root)
             - [`get_beacon_proposer_index`](#get_beacon_proposer_index)
             - [`merkle_root`](#merkle_root)
             - [`get_attestation_participants`](#get_attestation_participants)
@@ -168,6 +167,7 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted 
 | `MAX_CASPER_VOTES` | `2**10` (= 1,024) | votes |
 | `LATEST_BLOCK_ROOTS_LENGTH` | `2**13` (= 8,192) | block roots |
 | `LATEST_RANDAO_MIXES_LENGTH` | `2**13` (= 8,192) | randao mixes |
+| `LATEST_INDEX_ROOTS_LENGTH` | `2**13` (= 8,192) | index roots |
 | `LATEST_PENALIZED_EXIT_LENGTH` | `2**13` (= 8,192) | epochs | ~36 days |
 | `MAX_WITHDRAWALS_PER_EPOCH` | `2**2` (= 4) | withdrawals |
 
@@ -234,13 +234,6 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted 
 | `MAX_ATTESTATIONS` | `2**7` (= 128) |
 | `MAX_DEPOSITS` | `2**4` (= 16) |
 | `MAX_EXITS` | `2**4` (= 16) |
-
-### Validator registry delta flags
-
-| Name | Value |
-| - | - |
-| `ACTIVATION` | `0` |
-| `EXIT` | `1` |
 
 ### Signature domains
 
@@ -478,7 +471,6 @@ Unless otherwise indicated, code appearing in `this style` is to be interpreted 
     'validator_balances': ['uint64'],
     'validator_registry_update_slot': 'uint64',
     'validator_registry_exit_count': 'uint64',
-    'validator_registry_delta_chain_tip': 'hash32',  # For light clients to track deltas
 
     # Randomness and committees
     'latest_randao_mixes': ['hash32'],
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1382,16 +1378,7 @@ Note: All functions in this section mutate `state`.
...
-    state.validator_registry_delta_chain_tip = hash_tree_root(
-        ValidatorRegistryDeltaBlock(
-            latest_registry_delta_root=state.validator_registry_delta_chain_tip,
-            validator_index=index,
-            pubkey=validator.pubkey,
-            slot=validator.activation_slot,
-            flag=ACTIVATION,
-        )
-    )
+    validator.activation_slot = GENESIS_SLOT if genesis else (state.slot - state.slot % EPOCH_LENGTH + ENTRY_EXIT_DELAY)
```
djrtwo: I'm tempted to add a helper -- `epoch_start_slot(slot)` -- that takes input `slot` and returns `slot - slot % EPOCH_LENGTH`.  We're doing this in many places at this point and should increase readability.

specs/core/0_beacon-chain.md
```diff
@@ -1790,17 +1768,19 @@ and perform the following updates:
 
 * Set `state.previous_epoch_calculation_slot = state.current_epoch_calculation_slot`
 * Set `state.previous_epoch_start_shard = state.current_epoch_start_shard`
-* Set `state.previous_epoch_randao_mix = state.current_epoch_randao_mix`
+* Set `state.previous_epoch_seed = state.current_epoch_seed`
 * Set `state.current_epoch_calculation_slot = state.slot`
 * Set `state.current_epoch_start_shard = (state.current_epoch_start_shard + get_current_epoch_committee_count_per_slot(state) * EPOCH_LENGTH) % SHARD_COUNT`
-* Set `state.current_epoch_randao_mix = get_randao_mix(state, state.current_epoch_calculation_slot - SEED_LOOKAHEAD)`
+* Set `state.current_epoch_seed = hash(get_randao_mix(state, state.current_epoch_calculation_slot - SEED_LOOKAHEAD) + get_active_index_root(state, state.current_epoch_calculation_slot))`
```
djrtwo: I believe the invariant described below is broken here. 

Only in "Final Updates" are we updating the `active_index_root` for the the current `state.slot`. At this point in the execution, we haven't yet set `active_index_root` for `state.slot` but `state.current_epoch_calculation_slot == state.slot` and we are trying to access it. If we want the root of the current active val set, then we need to update `latest_index_roots` _before_ we attempt to access the value.

This should be safe to do at the top of the per-epoch change before the vset change. The vset change can't affect our active vset for this slot, only for `slot + ENTRY_EXIT_DELAY`

specs/core/0_beacon-chain.md
```diff
@@ -1790,17 +1768,19 @@ and perform the following updates:
...
 
 If a validator registry update does _not_ happen do the following:
 
 * Set `state.previous_epoch_calculation_slot = state.current_epoch_calculation_slot`
 * Set `state.previous_epoch_start_shard = state.current_epoch_start_shard`
 * Let `epochs_since_last_registry_change = (state.slot - state.validator_registry_update_slot) // EPOCH_LENGTH`.
-* If `epochs_since_last_registry_change` is an exact power of 2, set `state.current_epoch_calculation_slot = state.slot` and `state.current_epoch_randao_mix = state.latest_randao_mixes[(state.current_epoch_calculation_slot - SEED_LOOKAHEAD) % LATEST_RANDAO_MIXES_LENGTH]`. Note that `state.current_epoch_start_shard` is left unchanged.
+* If `epochs_since_last_registry_change` is an exact power of 2, set `state.current_epoch_calculation_slot = state.slot` and `state.current_epoch_seed = hash(get_randao_mix(state, state.current_epoch_calculation_slot - SEED_LOOKAHEAD) + get_active_index_root(state, state.current_epoch_calculation_slot))`. Note that `state.current_epoch_start_shard` is left unchanged.
```
djrtwo: Same comment as above.

specs/core/0_beacon-chain.md
```diff
@@ -1405,22 +1392,13 @@ def exit_validator(state: BeaconState, index: int) -> None:
     validator = state.validator_registry[index]
 
     # The following updates only occur if not previous exited
-    if validator.exit_slot <= state.slot + ENTRY_EXIT_DELAY:
+    if validator.exit_slot <= state.slot - state.slot % EPOCH_LENGTH + ENTRY_EXIT_DELAY:
```
JustinDrake: As discussed, we should probably round *up* to the nearest epoch boundary (as opposed to round down).

specs/core/0_beacon-chain.md
vbuterin: I just added a helper `entry_exit_effect_slot` which I think even more closely targets the computation that's being repeated a bunch of times.

specs/core/0_beacon-chain.md
```diff
@@ -1376,22 +1384,13 @@ def process_deposit(state: BeaconState,
 
 ### Routines for updating validator status
 
-Note: All functions in this section mutate `state`.
+Note: All functions in this section mutate `state`.
```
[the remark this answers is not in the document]
JustinDrake: Trailing space still there

specs/core/0_beacon-chain.md
```diff
@@ -1170,6 +1171,17 @@ def integer_squareroot(n: int) -> int:
...
 
+#### `entry_exit_effect_slot`
+
+```python
+def entry_exit_effect_slot(n: int) -> int:
+    """
+    An entry or exit triggered in the slot given by the input takes effect at
+    the slot given by the output.
+    """
+    return (n - n % EPOCH_LENGTH) + EPOCH_LENGTH + ENTRY_EXIT_DELAY
```
JustinDrake: Does this work when `n % EPOCH_LENGTH == 0`? In this case `entry_exit_effect_slot(n)` returns `n + EPOCH_LENGTH + ENTRY_EXIT_DELAY` when it should return `n + ENTRY_EXIT_DELAY`.

vbuterin: Is that what it should return? If it gets triggered during block `EPOCH_LENGTH * k`, then that means that the epoch transition for epoch `k` has already been processes, and so we are in epoch `k+1` now, so returning `(EPOCH_LENGTH * k) + EPOCH + ENTRY_EXIT_DELAY` is the _correct_ thing to do.

djrtwo: agreed

djrtwo:
What's the status here? Anything else need eyes on it? are we waiting on the epoch transition pr?

specs/core/0_beacon-chain.md
```diff
@@ -1709,13 +1699,14 @@ def process_ejections(state: BeaconState) -> None:
             exit_validator(state, index)
 ```
 
-### Validator registry
+### Validator registry and shuffling seed data
 
-First, update `previous_epoch_calculation_slot` and `previous_epoch_start_shard`:
+First, update `previous_epoch_calculation_slot`, `previous_epoch_start_shard` and `latest_index_roots`:
```
hwwhww: and + `previous_epoch_seed`.

Maybe just say something like:
```suggestion
First, update the previous epoch records:
```

Issue #409, opened 2019-01-08
get_shard_committees_at_slot out of range


terenc3t:
Not sure If this is a bug, let me use an example to illustrate:

Assume the following:
* `state.slot = 128`
*  in order to get `previous_epoch_attester_indices`, we call `get_attestation_participants` with attestations from slots `0..63`
*  `get_attestation_participants` calls `get_shard_committees_at_slot`, and calculates the following:
`earliest_slot_in_array = state.Slot - (state.Slot % EPOCH_LENGTH) - EPOCH_LENGTH`
 so earliest_slot_in_array = 64
* Then asserts the following `earliest_slot_in_array <= slot < earliest_slot_in_array + EPOCH_LENGTH * 2`

Would assert fail because attestation.data.slot is `0..63`? should the earliest_slot_in_array be?
 `earliest_slot_in_array = state.Slot - (state.Slot % EPOCH_LENGTH) - 2 * EPOCH_LENGTH`

djrtwo:
Yeah, this is a bug. I'll address it in #369 in one way or another

djrtwo:
Closed via #492

PR #480, opened 2019-01-21
epoch processing before block processing

djrtwo:
* Move epoch processing before block processing (resolves https://github.com/ethereum/eth2.0-specs/issues/267#issuecomment-446139113). Epoch transitions now happen at the _beginning_ of an epoch before any blocks are processed from the new epoch. 
* This resolves some issues with the shuffling available in state (#409, #92, #352) by ensuring epoch shuffling is updated before blocks are processed. Also changed the bounds of `get_crosslink_committees_at_slot` to be based upon the `current_` and `previous_epoch_calculation_slot` and to allow for accessing shufflings arbitrarily in the future.

Note: `get_crosslink_committees_at_slot` can access shufflings arbitrarily in the future, but this is not guaranteed to remain stable if a state transition changes the shuffling between `state.slot` and the future slot (validator registry change, or power of 2 reshuffling).

Note: The per-slot transition line "Set state.validator_registry[get_beacon_proposer_index(state, state.slot)].randao_layers += 1" is still broken at the state transition epochs because the shuffling is not yet updated at this point, but I'm leaving it in because it is very likely to be removed after our discussion about removing the hash-onion entirely.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -93,18 +93,6 @@
     - [Per-slot processing](#per-slot-processing)
         - [Misc counters](#misc-counters)
         - [Block roots](#block-roots)
-    - [Per-block processing](#per-block-processing)
-        - [Slot](#slot)
-        - [Proposer signature](#proposer-signature)
-        - [RANDAO](#randao)
-        - [Eth1 data](#eth1-data)
-        - [Operations](#operations)
-            - [Proposer slashings](#proposer-slashings-1)
-            - [Casper slashings](#casper-slashings-1)
-            - [Attestations](#attestations-1)
-            - [Deposits](#deposits-1)
-            - [Exits](#exits-1)
-            - [Custody](#custody)
     - [Per-epoch processing](#per-epoch-processing)
         - [Helpers](#helpers)
         - [Eth1 data](#eth1-data-1)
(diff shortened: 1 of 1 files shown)


djrtwo:
still WIP. I am looking at some of the shuffling problems before review

specs/core/0_beacon-chain.md
```diff
@@ -946,27 +946,26 @@ def get_crosslink_committees_at_slot(state: BeaconState,
...
     """
-    state_epoch_slot = state.slot - (state.slot % EPOCH_LENGTH) 
-    assert state_epoch_slot <= slot + EPOCH_LENGTH
-    assert slot < state_epoch_slot + EPOCH_LENGTH
-    offset = slot % EPOCH_LENGTH
+    assert slot >= state.previous_epoch_calculation_slot
 
-    if slot < state_epoch_slot:
+    if slot < state.current_epoch_calculation_slot:
         committees_per_slot = get_previous_epoch_committee_count_per_slot(state)
```
vbuterin: I'm pretty sure this is wrong, and `state_epoch_slot` is correct. The `current_epoch_calculation_slot` is the slot the data from which is used to calculate the current shuffling, which could be prior to the current slot.

specs/core/0_beacon-chain.md
```diff
@@ -1845,6 +1710,140 @@ def process_penalties_and_exits(state: BeaconState) -> None:
 * Let `e = state.slot // EPOCH_LENGTH`. Set `state.latest_penalized_balances[(e+1) % LATEST_PENALIZED_EXIT_LENGTH] = state.latest_penalized_balances[e % LATEST_PENALIZED_EXIT_LENGTH]`
 * Remove any `attestation` in `state.latest_attestations` such that `attestation.data.slot < state.slot - EPOCH_LENGTH`.
```
vbuterin: What is the rationale for doing these steps at the beginning of the new epoch and now at the end of the old epoch?

specs/core/0_beacon-chain.md
djrtwo: Right and if it is prior, we fall through to the `else` and use the `current_` seed, calc slot, etc.

I think this is essentially doing to same thing but more explicitly uses the slots we stored in state and I think is clearer.

The future slots is not generally safe so I'll add the restriction back in.

djrtwo:
closing in favor of #492

PR #481, opened 2019-01-21
Fix #479, make `item_index == new_index` a check to perform

### LINKED ISSUE #479: SSZ spec missing explicit checks container decoding
Regarding how `container`s are encoded and decoded, I looked at the py implementation (from the spec, not the actual implementation yet). 
```python
class Container:
    # this is the container; here we will define an empty class for demonstration
    pass

# get a reference to the type in some way...
container = Container()
typ = type(container)

assert(len(rawbytes) > current_index + LENGTH_BYTES)
total_length = int.from_bytes(rawbytes[current_index:current_index + LENGTH_BYTES], 'little')
new_index = current_index + LENGTH_BYTES + total_length
assert(len(rawbytes) >= new_index)
item_index = current_index + LENGTH_BYTES

values = {}
for field_name in get_field_names(typ):
    field_name_type = get_type_for_field_name(typ, field_name)
    values[field_name], item_index = deserialize(data, item_index, field_name_type)
assert item_index == new_index
return typ(**values), item_index
```

As I understand it, a `container` must be used if we want to to encode a list of dynamic-size things, e.g. `bytes`, like a list of `transactions`. The encoded form then becomes
`container_total_bytes | length_1 | data_1 | length_2 | data_2 | .. | data_n` . 
The implementation does verify that after decoding all items, the sums match up: `assert item_index == new_index`. However, 

1. This is not listed explicitly as a "Check to perform" in the spec. If this is not done, it would be possible to either overflow into the next element, or underflow so that after reading `n` items, the offset is not correct. Some implementations might choose to go with the total_length, some may choose to use the last `index`, and they'd be incompatible. 
2. The check implemented in python is made post-facto, meaning that if e.g. `length_1` overflows the buffer, there would be an out-of-bounds read, and possible crash. This is probably not a problem, if all lower-level decode-methods always checks bounds, however, an optimized implementation might skip that checking those individual fields if the envelope `container` is already checked.

ChihChengLiang:


specs/simple-serialize.md
```diff
@@ -325,6 +325,7 @@ Instantiate a container with the full set of deserialized data, matching each me
 |:------------------------------------------|:----------------------------------------------------------------|
 | rawbytes has enough left for length       | ``len(rawbytes) > current_index + LENGTH_BYTES``                |
 | list is not greater than serialized bytes | ``len(rawbytes) > current_index + LENGTH_BYTES + total_length`` |
+| item index is at end of serialized bytes  | ``item_index == current_index + LENGTH_BYTES + total_length``   |
```
holiman: Maybe also add something like

If the container has a fixed number of elements, verify that the number of items read matches with the expected number of elements. 

It may be a design-choice to allow extending types, and that a decoder which receives a type which has more fields than the definition should be forward-compatible? If so, that should clearly be specified.

ChihChengLiang: > If the container has a fixed number of elements, verify that the number of items read matches with the expected number of elements.
This can be only verified with item index? Let's say if the container has 2 expected elements and the raw_bytes has more, the former reads only 2 elements and is ignorant of whether the later has 1 or 2 more elements.

> A design-choice to allow extending types, and that a decoder which receives a type which has more fields than the definition should be forward-compatible?
cc @mratsim, would like to consult your expertise on this.

specs/simple-serialize.md
```diff
@@ -325,6 +326,7 @@ Instantiate a container with the full set of deserialized data, matching each me
 |:------------------------------------------|:----------------------------------------------------------------|
 | rawbytes has enough left for length       | ``len(rawbytes) > current_index + LENGTH_BYTES``                |
 | list is not greater than serialized bytes | ``len(rawbytes) > current_index + LENGTH_BYTES + total_length`` |
+| item index is at end of serialized bytes  | ``item_index == current_index + LENGTH_BYTES + total_length``   |
```
jannikluhn: This should probably checked for all types, not only for containers. So I'd not put it here, but at the beginning of the `Deserialize/Decode` section.

specs/simple-serialize.md
jannikluhn: I agree with 

> If the container has a fixed number of elements, verify that the number of items read matches with the expected number of elements.

but without the constraint on containers with a fixed number of elements (as all containers have a fixed number of elements).

> A design-choice to allow extending types, and that a decoder which receives a type which has more fields than the definition should be forward-compatible?

I don't think this intended in the current version of the spec, nor do I think it would be a good idea (even though I'd support some kind of versioning, but only if it's done explicitly).

holiman: > but without the constraint on containers with a fixed number of elements (as all containers have a fixed number of elements).

So, that's another question I have. Isn't this datatype also meant to encapsulate e.g. a list of transactions? That is, a non-fixed-length list of items which are dynamic-length (length-prefixed).

Because, 
- Either the docs should be updated to indicate that `containers` are used for both "variable-length lists of variable-length types` and also `fixed-length lists of arbitrary types`
- Or, the ssz spec needs another section to explain how to encode/decode such lists, as they most definitely will be needed, imo.

holiman: Oh, I see. I was misled by this comment on the page you link to:

> 1, Get the number of raw bytes to serialize: it is len(list) * sizeof(element)

Because that implies that all `element`s are of identical fixed size. So that's wrong then?

jannikluhn: Yes, that must be wrong. There's an [issue](https://github.com/ethereum/eth2.0-specs/issues/232) for that, but it hasn't been fixed yet. I opened a [PR](https://github.com/ethereum/eth2.0-specs/pull/484) for this yesterday.

specs/simple-serialize.md
```diff
@@ -326,6 +334,7 @@ Instantiate a container with the full set of deserialized data, matching each me
 | rawbytes has enough left for length       | ``len(rawbytes) > current_index + LENGTH_BYTES``                |
 | list is not greater than serialized bytes | ``len(rawbytes) > current_index + LENGTH_BYTES + total_length`` |
 
+
```
hwwhww: unnecessary newline

specs/simple-serialize.md
```diff
@@ -326,6 +334,7 @@ Instantiate a container with the full set of deserialized data, matching each me
 | rawbytes has enough left for length       | ``len(rawbytes) > current_index + LENGTH_BYTES``                |
```
hwwhww: Could you add ``code block`` to all the variable in the tables? e.g.:
```suggestion
| ``rawbytes`` has enough left for length       | ``len(rawbytes) > current_index + LENGTH_BYTES``                |
```

specs/simple-serialize.md
```diff
@@ -219,14 +219,22 @@ The decoding requires knowledge of the type of the item to be decoded. When
 performing decoding on an entire serialized string, it also requires knowledge
 of the order in which the objects have been serialized.
 
-Note: Each return will provide ``deserialized_object, new_index`` keeping track
-of the new index.
+Note: Each return will provide:
+- `deserialized_object`
+- `new_index`
```
hwwhww: 1. Could you add the definition of `deserialized_object` and `new_index` in `Terminology` section?
2. And actually, maybe `Terminology` should be rename to `Variables and Functions`...
2. Also, rename `byte_order` to `byteorder`.

hwwhww:
@holiman Thank you so much!

PR #482, opened 2019-01-21
clarify get_shuffling invariant

djrtwo:
`get_shuffling` invariant does not work for slot's arbitrarily in the future. Add slight modification of condition to clean up invariant

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -912,7 +912,7 @@ def get_shuffling(seed: Bytes32,
     return split(shuffled_active_validator_indices, committees_per_slot * EPOCH_LENGTH)
 ```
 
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -912,7 +912,7 @@ def get_shuffling(seed: Bytes32,
     return split(shuffled_active_validator_indices, committees_per_slot * EPOCH_LENGTH)
 ```
 
-**Invariant**: if `get_shuffling(seed, validators, slot)` returns some value `x`, it should return the same value `x` for the same `seed` and `slot` and possible future modifications of `validators` forever in phase 0, and until the ~1 year deletion delay in phase 2 and in the future.
+**Invariant**: if `get_shuffling(seed, validators, slot)` returns some value `x` and `slot <= state.slot + ENTRY_EXIT_DELAY`, it should return the same value `x` for the same `seed` and `slot` and possible future modifications of `validators` forever in phase 0, and until the ~1 year deletion delay in phase 2 and in the future.
```
vbuterin: I'd go with "for some `slot <= state.slot + ENTRY_EXIT_DELAY`" rather than "and". But looks good in either case.

djrtwo: updated! can I get a pr approval for merge?

PR #483, opened 2019-01-21
BLS-based RANDAO and custody (friendly to decentralised pools)

JustinDrake:
Unlock decentralised pools using m-of-n threshold BLS signatures for both RANDAO and custody. We also simplify a bunch:

* Remove `randao_commitment` and `custody_commitment`
* Remove miscellaneous logic such as `repeat_hash`

(Side note: Dfinity seems to be working to reduce the communication complexity of BLS DKG (Distributed Key Generation) so validator pools should also benefit from that.)

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -250,6 +250,7 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
 | `DOMAIN_ATTESTATION` | `1` |
 | `DOMAIN_PROPOSAL` | `2` |
 | `DOMAIN_EXIT` | `3` |
+| `DOMAIN_RANDAO` | `4` |
 
 ## Data structures
 
@@ -389,10 +390,6 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
     'pubkey': 'bytes48',
     # Withdrawal credentials
     'withdrawal_credentials': 'bytes32',
-    # Initial RANDAO commitment
-    'randao_commitment': 'bytes32',
-    # Initial custody commitment
-    'custody_commitment': 'bytes32',
     # A BLS signature of this `DepositInput`
     'proof_of_possession': 'bytes96',
 }
@@ -423,7 +420,7 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
     'slot': 'uint64',
     'parent_root': 'bytes32',
     'state_root': 'bytes32',
-    'randao_reveal': 'bytes32',
+    'randao_reveal': 'bytes96',
     'eth1_data': Eth1Data,
     'signature': 'bytes96',
 
@@ -520,8 +517,6 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
     'pubkey': 'bytes48',
     # Withdrawal credentials
     'withdrawal_credentials': 'bytes32',
-    # RANDAO commitment
-    'randao_commitment': 'bytes32',
     # Slots the proposer has skipped (i.e. layers of RANDAO expected)
     'randao_layers': 'uint64',
     # Slot when validator activated
@@ -536,8 +531,6 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
     'exit_count': 'uint64',
     # Status flags
     'status_flags': 'uint64',
-    # Custody commitment
-    'custody_commitment': 'bytes32',
     # Slot of latest custody reseed
     'latest_custody_reseed_slot': 'uint64',
     # Slot of second-latest custody reseed
@@ -946,7 +939,7 @@ def get_crosslink_committees_at_slot(state: BeaconState,
     """
     Returns the list of ``(committee, shard)`` tuples for the ``slot``.
     """
-    state_epoch_slot = state.slot - (state.slot % EPOCH_LENGTH) 
+    state_epoch_slot = state.slot - (state.slot % EPOCH_LENGTH)
     assert state_epoch_slot <= slot + EPOCH_LENGTH
     assert slot < state_epoch_slot + EPOCH_LENGTH
     offset = slot % EPOCH_LENGTH
@@ -1195,7 +1188,7 @@ A valid block with slot `GENESIS_SLOT` (a "genesis block") has the following val
     slot=GENESIS_SLOT,
     parent_root=ZERO_HASH,
     state_root=STARTUP_STATE_ROOT,
-    randao_reveal=ZERO_HASH,
+    randao_reveal=EMPTY_SIGNATURE,
     eth1_data=Eth1Data(
         deposit_root=ZERO_HASH,
         block_hash=ZERO_HASH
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1474,12 +1455,9 @@ Below are the processing steps that happen at every `block`.
 
 ### RANDAO
 
-* Let `repeat_hash(x, n) = x if n == 0 else repeat_hash(hash(x), n-1)`.
 * Let `proposer = state.validator_registry[get_beacon_proposer_index(state, state.slot)]`.
-* Verify that `repeat_hash(block.randao_reveal, proposer.randao_layers) == proposer.randao_commitment`.
+* Verify that `bls_verify(pubkey=proposer.pubkey, message=proposer.randao_layers, signature=block.randao_reveal, domain=get_domain(state.fork, state.slot, DOMAIN_RANDAO))`.
 * Set `state.latest_randao_mixes[state.slot % LATEST_RANDAO_MIXES_LENGTH] = hash(xor(state.latest_randao_mixes[state.slot % LATEST_RANDAO_MIXES_LENGTH], block.randao_reveal))`
```
djrtwo: Need to truncate the `randao_reveal` to 32 bytes by either doing `randao_reveal[0:32]` or `hash(randao_reveal)`. The former is cheaper and sufficient

specs/core/0_beacon-chain.md
```diff
@@ -1856,10 +1832,10 @@ This section is divided into Normative and Informative references.  Normative re
 ## Normative
 
 ## Informative
-<a id="ref-casper-ffg"></a> _**casper-ffg**_  
+<a id="ref-casper-ffg"></a> _**casper-ffg**_
```
djrtwo: These spaces are for formatting.

specs/core/0_beacon-chain.md
```diff
@@ -1856,10 +1832,10 @@ This section is divided into Normative and Informative references.  Normative re
...
  &nbsp; _Casper the Friendly Finality Gadget_. V. Buterin and V. Griffith. URL: https://arxiv.org/abs/1710.09437
 
-<a id="ref-python-poc"></a> _**python-poc**_  
+<a id="ref-python-poc"></a> _**python-poc**_
```
djrtwo: Same

specs/core/0_beacon-chain.md
JustinDrake: Having trailing spaces is not sustainable, partly because some editors (such as mine) as programmed to remove trailing spaces. Would it be sufficient to re-add the space and remove the following new line?

specs/core/0_beacon-chain.md
JustinDrake: I think this is fine :) The `randao_reveal` is a signature.

specs/core/0_beacon-chain.md
djrtwo: I added a `</br>` tag. It is much easier to mimic and will not lead to space cleaning editor modifications.

specs/core/0_beacon-chain.md
djrtwo: And is thus 96 bytes. Do we want the `latest_randao_mixes` to store 96 bytes per item rather than a `bytes32`?
`bytes32` is sufficient.

If we want `bytes96`, we need to fix the initialization of this array and make sure that the usage in shuffling of a 96 byte seed is fine. I'd lean toward truncating

JustinDrake: Oh I see this comment is about line 1460 (as opposed to line 1459). Still, `state.latest_randao_mixes` is storing a 32-byte `hash`, no? (As a side note, `randao_reveal[0:32]` isn't great because the top bits are flags which are not random.)

vbuterin: There's still a 32 byte thing being xored with a 96 byte thing. Personally I would just remove the xor and do `state.latest_randao_mixes[state.slot % LATEST_RANDAO_MIXES_LENGTH] = hash(state.latest_randao_mixes[state.slot % LATEST_RANDAO_MIXES_LENGTH] + block.randao_reveal)`

vbuterin:
Aside from the 32 vs 96 bytes xor thing which seems easy to fix, everything looks good to me!

specs/core/0_beacon-chain.md
djrtwo: `hash` of the concatenation looks good.
Otherwise take the last 32 bytes (instead of the first with the flags) and xor.

CarlBeek:
I'm a big fan of these changes! Definitely makes it much more d-pool friendly.  I think there still may be some other remaining challenges regarding custody though. (I'll elaborate once I properly understand the custody workings.) Am I being ridiculous or is there some custody logic missing? What is the secret that is being xored into the data and where is it updated?

JustinDrake:
> there still may be some other remaining challenges regarding custody

For custody we will also replace the hash-based commit-reveal with BLS. (Hence why `custody_commitment` goes away.)

> Am I being ridiculous or is there some custody logic missing?

That's phase 1! We do have placeholders for phase 0, but it's possible even these will go away.

> What is the secret that is being xored into the data and where is it updated?

The secret would be the BLS signature of `randao_layers` (which I will rename right now to something more generic), with a separate `DOMAIN_CUSTODY` signature domain.

CarlBeek:
> For custody we will also replace the hash-based commit-reveal with BLS. (Hence why custody_commitment goes away.)

I thought as much. My other issues are pertinent then. Basically, nothing is stopping anyone in a d-pool leaking the threshold-signature during the custody period. It is an issue inherent to committing the Merkle-root. I've had some ideas to combat this based on using a homomorphic Fiat-Shamir scheme, but I need to develop these a bit further.

>That's phase 1! We do have placeholders for phase 0, but it's possible even these will go away.

Aah, I see that makes sense.

> The secret would be the BLS signature of `randao_layers` (which I will rename right now to something more generic), with a separate DOMAIN_CUSTODY signature domain.

I very much like the idea of using `randao_layers` as the message.

JustinDrake:
> nothing is stopping anyone in a d-pool leaking the threshold-signature during the custody period

Aha, that's a good point. I guess the good news is that we have a bit more time to research custody. One possible approach is that custody be done trustlessly by a designated pool member who proves (to other pool members) correctness of the custody bit using a SNARK.

CarlBeek:
> I guess the good news is that we have a bit more time to research custody. 
Definitely.
> One possible approach is that custody be done trustlessly by a designated pool member who proves (to other pool members) correctness of the custody bit using a SNARK.
Absolutely, but that still is a centralisation vector as a pool is unable to operate sans this designated member. Furthermore, this member still has the ability to slash the entire pool (they could be specifically disincentivised against this though). To quote Kanye: "No one man should have all that power."

JustinDrake:
@CarlBeek: One way of making custody fully MPC-friendly is to replace "xor the leaves with a secret" with "BLS sign the leaves with an ephemeral key", and replace "reveal the secret" with "reveal the ephemeral private key". There are two m-of-n MPCs: one to sign, and a separate one to reconstruct the private key which is done after the timeout to not reveal has elapsed. What do you think?

Issue #232, opened 2018-12-04
SSZ question: clarification on the homogeneity of list elements


cleishm:
In the current [SSZ specification](https://github.com/ethereum/eth2.0-specs/blob/ff2551e754028260a619c1b538ac194568f1eff2/specs/simple-serialize.md#listvectors), lists are defined as "a collection of elements of the same homogeneous **type**".

By this definition, a list of byte arrays, a list of other lists, or a list of collections, are all homogeneous by *type* yet contain elements that may vary in serialized size.

The example implementations sum each elements serialized length to find the raw length, however the specification says the raw length is defined as `len(list) * sizeof(element)`, implying they all have the same serialized size. Additionally, the Tree Hash definition only uses the size of the first element in a list to calculate `items_per_chunk` (`SSZ_CHUNK_SIZE // len(lst[0])`).

The question is: should the specification define that lists contain elements of homogeneous type **and size**, or should we update the specification of raw length (e.g. to `sum(sizeof(elem) for elem in list)`) and the Tree Hash algorithm according?

cleishm:
I also wonder if an "easier" way to resolve this would be to change the list length prefix to _only_ `len(list)` (nElements), and not the raw length of the entire list encoding. This would simplify serialization, sacrificing only the ability to skip over a complete value when decoding, which is not a common use case afaik.

Thoughts?

jannikluhn:
I _think_ the spec is wrong and the example implementation is right. Otherwise lists of containers containing lists would not be possible, but those seem to be used in the phase 0 spec (e.g. `BeaconBlockBody.attestations`).

cleishm:
Yeah, I agree. Though that means the Tree Hash algorithm is wrong too.

Wdyt about just using the element count?

mkalinin:
> Wdyt about just using the element count?

When there is a need of decoding an object that have come from network it's good to know _whole size_ of encoded object. Otherwise, you have to pull list items in sequential fashion and calculate the size to mitigate overflows.

cleishm:
> you have to pull list items in sequential fashion and calculate the size to mitigate overflows

You’re going to do that anyway though, right? Unless you’re pulling the value and _not_ decoding it, which I can’t think of a use-case for, then you’ll be iterating and decoding the length of each part. And currently you have to iterate until you find the value whose end aligns with the overall raw length (or exceeds it, which would signal an encoding error).

mkalinin:
[...]

PR #484, opened 2019-01-22
SSZ: Lists of elements with non-constant element size

jannikluhn:
As pointed out a while ago in #232, the specification of list serialization fails if the element size is variable (even though the example implementation is correct). This PR fixes this by first serializing the elements and only then calculating the length and the prefix as I assume was the original intention.

This also makes the "note on efficiency" below irrelevant because knowing the number of elements does not help anymore. Therefore I removed it.


holiman:
Ah, didn't see this until now. The change makes sense

djrtwo:
Looks good. Can you handle merge conflict @jannikluhn ?

jannikluhn:
Resolved the merge conflicts

Issue #461, opened 2019-01-17
`MAX_CASPER_VOTES` clarification


djrtwo:
# Issue
`MAX_CASPER_VOTES` is a constant in an attempt to put a limit on the load of processing attestations within the context of checking slashing conditions. Right now there is a limit that the sum of the number of indices across the two slashable messages cannot exceed `MAX_CASPER_VOTES`.

To clarify -- This is not some sort of in protocol limit on the size of committees, but instead a limit on the size of the bulk slashing that can be performed in one message. If for example 1200 people signed the same slashable messages, multiple `CASPER_SLASHING` messages would have to be submit utilizing different subsets (and thus different aggregate signatures).

This is a bit unclear in the current spec and with the current names. I suggested we
* Change `MAX_CASPER_VOTES` to `MAX_INDICES_IN_CASPER_SLASHING` (a bit verbose but something more descriptive)
* Add a note to the validator guide on how to handle this case

djrtwo:
cc: @JustinDrake  to ensure I have the purpose of this constant correct

djrtwo:
I think this is related to #271

JustinDrake:
Yep, related to #271. Having a constant is possibly bad. Aggregated signatures that are exposed onchain cannot be unaggregated so with large committees the slashing would rely on more fine-grained aggregation to be cached offchain. One possibility is to cap the number of constituent signatures in the aggregated signatures that go onchain.

djrtwo:
Which would force splitting committee sigs across multiple attestations (on chain) when committee size (`total_val / SHARD_COUNT`) exceeds this constant.

hmm

vbuterin:
We can always set the constant to be >= 4096 so that it's always possible to slash a single committee. Anything wrong with that?

JustinDrake:
> We can always set the constant to be >= 4096

`MAX_CASPER_VOTES = 4096` is definitely simple. The worry would be "vote stuffing" DoS attacks where bogus validator indices are included in Casper slashing transactions. In the worst case:

* **CPU**: `MAX_CASPER_SLASHINGS * MAX_CASPER_VOTES` pubkeys need to be retrieved from memory and added. One benchmark suggest the overhead per addition is about 1,350ns so the ballpark is ~100ms.
* **Data**: Each validator index is 3 bytes so an attacker can create a beacon block with ~197kB of validator indices.

vbuterin:
[...]

PR #488, opened 2019-01-22
Reduce and revamp custody placeholder logic

JustinDrake:
Remove all custody and VDF placeholders (56 lines less!). A few notes on the placeholder fields and logic:

* All placeholder fields were dummy fields that can easily be restored in the phase 1 hard fork.
* One special case to the above is `custody_bit_0_validator_indices` in `SlashableVoteData` which was renamed to `validator_indices`. Renaming it back is *not* a spec change because SSZ field names are [no longer part of the spec](https://github.com/ethereum/eth2.0-specs/commit/a9328157a87451ee4f372df272ece158b386ec41#diff-8d8fe480a35579c7be2f976d9b321216).
* The placeholder logic was written using generalised functions (e.g. `bls_verify_multiple` vs `bls_verify`, and `indices(slashable_vote_data)` vs `slashable_vote_data.validator_indices`). This generality was unnecessary because it was not triggered when all custody bits were 0. This means we can simplify the logic without being inconsistent with phase 1.

Rationale:

* Keep phase 0 (likely far harder to deliver than phase 1) as clean as possible
* Focus on upgrade paths and incremental releases
* Custody is still under research—keep the design space open

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1109,8 +1109,7 @@ def verify_bitfield(bitfield: bytes, size: int) -> bool:
 
 ```python
 def verify_slashable_vote_data(state: BeaconState, slashable_vote_data: SlashableVoteData) -> bool:
-    # [TO BE REMOVED IN PHASE 1]
-    if len(slashable_vote_data.custody_bit_1_indices) > 0:
+    if len(slashable_vote_data.custody_bit_1_indices) > 0:  # [TO BE REMOVED IN PHASE 1]
         return False
 
     if len(indices(slashable_vote_data)) > MAX_CASPER_VOTES:
@@ -1532,6 +1531,7 @@ For each `attestation` in `block.body.attestations`:
     assert verify_bitfield(attestation.aggregation_bitfield)
     assert verify_bitfield(attestation.custody_bitfield)
     assert attestation.custody_bitfield & attestation.aggregation_bitfield == attestation.custody_bitfield
+    assert attestation.custody_bitfield == 0 # [TO BE REMOVED IN PHASE 1]
 ```
 
 * Verify aggregate signature:
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -261,7 +261,7 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
 ```python
 {
     # Proposer index
-    'proposer_index': 'uint24',
+    'proposer_index': 'uint64',
     # First proposal data
     'proposal_data_1': ProposalSignedData,
     # First proposal signature
@@ -291,7 +291,7 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
 ```python
 {
     # Validator indices
-    'validator_indices': '[uint24]',
+    'validator_indices': '[uint64]',
     # Custody bitfield
     'custody_bitfield': 'bytes',
     # Attestation data
@@ -402,7 +402,7 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
     # Minimum slot for processing exit
     'slot': 'uint64',
     # Index of the exiting validator
-    'validator_index': 'uint24',
+    'validator_index': 'uint64',
     # Validator signature
     'signature': 'bytes96',
 }
@@ -588,15 +588,15 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
 
 We define the following Python custom types for type hinting and readability:
 
-| Name | Type | Description |
+| Name | SSZ equivalent | Description |
 | - | - | - |
-| `SlotNumber` | unsigned 64-bit integer | the number of a slot |
-| `ShardNumber` | unsigned 64-bit integer | the number of a shard |
-| `ValidatorIndex` | unsigned 24-bit integer | the index number of a validator in the registry |
-| `Gwei` | unsigned 64-bit integer | an amount in Gwei |
-| `Bytes32` | 32-byte data | binary data with 32-byte length |
-| `BLSPubkey` | 48-byte data | a public key in BLS signature scheme |
-| `BLSSignature` | 96-byte data | a signature in BLS signature scheme |
+| `SlotNumber` | `uint64` | a slot number |
+| `ShardNumber` | `uint64` | a shard number |
+| `ValidatorIndex` | `uint64` | an index in the validator registry |
+| `Gwei` | `uint64` | an amount in Gwei |
+| `Bytes32` | `bytes32` | 32 bytes of binary data |
+| `BLSPubkey` | `bytes48` | a BLS public key |
+| `BLSSignature` | `bytes96` | a BLS signature |
 
 ## Ethereum 1.0 deposit contract
 
@@ -1058,7 +1058,7 @@ def get_attestation_participants(state: BeaconState,
     participants = []
     for i, validator_index in enumerate(crosslink_committee):
         aggregation_bit = get_bitfield_bit(bitfield, i)
-        if aggregation_bit == 1:
+        if aggregation_bit == 0b1:
             participants.append(validator_index)
     return participants
 ```
@@ -1121,7 +1121,7 @@ def verify_bitfield(bitfield: bytes, committee_size: int) -> bool:
         return False
 
     for i in range(committee_size + 1, committee_size - committee_size % 8 + 8):
-        if get_bitfield_bit(bitfield, i) != 0:
+        if get_bitfield_bit(bitfield, i) == 0b1:
             return False
 
     return True
@@ -1142,7 +1142,7 @@ def verify_slashable_vote(state: BeaconState, slashable_vote: SlashableVote) -> 
 
     for i in range(len(slashable_vote.validator_indices) - 1):
         if slashable_vote.validator_indices[i] >= slashable_vote.validator_indices[i + 1]:
-        return False
+            return False
 
     if not verify_bitfield(slashable_vote.custody_bitfield, len(slashable_vote.validator_indices)):
         return False
@@ -1153,7 +1153,7 @@ def verify_slashable_vote(state: BeaconState, slashable_vote: SlashableVote) -> 
     custody_bit_0_indices = []
     custody_bit_1_indices = []
     for i, validator_index in enumerate(slashable_vote.validator_indices):
-        if get_bitfield_bit(slashable_vote.custody_bitfield, i) == 0:
+        if get_bitfield_bit(slashable_vote.custody_bitfield, i) == 0b0:
             custody_bit_0_indices.append(validator_index)
         else:
             custody_bit_1_indices.append(validator_index)
@@ -1164,8 +1164,8 @@ def verify_slashable_vote(state: BeaconState, slashable_vote: SlashableVote) -> 
             bls_aggregate_pubkeys([state.validator_registry[i].pubkey for i in custody_bit_1_indices]),
         ],
         messages=[
-            hash_tree_root(AttestationDataAndCustodyBit(slashable_vote.data, 0)),
-            hash_tree_root(AttestationDataAndCustodyBit(slashable_vote.data, 1)),
+            hash_tree_root(AttestationDataAndCustodyBit(attestation_data=slashable_vote.data, custody_bit=0b0)),
+            hash_tree_root(AttestationDataAndCustodyBit(attestation_data=slashable_vote.data, custody_bit=0b1)),
         ],
         signature=slashable_vote.aggregate_signature,
         domain=get_domain(state.fork, slashable_vote.data.slot, DOMAIN_ATTESTATION),
@@ -1554,15 +1554,15 @@ For each `attestation` in `block.body.attestations`:
 * Verify bitfields and aggregate signature:
 
 ```python
-    assert attestation.custody_bitfield == 0  # [TO BE REMOVED IN PHASE 1]
+    assert attestation.custody_bitfield == b'\x00' * len(attestation.custody_bitfield)  # [TO BE REMOVED IN PHASE 1]
 
     for i in range(len(crosslink_committee)):
-        if get_bitfield_bit(attestation.aggregation_bitfield) == 0:
-            assert get_bitfield_bit(attestation.custody_bitfield) == 0
+        if get_bitfield_bit(attestation.aggregation_bitfield) == 0b0:
+            assert get_bitfield_bit(attestation.custody_bitfield) == 0b0
 
     participants = get_attestation_participants(state, attestation.data, attestation.aggregation_bitfield)
-    custody_bit_0_participants = get_attestation_participants(state, attestation.data, attestation.custody_bitfield)
-    custody_bit_1_participants = [i in participants for i not in custody_bit_0_participants]
+    custody_bit_1_participants = get_attestation_participants(state, attestation.data, attestation.custody_bitfield)
+    custody_bit_0_participants = [i in participants for i not in custody_bit_1_participants]
 
     assert bls_verify_multiple(
         pubkeys=[
@@ -1570,8 +1570,8 @@ For each `attestation` in `block.body.attestations`:
             bls_aggregate_pubkeys([state.validator_registry[i].pubkey for i in custody_bit_1_participants]),
         ],
         messages=[
-            hash_tree_root(AttestationDataAndCustodyBit(data=attestation.data, custody_bit=0)),
-            hash_tree_root(AttestationDataAndCustodyBit(data=attestation.data, custody_bit=1)),
+            hash_tree_root(AttestationDataAndCustodyBit(data=attestation.data, custody_bit=0b0)),
+            hash_tree_root(AttestationDataAndCustodyBit(data=attestation.data, custody_bit=0b1)),
         ],
         signature=attestation.aggregate_signature,
         domain=get_domain(state.fork, attestation.data.slot, DOMAIN_ATTESTATION),
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -302,10 +302,10 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
 
 ```python
 {
-    # Validator indices with custody bit equal to 0
-    'custody_bit_0_indices': ['uint24'],
-    # Validator indices with custody bit equal to 1
-    'custody_bit_1_indices': ['uint24'],
+    # Validator indices
+    'validator_indices': '[uint24]',
+    # Custody bitfield
+    'custody_bitfield': 'bytes',
     # Attestation data
     'data': AttestationData,
     # Aggregate signature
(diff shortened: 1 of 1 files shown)


vbuterin:
This feels a bit extreme to me. We're already getting feedback from developers that we're imposing too much change on them, and all this accomplishes is that it forces them to remove code that they're already written and then bring it back six months later.

Removing `CustodyReseed`, `CustodyChallenge`, and `CustodyResponse`, 'latest_vdf_outputs', and custody reseed slots seems totally fine because there was actually do still have some uncertainty.

JustinDrake:
> Removing `AttestationDataAndCustodyBit` makes it so that validators might have to sign an entirely different data structure at phase1 rather than just modifying fields.

Oh I see what you mean (and maybe Vitalik is making the same point?). The signing logic would be equivalent (modulo SSZ field additions) if it wasn't for the nesting discrepancy between `AttestationData` (1-layer deep) and `AttestationDataAndCustodyBit` (2-layer deep).

I've restored the nesting with `AttestationDataWrapped`. (A possibly cleaner alternative which avoids nesting is to define SSZ notation to add fields to an existing object.) I think the two spec versions are now isomorphic modulus SSZ field additions:

* Rename `validator_indices` to `custody_bit_0_indices` in `SlashableVoteData` (no-op)
* Add `custody_bit_0_indices` to `SlashableVoteData` (I moved indices to the end of `SlashableVoteData`)
* Add `custody_bitfield` to `Attestation`
* Rename `AttestationDataWrapped` to `AttestationDataAndCustodyBit` (no-op)
* Add `custody_bit` to `AttestationDataAndCustodyBit`
* Add `custody_reseeds`, `custody_challenges`, `custody_responses` to `BeaconBlockBody`
* Add `latest_vdf_outputs`, `custody_challenges`, `latest_custody_reseed_slot`, `penultimate_custody_reseed_slot` to `BeaconState`
* Add `custody_bitfield` to `PendingAttestation`

vbuterin:
I'd say that's still _too_ deep. It would still require awkward transition logic in phase 1, eg. attestations could be created during phase 0 but included during phase 1, we'd have to keep the `AttestationDataWrapped` and `AttestationDataAndCustodyBit` classes around...

What would it look like if we removed the PoC-specific stuff from only `State` and `Block` ? Benefits as I see them:

* No one needs to un-implement and then re-implement code
* The parts of the PoC that are stable and established (namely, that everyone signs an additional user-specific bit) remain, the parts that are under flux (how challenges and responses and commitments work) are cut, the parts that are not coming back anyway (randaos) are cut
* Removes the ugliest thing about the status quo, which is the need to have empty arrays of a placeholder class of which we don't know the current form yet

JustinDrake:
Fair enough :) I've reverted as you suggested and fixed a bunch of fixes (please check!):

**Bugs**

* Add missing `'` for `'data'` and `'custody_bit'`.
* Perform validation on all bitfields (notice the validation in `verify_bitfield` is stricter than what was in `get_attestation_participants`).
* Add two checks forcing custody bits to be 0 to be removed in phase 1.

**Cleanups**:

* Make `indices` into a helper
* Introduce helpers `get_bitfield_bit` and `verify_bitfield`
* Introduce contants `CUSTODY_BIT_ZERO` and `CUSTODY_BIT_ONE`
* Write up the generalised logic checking `attestation.aggregate_signature`

specs/core/0_beacon-chain.md
```diff
@@ -1489,7 +1507,6 @@ For each `casper_slashing` in `block.body.casper_slashings`:
 
 * Let `slashable_vote_data_1 = casper_slashing.slashable_vote_data_1`.
 * Let `slashable_vote_data_2 = casper_slashing.slashable_vote_data_2`.
-* Let `indices(slashable_vote_data) = slashable_vote_data.custody_bit_0_indices + slashable_vote_data.custody_bit_1_indices`.
```
JustinDrake: One edge case here is that the same index can be contained in both `custody_bit_0_indices` and `custody_bit_1_indices`. I don't think anything breaks but wanted to double check.

On second thought, I think there is something cleaner than `custody_bit_0_indices` and `custody_bit_1_indices`. We have instead `validator_indices` (sorted for compression, as discussed in the Telegram chat) and an accompanying `custody_bitfield` (corresponding to the sorted indices). This is both more consistent with the rest of the spec, and removes the edge case above.

@djrtwo, @vbuterin.

vbuterin: Yeah, I would support that.

specs/core/0_beacon-chain.md
```diff
@@ -1088,28 +1083,50 @@ def get_domain(fork: Fork,
...
-            bls_aggregate_pubkeys([state.validator_registry[i].pubkey for i in vote_data.custody_bit_0_indices]),
-            bls_aggregate_pubkeys([state.validator_registry[i].pubkey for i in vote_data.custody_bit_1_indices]),
+            bls_aggregate_pubkeys([state.validator_registry[i].pubkey for i in slashable_vote_data.custody_bit_0_indices]),
+            bls_aggregate_pubkeys([state.validator_registry[i].pubkey for i in slashable_vote_data.custody_bit_1_indices]),
         ],
         messages=[
-            hash_tree_root(AttestationDataAndCustodyBit(vote_data.data, False)),
-            hash_tree_root(AttestationDataAndCustodyBit(vote_data.data, True)),
+            hash_tree_root(AttestationDataAndCustodyBit(slashable_vote_data.data, CUSTODY_BIT_ZERO)),
+            hash_tree_root(AttestationDataAndCustodyBit(slashable_vote_data.data, CUSTODY_BIT_ONE)),
```
vbuterin: Is this really the cleanest way to express that zero and one are the custody bits? We could do it without adding new constants by doing `AttestationDataAndCustodyBit(data=slashable_vote_data.data, custody_bit=0)`; that way it requires the reader to hop around the spec less. I'm ultimately ok either way though.

specs/core/0_beacon-chain.md
```diff
@@ -1509,10 +1527,36 @@ For each `attestation` in `block.body.attestations`:
...
-* `aggregate_signature` verification:
-    * Let `participants = get_attestation_participants(state, attestation.data, attestation.aggregation_bitfield)`.
-    * Let `group_public_key = bls_aggregate_pubkeys([state.validator_registry[v].pubkey for v in participants])`.
-    * Verify that `bls_verify(pubkey=group_public_key, message=hash_tree_root(AttestationDataAndCustodyBit(attestation.data, False)), signature=attestation.aggregate_signature, domain=get_domain(state.fork, attestation.data.slot, DOMAIN_ATTESTATION))`.
+* Verify bitfields:
+
+```python
+    assert verify_bitfield(attestation.aggregation_bitfield)
+    assert verify_bitfield(attestation.custody_bitfield)
+    assert attestation.custody_bitfield & attestation.aggregation_bitfield == attestation.custody_bitfield
```
JustinDrake: Is that the right way of doing bitwise AND? @djrtwo

specs/core/0_beacon-chain.md
```diff
@@ -1088,28 +1075,75 @@ def get_domain(fork: Fork,
...
 #### `verify_slashable_vote_data`
 
 ```python
-def verify_slashable_vote_data(state: BeaconState, vote_data: SlashableVoteData) -> bool:
-    if len(vote_data.custody_bit_0_indices) + len(vote_data.custody_bit_1_indices) > MAX_CASPER_VOTES:
+def verify_slashable_vote_data(state: BeaconState, slashable_vote_data: SlashableVoteData) -> bool:
+    """
+    Verify validity of ``slashable_vote_data`` fields.
+    """
+    if slashable_vote_data.custody_bitfield != 0:  # [TO BE REMOVED IN PHASE 1]
```
JustinDrake: Is that the right way of checking the bitfield is all zeroes? @djrtwo

specs/core/0_beacon-chain.md
djrtwo: `bytes` in python is a list of immutable ints. The operation would need to be performed on each position. I think the following is the functionality you are looking for but isn't pretty:

```python
and_bitfield = [
    custody_byte & aggregate_byte 
    for custody_byte, aggregate_byte in zip(attestation.custody_bitfield, attestation.aggregation_bitfield)
]
assert and_bitfield == attestation.custody_bitfield
```

specs/core/0_beacon-chain.md
djrtwo: Need to do check each byte in the bytes.

Can do:
```python
if not all(custody_byte == ZERO_HASH for custody_byte in slashable_vote_data.custody_bitfield):
```

`ZERO_HASH` might be poor usage because of the name `HASH` in it. Could use `b'\x00'` as well

vbuterin: I would do `!= b'\x00' * len(slashable_vote_data.custody_bitfield)`

specs/core/0_beacon-chain.md
```diff
@@ -1509,10 +1533,36 @@ For each `attestation` in `block.body.attestations`:
...
 * Verify that `attestation.data.justified_block_root` is equal to `get_block_root(state, attestation.data.justified_slot)`.
 * Verify that either `attestation.data.latest_crosslink_root` or `attestation.data.shard_block_root` equals `state.latest_crosslinks[shard].shard_block_root`.
-* `aggregate_signature` verification:
-    * Let `participants = get_attestation_participants(state, attestation.data, attestation.aggregation_bitfield)`.
-    * Let `group_public_key = bls_aggregate_pubkeys([state.validator_registry[v].pubkey for v in participants])`.
-    * Verify that `bls_verify(pubkey=group_public_key, message=hash_tree_root(AttestationDataAndCustodyBit(attestation.data, False)), signature=attestation.aggregate_signature, domain=get_domain(state.fork, attestation.data.slot, DOMAIN_ATTESTATION))`.
+* Verify bitfields:
+
+```python
+    assert verify_bitfield(attestation.aggregation_bitfield)
```
vbuterin: `verify_bitfield` takes size as an argument. I think the size here is `len(crosslink_committee)` though you want to pull out the logic for calculating that.

specs/core/0_beacon-chain.md
vbuterin:
```python
for i in range(len(crosslink_committee)):
    if get_bitfield_bit(attestation.aggregation_bitfield) == 0:
        assert get_bitfield_bit(attestation.custody_bitfield) == 0
```

More verbose but it clarifies what we're trying to do much better imo.

specs/core/0_beacon-chain.md
```diff
@@ -1088,28 +1075,75 @@ def get_domain(fork: Fork,
     ) * 2**32 + domain_type
 ```
 
+#### `get_bitfield_bit`
+
+```python
+def get_bitfield_bit(bitfield: bytes, i: int) -> int:
```
hwwhww: What do you think about renaming `i` to `index`?

specs/core/0_beacon-chain.md
```diff
@@ -1088,28 +1075,75 @@ def get_domain(fork: Fork,
...
+    """
+    Extract the bit in ``bitfield`` at position ``i``.
+    """
+    return (bitfield[i // 8] >> (7 - (i % 8))) % 2
+```
+
+#### `verify_bitfield`
+
+```python
+def verify_bitfield(bitfield: bytes, size: int) -> bool:
```
hwwhww: Maybe `size` => `bit_count` or something else to indicate it's the size of a committee?

specs/core/0_beacon-chain.md
hwwhww: Suggestion:
We return `bool` in Python codebase since the use cases are usually checking if the bit is 0 or not.
```python
def has_voted(bitfield: Bitfield, index: int) -> bool:
    return bool(bitfield[index // 8] & (128 >> (index % 8)))
```

JustinDrake: I'm trying to reserve `index` for validator indices.

JustinDrake:
I think this is ready for review :) Also addresses https://github.com/ethereum/eth2.0-specs/issues/461

specs/core/0_beacon-chain.md
```diff
@@ -1112,28 +1100,75 @@ def get_domain(fork: Fork,
...
+    if slashable_vote.custody_bitfield != b'\x00' * len(slashable_vote.custody_bitfield):  # [TO BE REMOVED IN PHASE 1]
+        return False
+
+    if len(slashable_vote.validator_indices) == 0:
         return False
 
-    return bls_verify_multiple(
+    for i in range(len(slashable_vote.validator_indices) - 1):
+        if slashable_vote.validator_indices[i] >= slashable_vote.validator_indices[i + 1]:
+        return False
```
vbuterin: Indent the inner "return False" a bit more?

specs/core/0_beacon-chain.md
```diff
@@ -1526,10 +1551,33 @@ For each `attestation` in `block.body.attestations`:
...
+
+```python
+    assert attestation.custody_bitfield == 0  # [TO BE REMOVED IN PHASE 1]
+
+    for i in range(len(crosslink_committee)):
+        if get_bitfield_bit(attestation.aggregation_bitfield) == 0:
+            assert get_bitfield_bit(attestation.custody_bitfield) == 0
+
+    participants = get_attestation_participants(state, attestation.data, attestation.aggregation_bitfield)
+    custody_bit_0_participants = get_attestation_participants(state, attestation.data, attestation.custody_bitfield)
```
vbuterin: Shouldn't this be flipped? So these are the custody bit 1 participants and the others are the bit 0 participants?

specs/core/0_beacon-chain.md
```diff
@@ -1526,10 +1551,33 @@ For each `attestation` in `block.body.attestations`:
...
 * Verify that `attestation.data.justified_block_root` is equal to `get_block_root(state, attestation.data.justified_slot)`.
 * Verify that either `attestation.data.latest_crosslink_root` or `attestation.data.shard_block_root` equals `state.latest_crosslinks[shard].shard_block_root`.
-* `aggregate_signature` verification:
-    * Let `participants = get_attestation_participants(state, attestation.data, attestation.aggregation_bitfield)`.
-    * Let `group_public_key = bls_aggregate_pubkeys([state.validator_registry[v].pubkey for v in participants])`.
-    * Verify that `bls_verify(pubkey=group_public_key, message=hash_tree_root(AttestationDataAndCustodyBit(attestation.data, False)), signature=attestation.aggregate_signature, domain=get_domain(state.fork, attestation.data.slot, DOMAIN_ATTESTATION))`.
+* Verify bitfields and aggregate signature:
+
+```python
+    assert attestation.custody_bitfield == 0  # [TO BE REMOVED IN PHASE 1]
```
vbuterin: Doesn't this line still need to be removed?

JustinDrake: Changing it to `assert attestation.custody_bitfield == b'\x00' * len(attestation.custody_bitfield)`. In phase 0, don't we want to enforce all custody bits to be zero?

JustinDrake:
Made fixes as per @vbuterin.

Also addressed https://github.com/ethereum/eth2.0-specs/issues/192#issuecomment-455481304 by making validator indices uint64 (as opposed to uint24). We can do thanks `validator_indices` (in a `SlashableVote`) be compressable at the network layer.

hwwhww:
@JustinDrake I think this PR can also fix #461?

JustinDrake:
@hwwhww Yes :) https://github.com/ethereum/eth2.0-specs/pull/488#issuecomment-457857018

specs/core/0_beacon-chain.md
```diff
@@ -1158,23 +1146,74 @@ def get_domain(fork: Fork,
...
         pubkeys=[
-            bls_aggregate_pubkeys([state.validator_registry[i].pubkey for i in vote_data.custody_bit_0_indices]),
-            bls_aggregate_pubkeys([state.validator_registry[i].pubkey for i in vote_data.custody_bit_1_indices]),
+            bls_aggregate_pubkeys([state.validator_registry[i].pubkey for i in custody_bit_0_indices]),
+            bls_aggregate_pubkeys([state.validator_registry[i].pubkey for i in custody_bit_1_indices]),
         ],
         messages=[
-            hash_tree_root(AttestationDataAndCustodyBit(vote_data.data, False)),
-            hash_tree_root(AttestationDataAndCustodyBit(vote_data.data, True)),
+            hash_tree_root(AttestationDataAndCustodyBit(attestation_data=slashable_vote.data, custody_bit=0b0)),
```
djrtwo: This has an implicit conversion from a bit (integer type) to a bool with the assignment `custody_bit=0b0`.

We should probably handle this more explicitly.

specs/core/0_beacon-chain.md
```diff
@@ -1572,10 +1601,33 @@ For each `attestation` in `block.body.attestations`:
...
+        pubkeys=[
+            bls_aggregate_pubkeys([state.validator_registry[i].pubkey for i in custody_bit_0_participants]),
+            bls_aggregate_pubkeys([state.validator_registry[i].pubkey for i in custody_bit_1_participants]),
+        ],
+        messages=[
+            hash_tree_root(AttestationDataAndCustodyBit(data=attestation.data, custody_bit=0b0)),
+            hash_tree_root(AttestationDataAndCustodyBit(data=attestation.data, custody_bit=0b1)),
+        ],
+        signature=attestation.aggregate_signature,
+        domain=get_domain(state.fork, attestation.data.slot, DOMAIN_ATTESTATION),
```
djrtwo: use `slot_to_epoch` here

```suggestion
        domain=get_domain(state.fork, slot_to_epoch(attestation.data.slot), DOMAIN_ATTESTATION),
```

specs/core/0_beacon-chain.md
```diff
@@ -1572,10 +1601,33 @@ For each `attestation` in `block.body.attestations`:
...
 * Verify that `attestation.data.justified_block_root` is equal to `get_block_root(state, get_epoch_start_slot(attestation.data.justified_epoch))`.
 * Verify that either `attestation.data.latest_crosslink_root` or `attestation.data.shard_block_root` equals `state.latest_crosslinks[shard].shard_block_root`.
-* `aggregate_signature` verification:
-    * Let `participants = get_attestation_participants(state, attestation.data, attestation.aggregation_bitfield)`.
-    * Let `group_public_key = bls_aggregate_pubkeys([state.validator_registry[v].pubkey for v in participants])`.
-    * Verify that `bls_verify(pubkey=group_public_key, message=hash_tree_root(AttestationDataAndCustodyBit(attestation.data, False)), signature=attestation.aggregate_signature, domain=get_domain(state.fork, slot_to_epoch(attestation.data.slot), DOMAIN_ATTESTATION))`.
+* Verify bitfields and aggregate signature:
+
+```python
+    assert attestation.custody_bitfield == b'\x00' * len(attestation.custody_bitfield)  # [TO BE REMOVED IN PHASE 1]
```
djrtwo: It is non-standard in the 0_beacon-chain spec to define an extended codeblock outside of the context of a function def. Should probably define it as `verify_attestation_bitfields_and_signature(state, attestation)`.

djrtwo:
Awesome work! just a few minor comments.

specs/core/0_beacon-chain.md
JustinDrake: > It is non-standard in the 0_beacon-chain spec to define an extended codeblock outside of the context of a function def

I was hoping to do this more in a future cleanup :)

(We also have one with `process_deposit` under "Run the following:".)

specs/core/0_beacon-chain.md
JustinDrake: I feel like `custody_bit=bool(0b0))` is a step down. It's more verbose and now there's a semantic mismatch between `_bit` and `bool()`.

Maybe we could rename SSZ `bool` to `bit`? After all, we use it more as a `uint1` than a `bool`.

djrtwo: I think it's worth avoiding `uint1`... 

We can merge as is and decide on ssz type outside this pr

JustinDrake: Sounds good. I'll let me you approve/merge.

PR #490, opened 2019-01-23
Updated deposit contract for gas efficiency

vbuterin:
Using the progressive merkle root calculation algo with a sample implementation here:

https://github.com/ethereum/research/blob/master/beacon_chain_impl/progressive_merkle_tree.py

Should save a huge amount on gas costs, as with this algo on average only ~2 hashes need to be rewritten for each deposit with 0 new storage (compared with 32 rewritten and 2 new storage before). However, we lose the benefit of the current scheme that updated Merkle branches can be computed from the state.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -708,7 +708,8 @@ def deposit(deposit_input: bytes[512]):
 
     self.deposit_count += 1
 
-    log.Deposit(self.get_deposit_root(), deposit_data, merkle_tree_index, self.branch)
+    new_deposit_root:bytes32 = self.get_deposit_root()
+    log.Deposit(new_deposit_root, deposit_data, merkle_tree_index, self.branch)
 
     if msg.value == as_wei_value(MAX_DEPOSIT_AMOUNT, "gwei"):
         self.full_deposit_count += 1
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -653,7 +653,7 @@ DEPOSIT_CONTRACT_TREE_DEPTH: constant(uint256) = 32
 TWO_TO_POWER_OF_TREE_DEPTH: constant(uint256) = 4294967296  # 2**32
 SECONDS_PER_DAY: constant(uint256) = 86400
 
-Deposit: event({deposit_root: bytes32, data: bytes[2064], merkle_tree_index: bytes[8], branch: bytes32[32]})
+Deposit: event({deposit_root: bytes32, data: bytes[528], merkle_tree_index: bytes[8], branch: bytes32[32]})
 ChainStart: event({deposit_root: bytes32, time: bytes[8]})
 
 zerohashes: bytes32[32]
@@ -682,14 +682,14 @@ def get_deposit_root() -> bytes32:
 
 @payable
 @public
-def deposit(deposit_input: bytes[2048]):
+def deposit(deposit_input: bytes[512]):
     assert msg.value >= as_wei_value(MIN_DEPOSIT_AMOUNT, "gwei")
     assert msg.value <= as_wei_value(MAX_DEPOSIT_AMOUNT, "gwei")
 
     index: uint256 = self.deposit_count
     deposit_amount: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
     deposit_timestamp: bytes[8] = slice(concat("", convert(block.timestamp, bytes32)), start=24, len=8)
-    deposit_data: bytes[2064] = concat(deposit_amount, deposit_timestamp, deposit_input)
+    deposit_data: bytes[528] = concat(deposit_amount, deposit_timestamp, deposit_input)
     merkle_tree_index: bytes[8] = slice(concat("", convert(index, bytes32)), start=24, len=8)
 
     # add deposit to merkle tree
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -656,9 +656,29 @@ SECONDS_PER_DAY: constant(uint256) = 86400
 Deposit: event({previous_deposit_root: bytes32, data: bytes[2064], merkle_tree_index: bytes[8]})
 ChainStart: event({deposit_root: bytes32, time: bytes[8]})
 
-deposit_tree: map(uint256, bytes32)
+zerohashes: bytes32[32]
+branch: bytes32[32]
 deposit_count: uint256
 full_deposit_count: uint256
+
+@public
+def __init__():
+    for i in range(31):
+        self.zerohashes[i+1] = sha3(concat(self.zerohashes[i], self.zerohashes[i]))
+        self.branch[i+1] = self.zerohashes[i+1]
+
+@public
+@constant
+def get_deposit_root() -> bytes32:
+    root:bytes32 = 0x0000000000000000000000000000000000000000000000000000000000000000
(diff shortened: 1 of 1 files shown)


JustinDrake:
Oh nice! Let's see if I get this right. The intuition is that we have a branch in a sparse Merkle tree moving from left to right, "following" the Merkle updates. This is possible because:

1. We only write at index `deposit_count`, which only ever increments by one
2. We only read the most recent branch

A couple suggestions:

1. Log `branch` at well-chosen points (e.g. when `deposit_count % 128 == 0`—possibly off by 1). This allows Eth1-light-client validators to grab the most recent branch prior to the last processed deposit, and run through the subsequent deposits up to `latest_eth1_data.deposit_root`. (Note one cool thing is that we don't need to log the part of `branch` which is `zerohashes` so we save ourselves from logging log_2(128)*32 bytes.)
2. I think the assignments `self.branch[j] = self.zerohashes[j]` can be optimised away to save gas.

JustinDrake:
> I think those are necessary to facilitate Merkle root calculation.

In the Merkle root calculation can we reuse the `power_of_two` logic (or cache it)? Something like:

```
def get_deposit_root() -> bytes32:
    root:bytes32 = 0x0000000000000000000000000000000000000000000000000000000000000000
    size:uint256 = self.deposit_count
    i: int128 = 0
    power_of_two: uint256 = 2
    for _ in range(32):
        if (size+1) % power_of_two != 0:
            break
        i += 1
        power_of_two *= 2
    value:bytes32 = sha3(deposit_data)

    for h in range(32):
        if size % 2 == 1:
            if j < i:
                root = sha3(concat(self.zerohashes[h], root))
            else:
                root = sha3(concat(self.branch[h], root))
        else:
            if j < i:
                root = sha3(concat(root, self.zerohashes[h]))
            else:
                root = sha3(concat(root, self.branch[h]))
        size /= 2
    return root

    for j in range(32):
        if j < i:
            value = sha3(concat(self.branch[j], value))
            self.branch[j] = self.zerohashes[j]
```

vbuterin:
Aaah you're right, there is a way to do it, and it's not even super complicated. See python here: https://github.com/ethereum/research/blob/master/beacon_chain_impl/progressive_merkle_tree.py#L32

Adding this to the PR now.

specs/core/0_beacon-chain.md
```diff
@@ -653,57 +653,73 @@ DEPOSIT_CONTRACT_TREE_DEPTH: constant(uint256) = 32
 TWO_TO_POWER_OF_TREE_DEPTH: constant(uint256) = 4294967296  # 2**32
 SECONDS_PER_DAY: constant(uint256) = 86400
 
-Deposit: event({previous_deposit_root: bytes32, data: bytes[2064], merkle_tree_index: bytes[8]})
+Deposit: event({previous_deposit_root: bytes32, data: bytes[2064], merkle_tree_index: bytes[8], branch: bytes32[32]})
```
NIC619: Should we rename `previous_deposit_root` to `deposit_root`? or we still need to log previous deposit root?

Nashatyrev:
Just want to mention another issue this PR solves (may be as a side effect) by adding `merkle_branch` to the `Deposit` event: 
In the old version one can obtain `merkle_branch` only by calling const function `get_branch`, so deposits created with a single tx (or in a single block) could be verified only against their common `deposit_root`, because we couldn't retrieve `merkle_branch`es for 'intermediate' `deposit_roots`. However deposits still could be verified. 
The problem might occur if the `ChainStart` event happens on some non-last deposit within a single tx. Thus having such 'intermediate' `deposit_root` as a `ChainStart` argument

specs/core/0_beacon-chain.md
Nashatyrev: Yes, in this version the argument is not the previous `root`. 
If indeed the current `root` is wanted here then it can be removed at all, since it can be calculated from the `branch`

Nashatyrev:
We can cut topmost zeroes from the logged branch to reduce log size

NIC619:
@djrtwo My apology. My comment was a bit confusing and I've updated the comment.

The basic transaction cost would be `21000 + 68 * 2048 = 160264` if `len(deposit_input) == 2048`.
Logging accounts for another `8 * 2064 = 16512` gas.

djrtwo:
I see @NIC619. I think I just read through it too quickly.

@vbuterin pointed out that most of the 2048 max will be unused and that a majority of the bytes are for forward compatibility if we change the deposit requirements in some future fork. We can bring it down to 1024 or 512 to lower the _max_ gas cost, but it's not entirely necessary.

specs/core/0_beacon-chain.md
```diff
@@ -653,7 +653,7 @@ DEPOSIT_CONTRACT_TREE_DEPTH: constant(uint256) = 32
 TWO_TO_POWER_OF_TREE_DEPTH: constant(uint256) = 4294967296  # 2**32
 SECONDS_PER_DAY: constant(uint256) = 86400
 
-Deposit: event({deposit_root: bytes32, data: bytes[2064], merkle_tree_index: bytes[8], branch: bytes32[32]})
+Deposit: event({deposit_root: bytes32, data: bytes[528], merkle_tree_index: bytes[8], branch: bytes32[32]})
```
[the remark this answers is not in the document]
vbuterin: The real constant is 512; 528 is just 512 plus the size of the auxiliary data.

Though I suppose we could make both constants...

vbuterin:
So `MAX_INPUT_SIZE: constant(uint256) = 512`, `MAX_DATA_SIZE: constant(uint256) = 528` # = MAX_INPUT_SIZE + 16

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
vbuterin: So `MAX_INPUT_SIZE: constant(uint256) = 512`, `MAX_DATA_SIZE: constant(uint256) = 528` # = MAX_INPUT_SIZE + 16

specs/core/0_beacon-chain.md
```diff
@@ -653,57 +653,73 @@ DEPOSIT_CONTRACT_TREE_DEPTH: constant(uint256) = 32
...
+        power_of_two *= 2
+    value:bytes32 = sha3(deposit_data)
+    for j in range(32):
+        if j < i:
+            value = sha3(concat(self.branch[j], value))
+    self.branch[i] = value
 
     self.deposit_count += 1
+
+    log.Deposit(self.get_deposit_root(), deposit_data, merkle_tree_index, self.branch)
```
NIC619: Note: this syntax depends on next Vyper release which includes https://github.com/ethereum/vyper/pull/1213.

specs/core/0_beacon-chain.md
```diff
@@ -653,57 +653,73 @@ DEPOSIT_CONTRACT_TREE_DEPTH: constant(uint256) = 32
...
-@public
-@constant
-def get_branch(leaf: uint256) -> bytes32[32]: # size is DEPOSIT_CONTRACT_TREE_DEPTH (symbolic const not supported)
-    branch: bytes32[32] # size is DEPOSIT_CONTRACT_TREE_DEPTH
-    index: uint256 = leaf + TWO_TO_POWER_OF_TREE_DEPTH
-    for i in range(DEPOSIT_CONTRACT_TREE_DEPTH):
-        branch[i] = self.deposit_tree[bitwise_xor(index, 1)]
-        index /= 2
-    return branch
+            log.ChainStart(self.get_deposit_root(), chainstart_time)
```
NIC619: ditto

Issue #224, opened 2018-12-04
Integer overflow considerations


paulhauner:
I'm considering integer overflows in Lighthouse and there are cases where I'm making some assumptions about what I _think_ the spec means. 

For example, consider this code in "verify attestations":

> Verify that `attestation.data.slot <= block.slot - MIN_ATTESTATION_INCLUSION_DELAY`

As all of these variables are declared as `uint64` it would be fair to assume that we're going to see potential under-flows in the first few blocks and the behavior of the function will be significantly different depending on whether or not the subtraction saturates or under-flows.

From my perspective it would be nice to see some explicitness around what to do in the case of overflows.

Or maybe all overflow scenarios have been considered and I'm missing some other checks that are normalizing behavior?

JustinDrake:
In this particular instance it's an easy fix:

> Verify that `attestation.data.slot + MIN_ATTESTATION_INCLUSION_DELAY <= block.slot`

I will go through the spec and fix what I find. The "harder" underflows are regarding balances and penalties. The spec has the following notice, admittedly not completely satisfactory:

> Note: When applying penalties in the following balance recalculations implementers should make sure the `uint64` does not underflow.

paulhauner:
Thank you! :) I'll keep an eye out for your changes, then let you know if I notice any more.

JustinDrake:
So one idea is to have `INITIAL_SLOT_NUMBER = 3 * EPOCH_LENGTH`. Do you agree this solves the underflow problems with slots?

arnetheduck:
I'm not familiar enough with the spec to say that this is sufficient to fully solve underflows, but gut-feeling-wise, it seems.. brittle.

Many of these issues happen in comparisons - another way to address those at least is to simply disallow subtraction and instead only allow addition. This solves the underflow problem, but  not overflow problem. 

#192 expresses a desire to be able to make formal claims about overflows as well - for example, one could make a formal claim about the system being both under- and overflow-free up to slot X.

JustinDrake:
Paging in @djrtwo.

> This solves the underflow problem, but not overflow problem.

What is the overflow problem? Can you point to an instance in the code?

djrtwo:
When you flip a conditional that might have underflowed due to subtraction, you always get a conditional that might theoretically overflow due to addition.

[...]

PR #492, opened 2019-01-23
epoch trans at end of epoch and fix finality conditions

### LINKED ISSUE #352: `shard_committees_at_slots` missing when processing previous epoch attestations

### LINKED ISSUE #410: Finalization rule 3 would return True in no case
The spec says:
> Set `state.finalized_slot = state.previous_justified_slot` if any of the following are true:

where the third condition is: 

`state.previous_justified_slot == state.slot - 4 * EPOCH_LENGTH and state.justification_bitfield % 16 in (15, 14)`

Not sure if I misunderstood something, but I couldn't find a scenario that makes this condition true.

If we introduce the notation B4, ..., B1 from the [minispec](https://ethresear.ch/t/beacon-chain-casper-mini-spec/2760). This condition requires that `previous_justified_slot` stay at 4 epochs ago (B1), and the last 4 bits looks like `1110` or `1111`, that is B1, B2, and B3 in justification hashes.

Before this epoch, the bits for (B1, B2, B3) might look like `111`, `110`, or `10x`, but none of them is the case.
- `111`: The B1 would be finalized by rule 2 so no need rule 3 in next epoch.
- `110`: The previous justified slot would become B2.
- `10x`: The spec has `state.justification_bitfield |= 2` and `state.justification_bitfield |= 1` but no `state.justification_bitfield |= 4`. No way for B2 to become 1 in next epoch.

See the [test cases](https://github.com/ethereum/py-evm/pull/1690/files#diff-2f13bfb47df39a36b2fb37d93d170fcfR48).

screen shot 2019-01-08 at 8 40 47 pm

djrtwo:
* moves epoch transition to last slot in epoch (addresses https://github.com/ethereum/eth2.0-specs/issues/267#issuecomment-446139113)
* fixes #410 finality conditions
* fixes #352 and #409 (shard committees for prev epoch still available when processing attestations)

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1570,9 +1570,12 @@ For each `exit` in `block.body.exits`:
 
 ## Per-epoch processing
 
-The steps below happen when `state.slot % EPOCH_LENGTH == 0`.
+The steps below happen when `state.slot % EPOCH_LENGTH == EPOCH_LENGTH - 1`.
 
 ### Helpers
+
+* Let `current_epoch_start_slot = state.slot - (EPOCH_LENGTH + 1)`.
+* Let `previous_epoch_start_slot = state.slot - 2 * EPOCH_LENGTH + 1` if `state.slot > EPOCH_LENGTH` else `current_epoch_start_slot`.
 
 All [validators](#dfn-validator):
 
@@ -1581,23 +1584,23 @@ All [validators](#dfn-validator):
 
 [Validators](#dfn-Validator) attesting during the current epoch:
 
-* Let `current_epoch_attestations = [a for a in state.latest_attestations if state.slot - EPOCH_LENGTH <= a.data.slot < state.slot]`. (Note: this is the set of attestations of slots in the epoch `state.slot-EPOCH_LENGTH...state.slot-1`, _not_ attestations that got included in the chain during the epoch `state.slot-EPOCH_LENGTH...state.slot-1`.)
+* Let `current_epoch_attestations = [a for a in state.latest_attestations if current_epoch_start_slot <= a.data.slot <= state.slot]`. (Note: this is the set of attestations of slots in the epoch `current_epoch_start_slot...state.slot`, _not_ attestations that got included in the chain during the epoch `current_epoch_start_slot...state.slot`.)
 * Validators justifying the epoch boundary block at the start of the current epoch:
-  * Let `current_epoch_boundary_attestations = [a for a in current_epoch_attestations if a.data.epoch_boundary_root == get_block_root(state, state.slot-EPOCH_LENGTH) and a.data.justified_slot == state.justified_slot]`.
+  * Let `current_epoch_boundary_attestations = [a for a in current_epoch_attestations if a.data.epoch_boundary_root == get_block_root(state, current_epoch_start_slot) and a.data.justified_slot == state.justified_slot]`.
   * Let `current_epoch_boundary_attester_indices` be the union of the [validator](#dfn-validator) index sets given by `[get_attestation_participants(state, a.data, a.aggregation_bitfield) for a in current_epoch_boundary_attestations]`.
   * Let `current_epoch_boundary_attesting_balance = sum([get_effective_balance(state, i) for i in current_epoch_boundary_attester_indices])`.
 
 [Validators](#dfn-Validator) attesting during the previous epoch:
 
 * Validators that made an attestation during the previous epoch:
-  * Let `previous_epoch_attestations = [a for a in state.latest_attestations if state.slot - 2 * EPOCH_LENGTH <= a.data.slot < state.slot - EPOCH_LENGTH]`.
+  * Let `previous_epoch_attestations = [a for a in state.latest_attestations if previous_epoch_start_slot <= a.data.slot < current_epoch_start_slot]`.
   * Let `previous_epoch_attester_indices` be the union of the validator index sets given by `[get_attestation_participants(state, a.data, a.aggregation_bitfield) for a in previous_epoch_attestations]`.
 * Validators targeting the previous justified slot:
   * Let `previous_epoch_justified_attestations = [a for a in current_epoch_attestations + previous_epoch_attestations if a.data.justified_slot == state.previous_justified_slot]`.
   * Let `previous_epoch_justified_attester_indices` be the union of the validator index sets given by `[get_attestation_participants(state, a.data, a.aggregation_bitfield) for a in previous_epoch_justified_attestations]`.
   * Let `previous_epoch_justified_attesting_balance = sum([get_effective_balance(state, i) for i in previous_epoch_justified_attester_indices])`.
 * Validators justifying the epoch boundary block at the start of the previous epoch:
-  * Let `previous_epoch_boundary_attestations = [a for a in previous_epoch_justified_attestations if a.data.epoch_boundary_root == get_block_root(state, state.slot - 2 * EPOCH_LENGTH)]`.
+  * Let `previous_epoch_boundary_attestations = [a for a in previous_epoch_justified_attestations if a.data.epoch_boundary_root == get_block_root(state, previous_epoch_start_slot)]`.
   * Let `previous_epoch_boundary_attester_indices` be the union of the validator index sets given by `[get_attestation_participants(state, a.data, a.aggregation_bitfield) for a in previous_epoch_boundary_attestations]`.
   * Let `previous_epoch_boundary_attesting_balance = sum([get_effective_balance(state, i) for i in previous_epoch_boundary_attester_indices])`.
 * Validators attesting to the expected beacon chain head during the previous epoch:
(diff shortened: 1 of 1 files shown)


vbuterin:
Here's a PR to your PR :rofl: 

https://github.com/ethereum/eth2.0-specs/pull/495

djrtwo:
Thanks! Merged the PR into my PR

vbuterin:
Looks good to me at the moment. Maybe after merging this one and the light client one we can do a final pass on the epoch transition logic. I think for cleanliness reasons it may be a good idea to remove use of `state.slot` entirely (in favor of current/next/previous epoch start slot) but best do that after these two PRs as they both modify the epoch transition section quite a bit.

specs/core/0_beacon-chain.md
```diff
@@ -942,24 +942,25 @@ def get_crosslink_committees_at_slot(state: BeaconState,
     state_epoch_slot = state.slot - (state.slot % EPOCH_LENGTH)
     assert state_epoch_slot <= slot + EPOCH_LENGTH
     assert slot < state_epoch_slot + EPOCH_LENGTH
```
vbuterin: I think this should be changed to:
```python
previous_epoch_start_slot = state.slot - state.slot % EPOCH_LENGTH - EPOCH_LENGTH
next_epoch_start_slot =  state.slot - state.slot % EPOCH_LENGTH + EPOCH_LENGTH
assert previous_epoch_start_slot <= slot < next_epoch_start_slot
```

hwwhww: The logic is identical, right?
I do see it's also much more understandable, but also, the intention of `state_epoch_slot` is the avoid uint underflow: https://github.com/ethereum/eth2.0-specs/pull/429.

djrtwo: I definitely prefer the readability...

djrtwo: but yes, I believe it is the same functionality.

The underflow can be avoided using something like 
```python
current_epoch_start_slot = state.slot - state.slot % EPOCH_LENGTH
previous_epoch_start_slot = current_epoch_start_slot - EPOCH_LENGTH if state.slot >= EPOCH_LENGTH else current_epoch_start_slot
next_epoch_start_slot =  state.slot - state.slot % EPOCH_LENGTH + EPOCH_LENGTH
assert previous_epoch_start_slot <= slot < next_epoch_start_slot

**Comment by djrtwo** [comment]
* handled merge conflict
* modified where `current_epoch_calculation_slot` is being updated to be the start of the _next_ epoch when it is updated.
* updating `latest_index_roots` for the `next_epoch` instead of the current. using this root in the seed generation for the shuffling starting at next epoch
* created a function `generate_seed` to handle the complexity of the seed generation. Also handles the edge case when `slot < SEED_LOOKAHEAD` to prevent underflow
* set `latest_index_roots` for genesis slot/epoch
* use `generate_seed` to properly embed the active index root in the initial shuffling seed

**Comment by djrtwo** [comment]
I'm working on a PR to this PR that makes epochs a first class citizen. Seeing if it cleans things up.

File: specs/core/0_beacon-chain.md [hunk 2]
```diff
@@ -998,28 +1032,43 @@ def get_block_root(state: BeaconState,
...
+    """
+    Generate a seed for the given ``epoch``.
+    """
+    randao_mix_epoch = epoch + LATEST_RANDAO_MIXES_LENGTH - SEED_LOOKAHEAD
+
+    return hash(
+        get_randao_mix(state, randao_mix_epoch) +
+        get_active_index_root(state, epoch)
+    )
+
```
JustinDrake: extra line

specs/core/0_beacon-chain.md
```diff
@@ -998,28 +1032,42 @@ def get_block_root(state: BeaconState,
...
+#### `generate_seed`
+
+```python
+def generate_seed(state: BeaconState,
+                  epoch: EpochNumber) -> Bytes32:
+    """
+    Generate a seed for the given ``epoch``.
+    """
+    randao_mix_epoch = epoch + LATEST_RANDAO_MIXES_LENGTH - SEED_LOOKAHEAD
```
vbuterin: Huh? Why `+ LATEST_RANDAO_MIXES_LENGTH`?

If it's to avoid underflow IMO this is going too far and sacrifices clarity; we should have the anti-underflow logic in the `get_randao_mix` function.

specs/core/0_beacon-chain.md
```diff
@@ -1636,54 +1687,62 @@ Define the following helpers to process attestation inclusion rewards and inclus
 
 ### Eth1 data
 
-If `state.slot % ETH1_DATA_VOTING_PERIOD == 0`:
+If `next_epoch % ETH1_DATA_VOTING_PERIOD == 0`:
 
-* Set `state.latest_eth1_data = eth1_data_vote.data` if `eth1_data_vote.vote_count * 2 > ETH1_DATA_VOTING_PERIOD` for some `eth1_data_vote` in `state.eth1_data_votes`.
+* Set `state.latest_eth1_data = eth1_data_vote.data` if `eth1_data_vote.vote_count * 2 > ETH1_DATA_VOTING_PERIOD * EPOCH_LENGTH` for some `eth1_data_vote` in `state.eth1_data_votes`.
```
vbuterin: Maybe:

* If `eth1_data_vote.vote_count * 2 > ETH1_DATA_VOTING_PERIOD * EPOCH_LENGTH` for some `eth1_data_vote` in `state.eth1_data_votes` (ie. more than half the votes in this voting period were for that value), set `state.latest_eth1_data = eth1_data_vote.data`

specs/core/0_beacon-chain.md
djrtwo: Ah, the epoch starting at current unix time will fix this. Making an issue for that now

djrtwo: do we want to change this in this PR?

hwwhww:
I think #505 could close #224.

specs/core/0_beacon-chain.md
```diff
@@ -189,24 +190,36 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
 | Name | Value |
 | - | - |
 | `GENESIS_FORK_VERSION` | `0` |
-| `GENESIS_SLOT` | `0` |
+| `GENESIS_SLOT` | `2**19` |
+| `GENESIS_EPOCH` | `slot_to_epoch(GENESIS_SLOT)` |
```
ChihChengLiang: Note that `slot_to_epoch` use `EPOCH_LENGTH` that is defined in the next section. So in the chain config, this will look like:
```python
epoch_length = 2**6
SERENITY_CONFIG = BeaconConfig(
    ...
    GENESIS_EPOCH=slot_to_epoch(GENESIS_SLOT, epoch_length)
    ...
    EPOCH_LENGTH=epoch_length
    ...
)
```

specs/core/0_beacon-chain.md
```diff
@@ -946,27 +978,32 @@ def get_crosslink_committees_at_slot(state: BeaconState,
...
-        committees_per_slot = get_previous_epoch_committee_count_per_slot(state)
-        shuffling = get_shuffling(
-            state.previous_epoch_seed,
-            state.validator_registry,
-            state.previous_epoch_calculation_slot,
-        )
-        slot_start_shard = (state.previous_epoch_start_shard + committees_per_slot * offset) % SHARD_COUNT
+    epoch = slot_to_epoch(slot)
+    current_epoch = get_current_epoch(state)
+    previous_epoch = current_epoch - 1 if epoch > GENESIS_EPOCH else current_epoch
```
hwwhww: It seems the "previous" is relative to "current". Should we use `current_epoch` instead of `epoch` in this line?
```suggestion
    previous_epoch = current_epoch - 1 if current_epoch > GENESIS_EPOCH else current_epoch
```

specs/core/0_beacon-chain.md
djrtwo: Not the prettiest. I think we want to move to #504 which will use chain start time to seed the initial state rather than config.

PR #493, opened 2019-01-24
"proposer_slots" -> "proposer_nonce"

paulhauner:
## What
 
Rename `proposer_slots` to `proposer_nonce`.

## Why

The naming "proposer_slots" is confusing when considering the API between the Beacon Node and the Validator. Consider that the validator must ask the following questions of the Beacon Node:

- What slots should I propose on? (shuffling)
- What is my proposer slots? (`proposer_slots`)

AFAIK we don't care how many slots that proposer has proposed, only that the number is different for each proposal. Nonce sounds like a better descriptor to me.


paulhauner:
Closed in favor of #496

PR #496, opened 2019-01-24
Replace `proposer_slots` with `slot`

paulhauner:
## What

Use `slot` instead of `proposer_slots` when producing the `randao_reveal` signature.

## Why

In the separate Beacon Node (BN) & Validator Client (VC) model, it's difficult for a VC to know the BN isn't causing the VC to reveal future values. This is because the BN is the source of truth for the current `proposer_slots`.

If `slot` is used, a validator can use their system clock to determine if a reveal is a "future reveal" (bad) or a "present reveal" (good).

Additionally, using `slot` removes one API call from the VC to the BN. It already knows the `slot` but it must request `proposer_slots`.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -517,8 +517,6 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
     'pubkey': 'bytes48',
     # Withdrawal credentials
     'withdrawal_credentials': 'bytes32',
-    # Number of proposer slots since genesis
-    'proposer_slots': 'uint64',
     # Slot when validator activated
     'activation_slot': 'uint64',
     # Slot when validator exited
@@ -1333,7 +1331,6 @@ def process_deposit(state: BeaconState,
         validator = Validator(
             pubkey=pubkey,
             withdrawal_credentials=withdrawal_credentials,
-            proposer_slots=0,
             activation_slot=FAR_FUTURE_SLOT,
             exit_slot=FAR_FUTURE_SLOT,
             withdrawal_slot=FAR_FUTURE_SLOT,
@@ -1430,7 +1427,6 @@ Below are the processing steps that happen at every slot.
 ### Misc counters
 
 * Set `state.slot += 1`.
-* Set `state.validator_registry[get_beacon_proposer_index(state, state.slot)].proposer_slots += 1`.
 * Set `state.latest_randao_mixes[state.slot % LATEST_RANDAO_MIXES_LENGTH] = state.latest_randao_mixes[(state.slot - 1) % LATEST_RANDAO_MIXES_LENGTH]`
 
 ### Block roots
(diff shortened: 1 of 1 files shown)


djrtwo:
Copied from a text offline (well online, but not on github):

>The problem with the using `slot` is that it slightly opens up a malicious validator's range for controlling future shufflings.

> Say, I can manipulate the RNG in such a way that I can certainly get validator A in the next epoch. If we use `proposer_slots`, validator A only has one option for the value I can get them to reveal next epoch. If we use `slot`, then I now how more play beacuse the reveal is different depending on the slot I manage to get validator A into. This is a small but distinct increase in the attack surface when an attacker has some amount of control over the RNG. That's why we opted for `proposer_slots` akin to the previous `randao_layers` instead of `slot`.

JustinDrake:
I think I have a good mental model. Basically an attacker in the previous epoch has some influence (by choosing to reveal/not reveal at strategic places) over the shuffling for the next epoch. Let's say the attacker can choose from `n` different combinations of 64 proposers for the next epoch. From these `n` different combinations will result `m <= n` different RANDAO seeds. Our job is now to minimise `m` to minimise the grinding opportunity on the seed.

Now suppose we have validators `v_1` and `v_2` where `v_1` is controlled by the attacker and `v_2` is offline. Suppose also that the attacker how two combinations of 64 proposers which end in `v_1, v_2` and `v_2, v_1`. If we use `slot` then that results in two distinct RANDAO values, but if we use `proposer_slots` the two RANDAO values would be equivalent. So `proposer_slots` is strictly more secure.

Now this analysis raises another point, namely that we want the mixing process to be commutative. Indeed, if the attacker controls both `v_1` and `v_2` then we want the two combinations ending in `v_1, v_2` and `v_2, v_1` to be equivalent if both `v_1` and `v_2` reveal. This used to be the case when the mixing was xor, and is not longer the case with the hash chain. So maybe we should revert back to xor mixing.

One thing worth pointing is that these subtle analytical issues disappear with VDFs. Two options which could make sense:

1. Favour simplicity and convenience by moving to `slot`, sticking with the hash chain, and waiting for VDFs to clean things up.
2. Favour security of the base RANDAO by sticking with `proposer_slots` and reverting to xor mixing.

@vbuterin, @djrtwo

JustinDrake:
Here's a possibility to get the pros of both the hash chain and xor:

* Store the `randao_reveal` in the current epoch
* At the end of the epoch:
     * Sort the `randao_reveal`
     * Perform tree hashing on the sorted list
     * Use the tree hash root as the seed

Sorting provides commutativity. Tree hashing provides convenient witnessing of the `randao_reveal` for L2 applications such as low influence functions.

JustinDrake:
Oh! Instead of `proposer_slots` we can simply use `slot / EPOCH_LENGTH`! That gives us the security of `proposer_slots` and the simplicity of `slot` 😃

JustinDrake:
Oh, and another realisation: hash-then-xor gives us the same convenient witnessing as the hash chain! The reason is that finding two sets of size at most 64 that give the same answer when hashed-then-xored feels almost as strong (may need to check with cryptographer) as collision resistance of the hash function. So my concrete proposal is:

* Remove `proposer_slots`
* Use `slot / EPOCH_LENGTH` instead
* Stick with hashing the `randao_reveal`
* Revert back to xor as the mixing function

djrtwo:
closing in favor of #498

JustinDrake:
> I think it suffices that the proposer not be able to propose twice within a single epoch

We assume an attacker can control up to `ENTRY_EXIT_DELAY` consecutive slots in a period of low liveness (i.e. when most validators are offline). It is therefore possible for an attacker to get different values for `slot // EPOCH_LENGTH` within `ENTRY_EXIT_DELAY`, and hence grind more than with `proposer_slots`. Does that make sense? It's an edge case, but I wonder if we can address it cleanly.

PR #498, opened 2019-01-25
randao reveal is signed epoch number

djrtwo:
Addresses changes to randao reveal noted [here](https://github.com/ethereum/eth2.0-specs/pull/496#issuecomment-457569088).

- remove `proposer_slots`
- sign `block.slot // EPOCH_LENGTH` (the epoch number) as randao reveal
- Hash the reveal to reduce to 32 bytes and create a reveal chain
- XOR with previous randao mix to get new mix

Thanks @JustinDrake and @paulhauner for the ideas and conversation!


paulhauner:
GGWP

PR #499, opened 2019-01-26
make epochs first class citizens

djrtwo:
* changes the validator change slots to epochs
* uses epochs in places where we were previously normalizing to epoch boundaries anyway
* does _not_ change state vars for finality to epochs although we can
* does _not_ change crosslinks to epochs although we can

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -60,6 +60,7 @@
         - [Helper functions](#helper-functions)
             - [`hash`](#hash)
             - [`hash_tree_root`](#hash_tree_root)
+            - [`slot_to_epoch`](#slot_to_epoch)
             - [`is_active_validator`](#is_active_validator)
             - [`get_active_validator_indices`](#get_active_validator_indices)
             - [`shuffle`](#shuffle)
@@ -84,7 +85,7 @@
             - [`is_double_vote`](#is_double_vote)
             - [`is_surround_vote`](#is_surround_vote)
             - [`integer_squareroot`](#integer_squareroot)
-            - [`entry_exit_effect_slot`](#entry_exit_effect_slot)
+            - [`entry_exit_effect_epoch`](#entry_exit_effect_epoch)
             - [`bls_verify`](#bls_verify)
             - [`bls_verify_multiple`](#bls_verify_multiple)
             - [`bls_aggregate_pubkeys`](#bls_aggregate_pubkeys)
@@ -191,7 +192,7 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
 | `GENESIS_FORK_VERSION` | `0` |
 | `GENESIS_SLOT` | `0` |
 | `GENESIS_START_SHARD` | `0` |
-| `FAR_FUTURE_SLOT` | `2**64 - 1` |
+| `FAR_FUTURE_EPOCH` | `2**64 - 1` |
 | `ZERO_HASH` | `int_to_bytes32(0)` |
 | `EMPTY_SIGNATURE` | `int_to_bytes96(0)` |
 | `BLS_WITHDRAWAL_PREFIX_BYTE` | `int_to_bytes1(0)` |
@@ -203,10 +204,10 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
 | `SLOT_DURATION` | `6` | seconds | 6 seconds |
 | `MIN_ATTESTATION_INCLUSION_DELAY` | `2**2` (= 4) | slots | 24 seconds |
 | `EPOCH_LENGTH` | `2**6` (= 64) | slots | 6.4 minutes |
-| `SEED_LOOKAHEAD` | `2**6` (= 64) | slots | 6.4 minutes |
-| `ENTRY_EXIT_DELAY` | `2**8` (= 256) | slots | 25.6 minutes |
+| `SEED_LOOKAHEAD` | `2**0` (= 1) | epochs | 6.4 minutes |
+| `ENTRY_EXIT_DELAY` | `2**2` (= 4) | epochs | 25.6 minutes |
 | `ETH1_DATA_VOTING_PERIOD` | `2**10` (= 1,024) | slots | ~1.7 hours |
-| `MIN_VALIDATOR_WITHDRAWAL_TIME` | `2**14` (= 16,384) | slots | ~27 hours |
+| `MIN_VALIDATOR_WITHDRAWAL_EPOCHS` | `2**8` (= 256) | epochs | ~27 hours |
 
 ### Reward and penalty quotients
 
@@ -468,7 +469,7 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
     # Validator registry
     'validator_registry': [Validator],
     'validator_balances': ['uint64'],
-    'validator_registry_update_slot': 'uint64',
+    'validator_registry_update_epoch': 'uint64',
     'validator_registry_exit_count': 'uint64',
 
     # Randomness and committees
@@ -476,8 +477,8 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
     'latest_vdf_outputs': ['bytes32'],
     'previous_epoch_start_shard': 'uint64',
     'current_epoch_start_shard': 'uint64',
-    'previous_epoch_calculation_slot': 'uint64',
-    'current_epoch_calculation_slot': 'uint64',
+    'previous_calculation_epoch': 'uint64',
+    'current_calculation_epoch': 'uint64',
     'previous_epoch_seed': 'bytes32',
     'current_epoch_seed': 'bytes32',
 
@@ -514,14 +515,14 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
     'withdrawal_credentials': 'bytes32',
     # Number of proposer slots since genesis
     'proposer_slots': 'uint64',
-    # Slot when validator activated
-    'activation_slot': 'uint64',
-    # Slot when validator exited
-    'exit_slot': 'uint64',
-    # Slot when validator withdrew
-    'withdrawal_slot': 'uint64',
-    # Slot when validator was penalized
-    'penalized_slot': 'uint64',
+    # Epoch when validator activated
+    'activation_epoch': 'uint64',
+    # Epoch when validator exited
+    'exit_epoch': 'uint64',
+    # Epoch when validator withdrew
+    'withdrawal_epoch': 'uint64',
+    # Epoch when validator was penalized
+    'penalized_epoch': 'uint64',
     # Exit counter when validator exited
     'exit_count': 'uint64',
     # Status flags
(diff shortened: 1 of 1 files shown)


djrtwo:
@JustinDrake @vbuterin @hwwhww 

This is not a major fundamental change, but it is pretty invasive and would require a bunch of little code changes for existing implementations. That said, I think this approach reasonably cleans up a lot of the epoch normalizing arithmetic that was ending up everywhere. I think doing the same for finality and crosslinks would make it even cleaner.

Not set on it. Just wanted to get this out there for discussion.

vbuterin:
Yeah I definitely think this is a good idea. I'm happy to get this out now or later.

> could also clean up a lot of calls of slot_to_epoch(state.slot) if we just store the current epoch in state.

I'd say it would be simpler to just make a helper `current_epoch(state)`; it fits with our tradition of being state-minimalist lately.

specs/core/0_beacon-chain.md
```diff
@@ -988,33 +1002,31 @@ def get_randao_mix(state: BeaconState,
 
 ```python
 def get_active_index_root(state: BeaconState,
-                          slot: int) -> Bytes32:
+                          epoch: int) -> Bytes32:
     """
     Returns the index root at a recent ``slot``.
```
hwwhww: slot -> epoch

```suggestion
    Returns the index root at a recent ``epoch``.
```

specs/core/0_beacon-chain.md
```diff
@@ -866,29 +883,26 @@ def get_committee_count_per_slot(active_validator_count: int) -> int:
...
-    seed = xor(seed, int_to_bytes32(slot))
+    seed = xor(seed, int_to_bytes32(epoch))
     shuffled_active_validator_indices = shuffle(active_validator_indices, seed)
 
     # Split the shuffled list into epoch_length * committees_per_slot pieces
     return split(shuffled_active_validator_indices, committees_per_slot * EPOCH_LENGTH)
 ```
 
-**Invariant**: if `get_shuffling(seed, validators, slot)` returns some value `x` for some `slot <= state.slot + ENTRY_EXIT_DELAY`, it should return the same value `x` for the same `seed` and `slot` and possible future modifications of `validators` forever in phase 0, and until the ~1 year deletion delay in phase 2 and in the future.
+**Invariant**: if `get_shuffling(seed, validators, epoch)` returns some value `x` for some `epoch <= current_epoch(state) + ENTRY_EXIT_DELAY`, it should return the same value `x` for the same `seed` and `epoch` and possible future modifications of `validators` forever in phase 0, and until the ~1 year deletion delay in phase 2 and in the future.
```
vbuterin: Shouldn't ENTRY_EXIT_DELAY be denominated in epochs?

specs/core/0_beacon-chain.md
```diff
@@ -988,33 +1002,31 @@ def get_randao_mix(state: BeaconState,
...
 def generate_seed(state: BeaconState,
-                  slot: int) -> Bytes32:
+                  epoch: int) -> Bytes32:
     """
-    Generate a seed for the given ``slot``.
+    Generate a seed for the given ``epoch``.
     """
-    if slot < SEED_LOOKAHEAD:
-        randao_mix_slot = GENESIS_SLOT
+    if epoch < SEED_LOOKAHEAD:
```
vbuterin: Wait, why are we checking `slot` or `epoch` against a constant anyway? Shouldn't this be `epoch < current_epoch(state) + SEED_LOOKAHEAD`? Also, shouldn't `SEED_LOOKAHEAD` be denominated in epochs?

specs/core/0_beacon-chain.md
```diff
@@ -1367,10 +1380,10 @@ def process_deposit(state: BeaconState,
             pubkey=pubkey,
             withdrawal_credentials=withdrawal_credentials,
             proposer_slots=0,
-            activation_slot=FAR_FUTURE_SLOT,
-            exit_slot=FAR_FUTURE_SLOT,
-            withdrawal_slot=FAR_FUTURE_SLOT,
-            penalized_slot=FAR_FUTURE_SLOT,
+            activation_epoch=FAR_FUTURE_EPOCH,
+            exit_epch=FAR_FUTURE_EPOCH,
```
vbuterin: epch -> epoch

specs/core/0_beacon-chain.md
```diff
@@ -1644,37 +1657,37 @@ Define the following helpers to process attestation inclusion rewards and inclus
 
 ### Eth1 data
 
-If `state.slot % ETH1_DATA_VOTING_PERIOD == ETH1_DATA_VOTING_PERIOD - 1`:
+If `current_epoch(state) % ETH1_DATA_VOTING_PERIOD == 0`:
```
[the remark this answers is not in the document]
vbuterin: Not ETH1_DATA_VOTING_PERIOD-1? I guess it's ok either way...

JustinDrake:
Feels like a natural move to segregate slot-native stuff (block proposals/roots, attestations, inclusion distance, randao). A few comments:

1) In the "Beacon chain fork choice rule" section there are a few places where epochs make more sense than slots. For example:
    * be the finalized block with the highest slot number => highest epoch
    * at least EPOCH_LENGTH slots => at least 1 epoch
    * justified_head be the descendant of finalized_head with the highest slot number => highest epoch
2) In `Exit` objects I suggest using `epoch`. The reason is that voluntary exits are processed at epoch boundaries. It's also more consistent with `Validator.exit_epoch`.
3) Suggest a new (admittedly redundant) constant `GENESIS_EPOCH = slot_to_epoch(GENESIS_SLOT)`.
4) Returns the index root at a recent ``slot``. => epoch
5) `get_randao_mix` => only called at epoch boundaries, use epochs
6) I suggest using epochs in `Fork` objects. It's usual for forks to fall at round numbers, and having forks at epoch boundaries seems like a sane thing to have.
7) Need a new type (in the context of typehinting https://github.com/ethereum/eth2.0-specs/pull/477)

specs/core/0_beacon-chain.md
```diff
@@ -1585,26 +1598,26 @@ For each `exit` in `block.body.exits`:
...
-The steps below happen when `state.slot % EPOCH_LENGTH == EPOCH_LENGTH - 1`.
+The steps below happen when `(state.slot + 1) % EPOCH_LENGTH == 0`.
 
 ### Helpers
 
-* Let `next_epoch_start_slot = state.slot + 1`.
-* Let `current_epoch_start_slot = state.slot - (EPOCH_LENGTH + 1)`.
-* Let `previous_epoch_start_slot = state.slot - 2 * EPOCH_LENGTH + 1` if `state.slot > GENESIS_SLOT + EPOCH_LENGTH` else `current_epoch_start_slot`.
-* Let `next_epoch = next_epoch_start_slot // EPOCH_LENGTH`.
-* Let `current_epoch = current_epoch_start_slot // EPOCH_LENGTH`.
```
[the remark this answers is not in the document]
djrtwo: you're right. looks good

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
djrtwo: hmmm it would have a no-op essentially at the end of the 0th epoch and then get on track.
I kind of like not having `- 1` but don't feel strongly.

specs/core/0_beacon-chain.md
djrtwo: This is preventing underflow on the `randao_mix_epoch` when current_epoch is less than `SEED_LOOKAHEAD`.

djrtwo:
for `5`, we actually should be using `get_randao_mix` in per-slot processing instead of directly accessing the array. I vote keep as is

specs/core/0_beacon-chain.md
```diff
@@ -922,25 +936,25 @@ def get_crosslink_committees_at_slot(state: BeaconState,
     """
     Returns the list of ``(committee, shard)`` tuples for the ``slot``.
     """
-    state_epoch_slot = state.slot - (state.slot % EPOCH_LENGTH)
-    assert state_epoch_slot <= slot + EPOCH_LENGTH
-    assert slot < state_epoch_slot + EPOCH_LENGTH
+    current_epoch_slot = current_epoch(state) * EPOCH_LENGTH
+    assert current_epoch_slot <= slot + EPOCH_LENGTH
+    assert slot < current_epoch_slot + EPOCH_LENGTH
```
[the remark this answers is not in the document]

djrtwo:
Addressed all feedback except Justin's `5`, and a question from Vitalik I replied to.

Ready for another look @JustinDrake @hwwhww @vbuterin

JustinDrake:
1. Can you sync with Master? I don't see the `Epoch` type.
2. "update last finalized slot if possible" => epoch
3. Feels a bit strange to have `previous_epoch_start_slot`, `current_epoch_start_slot`, `next_epoch_start_slot` within the per-epoch processing
4. Can we get rid of the "per_slot" in the various helpers, e.g. `get_current_epoch_committee_count_per_slot`. Seems strange to work per slot, and then multiply by `EPOCH_LENGTH`, e.g. `range(get_current_epoch_committee_count_per_slot(state) * EPOCH_LENGTH` (done for two out of the three times this function is called).

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
vbuterin: `if next_epoch % ETH1_DATA_VOTING_PERIOD == 0` ?

specs/core/0_beacon-chain.md
vbuterin: Aah I see.

djrtwo:
handled 1, 2, 3, and 5

I'm a bit unsure of changing 4 before we have the locally calculable shuffling. I think we might want the per slot helper in that case.

specs/core/0_beacon-chain.md
```diff
@@ -787,23 +792,44 @@ Note: We aim to migrate to a S[T/N]ARK-friendly hash function in a future Ethere
...
+```python
+def get_current_epoch(state: BeaconState) -> EpochNumber:
+    return slot_to_epoch(state.slot)
+```
+
+#### `get_epoch_start_slot`
+
+```python
+def get_epoch_start_slot(epoch: EpochNumber) -> SlotNumber:
+    return slot_to_epoch(state.slot)
```
benjaminion: Ought this be

```python
    return epoch * EPOCH_LENGTH
```

or something like that?

JustinDrake:
I'm tempted to only store epoch boundary values in `latest_randao_mixes`. (And then make `get_randao_mix` use epochs.)

specs/core/0_beacon-chain.md
JustinDrake: It's frustating that `get_block_root` is called only at epoch boundaries, with only one exception.

djrtwo: Yup, that's a copy pasta error.

`epoch_start_to_slot` is okay, but I think `epoch_to_slot` is not descriptive enough.

djrtwo:
@JustinDrake I'm in favor. At epoch transition, copy over the current epoch to next in the array. just xor into that index at each block during the epoch.

specs/core/0_beacon-chain.md
djrtwo: changed this to remove the special case.

```python
randao_mix_epoch = epoch + LATEST_RANDAO_MIXES_LENGTH - SEED_LOOKAHEAD
```

djrtwo:
okay, addressed everything. ready for review again

specs/core/0_beacon-chain.md
```diff
@@ -168,9 +171,9 @@ Code snippets appearing in `this style` are to be interpreted as Python code. Be
 | `MAX_BALANCE_CHURN_QUOTIENT` | `2**5` (= 32) | - |
 | `BEACON_CHAIN_SHARD_NUMBER` | `2**64 - 1` | - |
 | `MAX_CASPER_VOTES` | `2**10` (= 1,024) | votes |
-| `LATEST_BLOCK_ROOTS_LENGTH` | `2**13` (= 8,192) | block roots |
-| `LATEST_RANDAO_MIXES_LENGTH` | `2**13` (= 8,192) | randao mixes |
-| `LATEST_INDEX_ROOTS_LENGTH` | `2**13` (= 8,192) | index roots |
+| `LATEST_BLOCK_ROOTS_LENGTH` | `2**13` (= 8,192) | slots |
```
JustinDrake: Move to "Time constants"?

PR #500, opened 2019-01-26
Fix field name error on eth1_data

paulhauner:
## What

Change `eth1_data_vote.data` > `eth1_data_vote.eth1_data`

## Why

`Eth1DataVote` does not have a field `data`.


hwwhww:
Maybe we should rename `Eth1DataVote.eth1_data` to `Eth1DataVote.data`?

decanus:
@JustinDrake I feel like short form is clear enough already, so I don't see the value in long form everywhere.

djrtwo:
I lean toward the short form -- `data`

PR #502, opened 2019-01-27
Update `Beacon chain state transition function` chapter intro

hwwhww:
- The current `per-slot` description should be for `per-block`.
- Add the real `per-slot` transition description.
- Personally feel that indicating "only affects a parts of the `state`" and "affects the entire `state`"  is not very helpful information here - they all affect the state anyway. 😅


djrtwo:
I agree with these!
But let's wait for #492. I made a couple of modifications to this section there.

Issue #508, opened 2019-01-28
[BLS] Modular square root issue


mratsim:
### Problem

The BLS test vectors in https://github.com/ethereum/eth2.0-tests/pull/12 and https://github.com/ethereum/eth2.0-tests/pull/13 reuse the Trinity `modular_squareroot`implementation at https://github.com/ethereum/trinity/blob/a1b0f058e7bc8e385c8dac3164c49098967fd5bb/eth2/_utils/bls.py#L63-L77

```python
def modular_squareroot(value: FQ2) -> FQP:
    """
    ``modular_squareroot(x)`` returns the value ``y`` such that ``y**2 % q == x``,
    and None if this is not possible. In cases where there are two solutions,
    the value with higher imaginary component is favored;
    if both solutions have equal imaginary component the value with higher real
    component is favored.
    """
    candidate_squareroot = value ** ((FQ2_order + 8) // 16)
    check = candidate_squareroot ** 2 / value
    if check in eighth_roots_of_unity[::2]:
        x1 = candidate_squareroot / eighth_roots_of_unity[eighth_roots_of_unity.index(check) // 2]
        x2 = FQ2([-x1.coeffs[0], -x1.coeffs[1]])  # x2 = -x1
        return x1 if (x1.coeffs[1], x1.coeffs[0]) > (x2.coeffs[1], x2.coeffs[0]) else x2
    return None
```

Those are a copy paste of the current spec https://github.com/ethereum/eth2.0-specs/blob/928f9772fa51ed2e251b99c4795097da976c4bde/specs/bls_signature.md#modular_squareroot

```python
Fq2_order = q ** 2 - 1
eighth_roots_of_unity = [Fq2([1,1]) ** ((Fq2_order * k) // 8) for k in range(8)]

def modular_squareroot(value: int) -> int:
    candidate_squareroot = value ** ((Fq2_order + 8) // 16)
    check = candidate_squareroot ** 2 / value
    if check in eighth_roots_of_unity[::2]:
        x1 = candidate_squareroot / eighth_roots_of_unity[eighth_roots_of_unity.index(check) // 2]
        x2 = -x1
        return x1 if (x1.coeffs[1].n, x1.coeffs[0].n) > (x2.coeffs[1].n, x2.coeffs[0].n) else x2
    return None
```

However, our experiments using Milagro modular square root showed that the spec does not match with established crypto library.

[...]

PR #509, opened 2019-01-28
Modular squareroot clarification

vbuterin:
Clarifies #508


djrtwo:
CC @mratsim

hwwhww:
@vbuterin 

In `modular_squareroot`:

> return x1 if (x1.coeffs[1].n, x1.coeffs[0].n) > (x2.coeffs[1].n, x2.coeffs[0].n) else x2

In Python implementation, it seems `FQP.coeffs` could be (i) a tuple of `int` **or** (ii) a tuple of `FQ`? In this use case (`hash_to_G2` to `modular_squareroot`), `x1.coeffs` is set by integers. So I think either:
1. There's no `.n` here: `return x1 if (x1.coeffs[1], x1.coeffs[0]) > (x2.coeffs[1], x2.coeffs[0]) else x2`
2. I missed to cast something.
3. `py_ecc` bug

vbuterin:
Gah, that's ugly. OK, see latest commit.

PR #512, opened 2019-01-28
Cleanup state transition org and explanation

djrtwo:
Moves a bunch of stuff out of the state transition section and clarifies the intention around how to process the block by block transition.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -50,80 +50,86 @@
             - [`Eth1Data`](#eth1data)
             - [`Eth1DataVote`](#eth1datavote)
     - [Custom Types](#custom-types)
+    - [Helper functions](#helper-functions)
+        - [`hash`](#hash)
+        - [`hash_tree_root`](#hash_tree_root)
+        - [`slot_to_epoch`](#slot_to_epoch)
+        - [`get_current_epoch`](#get_current_epoch)
+        - [`get_epoch_start_slot`](#get_epoch_start_slot)
+        - [`is_active_validator`](#is_active_validator)
+        - [`get_active_validator_indices`](#get_active_validator_indices)
+        - [`shuffle`](#shuffle)
+        - [`split`](#split)
+        - [`get_epoch_committee_count`](#get_epoch_committee_count)
+        - [`get_shuffling`](#get_shuffling)
+        - [`get_previous_epoch_committee_count`](#get_previous_epoch_committee_count)
+        - [`get_current_epoch_committee_count`](#get_current_epoch_committee_count)
+        - [`get_crosslink_committees_at_slot`](#get_crosslink_committees_at_slot)
+        - [`get_block_root`](#get_block_root)
+        - [`get_randao_mix`](#get_randao_mix)
+        - [`get_active_index_root`](#get_active_index_root)
+        - [`generate_seed`](#generate_seed)
+        - [`get_beacon_proposer_index`](#get_beacon_proposer_index)
+        - [`merkle_root`](#merkle_root)
+        - [`get_attestation_participants`](#get_attestation_participants)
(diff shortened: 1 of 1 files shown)


djrtwo:
Just added some notes to help parse what's going on here

djrtwo:
just handled merge conflict from master

specs/core/0_beacon-chain.md
```diff
@@ -1513,54 +1266,319 @@ def penalize_validator(state: BeaconState, index: ValidatorIndex) -> None:
...
+    while 1:
+        children = get_children(store, head)
+        if len(children) == 0:
+            return head
+        head = max(children, key=get_vote_count)
+```
+
+## Beacon chain state transition function
+
+We now define the state transition function. This is a parent to child block transition. At a high level the state transition is made up of three parts:
```
JustinDrake: What does "This is a parent to child block transition." mean?

specs/core/0_beacon-chain.md
```diff
@@ -1513,54 +1266,319 @@ def penalize_validator(state: BeaconState, index: ValidatorIndex) -> None:
...
+
+1. The per-slot transitions, which happens every slot.
+2. The per-block transitions, which happens at every block.
+3. The per-epoch transitions, which happens at every epoch boundary (i.e. `(state.slot + 1) % EPOCH_LENGTH == 0`).
```
JustinDrake: "which happens at every epoch boundary" => More specifically, at the last slot of every epoch.

specs/core/0_beacon-chain.md
```diff
@@ -1680,11 +1698,11 @@ For each `exit` in `block.body.exits`:
...
-## Per-epoch processing
+### Per-epoch processing
 
 The steps below happen when `(state.slot + 1) % EPOCH_LENGTH == 0`.
 
-### Helpers
+#### Helpers
 
 * Let `current_epoch = get_current_epoch(state)`.
 * Let `previous_epoch = current_epoch - 1 if current_epoch > GENESIS_EPOCH else current_epoch`.
```
JustinDrake: Can this be simplified now that `GENESIS_SLOT` is 2**19?

specs/core/0_beacon-chain.md
```diff
@@ -1947,13 +1965,13 @@ def process_penalties_and_exits(state: BeaconState) -> None:
...
 
-### Final updates
+#### Final updates
 
 * Set `state.latest_penalized_balances[(next_epoch) % LATEST_PENALIZED_EXIT_LENGTH] = state.latest_penalized_balances[current_epoch % LATEST_PENALIZED_EXIT_LENGTH]`.
 * Set `state.latest_randao_mixes[next_epoch % LATEST_RANDAO_MIXES_LENGTH] = get_randao_mix(state, current_epoch)`.
 * Remove any `attestation` in `state.latest_attestations` such that `slot_to_epoch(attestation.data.slot) < current_epoch`.
 
-## State root processing
+### State root verification
```
JustinDrake: It's tempting to put this at the end of per-block processing, .e.g. "Verify `block.state_root == hash_tree_root(state)`   after the per-epoch processing below."

vbuterin:
The last unresolved "must do" in #358 is `slashable_vote_data` -> `vote_data`. Is this even a good idea? I think maybe no, because the only place we use that variable and its type is for slashing, so "slashable" is a very accurate and useful descriptor. But if it is, then we should do it.

specs/core/0_beacon-chain.md
djrtwo: eh, I was trying to clarify that the transition function is a from post state of parent block to post state of child block. I'll clean it up

specs/core/0_beacon-chain.md
djrtwo: I see the reasoning, but at the end of the day, the implementation will have it here (after the epoch processing) so I'd lean toward keeping as is.

PR #518, opened 2019-01-29
validator guide to recent spec changes

djrtwo:
* Catches up the phase 0 validator guide to the recent spec changes. 
* fixes a minor bug in spec when calling `get_bitfield_bit` in a couple of places

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1638,8 +1638,8 @@ For each `attestation` in `block.body.attestations`:
     assert attestation.aggregation_bitfield != b'\x00' * len(attestation.aggregation_bitfield)
 
     for i in range(len(crosslink_committee)):
-        if get_bitfield_bit(attestation.aggregation_bitfield) == 0b0:
-            assert get_bitfield_bit(attestation.custody_bitfield) == 0b0
+        if get_bitfield_bit(attestation.aggregation_bitfield, i) == 0b0:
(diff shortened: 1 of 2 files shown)


specs/validator/0_beacon-chain-validator.md
```diff
@@ -312,40 +312,40 @@ Set `attestation.aggregate_signature = signed_attestation_data` where `signed_at
 ```python
 attestation_data_and_custody_bit = AttestationDataAndCustodyBit(
     attestation.data,
-    False,
+    0b0,
```
vbuterin: Maybe just use 0? 0b0 looks better to an experienced python coder but may give a "wtf m8" reaction to non-pythonistas.

specs/core/0_beacon-chain.md
```diff
@@ -946,7 +946,7 @@ def get_attestation_participants(state: BeaconState,
     participants = []
     for i, validator_index in enumerate(crosslink_committee):
         aggregation_bit = get_bitfield_bit(bitfield, i)
-        if aggregation_bit == 0b1:
+        if aggregation_bit == 1:
```
JustinDrake: Why change from `0b1` to `1`? Making it clear that it's a bit with the `0b` prefix seems like a win to me.

djrtwo: Can debate @vbuterin when he gets here.
I lean toward `0b` prefix but don't have a dog in this fiht.

PR #520, opened 2019-01-30
[WIP] helpers and notes for shuffling lookahead

djrtwo:
WIP

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -800,6 +800,17 @@ def get_current_epoch_committee_count(state: BeaconState) -> int:
     return get_epoch_committee_count(len(current_active_validators))
 ```
 
+### `get_next_epoch_committee_count`
+
+```python
+def get_next_epoch_committee_count(state: BeaconState) -> int:
+    next_active_validators = get_active_validator_indices(
+        state.validator_registry,
+        get_current_epoch(state) + 1,
+    )
+    return get_epoch_committee_count(len(next_active_validators))
+```
+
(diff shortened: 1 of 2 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -925,6 +957,7 @@ def get_beacon_proposer_index(state: BeaconState,
     """
     first_committee, _ = get_crosslink_committees_at_slot(state, slot)[0]
     return first_committee[slot % len(first_committee)]
+
```
ralexstokes: remove newline :)

specs/core/0_beacon-chain.md
```diff
@@ -966,6 +998,19 @@ def get_attestation_participants(state: BeaconState,
...
+```
+def is_power_of_two(value: int) -> bool:
+    """
+    Check if ``value`` is a power of two integer.
+    """
+    if value == 0:
+        return False
+    else:
+        return 2**int(math.log2(value)) == value
```
vbuterin: IMO this is dangerous at least if you do it in python. I would do:

```
def is_power_of_two(value: int) -> bool:
    """
    Check if ``value`` is a power of two integer.
    """
    if value == 0:
        return False
    elif value == 1:
        return True
    else:
        return value % 2 == 0 and is_power_of_two(value // 2)
```

vbuterin: Actually never mind, I verified that it works fine for all powers of 2 up to 10000 and I think because of how floating point rounding works it actually should always work. So the existing definition could be ok.

PR #530, opened 2019-01-30
Fix missed updates from slot to epoch

benjaminion:
And a `shard_committee` to `crosslink_committee`.


specs/validator/0_beacon-chain-validator.md
```diff
@@ -275,13 +275,13 @@ Set `attestation_data.latest_crosslink_root = state.latest_crosslinks[shard].sha
 
 ##### Justified slot
```
vbuterin:
```suggestion
##### Justified epoch
```

PR #532, opened 2019-01-30
Fix to verify bitfields and aggregate signature.

benjaminion:
In block processing, `crosslink_committee` is not defined. This is a fix.

In Phase 0 this check is redundant since we've already asserted that the
custody bitfield is all zero, but it will matter in later phases.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1632,7 +1632,8 @@ For each `attestation` in `block.body.attestations`:
     assert attestation.custody_bitfield == b'\x00' * len(attestation.custody_bitfield)  # [TO BE REMOVED IN PHASE 1]
     assert attestation.aggregation_bitfield != b'\x00' * len(attestation.aggregation_bitfield)
 
-    for i in range(len(crosslink_committee)):
+    assert len(attestation.aggregation_bitfield) <= len(attestation.custody_bitfield)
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1631,8 +1631,15 @@ For each `attestation` in `block.body.attestations`:
...
-    for i in range(len(crosslink_committee)):
+    
+    crosslink_committee = [
+        committee for committee, shard in get_crosslink_committees_at_slot(state, attestation.data.slot)
+        if shard == attestation.data.shard
+    ][0]
+    verify_bitfield(attestation.aggregation_bitfield, len(crosslink_committee))
+    verify_bitfield(attestation.custody_bitfield, len(crosslink_committee))
+    
+    for i in range(len(crosslink_committee):
```
djrtwo: @JustinDrake Is this for loop necessary? Aren't we already checking this using 
`assert attestation.custody_bitfield == b'\x00' * len(attestation.custody_bitfield)`?

hwwhww: I have the same question here. Can we remove this for-loop? @djrtwo @JustinDrake

hwwhww: Oh I got it, because `assert attestation.custody_bitfield == b'\x00' * len(attestation.custody_bitfield)` is only valid in Phase 0.

PR #534, opened 2019-01-31
name changes

djrtwo:
- [x] penalized -> slashed
- [x] exit -> voluntary exit
- [x] operations -> transactions
- [x] initial, startup -> genesis
- [x] `ChainStart` -> `Eth2Genesis`
- [x] Remove `Number` from custom types 
- [x] `SEED_LOOKAHEAD` -> `MIN_SEED_LOOKAHED`
- [x] `ENTRY_EXIT_DELAY` -> `ACTIVATION_EXIT_DELAY`
- [x] `INCLUDER_REWARD_QUOTIENT` -> `ATTESTATION_INCLUSION_REWARD_QUOTIENT`
- [x] `latest_index_roots` -> `latest_active_index_roots`
- [x] `current_epoch_start_shard` -> `current_shuffling_start_shard`, `current_shuffling_epoch`, `current_shuffling_seed`
- [x] `MIN_VALIDATOR_WITHDRAWAL_EPOCHS` -> `MIN_VALIDATOR_WITHDRAWAL_DELAY`
- [x] `MAX_WITHDRAWALS_PER_EPOCH` -> `MAX_EXIT_DEQUEUES_PER_EPOCH`
- [x] `ETH1_DATA_VOTING_PERIOD` -> `EPOCHS_PER_ETH1_VOTING_PERIOD`
- [x] `SLOT_DURATION` -> `SECONDS_PER_SLOT`
- [x] `EPOCH_LENGTH` -> `SLOTS_PER_EPOCH`

**Exits/penalties clean-up recap**

* exits
	* voluntary exits
	* forceful exits
		* slashings
		* ejections
* penalties
	* base penalties
	* inactivity penalties
	* (partial) slashings

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -19,10 +19,10 @@
         - [State list lengths](#state-list-lengths)
         - [Reward and penalty quotients](#reward-and-penalty-quotients)
         - [Status flags](#status-flags)
-        - [Max operations per block](#max-operations-per-block)
+        - [Max transactions per block](#max-transactions-per-block)
         - [Signature domains](#signature-domains)
     - [Data structures](#data-structures)
-        - [Beacon chain operations](#beacon-chain-operations)
+        - [Beacon chain transactions](#beacon-chain-transactions)
             - [Proposer slashings](#proposer-slashings)
                 - [`ProposerSlashing`](#proposerslashing)
             - [Attester slashings](#attester-slashings)
@@ -36,8 +36,8 @@
                 - [`Deposit`](#deposit)
                 - [`DepositData`](#depositdata)
                 - [`DepositInput`](#depositinput)
-            - [Exits](#exits)
-                - [`Exit`](#exit)
+            - [Voluntary exits](#voluntary-exits)
+                - [`VoluntaryExit`](#voluntaryexit)
         - [Beacon chain blocks](#beacon-chain-blocks)
             - [`BeaconBlock`](#beaconblock)
             - [`BeaconBlockBody`](#beaconblockbody)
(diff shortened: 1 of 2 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -606,14 +606,14 @@ We define the following Python custom types for type hinting and readability:
 
 | Name | SSZ equivalent | Description |
 | - | - | - |
-| `SlotNumber` | `uint64` | a slot number |
-| `EpochNumber` | `uint64` | an epoch number |
-| `ShardNumber` | `uint64` | a shard number |
+| `Slot` | `uint64` | a slot number |
+| `Epoch` | `uint64` | an epoch number |
+| `Shard` | `uint64` | a shard number |
```
hwwhww: The reason why I named it `ShardNumber` is that I assumed that we will have other ID schema for identifying the shard chain, maybe `ShardID`.

djrtwo: I see... I think in terms of the spec, we would treat those as the same type and value. I understand implementation might choose to use two types.

djrtwo:
cc: @vbuterin regarding "operations -> transactions".

JustinDrake:
😂 

screenshot 2019-02-12 at 19 48 39

JustinDrake:
Substrates calls the operations/transactions "extrinsics". See https://www.youtube.com/watch?v=GDvV-Vn2u3w&feature=youtu.be&t=928

prestonvanloon:
FYI: we are not renaming operations to transactions in Prysm. We generally think this should be reserved for user level and operations is the more intuitive name. 

We are also not renaming chainstart to eth2genesis. For the same reasons: more intuitive with the previous name.

PR #537, opened 2019-01-31
Use 2*63 for the genesis slot

JustinDrake:


Nashatyrev:
Slot `#‭576460752303423590‬6` doesn't look too user friendly. 
From engineering point of view: the only reference point when understanding/debugging Eth1 blockchain issues was always the block number. This change will complicate the things. 

BTW `state.slot << 1` wouldn't be `[slots since genesis] * 2` then ?

JustinDrake:
> Slot #‭576460752303423590‬6 doesn't look too user friendly.

Right, this number should not be exposed to users. Abstracting the most significant bit at a very low level would make sense.

> `state.slot << 1` wouldn't be `[slots since genesis] * 2` then ?

`(state.slot << 1) >> 1` is what you want :)

djrtwo:
@Nashatyrev Given the `(slot << 1) >>1`  or `slot - 2**63` does this satisfy the debugging requirement? 

Starting at slot/epoch 0 requires adding a bunch of conditional underflow logic that will only be needed in the first few epochs.

vbuterin:
Seems reasonable to me. As said above, we can have a UI convention that we subtract 2**63.

PR #543, opened 2019-02-01
Hash_tree_root -> hash_tree_root_internal

vbuterin:
Clarifies the distinction between "internal" hash roots (may be < 32 bytes for trivial objects) and "external" ones (zpadded to 32).


hwwhww:
@vbuterin  could you change it to be on `dev` branch?

PR #548, opened 2019-02-01
Fix typo `tree_hash_root` -> `hash_tree_root`

hwwhww:


specs/core/0_beacon-chain.md
```diff
@@ -1660,7 +1660,7 @@ Below are the processing steps that happen at every slot.
 
 #### Block roots
 
-* Let `previous_block_root` be the `tree_hash_root` of the previous beacon block processed in the chain.
+* Let `previous_block_root` be the `hash_tree_root` of the previous beacon block processed in the chain.
```
JustinDrake: Somewhat inconsistent with "Tree hash" in https://github.com/ethereum/eth2.0-specs/blob/master/specs/simple-serialize.md#tree-hash

djrtwo: I'd say call that section "Hash Tree"

djrtwo: * It's a hash-tree. -- we get the hash-tree root.
* It's a special type of hashing, "tree hashing" -- we tree hash the object.

Both are fine but we should pick one and be consistent. I'm in favor of `hash_tree_root` and calling it a hash tree. For the verb we "get the root of the hash-tree"

djrtwo: So what do we want to do here?

hwwhww: Personally, I'm fine with having some "tree hash" in the SSZ spec, kind of getting used to it, and I think we say it frequently...?

And we also have `tree_hash.py` in `py-ssz`...

djrtwo: yeah, that's fine. approved? Merge?

PR #562, opened 2019-02-02
Removes the source of the Vyper contract from the spec.

ralexstokes:
Following up on this comment: https://github.com/ethereum/eth2.0-specs/pull/388#pullrequestreview-191008164

This change allows for easier maintenance of the code and the spec by
uncoupling them. Before any edit to either document resulted in having to
synchronize the other. By adding a reference to the canonical repo for the code
we avoid having to maintain a duplicate copy here.

Some things to consider:
* do we like how the interface is specified here? should it just be dropped?
* do we want a direct link to the Vyper source file? (right now just links to repo)

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1378,7 +1378,7 @@ When sufficiently many full deposits have been made the deposit contract emits t
 
 ### Vyper code
 
-The source for the Vyper contract lives in a separate repository at [https://github.com/ethereum/deposit_contract](https://github.com/ethereum/deposit_contract).
+The source for the Vyper contract lives in a [separate repository](https://github.com/ethereum/deposit_contract) at [https://github.com/ethereum/deposit_contract/blob/master/deposit_contract/contracts/validator_registration.v.py](https://github.com/ethereum/deposit_contract/blob/master/deposit_contract/contracts/validator_registration.v.py).
 
 Note: to save ~10x on gas this contract uses a somewhat unintuitive progressive Merkle root calculation algo that requires only O(log(n)) storage. See https://github.com/ethereum/research/blob/master/beacon_chain_impl/progressive_merkle_tree.py for an implementation of the same algo in python tested for correctness.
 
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1378,86 +1378,15 @@ When sufficiently many full deposits have been made the deposit contract emits t
...
-        i += 1
-        power_of_two *= 2
-    value:bytes32 = sha3(deposit_data)
-    for j in range(32):
-        if j < i:
-            value = sha3(concat(self.branch[j], value))
-    self.branch[i] = value
-
-    self.deposit_count += 1
+The source for the Vyper contract lives in a separate repository at [https://github.com/ethereum/deposit_contract](https://github.com/ethereum/deposit_contract).
```
djrtwo: I think we should link directly to the contract here.
It's kind of buried in the repo and will be the thing most likely to be found when reading this spec.

specs/core/0_beacon-chain.md
```diff
@@ -1378,86 +1378,15 @@ When sufficiently many full deposits have been made the deposit contract emits t
...
-            chainstart_time: bytes[8] = slice(concat("", convert(timestamp_day_boundary, bytes32)), start=24, len=8)
-            log.ChainStart(new_deposit_root, chainstart_time)
-            self.chainStarted = True
-```
+For convenience, we provide the interface to the contract here:
 
-Note: to save ~10x on gas this contract uses a somewhat unintuitive progressive Merkle root calculation algo that requires only O(log(n)) storage. See https://github.com/ethereum/research/blob/master/beacon_chain_impl/progressive_merkle_tree.py for an implementation of the same algo in python tested for correctness.
+* `__init__()`: initializes the contract
+* `get_deposit_root() -> bytes32`: returns the current root of the deposit tree
+* `deposit(bytes[512])`: adds a deposit instance to the deposit tree, incorporating the input argument and the value transferred in the given call.
```
djrtwo: Worth noting the gwei value transfered needs to be within `MIN_DEPOSIT_AMOUNT` and `MAX_DEPOSIT_AMOUNT`. I think this is pretty crucial to the short description of the api.

Issue #556, opened 2019-02-02
Little vs big endian


jannikluhn:
SSZ has been updated to little endian integer encoding, but the phase 0 spec still uses big endian encoding at a few of sites. Is there a specific reason for this? If not I'd advocate for using little endian here as well for the sake of consistency.

djrtwo:
Good call. Agreed on consistency.

hwwhww:
Hmmm, I feel *not* fully comfortable with this change for some reasons:
1. eth1.0 is still a big-endian world - so all the bytes conversion on deposit contract are still in big-endian.
2. Our BLS spec uses big-endian now. Moreover, [`BLS12-381`curve](https://github.com/zkcrypto/pairing/blob/183a64b08e9dc7067f78624ec161371f1829623e/src/bls12_381/README.md#serialization) uses big-endian now.

That's said, IMHO it's okay to only use little-endian in SSZ.

djrtwo:
Yeah, there is actually a bug in the spec right now. The amount and timestamp logged in `deposit_data` from the deposit contract is assumed to be in a well formatted SSZ `DepositData` object. If we are to keep ssz little endian, then we need to either do the conversion in the deposit contract or clean it up on the 2.0 side. I asked the vyper guys if there is a built-in for `int` to little-endian bytes and they said no. This is definitely an argument for big endian...


contract code snippet below:
```python
deposit_amount: bytes[8] = slice(concat("", convert(msg.value / GWEI_PER_ETH, bytes32)), start=24, len=8)
deposit_timestamp: bytes[8] = slice(concat("", convert(block.timestamp, bytes32)), start=24, len=8)
deposit_data: bytes[528] = concat(deposit_amount, deposit_timestamp, deposit_input)
```

arnetheduck:
My general take on this would be that for types that have a `canonical` serialization defined outside of the logic of the Eth2 spec, we should use `bytesX` types in the Eth2 spec - for example, a keypair is generally not considered to be an integer or a point on the curve, from the point of view of the consumer of the key - instead, it's just a blob that you pass as-is to the cryptography black box, and you let the cryptography black box define that serialization on its own terms, compartmentalized.

This goes for hashes, keys and other similar types.

[...]

PR #564, opened 2019-02-03
convert int_to_bytes to little endian

djrtwo:
addresses #556

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -710,8 +710,8 @@ def shuffle(values: List[Any], seed: Bytes32) -> List[Any]:
             if remaining == 1:
                 break
 
-            # Read 3-bytes of `source` as a 24-bit big-endian integer.
-            sample_from_source = int.from_bytes(source[position:position + rand_bytes], 'big')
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -710,8 +710,8 @@ def shuffle(values: List[Any], seed: Bytes32) -> List[Any]:
             if remaining == 1:
                 break
 
-            # Read 3-bytes of `source` as a 24-bit big-endian integer.
-            sample_from_source = int.from_bytes(source[position:position + rand_bytes], 'big')
+            # Read 3-bytes of `source` as a 24-bit little-endian integer.
+            sample_from_source = int.from_bytes(source[position:position + rand_bytes], 'little')
```
```suggestion
            sample_from_source = int_to_bytes3(source[position:position + rand_bytes])
```

djrtwo: We want to convert bytes-to-int, not int-to-bytes. `int.from_bytes` is correct, but definitely very pythonic

hwwhww: Oooh, sorry for the misreading.
And I added `bytes_to_int` in #576.

PR #571, opened 2019-02-04
Update fork choice rule

vbuterin:
* Use validator indices instead of validators
* Weight each validator by `int(balance)` at the position of the start state instead of one weight unit regardless of balance (using `int(balance)` instead of balance to reduce the number of recalculation events implementers will have to deal with).

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1578,7 +1578,7 @@ def lmd_ghost(store: Store, start_state: BeaconState, start_block: BeaconBlock) 
 
     def get_vote_count(block: BeaconBlock) -> int:
         return sum([
-            int(start_state.validator_balances[validator_index])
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1570,19 +1570,16 @@ def lmd_ghost(store: Store, start_state: BeaconState, start_block: BeaconBlock)
...
+        (validator_index, get_latest_attestation_target(store, validator_index))
+        for validator_index in active_validator_indices
     ]
 
     def get_vote_count(block: BeaconBlock) -> int:
-        return len([
-            target
-            for target in attestation_targets
+        return sum([
+            int(start_state.validator_balances[validator_index])
```
djrtwo: This should be `get_effective_balance(start_state.validator_balances[validator_index])`

```suggestion
            int(get_effective_balance(start_state.validator_balances[validator_index]))
```

PR #575, opened 2019-02-05
Use little endian bit order

arnetheduck:
Slightly more simple and common to find bit n at position `1 << n`

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1059,7 +1059,7 @@ def get_bitfield_bit(bitfield: bytes, i: int) -> int:
     """
     Extract the bit in ``bitfield`` at position ``i``.
     """
-    return (bitfield[i // 8] >> (7 - (i % 8))) % 2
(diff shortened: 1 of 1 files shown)


arnetheduck:
Relevant ref: https://en.wikipedia.org/wiki/Endianness#Bit_endianness

hwwhww:
Ah, sorry, I just realize, maybe we need to update `verify_bitfield` too.
```python
def verify_bitfield(bitfield: bytes, committee_size: int) -> bool:
    """
    Verify ``bitfield`` against the ``committee_size``.
    """
    if len(bitfield) != (committee_size + 7) // 8:
        return False

    # Check `bitfield` is padded with zero bits only
    for i in range(committee_size, len(bitfield) * 8):
        if get_bitfield_bit(bitfield, i) == 0b1:
            return False

    return True
```

The padding zero part.

hwwhww:
^^^^^ False alarm. 🤦‍♀️ 

@djrtwo @vbuterin can we merge it?

JustinDrake:
Agreed this is simpler

Issue #563, opened 2019-02-03
Swap-or-not shuffle


vbuterin:
For historical background see https://github.com/ethereum/eth2.0-specs/issues/323

Dan Boneh recommended in discussions that we consider the swap-or-not method for shuffling.

Here is the paper: https://link.springer.com/content/pdf/10.1007%2F978-3-642-32009-5_1.pdf

Especially see the "generalized domain" algorithm on page 3.

Here's an implementation:

```python
def hash(x): return blake2s(x).digest()[:32]

def values_at_position(n, positions, seed, precompute=False):
    values = positions[::]
    for round in range(32):
        if precompute:
            hashvalues = b''.join([
                hash(seed + round.to_bytes(1, 'big') + i.to_bytes(4, 'big'))
                for i in range((n + 255) // 256)
            ])
        pivot = int.from_bytes(hash(seed + round.to_bytes(1, 'big')), 'big') % n
        if precompute:
            def permute(pos):
                flip = (pivot - pos) % n
                maxpos = max(pos, flip)
                bit = (hashvalues[maxpos // 8] >> (maxpos % 8)) % 2
                return flip if bit else pos
        else:
            def permute(pos):
                flip = (pivot - pos) % n
                maxpos = max(pos, flip)
                h = hash(seed + round.to_bytes(1, 'big') + (maxpos // 256).to_bytes(4, 'big'))
                byte = h[(maxpos % 256) // 8]
                bit = (byte >> (maxpos % 8)) % 2
                return flip if bit else pos

        values = [permute(v) for v in values]
    return values

def swap_or_not_shuffle(values, seed):
    return [values[i] for i in values_at_position(len(values), list(range(len(values))), seed, True)]

def swap_or_not_shuffle_partial(values, seed, count):
    return [values[i] for i in values_at_position(len(values), list(range(count)), seed, False)]
```

It passes basic statistical tests for shuffling 9 elements. It seems to have all of the desired properties that we want (including being invertible: just run the rounds in reverse order), and is only slightly slower than the prime shuffle .

djrtwo:
So simple!

To enumerate the "desired properties":
* Provably (close to) uniform
* Ability to compute P(x) in O(1)
* Ability to compute P_inverse(x) in O(1)

djrtwo:
closed via #576

JustinDrake:
Possibly relevant paper from Eurocrypt https://eprint.iacr.org/2018/1011.pdf

PR #576, opened 2019-02-06
Introduce swap-or-not shuffle

vbuterin:
See #563 for discussion.

Here is a more efficient implementation for shuffling an entire set; it can live here until we come up with an explicit "efficient implementation" doc:

```python
def shuffle(list_size, seed):
    indices = list(range(list_size))
    for round in range(90):
        hash_bytes = b''.join([
            hash(seed + round.to_bytes(1, 'little') + (i).to_bytes(4, 'little'))
            for i in range((list_size + 255) // 256)
        ])
        pivot = int.from_bytes(hash(seed + round.to_bytes(1, 'little')), 'little') % list_size

        powers_of_two = [1, 2, 4, 8, 16, 32, 64, 128]
            
        for i, index in enumerate(indices):
            flip = (pivot - index) % list_size
            hash_pos = index if index > flip else flip
            byte = hash_bytes[hash_pos // 8]
            if byte & powers_of_two[hash_pos % 8]:
                indices[i] = flip
    return indices
```

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -718,7 +718,7 @@ def shuffle(values: List[Any], seed: Bytes32) -> List[Any]:
             return flip if bit else pos
 
         indices = [permute(v) for v in indices]
-    return [v[index] for index in indices]
+    return [values[index] for index in indices]
 ```
 
 ### `split`
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -697,53 +697,22 @@ def get_active_validator_indices(validators: List[Validator], epoch: EpochNumber
 
 ```python
 def shuffle(values: List[Any], seed: Bytes32) -> List[Any]:
-    """
-    Return the shuffled ``values`` with ``seed`` as entropy.
-    """
-    values_count = len(values)
+    indices = list(range(len(values)))
+    for round in range(90):
+        hashvalues = b''.join([
+            hash(seed + round.to_bytes(1, 'big') + i.to_bytes(4, 'big'))
+            for i in range((n + 255) // 256)
+        ])
+        pivot = int.from_bytes(hash(seed + round.to_bytes(1, 'big')), 'big') % n
+        
+        def permute(pos):
+            flip = (pivot - pos) % n
+            maxpos = max(pos, flip)
+            bit = (hashvalues[maxpos // 8] >> (maxpos % 8)) % 2
+            return flip if bit else pos
 
-    # Entropy is consumed from the seed in 3-byte (24 bit) chunks.
-    rand_bytes = 3
-    # The highest possible result of the RNG.
-    rand_max = 2 ** (rand_bytes * 8) - 1
-
-    # The range of the RNG places an upper-bound on the size of the list that
-    # may be shuffled. It is a logic error to supply an oversized list.
-    assert values_count < rand_max
-
-    output = [x for x in values]
-    source = seed
-    index = 0
-    while index < values_count - 1:
-        # Re-hash the `source` to obtain a new pattern of bytes.
-        source = hash(source)
-        # Iterate through the `source` bytes in 3-byte chunks.
-        for position in range(0, 32 - (32 % rand_bytes), rand_bytes):
-            # Determine the number of indices remaining in `values` and exit
-            # once the last index is reached.
-            remaining = values_count - index
-            if remaining == 1:
-                break
-
-            # Read 3-bytes of `source` as a 24-bit big-endian integer.
-            sample_from_source = int.from_bytes(source[position:position + rand_bytes], 'big')
-
-            # Sample values greater than or equal to `sample_max` will cause
-            # modulo bias when mapped into the `remaining` range.
-            sample_max = rand_max - rand_max % remaining
-
-            # Perform a swap if the consumed entropy will not cause modulo bias.
-            if sample_from_source < sample_max:
-                # Select a replacement index for the current index.
-                replacement_position = (sample_from_source % remaining) + index
-                # Swap the current index with the replacement index.
-                output[index], output[replacement_position] = output[replacement_position], output[index]
-                index += 1
-            else:
-                # The sample causes modulo bias. A new sample should be read.
-                pass
-
-    return output
+        indices = [permute(v) for v in indices]
+    return [v[index] for index in indices]
 ```
 
 ### `split`
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -707,10 +707,10 @@ def shuffle(values: List[Any], seed: Bytes32) -> List[Any]:
     indices = list(range(len(values)))
     for round in range(90):
         hashvalues = b''.join([
-            hash(seed + round.to_bytes(1, 'big') + i.to_bytes(4, 'big'))
+            hash(seed + round.to_bytes(1, 'little') + i.to_bytes(4, 'little'))
             for i in range((len(values) + 255) // 256)
         ])
-        pivot = int.from_bytes(hash(seed + round.to_bytes(1, 'big')), 'big') % n
+        pivot = int.from_bytes(hash(seed + round.to_bytes(1, 'little')), 'little') % n
         def permute(pos):
             flip = (pivot - pos) % n
             maxpos = max(pos, flip)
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -697,53 +697,22 @@ def get_active_validator_indices(validators: List[Validator], epoch: EpochNumber
...
-            else:
-                # The sample causes modulo bias. A new sample should be read.
-                pass
-
-    return output
+    indices = list(range(len(values)))
+    for round in range(90):
+        hashvalues = b''.join([
+            hash(seed + round.to_bytes(1, 'big') + i.to_bytes(4, 'big'))
+            for i in range((n + 255) // 256)
```
hwwhww: undefined `n`, should we set `n = len(values)` at the beginning?

specs/core/0_beacon-chain.md
```diff
@@ -697,53 +697,22 @@ def get_active_validator_indices(validators: List[Validator], epoch: EpochNumber
...
+        ])
+        pivot = int.from_bytes(hash(seed + round.to_bytes(1, 'big')), 'big') % n
+
```
```suggestion

```

specs/core/0_beacon-chain.md
```diff
@@ -697,53 +697,22 @@ def get_active_validator_indices(validators: List[Validator], epoch: EpochNumber
...
+        pivot = int.from_bytes(hash(seed + round.to_bytes(1, 'big')), 'big') % n
+        
+        def permute(pos):
+            flip = (pivot - pos) % n
+            maxpos = max(pos, flip)
+            bit = (hashvalues[maxpos // 8] >> (maxpos % 8)) % 2
+            return flip if bit else pos
+
+        indices = [permute(v) for v in indices]
+    return [v[index] for index in indices]
```
```suggestion
    return [values[index] for index in indices]
```

specs/core/0_beacon-chain.md
```diff
@@ -697,53 +697,22 @@ def get_active_validator_indices(validators: List[Validator], epoch: EpochNumber
...
-                replacement_position = (sample_from_source % remaining) + index
-                # Swap the current index with the replacement index.
-                output[index], output[replacement_position] = output[replacement_position], output[index]
-                index += 1
-            else:
-                # The sample causes modulo bias. A new sample should be read.
-                pass
-
-    return output
+    indices = list(range(len(values)))
```
djrtwo: keep doc string here

```suggestion
    """
    Return the shuffled ``values`` with ``see`` as entropy.
    
    Utilizes 'swap or not' shuffling found in
    https://link.springer.com/content/pdf/10.1007%2F978-3-642-32009-5_1.pdf
    See the 'generalized domain' algorithm on page 3.
    """
    indices = list(range(len(values)))
```

djrtwo:
should we use little endian for consistency? (sorry @vbuterin)

vbuterin:
Sure, sounds fine to me.

djrtwo:
done

hwwhww:
@vbuterin @djrtwo 
1. Updated to the latest optimization version, would you like to check the docstring again?
2. For the shuffling tests, not sure if that's necessary, what do you think about parameterizing the round number `90`? i.e.,`def shuffle(list_size: int, seed: Bytes32, rounds: int=90) -> List[int]`

djrtwo:
we lost the `get_permuted_index` to optimizations? We'll likely want to make `get_permuted_index` exposed in the spec at some point.

(1) -- docstring looks good
(2) -- rounds as param looks good

hwwhww:
@djrtwo 
Fair enough. 😄 
Reverted and refactored it.

specs/core/0_beacon-chain.md
```diff
@@ -693,57 +694,29 @@ def get_active_validator_indices(validators: List[Validator], epoch: EpochNumber
...
-                index += 1
-            else:
-                # The sample causes modulo bias. A new sample should be read.
-                pass
+    Utilizes 'swap or not' shuffling found in
+    https://link.springer.com/content/pdf/10.1007%2F978-3-642-32009-5_1.pdf
+    See the 'generalized domain' algorithm on page 3.
+    """
+    for round in range(round_count):
+        pivot = bytes_to_int(hash(seed + int_to_bytes1(round))) % list_size
```
djrtwo: We should probably take the first 8 bytes from the hash when converting from bytes to int to avoid big int requirement on clients.

```suggestion
        pivot = bytes_to_int(hash(seed + int_to_bytes1(round))[0:8]) % list_size
```

hwwhww:
**To the implementers**: https://github.com/ethereum/eth2.0-specs/pull/576/commits/70e482be28a10ca0d37de34fb3a0b5cd94108ad3 is the optimized version for shuffling the whole validator registry.

protolambda:
I'm working on a new implementation experiment-repo, written in Go. For the new "swap-or-not" shuffling algorithm. Love optimizing this stuff (and don't have a ETH 2.0 client to work on more general stuff :disappointed: ).
Although the algorithm for set-shuffling by Vitalik looked promising, it still allocated O(n) hashes every round (n = list size, talking about "hash_bytes" in original).
I optimized it to only have 1 single hash (!!!) on the stack, since you can go through indices with the same hash (there are windows) consecutively.
I probably still have some off-by-1 errors in the shuffling itself, but it's looking good so far.

Here are the preliminary benchmark results (ran locally on a laptop, more about comparison):
```
BenchmarkPermuteIndex4M-8                               100000        157231 ns/op
BenchmarkPermuteIndex40K-8                              100000        157304 ns/op
BenchmarkPermuteIndex400-8                              100000        156259 ns/op
BenchmarkPermuteIndexComparison40K-8                         2    6247616777 ns/op
BenchmarkPermuteIndexComparison400-8                       200	    64324372 ns/op
BenchmarkShuffleList4M-8                                      10    1667117128 ns/op
BenchmarkShuffleList40K-8                                   1000      16748448 ns/op
BenchmarkShuffleList400-8                                  50000        390820 ns/op


 4M = 4000000
40K =   40000
400 =     400

BenchmarkPermuteIndexN = simply check how many times you can permute the index in a list sized N
BenchmarkPermuteIndexComparisonN = do the index-wise permutations, but for every list item, i.e. N times. For comparison with BenchmarkShuffleListN
BenchmarkShuffleListN = shuffle a complete list of N items.

```

1. Note how slow index-wise permutations are if you do a complete list shuffle with it, compared to shuffling the list collectively.
Shuffling 4 *million* items collectively, is more efficient (> 3x) than shuffling a list of 1/100th the size individually.
2. Note how shuffling individual indices isn't affected much (if at all) by the list-size.


Repo here: https://github.com/protolambda/eth2-shuffle
Warning: still work-in-progress, no guarantees on correctness (likely a off-by-1 somewhere, gonna test later)

protolambda:
Posting some relevant gitter chatter here:

> @tersec
> regarding the new shuffling algorithm, https://link.springer.com/content/pdf/10.1007%2F978-3-642-32009-5_1.pdf specifically in/around figure 3 states that the choice from the domain should be uniform, but mod list_size won't be?

> @protolambda
> 
> Had my doubts first too, but it helps if you print out the pairs for any pivot + listsize, you can see a mirror pattern (e.g. pivot = 5: 05, 14, 23) on each side of the pivot. And for each pivot, there's a different pairing for any given index. (Afaik)
> 
> And pivot choice (your mod list_size concern) is indeed not perfect, but is close to uniform for large enough random numbers
> 
> 2^64/4000000=~4.6e12 list_sizes. And only 1 incomplete. So most of the time it's uniform.
> And even more if you would take a 2^256 int from the hash
> 
> Ok, looks like that's exactly what Python's int.from_bytes does. So, the fraction of time that it's not uniform is ridiculously low. It may even be worth it to consider taking only 64 bits of the hash, for speed. 99.999999...% is still good.

> @tersec
> Fair enough, reasonable tradeoff

> @protolambda
> Interesting, it's both 64 bits and 256 bits :joy: The spec has a [0:8], but this is not reflected in the set-shuffle implementation in the PR. I prefer the 64 bits, it's sufficient and fast

Now, the issue is:
Post above by @vbuterin says:
```python
pivot = int.from_bytes(hash(seed + round.to_bytes(1, 'little')), 'little') % list_size
```
Spec currently says:
```python
pivot = bytes_to_int(hash(seed + int_to_bytes1(round))[0:8]) % list_size
```
There's two inconsistency issues:
- spec is not explicit about little endianness (although it is globally)
- version by Vitalik, that every implementer team is copying for efficiency, does not have the `[0:8]`: i.e. only using 64 bits of the hash, for efficiency (without much loss in it being uniform)

Some test-vectors would be really useful.

djrtwo:
* `bytes_to_int` uses little in the definition.
* spec avoids big int arithmetic so the vitalik version should be `[0:8]`. Fixing now

djrtwo:
Here is an issue for reworking the test vectors to the swap or not -- https://github.com/ethereum/eth2.0-test-generators/issues/10

PR #584, opened 2019-02-07
Pass in `current_epoch`

decanus:
Considering we calculate the current epoch in the `per epoch processing` section, we should be able to pass it to the `update_validator_registry` function.


vbuterin:
What would the benefit of this change be?

decanus:
I don't think there is a point in recalculating things that have already been calculated.

vbuterin:
Sure, though `get_current_epoch` is hardly a calculation; it's just `slot // EPOCH_LENGTH`.

PR #586, opened 2019-02-08
Updated phase 1: fork choice rule

vbuterin:
Adds the crosslink committee to the fork choice rule. This is useful because it means that even if a proposal committee is byzantine and attempts to prevent a crosslink via a "balance attack" (alternating between chain A and chain B being the canonical chain), the crosslink committee can force the equilibrium to flip to one side or the other.


djrtwo:
Is this saying to use the shard-attestations from the persistent committee + use the beaconchain-attestations from the current crosslink committee? It is not super clear in the current description that the fork choice would be using votes from these two separate sources (and data structures).

nitpick: We should use the term "persistent committee" imo rather than "proposal committee"

vbuterin:
> Is this saying to use the shard-attestations from the persistent committee + use the beaconchain-attestations from the current crosslink committee?

Yes.

Issue #568, opened 2019-02-03
Proof of custody game design


vbuterin:
The basic proof of custody mechanism works as follows:

* For every (eg. 2 week) period `p`, each validator calculates `rk[p] = BLS_sign(key=validator_privkey, msg=p)` as their round key.
* For every crosslink they participate in during that period, if that crosslink has data with chunks `D[0] ... D[size(D)-1]`, for each `D[i]` they compute `C[i] = mix(rk[p], D[i])` where `mix` is some function where for any `rk`, `mix(rk, x)` depends on `x`. They compute `R = custody_root(D)` as Merkle root of the `C[i]` values and sign `R mod 2` along with the crosslink (this is their "custody bit")
* During period `p + 1`, the validator must reveal `prevseed=rk[p]`, and the chain can verify that it is a correct value via `BLS_verify(key=validator.pubkey, msg=p, sig=prevseed)`

There are three types of byzantine behavior that we want to penalize:

1. Revealing `rk[p]` too early
2. Signing with some random custody bit because D is unavailable
3. Signing with the incorrect custody bit because even if D is available, the validator wishes to be malicious

For (1) we can have a special-purpose slashing condition; the only requirement is that _anyone_ must be able to trigger the condition and gain from doing so, so there should be a commit-reveal game to allow participants to trigger it without the block proposers being able to frontrun them.

(2) and (3) are similar cases, except (3) is more tractable. In case (3), once `rk[p]` is revealed, any third party can recompute `custody_root(D)`, check if the bit matches, and if the bit does not match start a challenge game, which they know ahead of time they will be able to win. First, they challenge to ask for the complete custody root, which must match the bit that has already been provided. At this point, the responder is free to respond with an arbitrary value of `R` that matches the same bit; for any `D`, it is almost always possible to find some data `D'` which differs from `D` in only one block where `custody_root(D') mod 2` is either 0 or 1. At this point, the responder has (and the challenger knows that the responder has) committed to some value which is _not_ equal to `custody_root(D)`.

[...]

PR #587, opened 2019-02-08
Updated phase 1: branch and early subkey reveal challenges

vbuterin:
Includes:

* Branch challenges (ask for a branch of some signed attestation)
* Responses to branch challenges
* Code for calculating the subkey for some custody period (~9 days)
* Mechanism for revealing subkeys, both punitively (if the subkey is revealed early) and legitimately (if the subkey is revealed after it's expired), with a proof-of-knowledge-based mechanism for preventing proposer stealing of early subkey whistleblower rewards
* Cannot leave the withdraw queue until all branch challenges answered and all subkeys revealed
* After a validator leaves the withdraw queue, they still cannot withdraw for ~1 day; this period can be used as an opportunity to initiate an interactive proof of custody challenge game (not yet included)

So far all of the machinery is agnostic to what is discussed in #568; everything after this point, however, will not be.

Not included:

* The actual proof of custody game

### DIFF
--- a/specs/core/1_shard-data-chains.md
+++ b/specs/core/1_shard-data-chains.md
@@ -236,6 +236,8 @@ Define a `SubkeyReveal` as follows:
     'validator_index': 'uint64',
     'period': 'uint64',
     'subkey': 'bytes96',
+    'mask': 'bytes32',
+    'revealer_index': 'uint64'
 }
 ```
 
@@ -248,13 +250,30 @@ def get_current_custody_period(state: BeaconState) -> int:
     return get_current_epoch(state) // CUSTODY_PERIOD_LENGTH
 ```
 
-### `verify_custody_subkey`
+### `verify_custody_subkey_reveal`
 
 ```python
-def verify_custody_subkey(pubkey: bytes48, subkey: bytes96, period: int) -> bool:
-    return bls_verify(
-        pubkey=pubkey,
-        message_hash=hash(int_to_bytes8(period)),
+def verify_custody_subkey_reveal(pubkey: bytes48,
+                                 subkey: bytes96,
+                                 mask: bytes32,
+                                 mask_pubkey: bytes48,
+                                 period: int) -> bool:
+    # Legitimate reveal: checking that the provided value actually is the subkey
+    if mask == ZERO_HASH:
+        pubkeys=[pubkey]
+        message_hashes=[hash(int_to_bytes8(period))]
+        
+    # Punitive early reveal: checking that the provided value is a valid masked subkey
+    # (masking done to prevent "stealing the reward" from a whistleblower by block proposers)
+    # Secure under the aggregate extraction infeasibility assumption described on page 11-12
+    # of https://crypto.stanford.edu/~dabo/pubs/papers/aggreg.pdf
+    else:
+        pubkeys=[pubkey, mask_pubkey]
+        message_hashes=[hash(int_to_bytes8(period)), mask]
+        
+    return bls_multi_verify(
+        pubkeys=pubkeys,
+        message_hashes=message_hashes,
         signature=subkey,
         domain=get_domain(
             fork=state.fork,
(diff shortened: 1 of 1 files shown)


specs/core/1_shard-data-chains.md
```diff
@@ -209,3 +272,275 @@ The `shard_chain_commitment` is only valid if it equals `compute_commitment(head
...
+    'open_branch_challenges': [BranchChallengeRecord],
+    'next_subkey_to_reveal': 'uint64',
+    'reveal_max_periods_late': 'uint64',
+```
+
+And the initializers:
+
+```python
+    'open_branch_challenges': [],
+    'next_subkey_to_reveal': 0,
```
djrtwo: This value is going to have to be initialized to the reveal period at the fork -- `get_current_custody_reveal_period(state)`

specs/core/1_shard-data-chains.md
```diff
@@ -209,3 +272,275 @@ The `shard_chain_commitment` is only valid if it equals `compute_commitment(head
...
+## Data structures
+
+### `Validator`
+
+Add member values to the end of the `Validator` object:
+
+```python
+    'open_branch_challenges': [BranchChallengeRecord],
+    'next_subkey_to_reveal': 'uint64',
+    'reveal_max_periods_late': 'uint64',
```
djrtwo: What do we intend to use this value for?
Currently it is only set @ reveals but not used in any other conditionals or calculations

specs/core/1_shard-data-chains.md
```diff
@@ -209,3 +272,275 @@ The `shard_chain_commitment` is only valid if it equals `compute_commitment(head
...
+Verify that `len(block.body.branch_responses) <= MAX_BRANCH_RESPONSES`.
+
+For each `response` in `block.body.branch_responses`:
+
+* Find the `BranchChallengeRecord` in `state.validator_registry[response.responder_index].open_branch_challenges` whose (`root`, `data_index`) match the (`root`, `data_index`) of the `response`. Verify that one such record exists (it is not possible for there to be more than one), call it `record`.
+* Verify that `verify_merkle_branch(leaf=response.data, branch=response.branch, depth=record.depth, index=record.data_index, root=record.root)` is True.
+* Verify that `get_current_epoch(state) >= record.inclusion_epoch + ENTRY_EXIT_DELAY`.
+* Remove the `record` from `state.validator_registry[response.responder_index].open_branch_challenges`
+* Determine the proposer `proposer_index = get_beacon_proposer_index(state, state.slot)` and set `state.validator_balances[proposer_index] += base_reward(state, index) // MINOR_REWARD_QUOTIENT`.
+
```
djrtwo: Should we penalize the `challenger` a bit here? 
To reduce incentive to grief validators from being able to withdraw by keeping the `len(open_branch_challenges) > 0`

specs/core/1_shard-data-chains.md
vbuterin: The idea is that if you only reveal the subkey for eg. epoch 2 during epoch 10, then there's an 8 epoch extra delay during which PoC challenges can be made against you. It's not in the code because PoC challenges are not yet in the code.

specs/core/1_shard-data-chains.md
```diff
@@ -209,3 +272,275 @@ The `shard_chain_commitment` is only valid if it equals `compute_commitment(head
...
+And the initializers:
+
+```python
+    'open_branch_challenges': [],
+    'next_subkey_to_reveal': 0,
+    'reveal_max_periods_late': 0,
+```
+
+Rename `withdrawal_epoch` to `withdrawable_epoch`.
```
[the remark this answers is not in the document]
vbuterin: Another possibility is we get rid of the WITHDRAWABLE flag, and make the `withdrawal_epoch -> withdrawable_epoch` change in phase 0 (or, add both a withdrawal and withdrawable epoch), and move the change that penalized validators are not part of the queue to phase 0 as well. This would reduce total complexity including the complexity of the phase 0 -> phase 1 transition.

specs/core/1_shard-data-chains.md
vbuterin: I would say no. Otherwise we run into fisherman's dilemma issues. I think the goal should be that even if challengers find nothing worth challenging, they just query for random branches; this keeps this part of the system being tested and keep validators on their toes.Also there is a limit to how long each validator can be griefed because of `MAX_BRANCH_CHALLENGE_DELAY`.

specs/core/1_shard-data-chains.md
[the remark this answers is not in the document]
vbuterin: See #615

PR #589, opened 2019-02-08
Silently skip deposits with invalid proof in process_deposit

Nashatyrev:
### What's wrong:
`process_deposit` has the 
```python
assert validate_proof_of_possession(...)
```
statement which aborts processing of the block containing a deposit with invalid proof resulting this block being invalid. 
Thus a proposer needs to skip deposits with invalid proofs when composing a block. The same way malicious/incorrect proposer may skip a number of valid deposits. 

### Suggestion:
Include *all* deposits to blocks (this is verified by constantly incremented deposit index) but silently ignore deposits with invalid proofs on block transition: i.e. not change the validator registry/balances

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1252,36 +1252,37 @@ def process_deposit(state: BeaconState,
     Note that this function mutates ``state``.
     """
     # Validate the given `proof_of_possession`
-    assert validate_proof_of_possession(
+    valid_proof = validate_proof_of_possession(
         state,
         pubkey,
         proof_of_possession,
         withdrawal_credentials,
     )
 
-    validator_pubkeys = [v.pubkey for v in state.validator_registry]
-
-    if pubkey not in validator_pubkeys:
-        # Add new validator
-        validator = Validator(
-            pubkey=pubkey,
-            withdrawal_credentials=withdrawal_credentials,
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1252,36 +1252,37 @@ def process_deposit(state: BeaconState,
     Note that this function mutates ``state``.
     """
     # Validate the given `proof_of_possession`
-    assert validate_proof_of_possession(
+    valid_proof = validate_proof_of_possession(
```
```suggestion
    proof_is_valid = validate_proof_of_possession(
```

To avoid confusing the reader into thinking the `valid_proof` variable contains a valid proof as opposed to just being a bool.

specs/core/0_beacon-chain.md
```diff
@@ -1252,36 +1252,37 @@ def process_deposit(state: BeaconState,
...
-        # Note: In phase 2 registry indices that have been withdrawn for a long time will be recycled.
-        state.validator_registry.append(validator)
-        state.validator_balances.append(amount)
-    else:
-        # Increase balance by deposit amount
-        index = validator_pubkeys.index(pubkey)
-        assert state.validator_registry[index].withdrawal_credentials == withdrawal_credentials
-
-        state.validator_balances[index] += amount
+    if valid_proof:
```
```suggestion
    if proof_is_valid:
```

PR #591, opened 2019-02-08
Weekly release of dev into master

djrtwo:
changelog will be available in tagged release


specs/core/0_beacon-chain.md
```diff
@@ -639,6 +650,19 @@ def slot_to_epoch(slot: SlotNumber) -> EpochNumber:
     return slot // EPOCH_LENGTH
 ```
 
+### `get_previous_epoch`
+
+```python
+def get_previous_epoch(state: BeaconState) -> EpochNumber:
+    """`
+    Return the previous epoch of the given ``state``.
+    If the current epoch is  ``GENESIS_EPOCH``, return ``GENESIS_EPOCH``.
```
prestonvanloon: This logic doesn't match the implementation.

specs/validator/0_beacon-chain-validator.md
```diff
@@ -344,24 +344,48 @@ Either (2) or (3) occurs if (1) fails. The choice between (2) and (3) is determi
...
+                assignment += (is_proposer,)
+
+                potential_assignments.append(assignment)
                 break
 
     return potential_assignments
 ```
 
-`get_next_epoch_crosslink_committees` should be called at the beginning of each epoch to plan for the next epoch. A validator should always plan for both values of `registry_change` as a possibility unless the validator can concretely eliminate one of the options. Planning for a future shuffling involves noting at which slot one might have to attest and propose and also which shard one should begin syncing (in phase 1+).
+`get_next_epoch_committee_assignments` should be called at the beginning of each epoch to plan for the next epoch. A validator should always plan for both values of `registry_change` as a possibility unless the validator can concretely eliminate one of the options. Planning for a future shuffling involves noting at which slot one might have to attest and propose and also which shard one should begin syncing (in phase 1+).
```
prestonvanloon: Can you clarify this bit `beginning of each epoch to plan for the next epoch`?

Does this mean that I call get_next_epoch_committee_assignments at the first slot in epoch N and it returns the assignments for epoch N+1?

specs/core/0_beacon-chain.md
terenc3t: This is diff than our implementation, I wrote the code before putting this helper function in spec. Let's update our implementation, will open an issue

specs/validator/0_beacon-chain-validator.md
prestonvanloon: Just read over the details in line 347, is it is clear there. But would helpful for clarity in this language here

specs/validator/0_beacon-chain-validator.md
```diff
@@ -344,24 +344,48 @@ Either (2) or (3) occurs if (1) fails. The choice between (2) and (3) is determi
...
 
 ```python
-def get_next_epoch_crosslink_committees(state: BeaconState,
-                                        validator_index: ValidatorIndex) -> List[Tuple[ValidatorIndex], ShardNumber]:
+def get_next_epoch_committee_assignments(
+        state: BeaconState,
+        validator_index: ValidatorIndex) -> List[Tuple[List[ValidatorIndex], ShardNumber, SlotNumber, bool]]:
+    """
+    Return a list of the two possible committee assignments for ``validator_index`` at the next epoch.
+    Possible committee ``assignment`` is of the form (List[ValidatorIndex], ShardNumber, SlotNumber, bool).
```
prestonvanloon: _Possible_? Can we use more assertive language and a structured object? 

It would be easier for understanding to have something like this (JSON as example):
```JSON
{
 "assignments": [{
    "validator_indices": [0, 1, 2],
    "shard": 5,
    "slot": 100,
    "is_proposer": true
  }]
}
```
This is much more clear than
```json
[[0,1,2], 5, 100, true]]
```

specs/validator/0_beacon-chain-validator.md
```diff
@@ -344,24 +344,48 @@ Either (2) or (3) occurs if (1) fails. The choice between (2) and (3) is determi
...
+        * ``assignment[0]`` is the list of validators in the committee
+        * ``assignment[1]`` is the shard to which the committee is assigned
+        * ``assignment[2]`` is the slot at which the committee is assigned
+        * ``assignment[3]`` is a bool signalling if the validator is expected to propose
```
prestonvanloon: All of the validators in the committee are proposers for the same shard at the same slot?

specs/validator/0_beacon-chain-validator.md
```diff
@@ -344,24 +344,48 @@ Either (2) or (3) occurs if (1) fails. The choice between (2) and (3) is determi
...
+                assignment = selected_committees[0]
+                assignment += (slot,)
+                first_committee_at_slot = crosslink_committees[0][0]  # List[ValidatorIndex]
+                is_proposer = first_committee_at_slot[slot % len(first_committee_at_slot)] == validator_index
+                assignment += (is_proposer,)
+
+                potential_assignments.append(assignment)
                 break
 
     return potential_assignments
```
prestonvanloon: Why is it "potential"? Aren't these the assignments?

specs/validator/0_beacon-chain-validator.md
djrtwo: The function is for a particular `validator_index`

specs/core/0_beacon-chain.md
prestonvanloon: What i am saying is this comment says 
```
 If the current epoch is  ``GENESIS_EPOCH``, return ``GENESIS_EPOCH``.
```
but the pseudo code below says
```
 if the current epoch X is less than GENESIS_EPOCH, return X
```

specs/validator/0_beacon-chain-validator.md
prestonvanloon: Why return a list of validators in the committee? If you see my JSON example above, you can see how this is a confusing response from this method.

specs/validator/0_beacon-chain-validator.md
djrtwo: The above section explains that at the start of the current epoch, there are two potential assignments for a validator for the next -- if a validator registry change occurs or not. This function gives you both.

You are a in a super-position of both futures until either the epoch change or you locally figure out if `registry_change` happens (which you can check by eagerly processing the next epoch transition). In the normal case, you can collapse this super position about the next epoch within `MIN_ATTESTATION_INCLUSION_DELAY` slots into the current epoch.

This section is not new. This release just has some slightly modified functionality and bug fixes. We can address these comments and clarifications in a separate issue/pr

specs/validator/0_beacon-chain-validator.md
djrtwo: You need to know the entire list to perform certain actions such as aggregating. Arguably it could be left out here, but it will need to be gathered eventually.

specs/core/0_beacon-chain.md
djrtwo: The code actually implemented the same functionality as the comment but the semantics were way different. Updated it for clarity.

PR #594, opened 2019-02-09
Add mandatory deposit index ordering

vbuterin:
Co-requisite with #589

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -526,7 +526,7 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
     # Ethereum 1.0 chain data
     'latest_eth1_data': Eth1Data,
     'eth1_data_votes': [Eth1DataVote],
-    'deposit_count': 'uint64'
+    'deposit_index': 'uint64'
 }
 ```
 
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -526,6 +526,7 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
     # Ethereum 1.0 chain data
     'latest_eth1_data': Eth1Data,
     'eth1_data_votes': [Eth1DataVote],
+    'deposit_count': 'uint64'
```
djrtwo: This is named `deposit_index` in all other locations.

specs/core/0_beacon-chain.md
```diff
@@ -1478,6 +1479,7 @@ def get_initial_beacon_state(initial_validator_deposits: List[Deposit],
         # Ethereum 1.0 chain data
         latest_eth1_data=latest_eth1_data,
         eth1_data_votes=[],
+        deposit_index=0
```
djrtwo: This should default to `len(initial_validator_deposits)`

vbuterin: Done. Though note that all this introduces one issue: if there are bad deposits before launch, then there will be a bunch of deposits that have to be reprocessed before any new deposits can be processed. I guess it's fine because they'll all fail due to the pubkey uniqueness check?

djrtwo: I don't think so. This will ensure that we don't have to attempt to process any failed deposits -- thus starting from the next deposit after the initial.

I'm assuming that `initial_validator_deposits` includes invalid ones  that will just get skipped over in `process_deposit` below

djrtwo: or am I missing something

vbuterin: Aah I see, if initial_validator_deposits includes invalid ones then there's no issue.

PR #595, opened 2019-02-10
Merkleise SSZ container elements

JustinDrake:
Reasons to use `merkle_hash` instead of  `hash` for containers:

1) **Smaller witnesses**: `BeaconState` is a somewhat wide container (26 fields as of now, likely 30+ in phase 2). With concatenation and plain concatenation the size of the Merkle witnesses for the top level are ~32 bytes per field element.
2) **Faster incremental hashing**
3) **Consistency**: Consistent with `merkle_hash` for lists/vectors.

### DIFF
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -434,10 +434,10 @@ Where the inner `hash_tree_root_internal` is a recursive application of the tree
 
 #### Container
 
-Recursively tree hash the values in the container in the same order as the fields, and return the hash of the concatenation of the results.
+Recursively tree hash the values in the container in the same order as the fields, and Merkle hash the results.
 
 ```python
-return hash(b''.join([hash_tree_root_internal(getattr(x, field)) for field in value.fields]))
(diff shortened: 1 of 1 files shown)


vbuterin:
The main practical downside I see is that it would increase the cost of re-hashing a validator record. That said, if we use the existing `merkle_hash` function with a 128 byte chunk size, then I suppose the slowdown is not large, and we still get the largest part of the gain in terms of reducing the portion of the Merkle proof that passes through the top-level state object....

djrtwo:
We don't frequently re-hash individual validator records because we pulled balances out to a separate array so I think this is fine.

JustinDrake:
> it would increase the cost of re-hashing a validator record

Would it not *decrease* the cost of re-hashing? A validator container has 7 fields. For the sake of argument, let's say we update one field and that SSZ chunks are 32 bytes. Without Merkleisation we would have 7*32 = 224 bytes of hashing overhead. With Merkleisation we would have log2(ceil(7))*32 = 96 bytes of overhead.

PR #596, opened 2019-02-10
Do not duplicate `bytes` length when tree-hashing

JustinDrake:
When tree-hashing a `bytes` the length currently appears twice:

1) in the first four bytes of the serialisation
2) in the final `hash(chunkz[0] + datalen)`

This is suboptimal:

1) It is unnecessary and causes an awkward 4-byte misalignment.
2) It is inconsistent with the way lists are handled. (`bytes` is just a list of `uint8`.)
3) Removing the 4 leading bytes is an optimisation opportunity when the Merkle root of the `bytes` object is already known.

### DIFF
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -9,7 +9,7 @@ deserializing objects and data types.
 ## ToC
 
 * [About](#about)
-* [Terminology](#terminology)
+* [Variables and Functions](#variables-and-functions)
 * [Constants](#constants)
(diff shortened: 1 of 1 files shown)


vbuterin:
I'd support making `bytes` a de facto synonym for `List[bytes1]` globally in SSZ; it _does_ make it simple.

JustinDrake:
Yay, 47 lines removed!

PR #597, opened 2019-02-11
Reward clarifications/fix

djrtwo:
* only reward attestations for slots from previous epoch.
* clarify that rewards/penalties are for the active validator set from the previous epoch

Note: this does not affect justification/finalization as those were already bound to the `epoch_boundary_attestations` which split them by epoch

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1779,7 +1779,7 @@ The steps below happen when `(state.slot + 1) % EPOCH_LENGTH == 0`.
 [Validators](#dfn-Validator) attesting during the current epoch:
 
 * Let `current_total_balance = get_total_balance(state, get_active_validator_indices(state.validator_registry, current_epoch))`.
-* Let `current_epoch_attestations = [a for a in state.latest_attestations if current_epoch == slot_to_epoch(a.data.slot)]`. (Note: this is the set of attestations of slots in the epoch `current_epoch`, _not_ attestations that got included in the chain during the epoch `current_epoch`.)
+* Let `current_epoch_attestations = [a for a in state.latest_attestations if current_epoch == slot_to_epoch(a.data.slot)]`. (Note: Each of these attestations votes for the current justified epoch/block root because of the [attestation block validity rules](#attestations-1).)
 * Validators justifying the epoch boundary block at the start of the current epoch:
   * Let `current_epoch_boundary_attestations = [a for a in current_epoch_attestations if a.data.epoch_boundary_root == get_block_root(state, get_epoch_start_slot(current_epoch))]`.
   * Let `current_epoch_boundary_attester_indices` be the union of the [validator](#dfn-validator) index sets given by `[get_attestation_participants(state, a.data, a.aggregation_bitfield) for a in current_epoch_boundary_attestations]`.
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1781,22 +1781,19 @@ The steps below happen when `(state.slot + 1) % EPOCH_LENGTH == 0`.
 * Let `current_total_balance = get_total_balance(state, get_active_validator_indices(state.validator_registry, current_epoch))`.
 * Let `current_epoch_attestations = [a for a in state.latest_attestations if current_epoch == slot_to_epoch(a.data.slot)]`. (Note: this is the set of attestations of slots in the epoch `current_epoch`, _not_ attestations that got included in the chain during the epoch `current_epoch`.)
 * Validators justifying the epoch boundary block at the start of the current epoch:
-  * Let `current_epoch_boundary_attestations = [a for a in current_epoch_attestations if a.data.epoch_boundary_root == get_block_root(state, get_epoch_start_slot(current_epoch)) and a.data.justified_epoch == state.justified_epoch]`.
+  * Let `current_epoch_boundary_attestations = [a for a in current_epoch_attestations if a.data.epoch_boundary_root == get_block_root(state, get_epoch_start_slot(current_epoch))]`.
```
vbuterin: Given that the required justified epoch is contained in the block root, isn't it the case that all honest attestations with the same epoch boundary root will have the same justified epoch?

Also, however, attestations whose justified epoch is "wrong" even if their epoch boundary root is correct should **not** count toward the 2/3 for making a new justified block. The reason is that the safety proofs depend on the existence of a chain of "supermajority links" with 2/3 of validators sharing the same source *and* target.

specs/core/0_beacon-chain.md
```diff
@@ -1781,22 +1781,19 @@ The steps below happen when `(state.slot + 1) % EPOCH_LENGTH == 0`.
...
-* Validators that made an attestation during the previous epoch:
+* Validators that made an attestation during the previous epoch, targeting the previous justified slot:
   * Let `previous_epoch_attestations = [a for a in state.latest_attestations if previous_epoch == slot_to_epoch(a.data.slot)]`.
   * Let `previous_epoch_attester_indices` be the union of the validator index sets given by `[get_attestation_participants(state, a.data, a.aggregation_bitfield) for a in previous_epoch_attestations]`.
-* Validators targeting the previous justified slot:
-  * Let `previous_epoch_justified_attestations = [a for a in current_epoch_attestations + previous_epoch_attestations if a.data.justified_epoch == state.previous_justified_epoch]`.
-  * Let `previous_epoch_justified_attester_indices` be the union of the validator index sets given by `[get_attestation_participants(state, a.data, a.aggregation_bitfield) for a in previous_epoch_justified_attestations]`.
-  * Let `previous_epoch_justified_attesting_balance = get_total_balance(state, previous_epoch_justified_attester_indices)`.
+  * Let `previous_epoch_attesting_balance = get_total_balance(state, previous_epoch_attester_indices)`.
```
vbuterin: Don't understand why we're removong the matching justified epoch requirement here...

djrtwo: Any attestation that is in `latest_attestations` is guaranteed to have the expected justified block/epoch because of the inclusion rules in the block.attestations section.
Keeping it in is strictly redundant logic

specs/core/0_beacon-chain.md
djrtwo: > Given that the required justified epoch is contained in the block root, isn't it the case that all honest attestations with the same epoch boundary root will have the same justified epoch?

Yes.

> Also, however, attestations whose justified epoch is "wrong" even if their epoch boundary root is correct should **not** count toward the 2/3 for making a new justified block.

These cannot even be included on chain due to the block processing rules

vbuterin: Right, got it. Though this seems like something that should be documented somewhere explicitly or else we might easily accidentally remove it...

vbuterin:
@djrtwo any reason for the phase 1 label on this one? It feels phase 0 specific.

djrtwo:
> any reason for phase 1 label?

Mistake! adding an explicit note about the attestation inclusion now

JustinDrake:
@djrtwo: I still find most of the "Per-epoch processing" section daunting and hard to read. I intend to sit down to review the section carefully to weed out further bugs, and possibly refactor things for readability. I've "approved" the PR to keep the ball rolling but haven't really wrapped my head around all the subtle behaviour.

PR #598, opened 2019-02-11
Add sanity checks to is_surround_vote()

tarasbob:


JustinDrake:
Why? One of the design goals is to keep the spec as simple as possible.

Note that this does change the consensus rules (some beacon blocks that used to be valid are no longer valid) so it is not merely a sanity check.

tarasbob:
Maybe this is not the right place to check this, but it doesn't make sense to me to allow attestations where source_epoch > target_epoch

JustinDrake:
> it doesn't make sense to me to allow attestations where source_epoch > target_epoch

There's a bunch of unexpected-yet-harmless validator behaviour we allow. (For example, it wouldn't make sense for a validator to use slots less than `GENESIS_SLOT` in various places.) Figuring out all unexpected behaviour is not part of the remit of the spec: it would add complexity, overhead and potentially DoS vectors.

> Maybe this is not the right place to check this

This check may make sense client-side, but that would be an implementation detail :)

PR #601, opened 2019-02-11
Add support for transfers between withdrawn accounts

vbuterin:


specs/core/0_beacon-chain.md
```diff
@@ -440,6 +444,25 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
...
 ```
 
+##### `Transfer`
+
+```python
+{
+    # Sending from
+    'sender': 'uint64',
+    # Sending to
+    'to': 'uint64',
```
hwwhww: Question to all eth1.0 and eth2.0 devs: do we want to keep using `to` in eth2.0...?

specs/core/0_beacon-chain.md
```diff
@@ -1766,6 +1790,27 @@ For each `exit` in `block.body.exits`:
...
+
+* Verify that `state.validator_balances[transfer.from] >= transfer.amount`.
+* Verify that `state.validator_balances[transfer.from] >= transfer.fee`.
+* Verify that `state.validator_balances[transfer.from] == transfer.amount + transfer.fee` or `state.validator_balances[transfer.from] >= transfer.amount + transfer.fee + MIN_DEPOSIT_AMOUNT`.
+* Verify that `transfer.slot == state.slot`.
+* Verify that `transfer.from` does not match the `from` field of another transfer in `block.body.transfers`.
+* Verify that `get_current_epoch(state) <= state.validator_registry[transfer.from].withdrawal_epoch`.
+* Verify that `get_current_epoch(state) <= state.validator_registry[transfer.to].withdrawal_epoch`.
+* Let `transfer_message = hash_tree_root(Transfer(from=transfer.from, to=transfer.to, value=transfer.value, fee=transfer.fee, slot=transfer.slot, signature=EMPTY_SIGNATURE))`.
+* Verify that `bls_verify(pubkey=state.validator_registry[transfer.from].pubkey, message_hash=transfer_message, signature=transfer.signature, domain=get_domain(state.fork, slot_to_epoch(transfer.slot), DOMAIN_TRANSFER))`.
```
djrtwo: Need to make sure signature is from the `withdrawal_key` -- to do so we need to actually include the key in the message, right?

specs/core/0_beacon-chain.md
```diff
@@ -1766,6 +1790,27 @@ For each `exit` in `block.body.exits`:
...
+Verify that `len(block.body.transfers) <= MAX_TRANSFERS`.
+
+For each `transfer` in `block.body.transfers`:
+
+* Verify that `state.validator_balances[transfer.from] >= transfer.amount`.
+* Verify that `state.validator_balances[transfer.from] >= transfer.fee`.
+* Verify that `state.validator_balances[transfer.from] == transfer.amount + transfer.fee` or `state.validator_balances[transfer.from] >= transfer.amount + transfer.fee + MIN_DEPOSIT_AMOUNT`.
+* Verify that `transfer.slot == state.slot`.
+* Verify that `transfer.from` does not match the `from` field of another transfer in `block.body.transfers`.
+* Verify that `get_current_epoch(state) <= state.validator_registry[transfer.from].withdrawal_epoch`.
```
djrtwo: In phase 0, these two conditions are always true.

We need to add a condition to ensure that the validator is exited.

```suggestion
* Verify that `get_current_epoch(state) >= state.validator_registry[transfer.to].exit_epoch`.
* Verify that `get_current_epoch(state) <= state.validator_registry[transfer.from].withdrawal_epoch`.
```

specs/core/0_beacon-chain.md
JustinDrake: Yes.

(In theory we could use a zero-knowledge scheme that doesn't reveal `withdrawal_key`. This has the benefit of opening the door for quantum-secure signatures.)

specs/core/0_beacon-chain.md
JustinDrake: `exit_epoch` doesn't work because we want to allow some time for validators to be slashed. My suggestion would be `exit_epoch + 2**13` which adds a security margin of 34 days.

djrtwo: Ah, you're right. Need a new constant for this then. `MIN_EXIT_EPOCHS_BEFORE_TRANSFER`

JustinDrake: Note that the `MIN_EXIT_EPOCHS_BEFORE_TRANSFER` restriction is mainly for the sender (as opposed to the recipient).

Question: why restrict the recipient at all?

vbuterin: Now that I think about it, you're right, the recipient doesn't need to be restricted...

I think what we're looking for is that the WITHDRAWABLE flag of the sender needs to be set.

specs/core/0_beacon-chain.md
JustinDrake: I changed `sender` to `from` to be consistent with `to`. (We could also use `sender`/`recipient` but that's a bit more verbose.)

specs/core/0_beacon-chain.md
JustinDrake: I've removed the constraint on the recipient. I've also added  `MIN_EXIT_EPOCHS_BEFORE_TRANSFER` as per above.

JustinDrake: > I think what we're looking for is that the WITHDRAWABLE flag of the sender needs to be set.

I'm tempted to remove the logic related to `WITHDRAWABLE` until phase 2. This includes removing [`process_withdrawals`](https://github.com/ethereum/eth2.0-specs/blob/d4f79c8878375a9e146f3491e2b4d3043a35a30a/specs/core/0_beacon-chain.md), `MIN_VALIDATOR_WITHDRAWAL_EPOCHS` and the `WITHDRAWABLE` flag.

vbuterin: So we _do_ need a flag to say "we're done passing through the exit queue, we're now in the stage where we're waiting for custody challenges". And we _do_ also need a concept of "withdrawableness" because transferring ETH from one account to another _is_ a kind of withdrawal. 

My inclination is to take the withdrawable logic from the phase 1 PR and import it into phase 0. That is, when a validator exits the queue, their `withdrawable_epoch` gets set to `now + `MIN_VALIDATOR_WITHDRAWAL_EPOCHS`. This gets us handling for two state changes for the price of one.

JustinDrake: > My inclination is to take the withdrawable logic from the phase 1 PR and import it into phase 0.

Sounds good 👍 

Nomenclature side note: Talking about the "exit queue" is much better than the language we had before. Did some renaming in #534:

* `process_withdrawals` => `process_exit_queue`
* `MAX_WITHDRAWALS_PER_EPOCH` => `MAX_EXIT_DEQUEUES_PER_EPOCH`
* `withdrawn_so_far` => `dequeues`

specs/core/0_beacon-chain.md
```diff
@@ -1766,6 +1793,27 @@ For each `exit` in `block.body.exits`:
...
+
+For each `transfer` in `block.body.transfers`:
+
+* Verify that `state.validator_balances[transfer.from] >= transfer.amount`.
+* Verify that `state.validator_balances[transfer.from] >= transfer.fee`.
+* Verify that `state.validator_balances[transfer.from] == transfer.amount + transfer.fee` or `state.validator_balances[transfer.from] >= transfer.amount + transfer.fee + MIN_DEPOSIT_AMOUNT`.
+* Verify that `transfer.slot == state.slot`.
+* Verify that `transfer.from` does not match the `from` field of another transfer in `block.body.transfers`.
+* Verify that `get_current_epoch(state) >= state.validator_registry[transfer.from].exit_epoch + MIN_EXIT_EPOCHS_BEFORE_TRANSFER`.
+* Verify that `hash(transfer.pubkey) == BLS_WITHDRAWAL_PREFIX_BYTE + state.validator_registry[transfer.from].withdrawal_credentials[1:]`
```
djrtwo: This isn't quite right -- see here https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#withdrawal-credentials

Should be
```suggestion
* Verify that `state.validator_registry[transfer.from].withdrawal_credentials == BLS_WITHDRAWAL_PREFIX_BYTE + hash(transfer.pubkey)[1:]`
```

JustinDrake: Fixed!

PR #603, opened 2019-02-11
(Do not merge yet) Reduce SSZ_CHUNK_SIZE to 32

JustinDrake:
The intent of setting SSZ_CHUNK_SIZE = 128 is to reduce the cost of hashing. It is possible (to be confirmed once we have more concrete numbers from implementers) that the gains are limited and not worth the complexity of using a non-standard chunk size.

If the gains indeed are limited, this PR aligns with the Eth2 goal "to minimize complexity, even at the cost of some losses in efficiency".


JustinDrake:
screenshot 2019-02-11 at 14 57 31

JustinDrake:
Closing in favour of #675.

PR #607, opened 2019-02-12
Minor modification to reduce lines of code

paulhauner:
## What

Replaces the counter with the index of the loop.

## Why

It is functionally the same but uses two less lines of code.

## Extra Info

My Rust linter picked this one up, I thought I might share.

Take it or leave it, no objections either way :)

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -2030,10 +2030,8 @@ def process_penalties_and_exits(state: BeaconState) -> None:
     eligible_indices = filter(eligible, all_indices)
     # Sort in order of exit epoch, and validators that exit within the same epoch exit in order of validator index
     sorted_indices = sorted(eligible_indices, key=lambda index: state.validator_registry[index].exit_epoch)
-    withdrawn_so_far = 0
-    for index in sorted_indices:
+    for withdrawn_so_far, index in enumerate(sorted_indices):
         prepare_validator_for_withdrawal(state, index)
-        withdrawn_so_far += 1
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -2030,10 +2030,8 @@ def process_penalties_and_exits(state: BeaconState) -> None:
     eligible_indices = filter(eligible, all_indices)
     # Sort in order of exit epoch, and validators that exit within the same epoch exit in order of validator index
     sorted_indices = sorted(eligible_indices, key=lambda index: state.validator_registry[index].exit_epoch)
-    withdrawn_so_far = 0
-    for index in sorted_indices:
+    for withdrawn_so_far, index in enumerate(sorted_indices):
         prepare_validator_for_withdrawal(state, index)
-        withdrawn_so_far += 1
         if withdrawn_so_far >= MAX_WITHDRAWALS_PER_EPOCH:
```
hwwhww: Since `withdrawn_so_far` starts from `0`, so it should be:
```python
        if withdrawn_so_far + 1 >= MAX_WITHDRAWALS_PER_EPOCH:
```
We could also move `prepare_validator_for_withdrawal` after the `if withdrawn_so_far >= MAX_WITHDRAWALS_PER_EPOCH: break`. However, the name `withdrawn_so_far` is not accurate in this context ...
What do you think about it?

JustinDrake: Solved by putting `prepare_validator_for_withdrawal(state, index)` lower down 👍

JustinDrake:
> Take it or leave it, no objections either way :)

More please 👍

PR #608, opened 2019-02-12
Provide definition for undefined `index`

paulhauner:
## What

Define the `index` variable in [per-epoch processing/crosslinks-1](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#crosslinks-1).

## Why

It took me some time to figure out what the `index` was.

## Extra Info

I'm not sure if this is the best way to word it. I wanted to go in an extra indent but that was forbidden.

If you don't want to merge, please let me know that my assumption is correct :)


JustinDrake:
LGTM 👍

PR #615, opened 2019-02-13
Bring forward changes to withdrawability from phase 1

vbuterin:
* The `WITHDRAWABLE` flag is removed; instead, a validator's withdrawability is determined through the `withdrawable_epoch` field (renamed and re-purposed from `withdrawal_epoch` which was not used)
* When a validator passes through the withdrawal queue, the `prepare_validator_for_withdrawal` function does not let them withdraw immediately; instead, they have to wait `MIN_VALIDATOR_WITHDRAWAL_EPOCHS`. This extra minimum delay serves no value in phase 0, but is crucial for phase 1 as the period between a validator passing through the queue and the validator being eligible to withdraw is where proof of custody challenges can come in; adding it in phase 0 is only half a line of code so easier to add it now.
* If a validator is penalized, they are no longer subject to the exit queue; instead, their `withdrawable_epoch` is set `LATEST_PENALIZED_EXIT_LENGTH` into the future and this is used to determine when the validator can withdraw
* Changes the eligibility condition for a transfer to use the `withdrawable_epoch`

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1369,6 +1369,7 @@ def penalize_validator(state: BeaconState, index: ValidatorIndex) -> None:
     Penalize the validator of the given ``index``.
     Note that this function mutates ``state``.
     """
+    assert validator.withdrawable_epoch == FAR_FUTURE_EPOCH  # [TO BE REMOVED IN PHASE 2]
     exit_validator(state, index)
     validator = state.validator_registry[index]
     state.latest_penalized_balances[get_current_epoch(state) % LATEST_PENALIZED_EXIT_LENGTH] += get_effective_balance(state, index)
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -233,7 +233,6 @@ Code snippets appearing in `this style` are to be interpreted as Python code.
 | `ENTRY_EXIT_DELAY` | `2**2` (= 4) | epochs | 25.6 minutes |
 | `ETH1_DATA_VOTING_PERIOD` | `2**4` (= 16) | epochs | ~1.7 hours |
 | `MIN_VALIDATOR_WITHDRAWAL_EPOCHS` | `2**8` (= 256) | epochs | ~27 hours |
-| `MIN_EXIT_EPOCHS_BEFORE_TRANSFER` | `2**13` (= 8,192) | epochs | ~36 days |
 
 ### State list lengths
 
@@ -261,7 +260,6 @@ Code snippets appearing in `this style` are to be interpreted as Python code.
 | Name | Value |
 | - | - |
 | `INITIATED_EXIT` | `2**0` (= 1) |
-| `WITHDRAWABLE` | `2**1` (= 2) |
 
 ### Max operations per block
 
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -2065,11 +2068,12 @@ def process_penalties_and_exits(state: BeaconState) -> None:
...
-        if validator.penalized_epoch <= current_epoch:
-            penalized_withdrawal_epochs = LATEST_PENALIZED_EXIT_LENGTH // 2
-            return current_epoch >= validator.penalized_epoch + penalized_withdrawal_epochs
+        # Cannot exit if you already have
+        if validator.withdrawable_epoch < FAR_FUTURE_EPOCH:
+            return False
+        # Can exit if at least the minimum amount of time has passed
         else:
-            return current_epoch >= validator.exit_epoch + MIN_VALIDATOR_WITHDRAWAL_EPOCHS
+            return current_epoch >= validator.exit_epoch + MIN_VALIDATOR_WITHDRAWABILITY_DELAY
```
JustinDrake: As the code is written it seems that if `current_epoch >= validator.exit_epoch + MIN_VALIDATOR_WITHDRAWABILITY_DELAY` then necessarily `validator.withdrawable_epoch < FAR_FUTURE_EPOCH` so that `eligible` always returns False.

djrtwo: The first condition is about `withdrawable_epoch` whereas the second is about `exit_epoch`.
@JustinDrake I think you were conflating these two variables, no?

specs/core/0_beacon-chain.md
```diff
@@ -1377,18 +1378,20 @@ def penalize_validator(state: BeaconState, index: ValidatorIndex) -> None:
     state.validator_balances[whistleblower_index] += whistleblower_reward
     state.validator_balances[index] -= whistleblower_reward
     validator.penalized_epoch = get_current_epoch(state)
+    validator.withdrawable_epoch = get_current_epoch(state) + LATEST_PENALIZED_EXIT_LENGTH
```
djrtwo: Do we need a condition that validators can only be penalized if they have not already entered the `withdrawable` state? I'm imagining someone exits, enters withdrawable, transfers some amount of funds, and then is found to be slashable. Although it would be nice to penalize them a bit, I think we should not allow this and other unexpected flows because it opens up unexpected state transitions (hard to test, potential attack vectors, etc)

specs/core/0_beacon-chain.md
JustinDrake: I was making another silly mistake 🤦‍♂️

specs/core/0_beacon-chain.md
JustinDrake: > Do we need a condition that validators can only be penalized if they have not already entered the `withdrawable` state?

Automatic withdrawals in phase 2 effectively provide that. I'd support adding extra conditions (to be removed in phase 2) for consistency with phase 2.

JustinDrake: Added `assert validator.withdrawable_epoch == FAR_FUTURE_EPOCH  # [TO BE REMOVED IN PHASE 2]` in `penalize_validator`.

vbuterin: It needs to be possible to penalize validators that are in the "waiting for PoC challenges" stage. So I'd recommend the simple `assert validator.withdrawable_epoch >= state.slot` or `assert validator.withdrawable_epoch > state.slot + ENTRY_EXIT_DELAY`.

PR #622, opened 2019-02-14
Add cautionary assert for divide-by-zero

paulhauner:
## What

Adds a cautionary assert for divide-by-zero on `get_permutated_index`

## Why

safety first

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -742,7 +742,7 @@ def get_permuted_index(index: int, list_size: int, seed: Bytes32) -> int:
     https://link.springer.com/content/pdf/10.1007%2F978-3-642-32009-5_1.pdf
     See the 'generalized domain' algorithm on page 3.
     """
-    assert(index < list_size)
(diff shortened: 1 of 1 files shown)


JustinDrake:
There is already is an implicit `assert` with the modulo

> pivot = bytes_to_int(hash(seed + int_to_bytes1(round))[0:8]) % list_size

This produces

> ZeroDivisionError: integer division or modulo by zero

paulhauner:
My objective was to make it explicit.

Maybe I've been spending too much time in a language focused on safety.

paulhauner:
Now I think about it, the spec is full of implicit asserts.

Happy to close this one.

vbuterin:
Technically, we really should have an assert that `index < list_size`, which would implicitly require `list_size > 0`.

paulhauner:
> Technically, we really should have an assert that `index < list_size`, which would implicitly require `list_size > 0`.

I was just about to raise this actually! It still returns a result when `index >= list_size`.

Issue #618, opened 2019-02-13
off-by-one validating justified_epoch for attestations


arnetheduck:
3d5aa35

Currently, spec has this rule for validating attestations:

> Verify that `attestation.data.justified_epoch` is equal to `state.justified_epoch if attestation.data.slot >= get_epoch_start_slot(get_current_epoch(state)) else state.previous_justified_epoch`.

Epoch transitions happen at `(state.slot + 1) % EPOCH_LENGTH == 0`. This means that when `attestation.data.slot % EPOCH_LENGTH == 0`, the epoch state update will not yet have happened but `get_epoch_start_slot` will already return a slot from the new epoch, and this check will fail due to an off-by-one.. it can be fixed by an extra + 1:

> Verify that `attestation.data.justified_epoch` is equal to `state.justified_epoch if attestation.data.slot + 1 >= get_epoch_start_slot(get_current_epoch(state)) else state.previous_justified_epoch`.

It should be noted however that this feels like a workaround - the deeper issue here is that `get_current_epoch` returns the new epoch one slot earlier than the `justified_epoch` update happens, so we have a situation where according to one function, we are in a new epoch, but `state` has not yet been updated to include the changes that pertain to that epoch.

arnetheduck:
related commit: f943361a23883b387cd52c00bea24d7605fccc5b

djrtwo:
It would be much cleaner to make this epoch based and arguably should have been done in the epoch rework

```
Verify that attestation.data.justified_epoch is equal to state.justified_epoch if slot_to_epoch(attestation.data.slot) >= get_current_epoch(state) else state.previous_justified_epoch.
```

If the attestation is included in the epoch during the slot it was for, then it uses `justified_epoch`. If not, it is from the prev, and uses `previous_justified_epoch`

djrtwo:
This is clean because we don't actually transition into the next upon until the end of the state transition. so when processing the block, the current epoch check is valid

djrtwo:
> This means that when `attestation.data.slot % EPOCH_LENGTH == 0`, the epoch state update will not yet have happened but `get_epoch_start_slot` will already return a slot from the new epoch, and this check will fail due to an off-by-one.. it can be fixed by an extra + 1:

I'm not sure I follow here

* An attestation with`attestation.data.slot % EPOCH_LENGTH == 0` can at earliest be included in a block at the next slot (assuming `MIN_ATTESTATION_INCLUSION_DELAY == 1`
[...]

PR #627, opened 2019-02-14
fix off by one attestaton issue

djrtwo:
addresses #618 

Note that this is the quick fix. There is a discussion to be had on when exactly the epoch transition should occur. An alternative would be to be at the _start_ of the 0th slot of an epoch rather than at the _end_ of the last slot of an epoch. I'd like to get this fix merged now for the spec release because this is a critical bug and can save any debate for over this next week.


djrtwo:
cc: @arnetheduck

JustinDrake:
> I'd like to get this fix merged now

Done! :)

arnetheduck:
what a beauty! :)

Issue #540, opened 2019-01-31
Refactor signature handling


JustinDrake:
Right now the signature logic across SSZ object is ad-hoc. The following are dealt differently:

* Beacon blocks: custom signed header
* Attestations: external signature
* Exits: overwrite `signature` field with 0x0
* RANDAO reveals: sign raw data

Here's one way to unify the signature logic (discussed with Danny and Vitalik):

* Remove signatures from the existing SSZ objects
* For every SSZ object `o` have a corresponding "signature header" `h` and canonical signature `h.signature`.
* The signature header `h` is defined below, where `h.root = tree_hash_root(o)` and `h.signature = bls_sign(h)` of where `h.signature` is first set to `EMPTY_SIGNATURE `.

```
{
    epoch: Epoch,
    shard: Shard,
    root: Bytes32,
    signature: Signature,
}
```

vbuterin:
Adding an empty signature field to _every_ SSZ object seems like a bad idea; it would just add a lot of hashing overhead. The cleanest would just be for every type `T` have a type `SignedT = {data: T, signature: Signature}`.

However there is another issue which is that for attestations, we need to have the attestation and custody bitfields not be signed over, because the signers don't know who each other are, so the signature mechanism is not going to fit it well no matter how it is done.

Another possibility is to establish a rule that the signature is the signature of all fields that come before it. This way we can just use order of appearance in the object to denote order of signing and what does and does not need to be signed.

JustinDrake:
Closing in favour of #649.

PR #633, opened 2019-02-15
Refactor signature handling

JustinDrake:
Goes with #625.

Related issue: #540

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -471,7 +471,7 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
 ```python
 {
     # Header
-    'slot': Slot,
+    'slot': 'uint64',
     'parent_root': 'bytes32',
     'state_root': 'bytes32',
     'randao_reveal': 'bytes96',
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -44,7 +44,6 @@
             - [`BeaconBlock`](#beaconblock)
             - [`BeaconBlockBody`](#beaconblockbody)
             - [`Proposal`](#proposal)
-            - [`RandaoReveal`](#randaoreveal)
         - [Beacon chain state](#beacon-chain-state)
             - [`BeaconState`](#beaconstate)
             - [`Validator`](#validator)
@@ -509,17 +508,6 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
     'block_root': 'bytes32',
     # Signature
     'signature': 'bytes96',
-}
-```
-
-#### `RandaoReveal`
-
-```python
-{
-    # Epoch number
-    'epoch': 'uint64',
-    # RANDAO reveal
-    'randao_reveal': 'bytes96',
 }
 ```
 
@@ -1641,8 +1629,7 @@ Below are the processing steps that happen at every `block`.
 
 #### RANDAO
 
-* Let `randao_reveal = RandaoReveal(epoch=get_current_epoch(state), randao_reveal=block.randao_reveal)`.
-* Verify that `bls_verify(pubkey=proposer.pubkey, message_hash=signed_root(randao_reveal, "randao_reveal"), signature=randao_reveal.randao_reveal, domain=get_domain(state.fork, get_current_epoch(state), DOMAIN_RANDAO))`.
+* Verify that `bls_verify(pubkey=proposer.pubkey, message_hash=hash_tree_root(get_current_epoch(state)), signature=block.randao_reveal, domain=get_domain(state.fork, get_current_epoch(state), DOMAIN_RANDAO))`.
 * Set `state.latest_randao_mixes[get_current_epoch(state) % LATEST_RANDAO_MIXES_LENGTH] = xor(get_randao_mix(state, get_current_epoch(state)), hash(block.randao_reveal))`.
 
 #### Eth1 data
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -500,16 +497,29 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
...
     'slot': 'uint64',
     # Shard number (`BEACON_CHAIN_SHARD_NUMBER` for beacon chain)
     'shard': 'uint64',
-    # Block's hash of root
+    # Block root
     'block_root': 'bytes32',
+    # Signature
+    'signature': 'bytes96',
+}
+```
```
vbuterin: Why do we even need to do this? Why not be more radical and just remove the Proposal class entirely and directly sign the block? Or is that to reduce the number of slashing conditions down the line when we have shard blocks?

specs/core/0_beacon-chain.md
```diff
@@ -1652,16 +1633,16 @@ Below are the processing steps that happen at every `block`.
...
+* Let `proposer = state.validator_registry[get_beacon_proposer_index(state, state.slot)]`.
+* Let `proposal = Proposal(block.slot, BEACON_CHAIN_SHARD_NUMBER, signed_root(block, "signature"), block.signature)`.
+* Verify that `bls_verify(pubkey=proposer.pubkey, message_hash=signed_root(proposal, "signature"), signature=proposal.signature, domain=get_domain(state.fork, get_current_epoch(state), DOMAIN_PROPOSAL))`.
 
 #### RANDAO
 
-* Let `proposer = state.validator_registry[get_beacon_proposer_index(state, state.slot)]`.
-* Verify that `bls_verify(pubkey=proposer.pubkey, message_hash=int_to_bytes32(get_current_epoch(state)), signature=block.randao_reveal, domain=get_domain(state.fork, get_current_epoch(state), DOMAIN_RANDAO))`.
+* Let `randao_reveal = RandaoReveal(epoch=get_current_epoch(state), randao_reveal=block.randao_reveal)`.
+* Verify that `bls_verify(pubkey=proposer.pubkey, message_hash=signed_root(randao_reveal, "randao_reveal"), signature=randao_reveal.randao_reveal, domain=get_domain(state.fork, get_current_epoch(state), DOMAIN_RANDAO))`.
```
vbuterin: This part of the reform feels a bit needlessly complexifying. Not everything that gets signed needs to be a `signed_root` of something. Attestations still sign regular roots of blocks for example. So I'd say signing `int_to_bytes32(get_current_epoch(state))` (or maybe better for security: `hash(int_to_bytes8(get_current_epoch(state)))`, so that way we don't trip up security assumptions that expect the `message_hash` to be an actual hash) is still fine.

vbuterin:
Yay! Love PRs that delete more LoCs than they add as usual. Made two comments, otherwise can't find any issues yet.

specs/core/0_beacon-chain.md
JustinDrake: We need the `Proposal` class for attester slashings. Maybe there's a way to clean this up.

specs/core/0_beacon-chain.md
JustinDrake: Would you be happy with `hash_tree_root(get_current_epoch(state))`? I don't like doing serialisation outside of the ssz functions.

JustinDrake: Now 40 lines removed with this simplification.

specs/core/0_beacon-chain.md
JustinDrake: I do like the idea of having two separate slashing conditions for two reasons:

1) It's somewhat cleaner (don't need `Proposal` and `BEACON_CHAIN_SHARD_NUMBER`)
2) Beacon chain proposer slashing is arguably more important than shard proposer slashing. So with two separate transaction types we have guaranteed throughput for beacon chain proposer slashings.

JustinDrake: It's not completely trivial if we remove `Proposal`. The reason is that for the proposer slashing you would need a Merkle path to the block root to prove the slot number.

vbuterin: Wait, why? Why not just have a `ProposerSlashing` object that contains two block headers? Sure it's maybe a hundred bytes longer, but that's nothing compared to attester slashings in any case...

JustinDrake: We could replace `Proposal` with a `BeaconBlockHeader` object but would that be more verbose (`Proposal` only have 4 fields, `BeaconBlockHeader` would have 7) and slightly less efficient?

vbuterin: Aah I see, `BeaconBlock` objects themselves are too big, and we _technically_ don't have a `BeaconBlockHeader` (well, from any block you can create a block header with an equivalent SSZ root by replacing the body with a hash of the body, but if we're going to take advantage of this fact in the spec it would be good to document it well...

I see a few possibilities:

1. Modify the `BeaconBlock` object so it has an explicit header that contains `hash(body)` and a separate body
2. _Define_ a `BeaconBlockHeader` as a header that can be derived from any `BeaconBlock`
3. Keep the current approach

Note that in terms of hashes and bytes, (1) and (2) are exactly equivalent; object1 containing the SSZ root of object2 with object2 living "somewhere else", and obect1 literally containing object2 in its entirety, are in some sense exactly the same thing and will have exactly the same SSZ root. So it's a choice of presentation rather than content.

(3) would change things less relative to the status quo and it would not introduce this weirdness of the header being kinda part of and kinda separate from the rest of the block, though it would lead to a slightly longer spec. I guess my "don't change things at this stage of the game unless it's a really really good idea" bias does make me lean slightly toward (3)....

JustinDrake: Ok, I'll keep as is and mull over it a bit.

JustinDrake: Can you approve the PR?

djrtwo:
I'd like to take a quick look at this before merge. Give me today

PR #634, opened 2019-02-15
Add explicit assert to get_permutated_index

paulhauner:
## What

Add an explicit assertion that the list size is within the bounds of the shuffler.

## Why

There is an implicit assert in `int_to_bytes4` but I don't think it' immediately obvious.

To find it you would need to be thinking about how Python handles `x.to_bytes(len, ..)` when `x >= 2**(len * 8)`.

Also, it illustrates the bounds of the algorithm.

## Additional Info

You could `assert SHUFFLE_ROUND_COUNT < 2**8` for `int_to_bytes1` but it's a constant so I omitted that.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -745,7 +745,7 @@ def get_permuted_index(index: int, list_size: int, seed: Bytes32) -> int:
     See the 'generalized domain' algorithm on page 3.
     """
     assert index < list_size
-    assert list_size < 2**24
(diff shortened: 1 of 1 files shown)


JustinDrake:
Can we go up to `list_size <= 2**40` (4 bytes + 1 byte from the `// 258`).

paulhauner:
Oh yeah, sorry my bad.

I did `2**32 / 256 == 2**24` instead of `2**32 * 256 == 2**40`.

I'll change it now

PR #638, opened 2019-02-15
Remove hashing bottleneck from swap or not

JustinDrake:
Optimise hashing in swap or not:

1) `hash(seed + int_to_bytes32(round))` can now be computed once and cached for all values of `round` (90 of them).
2) `hash(int_to_bytes32(position // 256))` can be precomputed and cached for all relevant values of `position`. Even with 4,000,000 validators that's only 0.5MB (!!!) of cached data.

I also use `int_to_bytes32` instead of `int_to_bytes1` and `int_to_bytes4` to be more generic.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -747,10 +747,10 @@ def get_permuted_index(index: int, list_size: int, seed: Bytes32) -> int:
     assert index < list_size
     
     for round in range(SHUFFLE_ROUND_COUNT):
-        pivot = bytes_to_int(hash(seed + int_to_bytes2(round))[0:8]) % list_size
+        pivot = bytes_to_int(hash(seed + int_to_bytes32(round))[0:8]) % list_size
         flip = (pivot - index) % list_size
         position = max(index, flip)
-        source = xor(hash(seed + int_to_bytes2(round)), hash(int_to_bytes32(position // 256)))
+        source = xor(hash(seed + int_to_bytes32(round)), hash(int_to_bytes32(position // 256)))
         byte = source[(position % 256) // 8]
         bit = (byte >> (position % 8)) % 2
         index = flip if bit else index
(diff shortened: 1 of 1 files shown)


vbuterin:
Are we sure this is safe? This seems scary to me; in particular, `xor(bit for position 1, bit for position 2)` would now be always the same at least within a given round (I think), and this seems like it could introduce a lot of predictability and would definitely break the provable security.

vbuterin:
And if we can't find a crazy optimization of this type that's fine, then keeping int_to_bytes1 and int_to_bytes4 is prudent; it keeps the data to be hashed well under 64 bytes, making it likely to fit into one block (of whatever the hash function is).

vbuterin:
I don't think shuffling is that bad in the grand scheme of things; the overhead is only ~11 bytes of hashing per validator per epoch, which is about the same as recalculating validator balances.

vbuterin:
One _implementation-level optimization_ that can be done to cut hashing down by 50% is to only compute the hash bits for the portions of the values that correspond to feasible outputs of `max(x, pivot - x)` (ie. `pivot/2.....pivot` and `n/2+pivot/2.....n`); I don't think anyone has done that yet.

protolambda:
> One _implementation-level optimization_ that can be done to cut hashing down by 50% is to only compute the hash bits for the portions of the values that correspond to feasible outputs of `max(x, pivot - x)` (ie. `pivot/2.....pivot` and `n/2+pivot/2.....n`); I don't think anyone has done that yet.

@vbuterin Thought of that, done that, just a few days ago :smile:

Also implemented it in such a way that you don't have to do the same hash twice, but also don't have to allocate a cache.

Implementation + more explanation here:
https://github.com/protolambda/eth2-shuffle

Probably still an off-by-1 bug or two lurking somewhere, going to check with test-vectors later.

PR #641, opened 2019-02-16
Fix out-of-bounds in `get_shuffling`

paulhauner:
## What

Change `get_shuffling(...)` so it gives shuffles using an _index_ of `active_validator_indices` instead of a _value_ of `active_validator_indices`.

## Why

The following fails with a `IndexError: list index out of range`:

```python
validator_indices = [
    0,    # inactive
    1,    # active
    2,    # inactive
    3,    # active
]

# ... therefore ...

active_validator_indices = [
    1,    # active
    3,    # active
]

def get_permuted_index(i, length):
    return i

shuffled_active_validator_indices = [
        active_validator_indices[get_permuted_index(i, len(active_validator_indices))]
        for i in active_validator_indices
]
```

Whereas this does not:

```python
shuffled_active_validator_indices = [
        active_validator_indices[get_permuted_index(i, len(active_validator_indices))]
        for i in range(len(active_validator_indices))
]
```

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -795,7 +795,7 @@ def get_shuffling(seed: Bytes32,
                   validators: List[Validator],
                   epoch: Epoch) -> List[List[ValidatorIndex]]
     """
-    Shuffle (seeded by ``seed`` and ``epoch``) active validators and split into crosslink committees.
+    Shuffle active validators and split into crosslink committees.
     Return a list of committees (each a list of validator indices).
     """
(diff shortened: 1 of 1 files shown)


JustinDrake:
@djrtwo Are we no longer seeding with `epoch`?

djrtwo:
We moved `epoch` into the mix in `generate_seed` rather than on the fly in `shuffle`. This was clearer and still accomplishes the goal of ensuring that the shuffling changes at points of reshuffling even in the case that no new randao reveals occur.

Issue #626, opened 2019-02-14
The integer debate


protolambda:
Follow up issue after earlier gitter chat and implementers call today (https://github.com/ethereum/eth2.0-pm/issues/29)

To pick the best solution, we need a more structural approach, and list the arguments for each "problem-class".

Note that I highlight Java and javascript as special considerations, due to the different support for integers than most other languages provide.

**Go, Rust, Swift, Nim all support signed/unsigned 32/64 bit numbers.**

## Problem classes

1) Slot numbers
2) Validator indices
3) Balances

## Slot numbers

Range here: 1 slot per 6 seconds, for a few thousand years (could upgrade earlier...) = approx. `1000*365*24*60*60/6=5,256,000,000`.

Signed numbers: No. However, there is a case for the first few epochs, where logic is looking back at old history, beyond the genesis. This could potentially result in an underflow.

Ways to catch the underflow:
- `x < 0`: signed number check
- `x < genesis_slot`, with `genesis_slot >> 0` (i.e. sufficiently large offset from 0.) One potential offset is `1 << 63`, **_effectively simulating a signed number_** (but worse: sign bit inverted, if you think of it as two's complement), but not human readable in decimal base. If this becomes the "blocknumber" of the future, we want it to be readable, right?

If we don't care about very long-term consistency, we can go for 32 bits. It seems unnecessary however, as there's more range available to every language/platform. E.g. we could even opt for an imaginary 48 bit (un)signed integer.

### Javascript

Highlight from earlier gitter chat:

> Well, if we're not using the higher bits (i.e. [53, 63]) of the signed 64 bit number for the next ~ 800K years, javascript can just keep using 52 bit mantissa based numbers (the 11 unused bits being the exponent part of the float). And then we can choose for a signed 64 bit number. With genesis at a clean 0.

I.e. by exploiting the range of slot numbers, we could choose for a 64 bit number (signed or unsigned), and have it be compatible with the 52 bit (excl. sign bit) range of javascript float-based (i.e. mantissa) integers.

Alternative would be to use big-numbers, like the alternative that Java also has, see below.

### Java

[...]

PR #642, opened 2019-02-16
Change if statements in get_crosslink_committes

### LINKED ISSUE #639: get_crosslink_committees reads previous_seed during genesis_epoch
### Summary

During `GENESIS_EPOCH`, `get_crosslink_committees_at_slot(..)` reads `state.previous_shuffling_epoch` when `epoch == current_epoch`.

### Detail

`get_crosslink_committees_at_slot(...)` first checks for `epoch == previous_epoch`, then checks for `epoch == current_epoch`.

During the genesis epoch, `current_epoch == previous_epoch` and the function so uses `state.previous_shuffling_epoch == ZERO_HASH` for shuffling.

During the next epoch (`GENESIS_EPOCH + 1`), when you wish to get the shuffling for `GENESIS_EPOCH`, you also trigger the `current_epoch == previous_epoch` path. However, `state.previous_shuffling_epoch != ZERO_HASH` anymore -- it is now the value that was set during `get_genesis_beacon_state(...)` (a.k.a. `current_shuffling_seed` during `GENESIS_EPOCH`).

## Proposed Solutions

1. (Preferred) Swap the `if` statements for so we first check for `epoch == current_epoch`, then `epoch == previous_epoch`.
2. (Seems wrong) During `get_genesis_beacon_state(..)`  ensure `state.previous_shuffling_seed` is also set to `generate_seed(state, GENESIS_EPOCH)`.

paulhauner:
Fixes #639.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -879,16 +879,16 @@ def get_crosslink_committees_at_slot(state: BeaconState,
 
     assert previous_epoch <= epoch <= next_epoch
 
-    if epoch == previous_epoch:
-        committees_per_epoch = get_previous_epoch_committee_count(state)
-        seed = state.previous_shuffling_seed
-        shuffling_epoch = state.previous_shuffling_epoch
-        shuffling_start_shard = state.previous_shuffling_start_shard
-    elif epoch == current_epoch:
+    if epoch == current_epoch:
         committees_per_epoch = get_current_epoch_committee_count(state)
         seed = state.current_shuffling_seed
         shuffling_epoch = state.current_shuffling_epoch
         shuffling_start_shard = state.current_shuffling_start_shard
+    elif epoch == previous_epoch:
+        committees_per_epoch = get_previous_epoch_committee_count(state)
+        seed = state.previous_shuffling_seed
+        shuffling_epoch = state.previous_shuffling_epoch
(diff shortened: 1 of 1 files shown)


JustinDrake:
Why doesn't `get_previous_epoch(state)` always return `get_current_epoch(state) - 1`?

djrtwo:
^ remnant from when we started epoch at 0.
We should not resolve #626 before we modify `get_previous_epoch`. Seems like people want to see epochs/slots to start at 0

JustinDrake:
I see three ways to prevent underflows:

1) Add a bunch of subtle and crufty `if` special cases across the codebase. We already tried this and In hindsight it feels like the wrong way forward.
2) As suggested in #626, use `int64` (as opposed to `uint64`) for `slot` and `epoch` and start with `GENESIS_SLOT = 0`. Definitely better than 1) but maybe not quite as clean as 2).
3) Have `GENESIS_SLOT` be large enough. It was my mistake to not realise that `2**63` would cause complications with Java and it's unfortunate there's no obvious choice. My latest proposal is `2**n * SECONDS_PER_DAY/SECONDS_PER_SLOT` where `n` is the smallest sufficiently large exponent.

djrtwo:
@paulhauner #655 addresses this immediate issue. I'm combing the spec to make sure that the initial previous epoch being before genesis_epoch doesn't open up any unexpected new bugs

PR #647, opened 2019-02-17
Explicitly `parent_root` check in state transition

JustinDrake:


hwwhww:
1. But isn't it has been checked in the "processing block conditions"?
    https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#beacon-chain-processing
    - The parent block with root `block.parent_root` has been processed and accepted.
2. For nodes (not only validators), it also implied that the they'd switch the local view of `BeaconState` to the view of block `parent_root` (`parent_block.state_root`) before applying state transition.

JustinDrake:
I see the sections "Beacon chain processing" and "Beacon chain state transition function" as serving two different functions.

The first section involves subjectivity (time, observed blocks, observed offchain attestations, ...) whereas the second is a purely objective state transition function. The `parent_root` check is an objective check part of the state transition function. It doesn't hurt to mention `parent_root` in the first section to help the presentation.

JustinDrake:
Closing in favour of #649.

PR #649, opened 2019-02-18
Parent root reconstruction + granular state roots + body/header segregation

JustinDrake:
This PR introduces three significant changes:

1. **Parent root reconstruction**: Instead of relying on a parent root "oracle" in the state transition function, the state transition function now only has two argument: (current) `state` and (latest) `block`.
2. **Granular state roots**: To be more friendly to clients we expose granular state roots at every slot independently of block production. Having the state roots at every epoch boundary is especially helpful.
3. **Body/header segregation**: This follows the discussion [here](https://github.com/ethereum/eth2.0-specs/pull/633#discussion_r257070067), and allows us to remove `Proposal` and related things.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1626,7 +1626,7 @@ Below are the processing steps that happen at every `block`.
 
 * Let `block_root = tree_hash_root(block.body)`.
 * Set `state.latest_block_roots[state.slot % LATEST_BLOCK_ROOTS_LENGTH] = block_root`.
-* Verify that `block.parent_root == get_block_root(state, state.slot - 1)`.
+* Verify that `block.header.parent_root == get_block_root(state, state.slot - 1)`.
 
 #### Block signature
 
@@ -1635,21 +1635,21 @@ Below are the processing steps that happen at every `block`.
 
 #### RANDAO
 
-* Verify that `bls_verify(pubkey=proposer.pubkey, message_hash=hash_tree_root(get_current_epoch(state)), signature=block.randao_reveal, domain=get_domain(state.fork, get_current_epoch(state), DOMAIN_RANDAO))`.
-* Set `state.latest_randao_mixes[get_current_epoch(state) % LATEST_RANDAO_MIXES_LENGTH] = xor(get_randao_mix(state, get_current_epoch(state)), hash(block.randao_reveal))`.
+* Verify that `bls_verify(pubkey=proposer.pubkey, message_hash=hash_tree_root(get_current_epoch(state)), signature=block.body.randao_reveal, domain=get_domain(state.fork, get_current_epoch(state), DOMAIN_RANDAO))`.
+* Set `state.latest_randao_mixes[get_current_epoch(state) % LATEST_RANDAO_MIXES_LENGTH] = xor(get_randao_mix(state, get_current_epoch(state)), hash(block.body.randao_reveal))`.
 
 #### Eth1 data
 
-* If there exists an `eth1_data_vote` in `state.eth1_data_votes` for which `eth1_data_vote.eth1_data == block.eth1_data` (there will be at most one), set `eth1_data_vote.vote_count += 1`.
-* Otherwise, append to `state.eth1_data_votes` a new `Eth1DataVote(eth1_data=block.eth1_data, vote_count=1)`.
+* If there exists an `eth1_data_vote` in `state.eth1_data_votes` for which `eth1_data_vote.eth1_data == block.body.eth1_data` (there will be at most one), set `eth1_data_vote.vote_count += 1`.
+* Otherwise, append to `state.eth1_data_votes` a new `Eth1DataVote(eth1_data=block.body.eth1_data, vote_count=1)`.
 
 #### Transactions
 
 ##### Proposer slashings
 
-Verify that `len(block.proposer_slashings) <= MAX_PROPOSER_SLASHINGS`.
+Verify that `len(block.body.proposer_slashings) <= MAX_PROPOSER_SLASHINGS`.
 
-For each `proposer_slashing` in `block.proposer_slashings`:
+For each `proposer_slashing` in `block.body.proposer_slashings`:
 
 * Let `proposer = state.validator_registry[proposer_slashing.proposer_index]`.
 * Verify that `proposer_slashing.block_header_1.slot == proposer_slashing.block_header_2.slot`.
@@ -1661,9 +1661,9 @@ For each `proposer_slashing` in `block.proposer_slashings`:
 
 ##### Attester slashings
 
-Verify that `len(block.attester_slashings) <= MAX_ATTESTER_SLASHINGS`.
+Verify that `len(block.body.attester_slashings) <= MAX_ATTESTER_SLASHINGS`.
 
-For each `attester_slashing` in `block.attester_slashings`:
+For each `attester_slashing` in `block.body.attester_slashings`:
 
 * Let `slashable_attestation_1 = attester_slashing.slashable_attestation_1`.
 * Let `slashable_attestation_2 = attester_slashing.slashable_attestation_2`.
@@ -1677,9 +1677,9 @@ For each `attester_slashing` in `block.attester_slashings`:
 
 ##### Attestations
 
-Verify that `len(block.attestations) <= MAX_ATTESTATIONS`.
+Verify that `len(block.body.attestations) <= MAX_ATTESTATIONS`.
 
-For each `attestation` in `block.attestations`:
+For each `attestation` in `block.body.attestations`:
 
 * Verify that `attestation.data.slot <= state.slot - MIN_ATTESTATION_INCLUSION_DELAY < attestation.data.slot + SLOTS_PER_EPOCH`.
 * Verify that `attestation.data.justified_epoch` is equal to `state.justified_epoch if slot_to_epoch(attestation.data.slot + 1) >= get_current_epoch(state) else state.previous_justified_epoch`.
@@ -1722,12 +1722,12 @@ For each `attestation` in `block.attestations`:
 
 ##### Deposits
 
-Verify that `len(block.deposits) <= MAX_DEPOSITS`.
+Verify that `len(block.body.deposits) <= MAX_DEPOSITS`.
 
 [TODO: add logic to ensure that deposits from 1.0 chain are processed in order]
 [TODO: update the call to `verify_merkle_branch` below if it needs to change after we process deposits in order]
 
-For each `deposit` in `block.deposits`:
+For each `deposit` in `block.body.deposits`:
 
 * Let `serialized_deposit_data` be the serialized form of `deposit.deposit_data`. It should be 8 bytes for `deposit_data.amount` followed by 8 bytes for `deposit_data.timestamp` and then the `DepositInput` bytes. That is, it should match `deposit_data` in the [Ethereum 1.0 deposit contract](#ethereum-10-deposit-contract) of which the hash was placed into the Merkle tree.
 * Verify that `deposit.index == state.deposit_index`.
@@ -1757,9 +1757,9 @@ process_deposit(state, deposit)
 
 ##### Voluntary exits
 
-Verify that `len(block.voluntary_exits) <= MAX_VOLUNTARY_EXITS`.
+Verify that `len(block.body.voluntary_exits) <= MAX_VOLUNTARY_EXITS`.
 
-For each `exit` in `block.voluntary_exits`:
+For each `exit` in `block.body.voluntary_exits`:
 
 * Let `validator = state.validator_registry[exit.validator_index]`.
 * Verify that `validator.exit_epoch > get_entry_exit_effect_epoch(get_current_epoch(state))`.
@@ -1771,9 +1771,9 @@ For each `exit` in `block.voluntary_exits`:
 
 Note: Transfers are a temporary functionality for phases 0 and 1, to be removed in phase 2.
 
-Verify that `len(block.transfers) <= MAX_TRANSFERS` and that all transfers are distinct.
+Verify that `len(block.body.transfers) <= MAX_TRANSFERS` and that all transfers are distinct.
 
-For each `transfer` in `block.transfers`:
+For each `transfer` in `block.body.transfers`:
 
 * Verify that `state.validator_balances[transfer.from] >= transfer.amount`.
 * Verify that `state.validator_balances[transfer.from] >= transfer.fee`.
@@ -2070,7 +2070,7 @@ def process_exit_queue(state: BeaconState) -> None:
 
 ### State root verification
 
-Verify `block.state_root == hash_tree_root(state)` if there exists a `block` for the slot being processed.
+Verify `block.header.state_root == hash_tree_root(state)` if there exists a `block` for the slot being processed.
 
 # References
 
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -298,9 +298,9 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
     # Proposer index
     'proposer_index': 'uint64',
     # First block metadata
-    'block_metadata_1': BeaconBlockMetadata,
+    'metadata_1': BeaconBlockMetadata,
     # Second block metadata
-    'block_metadata_2': BeaconBlockMetadata,
+    'metadata_2': BeaconBlockMetadata,
 }
 ```
 
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -112,8 +112,6 @@
         - [Beacon chain fork choice rule](#beacon-chain-fork-choice-rule)
     - [Beacon chain state transition function](#beacon-chain-state-transition-function)
         - [Per-slot processing](#per-slot-processing)
-            - [Slot](#slot)
-            - [Block roots](#block-roots)
         - [Per-block processing](#per-block-processing)
             - [Block header](#block-header)
             - [RANDAO](#randao)
@@ -230,13 +228,13 @@ Code snippets appearing in `this style` are to be interpreted as Python code.
 | `MIN_SEED_LOOKAHEAD` | `2**0` (= 1) | epochs | 6.4 minutes |
 | `ACTIVATION_EXIT_DELAY` | `2**2` (= 4) | epochs | 25.6 minutes |
 | `EPOCHS_PER_ETH1_VOTING_PERIOD` | `2**4` (= 16) | epochs | ~1.7 hours |
+| `SLOTS_PER_BATCHING` | `2**13` (= 8,192) | slots | ~13 hours |
 | `MIN_VALIDATOR_WITHDRAWAL_DELAY` | `2**8` (= 256) | epochs | ~27 hours |
 
 ### State list lengths
 
 | Name | Value | Unit | Duration |
 | - | - | :-: | :-: |
-| `LATEST_BLOCK_ROOTS_LENGTH` | `2**13` (= 8,192) | slots | ~13 hours |
 | `LATEST_RANDAO_MIXES_LENGTH` | `2**13` (= 8,192) | epochs | ~36 days |
 | `LATEST_ACTIVE_INDEX_ROOTS_LENGTH` | `2**13` (= 8,192) | epochs | ~36 days |
 | `LATEST_SLASHED_EXIT_LENGTH` | `2**13` (= 8,192) | epochs | ~36 days |
@@ -533,10 +531,11 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
     # Recent state
     'latest_crosslinks': [Crosslink],
     'latest_block_roots': ['bytes32'],
+    'latest_state_roots': ['bytes32'],
     'latest_active_index_roots': ['bytes32'],
     'latest_slashed_balances': ['uint64'],  # Balances slashed at every withdrawal period
     'latest_attestations': [PendingAttestation],
-    'batched_block_roots': ['bytes32'],
+    'historical_batchings': ['bytes32'],
 
     # Ethereum 1.0 chain data
     'latest_eth1_data': Eth1Data,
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -502,10 +495,12 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
 {
```
djrtwo: Should probably rename this `BeaconBlockHeader`

specs/core/0_beacon-chain.md
```diff
@@ -1412,34 +1407,48 @@ For convenience, we provide the interface to the contract here:
...
-    parent_root=ZERO_HASH,
-    state_root=GENESIS_STATE_ROOT,
-    randao_reveal=EMPTY_SIGNATURE,
-    eth1_data=Eth1Data(
-        deposit_root=ZERO_HASH,
-        block_hash=ZERO_HASH
-    ),
-    signature=EMPTY_SIGNATURE,
-    body=BeaconBlockBody(
+def get_genesis_beacon_block(state: BeaconState) -> BeaconBlock
```
djrtwo: What `state` is passed in here? `get_genesis_beacon_state` requires the `genesis_block_root` so if this takes in the `genesis_beacon_state` then there's a circular dependency.

specs/core/0_beacon-chain.md
```diff
@@ -502,10 +495,12 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
 {
     # Slot number
     'slot': 'uint64',
-    # Shard number (`BEACON_CHAIN_SHARD_NUMBER` for beacon chain)
-    'shard': 'uint64',
```
hwwhww: Question 🙋‍♀️: does that mean we can remove `shard` from `BeaconBlock` either with or without this PR?

JustinDrake: Yes. Although it's natural to separate the beacon and shard proposer slashing conditions with this PR because we now have a notion of header. See discussion [here](https://github.com/ethereum/eth2.0-specs/pull/633#discussion_r257070067).

specs/core/0_beacon-chain.md
JustinDrake: Rewrote things so that it's crystal clear there is no circular dependency :) In short, `block_body` does *not* depend on `state`, and only `block_body` is referenced in `state`.

specs/core/0_beacon-chain.md
JustinDrake: Yes, done! 👍(As mentioned in the PR description I didn't rename to `BeaconBlockHeader` to keep the diff small.)

specs/core/0_beacon-chain.md
```diff
@@ -1412,34 +1407,48 @@ For convenience, we provide the interface to the contract here:
...
         exits=[],
-    ),
+        transfers=[],
+    )
+
+    return BeaconBlock(
+        header=Proposal(
+            slot=GENESIS_SLOT,
+            parent_root=ZERO_HASH,
+            block_root=tree_hash_root(body),
```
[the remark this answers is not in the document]
JustinDrake: Good catch 👍

specs/core/0_beacon-chain.md
```diff
@@ -1604,33 +1618,31 @@ Below are the processing steps that happen at every slot.
...
-
-* Verify that `block.slot == state.slot`.
-
-#### Block signature
+#### Block header
 
+* Verify that `block.header.slot == state.slot`.
+* Verify that `block.header.parent_root == get_block_root(state, state.slot - 1)`.
+* Verify that `block.header.block_root == hash_tree_root(block.body)`.
+* Set `state.latest_block_roots[state.slot % LATEST_BLOCK_ROOTS_LENGTH] = block.header.block_root`.
```
vbuterin: One problem I see with this approach is that if only block bodies are put into the block roots accumulator and state roots are not, then we lose the implicit accumulator of state roots. I would say the ability to access historical state roots through the accumulator is extremely important; along with many other arguments, it's where receipts will ultimately live.

JustinDrake: Good catch! Implemented `historical_batchings` as discussed on the call today. cc @djrtwo

specs/core/0_beacon-chain.md
```diff
@@ -1412,35 +1398,43 @@ For convenience, we provide the interface to the contract here:
...
-    parent_root=ZERO_HASH,
-    state_root=GENESIS_STATE_ROOT,
-    randao_reveal=EMPTY_SIGNATURE,
-    eth1_data=Eth1Data(
-        deposit_root=ZERO_HASH,
-        block_hash=ZERO_HASH
-    ),
-    signature=EMPTY_SIGNATURE,
-    body=BeaconBlockBody(
+def get_genesis_beacon_block_body() -> BeaconBlock
```
```suggestion
def get_genesis_beacon_block_body() -> BeaconBlockBody:
```

specs/core/0_beacon-chain.md
```diff
@@ -1476,18 +1470,21 @@ def get_genesis_beacon_state(genesis_validator_deposits: List[Deposit],
...
+        historical_batchings=[],
 
         # Ethereum 1.0 chain data
-        latest_eth1_data=latest_eth1_data,
+        latest_eth1_data=genesis_eth1_data,
         eth1_data_votes=[],
         deposit_index=len(genesis_validator_deposits)
     )
 
+    state.latest_block_roots[GENESIS_SLOT % SLOTS_PER_BATCHING] = hash_tree_root(genesis_block_body)
```
djrtwo: I would consider filling this array with `hash_tree_root(genesis_block_body)` (similar to the active_index_roots below). The implication here is that the only thing before the genesis_block is the genesis_block. This is more failsafe for any initial state operations that might subtract to a previous epoch/slot.

djrtwo: We should initialize `latest_state_roots` in the same way (currently not initialized at all)

specs/core/0_beacon-chain.md
```diff
@@ -1505,6 +1502,24 @@ def get_genesis_beacon_state(genesis_validator_deposits: List[Deposit],
     return state
 ```
 
+```python
+def get_genesis_beacon_block(genesis_state: BeaconState,
+                             genesis_block_body: BeaconBlockBody) -> BeaconBlock
```
```suggestion
                             genesis_block_body: BeaconBlockBody) -> BeaconBlock:
```

specs/core/0_beacon-chain.md
```diff
@@ -1596,41 +1611,35 @@ _Note_: If there are skipped slots between a block and its parent block, run the
 
 ### Per-slot processing
 
-Below are the processing steps that happen at every slot.
-
-#### Slot
+Below are the processing steps that happen at every `slot > GENESIS_SLOT`.
 
+* Set `state.latest_state_roots[state.slot % SLOTS_PER_BATCHING] = hash(state)`.
```
```suggestion
* Set `state.latest_state_roots[state.slot % SLOTS_PER_BATCHING] = hash_tree_root(state)`.
```

specs/core/0_beacon-chain.md
```diff
@@ -1596,41 +1611,35 @@ _Note_: If there are skipped slots between a block and its parent block, run the
...
+* Set `state.latest_block_roots[state.slot % SLOTS_PER_BATCHING] = get_block_root(state, state.slot)`.
```
djrtwo: The call to `get_block_root` will fail here due to the assertion -- `assert slot < state.slot <= slot + SLOTS_PER_BATCHING`. This is an unfortunate consequence of the validity conditions of some helpers are dependent upon _when_ in the state transition function they are called.

specs/core/0_beacon-chain.md
```diff
@@ -1596,41 +1611,35 @@ _Note_: If there are skipped slots between a block and its parent block, run the
...
+* If `state.slot % SLOTS_PER_BATCHING == 0` append `merkle_root(state.latest_block_roots + state.latest_state_roots)` to `state.historical_batchings`.
```
djrtwo: Should we use `hash_tree_root` on the concatenated list instead?

specs/core/0_beacon-chain.md
djrtwo: Also, at best, this just overwrites a value with itself. What is the intention here?

specs/core/0_beacon-chain.md
```diff
@@ -1596,41 +1611,35 @@ _Note_: If there are skipped slots between a block and its parent block, run the
 
 ### Per-slot processing
 
-Below are the processing steps that happen at every slot.
-
-#### Slot
+Below are the processing steps that happen at every `slot > GENESIS_SLOT`.
```
djrtwo: So first we do book-keeping for the previous slot? This smells a bit funny

specs/core/0_beacon-chain.md
djrtwo: Seems like maybe we should be doing the following after the `state.slot += 1`

```
* Set `state.latest_block_roots[state.slot % SLOTS_PER_BATCHING] = get_block_root(state, state.slot - 1)`.
```

specs/core/0_beacon-chain.md
JustinDrake: The reason I went with `merkle_root` is that if we consider the concatenation as a `List` then we need to include the list size in the `hash_tree_root`. Here we have a fixed size data structure (let's call that an "array") so including the size is a nuisance. It's the same problem we have with shard data.

This is arguably a bug with `hash_tree_root`. We already support byte arrays (fixed size) I'll file an issue to generalise friendly Merkleisation to arbitrary arrays. 👍

vbuterin: I'd say `merkle_root` is fine here.

specs/core/0_beacon-chain.md
JustinDrake: Agreed. Fixed!

specs/core/0_beacon-chain.md
JustinDrake: I changed things around. Does that make more sense?

specs/core/0_beacon-chain.md
JustinDrake: > This is more failsafe for any initial state operations that might subtract to a previous epoch/slot.

Interesting! I guess we can review this when we finalise the slot/epoch underflow issue.

> the only thing before the genesis_block is the genesis_block

Semantically, the only thing before the `genesis_block` is nothingness, i.e. `ZERO_HASH`.

> currently not initialized at all

`latest_state_root` is initialised to `[ZERO_HASH for _ in range(SLOTS_PER_BATCHING)]`. I think that's the "semantically correct" way of doing it.

specs/core/0_beacon-chain.md
```diff
@@ -1596,41 +1611,35 @@ _Note_: If there are skipped slots between a block and its parent block, run the
...
 
+* Set `state.latest_state_roots[state.slot % SLOTS_PER_BATCHING] = hash_tree_root(state)`.
 * Set `state.slot += 1`.
-
-#### Block roots
-
-* Let `previous_block_root` be the `hash_tree_root` of the previous beacon block processed in the chain.
-* Set `state.latest_block_roots[(state.slot - 1) % LATEST_BLOCK_ROOTS_LENGTH] = previous_block_root`.
-* If `state.slot % LATEST_BLOCK_ROOTS_LENGTH == 0` append `merkle_root(state.latest_block_roots)` to `state.batched_block_roots`.
+* Set `state.latest_block_roots[state.slot % SLOTS_PER_BATCHING] = get_block_root(state, state.slot - 1)`.
```
djrtwo: These are really `block_body_roots` now. I think we should reserve `block_root` for the actual root of the block.

I'm a bit concerned about the mismatch between storing block body roots in the state vs actual block roots which will likely be used on the p2p network.

Also this changes what committees are attesting to. They used to attest to the actual block root. Now the validity rule is about attesting to this body (which doesn't even include the state/header). We are certainly changing meaning here and need to make sure it's actually the intention.

djrtwo: Also due to lines like "Verify that `attestation.data.justified_block_root` is equal to `get_block_root(state, get_epoch_start_slot(attestation.data.justified_epoch))`", we are really just finalizing the block body which isn't even a commitment to a chain (no `parent_root` in block body) and we could see the same block body in multiple chains.

JustinDrake: These are all great points! We discussed and agreed to rollback to block roots with the following changes:

* Add `latest_block_body_root` so that `parent_block_root` can be reconstructed from `state`.
* Define `block_root` to be `signed_root(block.header, "signature")`.

One discussion point is that we need canonical identifiers for various signed objects (e.g. blocks, attestations, transfers), for example in the context of block explorers. Our possibility is to use `signed_root(object, "signature")` as that canonical identifier.

JustinDrake: I think defining `block_root = signed_root(block.header, "signature")` isn't ideal so I've reverted back to `block_root = hash_tree_root(block.header)`. Technically it's a `block_header_root` though...

JustinDrake: Ok no `block_root = hash_tree_root(block)` and order is restored 👍 :)

specs/core/0_beacon-chain.md
```diff
@@ -1592,41 +1598,38 @@ _Note_: If there are skipped slots between a block and its parent block, run the
...
-#### Slot
-
-* Verify that `block.slot == state.slot`.
-
-#### Block signature
+#### Block header
 
+* Verify that `block.header.slot == state.slot`.
+* Verify that `block.header.block_body_root == hash_tree_root(block.body)`.
+* Set `state.latest_partial_block.header.state_root = get_state_root(state, state.slot - 1)`.
```
djrtwo: This section is pretty opaque. The block roots that go into state when there are empty slots involved having a modified state root? And the block header is supposed to point to this modified block instead of it's actual parent?

The block needs to point to another real block, otherwise the client cannot figure out where the state transition to the block is from. It would have to state transition from all possible blocks to the block's prev-slot to try to find the possible parent root.

I might be missing something here. can you write out in a few plain english steps what we are trying to accomplish here?

JustinDrake: The idea is that `state.latest_partial_block` contains the latest block with `header.state_root` zeroed out (to avoid the recursion). On line 1616 we reconstruct the (full) previous block with `header.state_root` filled in, and on line 1617 we check that the `previous_block_root` matches. I'll create the `previous_block` variable to clear things up.

specs/core/0_beacon-chain.md
```diff
@@ -1592,41 +1614,38 @@ _Note_: If there are skipped slots between a block and its parent block, run the
...
 
-#### Slot
+#### Block header
 
 * Verify that `block.slot == state.slot`.
-
-#### Block signature
-
+* Let `previous_block_header = state.latest_block_header`.
+* Set `previous_block_header.state_root = get_state_root(state, state.slot - 1)`.
```
djrtwo: This will not rebuild the "parent block header" in all cases (skipped slots). It will rebuild some weird intermediate made up block header because the way state is being pulled here. My first guess on fixing this would be to use the slot from the header rather than just the previous slot. Dumping this thought here and thinking about it tomorrow.

```suggestion
* Set `previous_block_header.state_root = get_state_root(state, previous_block_header.slot)`.
```

JustinDrake: Well spotted! Unfortunately `get_state_root` won't work when we have too many empty slots in a row. My suggestion would be to build the full `state.latest_block_header` (with `state_root`) on the next slot. Will fix soon :)

specs/core/0_beacon-chain.md
```diff
@@ -1582,41 +1604,36 @@ _Note_: If there are skipped slots between a block and its parent block, run the
...
 ### Per-slot processing
 
-Below are the processing steps that happen at every slot.
-
-#### Slot
+Below are the processing steps that happen at every `slot >= GENESIS_SLOT`.
 
+* Set `state.latest_state_roots[state.slot % SLOTS_PER_BATCHING] = hash_tree_root(state)`.
+* Set `state.latest_block_roots[state.slot % SLOTS_PER_BATCHING] = get_block_root(state, state.slot - 1)`.
+* Set `state.previous_block_header.state_root = get_state_root(state, state.slot)` if `state.previous_block_header.state_root == ZERO_HASH`.
```
```suggestion
* Set `state.latest_block_header.state_root = get_state_root(state, state.slot)` if `state.latest_block_header.state_root == ZERO_HASH`.
```

JustinDrake: Well spotted!

JustinDrake:
> Is that correct?

Yes! :)

specs/core/0_beacon-chain.md
```diff
@@ -482,6 +485,8 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
 
 ```python
 {
+    'randao_reveal': 'bytes96',
```
djrtwo: Why `randao_reveal` and `Eth1Data` out of header?
I understand because they are more of "operations", but is a client likely to want the randao reveals for tracking the randomness when building a header chain even if they don't want all the operations?

djrtwo: Why `randao_reveal` and `Eth1Data` out of header?
I understand because they are more of "operations", but is a client likely to want the randao reveals for tracking the randomness when building a header chain even if they don't want all the operations?

specs/core/0_beacon-chain.md
```diff
@@ -1403,35 +1410,47 @@ For convenience, we provide the interface to the contract here:
 
 ## On genesis
 
-A valid block with slot `GENESIS_SLOT` (a "genesis block") has the following values. Other validity rules (e.g. requiring a signature) do not apply.
+When enough full deposits have been made to the deposit contract a `Eth2Genesis` log is emitted. Construct a corresponding `genesis_state` and `genesis_block` as follows:
```
```suggestion
When enough full deposits have been made to the deposit contract, an `Eth2Genesis` log is emitted. Construct a corresponding `genesis_state` and `genesis_block` as follows:
```

specs/core/0_beacon-chain.md
```diff
@@ -1403,35 +1410,47 @@ For convenience, we provide the interface to the contract here:
...
+
+* Let `genesis_validator_deposits` be the list of deposits, ordered chronologically, up to and including the deposit that triggered the `Eth2Genesis` log.
+* Let `genesis_time` be the timestamp specified in the `Eth2Genesis` log.
+* Let `genesis_eth1_data` be the `Eth1Data` object where:
+    * `genesis_eth1_data.deposit_root` is the deposit root when the `Eth2Genesis` log was emitted
```
```suggestion
    * `genesis_eth1_data.deposit_root` is the `deposit_root` contained in the `Eth2Genesis` log.
```

specs/core/0_beacon-chain.md
```diff
@@ -1403,35 +1410,47 @@ For convenience, we provide the interface to the contract here:
...
+    * `genesis_eth1_data.block_hash` is the hash of the block that emitted the `Eth2Genesis` log
```
```suggestion
    * `genesis_eth1_data.block_hash` is the hash of the Ethereum 1.0 block that emitted the `Eth2Genesis` log.
```

specs/core/0_beacon-chain.md
```diff
@@ -464,16 +461,22 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
...
+    'signature': 'bytes96',
+}
+```
+
+#### `BeaconBlockHeader`
+
+```python
+{
+    'slot': 'uint64',
+    'previous_block_root': 'bytes32',
```
djrtwo: why `previous_` instead of `parent_`?

specs/core/0_beacon-chain.md
```diff
@@ -1587,41 +1608,36 @@ _Note_: If there are skipped slots between a block and its parent block, run the
 
 ### Per-slot processing
 
-Below are the processing steps that happen at every slot.
-
-#### Slot
+Below are the processing steps that happen at every `slot >= GENESIS_SLOT`.
```
djrtwo: I think it's actually `slot > GENESIS_SLOT` as the genesis state is the post state of the genesis block.

```suggestion
Below are the processing steps that happen at every `slot > GENESIS_SLOT`.
```

specs/core/0_beacon-chain.md
```diff
@@ -1587,41 +1608,36 @@ _Note_: If there are skipped slots between a block and its parent block, run the
...
 
+* Set `state.latest_state_roots[state.slot % SLOTS_PER_BATCHING] = hash_tree_root(state)`.
```
djrtwo: This means that the `latest_state_roots` for a `slot` is the post state of `slot - 1` and the _pre_ state of `slot`.
Is this the intention?

djrtwo: It seems a bit off on the expectation of looking up a `state` at `slot`

specs/core/0_beacon-chain.md
JustinDrake: Consistency—we use `previous_` everywhere else (`previous_epoch`, `previous_version `, `previous_shuffling_start_shard`, etc.)

specs/core/0_beacon-chain.md
JustinDrake: > is a client likely to want the randao reveals for tracking the randomness when building a header chain even if they don't want all the operations?

I can't think of a reason off the top of my head why randao reveals would be better placed in the header. One thing to point out is that the randao reveal is just a tiny Merkle path away from the `block_body_root`, itself part of the header.

> Why `randao_reveal` and `Eth1Data` out of header?

Original motivations:

**Aesthetics**: Keeps `BeaconBlockHeader` minimal, and as you mention it may be semantically cleaner.
**Overhead**: Smaller headers mean a bit less overhead in various places (header gossiping, proposer slashings, `state.latest_block_header`).

specs/core/0_beacon-chain.md
JustinDrake: After thinking about it, I think the current genesis approach is fine, and you're right that we want `>` (as opposed to `>=`).

specs/core/0_beacon-chain.md
JustinDrake: We discussed this yesterday. Notice that `slot += 1` isn't yet applied, so all is good :)

JustinDrake:
```
    for index in range(LATEST_ACTIVE_INDEX_ROOTS_LENGTH):
        state.latest_active_index_roots[index] = genesis_active_index_root
```

@djrtwo Can we just do `latest_active_index_roots[0] = genesis_active_index_root`?

djrtwo:
close in favor of #711

PR #655, opened 2019-02-19
Reduce GENESIS_SLOT to 2**32

JustinDrake:
* Address the slot/epoch underflow problem, even for Java implementers! 🎉
* Squash a bug with `get_previous_epoch`
* Fix #642
* Address #626 (Vitalik, Danny, myself agree that avoiding signed integers is probably best)

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -212,7 +212,7 @@ Code snippets appearing in `this style` are to be interpreted as Python code.
 | Name | Value |
 | - | - |
 | `GENESIS_FORK_VERSION` | `0` |
-| `GENESIS_SLOT` | `2**63` |
+| `GENESIS_SLOT` | `2**32` |
 | `GENESIS_EPOCH` | `slot_to_epoch(GENESIS_SLOT)` |
 | `GENESIS_START_SHARD` | `0` |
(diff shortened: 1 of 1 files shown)


Nashatyrev:
I would vote for `2**20`. It still addresses underflows, but is easier to remember and perform decimal calculations in mind.

JustinDrake:
We will reach `2**20` slots ~70 days after genesis, so `- 2**20` cannot be done by zeroing a single bit. On the other hand it will take 800+ years to reach slot `2**32` so there's no possibility of overflowing past the 32nd bit.

Nashatyrev:
@JustinDrake thanks for clarification! 
Sorry, I probably missed the rationale for zeroing a single bit instead of doing `- GENESIS_SLOT`, can you please recall?

JustinDrake:
It's friendlier to bit operations. For example, we can cast the `slot` to `uint32` (stripping off the top 32 bits), or equivalently `(slot << 32) >> 32`, or equivalent bitwise XOR `slot ^ 2**32`.

Unfortunately there isn't a single canonical way forward so it's liable to bikeshed. Vitalik independently suggested 2**32 on a call recently 😂 🤷‍♂️I think we all don't care too much.

PR #663, opened 2019-02-20
Remove Split Function

nisdas:
This is an outdated function which was used alongside the previous shuffle algorithm.


hwwhww:
Isn't it still being used in `get_shuffling`?

nisdas:
@hwwhww Oh wow it is, I completely missed that :expressionless:. Closing this now

PR #669, opened 2019-02-21
Fix slot range for attestation inclusion

djrtwo:
Attestations should only be able to be included in the current or next epoch. Never in the next next epoch which was being allowed in some cases due to the `+ MIN_ATTESTATION_INCLUSION_DELAY` on the upper end of the range.

#bugz.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1671,7 +1671,8 @@ Verify that `len(block.body.attestations) <= MAX_ATTESTATIONS`.
 
 For each `attestation` in `block.body.attestations`:
 
-* Verify that `attestation.data.slot <= state.slot - MIN_ATTESTATION_INCLUSION_DELAY < attestation.data.slot + SLOTS_PER_EPOCH`.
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1671,7 +1671,8 @@ Verify that `len(block.body.attestations) <= MAX_ATTESTATIONS`.
 
 For each `attestation` in `block.body.attestations`:
 
-* Verify that `attestation.data.slot <= state.slot - MIN_ATTESTATION_INCLUSION_DELAY < attestation.data.slot + SLOTS_PER_EPOCH`.
+* Verify that `attestation.data.slot <= state.slot - MIN_ATTESTATION_INCLUSION_DELAY`.
```
```suggestion
* Verify that `attestation.data.slot + MIN_ATTESTATION_INCLUSION_DELAY <= state.slot`.
```
Reasons:
1. So we can see both the lower-bound and upper-bound of `state.slot` clearly when reading these two lines.
2. The overflow/underflow discussion. Do our best to prevent it anyway.

PR #674, opened 2019-02-22
Update is_power_of_two to not use log2

decanus:


djrtwo:
Works for me. Definitely better and simpler.
I'll note: this is distinctly and implementation detail -- use whatever implementation you want!

PR #678, opened 2019-02-23
Allow transfers from non-activated validators

JustinDrake:
See item 6 in https://github.com/ethereum/eth2.0-specs/issues/675. This makes transfers more generic.


Takenforeverything:
I don't know who the f*** this is or what the f*** you're doing but stop
sending s*** to my goddamn email f****** hackers I was one of the original
stop your s*** yeah

PR #680, opened 2019-02-24
removes redundant checks

decanus:


djrtwo:
I believe this is an extra check to prevent overflow scenarios.

CC: @JustinDrake

JustinDrake:
> I believe this is an extra check to prevent overflow scenarios.

That's right. We could have a blanket statement like "`uint64` additions should never overflow` but I think this is the only practical place where an overflow can be triggered from an addition.

Issue #605, opened 2019-02-11
Cross-blockchain BLS12-381 standardisation


JustinDrake:
This is a heads-up that [the BLS spec](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/bls_signature.md) is likely to change. Many blockchain projects are converging towards BLS12-381 (Zcash Sapling, Chia, Ethereum 2.0, Filecoin, Dfinity, Algorand, etc.) and there is a strong desire for everyone to standardise on the fine details. This include standardising

- hash functions (to G1 and G2)
- generator points (for G1 and G2)
- point serialisation (for G1 and G2)
- possibly message hashing

Given that Zcash Sapling is already in production, one sensible approach may be to start from what Zcash has.

Dan Boneh and folks from Algorand have a [IETF draft](https://tools.ietf.org/id/draft-boneh-bls-signature-00.txt). I encourage anyone interested in this discussion to email Sergey Gorbunov (sgorbunov@uwaterloo.ca) who seems to be leading the effort.

axic:
cc @Schaeff

JustinDrake:
Announcement [tweet](https://twitter.com/Algorand/status/1095038841323483137) and [Medium post](https://medium.com/algorand/posting-bls-standard-draft-30a89723f386).

JustinDrake:
Closing in favour of #675.

PR #681, opened 2019-02-24
bls: Use int_to_bytes8 to serialize domain

wemeetagain:
The function `bytes8` does not exist in the spec. Using `int_to_bytes8`
has a clearer meaning: serialize into 8 bytes as little endian.

### DIFF
--- a/specs/bls_signature.md
+++ b/specs/bls_signature.md
@@ -71,8 +71,8 @@ q = 4002409555221667393417789825735904156556882819939007885332058136124031650490
 
 def hash_to_G2(message_hash: Bytes32, domain: uint64) -> [uint384]:
     # Initial candidate x coordinate
-    x_re = int.from_bytes(hash(message_hash + bytes8(domain) + b'\x01'), 'big')
-    x_im = int.from_bytes(hash(message_hash + bytes8(domain) + b'\x02'), 'big')
+    x_re = int.from_bytes(hash(message_hash + int_to_bytes8(domain) + b'\x01'), 'big')
+    x_im = int.from_bytes(hash(message_hash + int_to_bytes8(domain) + b'\x02'), 'big')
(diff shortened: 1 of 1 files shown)


hwwhww:
1. IMO we would avoid BLS spec referencing core specs.
2. `int_to_bytes8` is little endian converting, while `x_re` and `x_im` are still in big-endian. IMO BLS spec is part of cryptography black box (https://github.com/ethereum/eth2.0-specs/issues/556#issuecomment-460605704). Also, related to #605.

What do you think about explicitly:
```python
    x_re = int.from_bytes(hash(message_hash + domain.to_bytes(8, 'big') + b'\x01'), 'big')
    x_im = int.from_bytes(hash(message_hash + domain.to_bytes(8, 'big') + b'\x02'), 'big')
```

wemeetagain:
If we clarify as you mentioned, that gets at the heart of the issue, that the endian-ness is currently unspecified/unclear.

Perhaps an even cleaner solution would be to pass in domain as a bytes8 rather than as a uint64, and have the bls spec not need to worry about integer serialization at all.

eg:
```python
def hash_to_G2(message_hash: Bytes32, domain: Bytes8) -> [uint384]:
    x_re = int.from_bytes(hash(message_hash + domain + b'\x01'), 'big')
    x_im = int.from_bytes(hash(message_hash + domain + b'\x02'), 'big')
```

hwwhww:
Hey @wemeetagain, I like your second proposal more. :)

However, it will change the interface, and https://github.com/ethereum/eth2.0-specs/issues/605 may change it again later. Therefore, it may not worth it to change the interface recently before #605 due to the overhead of updating test vectors.

wemeetagain:
Oky, I'll close this PR. When the #605 change rolls around, we can give the appropriate input.

PR #687, opened 2019-02-25
remove deposit order todo

djrtwo:
What is the remaining TODO referencing?

> [TODO: update the call to `verify_merkle_branch` below if it needs to change after we process deposits in order]


JustinDrake:
Isn't this a much bigger change? For example we still have `assert bls_verify(`.

djrtwo:
ah, I remember. The change was in effect but a bad merge crushed it. I'll clean it up.

djrtwo:
ready for review

PR #688, opened 2019-02-25
Use `slashed` bool more Pythonicly

decanus:


vbuterin:
Not sure I like this. "Verify that `not proposer.slashed`" doesn't sound right to anyone who's not an expert pythonista, and the spec's primary goal is being human-readable, not being pythonista-readable. Maybe one more moderate pythonization would be `is False` instead of `== False`?

decanus:
@vbuterin updated

PR #695, opened 2019-02-27
 Use custom types in data structures and constants

Nashatyrev:
Original issue: https://github.com/ethereum/eth2.0-specs/issues/667

The goal is to improve Spec readability by using alias custom types in data structures and as constant units instead of generic types.

## What was done 
- Add `Hash` and `Bitfield` custom types. The `Hash` type is treated as a result of any 256-bit cryptographic hash function or its derivative preserving hash properties (like `xor`ing two hashes)
- Replace generic SSZ types in data structures with custom types where appropriate
- Add/update constant units according to custom types 

## What else can be done
- Add more custom types (like `Time`, `ForkVersion`, `Bitfield64`) which are now referred at 1-2 places in the Spec
- As per @hwwhww suggestion in https://github.com/ethereum/eth2.0-specs/issues/667 cover *all* elements with custom types 
- In some way describe array index types (like  `'validator_balances': [ValidatorIndex => Gwei]`)

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -575,7 +575,7 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
     # BLS public key
     'pubkey': BLSPubkey,
     # Withdrawal credentials
-    'withdrawal_credentials': Hash,
+    'withdrawal_credentials': Bytes32,
     # Epoch when validator activated
     'activation_epoch': Epoch,
     # Epoch when validator exited
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -208,7 +208,7 @@ Code snippets appearing in `this style` are to be interpreted as Python code.
 
 | Name | Value | Unit | 
 | - | - | - |
-| `GENESIS_FORK_VERSION` | `int_to_bytes4(0)` | uint64 | 
+| `GENESIS_FORK_VERSION` | `int_to_bytes4(0)` | bytes4 | 
 | `GENESIS_SLOT` | `2**32` | Slot | 
 | `GENESIS_EPOCH` | `slot_to_epoch(GENESIS_SLOT)` | Epoch | 
 | `GENESIS_START_SHARD` | `0` | Shard | 
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -229,9 +229,9 @@ Code snippets appearing in `this style` are to be interpreted as Python code.
 | `MIN_SEED_LOOKAHEAD` | `2**0` (= 1) | Epoch | 6.4 minutes |
 | `ACTIVATION_EXIT_DELAY` | `2**2` (= 4) | Epoch | 25.6 minutes |
 | `EPOCHS_PER_ETH1_VOTING_PERIOD` | `2**4` (= 16) | Epoch | ~1.7 hours |
-| `SLOTS_PER_HISTORICAL_ROOT` | `2**13` (= 8,192) | slots | ~13 hours |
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -556,15 +573,15 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
 ```python
 {
     # BLS public key
-    'pubkey': 'bytes48',
+    'pubkey': BLSPubkey,
     # Withdrawal credentials
-    'withdrawal_credentials': 'bytes32',
+    'withdrawal_credentials': Hash,
```
hwwhww: @JustinDrake 
Should it be `Bytes32`?

Nashatyrev: Already returned it back to `Bytes32` in this commit https://github.com/ethereum/eth2.0-specs/pull/695/commits/a693737275a4180250b4779e9952f3f84d025609

hwwhww: But there are two `withdrawal_credentials`s: one is in `DepositInput` and another one is in `Validator`.

Nashatyrev: Oh, my bad. 
Sure this one should also be `Bytes32`
Thanks for spotting this!

JustinDrake: I'm tempted to nuke `Bytes32` (which is redundant with `bytes32`) and use `Hash` everywhere. After all, `withdrawal_crendentials` fits with the definition: "result of hashing and its derivatives"

Nashatyrev: I'm fine with it. Commited

vbuterin: > After all, `withdrawal_crendentials` fits with the definition: "result of hashing and its derivatives"

Technically the first byte is a version number, but hashes have had built-in version numbers before (see ipfs multi hash) so that's not too terrible...

hwwhww:
I’m inclined to have custom types cover all elements. Compromise: define a general `Uint64`?

Nashatyrev:
IMHO, defining aliases for types which are used just in a couple of places in the spec wouldn't make the spec clearer. 
As well I see no big reason to make `UInt64` => `uint64` alias. 
We probably need someone else opinion here. @djrtwo ?

hwwhww:
Hmm, I think this PR would break CI tests...
Basically, it requires (i) defining a custom type in Python and also (ii) converting Python types to SSZ types when defining SSZ classes.

@Nashatyrev could you rebase it and then let's see how to fix it in `function_puller.py`?

Nashatyrev:
@hwwhww rebased. Let me know if I can help you with CI tests

protolambda:
1) Is this still up to date? (I don't think so)
2) Do we still plan to merge this?
3) FYI: this breaks the typing syntax in many of the SSZ utilities in the pyspec. I.e. the default zero values, random instantiation of ssz, and `minimal_ssz.py` itself. These basic types are declared with strings, not types: they instruct what the typing looks like, they are not the type itself. I recommend to add an alias functionality, where we just lookup the "raw" type for any given type used in the SSZ container type declarations, and only then serialize/hash it etc.

JustinDrake:
> Do we still plan to merge this?

I'm very pro adding aliases for readability. Do you want to go ahead @protolambda and fix whatever needs fixing? We can then merge.

protolambda:
I think we addressed most of the typing ideas with #1077. We can go further with it and use NewType here and there to *describe* some things in addition to typing them (e.g. `Slot`), but I think it's clean enough right now with the current data-types. So I'm closing this PR. Anyone who is interested is welcome to open a new PR for typing improvements, if anything can be improved on the *current* dev branch (this PR is quite out of date now). 

:clap: for bringing this up so early, and apologies from us all that typing didn't make it into the spec so quick, I hope the new more complete typing support makes up for it.

Issue #679, opened 2019-02-24
SSZ TLDR


JustinDrake:
Below is all the SSZ logic needed for the phase 0 spec. We define types ("basic" and "composite"), serialisation of basic types, and `hash_tree_root`. (The TLDR below incorporates tuples—see #665—and a chunk size reduction to 32 bytes.)

**Basic types**

* `uint64` (8-byte little-endian serialisation)
* `byte` (1-byte serialisation)
* `bool` (1-byte serialisation to `0x00` or `0x01`)

**Composite types**

* `Container` (inhomogeneous, fixed sized)
* `Tuple` (homogenous, fixed size)
* `List` (homogeneous, variable size)

**Helper functions**

* `pack`: Given ordered objects of the same basic type, serialise them, pack them into 32-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
* `merkleize`: Given ordered 32-byte chunks, right-pad them with zero chunks to the closest power of two, Merkleize the chunks, and return the root.
* `mix_in_length`: Given a Merkle root `r` and a length `l` (32-byte little-endian serialisation) return `hash(r + l)`.

**`hash_tree_root`**

Let `o` be an object. We recursively define `hash_tree_root(o)` as:

* `merkleize(pack(o))` if `o` is a basic object or a tuple of basic objects
* `mix_in_length(merkleize(pack(o)), len(o))` if `o` is a list of basic objects
* `merkleize([hash_tree_root(element) for element in o])` if `o` is a tuple of composite objects or a container
* `mix_in_length(merkleize([hash_tree_root(element) for element in o]), len(o))` if `o` is a list of composite objects

vbuterin:
I recall the 32 -> 128 change only reducing hashing time by ~1/3, so reducing back to seems reasonable. I actually think that if we decide re-hashing the (max 32 MB) validator balances array every epoch is too much, we can eke out a better efficiency/ugliness tradeoff by constricting validator balances from 8 bytes to 4 (or even storing the lower 4 bytes in the balances array and the upper 4 bytes in the registry; doing so would actually be a significant boon for light clients [1/3 Merkle proof size decrease] and may even be worth just doing now!)

JustinDrake:
> or even storing the lower 4 bytes in the balances array and the upper 4 bytes in the registry

[...]

Issue #665, opened 2019-02-20
Full support for tuples


JustinDrake:
SSZ has full support for lists, but only partial support for arrays (specifically, it can support arrays of bytes with `bytesN`). I'd argue we want full support for fixed-sized arrays. The reason is that fixed-sized arrays appear in various places where we do not want to mix in the array length. One example is when computing the `shard_data_root`, and another one appeared [here](https://github.com/ethereum/eth2.0-specs/pull/649/files#r258565227). Other examples include `latest_randao_mixes`, `latest_block_roots`, `latest_active_index_roots`, `latest_slashed_balances` which are all fixed size, and where including the data length is a nuisance.

Ideally we want to use `hash_tree_root` over `merkle_root` pretty much everywhere, and support for fixed-sized arrays will allow this.

(Side note regarding Go: In go we have "arrays" with the `[n]` notation for a fixed `n`. For example `[32]byte` would be the equivalent of `Bytes32`. We also have "slices" with the `[]` which have dynamic size. My suggestion is to define SSZ for slices *and* arrays.)

djrtwo:
I would lean toward favoring pythonic notation because the rest of the spec uses it.
"slices" in the go nomenclature are just dynamically sized arrays?

Why is including the data length on these arrays a "nuisance"? Is there any loss other than a tiny increase in size of the hashed item?

JustinDrake:
I think the idea to use SSZ everywhere (as opposed to the lower-level `merkle_root`) was dismissed some time ago (on our Telegram, or maybe our weekly call) because:

* Having the length complicates the proof-of-custody challenges
* Having the length complicates STARK-based erasure coding

In addition to the above, I would say that we want SSZ to be maximally friendly to merkleisation with the hope of SSZ becoming a standard. It's also slightly more efficient and friendly for the fixed-size components of `state` to be exposed as pure Merkle roots.

> I would lean toward favoring pythonic notation

Right, we should use whatever notation is most Pythonic! :)

JustinDrake:
> I'm not sure what the Python equivalent is.

I think in Python a fixed-size list is called a "tuple".

hwwhww:
> `latest_randao_mixes`, `latest_block_roots`, `latest_active_index_roots`, `latest_slashed_balances` which are all fixed size

*Off this topic*

1. Since we use the fixed size arrays for recording recent data. It seems we assume that "State list lengths" are un-changeable constants in the future forks?
[...]

PR #696, opened 2019-02-27
Rewrite SSZ spec

JustinDrake:
* Simplify presentation where appropriate. For example, deserialisation is implicit from serialisation (similar to `bls_sign` being implicit from `bls_verify`) and is left as an implementation exercise.
* Dramatically reduce spec size and hopefully improve readability.

**Spec changes**:

* Support for tuples (see #665)
* Reduce chunk-size reduction 32 bytes (see #679)
* Simplify logic of length prefix serialisation (the same logic is used for lists and containers)
* Restrict the `uintN` type (`N` must be in `[8, 16, 32, 64, 128, 256]`).
* Modify `signed_root` to take single ssz object, returning the `hash_tree_root` of the object less the last field
* Do _not_ length pre-fix fixed sized containers (i.e. does not embed a list)

### DIFF
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -1,6 +1,6 @@
 # [WIP] SimpleSerialiZe (SSZ)
 
-This is a **work in progress** describing typing, serialisation and Merkleisation of Ethereum 2.0 objects.
+This is a **work in progress** describing typing, serialization and Merkleization of Ethereum 2.0 objects.
 
 ## Table of contents
 
@@ -25,8 +25,8 @@ This is a **work in progress** describing typing, serialisation and Merkleisatio
 
 | Name | Value | Definition |
 |-|:-:|-|
-| `LENGTH_BYTES` | 4 | Number of bytes for the length of variable-length serialized objects. |
-| `MAX_LENGTH` | 2**(8 * LENGTH_BYTES) | Maximum serialization length. |
+| `LENGTH_BYTES` | `4` | Number of bytes for the length of variable-length serialized objects. |
+| `MAX_LENGTH` | `2**(8 * LENGTH_BYTES)` | Maximum serialization length. |
 
 ## Types
 
@@ -104,15 +104,15 @@ return serialized_length + serialized_bytes
 
 ## Deserialization
 
-Given a type, serialisation is an injective function from objects of that type to byte strings. That is, deserialisation—the inverse function—is well-defined.
+Given a type, serialization is an injective function from objects of that type to byte strings. That is, deserialization—the inverse function—is well-defined.
 
 ## Merkleization
 
 We first define helper functions:
 
-* `pack`: Given ordered objects of the same basic type, serialise them, pack them into 32-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
-* `merkleise`: Given ordered 32-byte chunks, right-pad them with zero chunks to the closest power of two, Merkleize the chunks, and return the root.
-* `mix_in_length`: Given a Merkle root `r` and a length `l` (32-byte little-endian serialisation) return `hash(r + l)`.
+* `pack`: Given ordered objects of the same basic type, serialize them, pack them into 32-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
+* `merkleize`: Given ordered 32-byte chunks, right-pad them with zero chunks to the closest power of two, Merkleize the chunks, and return the root.
+* `mix_in_length`: Given a Merkle root `root` and a length `length` (32-byte little-endian serialization) return `hash(root + length)`.
 
 Let `o` be an object. We now define object Merkleization `hash_tree_root(o)` recursively:
 
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -4,11 +4,9 @@ This is a **work in progress** describing typing, serialization and Merkleizatio
 
 ## Table of contents
 
-- [Constants](#constants)
 - [Typing](#typing)
     - [Basic types](#basic-types)
     - [Composite types](#composite-types)
-    - [Notation](#notation)
     - [Aliases](#aliases)
 - [Serialization](#serialization)
     - [`uintN`](#uintn)
@@ -19,31 +17,18 @@ This is a **work in progress** describing typing, serialization and Merkleizatio
 - [Self-signed containers](#self-signed-containers)
 - [Implementations](#implementations)
 
-## Constants
-
-| Name | Value | Description |
-|-|-|-|
-| `LENGTH_BYTES` | `4` | Number of bytes for the length of variable-length serialized objects. |
-| `MAX_LENGTH` | `2**(8 * LENGTH_BYTES)` | Maximum serialization length. |
-
 ## Typing
 
-#### Basic types
+### Basic types
 
 * `uintN`: `N`-bit unsigned integer (where `N in [8, 16, 32, 64, 128, 256]`)
 * `bool`: 1-bit unsigned integer
 
-#### Composite types
+### Composite types
 
-* **container**: ordered heterogenous collection of values
-* **tuple**: ordered fixed-length homogeneous collection of values
-* **list**: ordered variable-length homogenous collection of values 
-
-#### Notation
-
-* **container**: key-pair curly braket notation `{}` (e.g. `{'key1': uint64, 'key2': bool}`)
-* **tuple**: angle braket notation `[N]` (e.g. `uint64[N]`)
-* **list**: angle braket notation `[]` (e.g. `uint64[]`)
+* **container**: ordered heterogenous collection of values (key-pair curly braket notation `{}`, e.g. `{'foo': uint64, 'bar': bool}`)
+* **tuple**: ordered fixed-length homogeneous collection of values (angle braket notation `[N]`, e.g. `uint64[N]`)
+* **list**: ordered variable-length homogenous collection of values (angle braket notation `[]`, e.g. `uint64[]`)
 
 #### Aliases
 
@@ -52,7 +37,6 @@ For convenience we alias:
 * `byte` to `uint8`
 * `bytes` to `byte[]`
 * `bytesN` to `byte[N]`
-* `bit` to `bool`
 
 ## Serialization
 
@@ -75,8 +59,9 @@ return b'\x01' if value is True else b'\x00'
 #### Containers, tuples, lists
 
 ```python
+LENGTH_BYTES = 4
 serialized_bytes = ''.join([serialize(element) for element in value])
-assert len(serialized_bytes) < MAX_LENGTH
+assert len(serialized_bytes) < 2**(8 * LENGTH_BYTES)
 serialized_length = len(serialized_bytes).to_bytes(LENGTH_BYTES, 'little')
 return serialized_length + serialized_bytes
 ```
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -6,7 +6,7 @@ This is a **work in progress** describing typing, serialization and Merkleizatio
 
 - [Constants](#constants)
 - [Typing](#typing)
-    - [Primitive types](#primitive-types)
+    - [Basic types](#basic-types)
     - [Composite types](#composite-types)
     - [Notation](#notation)
     - [Aliases](#aliases)
@@ -30,7 +30,7 @@ This is a **work in progress** describing typing, serialization and Merkleizatio
 
 ## Typing
 
-#### Primitive types
+#### Basic types
 
 * `uintN`: `N`-bit unsigned integer (where `N in [8, 16, 32, 64, 128, 256]`)
 * `bool`: 1-bit unsigned integer
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -64,7 +64,7 @@ We reccursively define the `serialize` function which consumes an object `object
 
 ```python
 assert N in [8, 16, 32, 64, 128, 256]
-return object.to_bytes(N / 8, 'little')
+return object.to_bytes(N // 8, 'little')
 ```
 
 #### `bool`
@@ -111,7 +111,7 @@ Given a type, serialization is an injective function from objects of that type t
 We first define helper functions:
 
 * `pack`: Given ordered objects of the same basic type, serialize them, pack them into 32-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
-* `merkleize`: Given ordered 32-byte chunks, right-pad them with zero chunks to the closest power of two, Merkleize the chunks, and return the root.
+* `merkleize`: Given ordered 32-byte chunks, right-pad them with zero chunks to the next power of two, Merkleize the chunks, and return the root.
 * `mix_in_length`: Given a Merkle root `root` and a length `length` (32-byte little-endian serialization) return `hash(root + length)`.
 
 Let `object` be an object. We now define object Merkleization `hash_tree_root(object)` recursively:
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -18,7 +18,7 @@ This is a **work in progress** describing typing, serialization and Merkleizatio
     - [Lists](#lists)
 - [Deserialization](#deserialization)
 - [Merkleization](#merkleization)
-- [Signed containers](#signed-containers)
+- [Self-signed containers](#self-signed-containers)
 - [Implementations](#implementations)
 
 ## Constants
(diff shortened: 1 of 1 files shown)


specs/simple-serialize.md
```diff
@@ -1,424 +1,141 @@
...
+#### Tuples
 
 ```python
-byte_length = int_size / 8
-new_index = current_index + byte_length
-assert(len(rawbytes) >= new_index)
-return int.from_bytes(rawbytes[current_index:current_index+byte_length], 'little'), new_index
+serialized_elements = [serialize(element) for element in object]
+serialized_bytes = reduce(lambda x, y: x + y, serialized_elements)
+return serialized_bytes
```
jannikluhn: Why don't tuples have a length prefix? I thought that maybe the reason is that it's not necessary because one knows the number of elements in advance, but the same is true for containers which do have the prefix.

specs/simple-serialize.md
```diff
@@ -1,424 +1,141 @@
...
 ## Constants
 
-| Constant          | Value | Definition                                                                            |
-|:------------------|:-----:|:--------------------------------------------------------------------------------------|
-| `LENGTH_BYTES`    |   4   | Number of bytes used for the length added before a variable-length serialized object. |
-| `SSZ_CHUNK_SIZE`  |  128  | Number of bytes for the chunk size of the Merkle tree leaf.                           |
+| Name | Value | Description |
+|-|-|-|
+| `LENGTH_BYTES` | `4` | Number of bytes for the length of variable-length serialized objects. |
+| `MAX_LENGTH` | `2**(8 * LENGTH_BYTES)` | Maximum serialization length. |
```
jannikluhn: Should the chunk size be defined here as well (with a note that it must be greater or equal than the maximum size of a basic type)

specs/simple-serialize.md
```diff
@@ -1,424 +1,141 @@
...
-
-#### Container
-
-Recursively tree hash the values in the container in the same order as the fields, and Merkle hash the results.
-
-```python
-return merkle_hash([hash_tree_root_internal(getattr(x, field)) for field in value.fields])
-```
+* `pack`: Given ordered objects of the same basic type, serialize them, pack them into 32-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
+* `merkleize`: Given ordered 32-byte chunks, right-pad them with zero chunks to the closest power of two, Merkleize the chunks, and return the root.
```
```suggestion
* `merkleize`: Given ordered 32-byte chunks, right-pad them with zero chunks to the next power of two, Merkleize the chunks, and return the root.
```

as the closest power of two might be smaller

specs/simple-serialize.md
JustinDrake: > the reason is that it's not necessary because one knows the number of elements in advance

Right! I think the idea is that developers can free up a buffer of the right size when reading data on the wire.

> the same is true for containers which do have the prefix

For containers `len(serialized_bytes)` is not always known ahead of time.

It's a bit awkward that length is handled in three different ways for contains, lists and tuples. Maybe we can just use `len(serialized_bytes)` everywhere.

JustinDrake: @jannikluhn: I've gone ahead and used the exact same serialisation logic for containers, lists and tuples. It's slightly less efficient but goes well with the "Simple" theme of SSZ.

cc @djrtwo, @vbuterin

specs/simple-serialize.md
JustinDrake: The chunk size is chosen to match the output size of `hash` so is not really configurable.

specs/simple-serialize.md
```diff
@@ -1,424 +1,108 @@
...
-| `uintN`   | Type of `N` bits unsigned integer, where ``N % 8 == 0``.   |
-
-Convert directly to bytes the size of the int. (e.g. ``uint16 = 2 bytes``)
-
-All integers are serialized as **little endian**.
-
-| Check to perform       | Code                  |
-|:-----------------------|:----------------------|
-| Size is a byte integer | ``int_size % 8 == 0`` |
+#### `uintN`
```
djrtwo: reduce header level to `###`

```suggestion
### `uintN`
```

specs/simple-serialize.md
```diff
@@ -1,424 +1,108 @@
...
 ```
 
-#### bool
-
-Convert directly to a single 0x00 or 0x01 byte.
-
-| Check to perform  | Code                       |
-|:------------------|:---------------------------|
-| Value is boolean  | ``value in (True, False)`` |
+#### `bool`
```
djrtwo:
```suggestion
### `bool`
```

specs/simple-serialize.md
```diff
@@ -1,424 +1,108 @@
...
-|:--------------------------------------------|:----------------------------|
-| Length of serialized list fits into 4 bytes | ``len(serialized) < 2**32`` |
-
-1. Serialize all list elements individually and concatenate them.
-2. Prefix the concatenation with its length encoded as a `4-byte` **little-endian** unsigned integer.
-
-We define `bytes` to be a synonym of `List[bytes1]`.
-
-**Example in Python**
+#### Containers, tuples, lists
```
djrtwo:
```suggestion
### Containers, tuples, lists
```

specs/simple-serialize.md
```diff
@@ -1,424 +1,108 @@
...
-
-for item in value:
-   serialized_list_string += serialize(item)
-
-assert(len(serialized_list_string) < 2**32)
-
-serialized_len = (len(serialized_list_string).to_bytes(LENGTH_BYTES, 'little'))
-
-return serialized_len + serialized_list_string
+serialized_bytes = ''.join([serialize(element) for element in value])
```
djrtwo: Note sure that I like the ssz type to be implicit in `element`. I think the more explicit `def serialize(value, ssz_type)` is clearer.

specs/simple-serialize.md
```diff
@@ -1,424 +1,108 @@
...
+LENGTH_BYTES = 4
```
djrtwo: I think it clearer to keep the constant section at the top and define `LENGTH_BYTES` and `CHUNK_SIZE` there. And then in `Merkleization` section to call them "`CHUNK_SIZE`-byte chunks"

specs/simple-serialize.md
```diff
@@ -1,424 +1,108 @@
...
-
-4. Prepend the length to the serialized buffer.
-
-**Example in Python**
-
-```python
-def get_field_names(typ):
-    return typ.fields.keys()
+* `pack`: Given ordered objects of the same basic type, serialize them, pack them into 32-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
+* `merkleize`: Given ordered 32-byte chunks, right-pad them with zero chunks to the next power of two, Merkleize the chunks, and return the root.
```
djrtwo: Make it clearer that the number of chunks should be a power of two

specs/simple-serialize.md
```diff
@@ -1,424 +1,108 @@
...
+* `mix_in_length`: Given a Merkle root `root` and a length `length` (32-byte little-endian serialization) return `hash(root + length)`.
```
djrtwo:
```suggestion
* `mix_in_length`: Given a Merkle root `root` and a length `length` (`uint256` little-endian serialization) return `hash(root + length)`.
```

specs/simple-serialize.md
JustinDrake: How do you suggest changing "right-pad them with zero chunks to the next power of two"?

specs/simple-serialize.md
```diff
@@ -1,424 +1,108 @@
...
-* [Implementations](#implementations)
+## Typing
 
-## About
+### Basic types
 
-`SimpleSerialize` was first proposed by Vitalik Buterin as the serialization
-protocol for use in the Ethereum 2.0 Beacon Chain.
+* `uintN`: `N`-bit unsigned integer (where `N in [8, 16, 32, 64, 128, 256]`)
+* `bool`: 1-bit unsigned integer
```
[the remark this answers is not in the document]
JustinDrake: Conceptually it's 1-bit. See [here](https://github.com/ethereum/eth2.0-specs/blob/bac352e70f825da9ac5a7155a5a8581033de15c3/specs/simple-serialize.md#bool) for the 1-byte serialisation.

specs/simple-serialize.md
```diff
@@ -1,424 +1,108 @@
...
 
-The core feature of `ssz` is the simplicity of the serialization with low
-overhead.
+### Composite types
 
-## Variables and Functions
+* **container**: ordered heterogenous collection of values
+    * key-pair curly braket notation `{}`, e.g. `{'foo': uint64, 'bar': bool}`
```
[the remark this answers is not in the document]
JustinDrake: I see that [Wikipedia says](https://en.wikipedia.org/wiki/Bracket#Curly_brackets) it's "braces" in the US :joy:

specs/simple-serialize.md
```diff
@@ -1,424 +1,108 @@
...
-| `SSZ_CHUNK_SIZE`  |  128  | Number of bytes for the chunk size of the Merkle tree leaf.                           |
+* `byte` to `uint8`
+* `bytes` to `byte[]`
+* `bytesN` to `byte[N]`
 
-## Overview
+## Serialization
 
-### Serialize/Encode
+We reccursively define the `serialize` function which consumes an object `value` (of the type specified) and returns a byte string of type `bytes`.
```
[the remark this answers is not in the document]
JustinDrake: > the language implementation might interpret it rather than defining the output as an SSZ type itself

That doesn't work great because we define `serialize` recursively, so the output of `serialize` itself needs to be an object of a well-defined type.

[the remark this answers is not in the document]
djrtwo: I see. sounds good

specs/simple-serialize.md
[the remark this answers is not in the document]
djrtwo: lol. I just meant "braket" without the expected "c". "Bracket" I think is fine

specs/simple-serialize.md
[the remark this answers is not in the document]
djrtwo: But in practice it's 1 byte.
Not sure if noting that it is "1-bit" is worthwhile at this point in the spec. A reader will naturally think (or at least ponder if) bools are serialized as 1 bit.

specs/simple-serialize.md
vbuterin: We already to `tree_hash_root(value)` all over the place with the type implicit; I feel like it's been clear enough for our purposes.

specs/simple-serialize.md
```diff
@@ -1,424 +1,114 @@
...
+return serialized_length + serialized_bytes
 ```
 
-To `hash_tree_root_internal` a list, we simply do:
+## Deserialization
 
-```python
-return merkle_hash([hash_tree_root_internal(item) for item in value])
-```
+Given a type, serialization is an injective function from objects of that type to byte strings. That is, deserialization—the inverse function—is well-defined.
```
vbuterin: This sounds confusing and not sure how many people will get what this means.

Maybe something like "Because serialization is an injective function (ie. if two objects of the same type are not exactly identical they will serialize to different values), for any bytestring there will always be at most one object of any given type that it could deserialize to (or zero if a bytestring is not valid SSZ). An efficient algorithm for computing this object can be found in the implementation [here](http://github.com/fill/in/the/correct/link)."

specs/simple-serialize.md
```diff
@@ -1,424 +1,114 @@
...
-```
-
-### Signed roots
+We now define Merkleization `hash_tree_root(value)` of an object `value` recursively:
 
-Let `field_name` be a field name in an SSZ container `container`. We define `truncate(container, field_name)` to be the `container` with the fields from `field_name` onwards truncated away. That is, `truncate(container, field_name) = [getattr(container, field)) for field in value.fields[:i]]` where `i = value.fields.index(field_name)`.
+* `merkleize(pack(value))` if `value` is a basic object or a tuple of basic objects
+* `mix_in_length(merkleize(pack(value)), len(value))` if `value` is a list of basic objects
+* `merkleize([hash_tree_root(element) for element in value])` if `value` is a tuple of composite objects or a container
+* `mix_in_length(merkleize([hash_tree_root(element) for element in value]), len(value))` if `value` is a list of composite objects
```
vbuterin: > if `value` is a list of basic objects

I'd personally prefer "if `value` is a list of objects whose serialized values have a length of <32 bytes. The reason is that in the future we may want to serialize arrays of tuples of small variables and have them be packed efficiently; for one example of this see the [CBC proposal](https://github.com/ethereum/eth2.0-specs/issues/701) which includes 13-byte containers.

I would say consider something like this:

* `pack`: Given ordered objects of the same type, serialize them, right-pad each object with zero bytes to the next power of 2 if the size of the object is not a power of 2, pack them into `BYTES_PER_CHUNK`-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.

* `merkleize(pack(value))` if `value` is an object or tuple of objects where each object has serialized size <= 32 bytes.
 * `mix_in_length(merkleize(pack(value)), len(value))` if `value` is a list of objects where each object has serialized size <= 32 bytes.
* `merkleize([hash_tree_root(element) for element in value])` if value is a tuple of objects with serialized size > 32 bytes or a container.
* `mix_in_length(merkleize([hash_tree_root(element) for element in value]), len(value))` if `value` is a list of objects with serialized size > 32 bytes.

specs/simple-serialize.md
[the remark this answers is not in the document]
vbuterin: Yeah, I'd say just use bracket, with modifiers (square, curly, angle) where needed, except "parentheses" for the round ones (ie. these).

specs/simple-serialize.md
JustinDrake: Agreed with Vitalik :) I'm adding the following note as a compromise.

> *Note*: In the function definitions below (`serialize`, `hash_tree_root`, `signed_root`, etc.) objects implicitly carry their type.

specs/simple-serialize.md
JustinDrake: screenshot 2019-02-28 at 11 28 25

specs/simple-serialize.md
JustinDrake: Reworded to "if necessary append zero chunks so that the number of chunks is a power of two" 👍

specs/simple-serialize.md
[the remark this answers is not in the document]
JustinDrake: Changed it to

> * `bool`: `True` or `False`

specs/simple-serialize.md
JustinDrake: Thanks @jannikluhn, @djrtwo, @vbuterin for the reviews. All comments have been addressed :)

djrtwo:
"Implement tuples and a chunk-size reduction to 32 bytes (see #665 and #679)"
^ Are these the only actual changes that an implementer will need to make?

JustinDrake:
I've updated the description with the changes:

**Spec changes**:

* Support for tuples (see #665)
* Reduce chunk-size reduction 32 bytes (see #679)
* Simplify logic of length prefix serialisation (the same logic is used for lists, tuples and containers)
* Restrict the `uintN` type (`N` must be in [8, 16, 32, 64, 128, 256]`).

zilm13:
I'm trying to implement `hash_tree_root` according to this spec and have few questions/thoughts:
1. `Given ordered objects of the same basic type`  
Could we clarify here that `type and size`, just to make it clear

2. `Merkleize the chunks, and return the root`  
It's not clear from spec, what is `merkleization`

3. Given the very simple container `{i=1}`, it's `hash_tree_root` is `0100000000..` is it ok? So, hash function is never applied. Or did I miss something?

specs/simple-serialize.md
```diff
@@ -1,424 +1,116 @@
...
 
-#### uintN
+### Composite types
 
-| uint Type | Usage                                                      |
-|:---------:|:-----------------------------------------------------------|
-| `uintN`   | Type of `N` bits unsigned integer, where ``N % 8 == 0``.   |
+* **container**: ordered heterogenous collection of values
+    * key-pair curly bracket notation `{}`, e.g. `{'foo': "uint64", 'bar': "bool"}`
+* **tuple**: ordered fixed-length homogeneous collection of values
```
arnetheduck: aren't these better called `array`:s? `tuples` tend to allow heterogeneous types

JustinDrake: I don't care either way. Implementations can use whatever terminology works best. For example, the natural terminology for Go is "array" (instead of tuple) and "slice" (instead of list). Notice that in Python lists also allow heterogeneous types 🤷‍♂️

mjkeating:
`hash_tree_root(value)` is referenced several times in this new spec, but it appears to no longer be defined in this new spec.  Maybe I'm missing something.

JustinDrake:
@mjkeating "We now define Merkleization hash_tree_root(value)"

PR #699, opened 2019-02-27
fix committee assignment bugs

djrtwo:
2 bugs:
* We were modifying `current_shuffling_epoch` and then calling `get_current_epoch_committee_count` expecting that it was still getting the committee count from the previous value of `current_shuffling_epoch`. Switch operations to prevent this error
* when getting next epoch committee assignments, we were sometimes modifying `committees_per_epoch` and `shuffling_epoch` when these things should have remained constant.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -55,6 +55,7 @@
     - [Helper functions](#helper-functions)
         - [`hash`](#hash)
         - [`hash_tree_root`](#hash_tree_root)
+        - [`signed_root`](#signed_root)
         - [`slot_to_epoch`](#slot_to_epoch)
         - [`get_previous_epoch`](#get_previous_epoch)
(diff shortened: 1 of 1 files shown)


JustinDrake:
At this point `get_crosslink_committees_at_slot` is black magic to me. I wish it could somehow be massively simplified, similar to how we massively simplified the validator state machine (which had a bunch of subtle bugs, and was brittle).

djrtwo:
These fixes are pretty critical. Let's get it merged and try to simplify after

terenc3t:
> At this point `get_crosslink_committees_at_slot` is black magic to me. I wish it could somehow be massively simplified, similar to how we massively simplified the validator state machine (which had a bunch of subtle bugs, and was brittle).

Can confirm, whenever there's a validator run time failure we immediately look at this function then begin to scratch our head 😆

JustinDrake:
> Let's get it merged

Approved!

PR #709, opened 2019-03-01
Unified deposit processing between genesis and transaction

vbuterin:
Only a change of presentation, does NOT make any substantive changes (please verify that this is actually true).

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -387,7 +387,7 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
 ```python
 {
     # Branch in the deposit tree
-    'branch': ['bytes32'],
+    'proof': ['bytes32'],
     # Index in the deposit tree
     'index': 'uint64',
     # Data
@@ -994,16 +994,17 @@ def merkle_root(values: List[Bytes32]) -> Bytes32:
 ### `verify_merkle_branch`
 
 ```python
-def verify_merkle_branch(leaf: Bytes32, branch: List[Bytes32], depth: int, index: int, root: Bytes32) -> bool:
+def verify_merkle_branch(leaf: Bytes32, proof: List[Bytes32], depth: int, index: int, root: Bytes32) -> bool:
     """
-    Verify that the given ``leaf`` is on the merkle branch ``branch``.
+    Verify that the given ``leaf`` is on the merkle branch ``proof``
+    starting with the given ``root``.
     """
     value = leaf
     for i in range(depth):
         if index // (2**i) % 2:
-            value = hash(branch[i] + value)
+            value = hash(proof[i] + value)
         else:
-            value = hash(value + branch[i])
+            value = hash(value + proof[i])
     return value == root
 ```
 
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1247,7 +1290,7 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
         )
     )
     
-    if not proof_is_valid:
+    if not not proof_is_valid:
```
djrtwo: Don't need the double negative.

```suggestion
    if not proof_is_valid:
```

specs/core/0_beacon-chain.md
```diff
@@ -1236,6 +1253,32 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
     """
     deposit_input = deposit.deposit_data.deposit_input
 
+    # Should equal 8 bytes for deposit_data.amount +
+    #              8 bytes for deposit_data.timestamp +
+    #              176 bytes for deposit_data.deposit_input
+    # It should match the deposit_data in the eth1.0 deposit contract
+    serialized_deposit_data = ssz_serialize(deposit.deposit_data)
```
djrtwo: function is just defined as `serialize`

```suggestion
    serialized_deposit_data = serialize(deposit.deposit_data)
```

specs/core/0_beacon-chain.md
```diff
@@ -1707,35 +1750,7 @@ For each `attestation` in `block.body.attestations`:
...
-```
-
-* Run the following:
-
-```python
-process_deposit(state, deposit)
-```
-
-* Set `state.deposit_index += 1`.
+For each `deposit` in `block.body.deposits`, run `process_deposit(state, deposit)`
```
```suggestion
For each `deposit` in `block.body.deposits`, run `process_deposit(state, deposit)`.
```

specs/core/0_beacon-chain.md
```diff
@@ -1236,6 +1253,32 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
...
+    
+    # Verify the Merkle branch
+    merkle_branch_is_valid = verify_merkle_branch(
+        leaf=hash(serialized_deposit_data),
+        branch=deposit.branch,
+        depth=DEPOSIT_CONTRACT_TREE_DEPTH,
+        index=deposit.index,
+        root=state.latest_eth1_data.deposit_root
+    )
+    if not merkle_branch_is_valid:
```
djrtwo: I think we should actually throw here with an `assert` rather than fail silently.
Block should be invalid if the block producer provided an invalid data/branch for the deposit_index/root

specs/core/0_beacon-chain.md
```diff
@@ -1236,6 +1253,31 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
...
+    # Deposits must be processed in order
+    assert deposit.index == state.deposit_index
+    
+    # Verify the Merkle branch
+    merkle_branch_is_valid = verify_merkle_branch(
+        leaf=hash(serialized_deposit_data),
+        branch=deposit.branch,
+        depth=DEPOSIT_CONTRACT_TREE_DEPTH,
+        index=deposit.index,
+        root=state.latest_eth1_data.deposit_root
```
hwwhww: dangling comma:

```suggestion
        root=state.latest_eth1_data.deposit_root,
```

specs/core/0_beacon-chain.md
```diff
@@ -990,6 +991,22 @@ def merkle_root(values: List[Bytes32]) -> Bytes32:
     return o[1]
 ```
 
+### `verify_merkle_branch`
+
+```python
+def verify_merkle_branch(leaf: Bytes32, branch: List[Bytes32], depth: int, index: int, root: Bytes32) -> bool:
```
hwwhww: Can we rename `branch` to `proof`? IMO that makes more sense.

PR #713, opened 2019-03-02
fix randao mixes type, bytes32 like in function and state spec

protolambda:
Simple fix: looks like the state container definition and function say `latest_randao_mixes` should be a `Bytes32`:

```python
'latest_randao_mixes': ['bytes32'],

...

def get_randao_mix(state: BeaconState,
                   epoch: Epoch) -> Bytes32:
```

But the genesis list was initialized with empty signatures (96 bytes).

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1451,7 +1451,7 @@ def get_genesis_beacon_state(genesis_validator_deposits: List[Deposit],
         validator_registry_update_epoch=GENESIS_EPOCH,
 
         # Randomness and committees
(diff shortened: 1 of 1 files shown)


protolambda:
Wait, I see it was ZERO_HASH previously, why was it changed, but not elsewhere?

protolambda:
See #654

JustinDrake:
Thanks @protolambda 👍

(Oh boy, I approved that bad PR! cc @terenc3t)

terenc3t:
> Thanks @protolambda 👍
> 
> (Oh boy, I approved that bad PR! cc @terenc3t)

My bad! I thought we were xor'ing `block.randao_reveal` but it's actually `hash(block.randao_reveal)` 😟

PR #719, opened 2019-03-05
Fix float issue

hwwhww:
### Issue
`1e9` is `float` in Python.

### Proposed solution
Change it to `10**9`.


hwwhww:
close via https://github.com/ethereum/eth2.0-specs/pull/711/commits/ecb76420ecd124c39849aef478d71165633acc68

PR #721, opened 2019-03-06
rename modular_squareroot and add a_flag1 rule

ChihChengLiang:
- modular_squareroot  -> modular_squareroot_in_FQ2
- a_flag1: use the bit in `y_re` when `y_im ==0`.

### DIFF
--- a/specs/bls_signature.md
+++ b/specs/bls_signature.md
@@ -13,7 +13,7 @@
         - [G2 points](#g2-points)
     - [Helpers](#helpers)
         - [`hash_to_G2`](#hash_to_g2)
-        - [`modular_squareroot`](#modular_squareroot)
+        - [`modular_squareroot_in_FQ2`](#modular_squareroot_in_FQ2)
     - [Aggregation operations](#aggregation-operations)
         - [`bls_aggregate_pubkeys`](#bls_aggregate_pubkeys)
         - [`bls_aggregate_signatures`](#bls_aggregate_signatures)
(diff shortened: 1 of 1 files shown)


specs/bls_signature.md
```diff
@@ -59,7 +59,7 @@ We require:
 * `a_flag2 == b_flag2 == c_flag2 == 0`
 * `c_flag1 == 1`
 * if `b_flag1 == 1` then `a_flag1 == x1 == x2 == 0` and `(z1, z2)` represents the point at infinity
-* if `b_flag1 == 0` then `(z1, z2)` represents the point `(x1 * i + x2, y)` where `y` is the valid coordinate such that the imaginary part `y_im` of `y` satisfies `(y_im * 2) // q == a_flag1`
+* if `b_flag1 == 0` then `(z1, z2)` represents the point `(x1 * i + x2, y)` where `y` is the valid coordinate such that the imaginary part `y_im` of `y` satisfies `(y_im * 2) // q == a_flag1`. If `y_im` happens to be `0`, use the real part to set `(y_re * 2) // q == a_flag1`.
```
ChihChengLiang: Not sure how to find an instance of (x, y) whose y_im is zero for the test case 😓.

vbuterin: Now that I think about it, it may actually be impossible to find a test case..... so it's more a "just in case" rule.

I'm inclined to say don't worry about it much because we'll walk through all of this again once the BLS standardization process gets further.

djrtwo:
What's the status here @ChihChengLiang @vbuterin

ChihChengLiang:
Closed in favor of BLS standardization in the future

PR #723, opened 2019-03-06
Add tuple lengths; update list typedef format

dankrad:
Adding tuple lengths in BeaconState object (this changes merkleization/serialization as no length mixin required)


djrtwo:
This is the format that we recently discussed, but note that it is _not_ the format currently found in the SSZ spec. Need to decide one way or another and unify in this PR.

I lean toward this format for easy parsing.

djrtwo:
Shall we make `Deposit.proof` of fixed size length as well? It is a known length of `DEPOSIT_CONTRACT_TREE_DEPTH`

dankrad:
Sounds very reasonable :) Done

Issue #685, opened 2019-02-25
A possible change to how validator balances are stored


vbuterin:
Add to the validator registry a value `rounded_balance`, and reduce the `uint64` in `validator_balances` to `unit32`, renaming the array `validator_fractional_balances`.

Redefine `validator_balances[i]` with the following function:

```python
def get_balance(state: BeaconState, index: int) -> int:
    return (
        state.validator_registry[index].rounded_balance * 10**9 + 
        state.validator_fractional_balances[index]
    )
```

Replace `validator_balances[i] = x` (or `+= x`) with the following function:

```python
def set_balance(state: BeaconState, index: int, new_balance: int):
    validator = state.validator_registry[index]
    if (
        validator.rounded_balance * 10**9 > new_balance or
        validator.rounded_balance * 10**9 + (15 * 10**8) < new_balance
    ):
        validator.rounded_balance = new_balance // 10**9
    state.validator_fractional_balances[index] = (
        new_balance - validator.rounded_balance * 10**9
    )
```

### Motivation

The goal is to split the balance into two components, one an integer number of ETH and the other a fractional part. This accomplishes three functions:

* Reduces the per-epoch re-hashing cost of `validator_balances` by 50%
* Maintains an "approximate balance" that can be used by light clients in the `validator_registry`, reducing the number of Merkle branches per validator they need to download from 3 to 2 (actually often from ~2.01 to ~1.01, because when fetching a committee the Merkle branches in `active_index_roots` are mostly shared), achieving a very significant decrease in light client bandwidth costs
* Maintains an "approximate balance" that can be used in the fork choice rule (this does not technically need to be done in the state, but if it's done here it doesn't need to be done elsewhere...)

The "slack mechanism" (the rounded balance is adjusted down when the balance falls below x, but is adjusted up when it goes above x + 1.5) ensures that a full 0.5 ETH of balance adjustment is required to cause the rounded balance to change. This prevents "on-the-edge attacks" where an attacker maintains a large set of validators whose balances oscillate between x+0.000001 and x-0.000001 for some integer x, causing very frequent changes of the rounded balance and hence very frequent (i) relatively expensive validator registry re-hashings, and (ii) relatively expensive fork choice metadata adjustments.

JustinDrake:
Minor suggestions:

1) Stick with the `Gwei` type for consistency. (Also simplifies code.)
2) Rename to `validator.high_balance` and `state.low_balances`.
3) Minor code cleanups.

```python
[...]

Issue #727, opened 2019-03-07
Prevent balance underflow 


hwwhww:
### Issue
- There are several places that we deduct the `state.validator_balances[validator_index]` for penalties.
- Logically, the validator would be ejected before the balance field is under the risk of underflow.
- But, when we read the function separately, do we need to add some protection to prevent balance underflow? For example, change
```python
state.validator_balances[index] -= penalty
```
to
```python
state.validator_balances[index] = max(state.validator_balances[index] - penalty, 0)
```

vbuterin:
One way to fix this generally would be to define `decrement_balance` in that way in #728.

JustinDrake:
`decrement_balance` implemented 🎉

PR #711, opened 2019-03-01
Pythonize epoch transitions

vbuterin:
This is first and foremost a refactor of the epoch transition to define it entirely using python code.

To make the code executable, a number of substantive changes were also included. See below for a detailed change log of such changes.

## Substantive changes 
* Merkleization friendly pending attestations (#697)
    * Motivation
        * cleaner selection of attestations from previous and current epoch
        * merkleization friendly when cleaning up attestations at end of epoch transition
    * Changelog
        * remove `latest_attestations` from `BeaconState`
        * add `previous_epoch_attestations` and `current_epoch_attestations` to `BeaconState`
        * when processing attestations from block, append to previous/current epoch attestations array depending on if `attestation.data.slot` is from previous or current epoch
        * at end of epoch transition clean up attestations via the following:
            * Set `state.previous_epoch_attestations = state.current_epoch_attestations`.
            * Set `state.current_epoch_attestations = []`.
* parent root reconstruction + granular state roots + body/header segregation (#649)
    * Motivation
        * We want state transition function signature to look like `(state, block) -> state` but there was an implicit third parameter -- `parent_root`
        * This makes a number of small changes allowing us to cache the previous block header in state to make the state "self sufficient" in the state transition
        * To this end, we added `latest_state_roots` to state which is also a nice historic feature in general.
    * Changelog
        * replace `Proposal` with `BeaconBlockHeader`. Remove `shard` field.
            * note that `hash_tree_root(block) == hash_tree_root(block_header)`
        * change `parent_root` to `previous_block_root` in `BeaconBlock`
        * move `eth1_data` and `randao_reveal` into `BeaconBlockBody`
        * add `latest_state_roots` to `BeaconState`
        * change `batched_block_roots` to `historical_roots` and use this to batch `(block, state)` rather than just `block`. Do batching at end of epoch transition.
        * add `get_temporary_block_header` helper that produces a block header from a block with a `ZERO_HASH` for state root
        * modify `get_genesis_beacon_state` to handle modified state fields
        * per-slot processing
            * Do bookkeeping related to state, block, and latest_block_header
        * per-block processing
            * add "block header section" to handle all block header verification/bookkeeping
* `get_previous_epoch` returns `current_epoch - 1` in all cases
* tie break for `winning_root` favors lexicographically _higher_ hashes rather than lower
[...]

PR #728, opened 2019-03-07
High/low balance separation

vbuterin:
See #685 for reasoning

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -64,6 +64,10 @@
         - [`get_epoch_start_slot`](#get_epoch_start_slot)
         - [`is_active_validator`](#is_active_validator)
         - [`get_active_validator_indices`](#get_active_validator_indices)
+        - [`get_balance`](#get_balance)
+        - [`set_balance`](#set_balance)
+        - [`increase_balance`](#increase_balance)
+        - [`decrease_balance`](#decrease_balance)
         - [`get_permuted_index`](#get_permuted_index)
         - [`split`](#split)
         - [`get_epoch_committee_count`](#get_epoch_committee_count)
@@ -205,10 +209,10 @@ Code snippets appearing in `this style` are to be interpreted as Python code.
 
 | Name | Value | Unit |
 | - | - | :-: |
-| `MIN_DEPOSIT_AMOUNT` | `2**0 * 10**9` (= 1,000,000,000) | Gwei |
+| `MIN_DEPOSIT_AMOUNT` | `2**0 * 10**9` (= 2,000,000,000) | Gwei |
 | `MAX_DEPOSIT_AMOUNT` | `2**5 * 10**9` (= 32,000,000,000) | Gwei |
-| `FORK_CHOICE_BALANCE_INCREMENT` | `2**0 * 10**9` (= 1,000,000,000) | Gwei |
 | `EJECTION_BALANCE` | `2**4 * 10**9` (= 16,000,000,000) | Gwei |
+| `HIGH_BALANCE_INCREMENT` | `10 ** 9` (= 1,000,000,000) | Gwei |
 
 ### Initial values
 
@@ -516,7 +520,7 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
 
     # Validator registry
     'validator_registry': [Validator],
-    'validator_balances': ['uint64'],
+    'low_balances': ['uint32'],
     'validator_registry_update_epoch': 'uint64',
 
     # Randomness and committees
@@ -570,6 +574,8 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
     'initiated_exit': 'bool',
     # Was the validator slashed
     'slashed': 'bool',
+    # Rounded balance
+    'high_balance': 'uint32'
 }
 ```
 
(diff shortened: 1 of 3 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -570,6 +574,8 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
     'initiated_exit': 'bool',
     # Was the validator slashed
     'slashed': 'bool',
+    # Rounded balance
+    'high_balance': 'uint32'
```
JustinDrake: Can we make `high_balance` of type `Gwei`? Benefits:

* It avoids introducing a new semantic type (`ETH`).
* It simplifies `get_balance` and `set_balance` (see [here](https://github.com/ethereum/eth2.0-specs/issues/685#issuecomment-467025467)).
* It is slightly more future-proof.

While the 2x performance benefit for `hash_tree_root(state.low_balances)` (saving 32MB of hashing per epoch in the worst case) is nice, I'm also tempted to make `low_balances` of type `Gwei` for simplicity:

* It avoids introducing a new SSZ type (`uint32`).
* It avoids introducing a new semantic type (`Uint32Gwei`).
* We still get the significant light client optimisation.

specs/core/0_beacon-chain.md
```diff
@@ -749,6 +755,45 @@ def get_active_validator_indices(validators: List[Validator], epoch: Epoch) -> L
...
+````python
+def increase_balance(state: BeaconState, index: int, delta: int) -> None:
+    set_balance(state, index, get_balance(state, index) + delta)
+````
+
+#### `decrease_balance`
+
+````python
+def decrease_balance(state: BeaconState, index: int, delta: int) -> None:
+    set_balance(state, index, get_balance(state, index) - delta)
```
JustinDrake: Consider checking for underflows here.

specs/core/0_beacon-chain.md
```diff
@@ -205,10 +209,10 @@ Code snippets appearing in `this style` are to be interpreted as Python code.
 
 | Name | Value | Unit |
 | - | - | :-: |
-| `MIN_DEPOSIT_AMOUNT` | `2**0 * 10**9` (= 1,000,000,000) | Gwei |
+| `MIN_DEPOSIT_AMOUNT` | `2**0 * 10**9` (= 2,000,000,000) | Gwei |
```
JustinDrake: `2**0 * 10**9` doesn't match `2,000,000,000`.

specs/core/0_beacon-chain.md
vbuterin: The benefits of halving the validator registry recomputation time are large. If we can derive from protolambda's shuffling statistics that it takes 1.25s to hash 45 bits * 4 million validators, then cutting low balances to 32 bits would cut the maximum validator balance re-hashing time from 1.25 * 128/45 = 3.55 seconds to 1.25 * 64/45 = 1.78 seconds, which is potentially the difference between the chain having a hiccup every six minutes and not.

> It avoids introducing a new SSZ type (uint32).

SSZ already has a `unit32` type, and it's a type we'll need anyway if/when we end up [doing CBC](https://github.com/ethereum/eth2.0-specs/issues/701), for example.

I'd be ok with the high balance being a uint64. But note that it should not be of type Gwei because it's not counting Gwei, it's counting multiples of a billion Gwei. This is the same reasoning as why you would not use the Epoch type to refer to, say, proof of custody periods. Perhaps we should just call the types of the variables in the state uint32 and/or uint64, and use the Gwei type only for numbers that get passed into or out of `get_balance` and `set_balance` calls.

specs/core/0_beacon-chain.md
vbuterin: Suggested the same in #727 :smile:

specs/core/0_beacon-chain.md
JustinDrake: > cutting low balances to 32 bits would cut the maximum validator balance re-hashing time from 1.25 * 128/45 = 3.55 seconds to 1.25 * 64/45 = 1.78 seconds

Shuffling is somewhat different than Merkleisation. I just did a quick benchmark in Go where `hash_tree_root(state.low_balances)` takes 544ms for 2^22 validators when using `uint64`. So going down to `uint32` would at most shave 272ms per epoch. (My benchmark is single-thread and not necessarily optimised.)

> it's not counting Gwei

Right. My suggestion is to do the accounting in Gwei (not multiples of a billion Gwei). This simplifies both the typing and the `get_balance `/`set_balance` code.

JustinDrake: Just to provide a bit more context here, shuffling does 37-byte input hashes which take 300ns and merkleisation does 64-byte input hashes which take 500ns.

Another consideration is that in the current 1.25s shuffling the hash function consumes at most 25% of the total time. We discussed a possible optimisation with @protolambda to improve the 1.25s number.

djrtwo:
switched to be based on `dev` since just merged in #711

specs/core/0_beacon-chain.md
vbuterin: > We discussed a possible optimisation with @protolambda to improve the 1.25s number.

Have you done the optimization that requires you to only hash half the whole range of indices every round already? (This is the one that allows shuffling to be computed with 45 bits of hashing per validator instead of 90)

BTW I ran it in python and you are right, going down from 64 to 32 shaves only 350ms (in python as usual, hashing is artificially negligible relative to everything else because it's one of the few things that's C-optimized; the actual reduction was from 4150 ms to 3700).

Would it satisfy your uint64 purism bone to have the value be specified as bytes4 instead of uint32? I see the "low balance" as not being a real standalone "number" but rather a backend thing whose role is abstracted away by the `get_balance` and `set_balance` functions.

JustinDrake: > Have you done the optimization that requires you to only hash half the whole range of indices every round already?

I believe that optimisation was done from the start. More advanced optimisations are being studied. We're somewhat hopeful we can get shuffling of 4,000,000 validators closer to 0.5s. cc @protolambda

> going down from 64 to 32 shaves only 350ms

One of the things we are looking into is drastically reducing that number. Merkleisation is extremely parallelisable so the data to be hashed can be pipelined to the CPU to max out the bandwidth of the SHA256 instructions. (Further gains can be had with things like AVX2 and AVX512.)

One good reason to optimise data structures for hashing is in preparation for a STARK-friendly hash function, so I'm definitely weakening my stance that this is a micro-optimisation.

> Would it satisfy your uint64 purism bone to have the value be specified as bytes4 instead of uint32?

Interesting! Unfortunately the Merkleisation spec only does packing on basic types (`byte[4]` is a tuple). My purism bone is merely a "strong opinion weakly held" to guide intuition. It should be jettisoned if necessary.

Do we at least agree that `high_balance` should be `Gwei`? 😄

vbuterin: Sure, I'm fine with `high_balance` being Gwei. If we do end up not wanting to use a uint32 for low balance, then I think we should simplify and just store the entire balance in the low balance field, so balance reads are unchanged and it's only writes that have an extra update step.

PR #732, opened 2019-03-07
epoch transition at start of epoch

djrtwo:
Having epoch transitions during the last slot of the epoch leaves the state transitioned halfway between the two epochs. Fields related to shuffling, seed, justification, etc are updated for the following epoch but the `state.slot` still is set for the current epoch. This disparity between parts of the state splitting across epochs creates a number of off-by-one issues throughout helpers and verification.

For example, the attestation off by one issue that was fixed with a kludge was due to this disparity (https://github.com/ethereum/eth2.0-specs/pull/627).

A more important issue that Prysmatic has run into is that the shuffling assignments break when in this in between state. At the end of an epoch, say slot 63, all of the fields related to shuffling (`current_shuffling_seed`, `current_shuffling_start_shard`, etc) have been updated to reflect values for epoch starting at 64, but the state is still at slot 63. If in this state and you call something like `get_crosslink_committees_for_slot` for slot 64, the helper thinks you are requesting for the `next_epoch` and regenerates incorrect seeds.

## changelog

* Move epoch transition before slot transition and handle at the _start_ of each rather than at the end of each epoch.
* Create new subsection of state transition called "state-root caching". This is done before everything including epoch & slot transitions.
* Collapse state root verification section into the end of the block processing section
* remove attestation kludge https://github.com/ethereum/eth2.0-specs/pull/627

_note: the diff looks nasty but this is primarily just a reordering of components of the state transition function. removing the attestation kludge (+ 1) is the only non-ordering change_

For clarity, the state transition now looks like this:
https://github.com/ethereum/research/blob/d15e5c78a9a296c2ebe641034c3a3998e9a9c87f/spec_pythonizer/state_transition.py#L75-L84

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -115,9 +115,20 @@
     - [Beacon chain processing](#beacon-chain-processing)
         - [Beacon chain fork choice rule](#beacon-chain-fork-choice-rule)
     - [Beacon chain state transition function](#beacon-chain-state-transition-function)
+        - [Per-epoch processing](#per-epoch-processing)
+            - [Helper functions](#helper-functions-1)
+            - [Justification](#justification)
+            - [Crosslinks](#crosslinks)
+            - [Eth1 data](#eth1-data-1)
+            - [Rewards and penalties](#rewards-and-penalties)
+                - [Justification and finalization](#justification-and-finalization)
+                - [Crosslinks](#crosslinks-1)
+            - [Apply rewards](#apply-rewards)
+            - [Ejections](#ejections)
+            - [Validator registry and shuffling seed data](#validator-registry-and-shuffling-seed-data)
+            - [Slashings and exit queue](#slashings-and-exit-queue)
+            - [Final updates](#final-updates)
         - [Per-slot processing](#per-slot-processing)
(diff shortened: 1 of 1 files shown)


terenc3t:
Can we get some background on why we need this change?

JustinDrake:
Minor comments:

* The sections "Block header" and "State root verification" can now be merged. (If merged, the "Block header" section would go at the very end.)
* `get_state_root` is used in exactly one place. It's also a pain because the `assert` is preventing us from doing `state.slot += 1` at the end of `advance_slot`, simplifying two `state.slot - 1` with `state.slot`. I'm tempted to nuke `get_state_root` and do the various simplifications.

djrtwo:
The get helpers are particularly useful being defined for validators/clients outside of the state transition.

We can just do the array access directly within the state transition here knowing that it is safe in this context if you really want to move these lines around.

djrtwo:
> The sections "Block header" and "State root verification" can now be merged. (If merged, the "Block header" section would go at the very end.)

Seems like 'eth1 data', 'randao', and 'block header' can be processed in parallel to all of the beacon transactions with just a final state root check at the end. I think if we merge the header and state root sections, we might encourage missing this parallelization opportunity.

PR #734, opened 2019-03-08
Persistent committee size per slot reduced to max 128

vbuterin:
Cuts down the cost of verifying a shard chain and aggregating signatures for a shard chain, and also makes the shard chain signatures more usable by light clients for verification as they would only need to keep track of a max 256-sized committee.


specs/core/1_shard-data-chains.md
```diff
@@ -147,15 +168,24 @@ def get_shuffled_committee(state: BeaconState,
 ```python
 def get_persistent_committee(state: BeaconState,
                              shard: Shard,
-                             epoch: Epoch) -> List[ValidatorIndex]:
+                             slot: Slot) -> List[ValidatorIndex]:
```
hwwhww: Should we add an assertion for the domain range of the given`slot`?

specs/core/1_shard-data-chains.md
```diff
@@ -147,15 +168,24 @@ def get_shuffled_committee(state: BeaconState,
...
+        
+    earlier_start_epoch = epoch - (epoch % PERSISTENT_COMMITTEE_PERIOD) - PERSISTENT_COMMITTEE_PERIOD * 2
+    later_start_epoch = epoch - (epoch % PERSISTENT_COMMITTEE_PERIOD) - PERSISTENT_COMMITTEE_PERIOD
+
+    committee_count = max(
+        len(get_active_validator_indices(state.validator_registry, earlier_start_epoch)) //
+        (SHARD_COUNT * TARGET_COMMITTEE_SIZE),
+        len(get_active_validator_indices(state.validator_registry, later_start_epoch)) //
+        (SHARD_COUNT * TARGET_COMMITTEE_SIZE),
+    ) + 1
```
hwwhww: Could you explain why "committee_count adds 1 instead of maxing 1" commit? Or what you wanted was a `ceil` function here?

vbuterin: The goal is to have a committee size which is always not greater than `TARGET_COMMITTEE_SIZE`; `max(1, ...)` does not do that, ,`1 + ...` does. So yeah, ceil would work too, but it doesn't make a big difference either way.

specs/core/1_shard-data-chains.md
vbuterin: Technically we don't need to because `generate_seed` and specifically `get_randao_mix` and `get_active_index_root` inside there are doing the check. I suppose we could add a comment saying how far back it can check.

PR #735, opened 2019-03-08
Proposer rewards increased and proportional to balance

vbuterin:


djrtwo:
Weren't we concerned about this increasing above a the attestation inclusion reward (1/5 vs 1/4)?
I can't remember the exact reasoning..

vbuterin:
I don't think we were...

djrtwo:
oh, I remember. We were concerned about proposer reward being too high such that it incentivizes trying to not build on the head and instead bypass a block or two to gather up the reward by including the attestations in those (potentially) orphaned blocks

vbuterin:
Not building on the head is very risky in the current setup because the head would dominate with many attestations very quickly, but I definitely see the rationale in not encouraging trying.

BTW the need for this patch is arguably removed by the fair proposer selection patch. So if we feel the 1/8 reward is enough we could just close this?

JustinDrake:
Closing in favour of #781.

Issue #675, opened 2019-02-22
Miscellaneous beacon chain changes—take 4


JustinDrake:
(See #128, #218, #322 for takes 1, 2, 3.)

Below is a list of miscellaneous suggestions for phase 0, most of which were discussed on the researcher's call on Feb 19. This issue keeps track of some of the phase 0 work remaining.

- [x] 1. **Friendlier GENESIS_SLOT**: Implemented in #655.
- [x] 2. **Granular state roots**: Expose state roots at every slot. Implemented in #649.
- [x] 3. **Previous block root reconstruction**: Provide enough information in `state` to reconstruct the current block's `previous_block_root`. Implemented in #649.
- [x] 4. **Define genesis Eth1 data**: Implemented in #649.
- [x] 5. **Mandatory deposits**: Mandatory processing of pending deposits.
- [x] 6. **Transfers during pre-activation**: Allow not-yet-activated validators to make transfers.
- [x] 7. **LMD GHOST tie breaker**: Compare block hashes to tie-break LMD GHOST.
- [ ] 8. **Maximum reversions**: Enshrine dynamic weak subjectivity revert period. See #577.
- [x] 9. **Double justifications**: Specify fork choice rule when there are two justified blocks at the same height. (Possible solution: ignore both and fallback to the previous highest justified block.)
- [x] 10. **Move to SHA256**: See #612.
- [ ] 11. **Standardise BLS12-381**: See #605.
- [ ] 12. **Performance parameters tuning**: Fine-tune `SECONDS_PER_SLOT`, `SHARD_COUNT`, etc. based on benchmarks.
- [ ] 13a. **Increase proposer rewards**: See #621. Need to check incentive compatibility with inclusion distance reward.
- [x] 13b. **Incentive-compatible proposer rewards**: Make proposer rewards proportional to balance. 
- [ ] 14. **Increase rewards in general**: Calculate theoretical max issuance rate and work backwards to expected issuance.
- [x] 15. **Reduce SSZ_CHUNK_SIZE to 32**: See #603 and #696.
- [x] 16. **SSZ tuples**: See #665 and #696.
- [x] <s>17. **Immediately withdrawable if bad proof of possession**: See #657.</s>
- [x] 18. **4-byte working balance**: See #685.
- [x] 19. **Merkleisation-friendly pending attestations**: See #697.
- [ ] 20. **Fine-tune container field ordering**: To do with constants fine-tuning.
[...]

PR #736, opened 2019-03-08
Cannot exit until 2048 epochs (#675 item 21)

vbuterin:
See #675 item 21


djrtwo:
This is supposed to "(i) ensure stability of persistent committees" but there are other ways a validator might be exited -- slashing and balance ejections.  Not sure if the balance ejection could happen in the time frame, but a slashing certainly could.

Wanted to make sure the intention is to just provide more stability but not to prevent quick exits all together.

Also, we should probably just rebuild this PR from scratch against current `dev`. The merge conflict is nasty for such a simple PR

djrtwo:
Closing in favor of #746

Issue #706, opened 2019-02-28
Use all 64 bits in fork_version


JustinDrake:
This was brought up by @protolambda. We are inconsistent on the handling of `fork_version`:

* We define it as a `uint64`.
* We use it as a `uint32` when doing `fork_version * 2**32 + domain_type`.

I think it would be cleaner if the final signing logic used `hash(message_hash + bytes8(fork_version) + byte1(domain))` as opposed to `hash(message_hash + bytes8(domain))`.

djrtwo:
Defined `fork_version` as a `uint64` for consistency of 64-bit types everywhere, but clearly only intend to use it up to 32bits.

To be clear, you are suggesting changing the bls functions to accept `fork_version, domain` rather than just `domain`?

JustinDrake:
> you are suggesting changing the bls functions to accept `fork_version, domain` rather than just `domain`?

Right, but ideally there's something cleaner.

JustinDrake:
A clean solution (idea taken from @vbuterin) is to define `fork_version` as a `byte[4]`. It also makes more sense semantically because we're not doing no arithmetic on it. We can also define `domain_type` as a `byte[4]`, which again makes more sense semantically. We then replace `return fork_version * 2**32 + domain_type` with `return fork_version + domain_type`.

PR #738, opened 2019-03-08
Semantic fork versions and signature domains

JustinDrake:
Address #706

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1517,8 +1517,8 @@ def get_genesis_beacon_state(genesis_validator_deposits: List[Deposit],
         slot=GENESIS_SLOT,
         genesis_time=genesis_time,
         fork=Fork(
-            previous_version=GENESIS_FORK_VERSION,
-            current_version=GENESIS_FORK_VERSION,
+            previous_version=int_to_bytes4(GENESIS_FORK_VERSION),
+            current_version=int_to_bytes4(GENESIS_FORK_VERSION),
             epoch=GENESIS_EPOCH,
         ),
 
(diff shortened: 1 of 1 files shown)


hwwhww:
It requires BLS signature spec API updates and I think we agreed not changing that before the standardized spec?

JustinDrake:
@hwwhww No changes to BLS spec required now 👍

specs/core/0_beacon-chain.md
```diff
@@ -1107,7 +1107,7 @@ def get_total_balance(state: BeaconState, validators: List[ValidatorIndex]) -> G
 
 ```python
 def get_fork_version(fork: Fork,
-                     epoch: Epoch) -> int:
+                     epoch: Epoch) -> bytes4:
```
hwwhww: Ah sorry, one sec, we haven't defined `Bytes4` custom type yet, so either changed this line to
```suggestion
                     epoch: Epoch) -> bytes:
```
Or defined `Bytes4` and then here we can do:
```suggestion
                     epoch: Epoch) -> Bytes4:
```

What do you think about it? :)

specs/core/0_beacon-chain.md
```diff
@@ -1122,12 +1122,11 @@ def get_fork_version(fork: Fork,
 ```python
 def get_domain(fork: Fork,
                epoch: Epoch,
-               domain_type: int) -> int:
+               domain_type: bytes4) -> int:
```
hwwhww: ditto, either `domain_type: bytes` or `domain_type: Bytes4`.

specs/core/0_beacon-chain.md
JustinDrake: `bytesN` (hence `bytes4`) is defined [here](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/simple-serialize.md#aliases). Defining `Bytes4` also is fine I guess.

djrtwo: `bytes4` isn't valid python outside of the string context unless we define a new type in custom types. I think I'd rather just leave this as `bytes`

JustinDrake: Done! Also makes the diff super small.

PR #739, opened 2019-03-08
Use hash_tree_root everywhere

JustinDrake:
And get rid of merkle_root. This is possible because of SSZ tuples.


vbuterin:
I don't see how this can be valid python. `hash_tree_root` for non-SSZ-containers doesn't distinguish between fixed size and variable size lists, but the spec does. Maybe add an optional `typ` argument to `hash_tree_root` ?

JustinDrake:
I defined a new SSZ object :)

djrtwo:
I'll make sure this runs in the executable spec before merging

djrtwo:
lol found a couple of other bugs when running the state transition through many empty epochs! patching a fix in another PR

PR #740, opened 2019-03-08
Fix #710

### LINKED ISSUE #710: Deposit contract needs to check length of deposit data
The deposit contract needs to actually verify that the deposit data is **exactly** 192 bytes long (that's `len(ssz_serialize(deposit_data))`), 192 = 8 (amount) + 8 (timestamp) + 48 (pubkey) + 32 (withdrawal credentials) + 96 (proof of possession).

The reason is that if a deposit data is submitted that is NOT 192 bytes, it is impossible to create a DepositData that serializes to the hash put into the Merkle branch, which could be used to prevent progress of deposit processing.

JustinDrake:


djrtwo:
This actually needs to happen on deposit contract side. It is not use to throw once you get over here.

JustinDrake:
🤦‍♂️

PR #743, opened 2019-03-08
Minor cleanups

JustinDrake:


djrtwo:
eh, I find object equality a little bit opaque but I suppose it's fine...
We are doing it in process slashings too -- `assert attestation1.data != attestation2.data`

PR #745, opened 2019-03-08
Fix table of contents links

terenc3t:


djrtwo:
Thanks! but was just addressed in #743

PR #746, opened 2019-03-08
Cannot exit until 2048 epochs (#675 item 21) [redo]

djrtwo:
See #675 item 21

(Supersedes #736)


djrtwo:
[transfered comment from #736]

This is supposed to "(i) ensure stability of persistent committees" but there are other ways a validator might be exited -- slashing and balance ejections. Not sure if the balance ejection could happen in the time frame, but a slashing certainly could.

Wanted to make sure the intention is to just provide more stability but not to prevent quick exits all together.

Also, we should probably just rebuild this PR from scratch against current dev. The merge conflict is nasty for such a simple PR

vbuterin:
> Wanted to make sure the intention is to just provide more stability but not to prevent quick exits all together.

Correct.

Moonshieldgru:
**Export-50edd684-148c-4110-a6a2-5e6672c368cd.zip

**

Moonshieldgru (approved):

PR #750, opened 2019-03-10
Do not check withdrawal credentials for existing validators

JustinDrake:
We should not invalidate blocks that contain a deposit with an inconsistent withdrawal credential as that would stall the chain.


djrtwo:
Should we do PoP check if the pubkey already exists in the beacon chain?

JustinDrake:
> Should we do PoP check if the pubkey already exists in the beacon chain?

Good point!

Issue #742, opened 2019-03-08
Store prev/current justified root


djrtwo:
# issue

Attestation processing assumes that the previous/current block root will be available in `state.latest_block_roots` when processing attestations. 

```python
attestation.data.justified_block_root == get_block_root(
        state, get_epoch_start_slot(attestation.data.justified_epoch)
)
```

The call to `get_block_root` fails if more than `SLOTS_PER_HISTORICAL_ROOT` have passed since latest justified block. This is currently only about ~13 hours which requires liveness wrt justification in this period for the chain to continue to function. Oh no!

## proposed solution

Add `previous_justified_root` and `current_justified_root` directly to state to eliminate this lookup to a potentially unavailable slot.

(Thanks @arnetheduck for running your non-justifying chain for 8192 slots!)

arnetheduck:
It might not necessarily have to be part of the state (duplicate info, kind of).. we can cache it somewhere in the client as well - we already keep a few other book-keeping fields like this around anyway.

djrtwo:
State transition should be self contained wrt `pre_state, block`.
I'm not currently aware of any other book keeping fields that are _required_ in the state transition (that said, I expect clients to cache things in various ways as they decide worthwhile).

arnetheduck:
fair enuff, I was still on a version that required the previous block.

arnetheduck:
actually, there's nuance here. it doesn't look like the fields are needed to *apply* a block state transition - they're needed to fill out a block when you're proposing one, no?

tdmllc1:
Deposit 500000000000000 in binanace account for account tdmllc1@outlook.com liquid trading capital NFA ID 0356535

tdmllc1:
> Deposit 500000000000000 in binanace account for account [tdmllc1@outlook.com](mailto:tdmllc1@outlook.com) liquid trading capital NFA ID 0356535

Thank You

tdmllc1:
Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

djrtwo:
> actually, there's nuance here. it doesn't look like the fields are needed to apply a block state transition - they're needed to fill out a block when you're proposing one, no?

@arnetheduck which fields? access to the root of the justified epoch is needed both when creating an attestation and when validating it.

tdmllc1:
"1518368749000" deposit tdmllc1@outlook.com USD

arnetheduck:
[...]

PR #756, opened 2019-03-11
cache justified roots in state/rename attestation fields

djrtwo:
addresses #742

Two things:

### correct value for genesis justified roots
Right now, we just have `ZERO_HASH` for the genesis justified roots. This does not actually reflect the value of the root of the block at genesis. The value of the block at the genesis slot is looks like 

```python
block = get_empty_block()
genesis_state = get_genesis_state()
block.state_root = hash_tree_root(genesis_state)
```

Due to the normal chicken-and-egg state/block-root problem, we can't embed the genesis block root in as the initial justified roots when constructing the genesis state. If we want to enforce this, we can do the following in `cache_state`. This is nice because the value reflects an actual block root but it's not really necessary imo.

```python
if state.justified_root == ZERO_HASH:
    state.justified_root = hash_tree_root(state.latest_block_header)
    state.previous_justified_root = state.justified_root
```

### finalized root
Should we add another field `finalized_root` for symmetry? It is not needed in the state transition but might be nice to be available in state regardless of the number of slots since finality. (if greater than `HISTORICAL_ROOTS_LENGTH` you would no longer have it available in state)

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -335,22 +335,19 @@ The types are defined topologically to aid in facilitating an executable version
 
 ```python
 {
-    # Slot number
+    # LMD GHOST vote
     'slot': 'uint64',
-    # Shard number
-    'shard': 'uint64',
-    # Root of the signed beacon block
     'beacon_block_root': 'bytes32',
-    # Root of the ancestor at the epoch boundary
-    'epoch_boundary_root': 'bytes32',
-    # Data from the shard since the last attestation
+
+    # FFG vote
+    'source_epoch': 'uint64',
+    'source_root': 'bytes32',
+    'target_root': 'bytes32',
+
+    # Crosslink vote
+    'shard': 'uint64',
+    'previous_crosslink': Crosslink,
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1830,7 +1834,9 @@ def update_justification_and_finalization(state: BeaconState) -> None:
 
     # Rotate justified epochs
     state.previous_justified_epoch = state.justified_epoch
+    state.previous_justified_root = state.justified_root
     state.justified_epoch = new_justified_epoch
+    state.justified_root = get_block_root(state, get_epoch_start_slot(new_justified_epoch))
```
arnetheduck: won't this fail if `state.justified_epoch` is more than 8k slots old and no new justification happens?

djrtwo: Right, should conditionally overwrite

JustinDrake:
> Should we add another field `finalized_root`

Good idea! At the very least we want an EVM2.0 opcode to fetch the finalized root.

PR #762, opened 2019-03-12
Epoch-based proposer slashing

JustinDrake:
See #675 item 25.


specs/core/0_beacon-chain.md
```diff
@@ -2306,7 +2306,7 @@ def process_proposer_slashing(state: BeaconState,
     """
     proposer = state.validator_registry[proposer_slashing.proposer_index]
     # Verify that the slot is the same
```
```suggestion
    # Verify that the epoch is the same
```

Issue #692, opened 2019-02-26
Phase 0 Wire Protocol


mslipper:
Follow-on to #593.

This specification describes Ethereum 2.0's networking wire protocol.

The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL", NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.

# Use of `libp2p`

This protocol uses the `libp2p` networking stack. `libp2p` provides a composable wrapper around common networking primitives, including:

1. Transport.
2. Encryption.
3. Stream multiplexing.

Clients MUST be compliant with the corresponding `libp2p` specification whenever `libp2p` specific protocols are mentioned. This document will link to those specifications when applicable.

# Client Identity

## Identification

Upon first startup, clients MUST generate an RSA key pair in order to identify the client on the network. The SHA-256 `multihash` of the public key is the clients's Peer ID, which is used to look up the client in `libp2p`'s peer book and allows a client's identity to remain constant across network changes.

## Addressing

Clients on the Ethereum 2.0 network are identified by`multiaddr`s. `multiaddr`s are self-describing network addresses that include the client's location on the network, the transport protocols it supports, and its peer ID. For example, the human-readable `multiaddr` for a client located at `example.com`, available via TCP on port `8080`, and with peer ID `QmUWmZnpZb6xFryNDeNU7KcJ1Af5oHy7fB9npU67sseEjR` would look like this:

```
/dns4/example.com/tcp/8080/p2p/QmUWmZnpZb6xFryNDeNU7KcJ1Af5oHy7fB9npU67sseEjR
```

We refer to the `/dns4/example.com` part as the 'lookup protocol', the `/tcp/8080` part as the networking protocol, and the `/p2p/<peer ID>` part as the 'identity protocol.' 

Clients MAY use either `dns4` or `ip4` lookup protocols. Clients MUST set the identity protocol to `/tcp` followed by a port of their choosing. It is RECOMMENDED to use the default port of `9000`. Clients MUST set the identity protocol to `/p2p/` followed by their peer ID.

## Relevant `libp2p` Specifications

- [multiaddr](https://github.com/multiformats/multiaddr)
- [multihash](https://multiformats.io/multihash/)
- [peer-id](https://github.com/libp2p/js-peer-id)

# Transport

[...]

PR #763, opened 2019-03-12
Phase 0 Networking Specifications

mslipper:
Supplants #692. Notable changes:

- Specs are split into messaging, identity, and RPC documents. I've left out broadcast for now so that we can discuss that separately.
- The RPC protocol uses something more akin to JSON-RPC on the wire in order to identify individual methods to call rather than using a separate `libp2p` protocol name for each. This fits better with `libp2p`'s stream selection logic.


specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+and their corresponding responses are wrapped in a "response" structure:
+
+```
+(
+    id: uint64
+    result: Response
+)
+```
+
+If an error occurs, a variant of the response structure is returned:
```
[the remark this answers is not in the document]
mslipper: Ah, this is a good point. SSZ doesn't support `null` values either - let me think on this one for a little bit and come up with a solution.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+Message body schemas are notated like this:
+
+```
+(
+    field_name_1: type
+    field_name_2: type
+)
+```
+
+SSZ serialization is field-order dependent. Therefore, fields MUST be  encoded and decoded according to the order described in this document. The encoded values of each field are concatenated to form the final encoded message body. Embedded structs are serialized as Containers unless otherwise noted.
```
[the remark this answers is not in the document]
mslipper: Do you think it's clear enough that the field order matters if we don't mention SSZ? Technically speaking the message version (in `messaging.md`) implies the encoding format, so you're right that this is a bit redundant.

Correct, "embedded structs" means containers inside other containers.

specs/networking/node-identification.md
```diff
@@ -0,0 +1,32 @@
+ETH 2.0 Networking Spec - Node Identification
```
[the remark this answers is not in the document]
mslipper: Cool, will remove.

specs/networking/messaging.md
```diff
@@ -0,0 +1,41 @@
...
+
+# Motivation
+
+This specification seeks to define a messaging protocol that is flexible enough to be changed easily as the ETH 2.0 specification evolves.
+
+# Specification
+
+## Message Structure
+
+An ETH 2.0 message consists of a single byte representing the message version followed by the encoded, potentially compressed body. We separate the message's version from the version included in the `libp2p` protocol path in order to allow encoding and compression schemes to be updated independently of the `libp2p` protocols themselves.
```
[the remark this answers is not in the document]
mslipper: This is similar to what Raul K. from IPFS was saying - I can update the spec to reflect that exchange. What do you think about upgradeability paths if we go this route? From what I understand, there's a possibility that hybrid encoding formats will be supported and from an implementation perspective it might be easier to simply read the first byte of the message to determine how messages will be serialized.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+    error: (
+        code: uint16
+        data: bytes
+    )
+)
+```
+
+The details of the RPC-Over-`libp2p` protocol are similar to [JSON-RPC 2.0](https://www.jsonrpc.org/specification). Specifically:
+
+1. The `id` member is REQUIRED.
```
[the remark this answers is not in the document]
mslipper: Was thinking about including these just in case SSZ/whatever encoding format is used in the future start supporting `null` values.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+    best_root: bytes32
+    best_slot: uint64
+)
+```
+
+Clients exchange `hello` messages upon connection, forming a two-phase handshake. The first message the initiating client sends MUST be the `hello` message. In response, the receiving client MUST respond with its own `hello` message.
+
+Clients SHOULD immediately disconnect from one another following the handshake above under the following conditions:
+
+1. If `network_id` belongs to a different chain, since the client definitionally cannot sync with this client.
```
atoulme: You should consider spelling out network ID and chain ID as separate fields. Chain ID should be set to a fixed number "1" for ETH, and if others want to run their own chain they can change that ID.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+       |   |  +-----------+ Epoch 2
+       +-+-+
+         ^
+         |
+       +-+-+
+       |   |  +-----------+ Epoch 1
+       +---+
+```
+
+Once the handshake completes, the client with the higher `latest_finalized_epoch` or `best_slot` (if the clients have equal `latest_finalized_epoch`s) SHOULD send beacon block roots to its counterparty via `beacon_block_roots` (i.e., RPC method `10`).
```
atoulme: This spec handles the initial sync between nodes. How are new blocks handled? Will they be pushed proactively across? Do you just push the block root?

specs/networking/rpc-interface.md
FrankSzendzielarz: NetworkId vs ChainId +1.
Also, message body compression algorithm indicator.
Also, upgrade paths for SSZ (I get the feeling this might change on the wire)..maybe a sorted list of serialization method preferences, the highest mutual being selected?

specs/networking/rpc-interface.md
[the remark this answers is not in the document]
jannikluhn: > Do you think it's clear enough that the field order matters if we don't mention SSZ?

If we'd replace SSZ with something that doesn't care about field order, the rest of the protocol wouldn't be affected at all. So yes, I think we don't need to mention in explicitly.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+
+**Response Body:**
+
+```
+(
+    headers: []BlockHeader
+)
+```
+
+Requests beacon block headers from the peer starting from `(start_root, start_slot)`. The response MUST contain no more than `max_headers` headers. `skip_slots` defines the maximum number of slots to skip between blocks. For example, requesting blocks starting at slots `2` a `skip_slots` value of `2` would return the blocks at `[2, 4, 6, 8, 10]`. In cases where a slot is undefined for a given slot number, the closest previous block MUST be returned. For example, if slot `4` were undefined in the previous example, the returned array would contain `[2, 3, 6, 8, 10]`. If slot three were further undefined, the array would contain `[2, 6, 8, 10]` - i.e., duplicate blocks MUST be collapsed.
```
jannikluhn: `skip_slots` sounds off-by-one to me: skipping 0 blocks should result in a continuous sequence.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+
+### Beacon Block Headers
+
+**Method ID:** `11`
+
+**Request Body**
+
+```
+(
+    start_root: HashTreeRoot
```
jannikluhn: Do we need the root? It seems redundant to me, except for the case of chain reorgs which shouldn't happen frequently at sync (and even then, it's probably better to get blocks from the current chain that we'll be able to use later, instead of getting outdated ones).

specs/networking/rpc-interface.md
jannikluhn:
```suggestion
Requests beacon block headers from the peer starting from `(start_root, start_slot)`. The response MUST contain no more than `max_headers` headers. `skip_slots` defines the maximum number of slots to skip between blocks. For example, requesting blocks starting at slots `2` a `skip_slots` value of `2` would return the blocks at `[2, 4, 6, 8, 10]`. In cases where a slot is empty for a given slot number, the closest previous block MUST be returned. For example, if slot `4` were empty in the previous example, the returned array would contain `[2, 3, 6, 8, 10]`. If slot three were further empty, the array would contain `[2, 6, 8, 10]` - i.e., duplicate blocks MUST be collapsed.
```

as technically the slots are defined, there's just no block

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+
+```
+(
+    headers: []BlockHeader
+)
+```
+
+Requests beacon block headers from the peer starting from `(start_root, start_slot)`. The response MUST contain no more than `max_headers` headers. `skip_slots` defines the maximum number of slots to skip between blocks. For example, requesting blocks starting at slots `2` a `skip_slots` value of `2` would return the blocks at `[2, 4, 6, 8, 10]`. In cases where a slot is undefined for a given slot number, the closest previous block MUST be returned. For example, if slot `4` were undefined in the previous example, the returned array would contain `[2, 3, 6, 8, 10]`. If slot three were further undefined, the array would contain `[2, 6, 8, 10]` - i.e., duplicate blocks MUST be collapsed.
+
+The function of the `skip_slots` parameter helps facilitate light client sync - for example, in [#459](https://github.com/ethereum/eth2.0-specs/issues/459) - and allows clients to balance the peers from whom they request headers. Client could, for instance, request every 10th block from a set of peers where each per has a different starting block in order to populate block data.
```
jannikluhn:
```suggestion
The function of the `skip_slots` parameter helps facilitate light client sync - for example, in [#459](https://github.com/ethereum/eth2.0-specs/issues/459) - and allows clients to balance the peers from whom they request headers. Clients could, for instance, request every 10th block from a set of peers where each per has a different starting block in order to populate block data.
```

specs/networking/node-identification.md
```diff
@@ -0,0 +1,32 @@
...
+
+Clients use Ethereum Node Records (as described in [EIP-778](http://eips.ethereum.org/EIPS/eip-778)) to discover one another. Each ENR includes, among other things, the following keys:
+
+- The node's IP.
+- The node's TCP port.
+- The node's public key.
+
+For clients to be addressable, their ENR responses MUST contain all of the above keys. Client MUST verify the signature of any received ENRs, and disconnect from peers whose ENR signatures are invalid. Each node's public key MUST be unique.
+
+The keys above are enough to construct a [multiaddr](https://github.com/multiformats/multiaddr) for use with the rest of the `libp2p` stack.
```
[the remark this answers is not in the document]
raulk: libp2p peer IDs are derived from the public key protobuf, which is just key type + bytes. Here's the spec: https://github.com/libp2p/specs/pull/100. Both SECIO and TLS 1.3 validate peer IDs against the pubkey, so following the spec is important or connections will fail.

specs/networking/node-identification.md
```diff
@@ -0,0 +1,32 @@
...
+
+It is RECOMMENDED that clients set their TCP port to the default of `9000`.
+
+## Peer ID Generation
+
+The `libp2p` networking stack identifies peers via a "peer ID." Simply put, a node's Peer ID is the SHA2-256 `multihash` of the node's public key. `go-libp2p-crypto` contains the canonical implementation of how to hash `secp256k1` keys for use as a peer ID.
```
raulk:
```suggestion
The `libp2p` networking stack identifies peers via a "peer ID." Simply put, a node's Peer ID is the SHA2-256 `multihash` of the node's public key struct (serialized in protobuf, refer to the [Peer ID spec](https://github.com/libp2p/specs/pull/100)). `go-libp2p-crypto` contains the canonical implementation of how to hash `secp256k1` keys for use as a peer ID.
```

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+)
+```
+
+SSZ serialization is field-order dependent. Therefore, fields MUST be  encoded and decoded according to the order described in this document. The encoded values of each field are concatenated to form the final encoded message body. Embedded structs are serialized as Containers unless otherwise noted.
+
+All referenced data structures can be found in the [0-beacon-chain](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#data-structures) specification.
+
+## `libp2p` Protocol Names
+
+A "Protocol Name" in `libp2p` parlance refers to a human-readable identifier `libp2p` uses in order to identify sub-protocols and stream messages of different types over the same connection. A client's supported protocol paths are negotiated by the `libp2p` stack at connection time; as such they are not part of individual message bodies.
```
raulk:
```suggestion
A "Protocol ID" in `libp2p` parlance refers to a human-readable identifier `libp2p` uses in order to identify sub-protocols and stream messages of different types over the same connection. Peers exchange supported protocol IDs via the `Identify` protocol upon connection. When opening a new stream, peers pin a particular protocol ID to it, and the stream remains contextualised thereafter. Since messages are sent inside a stream, they do not need to bear the protocol ID.
```

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+        code: uint16
+        data: bytes
+    )
+)
+```
+
+The details of the RPC-Over-`libp2p` protocol are similar to [JSON-RPC 2.0](https://www.jsonrpc.org/specification). Specifically:
+
+1. The `id` member is REQUIRED.
+2. The `id` member in the response MUST be the same as the value of the `id` in the request.
```
raulk: Ideally we'd define the properties of this `id`. Is it monotonically increasing? Unique within a session?

specs/networking/rpc-interface.md
atoulme: What should they do?

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+
+```
+(
+    id: uint64
+    method_id: uint16
+    body: Request
+)
+```
+
+and their corresponding responses are wrapped in a "response" structure:
```
atoulme: Is it always a response, or would you send header and bodies unprompted?

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+### Provide Beacon Block Roots
+
+**Method ID:** `10`
+
+**Body:**
+
+```
+# BlockRootSlot
+(
+    block_root: HashTreeRoot
```
atoulme: Is a `HashTreeRoot` a `Bytes32`?

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+(
+    block_roots: []HashTreeRoot
+)
+```
+
+**Response Body:**
+
+```
+(
+    block_bodies: []BeaconBlockBody
```
atoulme: Sorry - is there a link to what a BeaconBlockBody is?

specs/networking/rpc-interface.md
djrtwo: Yes, should be updated to that or defined in this doc

specs/networking/rpc-interface.md
djrtwo: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#beaconblockbody

specs/networking/rpc-interface.md
mslipper: According to the JSON-RPC spec it need only be unique within a session. I'll update the doc to reflect this requirement and provide some `RECOMMENDED` options.

Having an `id` parameter like this is important since it allows RPC messages to be pipelined.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+    best_slot: uint64
+)
+```
+
+Clients exchange `hello` messages upon connection, forming a two-phase handshake. The first message the initiating client sends MUST be the `hello` message. In response, the receiving client MUST respond with its own `hello` message.
+
+Clients SHOULD immediately disconnect from one another following the handshake above under the following conditions:
+
+1. If `network_id` belongs to a different chain, since the client definitionally cannot sync with this client.
+2. If the `latest_finalized_root` shared by the peer is not in the client's chain at the expected epoch. For example, if Peer 1 in the diagram below has `(root, epoch)` of `(A, 5)` and Peer 2 has `(B, 3)`, Peer 1 would disconnect because it knows that `B` is not the root in their chain at epoch 3:
```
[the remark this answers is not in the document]
mslipper: Cool got it, will do.

specs/networking/rpc-interface.md
mslipper: I think this makes more sense from an API perspective. Will fix.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+    max_headers: uint64
+    skip_slots: uint64
+)
+```
+
+**Response Body:**
+
+```
+(
+    headers: []BlockHeader
```
atoulme: I note BlockHeader is not defined in the beacon chain spec. I opened a PR to define it as a struct.

specs/networking/messaging.md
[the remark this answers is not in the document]
AgeManning: - Should the encoded message also be length-prefixed? 
SSZ doesn't do this out of the box, and this will allow implementations to know incoming message sizes and therefore how much to read or reject excessive sized messages (potential DOS). 
- Should there be a size limit to messages?

specs/networking/rpc-interface.md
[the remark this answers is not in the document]
AgeManning: With SSZ we need the schema to decode. If I get a string of bytes, I can assume an `id` in the initial bytes of the message. How do I determine if the packet is a request/response in order to decode the rest of the message?

specs/networking/rpc-interface.md
jannikluhn: Send beacon block roots without an explicit request. I think RPC communication should always be request response pairs, never just a response.

specs/networking/rpc-interface.md
jannikluhn: If it's not needed specifically for the spec, we could also just define it here.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+
+All referenced data structures can be found in the [0-beacon-chain](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#data-structures) specification.
+
+## `libp2p` Protocol Names
+
+A "Protocol ID" in `libp2p` parlance refers to a human-readable identifier `libp2p` uses in order to identify sub-protocols and stream messages of different types over the same connection. Peers exchange supported protocol IDs via the `Identify` protocol upon connection. When opening a new stream, peers pin a particular protocol ID to it, and the stream remains contextualised thereafter. Since messages are sent inside a stream, they do not need to bear the protocol ID.
+
+## RPC-Over-`libp2p`
+
+To facilitate RPC-over-`libp2p`, a single protocol path is used: `/eth/serenity/rpc/1.0.0`. Remote method calls are wrapped in a "request" structure:
```
arnetheduck: what's the semantic meaning of these long version numbers?

specs/networking/node-identification.md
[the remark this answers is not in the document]
djrtwo: cc: @fjl

specs/networking/rpc-interface.md
notasecret: i would imagine it's there because over time it will be bugfixed (bugfix version), updated with a commitment to being backward compatible (minor version), and updated with complete disregard for any backward compatibility, for the sake of progress (major version)

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+2. The `id` member in the response MUST be the same as the value of the `id` in the request.
+3. The `method_id` member is REQUIRED.
+4. The `result` member is required on success, and MUST NOT exist if there was an error.
+5. The `error` member is REQUIRED on errors, and MUST NOT exist if there wasn't an error.
+
+Structuring RPC requests in this manner allows multiple calls and responses to be multiplexed over the same stream without switching.
+
+The "method ID" fields in the below messages refer to the `method` field in the request structure above.
+
+The first 1,000 values in `error.code` are reserved for system use. The following error codes are predefined:
```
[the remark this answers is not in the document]
notasecret: sort of like port numbers :)

specs/networking/messaging.md
```diff
@@ -0,0 +1,41 @@
...
++--------------------------+
+```
+
+Clients MUST ignore messages with mal-formed bodies. The `version` byte MUST be one of the below values:
+
+## Version Byte Values
+
+### `0x01`
+
+- **Encoding Scheme:** SSZ
```
notasecret: why not just drop in a standard binary serializer, like protobuf or captainproto?

specs/networking/rpc-interface.md
mslipper: Yes, the idea is that they follow semver. In practice I'd estimate that the only time we'd change these version numbers is if there was a backwards-incompatible change to the serialization/compression scheme.

See @raulk's point above re: message envelopes.

specs/networking/messaging.md
mslipper: Protobuf/captainproto were decided against due to their lack of support for various Ethereum-specific data structures, namely fixed-length arrays and Merkle roots.

specs/networking/messaging.md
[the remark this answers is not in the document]
mslipper: `libp2p` length-prefixes the message automatically.

I do think there should be a size limit.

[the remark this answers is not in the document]
AgeManning: I'm not sure about go-libp2p, but I've implemented a basic implementation of this in rust-libp2p. I had to specify length prefixes and size limits (to know when to stop reading from the stream). This may be the case for other implementations, I didn't see length prefixes in the libp2p specs.

mslipper:
We have a call tomorrow to go over the wire protocol, after which I'll update this PR with the group's decision. Ping me on Gitter if you'd like an invite to the call.

specs/networking/messaging.md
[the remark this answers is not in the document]
mslipper: It's defined in here, I believe: https://github.com/multiformats/multistream-select.

specs/networking/rpc-interface.md
mslipper: I'm in agreement with @jannikluhn here - this should be the result of an RPC call. I'll update to fix.

specs/networking/rpc-interface.md
mslipper: See below comment re: making a request portion for this.

specs/networking/rpc-interface.md
mslipper: There will be a separate spec for gossip.

specs/networking/rpc-interface.md
arnetheduck: A version number can either be interpreted or not. If we rely on semver, we should specify the correct behavior for clients: consider client a that supports 1.0.0 - should it also accept 1.0.1 messages as valid automatically, or discard them? This matters for forwards compatibility.

specs/networking/rpc-interface.md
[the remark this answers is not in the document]
mslipper: Added an `is_error` boolean field. Note that with SSZ at least you can read the `is_error` field prior to the contents of the `result` via offsets. This allows clients to switch the serialized type based on the `is_error` value.

arnetheduck:
Can someone familiar with libp2p draw a diagram of all payload bytes included in a typical packet, all the way from.. transport (tcp), including envelopes, wrappers, multiplexers etc?

specs/networking/node-identification.md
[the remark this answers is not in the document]
arnetheduck: As I mention in https://github.com/libp2p/specs/pull/100/files#r266291995 - protobuf is not deterministic, and thus not great for feeding into a hashing function to determine an ID, unless you used a modified protobuf version that's locked down.

specs/networking/rpc-interface.md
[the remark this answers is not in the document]
arnetheduck: the alternative would be to use a list - empty if there's no error, and one item if there is.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,257 @@
...
+
+**Body**:
+
+```
+(
+    network_id: uint8
+    latest_finalized_root: bytes32
+    latest_finalized_epoch: uint64
+    best_root: bytes32
+    best_slot: uint64
```
arnetheduck: slots are based on wall time - what's the best_slot field for?

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,257 @@
...
+    block_root: bytes32
+    slot: uint64
+)
+
+(
+    roots: []BlockRootSlot
+)
+```
+
+Send a list of block roots and slots to the requesting peer.
```
arnetheduck: which beacon block roots? known heads? all?

specs/networking/rpc-interface.md
arnetheduck: ẁe need a mechanism for recovering blocks, in case something is lost or the client goes offline for a short bit and loses a few (computer went to sleep / ISP went down for 10 minutes). 

I argue in the original issue (https://github.com/ethereum/eth2.0-specs/issues/692#issuecomment-469356620) that  it's often natural to request blocks backwards for this reason: the data structure we're syncing is a singly linked list pointing backwards in time and we receive attestations and blocks that let us discover heads "naturally" by listening to the broadcasts. With a `block_root+previous_n_blocks` kind of request we can both sync and recover, and for example use attestations to discover "viable" heads to work on, from a sync or recovery perspective. Indeed, negotiating finalized epochs in the handshake is somewhat redundant in that case, albeit a nice optimization (except for the chain id) - we could equally well request blocks from the peer that gossiped us the block or attestation whose parent we're missing - they should not be gossiping attestations they have not linked to a finalized epoch of value.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,257 @@
...
+
+**Response Body:**
+
+```
+(
+    block_bodies: []BeaconBlockBody
+)
+```
+
+Requests the `block_bodies` associated with the provided `block_roots` from the peer. Responses MUST return `block_roots` in the order provided in the request. If the receiver does not have a particular `block_root`, it must return a zero-value `block_body` (i.e., a `block_body` container with all zero fields).
```
arnetheduck: It seems to me that when everything is going smoothly, block bodies consist of very few attestations (they should be combined by then), and a few minor items like the transfers etc. has anything looked at the numbers to see how much value there is in having separate requests for headers and bodies? Requesting headers then bodies creates additional round-trips which are a cost on its own.

specs/networking/rpc-interface.md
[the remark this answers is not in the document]
arnetheduck: just to be clear - when encoding or decoding ssz, there generally exists no provision for skipping fields - even if `is_error` is false, `data` must contain bytes. embedding a `StatusData` in the data field seems to go against the spirit of SSZ generally, as SSZ decoders in general expect to know the exact type of each field, thus would not fit "naturally" in "normal" ssz code.

That said, this issue stems from using SSZ in a wire protocol setting for which it is not.. great.

specs/networking/rpc-interface.md
jannikluhn: Interesting! To summarize my understanding of your comment: Syncing forwards is safer as we can verify each block immediately when we receive it, but syncing backwards is more efficient/doesn't require additional database indexing (and I guess syncing forwards may require a negotiating phase to discover the best shared block). You're proposing to interpret the fact that I see lots of attestations on top of my sync peer's head flying around the network as evidence that their head is valid? And therefore, I'd be pretty safe syncing backwards?

That sounds reasonable. My original concern was that this requires me to know (at least some good fraction of) the validator set as otherwise my sync peer could create lots of fraudulent attestations for free that I have no chance of verifying. But I would notice this if I have at least one single honest peer (if I try to sync from them or compare the attestations coming from them).

Do you think having only a backwards sync is fine or do we need both (e.g. for highly adversarial environments, or resource constrained devices that don't participate in gossiping?).

specs/networking/rpc-interface.md
jannikluhn: pretty sure this is supposed to refer to the slot of the head block. Maybe rename `best_root` and `best_slot` to `head_root` and `head_slot` (or to be even more clear `head_block_root/slot`)?

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,257 @@
...
+
+All referenced data structures can be found in the [0-beacon-chain](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#data-structures) specification.
+
+## `libp2p` Protocol Names
+
+A "Protocol ID" in `libp2p` parlance refers to a human-readable identifier `libp2p` uses in order to identify sub-protocols and stream messages of different types over the same connection. Peers exchange supported protocol IDs via the `Identify` protocol upon connection. When opening a new stream, peers pin a particular protocol ID to it, and the stream remains contextualised thereafter. Since messages are sent inside a stream, they do not need to bear the protocol ID.
+
+## RPC-Over-`libp2p`
+
+To facilitate RPC-over-`libp2p`, a single protocol path is used: `/eth/serenity/beacon/rpc/1.0.0`. Remote method calls are wrapped in a "request" structure:
```
mhchia: Just curious about the reason why we have only one protocol id, instead of "one protocol for every type of request". If we have only one protocol id and use SSZ as the codec, from my understanding we might need a `method_id` prefix(either as a varint, or a fixed byte like in RLPx) for each message. This way we can read all of the bytes for a message, and then decode them according to their `method_id`

specs/networking/rpc-interface.md
arnetheduck: > more efficient

In terms of network / bandwidth, I'd say it's about the same but there are some nuances:

* in forward sync, I can ask for "more" slots than already exist, potentially saving round trips - a client could use this to request "all" data at the time of request arrival. consider the following race: A sends request, a new block is produced, B receives request (similar: B starts sending response which takes time, new block is produced).
* in backward sync, one could have an `(latest-head, known_slot_number)` request ("give me the block you consider to be the head, and history back to slot N") to alleviate this race, but then the server selects the head.
* both above races are generally solved by collecting data from broadcasts while syncing (classic subscribe-then-sync pattern) - they are mainly concerns if you don't want to subscribe or want to delay subscribing.
* in forward sync, I might end up on a different branch / head than I thought I would - the request itself does not point one out

In terms of client implementations, I think of backward sync as biased to make it cheaper for the server: the server already has the data necessary - also because the head is kept hot - while the client has to keep a chain of "unknown" blocks around / can't validate eagerly. An additional rule that the response must be forward-ordered could help the client apply / validate the blocks eagerly.

The backwards sync can be seen as more passive/reactive/lazy while forward sync is more active..

> attestations on top of my sync peer's head flying around the network as evidence that their head is valid

right. the assumption rests on two premises (thanks @djrtwo!):

* honest clients will not propagate invalid data (not signed by a validator they know)
* there's a slashing condition on creating unviable attestations - there's currently no penalty to create & sign an unviable block so one can perhaps imagine a malicious group of validators creating lots of these and for example spam everyone during important periods "for free". It sounds a bit far fetched though, tbh, to be creating blocks this way - would love to hear thoughts.
* I've weak-subjectively selected an initial state that contains some validators. I'd primarily look for anything signed by those validators as another heuristic for where to start syncing (even if the validator set might have changed from there).

> Do you think having only a backwards sync is fine or do we need both (e.g. for highly adversarial environments, or resource constrained devices that don't participate in gossiping?).

I'm not sure :) I'm curious to hear feedback on this point, but here are some thoughts:
* it's important that we have a request like Hello to ask for what clients consider to be the head for non-gossiping use cases - but I think that's orthogonal to the sync direction.
* clients should be free to send that request at any time, not just during the initial negotiation phase
* direction can be different for request and response - if different, requires a slightly "smarter" server
* there's a cost for each direction, in terms of implementation. I'd start with one and look for strong motivations before implementing the other, as the returns are not that great. Either direction is sufficient, really.

specs/networking/rpc-interface.md
mslipper: `libp2p` incurs stream switching overhead every time the protocol changes. We do have to extract the method ID here, however this can be done with byte offsets in SSZ.

specs/networking/rpc-interface.md
mslipper: I've updated this. See latest changes.

specs/networking/rpc-interface.md
raulk: Let me correct an imprecision there. There’s no such thing as protocol changes inside a stream. When you open a stream, you pin a protocol to it and it stays invariant. So if you have N protocols and want to exchange messages for all of them, you open at least N streams.

mslipper: Am I misunderstanding the `multistream-select` spec? From my interpretation of https://github.com/multiformats/multistream-select, for one TCP connection to send/receive messages for a given protocol a new stream needs to be selected by sending a 'switch protocols' message. Is this accurate?

mslipper: While my nomenclature may be incorrect (apologies if it is), I'm referring to any overhead incurred on-the-wire to make two different clients speak the same protocol.

zscole:
For the sake of expediency, we're going to be implementing a few changes in [Hobbits](https://github.com/Whiteblock/hobbits) to provide an easy and modular mechanism that allows clients to start talking to one another. For the time being, connections will be established via basic TCP with a fixed port (9000 is what we had discussed and agreed upon in the call). We'll be changing the Hobbits spec to assume the default serialization method SSZ and we're going to convert it from being a text to a binary protocol in version 0.2.

mslipper:
I've also brought back the message envelope as discussed during the call. This should match what goes into Hobbits version 0.2. See `messaging.md` for details.

specs/networking/rpc-interface.md
nisdas: I think `head_block_root` and `head_slot` would be clearer

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,257 @@
...
+       |   |  +-----------+ Epoch 2
+       +-+-+
+         ^
+         |
+       +-+-+
+       |   |  +-----------+ Epoch 1
+       +---+
+```
+
+Once the handshake completes, the client with the higher `latest_finalized_epoch` or `best_slot` (if the clients have equal `latest_finalized_epoch`s) SHOULD request beacon block roots from its counterparty via `beacon_block_roots` (i.e., RPC method `10`).
```
nisdas: How would this be handled if the clients both have equal `latest_finalized_epoch` and `best_slot` ?

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,257 @@
...
+
+```
+# BlockRootSlot
+(
+    block_root: bytes32
+    slot: uint64
+)
+
+(
+    roots: []BlockRootSlot
```
nisdas: Is there any size limit to the block roots in a response ? If there are no limits, I would imagine it being pretty easy to overwhelm any peer with it.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,246 @@
...
+
+**Body:**
+
+```
+(
+    reason: uint64
+)
+```
+
+Client MAY send `goodbye` messages upon disconnection. The reason field MUST be one of the following values:
```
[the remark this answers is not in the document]
nisdas: shouldn't we still be connected after `sync finished` ? We would still need to propagate any newly proposed blocks to our peers

specs/networking/rpc-interface.md
arnetheduck: then you discard the client because you know it's either at genesis or providing invalid data.. finalization happens several epochs behind `best_slot` in the happy case.

specs/networking/rpc-interface.md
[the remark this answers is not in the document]
arnetheduck: generally, the standard way to sync in these kinds of "live" protocols is to start listening to broadcasts, then initiate sync.. else you'll miss packets during sync and will have to recover again.

specs/networking/rpc-interface.md
nisdas: @arnetheduck ah no, I was referring to if both clients have the same `best_slot` and `finalized_epoch` together . So FinalizedEpoch<sub> A </sub>  == FinalizedEpoch<sub>B</sub> and BestSlot<sub> A </sub> == BestSlot<sub> B </sub>

zah:
> I've also brought back the message envelope as discussed during the call. This should match what goes into Hobbits version 0.2. See `messaging.md` for details.

When libp2p's multistream implementation is upgraded to 2.0 (https://github.com/libp2p/specs/pull/95), each request (a.k.a. new stream in libp2p's lingo) will be identified with a short numeric identifier. Woudn't this value be enough to determine the compression and encoding being used and expected response length? To achieve this in practice, we would just need to define additional protocol identifiers such as `/eth/serenity/beacon/rpc/hello/snappy:ssz`

cc @raulk to confirm that this is the best way to do it.

specs/networking/node-identification.md
```diff
@@ -0,0 +1,32 @@
...
+
+For clients to be addressable, their ENR responses MUST contain all of the above keys. Client MUST verify the signature of any received ENRs, and disconnect from peers whose ENR signatures are invalid. Each node's public key MUST be unique.
+
+The keys above are enough to construct a [multiaddr](https://github.com/multiformats/multiaddr) for use with the rest of the `libp2p` stack.
+
+It is RECOMMENDED that clients set their TCP port to the default of `9000`.
+
+## Peer ID Generation
+
+The `libp2p` networking stack identifies peers via a "peer ID." Simply put, a node's Peer ID is the SHA2-256 `multihash` of the node's public key struct (serialized in protobuf, refer to the [Peer ID spec](https://github.com/libp2p/specs/pull/100)). `go-libp2p-crypto` contains the canonical implementation of how to hash `secp256k1` keys for use as a peer ID.
```
zah: As mentioned in other comments, using SSZ hashing here would both reduce the footprint of the client (because existing code will be reused) and it would give us determinism (ProtoBuf can encode the same data in multiple ways, which will lead to multiple possible hash values).

specs/networking/rpc-interface.md
zah: It's true that at the moment, libp2p will write the textual identifier of the protocol at the start of each opened stream and this can be considered "stream switching overhead" when making requests. The problem will be addressed in MultiStream 2.0 though, so I think we can design the wire protocol with this upgrade in mind:

https://github.com/libp2p/specs/pull/95

With Multistream 2.0, there will be an initial handshake step determining the shared protocols between two peers. Then all shared protocols will be assigned numeric IDs which will be used when opening new streams.

zah: In other words, the ETH clients will be much simpler if `RPC-Over-libp2p` is defined to just create a new stream for each requests. We don't need to duplicate the multiplexing capabilities of libp2p in our own code.

specs/networking/node-identification.md
mslipper: This paragraph is referencing `libp2p`'s requirements. Any custom hashing scheme we use here will need to either be compatible with `libp2p` out-of-the-box or require changes to `libp2p` itself.

specs/networking/rpc-interface.md
mslipper: I can't tell if there is a timeline for Multistream 2.0 to be released. We can always update this spec to use Multistream 2.0 when the spec is ready, but for now I'm in favor of using the stable `libp2p` to avoid introducing implied dependencies.

mslipper:
> > I've also brought back the message envelope as discussed during the call. This should match what goes into Hobbits version 0.2. See `messaging.md` for details.
> 
> When libp2p's multistream implementation is upgraded to 2.0 ([libp2p/specs#95](https://github.com/libp2p/specs/pull/95)), each request (a.k.a. new stream in libp2p's lingo) will be identified with a short numeric identifier. Woudn't this value be enough to determine the compression and encoding being used and expected response length? To achieve this in practice, we would just need to define additional protocol identifiers such as `/eth/serenity/beacon/rpc/hello/snappy:ssz`
> 
> cc @raulk to confirm that this is the best way to do it.

Correct, we don't need this enveloping at all with `libp2p`. However some clients don't have `libp2p` implementations yet and specifically requested a way to communicate without it in order to test their networking stack. Thus, we need a way of representing compression and serialization that's agnostic of `libp2p`.

arnetheduck:
I think this might be a good time to split up this spec: 

* one that focuses on application level protocol: how to get and gossip blocks etc. Basically the SSZ-encoded RPC requests and other payloads.
* one that focuses on mapping eth2 to a transport layer. the working theory is that this transport layer is libp2p, but it could also be a simpler clear-test testnet protocol some have asked for, or even devp2p or whatever.

then, from the application level protocol we have a list of properties our transport layer must have: `version and feature negotiation`, `broadcast`, `peer-to-peer` etc. for each of these, we look at how they map to the underlying transport.

In summary, we'd have several documents (or sections of appendicec etc):
* 0-eth2-application.txt - blocks, attestations, etc
* 0-eth2-libp2p-mapping.txt - peer id format, gossipsub vs floodsub, minimal supported encryption methods, etc - basically a description of how eth uses libp2p specifically, starting from something like https://github.com/ethresearch/p2p/issues/4
* 0-eth2-temporary-mapping.txt - temporary pre-libp2p network for those that feel this is useful.

Some of the coordination issues we see stem from it being pretty early for protocol discussions, specially cross-client ones. Splitting it up will allow us to make progress on the application-level messages without being too distracted by other ongoings.

dreamcodez:
@mslipper,
@mslipper general current critiques:

1. regarding the lack of defined response codes, it seems that without defined generic response codes each rpc requires a pair of messages to be defined even if it is just to say 'yes that thing was successfully done'

2. regarding the lack of protocol extensibility in the form of protocol headers (are we so certain that people working our network protocol will never find our data structures wanting?)

3. regarding lack of separation of lightweight headers from message payload ; it is not enough to just know which byte offsets to read -- sometimes its useful to know metadata about a message without asking the sender to push the whole thing -- its also useful to make decisions (especially with regards to routing) without unwrapping the payload itself -- this should be in the form of lightweight headers.

@mslipper re: binary protocol 0.2
1. I would prefer request compression to be one value, and response compressions to be a list (list of compression to be accepted) -- this allows for future experimentation and upgradeability to other compression protocols without breaking shit. (this is of course applicable to just rpc messages)

2. If we are going to use a single binary byte to represent compression -- please at least specify the most popular 5-10ish codecs' bytecodes.

3. once again -- headers???

;)

*awaits feedback in bunker*

mslipper:
@dreamcodez 

> regarding the lack of defined response codes, it seems that without defined generic response codes each rpc requires a pair of messages to be defined even if it is just to say 'yes that thing was successfully done'

Note sure what you mean by this... the 'responses' as defined here contain the result of the procedure call or an error. Are you referring to defined error codes?

> regarding the lack of protocol extensibility in the form of protocol headers (are we so certain that people working our network protocol will never find our data structures wanting?)

What do you mean by headers in this case? If you want an additional 'command' (in Hobbits parlance, anyway) you'd just need to define a new `method_id`. Note that the serialization of that command is handled by the message envelope right now, and by upfront negotiation in the future.

> 

regarding lack of separation of lightweight headers from message payload ; it is not enough to just know which byte offsets to read -- sometimes its useful to know metadata about a message without asking the sender to push the whole thing -- its also useful to make decisions (especially with regards to routing) without unwrapping the payload itself -- this should be in the form of lightweight headers.

Can you give an example of the type of metadata you are referring to?

> I would prefer request compression to be one value, and response compressions to be a list (list of compression to be accepted) -- this allows for future experimentation and upgradeability to other compression protocols without breaking shit. (this is of course applicable to just rpc messages)

This would mean the introduction of a handshake process, which defeats the purpose of having a simple intermediary wire format to tide us over until `libp2p` is ready. You can already change the compression/encoding protocols by changing the nibble values in the message envelope.

> If we are going to use a single binary byte to represent compression -- please at least specify the most popular 5-10ish codecs' bytecodes.

I'd honestly prefer _not_ to do this, since defining them would create an implicit requirement within this spec to support multiple compression algorithms. I'd much rather we decide on one as a group to reduce implementation overhead. This is why `0x0` (i.e., no compression) is the only defined compression nibble at this time.

specs/networking/rpc-interface.md
mslipper: Frankly, I'm in favor of simply having integer version numbers and have a blanket statement that sub-protocol version numbers are neither forwards nor backwards compatible.

specs/networking/node-identification.md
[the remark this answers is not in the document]
mslipper: Wouldn't this be handled at the `libp2p` layer? Here we're describing how to construct a `multiaddr` from an ENR; the actual handling of the `multiaddr` itself and the underlying hash construction would be the responsibiliy of `libp2p`.

specs/networking/rpc-interface.md
arnetheduck: that would be my preference as well, with how the encoding and protocol looks today.

if we had a serialization format that allowed forwards/backwards compatible additions (ie adding fields) we could maybe consider two-level numbering here, where the first number would be the blanket statement, while the second would signal the version with additional fields added, which would still be compatible with previous clients.

Such an encoding is generally a good thing in the wire use case, which would be a reason to look to extensions to SSZ when used outside consensus (a super-set of SSZ for example).

mslipper:
> I think this might be a good time to split up this spec:
> 
>     * one that focuses on application level protocol: how to get and gossip blocks etc. Basically the SSZ-encoded RPC requests and other payloads.
> 
>     * one that focuses on mapping eth2 to a transport layer. the working theory is that this transport layer is libp2p, but it could also be a simpler clear-test testnet protocol some have asked for, or even devp2p or whatever.
> 
> 
> then, from the application level protocol we have a list of properties our transport layer must have: `version and feature negotiation`, `broadcast`, `peer-to-peer` etc. for each of these, we look at how they map to the underlying transport.
> 
> In summary, we'd have several documents (or sections of appendicec etc):
> 
>     * 0-eth2-application.txt - blocks, attestations, etc
> 
>     * 0-eth2-libp2p-mapping.txt - peer id format, gossipsub vs floodsub, minimal supported encryption methods, etc - basically a description of how eth uses libp2p specifically, starting from something like [ethresearch/p2p#4](https://github.com/ethresearch/p2p/issues/4)
> 
>     * 0-eth2-temporary-mapping.txt - temporary pre-libp2p network for those that feel this is useful.
> 
> 
> Some of the coordination issues we see stem from it being pretty early for protocol discussions, specially cross-client ones. Splitting it up will allow us to make progress on the application-level messages without being too distracted by other ongoings.

@arnetheduck I really like this idea... what do you think of doing this separately from this PR? I'm sure that there will be additional feedback generated from those documents.

specs/networking/node-identification.md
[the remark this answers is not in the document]
arnetheduck: it would, but libp2p itself looks broken in this case - we need to keep an eye on that upstream issue so that we don't spread the breakage further.

Does using ENR require decoding RLP in this context?

FrankSzendzielarz:
Yes. I think we need flexibility and having aspects like serialisation method separated from message format and semantics will be of more immediate utility than say wire performance.

arnetheduck:
> @arnetheduck I really like this idea... what do you think of doing this separately from this PR? I'm sure that there will be additional feedback generated from those documents.

I think we're pretty close already in terms of how the spec is sectioned (nice work!) - just need to make the distinction slightly more explicit. It's mostly an administrative matter to facilitate discussion, make a dedicated space for separate non-libp2p discussion, and establish terminology once and for all (what's "wire" and "application" and so on). Can be done at any time, whenever feels convenient. 

Eventually, I also hope it will lead to a more clearly documented interface between eth2 and the underlying overlay (haha).

dreamcodez:
> @dreamcodez
> 
> > regarding the lack of defined response codes, it seems that without defined generic response codes each rpc requires a pair of messages to be defined even if it is just to say 'yes that thing was successfully done'
> 
> Note sure what you mean by this... the 'responses' as defined here contain the result of the procedure call or an error. Are you referring to defined error codes?
> 
Omitting a generic rpc response codes will require us produce an explicitly coded 'response' message to be paired with each said request.. If message A gets sent via rpc, then its nice to not have to create a message AResponse to indicate the response permutations of 'success' and 'error' at the very least.

> > regarding the lack of protocol extensibility in the form of protocol headers (are we so certain that people working our network protocol will never find our data structures wanting?)
> 
> What do you mean by headers in this case? If you want an additional 'command' (in Hobbits parlance, anyway) you'd just need to define a new `method_id`. Note that the serialization of that command is handled by the message envelope right now, and by upfront negotiation in the future.
Not referring to commands -- referring to pieces of metadata which our core protocol could not have possibly foreseen use cases for in the future -- for example -- proxies and/or routing topologies will need to tag extra metadata to efficiently push decisions and context to the next hop. rather than having to modify our protocol, they can just adopt new header extensions.
> > regarding lack of separation of lightweight headers from message payload ; it is not enough to just know which byte offsets to read -- sometimes its useful to know metadata about a message without asking the sender to push the whole thing -- its also useful to make decisions (especially with regards to routing) without unwrapping the payload itself -- this should be in the form of lightweight headers.
> 
> Can you give an example of the type of metadata you are referring to?
> 
See above -- I know there are more use cases but I cannot express them at the moment. The main idea is i should be able to grab lightweight metadata when looking up objects to see if i need the full object (maybe it hasn't changed since last time I asked).

> > I would prefer request compression to be one value, and response compressions to be a list (list of compression to be accepted) -- this allows for future experimentation and upgradeability to other compression protocols without breaking shit. (this is of course applicable to just rpc messages)
> 
> This would mean the introduction of a handshake process, which defeats the purpose of having a simple intermediary wire format to tide us over until `libp2p` is ready. You can already change the compression/encoding protocols by changing the nibble values in the message envelope.
I disagree -- this does not force anyone to handshake in a way which would create any additional overhead -- a rpc requestor peer can simply say 'hey i'm sending this in spdy, and i can receive responses in X,Y,Z' -- if the response compression preference list gives a few choices -- this creates a high probability of interoperability -- especially if the scheme chosen in the initial request is spdy or something non-experimental.

> > If we are going to use a single binary byte to represent compression -- please at least specify the most popular 5-10ish codecs' bytecodes.
> 
> I'd honestly prefer _not_ to do this, since defining them would create an implicit requirement within this spec to support multiple compression algorithms. I'd much rather we decide on one as a group to reduce implementation overhead. This is why `0x0` (i.e., no compression) is the only defined compression nibble at this time.
Let's say that our request looks like:
`EWP 0.1 snappy snappy,gzip` -- this means that I will be able to work with any endpoint which supports snappy, but i can also support gzip coming back...
I don't really see any drawbacks with supporting this...

jannikluhn:
> I think this might be a good time to split up this spec

I support this. Maybe we can even abstract away the serialization format from the application level protocol in order to isolate another heavily discussed component.

specs/networking/rpc-interface.md
raulk: +1 on integers to signal a **generation** (generation 1, generation 2...). Any reason you wouldn’t have a varint style bitmap in the HELLO message to communicate finer-grained capabilities? @arnetheduck 

I would model serialisation format and compression as part of the protocol ID. Then allow Multistream to negotiate.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,292 @@
...
+### Hello
+
+**Method ID:** `0`
+
+**Body**:
+
+```
+(
+    network_id: uint8
+    chain_id: uint8
```
arnetheduck: the rest of the eth2 spec favors uint64 for simplicity and uniformity - what's the rationale behind `uint8` here?

mslipper: I think it's unlikely that we'll need to support mode than 256 chain IDs, so we can save the bytes.

specs/networking/rpc-interface.md
mslipper: Added a limit.

specs/networking/rpc-interface.md
jannikluhn: A possible compatible change could be added message types, so I think minor version numbers could be useful in some cases.

> Any reason you wouldn’t have a varint style bitmap in the HELLO message to communicate finer-grained capabilities?

Capabilities are known from the discovery protocol/ENRs already (but we need to define what types of capabilities we need). So I don't think we need it in the HELLO message.

specs/networking/rpc-interface.md
jannikluhn: There will be 1024 shards at least. We haven't talked about an id scheme for those yet, but to be safe I think we should go with uint64 (also considering that HELLO messages are sent only infrequently). Same for network id.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,292 @@
...
+
+1. `0`: No error.
+2. `10`: Parse error.
+2. `20`: Invalid request.
+3. `30`: Method not found.
+4. `40`: Server error.
+
+### Alternative for Non-`libp2p` Clients
+
+Since some clients are waiting for `libp2p` implementations in their respective languages. As such, they MAY listen for raw TCP messages on port `9000`. To distinguish RPC messages from other messages on that port, a byte prefix of `ETH` (`0x455448`) MUST be prepended to all messages. This option will be removed once `libp2p` is ready in all supported languages.
```
atoulme: I think we should allow using a separate port. It's entirely possible a client will allow both communication modes.

atoulme: Also, this paragraph should go to the top since it relates to the envelope and port 9000.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,292 @@
...
+
+### Get Status
+
+**Method ID:** `2`
+
+**Request Body:**
+
+```
+(
+	sha: bytes32
```
atoulme: I didn't ask when this got drafted, sorry. What is sha here?

specs/networking/messaging.md
```diff
@@ -0,0 +1,46 @@
...
+
+Clients MUST ignore messages with mal-formed bodies. The compression/encoding nibbles MUST be one of the following values:
+
+## Compression Nibble Values
+
+- `0x0`: no compression
+
+## Encoding Nibble Values
+
+- `0x1`: SSZ
```
atoulme: Permission to add `0x2` for BSON please?

specs/networking/messaging.md
```diff
@@ -0,0 +1,46 @@
...
+|           body           |
+|                          |
++--------------------------+
+```
+
+Clients MUST ignore messages with mal-formed bodies. The compression/encoding nibbles MUST be one of the following values:
+
+## Compression Nibble Values
+
+- `0x0`: no compression
```
atoulme: Permission to add `0x1` for snappy compression please?

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,292 @@
...
+    block_root: bytes32
+    slot: uint64
+)
+
+(
+    roots: []BlockRootSlot
+)
+```
+
+Requests a list of block roots and slots from the peer. The `count` parameter MUST be less than or equal to `32768`. The slots MUST be returned in ascending slot order.
```
paulhauner: I have some questions regarding the response:

- Can you skip a slot (e.g., `1, 3, 4`)?
- Can you return less/more than the requested roots?
- Can you start at higher slot?
- Can you return a slot outside of the `start_slot + count` bounds?

Maybe _"The slots MUST be returned in ascending slot order."_ is succinct already? If this is the case we could add something like "The only requirements for `roots` are ...".

P.S. this is my first comment, thanks for making this doc!

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,292 @@
...
+(
+    block_roots: []HashTreeRoot
+)
+```
+
+**Response Body:**
+
+```
+(
+    block_bodies: []BeaconBlockBody
```
paulhauner: To obtain blocks from a peer I expect to first request headers, then bodies. After receiving the bodies, headers are combined with the bodies and `chain.import_block(..)` (or w/e) is called.

Given that we don't have any information linking `body` to `header`, we need to iterate headers to find `header.block_body_root == hash_tree_root(body)`.

You could reduce this iteration to "all headers from that `peer_id`" rather than _all_ headers, I guess.

Nonetheless, thoughts on returning `(block_body, block_root)` to allow `O(1)` lookup of corresponding header?

paulhauner: Oh ignore this, I can just key the headers by `block_body_root`!

specs/networking/messaging.md
jannikluhn: I believe it would be best if we could agree on a single encoding format to maximize compatibility.

FrankSzendzielarz:
Yes but the protocol could allow many😀

On 23 Mar 2019 20:04, jannikluhn <notifications@github.com> wrote:

@jannikluhn commented on this pull request.

In specs/networking/messaging.md<https://github.com/ethereum/eth2.0-specs/pull/763#discussion_r268405597>:

> ++--------------------------+
+|                          |
+|           body           |
+|                          |
++--------------------------+
+```
+
+Clients MUST ignore messages with mal-formed bodies. The compression/encoding nibbles MUST be one of the following values:
+
+## Compression Nibble Values
+
+- `0x0`: no compression
+
+## Encoding Nibble Values
+
+- `0x1`: SSZ


I believe it would be best if we could agree on a single encoding format to maximize compatibility.

specs/networking/messaging.md
mslipper: Would prefer not, for the same reasons articulated on our call - we should agree together on a single encoding scheme.

specs/networking/rpc-interface.md
mslipper: The commit hash of the node.

specs/networking/messaging.md
mslipper: See above - I don't want to commit teams to an implementation without getting consensus.

specs/networking/rpc-interface.md
mslipper: Sounds good. I'll update.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,292 @@
...
+)
+```
+
+and their corresponding responses are wrapped in a "response" structure:
+
+```
+(
+    id: uint64
+    response_code: uint16
+    result: bytes
```
atoulme: Can we just use one structure? You can tell if it's a response by comparing request IDs.

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,292 @@
...
+)
+```
+
+Requests beacon block headers from the peer starting from `(start_root, start_slot)`. The response MUST contain no more than `max_headers` headers. `skip_slots` defines the maximum number of slots to skip between blocks. For example, requesting blocks starting at slots `2` a `skip_slots` value of `1` would return the blocks at `[2, 4, 6, 8, 10]`. In cases where a slot is empty for a given slot number, the closest previous block MUST be returned. For example, if slot `4` were empty in the previous example, the returned array would contain `[2, 3, 6, 8, 10]`. If slot three were further empty, the array would contain `[2, 6, 8, 10]` - i.e., duplicate blocks MUST be collapsed. A `skip_slots` value of `0` returns all blocks.
+
+The function of the `skip_slots` parameter helps facilitate light client sync - for example, in [#459](https://github.com/ethereum/eth2.0-specs/issues/459) - and allows clients to balance the peers from whom they request headers. Clients could, for instance, request every 10th block from a set of peers where each per has a different starting block in order to populate block data.
+
+### Beacon Block Bodies
+
+**Method ID:** `12`
```
atoulme: Just like in LES, I would use different method ids for requests and responses. So it's possible for me to send you proactively blocks and headers using RPC, and you don't need to know about it in advance.

specs/networking/rpc-interface.md
arnetheduck: ah. right. I find that the situation is somewhat analogous to receiving a block whose parent is unknown to it - you have to make a very similar decision there - the information in this hello, just like in the parentless block, is essentially useless, from a trust perspective, and you need to turn to other sources. 

@djrtwo suggested that attestations might be a good heurestic as signing one carries risk for the validator that does so. The information here can, from what I can see, be used to quickly disconnect from a client, if they're saying the network is different. The rest is advisory, and you're hoping for the best.

specs/networking/rpc-interface.md
arnetheduck: @raulk taking a step back, my initial understanding of the libp2p setup was that you would negotiate capabilities with discovery mainly and connect to clients you know you have common ground with - then merely verify the support by signing up to the various streams here and that each stream would be a protocol on its with libp2p dealing with the mulitplexing etc - that has changed now I see, and it looks like there's another layer of protocol negotiation within the stream to discover capabilities - that feels.. redundant, to do the same work twice, and somewhat limiting, because how does a client add a completely new message they want to test or use in some client-specific scenario (for example to establish / evaluate its usefulness) - but it seems I need to reread the newer spec.

I'll be honest though and say that I don't fully understand where the varint would go at this point with the various layers, but integers tend to be harder to negotiate than strings, in a decentralized manner - a string, you just pick one and start using it - if it becomes popular, people will avoid it. Numbers.. you need a registry and the associated maintenance.

raulk: @arnetheduck my intention isn't to propose any changes here; was just curious to hear the rationale of the Eth2.0 community re: finer-grained protocols vs. coarse-grained protocol with capabilities. We also debate this occasionally in the libp2p community ;-) [libp2p supports both patterns].

Re: semver vs. protocol generations. libp2p does not impose a specific version token (if any). When registering a handler, you can attach a [predicate](https://github.com/multiformats/go-multistream/blob/master/multistream.go#L124) to evaluate the match. So handler X could match N versions, and when receiving the callback, you can inspect the protocol pinned on the stream to infer which abilities you activate, etc.

We've traditionally used semver it in libp2p, IPFS, etc., but a few of us are not convinced of its aptness. Semver is good to convey the severity of changes in APIs, but protocol evolution is a different beast.

You generally strive to keep things backwards compatible, yet enable feature upgrades/opt-ins over time that may not be rolling cumulative, e.g. if 1.14.0 and 1.15.0 introduce feature X and Y respectively, how do I convey that a given peer supports Y but not X?

That's where protocol granularity comes into play: potentially make each feature/message/RPC a different protocol, and track "generations/revisions" of those protocols. libp2p supports that design. A few thoughts:

* We're evolving Multistream to avoid round trips when you're certain the other party supports a protocol (_selection_ vs _negotiation_), for example, via an ENR.
* If two messages have timing dependencies (can't send message B until after A), and they segregated across protocols, it may make state-tracking a bit more complicated.

---

> integers tend to be harder to negotiate than strings, in a decentralized manner - a string, you just pick one and start using it - if it becomes popular, people will avoid it.

I meant replacing semver by protocol generations, e.g. `/eth/serenity/rpc/v10`. Sorry for not being clear!

raulk:
@zah 

> When libp2p's multistream implementation is upgraded to 2.0 (libp2p/specs#95), each request (a.k.a. new stream in libp2p's lingo) will be identified with a short numeric identifier. Woudn't this value be enough to determine the compression and encoding being used and expected response length? To achieve this in practice, we would just need to define additional protocol identifiers such as /eth/serenity/beacon/rpc/hello/snappy:ssz
> 
> cc @raulk to confirm that this is the best way to do it.

You don't need to wait for multistream 2.0. When you open a new stream, multistream 1.0 is in charge of associating that stream with a protocol. To be clear:

1. When opening a stream, you tell libp2p which protocol `P` you want it to be associated with via the API.
2. libp2p opens a stream with the underlying multiplexer (yamux, mplex, spdystream, quic, etc.). At this point, the stream is _just a number_.
3. multistream 1.0 kicks in and selects protocol `P`.
4. libp2p returns the stream, contextualised for that protocol `P`.

arnetheduck:
Generally, I see we've gained a lot of different `uintXX` sizes, while eth2 has opted to go with `uint64` exclusively (almost?) - network/chain/method id etc. The motivation seems to be to save a few bytes here and there, but they look like one-offs in requests and the like, nothing that will fundamentally make an actual difference.

Considering sizes of payloads and taking into consideration the departure from rest of the eth2 spec, how do we feel about sticking with `uint64` unless there's strong (aka motivated by compelling numbers) reason not to?

arnetheduck:
Generally, I see we've gained a lot of different `uintXX` sizes, while eth2 has opted to go with `uint64` exclusively (almost?) - network/chain/method id etc. The motivation seems to be to save a few bytes here and there, but they look like one-offs in requests and the like, nothing that will fundamentally make an actual difference.

Considering sizes of payloads and taking into consideration the departure from rest of the eth2 spec, how do we feel about sticking with `uint64` unless there's strong (aka motivated by compelling numbers) reason not to?

specs/networking/rpc-interface.md
jrhea: > Do you think having only a backwards sync is fine or do we need both (e.g. for highly adversarial environments, or resource constrained devices that don't participate in gossiping?).

It seems reasonable to sync backwards from the latest received gossiped block (at least as an initial implementation)

> in backward sync, one could have an (latest-head, known_slot_number) request ("give me the block you consider to be the head, and history back to slot N") to alleviate this race, but then the server selects the head.

i would argue that we should drop start_slot.  if we give clients the option to request a block by either start_slot or start_root then that forces us to maintain a lookup or search mechanism for both.  if we are saying that both fields (start_slot and start_root) required to sync, then I would disagree.  we should be able to simply perform a lookup by block_root and walk the chain backwards until we reach max_headers.

specs/networking/rpc-interface.md
atoulme: What is that? Best root?

specs/networking/rpc-interface.md
```diff
@@ -0,0 +1,292 @@
...
+
+**Body**:
+
+```
+(
+    network_id: uint8
+    chain_id: uint64
+    latest_finalized_root: bytes32
+    latest_finalized_epoch: uint64
+    best_root: bytes32
```
arnetheduck: Taking into account the backwards sync suggested elsewhere, and that we can use attestations as a (strong) heuristic that a block is valid and useful, it seems prudent to include (some) attestations here - instead of simply supplying some data like best_root that cannot be trusted anyway, a recent attestation would help the connecting client both with head / fork selection and to know with a higher degree of certainty that the root sent "makes sense" and should be downloaded.

The details of this are TBD - but probably we're looking at something like `attestations: [Attestation]` where it's up to the client to choose a representative and recent set (or none, which is also fine, because then one can listen to broadcasts).

specs/networking/rpc-interface.md
arnetheduck: >  latest received gossiped block

or even better, latest gossiped attestation

> Do we really need start_slot?

I would say that if we go with backwards sync, we should not implement forwards sync here or elsewhere unless there's a strong case for that direction. Having to implement both directions negates some of the benefits of backward sync and adds implementation surface.

It is quite possible to add forward sync in a later version of the protocol as well should it prove necessary.

jrhea: > or even better, latest gossiped attestation

I can dig that

@arnetheduck or anyone else. Why do we need start_slot

PR #764, opened 2019-03-12
Clean up GENESIS_FORK_VERSION

JustinDrake:


djrtwo:
I'd rather fork versions be human readable numbers from `[0, 2**32)` that are converted to 4 byte form in the state. Rather than 4 byte values we have to think about.

JustinDrake:
We can do `GENESIS_FORK_VERSION = int_to_bytes4(i)` where `i` is the human-readable number in `[0, 2**32)`. Minor detail—feel free to close :)

PR #766, opened 2019-03-13
Added light client related files

vbuterin:


zah:
This new proposal is really good and easy to follow, but I do have some questions:

1) Consider the situation where the light client receives a `[BlockValidityProof](https://github.com/ethereum/eth2.0-specs/blob/b566722b52ad2cd9f92a0b48953e3b784a04853d/specs/light_client/sync_protocol.md#verifying-blocks)` from a valid committee majority, but the referenced block did not become part of the canonical chain. As I see it, this may happen if the proposer for the next slot didn't notice the referenced block in time and some alternative history was created that got eventually selected due to the fork choice rules.

    How will the light client handle this situation? Perhaps it has to keep requesting BlockValidityProofs from other peers until it gets one at a better height? The situation is somewhat bad, because the light client will trust a crosslink that was not officially confirmed.

2) The sync procedure allows you to keep monitoring the changes in a particular shard's committee (and a state root the validators agreed on), but how do you get the very latest crosslink of a different shard? Presumably, you can use the known state root to verify a merkle proof for the members of another committee, but perhaps the spec needs to be refactored a bit to facilitate this (we need a function that computes the committee for a particular shard).

specs/light_client/sync_protocol.md
```diff
@@ -0,0 +1,171 @@
...
+    epoch = slot_to_epoch(header.slot)
+    
+    actual_committee_count = max(
+        earlier_validator_count // (SHARD_COUNT * TARGET_COMMITTEE_SIZE),
+        later_validator_count // (SHARD_COUNT * TARGET_COMMITTEE_SIZE),
+    ) + 1
+    
+    def get_offset(count, end:bool):
+        return get_split_offset(count,
+                                SHARD_COUNT * committee_count,
```
zah: This seems like a simple oversight. `committee_count` is not defined in the scope here.

specs/light_client/sync_protocol.md
```diff
@@ -0,0 +1,171 @@
...
+                                validator_memory.shard_id * committee_count + (1 if end else 0))
+                                
+    actual_earlier_committee = maximal_earlier_committee[
```
zah: There are some unfinished renames here: `maximal_earlier_committee` vs `earlier_committee`, `get_maximal_earlier_committee` vs `get_earlier_period_data`

vbuterin:
> Consider the situation where the light client receives a BlockValidityProof from a valid committee majority, but the referenced block did not become part of the canonical chain. As I see it, this may happen if the proposer for the next slot didn't notice the referenced block in time and some alternative history was created and got eventually selected due to the fork choice rules. How will the light client handle this situation?

I'm inclined to say that light clients can just keep asking for headers, and internally apply LMD GHOST based on the committee they're following. It is possible for this fork choice to diverge from the "official" one, but only in cases where close to 1/2 of validators or the committee are byzantine.

> The sync procedure allows you to keep monitoring the changes in a particular shard's committee (and a state root the validators agreed on), but how do you get the very latest crosslink of a different shard? 

You authenticate the Merkle path shard block header -> beacon block header -> state -> latest_crosslink_data.

zah:
So, the light client has a slightly weaker notion of finality. Isn't this a problem for the use case of authenticating ETH2 data in other blockchains? You cannot rely on heuristics like requesting data from other peers in such contexts.

> You authenticate the Merkle path shard block header -> beacon block header -> state -> latest_crosslink_data.

But the `latest_crosslink_data` is not necessarily the very latest crosslink. Two scenarios can happen:

1) The monitored shard did not receive block proposal in the most recent epochs. The best `BlockValidityProofs` proof will then sync the client with a slightly older version of the state. There might even be some new crosslinks that are finalized at this point.

2) The client is interested in the non-finalized crosslinks appearing in the slots after the block in the `BlockValidityProofs`.

Both problems are solvable if the client is able to reconstruct another committee and then use that to authenticate a recent block header.

vbuterin:
So if you want to learn about finalized blocks, then you can authenticate the Merkle path shard block header -> beacon block header -> state -> finalized_epoch -> historical block root[finalized_epoch], and there's your finalized block. Even if that shard block gets reverted later, if you assume honest majority of the persistent committee, the fact that that block exists at all implies that the finalized block actually is finalized.

zah:
Fair enough, thanks for the clarification.

Am I right to assume that one of the next required steps would be to develop an automatic calculator that, given an arbitrary SSZ expression/function, can compute the size of the required Merkle proof and perhaps the minimal amount of time (gas) to verify it in a certain environment? Such a tool can also try to produce a zk-snark circuit to test the feasibility of compressing the proof. I'll be very happy to work on something like this.

Having such a tool would allow us to compare in more concrete terms the various possible ways to arrange the signed data. I think there might be some reasonable optimisations brought by denormalizing the data a little bit (i.e. including an easier to reach copy of something, that can be otherwise obtained through a longer chain of links). This affects the size of the Merkle proofs, the required verification time and the size of the zk-snark circuits.

As an example, even here, you are suggesting that the `BlockValidityProof` will include a signed block header, but in reality this will have to be the `AttestationData` that the validators sign and broadcast. By adding additional fields or signatures in the attestations, we make certain trade-offs in the required computation time and network traffic for both light clients and full nodes.

vbuterin:
> Am I right to assume that one of the next required steps would be to develop an automatic calculator that, given an arbitrary SSZ expression/function, can compute the size of the required Merkle proof and perhaps the minimal amount of time (gas) to verify it in a certain environment?

It depends on what your goals are :)

If you just want to build a light client, then none of that is needed, you just need a tool to create and verify multiproofs of arbitrary SSZ objects (that happens to be on my own todolist as well).

> As an example, even here, you are suggesting that the BlockValidityProof will include a signed block header, but in reality this will have to be the AttestationData that the validators sign and broadcast

Don't think this has anything to do with `AttestationData`. This particular light client design is based off of persistent committee shard attestations, not crosslink committee block attestations (it's done that way to cut costs, as you would only need to download a new committee every ~9 days).

djrtwo:
What's the status of these vs the files already in `dev` branch?
It seemed that some were accidentally direct committed to master at some point

vbuterin:
These files are newer. I feel like we should just merge this PR in soon; it's not final but it's definitely closer to final than what exists there now.

specs/light_client/merkle_proofs.md
```diff
@@ -127,8 +154,38 @@ def verify_multi_proof(root, indices, leaves, proof):
...
+
+```python
+def merkle_partial_from_paths(obj, paths):
+    indices = set()
+    for path in paths:
+        indices = indices.union(get_generalized_indices(obj, path))
+    return MerklePartial(
+        root=hash_tree_root(obj),
+        indices=indices,
+        values= mk_multi_proof
```
djrtwo: `mk_multi_proof` is undefined, and this function does not set `proof`

vbuterin: Removed that section entirely for now. I think the right time to bring it back would be after we have executable SSZ serialization, as then we would be able to actually generate the proofs as part of the spec (right now I'm not sure what the "final" interface in the executable spec would be...)

PR #772, opened 2019-03-14
Fair proposer selection probability

vbuterin:
Makes a validator's probability of being selected as a proposer proportional to a validator's balance. This ensures fairness between validators of different deposit sizes, making proposer manipulation attacks more expensive and ensuring that there is no advantage to be gained by concentrating one's ETH in more vs fewer validator slots. This allows us to widen the gap between the min and max deposit size.

Note that as a side effect, proposer selection becomes less predictable, but I don't feel like this is a large downside. Specifically, the proposer selection does not have a guaranteed 1 epoch lookahead like the committees do. This change is noted in the validator guide within this PR.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1014,14 +1014,14 @@ def get_beacon_proposer_index(state: BeaconState,
 
     assert previous_epoch <= epoch <= next_epoch
 
-    first_committee, _ = get_crosslink_committees_at_slot(state, slot, registry_change)[0]    i = 0
-    while i < len(first_committee):
+    first_committee, _ = get_crosslink_committees_at_slot(state, slot, registry_change)[0]
+    i = 0
+    while True:
         rand_byte = hash(generate_seed(get_current_epoch(state)) + int_to_bytes8(i // 32))[i % 32]
         candidate = first_committee[(epoch % i) % len(first_committee)]
         if get_effective_balance(state, candidate) * 256 > MAX_DEPOSIT_AMOUNT * rand_byte:
             return candidate
         i += 1
-    return first_committee[epoch % len(first_committee)]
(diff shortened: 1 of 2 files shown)


JustinDrake:
I refactored the logic to remove the line `return first_committee[epoch % len(first_committee)]` because it would never get triggered in practice (not friendly to testing), and it's also simpler :)

vbuterin:
OK that works. I had thought you might favor a constant bound on runtime, but I suppose the probability of it actually going longer than ~100 rounds is cryptographically securely low in any case.

djrtwo:
Can you write a note to implementers on why this is needed/desirable?

djrtwo:
> Note that as a side effect, proposer selection becomes less predictable, but I don't feel like this is a large downside.

Ah, this means there is no longer a guaranteed 1 epoch lookahead. You can know your committee but don't know if you are the proposer until start of epoch because of potential changes in balance. We need to make note of this in V-guide -- you cannot tell if proposer until epoch-of even if you used lookahead on your committee/slot

vbuterin:
> Can you write a note to implementers on why this is needed/desirable?

Done.

> Ah, this means there is no longer a guaranteed 1 epoch lookahead. You can know your committee but don't know if you are the proposer until start of epoch because of potential changes in balance. We need to make note of this in V-guide -- you cannot tell if proposer until epoch-of even if you used lookahead on your committee/slot

Correct. There's a guaranteed 1-epoch lookahead for which committee you will be in, but not proposer status. Although we expect that ~99.99%+ of the time, the proposer will nevertheless be the same in two different branches, because balances don't change _that_ much.

djrtwo:
Fixed a couple of minor bugs and modified the validator guide to not have lookahead on proposing. Just on committee

PR #773, opened 2019-03-14
Create an block header struct

atoulme:
This is so we can implement light client transporting block headers only.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -463,15 +463,23 @@ The following data structures are defined as [SimpleSerialize (SSZ)](https://git
 
 ```python
 {
+    'header': BeaconBlockHeader,
+
+    # Body
+    'body': BeaconBlockBody,
+}
+```
+
+#### `BeaconBlockHeader`
+
+```python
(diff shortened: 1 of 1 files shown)


vbuterin:
I feel like the better way to do this is to standardize the notion of "expansions" and "summaries" that I define in this doc: https://github.com/ethereum/eth2.0-specs/blob/154eec0d027468d1a228f7f76e233eec9c4320ff/specs/light_client/sync_protocol.md

The "header contains root of body" thing was done that way for a reason (namely, so that signing over the header also signs over the body, but at the same time you could verify signatures using just the header) and should probably stay that way. It's easy to convert any `BeaconBlock` into its corresponding `BeaconBlockHeader` already in any case.

JustinDrake:
@atoulme Can you please issue pull requests on `dev` (as opposed to `master`). Also, as pointed out by Vitalik, there *is* a [block header struct](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#beaconblockheader) which fits the "summary"/"expansion" paradigm.

PR #777, opened 2019-03-15
Added FixedSizeList wrappers

vbuterin:
Requires corresponding changes to the spec pythonizer.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1528,7 +1528,7 @@ def get_genesis_beacon_state(genesis_validator_deposits: List[Deposit],
         validator_registry_update_epoch=GENESIS_EPOCH,
 
         # Randomness and committees
-        latest_randao_mixes=[ZERO_HASH for _ in range(LATEST_RANDAO_MIXES_LENGTH)],
+        latest_randao_mixes=FixedSizeList([ZERO_HASH for _ in range(LATEST_RANDAO_MIXES_LENGTH)]),
         previous_shuffling_start_shard=GENESIS_START_SHARD,
         current_shuffling_start_shard=GENESIS_START_SHARD,
         previous_shuffling_epoch=GENESIS_EPOCH,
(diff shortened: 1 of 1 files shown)


JustinDrake:
Should we go with `Arrray`? (It's what Golang uses. Also see [this comment](https://github.com/ethereum/eth2.0-specs/pull/696#discussion_r261393027).)

CarlBeek:
> Should we go with `Arrray`? (It's what Golang uses. Also see [this comment](https://github.com/ethereum/eth2.0-specs/pull/696#discussion_r261393027).)

I disagree with this notation. Many, if not the majority of, languages seem to deem `Array` to be dynamically sized. [[Source](https://en.wikipedia.org/w/index.php?title=Comparison_of_programming_languages_(array)#Array_system_cross-reference_list)]

JustinDrake:
That table is pretty cool. It looks like it's roughly 50/50 (what they call "init" and "heap" is a special case of fixed sized).

CarlBeek:
Agreed. My point is more that the terminology may be ambiguous to users of certain languages. (A quick poll of the 3 CS majors in the room with me had them all declaring it to be dynamic. (I am aware this is not a strong argument))

vbuterin:
If we want clarity, the name needs to evoke the property that actually make static and dynamic sized lists different. List vs tuple, array vs slice, list vs array all fail to do that.

I'll keep thinking for better alternatives...

protolambda:
Ideas: Mutable List vs Immutable list. Or appendable list vs fixed list. Or list vs vector.

Honestly I think it's better to keep it simple and be clear in the definition of the type itself.
Slice vs Array is clear enough for me, and so is List vs Tuple.

vbuterin:
If one-word-ness is a desideratum then list vs vector is definitely the least bad of what I've seen so far.

JustinDrake:
+1 for list vs vector. Happy to rename "tuple" to "vector" in the SSZ spec.

PR #779, opened 2019-03-15
Move to SHA256

JustinDrake:
SHA256 is de facto blockchain standard. Standardisation of the hash function is a prerequisite for [full standardisation of BLS12-381 signatures](https://github.com/ethereum/eth2.0-specs/issues/605). Blockchain projects are likely to provide a cheap SHA256 opcode/precompile, and unlikely to provide a Keccak256 equivalent. (Even WASM-enabled blockchains are likely to provide a SHA256 opcode/precompile since WASM does *not* natively support optimised SHA256 CPU instructions.) With Ethereum 2.0 embracing SHA256 the wider industry is more likely to converge towards a unified cross-blockchain communication scheme via Merkle receipts.

There are no security blockers with SHA256 (see comments by Dan Boneh [here](https://github.com/ethereum/eth2.0-specs/issues/612#issuecomment-470452562)).

PR #780, opened 2019-03-15
Simplify deposits

### LINKED ISSUE #760: Remove DepositData
Currently `DepositData` is:

```python
{
    # Amount in Gwei
    'amount': 'uint64',
    # Timestamp from deposit contract
    'timestamp': 'uint64',
    # Deposit input
    'deposit_input': DepositInput,
}
```

We already discussed removing `timestamp`. Further, we can put `amount` inside `DepositInput` because (unlike `timestamp`) it is known ahead of time and can be signed over by the `proof_of_possession`.

This would allow us to reduce to two objects for deposits: `Deposit` and `DepositInput`. (We could then rename `DepositInput` to `DepositData` for consistency.)

JustinDrake:
Fix #760

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1271,7 +1271,7 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
 
     # Verify the Merkle branch
     merkle_branch_is_valid = verify_merkle_branch(
-        leaf=hash(serialize(deposit.data)),  # 48 + 32 + 8 + 96 = 184 bytes serialisation
+        leaf=hash(serialize(deposit.data)),  # 48 + 32 + 8 + 96 = 184 bytes serialization
         proof=deposit.proof,
         depth=DEPOSIT_CONTRACT_TREE_DEPTH,
         index=deposit.index,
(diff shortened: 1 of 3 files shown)


JustinDrake:
Opened an issue :)

https://github.com/ethereum/deposit_contract/issues/24

NIC619:
Opened the  PR at 
https://github.com/ethereum/deposit_contract/pull/25 .

I didn't change much: first compute `amount` from `msg.value` then concatenate it with `deposit_input`.
But this differs from the spec in the order of the data in `deposit_data`:
this PR has `amount + pubkey + withdrawal_credentials + proof_of_possession`,
spec has `pubkey + withdrawal_credentials + amount  + proof_of_possession`.
I can change the order if the spec one is preferred.

JustinDrake:
@NIC619 The spec ordering was chosen to be consistent with the `Validator` object which has `pubkey` and `withdrawal_credentials` as the first two fields (those are the two fixed fields).

NIC619:
@JustinDrake Updated in 
https://github.com/ethereum/deposit_contract/pull/25/commits/4c8ad52cb2ef6f86864e9ef862b7f3652352c8d8

djrtwo:
Similar to #779 , @NIC619 I think we should merge https://github.com/ethereum/deposit_contract/pull/25 into `dev` only to be released when this is released via v0.6.0 as to not confuse those targeting v0.5.x

PR #781, opened 2019-03-15
Increase proposer rewards

JustinDrake:
Replacement for #735.


djrtwo:
why not the proportional balance portion of this PR?

JustinDrake:
> why not the proportional balance portion of this PR?

Because of https://github.com/ethereum/eth2.0-specs/pull/772

djrtwo:
Comment from @vbuterin  on the other thread

> Not building on the head is very risky in the current setup because the head would dominate with many attestations very quickly, but I definitely see the rationale in not encouraging trying.

> BTW the need for this patch is arguably removed by the fair proposer selection patch. So if we feel the 1/8 reward is enough we could just close this?

djrtwo:
I believe there was rough consensus to close this?

JustinDrake:
I guess we can review this with the other economic constants.

PR #784, opened 2019-03-15
Milder ejections

JustinDrake:
See item 22 in https://github.com/ethereum/eth2.0-specs/issues/675.

Two reasons for this PR:

1) Send ejections to the exit churn mechanism (partially addressing https://github.com/ethereum/eth2.0-specs/issues/527).
2) Avoid sudden drops in the active validator count. (Somewhat related to 1.)


djrtwo:
@JustinDrake Can you add a line or two in the PR comment on why this is desirable?

JustinDrake:
Two reasons:

1) Send ejections to the exit churn mechanism.
2) Avoid sudden drops in the active validator count. (Somewhat related to 1.)

Issue #527, opened 2019-01-30
Include ejections and penalties when calculating validator churn


JustinDrake:
As title suggests, update churn logic in `update_validator_registry`.

JustinDrake:
This is being addressed in the withdrawal/exit queue refactor https://github.com/ethereum/eth2.0-specs/pull/850

PR #785, opened 2019-03-15
Include recently slashed churn in exit churn queue

JustinDrake:
Addresses #527 in combination with #784.

These changes were made to keep a tighter bound on validator churn to ensure adequate safety margins when changing v-sets

* add the recently slashed balances to the exit churn to prevent too much turn over in the case of slashed exits
* do not process exits at all if too long since last validator registry change to avoid counting errors wrt slashed balance array

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -595,6 +595,7 @@ The types are defined topologically to aid in facilitating an executable version
     'validator_registry': [Validator],
     'validator_balances': ['uint64'],
     'validator_registry_update_epoch': 'uint64',
+    'validator_registry_update_slashed_balances': 'uint64',
 
     # Randomness and committees
     'latest_randao_mixes': ['bytes32', LATEST_RANDAO_MIXES_LENGTH],
@@ -2116,6 +2117,7 @@ def update_validator_registry(state: BeaconState) -> None:
             exit_validator(state, index)
 
     state.validator_registry_update_epoch = current_epoch
+    state.validator_registry_update_slashed_balances = total_at_end
 ```
 
 Run the following function:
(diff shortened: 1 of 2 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -2114,6 +2117,7 @@ def update_validator_registry(state: BeaconState) -> None:
             exit_validator(state, index)
 
     state.validator_registry_update_epoch = current_epoch
+    state.validator_registry_update_slashed_balances = total_at_end
```
vbuterin: Another way to do this that does not require a new variable is as follows:

```
if state.validator_registry_update_epoch > state.current_epoch - LATEST_SLASHED_EXIT_LENGTH:
    total_at_start = state.latest_slashed_balances[state.validator_registry_update_epoch % LATEST_SLASHED_EXIT_LENGTH]
    total_at_end = state.latest_slashed_balances[current_epoch % LATEST_SLASHED_EXIT_LENGTH]
    balance_churn = total_at_end - total_at_start
        if validator.exit_epoch == FAR_FUTURE_EPOCH and validator.initiated_exit:
            # Check the balance churn would be within the allowance
            balance_churn += get_effective_balance(state, index)
            if balance_churn > max_balance_churn:
                break

            # Exit validator
            exit_validator(state, index)

```

specs/core/0_beacon-chain.md
```diff
@@ -2102,16 +2102,21 @@ def update_validator_registry(state: BeaconState) -> None:
...
 
     # Exit validators within the allowable balance churn
-    balance_churn = 0
-    for index, validator in enumerate(state.validator_registry):
-        if validator.exit_epoch == FAR_FUTURE_EPOCH and validator.initiated_exit:
-            # Check the balance churn would be within the allowance
-            balance_churn += get_effective_balance(state, index)
-            if balance_churn > max_balance_churn:
-                break
+    if state.current_epoch < state.validator_registry_update_epoch + LATEST_SLASHED_EXIT_LENGTH:
```
djrtwo: Why are you blocking exits if there have been `LATEST_SLASHED_EXIT_LENGTH` epochs since last registry update?

JustinDrake: See discussing [here](https://github.com/ethereum/eth2.0-specs/pull/785#discussion_r266187726). It's a simplification to avoid caching a `slashed_balance`.

PR #792, opened 2019-03-17
Set `GENESIS_FORK_VERSION` to a `bytes4` constant

hwwhww:
#### Issue
It's suboptimal that we have to convert `GENESIS_FORK_VERSION` with `int_to_bytes4(GENESIS_FORK_VERSION)` everytime we use it.

#### Solution
We define that a "fork version" is a `bytes4` object universally.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -208,7 +208,7 @@ Code snippets appearing in `this style` are to be interpreted as Python code.
 
 | Name | Value |
 | - | - |
-| `GENESIS_FORK_VERSION` | `0` |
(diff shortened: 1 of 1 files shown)


JustinDrake:
Agreed this is an improvement :)

FYI, @djrtwo disagrees with us—see [here](https://github.com/ethereum/eth2.0-specs/pull/764). I still think this is the way forward :)

hwwhww:
@JustinDrake @djrtwo 
Whoops! Sorry for that I missed that #764 was there!
If human-readability is a problem, we *could* have both `ForkVersion: bytes4` and `ForkNumber: uint32`.

PR #794, opened 2019-03-17
Rename "tuple" to "vector" in SSZ spec

JustinDrake:
To be done in combination with #777.

Also:

* Define "fixed-size" and "variable-size" more rigorously
* Use `"` vs `'` consistently
* Add missing `"`

### DIFF
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -12,7 +12,7 @@ This is a **work in progress** describing typing, serialization and Merkleizatio
 - [Serialization](#serialization)
     - [`"uintN"`](#uintn)
     - [`"bool"`](#bool)
-    - [Tuples, containers, lists](#tuples-containers-lists)
+    - [Vectors, containers, lists](#vectors-containers-lists)
 - [Deserialization](#deserialization)
 - [Merkleization](#merkleization)
 - [Self-signed containers](#self-signed-containers)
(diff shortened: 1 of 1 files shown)


pipermerriam:
I know this was already merged, but what is the motivation on using `vector` instead of `tuple`?  The `tuple` language lines up nicely with the concept that already exists in the ABI encoding spec and python's concept of Tuple types.

pipermerriam:
Was `struct` considered?  Seems like "Struct" would likely carry the most semantic meaning across many languages.

JustinDrake:
See https://github.com/ethereum/eth2.0-specs/pull/777. Basically we want to convey an "ordered fixed-length homogeneous collection of values" in one word. "tuple", "array", "vector", etc. were considered. Python tuples are not homogeneous, and using `Tuple` may cause a clash with type hinting since we want to be explicit about the length.

`struct` would be fine for containers. (That's the terminology used in Go.)

pipermerriam:
:+1: I think struct is significantly better than vector, I'll create a PR today unless someone is opposed.

JustinDrake:
Wait, `struct` is good for ordered *heterogenous* collection of values, right? Vector is for homogeneous stuff.

It's unfortunate there's a bikeshedding surface here. At the end of the day, the spec mostly need to be consistent and implementers will use whatever terminology is most appropriate for their language. (E.g. in Go if would be `struct`, `slice`, `array` instead of `container`, `vector`, `list`.)

pipermerriam:
blarg, ignore me, shouldn't comment on pull requests first thing in the morning before shower and coffee.

PR #796, opened 2019-03-18
Use `EMPTY_SIGNATURE` for temporary block header

paulhauner:
## What

Changes `get_temporary_block_header(..)` to use `EMPTY_SIGNATURE` instead of `block.signature`.

## Why

`state.latest_block_header.signature` is different between producing and processing a block.

### Producing

### Processing


paulhauner:
Now I run this on more blocks I can see that this PR doesn't fix this. I'm going to move this to #797.

PR #798, opened 2019-03-18
Replace previous_block_root with previous_block_signature

### LINKED ISSUE #797: Block header signature issue
`state.latest_block_header.signature` is different between producing and processing a block and this is causing block processing to fail because of a mismatch on `block.state_root`.

### Producing

### Processing

JustinDrake:
Possible approach to fix #797. (Not lobbying for this, merely putting it up as an option.)

Note that the extra `bls_verify` can be optimised away by implementers.


JustinDrake:
Hum, not sure this works because `latest_block_roots` cannot be populated without the signature...

PR #799, opened 2019-03-18
Remove latest_block_header

### LINKED ISSUE #797: Block header signature issue

JustinDrake:
Fix #797:

* Remove `latest_block_header`
* Maintain `(pre_state, block) -> (post_state, error)` state transition function by caching the block root *after* checking the state root

```python
def final_updates(state: BeaconState, block: BeaconBlock) -> None:
    # Check state root
    assert block.state_root == hash_tree_root(state)
    # Cache block root
    state.latest_block_roots[state.slot % SLOTS_PER_HISTORICAL_ROOT] = hash_tree_root(block)
```

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -622,7 +622,6 @@ The types are defined topologically to aid in facilitating an executable version
     'latest_state_roots': ['bytes32', SLOTS_PER_HISTORICAL_ROOT],
     'latest_active_index_roots': ['bytes32', LATEST_ACTIVE_INDEX_ROOTS_LENGTH],
     'latest_slashed_balances': ['uint64', LATEST_SLASHED_EXIT_LENGTH],  # Balances slashed at every withdrawal period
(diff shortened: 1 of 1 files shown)


djrtwo:
This has the same problem as described in #797. 

When "producing" a block, you don't set `block.state_root` until entirely out of the state transition function (and just avoid this state_root validity check) so at this point _within_ the state transition function the `state_root` is not yet set in the block.

My intuition is telling me `(pre_state, prev_block_root, block) -> post_state` is the way to go.

djrtwo:
I see, you're implying that this happens "between" two state transitions... 
I think just being explicit about the oracle is cleaner.

vbuterin:
I don't like this either. I remember implementing this exact fix years ago when doing an early Casper PoC, and it just led to a lot of confusion and client complexity around "signing post-state root" vs "real post-state root".

PR #800, opened 2019-03-18
ci tests

djrtwo:
Ports testing and spec build script from research repo. Note this PR is against master and will be backported into dev

Things:
* build scrip in `scripts/phase0`
* added Makefile to build executable spec and compile dependencies
* dependencies and stubs added to `utils/phase0`. This includes minimal ssz, merklization, etc.
* tests
  * ported `sanity_check` from research repo. no additional tests
  * two "marks" exist -- `sanity` and `minimal_config` -- for running quick tests locally. Run these limited marks via `make test`
  * CI runs full tests against minimal config and default config. If this gets out of hand, we can strip it down and just have releases have to pass the full tests.
  * note: these tests currently bypass the bug found in #797 by considering the root of a block to be the `hash_tree_root` of the block with EMPTY_SIGNATURE. Will update these tests depending on which way #797 goes.

Note, this will probably get a round or two of cleanups but wanted to get this up for merge now because its a blocker on all other PRs


utils/phase0/state_transition.py
```diff
@@ -0,0 +1,100 @@
...
+def state_transition(state: BeaconState,
+                     block: BeaconBlock,
+                     verify_state_root: bool=False) -> BeaconState:
+    while state.slot < block.slot:
+        spec.cache_state(state)
+        if (state.slot + 1) % spec.SLOTS_PER_EPOCH == 0:
+            process_epoch_transition(state)
+        spec.advance_slot(state)
+        if block.slot == state.slot:
+            process_block(state, block)
```
hwwhww: It seems `verify_state_root` should be passed to `process_block`:
```suggestion
            process_block(state, block, verify_state_root)
```

p.s. however, `verify_state_root` is never set to `True` in the current tests.

tests/phase0/helpers.py
```diff
@@ -0,0 +1,144 @@
...
+    deposit_data_list = []
+    for i in range(num_validators):
+        pubkey = pubkeys_list[i]
+        privkey = pubkey_to_privkey[pubkey]
+        deposit_data = DepositData(
+            amount=spec.MAX_DEPOSIT_AMOUNT,
+            timestamp=deposit_timestamp,
+            deposit_input=DepositInput(
+                pubkey=pubkey,
+                withdrawal_credentials=privkey.to_bytes(32, byteorder='big'),
```
hwwhww: It seems incorrect or just a stub. Assume that we use validator pubkeys as the withdrawal pubkeys, than it can be:

```python
        withdrawal_credentials=BLS_WITHDRAWAL_PREFIX_BYTE  + hash(pubkey)[1:],
```

or add a comment to note that it's a stub and use a made-up value e.g., `b'\x12' * 32`.

I prefer to use `BLS_WITHDRAWAL_PREFIX_BYTE` here.

djrtwo: Seems like a couple of the teams (lighthouse & nimbus) put the privkey in withdrawal credentials as a convention. To do so validly, needed to go bigint to ensure the 0-byte for these lower values.

It doesn't actually buy us anything really. it might just be better to assume that clients have a rainbow table for the first 100k pubkeys to do reverse lookup

PR #807, opened 2019-03-19
Avoid underflow in voluntary exits

JustinDrake:


JustinDrake:
I went with `assert is_active_validator`.

PR #808, opened 2019-03-19
Replace committee exponential backoff with max progress

vbuterin:
Removes the mechanism that only rotates committees if blocks have been finalized and every shard has been crosslinked or at exponentially decreasing intervals, and replaces it with a rule that shard committees can only progress a maximum of 64 epochs at a time to preserve the invariant that maximum possible work required per epoch for a validator is O(1).

This also fixes a bug in `process_crosslink` regarding the use of `previous_epoch`.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -74,7 +74,6 @@
         - [`get_beacon_proposer_index`](#get_beacon_proposer_index)
         - [`verify_merkle_branch`](#verify_merkle_branch)
         - [`get_attestation_participants`](#get_attestation_participants)
-        - [`is_power_of_two`](#is_power_of_two)
         - [`int_to_bytes1`, `int_to_bytes2`, ...](#int_to_bytes1-int_to_bytes2-)
         - [`bytes_to_int`](#bytes_to_int)
         - [`get_effective_balance`](#get_effective_balance)
@@ -861,16 +860,12 @@ def get_crosslink_committees_at_slot(state: BeaconState,
     ))
 
     if epoch == current_epoch:
-        shuffling_start_shard = state.latest_start_shard
+        start_shard = state.latest_start_shard
     elif epoch == previous_epoch:
-        shuffling_start_shard = (
-            state.latest_start_shard - SLOTS_PER_EPOCH * committees_per_epoch
-        ) % SHARD_COUNT
+        start_shard = (state.latest_start_shard - SLOTS_PER_EPOCH * committees_per_epoch) % SHARD_COUNT
     elif epoch == next_epoch:
         current_epoch_committees = get_current_epoch_committee_count(state)
-        shuffling_start_shard = (
-            state.latest_start_shard + EPOCH_LENGTH * current_epoch_committees
-        ) % SHARD_COUNT
+        start_shard = (state.latest_start_shard + EPOCH_LENGTH * current_epoch_committees) % SHARD_COUNT
 
     shuffling = get_shuffling(
         generate_seed(state, epoch),
@@ -879,7 +874,7 @@ def get_crosslink_committees_at_slot(state: BeaconState,
     )
     offset = slot % SLOTS_PER_EPOCH
     committees_per_slot = committees_per_epoch // SLOTS_PER_EPOCH
-    slot_start_shard = (shuffling_start_shard + committees_per_slot * offset) % SHARD_COUNT
+    slot_start_shard = (start_shard + committees_per_slot * offset) % SHARD_COUNT
 
     return [
         (
@@ -1015,16 +1010,6 @@ def get_attestation_participants(state: BeaconState,
         if aggregation_bit == 0b1:
             participants.append(validator_index)
     return participants
-```
-
-### `is_power_of_two`
-
-```python
-def is_power_of_two(value: int) -> bool:
-    """
-    Check if ``value`` is a power of two integer.
-    """
-    return (value > 0) and (value & (value - 1) == 0)
 ```
 
 ### `int_to_bytes1`, `int_to_bytes2`, ...
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -185,8 +185,10 @@ Code snippets appearing in `this style` are to be interpreted as Python code.
 | `MAX_INDICES_PER_SLASHABLE_VOTE` | `2**12` (= 4,096) |
 | `MAX_EXIT_DEQUEUES_PER_EPOCH` | `2**2` (= 4) |
 | `SHUFFLE_ROUND_COUNT` | 90 |
+| `MAX_CROSSLINK_EPOCHS` | `2**6` (= 64) |
 
 * For the safety of crosslinks `TARGET_COMMITTEE_SIZE` exceeds [the recommended minimum committee size of 111](https://vitalik.ca/files/Ithaca201807_Sharding.pdf); with sufficient active validators (at least `SLOTS_PER_EPOCH * TARGET_COMMITTEE_SIZE`), the shuffling algorithm ensures committee sizes of at least `TARGET_COMMITTEE_SIZE`. (Unbiasable randomness with a Verifiable Delay Function (VDF) will improve committee robustness and lower the safe minimum committee size.)
+* `MAX_CROSSLINK_EPOCHS` should be a small constant times `SHARD_COUNT // EPOCH_LENGTH`
 
 ### Deposit contract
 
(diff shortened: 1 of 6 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -2119,30 +2059,17 @@ def update_validator_registry(state: BeaconState) -> None:
...
 
 ```python
-def update_registry_and_shuffling_data(state: BeaconState) -> None:
-    # First set previous shuffling data to current shuffling data
-    state.previous_shuffling_epoch = state.current_shuffling_epoch
-    state.previous_shuffling_start_shard = state.current_shuffling_start_shard
-    state.previous_shuffling_seed = state.current_shuffling_seed
+def update_registry(state: BeaconState) -> None:
     next_epoch = current_epoch + 1
```
JustinDrake: This line can go

specs/core/0_beacon-chain.md
```diff
@@ -2119,30 +2059,17 @@ def update_validator_registry(state: BeaconState) -> None:
...
-        state.current_shuffling_seed = generate_seed(state, state.current_shuffling_epoch)
-    else:
-        # If processing at least one crosslink keeps failing, then reshuffle every power of two,
-        # but don't update the current_shuffling_start_shard
-        epochs_since_last_registry_update = current_epoch - state.validator_registry_update_epoch
-        if epochs_since_last_registry_update > 1 and is_power_of_two(epochs_since_last_registry_update):
-            state.current_shuffling_epoch = next_epoch
-            state.current_shuffling_seed = generate_seed(state, state.current_shuffling_epoch)
+        # If we update the registry, update the shuffling data 2/3 or and shards as well
+    state.current_shuffling_start_shard = (
```
JustinDrake: Indentation doesn't match comment

specs/core/0_beacon-chain.md
```diff
@@ -2119,30 +2059,17 @@ def update_validator_registry(state: BeaconState) -> None:
...
-        )
-        state.current_shuffling_seed = generate_seed(state, state.current_shuffling_epoch)
-    else:
-        # If processing at least one crosslink keeps failing, then reshuffle every power of two,
-        # but don't update the current_shuffling_start_shard
-        epochs_since_last_registry_update = current_epoch - state.validator_registry_update_epoch
-        if epochs_since_last_registry_update > 1 and is_power_of_two(epochs_since_last_registry_update):
-            state.current_shuffling_epoch = next_epoch
-            state.current_shuffling_seed = generate_seed(state, state.current_shuffling_epoch)
+        # If we update the registry, update the shuffling data 2/3 or and shards as well
```
JustinDrake: Don't understand "update the shuffling data 2/3 or and shards as well"

specs/core/0_beacon-chain.md
```diff
@@ -2397,7 +2324,8 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
         # the attestation is trying to create
         Crosslink(
             crosslink_data_root=attestation.data.crosslink_data_root,
-            epoch=slot_to_epoch(attestation.data.slot)
+            epoch=min(slot_to_epoch(attestation.data.slot),
```
JustinDrake: Should this be replaced with `assert slot_to_epoch(attestation.data.slot) <= attestation.data.previous_crosslink.epoch + MAX_CROSSLINK_EPOCHS`? The reason is that the newly created crosslink will have a mismatch between `crosslink_data_root` and `epoch`.

specs/core/0_beacon-chain.md
```diff
@@ -2119,30 +2059,17 @@ def update_validator_registry(state: BeaconState) -> None:
...
-        # If processing at least one crosslink keeps failing, then reshuffle every power of two,
-        # but don't update the current_shuffling_start_shard
-        epochs_since_last_registry_update = current_epoch - state.validator_registry_update_epoch
-        if epochs_since_last_registry_update > 1 and is_power_of_two(epochs_since_last_registry_update):
-            state.current_shuffling_epoch = next_epoch
-            state.current_shuffling_seed = generate_seed(state, state.current_shuffling_epoch)
+        # If we update the registry, update the shuffling data 2/3 or and shards as well
+    state.current_shuffling_start_shard = (
+        state.current_shuffling_start_shard +
+        get_current_epoch_committee_count(state) % SHARD_COUNT
```
JustinDrake: The `% SHARD_COUNT` should go one line down, after the closing bracket `)`.

specs/core/0_beacon-chain.md
```diff
@@ -1529,12 +1483,7 @@ def get_genesis_beacon_state(genesis_validator_deposits: List[Deposit],
 
         # Randomness and committees
         latest_randao_mixes=Vector([ZERO_HASH for _ in range(LATEST_RANDAO_MIXES_LENGTH)]),
-        previous_shuffling_start_shard=GENESIS_START_SHARD,
         current_shuffling_start_shard=GENESIS_START_SHARD,
```
JustinDrake: Suggest renaming `current_shuffling_start_shard` to `shuffling_start_shard`.

specs/core/0_beacon-chain.md
```diff
@@ -2119,30 +2059,17 @@ def update_validator_registry(state: BeaconState) -> None:
...
-def update_registry_and_shuffling_data(state: BeaconState) -> None:
-    # First set previous shuffling data to current shuffling data
-    state.previous_shuffling_epoch = state.current_shuffling_epoch
-    state.previous_shuffling_start_shard = state.current_shuffling_start_shard
-    state.previous_shuffling_seed = state.current_shuffling_seed
+def update_registry(state: BeaconState) -> None:
     current_epoch = get_current_epoch(state)
     next_epoch = current_epoch + 1
     # Check if we should update, and if so, update
     if should_update_validator_registry(state):
```
JustinDrake: Consider deleting `should_update_validator_registry` and replacing this line with `state.finalized_epoch > state.validator_registry_update_epoch`.

specs/core/0_beacon-chain.md
```diff
@@ -185,8 +183,10 @@ Code snippets appearing in `this style` are to be interpreted as Python code.
 | `MAX_INDICES_PER_SLASHABLE_VOTE` | `2**12` (= 4,096) |
 | `MAX_EXIT_DEQUEUES_PER_EPOCH` | `2**2` (= 4) |
 | `SHUFFLE_ROUND_COUNT` | 90 |
+| `MAX_CROSSLINK_EPOCHS` | `2**6` (= 64) |
```
JustinDrake: Move to "Time parameters".

specs/core/0_beacon-chain.md
vbuterin: No because `attestation.data.slot` is meant to reflect the _current_ slot, not the slot of the end of the crosslink it is creating. The epoch of the end of the crosslink is specified by `min(slot_to_epoch(attestation.data.slot), attestation.latest_crosslink.epoch + MAX_CROSSLINK_EPOCHS)` so we don't need a separate variable for it.

JustinDrake:
Can we get rid of `state.validator_registry_update_epoch` and instead rely on `state.finalized_epoch`?

(Loving the simplifications in this PR!)

djrtwo:
I don't think you can just get rid of `previous` and `current_shuffling_epoch`. In the case of non-finality, your v-set needs to remain stable against some epoch in the past and we need to remember two epochs worth of the shuffling epoch.

vbuterin:
> I don't think you can just get rid of `previous` and `current_shuffling_epoch`. In the case of non-finality, your v-set needs to remain stable against some epoch in the past and we need to remember two epochs worth of the shuffling epoch.

Your _validator set_ needs to stay stable (at least until I write the withdrawal queue -> exit queue change), but the _shuffling_ does not.

djrtwo:
> Your validator set needs to stay stable (at least until I write the withdrawal queue -> exit queue change), but the shuffling does not.

Right, but your validator set won't stay stable in this method. [`get_shuffling`](https://github.com/ethereum/eth2.0-specs/blob/b50e148642d4a19d5517ab1ab689708b33ed7b53/specs/core/0_beacon-chain.md#get_shuffling) pulls the active validator set from the epoch passed in. Validator activations also have a 4 epoch delay. So if the epoch that you are getting your v-set from keeps incrementing even without finality, then your v-set can change without finality.

vbuterin:
> EDIT: You are also going to run into issues doing the randomness dynamically. If there are short range forks, then your shuffling could be different which would lead to unslashable attestations across forks.

This is why slashing uses SlashableAttestations (possibly being renamed StandaloneAttestations) instead of using regular attestations; that solves this problem. Note that the need to use SlashableAttestations is unavoidable anyway because even in the current system, we reshuffle with exponential backoff even if we don't finalize, and you can get finality breaks from equivocations between two different epochs.

Shuffling is relevant only to crosslinks and slot selection; it has near-zero relevance to the Casper finalization mechanism, which treats all active validators in each epoch identically (not exactly zero because if you're close to the end of an epoch it's more likely your attestation will have to wait an extra epoch to be processed, but that's a small detail...)

PR #814, opened 2019-03-20
fix and extend header tests

djrtwo:
The previous header test for slashed proposer was not setup properly and was passing incorrectly. Fixed that and added more quick tests on header

### DIFF
--- a/tests/phase0/block_processing/test_process_block_header.py
+++ b/tests/phase0/block_processing/test_process_block_header.py
@@ -4,6 +4,8 @@ import pytest
 
 from build.phase0.spec import (
     get_beacon_proposer_index,
+    cache_state,
+    advance_slot,
     process_block_header,
 )
 from tests.phase0.helpers import (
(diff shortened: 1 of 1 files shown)


tests/phase0/block_processing/test_process_block_header.py
```diff
@@ -14,13 +16,93 @@
...
+    advance_slot(pre_state)
+
+    post_state = deepcopy(pre_state)
+
+    #
+    # test block header
+    #
+    process_block_header(post_state, block)
+
+    return state, [block], post_state
```
hwwhww: Pythonista speaking, it's a little weird to see that no assertion in this testcase. These three cases (`test_sucess`, `test_invalid_slot `, `test_invalid_previous_block_root`) can be combined into one testcase + maybe some setup. 😄

tests/phase0/block_processing/test_process_block_header.py
```diff
@@ -14,13 +16,93 @@
...
+
+
+def test_invalid_slot(state):
+    pre_state = deepcopy(state)
+
+    # mess up previous block root
+    block = build_empty_block_for_next_slot(pre_state)
+    block.previous_block_root = b'\12'*32
```
hwwhww: It seems should only be set in the `test_invalid_previous_block_root` case?

tests/phase0/block_processing/test_process_block_header.py
djrtwo: ah right. it should for sure

tests/phase0/block_processing/test_process_block_header.py
djrtwo: `build_empty_block_for_next_slot` builds a valid block including the previous_block_root. Just overriding in the case where we want to trigger an error.

hwwhww: But the only difference between `test_invalid_slot` and `test_invalid_previous_block_root` is that `test_invalid_slot` has one more `advance_slot(pre_state)`; and they both throw `AssertionError`. So we are not able to see what's the root cause that triggers the exception in `test_invalid_slot` case.

djrtwo: I missed the error you were actually pointing out. Fixed!

tests/phase0/block_processing/test_process_block_header.py
djrtwo: On second thought.. I'm not sure what we would assert. `process_block_header` actually has no side effects.

PR #826, opened 2019-03-21
Simplify justification and finalization accounting logic

vbuterin:
Much of the simplification is cosmetic. The following changes are substantive:

* Inactivity leak penalty specifically on missing the target, not both the target and the source
* Even outside of inactivity leak scenarios, slashing victims suffer offline penalties
* Max base-reward penalties in case of inactivity leak scenarios are greater

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1886,7 +1886,7 @@ def get_inactivity_penalty(state: BeaconState, index: ValidatorIndex, epochs_sin
     if epochs_since_finality <= 4:
         extra_penalty = 0
     else:
-        extra_penalty = get_effective_balance(state, index) * min(epochs_since_finality // INACTIVITY_PENALTY_QUOTIENT // 2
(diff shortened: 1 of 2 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1883,10 +1883,11 @@ def get_base_reward(state: BeaconState, index: ValidatorIndex) -> Gwei:
...
 ```python
 def get_inactivity_penalty(state: BeaconState, index: ValidatorIndex, epochs_since_finality: int) -> Gwei:
-    return (
-        get_base_reward(state, index) +
-        get_effective_balance(state, index) * epochs_since_finality // INACTIVITY_PENALTY_QUOTIENT // 2
-    )
+    if epochs_since_finality <= 4:
+        extra_penalty = 0
+    else:
+        extra_penalty = get_effective_balance(state, index) * min(epochs_since_finality // INACTIVITY_PENALTY_QUOTIENT // 2
```
hwwhww: Unfinished line. What's the second argument for `min()` here?

PR #831, opened 2019-03-22
Merge attestation verification logic

vbuterin:
Also rename slashable attestation to indexed attestation to reflect its broader functionality in phase 1.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -77,6 +77,7 @@
         - [`generate_seed`](#generate_seed)
         - [`get_beacon_proposer_index`](#get_beacon_proposer_index)
         - [`verify_merkle_branch`](#verify_merkle_branch)
+        - [`get_crosslink_committee_for_attestation`](#get_crosslink_committee_for_attestation)
         - [`get_attestation_participants`](#get_attestation_participants)
         - [`int_to_bytes1`, `int_to_bytes2`, ...](#int_to_bytes1-int_to_bytes2-)
         - [`bytes_to_int`](#bytes_to_int)
@@ -85,6 +86,7 @@
         - [`get_fork_version`](#get_fork_version)
         - [`get_domain`](#get_domain)
         - [`get_bitfield_bit`](#get_bitfield_bit)
+        - [`set_bitfield_bit`](#set_bitfield_bit)
         - [`verify_bitfield`](#verify_bitfield)
         - [`convert_to_indexed`](#convert_to_indexed)
         - [`verify_indexed_attestation`](#verify_indexed_attestation)
@@ -1037,6 +1039,20 @@ def verify_merkle_branch(leaf: Bytes32, proof: List[Bytes32], depth: int, index:
     return value == root
 ```
 
+### `get_crosslink_committee_for_attestation`
+
+```python
+def get_crosslink_committee_for_attestation(state: BeaconState,
+                                        attestation_data: AttestationData) -> List[ValidatorIndex]:
+    # Find the committee in the list with the desired shard
+    crosslink_committees = get_crosslink_committees_at_slot(state, attestation_data.slot)
+
+    assert attestation_data.shard in [shard for _, shard in crosslink_committees]
+    crosslink_committee = [committee for committee, shard in crosslink_committees if shard == attestation_data.shard][0]
+
+    return crosslink_committee
+```
+
 ### `get_attestation_participants`
 
 ```python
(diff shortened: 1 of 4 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1214,30 +1215,44 @@ def verify_bitfield(bitfield: bytes, committee_size: int) -> bool:
...
 
-### `verify_slashable_attestation`
+### `convert_to_standalone`
 
 ```python
-def verify_slashable_attestation(state: BeaconState, slashable_attestation: SlashableAttestation) -> bool:
+def convert_to_standalone(state: BeaconState, attestation: Attestation):
     """
-    Verify validity of ``slashable_attestation`` fields.
+    Converts an attestation to (almost) standalone-verifiable form
```
hwwhww: Nitpick: to align with the rest of this spec:
```suggestion
    Convert an attestation to (almost) standalone-verifiable form
```

vbuterin:
I still like `StandaloneAttestation`; technically both are verifiable, one is just verifiable with less auxiliary info than the other :smile:

Let's argue about it in the next call?

specs/core/0_beacon-chain.md
```diff
@@ -1214,30 +1215,44 @@ def verify_bitfield(bitfield: bytes, committee_size: int) -> bool:
...
     """
-    Verify validity of ``slashable_attestation`` fields.
+    Convert an attestation to (almost) standalone-verifiable form
     """
-    if slashable_attestation.custody_bitfield != b'\x00' * len(slashable_attestation.custody_bitfield):  # [TO BE REMOVED IN PHASE 1]
-        return False
+    return StandaloneAttestation(
+        validator_indices=get_attestation_participants(state, attestation.data, attestation.aggregation_bitfield),
+        data=attestation.data,
+        custody_bitfield=attestation.custody_bitfield,
```
hwwhww: To confirm with @vbuterin:

1. I *assumed* that before this conversion, the committee size of this `Attestation.custody_bitfield` is the committee of the given crosslink committee - but actually, the size of `Attestation.custody_bitfield` is not restricted in the `dev` branch.
2. And after this convertion, in `verify_standalone_attestation`, it expects that the size of `StandaloneAttestation.custody_bitfield` is `len(StandaloneAttestation.validator_indices)`?

The benefit of having `len(Attestation.custody_bitfield) == len(Attestation.aggregation_bitfield) == len(crosslink_committee)` is that we can simply merge the bitfields with `or` operator when aggregating 
 many `Attestation`s. If the size of `len(Attestation.custody_bitfield) != len(crosslink_committee)`, it seems it would add additional overhead to generate the new `custody_bitfield`?

vbuterin: > I assumed that before this conversion, the committee size of this Attestation.custody_bitfield is the committee of the given crosslink committee - but actually, the size of Attestation.custody_bitfield is not restricted in the dev branch.

An attestation committee will not exceed 4096 already because that's the "all ETH is validating" threshold. In any case, having the same threshold in both regular attestations and standalone attestations seems necessary for slashing to be possible in any case.

> it expects that the committee
size of StandaloneAttestation.custody_bitfield is len(StandaloneAttestation.validator_indices)?

You are right that the proposed change does increase strictness because the current approach requires the custody bitfield to be _any_ quantity of zero bytes. I would argue again that the current approach is wrong and the new approach is correct.

JustinDrake:
Brainstorming names:

* `IndexedAttestation` (suggesting that attester indices are provided)
* `ExternalAttestation` (suggesting the attestation can come from an external fork)
* `ForkAttestation` <-- my favourite

specs/core/0_beacon-chain.md
```diff
@@ -1147,30 +1148,44 @@ def verify_bitfield(bitfield: bytes, committee_size: int) -> bool:
...
     """
-    Verify validity of ``slashable_attestation`` fields.
+    Convert an attestation to (almost) indexed-verifiable form
     """
-    if slashable_attestation.custody_bitfield != b'\x00' * len(slashable_attestation.custody_bitfield):  # [TO BE REMOVED IN PHASE 1]
-        return False
+    return IndexedAttestation(
+        validator_indices=get_attestation_participants(state, attestation.data, attestation.aggregation_bitfield),
+        data=attestation.data,
+        custody_bitfield=attestation.custody_bitfield,
```
djrtwo: This `custody_bitfield` here is incorrect. We need to truncate the bitfield to be the `ceildiv8`of the size of the participants (`validator_indices`) and if we sort the `validator_indices`, the newly constructed custody_bitfield needs to reflect this sort order.

djrtwo: Otherwise it:
* fails `verify_bitfield` because the length of `validator_indices` can easily be less than the length of the original committee 
* even if all validators participated the custody_bitfield is then in the incorrect order wrt `validator_indices`

vbuterin: The best I've thought of so far is to modify `IndexedAttestation` so that it has `bit_0_validator_indices` and `bit_1_validator_indices`, and get rid of the custody bitfield.

djrtwo: That will make `IndexedAttestation` inside of `AttesterSlashing` almost 2x in size.
Did you check out the fix I put in place?

vbuterin: > That will make IndexedAttestation inside of AttesterSlashing almost 2x in size.

Why would it? The total size of the two lists would still equal the size of the original list.

vbuterin: I don't like the fix; adding two whole new functions in a PR that's supposed to be simplifying things feels very not-worth-it.

djrtwo: whoops, missed that you wanted to get rid of custody bitfield. 
Making the change, thanks!

PR #834, opened 2019-03-22
Remove timestamp

arnetheduck:


hwwhww:
It seems duplicate to #780.

arnetheduck:
whoops, didn't see that one. thanks @hwwhww !

PR #839, opened 2019-03-25
cleanup tests to more easily dump output

djrtwo:
Did a little bit of cleaning of tests to more easily manually output test vectors until we get the CI test gen flow in place. The main change is to ensure that the only required arg for tests is an initial `state`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /.pytest_cache
 
 build/
+output/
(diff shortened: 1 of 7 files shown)


djrtwo:
Agreed. We'll be able to wrap the pytest functions no problem to output vectors when run with `pytest --output-yaml` or something, and in that case, because the functions will be run in the context of pytest, the fixtures can be used freely.

PR #841, opened 2019-03-26
Missed filter condition in get_winning_root_and_participants

mkalinin:
### What's done
Adds an extra condition to filter attesters by `shard` passed to `get_winning_root_and_participants`.

### Why
Otherwise `get_winning_root_and_participants` may return validator indices that were not assigned to the committee of given `shard`.

Spec v0.4.0 has that condition either:
> Let `attesting_validator_indices(crosslink_committee, crosslink_data_root)` be the union of the validator index sets given by `[get_attestation_participants(state, a.data, a.aggregation_bitfield) for a in current_epoch_attestations + previous_epoch_attestations if a.data.shard == shard and a.data.crosslink_data_root == crosslink_data_root]`.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1800,7 +1800,8 @@ def get_previous_epoch_matching_head_attestations(state: BeaconState) -> List[Pe
 def get_winning_root_and_participants(state: BeaconState, shard: Shard) -> Tuple[Bytes32, List[ValidatorIndex]]:
     all_attestations = state.current_epoch_attestations + state.previous_epoch_attestations
     valid_attestations = [
-        a for a in all_attestations if a.data.previous_crosslink == state.latest_crosslinks[shard]
+        a for a in all_attestations 
+        if a.data.shard == shard and a.data.previous_crosslink == state.latest_crosslinks[shard]
     ]
     all_roots = [a.data.crosslink_data_root for a in valid_attestations]
 


mkalinin:
There is an edge case with `GENESIS_EPOCH` where this condition is not enough in order to get participants correctly. This case takes place when `process_crosslinks` runs through slots of `previous_epoch = GENESIS_EPOCH - 1`:
1. If `slot < GENESIS_SLOT` then `get_crosslink_committees_at_slot` returns tuples with empty committees but with same shard numbers as for `slot >= GENESIS_SLOT`.
2. Then `get_winning_root_and_participants` returns non-empty participants for slots that nobody attested to, i.e. `slot < GENESIS_SLOT`.

I am not sure what would be the best fix here: 
1. Do not run through slots preceding `GENESIS_SLOT` in `process_crosslinks`.
2. Check participants against crosslink committees indices.
3. Return empty tuple list if `get_crosslink_committees_at_slot` is called with `slot < GENESIS_SLOT`.

(2) sorts things out in a strict way and looks like the better solution than (1).
Is there anything against a fix in `get_crosslink_committees_at_slot`? Potentially, it could prevent issues related to committees of `epoch = GENESIS_EPOCH - 1` in the other places.

JustinDrake:
I think you're right that we need extra check in phase 0 where `crosslink_data_root == ZERO_HASH` for all shards. In phase 1 we won't need this, but it doesn't hurt.

djrtwo:
closing and reopening to trigger ci build

mkalinin:
Closing it in favour of https://github.com/ethereum/eth2.0-specs/pull/855

PR #843, opened 2019-03-26
Make use of object initial values

JustinDrake:
* 71 lines removed in phase 0 if `GENESIS_SLOT != 0`
* 92 lines removed in phase 0 if `GENESIS_SLOT == 0` (upcoming PR from @djrtwo)

Also bunch of lines also removed in phase 1. Golang has such initial values baked into the language.

Screenshot 2019-03-26 at 20 26 41


djrtwo:
Internally debating this.
Do you plan to define initial values for each type somewhere?

JustinDrake:
> Do you plan to define initial values for each type somewhere?

Yes, we can define initial values immediately after defining types. It's a one-liner: "Recursively initialize objects using `0` for `"uintN"`, `False` for `"bool"`, and `[]` for lists."

JustinDrake:
Close in favour of #963

PR #850, opened 2019-03-28
Withdrawal queue -> exit queue

vbuterin:


specs/core/0_beacon-chain.md
```diff
@@ -1973,80 +1963,43 @@ def apply_rewards(state: BeaconState) -> None:
...
 ```python
 def process_ejections(state: BeaconState) -> None:
     """
     Iterate through the validator registry
-    and eject active validators with balance below ``EJECTION_BALANCE``.
+    and deposit or eject active validators with sufficiently high or low balances
     """
-    for index in get_active_validator_indices(state.validator_registry, get_current_epoch(state)):
-        if get_balance(state, index) < EJECTION_BALANCE:
+    for index, validator in enumeratE(state.validator_registry):
```
JustinDrake: `enumerate` (not `enumeratE`)

specs/core/0_beacon-chain.md
```diff
@@ -418,14 +415,14 @@ The types are defined topologically to aid in facilitating an executable version
     'pubkey': 'bytes48',
     # Withdrawal credentials
     'withdrawal_credentials': 'bytes32',
+    # Epoch when became eligible for activation
+    'activation_eligibility_epoch': 'uint64',
```
JustinDrake: Need to define initial value in `process_deposit`.

specs/core/0_beacon-chain.md
```diff
@@ -418,14 +415,14 @@ The types are defined topologically to aid in facilitating an executable version
...
     # Epoch when validator activated
     'activation_epoch': 'uint64',
     # Epoch when validator exited
     'exit_epoch': 'uint64',
     # Epoch when validator is eligible to withdraw
     'withdrawable_epoch': 'uint64',
-    # Did the validator initiate an exit
-    'initiated_exit': 'bool',
```
JustinDrake: There's a few `initiated_exit` lingering

specs/core/0_beacon-chain.md
```diff
@@ -1352,22 +1353,20 @@ def initiate_validator_exit(state: BeaconState, index: ValidatorIndex) -> None:
...
-        validator.exit_epoch = get_delayed_activation_exit_epoch(get_current_epoch(state))
+        # Update exit queue counters
+        if state.exit_epoch < get_delayed_activation_exit_epoch(get_current_epoch(state)):
+            state.exit_epoch = get_delayed_activation_exit_epoch(get_current_epoch(state))
+        if state.exit_queue_filled >= MAX_EXIT_DEQUEUES_PER_EPOCH:
+            state.exit_epoch += 1
+            state.exit_queue_filled = 0
+        # Set validator exit epoch and withdrawable epoch
+        if validator.exit_epoch > state.exit_epoch:
+            validator.exit_epoch = state.exit_epoch
```
JustinDrake: Should we set `exit_epoch` to something even if `validator.exit_epoch <= state.exit_epoch`?

vbuterin: Why? That implies that the validator is already in the exit queue.

JustinDrake: If `initiate_validator_exit` is called (e.g. because `slash_validator` is called) then we want that event to be recorded somehow in the `Validator` object. Do we need `validator.initiated_exit` or `validator.exitable_epoch`?

JustinDrake:
Should we be ignoring attestations of slashed-and-queued validators?

CarlBeek:
> Should we be ignoring attestations of slashed-and-queued validators?

I'd argue so. Including them would seemingly only decrease fork choice stability by allowing slashed validators to attest to forks.

JustinDrake:
If `MAX_EXIT_DEQUEUES_PER_EPOCH = 4` and we have 1,000,000 active validators then it would take 3 years (!!) to dequeue them all. Why not have a churn queue mechanism which is percentage-based (similar to the `MAX_BALANCE_CHURN_QUOTIENT` logic)?

JustinDrake:
Another consideration is that entering the exit queue as a "hedge" (with the intent of rejoining under a different pubkey after exiting) is close to free. Should we increase `MIN_VALIDATOR_WITHDRAWABILITY_DELAY` or add an exit fee to disincentive entering the exit queue prematurely?

specs/core/0_beacon-chain.md
vbuterin: My thinking is that you can check that a validator has initiated the exit process via `validator.exit_epoch != FAR_FUTURE_EPOCH`.

vbuterin:
> If MAX_EXIT_DEQUEUES_PER_EPOCH = 4 and we have 1,000,000 active validators then it would take 3 years (!!) to dequeue them all. Why not have a churn queue mechanism which is percentage-based (similar to the MAX_BALANCE_CHURN_QUOTIENT logic)?

Personally I'd be inclined to favor a hybrid formula, eg. `max_dequeues = max(4, len(validator set) / 65536)`. That would guarantee that all validators can cycle out in ~9.5 months, giving a weak subjectivity period of ~3 months, except in the case where there is less than 8m ETH validating, the validator set can clear faster, providing a mechanism so that if validating turns out to be unexpectedly unpopular the system compromises in the direction of lower weak subjectivity periods to keep the absolute quantity of ETH validating not-too-low.

> Should we increase MIN_VALIDATOR_WITHDRAWABILITY_DELAY or add an exit fee to disincentivise entering the exit queue prematurely?

Premature exit fee seems reasonable to me! Though keep in mind that if we want to do it via min delays instead it's not `MIN_VALIDATOR_WITHDRAWABILITY_DELAY` that we want to tweak, but rather `PERSISTENT_COMMITTEE_PERIOD`, due to this line:

```
    assert get_current_epoch(state) - validator.activation_epoch >= PERSISTENT_COMMITTEE_PERIOD
```

`PERSISTENT_COMMITTEE_PERIOD` is currently 9 days, but we were considering extending it to 18 days for efficiency reasons. So you can't exactly exit *immediately*. But an exit fee to make exiting quickly even more undesirable does seem like a good idea.

vbuterin:
> Should we be ignoring attestations of slashed-and-queued validators?

The PR is doing that already, see the [`get_unslashed_attesting_indices`](https://github.com/ethereum/eth2.0-specs/pull/850/files#diff-51a43328a58414e132a744f3771f018cR1537) definition. Or do you mean ignoring as in not even allowing the bitfields on chain?

specs/core/0_beacon-chain.md
JustinDrake: I think what was tripping me up is the condition `validator.exit_epoch > state.exit_epoch`. It seems that this condition is always true since it is inside `if validator.exit_epoch == FAR_FUTURE_EPOCH`, and we always have `FAR_FUTURE_EPOCH > state.exit_epoch`.

vbuterin: Ah, you're right there! fixed.

JustinDrake:
I think we can get rid of `validator_registry_update_epoch` because it can be derived from the `activation_epoch`s. (And this is a slight optimisation if `indices_for_activation == []` since activations can happen in the next epoch regardless of finality.)

vbuterin:
> I think we can get rid of validator_registry_update_epoch because it can be derived from the activation_epochs. 

By using `max(activation_epoch) - ACTIVATION_EXIT_DELAY`? Seems reasonable to me!

specs/core/0_beacon-chain.md
```diff
@@ -2002,80 +1993,49 @@ def apply_rewards(state: BeaconState) -> None:
...
 ```python
 def update_registry(state: BeaconState) -> None:
     # Check if we should update, and if so, update
-    if state.finalized_epoch > state.validator_registry_update_epoch:
-        update_validator_registry(state)
+
+    activations_since_finalization = len([index in state.validator_registry if
+        state.validator_registry[index].activation_epoch > state.finalized_epoch + ACTIVATION_EXIT_DELAY
+    ])
```
vbuterin: I think we can remove this statement and the if statement entirely, because the line of code `for index in indices_for_activation[:MAX_EXIT_DEQUEUES_PER_EPOCH - activations_since_finalization]:` does that function already.

JustinDrake: Reworked, saving a bunch of lines :+1:.

Why does `get_delayed_activation_exit_epoch` have a `+ 1`? cc @djrtwo

specs/core/0_beacon-chain.md
```diff
@@ -1939,7 +1931,7 @@ def get_justification_and_finalization_deltas(state: BeaconState) -> Tuple[List[
     for index in eligible_validators:
```
JustinDrake: Should `eligible_validators` be `get_active_validator_indices(state.validator_registry, current_epoch)`?

specs/core/0_beacon-chain.md
vbuterin: It's `epoch + 1 + ACTIVATION_EXIT_DELAY` because `ACTIVATION_EXIT_DELAY` is the _minimum_ activation delay, so if you're currently inside epoch n, to ensure a delay of four full epochs you want to return n+5 so that you get the entire span of epochs [n+1 .... n+4] in between.

djrtwo:
@vbuterin Can you link to or add rationale for client implementers?

hwwhww:
Hey @JustinDrake 

Question: what's the rationale of removing `validator.activation_epoch == FAR_FUTURE_EPOCH` in
https://github.com/ethereum/eth2.0-specs/pull/850/commits/47464f28dc014ea2bb2a20121a1353fc80180174 ?

It was found by @NIC619 that it seems after removing it, the `validator.activation_epoch` will be updated again and again.

e.g. when `ACTIVATION_EXIT_DELAY: 4`, finalized epoch 1 when processing epoch 4. pre/post-activation_epoch represents before/after `process_registry_updates`.

| current_epoch | finalized_epoch | get_delayed_activation_exit_epoch(state.finalized_epoch) | pre-activation_epoch | post-activation_epoch |
|---------------|-----------------|----------------------------------------------------------|----------------------|-----------------------|
| 0             | 0               | 5                                                        | FAR_FUTURE_EPOCH     | 5                     |
| 1             | 0               | 5                                                        | 5                    | 6                     |
| 2             | 0               | 5                                                        | 6                    | 7                     |
| 3             | 0               | 5                                                        | 7                    | 8                     |
| 4             | 1               | 6                                                        | 8                    | 9                     |

If so, should we either add `validator.activation_epoch == FAR_FUTURE_EPOCH` back or modify `validator.activation_epoch >= get_delayed_activation_exit_epoch(state.finalized_epoch)` condition?

PR #852, opened 2019-03-28
Updated BASE_REWARD_QUOTIENT

vbunin:
The BASE_REWARD_QUOTIENT appears to have been set for 1024, but this was not reflected in the spec as it still showed it as 32.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -250,7 +250,7 @@ Code snippets appearing in `this style` are to be interpreted as Python code.
 
 | Name | Value |
 | - | - |
-| `BASE_REWARD_QUOTIENT` | `2**5` (= 32) |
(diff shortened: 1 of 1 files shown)


djrtwo:
Where are you seeing that the quotient was "set for 1024"?

vbunin:
Saw 2048 being used here -> https://ethresear.ch/t/the-economic-incentives-of-staking-in-serenity/4157

Then 1024 in an updated version here -> https://docs.ethhub.io/ethereum-roadmap/ethereum-2.0/eth-2.0-economics/

And figured the documentation is out of date since the spec had it set at 32 for months. Am I missing something?

djrtwo:
base reward quotient is set to 32 here -- https://docs.ethhub.io/ethereum-roadmap/ethereum-2.0/eth-2.0-economics/

The constant was changed from 2048 to 32 when we switched the base unit of account from eth to gwei. I believe 32 is correct.

Issue #844, opened 2019-03-27
Define vector as an alias


jannikluhn:
A vector is serialized in the same way a container is that uses the same type for all elements. We therefore could define it as an alias. This would simplify the spec and reduce the number of test vectors.

The only difference between the two is the way the hash tree root algorithm works in the case of basic elements: Containers hash each element and merkleize, vectors pack and merkleize. With composite elements there's no difference. I don't know if this is an important special case, but I can't see a reason for it. So I think we can just use pack for containers of basic types as well.

JustinDrake:
Interesting! Packing vectors of basic types makes sense, and the spec simplification is nice 👍

JustinDrake:
Closing with #853.

PR #853, opened 2019-03-28
Define vectors as special containers

jannikluhn:
See #844

### DIFF
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -33,9 +33,9 @@ This is a **work in progress** describing typing, serialization and Merkleizatio
 
 ### Composite types
 
-* **container**: ordered fixed-length heterogenous collection of values
+* **container**: ordered fixed-length homogeneous or heterogeneous collection of values
     * key-pair curly bracket notation `{}`, e.g. `{"foo": "uint64", "bar": "bool"}`
-* **list**: ordered variable-length homogenous collection of values
+* **list**: ordered variable-length homogeneous collection of values
     * angle bracket notation `[type]`, e.g. `["uint64"]`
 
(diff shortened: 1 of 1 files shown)


specs/simple-serialize.md
```diff
@@ -101,9 +100,9 @@ We first define helper functions:
 
 We now define Merkleization `hash_tree_root(value)` of an object `value` recursively:
 
-* `merkleize(pack(value))` if `value` is a basic object or a vector of basic objects
+* `merkleize(pack(value))` if `value` is a basic object or a container of only basic objects
```
JustinDrake: I'm not sure packing of non-homogeneous basic-type containers is particularly clean. Maybe we do want to restrict packing to vectors.

djrtwo:
closing and reopening to trigger ci

specs/simple-serialize.md
jannikluhn: yeah, I think I agree. Reverted to the original logic, but changed the phrasing to not use `vector` anymore (to avoid confusion if an object is not explicitly defined to be a vector, but has the same structure).

JustinDrake: I'm not sure it's a win anymore. The reason is that if developers want packing of their homogenous containers (something we don't have anywhere in the Eth2.0 spec!), then just pick a vector. And for most languages (e.g. Golang) there's a natural distinction between containers (`struct`) and vectors (`slice`) which breaks down because now you have to do this special check on `struct`s to see if they are actually a `slice`.

JustinDrake: @jannikluhn Do you still want to pursue this PR?

JustinDrake: @jannikluhn Do you still want to pursue this PR?

jannikluhn: no, I don't think this is particularly important.

PR #855, opened 2019-03-29
Rework crosslink processing

mkalinin:
## What was wrong?
Addresses a number of issues in crosslink processing encountered during update to v0.5.1:
1. Attestation created in one epoch but not included into any block of that epoch had a chance to eventually become invalid. It happened because of crosslink update on the closest epoch transition which made `previous_crosslink` in the attestation outdated. This PR, also, adds a test to a spec that covers this case. Described in this issue https://github.com/ethereum/eth2.0-specs/issues/845.
2. Penalties and rewards for crosslinks were calculated _after_ crosslinks have been processed, therefore `get_winning_root_and_participants` returned empty participation lists for crosslinks of updated shards as this update made `previous_crosslink` in attestations out of date.
3. There is a case when two attestations for a same shard, i.e. for previous and current epoch, are included into a state during _current_ epoch. In this case attestation from _previous_ epoch had a chance to win cause epoch was not counted in the comparison of crosslink data roots. Mentioned here: https://github.com/ethereum/eth2.0-specs/issues/845#issuecomment-477482090.
4. Shard was not counted in attestation filtering. Supplants https://github.com/ethereum/eth2.0-specs/pull/841.
5. An edge case with `GENESIS_EPOCH` caused incorrect crosslink processing. Mentioned here: https://github.com/ethereum/eth2.0-specs/pull/841#issuecomment-476732037.

## How it was fixed.
1. Outdated `previous_crosslink` was fixed by breaking crosslinks down to `previous_epoch_crosslinks` and `current_epoch_crosslinks`. Now attestation is checked against both lists which prevents it from being spoiled. There was another solution of this issue, to process crosslinks only for previous epoch. But this approach adds a delay to a shard crosslinking, IMO, a speed of crosslinking process is invariant that is not desired to be broken.
2. To work around rewards issue crosslinks are calculated and stored in memory _before_ rewards calculation and got applied _after_ `apply_reward` has been called during the epoch processing. It preserves an invariant of calculating voting balances _before_ validator balances get changed.
3. Other issues have been fixed by adding a strict order to crosslink processing: current epoch is processed after previous one.

### Crosslinks processing flow
This is how this PR affects epoch processing:
```python
def process_epoch_transition(state: BeaconState) -> None:
    spec.update_justification_and_finalization(state)
    
    # calculating crosslinks:
    latest_crosslinks = spec.get_latest_crosslinks(state)
    
    spec.maybe_reset_eth1_period(state)

    # applying rewards and crosslinks after them
    spec.apply_rewards(state)
    spec.apply_crosslinks(state, latest_crosslinks)
    ...
```

### Complexity
This PR adds a certain degree of complexity to crosslink and epoch processing. This complexity has been added in order to preserve invariants that from my perspective are important for the beacon chain processing. It may be there are other ways of solving these issues, would be happy to discuss them here.

### Naming and types
**Disclaimer:** I am not pythonist. In this PR you can meet bad namings, not that idiomatic function decomposition alongside with inappropriate type usage (esp. `Vector`).

### Testing
These changes have been tested in our simulator. It works well for ideal case when all attestations are included in the next coming block, but I believe those are good for sophisticated cases as well. Corresponding commit https://github.com/harmony-dev/beacon-chain-java/commit/a1ee101f135a20e89303087a3adb7d05be2e25eb

### Misc
This PR a bit affects Phase 1 spec. There is also renaming of `previous_crosslink` in attestation data to `source_crosslink`; this is done because `previous_crosslink` name started to clash with `previous_epoch_crosslinks`. Not sure that I like `source_crosslink` but gave up on picking something else.

### Related issues and PRs
Addresses https://github.com/ethereum/eth2.0-specs/issues/845, supplants https://github.com/ethereum/eth2.0-specs/pull/841.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -354,7 +354,7 @@ The types are defined topologically to aid in facilitating an executable version
 
     # Crosslink vote
     'shard': 'uint64',
-    'previous_crosslink': Crosslink,
+    'source_crosslink': Crosslink,
     'crosslink_data_root': 'bytes32',
 }
 ```
@@ -1733,12 +1733,12 @@ def get_previous_epoch_matching_head_attestations(state: BeaconState) -> List[Pe
 
 ```python
 def get_winning_root_and_participants(state: BeaconState, shard: Shard, epoch: Epoch) -> Tuple[Bytes32, List[ValidatorIndex]]:
-    previous_crosslinks = state.current_epoch_crosslinks if epoch == slot_to_epoch(state.slot) else state.previous_epoch_crosslinks
+    source_crosslinks = state.current_epoch_crosslinks if epoch == slot_to_epoch(state.slot) else state.previous_epoch_crosslinks
     attestations = state.current_epoch_attestations if epoch == slot_to_epoch(state.slot) else state.previous_epoch_attestations
 
     valid_attestations = [
         a for a in attestations 
-        if a.data.shard == shard and a.data.previous_crosslink == previous_crosslinks[shard]
+        if a.data.shard == shard and a.data.source_crosslink == source_crosslinks[shard]
     ]
     all_roots = [a.data.crosslink_data_root for a in valid_attestations]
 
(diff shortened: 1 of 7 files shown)


djrtwo:
Agreed that there are a few things that will be better served with dual crosslink arrays, but I'm going to take a stab at simplifying this before we merge.

Thanks!

mkalinin:
> but I'm going to take a stab at simplifying this before we merge.

Happy to implement your simplification when it's done, would be helpful to check it in a simulation before the merge.

mkalinin:
Superseded by https://github.com/ethereum/eth2.0-specs/pull/874

PR #858, opened 2019-03-30
Update the descriptive text to refer to the correct type of root

ralexstokes:
The text was outdated in referring to the `hash_tree_root` when instead we now use the `signed_root`.

This PR fixes this.


JustinDrake:
This line shouldn't really be part of the spec. Danny started an "invariants" document [here](https://notes.ethereum.org/oGP2JMPpTZC7h0GW3lnPLA#).

Issue #730, opened 2019-03-07
Enabling compounding growth for validators' stakes


CarlBeek:
Receiving compounding returns on one's stake is something I deem to be vital in a system that should be competitive as an investment instrument (something I argue is essential for drawing in as much staking capital as possible). 

In the current spec, the incentives could be deemed as unfair as `get_effecitve_balance` gives compounding returns to those who have less than 32 Eth (who have been mediocre validators at best) while giving constant (as a function of your balance) returns for those with more than 32 (those who are great validators).

A lack of compounding returns means that there is some point where it is worthwhile for a validator to logout and transfer their excess balance (their balance over 32 Eth) and then become a new active validator again. This will naturally increase the number of transactions in the beacon chain. Precisely where this point is depends on: the length of the exit queue, the fungibility of BETH, and the transfer fee market. 

Validators with smaller holdings (say only 32 Eth) are encouraged to withdraw their excess Eth periodically and either send it to a staking pool or some other open-finance alternative. Validators with large stakes, say 32x 32Eth, would be behaving economically irrationally if they do not withdraw all of their stakes when they reach 33 Eth each and launch 33 validator instances of 32 Eth again.

One possible solution would be to change `get_effective_balance` to clip your balance into [16,64]. By raising the maximum to 64 Eth, honest validators can keep staking until each validator has sufficient rewards to spilt their validator instance into 2 separate instances. This would reduce the number of transfers (and thus size and processing) in the beacon chain whilst making staking a more attractive use of capital.

When implementing this proposal, it would also be sensible to revert transfers to accounts with >32 Eth. Otherwise, validators could consolidate their ether into fewer instances which is accentuates the issue described below:

The downside of this proposal is that validators with more that 32 Eth have to do less work than they are payed for (for as block proposers etc) but they are payed appropriately for the security that they provide.

Another issue is that honest validators (with large accrued stakes) will be shuffled to a single shard (instead of multiple shards as per the economically rational behaviour described above). This means that the minimum number of validators per shard will need to increase.

**TLDR:**
Currently validator balances over 32 Eth are not rewarded which represents capital inefficiency. Due to the design of the system, the better a validator's behaviour, the greater the burden of this capital inefficiency they occur. I propose providing rewards all the way up to 64 Eth. This does have some downsides though.

djrtwo:
I believe we decided to enable transfers for balances beyond `MAX_EFFECTIVE_BALANCE` rather than to enable compounding.

[...]

PR #862, opened 2019-03-31
Rounded effective balance

JustinDrake:
Use the rounded balance for the effective balance to homogenise the on-onchain finality and crosslink gadgets with the fork choice rule and the light client logic.

Minor cleanups:

* Use `increase_balance` and `decrease_balance` in `apply_rewards` instead of `set_balance`.
* Use `get_effective_balance` for activations and ejections.
* Rename `high_balance` to `rounded_balance` (we no longer have `high_balance` and `low_balance`).

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -2004,7 +2004,7 @@ def process_ejections(state: BeaconState) -> None:
     and eject active validators with balance below ``EJECTION_BALANCE``.
     """
     for index in get_active_validator_indices(state.validator_registry, get_current_epoch(state)):
-        if get_effective_balance(state, index) < EJECTION_BALANCE:
+        if get_effective_balance(state, index) <= EJECTION_BALANCE:
             initiate_validator_exit(state, index)
 ```
 
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -430,7 +430,7 @@ The types are defined topologically to aid in facilitating an executable version
     # Was the validator slashed
     'slashed': 'bool',
     # Rounded balance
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1091,7 +1088,7 @@ def get_effective_balance(state: BeaconState, index: ValidatorIndex) -> Gwei:
     """
     Return the effective balance (also known as "balance at stake") for a validator with the given ``index``.
     """
-    return min(get_balance(state, index), MAX_DEPOSIT_AMOUNT)
+    return min(state.validator_registry[index].rounded_balance, MAX_DEPOSIT_AMOUNT)
```
djrtwo: This function is used for reward/penalty calculations. Do we want to use rounded balance for reward/penalties?

JustinDrake: I'd argue so. The balance at stake is rounded so rewards/penalties should also be.

CarlBeek: This does effect a validator's effective rate of return as rounded_balance <= actual_balance. In the event that a validator has >= 32ETH and has always had so, then this doesn't matter. In the event that a validator's balance drops to 16ETH, however, by this makes the difference between having 64 ETH and 65 ETH after validating for a long time.

If we allow balances to grow exponentially as per #730, then this always matters. See graphs attached below:

econ_test_init_bal16
econ_test_init_bal32

JustinDrake: Right. I guess a downside of using rounded balance for rewards is that it creates a misalignment between capital costs (not rounded) and rewards (rounded).

JustinDrake:
Close in favour of #949

PR #869, opened 2019-03-31
Stagger reveal slots

vbuterin:
Depends on #868 for `get_switchover_epoch`.

Also changes the condition for punitive reveals to `assert reveal.period >= current_custody_period and is_active_validator(revealer, get_current_epoch(state))` because revealing a subkey during the period that it is being used to make custody bits should also count as punitive.

Issue #864, opened 2019-03-31
Custody game checklist


JustinDrake:
**TODO**

- [x] 1. **Merkleisation-friendly challenge records**: Zero-out and over-write instead of `remove()` and `append()`.
- [ ] 2. **Worst-case key reveals**: Increase `MAX_CUSTODY_KEY_REVEALS` and `EPOCHS_PER_CUSTODY_PERIOD` to deal with worst-case key reveal bandwidth.
- [ ] 3. **Increase challenge bandwidth**: Increase `MAX_CUSTODY_CHUNK_CHALLENGES` and `MAX_CUSTODY_MIX_CHALLENGES`.
- [x] 4. **Fix withdrawability**: Validators need to regain withdrawability after `responder.withdrawable_epoch = FAR_FUTURE_EPOCH`.
- [x] 5. **MPC-friendly chunk bit**: Move to something like the Legendre symbol.
- [x] 6. **Force timely key reveals**: Similar to timely challenge responses.
- [x] 7. **Multiple challenges per validator**: Only the first challenge that passes the waiting period rewards the challenger.
- [x] 8. **Stagger reveal periods**: To avoid large key reveal bursts at period boundaries.
- [x] 9. **Remove `max_reveal_lateness`**: Find cleaner alternative.
- [ ] 10. **Redefine `BYTES_PER_SHARD_BLOCK`**: Replace by `BYTES_PER_SHARD_BLOCK = 2 * BYTES_PER_SHARD_BLOCK_BODY`.
- [ ] 11. **Economic review**: Particularly around sizes and structure of rewards and penalties.
- [ ] 12. **Review `BYTES_PER_CUSTODY_CHUNK`**: There are pros and cons to increasing/decreasing.
- [x] 13. **Executable spec**: Make custody game executable and add tests.
- [x] 14. **Chunk response delay**: Add `MIN_ACTIVATION_EXIT_DELAY` for chunk responses. (To avoid proposers responding to their own challenges.)

**Ideas to consider**

- [ ] 1. **G1 key reveals**: 2x smaller key reveals.
- [x] 2. **Merge RANDAO and key reveals**: Slightly reduced key reveal overhead.
- [x] 3. ~**Batched reveals**: Batch reveals in case of lateness.~
- [ ] 4. **Separate reveal per epoch**: Derive a new key per epoch from a single period reveal.
- [ ] 5. **Forced responses**: Force proposers to respond to their challenges.
[...]

PR #872, opened 2019-04-01
Custody punishments for late reveals & deterministic staggering

CarlBeek:
This PR implements the following:
* **Forced timely key reveals** Validators who have not revealed by the end of their reveal period are slashed.
* **Staggers reveal periods** Reveal periods are deterministically staggered by taking the `validator_index%EPOCHS_PER_CUSTODY_PERIOD`. This introduces modulo-bias and the earlier epochs in a period could have one more reveal each. The staggering strategy from #869 is not used for two reasons: it requires storing when each validator's latest reveal period started (as it cannot be calculated later as it is dependent on the RANDAO state) and secondly, this mechanism allows validators to know when (approximately) their reveal period will begin.
* **Removes `max_reveal_lateness`** I think all the functionality that was provided by `max_reveal_lateness` has been replaced, but please check my logic.

Hereby items 6, 8, and 9 from #864 are addressed.

### DIFF
--- a/specs/core/1_custody-game.md
+++ b/specs/core/1_custody-game.md
@@ -489,7 +489,8 @@ Run `process_reveal_deadlines(state)` immediately after `process_ejections(state
 ```python
 def process_reveal_deadlines(state) -> None:
     for idx, validator in enumerate(state.validator_registry):
-        if validator.latest_custody_reveal_period + 1 < get_validators_current_custody_reveal_period(state, idx):
+        if (validator.latest_custody_reveal_period + CUSTODY_RESPONSE_DEADLINE // EPOCHS_PER_CUSTODY_PERIOD 
+            < get_validators_current_custody_reveal_period(state, idx)):
             slash_validator(state, idx)
 ```
 
--- a/specs/core/1_custody-game.md
+++ b/specs/core/1_custody-game.md
@@ -293,11 +293,11 @@ def process_custody_reveal(state: BeaconState,
                            reveal: CustodyKeyReveal) -> None:
     assert verify_custody_key(state, reveal)
     revealer = state.validator_registry[reveal.revealer_index]
-    validators_current_custody_reveal_period = get_validators_current_custody_reveal_period(state, reveal.revealer_index)
+    validators_next_custody_reveal_period = validator.latest_custody_reveal_period + 1
 
     # Case 1: non-masked non-punitive non-early reveal
     if reveal.mask == ZERO_HASH:
-        assert reveal.period == validators_current_custody_reveal_period
+        assert reveal.period == validators_next_custody_reveal_period
         # Revealer is active or exited, but not withdrawn
         assert is_slashable_validator(revealer, get_current_epoch(state))
         revealer.latest_custody_reveal_period += 1
@@ -307,7 +307,7 @@ def process_custody_reveal(state: BeaconState,
 
     # Case 2: masked punitive early reveal
     else:
-        assert reveal.period > validators_current_custody_reveal_period
+        assert reveal.period >= validators_next_custody_reveal_period
         assert is_slashable_validator(revealer)
         slash_validator(state, reveal.revealer_index, reveal.masker_index)
 ```
--- a/specs/core/1_custody-game.md
+++ b/specs/core/1_custody-game.md
@@ -31,6 +31,7 @@
         - [`get_crosslink_chunk_count`](#get_crosslink_chunk_count)
         - [`get_custody_chunk_bit`](#get_custody_chunk_bit)
         - [`epoch_to_custody_period`](#epoch_to_custody_period)
+        - [`get_validators_current_custody_reveal_period`](#get_validators_current_custody_reveal_period)
         - [`verify_custody_key`](#verify_custody_key)
     - [Per-block processing](#per-block-processing)
         - [Transactions](#transactions)
@@ -39,6 +40,9 @@
             - [Bit challenges](#bit-challenges)
             - [Custody responses](#custody-responses)
     - [Per-epoch processing](#per-epoch-processing)
+        -[Handling of custody-related deadlines](#handling_of_custody-related_deadlines)
+        -[Miscellaneous processing](#miscellaneous_processing)
+
 
 <!-- /TOC -->
 
@@ -180,8 +184,12 @@ Add the following fields to the end of the specified container objects. Fields w
 #### `Validator`
 
 ```python
-    'custody_reveal_index': 'uint64',
-    'max_reveal_lateness': 'uint64',
+    '''
+    latest_custody_reveal_period is initialised to the custody period
+    (of the particular validator) *AFTER* which the validator was activated.
+    = get_validators_current_custody_reveal_period(...) + 1
+    '''
+    'latest_custody_reveal_period': 'uint64',
 ```
 
 #### `BeaconState`
@@ -229,6 +237,17 @@ def epoch_to_custody_period(epoch: Epoch) -> int:
     return epoch // EPOCHS_PER_CUSTODY_PERIOD
 ```
 
+### `get_validators_current_custody_reveal_period`
+
+```python
+def get_validators_current_custody_reveal_period(state: BeaconState, validator_index: ValidatorIndex) -> int:
+    '''
+    Note: This function implicitly requires that validators are not removed from the
+    validator set in fewer than EPOCHS_PER_CUSTODY_PERIOD epochs
+    '''
+    return epoch_to_custody_period(get_current_epoch(state) + validator_index % EPOCHS_PER_CUSTODY_PERIOD)
+```
+
 ### `verify_custody_key`
 
 ```python
@@ -274,22 +293,22 @@ def process_custody_reveal(state: BeaconState,
                            reveal: CustodyKeyReveal) -> None:
     assert verify_custody_key(state, reveal)
     revealer = state.validator_registry[reveal.revealer_index]
-    current_custody_period = epoch_to_custody_period(get_current_epoch(state))
+    validators_current_custody_reveal_period = get_validators_current_custody_reveal_period(state, reveal.revealer_index)
 
     # Case 1: non-masked non-punitive non-early reveal
     if reveal.mask == ZERO_HASH:
-        assert reveal.period == epoch_to_custody_period(revealer.activation_epoch) + revealer.custody_reveal_index
-        # Revealer is active or exited
-        assert is_active_validator(revealer, get_current_epoch(state)) or revealer.exit_epoch > get_current_epoch(state)
-        revealer.custody_reveal_index += 1
-        revealer.max_reveal_lateness = max(revealer.max_reveal_lateness, current_custody_period - reveal.period)
+        assert reveal.period == validators_current_custody_reveal_period
+        # Revealer is active or exited, but not withdrawn
+        assert is_slashable_validator(revealer, get_current_epoch(state))
+        revealer.latest_custody_reveal_period += 1
+        # Reward block proposer
         proposer_index = get_beacon_proposer_index(state, state.slot)
         increase_balance(state, proposer_index, base_reward(state, index) // MINOR_REWARD_QUOTIENT)
 
     # Case 2: masked punitive early reveal
     else:
-        assert reveal.period > current_custody_period
-        assert revealer.slashed is False
+        assert reveal.period > validators_current_custody_reveal_period
+        assert is_slashable_validator(revealer)
         slash_validator(state, reveal.revealer_index, reveal.masker_index)
 ```
 
@@ -324,7 +343,7 @@ def process_chunk_challenge(state: BeaconState,
     state.custody_chunk_challenge_records.append(CustodyChunkChallengeRecord(
         challenge_index=state.custody_challenge_index,
         challenger_index=get_beacon_proposer_index(state, state.slot),
-        responder_index=challenge.responder_index
+        responder_index=challenge.responder_index,
         deadline=get_current_epoch(state) + CUSTODY_RESPONSE_DEADLINE,
         crosslink_data_root=challenge.attestation.data.crosslink_data_root,
         depth=depth,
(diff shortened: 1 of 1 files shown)


dankrad:
Cool! One question: Theoretically, we do not need latest_custody_reveal_period to be an integer, right? It could be a boolean that stores whether a reveal was published in this period.
This might be annoying as it would require one additional validator update per period, so we could just make this number mod 2? I think that information would be sufficient?

specs/core/1_custody-game.md
```diff
@@ -464,7 +482,18 @@ def process_bit_challenge_response(state: BeaconState,
...
-Run `process_challenge_deadlines(state)` immediately after `process_ejections(state)`:
+### Handling of custody-related deadlines
+
+Run `process_reveal_deadlines(state)` immediately after `process_ejections(state)`:
+
+```python
+def process_reveal_deadlines(state) -> None:
+    for idx, validator in enumerate(state.validator_registry):
+        if validator.latest_custody_reveal_period + 1 < get_validators_current_custody_reveal_period(state, idx):
+            slash_validator(state, idx)
```
vbuterin: This looks dangerous. This basically means that being offline for ~9 days will get you slashed, whereas the standard in the rest of the protocol is that you need for be offline for ~73 days for that to happen. So I think it needs to be `+ CUSTODY_RESPONSE_DEADLINE // EPOCHS_PER_CUSTODY_PERIOD`, not `+ 1`.

vbuterin:
>The staggering strategy from #869 is not used for two reasons: it requires storing when each validator's latest reveal period started (as it cannot be calculated later as it is dependent on the RANDAO state) and secondly, this mechanism allows validators to know when (approximately) their reveal period will begin.

Wait, why do you need to store when each validator's latest reveal period started? In my proposal, the only thing that gets staggered is the minimum reveal block, custody _periods_ are still aligned.

If custody periods are all aligned, then I would argue that my staggering strategy is superior because relying on validator index means that there's persistent unfairness where some validators get always earlier reveal times than others. If custody periods are _not_ aligned, I agree your approach is best. What do you see as being the benefits and costs either way?

> Cool! One question: Theoretically, we do not need latest_custody_reveal_period to be an integer, right? It could be a boolean that stores whether a reveal was published in this period.

No because the +1 lateness bound is too tight to be safe (see my other comment), it needs to be +8.

specs/core/1_custody-game.md
dankrad: Yes, that sounds reasonable. Just a question, do we want to set any deadlines for bitfield challenges? I just checked the logic, and currently there are none. We probably don't want a logic to remember when reveals happened, so it would make sense to just set this to 2 * CUSTODY_RESPONSE_DEADLINE?

CarlBeek:
> Wait, why do you need to store when each validator's latest reveal period started? In my proposal, the only thing that gets staggered is the minimum reveal block, custody _periods_ are still aligned.

This comes down to the stringency of my slashing mechanism. As everyone who has not revealed is slashed by the start of their next period, it would be unfair to allow certain validators only 1 block to reveal their key.

> If custody periods are not aligned, I agree your approach is best.

I was staggering the periods of each validator (up to 1 custody_period phase difference between any two validators). I was thus targeting *non*-aligned custody periods hence the ridiculous naming of `get_validators_current_custody_reveal_period()`. By increasing the reveal time (which I agree should be done), the phase difference of your stochastic minimum_reveal_block has less of an impact on the available reveal time for a validator.

> What do you see as being the benefits and costs either way?

I don't think it really matters much either way. The downsides of doing it your way are increased hashing (minor) and lack of predictability of when a particular validator's reveal period begins. On the other hand, my system has constant bias (as described by me as "modulo-bias" in my initial comment) in the sense that blocks at the start of a reveal period could contain 1 more reveal every time.

specs/core/1_custody-game.md
CarlBeek: > This looks dangerous. This basically means that being offline for ~9 days will get you slashed, whereas the standard in the rest of the protocol is that you need for be offline for ~73 days for that to happen. So I think it needs to be `+ CUSTODY_RESPONSE_DEADLINE // EPOCHS_PER_CUSTODY_PERIOD`, not `+ 1`.

Agreed and implemented.

There are deadlines for bitfield challenges. See line 506 below

specs/core/1_custody-game.md
```diff
@@ -274,22 +293,22 @@ def process_custody_reveal(state: BeaconState,
                            reveal: CustodyKeyReveal) -> None:
     assert verify_custody_key(state, reveal)
     revealer = state.validator_registry[reveal.revealer_index]
-    current_custody_period = epoch_to_custody_period(get_current_epoch(state))
+    validators_current_custody_reveal_period = get_validators_current_custody_reveal_period(state, reveal.revealer_index)
```
dankrad:
```suggestion
    validators_next_custody_reveal_period = validator.latest_custody_reveal_period + 1
```
I think this needs to be changed to be consistent with having more time to reveal the secret

specs/core/1_custody-game.md
```diff
@@ -274,22 +293,22 @@ def process_custody_reveal(state: BeaconState,
...
 
     # Case 1: non-masked non-punitive non-early reveal
     if reveal.mask == ZERO_HASH:
-        assert reveal.period == epoch_to_custody_period(revealer.activation_epoch) + revealer.custody_reveal_index
-        # Revealer is active or exited
-        assert is_active_validator(revealer, get_current_epoch(state)) or revealer.exit_epoch > get_current_epoch(state)
-        revealer.custody_reveal_index += 1
-        revealer.max_reveal_lateness = max(revealer.max_reveal_lateness, current_custody_period - reveal.period)
+        assert reveal.period == validators_current_custody_reveal_period
```
dankrad:
```suggestion
        assert reveal.period == validators_next_custody_reveal_period
```

specs/core/1_custody-game.md
```diff
@@ -274,22 +293,22 @@ def process_custody_reveal(state: BeaconState,
...
+        revealer.latest_custody_reveal_period += 1
+        # Reward block proposer
         proposer_index = get_beacon_proposer_index(state, state.slot)
         increase_balance(state, proposer_index, base_reward(state, index) // MINOR_REWARD_QUOTIENT)
 
     # Case 2: masked punitive early reveal
     else:
-        assert reveal.period > current_custody_period
-        assert revealer.slashed is False
+        assert reveal.period > validators_current_custody_reveal_period
```
dankrad:
```suggestion
        assert reveal.period >= validators_next_custody_reveal_period
```

specs/core/1_custody-game.md
```diff
@@ -358,8 +377,7 @@ def process_bit_challenge(state: BeaconState,
     assert verify_standalone_attestation(state, convert_to_standalone(state, challenge.attestation))
     # Verify the attestation is eligible for challenging
     responder = state.validator_registry[challenge.responder_index]
-    min_challengeable_epoch = responder.exit_epoch - EPOCHS_PER_CUSTODY_PERIOD * (1 + responder.max_reveal_lateness)
-    assert min_challengeable_epoch <= slot_to_epoch(challenge.attestation.data.slot) 
+    assert responder.latest_custody_reveal_period > epoch_to_custody_period(slot_to_epoch(challenge.attestation.data.slot))
```
dankrad: I think in the current implementation there is no time limit on bit challenges. Do we want to set this to 2*CUSTODY_RESPONSE_DEADLINE ? (That means after the custody reveal there is at least CUSTODY_RESPONSE_DEADLINE to post a bit challenge)

specs/core/1_custody-game.md
dankrad: > There are deadlines for bitfield challenges. See line 506 below

This deadline concerns the response to the bit challenge. I am suggesting we should not allow bit challenges arbitrarily into the past (see my comment on line 380)

dankrad:
> > What do you see as being the benefits and costs either way?
> 
> I don't think it really matters much either way. The downsides of doing it your way are increased hashing (minor) and lack of predictability of when a particular validator's reveal period begins. On the other hand, my system has constant bias (as described by me as "modulo-bias" in my initial comment) in the sense that blocks at the start of a reveal period could contain 1 more reveal every time.

I like the approach that Carl is suggesting here, because it minimises the time from when a custody secret is used to its corresponding reveal period. The expected rational behaviour is to reveal a custody secret as soon as possible after using it, so this approach has two advantages:
 * The time to check a custody secret is minimised (less load on validators checking secrets)
 * Validators can reveal as soon as possible, minimising the time when they can get slashed

Also the logic does seem to have very little overhead. But overall, the difference is fairly minor, so I don't feel very strongly about it.

specs/core/1_custody-game.md
```diff
@@ -479,16 +512,21 @@ def process_challenge_deadlines(state: BeaconState) -> None:
...
     validator = state.validator_registry[index]
     # Cannot exit if there are still open chunk challenges
     if len([record for record in state.custody_chunk_challenge_records if record.responder_index == index]) > 0:
         return False
+    # Cannot exit if there are still open bit challenges
+    if len([record for record in state.custody_bit_challenge_records if record.responder_index == index]) > 0:
+        return False
     # Cannot exit if you have not revealed all of your custody keys
-    elif epoch_to_custody_period(revealer.activation_epoch) + validator.custody_reveal_index <= epoch_to_custody_period(validator.exit_epoch):
+    elif validator.latest_custody_reveal_period <= epoch_to_custody_period(validator.exit_epoch):
```
vbuterin: I think if it's _latest reveal_ rather than _next reveal_, the inequality should be <, not  <=. The rationale is that if you exit during epoch N, that means that you used the subkey for epoch N and previous epochs, so you need to reveal all the subkeys up to and including N, so the "latest reveal" should be N for an exit to be valid.

vbuterin:
> This comes down to the stringency of my slashing mechanism. As everyone who has not revealed is slashed by the start of their next period, it would be unfair to allow certain validators only 1 block to reveal their key.

So if the slashing mechanism only works after 8 epochs, this seems like it matters less.

> because it minimises the time from when a custody secret is used to its corresponding reveal period

I like how it _stabilizes_ the time between use and reveal, but I actually think _minimizing_ the time could be a bad idea. Revealing data literally a block after the period ends opens you up to slashing if the chain reorgs. That said, I suppose this is something that we could leave open as a tradeoff to the user - if they wait longer, then they get more safety, but then they would have to store more data in the meantime.

specs/core/1_custody-game.md
```diff
@@ -353,13 +372,15 @@ def process_bit_challenge(state: BeaconState,
         domain=get_domain(state, get_current_epoch(state), DOMAIN_CUSTODY_BIT_CHALLENGE),
     )
     # Verify the challenger is not slashed
-    assert challenger.slashed is False
+    assert is_slashable_validator(challenger)
     # Verify the attestation
     assert verify_standalone_attestation(state, convert_to_standalone(state, challenge.attestation))
     # Verify the attestation is eligible for challenging
     responder = state.validator_registry[challenge.responder_index]
-    min_challengeable_epoch = responder.exit_epoch - EPOCHS_PER_CUSTODY_PERIOD * (1 + responder.max_reveal_lateness)
```
vbuterin: We can't get rid of the max delay because then the validator would be liable for an unboundedly growing period of data. We also can't set max delay to equal the custody response deadline because that would just be too much data to store (2 MB per epoch * 2**14 epochs = 32 GB); we want a validator to only be responsible for the data in a single period (unless the validator screws up eg. by delaying subkey reveals). So we need a max delay of 1 in the normal case, but if the validator does delay subkey reveals we need to increase the max delay to make sure that when the subkey is revealed the data can still be challenges. This was the rationale for `max_reveal_lateness`.

vbuterin: @JustinDrake and I thought about this and couldn't come up with a way to solve this issue without having that variable, despite its ugliness, though what we _can_ do is ameliorate the problem where if you're late once you're stuck with a longer period forever, by having `max_reveal_lateness` be measured in a finer-grained time interval (eg. epochs) and have it slowly decrease over time if a validator shows good behavior (another isomorphic approach is to have a `min_challengeable_epoch` that, if it is more than a full period behind the current epoch, moves forward eg. 2 epochs per epoch.

specs/core/1_custody-game.md
```diff
@@ -180,8 +184,12 @@ Add the following fields to the end of the specified container objects. Fields w
...
 
 ```python
-    'custody_reveal_index': 'uint64',
-    'max_reveal_lateness': 'uint64',
+    '''
+    latest_custody_reveal_period is initialised to the custody period
+    (of the particular validator) *AFTER* which the validator was activated.
+    = get_validators_current_custody_reveal_period(...) + 1
+    '''
+    'latest_custody_reveal_period': 'uint64',
```
vbuterin: So the first subkey that the validator will be revealing is from the period that is _two_ periods ahead of the current one? Not sure I see the logic here. You sure we don't want `- 1`?

And if it is -1, then it seems better to avoid negative numbers and keep track of `next_custody_reveal_period` instead of `latest_custody_reveal_period`.

dankrad:
> Revealing data literally a block after the period ends opens you up to slashing if the chain reorgs.
Good point. Would it make sense to only allow slashing while the custody secret is actually in use? Then you can never get slashed for revealing your secret, which makes sense, and if it is accidentally disclosed after the custody period it is presumably not really an offense worth slashing.

specs/core/1_custody-game.md
dankrad: Ah, now I see the rationale for max_reveal_lateness. I like the idea of having it decrease over time.

vbuterin:
> Would it make sense to only allow slashing while the custody secret is actually in use? 

Aren't we doing that already? At least we were when I wrote the original code; that's definitely how it should work IMO.

specs/core/1_custody-game.md
```diff
@@ -274,22 +293,22 @@ def process_custody_reveal(state: BeaconState,
...
+        revealer.latest_custody_reveal_period += 1
+        # Reward block proposer
         proposer_index = get_beacon_proposer_index(state, state.slot)
         increase_balance(state, proposer_index, base_reward(state, index) // MINOR_REWARD_QUOTIENT)
 
     # Case 2: masked punitive early reveal
     else:
-        assert reveal.period > current_custody_period
-        assert revealer.slashed is False
+        assert reveal.period >= validators_next_custody_reveal_period
```
dankrad:
```suggestion
        assert reveal.period >= get_validators_current_custody_reveal_period(state, reveal.masker_index)
```
Only allow slashing for secrets that are still in use or will be used in the future

dankrad:
> Aren't we doing that already? At least we were when I wrote the original code; that's definitely how it should work IMO.

Yup, probably my bad, I accidentally removed that logic in a suggestion earlier

CarlBeek:
This is being deprecated due to the improved rewrite that is #934 :)

PR #873, opened 2019-04-02
minor bug `process_attester_slashings`

djrtwo:
Fixes a minor bug in `process_attester_slashings` and adds  a number of attester slashing tests.

Thanks for finding this @protolambda

### DIFF
--- a/tests/phase0/test_sanity.py
+++ b/tests/phase0/test_sanity.py
@@ -40,6 +40,7 @@ from tests.phase0.helpers import (
     build_empty_block_for_next_slot,
     force_registry_change_at_next_epoch,
     get_valid_attestation,
+    get_valid_attester_slashing,
     get_valid_proposer_slashing,
     privkeys,
     pubkeys,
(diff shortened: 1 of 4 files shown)


protolambda:
Good to see more tests for this, the sanity check missed it :+1:

djrtwo:
I apparently skipped attester-slashing tests when I did the first pass on sanity tests 😬

tests/phase0/test_sanity.py
```diff
@@ -140,6 +141,32 @@ def test_proposer_slashing(state):
...
+    #
+    block = build_empty_block_for_next_slot(test_state)
+    block.body.attester_slashings.append(attester_slashing)
+    state_transition(test_state, block)
+
+    assert not state.validator_registry[validator_index].initiated_exit
+    assert not state.validator_registry[validator_index].slashed
+
+    slashed_validator = test_state.validator_registry[validator_index]
+    assert not slashed_validator.initiated_exit
```
hwwhww: It seems meaningless check since right now it doesn't trigger `initiate_validator_exit` until we have #850.

tests/phase0/test_sanity.py
```diff
@@ -140,6 +141,32 @@ def test_proposer_slashing(state):
...
+    assert slashed_validator.slashed
+    assert slashed_validator.exit_epoch < spec.FAR_FUTURE_EPOCH
+    assert slashed_validator.withdrawable_epoch < spec.FAR_FUTURE_EPOCH
+    # lost whistleblower reward
+    assert get_balance(test_state, validator_index) < get_balance(state, validator_index)
```
hwwhww: Maybe also check `proposer_index`? It happened to me when pseudo-randomly, `proposer == slashed_validator` in unit tests. 😆

tests/phase0/test_sanity.py
djrtwo: Just making sure that slashing is following a different path than "initiated_exit"

PR #875, opened 2019-04-02
Fix epoch overflow in process_exit_queue

mkalinin:
## What was wrong?
There was an overflow in checking whether validator is eligible for exit; it made validators exit at the first epoch transition.


vbuterin:
The entire mechanism will be revamped by #850; and I don't think this applies there. In particular, that entire function no longer exists in that PR.

PR #851, opened 2019-03-28
Combine specs and test-generators

protolambda:
See Issue #836 for motivation and reasoning on basic structure.

~~_Blocked by #877_~~ (Supporting new format, work in progress, see comment)

## `specs`

The specs directory, unchanged. Where the markdown specs live.

## `test_generators`

Full documentation on generators can be found in `/test_generators/README.md`.
If there is anything magic to remember, it's that you run `make gen_yaml_tests` to run the generators, the makefile will handle virtual environments/dependencies etc.

## `py-tests`

Py-tests are meant to validate basic consistency and sanity expectations of the spec itself.
Documentation can be found here: `/pytests/README.md`. Generally not used by client teams, it just helps to make sure a PR is valid.

## `scripts`

This contains the (basic, no dependencies) code to generate the pyspec from the markdown version.

## `test_libs`

This contains all the python packages that one may use during any kind of testing (pyspec, test-generator utilities, and more in the future)

### `pyspec`

Most of the pyspec is rather static, it is just the `eth2/phase0/spec.py` file being dynamically generated from the markdown for now. This file is git-ignored. The other files are just kept in place, to keep the build system simple. You run `make pyspec` to build the pyspec from the markdown.

More documentation in `/test_libs/pyspec/README.md`

### `gen_helpers`

This is just a module to require in your yaml generator to make use of the base generator etc. Documentation can be found in `/test_generators/README.md`.

## `.circleci`

The Circle CI configuration. Similar to before, but updated to handle yaml test generation, and make use of the new makefile commands.

PR #877, opened 2019-04-03
A new more complete test format for ETH 2.0 testing

protolambda:
# The format

**See `specs/test_formats/README.md` file in PR**, includes new testing glossary, configuration considerations, and general format.

Test type formats will be documented in markdown files in the `specs/test_formats/` folder.

Upgrade path:
- First, we just need consensus on the general format.
- Update the CI/pyspec PR (#851), it includes a base test-generator to make life easy
- Upgrade the existing test generators, documenting them in the progress
- Introduce new test types

## Test Types

The following test-types are proposed:

- Existing types
  - BLS
  - SSZ
  - Shuffling
- Next up
  - general state transition, `(pre state, [block...]) -> post state`
  - transactions (operations), one handler per type `(pre state, tx) -> post state`
  - epoch sub-transitions (not the ones that just process a batch)
- To be looked into
  - fork choice tests
  - "deltas": test computed changes in balances
  - suggestions welcome

#### BLS

To be updated to move the "sub types" to separate test suites, each with their own handler. I.e. it would look like:

```
├── bls
│   ├── signing
│   │   └── sign_msg.yml
│   ├── verify_single
│   │   ├── invalid.yml
│   │   └── valid.yml
│   ├── verify_multiple
│   │   ├── invalid.yml
│   │   └── valid.yml
│   ├── aggregate_sigs
│   │   ├── single.yml
│   │   ├── mixed_invalid.yml
│   │   └── multiple.yml
│   ├── aggregate_pubkeys
│   │   ├── single.yml
│   │   ├── two.yml
│   │   ├── multiple.yml
│   │   └── duplicate.yml
│   └── ...
...
```

#### SSZ

The SSZ tests are compatible as is, and the base generator is based of of the design of this generator.
Easy to port over to use the new base generator.

#### Shuffling

Although the current testing is good, it sort of mixes validator activation with the shuffling algorithm.
We may decide to change it to just test shuffling based on a list of indices later down the road.


protolambda:
**Work in progess**, suggestions welcome, aiming at fast-paced progress to get new state transition tests in

specs/test_formats/README.md
```diff
@@ -0,0 +1,168 @@
...
+- prevent duplication of a minimal set of tests
+- make all tests easy to upgrade when a new config constant is introduced.
+- clearly define which constants to use
+- share-able between clients, for cross-client short or long lived test-nets
+- minimize the amounts of different constants permutations to compile as a client. \**
+ 
+\**: Some clients prefer compile-time constants and optimizations.
+They should compile for each configuration once, and run the corresponding tests per build target.
+
+
```
djrtwo: remove one blank line.

```suggestion
```

specs/test_formats/README.md
```diff
@@ -0,0 +1,168 @@
...
+Note that phases are considered to be "super forks",
+ i.e. they will just have a fork name, and be more heavy on changes.
+
+## Config sourcing
+
+The constants configurations are located in:
+
+```
+<specs repo root>/configs/contants/<config name>.yaml
```
djrtwo: typo
```suggestion
<specs repo root>/configs/constants/<config name>.yaml
```

specs/test_formats/README.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+And copied by CI for testing purposes to:
+
+```
+<tests repo root>/configs/contants/<config name>.yaml
```
djrtwo:
```suggestion
<tests repo root>/configs/constants/<config name>.yaml
```

specs/test_formats/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Glossary
+
+- `generator`: a program that outputs one or more `suite` files.
+- `type`: the specialization of one single `generator`.
+- `suite`: a YAML file with:
+  - a header: describes the `suite`, and defines what the `suite` is for
+  - a list of test cases
+- `runner`: where a generator is a "producer", this is the 1-to-1 "consumer". A `runner` focuses on one `type`.
+- `handler`: a `runner` may be too limited sometimes, you may have a `suite` with a specific focus that requires a different format.
```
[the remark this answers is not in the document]
djrtwo: If so, be explicit in this def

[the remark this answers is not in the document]
protolambda: See new description

coreyandersonbuilders:
Hi-jacked accounts fraud has need done

coreyandersonbuilders:
Fraud has been done against me

specs/test_formats/README.md
```diff
@@ -0,0 +1,170 @@
...
+### Fork config design
+
+There are two types of fork-data:
+1) timeline: when does a fork take place?
+2) coverage: what forks are covered by a test?
+
+The first is neat to have as a separate form: we prevent duplication, and can run with different presets
+ (e.g. fork timeline for a minimal local test, for a public testnet, or for main-net)
+
+The second is still somewhat ambiguous: some tests may want cover multiple forks, and can do so in different ways:
```
hwwhww: IMO it's unlikely that we *need* to have one test suite for multiple forks. Even the same scenario of Petersburg and Byzantium happen (fork on the same block) in eth2, the test suite only need to provide  Byzantium tests.

specs/test_formats/README.md
```diff
@@ -0,0 +1,170 @@
...
+With this approach, configurations are backwards compatible (older clients ignore unknown variables), and easy to maintain.
+
+### Fork config design
+
+There are two types of fork-data:
+1) timeline: when does a fork take place?
+2) coverage: what forks are covered by a test?
+
+The first is neat to have as a separate form: we prevent duplication, and can run with different presets
+ (e.g. fork timeline for a minimal local test, for a public testnet, or for main-net)
```
hwwhww: It sounds more like "Chain config" to me. What's the use case here?

specs/test_formats/README.md
```diff
@@ -0,0 +1,170 @@
...
+ 
+\**: Some clients prefer compile-time constants and optimizations.
+They should compile for each configuration once, and run the corresponding tests per build target.
+
+## Fork-timeline
+
+A fork timeline is (preferably) loaded in as a configuration object into a client, as opposed to the constants configuration:
+ - we do not allocate or optimize any code based on slot numbers
+ - when we transition from one fork to the other, it is preferred to stay online.
+ - we may decide on a slot number for a fork based on external events (e.g. Eth1 log event),
```
hwwhww: Eth1 log event defines the genesis time, not the slot number.

protolambda: Yes, just want to highlight that it can be a dynamically chosen slot, and doesn't have to be a pre-determined thing when compiling.

specs/core/0_beacon-chain.md
```diff
@@ -178,6 +178,9 @@ Code snippets appearing in `this style` are to be interpreted as Python code.
 
 ## Constants
 
+Note: the default main-net values for the constants are included here for illustrative purposes.
+The different configurations for main-net, test-nets, and yaml-based testing can be found in the `configs/contants/` directory.
```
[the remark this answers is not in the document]
protolambda: It feels strange to write mainnet before test-net. Can we decide on one wording for main(-)net and test(-)net? I'll fix all occurrences based on that then.

specs/test_formats/README.md
protolambda: The idea is that for some tests, they should be exactly the same for some consecutive forks. And instead of duplicating the file to ensure it runs as expected for multiple forks, I would use the forks-list for these types of tests to indicate this consistency explicitly. And, it adds to another property: we can really easily determine exactly which forks are covered by which tests.

protolambda:
@hwwhww 

> Side note: eth 1.0 does have WIP standardized genesis format

Yes, the timeline is very similar to the "params" section of the genesis format proposed for eth 1. But with slots, because we can :)

The "genesis"  and "accounts" sections are not so compatible with eth 2: genesis of eth 2 is based on contract events on some eth-1-like chain. And I assume this is the case for a testnet as well. Either based on a connection with a real net, or just some raw contract log event data to inject in an eth 2 genesis setup.
We're not quite there yet with cross-client testnets, so I prefer to just focus on the forks timeline, and ensure the standard has a way of telling tests what fork to use when running e.g. a state-transition test with some arbitrary slot number in it.
Also, since tests are just based on input data, and there's no need for genesis data in most tests, maybe we should just keep it lean. We can already define test-nets with just a combination of a fork-timeline and a pointer to a deposit contract on a testnet or something.

specs/test_formats/README.md
protolambda: See issue comment on genesis format / timelines. Let me know if there's any way to improve it, without making the format too heavy, or too far from an eth-2 type genesis.

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
djrtwo: Seems like all the clients use "testnet" in their articles about testnets. I also tend to write "testnet".

Lets go with no-space/no-dash

specs/test_formats/README.md
djrtwo: the 1.0 tests actually do a version of this by specifying the results of a test for multiple fork versions within the same file and leave them blank for unsupported forks. Very often this just ends up with overly verbose test files just repeating the same outputs IMO. (example here https://github.com/ethereum/tests/blob/develop/TransactionTests/ttVValue/V_equals38.json)

This instead is a list to make explicit which forks a test is intended to run against. So if it were an SSZ test for a feature that was supported in three forks, the three forks would be noted. If the feature changed in a forth fork, the forth fork would not be added to the list and instead a new tests would be written noting the forth fork.

hwwhww: Ahh sorry I misunderstood the "cover" here; I mixed up it with the timeline thing and thought you want to make one test case includes multiple state transition**s** with different forks. 😅 

Now I understand it's about compatibility and yes that makes sense to me. 😃 👍 👍 
FYI in eth1, they also use `>=` notation for compatibility like:
https://github.com/ethereum/tests/blob/5cef36e2844d4cfdc6b52362ba09261f0a09b76e/src/GeneralStateTestsFiller/stRevertTest/TouchToEmptyAccountRevert3Filler.json#L18

protolambda:
>  In eth2, it looks like we can define it in a list of Fork objects

@hwwhww Interesting, but stating both previous and current version for every entry in the list seems very verbose. Like spelling out a raw linked list. It does bring up an interesting point however: do we want forks based on epoch number (less flexible, current `Fork` format) or slot number (very exact, old `Fork` format)

specs/test_formats/README.md
protolambda: With conditions it's somewhat harder to parse. I'm open for it, but just feel like a list works well enough. And we won't have any ambiguity when there are two forks at the same time, and you need to decide what to include/exclude.

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
protolambda: Done, see 09cecca

hwwhww:
> The "genesis" and "accounts" sections are not so compatible with eth 2: genesis of eth 2 is based on contract events on some eth-1-like chain. And I assume this is the case for a testnet as well. Either based on a connection with a real net, or just some raw contract log event data to inject in an eth 2 genesis setup.

Yes, agreed that eth2 genesis format will be quite different from eth1. Just FYI. 🙂  

> a way of telling tests what fork to use when running e.g. a state-transition test with some arbitrary slot number in it.

For the state transition test, my gut tells me that one fork setting (i.e., set that this state transition happens with which fork) is good enough for *most* cases. But not sure if there's any case of mulit-fork timeline requirement?

> but stating both previous and current version for every entry in the list seems very verbose

Right, but state transition tests probably only need one `Fork` object!

> do we want forks based on epoch number (less flexible, current Fork format) or slot number (very exact, old Fork format)

I'd say based on current phase 0 design, it makes more sense to do with `epoch` only - unless we do another new fundemental change in the future... 😢

---

After all, the two different angles of test format designs so far:
1. Generic: as this PR, one general format for all cases
2. Categories: the current test formats, different formats for different tests (e.g., bls, shuffling, ssz)

I do like this solution more! 👍

djrtwo:
> I'd say based on current phase 0 design, it makes more sense to do with epoch only - unless we do another new fundemental change in the future...

Agreed -- epoch number.

Also agree that using the `Fork` data structure to specify a chain is overkill. We just maintain current and prev version in the BeaconState to ensure we can process signatures from right before and after the fork. The fork timeline list will naturally fit into the `Fork` construct as needed.

> But not sure if there's any case of mulit-fork timeline requirement?

Might be useful to test a fork in rapid succession (kind of like the constantinople and petersburg forks).

protolambda:
Changed fork timeline definition to use epoch numbers

specs/test_formats/README.md
protolambda: Fixed a bit of wording, and included example configs for fork-timelines and constant presets, with format description.

configs/constant_presets/minimal.yaml
```diff
@@ -0,0 +1,127 @@
+# Minimal preset
+
+
+# Misc
+# ---------------------------------------------------------------
+
+
+# Just 8 shards in the minimal testing setup
+SHARD_COUNT: 8
```
djrtwo: Let's have this file at least mirror the changes here -  https://github.com/ethereum/eth2.0-specs/blob/199e7849dafc48145dfbc6779ef2ccfa04e4a718/tests/phase0/conftest.py#L12-L22

specs/core/0_beacon-chain.md
djrtwo: use "mainnet" instead of "main-net". same for "test-net"

protolambda: Missed that, good spot

protolambda: Fixed, see 117e157

configs/constant_presets/minimal.yaml
protolambda: Fixed, see c5d2696

PR #880, opened 2019-04-05
RANDAO reveal slashing, custody period staggering and integration of custody and RANDAO reveals

dankrad:
Currently, there is no mechanism to punish validators for revealing their Randao reveals. This could lead to a situation where these can easily bought and sold without risk for a long time in the future, which aggravates the attacks on Randao.

The logic is very similar to the early custody reveal logic.


specs/core/0_beacon-chain.md
```diff
@@ -2310,6 +2327,44 @@ def process_attester_slashing(state: BeaconState,
...
+        message_hashes=message_hashes,
+        signature=randao_reveal_slashing.reveal,
+        domain=get_domain(
+            fork=state.fork,
+            epoch=randao_reveal_slashing.epoch,
+            domain_type=DOMAIN_RANDAO,
+        ),
+    )
+
+    assert randao_reveal_slashing.epoch > get_current_epoch(state)
```
djrtwo: This inequality was flipped. I believe it is correct now.

djrtwo: Do we want to allow for slashing in the case that it is just from 1 epoch in the future? Without a buffer, it is feasible that minor malicious reorgs could potentially be successful in including slashings. 

Say for example 64 builds on 62. Then 63 attempts to build on 62. 63 can then include the reveal from 64 as a slashing. Although they are likely to not make it into the canonical chain, times of medium latency might make the chances of success here non-negligible.

I would consider adding a constant `RANDAO_SLASHING_EPOCHS` or something similar that allows us to tune the constant via `assert randao_reveal_slashing.epoch > get_current_epoch(state) + RANDAO_SLASHING_EPOCHS` where `RANDAO_SLASHING_EPOCHS >= 1`

specs/core/0_beacon-chain.md
```diff
@@ -2310,6 +2327,44 @@ def process_attester_slashing(state: BeaconState,
...
+    """
+    revealer = state.validator_registry[randao_reveal_slashing.revealer_index]
+    masker = state.validator_registry[randao_reveal_slashing.masker_index]
+    pubkeys = [revealer.pubkey, masker.pubkey]
+    message_hashes = [
+        hash_tree_root(randao_reveal_slashing.epoch),
+        randao_reveal_slashing.mask,
+    ]
+
+    assert bls_verify_multiple(
```
hwwhww: suggest moving `assert bls_verify_multiple` to the last assertion since it's higher cost calculation than others.

vbuterin:
Can we think about this a lot more before we put something like this in?

I'm very worried about this approach because it means that if you propose a block, you are making yourself vulnerable to slashings if the chain forks, even if you did nothing wrong. In all other cases, the chain would need to reorg at least a week for that to happen.

I would propose this alternative: if you make an early randao reveal, you face a *small* penalty (eg. a week's worth of rewards), with a rule that this penalty can be applied a maximum of once every 72 days.

vbuterin:
Also want to think more about combining the two kinds of slashing reveals as a single message type and a single verification function.

vbuterin:
> Also want to think more about combining the two kinds of slashing reveals as a single message type and a single verification function.

I have an idea in that regard: the custody key for period N is also the randao subkey for the slot at the start of period N+2. Then, we could have a rule that revealing a randao subkey early (ie. any time before the start of period N+2) would only have a small penalty, but revealing any randao subkey more than one full period early (which coincides with when the randao subkey would be used as a PoC key if it's used that way) gets you slashed. So we get both covered with one condition.

dankrad:
> I have an idea in that regard: the custody key for period N is also the randao subkey for the slot at the start of period N+2. Then, we could have a rule that revealing a randao subkey early (ie. any time before the start of period N+2) would only have a small penalty, but revealing any randao subkey more than one full period early (which coincides with when the randao subkey would be used as a PoC key if it's used that way) gets you slashed. So we get both covered with one message type and one function.

Very interesting. I like this solution. So actually the custody key for the custody period starting at epoch N would then be the Randao reveal starting at epoch N+EPOCHS_PER_CUSTODY_PERIOD+2, right?

> I would propose this alternative: if you make an early randao reveal, you face a small penalty (eg. a week's worth of rewards), with a rule that this penalty can be applied a maximum of once every 72 days.
What's the reason for the 72 day rule?

dankrad:
I now unified this with the Custody reveal and removed the masking logic from the Custody Reveal scheme, since all slashing can now be performed through the RANDAO reveal logic. I had to introduce another constant to do this:

- The custody reveal is now the RANDAO reveal at period is now period * EPOCHS_PER_CUSTODY_PERIOD + CUSTODY_PERIOD_TO_RANDAO_PADDING (with #872 this will have to be adapted)
- if the reveal is less than RANDAO_SLASHING_EPOCHS (=2) early, it is invalid
- If the reveal is more than CUSTODY_PERIOD_TO_RANDAO_PADDING early, then the validator gets slashed
- otherwise, a minor penalty of 1 / RANDAO_REVEAL_PENALTY_QUOTIENT (1/32) is applied

dankrad:
Consensus of today was that we will merge this. Vitalik had some reservation about the length of CUSTODY_PERIOD_TO_RANDAO_PADDING and suggested possibly setting CUSTODY_PERIOD_TO_RANDAO_PADDING = EPOCHS_PER_CUSTODY_PERIOD. I prefer it being shorter, the code is written so that we can adjust this constant as we see fit.

specs/core/1_custody-game.md
```diff
@@ -168,8 +166,6 @@ This document details the beacon chain additions and changes in Phase 1 of Ether
     'revealer_index': ValidatorIndex,
     'period': 'uint64',
     'key': BLSSignature,
-    'masker_index': ValidatorIndex,
```
JustinDrake: Don't we want to completely nuke `CustodyKeyReveal`? That is, do all key reveals (early or not) with the same operation type.

specs/core/0_beacon-chain.md
```diff
@@ -1343,7 +1366,8 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
             withdrawable_epoch=FAR_FUTURE_EPOCH,
             initiated_exit=False,
             slashed=False,
-            high_balance=0
+            high_balance=0,
+            exposed_randao_reveals=[],
```
vbuterin: I dislike the idea of the validator record object not being constant-sized; it intuitively feels like a good invariant to keep that it's constant. In other places (eg. phase 1), we're keeping challenge records as separate lists too; would it make sense to make it be a separate list here too?

specs/core/0_beacon-chain.md
```diff
@@ -2310,6 +2341,60 @@ def process_attester_slashing(state: BeaconState,
...
+        message_hashes=message_hashes,
+        signature=randao_reveal_slashing.reveal,
+        domain=get_domain(
+            fork=state.fork,
+            epoch=randao_reveal_slashing.epoch,
+            domain_type=DOMAIN_RANDAO,
+        ),
+    )
+
+    if randao_reveal_slashing.epoch >= get_current_epoch(state) + CUSTODY_PERIOD_TO_RANDAO_PADDING:
```
vbuterin: As discussed in person, I'd support just setting CUSTODY_PERIOD_TO_RANDAO_PADDING to equal EPOCHS_PER_CUSTODY_PERIOD for simplicity and added conservatism.

specs/core/0_beacon-chain.md
```diff
@@ -2310,6 +2341,60 @@ def process_attester_slashing(state: BeaconState,
...
+        # Replacement for custody reveal slashing
+        slash_validator(state, randao_reveal_slashing.revealer_index, randao_reveal_slashing.masker_index)
+    else:
+        # Only a small penalty for RANDAO reveal that does not interfere with the custody period
+        penalty = get_effective_balance(state, randao_reveal_slashing.revealer_index) // RANDAO_REVEAL_PENALTY_QUOTIENT
```
```suggestion
        max_proposer_slot_reward = (
            get_base_reward(state, randao_reveal_slashing.revealer_index) *
            len(get_active_validators(state, get_current_epoch(state))) //
            EPOCH_LENGTH  //
            PROPOSER_REWARD_QUOTIENT
        )
        penalty = max_proposer_slot_reward * 2
```

vbuterin: Propose to make the penalty be proportional to the reward a proposer can receive in a slot.

specs/core/1_custody-game.md
vbuterin: I'd support that. Though if so, `verify_custody_key` would need to be removed, and instead phase 1 would have an instruction to change how `process_randao_reveal_slashing` works.

dankrad: Integrated this

specs/core/0_beacon-chain.md
dankrad: So I did this but I noticed there is a problem with this approach: If we do this, then we will have to wait for a whole custody period after the round key has been used to reveal the key (or else the validator would be subject to RANDAO slashing)! So as a compromise between the two I chose CUSTODY_PERIOD_TO_RANDAO_PADDING=2**8 (13 hours)

specs/core/0_beacon-chain.md
dankrad: OK, I assume you meant
```
        max_proposer_slot_reward = (
            get_base_reward(state, randao_key_reveal.revealed_index) *
            SLOTS_PER_EPOCH //
            len(get_active_validator_indices(state, get_current_epoch(state))) //
            PROPOSER_REWARD_QUOTIENT
        )
```
(the quotient is inverted)

specs/core/0_beacon-chain.md
dankrad: I changed this now to be a separate list. To make it a bit more efficient I use the epoch as a primary index, so in the typical case it should have very little overhead.

JustinDrake:
On second thought it may be best to slate all this logic for phase 1. Several reasons:

* RANDAO reveals are tightly integrated with phase 1 custody key reveals
* RANDAO reveals are not critical to phase 0
* RANDAO reveals are somewhat subtle because of the imperfect fault attribution so should not be rushed

All in all we'll have more room to manoeuvre by pushing this to phase 1. We'd also help ship phase 0 faster.

dankrad:
> * RANDAO reveals are not critical to phase 0
Well, they do determine the shuffling so not entirely uncritical, right?

> * RANDAO reveals are somewhat subtle because of the imperfect fault attribution so should not be rushed
In what way is fault attribution imperfect?

JustinDrake:
> not entirely uncritical, right?

RANDAO reveals allows to have a slightly weaker security model beyond pure honesty. It's hard to tell how critical it is. (As a side note, VDFs should clean this up.)

> In what way is fault attribution imperfect?

From the point of the view of the beacon chain the following two events are indistinguishable:

1) Malicious early reveal
2) Reorg with non-malicious non-early reveal included "early"

dankrad:
> (As a side note, VDFs should clean this up regardless.)
The problem is that with no punishment at all, any single validator can sell their RANDAO secrets and make an extra income without any fear. So even VDFs might not fix this if the secret market becomes too efficient.

> From the point of the view of the beacon chain the following two events are indistinguishable:
> 
>     1. Malicious early reveal
> 
>     2. Reorg with non-malicious non-early reveal included "early"

So in order to get punished at all, the reorg would have to be more than 1 full epoch (that is if the reveal was the first reveal of an epoch and the reorg went two the end of the previous previous epoch.

In order to get slashed, the reorg would have to be for a full 2**7 epochs.

I think these events are sufficiently unlikely that we have good attribution here.

JustinDrake:
> I think these events are sufficiently unlikely that we have good attribution here.

I'm expecting (potentially very major) reorgs in phase 0 (the experimental "testnet"). Phase 1 will hopefully be stable enough.

djrtwo:
> On second thought it may be best to slate all this logic for phase 1.

This is my intuition as well from the perspective of managing complexity and keeping the phase 1  custody components in phase 1.

dankrad:
Should we discuss this on the next call? I am happy either way, but would like to know the consensus before I make the adaptations for phase 1 integration.

JustinDrake:
Danny signalled in the last public call that moving RANDAO reveals to phase 1 makes sense to him. Does that make sense to you @vbuterin?

dankrad:
I merged #934 into this, so now this implements:
 - Merging of RANDAO and custody reveals
 - Punishing/slashing for early RANDAO reveals
 - Staggering of custody periods

specs/core/1_custody-game.md
```diff
@@ -509,16 +628,28 @@ def process_challenge_deadlines(state: BeaconState) -> None:
             records[records.index(challenge)] = CustodyBitChallengeRecord()
 ```
 
+Run `clean_up_exposed_randao_key_reveals(state)` after `process_final_updates(state)`:
+
+```python
+def clean_up_exposed_randao_key_reveals(state: BeaconState) -> None:
+
+    # Clean up exposed RANDAO key reveals
+    state.exposed_randao_reveals[current_epoch % RANDAO_PENALTY_MAX_FUTURE_EPOCHS] = []
```
vbuterin: Why not just make this line be part of `process_final_updates`?

specs/core/1_custody-game.md
```diff
@@ -233,13 +259,27 @@ def get_custody_chunk_bit(key: BLSSignature, chunk: bytes) -> bool:
...
 ```
 
-### `epoch_to_custody_period`
+### `get_randao_epoch_for_custody_period`
 
 ```python
-def epoch_to_custody_period(epoch: Epoch) -> int:
-    return epoch // EPOCHS_PER_CUSTODY_PERIOD
+def get_randao_epoch_for_custody_period(period: int, validator_index: ValidatorIndex) -> Epoch:
+    return (period + 1) * EPOCHS_PER_CUSTODY_PERIOD - validator_index % EPOCHS_PER_CUSTODY_PERIOD + CUSTODY_PERIOD_TO_RANDAO_PADDING
```
```suggestion
    next_period_start = (period + 1) * EPOCHS_PER_CUSTODY_PERIOD - validator_index % EPOCHS_PER_CUSTODY_PERIOD
    return next_period_start + CUSTODY_PERIOD_TO_RANDAO_PADDING
```

vbuterin: Makes the lines shorter and makes it clearer what's actually going on this way.

specs/core/1_custody-game.md
```diff
@@ -288,32 +329,93 @@ Add the following operations to the per-block processing, in order the given bel
...
+
+```python
+def process_randao_key_reveal(state: BeaconState,
+                              reveal: RandaoKeyReveal) -> None:
+    """
+    Process ``RandaoKeyReveal`` operation.
+    Note that this function mutates ``state``.
+    """
+    # Verify signature correctness
     assert verify_custody_key(state, reveal)
```
vbuterin: Maybe call it something other than `verify_custody_key`? Like `verify_randao_reveal` or something similar.

specs/core/1_custody-game.md
```diff
@@ -393,9 +494,14 @@ def process_bit_challenge(state: BeaconState,
...
-        revealer_index=challenge.responder_index,
-        period=epoch_to_custody_period(slot_to_epoch(attestation.data.slot)),
+    assert verify_custody_key(state, RandaoKeyReveal(
+        revealed_index=challenge.responder_index,
+        epoch=get_randao_epoch_for_custody_period(
+            get_validators_custody_reveal_period(
+                state=state,
+                index=challenge.responder_index,
+                epoch=slot_to_epoch(attestation.data.slot)),
+            challenge.responder_index),
```
vbuterin: IMO we don't even need the `verify_custody_key` function here; just use the one-liner we use in phase 0:

```python
    epoch=get_randao_epoch_for_custody_period(
            get_validators_custody_reveal_period(
                state=state,
                index=challenge.responder_index,
                epoch=slot_to_epoch(attestation.data.slot)),
            challenge.responder_index)
    assert bls_verify(responder.pubkey, hash_tree_root(epoch), challenge.responder_key, get_domain(state, DOMAIN_RANDAO, epoch))
```

This way we can remove the complexity of wrapping the custody key in a pointless reveal object with extraneous fields (the empty mask) and furthermore reduce the number of `verify_custody_key` calls to 1, which allows the function to be removed entirely (see comment above).

specs/core/1_custody-game.md
```diff
@@ -288,32 +329,93 @@ Add the following operations to the per-block processing, in order the given bel
...
-        # Revealer is active or exited
-        assert is_active_validator(revealer, get_current_epoch(state)) or revealer.exit_epoch > get_current_epoch(state)
-        revealer.custody_reveal_index += 1
-        revealer.max_reveal_lateness = max(revealer.max_reveal_lateness, current_custody_period - reveal.period)
-        proposer_index = get_beacon_proposer_index(state)
-        increase_balance(state, proposer_index, base_reward(state, index) // MINOR_REWARD_QUOTIENT)
-
-    # Case 2: masked punitive early reveal
+        # Handle the case of a regular custody reveal
+        process_custody_reveal(state, reveal)
```
vbuterin: Maybe for symmetry put the logic below this into a separate function  `process_randao_key_punitive_reveal` and have the top-level function just be:

```
    revealed_validator = state.validator_registry[reveal.revealed_index]

    if reveal.mask == ZERO_HASH: 
        pubkeys = [revealed_validator.pubkey]
        message_hashes = [hash_tree_root(reveal.epoch)]
    else:
        masker = state.validator_registry[reveal.masker_index]
        pubkeys = [revealed_validator.pubkey, masker.pubkey]
        message_hashes = [
            hash_tree_root(reveal.epoch),
            reveal.mask,
        ]

    assert bls_verify_multiple(
        pubkeys=pubkeys,
        message_hashes=message_hashes,
        signature=reveal.reveal,
        domain=get_domain(
            state=state,
            domain_type=DOMAIN_RANDAO,
            message_epoch=reveal.epoch,
        ),
    )
    if reveal.mark == ZERO_HASH:
        process_custody_reveal(state, reveal)
    else:
        process_randao_key_punitive_reveal(state, reveal)
```

This, plus the suggestion below, would allow the `verify_custody_key` function to be cut out entirely.

vbuterin: Though honestly if we have this parallel structure the right approach may be to just have two objects, a `CustodyKeyReveal` and a `EarlyReveal`, where only the latter has the masking mechanism and the former does not (the former would just be a thin wrapper `{validator_index: uint64, custody_key: bytes96}`). This way we don't need the empty mask fields in legitimate key reveals, and the code can be separated out more instead of being an if statement that runs totally different code depending on what role the object is serving.

specs/core/1_custody-game.md
dankrad: It essentially is -- I thought it would be nicer in terms of testing to have a functional separation of phase 0 and phase1 stuff. Happy to go back to the "patch" approach, but we will have to find a way to make it work with the function puller

dankrad: Done it your way for now, I don't really have a strong opinion :)

specs/core/1_custody-game.md
dankrad: I did this using two objects now. Pls review and check if this is your preferred method :)

specs/core/1_custody-game.md
dankrad: Yes, done

specs/core/1_custody-game.md
dankrad: Removed the function now as it seemed pointless

dankrad:
TBD after call:
Randao key reveal -> Early derived key reveal
Make penalty proportional to the number of validators already punished for epoch

dankrad:
The penalty for the early RANDAO reveal (that is not valid as a custody key anymore) is now proportional to the number of secrets already exposed. This is the simplest way of increasing punishment for a coordinated attack, and probably quite good because RANDAO reveals are useful to an attacker if they are all for the same epoch by different validators.
Possible changes:
 - This is a bit unfair because the first one to get exposed gets the smallest penalty. We could extend it to also work retroactively on the ones which are already exposed, at the cost of a bit of complexity [and having to think if we want to let them exit if a validator has a secret exposed]
 - Another approach would be to make it proportional to all reveals (not just from the same epoch)

specs/core/1_custody-game.md
```diff
@@ -233,13 +274,28 @@ def get_custody_chunk_bit(key: BLSSignature, chunk: bytes) -> bool:
...
+                                         epoch: Epoch=None) -> int:
+    '''
+    This function returns the reveal period for a given validator.
+    If no epoch is supplied, the current epoch is assumed.
+    Note: This function implicitly requires that validators are not removed from the
+    validator set in fewer than EPOCHS_PER_CUSTODY_PERIOD epochs
+    '''
+    epoch = get_current_epoch(state) if epoch is None else epoch
+    return (epoch + validator_index % EPOCHS_PER_CUSTODY_PERIOD) // EPOCHS_PER_CUSTODY_PERIOD
 ### `replace_empty_or_append`
```
````suggestion
```

### `replace_empty_or_append`

````

This code-block is not closed.

File: specs/core/0_beacon-chain.md [hunk 13]
```diff
@@ -234,6 +236,9 @@ These configurations are updated for releases, but may be out of sync during `de
 | `MIN_VALIDATOR_WITHDRAWABILITY_DELAY` | `2**8` (= 256) | epochs | ~27 hours |
 | `PERSISTENT_COMMITTEE_PERIOD` | `2**11` (= 2,048)  | epochs | 9 days  |
 | `MAX_CROSSLINK_EPOCHS` | `2**6` (= 64) | epochs | ~7 hours |
+| `RANDAO_PENALTY_EPOCHS` | `2` | epochs | 12.8 minutes |
```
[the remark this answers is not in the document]
dankrad: Now in 1_custody-game -- looks consistent to me? What is the problem?

specs/core/0_beacon-chain.md
```diff
@@ -2161,6 +2190,74 @@ def process_attester_slashing(state: BeaconState,
...
+        ),
+    )
+
+    if randao_key_reveal.epoch >= get_current_epoch(state) + CUSTODY_PERIOD_TO_RANDAO_PADDING:
+        # Replacement for custody reveal slashing
+        slash_validator(state, randao_key_reveal.revealer_index, randao_key_reveal.masker_index)
+    else:
+        # Only a small penalty for RANDAO reveal that does not interfere with the custody period
+        penalty = get_effective_balance(state, randao_key_reveal.revealer_index) // RANDAO_REVEAL_PENALTY_QUOTIENT
+        proposer_index = get_beacon_proposer_index(state, state.slot)
```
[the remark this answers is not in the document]
dankrad: The logic is quite different now and much less is shared. Suggest factoring out if we add similar penalties in the future.

specs/core/1_custody-game.md
```diff
@@ -79,24 +84,32 @@ This document details the beacon chain additions and changes in Phase 1 of Ether
 | Name | Value | Unit | Duration |
 | - | - | :-: | :-: |
 | `MAX_CHUNK_CHALLENGE_DELAY` | `2**11` (= 2,048) | epochs | ~9 days |
-| `EPOCHS_PER_CUSTODY_PERIOD` | `2**11` (= 2,048) | epochs | ~9 days |
 | `CUSTODY_RESPONSE_DEADLINE` | `2**14` (= 16,384) | epochs | ~73 days |
+| `RANDAO_PENALTY_EPOCHS` | `2` | epochs | 12.8 minutes |
```
```suggestion
| `RANDAO_PENALTY_EPOCHS` | `2**1` (= 2) | epochs | 12.8 minutes |
```

specs/core/1_custody-game.md
```diff
@@ -79,24 +84,32 @@ This document details the beacon chain additions and changes in Phase 1 of Ether
...
+| `EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS` | `2**14` | epochs | ~73 days |
```
```suggestion
| `EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS` | `2**14` (= 16, 384) | epochs | ~73 days |
```

specs/core/1_custody-game.md
```diff
@@ -79,24 +84,32 @@ This document details the beacon chain additions and changes in Phase 1 of Ether
...
+| `EPOCHS_PER_CUSTODY_PERIOD` | `2**11` (= 2,048) | epochs | ~9 days |
+| `CUSTODY_PERIOD_TO_RANDAO_PADDING` | `2**11` (= 2,048) | epochs | ~9 days |
+| `MAX_REVEAL_LATENESS_DECREMENT` | `2**7` (= 128) | epochs | ~14 hours |
 
 ### Max operations per block
 
 | Name | Value |
 | - | - |
 | `MAX_CUSTODY_KEY_REVEALS` | `2**4` (= 16) |
+| `MAX_EARLY_DERIVED_SECRET_REVEALS` | `1` |
```
```suggestion
| `MAX_EARLY_DERIVED_SECRET_REVEALS` | `2**0` (=1) |
```

specs/core/1_custody-game.md
```diff
@@ -79,24 +84,32 @@ This document details the beacon chain additions and changes in Phase 1 of Ether
...
 | `MAX_CUSTODY_CHUNK_CHALLENGES` | `2**2` (= 4) |
 | `MAX_CUSTODY_BIT_CHALLENGES` | `2**2` (= 4) |
 | `MAX_CUSTODY_RESPONSES` | `2**5` (= 32) |
 
+### Reward and penalty quotients
+
+| `EARLY_DERIVED_SECRET_REVEAL_SLOT_REWARD_MULTIPLE` | `2` |
```
```suggestion
| `EARLY_DERIVED_SECRET_REVEAL_SLOT_REWARD_MULTIPLE` | `2**1` (= 2) |
```

PR #884, opened 2019-04-07
Homogenised start shard

JustinDrake:
When the committee count equals `SHARD_COUNT`, add an offset. Fix item 28 in #675.

Screenshot 2019-04-07 at 10 47 39

(Went with a slightly improved approach which still allows for committee count to equal `SHARD_COUNT`.)

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -69,6 +69,7 @@
         - [`get_epoch_committee_count`](#get_epoch_committee_count)
         - [`compute_committee`](#compute_committee)
         - [`get_current_epoch_committee_count`](#get_current_epoch_committee_count)
+        - [`get_epoch_start_shard`](#get_epoch_start_shard)
         - [`get_crosslink_committees_at_slot`](#get_crosslink_committees_at_slot)
         - [`get_block_root`](#get_block_root)
         - [`get_state_root`](#get_state_root)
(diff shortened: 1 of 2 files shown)


djrtwo:
Not terribly happy about the cleanliness here, but it works.

vbuterin:
This seems far too bulky for the purpose. Why not just:

```
def get_shard_delta(active_validator_count: int) -> int:
    committee_count = get_epoch_committee_count(active_validator_count)
    return min(committee_count, SHARD_COUNT - SHARD_COUNT // EPOCH_LENGTH)
```

Alternatively, we can replace the last line with `return committee_count if committee_count < SHARD_COUNT else SHARD_COUNT // EPOCH_LENGTH` to keep exact functionality.

djrtwo:
Is this what you meant?

JustinDrake:
I'd say `return committee_count if committee_count < SHARD_COUNT else SHARD_COUNT // EPOCH_LENGTH` is better than `min(committee_count, SHARD_COUNT - SHARD_COUNT // EPOCH_LENGTH)`. The reason is that the former allows for all shards to get crosslinked in every epoch whereas the latter always leaves `SHARD_COUNT // EPOCH_LENGTH` shards behind.

vbuterin:
Ah no, I meant not adding any of the other logic in the PR and instead _just_ adding the get_shard_delta function.

djrtwo:
closed in favor of #889

PR #886, opened 2019-04-08
Prevents transfers to active validators

CarlBeek:
Currently it is possible to transfer to active validators. This should not be the case as it allows for rapid changes to the validator weights. This PR changes that. (It also fixes a minor typo.)


specs/core/0_beacon-chain.md
```diff
@@ -2428,11 +2428,16 @@ def process_transfer(state: BeaconState, transfer: Transfer) -> None:
...
     # A transfer is valid in only one slot
     assert state.slot == transfer.slot
-    # Only withdrawn or not-yet-deposited accounts can transfer
+    # Only withdrawn or not-yet-activated accounts can transfer
     assert (
         get_current_epoch(state) >= state.validator_registry[transfer.sender].withdrawable_epoch or
         state.validator_registry[transfer.sender].activation_epoch == FAR_FUTURE_EPOCH
     )
+    # Verify that the validator recieving the transfer is either withdrawn or not-yet-activated
+    assert (
```
JustinDrake: Why not `assert not is_active_validator(state.validator_registry[transfer.recipient])`?

specs/core/0_beacon-chain.md
```diff
@@ -2428,11 +2428,9 @@ def process_transfer(state: BeaconState, transfer: Transfer) -> None:
     )
     # A transfer is valid in only one slot
     assert state.slot == transfer.slot
-    # Only withdrawn or not-yet-deposited accounts can transfer
-    assert (
-        get_current_epoch(state) >= state.validator_registry[transfer.sender].withdrawable_epoch or
-        state.validator_registry[transfer.sender].activation_epoch == FAR_FUTURE_EPOCH
-    )
+    # Verify sender and receipient are not active
+    assert not is_active_validator(state.validator_registry[transfer.sender])
```
JustinDrake: That's not going to work because senders are liable to being slashed between exiting and withdrawing so we don't want them to transfer their balance away.

CarlBeek:
With the new balance structure this is not needed. :)

PR #889, opened 2019-04-09
Homogenised start shard

vbuterin:
Alternative presentation to #884, should be substantively equivalent

This addresses item 28 in #675. This ensures that in the case where all shards are crosslinked each epoch that asymmetries between shards being at the start or end of the epoch do not persist.


djrtwo:
I'll port the tests after my talk tomorrow!

PR #890, opened 2019-04-10
limit bit-length of justification bitfield to strict 64

protolambda:
Limit bit-length of justification bitfield to strict 64, prevent SSZ encoding crash due to too large integer size.

The pyspec crashed after running it for a long series of epochs because of this :(. It's not a pretty fix, but otherwise Python just grows the integer out of uint64 bounds, making it non-encodeable in 64 bits.


vbuterin:
Maybe compress both lines into `state.justification_bitfield = (state.justification_bitfield << 1) % 2**64`?

PR #891, opened 2019-04-10
Increases possible balance at stake to 64ETH

CarlBeek:
Prerequisite: #886 (Prevents transfers to active validators)
Closes: #730

As it says in the tin, this increases the balance at stake to 64ETH. This means that rewards compound when validator balances are > 32 ETH. Activations are still limited to 32 ETH so the only way to reach > 32ETH as an active validator is by receiving rewards.


dankrad:
I support this :)

prestonvanloon:
Can you add a bit more context to this issue? 32 ETH deposit = 64 ETH at stake?

CarlBeek:
Maybe I titled the PR poorly. What I meant is "Increases **possible** balance at stake to 64ETH". The idea is that a validator can deposit 32 ETH, but by being an honest validator, they can accrue up to 64 ETH whilst obtaining compounding interest (after which their balance will increase linearly). I have a longer discussion in #730.

The difference is basically in how these two validators' balances increase over time:
econ_test_init_bal32

dankrad:
So I was talking this through with Carl over lunch, and I have another suggestion: How would it be if we instead left the cap at 32 Eth for the balance at stake, but made all balances over 32 Eth transferrable? In that case, the balance over 32 would no longer be locked up and can be used, so there is no need to reward it.

In the same spirit, is there a reason why the lower limit for being ejected is as low as 32 Eth? This is an awful lot of penalties before someone gets ejected. If we made it something like 28 Eth, it would let us take the assumption that all votes are equal, simplifying the logic quite a bit.

CarlBeek:
> So I was talking this through with Carl over lunch, and I have another suggestion...

What the basic proposal here is is to restrict the range over which a validator's balance could exist. By limiting everything to [32, 28], we could assume all validators have the same balance and thus not bother calculating exactly what balance is attesting to what. (Maybe only Light Clients make this assumption.)

The upper bound is achieved by allowing the portion of the balance that is over 32 to be immediately withdrawal-able and only using >=32ETH as the amount at stake. This allows for the compounding growth I keep hankering after by means of withdrawing into new validator instances (or alternative Defi solutions).

The lower bound would be enforced by ejections, as is done now. This proposal is significantly more stringent that the current one, but I argue that the current one is too low. At the moment a validator needs to be offline for a ridiculously long amount of time before being ejected.

Additionally, this proposal would remove the need to prevent deposits & transfers to active validators.

This idea can be taken even further by assuming that the majority of validators will have a balance >=32 and thus the need for rounded balances is greatly reduced. (As the majority of balances would not change.)

CarlBeek:
After discussion today, there are better alternatives that achieve the same goals and more.

PR #893, opened 2019-04-10
Add union type and null to SSZ

dankrad:
This is a suggestion to add an option type and a null type to the SSZ spec. While we will not use it in the spec at the moment, this could be a very simple addition that would make SSZ much more powerful. 

In a discussion with @mslipper it transpired that this was one of the main stumbling blocks for making more extensive use of SSZ in the networking. I wonder if this would solve the main problems?


specs/simple-serialize.md
```diff
@@ -70,7 +73,13 @@ assert value in (True, False)
 return b"\x01" if value is True else b"\x00"
 ```
 
-### Vectors, containers, lists
+### `"null"`
+
+```python
+return b""
```
jannikluhn: I think this would run into the same problem as [this one](https://github.com/ethereum/eth2.0-specs/issues/854). So either we should define null as some non-empty string (e.g., `b"\x00"`) or allow it only as a container or vector element.

specs/simple-serialize.md
```diff
@@ -87,6 +96,17 @@ serialized_length = len(serialized_bytes).to_bytes(BYTES_PER_LENGTH_PREFIX, "lit
...
 ```
 
+If `value` is an option type:
+
+Define value as an object that has properties `value.value` with the contained value, and `value.type_index` which indexes the type.
+
+```python
+serialized_bytes = serialize(value.value)
+serialized_type_index = value.type_index.to_bytes(BYTES_PER_LENGTH_PREFIX, "little")
+return serialized_length + serialized_bytes
```
```suggestion
return serialized_type_index + serialized_bytes
```

specs/simple-serialize.md
dankrad: Yes, I agree. I think null only makes sense inside an option type and then it should just be an empty string (so it does not take any space in serialization). And we should forbid all other empty container objects.

snario:
Hi there! I’m just commenting to make a point that in state channels we need to have a single data type to represent the state in the channel and we need it to be of dynamic length. Also we need this to be in solidity early as possible

specs/simple-serialize.md
vbuterin: To clarify, you mean keep this code exactly as is, so in any actual implementation the rules for union types below would apply, and a `null` would resolve as just the 4 bytes representing whatever its index in the union type is?

dankrad: Yes, exactly.

mslipper:
This would solve _some_ of the issues in the networking spec, but not all of them. While we can now represent nullable values (which would clean up some of the data structures in the networking spec), it doesn't solve the problem of needing to know an upfront schema to serialize/deserialize SSZ-encoded data. For things like the RPC wrapper, we need a serialization scheme that can model something similar to generics in object-oriented languages. Right now, the easiest way to do this is to not use any serialization scheme for the RPC wrapper and instead just prepend bytes to whatever SSZ-encoded object is being sent.

Since the introduction of a `null` type helps us clean up data structures elsewhere, I'd still like to merge this.

dankrad:
> For things like the RPC wrapper, we need a serialization scheme that can model something similar to generics in object-oriented languages.

Would it not be possible to do this using the Union type? If the call can take typs A, B, C or D, you can do this bye defining a union type of Union(A, B, C, D) and defining your RPC call on that. The type index also makes it easy to add more types to the union later without breaking compatibility with the old type.

mslipper:
> Would it not be possible to do this using the Union type? If the call can take typs A, B, C or D, you can do this bye defining a union type of Union(A, B, C, D) and defining your RPC call on that. The type index also makes it easy to add more types to the union later without breaking compatibility with the old type.

This is possible, but unwieldy - there are something like 7 different types that the RPC wrapper can include in the request/response body field. Defining a union type on all of them seems error-prone and difficult to maintain.

dankrad:
@JustinDrake @vbuterin I am not sure if you think this is a strong enough endorsement to merge this change. I would be in favour of it as it is fairly simple and will not change anything in the consensus layer, but I don't feel very strongly either way.

djrtwo:
Should we discuss this on tomorrow's call?

dankrad:
> Should we discuss this on tomorrow's call?

Sounds like a good idea

Nashatyrev:
Don't we need to adjust 'fixed/variable-size' definition?
I believe we should treat the union as a variable-size type
Should we treat the union of fixed-size types of the same size as fixed-type?

dankrad:
> Don't we need to adjust 'fixed/variable-size' definition?

I think it is already correct?

> I believe we should treat the union as a variable-size type

Yes.

> Should we treat the union of fixed-size types of the same size as fixed-type?

Hmm, I can see that, but not sure if it would be hugely beneficial to add this special case. I can see it would simplify things if we say that a union of several fixed-size object has the (fixed) size of the max size of those objects.

Nashatyrev:
>>    I believe we should treat the union as a variable-size type

> Yes.

`
We recursively define "variable-size" types to be lists and all types that contains a variable-size type. All other types are said to be "fixed-size"
`
According to this statement a union of fixed-size types is itself fixed-size. Am I missing something?

dankrad:
>  According to this statement a union of fixed-size types is itself fixed-size. Am I missing something?

Ah, yes, I overlooked that sentence. Fixed now.

JustinDrake:
Can we alias `null` to `{}` to avoid creating a new type?

dankrad:
> Can we alias `null` to `{}` to avoid creating a new type?

You mean the empty container? That sounds like an interesting idea.

JustinDrake:
> You mean the empty container?

Yes

dankrad:
> Can we alias `null` to `{}` to avoid creating a new type?

Done.

specs/simple-serialize.md
```diff
@@ -97,6 +113,16 @@ fixed_parts = [part if part != None else variable_offsets[i] for i, part in enum
 return b"".join(fixed_parts + variable_parts)
 ```
 
+If `value` is an union type:
```
hwwhww:
```suggestion
If `value` is a union type:
```

specs/simple-serialize.md
```diff
@@ -97,6 +113,16 @@ fixed_parts = [part if part != None else variable_offsets[i] for i, part in enum
...
+
+Define value as an object that has properties `value.value` with the contained value, and `value.type_index` which indexes the type.
+
+```python
+serialized_bytes = serialize(value.value)
+serialized_type_index = value.type_index.to_bytes(BYTES_PER_LENGTH_PREFIX, "little")
```
hwwhww: There's only `BYTES_PER_LENGTH_OFFSET` after SOS.

```suggestion
serialized_type_index = value.type_index.to_bytes(BYTES_PER_LENGTH_OFFSET, "little")
```

specs/simple-serialize.md
```diff
@@ -41,8 +42,14 @@ This is a **work in progress** describing typing, serialization and Merkleizatio
     * angle bracket notation `[type, N]`, e.g. `["uint64", N]`
 * **list**: ordered variable-length homogeneous collection of values
     * angle bracket notation `[type]`, e.g. `["uint64"]`
+* **union**: union type containing one of the given subtypes
+    * round bracket notation `(type1, type2, ...)`, e.g. `("uint64", "null")`
 
-We recursively define "variable-size" types to be lists and all types that contains a variable-size type. All other types are said to be "fixed-size".
+We recursively define "variable-size" types to be lists and unions and all types that contain a variable-size type. All other types are said to be "fixed-size".
```
hwwhww: Suggestion: add a header for this sentence. (and also update ToC if it's applied)

```suggestion
#### Variable-size and fixed-size

We recursively define "variable-size" types to be lists and unions and all types that contain a variable-size type. All other types are said to be "fixed-size".
```

specs/simple-serialize.md
```diff
@@ -97,6 +113,16 @@ fixed_parts = [part if part != None else variable_offsets[i] for i, part in enum
 return b"".join(fixed_parts + variable_parts)
 ```
 
+If `value` is an union type:
+
+Define value as an object that has properties `value.value` with the contained value, and `value.type_index` which indexes the type.
```
hwwhww: IMO maybe too many "value"s in these sentences and code. Perhaps using `value.contained_value` would be more readable?

JustinDrake: Shall we use `object` instead? (That's what these things are after all, objects.)

dankrad: Feels wrong though to change the naming of the property though, it's really the first "value" that is only a placeholder.

dankrad: You mean `value.object` or `object.value` @JustinDrake ?

JustinDrake: I was thinking `object.value`

dankrad: I'm happy to go with that

dankrad: Oh but 'object' is a type in python :(

specs/simple-serialize.md
```diff
@@ -51,18 +57,22 @@ For convenience we alias:
 * `"byte"` to `"uint8"` (this is a basic type)
 * `"bytes"` to `["byte"]` (this is *not* a basic type)
 * `"bytesN"` to `["byte", N]` (this is *not* a basic type)
+* `"null"`: `{}`, i.e. the empty container
 
 ### Default values
```
JustinDrake: Default values for `union` types are not defined.

dankrad: Good point, I would suggest making it the the first type in the union, and requiring that if null is present, it should be the first type.

PR #895, opened 2019-04-11
Clean up light client spec

hwwhww:


djrtwo:
@hwwhww You have this marked as "Draft". Is there anything else to do before merge?

hwwhww:
@djrtwo yessss will add more commits!

specs/light_client/sync_protocol.md
```diff
@@ -129,21 +133,20 @@ def compute_committee(header: BeaconBlockHeader,
         [i for i in earlier_committee if epoch % PERSISTENT_COMMITTEE_PERIOD < get_switchover_epoch(i)] +
         [i for i in later_committee if epoch % PERSISTENT_COMMITTEE_PERIOD >= get_switchover_epoch(i)]
```
hwwhww: @vbuterin 
I think they should be:
```python
        [i for i in actual_earlier_committee if epoch % PERSISTENT_COMMITTEE_PERIOD < get_switchover_epoch(i)] +
        [i for i in actual_later_committee if epoch % PERSISTENT_COMMITTEE_PERIOD >= get_switchover_epoch(i)]
```
?

specs/light_client/merkle_proofs.md
```diff
@@ -130,35 +137,35 @@ def get_proof_indices(tree_indices: List[int]) -> List[int]:
...
-````
+```
 
 Generating a proof is simply a matter of taking the node of the SSZ hash tree with the union of the given generalized indices for each index given by `get_proof_indices`, and outputting the list of nodes in the same order.
 
 Here is the verification function:
 
 ```python
-def verify_multi_proof(root, indices, leaves, proof):
+def verify_multi_proof(root: Bytes32, indices: List[int], leaves: List[Bytes32], proof: List[bytes]) -> bool:
```
djrtwo: Is a `proof` not a list of hashes -- `List[Bytes32]`?

```suggestion
def verify_multi_proof(root: Bytes32, indices: List[int], leaves: List[Bytes32], proof: List[Bytes32]) -> bool:
```

PR #900, opened 2019-04-12
Update python poc link

terenc3t:
Should we update the python poc link to https://github.com/ethereum/trinity/tree/master/eth2?


JustinDrake:
I'm tempted to nuke the "References" and "Copyright" sections.

PR #902, opened 2019-04-12
Move the ordering of the definition of `process_deposit`.

ralexstokes:
This move matches the symmetry of the other operation processing definitions.

Note: this PR *only* changes the ordering of the text in the document, not any content.


djrtwo:
`process_deposit` is also used in `get_genesis_beacon_state` thus the def in helpers. I'm inclined to keep it that way

PR #912, opened 2019-04-13
Leftover DepositInput → DepositData

JSON:
Due to https://github.com/ethereum/eth2.0-specs/pull/780:
DepositInput → DepositData
deposit_input → deposit_data

For line 107, serialize(deposit_input) is changed to serialize(deposit.data) since . rather than _ is used here: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#deposits Please correct if wrong


djrtwo:
`Deposit` is a higher level wrapper created by the deposit contract. The validator sends the serialized object `DepositData` (named `deposit_data` here) as the payload of the eth1 tx so it should be `serialize(deposit_data)`. I'll clean it up

JSON:
thank you!

PR #916, opened 2019-04-14
fix transfer invariant, credits to @holiman for finding the edge case

protolambda:
See https://github.com/protolambda/zrnt/issues/4


protolambda:
@holiman Yes, thought about it, and asked Justin if allowing self-transfers (but with adjusted rules) would be better. Essentially, the beacon-chain is just a system chain, and the possibility of not having transfers for phase-0 is still up for discussion. Hence the choice for a more straight-forward fix to get the edge-case over with, in a simple way. If we would want beacon transfers for actual more eth1-like functionality, I'm sure we would go for the other more complete solution you mentioned.

PR #917, opened 2019-04-14
Cleanup containers

JustinDrake:
* Remove unnecessary per-field comments, focus on comments for field grouping
* Group `Validator` fields relevant to light-clients together (small optimisation for light clients)
* Rework grouping of `BeaconState` fields to "tell a story":
    * `# Versioning` (in space and time) specifies a blockchain
    * `# History` (block/state roots) is the "bones" of any blockchain
    * `# Eth1` voting leads to deposits
    * `# Registry` emerges from deposits
    * `# Shuffling` the registry leads to committees
    * `# Attestations` are committee votes
    * `# Crosslinks` emerge from attestations
    * `# Finality` also emerges from attestations
* `deposit_index` => `eth1_deposit_index`

Do not merge before:

* #695 which specifies custom types for individual container fields, offsetting the removal of comments in some instances
* #896 and #843 so that we don't have to continuously maintain the genesis `BeaconState`
* #874 which introduces `current_crosslinks` and `previous_crosslinks`
* #865 which renames `proof_of_possession` to `signature`

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -534,11 +534,16 @@ The types are defined topologically to aid in facilitating an executable version
     'slot': 'uint64',
     'fork': Fork,
 
-    # Roots
+    # History
     'latest_block_header': BeaconBlockHeader,
     'latest_block_roots': ['bytes32', SLOTS_PER_HISTORICAL_ROOT],
     'latest_state_roots': ['bytes32', SLOTS_PER_HISTORICAL_ROOT],
     'historical_roots': ['bytes32'],
+
+    # Eth1
+    'latest_eth1_data': Eth1Data,
+    'eth1_data_votes': [Eth1DataVote],
+    'eth1_deposit_index': 'uint64',
 
     # Registry
     'validator_registry': [Validator],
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -411,36 +392,27 @@ The types are defined topologically to aid in facilitating an executable version
 
 ```python
 {
-    # BLS public key
+    # Fields relevant to light clients
```
djrtwo: Not sure I agree with the grouping here. pubkey and slashed are relevant to far more than light clients

dankrad:
So I kind of disagree with this. I believe it is good to have a human-readable interpretation of a field next to it, I just think the comments should be clearer than they are now. This goes exactly against what I propose in #942. 

My suggestion is that in an ideal world, a field has a corresponding description that can be checked against the actual implementing algorithm. Let's give an example:
```python
{
    # Validator indices
    'custody_bit_0_indices': ['uint64'],
    'custody_bit_1_indices': ['uint64'],
    # Attestation data
    'data': AttestationData,
    # Aggregate signature
    'aggregate_signature': 'bytes96',
}
```
I agree that the comment "Validator indices" is useless. However, if we change it to
```python
{
    # Indices of all the validators that voted attested with custody bit = 0
    'custody_bit_0_indices': ['uint64'],
    # Indices of all the validators that voted attested with custody bit = 1
    'custody_bit_1_indices': ['uint64'],
    # Attestation data
    'data': AttestationData,
    # Aggregate signature
    'aggregate_signature': 'bytes96',
}
```
then it actually becomes more readable and we can use this information to check whether the algorithm computing this actually does what it claims.

protolambda:
@JustinDrake 
Now that #695 and #1077 are complete, can we get this PR ready again, if you still like to merge it? It needs an update to account for the new fields, and the syntax is effectively just `name: type` now.

JustinDrake:
Close in favour of #1156

PR #919, opened 2019-04-14
Cleaner dust checking in transfers

JustinDrake:
Inspired by [this comment](https://github.com/ethereum/eth2.0-specs/pull/916#commitcomment-33170877).


protolambda:
@holiman Looks like Justin liked your idea. LGTM

holiman:
Now zero is disallowed - is that intentional? Just checking...

holiman:
Ah no sorry, disregard that!

Issue #906, opened 2019-04-12
Bug in proposer reward


djrtwo:
When applying proposer rewards within an epoch transition in `get_justification_and_finalization_deltas`, we need to get the proposer index from the _previous_ epoch. Right now there is an assertion in `get_beacon_proposer_index` that it has to be from the current epoch. This was added when we changed the sampling of proposer to be based upon balance because the balance can only be deterministically know in the current epoch.

If we remove the assertion, we can't guarantee that the proposer we fetch from the previous epoch is actually correct (although it will be 99.9% of the time).

```python
# Proposer bonus
if index in get_attesting_indices(state, state.previous_epoch_attestations):
    proposer_index = get_beacon_proposer_index(state, inclusion_slot(state, index))
```

```python
def get_beacon_proposer_index(state: BeaconState,
                              slot: Slot) -> ValidatorIndex:
    current_epoch = get_current_epoch(state)
    assert slot_to_epoch(slot) == current_epoch
```

Going to bed. Will think about this tomorrow

djrtwo:
@JustinDrake Suggests just giving proposer rewards immediately. We already do this with Transfers and with Slashing rewards. This avoids the issues of having to calculate proposers outside of the epoch in which they propose and adds negligible effects to state hashing.

cc: @vbuterin

JustinDrake:
So @djrtwo found out that we must process proposer rewards immediately because every block can change the balance.

JustinDrake:
Fixed in #920.

PR #920, opened 2019-04-14
attesation proposer rewards to block processing

djrtwo:
Handles #906 

* move attestation proposer rewards to block processing
* beacon block proposer can only be deterministically checked during the slot in question because balances might change at every block. thus removal of `slot` param from `get_beacon_proposer_index`
* extend some helper functionality to be more modular to be used in this context

todo: write some tests


hwwhww:
```python
- def get_beacon_proposer_index(state: BeaconState,
-                               slot: Slot) -> ValidatorIndex:
+ def get_beacon_proposer_index(state: BeaconState) -> ValidatorIndex:
```

But it is useful for block proposal!
https://github.com/ethereum/eth2.0-specs/blob/442aa0e7f34012b37e61f19e156246694ac76997/specs/validator/0_beacon-chain-validator.md#block-proposal

JustinDrake:
> But it is useful for block proposal!

It's a fundamental constraint with the current proposer sampling logic.

Screenshot 2019-04-16 at 14 30 15

JustinDrake:
Removed ~30 lines and made things more readable... but broke the tests :joy:

PR #923, opened 2019-04-14
Update 0_beacon-chain.md

ralexstokes:
Fix another typo w/ the withdrawal ~> exit queue PR

Specifically: the rest of this code expects the index, not the validator object so we collect the index in the list comprehension in lieu of the validator object


JustinDrake:
Thank you Stokes :) How did we miss all these bugs? :joy: Need more testing @djrtwo @protolambda

PR #924, opened 2019-04-14
Remove serialization from consensus

JustinDrake:
Consensus now only cares about Merkleisation (i.e. `hash_tree_root`), not about serialization (i.e. `serialize`). This simplifies consensus code by a few tens of lines, is conceptually cleaner, and is more future proof.

A corresponding [change in the deposit contract](https://github.com/ethereum/deposit_contract/pull/31).


hwwhww:
@JustinDrake 
In the current deposit contract, the API is `def deposit(deposit_data: bytes[184])`, where the input is the serialized `DepositData` and calculate the leaf value by `sha3` in the contract.

So regarding the deposit contract side, my understanding is:
1. Keep having serialization input -> build tree on chain
2. Need to implement tree hash on contract (only for `deposit_data` case)

Is it what you think?

JustinDrake:
My suggestion would be to change the deposit contract API to `def deposit(pubkey: bytes[48], withdrawal_credentials: bytes[32], amount: bytes[8], signature: bytes[96])` and then manually compute `hash_tree_root` (which uses `sha2`).

hwwhww:
@JustinDrake ahh right that would be more reasonable. 😅

Note that, having the fields in API would limit our options to update eth2 fields without forking eth1 or creating a new deposit contract. Although I remember the last time we discuss the upgradability of the deposit contract, the solution would just be just moving to a new contract, right?

specs/core/0_beacon-chain.md
```diff
@@ -2230,7 +2230,7 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
 
     # Verify the Merkle branch
     merkle_branch_is_valid = verify_merkle_branch(
-        leaf=hash(serialize(deposit.data)),  # 48 + 32 + 8 + 96 = 184 bytes serialization
+        leaf=hash_tree_root(deposit.data),
```
vbuterin: Doesn't this mean that we also need to calculate the SSZ root of the deposit data in the Vyper contract?

JustinDrake: Yes, we need to calculate `hash_tree_root` in the contract. Assuming `def deposit(pubkey: bytes[48], withdrawal_credentials: bytes[32], amount: bytes[8], signature: bytes[96])` it's something like:

```python
pubkey_hash = hash(hash(pubkey[0:32]) + hash(pubkey[32:48] + zero_hash[0:16]))
signature_hash = hash(
    hash(hash(signature[0:32]) + hash(signature[32:64])) +
    hash(hash(signature[64:96]) + hash(zero_hash))
)
hash_tree_root = hash(
    hash(pubkey_hash + withdrawal_credentials) +
    hash(amount + zero_hash[0:24] + signature_hash)
)
```

JustinDrake:
> the solution would just be just moving to a new contract, right?

Right :)

tests/phase0/helpers.py
```diff
@@ -61,7 +61,7 @@ def create_mock_genesis_validator_deposits(num_validators, deposit_data_leaves=N
             amount=spec.MAX_DEPOSIT_AMOUNT,
             proof_of_possession=proof_of_possession,
         )
-        item = hash(deposit_data.serialize())
+        item = deposit_data.hash_tree_root()
```
vbuterin:
```suggestion
        item = hash_tree_root(deposit_data)
```

tests/phase0/helpers.py
```diff
@@ -180,7 +180,7 @@ def build_deposit(state,
                   amount):
     deposit_data = build_deposit_data(state, pubkey, privkey, amount)
 
-    item = hash(deposit_data.serialize())
+    item = deposit_data.hash_tree_root()
```
vbuterin:
```suggestion
    item = hash_tree_root(deposit_data)
```

hwwhww:
Opened https://github.com/ethereum/deposit_contract/issues/29.
I'll take it on the deposit contract side and then come back to clear and merge this PR.

hwwhww:
link the contract side: https://github.com/ethereum/deposit_contract/pull/31

hwwhww:
@djrtwo https://github.com/ethereum/deposit_contract/pull/31 is ready. :)

PR #934, opened 2019-04-16
Custody punishments & period staggering Take 2

CarlBeek:
This PR implements the following (and supersedes my previous attempt #872):

* Forced timely key reveals Validators who have not revealed by the end of their reveal period are slashed.
* Staggers reveal periods Reveal periods are deterministically staggered by taking the validator_index%EPOCHS_PER_CUSTODY_PERIOD.
* `max_reveal_lateness` is gradually reduced for timely responses

Hereby items 6 and 8 from #864 are addressed.


vbuterin:
Would it make sense to just merge in the proposal to make the custody key be the RANDAO 2 periods in the future (#880) into this PR?

So `CustodyKeyReveal` would be removed, it would just be `RandaoReveal`, and the masked reveal clause would have two sub-clauses, a partial penalty if the reveal is early by less than 2 custody periods, and a slashing if the reveal is early by more than 2 custody periods.

specs/core/1_custody-game.md
```diff
@@ -251,7 +274,7 @@ def verify_custody_key(state: BeaconState, reveal: CustodyKeyReveal) -> bool:
         signature=reveal.key,
         domain=get_domain(
             fork=state.fork,
-            epoch=reveal.period * EPOCHS_PER_CUSTODY_PERIOD,
+            epoch=reveal.period * EPOCHS_PER_CUSTODY_PERIOD + reveal.revealer_index % EPOCHS_PER_CUSTODY_PERIOD,
```
dankrad: I think this is a mistake (no need to change this line)

dankrad:
I would suggest closing this in favour of #880, where I made an attempt to merge all the changes suggested here

CarlBeek:
This has been incorporated into #880 and thus is no longer relevant.

PR #935, opened 2019-04-16
Add initial libp2p standardization

AgeManning:
Ahead of multi-client testnets, I think it will be useful to start a discussion on standardizing the libp2p protocols and their respective message formats. 

This is a minimal initial draft which outlines some of the protocols, their id's and suggested formats which could be used for eth2.0 clients. 

The protocols and formats listed here are currently in use by lighthouse but are entirely malleable and currently exist as suggestions for client implementations.

### DIFF
--- a/specs/networking/libp2p-standardization.md
+++ b/specs/networking/libp2p-standardization.md
@@ -77,7 +77,7 @@ optionally [yamux](https://github.com/hashicorp/yamux/blob/master/spec.md)
 study*
 
 The [Gossipsub](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub)
-protocol will be used for block and attestation propagation across the
+protocol is used for block and attestation propagation across the
 network.
 
 ### Configuration Parameters
@@ -111,7 +111,7 @@ propagation times and message duplication. Current network-related defaults are:
 *The Go and Js implementations use string topics - This is likely to be
 updated to topic hashes in later versions - https://github.com/libp2p/rust-libp2p/issues/473*
 
-For Eth2.0 clients, topics will be sent as `SHA2-256` hashes of the topic string.
+For Eth2.0 clients, topics are sent as `SHA2-256` hashes of the topic string.
 
 There are two main topics used to propagate attestations and beacon blocks to
 all nodes on the network.
@@ -120,12 +120,12 @@ all nodes on the network.
 	beacon blocks to all nodes on the networks.
 - The `beacon_attestation` topic - This topic is used to propagate
 	aggregated attestations to subscribing nodes (typically block proposers) to
-	be included into future blocks. Attestations will be aggregated in their
+	be included into future blocks. Attestations are aggregated in their
 	respective subnets before publishing on this topic.
 
-Shards will be grouped into their own subnets (defined by a shard topic). The
-number of shard subnets will be defined via `SHARD_SUBNET_COUNT` and the shard
-`shard_number % SHARD_SUBNET_COUNT` will be assigned to the topic:
+Shards are grouped into their own subnets (defined by a shard topic). The
+number of shard subnets is defined via `SHARD_SUBNET_COUNT` and the shard
+`shard_number % SHARD_SUBNET_COUNT` is assigned to the topic:
 `shard{shard_number % SHARD_SUBNET_COUNT}`.
 
 ### Messages
@@ -138,7 +138,7 @@ has a maximum size of 512KB (estimated from expected largest uncompressed block
 size).
 
 The `data` field of a Gossipsub `Message` is an SSZ-encoded object. For the `beacon_block` topic,
-this will be a `beacon_block`. For the `beacon_attestation` topic, this will be
+this is a `beacon_block`. For the `beacon_attestation` topic, this is
 an `attestation`.
 
 ## Eth-2 RPC
--- a/specs/networking/libp2p-standardization.md
+++ b/specs/networking/libp2p-standardization.md
@@ -9,6 +9,57 @@ configuration and messaging formats.
 
 # Libp2p Protocols
 
+## Transport
+
+This section details the libp2p transport layer that underlies the
+[protocols](#protocols) that are listed in this document.
+
+Libp2p allows composition of multiple transports. Eth2.0 clients should support
+TCP/IP and optionally websockets. Websockets are useful for implementations
+running in the browser and therefore native clients would ideally support these implementations 
+by supporting websockets.
+
+An ideal libp2p transport would therefore be TCP/IP with a fallback to
+websockets.
+
+### Encryption
+
+Libp2p currently offers [Secio](https://github.com/libp2p/specs/pull/106) which
+can upgrade a transport which will then encrypt all future communication. Secio
+generates a symmetric ephemeral key which peers use to encrypt their
+communication. It can support a range of ciphers and currently supports key
+derivation for elliptic curve-based public keys.
+
+Current defaults are:
+- Key agreement: `ECDH-P256` (also supports `ECDH-P384`)
+- Cipher: `AES-128` (also supports `AES-256`, `TwofishCTR`)
+- Digests: `SHA256` (also supports `SHA512`)
+
+
+## Protocols
+
+This section lists the necessary libp2p protocols required by Ethereum 2.0
+running a libp2p network stack.
+
+## Multistream-select
+
+#### Protocol id: `/multistream/1.0.0`
+
+Clients running libp2p should support the [multistream-select](https://github.com/multiformats/multistream-select/)
+protocol which allows clients to negotiate libp2p protocols establish streams
+per protocol.
+
+## Multiplexing
+
+Libp2p allows clients to compose multiple multiplexing methods. Clients should
+support [mplex](https://github.com/libp2p/specs/tree/master/mplex) and
+optionally [yamux](https://github.com/hashicorp/yamux/blob/master/spec.md)
+(these can be composed).
+
+**Mplex protocol id: `/mplex/6.7.0`**
+
+**Yamux protocol id: `/yamux/1.0.0`**
+
 ## Gossipsub
 
 #### Protocol id: `/meshsub/1.0.0`
@@ -70,11 +121,14 @@ number of shard subnets will be defined via `SHARD_SUBNET_COUNT` and the shard
 
 ### Messages
 
-Messages sent across gossipsub are fixed-size length-prefixed byte arrays.
-Each message has a maximum size of 512KB (estimated from expected largest uncompressed
-block size).
+#### Libp2p Specification
 
-The byte array is prefixed with a unsigned 64 bit length number encoded as an
+*This section simply outlines the data sent across the wire as specified by
+libp2p - this section is aimed at gossipsub implementers to standardize their implementation of this protocol*
+
+Libp2p raw gossipsub messages are sent across the wire as fixed-size length-prefixed byte arrays.
+
+The byte array is prefixed with an unsigned 64 bit length number encoded as an
 `unsigned varint` (https://github.com/multiformats/unsigned-varint). Gossipsub messages therefore take the form:
 ```
 +--------------------------+
@@ -86,7 +140,17 @@ The byte array is prefixed with a unsigned 64 bit length number encoded as an
 +--------------------------+
 ```
 
-The body of the message is an SSZ-encoded object. For the `beacon_block` topic,
+The body represents a protobuf-encoded [Message](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/rpc.proto#L17-L24). 
+
+In the following section we discuss the data being sent in the `data` field of
+the protobuf gossipsub `Message`.
+
+#### Eth2.0 Specifics
+
+Each message has a maximum size of 512KB (estimated from expected largest uncompressed
+block size).
+
+The `data` that is sent in a Gossipsub message is an SSZ-encoded object. For the `beacon_block` topic,
 this will be a `beacon_block`. For the `beacon_attestation` topic, this will be
 an `attestation`.
 
@@ -99,6 +163,9 @@ The [RPC Interface](./rpc-interface.md) is specified in this repository.
 ## Identify
 
 #### Protocol Id: `/ipfs/id/1.0.0` (to be updated to `/p2p/id/1.0.0`)
+
+*To be updated to incorporate discv5*
+
 
 The Identify protocol (defined in go - [identify-go](https://github.com/ipfs/go-ipfs/blob/master/core/commands/id.go) and rust [rust-identify](https://github.com/libp2p/rust-libp2p/blob/master/protocols/identify/src/lib.rs))
 allows a node A to query another node B which information B knows about A. This also includes the addresses B is listening on.
@@ -119,5 +186,7 @@ type of connecting node. Suggested format:
 ## Discovery
 
 #### Protocol Id: `/eth/serenity/disc/1.0.0`
+
+*To be updated to incorporate discv5*
 
 The discovery protocol to be determined.
--- a/specs/networking/libp2p-standardization.md
+++ b/specs/networking/libp2p-standardization.md
@@ -53,12 +53,20 @@ updated to topic hashes in later versions - https://github.com/libp2p/rust-libp2
 
 For Eth2.0 clients, topics will be sent as `SHA2-256` hashes of the topic string.
 
-There is one dedicated topic for propagating beacon blocks and aggregated
-attestations across the network. This topic will have the string
-`beacon_chain`. Each shard will have it's own topic allowing relevant parties
-to subscribe to in order to receive local shard attestations. The shard topics are
-prefixed with `shard` followed by the number of the shard. For example,
-messages relating to shard 10, will have the topic string `shard10`.
+There are two main topics used to propagate attestations and beacon blocks to
+all nodes on the network.
+
+- The `beacon_block` topic - This topic is used solely for propagating new
+	beacon blocks to all nodes on the networks.
+- The `beacon_attestation` topic - This topic is used to for propagate
+	aggregated attestations to subscribing nodes (typically block proposers) to
+	be included into future blocks. Attestations will be aggregated in their
+	respective subnets before publishing on this topic.
+
+Shards will be grouped into their own subnets (defined by a shard topic). The
+number of shard subnets will be defined via `SHARD_SUBNET_COUNT` and the shard
+`shard_number % SHARD_SUBNET_COUNT` will be assigned to the topic:
+`shard{shard_number % SHARD_SUBNET_COUNT}`.
 
 ### Messages
 
@@ -78,31 +86,15 @@ The byte array is prefixed with a unsigned 64 bit length number encoded as an
 +--------------------------+
 ```
 
-The body of the message is an SSZ-encoded object representing either a
-beacon block or attestation. The type of objected is determined via a prefixed
-nibble. Currently there are two objects that are sent across the gossip
-network. They are (with their corresponding nibble specification):
-
-- `0x1`: Beacon block
-- `0x2`: Attestation
-
-The body therefore takes the form:
-```
-+--------------------------+
-|       type nibble        |
-+--------------------------+
-|                          |
-|    SSZ-encoded object    |
-|                          |
-+--------------------------+
-```
+The body of the message is an SSZ-encoded object. For the `beacon_block` topic,
+this will be a `beacon_block`. For the `beacon_attestation` topic, this will be
+an `attestation`.
 
 ## Eth-2 RPC
 
 #### Protocol Id: `/eth/serenity/beacon/rpc/1`
 
 The [RPC Interface](./rpc-interface.md) is specified in this repository.
-
 
 ## Identify
 
--- a/specs/networking/libp2p-standardization.md
+++ b/specs/networking/libp2p-standardization.md
@@ -7,7 +7,7 @@ Ethereum 2.0 clients plan to use the libp2p protocol networking stack for
 mainnet release. This document aims to standardize the libp2p client protocols,
 configuration and messaging formats.
 
-# Libp2p Protocols
+# Libp2p Components 
 
 ## Transport
 
@@ -19,8 +19,11 @@ TCP/IP and optionally websockets. Websockets are useful for implementations
 running in the browser and therefore native clients would ideally support these implementations 
 by supporting websockets.
 
-An ideal libp2p transport would therefore be TCP/IP with a fallback to
-websockets.
+An ideal libp2p transport would therefore support both TCP/IP and websockets.
+
+*Note: There is active development in libp2p to facilitate the
+[QUIC](https://github.com/libp2p/go-libp2p-quic-transport) transport, which may
+be adopted in the future*
 
 ### Encryption
 
@@ -35,6 +38,11 @@ Current defaults are:
 - Cipher: `AES-128` (also supports `AES-256`, `TwofishCTR`)
 - Digests: `SHA256` (also supports `SHA512`)
 
+*Note: Secio is being deprecated in favour of [TLS
+1.3](https://github.com/libp2p/specs/blob/master/tls/tls.md). It is our
+intention to transition to use TLS 1.3 for encryption between nodes, rather
+than Secio.*
+
 
 ## Protocols
 
@@ -45,7 +53,8 @@ running a libp2p network stack.
 
 #### Protocol id: `/multistream/1.0.0`
 
-Clients running libp2p should support the [multistream-select](https://github.com/multiformats/multistream-select/)
+Clients running libp2p should support the
+[multistream-select](https://github.com/multiformats/multistream-select/)
 protocol which allows clients to negotiate libp2p protocols establish streams
 per protocol.
 
@@ -62,7 +71,7 @@ optionally [yamux](https://github.com/hashicorp/yamux/blob/master/spec.md)
 
 ## Gossipsub
 
-#### Protocol id: `/meshsub/1.0.0`
+#### Protocol id: `/eth/serenity/gossipsub/1.0.0`
 
 *Note: Parameters listed here are subject to a large-scale network feasibility
 study*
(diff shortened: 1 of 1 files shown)


JustinDrake:
Should we standardise on peer ids?

prestonvanloon:
At prysmatic labs, we are using multiple pubsub topics where each topic name is a 1:1 mapping to a message type. An example would be that attestation announcements come across the wire as `/eth/serenity/beacon/v1/attestationannounce`. Likewise with RPC calls, we have a stream for each RPC method/topic. 

Has there been any analysis of the trade-offs for this? Subscribing to only the topics you care about may reduce your network traffic while adding some potential complexity

prestonvanloon:
Having 1:1 topic mappings also reduces the need for a type nipple. The message is either the expected type or not.

AgeManning:
@JustinDrake - PeerId's seem to be specified here: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/networking/node-identification.md#peer-id-generation
So I have left them out of this particular doc. Perhaps we merge the two.

AgeManning:
@prestonvanloon - Agree that a potentially we want a topic for each message type (which removes the need for a nibble). In gossipsub for each topic, we have to maintain a set of mesh peers. So there may be some overhead in managing two sets of overlay networks for the blocks/attestations. I guess this is the point of this PR to agree on how we should proceed. I believe the WhiteBlock guys will be doing some testing for gossipsub in the near future and we will be able to have a more informed opinion on which way to proceed. 

In regards to having a separate protocol id for requests/responses in the RPC protocol seems like a useful separation as we will not need to encode the different types. However it seems to go against current libp2p protocol designs which typically have a singular protocol id per protocol. I'd opt to go whichever is the most performant.

JustinDrake:
> go-libp2p-crypto contains the canonical implementation of how to hash secp256k1 keys for use as a peer ID

Would it make sense to push for BLS12-381 keys in the longer term, to unify the consensus and networking layers?

prestonvanloon:
@AgeManning Agreed on the RPC part, it seems sub-optimal to open multiple streams between two peers.

cc: @zscole is this on your radar?

AgeManning:
@JustinDrake - Makes sense for consistency. The caveats might be that BLS will likely take longer to verify (compared to secp256k1) and there is talk of generating new key pairs for every message sent across the pubsub network (to anonymize validators) (I believe BLS generation is comparable to secp256k1 keys so this shouldn't be an issue).

zscole:
> cc: @zscole is this on your radar?

I don't see why multiple streams would be necessary for anything other than FTP. I don't think it would present any inherent optimization, but this is something we can test for to validate if really necessary.

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,131 @@
...
+updated to topic hashes in later versions - https://github.com/libp2p/rust-libp2p/issues/473*
+
+For Eth2.0 clients, topics will be sent as `SHA2-256` hashes of the topic string.
+
+There is one dedicated topic for propagating beacon blocks and aggregated
+attestations across the network. This topic will have the string
+`beacon_chain`. Each shard will have it's own topic allowing relevant parties
+to subscribe to in order to receive local shard attestations. The shard topics are
+prefixed with `shard` followed by the number of the shard. For example,
+messages relating to shard 10, will have the topic string `shard10`.
```
djrtwo: We might want to launch with fewer shard subnets than shards to promote stability in networks and reduce complexity. To do so, we would introduce a constant `SHARD_SUBNET_COUNT` and shards would be assigned to `shard{shard_number % SHARD_SUBNET_COUNT)`.

If we launch the beacon chain with no notion of "persistent committees", then this will very likely need to be a requirement, otherwise each shard subnet would get almost entirely dropped each epoch.

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,131 @@
...
+|                          |
++--------------------------+
+```
+
+The body of the message is an SSZ-encoded object representing either a
+beacon block or attestation. The type of objected is determined via a prefixed
+nibble. Currently there are two objects that are sent across the gossip
+network. They are (with their corresponding nibble specification):
+
+- `0x1`: Beacon block
```
djrtwo: Not necessarily suggesting it, but if we use a different topic for beacon blocks and attestations, we would not need the nibble, correct?

The major downside I see in fragmenting topics is that it gives nodes more optionality on what to listen for which could reduce the robustness/density of any given topic.

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,131 @@
...
+## Eth-2 RPC
+
+#### Protocol Id: `/eth/serenity/beacon/rpc/1`
+
+The [RPC Interface](./rpc-interface.md) is specified in this repository.
+
+
+## Identify
+
+#### Protocol Id: `/ipfs/id/1.0.0` (to be updated to `/p2p/id/1.0.0`)
```
djrtwo: Is `/p2p/` not yet supported? and why not use `/eth/` here? Is this something global beyond the eth protocol?

specs/networking/libp2p-standardization.md
AgeManning: Yep sure. I'll add this in.

specs/networking/libp2p-standardization.md
AgeManning: Yep, this is what @prestonvanloon was suggesting. I think multiple topics is a better approach assuming we only have two objects (and therefore topics).  I'll modify this.

specs/networking/libp2p-standardization.md
AgeManning: Yep so, the protocols that come with libp2p have their own protocol IDs. Currently (to my knowledge) these are hard coded and not customisable. I'd have to make a pr to libp2p or rewrite the protocol to change the naming. (For rust-libp2p). 
I assume the same for go in order for the protocols to be interoperable. 
I'm curious if we want to make the protocol IDs customisable within the libp2p protocols.

AgeManning:
@djrtwo - I'm not sure this PR is suited for the eth2 specs, I made it to allow comments easily. Perhaps it's suited in a client's repo or merged in with other networking docs.

mslipper:
Looks like I missed this - I'm going to close my PR (https://github.com/ethereum/eth2.0-specs/pull/975) in favor of this one. @AgeManning would you mind if I committed changes to this one in order to merge our efforts?

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,131 @@
...
+prefixed with `shard` followed by the number of the shard. For example,
+messages relating to shard 10, will have the topic string `shard10`.
+
+### Messages
+
+Messages sent across gossipsub are fixed-size length-prefixed byte arrays.
+Each message has a maximum size of 512KB (estimated from expected largest uncompressed
+block size).
+
+The byte array is prefixed with a unsigned 64 bit length number encoded as an
```
arnetheduck: since ssz is binary little endian uint64 (8 bytes), why not follow suit here? fewer "necessary" encodings in the app layer

specs/networking/libp2p-standardization.md
arnetheduck: +1

AgeManning:
> Looks like I missed this - I'm going to close my PR (#975) in favor of this one. @AgeManning would you mind if I committed changes to this one in order to merge our efforts?

Of course. I made this mainly to start the discussion. All edits/commits welcome. :)

specs/networking/libp2p-standardization.md
AgeManning: Yep, I also don't like the idea of having extra encodings. The reason this is here, is because this is currently how `rust-libp2p` sends raw messages over floodsub/gossipsub. I have explicitly written it out for other implementers not using libp2p. 
I assume this is the same for the other libp2p implementations. This particular encoding wasn't a design choice, rather highlighting what currently exists. 
This being said, I'm happy to modify libp2p to suit our needs.

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,123 @@
...
+#### Protocol Id: `/eth/serenity/beacon/rpc/1`
+
+The [RPC Interface](./rpc-interface.md) is specified in this repository.
+
+## Identify
+
+#### Protocol Id: `/ipfs/id/1.0.0` (to be updated to `/p2p/id/1.0.0`)
+
+The Identify protocol (defined in go - [identify-go](https://github.com/ipfs/go-ipfs/blob/master/core/commands/id.go) and rust [rust-identify](https://github.com/libp2p/rust-libp2p/blob/master/protocols/identify/src/lib.rs))
+allows a node A to query another node B which information B knows about A. This also includes the addresses B is listening on.
```
jannikluhn: What's the relationship to `discv5`? Is this supposed to be a temporary substitute, a permanent complement, or a replacement?

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,123 @@
...
+The protocol has two configurable parameters, which can be used to identify the
+type of connecting node. Suggested format:
+```
+	version: `/eth/serenity/1.0.0`
+	user_agent: <client name and version>
+```
+
+## Discovery
+
+#### Protocol Id: `/eth/serenity/disc/1.0.0`
```
jannikluhn: I think the current plan for discv5 is to use an independent transport.

specs/networking/libp2p-standardization.md
AgeManning: This exists in libp2p to prevent arbitrary sized byte streams. Once a message is received, libp2p checks the initial size. If the initial size is lager than the maximum, it drops the message. It then proceeds to read `min(max_size, size_of_message)` bytes. 
We could modify this to deal only SSZ encoding, but I believe it's changing the actual floodsub/gossipsub spec (assuming the spec is the implementations :) ).

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,123 @@
...
+	beacon blocks to all nodes on the networks.
+- The `beacon_attestation` topic - This topic is used to for propagate
+	aggregated attestations to subscribing nodes (typically block proposers) to
+	be included into future blocks. Attestations will be aggregated in their
+	respective subnets before publishing on this topic.
+
+Shards will be grouped into their own subnets (defined by a shard topic). The
+number of shard subnets will be defined via `SHARD_SUBNET_COUNT` and the shard
+`shard_number % SHARD_SUBNET_COUNT` will be assigned to the topic:
+`shard{shard_number % SHARD_SUBNET_COUNT}`.
```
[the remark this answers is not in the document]
AgeManning: I had imagined each shard subnet would be aggregating attestations. In later phases will the subnets be propagating other objects?
I agree for consistency this should be renamed as you suggest.

specs/networking/libp2p-standardization.md
AgeManning: Yep, temporary placeholder. This is currently what lighthouse is using in place of a proper discovery protocol.

specs/networking/libp2p-standardization.md
AgeManning: Yes, agree. This entire document will be updated with discv5 soon.

specs/networking/libp2p-standardization.md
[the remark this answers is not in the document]
djrtwo: there's actually a tradeoff here. In phase 1, we are assuming that persistent committees make up the shard subnets and provide a stable backbone to the network that the crosslink committees can rapidly connect to and utilize.

If we fragment the attestations from the rest of the shard subnet, then we run the risk of the crosslink committee not being able to find the shard attestation subnet _if_ the persistent committee chooses not to listen to this topic. The easy answer is the dictate that persistent committees join and relay all topics related to their persistent shard. Is this simple dictation of honest behavior here enough?

specs/networking/libp2p-standardization.md
djrtwo: Let's add an explicit note that this is a placeholder

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,123 @@
...
++--------------------------+
+|                          |
+|       body (<1M)         |
+|                          |
++--------------------------+
+```
+
+The body of the message is an SSZ-encoded object. For the `beacon_block` topic,
+this will be a `beacon_block`. For the `beacon_attestation` topic, this will be
+an `attestation`.
```
mhchia: Is the body here referring to the whole [`Message`](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/rpc.proto#L17-L24)? If so, IMO listing all the fields, or at least the `signature` and `data` will be more clear. And we can probably add that we are able to just put the serialized data to that field directly without any prefix, let the pubsub take care of the [de]serialization, and parse the data we get according to the message type corresponding to the topic.

specs/networking/libp2p-standardization.md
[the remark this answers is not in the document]
jannikluhn: I'm not very concerned about this. But if this turns out to be a problem (or  if we want to play it safe), we could add a rule that nodes should disconnect from nodes that don't connect to the attestation topic (or even stronger, that don't send N attestations per minute). But we should be careful not to put too much burden on light clients.

specs/networking/libp2p-standardization.md
AgeManning: Yep you are right. This should be more specific. The body here was referring to the `data` field of the gossipsub message. I assume all other fields are handled at the libp2p level. I can add more details to avoid confusion.

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,123 @@
...
+updated to topic hashes in later versions - https://github.com/libp2p/rust-libp2p/issues/473*
+
+For Eth2.0 clients, topics will be sent as `SHA2-256` hashes of the topic string.
+
+There are two main topics used to propagate attestations and beacon blocks to
+all nodes on the network.
+
+- The `beacon_block` topic - This topic is used solely for propagating new
+	beacon blocks to all nodes on the networks.
+- The `beacon_attestation` topic - This topic is used to for propagate
```
ralexstokes:
```suggestion
- The `beacon_attestation` topic - This topic is used to propagate
```

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,192 @@
...
+
+This section details the libp2p transport layer that underlies the
+[protocols](#protocols) that are listed in this document.
+
+Libp2p allows composition of multiple transports. Eth2.0 clients should support
+TCP/IP and optionally websockets. Websockets are useful for implementations
+running in the browser and therefore native clients would ideally support these implementations 
+by supporting websockets.
+
+An ideal libp2p transport would therefore be TCP/IP with a fallback to
```
raulk: In libp2p there's no concept of a "fallback" transport. All transports are equal, and each advertises a 1..n multiaddrs. It's up to the initiator to decide which transport they'll dial.

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,192 @@
...
+websockets.
+
+### Encryption
+
+Libp2p currently offers [Secio](https://github.com/libp2p/specs/pull/106) which
```
raulk: libp2p now has a [spec for TLS 1.3](https://github.com/libp2p/specs/tree/master/tls), with a [reference implementation in Go](https://github.com/libp2p/go-libp2p-tls). It's currently undergoing an [experiment in IPFS](https://github.com/ipfs/go-ipfs/pull/6229), and once we deem it stable, we'll recommend TLS 1.3 over SecIO.

It's OK to start with SecIO for now, but we'll recommend and switch to TLS 1.3 in IPFS and Filecoin as a default once it's stable.

We're also have a security firm auditing TLS 1.3, amongst other libp2p core components.

raulk: One aspect that worries us is that libp2p peers negotiate about which encryption method to use in cleartext during connection bootstrapping. Consequently we may be vulnerable to censorship via deep packet inspection. We aim to make our protocols indistinguishable from HTTP/3, hence the preference for TLS 1.3.

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,192 @@
...
+## Protocols
+
+This section lists the necessary libp2p protocols required by Ethereum 2.0
+running a libp2p network stack.
+
+## Multistream-select
+
+#### Protocol id: `/multistream/1.0.0`
+
+Clients running libp2p should support the [multistream-select](https://github.com/multiformats/multistream-select/)
```
raulk: I'll be writing a primer about connection bootstrapping and protocol negotiation soon; it's confusing.

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,192 @@
...
+### Messages
+
+#### Libp2p Specification
+
+*This section simply outlines the data sent across the wire as specified by
+libp2p - this section is aimed at gossipsub implementers to standardize their implementation of this protocol*
+
+Libp2p raw gossipsub messages are sent across the wire as fixed-size length-prefixed byte arrays.
+
+The byte array is prefixed with an unsigned 64 bit length number encoded as an
```
raulk: 64-bit varint?

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+The body represents a protobuf-encoded [Message](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/rpc.proto#L17-L24). 
+
+In the following section we discuss the data being sent in the `data` field of
+the protobuf gossipsub `Message`.
+
+#### Eth2.0 Specifics
+
+Each message has a maximum size of 512KB (estimated from expected largest uncompressed
```
raulk: It's worth specifying that this is Eth2.0's choice, as gossipsub does not impose a max message length, to my knowledge.

specs/networking/libp2p-standardization.md
raulk: You should be able to customise all of these protocol IDs. Doing so will segregate your network from the IPFS or libp2p public networks at the protocol level, such that even if peers happen to cross-connect, they will share no protocols in common, and therefore, won't be able to interact.

Ideally you'll override the protocol ID as low as in `multistream-select` (to be renamed to `multiselect`). This is used during connection bootstrapping, and if these IDs don't match, peers will disconnect immediately.

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,192 @@
...
+This protocol allows nodes to discover addresses of other nodes to be added to
+peer discovery. It further allows nodes to determine the capabilities of all it's connected
+peers.
+
+### Configuration Parameters
+
+The protocol has two configurable parameters, which can be used to identify the
+type of connecting node. Suggested format:
+```
+	version: `/eth/serenity/1.0.0`
```
raulk: 👍

specs/networking/libp2p-standardization.md
raulk: I thought there was a plan to implement discv5 as a libp2p protocol?

raulk:
On the topic (pun intended) of dedicated topics per message type and shard vs. wildcard topics per shard, I'd suggest the former. A few reasons off the top of my head:

* It curtails irrelevant gossip amplification. The effort invested by each peer to maintain the mesh overlay is proportional to the traffic it's interested in.
* Validation functions can be type-specific versus coarse-grained god-functions.
* Easier debugging and traceability.
* Allows lighter clients to subscribe only to the data they need.

specs/networking/libp2p-standardization.md
arnetheduck: so fwiw, my comment was primarily aimed at keeping things natural for each layer - if the length is part of "libp2p" and not the "eth2 payload", it should of course use whatever libp2p tends to use. echoing @jannikluhn 's sentiment here - should this be part of an eth2 document, or just noted as "assume the transport is a block transport / provides length"?

specs/networking/libp2p-standardization.md
mhchia: Speaking of the maximum size of a message, even though there isn't a specified constant, I played with the max size of [`DelimitedReader`](https://github.com/libp2p/go-libp2p-pubsub/blob/master/comm.go#L31), and now suspect it is the upper bound of the size in go implementation, IIUC.

specs/networking/libp2p-standardization.md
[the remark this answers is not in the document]
djrtwo: agreed 👍

specs/networking/libp2p-standardization.md
AgeManning: Sure. I was referring to this:
https://github.com/libp2p/rust-libp2p/blob/master/core/src/transport/choice.rs#L62-L76
Which, given a multiaddr tries to dial the initial transport, if that fails, then it dials the second transport. By fail, the Mutliaddr may not support the first transport. 
I agree this is not really a fallback, it's selecting the transport associated with the multiaddr. 
I will re-word this.

specs/networking/libp2p-standardization.md
AgeManning: Totally agree. I wasn't aware how far along TLS 1.3 was. I've currently laid out a few protocols/choices here, mainly as a template so others can discuss which is the best to proceed with. I agree TLS 1.3 is preferred over Secio. Will add this into the section. For initial multiclient testnets, perhaps we start with secio, then upgrade.

specs/networking/libp2p-standardization.md
AgeManning: Awesome. Let's link to it here :)

specs/networking/libp2p-standardization.md
AgeManning: Yep. This is based off the current floodsub implementation in rust-libp2p (figure all implementations should be standardized). Data is sent via this function:
https://github.com/libp2p/rust-libp2p/blob/master/core/src/upgrade/transfer.rs#L32-L41
Notice the the len of the data is using the `unsigned_varint` package with a u64_buffer():
https://github.com/libp2p/rust-libp2p/blob/master/core/src/upgrade/transfer.rs#L45 and
https://github.com/libp2p/rust-libp2p/blob/master/core/src/upgrade/transfer.rs#L46

specs/networking/libp2p-standardization.md
AgeManning: Floodsub in rust-libp2p has a maximum size of 2048 bytes:
https://github.com/libp2p/rust-libp2p/blob/master/protocols/floodsub/src/protocol.rs#L60
I've made this configurable in the gossipsub implementation. 
The max size exists so that if we see a large amount of data coming through a stream we only read a maximum amount. I understand this to prevent potential DOS vectors from malicious user's sending arbitrary large streams of data.

specs/networking/libp2p-standardization.md
AgeManning: Ok, to my knowledge, this is currently not the case in rust-libp2p. I'll make a PR to make these configurable.

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,192 @@
...
+
+# Libp2p Protocols
+
+## Transport
+
+This section details the libp2p transport layer that underlies the
+[protocols](#protocols) that are listed in this document.
+
+Libp2p allows composition of multiple transports. Eth2.0 clients should support
+TCP/IP and optionally websockets. Websockets are useful for implementations
```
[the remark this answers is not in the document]
AgeManning: Will do.

specs/networking/libp2p-standardization.md
AgeManning: I'm currently working on this. My current plan is to mimic mDNS in the integration into libp2p. In this case, there will be no protocol id, but will allow implementations to use the `discv5 discovery gadget` which listens separately on a UDP port. Messages from this gadget will be available to the user to pass into other libp2p protocols. i.e we discover a node, this discovery can be used to dial a tcp connection or add to a kademlia DHT in libp2p. 
Once, I've built an implementation, I will update this document.

specs/networking/libp2p-standardization.md
AgeManning: I'm working on discv5 at the moment. Currently, I think this protocol may still be useful for NAT traversal and can be fed into the discv5 protocol. Potentially we don't need it at all. I'll add a placeholder note for now.

specs/networking/libp2p-standardization.md
AgeManning: Yep. I added it here because I was concerned about client interop and wanted to specifically lay out the bytes across the wire. I agree that it doesn't belong in this document and clients can sort it out if their libp2p implementations don't match. Will remove this.

specs/networking/libp2p-standardization.md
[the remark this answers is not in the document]
djrtwo: @AgeManning I think this is the only unaddressed comment

> I had imagined each shard subnet would be aggregating attestations. In later phases will the subnets be propagating other objects?

Yes, the shard subnets will have blocks, attestations, and user level transactions. Splitting up the subnet right now just to `_attestation` makes sense to me.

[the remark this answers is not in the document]
AgeManning: Awesome. Thanks

jrhea:
> Having 1:1 topic mappings also reduces the need for a type nipple. The message is either the expected type or not.

True. We should make sure there are only two types nipples. ✌️😂🤠

specs/networking/libp2p-standardization.md
arnetheduck: it would be nice to have at least the ability to run discv5 over other transports - what would that look like, in this setup?

specs/networking/libp2p-standardization.md
```diff
@@ -0,0 +1,175 @@
...
+#### Protocol Id: `/eth/serenity/beacon/rpc/1`
+
+The [RPC Interface](./rpc-interface.md) is specified in this repository.
+
+## Identify
+
+**Note: This protocol is a placeholder and will be updated once the discv5
+discovery protocol is added to this document**
+
+#### Protocol Id: `/eth/serentiy/id/1.0.0`
```
wemeetagain: typo

AgeManning: Yep will fix.

AgeManning:
@djrtwo - I consider this still a discussion PR and it will be updated in a week or so with discv5. However it might be useful as a guideline to start libp2p co-ordination so happy to merge this WIP PR if you like. I can make future PR's to update it :).

AgeManning:
I've added a small paragraph for Discv5. 
We no longer need the `identify` protocol as Discv5 handles peer and topic discovery for us. It does not maintain it's own protocol-id as it has no need for establishing long-lasting streams. For these reason's I use the word `standalone` but it could be built from libp2p-components. It's standalone from a `multistream-select` perspective.

fjl:
This is really cool. Why should clients optionally support yamux?

djrtwo:
Merging.
We can take subsequent convos to issues/PRs

AgeManning:
@fjl - I originally made this as there was no discussion on what clients should use. By default, in rust, we can support both simultaneously. I've not done any testing or benchmarking between either protocol, but my feel is that mplex is minimal and yamux is more complex and potentially used more in prod (I have no real knowledge here however). I opted for the minimal by default and left the other libp2p-default multiplexer as optional. This was in part to spark discussion with people who know more about this than me (which didn't happen).

Issue #941, opened 2019-04-16
get_base_reward in GENESIS_EPOCH is 0, so all rewards are zero


dankrad:
`get_base_reward` uses `get_previous_total_balance`, but this returns 0 during the genesis epoch. So all rewards are zero in the `GENESIS_EPOCH`.

Why are we using `get_previous_total_balance` here and not `get_total_balance`? The latter seems to make more sense to me.

If there is a good reason why we want to use `get_previous_total_balance`, we can set         `validator.activation_epoch = GENESIS_EPOCH - 1` for validators activated at genesis?

djrtwo:
The key here is that rewards are given for activity during the previous epoch. Because we want to give all attestations an entire epoch to be included on chain, we have to wait until the end of the following epoch to reward attestations from the previous.

Some of this changes here -- https://github.com/ethereum/eth2.0-specs/pull/896

Rewards are given for activity in the previous epoch so we just skip rewards calculations during the genesis epoch. It is using previous_total_balance to reward based upon your balance during the epoch in which you produced the `attestation`.

dankrad:
So ignoring #896, not sure if it has been fixed there, I think this is actually a bug.

At the end of GENESIS_EPOCH, we do the per-epoch processing for the GENESIS_EPOCH. So at that time, 64 blocks and the corresponding attestations will have been processed, and should be rewarded. However, get_base_reward will be called, and will always return 0 (you can test this by putting an assert False in the return 0 case, all test cases with epoch transitions will fail). This is because at GENESIS_EPOCH-1, the total balance was actually 0.

JustinDrake:
> At the end of GENESIS_EPOCH, we do the per-epoch processing for the GENESIS_EPOCH.

The processing of `GENESIS_EPOCH` attestation rewards happens at the end `GENESIS_EPOCH + 1` (as opposed to `GENESIS_EPOCH`).

dankrad:
> > At the end of GENESIS_EPOCH, we do the per-epoch processing for the GENESIS_EPOCH.
> 
> The processing of `GENESIS_EPOCH` attestation rewards happens at the end `GENESIS_EPOCH + 1` (as opposed to `GENESIS_EPOCH`).

I see, but then why are we using the current balance for the validator, but the previous epoch balance for the total balance? I don't understand the logic behind this.

Also, I would suggest we add more context for logic like this. so that code like this can be understood easily without needing to refer to previous discussion (also goes with #942).

[...]

PR #949, opened 2019-04-17
Revamp balances and incentivisation

JustinDrake:
Substantive changes:

* Use capped (at 32 ETH) and rounded balance (with hysteresis) as the "effective balance" (stored in `validator.effective_balance`) throughout beacon chain consensus
     * Consistency with fork choice rule and light clients 🎉
* Only update `validator.effective_balance` once per epoch
     * Ability to compute the proposer for the current epoch 🎉
     * Cleaner modularity and parallel processing of rewards
     * Restore per-epoch (as opposed to per-slot) proposer reward processing
* Introduce `PendingAttestation.proposer_index` for attestation proposer rewards

Cosmetic changes (mostly related to `get_attestation_deltas`):

* Significantly improve readability and remove 82 lines 🎉
* Remove and merge helper functions where appropriate.
* Improve naming of helper functions
* Abstract away common logic and rework for readability
* Add `MIN_EPOCHS_TO_INACTIVITY_PENALTY` and `BASE_REWARDS_PER_EPOCH` constants
* Rescale `INACTIVITY_PENALTY_QUOTIENT`


specs/core/0_beacon-chain.md
```diff
@@ -185,6 +185,7 @@ These configurations are updated for releases, but may be out of sync during `de
 | `MAX_ATTESTATION_PARTICIPANTS` | `2**12` (= 4,096) |
 | `MIN_PER_EPOCH_CHURN_LIMIT` | `2**2` (= 4) |
 | `CHURN_LIMIT_QUOTIENT` | `2**16` (= 65,536) |
+| `BASE_REWARDS_PER_EPOCH` | `5` |
```
dankrad: I obviously like this :) I was wondering whether this is actually an exact figure though: I am confused by PROPOSER_REWARD_QUOTIENT, which seems to make this 1/8 in one of the cases.

specs/core/0_beacon-chain.md
```diff
@@ -1771,80 +1774,58 @@ def maybe_reset_eth1_period(state: BeaconState) -> None:
...
+            (validator.slashed and previous_epoch < validator.withdrawable_epoch)
         )
     ]
-    # Process rewards or penalties for all validators
+    rewards = [0 for index in range(len(state.validator_registry))]
+    penalties = [0 for index in range(len(state.validator_registry))]
     for index in eligible_validators:
-        base_reward = get_base_reward(state, index)
-        # Expected FFG source
+        base_reward = get_base_reward(state, get_previous_epoch_total_balance(state), index)
```
dankrad: So I know this has not changed, but is there an explicit reason to use get_previous_epoch_total_balance rather than the current balance here? This has a nasty edge case in the GENESIS_EPOCH that needs to be fixed:  #941

specs/core/0_beacon-chain.md
```diff
@@ -2176,7 +2158,7 @@ def process_proposer_attestation_rewards(state: BeaconState) -> None:
     for pending_attestations in (state.previous_epoch_attestations, state.current_epoch_attestations):
         for index in get_unslashed_attesting_indices(state, pending_attestations):
             if get_earliest_attestation(state, pending_attestations, index).inclusion_slot == state.slot:
-                base_reward = get_base_reward_from_total_balance(state, get_current_total_balance(state), index)
+                base_reward = get_base_reward(state, get_current_epoch_total_balance(state), index)
```
dankrad: Similar to the other comment above, why are we using the current balance rather than the previous one now?

specs/core/0_beacon-chain.md
JustinDrake: There's actually only four base rewards now that the inclusion delay bonus is an inclusion delay penalty:

1) Matching FFG source
2) Matching FFG target
3) Matching head
4) Matching crosslink data root

Yay for powers of two 🤪

specs/core/0_beacon-chain.md
JustinDrake: I think I agree with Danny. See [here](https://github.com/ethereum/eth2.0-specs/issues/941#issuecomment-483947322).

specs/core/0_beacon-chain.md
JustinDrake: Proposer attestation rewards (which does not fall in the `BASE_REWARDS_PER_EPOCH` because it only affects 64 proposers, and is scaled down by `PROPOSER_REWARD_QUOTIENT`) are handled on a block-by-block basis because the beacon chain can only compute `get_beacon_proposer_index` for the current slot.

specs/core/0_beacon-chain.md
dankrad: So should this be changed to 4 then?

How about block proposer rewards?

specs/core/0_beacon-chain.md
dankrad: So why then use the current validator balance, if using the previous total balance? (sorry for the parallel discussion)

specs/core/0_beacon-chain.md
```diff
@@ -1771,80 +1774,58 @@ def maybe_reset_eth1_period(state: BeaconState) -> None:
...
-            penalties[index] += get_inactivity_penalty(state, index, epochs_since_finality)
-        # Expected head
-        if index in get_unslashed_attesting_indices(state, matching_head_attestations):
-            rewards[index] += base_reward * matching_head_balance // total_balance
-        else:
-            penalties[index] += base_reward
-        # Take away max rewards if we're not finalizing
-        if epochs_since_finality > 4:
-            penalties[index] += base_reward * 4
+            penalties[index] += base_reward * (1 - MIN_ATTESTATION_INCLUSION_DELAY // inclusion_delay)
```
dankrad:
```suggestion
            penalties[index] += base_reward if inclusion_delay >= MIN_ATTESTATION_INCLUSION_DELAY else 0
```
I would prefer using logic instead of arithmetics for readability

JustinDrake: Oh that's not what I intended. We always have `inclusion >= MIN_ATTESTATION_INCLUSION_DELAY`.

I probably want `base_reward * (inclusion_delay - MIN_ATTESTATION_INCLUSION_DELAY) // inclusion_delay`.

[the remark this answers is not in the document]
JustinDrake: I think you also had this idea :) The intent (which I screwed up) was to scale down `base_reward` based on `inclusion_delay`.

specs/core/0_beacon-chain.md
```diff
@@ -1771,80 +1774,58 @@ def maybe_reset_eth1_period(state: BeaconState) -> None:
...
+
+        # Inactivity penalty
+        epochs_since_finality = previous_epoch + 1 - state.finalized_epoch
```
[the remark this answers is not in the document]
JustinDrake: Vitalik wrote "This seems weird; I think +0 is fine." when it was `current_epoch` (see [here](https://github.com/ethereum/eth2.0-specs/issues/882#issuecomment-480582463)). I guess the main thing is that the name of the constant (`MIN_EPOCHS_TO_LEAK` or whatever we end up choosing) matches the code.

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
JustinDrake: I care too much either way :)

specs/core/0_beacon-chain.md
```diff
@@ -1579,38 +1581,28 @@ The steps below happen when `state.slot > GENESIS_SLOT and (state.slot + 1) % SL
...
 ```
 
-```python
-def get_unslashed_attesting_indices(state: BeaconState, attestations: List[PendingAttestation]) -> List[ValidatorIndex]:
-    output = set()
-    for a in attestations:
-        output = output.union(get_attesting_indices(state, a.data, a.aggregation_bitfield))
-    return sorted(filter(lambda index: not state.validator_registry[index].slashed, list(output)))
-```
+Note: The balance computed by `get_previous_epoch_total_balance` may be different to the actual total balance during the previous epoch transition. Due to the bounds on per-epoch validator churn and per-epoch rewards/penalties, the maximum balance difference is low and only marginally affects consensus safety.
```
vbuterin: Maybe just call it `estimate_previous_epoch_total_balance`?

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
JustinDrake: Reverted to reward

specs/core/0_beacon-chain.md
JustinDrake: The helper functions now have an `epoch` parameter which can be `current_epoch` or `previous_epoch` and we get exact effective balance :)

dankrad:
@JustinDrake could you give a summary of how exactly the rewards/penalty change through this? I would like to read it with this info side by side :)

JustinDrake:
See the "substantive changes" section (which I fleshed out a bit) in the PR description.

specs/core/0_beacon-chain.md
```diff
@@ -1628,74 +1582,55 @@ def process_crosslinks(state: BeaconState) -> None:
...
+            get_matching_source_attestations(state, previous_epoch),
+            get_matching_target_attestations(state, previous_epoch),
+            get_matching_source_attestations(state, previous_epoch),
+        ):
+            if index in get_unslashed_attesting_indices(state, attestations):
+                rewards[index] += base_reward * get_attesting_balance(state, attestations) // total_balance
+            else:
+                penalties[index] += base_reward
+
+        if index in get_unslashed_attesting_indices(state, get_matching_source_attestations(state, previous_epoch)):
```
djrtwo: This section doesn't work. `get_beacon_proposer_index` can only used for slots in the current epoch so this section needs to be revamped.

You can instead select all attestations with an `inclusion_slot` from the current epoch and then see if it is the first to be included for the validator index from either the previous_epoch_attestations or current (depending on slot of the attestation), and reward accordingly.

test_libs/pyspec/tests/test_sanity.py
```diff
@@ -438,7 +436,7 @@ def test_balance_driven_status_transitions(state):
     assert pre_state.validator_registry[validator_index].exit_epoch == spec.FAR_FUTURE_EPOCH
 
     # set validator balance to below ejection threshold
-    set_balance(pre_state, validator_index, spec.EJECTION_BALANCE - 1)
+    pre_state.validator_registry[validator_index].effective_balance = spec.EJECTION_BALANCE - 1
```
djrtwo: Make this exactly `EJECTION_BALANCE` now that the equality condition is changed

```suggestion
    pre_state.validator_registry[validator_index].effective_balance = spec.EJECTION_BALANCE
```

specs/core/0_beacon-chain.md
```diff
@@ -1628,74 +1582,55 @@ def process_crosslinks(state: BeaconState) -> None:
...
+def get_attestation_deltas(state: BeaconState) -> Tuple[List[Gwei], List[Gwei]]:
+    previous_epoch = get_previous_epoch(state)
+    total_balance = get_total_active_balance(state)
+    eligible_validator_indices = [
         index for index, validator in enumerate(state.validator_registry)
-        if (
-            is_active_validator(validator, current_epoch) or
-            (validator.slashed and current_epoch < validator.withdrawable_epoch)
-        )
+        if is_active_validator(validator, previous_epoch) or (validator.slashed and previous_epoch < validator.withdrawable_epoch)
```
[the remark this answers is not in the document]
JustinDrake: I'm spoiled at SigmaPrime! (They have these huge wide screen.) I just went through the whole codebase looking for lines over 120 characters long and found 18 matches in code. I'll fix this particular issue and in another PR address the other 17 instances :)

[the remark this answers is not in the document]
JustinDrake: It turns out there was a bug here. We should use `current_epoch < validator.withdrawable_epoch` to preserve the invariants that withdrawable validators (in the current epoch) do not receive penalties or rewards.

specs/core/0_beacon-chain.md
```diff
@@ -1628,74 +1582,55 @@ def process_crosslinks(state: BeaconState) -> None:
...
-
-    adjusted_quotient = integer_squareroot(total_balance) // BASE_REWARD_QUOTIENT
-    return get_effective_balance(state, index) // adjusted_quotient // 5
-```
-
 ```python
-    return get_base_reward_from_total_balance(state, get_previous_total_balance(state), index)
-```
+    adjusted_quotient = integer_squareroot(get_total_active_balance(state)) // BASE_REWARD_QUOTIENT
```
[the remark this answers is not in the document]
JustinDrake: Will update PR note. Vitalik suggested we stick with current epoch balance for simplicity.

specs/core/0_beacon-chain.md
```diff
@@ -1569,27 +1521,28 @@ def process_justification_and_finalization(state: BeaconState) -> None:
...
     old_current_justified_epoch = state.current_justified_epoch
 
     # Process justifications
     state.previous_justified_epoch = state.current_justified_epoch
     state.previous_justified_root = state.current_justified_root
     state.justification_bitfield = (state.justification_bitfield << 1) % 2**64
-    previous_boundary_attesting_balance = get_attesting_balance(state, get_previous_epoch_boundary_attestations(state))
-    if previous_boundary_attesting_balance * 3 >= get_previous_total_balance(state) * 2:
+    previous_epoch_matching_target_balance = get_attesting_balance(state, get_matching_target_attestations(state, previous_epoch))
+    if previous_epoch_matching_target_balance * 3 >= get_total_active_balance(state) * 2:
```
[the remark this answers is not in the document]
JustinDrake: Vitalik suggested we stick with current epoch balance for simplicity. I tried it out, and it is indeed simpler. We *could* add `state.previous_effective_balances` (8 bytes per validator) but I guess the heuristic "simplicity trumps small gains in efficiency" trumps.

specs/core/0_beacon-chain.md
JustinDrake: I opted to add `proposer_index` to `PendingAttestation` similarly to `inclusion_slot`.

specs/core/0_beacon-chain.md
```diff
@@ -945,19 +917,20 @@ def generate_seed(state: BeaconState,
...
-def get_beacon_proposer_index(state: BeaconState) -> ValidatorIndex:
+def get_beacon_proposer_index(state: BeaconState, slot: Slot=None) -> ValidatorIndex:
     """
-    Return the beacon proposer index at ``state.slot``.
+    Return the beacon proposer index at ``slot``.
     """
-    current_epoch = get_current_epoch(state)
-
-    first_committee, _ = get_crosslink_committees_at_slot(state, state.slot)[0]
+    epoch = slot_to_epoch(slot if slot != None else state.slot)
```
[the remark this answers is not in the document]
JustinDrake: I now only get the proposer for the current slot :)

specs/core/0_beacon-chain.md
```diff
@@ -1628,74 +1582,55 @@ def process_crosslinks(state: BeaconState) -> None:
...
         base_reward = get_base_reward(state, index)
-        # Expected FFG source
-        if index in get_unslashed_attesting_indices(state, state.previous_epoch_attestations):
-            rewards[index] += base_reward * total_attesting_balance // total_balance
-            # Inclusion speed bonus
-            earliest_attestation = get_earliest_attestation(state, state.previous_epoch_attestations, index)
+
+        # Micro-incentives for attestations matching FFG source, FFG target, and head
+        for attestations in (
+            get_matching_source_attestations(state, previous_epoch),
```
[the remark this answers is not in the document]
JustinDrake: Significantly improved performance with caching 👍

specs/core/0_beacon-chain.md
```diff
@@ -237,8 +236,8 @@ These configurations are updated for releases, but may be out of sync during `de
...
 | `WHISTLEBLOWING_REWARD_QUOTIENT` | `2**9` (= 512) |
 | `PROPOSER_REWARD_QUOTIENT` | `2**3` (= 8) |
-| `INACTIVITY_PENALTY_QUOTIENT` | `2**24` (= 16,777,216) |
-| `MIN_PENALTY_QUOTIENT` | `2**5` (= 32) |
+| `INACTIVITY_PENALTY_QUOTIENT` | `2**25` (= 33,554,432) |
+| `MIN_SLASHING_PENALTY_QUOTIENT` | `2**5` (= 32) |
 
 * **The `BASE_REWARD_QUOTIENT` is NOT final. Once all other protocol details are finalized it will be adjusted, to target a theoretical maximum total issuance of `2**21` ETH per year if `2**27` ETH is validating (and therefore `2**20` per year if `2**25` ETH is validating, etc etc)**
 * The `INACTIVITY_PENALTY_QUOTIENT` equals `INVERSE_SQRT_E_DROP_TIME**2` where `INVERSE_SQRT_E_DROP_TIME := 2**12 epochs` (~18 days) is the time it takes the inactivity penalty to reduce the balance of non-participating [validators](#dfn-validator) to about `1/sqrt(e) ~= 60.6%`. Indeed, the balance retained by offline [validators](#dfn-validator) after `n` epochs is about `(1 - 1/INACTIVITY_PENALTY_QUOTIENT)**(n**2/2)` so after `INVERSE_SQRT_E_DROP_TIME` epochs it is roughly `(1 - 1/INACTIVITY_PENALTY_QUOTIENT)**(INACTIVITY_PENALTY_QUOTIENT/2) ~= 1/sqrt(e)`.
```
hwwhww: If `INACTIVITY_PENALTY_QUOTIENT` changes from `2**25` to `2**24`, then `INACTIVITY_PENALTY_QUOTIENT != INVERSE_SQRT_E_DROP_TIME**2` now. This sentence needs to be updated or we change `INACTIVITY_PENALTY_QUOTIENT` back to `2**24` with `INACTIVITY_PENALTY_QUOTIENT // 2` in formula.

specs/core/0_beacon-chain.md
```diff
@@ -1615,74 +1569,55 @@ def process_crosslinks(state: BeaconState) -> None:
...
+        rewards[earliest_attestation.proposer_index] += get_base_reward(state, index) // PROPOSER_REWARD_QUOTIENT
+        inclusion_delay = earliest_attestation.inclusion_slot - earliest_attestation.data.slot
+        rewards[index] += get_base_reward(state, index) * MIN_ATTESTATION_INCLUSION_DELAY // inclusion_delay
+
+    # Inactivity penalty
+    finality_delay = previous_epoch - state.finalized_epoch
+    if finality_delay > MIN_EPOCHS_TO_INACTIVITY_PENALTY:
+        matching_target_attesting_indices = get_unslashed_attesting_indices(state, matching_target_attestations)
+        for index in eligible_validator_indices:
+            penalties[index] += BASE_REWARDS_PER_EPOCH * get_base_reward(state, index)
```
hwwhww: Could you elaborate `BASE_REWARDS_PER_EPOCH:= 5`? Does that mean:
1. attesting expected source
2. attesting expected target
3. attesting expected head
4. attesting shard block (in `get_crosslink_deltas`)
5. attestation inclusion

JustinDrake: 1. expected source
2. expected target
3. expected head
4. inclusion delay reward
5. expected crosslink data root

specs/core/0_beacon-chain.md
```diff
@@ -1461,53 +1423,43 @@ The steps below happen when `state.slot > GENESIS_SLOT and (state.slot + 1) % SL
...
-    for a in attestations:
-        output = output.union(get_attesting_indices(state, a.data, a.aggregation_bitfield))
-    return sorted(filter(lambda index: not state.validator_registry[index].slashed, list(output)))
-```
-
-```python
-def get_attesting_balance(state: BeaconState, attestations: List[PendingAttestation]) -> Gwei:
-    return get_total_balance(state, get_unslashed_attesting_indices(state, attestations))
+def get_matching_source_attestations(state: BeaconState, epoch: Epoch) -> List[PendingAttestation]:
+    return state.current_epoch_attestations if epoch == get_current_epoch(state) else state.previous_epoch_attestations
```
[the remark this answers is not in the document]
JustinDrake: Added an `assert` :)

specs/core/0_beacon-chain.md
JustinDrake: I think this sentence should go away. We should have a separate document explaining design decisions/rationales. (This will be included in the Transparent Paper, whenever I get round to writing it.)

hwwhww: Yeah, it's also mentioned in @vbuterin's https://notes.ethereum.org/s/rkhCgQteN#Slashing-and-anti-correlation-penalties. We could (i) update rationales doc (ii) remove this sentence from the spec and link to rationale doc for now.

Issue #944, opened 2019-04-16
Figure 5 in get_base_reward


dankrad:
```python
def get_base_reward(state: BeaconState, index: ValidatorIndex) -> Gwei:
    if get_previous_total_balance(state) == 0:
        assert False
        return 0

    adjusted_quotient = integer_squareroot(get_previous_total_balance(state)) // BASE_REWARD_QUOTIENT
    return get_effective_balance(state, index) // adjusted_quotient // 5
```
The division by 5 in the last step makes me feel very uneasy. I suspect it is because there are up to five times per epoch that the base reward can be applied? [but then this is not quite right because of PROPOSER_REWARD_QUOTIENT?]
I would suggest making this explicit by introducing a constant. This would improve readability and also alert someone changing this in the future that these two are connected.

djrtwo:
Agreed on the constant. It is indeed because there are 5 different potential rewards. This allows us to tune the reward/penalty constants to target a specific return without considering how the reward is divided up.

`TOTAL_REWARD_QUOTIENT`, `TOTAL_POTENTIAL_REWARDS`, `NUM_REWARD_TYPES` or something

JustinDrake:
Closing in favour of #949 which introduces `BASE_REWARDS_PER_EPOCH` (see https://github.com/ethereum/eth2.0-specs/pull/949/files#diff-51a43328a58414e132a744f3771f018cR176)

PR #950, opened 2019-04-17
Factor out NUMBER_OF_BASE_REWARDS

dankrad:
This factors out the number 5 in get_base_reward (see #944)


JustinDrake:
Three minutes too late it seems :joy: See #949

PR #960, opened 2019-04-18
SSZ impl fixes

protolambda:
Cleaned up the `minimal_ssz.py` a bit, and fixed some bugs in the progress.
- vectors/lists where mixed up in serialization somewhere, incorrect serialization
- vector <-> list typing unclear at times, resulting in more funny serialization
- improved structure, generalized container encoding a bit using helper functions
- documented the magic

### DIFF
--- a/test_libs/pyspec/eth2spec/utils/minimal_ssz.py
+++ b/test_libs/pyspec/eth2spec/utils/minimal_ssz.py
@@ -1,6 +1,5 @@
 from .hash_function import hash
 
-
 BYTES_PER_CHUNK = 32
 BYTES_PER_LENGTH_PREFIX = 4
 ZERO_CHUNK = b'\x00' * BYTES_PER_CHUNK
@@ -62,16 +61,17 @@ def is_basic(typ):
     if not isinstance(typ, str):
         return False
     # "uintN": N-bit unsigned integer (where N in [8, 16, 32, 64, 128, 256])
-    if typ[:4] == 'uint' and typ[4:] in ['8', '16', '32', '64', '128', '256']:
+    elif typ[:4] == 'uint' and typ[4:] in ['8', '16', '32', '64', '128', '256']:
         return True
     # "bool": True or False
-    if typ == 'bool':
+    elif typ == 'bool':
         return True
     # alias: "byte" -> "uint8"
-    if typ == 'byte':
+    elif typ == 'byte':
         return True
     # default
-    return False
+    else:
+        return False
 
 
 def is_constant_sized(typ):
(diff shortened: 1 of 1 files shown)


protolambda:
Tested the ssz implementation with my other SSZ work. I have ~10K generated test-cases (random, zeroed, nil-length, max value, chaotic mix, etc.) generated from this, and ZRNT (after a few bugfixes) produces the same SSZ serialization/tree-roots :tada:

test_libs/pyspec/eth2spec/utils/minimal_ssz.py
```diff
@@ -58,18 +58,39 @@ def __len__(self):
...
+    if typ[:4] == 'uint' and typ[4:] in ['8', '16', '32', '64', '128', '256']:
+        return True
+    # "bool": True or False
+    if typ == 'bool':
+        return True
+    # alias: "byte" -> "uint8"
+    if typ == 'byte':
+        return True
+    # default
+    return False
```
hwwhww: IMO using `if-elif-else` would be more clear.

test_libs/pyspec/eth2spec/utils/minimal_ssz.py
```diff
@@ -151,42 +198,71 @@ def mix_in_length(root, length):
     return hash(root + length.to_bytes(32, 'little'))
 
 
+# Note: defaults to uint64 for integer type inference due to lack of information.
+# Other integer sizes are still supported, see spec.
```
hwwhww: This comment could be moved to docstring.

test_libs/pyspec/eth2spec/utils/minimal_ssz.py
```diff
@@ -151,42 +198,71 @@ def mix_in_length(root, length):
...
+            return [infer_type(value[0])]
+        else:
+            # Element type does not matter, list-content size will be encoded regardless, list is empty.
+            return ['uint64']
     elif isinstance(value, (bytes, str)):
         return 'bytes'
     elif isinstance(value, int):
         return 'uint64'
+    elif isinstance(value, int):
+        return 'uint64'
```
hwwhww: Redundant to the previous `elif isinstance(value, int): ...`

test_libs/pyspec/eth2spec/utils/minimal_ssz.py
```diff
@@ -151,42 +198,71 @@ def mix_in_length(root, length):
...
     else:
         raise Exception("Failed to infer type")
 
-
```
hwwhww: Let's keep 2 new lines as PEP8!

PR #962, opened 2019-04-18
SSZ static testing [blocked by #960]

protolambda:
See `/specs/test_formats/ssz_static/README.md` and `/specs/test_formats/ssz_generic/README.md` to learn about the two types of SSZ testing.

SSZ itself is very much the same. It's just that the clients build towards efficient SSZ deserialization/serialization for their phase-0 structs, and generic SSZ would not cover this code.

Also, the generic SSZ test-suite was build on by:
- https://github.com/ethereum/eth2.0-test-generators/pull/32
- https://github.com/ethereum/eth2.0-test-generators/pull/34

But these are based on the older SSZ (afaik). Production-first clients have a high need to just test spec-SSZ for their ETH-2-phase0 types.

Generic SSZ would require to:
- standardize a SSZ type format
- write a type parser in every client, and an encoder for the test-gen
- make clients instantiate these dynamic types. Something you wouldn't do in Go or Rust. Instead, a generic Container/Vector/List object would be used, but this makes things less efficient, and introduces more overhead in the SSZ encoder/decoder.
I do see generic-SSZ be a thing in the future, but for now we really just need testing for the ETH-2.0 types, that every client is looking to serialize/deserialize anyway.

--------

The introduced tests:

For every type pulled from the phase-0 spec (dynamic, generator does not need to be updated), create a bunch of random instances (we use the SSZ typing to fill it with random data), and output test vectors for it.
The randomization has different modes, and a "chaos" switch to randomly change the mode during filling of the SSZ instance with random data.
(Note: some modes are more like edge-cases, e.g. all zeroes, and will only output 1 case, unless in chaos-mode and things are mixed up)

The test-format itself is documented in the `ssz_static` test-format docs included in the PR.

It produces roughly ~10K test-cases, split in different suites. Most are for the minimal testing-configuration. And a few random ones are generated for the mainnet-configuration (larger objects, requires more space).

---------

Note: this PR is branched of the `ssz-impl-fixes` (PR #960), to make use of the new fixed spec-conformant SSZ implementation. Wait for that PR to merge.


test_libs/pyspec/eth2spec/debug/random_value.py
```diff
@@ -0,0 +1,137 @@
...
+def get_random_bytes_list(rng: Random, length: int) -> bytes:
+    return bytes(rng.getrandbits(8) for _ in range(length))
+
+
+def get_random_basic_value(rng: Random, typ: str) -> Any:
+    if typ == 'bool':
+        return rng.choice((True, False))
+    if typ[:4] == 'uint':
+        size = int(typ[4:])
+        assert size in (8, 16, 32, 64, 128, 256)
```
djrtwo:
```suggestion
        assert size in UINT_SIZES
```

test_libs/pyspec/eth2spec/debug/random_value.py
```diff
@@ -0,0 +1,137 @@
...
+    else:
+        raise ValueError("Not a basic type")
+
+
+def get_min_basic_value(typ: str) -> Any:
+    if typ == 'bool':
+        return False
+    if typ[:4] == 'uint':
+        size = int(typ[4:])
+        assert size in (8, 16, 32, 64, 128, 256)
```
djrtwo:
```suggestion
        assert size in UINT_SIZES
```

test_libs/pyspec/eth2spec/debug/random_value.py
```diff
@@ -0,0 +1,137 @@
...
+    else:
+        raise ValueError("Not a basic type")
+
+
+def get_max_basic_value(typ: str) -> Any:
+    if typ == 'bool':
+        return True
+    if typ[:4] == 'uint':
+        size = int(typ[4:])
+        assert size in (8, 16, 32, 64, 128, 256)
```
djrtwo:
```suggestion
        assert size in UINT_SIZES
```

test_libs/pyspec/eth2spec/utils/minimal_ssz.py
```diff
@@ -16,8 +16,8 @@ def __init__(self, **kwargs):
 
         def __eq__(self, other):
             return (
-                self.fields == other.fields and
-                self.serialize() == other.serialize()
+                    self.fields == other.fields and
```
djrtwo: Why 8 space indentation here?

protolambda:
@djrtwo Still relatively small compared to the Eth 1.0 tests repo. Just 67MB, fine with git LFS.
```
du -sh yaml_tests/ssz_static
67M     yaml_tests/ssz_static
```

protolambda:
Rebased on dev to solve merge conflicts (this PR was based of my previous SSZ-PR, which still had some changes / rebases before being submitted). Also applied suggestions from Danny :+1:

test_libs/pyspec/eth2spec/utils/minimal_ssz.py
protolambda: Double indentation was from my IDE here, not sure, some style thing. I changed it to one line so it's consistent regardless of formatter.

djrtwo:
lgtm

Issue #898, opened 2019-04-12
BLS Hash to G2


kirk-baird:
# Issues
The hash and test method used for mapping a message to G2 point is not incremented correctly.
```
def hash_to_G2(message_hash: Bytes32, domain: uint64)
```

# How it currently behaves
Currently the function hashes to an x value then increments this x-value until a valid curve point is found.

# How it should behave
In the case where the initial hash(m) does not generate a valid curve point:

The message is modify either by:
a) Incrementing the appended \x01 and \x02 bytes for each failed attempt
b) Modifying the message in some other incremental way

Hash this new set of bytes and check if it is a valid curve point.

Modify the message again and repeat this process until a valid point is discovered

# Additional Info
It may be worth waiting for clarification from the standardization process before deciding how to modify the original message.

JustinDrake:
Judging from today's BLS standardisation call, it sounds like the constant-time Fouque-Tibouchi method may be preferable to hash and increment.

djrtwo:
My gut is to leave this bug until we have the standard we all intend to conform on (probably Fouque-Tibouchi) to prevent everyone from opening up their bls implementations.

That said, my gut could be wrong as it is knowingly leaving insecure crypto sitting around. Any other opinions?

hwwhww:
Per @JustinDrake's comment  https://github.com/ethereum/eth2.0-specs/issues/1000#issuecomment-489433518

> BLS standardisation is unlikely to be finalised by June 30, although significant progress will have been made. Any inconsistency with the final BLS standard will have to be reflected in the launch release.

Do you think this bugfix should be delayed/removed from June 30 Frozen?

JustinDrake:
I agree with Danny, i.e. "leave this bug until we have the standard". The latest standardisation direction is to move to the new Wahby-Boneh hash function. https://eprint.iacr.org/2019/403

Removed from `Phase 0 Frozen` milestone, and suggest closing.

djrtwo:
I don't suggest closing a known bug issue unless it is either fixed or a warning is added to the spec itself

PR #964, opened 2019-04-19
BLS—separate cryptographic-layer from application-layer

JustinDrake:
Cryptographic layer (under standardisation):
   * Do not deal with application-layer domains (as per initial standardisation rough consensus)
   * Allow for arbitrary long messages to `hash_to_G2` (as per initial standardisation rough consensus)
   * Use `raw_bls_verify` and `raw_bls_verify_multiple` (as opposed to `bls_verify` and `bls_verify_multiple` for application layer)

Application layer:
   * Abstract away self-signed object logic (only 3 inputs to `bls_verify` instead of 4)
   * Clean up domains (use `Bytes4` instead of `int`)
   * Add `RandaoReveal` object

### DIFF
--- a/specs/bls_signature.md
+++ b/specs/bls_signature.md
@@ -18,6 +18,8 @@
         - [`bls_aggregate_pubkeys`](#bls_aggregate_pubkeys)
         - [`bls_aggregate_signatures`](#bls_aggregate_signatures)
     - [Signature verification](#signature-verification)
+        - [`raw_bls_verify`](#raw_bls_verify)
+        - [`raw_bls_verify_multiple`](#raw_bls_verify_multiple)
         - [`bls_verify`](#bls_verify)
         - [`bls_verify_multiple`](#bls_verify_multiple)
 
--- a/specs/bls_signature.md
+++ b/specs/bls_signature.md
@@ -67,12 +67,25 @@ We require:
 
 ### `hash_to_G2`
 
+Note: The following function is an (insecure!) placeholder that will be replaced according to the recommendations of the ongoing BLS12-381 standardisation effort.
+
 ```python
 G2_cofactor = 305502333931268344200999753193121504214466019254188142667664032982267604182971884026507427359259977847832272839041616661285803823378372096355777062779109
 q = 4002409555221667393417789825735904156556882819939007885332058136124031650490837864442687629129015664037894272559787
 
 def hash_to_G2(message: bytes) -> (Bytes96, Bytes96):
-    # TBD (pending standardisation)
+    # Initial candidate x coordinate
+    x_re = int.from_bytes(hash(message + b'\x01'), 'big')
+    x_im = int.from_bytes(hash(message + b'\x02'), 'big')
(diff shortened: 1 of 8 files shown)


specs/bls_signature.md
```diff
@@ -69,19 +71,8 @@ We require:
...
-    
-    # Test candidate y coordinates until a one is found
-    while 1:
-        y_coordinate_squared = x_coordinate ** 3 + Fq2([4, 4])  # The curve is y^2 = x^3 + 4(i + 1)
-        y_coordinate = modular_squareroot(y_coordinate_squared)
-        if y_coordinate is not None:  # Check if quadratic residue found
-            return multiply_in_G2((x_coordinate, y_coordinate), G2_cofactor)
-        x_coordinate += Fq2([1, 0])  # Add 1 and try again
+def hash_to_G2(message: bytes) -> (Bytes96, Bytes96):
+    # TBD (pending standardisation)
```
vbuterin: I'd say don't remove it; keep it as is, as people need to have *something* to implement for the moment. But definitely put a "THIS WILL CHANGE" marker on it.

specs/bls_signature.md
```diff
@@ -65,14 +67,16 @@ We require:
 
 ### `hash_to_G2`
 
+Note: The following function is an (insecure!) placeholder that will be replaced according to the recommendations of the ongoing BLS12-381 standardisation effort.
```
ChihChengLiang: Note for the note: the hash-and-check function is insecure since it is easy to waste verifiers' or signers' time by accidentally or adversarially choosing a difficult to hash message. Related #898, ia.cr/2019/403 .

djrtwo:
@JustinDrake We can close this, right? I think we ended up cleaning up the applciation layer interface a bit in another PR and taht we didn't want to make changes to the bls spec until we have the standardization process done.

JustinDrake:
I'm intending to cherry pick things that don't change the BLS interface too much. There's a lot of BLS cleanup that's accumulated over and would be good to clear some of the backlog before the freeze.

Also @protolambda wants to clear up a few BLS things:

Screenshot 2019-06-28 at 19 24 03

protolambda:
Going to make a copy of this PR, rebase it on dev branch, and fine tune it. The contents as-is are quite outdated with typing. But otherwise support changes.

bytes8 seems like the better choice:
- given the bytes4 inputs (version, domain type), and then later addition to the raw message.
- cross platform, no unnecessary uint64 problems
- no complications with endianness, just a well defined order of 8 bytes.

protolambda:
We can freeze the spec as-is, and update the typing in the BLS tests to be consistent with the spec (we have a spec-test issue for it: https://github.com/ethereum/eth2.0-spec-tests/issues/8).
That would be some time after freeze (these are tests, non breaking, already stable for a long time, not crucial).

I created PR #1233 to discuss a new cleaned up rebased version of this PR. I'm removing the freeze-label here and closing it, as this PR is not going to make it as-is, nor is the change well formed yet, especially based on old types and assumptions (we still had the BLS stub, time flies). It can be non-breaking (regarding consensus), and introduced post-freeze.

PR #966, opened 2019-04-19
Added signing_root to ssz_static tests

zilm13:
This PR adds `signing_root` to the test cases with `hash_tree_root` if object contains `signature` field. I think it's good enough for initial validation of clients logic without adding extra complexity with generated structures.
cc: @protolambda


zilm13:
@protolambda merged your fixes and did little cleanup of References in `ssz_static/core.md`. I think it's done now

PR #967, opened 2019-04-20
update CI config: caching of repo, venv, split test install/run

protolambda:
- repo itself is cached now, for faster checkout
- venv for test running is cached now
- venv is installed first now, and tests run separately. Also makes local runs faster if you don't need to re-install dependencies (even if they are cached)
- `make citest` does the same as `make test`, but outputs a test-report for CI
  - Note: the test reporting can be improved, we would want one separate report for each sub-topic. But the tests need to be more structured for that first. This is at least an improvement on status-quo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,89 +1,97 @@
 version: 2.1
+commands:
+  restore_cached_venv:
+    description: "Restores a cached venv"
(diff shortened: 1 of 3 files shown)


protolambda:
@djrtwo LGTM?

Issue #833, opened 2019-03-22
Issuance rate parameterization under adversarial conditions


adlerjohn:
Thanks @Mikerah for initial discussions on this.

#251 provided some initial discussion on issuance _rate_, but does not discuss how to avoid the [two extremes presented in the design rationale](https://notes.ethereum.org/s/rkhCgQteN#Base-rewards), namely:

> two common models, (i) fixed reward rate, and (ii) fixed total reward

Each can be solved individually, but let's consider the first case for now:

>  The main argument against (i) is that it imposes too much uncertainty on the network of two kinds: uncertainty of the total level of issuance, and uncertainty of the total level of deposits (as if a fixed reward rate is set too low then almost no one will participate, threatening the network, and if a rate is set too high then very many validators will participate, leading to unexpectedly high issuance).

The solution to the above two extreme cases is a mid-way compromise of inverse-square root issuance based on the total balance. The design rationale states that "The inverse-square-root approach compromises between the two and avoids the worst consequences of each one," but I argue that it does not, continuing discussion from #832 [in this comment](https://github.com/ethereum/eth2.0-specs/issues/832#issuecomment-475628818). In the worst case, an inv-sqrt scaling on rewards _does not solve the problem of having set the parameter too low_. To actually solve it would require an inverse-exponential function, where the rewards increase to infinite with no validators.

A compromise solves neither problem, and instead leaves the worst case scenario for both intact. How can the rewards be scaled in a way that we have rigorous proofs of how the system performs in adversarial conditions?

Is there actually a real, provable problem with using a linear scaling reward function, other than the optics that inflation is too high?

vbuterin:
> A compromise solves neither problem, and instead leaves the worst case scenario for both intact.

Huh? That's definitely not true. First of all, `y = 1/sqrt(x)` _is_ unbounded with `y` approaching infinity as `x` approaches zero, so it is the case that if the parameter set is too low it will still reach an equilibrium with a significant number of validators. Second, as my [paper on discouragement attacks](https://github.com/ethereum/research/blob/master/papers/discouragement/discouragement.pdf) shows, using a square root formula instead of a simple `y = 1/x` _does_ make discouragement attacks 2x weaker, to the point where it does possibly solve the issue (see "this implies we need `p < 1/2` for epsilon attacks to be unprofitable" on page 6).

adlerjohn:
[...]

PR #971, opened 2019-04-20
Signal non-final status of base reward and desired issuance goal

vbuterin:
An issuance increase is proposed based on community feedback, to `2**21` ETH if `2**27` ETH is validating, along with an agreement to set the base reward quotient based on a pre-set max issuance bound once all protocol details are finalized.

Here's a table of new total issuance rates with this proposal:

| ETH validating | Max annual issuance | Max annual return rate |
| - | - | - |
| 1,000,000 | 181,019 | 18.10% |
| 3,000,000 | 313,534 | 10.45% |
| 10,000,000 | 572,433 | 5.72% |
| 30,000,000 | 991,483 | 3.30% |
|100,000,000 | 1,810,193 | 1.81% |
| 134,217,728 | 2,097,152 | 1.56% |

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -241,7 +241,7 @@ These configurations are updated for releases, but may be out of sync during `de
 | `INACTIVITY_PENALTY_QUOTIENT` | `2**24` (= 16,777,216) |
 | `MIN_PENALTY_QUOTIENT` | `2**5` (= 32) |
 
-* The `BASE_REWARD_QUOTIENT` parameter dictates the per-epoch reward. It corresponds to ~2.54% annual interest assuming 10 million participating ETH in every epoch.
+* **The `BASE_REWARD_QUOTIENT` is NOT final. Once all other protocol details are finalized it will be adjusted, to target a theoretical maximum total issuance of `2**21` ETH per year if `2**27` ETH is validating (and therefore `2**20` per year if `2**25` ETH is validating, etc etc)**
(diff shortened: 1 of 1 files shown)


JustinDrake:
Below's my rationalisation as to why the numbers are reasonable.

Targeting 2**25 ETH at stake (~32m ETH) for the long term feels about right for strong security. In such conditions, the base inflation would be ~1% and the base return ~%3.2%. Assuming each shard consumes on average 1,000 ETH in gas per year, with half of the gas burnt, then inflation would be ~0.5% and the validator return ~5%. Feels healthy :) If we're get significant less than 2**25 ETH at stake doubling the base inflation wouldn't be unreasonable :)

dankrad:
These numbers seem more reasonable to me than the numbers currently in the spec!

> If we get significant less than 2^25 ETH at stake then doubling the base inflation wouldn't be unreasonable :)

Wait, the way I see it at the moment, base inflation would actually be lower with less money staked, right? Rewards per staked money would be increased though.

adamluc:
@JustinDrake - why are a portion of fees burned?

JustinDrake:
@adamluc Eth2 will likely have a new gas mechanism which involves burning some fraction of the gas fees. See [EIP 1559](https://github.com/ethereum/EIPs/issues/1559) for one possible way forward. https://medium.com/@eric.conner/fixing-the-ethereum-fee-market-eip-1559-9109f1c1814b

adlerjohn:
> Feels healthy!

That...doesn't seem like a particularly rigorous metric. Neither does "this other blockchain has X issuance rate, and they haven't gotten attacked yet so X must be fine!" A chain is secure until it isn't. I attempted to solicit discussion around the metrics for issuance rate parametrization in #833, but it looks like it fell on deaf ears. Is the issuance rate really being influenced by "community feedback"? Why not just outsource the task to an independent panel of actual expert economists?

collinjmyers:
Happy hear a proposal has been put in place to get the conversation started - for those participating in staking (especially the Genesis stakers) it is also crucial to have more clarity around network fees and how they are divided up between pow & pos participants - I will provide a new validator profitability analysis and update my original articles based on the new proposal - however without a better undertaking of network fee dynamics at the 10M mark and at genesis/phase 1/phase 2 the analysis will remain incomplete....ish.

ehariton:
> 
> 
> > Feels healthy!
> 
> That...doesn't seem like a particularly rigorous metric. Neither does "this other blockchain has X issuance rate, and they haven't gotten attacked yet so X must be fine!" A chain is secure until it isn't. I attempted to solicit discussion around the metrics for issuance rate parametrization in #833, but it looks like it fell on deaf ears. Is the issuance rate really being influenced by "community feedback"? Why not just outsource the task to an independent panel of actual expert economists?

Strongly agree with @adlerjohn here. Choosing a fixed percentage based on a feeling is scary. Plus a fixed percentage is always going to supply too much security, or not enough. Too much means we're over paying to keep the system secure. Too little is also bad. Can we find an optimal solution space that provides just enough security?

EdsonAlcala:
sorry my ignorance but how do you calculate the Max annual issuance and Max annual return rate? @vbuterin

Issue #871, opened 2019-04-01
Network message SSZ schema


jannikluhn:
(Continuing discussion from the PR)

Two nitpicks regarding the network message serialization format:

1. If we use SSZ, we need a single base schema that embeds the payload message for network messages as otherwise receivers don't know how to decode it. However right now, the spec defines three message types:

```
Request: (
    id: uint64
    method_id: uint16
    body: Request
)

Reponse: (
    id: uint64
    response_code: uint16
    result: bytes
)

ErrorResponse: (
    id: uint64
    response_code: uint16
    result: bytes
)
```

`Response` and `ErrorResponse` are identical, but `Request` differs a little. It would also be nice to give them the same name and field names, because that's what implementations will use when they first decode a message (e.g. `Message` with `message_id`, `request_id`, and `body`). `body` would be a `bytes` that carries the SSZ serialized request/response/error. This adds an additional 4-byte prefix, but I think that's unavoidable.

2. The spec says

```
The result member is OPTIONAL on errors, and MAY contain additional information about the error.
```

There aren't optional fields in SSZ right now, so it should just be a blank string if there's no information.

@mslipper

mslipper:
Thanks @jannikluhn! These are points well taken. During the networking workshop tomorrow I intend on working through these with the implementation teams; it's becoming clear that SSZ isn't a good choice for network data.

jannikluhn:
Personally I totally agree with this, but many people seem to like using the same format for everything. If we go with something else, I'd suggest RLP: It works quite well for Eth1, there are tons of implementations, it doesn't require a schema (at least not for the first decoding step), and messages can be extended.

Sadly, I'm not going to be able to participate in the workshop, but I'm looking forward to hear about the results.

dankrad:
So I just overheard the discussion from the workshop yesterday, and if the most important thing that is missing from SSZ right now is an Option type and a Null object, I think it may be possible to add this with 10ish lines of code. I think having the same serialization format in the core and networking would be nice. However I am wondering what the future implications with regards to upgrade (#861) are?

arnetheduck:
We have an option type - the above messages could be defined as:
```
Request: (
    id: uint64
    method_id: uint16
    body: Request
)

Reponse: (
    id: uint64
    responseData: [Payload]
    error: [ErrorResponse]
)
```

arnetheduck:
>  it doesn't require a schema

[...]

PR #974, opened 2019-04-21
Use SSZ union-types in RPC request/response

mslipper:
This PR addresses #871 by using SSZ union types in request body

### DIFF
--- a/specs/networking/rpc-interface.md
+++ b/specs/networking/rpc-interface.md
@@ -39,32 +39,36 @@ To facilitate RPC-over-`libp2p`, a single protocol name is used: `/eth/serenity/
 Remote method calls are wrapped in a "request" structure:
 
 ```
-(
-    id: uint64
-    method_id: uint16
-    body: Request
-)
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                                                               |
++                          id (uint64)                          +
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|       method_id (uint16)      |       body_len (uint32)       |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                               |     body (body_len bytes)     |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ```
 
 and their corresponding responses are wrapped in a "response" structure:
 
 ```
-(
-    id: uint64
-    response_code: uint16
-    result: bytes
-)
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                                                               |
++                          id (uint64)                          +
+|                                                               |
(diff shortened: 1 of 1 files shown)


JustinDrake:
> Since the `body` and `result` fields depend on the value of `method_id` and `response_code`, a schema for the above structure cannot be known beforehand.

Doesn't SSZ unions (over the possible `method_id` and `response_code`) allow for this use case? It would be nice to use SSZ everywhere.

specs/networking/rpc-interface.md
```diff
@@ -39,32 +39,36 @@ To facilitate RPC-over-`libp2p`, a single protocol name is used: `/eth/serenity/
...
-    result: bytes
-)
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                                                               |
++                          id (uint64)                          +
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|     response_code (uint16)    |      result_len (uint32)      |
```
arnetheduck: can we keep these aligned a bit? ie life gets easier when types are aligned to their size, which in this case can be had by changing `response_code` to uint32.. alternatively, stick another `uint16` in there for future expansion set to 0 for now

djrtwo:
If we use specific stream for each RPC, we avoid having to do this correct?
Is there any downside to this approach?

arnetheduck:
SSZ unions can be done with:
```
{
  some_a: [AType]
  some_b: [BType]
}
```
if you know all the type possibilities up-front and don't mind that lists in theory can have more than one item. in practice, this tends not to be a big deal (see for example the protobuf encoding where the last value with matching key is used for single-item (optional) fields, and lists simply collect them instead - the only difference is a cosmetic one, in the schema.

in terms of forward/backward compatibility, it's a "heavy" solution - every change, even purely additive, needs a new "protocol" or "request".

specs/networking/rpc-interface.md
mslipper: Sure thing.

mslipper:
@djrtwo using separate streams allows us to avoid some of this. You still won't be able to parse responses from multiple RPC requests at once without some identifier linking a response to a request. It _does_ become much simpler, however.

jannikluhn:
> and simple byte sequences for everything else

We could also call everything else message header and encode it in SSZ. This wouldn't change anything on the wire (as long as we only use fixed size values in the header), but maybe it's a little bit simpler conceptually.

djrtwo:
Would a union type (https://github.com/ethereum/eth2.0-specs/pull/893) preclude us from having to define this external, non-ssz wrapper?

djrtwo:
@mslipper  See above question

mslipper:
My bad, missed that notification - a union type technically solves this, but introduces other problems. The union type would need to be over a list of all possible RPC request/responses, which IMO rapidly reduces maintainability.

Beyond the desire for consistency, is there another reason for why we're leaning towards SSZ everywhere?

JustinDrake:
> The union type would need to be over a list of all possible RPC request/responses

The list of all possible RPC request/responses should be readily available, right? For example, every request type needs to have a corresponding request handler. As such, maintenance of the union can be abstracted away programmatically.

mslipper:
@JustinDrake doesn't this seem needlessly complex to you? We're talking about creating tooling to work around SSZ's limitations in a specification that hasn't been implemented yet. Is there a specific reason for requiring SSZ everywhere? It seems to me that decoding a sequence of bytes on the wire _then_ decoding the SSZ payload is far simpler than maintaining a union type programmatically.

JustinDrake:
> We're talking about creating tooling to work around SSZ's limitations

I'm suggesting reusing the existing query/response infrastructure that needs to exist anyway. (In other words, no extra tooling needs to be created.)

> Is there a specific reason for requiring SSZ everywhere?

The reason is the long tail of benefits we get from standardisation :)

As a side note, judging from internal conversations, I think the design team (Danny, Vitalik, myself) are inclined to stick with SSZ. Having said that, I don't particularly care too much either way 😂

dankrad:
> doesn't this seem needlessly complex to you? We're talking about creating tooling to work around SSZ's limitations in a specification that hasn't been implemented yet.

Could you describe the tooling that would be needed? I also still don't get what the problem is here, as long as there is a finite enum of possible request types that's going to be used, it should be fine? It is even possible to maintain forwards compatibility by just adding new types to the union and make old clients ignore requests with unkown type numbers.

PR #975, opened 2019-04-21
Add spec for data propagation

mslipper:
This PR introduces the topics used to propagate data through the ETH 2.0 network. I only included beacon blocks and attestations for now. I've also not made any prescriptions around the specific pubsub implementation we use.

### DIFF
--- a/specs/networking/block-propagation.md
+++ b/specs/networking/block-propagation.md
@@ -32,7 +32,6 @@ We expect that over time `SHARD_SUBNET_COUNT` will be increased to equal the tot
 
 Messages are serialized using SSZ unless otherwise noted.
 
-## 
 
 ## Propagated Data
 
--- a/specs/networking/block-propagation.md
+++ b/specs/networking/block-propagation.md
@@ -23,7 +23,7 @@ There are a variety of network-based pubsub implementations available. `libp2p` 
 
 Topics involving single shards are mapped to "subnets." A shard's subnet is calculating by calculating `shard_number % SHARD_SUBNET_COUNT`. Grouping shards into subnets as described confers the following benefits:
 
-- Shards using smaller amounts network traffic are grouped with shards with more network traffic, thus increasing the stability of the network by reducing the likelihood that less-used shareds will be eclipsed.
+- Shards using smaller amounts network traffic are grouped with shards with more network traffic, thus increasing the stability of the network by reducing the likelihood that less-used shards will be eclipsed.
 - The existence of the `SHARD_SUBNET_COUNT` creates a quality-of-service parameter that we can tweak to ensure the health over the network.
 
 We expect that over time `SHARD_SUBNET_COUNT` will be increased to equal the total number of shards.
--- a/specs/networking/block-propagation.md
+++ b/specs/networking/block-propagation.md
@@ -1,0 +1,59 @@
+ETH 2.0 Networking Spec - Data Propagation
+===
+
+# Abstract
+
+This specification describes how data will be propagated ("gossiped") on the network among Ethereum 2.0 nodes.
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL”, NOT", “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in RFC 2119.
+
+# Dependencies
+
+This specification assumes familiarity with the [RPC Interface](./rpc-interface.md) and [Beacon Chain](../core/0_beacon-chain.md) specifications.
+
+# Specification
+
+## Pubsub
+
+Ethereum 2.0 uses a publish/subscribe ("pubsub") construction to propagate information throughout the network. In pubsub applications, nodes tag the messages they send with a topic. Only the nodes interested in that topic will process the message. This allows nodes to decide for themselves which information they are interested in.
+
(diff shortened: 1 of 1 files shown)


specs/networking/block-propagation.md
```diff
@@ -0,0 +1,59 @@
...
+- Shards using smaller amounts network traffic are grouped with shards with more network traffic, thus increasing the stability of the network by reducing the likelihood that less-used shareds will be eclipsed.
+- The existence of the `SHARD_SUBNET_COUNT` creates a quality-of-service parameter that we can tweak to ensure the health over the network.
+
+We expect that over time `SHARD_SUBNET_COUNT` will be increased to equal the total number of shards.
+
+## Encoding
+
+Messages are serialized using SSZ unless otherwise noted.
+
+##
```
hwwhww:
```suggestion
```

specs/networking/block-propagation.md
```diff
@@ -0,0 +1,59 @@
...
+(
+	block_root: HashTreeRoot
+)
+```
+
+Notifies interested nodes that a new beacon block has been created. Nodes are expected to use one of the RPC APIs (Method ID `12`, for example) to fill out the block header and block body.
+
+### Attestations
+
+**Topic:** `attestations/${subnet}`
```
hwwhww: What do you think about having `${subnet}/` as the prefix?

specs/networking/block-propagation.md
```diff
@@ -0,0 +1,58 @@
...
+
+There are a variety of network-based pubsub implementations available. `libp2p` includes three: `floodsub`, `gossipsub`, and most recently `EpiSub`. `gossipsub` is currently the most likely implementation to be adopted as the official standard, however this may change as a result of performance testing. Given these conditions, **this specification makes no assumptions about the underlying pubsub implementation.**
+
+## Shard Subnets
+
+Topics involving single shards are mapped to "subnets." A shard's subnet is calculating by calculating `shard_number % SHARD_SUBNET_COUNT`. Grouping shards into subnets as described confers the following benefits:
+
+- Shards using smaller amounts network traffic are grouped with shards with more network traffic, thus increasing the stability of the network by reducing the likelihood that less-used shareds will be eclipsed.
```
djrtwo:
```suggestion
- Shards using smaller amounts network traffic are grouped with shards with more network traffic, thus increasing the stability of the network by reducing the likelihood that less-used shards will be eclipsed.
```

specs/networking/block-propagation.md
```diff
@@ -0,0 +1,58 @@
...
+
+**Body:**
+
+```
+(
+	block_root: HashTreeRoot
+)
+```
+
+Notifies interested nodes that a new beacon block has been created. Nodes are expected to use one of the RPC APIs (Method ID `12`, for example) to fill out the block header and block body.
```
djrtwo: Gossiping of full blocks is the current expected behavior rather than multiple rounds of comms via RPC

mslipper: Roger, will update.

djrtwo:
safe to close @mslipper ?

specs/networking/block-propagation.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Pubsub
+
+Ethereum 2.0 uses a publish/subscribe ("pubsub") construction to propagate information throughout the network. In pubsub applications, nodes tag the messages they send with a topic. Only the nodes interested in that topic will process the message. This allows nodes to decide for themselves which information they are interested in.
+
+There are a variety of network-based pubsub implementations available. `libp2p` includes three: `floodsub`, `gossipsub`, and most recently `EpiSub`. `gossipsub` is currently the most likely implementation to be adopted as the official standard, however this may change as a result of performance testing. Given these conditions, **this specification makes no assumptions about the underlying pubsub implementation.**
+
+## Shard Subnets
+
+Topics involving single shards are mapped to "subnets." A shard's subnet is calculating by calculating `shard_number % SHARD_SUBNET_COUNT`. Grouping shards into subnets as described confers the following benefits:
```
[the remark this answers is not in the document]
mhchia: > Grouping shards into subnets as described confers the following benefits

Does the shard grouping here mean, shards with the same `shard_id % SHARD_SUBNET_COUNT` are "grouped" together?

specs/networking/block-propagation.md
```diff
@@ -0,0 +1,58 @@
...
+
+- Shards using smaller amounts network traffic are grouped with shards with more network traffic, thus increasing the stability of the network by reducing the likelihood that less-used shards will be eclipsed.
```
mhchia: > Shards using smaller amounts network traffic are grouped with shards with more network traffic

Sorry, could you elaborate more this? Does it assume smaller amounts networks are always grouped with more the ones with more traffic?

specs/networking/block-propagation.md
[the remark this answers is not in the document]
ralexstokes: yeah, just a way to batch shards into "subnets" in the case that there would be too few nodes in a subnet if we do the natural thing -- one subnet per shard.

so you can imagine there are `SHARD_COUNT` shards, where each has a unique number from `0..SHARD_COUNT` called the `shard_id`. here we introduce a `subnet_id` -- if you are curious about a particular shard `S` you can compute `S % SHARD_SUBNET_COUNT` to get the `subnet_id`

specs/networking/block-propagation.md
ralexstokes: yeah, how will this work given that `%` will produce a uniform distribution b/t namespaces?

djrtwo:
safe to close @mslipper ?

mslipper:
Safe to close.

PR #979, opened 2019-04-22
pytests use configuration system

protolambda:
- split `requirements.txt` in two: one for normal use (whatever that may be), one for running tests (has the normal `requirements.txt` as `-r` line to ensure you don't miss the dependencies)
- add command option to conftest to switch
- update makefile

Also: I fixed a minor testing bug that was uncovered because of a small constant change: minimal config has a shorter eth1-data voting period. The new winning vote set the eth1-deposit count to 0, and it resulted in a negative expected-deposit-count. Default test-blocks are prepared with a vote for whatever deposit count the state says it is at.
*Maybe we should add a voting requirement to the spec itself, so that you can't make the eth1-data deposit counter lower than the deposit index in the state.*

Note: I don't believe anyone has actively been maintaining the test-correctness for non-`minimal_config`. Currently, when running tests against the mainnet config (slower), some of them fail. I identified a few root causes: `assert verify_attestation(...)` fails for some reason, a `committee` for some shard was not found, and committee size <-> bitfield size is off.


hwwhww:
I made another PR on it: https://github.com/ethereum/eth2.0-specs/pull/981 so that we can organize all dependencies in `setup.py`. 😄

protolambda:
rebased on dev, PR should run cleanly now. Hope this resolves the strange local-execution issues @hwwhww was experiencing.

hwwhww:
It works now! Straaange.

djrtwo:
> I don't believe anyone has actively been maintaining the test-correctness for non-minimal_config

These failed tests had to do with the low number of validators (100) compared to the slots_per_epoch. When generating attestations, I am often assuming there is at least 2 validators per committee. I changed `num_validators` to scale with `slots_per_epoch` and tests are fixed. but now the mainnet tests take a _very_ long time to run.

PR #984, opened 2019-04-23
Update committee helper comments

terenc3t:


specs/core/0_beacon-chain.md
```diff
@@ -814,7 +814,7 @@ def get_epoch_committee_count(state: BeaconState, epoch: Epoch) -> int:
 ```python
 def get_shard_delta(state: BeaconState, epoch: Epoch) -> int:
     """
-    Return the minimum number of shards that get processed at ``epoch``.
+    Return the number of shards to increment ``state.shard_shard`` during ``epoch``.
```
hwwhww: @djrtwo typo? 😄

PR #989, opened 2019-04-23
Finality testing

djrtwo:
Add tests to explicitly test each of the 4 finality rules.
These are pretty heavy wrt timing but very important tests.

Note, built off of #988 so merge in #988 before review


test_libs/pyspec/tests/test_finality.py
```diff
@@ -0,0 +1,193 @@
...
+        blocks += new_blocks
+
+        if epoch == 0:
+            check_finality(test_state, prev_state, False, False, False)
+        elif epoch == 1:
+            check_finality(test_state, prev_state, False, False, False)
+        elif epoch == 2:
+            check_finality(test_state, prev_state, True, False, False)
+        elif epoch >= 3:
+            # rule 4 of finaliy
```
hwwhww:
```suggestion
            # rule 4 of finality
```

test_libs/pyspec/tests/test_finality.py
```diff
@@ -0,0 +1,193 @@
...
+            fill_aggregate_attestation(post_state, prev_attestation)
+            block.body.attestations.append(prev_attestation)
+
+        state_transition(post_state, block)
+        blocks.append(block)
+
+    return state, blocks, post_state
+
+
+def test_finality_from_genesis_rule_4(state):
```
hwwhww: I understand what you mean by "from_genesis", but since there's no other rule 4 test case, it seems okay to move it to the last and rename it to `def test_finality_rule_4(state)`.

PR #992, opened 2019-04-24
off by one error for eth1 data voting

djrtwo:
The logic for checking if it was the end of an eth1 data voting period was off by one. Fixed this off by one error and added a test


djrtwo:
I think I'm going to use my executive merge power and get this merged for v0.6 release.

PR #996, opened 2019-04-25
`_deltas` functions should return `Tuple` instead of `List`

hwwhww:


JustinDrake:
Can we use a tool to find these automatically?

djrtwo:
You can run mypy/a linter on the output pyspec file but probably not on the markdown

Issue #927, opened 2019-04-15
Test-generator: operations from blocks


protolambda:
Since the generators are now (waiting for #851 merge) integrated into the PySpec and hosted in this repository, we can push for much more completeness in YAML testing.

This issue is to track the progress of providing test-cases for block-operations ("transactions" previously).
Suggestions for test-cases are welcome. And better, open up a PR with implementation (once #851 is ready)

- [x] Deposits
  - [x] Valid new deposit
  - [x] Valid top-up
  - [x] Invalid: deposit index
  - [x] Invalid: deposit proof
  - [x] Valid new deposit but invalid signature (proof of possession)
- [x] Proposer slashings
  - [x] Valid proposer slashing
  - [x] Invalid proposer index
  - [x] Invalid: different epoch
  - [x] Invalid: Same block headers
  - [x] Invalid: Non-slashable proposer (not activated) [TODO: correct, maybe just slash them anyway, 
since they signed it?]
  - [x] Invalid: Non-slashable proposer (slashed)
  - [x] Invalid: Non-slashable proposer (withdrawn)
  - [x] Invalid: Invalid signatures (either header, or both)
- [x] Attester slashings
  - [x] Invalid: Same attestation
  - [x] Valid: double vote
  - [x] Valid: surround vote
  - [x] Invalid: no double or surround vote
  - [x] Invalid: either attestation has invalid indexed form (custody bit indices wrong)
  - [x] Invalid: no slashable indices
  - [x] Invalid: invalid signatures (either attestation, or both)
- [x] Attestations (inclusion)
  - [x] Valid attestation
  - [x] Valid previous epoch
  - [x] Invalid: shard out of range
  - [x] Invalid: incorrect source epoch (previous and current epoch)
  - [x] Invalid: incorrect source root (previous and current epoch)
  - [x] Invalid: non-zero crosslink root
  - [x] Invalid: incorrect previous crosslink root with state (previous)
  - [x] Invalid: source root is target root
  - [x] Invalid: attestation references non-zero custody bitfield
  - [x] Invalid: current and previous root switched
  - [x] Invalid: mismatch in participants/custody bits
  - [x] Invalid: invalid signature
- [x] Voluntary Exits
  - [x] Valid exit
  - [x] Exit queue
  - [x] Invalid: out of range validator index
  - [x] Invalid: not active (activation phase)
  - [x] Invalid: already exited
  - [x] Invalid: voluntary exit is in the future
[...]

PR #999, opened 2019-04-25
State tests for Attestations

zilm13:
I have tried to implement all cases or something similar for Attestations from #927 
Also this PR includes some changes in everything I've noticed mistyped or incorrect on the way, related to attestations. If it's better to detach something to separate PR, just ask me.
`full_suite` with mainnet config is made with reduced set of cases, included only one heavy for example, because most of them need a lot of time to complete, however I've tried all cases and they finished after half an hour or so.


protolambda:
> Also this PR includes some changes in everything I've noticed mistyped or incorrect on the way, related to attestations. If it's better to detach something to separate PR, just ask me.

I would prefer to have a separate PR for unrelated changes (like the validator doc update). I.e. copy the branch (checkout as new), `git rebase -i HEAD~n`, change `pick -> drop` for unrelated changes, and push the unrelated commits to a separate branch.

> full_suite with mainnet config is made with reduced set of cases, included only one heavy for example, because most of them need a lot of time to complete, however I've tried all cases and they finished after half an hour or so.

Honestly, if individual test suites take longer than a minute to run, they should probably not be included. In that case it is more productive to ensure that the tests based on the minimal config cover the same edge cases.

At first glance the attestation test-generation code looks good. However, I we may need to change things a bit, as part of an effort to reduce total testing code in the specs repository (sorry :disappointed:).

@djrtwo and I are looking into getting some of the test case "setup" work in test-generators and py-tests unified. To reduce duplicate effort. As you may know, we are testing the spec itself too, and the edge-cases being tested are mostly the same. See https://github.com/ethereum/eth2.0-specs/tree/dev/test_libs/pyspec/tests/block_processing
This is not to say it is 100% the same, but the "setup" part of the tests (edge case scenario being created) is. You can find an earlier proposal of mine to generalize it here: https://notes.ethereum.org/5N7MrvOESrKK3AIHYYvYww?view Since then I updated it with some newer ideas, but did not have the chance to fully implement it.
And we're still exploring the idea, there may be better ways to generalize it. Suggestions welcome :). I hope to implement something stable next week. I'm putting this PR on hold for now.

zilm13:
@protolambda ok, waiting for further instruction about this PR when merging of pyspec tests and state tests is over. Checked your proposal: everything looks good, but anyway there could be cases in the future which will not be able to fit in this decorator. However if it could be used for most of the tests, it's great already.

As for speed of generator - I could comment this line https://github.com/ethereum/eth2.0-specs/pull/999/files#diff-559e6b9bef3fe6d1a20b7945f38b1c00R335 to be sure, that generator runs under 1 minute.

protolambda:
Status update: we've unified the pytests; we can generate test-vectors from these tests now. So the format in this PR is not quite right anymore, as it would be more useful to make these tests part of the unified tests. I'll port over the code from here that presents new edge case coverage :+1:

protolambda:
New tests included in #1052, closing this.

PR #1009, opened 2019-04-29
Attestation committee refactor

vbuterin:
Mostly cosmetic and enabling some functionality needed in phase 1. A few substantive changes to `Attestation` data structure.

* Remove `get_crosslink_committees_at_slot` (that function's ugly man...)
* Make the "base" that everything works off instead be `get_crosslink_committee` (simpler and more generic)
* Attestations store epoch, start shard and shard, no longer slot (slot can be calculated from the other three)
* store `inclusion_distance` in `PendingAttestation` rather than `inclusion_slot` for more direct use in subsequent calculations.
* Retaining start shard in attestations allows `get_attesting_indices` to peek much further back into the past, making it useful for slashings (Phase 1)
* Some two-layer-deep nested loops become one-layer-deep loops
* alter `get_winning_crosslink_and_attesting_indices` function signature to mirror that of `get_crosslink_committee`
* `is_double_vote` and `is_surround_vote` replaced with single `is_slashable_attestation_data`


specs/core/0_beacon-chain.md
```diff
@@ -805,44 +808,30 @@ def compute_committee(validator_indices: List[ValidatorIndex],
...
+        return state.latest_start_shard
+    elif epoch == get_previous_epoch(state):
+        previous_shard_delta = get_shard_delta(state, epoch)
+        return (state.latest_start_shard - previous_shard_delta) % SHARD_COUNT
+    elif epoch == get_current_epoch(state) + 1:
+        current_shard_delta = get_shard_delta(state, get_current_epoch(state))
+        return (state.latest_start_shard + current_shard_delta) % SHARD_COUNT
+    else:
+        raise Exception("Not supported")
+```
```
```suggestion
def get_epoch_start_shard(state: BeaconState, epoch: Epoch) -> Shard:
    assert epoch <= get_current_epoch(state) + 1
    check_epoch = get_current_epoch(state) + 1
    shard = (state.latest_start_shard + get_shard_delta(state, get_current_epoch(state))) % SHARD_COUNT
    while check_epoch > epoch:
        check_epoch -= 1
        shard = (shard - get_shard_delta(state, check_epoch)) % SHARD_COUNT
    return shard
```

djrtwo: Might make this `(shard + SHARD_COUNT - get_shard_delta(state, check_epoch))` just to be true to the unsigned type

specs/core/0_beacon-chain.md
```diff
@@ -1511,18 +1427,19 @@ def get_attestation_deltas(state: BeaconState) -> Tuple[List[Gwei], List[Gwei]]:
...
-            for index in crosslink_committee:
-                base_reward = get_base_reward(state, index)
-                if index in attesting_indices:
-                    rewards[index] += base_reward * attesting_balance // committee_balance
-                else:
-                    penalties[index] += base_reward
+    epoch = get_previous_epoch(state)
+    for offset in range(get_epoch_committee_count(state, epoch)):
+        shard = (get_epoch_start_shard(epoch) + offset) % SHARD_COUNT
+        crosslink_committee =  get_crosslink_committee(state, epoch, offset)
```
djrtwo: `get_crosslink_committee` expects a shard rather tahn an `offset`

```suggestion
        crosslink_committee = get_crosslink_committee(state, epoch, shard)
```

specs/core/0_beacon-chain.md
```diff
@@ -956,6 +868,49 @@ def verify_merkle_branch(leaf: Bytes32, proof: List[Bytes32], depth: int, index:
...
+        bit = (byte >> (position % 8)) % 2
+        index = flip if bit else index
+
+    return index
+```
+
+### `get_crosslink_committee`
+
+```python
+def get_crosslink_committee(state: BeaconState, epoch: Epoch, shard: Shard) -> List[ValidatorIndex]:
```
djrtwo: Might want to add an assertion to `get_crosslink_committee` that the shard is in the expected range. A committee cannot be retrieved for all shards in each epoch

specs/core/0_beacon-chain.md
```diff
@@ -1091,39 +1042,10 @@ def verify_indexed_attestation(state: BeaconState, indexed_attestation: IndexedA
...
         signature=indexed_attestation.signature,
-        domain=get_domain(state, DOMAIN_ATTESTATION, indexed_attestation.data.epoch),
+        domain=get_domain(state, DOMAIN_ATTESTATION, indexed_attestation.data.target_epoch),
     )
 ```
 
-### `is_double_vote`
-
-```python
-def is_double_vote(attestation_data_1: AttestationData,
```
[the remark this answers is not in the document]
djrtwo: Specifically, they are used in the validator guide -- https://github.com/ethereum/eth2.0-specs/blob/dev/specs/validator/0_beacon-chain-validator.md#attester-slashing

djrtwo:
I'm working on some edits based on the above feedback

djrtwo:
and now working on some bugs :)

vbuterin:
I still am not fully comfortable with `compute_committee` being removed. It's definitely the right choice if you look at phase 0 in a vacuum, but the reason I separated out that function initially was that `get_persistent_committee` in phase 1 also uses it, so we're making both kinds of committees from the same basic building block.

Here is the usage in phase 1: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/1_shard-data-chains.md#get_period_committee

One possible compromise I see would be to define a function:

```python
def get_indices_of_chunk(list_size: int, chunks: int, chunk: int) -> List[int]:
    start = (len(list_size) * chunk) // chunks
    end = (len(list_size) * (chunk + 1)) // chunks
    return list(range(start, end))
```

Then the PR would be, in `get_crosslink_committee`:

```python
    return [
        active_indices[get_shuffled_index(i, len(active_indices), seed)]
        for i in get_indices_of_chunk(len(active_indices), committee_count, committee_index)
    ]
```

And the same in `get_period_committee`.

JustinDrake:
@vbuterin: I've exposed `get_committee` which abstracts away common logic for both `get_crosslink_committee` (now a 2-liner) and `get_period_committee` (now a 1-liner).

djrtwo:
I think the previous function signature of `compute_committee` is more straight forward and readable and also is much clearer to the client what the explicit inputs are which could be used to cache results.

I vote going back to that function signature.

vbuterin:
> I think the previous function signature of compute_committee is more straight forward and readable and also is much clearer to the client what the explicit inputs are which could be used to cache results.

Agreed.

JustinDrake:
`compute_committee` restored! :)

PR #1010, opened 2019-04-30
Proposer slashing tests

jannikluhn:
Add tests for propose slashing as described in #927.

Follows pretty much the structure of the already existing deposit tests.

What's missing are tests with invalid header signatures, and the existing ones aren't properly signed. I think particular for this and in general for all test generators it would be nice to have some additional helper functions that make it easier to create such objects (signed headers, state with validator i exited and j slashed, etc.)

### DIFF
--- a/test_generators/operations/genesis.py
+++ b/test_generators/operations/genesis.py
@@ -17,16 +17,21 @@ def create_genesis_state(deposits: List[spec.Deposit]) -> spec.BeaconState:
     )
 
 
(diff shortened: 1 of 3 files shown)


protolambda:
Similarly with test generation presented in #999 (comment [here](https://github.com/ethereum/eth2.0-specs/pull/999#issuecomment-491516123)):
> Status update: we've unified the pytests; we can generate test-vectors from these tests now. So the format in this PR is not quite right anymore, as it would be more useful to make these tests part of the unified tests. I'll port over the code from here that presents new edge case coverage :+1:

protolambda:
New tests included in #1052, closing this.

PR #1013, opened 2019-04-30
Generate seed once for get_beacon_proposer_index

terencechain:
Does it make sense to move `generate_seed` outside of the while loop so we'd only need to compute once?


vbuterin:
Reminder that the spec is optimized for readability, not execution speed :)

Though I can see how this could improve readability too.

PR #1014, opened 2019-05-01
Add 32-byte graffiti

JustinDrake:
Add 32-byte of arbitrary "graffiti" data in beacon blocks, in a similar vein to `extraData` in Eth1 headers. To be used in wonderful and unpredictable ways (permissionless innovation by block proposers).


pepoospina:
You need two bytes32 for storing a [cid](https://github.com/multiformats/cid). It would be nice to have the option to store ipfs hashes here.

JustinDrake:
@pepoospina: We're considering increasing the size of the graffiti significantly, possibly to something like 2kB.

PR #1015, opened 2019-05-01
Remove get_state_root from state transition doc

JustinDrake:
Remove `get_state_root` from the state transition function spec because it is not used by the state transition function. (Unused code affects spec testability and conciseness. Also note that `get_state_root` is slot-based which is inconsistent with its `get_block_root` counterpart.)

The `get_state_root` helper *is* used by the test suite and is now defined there. If useful to validators beyond the test suite, it may make sense to define `get_state_root` in the validator doc.


djrtwo:
Generally agree except with the "Also note that get_state_root is slot-based which is inconsistent with its get_block_root counterpart" argument. We store state roots per slot in state so makes sense to give slot granular access :)

Issue #946, opened 2019-04-16
Clarify "per-epoch processing"


dankrad:
When reading through the spec, I noticed that during epoch processing, the "current epoch" actually refers to the previous epoch, whereas in the block processing, the "current slot" corresponds to the slot of the given block.

To make this less confusing, I suggest renaming it to "end of epoch processing"?

JustinDrake:
We have

```python
def get_current_epoch(state: BeaconState) -> Epoch:
    return state.slot // SLOTS_PER_EPOCH
```

and (see [state_transition.py](https://github.com/ethereum/eth2.0-specs/blob/dev/test_libs/pyspec/eth2spec/phase0/state_transition.py))

```python
def state_transition_to(state: BeaconState, up_to: Slot) -> BeaconState:
    while state.slot < up_to:
        spec.cache_state(state)
        if (state.slot + 1) % spec.SLOTS_PER_EPOCH == 0:
            process_epoch_transition(state)
        spec.advance_slot(state)
```

In other words, `process_epoch_transition` happens at slot `SLOTS_PER_EPOCH * k + SLOTS_PER_EPOCH - 1` for `k = 0, 1, 2, ...`, where the "current epoch" is `k`.

This does seem to contract this sentence in the spec:

> The per-epoch transitions, which happens at the start of the first slot of every epoch."

@djrtwo: Should it say "The per-epoch transitions, which happens at the start of the last slot of every epoch."?

dankrad:
> @djrtwo: Should it say "The per-epoch transitions, which happens at the start of the last slot of every epoch."?

I think it should actually be

- after the last slot of every epoch
- before the first slot of every epoch

(I prefer the first variant and suggest, to make it clearer, to call them "end of epoch" transition, because the epoch number at the time of the transition is the old epoch rather than the new one)

djrtwo:
> The per-epoch transitions, which happens at the start of the first slot of every epoch."

At the start of the state transition of the first slot of every epoch.

yes, the slot isn't incremented yet at that point in the state transition but it is certainly during the state transition of the 1st slot of the epoch.

That said, the debate really seems to be in #1043 😂

JustinDrake:
Closing in favour of #1043

PR #1018, opened 2019-05-01
Incorporate state_transition.py into state transition spec

JustinDrake:
This PR incorporates `state_transition.py` (along with a significant cleanup) into the state transition spec. This limits "magic" outside of the spec and addresses #946. We generally reduce fluff, increase readability, and reduce line count (20 lines in the spec, 140 lines overall). A bug is also fixed with the addition of `assert state.slot < block.slot` in `state_transition`.


specs/core/0_beacon-chain.md
```diff
@@ -1264,50 +1263,56 @@ def get_genesis_beacon_state(genesis_validator_deposits: List[Deposit],
...
-* The state caching caches the state root of the previous slot and updates block and state roots records.
-* The per-epoch transitions focus on the [validator](#dfn-validator) registry, including adjusting balances and activating and exiting [validators](#dfn-validator), as well as processing crosslinks and managing block justification/finalization.
-* The per-slot transitions focus on the slot counter.
-* The per-block transitions generally focus on verifying aggregate signatures and saving temporary records relating to the per-block activity in the `BeaconState`.
-
-Beacon blocks that trigger unhandled Python exceptions (e.g. out-of-range list accesses) and failed `assert`s during the state transition are considered invalid.
-
-Note: If there are skipped slots between a block and its parent block, run the steps in the [state-root](#state-caching), [per-epoch](#per-epoch-processing), and [per-slot](#per-slot-processing) sections once for each skipped slot and then once for the slot containing the new block.
+```python
+def state_transition(state: BeaconState, block: BeaconBlock) -> BeaconState:
```
```suggestion
def state_transition(state: BeaconState, block: BeaconBlock):
```

specs/core/0_beacon-chain.md
```diff
@@ -1264,50 +1263,56 @@ def get_genesis_beacon_state(genesis_validator_deposits: List[Deposit],
...
-3. The per-slot transitions, which happens at every slot.
-4. The per-block transitions, which happens at every block.
-
-Transition section notes:
-* The state caching caches the state root of the previous slot and updates block and state roots records.
-* The per-epoch transitions focus on the [validator](#dfn-validator) registry, including adjusting balances and activating and exiting [validators](#dfn-validator), as well as processing crosslinks and managing block justification/finalization.
-* The per-slot transitions focus on the slot counter.
-* The per-block transitions generally focus on verifying aggregate signatures and saving temporary records relating to the per-block activity in the `BeaconState`.
-
-Beacon blocks that trigger unhandled Python exceptions (e.g. out-of-range list accesses) and failed `assert`s during the state transition are considered invalid.
```
hwwhww: IMO it's still worth keeping this sentence for the readers.

specs/core/0_beacon-chain.md
```diff
@@ -1264,50 +1263,56 @@ def get_genesis_beacon_state(genesis_validator_deposits: List[Deposit],
...
-
-Note: If there are skipped slots between a block and its parent block, run the steps in the [state-root](#state-caching), [per-epoch](#per-epoch-processing), and [per-slot](#per-slot-processing) sections once for each skipped slot and then once for the slot containing the new block.
+```python
+def state_transition(state: BeaconState, block: BeaconBlock) -> BeaconState:
+    assert state.slot < block.slot
+    # Slot processing (including slots with no blocks)
```
hwwhww: Since you removed the notes of "skipped slots", maybe add more comment here?

specs/core/0_beacon-chain.md
```diff
@@ -1264,50 +1263,56 @@ def get_genesis_beacon_state(genesis_validator_deposits: List[Deposit],
 
 ## Beacon chain state transition function
 
-We now define the state transition function. At a high level, the state transition is made up of four parts:
+The post-state corresponding to a pre-state `state` and a block `block` is defined as `state_transition(state, block)`. State transitions that trigger an unhandled exception (e.g. a failed `assert` or an out-of-range list access) are considered invalid.
 
-1. State caching, which happens at the start of every slot.
-2. The per-epoch transitions, which happens at the start of the first slot of every epoch.
-3. The per-slot transitions, which happens at every slot.
-4. The per-block transitions, which happens at every block.
```
hwwhww: The `state_transition(state: BeaconState, block: BeaconBlock)` function assumes that to apply state transition, you need a `BeaconBlock`. But in the protocol, we should be able to apply per-slot transition without block. However, that's more difficult to present in code elegantly (with parameter realtime `slot`).

specs/core/0_beacon-chain.md
```diff
@@ -1632,19 +1623,17 @@ def process_final_updates(state: BeaconState) -> None:
...
 
 ```python
-def advance_slot(state: BeaconState) -> None:
-    state.slot += 1
+    process_randao(state, block)
+    process_eth1_data(state, block)
+    process_operations(state, block.body)
+    # verify_block_state_root(state, block)
```
hwwhww: I'd say enable it in the spec, and hack it in the tests.

/cc @djrtwo

specs/core/0_beacon-chain.md
JustinDrake: Agreed! :) See "State transitions that trigger an unhandled exception (e.g. a failed `assert` or an out-of-range list access) are considered invalid."

specs/core/0_beacon-chain.md
JustinDrake: Right. I think it's fine to define `state_transition` as a `(state: BeaconState, block: BeaconBlock) -> BeaconState` function. Two reasons:

* The per-slot state transition can be inferred by the per-block state transition. (This is similar to deserialisation being inferred from serialisation, or BLS signing being inferred from BLS verification.)
* We can clarify in the validator doc that validators are expected to run the per-slot state transition function in some instances.

JustinDrake:
Thanks @hwwhww for the good feedback :) I agree regarding comments. The `state_transition` function now roughly has one comment per line of code.

specs/core/0_beacon-chain.md
djrtwo: I think I lean toward

```python
def transition_to(state, slot):
    ...

def state_transition(state, block):
    transition_to(state, block.slot)
    process_block(state, block)
```

It is both readable and immediately useful in other areas of the spec. Right now, we kind of hand-wave how to transition through empty slots in the validator doc and the fork choice, but having this explicit function to point to would be more explicit. _Plus_ if we make the fork choice executable, we'll need this `transition_to` functionality

specs/core/0_beacon-chain.md
```diff
@@ -1691,11 +1683,25 @@ def process_eth1_data(state: BeaconState, block: BeaconBlock) -> None:
 
 #### Operations
 
-##### Proposer slashings
+```python
+def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
+    assert len(body.deposits) == min(MAX_DEPOSITS, state.latest_eth1_data.deposit_count - state.deposit_index)
```
djrtwo: This probably deserves a comment

```suggestion
    # require deposits if unprocessed deposits exist
    assert len(body.deposits) == min(MAX_DEPOSITS, state.latest_eth1_data.deposit_count - state.deposit_index)
```

specs/core/0_beacon-chain.md
```diff
@@ -1691,11 +1683,25 @@ def process_eth1_data(state: BeaconState, block: BeaconBlock) -> None:
...
+    assert len(body.transfers) == len(set(body.transfers))
```
djrtwo: maybe comment here

```suggestion
    # Verify all transfers are unique
    assert len(body.transfers) == len(set(body.transfers))
```

specs/core/0_beacon-chain.md
djrtwo: a bit torn here. 

options:
1. uncomment and leave as is
2. add `verify_state_root` boolean in `process_block` and `state_transition` and a note about in which context you would run it as True or False (receiving a block vs building a block)
3. have `verify_block_state_root` live outside of the state transition function entirely and as a note on block validity below.

I think I like (2) the best, but am open to debate.
*

specs/core/0_beacon-chain.md
JustinDrake: Here's the new organisation for `state_transition`:

* `process_slots`
  * `process_slot`
  * `process_epoch`
    * `process_justification_and_finalization`
    * `process_crosslinks`
    * `process_rewards_and_penalties`
    * `process_registry_updates`
    * `process_slashings`
    * `process_final_updates`
* `process_block`
  * `process_block_header`
  * `process_randao`
  * `process_eth1_data`
  * `process_operations`
    * `process_proposer_slashing`
    * `process_attester_slashing`
    * `process_attestation`
    * `process_deposit`
    * `process_voluntary_exit`
    * `process_transfer`
  * `process_state_root`

specs/core/0_beacon-chain.md
JustinDrake: My suggestion would be for the test bench to populate the `state_root` immediately prior to verification in case it is set to `ZERO_HASH`. That way we can keep the spec clean.

djrtwo: some sort of monkey patch?

JustinDrake: Yes

specs/core/0_beacon-chain.md
arnetheduck: fwiw, we have two separate functions for block and blockless transtions in nimbus: https://github.com/status-im/nim-beacon-chain/blob/17fa4997556d0a69814144744e389cfa97c8f5ee/beacon_chain/state_transition.nim#L1233

part of the reason is that validation of block data and application of it is mixed up, so we will have modified the old state before knowing that the block is valid, which means we must also be able to roll back. blockless transitions however can't fail, so there's a difference / simplification in error handling as well.

specs/core/0_beacon-chain.md
JustinDrake: I've added a bunch of comments :) For skip slots, I have "Process slots (including those with no blocks) since block"

specs/core/0_beacon-chain.md
```diff
@@ -1264,50 +1262,59 @@ def get_genesis_beacon_state(genesis_validator_deposits: List[Deposit],
...
-
-### State caching
+```python
+def state_transition(state: BeaconState, block: BeaconBlock) -> BeaconBlock:
+    # Process slots (including those with no blocks) since block
+    process_slots(state, block.slot)
+    # Process block
+    process_block(state, block)
+    # Return post-state
+    return state
```
hwwhww: As noted in #1059, `state` is mutable. I think we are mixing how we handle mutable and immutable objects.

For example, writing `process_block(state, block)` implies it's mutable, but then having `return state` is confusing and can be removed since we are expecting that the pyspec user/reader should know `state` is mutable.

JustinDrake: We are definitely mixing things. Conceptually thinking of `state_transition` as a `(BeaconState, BeaconBlock) -> BeaconState` is cleanest IMO. Will mull.

JustinDrake: Implemented your great suggestion to add a `try` `except` 🎉

djrtwo:
unsure of the exception handling. See https://github.com/ethereum/eth2.0-specs/issues/1059#issuecomment-490155236

djrtwo:
I think we should revert the exception handling here and get it merged and discuss otherwise in #1059

JustinDrake:
> I think we should revert the exception handling here

Done. Feel free to merge :)

djrtwo:
@JustinDrake We didn't decide on how to handle state root validation. I moved the `process_state_root` rfunction out to `state_transition` and added a flag. this seems natural to me for clients as they will certainly have a flag to be able to run either way, and defaulting to false translates easily to our current testing infrastructure.

Waiting for comment before merge

JustinDrake:
> I moved the process_state_root rfunction out to state_transition

Makes sense. I like the idea of inlining it!

> and added a flag

We probably want to clean this up eventually but don't want to block this PR. Suggest merging :)

Issue #667, opened 2019-02-20
Use custom types in data structures


Nashatyrev:
As soon as there are strict custom types for helper functions may be it makes sense to use them appropriately for data structures members like below.
```python
BeaconState
{
    # Misc
    'slot': Slot,
    'genesis_time': 'uint64',
    'fork': Fork,  # For versioning hard forks

    # Validator registry
    'validator_registry': [Validator],
    'validator_balances': [Gwei],
    'validator_registry_update_epoch': Epoch,

    # Randomness and committees
    'latest_randao_mixes': [Bytes32],
    'previous_shuffling_start_shard': Shard,
    'current_shuffling_start_shard': Shard,
    'previous_shuffling_epoch': Epoch,
    'current_shuffling_epoch': Epoch,
    'previous_shuffling_seed': Bytes32,
    'current_shuffling_seed': Bytes32,
...
```
In our Java implementation we already converted the structures members (like [this](https://github.com/harmony-dev/beacon-chain-java/blob/1ec9b9b10c9a474c7e06ff1f473e3b99c15605a5/core/src/main/java/org/ethereum/beacon/core/BeaconState.java#L39)) and constants (like [this](https://github.com/harmony-dev/beacon-chain-java/blob/335e0d91df44882adf4245e91bde102fea887c5a/core/src/main/java/org/ethereum/beacon/core/spec/InitialValues.java#L21)) to custom types (where applicable) which significantly improved readability and type safety.

JustinDrake:
Agreed this improves readability! We can even define a `Hash` custom type.

Nashatyrev:
Cool! I can take care of PR then. 
We also added `Time` type (aka number of seconds), though it's not too widespread.
Another question: is there any means to specify the array index type, like ?
```
  'validator_balances': [ValidatorIndex => Gwei],
...
  'latest_block_roots': [Slot => Hash],
  'latest_index_roots': [Epoch => Hash],
```

hwwhww:
I'm not sure if I like it or not. Some notes:
1. The current data structure section is defined with SSZ serialization types. I’d say let's not mix custom type and SSZ in the same section; so if we want to use custom types to describe data structure, do it **fully**. That said, some other types like `Bitfield: bytes`, `Bool: bool` (#658) would need to be defined as well.
    - Then we will find that `'justification_bitfield': 'uint64'` is suspicious but correct…ahhh..  (see #371)
[...]

Issue #371, opened 2018-12-28
Is justification_bitfield an integer or a `bytes` type?


ChihChengLiang:
`justification_bitfield` is specified `uint64` in the spec.

Would it make more sense to specify it as `bytes` type? Since the bitfield keyword indicates a byte type and it involves many bitwise operations as well.

vbuterin:
In python it's easier to do bit operations on integers than bytes :smile: 

If we go bytes, then to keep the spec being correctly executable python, we would need to make helpers for `bitwise_or`, `bitwise_and` and `shift`.

Issue #1008, opened 2019-04-28
Generalized Merkle Tree Index: use 0-based indexing


mratsim:
The current specs for generalized merkle tree uses 1-based indexing (https://github.com/ethereum/eth2.0-specs/blob/2787fea5feb8d5977ebee7c578c5d835cff6dc21/specs/light_client/merkle_proofs.md#generalized-merkle-tree-index)

The differences between 0 and 1 based indexing are the following:

|             | root at 0 | root at 1 |
|-------------|-----------|-----------|
| Left child  | i*2 + 1  | i*2      |
| Right child | i*2 + 2  | i*2 + 1  |
| Parent      | (i-1)/2  | i/2      |

## Convention

Using 0-based indexing would be using the common convention of most programming languages (save Fortran, Matlab and Julia).

## Errors

It is well known that mixing 0-based and 1-based indexing tends to create off-by-one errors, we already had several such spec bugs in the past. And using 1-based indexing is introducing new risks.

## Memory saving

I expect clients to allocate binary trees with power-of-2 sizes for efficiency reasons.
If we have N nodes in the merkle tree where N is a power of 2 we will waste N-1 node space.

Concrete example: 1 byte can store 256 values in the range [0, 255]
Assuming we have values in range [1, 256] instead like the current proposition we would need 2 bytes of storage and will waste the [257-511] range which represent 255 values.

### Alignment issues

Assuming we store merkle proofs in arrays of uint32 at a low-level, they would need to be aligned to 32-bit boundaries on many architecture that would benefit the most from light clients (ARM, MIPS, ...).

Furthermore Merkle proofs would probably be excellent candidates to be accelerated by SIMD/vector hardware which requires also requires alignment.

This off-by-one will cause unaligned loads/store that implementers will have to work around, probably via padding further increasing memory waste or via temporaries.

## Potential counter-argument: extra computation

Some might be worried that 0-based is slower than 1-based indexing
where we can just use `2*i` and `2*i+1` to find children instead of `2*i+1` and `2*i+2`
and `i/2` to find parents instead of `(i-1)/2`.
  1. It doesn't make a difference because at a low-level, CPUs support Scale*Index+Offset
     addressing mode if scale is a power of 2. Execution time is the same. (This is called SIB addressing in x86 and scaled register in ARM).
  2. The performance is dominated by fetching data from memory. During
     the previous memory fetch, the CPU can precompute the next location to fetch.
     as there is no data dependencies.

[...]

PR #1019, opened 2019-05-01
Cleaner bitfields

JustinDrake:
This draft PR tries to fix a longstanding wart (see #371, #667) regarding bitfields. We define bitfields as lists or vectors of `bit`s (as opposed to `bytes` and `uint64` respectively). Benefits:

* Remove ugly-as-hell helper functions `verify_bitfield` and `get_bitfield_bit` (save 28 lines)
* No more bit manipulation with `justification_bitfield`—more readable
* Merkleisation-friendly `justification_bitfield` (i.e. `justification_bitfield` does not change unless a new epoch is justified)
* Increased number of bits in `justification_bitfield` in case dApps and light clients find it useful (set to `EPOCHS_PER_HISTORICAL_VECTOR == 8192`)
* Semantically natural typing (`bytes` and `uint64` were hacks)
* Use `bit` type for `custody_bit`

This requires two minor SSZ changes:

1) Add `bit` as an alias to `bool` (cosmetic)
2) Pack lists and vectors of `bit`s similar to packing of `uintN` (substantive)

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -206,7 +206,7 @@ These configurations are updated for releases, but may be out of sync during `de
 | `PERSISTENT_COMMITTEE_PERIOD` | `2**11` (= 2,048)  | epochs | 9 days  |
 | `MAX_CROSSLINK_EPOCHS` | `2**6` (= 64) | epochs | ~7 hours |
 | `MIN_EPOCHS_TO_INACTIVITY_PENALTY` | `2**2` (= 4) | epochs | 25.6 minutes |
-| `EPOCHS_PER_HISTORY_OBJECT` | `2**13` (= 8,192) | epochs | ~36 days |
+| `EPOCHS_PER_HISTORICAL_VECTOR` | `2**13` (= 8,192) | epochs | ~36 days |
 
 * `MAX_CROSSLINK_EPOCHS` should be a small constant times `SHARD_COUNT // SLOTS_PER_EPOCH`
 
@@ -546,7 +546,7 @@ The types are defined topologically to aid in facilitating an executable version
     'balances': ['uint64'],
 
     # Randomness and committees
-    'latest_randao_mixes': ['bytes32', EPOCHS_PER_HISTORY_OBJECT],
+    'latest_randao_mixes': ['bytes32', EPOCHS_PER_HISTORICAL_VECTOR],
     'latest_start_shard': 'uint64',
 
     # Finality
@@ -556,7 +556,7 @@ The types are defined topologically to aid in facilitating an executable version
     'current_justified_epoch': 'uint64',
     'previous_justified_root': 'bytes32',
     'current_justified_root': 'bytes32',
-    'justification_bitfield': ['bit', EPOCHS_PER_HISTORY_OBJECT],
+    'justification_bitfield': ['bit', EPOCHS_PER_HISTORICAL_VECTOR],
     'finalized_epoch': 'uint64',
     'finalized_root': 'bytes32',
 
@@ -565,8 +565,8 @@ The types are defined topologically to aid in facilitating an executable version
     'previous_crosslinks': [Crosslink, SHARD_COUNT],
     'latest_block_roots': ['bytes32', SLOTS_PER_HISTORICAL_ROOT],
     'latest_state_roots': ['bytes32', SLOTS_PER_HISTORICAL_ROOT],
-    'latest_active_index_roots': ['bytes32', EPOCHS_PER_HISTORY_OBJECT],
-    'latest_slashed_balances': ['uint64', EPOCHS_PER_HISTORY_OBJECT],  # Balances slashed at every withdrawal period
+    'latest_active_index_roots': ['bytes32', EPOCHS_PER_HISTORICAL_VECTOR],
+    'latest_slashed_balances': ['uint64', EPOCHS_PER_HISTORICAL_VECTOR],  # Balances slashed at every withdrawal period
     'latest_block_header': BeaconBlockHeader,  # `latest_block_header.state_root == ZERO_HASH` temporarily
     'historical_roots': ['bytes32'],
 
@@ -878,8 +878,8 @@ def get_randao_mix(state: BeaconState,
     """
     Return the randao mix at a recent ``epoch``.
     """
-    assert get_current_epoch(state) - EPOCHS_PER_HISTORY_OBJECT < epoch <= get_current_epoch(state)
-    return state.latest_randao_mixes[epoch % EPOCHS_PER_HISTORY_OBJECT]
+    assert get_current_epoch(state) - EPOCHS_PER_HISTORICAL_VECTOR < epoch <= get_current_epoch(state)
+    return state.latest_randao_mixes[epoch % EPOCHS_PER_HISTORICAL_VECTOR]
 ```
 
 ### `get_active_index_root`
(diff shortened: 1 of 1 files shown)


dankrad:
I'm in favour of this because it adds clarity.

dankrad:
> We could alias bool to uint1 so that the packing logic only needs to be defined once for uintN. We could also alias bit to bool to fit with the bitfield nomenclature.

Not so sure about these. I think uint1 would still kind of be a special case as it needs to be bit-aligned, not byte-aligned; nice to mark this special case by a different name (bool).

I would suggest "boolfield" as it is much preferrable to have functional descriptors (the "bit" part is only an implementation detail after all); then again, bitfield is a much more common term.

JustinDrake:
Screenshot 2019-05-01 at 16 57 39

vbuterin:
Wouldn't this also require a special case in SSZ? It makes the phase 0 spec cleaner but the SSZ serialization more complicated by about the same amount...

Though you could argue that adding the bigint type to SSZ has a similar effect, and I've advocated for that :)

JustinDrake:
> Wouldn't this also require a special case in SSZ?

Yes, the substantive change involves packing lists and vectors of bits similar to packing lists and vectors of `uintN`. My best guess is that it is a 2-liner in SSZ, versus a 32-liner in the state transition function. I'll wait until after the SSZ rewrite is done so we can make a more informed decision on the SSZ complexity increase (cc @protolambda).

dankrad:
> My best guess is that it is a 2-liner in SSZ, versus a 32-liner in the state transition function. I'll wait until after the SSZ rewrite is done so we can make a more informed decision on the SSZ complexity increase (cc @protolambda).

I think I would argue that even if it is not a complexity reduction, we should just do it because SSZ is the right place to deal with data types, and this is clearly a question of data types. Not having to worry about bitfields at higher levels of the spec makes it more readable; basically it should be possible to read the Eth2.0 spec without reading SSZ and just assuming that SSZ will do the right thing when it comes to bools/bits.

vbuterin:
> Yes, the substantive change involves packing lists and vectors of bits similarly to packing lists and vectors of uintN (yay for consistency!) My best guess is that it is a 2-liner in SSZ, versus a 32-liner in the state transition function.

Probably a bit more than 2 lines. Here's my attempt:

```
def serialize_bits(bits):
    o = bytearray((len(bits) + 7) // 8)
    for i, bit in enumerate(bits):
        o[i // 8] |= bit << (i % 8)
    return bytes(o)
```
Note however that this is insufficient because unlike all other SSZ data structures, where we can deduce the number of items from the length, here we cannot do so, because an N byte list could correspond to a [8N .... 8N+7] size bitfield. So we'd need a custom length prefix (we don't need that today because we're feeding in the length from the committee size, so in some sense we have a "static" data type that's not really "static"). So perhaps we'd have to do something like `return bytes([len(bits) % 8]) + o`

My instinct is still that bitfields in SSZ are on net a good thing, but there definitely _are_ costs to doing it.

> basically it should be possible to read the Eth2.0 spec without reading SSZ and just assuming that SSZ will do the right thing when it comes to bools/bits.

Fully agree with this!

vbuterin:
Ooh, I have another idea for how to serialize variable-sized bitfields: use the generalized index scheme from https://github.com/ethereum/eth2.0-specs/blob/dev/specs/light_client/merkle_proofs.md#generalized-merkle-tree-index. As I describe [here](https://github.com/ethereum/eth2.0-specs/issues/1008#issuecomment-487445311), there is a natural correspondence between generalized indices and bitfields, in that a generalized index _is_ a bitfield, with 0s representing when to go left and 1s representing when to go right, but with an extra postfix of 1 to signify where the bitfield starts from. One could take this exact scheme and use it to re-interpret any bigint as a bitfield, and then we'd just need SSZ to support bigints.

JustinDrake:
> Probably a bit more than 2 lines. Here's my attempt:

Serialisation is not part of consensus :) Serialisation-specific implementation details (such as a custom length prefix—which is a good point) do not contribute towards the "strict" line count. This PR is a natural continuation to #924.

In the [SSZ rewrite in progress](https://colab.research.google.com/drive/1vqb3kLTNyXouUuUUkIiOvYpDWZwEcji_#scrollTo=vNN0iY1WHcwk) (where the short Merkleisation-specific code is to be incorporated in `0_beacon_chain.md`) the main thing to do is generalise the `pack` function from bytes to bits. I wouldn't be surprised if it ends up being 1-liner (or even a 0-liner) 😂Proof is in the pudding though.

> I have another idea for how to serialize variable-sized bitfields: use the generalized index scheme 

Interesting!

> and then we'd just need SSZ to support bigints

Can we simply alias `bigint` and `bitfield` to `list[bit]` and call it a day?

vbuterin:
>Can we simply alias bigint and bitfield to list[bit] and call it a day?

We could, though I would argue that `bigint` is the more natural "base-ish" type. The fact is that the way this serialization would work can't avoid being fundamentally different to how every other `List[x]` works, and we don't want to confuse people by making non-similar things look similar. That said, naming is secondary, if we're ok with (i) the encoding algorithm for bitfields defined by `def bitfield_to_int(bits): return (1 << len(bits)) + sum([(1 << i) * bit for i, bit in enumerate(bits)])` or its inverse `def int_to_bitfield(n): return [(n >> i) % 2 for i in range(log2(n))]` and (ii) adding an SSZ type that's identical to `bytes` except it enforces that leading zero bytes are illegal, then I'd be ok describing it either way, as they're equivalent under the hood.

> the main thing to do is generalise the pack function from bytes to bits

The big problem here is that in pretty much every computer language, and the underlying architectures, things can't be N bits long for N % 8 != 0. The very output of `serialize`, and the expected input to `hash`, are both `bytes`.

> Serialisation is not part of consensus :) 

Everything I said above applies to tree-hashing too :)

JustinDrake:
> `def bitfield_to_int(bits): return (1 << len(bits)) + sum([(1 << i) * bit for i, bit in enumerate(bits)])`
> `def int_to_bitfield(n): return [(n >> i) % 2 for i in range(log2(n))]`

This is neat :) The main question is whether we want 0-based or 1-based indexing of generalised indices (#1008). An argument in favour of 0-based indexing is that bigints have support for 0, so with 0-based indexing the concepts of "bigint" and "generalised index" become equivalent.

> adding an SSZ type that's identical to bytes except it enforces that leading zero bytes are illegal

It's still not clear to me why adding an SSZ type is necessary given we can use existing types.

> the expected input to `hash` is `bytes`

Right, the current `chunkify` function should be generalised to right-pad the last chunk with zero bits (as opposed to zero bytes). Maybe replace

```python
def chunkify(byte_string):
    byte_string += b'\x00' * (-len(byte_string) % 32)
    return [byte_string[i:i + 32] for i in range(0, len(byte_string), 32)]
```

with something like

```python
def chunkify(bits):
    bits += [0b0] * (-len(bits) % 256)
    byte_string = bytes([sum([byte[b] << b for b in range(0, 8)]) for byte in zip(*(iter(bits),) * 8)])
    return [byte_string[i:i + 32] for i in range(0, len(byte_string), 32)]
```

vbuterin:
I'm really inclined to say that making bits rather than bytes be the base unit of SSZ just for this single use case is overkill.

> It's still not clear to me why adding an SSZ type is necessary given we can use existing types.

To preserve the norm that there's only one valid way to serialize something. If we have an injective map from bitfields to bigints, then we want there to be only one valid way to serialize a bigint, which means that deserializing eg. `\x05\x00` should return an error (it should be `\x05`).

> An argument in favour of 0-based indexing is that bigints have support for 0, so with 0-based indexing the concepts of "bigint" and "generalised index" become equivalent.

That is interesting. So if we offset everything by 1 then we get a fully bijective map. Though that's only a really strong desideratum if the goal is to make `List[bit]` be the base type, and I still feel like that would confuse things, and we should stick to SSZ objects all being byte-based, in which case bigint would be the base type, and we'd only need an injective map (ie. each bitfield has a unique corresponding bigint, not necessarily vice versa). In that case, 1-based is better because the bit representation of the bigint would just be the bitfield with an extra 1 on the end, as opposed to do an awkward subtract-and-carry operation on a bitfield to deserialize it.

djrtwo:
> I'm really inclined to say that making bits rather than bytes be the base unit of SSZ just for this single use case is overkill.

Late to the party but very much agree with this.

djrtwo:
Are we closing this in favor of the bigint PR?

JustinDrake:
> Are we closing this in favor of the bigint PR?

I guess we want to merge the two. The bigint PR current only touches `simple-serialize.md`. This PR cleans up `0_beacon-chain.md` and can be adapted to however bigints get implemented.

djrtwo:
@dankrad Can we close this in favor of your coming PR?

JustinDrake:
Closing in favour of #1224

PR #1020, opened 2019-05-01
Revert 981 only setup py

djrtwo:


hwwhww:
@djrtwo @protolambda 
To make sure I understand it correctly, #981 was reverted because it was merged before fully review process? And I should open another PR for it?

@protolambda 

> - installs local module requirements, in a fresh copy (`make clean` or fresh git clone)

Do you mean in CI? It seems another issue, not related to `setup.py` route, correct?


> - doesn't move the requirements to the full test_libs folder. That's just a hack to avoid using local dependencies. I prefer to separate debug/util packages from the pyspec module.

Do you mean my latest comment:
https://github.com/ethereum/eth2.0-specs/pull/981#issuecomment-485638169: IMO the better solution may be using one single `setup.py` for all test_libs sub packages. What do you think about it?

Continue the discussion on #981, I'm still pro-single `setup.py` in `test_libs` than what #981 was (Installing `config_helpers` separately).

If we using one single `setup.py` in `test_libs`, we can still make deps to:
```python
deps = {
    'config_helpers': [
        "ruamel.yaml==0.15.87",
    ],
}
deps += {
    'pyspec': [
        "eth-utils>=1.3.0,<2",
        "eth-typing>=2.1.0,<3.0.0",
        "pycryptodome==3.7.3",
        "py_ecc>=1.6.0",
    ],
    'pyspec-test': [
        "pytest>=3.6,<3.7",
    ] + deps['config_helpers'],
    ...
}

deps += {
    'pyspec-dev': 'pyspec' + ''pyspec-test'
}
```
And install it with `pip3 install -e .[pyspec]` or `pip3 install -e .[pyspec-test]`.

I see there's more value in organizing all dependencies in one file:
- Easier to manage the dependencies. For example, both `gen_base` and `config_helpers` are depending on `ruamel.yaml==0.15.87`, you can update it once in the one-single `setup.py`.
- If we want to release it in pypi with spec release cycle, one-single `setup.py` makes releasing easier.
- Not a hacky way. :)

> don't trust CI for an upgrade, it caches things.

Right, that is a nightmare. Next time we can probably update the version: https://circleci.com/docs/2.0/caching/#clearing-cache when updating without changing checksum file.

protolambda:
@hwwhww Will continue discussion in #981

PR #1021, opened 2019-05-01
Clean up genesis

JustinDrake:
Various minor cosmetic changes.


djrtwo:
Not sure I think this is worthwhile.
Genesis state is super related to the state transition function and utilizes some of the functions in that doc

vbuterin:
Agree. This creates a weird inter-file dependency graph: genesis and phase 1 both depend on phase 0.

JustinDrake:
Agreed that a single file is preferred :) Reopening because there are genesis cleanups that I want incorporated. Will rework this PR back to a single file :)

PR #1024, opened 2019-05-01
implement SOS serialization in minimal_ssz.py

protolambda:
Note: PR into master, to get it into v0.6.1. Will make PRs into special dev branch for future v0.6.x updates after v0.6.1

Implemented SOS serialization for minimal_ssz.py (util of the executable spec).

Some notes:
- Container encoding differs slightly from previous implementation semantics: I call the helper function `encode_series`, since it does lists, vectors and containers. Not just containers.
- Removed unnecessary summing & list concatenation. Worth a few lines to get fast CI + tests.
- ~~The encoding seems funny, but is correct from what I understand. Example outlined below.~~ Edit: funny encoding was just incorrect.
- The spec can be improved somewhat:
  - `serialize(sum(fixed_lengths + variable_lengths[:i])) ` is not just inefficient, it is also missing explicit expression of the expected length of the serialization output. Should be `BYTES_PER_LENGTH_OFFSET`, which is 4, i.e. `uint32`, something different from most of the other parts of the spec.
  - typo: Rec**~~c~~**ursively

### DIFF
--- a/test_libs/pyspec/eth2spec/utils/minimal_ssz.py
+++ b/test_libs/pyspec/eth2spec/utils/minimal_ssz.py
@@ -163,7 +163,7 @@ def serialize_value(value, typ=None):
         return encode_series(value, [typ[0]] * len(value))
     # "bytes" (variable size)
     elif isinstance(typ, str) and typ == 'bytes':
-        return (b"\x00" * BYTES_PER_LENGTH_OFFSET) + coerce_to_bytes(value)
+        return coerce_to_bytes(value)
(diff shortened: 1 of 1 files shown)


protolambda:
@JustinDrake You worked on the SOS wording in the SSZ spec, and are working with bitfields in PR #1019, what do you think about the encoding example above?

protolambda:
Now that I think about it, the 0x0000... actually has to be 0x04000... To account for the offset itself. Can fix that later. Base point still stands: do we actually want it to be there?

Issue #1017, opened 2019-05-01
Underflow in generate_seed for GENESIS_EPOCH


mkalinin:
## Issue
Encountered in `v0.6.0`.

With `epoch = 0 (GENESIS_EPOCH)` [generate_seed](https://github.com/ethereum/eth2.0-specs/blob/v0.6.0/specs/core/0_beacon-chain.md#generate_seed) passes `epoch = GENESIS_EPOCH - 1` value to `get_randao_mix`:
```python
...
    return hash(
        get_randao_mix(state, epoch - MIN_SEED_LOOKAHEAD) +
...
```
With `epoch = -1` [get_randao_mix](https://github.com/ethereum/eth2.0-specs/blob/v0.6.0/specs/core/0_beacon-chain.md#get_randao_mix) throws an assertion error as condition 
`epoch <= get_current_epoch(state)` gets violated.

JustinDrake:
> With epoch = -1 get_randao_mix throws an assertion error as condition
epoch <= get_current_epoch(state) gets violated.

`epoch <= get_current_epoch(state)` corresponds to ` -1 <= 0` which is not violated. Still, we shouldn't underflow.

@djrtwo I remember you saying that in Python we can do some magic to the minus operator `-` so that if `a - b` is ever negative we `assert False`. That will allow the executable spec to automatically find those nasties, especially when fuzz testing comes along.

djrtwo:
> epoch <= get_current_epoch(state) corresponds to -1 <= 0 which is not violated. Still, we shouldn't underflow.

violated when it's actually `uints` instead of python's int type. Anything less than 0 is an underflow and will likely be an issue in strongly typed languages.

> @djrtwo I remember you saying that in Python we can do some magic to the minus operator - so that if a - b is ever negative we assert False

yes, it is possible to monkeypatch the `-` operator. Might be useful in this case.

----- 

nice find @mkalinin, I think it likely appropriate to put a min of genesis_epoch on that call to get_randao_mix. 

The actual assert in `get_randao_mix` is going to underflow as well with `get_current_epoch(state) - LATEST_RANDAO_MIXES_LENGTH`.

I'll clean up these two and another function in a PR now

djrtwo:
addressed in #1027

mkalinin:
> yes, it is possible to monkeypatch the - operator. Might be useful in this case.

[...]

PR #1027, opened 2019-05-01
Fix underflows from #1017

djrtwo:
prevent underflows in `generate_seed`, `get_randao_mix`, and `get_active_index_root`.

Note that for epochs less than `MIN_SEED_LOOKAHEAD` we use `ZERO_HASH` rather than just pulling from `get_randao_mix(state, GENESIS_EPOCH)` because the randao_mix at `GENESIS_EPOCH` is not static and would thus break lookahead.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -800,7 +800,8 @@ def get_randao_mix(state: BeaconState,
     """
     Return the randao mix at a recent ``epoch``.
     """
-    assert get_current_epoch(state) - LATEST_RANDAO_MIXES_LENGTH < epoch <= get_current_epoch(state)
+    min_epoch = epoch - LATEST_RANDAO_MIXES_LENGTH + 1 if epoch > LATEST_RANDAO_MIXES_LENGTH else GENESIS_EPOCH
+    assert min_epoch <= epoch <= get_current_epoch(state)
     return state.latest_randao_mixes[epoch % LATEST_RANDAO_MIXES_LENGTH]
 ```
 
@@ -812,7 +813,8 @@ def get_active_index_root(state: BeaconState,
     """
     Return the index root at a recent ``epoch``.
     """
-    assert get_current_epoch(state) - LATEST_ACTIVE_INDEX_ROOTS_LENGTH + ACTIVATION_EXIT_DELAY < epoch <= get_current_epoch(state) + ACTIVATION_EXIT_DELAY
+    min_epoch = get_current_epoch(state) + ACTIVATION_EXIT_DELAY - LATEST_ACTIVE_INDEX_ROOTS_LENGTH + 1 if epoch > LATEST_ACTIVE_INDEX_ROOTS_LENGTH - ACTIVATION_EXIT_DELAY else GENESIS_EPOCH
+    assert min_epoch <= epoch <= get_current_epoch(state) + ACTIVATION_EXIT_DELAY
     return state.latest_active_index_roots[epoch % LATEST_ACTIVE_INDEX_ROOTS_LENGTH]
 ```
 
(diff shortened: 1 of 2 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -812,7 +813,8 @@ def get_active_index_root(state: BeaconState,
     """
     Return the index root at a recent ``epoch``.
     """
-    assert get_current_epoch(state) - LATEST_ACTIVE_INDEX_ROOTS_LENGTH + ACTIVATION_EXIT_DELAY < epoch <= get_current_epoch(state) + ACTIVATION_EXIT_DELAY
+    min_epoch = get_current_epoch(state) + ACTIVATION_EXIT_DELAY - LATEST_ACTIVE_INDEX_ROOTS_LENGTH + 1 if epoch > LATEST_ACTIVE_INDEX_ROOTS_LENGTH - ACTIVATION_EXIT_DELAY else GENESIS_EPOCH
```
hwwhww: 1. I think you wanted to use both `get_current_epoch(state)` here? `... if get_current_epoch(state) > LATEST_ACTIVE_INDEX_ROOTS_LENGTH - ACTIVATION_EXIT_DELAY ...`
2. Though I can see `LATEST_ACTIVE_INDEX_ROOTS_LENGTH` definitely is greater than `ACTIVATION_EXIT_DELAY`, but it added a config validity condition https://github.com/ethereum/eth2.0-specs/issues/407.


Solution: keep it nonnegative
```suggestion
    min_epoch = get_current_epoch(state) + ACTIVATION_EXIT_DELAY - LATEST_ACTIVE_INDEX_ROOTS_LENGTH + 1 if get_current_epoch(state) + ACTIVATION_EXIT_DELAY > LATEST_ACTIVE_INDEX_ROOTS_LENGTH else GENESIS_EPOCH
```

specs/core/0_beacon-chain.md
```diff
@@ -800,7 +800,8 @@ def get_randao_mix(state: BeaconState,
     """
     Return the randao mix at a recent ``epoch``.
     """
-    assert get_current_epoch(state) - LATEST_RANDAO_MIXES_LENGTH < epoch <= get_current_epoch(state)
+    min_epoch = epoch - LATEST_RANDAO_MIXES_LENGTH + 1 if epoch > LATEST_RANDAO_MIXES_LENGTH else GENESIS_EPOCH
```
hwwhww: I think it should be `get_current_epoch(state)` instead of `epoch`:
```suggestion
    min_epoch = get_current_epoch(state) - LATEST_RANDAO_MIXES_LENGTH + 1 if get_current_epoch(state) > LATEST_RANDAO_MIXES_LENGTH else GENESIS_EPOCH
```

JustinDrake:
A few initial comments:

1) Lines such as `    min_epoch = get_current_epoch(state) + ACTIVATION_EXIT_DELAY - LATEST_ACTIVE_INDEX_ROOTS_LENGTH + 1 if epoch > LATEST_ACTIVE_INDEX_ROOTS_LENGTH - ACTIVATION_EXIT_DELAY else GENESIS_EPOCH` do not fit within the 120 character limit.
2) Readability for these helper functions has been significantly reduced.
3) Looking forward to removing the untriggerable `assert` in `get_block_root_at_slot`, `get_randao_mix`, `get_active_index_root`, `get_shuffled_index`, `integer_squareroot`, `get_matching_source_attestations` 😂Keen to find alternative testing solutions that do not pollute the state transition spec (in a similar vein to #1029).

specs/core/0_beacon-chain.md
```diff
@@ -825,7 +827,7 @@ def generate_seed(state: BeaconState,
     Generate a seed for the given ``epoch``.
     """
     return hash(
-        get_randao_mix(state, epoch - MIN_SEED_LOOKAHEAD) +
+        get_randao_mix(state, epoch - MIN_SEED_LOOKAHEAD) if epoch >= MIN_SEED_LOOKAHEAD else ZERO_HASH +
```
JustinDrake: Brainstorming alternatives: `get_randao_mix(state, epoch + LATEST_RANDAO_MIXES_LENGTH - MIN_SEED_LOOKAHEAD)`.

djrtwo:
@JustinDrake I'm okay taking the asserts out as they are untriggerable in the context of the state transition function. I added notes in the comment on the expected range for implementers whom I know utilize in other contexts

PR #1028, opened 2019-05-02
Clean up verify_indexed_attestation

JustinDrake:
Cosmetic changes:

* Add 4 lines of comments (now every statement has a comment)
* Avoid unnecessary `assert` (the end goal for me is for `assert`s to be exclusive to the operation processing helpers).
* Merge `return`s into one (increase readability, reduce verbosity)
* Use shorter-named `bit_0_indices` and `bit_1_indices` helper variables

Substantive change:

* Remove the condition that `len(0_indices) + len(1_indices) > 0`. This condition is redundant in the context of `process_attester_slashing` because of `slashed_any`. It is artificial in `process_attestation` where validators are incentivised to maximise new attestations.


specs/core/0_beacon-chain.md
```diff
@@ -1018,37 +1018,33 @@ def convert_to_indexed(state: BeaconState, attestation: Attestation) -> IndexedA
...
     """
-    custody_bit_0_indices = indexed_attestation.custody_bit_0_indices
-    custody_bit_1_indices = indexed_attestation.custody_bit_1_indices
+    bit_0_indices = indexed_attestation.custody_bit_0_indices
+    bit_1_indices = indexed_attestation.custody_bit_1_indices
 
-    # Ensure no duplicate indices across custody bits
-    assert len(set(custody_bit_0_indices).intersection(set(custody_bit_1_indices))) == 0
-
-    if len(custody_bit_1_indices) > 0:  # [TO BE REMOVED IN PHASE 1]
```
djrtwo: Tiered statements like this lends itself more to debugging. when you stack everything into one bool and one code path, you get no granularity of what was actually set off.

Even if one were debugging the pytests, this would become very difficult to even use print debugging or something similar

JustinDrake: Agreed that debugging takes a hit. There is a counter-balancing readability gain from avoiding the double negatives (`if not(...)` followed by `return False`) and the reduced verbosity.

My gut feel is that the spec prioritises readability over debugability. Don't care too much either way so will let you decide :)

hwwhww: > Even if one were debugging the pytests, this would become very difficult to even use print debugging or something similar

From my experience on implementing it, `return False` is also hard to debug for each check.

*If* we really want to make it debugging-friendly, I'd say it's better to (i) change it from `verify_indexed_attestation` to `validate_indexed_attestation` and (ii) change `return False` to either `raise Exception()` or `assert`.  

Using `raise` may be a good compromise between this PR and debugging-friendly since we can put comments in the Exception message. For example:

```python
    if len(bit_1_indices) != 0:
        raise Exception("Error during verifying no index has custody bit equal to 1 [to be removed in phase 1]")

    if len(bit_0_indices) + len(bit_1_indices) <= MAX_INDICES_PER_ATTESTATION:
        raise Exception("Error during verifying max number of indices")
    ...
```

djrtwo: I'm pro changing it to a `validate` form with `raise`. It is self-documenting and translates naturally for implementers (also will make my testing efforts easier).

I _did_ solicit the feedback of a couple of implementers on the tiered vs single boolean implementation and both responded as tiered being the preferable for reading, understanding, and translating the spec. I think `validate` w/ `raise` will just make that better.

JustinDrake: I've renamed to `validate` and gone ahead with `assert` everywhere. We get both readability and debugability 🎉 (I'm avoiding `raise` because we already have `assert` which works fine.)

djrtwo:
aw, you skipped the `raise`. Understandable I suppose as it's not anywhere else in the spec.

This looks good to me.

djrtwo:
> It is artificial in process_attestation where validators are incentivised to maximise new attestations

A bit torn on this. Can a validator then include any empty attestations in empty slots and unnecessarily bloat block size? Seems like it could become an unintended source of graffiti

JustinDrake:
> Seems like it could become an unintended source of graffiti

I've been thinking about this. It turns out there are other cheap ways to generate graffiti. For example, a block proposed can make `Transfer`s where `sender == recipient` and encode data in the amounts. Because graffiti cannot be avoided the Zen approach is to embrace them. The good news is that all operations are capped, so at a minimum any source of graffiti is bounded.

> and unnecessarily bloat block size?

Bloating is negligible, or even zero since the default rational behaviour is probably to max out the number of attestations in a block to maximise revenue. And using `Attestation`s for graffiti means forgoing on the opportunity to make fees, so there's an opportunity cost.

Issue #1030, opened 2019-05-02
Spec tests can't be easily parsed in javascript


wemeetagain:
The yaml spec tests generated here include very large integers that get silently changed/corrupted by `js-yaml`, the standard-ish js yaml parser.

Some possible solutions:
- wrap integers in quotes
- create a new randomization mode that caps the max uint
- make me write/extend a yaml parser -_-

wemeetagain:
Fixed by extending the yaml parser, no spec test changes required

PR #1031, opened 2019-05-02
Add randomization mode suitable for js numbers

wemeetagain:
See #1030

Add another randomization mode to create additional test cases which are more friendly to js consumption.

### DIFF
--- a/test_libs/pyspec/eth2spec/debug/random_value.py
+++ b/test_libs/pyspec/eth2spec/debug/random_value.py
@@ -7,7 +7,7 @@ UINT_SIZES = [8, 16, 32, 64, 128, 256]
 
 basic_types = ["uint%d" % v for v in UINT_SIZES] + ['bool', 'byte']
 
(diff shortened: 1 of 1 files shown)


protolambda:
Imho defining a lower maximum is not the correct way to test support for edge-cases. If it's 64 bits, you should test for 64 1s. Wrapping the number in quotes is doable, but technically it's perfect valid yaml. Oh javascript...

Anyone else problems with yaml parses? Should we quote them?

wemeetagain:
This PR is mostly for the lols.
I know its valid yaml, its just practically not favorable with our current tooling
(defacto standard yaml parsing library not compliant w the yaml spec == sad javascript :clown_face: )
We can/do parse arbitrary length strings into numbers, no problem.
We could also do our own pre-processing on the yaml files and wrap numbers in quotes ourselves, but ofc it would be preferable to have something nice out of the box.

PR #1035, opened 2019-05-03
Allow multiple bit challenges, and recover withdrawability

### LINKED ISSUE #864: Custody game checklist

vbuterin:
Resolves #864 items 4, 7, 14


specs/core/1_custody-game.md
```diff
@@ -688,6 +692,18 @@ Append this to `process_final_updates(state)`:
 ```python
     # Clean up exposed RANDAO key reveals
     state.exposed_derived_secrets[current_epoch % EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS] = []
+    # Reset withdrawable epochs if challenge records are empty
+        eligible = True
+        for records in (state.custody_chunk_challenge_records, state.bit_challenge_records):
+            for filter_func in (lambda rec: rec.challenger_index == index, lambda rec: rec.responder_index == index):
+                if len(list(filter(filter_func, records))) > 0:
+                    eligible = False
```
hwwhww: early `break`
```suggestion
                    eligible = False
                    break
```

specs/core/1_custody-game.md
```diff
@@ -626,6 +627,9 @@ def process_bit_challenge_response(state: BeaconState,
                                    challenge: CustodyBitChallengeRecord) -> None:
     # Verify chunk index
     assert response.chunk_index < challenge.chunk_count
+    # Verify responder has not been slashed
+    responder = state.validator_registry[record.responder_index]
```
hwwhww:
```suggestion
    responder = state.validator_registry[challenge.responder_index]
```

specs/core/1_custody-game.md
```diff
@@ -688,6 +692,18 @@ Append this to `process_final_updates(state)`:
...
+                if len(list(filter(filter_func, records))) > 0:
+                    eligible = False
+        if eligible:
+            if validator.exit_epoch == FAR_FUTURE_EPOCH:
+                validator.withdrawable_epoch = FAR_FUTURE_EPOCH
+            else:
+                validator.withdrawable_epoch = validator.exit_epoch + MIN_VALIDATOR_WITHDRAWABILITY_DELAY
 ```
 
 In `process_penalties_and_exits`, change the definition of `eligible` to the following (note that it is not a pure function because `state` is declared in the surrounding scope):
```
hwwhww: Both `process_penalties_and_exits` and `eligible` no longer exist in the phase 0 spec after #850.

specs/core/1_custody-game.md
vbuterin: I don't think there's much point in the break; it'll only escape the inner of the two loops anyway, and not sure it adds readability.

specs/core/1_custody-game.md
```diff
@@ -688,6 +692,18 @@ Append this to `process_final_updates(state)`:
...
+        eligible = True
+        for records in (state.custody_chunk_challenge_records, state.bit_challenge_records):
+            for filter_func in (lambda rec: rec.challenger_index == index, lambda rec: rec.responder_index == index):
+                if len(list(filter(filter_func, records))) > 0:
+                    eligible = False
+        if eligible:
+            if validator.exit_epoch == FAR_FUTURE_EPOCH:
+                validator.withdrawable_epoch = FAR_FUTURE_EPOCH
+            else:
+                validator.withdrawable_epoch = validator.exit_epoch + MIN_VALIDATOR_WITHDRAWABILITY_DELAY
```
hwwhww: Alternatively:
```python
records = state.custody_chunk_challenge_records + state.bit_challenge_records
validator_indices_in_records = set([record.challenger_index for record in records] + [record.responder_index for record in records])
    if index not in validator_indices_in_records:
        if validator.exit_epoch == FAR_FUTURE_EPOCH:
            validator.withdrawable_epoch = FAR_FUTURE_EPOCH
        else:
            validator.withdrawable_epoch = validator.exit_epoch + MIN_VALIDATOR_WITHDRAWABILITY_DELAY
```

vbuterin: I like this more.

specs/core/1_custody-game.md
```diff
@@ -688,6 +682,17 @@ Append this to `process_final_updates(state)`:
...
     state.exposed_derived_secrets[current_epoch % EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS] = []
+    # Reset withdrawable epochs if challenge records are empty
+    records = state.custody_chunk_challenge_records + state.bit_challenge_records
+    validator_indices_in_records = set(
+        [record.challenger_index for record in records] + [record.responder_index for record in records]
+    )
+    for index, validator in enumerate(state.validator_registry):
+        if index not in validator_indices_in_records:
+            if validator.exit_epoch == FAR_FUTURE_EPOCH:
+                validator.withdrawable_epoch = FAR_FUTURE_EPOCH
```
hwwhww: What's the case of that `validator.exit_epoch == FAR_FUTURE_EPOCH && validator.withdrawable_epoch != FAR_FUTURE_EPOCH` happens?

specs/core/1_custody-game.md
hwwhww: discuss in #1066

specs/core/1_custody-game.md
djrtwo: There is no place in the current code where this can happen.

`withdrawable_epoch` is only changed to `!= FAR_FUTURE_EPOCH` in `slash_validator` and `initiate_validator_exit`. Both of these instances also adjust `exit_epoch` to be `!= FAR_FUTURE_EPOCH`
The only place where `exit_epoch` is set is in [`initiate_validator_exit`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#initiate_validator_exit) and both `exit_epoch` and `withdrawable_epoch` are changed at the same time.

hwwhww: @djrtwo I feel the same way. :)

This if-else statement can be simplified to:

```python
if validator.exit_epoch != FAR_FUTURE_EPOCH:
    validator.withdrawable_epoch = validator.exit_epoch + MIN_VALIDATOR_WITHDRAWABILITY_DELAY
```

But, maybe worth to keep it due to precaution? /cc @vbuterin

djrtwo: I think you could actually do the following to very specifically fix the case we care about.

```python
if validator.exit_epoch != FAR_FUTURE_EPOCH and validator.withdrawable_epoch == FAR_FUTURE_EPOCH:
    validator.withdrawable_epoch = validator.exit_epoch + MIN_VALIDATOR_WITHDRAWABILITY_DELAY
```

PR #1044, opened 2019-05-05
Crosslink in AttestationData

JustinDrake:
Substantive change:

* Batch fields in `AttestationData` into a `Crosslink`
    * better encapsulation and reuse of `Crosslink` type
    * cleaner presentation (remove an awkward helper function, simpler code, 15 lines removed).

Cosmetic changes:

* rename `MAX_CROSSLINK_EPOCHS` to `MAX_EPOCHS_PER_CROSSLINK`
* avoid long lines (over 120 characters) in `process_attestation`


specs/core/0_beacon-chain.md
```diff
@@ -1705,30 +1695,30 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
...
         inclusion_delay=state.slot - attestation_slot,
         proposer_index=get_beacon_proposer_index(state),
     )
+
+    assert data.target_epoch in (get_previous_epoch(state), get_current_epoch(state))
+    if data.target_epoch == get_previous_epoch(state):
+        ffg_data = (state.previous_justified_epoch, state.previous_justified_root, get_previous_epoch(state))
+        previous_crosslink = state.previous_crosslinks[data.crosslink.shard]
+        state.previous_epoch_attestations.append(pending_attestation)
     if data.target_epoch == get_current_epoch(state):
```
hwwhww: `elif` would be more accurate.
```suggestion
    elif data.target_epoch == get_current_epoch(state):
```

specs/core/0_beacon-chain.md
```diff
@@ -1296,28 +1296,18 @@ def get_attesting_balance(state: BeaconState, attestations: List[PendingAttestat
...
-```
-
 ```python
 def get_winning_crosslink_and_attesting_indices(state: BeaconState, epoch: Epoch, shard: Shard) -> Tuple[Crosslink, List[ValidatorIndex]]:
-    shard_attestations = [a for a in get_matching_source_attestations(state, epoch) if a.data.shard == shard]
-    shard_crosslinks = [get_crosslink_from_attestation_data(state, a.data) for a in shard_attestations]
+    shard_attestations = [a for a in get_matching_source_attestations(state, epoch) if a.data.crosslink.shard == shard]
     candidate_crosslinks = [
-        c for c in shard_crosslinks
+        c for c in [a.data.crosslink for a in shard_attestations]
```
djrtwo: This double selection is strange and will certainly cause confusion to readers. I understand you did it so you can select crosslinks and then assign to a short var to a more manageable var to use in the boolean, but it's pretty non-obvious what is going on or why.

I suggest still doing the selection in a line above `crosslinks = [a.data.crosslink for a in shard_attestations]`
or make the if statement multiline like the following

```python
    candidate_crosslinks = [
        a.data.crosslink for a in shard_attestations
        if (
            hash_tree_root(state.current_crosslinks[shard]) in
            (a.data.crosslink.previous_crosslink_root, hash_tree_root(a.data.crosslink)
        )
```

specs/core/0_beacon-chain.md
```diff
@@ -1705,30 +1695,30 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
...
         state.current_epoch_attestations.append(pending_attestation)
-    else:
-        state.previous_epoch_attestations.append(pending_attestation)
+
+    # Check FFG data, crosslink data, and signature
+    assert ffg_data == (data.source_epoch, data.source_root, data.target_epoch)
+    assert data.crosslink.epoch == min(data.target_epoch, previous_crosslink.epoch + MAX_EPOCHS_PER_CROSSLINK)
+    assert data.crosslink.previous_crosslink_root == hash_tree_root(previous_crosslink)
+    assert data.crosslink.crosslink_data_root == ZERO_HASH  # [to be removed in phase 1]
+    assert verify_indexed_attestation(state, convert_to_indexed(state, attestation))
```
hwwhww: Python speaking, since `state` object is mutable, if the later assertion raises exception, the `state` (`state.previous_epoch_attestations`/`state.current_epoch_attestations`) update has been applied.

It may be a problem for pyspec when we have more complicated test cases, e.g., apply two blocks, the first one is invalid and the second one is valid. IMO putting all state updates to the last is the safest.

specs/core/0_beacon-chain.md
JustinDrake: The logic is equivalent given `assert data.target_epoch in (get_previous_epoch(state), get_current_epoch(state))`. I prefer `if` because it's more symmetric :)

specs/core/0_beacon-chain.md
djrtwo: I felt similarly but then looked at the rest of the doc and realized we aren't strictly doing that in all block operations

specs/core/0_beacon-chain.md
djrtwo: This is not strictly equivalent. At the genesis epoch, `get_previous_epoch == get_current_epoch` and so you would add the `pending_attestation` to both `previous_epoch_attestations` and `current_epoch_attestations`

hwwhww: Oooh right I forgot `get_previous_epoch == get_current_epoch` case! And... if so, this PR is a bugfix, right?

djrtwo: No, the previous if/else handled it fine

specs/core/0_beacon-chain.md
hwwhww: That seems anti-pattern sin. 🙈

specs/core/0_beacon-chain.md
hwwhww: @djrtwo Ah sorry I confused myself. (😵)

Do you mean we should only update `current_epoch_attestations` at the genesis epoch?
```python
if data.target_epoch == get_current_epoch(state):
    ....
elif if data.target_epoch == get_previous_epoch(state):
    ....
```

hwwhww: Is it okay to introduce `raise`?
```python
if data.target_epoch == get_current_epoch(state):
    ....
elif data.target_epoch == get_previous_epoch(state):
    ....
else:
    raise(Exception("Invalid target_epoch"))
```

specs/core/0_beacon-chain.md
JustinDrake: Addressed, and cleaned up `get_winning_crosslink_and_attesting_indices` :)

specs/core/0_beacon-chain.md
JustinDrake: Fixed genesis bug :)

specs/core/0_beacon-chain.md
JustinDrake: @hwwhww Do you want to open an issue to discuss (and potentially fix) outside of this PR?

hwwhww: @JustinDrake Yes, will do! :)

JustinDrake:
> Since `crosslink_data_root` only be used in `Crosslink`, what do you think about renaming it to `data_root`?

`data_root` is also used in the custody bit, but I think that's the right thing to do regardless :) I also think we should rename `previous_crosslink_root` to `parent_crosslink_root`. The reason is that there is a naming clash now with `state.previous_crosslink_roots`. For consistency this means reverting back to `parent_block_root` (as opposed to `previous_block_root`).

specs/core/0_beacon-chain.md
hwwhww: #1059

djrtwo:
yeah, let me do one more pass

djrtwo:
I cleaned up the validator guide (and fixed a couple of things there). Would appreciated a once over before we merge

JustinDrake:
Side note: making the validator guide executable may be useful, if anything to help with testing.

djrtwo:
Yeah, I've been thinking it would help keep things in sync for sure

PR #1052, opened 2019-05-05
Spectest decorator

protolambda:
Since the pytests share a lot of "edge case" setups with test-generation, I'm looking to unify these tests with the generator code, i.e. not duplicate testing work.

To achieve this, a few things have to be changed:
- generators don't know any "fixtures" like pytests do. However, we can implement an elegant decorator utility to provide test functions with things like a genesis-state setup. I included an implementation for this in `context.py`/`utils.py`: `with_args` (generalized) and `with_state` (built on top of generalized version, provides fresh genesis state).
- existing tests have to be ported over. Luckily, the format introduced here is similar, yet **shorter** (`yield from` is a blessing) and **faster** (less, if not none, deep-copies)

The new format:

We have a new decorator for tests: `@spectest(description='very informative')` (description is optional)
This decorator enables you to:
- call the test-function with an optional `generator_mode` param, when `True` it makes the test function collect yielded data into a dict, incl. description, and return it.
- ignore yields completely, and execute the full function like a normal test, when not in  `generator_mode` (default).

Example test:
```python
@spectest()
def test_foobar_thing():
  state = ... (preparation)
  state.special_thing = 123 # edge case set up
  yield 'pre', state
  operation = ...
  yield 'op', operation
  process_thing(state, operation)
  assert state.special_thing > 300 # assert changes like normal test
  yield 'post', state
```

Note that in some cases, some of it can be generalized with a local helper function, and then you just call `yield from run_processing_foobar(state, operation)` or similar. See POC attestations test update.

And, you can call the test to create a test-case vector:
```python
test_case = test_foobar_thing(generator_mode=True)

# returns:
{
  'description': 'foobar_thing'
  'pre': ...,
  'op': ...,
  'post': ...
}
```

Note that values yielded in generator mode are encoded before continuing in the function, so there's no deep-copy necessary: after encoding the original containers may mutate.
This also speeds up the normal test-runs, as values are not collected/encoded, nor deep-copied, so you get the minimal amount of object instances / memory allocation / etc.

### Some syntax sugar

99% of the pytests look like this:
```python
@with_state
@spectest()
def test_my_thing(state):
   ...
```
To make this shorter, we define a `@spec_state_test` that composes these two generators. It's literally a state-spec-test!
```python
from tests.context import spec_state_test

@spec_state_test
def test_my_thing(state):
   ...
```

For anyone interested in previous iterations of the edge-case unification idea, here they are: https://notes.ethereum.org/5N7MrvOESrKK3AIHYYvYww#
The important pivot made here is that we do not separate the edge-cases from testing, nor do we reference the processing function, and make the decorator run the behavior. This enables more freedom in testing, yet it is powerful enough to create test-vectors cleanly from the existing pytest work.

Next steps:
- update remaining tests
- make block-operation test-generator and epoch transition test-generator call these tests, and output vectors into spec-tests repo.


djrtwo:
ported all remaining tests.
Took a shortcut on finality tests. Did a bit of cleaning but kept the one `deepcopy` where it was to make comparison between pre and post state each epoch easier.

djrtwo:
whoops on the extra array nesting. nice catch

protolambda:
Tests are running, and I got block-operation generators to work based on the same tests :tada:
I think we can merge this in after review, and then we update the remaining part of test-generation later to call the epoch-transition tests, sanity tests etc. I don't want this to get too much out of sync with the `dev` tests to avoid large merge conflicts / people writing outdated style tests.

test_generators/operations/main.py
```diff
@@ -1,9 +1,31 @@
...
     gen_runner.run_generator("operations", [
-        mini_deposits_suite,
-        full_deposits_suite
+        create_suite('attestation',       'minimal', lambda: generate_from_tests(test_process_attestation)),
+        create_suite('attestation',       'mainnet', lambda: generate_from_tests(test_process_attestation)),
+        create_suite('attester_slashing', 'minimal', lambda: generate_from_tests(test_process_attester_slashing)),
+        create_suite('attester_slashing', 'mainnet', lambda: generate_from_tests(test_process_attester_slashing)),
+        create_suite('block_header',      'minimal', lambda: generate_from_tests(test_process_block_header)),
+        create_suite('block_header',      'mainnet', lambda: generate_from_tests(test_process_block_header)),
+        create_suite('deposits',          'minimal', lambda: generate_from_tests(test_process_deposit)),
```
djrtwo: use singular `'deposit'` here  to conform to others

```suggestion
        create_suite('deposit',          'minimal', lambda: generate_from_tests(test_process_deposit)),
```

test_generators/operations/main.py
```diff
@@ -1,9 +1,31 @@
...
+        create_suite('deposits',          'mainnet', lambda: generate_from_tests(test_process_deposit)),
```
djrtwo:
```suggestion
        create_suite('deposit',          'mainnet', lambda: generate_from_tests(test_process_deposit)),
```

test_generators/operations/suite_creator.py
```diff
@@ -0,0 +1,39 @@
...
+
+def create_suite(operation_name: str, config_name: str, get_cases: Callable[[], Iterable[gen_typing.TestCase]])\
+        -> Callable[[str], gen_typing.TestSuiteOutput]:
+    def suite_definition(configs_path: str) -> gen_typing.TestSuiteOutput:
+        presets = loader.load_presets(configs_path, config_name)
+        spec.apply_constants_preset(presets)
+
+        return ("%s_%s" % (operation_name, config_name), operation_name, gen_suite.render_suite(
+            title="%s operation" % operation_name,
+            summary="Test suite for deposit type operation processing",
```
djrtwo: This summary needs to be more generic

```suggestion
            summary="Test suite for %s type operation processing" % operation_name,
```

test_generators/operations/main.py
```diff
@@ -1,9 +1,31 @@
...
+    test_process_voluntary_exit
+)
+
 from gen_base import gen_runner
 
-from deposits import mini_deposits_suite, full_deposits_suite
+from suite_creator import generate_from_tests, create_suite
 
 if __name__ == "__main__":
     gen_runner.run_generator("operations", [
```
djrtwo: The pattern could drop the explicit create_suite per line and instead pass the creator in as a separate argument to run_generator.

In your assessment is the create_suite pattern fixed enough to do this to drop the verbosity a bit here? or do you think the gen_runner should not be concerned about how suites are created?

protolambda: I prefer a little bit of verbosity here, as it gives us a lot of flexibility: we can just add tests to any particular type by concatenating more to the list returned by `generate_from_tests`. Or if necessary, modify the returned list, to e.g. exclude a given test.

test_generators/operations/suite_creator.py
protolambda: Good spot there, missed it :+1:

protolambda:
Cleaned up PR suggestion commits

protolambda:
3189cf00797e5e9b76024428b32996fe5bc237b9 includes the new edge cases presented in #1010, in a more concise form, as part of the unified spec-tests. Thanks @jannikluhn. Hope to write many more spectests in this format with the other testers, sorry about the delay getting the tests in because of this PR.

protolambda:
b4be220a48ecc61853ac54aae01f9ed471324ba9 includes new attestation edge cases presented in #999. Thanks @zilm13. See above comment, the format is improving, and spec testing shares effort with test-vector generation now. Sorry about the delay of getting test cases in, hope it's worth it :)

djrtwo:
merging.
testing work continued on separate branches such as [`spectest-deco` PR 1052](https://github.com/ethereum/eth2.0-specs/pull/1052)

PR #1055, opened 2019-05-06
Fix deposit domain

protolambda:
Forks are ignored for deposit validity: deposits are always accepted (if coming from the correct contract(s)).

Credits to @cemozerr for pointing out that the Validator and Beacon spec were inconsistent here.
Credits to @schroedingerscode for pointing out that the domain format changed some time ago (fork version and domain type swapped)

After looking into the inconsistency discussion on gitter, it occurred to me that deposits are not bound to an eth 2.0 fork, but to the eth 1.0 contract, and maybe shouldn't be signed. Vaguely remembering previous discussions about singatures with @JustinDrake in Sydney. This was confirmed to stay zeroed across forks (or previous `GENESIS_FORK_VERSION`), and here's the fix.

Note: I swapped back fork-version and domain type, as the [change here](https://github.com/ethereum/eth2.0-specs/commit/16123685afd5c175cb1df35f46a822fa87ba30ee#diff-51a43328a58414e132a744f3771f018cR1165) looks unintentional, and signing/verifying with just the domain-type is more straight-forward this way.


specs/core/0_beacon-chain.md
```diff
@@ -1766,7 +1767,8 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
     validator_pubkeys = [v.pubkey for v in state.validator_registry]
     if pubkey not in validator_pubkeys:
         # Verify the deposit signature (proof of possession)
-        if not bls_verify(pubkey, signing_root(deposit.data), deposit.data.signature, get_domain(state, DOMAIN_DEPOSIT)):
+        # Note: deposits are valid regardless of fork version, hence the type is passed directly as domain.
+        if not bls_verify(pubkey, signing_root(deposit.data), deposit.data.signature, DOMAIN_DEPOSIT):
```
JustinDrake: Type mismatch

specs/core/0_beacon-chain.md
```diff
@@ -966,7 +966,8 @@ def get_domain(state: BeaconState,
     """
     epoch = get_current_epoch(state) if message_epoch is None else message_epoch
     fork_version = state.fork.previous_version if epoch < state.fork.epoch else state.fork.current_version
-    return bytes_to_int(fork_version + int_to_bytes4(domain_type))
+    # fork version is on the big-endian side: when signing using only the type (e.g. deposits), the type can be passed directly.
```
JustinDrake: Design rationales will go in another document. Suggest removing this comment. (As a side note, comments start with a capital.)

protolambda: Should have been `fork_version`, hence no capital. I'll remove it.

specs/core/0_beacon-chain.md
protolambda: It's a verbosity thing. In python it's just an integer. And it aligns to the lower part. And there's no way to explicitly cast from uint32 to uint64. Open for suggestions.

JustinDrake:
The current solution may be insufficient in the context of multiple deposit contracts. One approach is to assign each deposit contract a `version` and for deposits replace `state.fork` with `deposit_contract.version` in the calculation of domains.

djrtwo:
I'm inclined to assume we have one deposit contract and to handle the exceptional case _if_ we happen to change deposit contracts. Note that the rest of the eth1data machinery is not sufficient for multiple deposit contracts either

protolambda:
> Note that the rest of the eth1data machinery is not sufficient for multiple deposit contracts either

Deposit index is very much a singleton mechanism indeed.

Maybe it can be solved on the eth1 side, i.e. make the deposit contract flexible for permissioned integration of secondary contracts (for whatever reason you need a second contract to deal with deposits). However, I think the only reason for another contract is some kind of bug / unfortunate event, and in that case I suppose we fork the beacon chain for a new contract.

So what's the way forward? Create a helper function to more explicitly create a correct signature domain for deposits? I think this would be neat to just do with typing, if we had them...

A helper would be something like `bls_domain(domain_type, fork_version)`, call `bls_domain(DOMAIN_TYPE_DEPOSIT, b'\x00\x00\x00\x00')`

djrtwo:
so `get_domain` calls `bls_domain`, and when constructing a deposit domain, one calls `bls_domain` directly?

seems reasonable.

djrtwo:
Made the adjustment of pulling out `bls_domain` into separate function and calling it directly when handling deposit domains.

Ready for review @protolambda @JustinDrake

protolambda:
Defaulting to zero lets us avoid another pseudo-configurable constant. Generally also nicer to just read what the signature is about, i.e. the domain, and no more. And `bls_domain` nicely black-boxes the actual bytes being formatted into a BLS domain.

djrtwo:
lgtm

specs/core/0_beacon-chain.md
```diff
@@ -629,6 +630,16 @@ The `hash` function is SHA256.
 
 `def signing_root(object: SSZContainer) -> Bytes32` is a function defined in the [SimpleSerialize spec](../simple-serialize.md#self-signed-containers) to compute signing messages.
 
+### `bls_domain`
+
+```python
+def bls_domain(domain_type: int, fork_version=b'\x00\x00\x00\x00') -> int:
```
hwwhww: 1. use `bytes` for type hinting
```suggestion
def bls_domain(domain_type: int, fork_version: bytes=b'\x00\x00\x00\x00') -> int:
```

2. or define `ForkVersion` or `Bytes4` in Custom types section
```
| `ForkVersion` | `bytes4` | 4 bytes of binary data |
```
and then
```suggestion
def bls_domain(domain_type: int, fork_version: ForkVersion=b'\x00\x00\x00\x00') -> int:
```

djrtwo: rolled with `bytes` for now

PR #1060, opened 2019-05-07
Cleanup int_to_bytes

JustinDrake:
Reduce the amount of external magic.

### DIFF
--- a/scripts/phase0/build_spec.py
+++ b/scripts/phase0/build_spec.py
@@ -18,9 +18,6 @@ from eth2spec.utils.bls_stub import *
 # stub, will get overwritten by real var
 SLOTS_PER_EPOCH = 64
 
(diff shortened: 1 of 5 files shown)


scripts/phase0/build_spec.py
```diff
@@ -15,15 +15,11 @@ def build_phase0_spec(sourcefile, outfile):
...
-""")
-    for i in (1, 2, 3, 4, 8, 32, 48, 96):
-        code_lines.append("def int_to_bytes%d(x): return x.to_bytes(%d, 'little')" % (i, i))
-
-    code_lines.append("""
-
 # stub, will get overwritten by real var
 SLOTS_PER_EPOCH = 64
 
+def int_to_bytes(integer: int, length: int) -> bytes:
```
djrtwo: This no longer needs to be here, right?

PR #1061, opened 2019-05-07
Make phase 1 spec executable

dankrad:


CarlBeek:
So I did some work on this myself, but it seems like I was targeting lower-level functionality. My approach was to offer insert and replace functionality. The idea was to build the two pyfiles for phase0 and phase1 and then combine them.

It works in two stages: first all functions and global scope objects (with the same name) that exist in the newer file overwrite instances in the older file. Thereafter, instances of code that are wrapped in a certain tag in the new file are inserted into the old file where instances of the label are found.

In the interests of not duplicating work between us, I've condensed my code into a single file if you'd like to include/implement parts of it in your work. You can save yourself the regex fun this way: https://gist.github.com/CarlBeek/5e86bef2fbfa0c9e5668382bb1afe918

From a design standpoint, the way I see it, there are two major options: either add tags and labels to the code such that a single file is ultimately built for each phase or monkey-patch the previous phase into the current one for testing. The reason to prefer the tag-and-label option is that it is more extensible in the sense that by simply adding the tags to a given phase it is immediately upgradeable and executable. In addition, there will be no code duplication between the specs and the monkey patch file (which is a possible source for error). The existence of tags in the previous phase's file also serves to indicate which functions are expected to change in the future.

The argument for preferring the monkey patch option is that it keeps the specs cleaner and hides much of the ugliness in the monkey-patch file. This also means that the phase 0 specs are unaffected by changes to the phase 1 spec as no new tags are inserted.

Ultimately, I don't think this is a black and white decision so I am interested to hear what your thoughts are.

protolambda:
FYI: you can mark WIP-work as a 'draft' PR. :+1:

dankrad:
OK, so tests from phase 0 + EarlyDerivedSecretReveal tests now working. Currently I changed the imports in the tests to use the phase1 version of the spec, couldn't find an elegant way to have a "switch" for the test target in pytest, but given #1052 I will not invest time in that for now.

Before we proceed, what are people's opinion on using the Monkeypatch approach used in this PR vs. the source-level "patch" approach outlined by @CarlBeek ?

djrtwo:
I'm leaning toward the monkey patch but would like to consider how it might look in light of #1018 merged. I think it could make sense to just spell out the full state transition again in the phase 1 specs with the additional functionality, but it would be a little verbose and might open up some issues in maintaining each in separate places

specs/core/1_custody-game.md
```diff
@@ -468,13 +479,13 @@ def process_chunk_challenge(state: BeaconState,
             record.chunk_index != challenge.chunk_index
         )
     # Verify depth
-    depth = math.log2(next_power_of_two(get_custody_chunk_count(challenge.attestation.data.crosslink)))
+    depth = math.ceil(math.log2(get_custody_chunk_count(challenge.attestation.data.crosslink)))
```
CarlBeek: I am not sure this is a particularly nice way of doing things, but I couldn't find this function (I thought it was in SSZ somewhere) and because we already have `math` imported, I did this.

specs/core/1_custody-game.md
```diff
@@ -624,20 +637,21 @@ def process_bit_challenge_response(state: BeaconState,
     assert verify_merkle_branch(
         leaf=hash_tree_root(response.chunk),
         branch=response.data_branch,
-        depth=math.log2(next_power_of_two(challenge.chunk_count)),
+        depth=math.ceil(math.log2(challenge.chunk_count)),
```
CarlBeek: Again, a hack due to the lack of `next_power_of_two()`. See my above comment.

specs/core/1_custody-game.md
```diff
@@ -624,20 +637,21 @@ def process_bit_challenge_response(state: BeaconState,
...
         index=response.chunk_index,
         root=challenge.data_root,
     )
     # Verify the chunk bit leaf matches the challenge data
     assert verify_merkle_branch(
         leaf=response.chunk_bits_leaf,
         branch=response.chunk_bits_branch,
-        depth=math.log2(next_power_of_two(challenge.chunk_count) // 256),
+        depth=math.ceil(math.log2(challenge.chunk_count)) >> 8,
```
CarlBeek: I did this in a rush, can someone check my logic here please.

specs/core/1_custody-game.md
```diff
@@ -212,44 +212,50 @@ Add the following fields to the end of the specified container objects. Fields w
...
     'custody_responses': [CustodyResponse],
     'custody_key_reveals': [CustodyKeyReveal],
     'early_derived_secret_reveals': [EarlyDerivedSecretReveal],
+}
 ```
 
 ## Helpers
 
-### `typeof`
+### `type_of`
```
CarlBeek: Do we want to call this `typeof` or the more pythonic `type-of`?

specs/core/1_custody-game.md
```diff
@@ -126,7 +126,7 @@ This document details the beacon chain additions and changes in Phase 1 of Ether
     'attestation': Attestation,
     'challenger_index': ValidatorIndex,
     'responder_key': BLSSignature,
-    'chunk_bits': Bitfield,
+    'chunk_bits': "bytes",
```
```suggestion
    'chunk_bits': 'bytes',
```
Side question: Do we want a bitfield custom type?

specs/core/1_custody-game.md
```diff
@@ -155,7 +155,7 @@ This document details the beacon chain additions and changes in Phase 1 of Ether
     'inclusion_epoch': Epoch,
     'data_root': Hash,
     'chunk_count': 'uint64',
-    'chunk_bits_merkle_root': Hash,
+    'chunk_bits_merkle_root': 'bytes32',
```
CarlBeek: In my head we got rid of the `Hash` type, am I correct in this?

test_libs/pyspec/eth2spec/utils/minimal_ssz.py
```diff
@@ -6,6 +6,8 @@
 BYTES_PER_LENGTH_OFFSET = 4
 ZERO_CHUNK = b'\x00' * BYTES_PER_CHUNK
 
+cached_typedefs = {}
```
CarlBeek:
```suggestion
```

test_libs/pyspec/tests/phase0/conftest.py
```diff
@@ -3,34 +3,31 @@
...
 
-
 @pytest.fixture(autouse=True)
 def config(request):
     config_name = request.config.getoption("--config")
     presets = loader.load_presets('../../configs/', config_name)
     spec.apply_constants_preset(presets)
-
+    helpers.spec = spec
+    request.function.__globals__['spec'] = spec
```
CarlBeek: So this is my hack to get the spec version available to the tests to be the correct version independent of whether it is the phase 0 or phase 1 spec. It is not the prettiest, but it is the best solution I could come up with. It allows phase 0 tests to be imported into phase 1 and the tests to be re-run against the phase1 spec with the phase 1 spec objects in the test `helpers.py`. If someone finds a better solution, I am keen to hear about it.

test_libs/pyspec/tests/phase1/block_processing_phase1/test_phase0_block_processing.py
```diff
@@ -0,0 +1,7 @@
+from tests.phase0.block_processing.test_process_attestation import *
```
CarlBeek: This is how all tests from Phase 0 are imported into Phase 1. I do not like `*` imports (and they are not PEP8 compliant), but these files seem ridiculous when they are imported function-by-function.

scripts/build_spec.py
```diff
@@ -0,0 +1,220 @@
...
+If building phase 0:
+    1st argument is input spec.md
+    2nd argument is output spec.py
+
+If building phase 1:
+    1st argument is input spec_phase0.md
+    2nd argument is input spec_phase1.md
+    3rd argument is output spec.py
+'''
+    parser = ArgumentParser(description=description)
```
CarlBeek: Should we be bothering with this command line interface?

test_libs/pyspec/eth2spec/phase1/state_transition.py
```diff
@@ -0,0 +1,70 @@
+from . import spec
+
+from .spec import (
+    BeaconState,
+    BeaconBlock,
+    Slot,
+)
+
+from eth2spec.phase0.state_transition import (
```
CarlBeek: This whole file should be removed. (Mirroring what was done in #1018) It is actually unused as it stands and I forgot to remove it in my last commit. I will port whatever minimal functionality remains in it to other places tomorrow.

specs/core/1_custody-game.md
dankrad: This will be implemented in #1096

CarlBeek: Yup! I would like to keep this PR this way for now though as otherwise the tests fail (and I am rapidly reaching my failing-test exposure saturation point).

test_libs/pyspec/eth2spec/utils/minimal_ssz.py
dankrad: I think you don't need cached_typedefs anymore as you got rid of SSZTypeExtension

test_libs/pyspec/eth2spec/phase1/state_transition.py
dankrad: Yes, I would actually suggest creating a phase 1 state transition .md file, that "integrates" the different parts of phase 1 into the state transition.

specs/core/1_custody-game.md
dankrad: I just found a nice way to do this in python without relying on floating point arithmetic: Use the bit_length method of an integer:
```python
def ceillog2(x):
    return x.bit_length()
```

I suggest we use this as it would not add dependencies on FP math that can potentially be flaky in extreme conditions?

specs/core/1_custody-game.md
dankrad: Looks correct. See above for my comment how to remove the FP arithmetic

specs/core/1_custody-game.md
dankrad: `type_of` is easier to read. Could we also just use `type` and not need this definition at all? I guess this implicitly depends on how SSZ is implemented in python then?

test_libs/pyspec/tests/phase0/conftest.py
djrtwo: The `helpers.py` hack here isn't really sustainable. In #1102, we are dividing out helpers into better organized files.

I'll look into this a bit more tomorrow

specs/core/1_custody-game.md
dankrad: I think so. Although there is a suggestion to go back to using explicit aliases in #667, which I support, for now it's just "bytes32".

test_libs/pyspec/tests/phase1/block_processing_phase1/test_phase0_block_processing.py
hwwhww: What's the need of importing phase 0 tests in `tests/phase1`? I think all the phase 0 tests would be tested with `make test` command?

For your question, we can skip it by:
```suggestion
from tests.phase0.block_processing. test_process_attestation import *  # noqa: F401, F403
```

test_libs/pyspec/tests/phase0/conftest.py
CarlBeek: To be clear, is your objection to the use of a monolithic `helpers.py`, to the particular method by which I inject the correct version of `spec.py` into `helpers.py` or to the use of `request.function.__globals__['spec'] = spec` and `request.function.__globals__['helpers'] = helpers` to provide the spec and helpers to the tests?

I agree with the direction taken in #1102 w.r.t. breaking up `helpers.py`. I would prefer an alternative to my hack use of pytest globals for spec injection, but I am unable to come up with one myself. I think it is very important that the spec version used by the helpers is the same as the spec version being tested.

test_libs/pyspec/tests/phase1/block_processing_phase1/test_phase0_block_processing.py
CarlBeek: The reason for importing the tests from `phase0/*` into `phase1/*` is that it allows for the phase 0 tests to be run against the phase 1 spec. Unless specified otherwise by a pytest decorator, phase 0 tests will be run against phase 1. This prevent us from breaking things in phase 0 by changing phase 1. My new way of testing builds a completely new `phase1/spec.py` that is entirely independent of `phase1/spec.py`

test_libs/pyspec/tests/phase0/conftest.py
dankrad: I actually think this hack is ok, if there is no nice way to provide this in pytest :) I actually tried to do something like this before but failed because I was still using normal "import"s

Makefile
```diff
@@ -33,29 +38,40 @@ gen_yaml_tests: $(PY_SPEC_ALL_TARGETS) $(YAML_TEST_TARGETS)
...
+	python -m pytest --junitxml=test-reports/eth2spec/test_results_phase0.xml tests/phase1
 
 install_lint:
-	cd $(PY_SPEC_DIR); python3 -m venv venv; . venv/bin/activate; pip3 install flake8==3.5.0
+	cd $(PY_SPEC_DIR); python3 -m venv venv; . venv/bin/activate; pip3 install flake8==3.7.0
 
 lint: $(PY_SPEC_ALL_TARGETS)
 	cd $(PY_SPEC_DIR); . venv/bin/activate; \
-	flake8 --max-line-length=120 ./eth2spec;
+	flake8 --max-line-length=120 --per-file-ignores='./tests/*:F821,F403,F401,F405' ./eth2spec ./tests;
```
CarlBeek: Linter messages `F821, F401, F403, F405` will always be raised as they are a function of how I designed the tests. They are only ignored for the `test_libs/pyspec/tests/*` files.

`F821` is fundamental to this redesign and there are too many occurrences to handle individually with `# noqa`.
The linter doesn't raise any issues around the use of `spec.foo()` and `helpers.bar()` as, from its perspective, `spec` and `helpers` are undefined.

`F401, F403, F405` These could be ignored individually at the time of import, but I this is cleaner.


Are there any objections to doing it this way? Sidenote: using `--per-file-ignores='...'` is a feature that requires flake8 > v3.7.0. is there any reason why I can't use this version?

dankrad: Personally I'm ok with this. I think it's better than applying some other machinery to effectively circumvent the linter.

hwwhww:
@CarlBeek 
> The reason for importing the tests from phase0/* into phase1/* is that it allows for the phase 0 tests to be run against the phase 1 spec.

> This was the hack I came up with (to get the correct version of the spec to exist in the various test file): request.function.__globals__['spec'] = spec. I am not super happy with it, do you have any other ideas?

> I initially passed around function arguments to the tests that contained the spec version, but this became crazy very quickly.

I was thinking about how to get rid of these `import` and here is my attempt of making `spec` as fixture: https://github.com/ethereum/eth2.0-specs/commit/587f2ec2ceeb2f9198ecbc78f10bbc27ff1bf79e

1. Perhaps this idea is what you mean by "passed around function arguments to the tests", but here I pass the `spec`. 😄
2. It uses a workaround to [have `fixtures` in `pytest.mark.parametrize`](https://github.com/pytest-dev/pytest/issues/349) in
https://github.com/ethereum/eth2.0-specs/blob/587f2ec2ceeb2f9198ecbc78f10bbc27ff1bf79e/test_libs/pyspec/tests/conftest.py#L39-L43

3. Now I only make `test_libs/pyspec/tests/phase0/test_sanity.py` pass. Note that although its under `phase0` dir, it actually runs with both phase0 and phase1 spec. For example:
https://github.com/ethereum/eth2.0-specs/blob/587f2ec2ceeb2f9198ecbc78f10bbc27ff1bf79e/test_libs/pyspec/tests/phase0/test_sanity.py#L21-L27

4. My commit is very rough, but the goal is to organize the tests between phases/forks orderly.
    - For the phase0 & phase1 common tests, we can make `autouse=True` per file if we want. Or set it per test.
5.  Cons: the dumb passing-spec-variable approach would make `helpers.py` verbose - but I think it solves the question of [how to make future `helpers.py` work for different spec](https://github.com/ethereum/eth2.0-specs/pull/1061#discussion_r285808063).

CarlBeek:
@hwwhww Thank you for looking at it :)

> 1. Perhaps this idea is what you mean by "passed around function arguments to the tests", but here I pass the `spec`. 😄

Yes, this is what I meant by that. I actually implemented that somewhere in the past but it is very ugly. I find it obfuscates what is actually happening in the tests. This is further enhanced by the need of the helpers to also have access to the correct version of the spec. If you look back in my commit history eg. [`test_process_proposer_slashing.py` from "Passes spec & helpers module to tests phase0"](https://github.com/ethereum/eth2.0-specs/blob/7c8f83d5e84d04274eb61e3faea9f5d09b727c38/test_libs/pyspec/tests/phase0/block_processing/test_process_proposer_slashing.py) you can see just how crazy this gets.

> 5. Cons: the dumb passing-spec-variable approach would make helpers.py verbose - but I think it solves the question of how to make future helpers.py work for different spec.

As it currently exists, my different versions of the spec and different tests use different helpers, there is no problem with this. Tests for different phases and forks can be trivially activated/deactivated on a per-test level with `@pytest.mark.skipif(spec.phase <= 1)` etc. I think this solves #1101 quite nicely too.

test_libs/pyspec/eth2spec/phase1/state_transition.py
```diff
@@ -0,0 +1,70 @@
...
+
+    process_operations(state, block)
+    if verify_state_root:
+        spec.verify_block_state_root(state, block)
+
+
+def process_epoch_transition(state: BeaconState) -> None:
+    spec.process_justification_and_finalization(state)
+    spec.process_crosslinks(state)
+    # TODO: Eligible
```
dankrad: I think this still needs to be done -- the "eligible" function in 1_custody-game.md needs to be integrated into the process_penalties_and_exits function from the beacon chain.

dankrad:
How far are we on this? Seems like good progress has been made. From what I can tell, the TODOs would be:

- Integrate the "eligible" function
- add the 1_shard-data-chains.md file
- add some tests :)

test_libs/pyspec/eth2spec/phase1/state_transition.py
hwwhww: `process_penalties_and_exits` thing will be removed in #1116, so you can remove this TODO tag now!
```suggestion
```

CarlBeek: Yes, as I mentioned in a comment above, this whole file will be deleted. #1116 gets rid of `eligible`. It is on my list of TODOs

CarlBeek:
> How far are we on this? Seems like good progress has been made. From what I can tell, the TODOs would be:
> 
> * Integrate the "eligible" function
> * add the 1_shard-data-chains.md file
> * add some tests :)

The biggest thing that still need to happen to this is integration with #1111. Due to how significantly this changes how tests work, I think it is easier to merge dev into this branch once #1111 backports v06x. After that, I want to:

- [ ] Rework how the tests are configured and run. I'd like to switch to decorators like what is done in v06x and not necessarily have the implicit duplication of tests via imports into phase 1 of phase 0 tests.
- [ ] Transition away from my global variables hack so that static analysis tools can correctly analyse this code. (And have Flake8 stop shouting at me for having undefined variables.) I'd like to do this via the decorators mentioned above.
- [ ] Integrate `1_shard-data-chains.md` (as you mentioned)
- [ ] ~~Add some additional tests~~ I think I'd prefer to maybe do this in a separate PR as this one is already massive and like to limit it's scope at some point.
- [ ] Write up how to use my inserts and labels

specs/core/1_shard-data-chains.md
```diff
@@ -243,15 +252,20 @@ def compute_crosslink_data_root(blocks: List[ShardBlock]) -> Hash:
...
 
     return hash(
-        merkle_root(pad_to_power_of_2([
-            merkle_root_of_bytes(zpad(serialize(get_shard_header(block)), BYTES_PER_SHARD_BLOCK_BODY)) for block in blocks
-        ])) +
-        merkle_root(pad_to_power_of_2([
-                merkle_root_of_bytes(block.body) for block in blocks
+        hash_tree_root(pad_to_power_of_2([
+            hash_tree_root_of_bytes(
+                zpad(serialize(get_shard_header(block)), BYTES_PER_SHARD_BLOCK_BODY)
```
CarlBeek: I don't really know what `zpad` is supposed to do here. Based on the arguments, I implemented a `zpad` function above, but I am unsure whether this is correct.

More generally, I think the whole `compute_crosslink_data_root` function can be neatened up and simplified.

specs/core/1_shard-data-chains.md
```diff
@@ -137,7 +145,8 @@ def get_period_committee(state: BeaconState, epoch: Epoch, shard: Shard, index:
 ```python
 def get_switchover_epoch(state: BeaconState, epoch: Epoch, index: ValidatorIndex):
     earlier_start_epoch = epoch - (epoch % PERSISTENT_COMMITTEE_PERIOD) - PERSISTENT_COMMITTEE_PERIOD * 2
-    return bytes_to_int(hash(generate_seed(state, earlier_start_epoch) + bytes3(index))[0:8]) % PERSISTENT_COMMITTEE_PERIOD
```
CarlBeek: I am not exactly sure what is meant by `bytes3(index))[0:8]` here. I changed it to `int_to_bytes(index, length=3)[0:8]` as this seemed the most intuitive to me, but slicing bytes of length 3 into [0:8] doesn't make much sense.

test_libs/pyspec/eth2spec/test/context.py
```diff
@@ -80,3 +88,35 @@ def entry(*args, **kw):
...
+def with_all_phases(fn):
+    """
+    Decorator to run everything with all availible spec phases
+    """
+    def entry(*args, **kw):
+        kw['spec'] = spec_phase0
+        fn(*args, **kw)
+        kw['spec'] = spec_phase1
+        return fn(*args, **kw)
+    return entry
```
hwwhww: Basically, I like `with_all_phases` approach!

Some issues:
1. The state generator is broken now. I run:
```sh
make eth2.0-spec-tests/tests/sanity
```
(Not sure if I missed any new flag?)

Error message:
```sh
  File "main.py", line 34, in <module>
    create_suite('slots', 'mainnet', lambda: generate_from_tests(test_slots)),
  File "/Users/hwwang/ethereum/eth2.0-specs/test_generators/sanity/venv/lib/python3.6/site-packages/gen_base/gen_runner.py", line 98, in run_generator
    (output_name, handler, suite) = suite_creator(args.configs_path)
  File "main.py", line 25, in suite_definition
    test_cases=get_cases()))
  File "main.py", line 31, in <lambda>
    create_suite('blocks', 'minimal', lambda: generate_from_tests(test_blocks)),
  File "/Users/hwwang/ethereum/eth2.0-specs/test_generators/sanity/venv/lib/python3.6/site-packages/gen_from_tests/gen.py", line 19, in generate_from_tests
    test_case = tfn(generator_mode=True, bls_active=bls_active)
  File "/Users/hwwang/ethereum/eth2.0-specs/test_generators/sanity/venv/lib/python3.6/site-packages/eth2spec/test/context.py", line 121, in entry
    return fn(*args, **kw)
  File "/Users/hwwang/ethereum/eth2.0-specs/test_generators/sanity/venv/lib/python3.6/site-packages/eth2spec/test/context.py", line 13, in entry
    kw['state'] = create_genesis_state(spec=kw['spec'], num_validators=spec_phase0.SLOTS_PER_EPOCH * 8)
  File "/Users/hwwang/ethereum/eth2.0-specs/test_generators/sanity/venv/lib/python3.6/site-packages/eth2spec/test/helpers/genesis.py", line 35, in create_genesis_state
    state.validator_registry = [build_mock_validator(spec, i, state.balances[i]) for i in range(num_validators)]
  File "/Users/hwwang/ethereum/eth2.0-specs/test_generators/sanity/venv/lib/python3.6/site-packages/eth2spec/test/helpers/genesis.py", line 35, in <listcomp>
    state.validator_registry = [build_mock_validator(spec, i, state.balances[i]) for i in range(num_validators)]
  File "/Users/hwwang/ethereum/eth2.0-specs/test_generators/sanity/venv/lib/python3.6/site-packages/eth2spec/test/helpers/genesis.py", line 8, in build_mock_validator
    withdrawal_credentials = spec.BLS_WITHDRAWAL_PREFIX_BYTE + spec.hash(pubkey)[1:]
AttributeError: module 'eth2spec.phase1.spec' has no attribute 'BLS_WITHDRAWAL_PREFIX_BYTE'
generator /sanity finished
```

2. It looks like the state test generator output would be only the result of `spec_phase1`, no `spec_phase0`?

specs/core/0_beacon-chain.md
```diff
@@ -1278,8 +1278,11 @@ def process_epoch(state: BeaconState) -> None:
     process_crosslinks(state)
     process_rewards_and_penalties(state)
     process_registry_updates(state)
+    # @process_reveal_deadlines
+    # @process_challenge_deadlines
     process_slashings(state)
     process_final_updates(state)
+    # @after_process_final_updates
```
hwwhww: I think it worths some comment here to explain the labels here for the readers.

scripts/build_spec.py
```diff
@@ -0,0 +1,271 @@
...
+    return old_functions
+
+
+def combine_constants(old_constants: Dict[str, str], new_constants: Dict[str, str]) -> Dict[str, str]:
+    for key, value in new_constants.items():
+        old_constants[key] = value
+    return old_constants
+
+
+def dependency_order_ssz_objects(objects: Dict[str, str]) -> Dict[str, str]:
```
dankrad:
```suggestion
def dependency_order_ssz_objects(objects: Dict[str, str]) -> None:
```

specs/core/1_custody-game.md
dankrad: See my other comment below -- seems like we could just use python `type()` here, right? What's the argument against doing that?

specs/core/1_custody-game.md
```diff
@@ -369,15 +361,18 @@ def process_custody_key_reveal(state: BeaconState,
...
+        get_validators_custody_reveal_period(state, reveal.revealed_index) - revealer.next_custody_reveal_period
+    )
 
     # Process reveal
     revealer.next_custody_reveal_period += 1
 
     # Reward Block Preposer
     proposer_index = get_beacon_proposer_index(state)
-    increase_balance(state, proposer_index, base_reward(state, index) // MINOR_REWARD_QUOTIENT)
+    increase_balance(state, proposer_index, get_base_reward(state, proposer_index) // MINOR_REWARD_QUOTIENT)
```
djrtwo: This reawrd might better be a function of the `revealer_index` rather than `proposer_index`

specs/core/1_custody-game.md
```diff
@@ -463,13 +462,13 @@ For each `challenge` in `block.body.custody_chunk_challenges`, run the following
 def process_chunk_challenge(state: BeaconState,
                             challenge: CustodyChunkChallenge) -> None:
     # Verify the attestation
-    assert verify_indexed_attestation(state, convert_to_indexed(state, challenge.attestation))
+    assert validate_indexed_attestation(state, convert_to_indexed(state, challenge.attestation))
```
djrtwo: No need for `assert` here. `validate_indexed_attestation` handles the asserts.

```suggestion
    validate_indexed_attestation(state, convert_to_indexed(state, challenge.attestation))
```

specs/core/1_custody-game.md
```diff
@@ -518,14 +517,14 @@ def process_bit_challenge(state: BeaconState,
     assert is_slashable_validator(challenger, get_current_epoch(state))
 
     # Verify the attestation
-    assert verify_indexed_attestation(state, convert_to_indexed(state, challenge.attestation))
+    assert validate_indexed_attestation(state, convert_to_indexed(state, challenge.attestation))
```
djrtwo: ditto

```suggestion
    validate_indexed_attestation(state, convert_to_indexed(state, challenge.attestation))
```

specs/core/1_custody-game.md
```diff
@@ -518,14 +517,14 @@ def process_bit_challenge(state: BeaconState,
     assert is_slashable_validator(challenger, get_current_epoch(state))
 
     # Verify the attestation
```
djrtwo: worth assigning `attestation = challenge.attestation` to reduce line lengths throughout function

specs/core/1_custody-game.md
```diff
@@ -659,20 +661,25 @@ def process_bit_challenge_response(state: BeaconState,
...
+# begin insert @process_reveal_deadlines
+    process_reveal_deadlines(state)
+# end insert @process_reveal_deadlines
 def process_reveal_deadlines(state: BeaconState) -> None:
     for index, validator in enumerate(state.validator_registry):
-        if (validator.latest_custody_reveal_period +
-            (CUSTODY_RESPONSE_DEADLINE // EPOCHS_PER_CUSTODY_PERIOD) <
-            get_validators_custody_reveal_period(state, index)):
-                slash_validator(state, index)
+        if (validator.next_custody_reveal_period + (CUSTODY_RESPONSE_DEADLINE // EPOCHS_PER_CUSTODY_PERIOD)
```
djrtwo: consider assignment before `if` -- `reveal_deadline = validator.next_custody_reveal_period + (CUSTODY_RESPONSE_DEADLINE // EPOCHS_PER_CUSTODY_PERIOD`

specs/core/1_shard-data-chains.md
```diff
@@ -47,14 +47,16 @@ This document describes the shard data layer and the shard fork choice rule in P
 | `BYTES_PER_SHARD_BLOCK_BODY` | `2**14` (= 16,384) |
 | `MAX_SHARD_ATTESTIONS` | `2**4` (= 16) |
 | `PHASE_1_GENESIS_EPOCH` | **TBD** |
-| `PHASE_1_GENESIS_SLOT` | get_epoch_start_slot(PHASE_1_GENESIS_EPOCH) |
+| `PHASE_1_GENESIS_SLOT` | **TBD** |
```
djrtwo: Not quite relevant to this PR but i think "genisis" is a bad name here. Suggest "fork" -- `PHASE_1_FORK_*`

specs/core/1_shard-data-chains.md
```diff
@@ -47,14 +47,16 @@ This document describes the shard data layer and the shard fork choice rule in P
...
+| `GENESIS_SHARD_SLOT` | 0 |
 
 ### Time parameters
 
 | Name | Value | Unit | Duration |
 | - | - | :-: | :-: |
 | `CROSSLINK_LOOKBACK` | `2**0` (= 1) | epochs  | 6.2 minutes |
 | `PERSISTENT_COMMITTEE_PERIOD` | `2**11` (= 2,048) | epochs | ~9 days |
+| `SECONDS_PER_SLOT` | `2**1 * 3**1` (= 6) | 6 seconds |
```
djrtwo: This is now defined in two different places -- fork choice doc and this shard chain doc.

@JustinDrake would argue that the `assert` using `SECONDS_PER_SLOT` here should instead be in the fork choice to limit the tree under consideration.

test_generators/epoch_processing/main.py
```diff
@@ -14,7 +15,8 @@ def create_suite(transition_name: str, config_name: str, get_cases: Callable[[],
         -> Callable[[str], gen_typing.TestSuiteOutput]:
     def suite_definition(configs_path: str) -> gen_typing.TestSuiteOutput:
         presets = loader.load_presets(configs_path, config_name)
-        spec.apply_constants_preset(presets)
+        spec_phase0.apply_constants_preset(presets)
+        spec_phase1.apply_constants_preset(presets)
```
djrtwo: Does this generator actually use phase1 at all?

test_generators/operations/main.py
```diff
@@ -13,14 +13,16 @@
...
+from eth2spec.phase1 import spec as spec_phase1
 
 
 def create_suite(operation_name: str, config_name: str, get_cases: Callable[[], Iterable[gen_typing.TestCase]]) \
         -> Callable[[str], gen_typing.TestSuiteOutput]:
     def suite_definition(configs_path: str) -> gen_typing.TestSuiteOutput:
         presets = loader.load_presets(configs_path, config_name)
-        spec.apply_constants_preset(presets)
+        spec_phase0.apply_constants_preset(presets)
+        spec_phase1.apply_constants_preset(presets)
```
djrtwo: ditto for all generators

test_libs/pyspec/eth2spec/test/block_processing/test_process_attestation.py
```diff
@@ -191,111 +196,119 @@ def test_invalid_current_source_root(state):
...
     attestation.custody_bitfield = deepcopy(attestation.aggregation_bitfield) + b'\x00'
 
-    sign_attestation(state, attestation)
+    sign_attestation(spec, state, attestation)
 
-    yield from run_attestation_processing(state, attestation, False)
+    yield from run_attestation_processing(spec, state, attestation, False)
 
 
+@with_all_phases
```
djrtwo: This is only a phase 0 test

```suggestion
@with_phase0
```

(need to handle import as well)

test_libs/pyspec/eth2spec/test/block_processing/test_process_early_derived_secret_reveal.py
```diff
@@ -0,0 +1,130 @@
...
+    spec.process_early_derived_secret_reveal(state, randao_key_reveal)
+
+    slashed_validator = state.validator_registry[randao_key_reveal.revealed_index]
+
+    if randao_key_reveal.epoch >= spec.get_current_epoch(state) + spec.CUSTODY_PERIOD_TO_RANDAO_PADDING:
+        assert slashed_validator.slashed
+        assert slashed_validator.exit_epoch < spec.FAR_FUTURE_EPOCH
+        assert slashed_validator.withdrawable_epoch < spec.FAR_FUTURE_EPOCH
+    # lost whistleblower reward
+    # FIXME: Currently broken because get_base_reward in genesis epoch is 0
```
djrtwo: Why is base reward 0 at genesis epoch? total active balance should be non-zero as well as the effective balance of the validator

test_libs/pyspec/eth2spec/test/block_processing/test_process_early_derived_secret_reveal.py
```diff
@@ -0,0 +1,130 @@
...
+def test_reveal_from_current_epoch(spec, state):
+    randao_key_reveal = get_valid_early_derived_secret_reveal(spec, state, spec.get_current_epoch(state))
+
+    yield from run_early_derived_secret_reveal_processing(spec, state, randao_key_reveal, False)
+
+
+@with_phase1
+@spec_state_test
+def test_reveal_from_past_epoch(spec, state):
+    if spec.get_current_epoch(state) < 1:
```
djrtwo: Does this ever run? I'm pretty sure we are only passing in "genesis" states to tests currently.
Could just run `next_epoch` a few times to get past the start epoch

test_libs/pyspec/eth2spec/test/block_processing/test_process_early_derived_secret_reveal.py
```diff
@@ -0,0 +1,130 @@
+import pytest
```
djrtwo: Even though we have these nice decorators for specify which specs to run against, I'm still kind of in favor of having phase 1 tests in a separate directory

test_libs/pyspec/eth2spec/test/block_processing/test_process_early_derived_secret_reveal.py
djrtwo:
```
    next_epoch(spec, state)
    apply_empty_block(spec, state)
```

test_libs/pyspec/eth2spec/test/context.py
```diff
@@ -80,3 +88,35 @@ def entry(*args, **kw):
...
+
+def with_all_phases(fn):
+    """
+    Decorator to run everything with all availible spec phases
+    """
+    def entry(*args, **kw):
+        kw['spec'] = spec_phase0
+        fn(*args, **kw)
+        kw['spec'] = spec_phase1
+        return fn(*args, **kw)
```
djrtwo: Is this return necessary?

test_libs/pyspec/eth2spec/test/block_processing/test_process_attestation.py
```diff
@@ -191,111 +196,119 @@ def test_invalid_current_source_root(state):
...
 @spec_state_test
-def test_non_zero_crosslink_data_root(state):
-    attestation = get_valid_attestation(state)
+def test_non_zero_crosslink_data_root(spec, state):
```
djrtwo: This is also just `phase0` spec

test_libs/pyspec/eth2spec/test/phase_1/block_processing/test_process_early_derived_secret_reveal.py
```diff
@@ -0,0 +1,130 @@
...
+    spec.process_early_derived_secret_reveal(state, randao_key_reveal)
+
+    slashed_validator = state.validator_registry[randao_key_reveal.revealed_index]
+
+    if randao_key_reveal.epoch >= spec.get_current_epoch(state) + spec.CUSTODY_PERIOD_TO_RANDAO_PADDING:
+        assert slashed_validator.slashed
+        assert slashed_validator.exit_epoch < spec.FAR_FUTURE_EPOCH
+        assert slashed_validator.withdrawable_epoch < spec.FAR_FUTURE_EPOCH
+    # lost whistleblower reward
+    # FIXME: Currently broken because get_base_reward in genesis epoch is 0
```
djrtwo: Why is base reward 0 at genesis epoch? total active balance should be non-zero as well as the effective balance of the validator

CarlBeek: You are correct. I copied this over from the original phase 1 tests (before it was executable) and this came from there. It is fixed in the latest tests.

specs/core/1_custody-game.md
CarlBeek: Yeah, that shortens things a lot.

specs/core/1_custody-game.md
CarlBeek: Agreed

specs/core/1_shard-data-chains.md
CarlBeek: Yes, but I repeated it for the purposes of having the spec be executable as the fork choice not being executable means that `SECONDS_PER_SLOT ` is not in the scope of spec.py

I agree that it should not necessarily be a condition for block validity at a state-transition level. The reason I like having this check in the fork choice is less for tree limiting and more for having block validity purely dependent on state and the block's contents.

test_generators/epoch_processing/main.py
CarlBeek: So the reason for this is what @hwwhww  pointed out below in her comment about the decorators. Because of the way they are constructed, the the test generators now run against phase 0 and phase 1. That means that without this the state generator is broken. (i.e. running `make eth2.0-spec-tests/tests/sanity` fails)

test_libs/pyspec/eth2spec/test/context.py
```diff
@@ -80,3 +88,35 @@ def entry(*args, **kw):
...
+    """
+    Decorator to use phase 1's spec
+    """
+    def entry(*args, **kw):
+        kw['spec'] = spec_phase1
+        return fn(*args, **kw)
+    return entry
+
+
+def with_all_phases(fn):
```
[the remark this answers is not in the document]
CarlBeek: I like this idea and it may be a good way of resolving the state generator issue w.r.t. only running the appropriate phase.

test_generators/operations/main.py
CarlBeek: See generator comment above

test_libs/pyspec/eth2spec/test/phase_1/block_processing/test_process_early_derived_secret_reveal.py
dankrad: Ah that was me, I remember there was this issue at some point. Seems like it's fixed now.

test_libs/pyspec/eth2spec/test/context.py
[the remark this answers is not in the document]
CarlBeek: I swapped out the decorators for parameterised decorator factories so the behaviour you described is now the default. :)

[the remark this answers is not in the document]
djrtwo: whoops! I'm not certain I was recommending to change the default from `@with_all_phases`. I think `with_all_phases` is a better default because we assume functionality works when we add a new phase unless we explicitly make it not work. I just thought it might be useful when distinguishing from the default `@with_phases(['phase0'])` rather than `with_phase0`.

I guess I want the default behavior to have to explicitly pull tests out for a phase rather than have to put tests in for a phase. Thoughts? Sorry for my miscommunication in my vague comment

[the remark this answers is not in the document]
CarlBeek: I had this debate myself. I'll re-add `with_all_phases`. It'll basically be an alias for the `with_phases` decorator factory, but with all possible phases populating the list.

I could also add an opt out decorator. Something like `with_all_phases_except(['phase0'])`. What are your thoughts?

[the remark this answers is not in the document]
djrtwo: Yeah, that would likely work well for things that are `>=` some phase. Most of the phase1 will be "all phases except 0"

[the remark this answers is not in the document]
CarlBeek: 👍 I implemented in this in the [latest commit.](https://github.com/ethereum/eth2.0-specs/pull/1061/commits/67471a8d6e802115513c4014fa25c6abd28c7f35)

PR #1067, opened 2019-05-09
Fix get_shuffled_index rename in test generators

mratsim:
Additional update to https://github.com/ethereum/eth2.0-specs/pull/1045 following https://github.com/ethereum/eth2.0-specs/pull/1009


mratsim:
Closing, I didn't realize that it should go into `v06x` branch (it's fixed here) instead of `dev`

Issue #1012, opened 2019-04-30
BeaconNode <--> ValidatorClient API - Protocol


spble:
# ETH2.0 Beacon Node & Validator Client Protocol Discussion

**Further background, and actual protocol, is described in issue #1011**

It would be useful to choose a standard protocol for the `BeaconNode` and `ValidatorClient` API.

It was discussed during the [Client Architecture](https://notes.ethereum.org/iuPB2YjKQMGua0nwCZJVAQ) session at the [Sydney Implementers meeting](https://notes.ethereum.org/7w7diW1-St2_Yu-YHjg6NA) that the main decision is between [gRPC](https://grpc.io/) and [JSON-RPC](https://www.jsonrpc.org/specification).
This discussion was a follow on from the [Client Architecture Roundtable](https://hackmd.io/s/rJl05Lk6X) in Prague.

### gRPC
**Advantages**
 - Fast
 - Highly specified, compiled specification
 - Has explicit types and support for binary data

**Disadvantages**
 - Non human-readable transport
 - Requires specific framework for integration other software
 - Requires special software to interface with directly

### JSON-RPC
**Advantages**
- Human-readable transport
- Simple enough to be implemented by any language
- Well understood by many developers
- Can easily be called with `curl`
- Consistency with Eth1.0 clients

**Disadvantages**
- Slower
- Some ambiguity around data representations
- Specification of API not as rigerous, must be specified out-of-band

In conclusion, most people had a preference towards JSON-RPC mainly due to it's human readability and ease of implementation.

prestonvanloon:
For Prysm, we will be continuing to use protocol buffers for our beacon chain and validator implementation. 
The discussion within the team is that the API enforcement within generative code and performance gains outweigh the marginal benefit of using `curl` or other pre-installed tools rather than tools created for the ecosystem. 

Client interop may be achieved through a gRPC proxy gateway, but the bidirectional streaming would not work so we may not support JSON-RPC unless there is a very compelling reason to do so.

spble:
Thanks for the input @prestonvanloon - I definitely see the performance improvements with using gRPC, but I imagine the interface between the BeaconNode and ValidatorClient will be a fairly low-bandwidth interface. As such, doing a call in 10ms instead of 100ms would not bring any substantial benefit in my opinion.

[...]

Issue #1011, opened 2019-04-30
BeaconNode <--> ValidatorClient API - Messages


spble:
# ETH2.0 Beacon Node & Validator Client RPC Proposal

## Outline

This issue proposes a minimal communications specification between a beacon node and validator client, targeting _phase 0_ of the Eth2.0 specification. The protocol is designed to be a simple local communications interface, permitting two separate binaries to communicate.

This specification is intended to describe communication abstractly, without choosing any particular protocol. The protocol used (e.g. gRPC/JSON-RPC) is discussed in a separate issue: #1012.

This issue follows on from a discussion during the [Client Architecture](https://notes.ethereum.org/iuPB2YjKQMGua0nwCZJVAQ) session at the [Sydney Implementers meeting](https://notes.ethereum.org/7w7diW1-St2_Yu-YHjg6NA). This also follow from the [Client Architecture Roundtable](https://hackmd.io/s/rJl05Lk6X) in Prague.
There is an editable version of this document, here: https://hackmd.io/r7D61hs4RWKm8nz_O2iinQ

###  Background
The beacon node maintains the state of the beacon chain by communicating with other beacon nodes in the Ethereum Serenity network. Conceptually, it does not maintain keypairs that participate with the beacon chain.

The validator client is conceptually a separate entity which utilises private keys to perform validator related tasks on the beacon chain, which we call validator "duties". This includes the production of beacon blocks and signing of attestations.

Since it is recommended to separate these concerns in the client implementations, it is necessary for us to clearly define the communication between them.

The goal of this specification is to promote interoperability between beacon nodes and validator clients derived from different projects. For example, the validator client from Lighthouse, could communicate with a running instance of the beacon node from Prysm.

This proposal has been adapted from the [Lighthouse gRPC protocol specification](https://github.com/sigp/lighthouse/blob/master/protos/src/services.proto).

There is also another [WIP proposal for a Minimum Validator Interface](https://notes.ethereum.org/Ia2kvjy0RX2J-GxrWfoCAQ), which describes additional functions possibly necessary for phase 1 and beyond.

### Entities 
The following are the two entities that participate in this protocol:
 - **`BeaconNode`**:
 A beacon node instance, run with a `--rpc` flag to enable the RPC interface. Runs stand-alone.

 - **`ValidatorClient`**:
A validator client instance, which must connect to at least one instance of `BeaconNode`.

### Endpoints
[...]

PR #1069, opened 2019-05-09
Beacon Node API for Validator

spble:
This PR proposes an OpenAPI specification for communication between the beacon node and the validator client. This is intended to be as minimal a specification as possible, just enough to allow the validator client implementation to function without any extra features.

This PR formalises the discussions in issues #1011 and #1012.

This RFC/proposal creates an OpenAPI yaml file inside the `specs/validator` folder which is discussed by a short markdown document in the same folder. A link to this markdown document has been placed in the primary README.md file for the repository.

If you'd like to explore the API proposal directly, please take a look on [SwaggerHub](https://app.swaggerhub.com/apis/spble/beacon_node_api_for_validator)


specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,613 @@
...
+              type: string
+              format: byte
+              pattern: "^0x[a-fA-F0-9]{64}$"
+              description: "Root of the crosslinked shard data since the previous crosslink."
+
+  responses:
+    Success:
+      description: "Request successful."
+    InvalidRequest:
+      description: "Invalid request syntax."
```
hwwhww: Maybe define a new code `404` for the "Resource not found".

arnetheduck:
> A drawback is that we need to maintain both this yaml file and core spec object scheme for the same objects. We need to go through it again when freezing the spec.

if we formalize the schema format for the spec messages a bit, one can be generated from the other, potentially - that could be useful in the future as well, when defining new messages!

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,613 @@
+openapi: "3.0.2"
+info:
+  title: "Minimal Beacon Node API for Validator"
+  description: "A minimal API specification for the beacon node, which enables a validator to connect and perform its obligations on the Ethereum 2.0 phase 0 beacon chain."
+  version: "0.1"
+  license:
+    name: "Apache 2.0"
+    url: "https://www.apache.org/licenses/LICENSE-2.0.html"
+tags:
+  - name: Necessary for validator
```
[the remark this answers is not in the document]
spble: Good idea!

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,613 @@
...
+          description: "The index of the validator in the committee."
+        attestation_slot:
+          type: integer
+          format: uint64
+          description: "The slot at which the validator must attest."
+        attestation_shard:
+          type: integer
+          format: uint64
+          description: "The shard in which the validator must attest."
+        block_production_slot:
```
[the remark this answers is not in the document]
spble: Good idea, that name is more consistent.

specs/validator/beacon_node_oapi.yaml
spble: I agree that an error `404` should be returned if a non-existent API endpoint is requested, however I don't think that a '404 - NotFound' error would be an appropriate return code for any of the actual endpoints (since you aren't requesting data which may or may not exist).
Since the actual error code is defined on the `path` (endpoint) definitions, it didn't seem appropriate to list a `NotFound` in components without referring to it somewhere in the endpoint definition. I agree with you however, that it would be nice to have it listed with all the other responses so that implementers know it is actually a valid response.

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,613 @@
...
+                    - type: string
+                      format: byte
+                      pattern: "^0x[a-fA-F0-9]{64}$"
+                    - type: integer
+                      format: uint64
+                      enum: [32]
+                      example: 32
+                      description: "The DEPOSIT_CONTRACT_TREE_DEPTH value."
+                minItems: 2
+                maxItems: 2
```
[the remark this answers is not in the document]
spble: This is the only way I could figure out how to specify the [`Deposit`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#deposit) object, since it defines a two element array, where each element is a different type.

The definition I have provided is as restrictive as I could be while still enabling the spec. It says that it must be an array of length two, and items must either be bytes or the actual integer `32`.
Technically this means you could provide invalid things here, such as `[32, 32]` or `[bytes, bytes]`, but I wasn't sure how to be more specific in this case using OpenAPI 3. Certainly open to suggestion!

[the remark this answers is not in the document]
hwwhww: > the Deposit object, since it defines a two element array

This is a [SSZ Vector](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/simple-serialize.md#composite-types) notation,
The notation of `'proof': ['bytes32', DEPOSIT_CONTRACT_TREE_DEPTH],` means SSZ vector of fixed size DEPOSIT_CONTRACT_TREE_DEPTH`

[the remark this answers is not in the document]
hwwhww: > the Deposit object, since it defines a two element array

`'proof': ['bytes32', DEPOSIT_CONTRACT_TREE_DEPTH]` is a [SSZ Vector](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/simple-serialize.md#composite-types) notation, it means fixed-length (`DEPOSIT_CONTRACT_TREE_DEPTH`) array of `bytes32` values.

Now I see how confusing this notation is. 😬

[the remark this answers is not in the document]
spble: Ahh right! I thought it seemed strange. Thank you for clearing that up, I will push a fix for that.

spble:
Yes, it is a shame to have it duplicated and would be best to have it generated.
Of course, it is difficult to generate this YAML from the python/markdown, so I am more than happy to maintain this spec and update it after the big freeze!

shanejonas:
Just wanted to chime in and show that this is possible with `JSON-RPC` with the [OpenRPC Spec](https://open-rpc.org).

Here is an example of the [playground with the ethereum-json-rpc-specification](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/etclabscore/ethereum-json-rpc-specification/master/openrpc.json)

It simplifies a lot of what you have already since it follows familiar principles.

paulhauner:
> Just wanted to chime in and show that this is possible with `JSON-RPC` with the [OpenRPC Spec](https://open-rpc.org).

Thanks. We're not using REST/HTTP just for OpenAPI though (I think we'd use it even if OpenAPI didn't exist). See [here](https://github.com/ethereum/eth2.0-specs/issues/1012#issuecomment-489660765) for detailed reasoning.

shanejonas:
REST is great until you want to support other transports. Then it starts to fall apart. JSON-RPC is simple and transport agnostic, and now has the tools to compete.

The arguments outlined (like stateful across multiple calls) only apply to API design and  can be avoided.

I’d love to see ETH continue to use JSON-RPC with the rest of the blockchain ecosystem.

FrankSzendzielarz:
@shanejonas REST APIs (HTTP) are fairly transport agnostic. Can you clarify? Eth 1.x implementers see stateful rpc apis as having been something of a mistake, afaik.

shanejonas:
With HTTP (and swagger outlined here). When you do want to support websocket, IPC, postMessage/webworker or any other transport in the future you will need yet another spec and debate on how to best describe it.

JSON-RPC doesn't tell you how stateful your API should be. Those are implementation details.

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,609 @@
...
+      parameters:
+      - name: beacon_block
+        in: query
+        required: true
+        description: "The `BeaconBlock` object, as sent from the beacon node originally, but now with the signature field completed."
+        schema:
+          $ref: '#/components/schemas/BeaconBlock'
+      responses:
+        200:
+          $ref: '#/components/responses/Success'
```
djrtwo: Does this response code here mean that beacon node sees the block as valid, has incorporated into local view, and broadcast to network, or does it just mean that it was broadcast?

More specifically, do we expect a BN to validate the incoming block at this point before broadcast or to just assume it is valid and broadcast? In the general case, the BN would likely want to assume it is valid before broadcasting to reduce latency in propagating the block to network.

I suppose the only way a block would be invalid in between this call and response would be if the validator added an invalid sig or if the BN forgot about part of it's block tree between the call and response.  So maybe the validation expected in the response code is just the proposer signature.

The normal reorg scenario is not really an issue because the BN would still see the block as valid in the context of the entire block tree rather than just the head of the chain.

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,609 @@
...
+        400:
+          $ref: '#/components/responses/InvalidRequest'
+        500:
+          $ref: '#/components/responses/InternalError'
+        503:
+          $ref: '#/components/responses/CurrentlySyncing'
+    post:
+      tags:
+        - MinimalSet
+      summary: "Published a signed attestation."
```
djrtwo:
```suggestion
      summary: "Publish a signed attestation."
```

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,609 @@
...
+      description: "Instructs the beacon node to publish a newly signed IndexedAttestation object, to be incorporated into the beacon chain."
```
djrtwo: This is to be broadcast on the particular shard subnet. Let's that mention that here.

As for "to be incorporated into the beacon chain", it is only incorporated into the local chain when it is seen on a subsequent block rather than immediately added to local chain. Might just remove this part of the statement.

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,609 @@
...
+          $ref: '#/components/responses/InternalError'
+        503:
+          $ref: '#/components/responses/CurrentlySyncing'
+
+  /validator/attestation:
+    get:
+      tags:
+        - MinimalSet
+      summary: "Produce an attestation, without signature."
+      description: "Requests that the beacon node produce an IndexedAttestation, with a blank signature field, which the validator will then sign."
```
djrtwo: I'm assuming the custody bitfield is already filled with a 1 for the validator requesting the attestation?

In phase 1, this might be a two stage process.
* Request custody data from BN, and compute local proof of custody.
* Request attestation, passing the PoC bit as a param for the BN to fill in appropriately.

Since this is the case, does it make sense to add a the `custody_bit` param here and have the phase 0 validators always fill it in with 0?

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,609 @@
...
+      description: "The validator's BLS public key, uniquely identifying them. _48-bytes, hex encoded with 0x prefix, case insensitive._"
+      example: "0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc"
+    version:
+      type: string
+      description: "A string which uniquely identifies the client implementation and its version; similar to [HTTP User-Agent](https://tools.ietf.org/html/rfc7231#section-5.5.3)."
+      example: "Lighthouse / v0.1.5 (Linux x86_64)"
+    genesis_time:
+      type: integer
+      format: uint64
+      description: "The genesis_time configured for the beacon node, which is the time the Eth1.0 validator deposit smart contract has enough ETH staked (i.e. Eth2.0 begins)."
```
djrtwo: Removed the reference to the deposit contract because other potential eth2 chains might start without the connection to an eth1 chain.
```suggestion
      description: "The genesis_time configured for the beacon node, which is the unix time at which the Eth2.0 chain began."
```

specs/validator/beacon_node_oapi.yaml
djrtwo: If instead, the intention is for the validator to plug in their PoC bit when signing and passing the attestation back to BN, then we should note it here.

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,609 @@
...
+      example: 1557716289
+    ValidatorDuty:
+      type: object
+      properties:
+        validator_pubkey:
+          $ref: '#/components/schemas/pubkey'
+        committee_index:
```
djrtwo: Is the use here to fill in the poc bit?

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,609 @@
...
+        current_version:
+          type: string
+          format: byte
+          pattern: "^0x[a-fA-F0-9]{8}$"
+          description: "Current fork version."
+        epoch:
+          type: integer
+          format: uint64
+          description: "Fork epoch number."
+    IndexedAttestation:
```
djrtwo: `IndexedAttestation` is missing `signature` field

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,609 @@
...
+        crosslink:
+          title: CrossLink
+          type: object
+          description: "The [`Crosslink`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#crosslink) object from the Eth2.0 spec."
+          properties:
+            shard:
+              type: integer
+              format: uint64
+              description: "The shard number."
+            epoch:
```
djrtwo: `epoch` has been removed and `start_epoch` and `end_epoch` have been added to `Crosslink` in `dev` branch

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,609 @@
...
+    BeaconBlock:
+      description: "The [`BeaconBlock`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#beaconblock) object from the Eth2.0 spec."
+      allOf:
+        - $ref: '#/components/schemas/BeaconBlockCommon'
+        - type: object
+          properties:
+            body:
+              $ref: '#/components/schemas/BeaconBlockBody'
+    BeaconBlockHeader:
+      description: "The [`BeaconBlockHeader`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#beaconblockheader) object from the Eth2.0 spec."
```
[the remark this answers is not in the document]
spble: I agree, though I am concerned about this since I think the most usable version of this document is when it's listed on SwaggerHub (or some swagger viewer). The links won't be interpreted correctly if they are relative in this case.
Maybe there is a better absolute url that I can use, rather than the `dev` branch?
Otherwise, if you think they should be relative links regardless, happy to update them.

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,609 @@
...
+          description: "The slot in which a validator must propose a block, or `null` if block production is not required."
+    SyncingStatus:
+      type: object
+      nullable: true
+      properties:
+        starting_block:
+          type: integer
+          format: uint64
+          description: "The block at which syncing started (will only be reset after the sync reached its head)"
+        current_block:
```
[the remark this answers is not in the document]
spble: Yep, these should be changed to have slot numbers, not block numbers.

specs/validator/beacon_node_oapi.yaml
spble: Yes, this is so that the validator knows how to do the PoC bits.
Due to the other changes suggested/made here (offloading this responsibility to the beacon node), this is no longer necessary.

specs/validator/beacon_node_oapi.yaml
```diff
@@ -0,0 +1,609 @@
...
+          $ref: '#/components/responses/InternalError'
+
+  /validator/duties:
+    get:
+      tags:
+        - MinimalSet
+      summary: "Get validator duties for the requested validators."
+      description: "Requests the beacon node to provide a set of _duties_, which are actions that should be performed by validators. This API call should be polled at every slot, to ensure that any chain reorganisations are catered for, and to ensure that the currently connected beacon node is properly synchronized."
+      parameters:
+      - name: validator_pubkeys
```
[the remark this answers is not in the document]
spble: I have added an epoch parameter and updated the description here to make this clearer. Have also added a `406` status code, to indicate that the requested epoch is out of range.

specs/validator/beacon_node_oapi.yaml
spble: After some discussion with @paulhauner - we concluded that in our case, the beacon node will need to validate the incoming block so that it can be incorporated into the state, before it is broadcast. As such, we need to perform validation of the block.

However, we also concluded that it's not the beacon node's responsibility to prevent broadcast of a block if it deems it to be invalid; the validator should know what it's doing.

As such, I have clarified this behaviour in the description, and created a new status code `202` which indicates that the block was published; however it did not pass validity checks (and therefore has not been incorporated into the beacon node's database).

specs/validator/beacon_node_oapi.yaml
spble: Great suggestion, I have added an extra parameter to the attestation GET call, `poc_bit`, in which the validator provides their custody bit in their request to the beacon node. When the beacon node returns the attestation to be signed, it does so with the appropriate bit set, as provided by the validator.

specs/validator/beacon_node_oapi.yaml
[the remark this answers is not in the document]
djrtwo: Interesting.. `master` is generally more stable and contains the releases so that would be my preference here. I'll change it

specs/validator/0_beacon-node-validator-api.md
```diff
@@ -0,0 +1,28 @@
...
+
+This specification is derived from a proposal and discussion on Issues [#1011](https://github.com/ethereum/eth2.0-specs/issues/1011) and [#1012](https://github.com/ethereum/eth2.0-specs/issues/1012)
+
+
+## Specification 
+
+The API specification has been written in [OpenAPI 3.0](https://swagger.io/docs/specification/about/) and is provided in the [beacon_node_oapi.yaml](beacon_node_oapi.yaml) file alongside this document.
+
+For convenience, this specification has been uploaded to [SwaggerHub](https://swagger.io/tools/swaggerhub/) at the following URL:
+[https://app.swaggerhub.com/apis/spble/beacon_node_api_for_validator/0.1](https://app.swaggerhub.com/apis/spble/beacon_node_api_for_validator/0.1)
```
djrtwo: Should we remove the version in the url here? That makes it resolve to the latest version on swagger

spble: Yes, definitely! nice catch.

PR #1075, opened 2019-05-11
get_genesis_beacon_state is not defined in the deposit contract

jrhea:
i modified the spec to link to the function.


jrhea:
One problem with this is that it creates another maintenance point when doing a release.  You have to change the link to point to the tagged version.  Perhaps I should just mention that it is defined in: `0_beacon-chain.md`

djrtwo:
sending this to `dev` for now instead of `master`

Issue #1064, opened 2019-05-08
SSZ rework - static type hinting, minimal impl. - with POC


protolambda:
POC implementation here: https://gist.github.com/protolambda/ce2525774c7d1366b7820d4e57d865fa
You can load the POC into Google Colab (click badge on top) to play around with the code.

Credits to @JustinDrake for initial suggestions (Link: https://gist.github.com/protolambda/0cc62589bd4dba209da1b2478d39a419), and later feedback on the non-typed iteration.

Older non-typed iteration of rework: https://gist.github.com/protolambda/4dd66338a2f6214779cadb03c222171a

With this new typed approach, we have a more readable SSZ (ignoring the magic parts) than the previous `minimal_ssz.py`, and we enable static-type hints / linting / checks / IDE suggestions / etc. (as requested by @hwwhww in python call discussion)

I'm switching my efforts back to testing now. I think @JustinDrake can direct getting this into a PR form, and splitting it up. And I hope @hwwhww can review if it fulfills the typing requirements (and maybe iterate on the POC).

Example usage snippet:

```python
class Abc(SSZContainer):
  ab: List[uint64]
  x: uint32

class Foobar(SSZContainer):
  foo: Vector[Abc, SSZLen(5)]

class BeaconThing(SSZContainer):
  a: uint8
  b: Foobar
  baz: BytesN[SSZLen(3)]
  quix: bytes

x = BeaconThing()
x.b.foo[ 0].ab = [0x0123456789abcdef]
x.b.foo[ 0].x = 0x11223344
x.b.foo[4].ab = [0xdeadbeefdeadbeef, 0x3333333333333333]
x.b.foo[4].x = 0x55667788
x.baz = b'\xaa\xbb\xcc'
x.quix = b'\xff\xff'
x.a = 0xcc
print(x.serialize().hex())
print(x.hash_tree_root().hex())

y = BeaconThing()
y.a = 0x77
y.b = Foobar(foo=Vector(Abc(ab=[0x1122334455667788], x=0xaabbccdd), Abc(), Abc(x=0x12345678)))
y.quix = b'\x42\x42\x42\x42\x42\x42'
print(y.serialize().hex())

p = uint8(0xff)
print(serialize(p, uint8).hex())
q = uint16(0xaabb)
print(serialize(q, uint16).hex())
```

Note that the "magic" behind the scenes relies on:
- annotations being present
- generics being used to describe vector/list element types, and vector/bytesN sizes.
[...]

PR #835, opened 2019-03-23
RST/Sphinx docs

hwwhww:
Per discussion on #412 and #793, here's a branch to see the what would a Sphinx doc be: https://github.com/hwwhww/eth2.0-specs/tree/hwwhww/sphinx

(copied the docs setting from [ethereum-python-project-template](https://github.com/ethereum/ethereum-python-project-template).)

### Modulization
This PR includes some meta updates
1.  The `/eth2_specs` dir is copied from `/build` directory
2. Add `setup.py` to make `eth2_specs` as a Python module
3. Add `doctest` in CI

### How to setup `docs`
```sh
pip install -e .[dev,docs]

cd docs/; sphinx-build -W -T -E . _build/html
```
Then you can see the entry point would be at
`/docs/_build/html/index.html`. And the main preview page is `/docs/_build/html/core/0_beacon-chain/index.html`.

### Quick tutorial
1. The converted `.rst` file of phase 0 spec: https://github.com/hwwhww/eth2.0-specs/blob/hwwhww/sphinx/docs/core/0_beacon-chain/index.rst
2. I only updated
    - "Informative" section, using `.rst` formatting external links 
    - The helper functions part [here](https://github.com/hwwhww/eth2.0-specs/blob/hwwhww/sphinx/docs/core/0_beacon-chain/index.rst#helper-functions): to mirror the source code of a certain function, we can add:

```
.. literalinclude:: ../../../eth2_specs/phase0/spec.py
   :pyobject: <function_name>
```
#### How does it look like:

image

image

### Pros
- Nicer CI/tests environment: no need for `function_puller.py`!
- Better versioning than a pure markdown file.
- Better ToC and better reading experience.
- No matter what's the output of the CI workflow from #793, IMHO having `eth2_specs` module would make the tests much more natural.

### Cons
- The contributors will feel that they are hacking Python instead of proposing an idea; but also, everyone can open a well-documented doc in issue and Specs Elves can help implement it.
- @protolambda: it may be too far from "specification", it would be more like documentation.
- reStructuredText...

I might be biased, so I wonder what do the contributors and readers think about it?
/cc @protolambda @djrtwo @JustinDrake @vbuterin @pipermerriam

PR #1077, opened 2019-05-12
SSZ implementation for exec. spec - Support for Python 3 typing.

protolambda:
~~I'm working on tests now, and shifted away from the SSZ readability/typing ambitions after my POC presented in #1064. I hope others can pick up the SSZ typing idea now, with this code ready to transform the spec. See comments in issue.~~ Finished it (testing PRs feature-wise complete, merging coming along, I got to work on SSZ again)

What still needs work:
- ~~less hacky spec builder / type field translation~~
- ~~check that when constants are changed (config override), types adapt. Otherwise, re-introduce the re-init function we have now for old non-python type system.~~ Introduced re-init function back. May explore module-reloading as alternative later, after phase-1 spec-building is in.
- ~~fix integer/bytesN types being recognized or not. Just need to be unique declarations to work for the SSZ matching code. Otherwise no-overhead int/bytes objects. (Or maybe Dankrad wants to box them anyway for bounds checking?)~~
  - After discussion with @vbuterin, boxing just the less-used integer types seemed like the easy yet reasonable solution. We let `int` default to uint64, and don't have to worry about ugly slow integer boxes in 99% of the spec.
- ~~make it run outside of POC setting, in exec. spec. minor detail.~~  Fixed it. The `ssz_static` build output now matches that of dev, but with a type based ssz system :tada:


protolambda:
Let's don't forget about our oldest current open spec PR: #695 (from before executable-spec time :joy:). We finally get to do SSZ typing well with this PR, and have static-type checking in the spec. :tada:

protolambda:
- `is_list_type(typ)` can be `issubclass(typ, list)`, `is_vector_type(typ)` can be `issubclass(typ, Vector)`, and similar for Container
  - Although, not completely sure about `issubclass`, functions may be more readable for non-python programmers.
- `issubclass(typ, bool)` can `typ == bool` for readability in `serialize_basic`. Bools don't generally have subclasses (do we plan to?)
- type inference like `List[infer_type(obj[0])]` is not going to work. Integer typing is lost. And no annoations available sadly (reason for inferring). We can either default to `uint64`, or throw an error when someone tries to infer a bar List[uintX] or uintX type.
- `is_uint(obj.__class__)` is not going to work. NewType is not changing the `__class__` afaik
- is_list_type doesn't have to check attributes, it can do `issubclass`. Same for vector. Same for container.
- I like the `read_elem_typ` usage in hash-tree-root, abstracts away the difference between bytes and a List nicely.
- `zip(obj.get_field_values(), typ.get_field_types())` can be replaced with `obj.get_fields().items()`. Similar change for signing root
- we may want to rename `SSZContainer` to just `Container`, more consistent with `List` and `Vector`, nice and brief.
- we need to update the spec builder to just keep classes as is :)
  If we're ambitious, we can take the following steps:
    1. extract the constants from the spec (just define names + types), or define them in python style with values (comments still around for X& seconds?)
    2. import all helpers at once with a `from spec_things import *`, then manage helpers / type declarations / imports from elsewhere there.
    3. Run and edit the spec as a python notebook, converted back and forth to markdown. See https://github.com/aaren/notedown and issue #835. We should have had it during the experimentation phase months ago, to make life easy. Not sure if it is still a great time. (and if we can make it work for phase-1 modification compatibility, or still need the spec builder)

I'm pro `NewType`, as it is good for performance. I also know others are in favor of boxed integers (please make a POC whoever has the time), for nice type checking + under/overflow bounds checking, at a performance cost. Can we make a decision yet? @dankrad @hwwhww

vbuterin:
> is_list_type(typ) can be issubclass(typ, list), is_vector_type(typ) can be issubclass(typ, Vector), and similar for Container

I tried this and it actually didn't work:

```
>>> issubclass(ssz.List[ssz.uint64], list)
TypeError: issubclass() arg 1 must be a class
```

> type inference like List[infer_type(obj[0])] is not going to work. Integer typing is lost. And no annoations available sadly (reason for inferring). We can either default to uint64, or throw an error when someone tries to infer a bar List[uintX] or uintX type.

So the approach I was assuming we would take here is that plain int objects would be uint64's, and all other types of uints would need a wrapper class like `Vector`. As right now we don't even use any other kind of int anywhere, so it would not really cost much. I want to fight hard for full type inference capability :smile: 

> `zip(obj.get_field_values(), typ.get_field_types())` can be replaced with `obj.get_fields().items()`. Similar change for signing root

Great! Will change.

> we may want to rename SSZContainer to just Container, more consistent with List and Vector, nice and brief.

Agree! Will change.

hwwhww:
Questions:
1. Are you taking  `typing.List` as the SSZ `List` type?
2. Have you made the `serialize` work well with `NewType()` uint yet? The latest commit and 81cb4a2 are both broken, just wonder if the type issues are solved wrt using `NewType` instead of `class`. :)

--

* My *gut feeling* is that adding `ListMeta` and `UintMeta/Uint` would enable us to apply similar rules to deal with the types easily and more readable.
* TBH function attribute `NewType(...).byte_len` does look hacky. I’d say I’m fine with status quo if *it does work*. I’m interested in making the `UintMeta/Uint` PoC *after* the current code passes the tests. That is, I don’t think the box checks are required of "minimal" SSZ, but if the performance is not that bad, it may help increase more confidence in test generator, but that could be in other iterations.

protolambda:
@vbuterin 

> issubclass(ssz.List[ssz.uint64], list)

Right. It works for Vector, but List is very different. List is from the typing package, and more of an add-on to make a normal list appear to be typed.
```python
from typing import List, NewType

print(issubclass(List, list))
print(issubclass(List[str], list))
print(issubclass(list, List))
# print(issubclass(list, List[str])) # TypeError: Parameterized generics cannot be used with class or instance checks

uint8 = NewType('uint8', int)
print(issubclass(List[uint8], list))
# print(issubclass(list, List[uint8])) # TypeError: Parameterized generics cannot be used with class or instance checks
```
Prints all `True` in Python 3.7

I can't seem to reproduce your error however. I wonder if it's the Python version, or something else.

Here's the current Vector behavior:

```python
print(issubclass(Vector, Vector))
print(issubclass(Vector[str, 123], Vector))
print(issubclass(Vector, Vector[str, 123]))

uint8 = NewType('uint8', int)
print(issubclass(Vector[uint8, 123], Vector))
print(issubclass(Vector, Vector[uint8, 123]))
```
```txt
False
True
False
True
False
```

We may want to change the behavior of `issubclass(Vector, Vector)`, but creating an instance of `Vector`, without element type and length, is not allowed, so we don't normally see this case.

protolambda:
@hwwhww 
> Are you taking typing.List as the SSZ List type?

Sort of. We use `List[abc]` in type annotations, and those annotations are used to deal with SSZ encoding/hashing fields of our containers. However, values are just `list`, and there element type cannot be inferred (afaik) :(. We could just create another class similar to the `Vector` + `VectorMeta` metaclass hack, and have typed lists. But then we would need to do: `SSZList[Validator](validator0, validator1, ...)` when creating the list, and need to convert lists to typed lists. However, I believe one can still do something like `SSZList(foo.bar() for foo in foos)`, similar to normal `list` comprehension. And I'm writing `SSZList` here, but if don't deal with `typing.List`, we can just call ours `List`.
It would make SSZ-life a lot easier and consistent, but I don't think it's very Python at all :see_no_evil:.

> Have you made the serialize work well with NewType() uint yet? The latest commit and 81cb4a2 are both broken, just wonder if the type issues are solved wrt using NewType instead of class. :)

`uintX` NewTypes are very hard to deal with. They are not classes. They act like the `int` underneath in the type system during runtime (hence the speed boost over boxed integers). However, since e.g. the `uint8` declaration doesn't disappear, it's a unique object, and annotations still reference it during runtime, we get to use this non-class object as if it is a type, and just check type equivalence by checking the equivalence of the referenced `NewType` output. It's a hack, but gets us zero-overhead integers. At the cost of only being able to deal with these integers when the values are passed with either explicit type argument to the `serialize`/`hash_tree_root` functions or this typing is retrieved from annotations from the encompassing container.

One work-around would be to default to `uint64` for a bare integer. And then box (i.e. wrap in actual class) or remove the others. (note that we have `BytesN[123]` to deal with data of such sizes still.)
Either way, I feel the limits of Python typing here, and I'm open to suggestions.

> My gut feeling is that adding ListMeta and UintMeta/Uint would enable us to apply similar rules to deal with the types easily and more readable.

See above comment on custom `List` type. We can do it, but we're not using the default list anymore. On the other hand, we don't need a typed list everywhere, just in containers etc. So it may actually be a fine choice. (i.e. we only read the verbosity of our custom `List` in type declarations and when we create a new instance)

> TBH function attribute NewType(...).byte_len does look hacky.

It does work. But yes, it is super hacky. But the alternative of boxing integers doesn't sound great either. Maybe we can just go for the default `uint64` route, and box / remove the others? (We're only really using `uint64`, and `byte` only in a `bytes(N)` context. And then `BytesN[x]` covers all other raw data.

protolambda:
Reviewing can start, SSZ-static generator output matches the output by the dev branch with the old SSZ.

What may still need to change:
- fix `ssz_generic` (uint only really...) tests generator. Personally not a fan of this generator. It catches just ~5% of the possible SSZ bugs. Also covered by ssz-static. And has a large maintenance cost, and still needs to be feature-complete, after 4 months. (SSZ-static turned out to be much more direct to get down and work with)
- run tests. The CI is down due to devops changing limits temporarily.

hwwhww:
@protolambda
I made the `test_libs/pyspec/tests` pass on my local side. Could you check if f0ceefc  also works well for test generator?

Some ideas
1. `Container.hash_tree_root()` can be `@property`.
2. Cache the `Container.__hash__` result.

hwwhww:
@vbuterin 
I guess you are working on this PR now? The CI is back but broken with your latest commits. I think it's better to make the light green and then deal with the conflicts. Let me know if you're not on this PR. :)

protolambda:
@hwwhww I will try to solve the merge conflicts now that v0.6.2 is in, and get those tests passing again (mostly just formatting I imagine. Also need to update the format for some of the generation output in them).
About caching and hash-tree-root: We are all thinking the same here I think; SSZ-caching in some of our types would be nice. A `property` getter and some hash value are easy enough, but invalidating the cache is the real problem. I would like to see something more elegant than changing boolean values in the spec...
Let's get this PR in first, and then we can try ssz-caching separately.

protolambda:
Resolved merge conflicts, and tests pass. :tada:
There are some problems with linting however. I'll resolve the things except the ones in the ssz partials file (still work in progress code, or ready for review?)

protolambda:
@hwwhww The CI doesn't see the flake8 version in the Makefile (I updated it) since it's not in the requirements file (where the checksum is taken from), and still uses the old cached flake8, with bugged linting for type annotations. :(
Also see comment about CI + venv requirements in the other PR about merging in the contract. Similar problems there.

All issues (merge conflicts, testing, linting) resolved. Just need the partials finished + conforming to PEP-8, and we can get this merged.
I do think the partials are nice, but also a bit of scope-creep in this PR. We're not implementing SSZ-unions/nulls, yet we do partials already. @vbuterin your call when you like them to get in. PR is otherwise ready for review/merge.

protolambda:
Had to fix a dependency issue (fixed in commit 8e8ef2d), but verified that the `ssz_static` test suites generated by the old ssz implementation currently on `dev` are identical to the tests generated by the implementation in this branch. :tada:

protolambda:
@hwwhww Fixed the linting caching problem. Can we get this PR into final review, then merge?

test_libs/pyspec/eth2spec/utils/ssz/ssz_partials.py
```diff
@@ -0,0 +1,306 @@
...
+    if len(path) == 0:
+        return [root] if is_basic_type(typ) else list(ssz_leaves(obj, typ=typ, root=root).keys())
+    if path[0] == '__len__':
+        return [root * 2 + 1] if is_list_type(typ) else []
+    base = root * 2 if is_list_kind(typ) else root
+    if is_bottom_layer_kind(typ):
+        length = 1 if is_basic_type(typ) else len(obj)
+        index, _, _ = get_bottom_layer_element_position(typ, base, length, path[0])
+        return [index]
+    else:
```
djrtwo: This can be `elif/else` with jsut a final return, rather than the nesting here.

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -0,0 +1,506 @@
...
+    byte_len = 4
+
+    def __new__(cls, value, *args, **kwargs):
+        if value.bit_length() > 32:
+            raise ValueError("value out of bounds for uint16")
+        return super().__new__(cls, value)
+
+
+# We simply default to uint64. But do give it a name, for readability
+uint64 = NewType('uint64', int)
```
djrtwo: Seems prudent to create a child class with the `bit_length` bounds check like the other `uint`s

djrtwo: Is this a sacrifice to make it easier to deal with integer types in most cases?

test_libs/pyspec/eth2spec/test/sanity/test_blocks.py
```diff
@@ -91,7 +92,7 @@ def test_empty_epoch_transition(state):
 #     block = build_empty_block_for_next_slot(state)
 #     block.slot += spec.SLOTS_PER_EPOCH * 5
 #     sign_block(state, block, proposer_index=0)
-#     yield 'blocks', [block], [spec.BeaconBlock]
+#     yield 'blocks', [block], List[spec.BeaconBlock]
```
hwwhww: Any reason for commenting out this test case?

Tips: you can skip a test case with:
```python
@pytest.mark.skip(reason="<THE REASON>")
```

test_libs/pyspec/eth2spec/test/sanity/test_blocks.py
```diff
@@ -399,7 +400,7 @@ def test_historical_batch(state):
 #
 #     state_transition(state, block)
 #
-#     yield 'blocks', [block], [spec.BeaconBlock]
+#     yield 'blocks', [block], List[spec.BeaconBlock]
```
hwwhww: ditto

test_libs/pyspec/eth2spec/test/sanity/test_blocks.py
protolambda: Comment is from a previous PR. The 5 epoch long transition is horribly slow. Polishing these few tests is good for a future PR, maybe it's fast enough when we have some basic SSZ caching.

test_libs/pyspec/eth2spec/test/sanity/test_blocks.py
protolambda: see reply to ditto reference

test_libs/pyspec/eth2spec/debug/decode.py
```diff
@@ -1,28 +1,39 @@
...
-    elif isinstance(typ, list):
-        return [decode(element, typ[0]) for element in json]
-    elif isinstance(typ, str) and typ[:4] == 'byte':
-        return bytes.fromhex(json[2:])
-    elif hasattr(typ, 'fields'):
+def decode(data, typ):
+    if is_uint_type(typ):
+        return data
+    elif is_bool_type(typ):
+        assert data in (True, False)
```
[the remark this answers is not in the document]
protolambda: technically the `data` can be anything. But yes, some other parts are not checked as strictly. Let's just keep it as is.

test_libs/pyspec/eth2spec/debug/encode.py
```diff
@@ -1,27 +1,36 @@
...
+    if is_uint_type(typ):
+        if hasattr(typ, '__supertype__'):
+            typ = typ.__supertype__
+        # Larger uints are boxed and the class declares their byte length
+        if issubclass(typ, uint) and typ.byte_len > 8:
             return str(value)
         return value
-    elif typ == 'bool':
+    elif is_bool_type(typ):
         assert value in (True, False)
```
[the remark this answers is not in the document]
protolambda: ditto

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
protolambda: Previously we had `NewType` for every integer type. But those are not classes, i.e. we can't infer the type from the value, we have to get it from the annotations, and act as if the NewType reference is a type class. Then it changed to this, so we can infer a type from a value easily. We don't really use the non-uint64 types, and if we do, we may want to check their bounds more aggressively, so hence the choice to wrap them. And uint64 as default, to not box every single integer (performance + readability)

protolambda: We could make the uint types more strict in underflow/overflow checks, but that was originally an idea pushed by Dankrad, and I'd rather have him implement it his way, or pull in a library that does exactly this (fixed-size over/under-flow checked unsigned ints). This minimal form will do for now.

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -0,0 +1,525 @@
...
+        result = []
+    elif issubclass(typ, bool):
+        result = False
+    elif issubclass(typ, Vector):
+        result = typ()
+    elif issubclass(typ, BytesN):
+        result = typ()
+    elif issubclass(typ, bytes):
+        result = b''
+    elif issubclass(typ, Container):
```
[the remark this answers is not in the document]
protolambda: See 59137fd

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
protolambda: Discussed in call, uint types can be extended in a later PR when necessary.

test_libs/pyspec/eth2spec/utils/ssz/ssz_partials.py
protolambda: moved partials out of PR

specs/core/0_beacon-chain.md
```diff
@@ -599,9 +579,9 @@ We define the following Python custom types for type hinting and readability:
 | `Shard` | `uint64` | a shard number |
 | `ValidatorIndex` | `uint64` | a validator registry index |
 | `Gwei` | `uint64` | an amount in Gwei |
-| `Bytes32` | `bytes32` | 32 bytes of binary data |
-| `BLSPubkey` | `bytes48` | a BLS12-381 public key |
-| `BLSSignature` | `bytes96` | a BLS12-381 signature |
+| `Bytes32` | `Bytes32` | 32 bytes of binary data |
```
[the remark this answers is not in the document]
protolambda: See 9bdb182

protolambda:
Ran the SSZ static generators again, and they result in the same test-vectors as running the generator on the `dev` branch (`master` has different type definitions, but same ssz implementation as `dev` now).
Ready to merge, if reviewers are all good with it :+1:

protolambda:
> I’d (strongly) suggest moving the new merkle_minimal.py and ssz_partials.py to another PR so that it could get a proper review.

`merkle_minimal.py` didn't really change much. Just a few simple efficiency gains while we are modifying (and re-testing) the code around it anyway. Fixed the power-of-2 util to be even more efficient, yet readable (see https://github.com/ethereum/eth2.0-specs/issues/1115#issuecomment-495919431)

The partials implementation is moved out to the `ssz-partials` branch (based on discussion in the Tue 4 Jun call). When this is merged we'll rebase the partials branch, and open a PR for it. @vbuterin

PR #1081, opened 2019-05-13
fix lint CI race condition

djrtwo:
Lint CI requires `install_test` and was sometimes failing due to a race condition.

fixed!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,12 +102,12 @@ workflows:
   test_spec:
     jobs:
       - checkout_specs
+      - lint:
+          requires:
+            - checkout_specs
       - install_test:
           requires:
             - checkout_specs
       - test:
-          requires:
-            - install_test
-      - lint:
           requires:
             - install_test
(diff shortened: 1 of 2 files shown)


hwwhww:
`lint` job doesn't require `install_test` (`make install_test`). Instead, it installs `flake8` with `make intall_lint`.

But it seems we should make it requires `checkout_specs`.
```yaml
      - lint:
          requires:
            - checkout_specs
```

djrtwo:
neither `install_test` nor `checkout_specs` actually build the pyspec 

Here is an instance of the error. Because `test` builds the pyspec, I think we should just have it as a dependency of `test`. Thoughts?

hwwhww:
@djrtwo Ahh, you're right about that both `test` and `lint` jobs build `pyspec`. I also realized that only `install_test` does `save_cached_venv` so `flake8` installation isn't part of the stored cache... 

If we make `lint` requires `test`, the workflow becomes:

image

Sorry for the spam commits of the experiment on this branch. I was thinking about if we can utilize parallel jobs to:

image

But it seems to require reworking Makefile... so I'm inclined to keep serial workflow for now.

PR #1083, opened 2019-05-14
Added data availability proofs, earliest draft

vbuterin:
Includes description of binary fields, link to sample python implementation, and structure of extended data roots and slashings for incorrect construction. Depends on some pieces that are not fully specified yet.


djrtwo:
top of my list after spec freeze 😅

Issue #1082, opened 2019-05-13
get_crosslink_committees_at_slot is no longer valid


terencechain:
`get_crosslink_committees_at_slot` is no longer defined in the beacon chain spec, while validator spec still references it for committee assignment. We need to update the honest validator spec to use `get_crosslink_committee` based on slot

djrtwo:
on it

PR #1068, opened 2019-05-09
[WIP] Execution-ified fork choice, draft 1

vbuterin:
Not tested at all. Also missing one optimization that's probably crucial to have even enough efficiency for testing, which is a children cache.

Also missing weak-subjectivity-aware safety measurement and time-aware justified head switching.

PR #1085, opened 2019-05-14
fix up validator guide crosslink committee instructions

djrtwo:
addresses #1082

### DIFF
--- a/specs/validator/0_beacon-chain-validator.md
+++ b/specs/validator/0_beacon-chain-validator.md
@@ -148,7 +148,7 @@ def get_committee_assignment(
 
     committees_per_slot = get_epoch_committee_count(state, epoch) // SLOTS_PER_EPOCH
     epoch_start_slot = get_epoch_start_slot(epoch)
-    for slot in range(epoch_start_slot, epoch_start_slot + SLOTS_PER_EPOCH)
+    for slot in range(epoch_start_slot, epoch_start_slot + SLOTS_PER_EPOCH):
         slot_start_shard = get_epoch_start_shard(state, epoch) + committees_per_slot * (slot % SLOTS_PER_EPOCH)
         for i in range(committees_per_slot):
             shard = (slot_start_shard + i) % SHARD_COUNT
--- a/specs/validator/0_beacon-chain-validator.md
+++ b/specs/validator/0_beacon-chain-validator.md
@@ -20,6 +20,8 @@
         - [Process deposit](#process-deposit)
         - [Validator index](#validator-index)
         - [Activation](#activation)
+    - [Validator assignments](#validator-assignments)
+        - [Lookahead](#lookahead)
     - [Beacon chain responsibilities](#beacon-chain-responsibilities)
         - [Block proposal](#block-proposal)
             - [Block header](#block-header)
@@ -45,8 +47,6 @@
                 - [Aggregation bitfield](#aggregation-bitfield)
                 - [Custody bitfield](#custody-bitfield)
                 - [Aggregate signature](#aggregate-signature)
-    - [Validator assignments](#validator-assignments)
-        - [Lookahead](#lookahead)
     - [How to avoid slashing](#how-to-avoid-slashing)
         - [Proposer slashing](#proposer-slashing)
         - [Attester slashing](#attester-slashing)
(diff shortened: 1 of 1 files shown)


specs/validator/0_beacon-chain-validator.md
```diff
@@ -127,13 +127,61 @@ Once a validator is activated, the validator is assigned [responsibilities](#bea
...
+def is_proposer(state: BeaconState,
+                validator_index: ValidatorIndex) -> bool:
+    return get_beacon_proposer_index(state) == validator_index
+```
+
+*Note*: To see if a validator is assigned to proposer during the slot, the validator must run an empty slot transition from the previous state to the current slot using `process_slots(state, current_slot)`.
+
+### Lookahead
+
+The beacon chain shufflings are designed to provide a minimum of 1 epoch lookahead on the validator's upcoming committee assignments for attesting dictated by the shuffling and slot. Note that this lookahead does not apply to proposing, which must checked during the slot in question.
```
JustinDrake: "Note that this lookahead does not apply to proposing, which must checked during the slot in question." => This is no longer true with the latest `validator.effective_balance` logic.

specs/validator/0_beacon-chain-validator.md
```diff
@@ -127,13 +127,61 @@ Once a validator is activated, the validator is assigned [responsibilities](#bea
...
+
+A validator can use the following function to see if they are supposed to propose during their assigned committee slot. This function can only be run during the slot in question. Proposer selection is only stable within the context of the current epoch.
+
+```python
+def is_proposer(state: BeaconState,
+                validator_index: ValidatorIndex) -> bool:
+    return get_beacon_proposer_index(state) == validator_index
+```
+
+*Note*: To see if a validator is assigned to proposer during the slot, the validator must run an empty slot transition from the previous state to the current slot using `process_slots(state, current_slot)`.
```
JustinDrake: Transitioning to the previous slot is no longer required (see other comment).

specs/validator/0_beacon-chain-validator.md
```diff
@@ -127,13 +127,61 @@ Once a validator is activated, the validator is assigned [responsibilities](#bea
...
 
 ### Block proposal
 
-A validator is expected to propose a [`BeaconBlock`](../core/0_beacon-chain.md#beaconblock) at the beginning of any slot during which `get_beacon_proposer_index(state)` returns the validator's `validator_index`. To propose, the validator selects the `BeaconBlock`, `parent`, that in their view of the fork choice is the head of the chain during `slot - 1`. The validator is to create, sign, and broadcast a `block` that is a child of `parent` and that executes a valid [beacon chain state transition](../core/0_beacon-chain.md#beacon-chain-state-transition-function).
+A validator is expected to propose a [`BeaconBlock`](../core/0_beacon-chain.md#beaconblock) at the beginning of any slot during which `is_proposer(state, validator_index)` returns `True`. To propose, the validator selects the `BeaconBlock`, `parent`, that in their view of the fork choice is the head of the chain during `slot - 1`. The validator is to create, sign, and broadcast a `block` that is a child of `parent` and that executes a valid [beacon chain state transition](../core/0_beacon-chain.md#beacon-chain-state-transition-function).
```
JustinDrake: "To propose, the validator selects the `BeaconBlock`, `parent`, that in their view of the fork choice is the head of the chain during `slot - 1`." => This can be better formalised with the new executable fork choice (see `get_head` in #1068).

JustinDrake: "The validator is to create, sign, and broadcast a `block` that is a child of `parent` and that executes a valid [beacon chain state transition](../core/0_beacon-chain.md#beacon-chain-state-transition-function)."

"create, sign" is redundant given "that executes a valid state transition".

specs/validator/0_beacon-chain-validator.md
```diff
@@ -127,13 +127,61 @@ Once a validator is activated, the validator is assigned [responsibilities](#bea
...
+    """
+    next_epoch = get_current_epoch(state) + 1
+    assert epoch <= next_epoch
+
+    committees_per_slot = get_epoch_committee_count(state, epoch) // SLOTS_PER_EPOCH
+    epoch_start_slot = get_epoch_start_slot(epoch)
+    for slot in range(epoch_start_slot, epoch_start_slot + SLOTS_PER_EPOCH)
+        slot_start_shard = get_epoch_start_shard(state, epoch) + committees_per_slot * (slot % SLOTS_PER_EPOCH)
+        for i in range(committees_per_slot):
+            shard = (slot_start_shard + i) % SHARD_COUNT
```
hwwhww: nitpick: `slot_start_shard` is not `Shard` type since you move `(....) % SHARD_COUNT` to the inner for-loop.
I prefer to add an extra `% SHARD_COUNT` (similar to `get_beacon_proposer_index`) just for making it clear:

```python
        offset = committees_per_slot * (slot % SLOTS_PER_EPOCH)
        slot_start_shard = (get_epoch_start_shard(state, epoch) + offset) % SHARD_COUNT
        for i in range(committees_per_slot):
            shard = (slot_start_shard + i) % SHARD_COUNT
```

specs/validator/0_beacon-chain-validator.md
djrtwo: ah, right. balance _only_ changes at epoch's right?

djrtwo: gotcha. `effective_balance` is only adjusted at epoch boundaries. A few things still change `balance` but it isn't cleaned up against `effective_balance` until the boundaries.

This is a good invariant to note. from `state_A -> state_b` all `effective_balances` are unchanged unless `state_A -> state_B` is an epoch change.

specs/validator/0_beacon-chain-validator.md
djrtwo: Still does not provide a 1 epoch lookahead. It is instead just within the epoch in question

specs/validator/0_beacon-chain-validator.md
djrtwo: can mod in 1068

djrtwo: When validating, creating a block and then signing the block are separate actions. All clients have an option to run the state transition without verifying signatures because you can't have a valid signature until you've run the block transition.

I'll try to clean this up a little.

specs/validator/0_beacon-chain-validator.md
```diff
@@ -127,13 +127,61 @@ Once a validator is activated, the validator is assigned [responsibilities](#bea
...
+    return get_beacon_proposer_index(state) == validator_index
+```
+
+*Note*: To see if a validator is assigned to proposer during the slot, the validator must run an empty slot transition from the previous state to the current slot using `process_slots(state, current_slot)`.
+
+### Lookahead
+
+The beacon chain shufflings are designed to provide a minimum of 1 epoch lookahead on the validator's upcoming committee assignments for attesting dictated by the shuffling and slot. Note that this lookahead does not apply to proposing, which must checked during the slot in question.
+
+`get_committee_assignment` should be called at the start of each epoch to get the assignment for the next epoch (`current_epoch + 1`). A validator should plan for future assignments which involves noting at which future slot one will have to attest and also which shard one should begin syncing (in Phase 1+).
```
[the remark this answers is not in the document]
djrtwo: It actually should be `involves`. The intent here is that "planning involves noting". I can clean it up.

PR #1088, opened 2019-05-17
specs/networking: minimal wire protocol for testing

fjl:
This change repurposes the 'messaging' document to define a minimal wire
protocol for interoperability testing.

I think it's OK to repurpose the messaging document and redefine
envelopes because libp2p defines its own envelope format and doesn't
need the compression/encoding negotiation.

### DIFF
--- a/specs/networking/messaging.md
+++ b/specs/networking/messaging.md
@@ -1,45 +1,46 @@
-# Eth 2.0 Networking Spec - Messaging
+# Eth 2.0 Networking Spec - Minimal Messaging
 
 ## Abstract
 
-This specification describes how individual Ethereum 2.0 messages are represented on the wire.
+This specification describes how individual Ethereum 2.0 messages are represented on the
+wire.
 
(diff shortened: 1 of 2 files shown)


djrtwo:
Sanity check -- clients already with libp2p support _can_ utilize this message structure, correct?

fjl:
I think I misunderstood something in the spec structure when I wrote this down, need to recheck how this affects the libp2p-based spec. What we want is a protocol that can be implemented in a couple hours without using libp2p.

jrhea:
> I think I misunderstood something in the spec structure when I wrote this down, need to recheck how this affects the libp2p-based spec. What we want is a protocol that can be implemented in a couple hours without using libp2p.

I don't think it matters if this is compatible with libp2p right now.  We just want to accomplish the simplest possible wire protocol that enables us to accomplish an initial interop.  The first step to interop should be a static node topology with tcp connection and no discovery.  Once we have that, have shown that this base case works, then we can introduce more complexity.

jrhea:
@fjl i would argue that this IS ready to be merged.  thoughts?

djrtwo:
@fjl What;s the status here?

djrtwo:
closing this. let me know if there is anything here that still needs to be addressed

PR #1092, opened 2019-05-17
Fix #1090

### LINKED ISSUE #1090: Document all intermediate values that are signed integers
Values, like `(pivot - index)` in `get_shuffled_index`, are required to be signed (in the plain form represented there). So it would be great to document all cases of them (as some clients have the assumptions that integers are unsigned).

JustinDrake:
Avoid signed integers

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -887,7 +887,7 @@ def get_shuffled_index(index: ValidatorIndex, index_count: int, seed: Bytes32) -
     # See the 'generalized domain' algorithm on page 3
     for round in range(SHUFFLE_ROUND_COUNT):
         pivot = bytes_to_int(hash(seed + int_to_bytes(round, length=1))[0:8]) % index_count
-        flip = (pivot - index) % index_count
+        flip = (pivot + index_count - index) % index_count
         position = max(index, flip)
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -647,7 +647,7 @@ def get_previous_epoch(state: BeaconState) -> Epoch:
     Return the current epoch if it's genesis epoch.
     """
     current_epoch = get_current_epoch(state)
-    return (current_epoch - 1) if current_epoch > GENESIS_EPOCH else current_epoch
+    return GENESIS_EPOCH if current_epoch == GENESIS_EPOCH else current_epoch - 1
```
hwwhww: Tiny tiny nitpick: the previous version looks fine to me, and slightly more efficient since we expect in most cases, `current_epoch` is greater than `GENESIS_EPOCH`.

JustinDrake: I was trying to make it clear that `current_epoch - 1` is not even "considered" unless `current_epoch != GENESIS`. Not too fussed either way :)

djrtwo: So in python, the left side is never run if the conditional is False, but I think this makes it slightly clearer to be careful with the subtraction. I'm oky with as is

PR #1098, opened 2019-05-19
Set genesis_state.latest_block_header with the root of empty `BeaconBlockBody`

### LINKED ISSUE #1089: Set genesis block body root to ssz hash of the empty struct rather than empty hash
A block is not useful unless it has a way to point to its parent block, so that an importer knows how to import it. Although the spec does not define it, all block intrinsically has a block identifier. Currently our block identifier is defined by `previous_block_root`:

```
signing_root(BeaconBlockHeader {
  	slot: block.slot,
	previous_block_root: block.previous_block_root,
	state_root: self.0.state_root,
	block_body_root: if is_genesis {
		Default::default()
	} else {
		hash(&block.body)
	},
	..Default::default()
})
```

Notice the if statement just for genesis block. If, instead, we change genesis `block_body_root` to empty struct signing root, it can be avoided.

hwwhww:
Fix #1089.

I think it is a bug because before this fix, the test helper [`build_empty_block_for_next_slot`](https://github.com/ethereum/eth2.0-specs/blob/23b1cb225d4aab476bbeba06ab1d17755ccfd104/test_libs/pyspec/tests/helpers.py#L131-L132) *did not* build the correct `slot=GENESIS_SLOT+1` block which the parent block is point to [`genesis_block`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#genesis-block).

https://github.com/ethereum/eth2.0-specs/blob/23b1cb225d4aab476bbeba06ab1d17755ccfd104/test_libs/pyspec/tests/helpers.py#L131-L132

/cc @sorpaas

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1189,7 +1189,11 @@ Let `genesis_state = get_genesis_beacon_state(genesis_deposits, eth2genesis.gene
 
 ```python
 def get_genesis_beacon_state(deposits: List[Deposit], genesis_time: int, genesis_eth1_data: Eth1Data) -> BeaconState:
-    state = BeaconState(genesis_time=genesis_time, latest_eth1_data=genesis_eth1_data)
+    state = BeaconState(
+        genesis_time=genesis_time,
+        latest_eth1_data=genesis_eth1_data,
+        latest_block_header=BeaconBlockHeader(body_root=hash_tree_root(BeaconBlockBody())),
+    )
 
     # Process genesis deposits
     for deposit in deposits:


JustinDrake:
We probably also want to change `genesis_block`.

hwwhww:
@JustinDrake 
> We probably also want to change `genesis_block`. 

Doesn't SSZ default value work well for `genesis_block` now?
```python
genesis_block = BeaconBlock(state_root=hash_tree_root(genesis_state))
```
implies to
```python
genesis_block = BeaconBlock(
    slot=0,
    parent_root=ZERO_HASH,
    state_root=hash_tree_root(genesis_state),
    body=BeaconBlockBody(),
    signature=b'\x00' * 96,
)
```

hwwhww:
I'm merging this fix since it's blocking adding tests for #1068.

PR #1100, opened 2019-05-20
Added backward argument to block headers request

yglukhov:
The `backward` option is useful for obtaining the branches we're missing for some reason, but getting attestations to.

### DIFF
--- a/specs/networking/rpc-interface.md
+++ b/specs/networking/rpc-interface.md
@@ -241,7 +241,7 @@ Requests beacon block headers from the peer starting from `(start_root, start_sl
 
 The function of the `skip_slots` parameter helps facilitate light client sync - for example, in [#459](https://github.com/ethereum/eth2.0-specs/issues/459) - and allows clients to balance the peers from whom they request headers. Clients could, for instance, request every 10th block from a set of peers where each peer has a different starting block in order to populate block data.
 
-If `backward` is `1`, request the blocks preceding the requested one in reverse order. E.g. for block `10` return `[10, 9, 8, ...]`.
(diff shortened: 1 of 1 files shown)


atoulme:
Looks ok to me. LES uses the word `reverse` instead of `backward` though.

yglukhov:
I don't mind rewording to `reverse` but it seems a bit ambiguous to me, potentially meaning "reverse the order of the returned headers".

specs/networking/rpc-interface.md
```diff
@@ -225,6 +225,7 @@ Requests a list of block roots and slots from the peer. The `count` parameter MU
     start_slot: uint64
     max_headers: uint64
     skip_slots: uint64
+    backward: uint8 in {0, 1}
```
JustinDrake: Should this be of type `bool`?

yglukhov: Yes, semantically this is a `bool`. Spec-wise though I didn't find any occurrences of `bool` to date, and went kinda the same way it was done in [eth1 wire spec](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#getblockheaders-0x03). Should I change it to bool instead?

JustinDrake: We use `bool` in https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md

yglukhov: Indeed. Changed to `bool`

djrtwo:
closing in favor of  #1281

Issue #1054, opened 2019-05-06
Phase 0 state transition checklist—take 5


JustinDrake:
(See #128, #218, #322, #675 for takes 1, 2, 3, 4.)

This issue keeps track of possible phase 0 state transition function changes.

- [x] 1. (*post-freeze discussion*) **BLS12-381 standardisation**: See [here](https://github.com/pairingwg/bls_standard).
- [x] 2. (*post-freeze discussion*) **Big vs little byteorder of `hash_tree_root` inputs**: See also #556.
- [x] 3. (*post-freeze discussion*) **Performance parameters fine-tuning**: E.g. review `SLOTS_PER_EPOCH` and `SHARD_COUNT` based on benchmarks.
- [x] 4. **Issuance schedule**: Enact schedule whereby `2**21` ETH is issued per year when `2**27` ETH is at stake.
- [x] 5. <s>**Better aligned micro-incentives**: Make crosslink incentives proportional to number of epochs in a crosslink. Make matching head incentives proportional to head slot vs target slot.</s>
- [x] 6. <s>**Stateless withdrawal credentials**: Consider not storing withdrawal credentials. See #937.</s>
- [x] 7. **Remove `MIN_ATTESTATION_INCLUSION_DELAY`**: Set `MIN_ATTESTATION_INCLUSION_DELAY` to `1`.
- [x] 8. (*post-freeze discussion*) **Remove transfers**: Consider completely removing the `Transfer` operation, going beyond `MAX_TRANSFERS = 0`. We want to do a practice hard fork that add a field to an SSZ object (in this case, adding the `transfers` field to `BeaconBlockBody`).
- [x] 9. (*post-freeze discussion*) **Last slot vs first slot epoch transitions**: See #1043.
- [x] 10. <s>**Merged historical stats**: Consider merging `latest_randao_mixes`, `latest_active_index_roots`, `latest_slashed_balances` into a single SSZ object.</s>
- [x] 11. **More historical stats**: Store the last 8 months of historical lists so that historical start shards can be recomputed.
- [x] 12. **Crosslink lengths**: Fix #1034.
- [x] 13. **Bitfields/bigints/generalised indices**: See #1019.
- [x] 14 **Fine-tune object field ordering**: See #917.
- [x] 15 **`Crosslink` in `AttestationData`**: See #1044.
[...]

PR #1104, opened 2019-05-21
Active registry roots

JustinDrake:
Possible replacement of `latest_active_index_roots` with `latest_active_registry_roots` to allow for significantly more efficient light clients (see item 17 in #1054). Key changes:

* Validator information (including indices) is pre-shuffled and grouped by committee.
* Validator information relevant to light clients (`validator.effective_balance` and `validator.slashed`) is kept "close" to the validator index to avoid accessing `validator_registry`.
* Validator information (`index`, `effective_balance`, `slashed`) is compactified into a `uint64` which is friendly to SSZ packing.
* Shuffled pubkeys are available but kept separate from compact validator information. Light clients with pubkeys cached do not need to re-download and re-authenticate them.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -402,10 +402,12 @@ The types are defined topologically to aid in facilitating an executable version
 #### `ActiveRegistry`
 
 ```python
-    [{
+{
+    'committees': [{
         'pubkeys': ['bytes48'],
         'compact_validators': ['uint64'],
     }, SHARD_COUNT]
+}
 ```
 
 #### `PendingAttestation`
@@ -819,21 +821,20 @@ def get_randao_mix(state: BeaconState,
 ### `get_active_registry_root`
 
 ```python
-def get_active_registry_root(state: BeaconState) -> Bytes32:
+def get_active_registry_root(state: BeaconState, epoch: Epoch) -> Bytes32:
     """
     Return the active registry root for the current epoch.
     """
-    epoch = get_current_epoch(state)
-    active_registry = ActiveRegistry()
+    committees = ActiveRegistry().committees
     for committee_number in range(get_epoch_committee_count(state, epoch)):
         shard = (get_start_shard(state, epoch) + committee_number) % SHARD_COUNT
         for index in get_crosslink_committee(state, epoch, shard):
             validator = validator_registry[index]
-            active_registry[shard].pubkeys.append(validator.pubkey)
+            committees[shard].pubkeys.append(validator.pubkey)
             # index (top 7 bytes) + slashed (8th bit) + effective_balance (bottom 7 bits)
             compact_validator = index << 8 + validator.slashed << 7 + validator.effective_balance // GWEI_PER_ETH
-            active_registry[shard].compact_validators.append(compact_validator)
-    return hash_tree_root(active_registry)
+            committees[shard].compact_validators.append(compact_validator)
+    return hash_tree_root(committees)
 ```
 
 ### `generate_seed`
@@ -1226,7 +1227,7 @@ def get_genesis_beacon_state(deposits: List[Deposit], genesis_time: int, genesis
 
     # Populate active registry roots
     for index in range(LATEST_ACTIVE_REGISTRY_ROOTS_LENGTH):
-        state.latest_active_registry_roots[index] = get_active_registry_root(state)
+        state.latest_active_registry_roots[index] = get_active_registry_root(state, GENESIS_EPOCH)
 
     return state
 ```
@@ -1573,7 +1574,7 @@ def process_final_updates(state: BeaconState) -> None:
     state.latest_start_shard = (state.latest_start_shard + get_shard_delta(state, current_epoch)) % SHARD_COUNT
     # Set active registry root
     active_registry_index = (next_epoch + ACTIVATION_EXIT_DELAY) % LATEST_ACTIVE_REGISTRY_ROOTS_LENGTH
-    state.latest_active_registry_roots[active_registry_index] = get_active_registry_root(state)
+    state.latest_active_registry_roots[active_registry_index] = get_active_registry_root(state, next_epoch + ACTIVATION_EXIT_DELAY)
     # Set total slashed balances
     state.latest_slashed_balances[next_epoch % LATEST_SLASHED_EXIT_LENGTH] = (
         state.latest_slashed_balances[current_epoch % LATEST_SLASHED_EXIT_LENGTH]
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -402,12 +402,10 @@ The types are defined topologically to aid in facilitating an executable version
 #### `ActiveRegistry`
 
 ```python
-{
-    'committees': [
+    [{
         'pubkeys': ['bytes48'],
         'compact_validators': ['uint64'],
-    ],
-}
+    }, SHARD_COUNT]
 ```
 
 #### `PendingAttestation`
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -31,6 +31,7 @@
             - [`DepositData`](#depositdata)
             - [`BeaconBlockHeader`](#beaconblockheader)
             - [`Validator`](#validator)
+            - [`ActiveRegistry`](#activeregistry)
             - [`PendingAttestation`](#pendingattestation)
             - [`HistoricalBatch`](#historicalbatch)
         - [Beacon operations](#beacon-operations)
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -805,17 +816,24 @@ def get_randao_mix(state: BeaconState,
...
-    return state.latest_active_index_roots[epoch % LATEST_ACTIVE_INDEX_ROOTS_LENGTH]
+    epoch = get_current_epoch(state)
+    active_registry = ActiveRegistry()
+    for committee_number in range(get_epoch_committee_count(state, epoch)):
+        shard = (get_start_shard(state, epoch) + committee_number) % SHARD_COUNT
+        for index in get_crosslink_committee(state, epoch, shard):
+            validator = validator_registry[index]
+            active_registry[shard].pubkeys.append(validator.pubkey)
+            # index (top 7 bytes) + slashed (8th bit) + effective_balance (bottom 7 bits)
+            compact_validator = index << 8 + validator.slashed << 7 + validator.effective_balance // GWEI_PER_ETH
```
vbuterin: This does seem like a departure from our usual style. So far we have avoided the efficiencies of shaving off bytes even in places where they look very significant (balances modulo 1 ETH) to preserve our "purity" of just using nice clean uint64s for everything. Here, doing the same thing would only increase the byte count from 8 to 17 (8 for index, 8 for balance, 1 for slashed). What's the rationale for doing it differently here?

JustinDrake: > This does seem like a departure from our usual style.

Right. Below are my reasons (I don't feel too strongly):

1) It's a hashing saving (no individual container leaves, and four `uint64` get packed per SSZ leaf, as opposed to one `container`). This may seem like a small saving for the beacon chain, but it could be significant for ultra-light clients (e.g. Eth2.0 client in Eth1.0).
2) Light clients are "extra-consensus", in the sense that `latest_active_registry_roots ` is exposed but never read by the state transition function.

specs/core/0_beacon-chain.md
```diff
@@ -397,6 +399,15 @@ The types are defined topologically to aid in facilitating an executable version
 }
 ```
 
+#### `ActiveRegistry`
+
+```python
+    [{
```
djrtwo: I think this should be a container type with a single list field just so we can easily talk about the type

specs/core/0_beacon-chain.md
```diff
@@ -898,10 +916,10 @@ def get_shuffled_index(index: ValidatorIndex, index_count: int, seed: Bytes32) -
     return index
 ```
 
-### `compute_committee`
+### `get_committee_indices`
 
 ```python
-def compute_committee(indices: List[ValidatorIndex], seed: Bytes32, index: int, count: int) -> List[ValidatorIndex]:
+def get_committee_indices(indices: List[ValidatorIndex], seed: Bytes32, index: int, count: int) -> List[ValidatorIndex]:
```
djrtwo: This doesn't take a `BeaconState` input, so I would argue it is a `compute_` rather than `get_`

specs/core/0_beacon-chain.md
```diff
@@ -805,17 +816,24 @@ def get_randao_mix(state: BeaconState,
     return state.latest_randao_mixes[epoch % LATEST_RANDAO_MIXES_LENGTH]
 ```
 
-### `get_active_index_root`
+### `get_active_registry_root`
```
djrtwo: This name implies to me that it is going to get an active registry root from `latest_active_registry_roots` similar to the function `get_randao_mix`.

I actually think the term `compute_` is more appropriate here. We compute something to put it into the state.

specs/core/0_beacon-chain.md
```diff
@@ -805,17 +816,24 @@ def get_randao_mix(state: BeaconState,
...
+def get_active_registry_root(state: BeaconState) -> Bytes32:
     """
-    Return the index root at a recent ``epoch``.
-    ``epoch`` expected to be between
-    (current_epoch - LATEST_ACTIVE_INDEX_ROOTS_LENGTH + ACTIVATION_EXIT_DELAY, current_epoch + ACTIVATION_EXIT_DELAY].
+    Return the active registry root for the current epoch.
     """
-    return state.latest_active_index_roots[epoch % LATEST_ACTIVE_INDEX_ROOTS_LENGTH]
+    epoch = get_current_epoch(state)
+    active_registry = ActiveRegistry()
```
djrtwo: I think this _has_ to be a container type if you are going to name it and call it like this

specs/core/0_beacon-chain.md
```diff
@@ -805,17 +816,24 @@ def get_randao_mix(state: BeaconState,
...
+    for committee_number in range(get_epoch_committee_count(state, epoch)):
+        shard = (get_start_shard(state, epoch) + committee_number) % SHARD_COUNT
+        for index in get_crosslink_committee(state, epoch, shard):
+            validator = validator_registry[index]
+            active_registry[shard].pubkeys.append(validator.pubkey)
```
djrtwo: You're sorting validators by their _crosslink_ committee shard. This isn't going to be particularly useful for light clients _if_ we don't merge the two committees.

specs/core/0_beacon-chain.md
```diff
@@ -1554,11 +1571,9 @@ def process_final_updates(state: BeaconState) -> None:
...
     # Update start shard
     state.latest_start_shard = (state.latest_start_shard + get_shard_delta(state, current_epoch)) % SHARD_COUNT
-    # Set active index root
-    index_root_position = (next_epoch + ACTIVATION_EXIT_DELAY) % LATEST_ACTIVE_INDEX_ROOTS_LENGTH
-    state.latest_active_index_roots[index_root_position] = hash_tree_root(
-        get_active_validator_indices(state, next_epoch + ACTIVATION_EXIT_DELAY)
-    )
+    # Set active registry root
+    active_registry_index = (next_epoch + ACTIVATION_EXIT_DELAY) % LATEST_ACTIVE_REGISTRY_ROOTS_LENGTH
+    state.latest_active_registry_roots[active_registry_index] = get_active_registry_root(state)
```
djrtwo: This usage is incorrect. You are inserting the _current_ epoch active registry into the epoch position `ACTIVATION_EXIT_DELAY` in the future.

This is purposefully supposed to be populating the future epoch position with the future registry.

specs/core/0_beacon-chain.md
JustinDrake: Some time ago I argued along the same lines, i.e. reserve `get_` for functions that get state from `BeaconState`. I vaguely remember HW and V arguing for a laxer use of `get_` for pure functions.

I don't care either way so long as we're consistent. Given that `get_epoch_start_slot`, `get_shuffled_index`, `get_committee_indices`, `get_bitfield_bit`, `get_delayed_activation_exit_epoch` and `get_genesis_beacon_state` all have the `get_` prefix without taking a `BeaconState` input, I think the path of least resistance is to go with `get_` for pure functions :)

specs/core/0_beacon-chain.md
JustinDrake: That's a good point. Maybe we want to wait until phase 1 to put this light-client infrastructure in place.

djrtwo: The `active_index_roots` in combination with the historical randaomixes being in state was generic enough to serve proofs about the historic crosslink and persistent committees.
If we go this new path, we should either wait or intend to put both committees in the registry root.

specs/core/0_beacon-chain.md
```diff
@@ -805,17 +818,23 @@ def get_randao_mix(state: BeaconState,
...
     """
-    return state.latest_active_index_roots[epoch % LATEST_ACTIVE_INDEX_ROOTS_LENGTH]
+    committees = ActiveRegistry().committees
+    for committee_number in range(get_epoch_committee_count(state, epoch)):
+        shard = (get_start_shard(state, epoch) + committee_number) % SHARD_COUNT
+        for index in get_crosslink_committee(state, epoch, shard):
+            validator = validator_registry[index]
+            committees[shard].pubkeys.append(validator.pubkey)
+            # index (top 7 bytes) + slashed (8th bit) + effective_balance (bottom 7 bits)
+            compact_validator = index << 8 + validator.slashed << 7 + validator.effective_balance // GWEI_PER_ETH
```
```suggestion
            compact_validator = index << 8 + validator.slashed << 7 + validator.effective_balance // EFFECTIVE_BALANCE_INCREMENT
```
Same value, but it might not be in other configurations / future. So let's use the increment constant instead.

protolambda:
"pre-shuffling": could this simplify the spec more? I.e. we compute a list of shuffled active indices after the randao (or effectively the shuffling seed) is updated, and then just have the data (committees, active indices) available when we need it? Clients can already optimize it themselves, but it may actually reduce the spec complexity (if we can have this index list somewhere around...).

djrtwo:
> "pre-shuffling": could this simplify the spec more?

We actually used to store the shuffling entirely in state, but after much debate removed it to (1) reduce state size and (2) give optionality to clients esp in resource restricted environments.

JustinDrake:
Close in favour of #1219.

PR #1107, opened 2019-05-21
Avoid finalization when zero total stake

Nashatyrev:
### Problem

After applying some number of 'empty slot transitions' to any state all validators are finally exited due to inactivity penalties and the state has zero active validators. From this point the state becomes finalized due to the `>=` condition (see proposed change).
Of course this is not the case for the actual network but may cause client software issues. E.g. when the client is not synced yet but is trying calculate the state of the current slot (based on system clocks) by applying 'empty slot transitions'. When encountering a finalized epoch it may stick to this zero-balance fork and then just fail to sync correctly 

### Solution 
Change `>=` to `>` condition to avoid justifying epoch when the total stake balance is zero

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1394,7 +1394,7 @@ def process_crosslinks(state: BeaconState) -> None:
             shard = (get_epoch_start_shard(state, epoch) + offset) % SHARD_COUNT
             crosslink_committee = get_crosslink_committee(state, epoch, shard)
             winning_crosslink, attesting_indices = get_winning_crosslink_and_attesting_indices(state, epoch, shard)
-            if 3 * get_total_balance(state, attesting_indices) >= 2 * get_total_balance(state, crosslink_committee):
+            if 3 * get_total_balance(state, attesting_indices) > 2 * get_total_balance(state, crosslink_committee):
                 state.current_crosslinks[shard] = winning_crosslink
 ```
 
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1352,14 +1352,14 @@ def process_justification_and_finalization(state: BeaconState) -> None:
     previous_epoch_matching_target_balance = get_attesting_balance(
         state, get_matching_target_attestations(state, previous_epoch)
     )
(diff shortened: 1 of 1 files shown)


JustinDrake:
Side note: That the tests are still passing suggests we're missing an edge-case test for those thresholds.

djrtwo:
I'm a bit conflicted on this one. All validators exiting the chain is a failure scenario beyond the balance being equal to zero. The chain cannot even be built at this point.

Mathematically, the threshold for finality should be `>=` so I'm not certain I'm willing compromise that to handle the case where a chain that _can no longer_ be built can prevent finalizing itself.

@JustinDrake The reason this scenario is not tested is that there is no longer a chain to test here.

JustinDrake:
> All validators exiting the chain is a failure scenario beyond the balance being equal to zero.

I don't think that's the necessarily the case. The effective balances can be zero without all the validators having exited.

> handle the case where a chain that can no longer be built

I think the point is that there are scenarios where the effective balances are all zero and the chain can still be built. (Remember that effective balance is rounded down to `EFFECTIVE_BALANCE_INCREMENT`, and that there is an exit queue.)

djrtwo:
Still thinking through such a scenario. In such a case, the remaining zero-balance validators are on a death march unless new capital is deposited into the system, right?

If we need to handle this case, I would prefer the following blocker statement rather than compromise the mathematical integrity finality conditional statements.

```python
if get_total_active_balance(state) == 0:
    return
```

Nashatyrev:
My two cents: from my perspective `>=` and `>` are _almost_ equivalent for non-zero balance especially taking into account that `2/3` is kind of conceptual ratio. 
Adding another couple of `if`s looks less intrusive but makes the spec a bit more heavy 
Either way I will be pretty fine with any solution

djrtwo:
I'm pretty sure I'm going to go with the extra `if`. Taking care of it now with a test on the case.

JustinDrake:
> I'm pretty sure I'm going to go with the extra `if`.

It turns out there are other problems with rounding down the balance, namely divisions by zero in a couple places. I would want to avoid putting `if` statements everywhere. Maybe we can round the balance up instead?

djrtwo:
Closing in favor of #1123

PR #1108, opened 2019-05-21
minor fix to how eth1 data is counted in validator guide

djrtwo:
Fix minor issue in v-guide using an old method to count eth1 votes

### DIFF
--- a/specs/validator/0_beacon-chain-validator.md
+++ b/specs/validator/0_beacon-chain-validator.md
@@ -224,7 +224,7 @@ epoch_signature = bls_sign(
 
 `block.eth1_data` is a mechanism used by block proposers vote on a recent Ethereum 1.0 block hash and an associated deposit root found in the Ethereum 1.0 deposit contract. When consensus is formed, `state.latest_eth1_data` is updated, and validator deposits up to this root can be processed. The deposit root can be calculated by calling the `get_deposit_root()` function of the deposit contract using the post-state of the block hash.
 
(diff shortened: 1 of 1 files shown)


specs/validator/0_beacon-chain-validator.md
```diff
@@ -233,7 +233,7 @@ epoch_signature = bls_sign(
     * Let `deposit_root` and `deposit_count` be the deposit root and deposit count of the Eth 1.0 deposit contract in the post-state of the block referenced by `block_hash`
     * Let `best_vote_data = Eth1Data(block_hash=block_hash, deposit_root=deposit_root, deposit_count=deposit_count)`.
 * If `D` is nonempty:
-    * Let `best_vote_data` be the `eth1_data` of the member of `D` that has the highest `vote.vote_count`, breaking ties by favoring block hashes with higher associated block height.
+    * Let `best_vote_data` be the `eth1_data` member of `D` that has the highest vote count (`D.count(eth1_data)`), breaking ties by favoring block hashes with higher associated block height.
```
hwwhww: The previous sentence "Let `D` be the set of `Eth1DataVote` objects `vote` in `state.eth1_data_votes` where: ..." makes me thought `D` is a `set`, where `D.count(eth1_data)` would always be `1`. What do you think about changing to "Let `D` be the list of `Eth1DataVote` objects `vote`"?

djrtwo: You're right. will rearrange things a a bit

djrtwo:
Done @hwwhww, but the build is failing due to credit 😬

PR #1109, opened 2019-05-22
Increase length of historical vectors

JustinDrake:
Address item 11 of #1054. This stores 8-month old historical start shards so that historical committees to be recomputed (e.g. for attester slashings).

Notice this makes slashing quite harsh:

> state.validator_registry[slashed_index].withdrawable_epoch = current_epoch + EPOCHS_PER_HISTORICAL_VECTOR

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -16,7 +16,6 @@
         - [Gwei values](#gwei-values)
         - [Initial values](#initial-values)
         - [Time parameters](#time-parameters)
-        - [State list lengths](#state-list-lengths)
         - [Reward and penalty quotients](#reward-and-penalty-quotients)
         - [Max operations per block](#max-operations-per-block)
         - [Signature domains](#signature-domains)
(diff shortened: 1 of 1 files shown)


djrtwo:
Two options to reduce slashing severity

* Just use a different vector length for the slashing accumulation array
* Do not use the entire array of history for calculating the total slashed in the period.  Can make `total_at_start` be something like the following `total_at_start = state.latest_slashed_balances[(current_epoch + EPOCHS_PER_HISTORICAL_VECTOR - EPOCHS_PER_SLASHING_PERIOD) % EPOCHS_PER_HISTORICAL_VECTOR]`

vbuterin:
Agree that we don't need to make slashing lead to an 8 month delay; probably best to keep the current ~36 days.

djrtwo:
@JustinDrake What route do you want to go here?

JustinDrake:
> What route do you want to go here?

"Just use a different vector length for the slashing accumulation array" sounds fine :) (I'm at a zkSNARKs workshop this week—do you mind going ahead on this one?)

djrtwo:
on it

djrtwo:
closing in favor of #1196

PR #1113, opened 2019-05-23
Attestation block root is the truncated hash of block header, not the block itself

sorpaas:


djrtwo:
`signing_root(block_header) == signing_root(block)`. This is a nice property of `hash_tree_root` -- that an field can be replaced with it's hash-tree-root and the root of the outer object remains the same. So the replacement of `body` with `block_body_root` in the block header keeps the root the same.

sorpaas:
Thanks for clarifying!

PR #1114, opened 2019-05-23
Finish state tests intro

protolambda:
Introduces:
- test-format docs for block operations
- epoch processing test generator (unified py tests)
- sanity test generator (unified py tests)

TODO (draft):
- [x] epoch processing crosslinks in mainnet config
- [x] test-format docs for new generators
- [x] sanity tests may still have a few inconsistently formatted tests, to be updated


specs/test_formats/operations/README.md
```diff
@@ -2,9 +2,39 @@
...
+
+A handler of the `operations` test-runner should process these cases, 
+ calling the corresponding processing implementation.
+
+Operations:
+
+| *`operation-name`*      | *`operation-object`* | *`input name`*       | *`processing call`*                                    |
+|-------------------------|----------------------|----------------------|--------------------------------------------------------|
+| `attestation`           | `Attestation`        | `attestation`        | `process_deposit(state, attestation)`                  |
+| `attester_slashing`     | `AttesterSlashing`   | `attester_slashing`  | `process_deposit(state, attester_slashing)`            |
```
djrtwo:
```suggestion
| `attester_slashing`     | `AttesterSlashing`   | `attester_slashing`  | `process_attester_slashing(state, attester_slashing)`            |
```

specs/test_formats/operations/README.md
```diff
@@ -2,9 +2,39 @@
...
+## Condition
+
+A handler of the `operations` test-runner should process these cases, 
+ calling the corresponding processing implementation.
+
+Operations:
+
+| *`operation-name`*      | *`operation-object`* | *`input name`*       | *`processing call`*                                    |
+|-------------------------|----------------------|----------------------|--------------------------------------------------------|
+| `attestation`           | `Attestation`        | `attestation`        | `process_deposit(state, attestation)`                  |
```
djrtwo:
```suggestion
| `attestation`           | `Attestation`        | `attestation`        | `process_attestation(state, attestation)`                  |
```

specs/test_formats/sanity/README.md
```diff
@@ -0,0 +1,7 @@
+# Sanity tests
```
djrtwo: I'm not certain it is worth calling these "sanity" tests. `blocks` tests are what eth1 calls "blockchain" tests and will likely in the end have the most test cases as they test the entire state transition.

djrtwo: maybe just `state_transition` as the dir here and have state transitions triggered by blocks and those triggered by slots. (as is just rename).

The initial blocks tests are the `sanity` suite, but we might have many more blocks suites

test_libs/pyspec/eth2spec/test/epoch_processing/test_process_crosslinks.py
```diff
@@ -103,19 +104,24 @@ def test_single_crosslink_update_from_previous_epoch(state):
 
 @spec_state_test
 def test_double_late_crosslink(state):
+    if spec.SLOTS_PER_EPOCH < spec.SHARD_COUNT:
```
djrtwo: This isn't the proper condition here. should be

```suggestion
    if get_epoch_committee_count(state, get_current_epoch(state)) < spec.SHARD_COUNT:
```
```

File: test_libs/pyspec/eth2spec/test/epoch_processing/test_process_registry_updates.py [hunk 5]
```diff
@@ -19,15 +20,12 @@ def test_activation(state):
...
     for _ in range(spec.ACTIVATION_EXIT_DELAY + 1):
-        block = next_epoch(state)
-        blocks.append(block)
+        next_epoch(state)
+
+    yield 'pre', state
 
-    # provide extra type hinting here, since it is wrapped in a list.
-    yield 'blocks', blocks, [spec.BeaconBlock]
+    process_registry_updates(state)
```
djrtwo: It seems a little bit artificial to run `process_registry_updates` on a state that has not yet undergone the functions in the slot processing prior to it. (`process_slot` and the three preceding  functions in `process_epoch`).

Without getting the state to this mid-point in the epoch transition, many things we expect to happen might not. 

I would consider adding a `prepare_state_for_registry_update` helper that is called before yielding the `pre` state. This function would step up to the next epoch transition slot and trigger the preceding function calls.

specs/test_formats/sanity/README.md
protolambda: I think it's nice to have sanity apart from the others, as one may run these before anything else, to catch obvious problems first.
`slots` and `blocks` pretty much direct the format, like a handler should. It's not the intention to put future blockchain tests in there.

We can easily create a `state_transition` runner, with a handler that does the same (and possibly split that up into more descriptive handlers, e.g. `slow_blocks`, `orphan_blocks`, `slashable_blocks` and so on.)

test_libs/pyspec/eth2spec/test/epoch_processing/test_process_registry_updates.py
protolambda: It's more similar to crosslinks now (see 73f0f74), with a `run_process_registry_updates` function

specs/test_formats/epoch_processing/README.md
```diff
@@ -0,0 +1,34 @@
+# Epoch processing tests
+
+The different epoch sub-transitions are tested individually with test handlers.
+The format is similar to block-processing state-transition tests.
+There is no "change" factor however, the transitions are a pure functions with just the pre-state as input.
```
hwwhww:
```suggestion
There is no "change" factor however, the transitions are pure functions with just the pre-state as input.
```

test_generators/epoch_processing/main.py
```diff
@@ -0,0 +1,38 @@
...
+)
+
+from gen_base import gen_runner, gen_suite, gen_typing
+from gen_from_tests.gen import generate_from_tests
+from preset_loader import loader
+from eth2spec.phase0 import spec
+
+
+def create_suite(transition_name: str, config_name: str, get_cases: Callable[[], Iterable[gen_typing.TestCase]]) \
+        -> Callable[[str], gen_typing.TestSuiteOutput]:
```
hwwhww: nitpick: IMO avoid `\` here
```python
def create_suite(
        transition_name: str,
        config_name: str,
        get_cases: Callable[[],
        Iterable[gen_typing.TestCase]]) -> Callable[[str], gen_typing.TestSuiteOutput]:
```

test_generators/operations/main.py
```diff
@@ -8,9 +10,29 @@
...
 
-from gen_base import gen_runner
+from gen_base import gen_runner, gen_suite, gen_typing
+from gen_from_tests.gen import generate_from_tests
+from preset_loader import loader
+from eth2spec.phase0 import spec
+
+
+def create_suite(operation_name: str, config_name: str, get_cases: Callable[[], Iterable[gen_typing.TestCase]]) \
+        -> Callable[[str], gen_typing.TestSuiteOutput]:
```
hwwhww: ditto
```python
def create_suite(
        transition_name: str,
        config_name: str,
        get_cases: Callable[[],
        Iterable[gen_typing.TestCase]]) -> Callable[[str], gen_typing.TestSuiteOutput]:
```

test_generators/sanity/main.py
```diff
@@ -0,0 +1,35 @@
...
+            test_cases=get_cases()))
+    return suite_definition
+
+
+if __name__ == "__main__":
+    gen_runner.run_generator("sanity", [
+        create_suite('blocks', 'minimal', lambda: generate_from_tests(test_blocks)),
+        create_suite('blocks', 'mainnet', lambda: generate_from_tests(test_blocks)),
+        create_suite('slots', 'minimal', lambda: generate_from_tests(test_slots)),
+        create_suite('slots', 'mainnet', lambda: generate_from_tests(test_slots)),
```
hwwhww: There are repeated `'minimal'` and `'mainnet'` strings in the PR. Maybe these two strings can be defined as constants. For example: add `test_generators/constants.py` to define `MINIMAL_TESTS_NAME = 'minimal'` and `MAINNET_TESTS_NAME = 'mainnet'`.

test_libs/pyspec/eth2spec/test/epoch_processing/test_process_crosslinks.py
```diff
@@ -103,19 +104,24 @@ def test_single_crosslink_update_from_previous_epoch(state):
...
     fill_aggregate_attestation(state, attestation_1)
 
-    # add attestation_1 in the next epoch
+    # add attestation_1 to next epoch
     next_epoch(state)
     add_attestation_to_state(state, attestation_1, state.slot + 1)
 
     for slot in range(spec.SLOTS_PER_EPOCH):
-        attestation_2 = get_valid_attestation(state, signed=True)
+        attestation_2 = get_valid_attestation(state, signed=False)
```
hwwhww: Follow up with my previous comment: https://github.com/ethereum/eth2.0-specs/pull/1102#issuecomment-494232533
I just figured out you were replying to me in the third paragraph? https://github.com/ethereum/eth2.0-specs/pull/1102#issuecomment-494522605 

I think I failed to explain what I meant! 😅  Sorry, I meant making it more explicit! 

Either we can
1. have `def get_valid_attestation(state, signed, slot=None)`, so it's restricted to set signed *every time* you call the function; or
2. have `def get_valid_attestation(state, slot=None, signed=False)`, and make sure it's really clear that "default value is `False`". In that case, IMO having both `get_valid_attestation(state, signed=False)` and `get_valid_attestation(state)` exist are confusing. We should never set `signed=False` when calling it.

test_libs/pyspec/eth2spec/test/epoch_processing/test_process_registry_updates.py
```diff
@@ -5,10 +5,34 @@
...
+    # transition state to slot before state transition
+    slot = state.slot + (spec.SLOTS_PER_EPOCH - state.slot % spec.SLOTS_PER_EPOCH) - 1
+    block = build_empty_block_for_next_slot(state, signed=False)
+    block.slot = slot
+    sign_block(state, block)
+    state_transition(state, block)
+
+    # cache state before epoch transition
+    spec.cache_state(state)
+
```
djrtwo: My gut is to run the following functions prior to yield `pre` to get the state to the exact point in the normal state transition. What do you think?

```python
    process_justification_and_finalization(state)
    process_crosslinks(state)
    process_rewards_and_penalties(state)
```

specs/test_formats/sanity/README.md
djrtwo: works for me

test_generators/epoch_processing/main.py
protolambda: My IDE gives a PEP8 warning with that suggestion.

protolambda: And formatter adds 20 spaces of indentation to the last line of that suggestion, and then it's happy... I'd rather keep it as-is.

test_generators/sanity/main.py
protolambda: these are the names of the configuration files, and I don't see them changing. Nor do I like importing for just that really.

test_libs/pyspec/eth2spec/test/epoch_processing/test_process_crosslinks.py
protolambda: Option 2 then.

test_libs/pyspec/eth2spec/test/epoch_processing/test_process_registry_updates.py
protolambda: Have we formalized dependencies in any way? Some clients process them in different order / parallel.

test_libs/pyspec/eth2spec/test/epoch_processing/test_process_crosslinks.py
protolambda: See 
b12031b

test_generators/operations/main.py
protolambda: See above comment

test_libs/pyspec/eth2spec/test/epoch_processing/test_process_registry_updates.py
djrtwo: We have not, but epoch processing cannot in general be processed out of order. Rewards/penalties can be processed in parallel and most of the block operations can be validated in parallel but not fully processed.

`process_registry_updates` certainly has a dependency on rewards/penalties (which has a dependency on justification/finality/crosslinks) due to the ejection condition `validator.effective_balance <= EJECTION_BALANCE`.

In isolation, it is not strictly incorrect to not run the prior functions in this test, but it is creating a more contrived state rather than what would actually be seen in the wild

PR #1116, opened 2019-05-23
Clean up `process_penalties_and_exits` and fix ToC

hwwhww:
Replace #1066 since things are different after #1035.
(edit: false alarm)

### DIFF
--- a/specs/core/1_custody-game.md
+++ b/specs/core/1_custody-game.md
@@ -699,6 +699,6 @@ Append this to `process_final_updates(state)`:
     )
     for index, validator in enumerate(state.validator_registry):
         if index not in validator_indices_in_records:
-            if validator.exit_epoch != FAR_FUTURE_EPOCH:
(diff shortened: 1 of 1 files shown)


specs/core/1_custody-game.md
```diff
@@ -689,28 +699,6 @@ Append this to `process_final_updates(state)`:
     )
     for index, validator in enumerate(state.validator_registry):
         if index not in validator_indices_in_records:
-            if validator.exit_epoch != FAR_FUTURE_EPOCH and validator.withdrawable_epoch == FAR_FUTURE_EPOCH:
+            if validator.exit_epoch != FAR_FUTURE_EPOCH:
```
```suggestion
            if validator.exit_epoch != FAR_FUTURE_EPOCH and validator.withdrawable_epoch == FAR_FUTURE_EPOCH:
```

I don't think this line should be changed. The problem with removing `and validator.withdrawable_epoch == FAR_FUTURE_EPOCH` is that an exited validator's withdrawal epoch keeps being pushed out each epoch.

hwwhww: You're right, I'm dizzy now. 😵
There's only suspense, not delay in `process_*_challenge` functions.

PR #1117, opened 2019-05-23
update version used in test generators to get SHA-256 hash

ralexstokes:
this is blocked on the new release of `py_ecc` but should fix the issue w/ BLS tests having the wrong hash function


djrtwo:
Let's merge this to v06x to be released into master in next 24 hours

ralexstokes:
looks like it will be easiest to just cherry pick this commit onto `v06x` if that's best

PR #1122, opened 2019-05-24
Simplify deposits

JustinDrake:
Remove unnecessary index field from `Deposit` while retaining the requirement that deposits are processed in order.


specs/core/0_beacon-chain.md
```diff
@@ -1763,12 +1761,11 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
         leaf=hash_tree_root(deposit.data),
         proof=deposit.proof,
         depth=DEPOSIT_CONTRACT_TREE_DEPTH,
-        index=deposit.index,
+        index=state.deposit_index,
```
ralexstokes: we can now arbitrarily reorder unprocessed deposits as long as they are equivalent under `index // (2**i) % 2:` in the verification calculation... do we trust validators to present deposits in chronological order?

djrtwo:
> if we merge this we should remove the index from the log in the deposit contract...

Maybe? The client still needs to hold a local tree of all deposits so the index might be slightly useful?

ralexstokes:
> Maybe? The client still needs to hold a local tree of all deposits so the index might be slightly useful?

@djrtwo i guess this comes down to how strongly we want to have the `index` be part of the protocol -- if we take it out, clients will likely still have the index locally, it just won't be required for the deposit contract

PR #1126, opened 2019-05-27
Sort attester slashing indices to avoid arbitrary ordering

### LINKED ISSUE #1125: Attester slashings are done in an arbitrary order and do not commute
While testing Lighthouse against the [new test vectors](https://github.com/ethereum/eth2.0-spec-tests) I encountered a discrepancy in the attester slashing tests which seems to arise from the spec iterating over the indices to slash in the arbitrary order used by Python's hashset implementation. The buggy line is:

```python
for index in set(attesting_indices_1).intersection(attesting_indices_2):
```

in `process_attester_slashing`.

If slashings commuted the order could be arbitrary, but due to the `exit_epoch` being based upon the churn limit (and therefore the validators who have been previously slashed), you get different `exit_epoch`s and different final states for different slashing orders. One test that exemplifies this is `success_double` (mainnet) which slashes `[2, 143, 287, 353, 464, 475, 487, 496]`, which in Python set ordering is `{353, 2, 487, 143, 464, 496, 475, 287}`.

I think the best solution is to specify that the indices should be sorted before iteration, but I'm open to other ideas.

michaelsproul:
Closes #1125

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1707,7 +1707,7 @@ def process_attester_slashing(state: BeaconState, attester_slashing: AttesterSla
     slashed_any = False
     attesting_indices_1 = attestation_1.custody_bit_0_indices + attestation_1.custody_bit_1_indices
     attesting_indices_2 = attestation_2.custody_bit_0_indices + attestation_2.custody_bit_1_indices
-    for index in set(attesting_indices_1).intersection(attesting_indices_2):
+    for index in sorted(set(attesting_indices_1).intersection(attesting_indices_2)):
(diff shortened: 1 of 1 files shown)


dankrad:
This looks like a good solution. Have we checked if we are iterating over dicts/sets anywhere else where it could cause problems?

djrtwo:
Nice catch!

@dankrad I have a PR  in the works that turns some of the things in the spec to `set`s to make it explicit that order does not matter. I'll do the opposite check too to make sure that order is preserved where necessary.

michaelsproul:
If there's going to be a v0.6.2 release could it include this patch?

djrtwo:
Releasing it in v0.6.3!

PR #1127, opened 2019-05-27
Move deposit contract back

### LINKED ISSUE #980: Move deposit contract back to this repo
Deposit contract was moved to another [repo](https://github.com/ethereum/deposit_contract) because we wanted to make eth2.0-specs pure with just documents. But after some recent updates (thanks to @protolambda and @djrtwo), now `eth2.0-specs` repo is the CI/test-gen center. It seems moving deposit contract back can make the spec-syncing and release easier.

hwwhww:
Fix #980

### Changelogs
- Copy the contract and tests code from deposit contract repo https://github.com/ethereum/deposit_contract/commit/fad5c329bde664510132a2179560a0ecbc2fbeb3
- The tests of deposit contract require pyspec's `minimal_ssz` and `DepositData`.
- Update Makefile:
    - Add `install_deposit_contract_test` command to install the testing requirements of `deposit_contract`
    - Add `test_deposit_contract` command to run the `deposit_contract` tests
- Update CircleCI config:
    - Add `restore_default_cached_venv` and `save_default_cached_venv` to avoid setting `venv_name` and `reqs_checksum` in multiple places.

### TODOs
- Weird dependency issue
    - Currently, commit https://github.com/ethereum/eth2.0-specs/commit/3e6f7a2e961183b0f1d65e4bf1a7f694f24574e9 fails in CI but works when executing `make clean && make pyspec && make install_test && make install_deposit_contract_test && make test_deposit_contract` on my local side.
    - Moreover, it works in CI after I added https://github.com/ethereum/eth2.0-specs/commit/0ec03db0fb9480e19d6b40098cf3b141fdc4ac15 which I inserted `pip3 install -r requirements-testing.txt` in `test_deposit_contract` command.
    - It looks like something is wrong when I dealt with the local dependencies+cache between CircleCI jobs. @protolambda do you have any insight on it? How would `test_generators` deal with it in CI?
- Add README
- Create aliases for `venv_name` and `reqs_checksum` in CiricleCI config?


.circleci/config.yml
```diff
@@ -111,3 +126,7 @@ workflows:
       - lint:
           requires:
             - test
+      - deposit_contract:
+          requires:
+            - install_env
+            - lint
```
djrtwo: Why require `lint`?

Makefile
```diff
@@ -46,6 +47,17 @@ lint: $(PY_SPEC_ALL_TARGETS)
...
 
+install_deposit_contract_test:
+	cd $(PY_SPEC_DIR); python3 -m venv venv; . venv/bin/activate; \
+	cd ../..; cd $(DEPOSIT_CONTRACT_DIR); \
+	pip3 install -r requirements-testing.txt
+
+test_deposit_contract: $(PY_SPEC_ALL_TARGETS)
+	cd $(PY_SPEC_DIR); python3 -m venv venv; . venv/bin/activate; \
+	cd ../.. && cd $(DEPOSIT_CONTRACT_DIR); \
+	pip3 install -r requirements-testing.txt; \
```
djrtwo: Is this necessary? Is the intention to make it fail safe?

Right now, `install_test` and `test` are independent. Maybe we should ask @protolambda his logic here.

deposit_contract/contracts/validator_registration.json
```diff
@@ -0,0 +1 @@
...805160200180610fe0828460006004600a8704601201f161106257600080fd5b505060606110c06024638067328961104052610e00516110605261105c6000305af161108d57600080fd5b6110e0805160200180611120828460006004600a8704601201f16110b057600080fd5b5050610e60516111e05260606111a0526111a05161120052610fe08051602001806111a0516111e001828460006004600a8704601201f16110f057600080fd5b50506111a0516111e0015160206001820306601f82010390506111a0516111e00161118081516020818352015b836111805110151561112e5761114b565b6000611180516020850101535b815160010180835281141561111d575b5050505060206111a0516111e0015160206001820306601f82010390506111a05101016111a0526111a051611220526111208051602001806111a0516111e001828460006004600a8704601201f16111a257600080fd5b50506111a0516111e0015160206001820306601f82010390506111a0516111e00161118081516020818352015b83611180511015156111e0576111fd565b6000611180516020850101535b81516001018083528114156111cf575b5050505060206111a0516111e0015160206001820306601f82010390506111a05101016111a0527f08b71ef3f1b58f7a23ffb82e27f12f0888c8403f1ceb0ea7ea26b274e2189d4c6111a0516111e0a160016004555b5b005b63845980e8600051141561127c57341561126f57600080fd5b60045460005260206000f3005b60006000fd5b61018561140703610185600039610185611407036000f3"}
```
djrtwo: I've been wondering if instead of including this in the repo, we should build it dynamically. upon testing.

At the very least, we should add a command for building this in the `Makefile`

deposit_contract/tests/contracts/test_deposit.py
```diff
@@ -0,0 +1,242 @@
...
+from eth2spec.utils.minimal_ssz import (
+    SSZType,
+    hash_tree_root,
+)
+from eth2spec.phase0.spec import (
+    DepositData,
+)
+
+
+def hash(data):
```
djrtwo: Should probably import `hash` from `spec` and use that instead

deposit_contract/tests/contracts/utils.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def get_deposit_contract_code():
+    file_path = os.path.join(DIR, './../../contracts/validator_registration.v.py')
+    deposit_contract_code = open(file_path).read()
+    return deposit_contract_code
+
+
+def get_deposit_contract_json():
+    file_path = os.path.join(DIR, './../../contracts/validator_registration.json')
```
djrtwo: If we remove the ABI from the repo, we can build this dynamically at the start of testing instead

deposit_contract/tool/compile_deposit_contract.py
```diff
@@ -0,0 +1,33 @@
...
+import os
+
+from vyper import (
+    compiler,
+)
+
+DIR = os.path.dirname(__file__)
+
+
+def generate_compiled_json(file_path: str):
```
djrtwo: Lets add a target in `Makefile` for this

deposit_contract/tests/contracts/utils.py
hwwhww: I see the benefits of having ABI in the repo are:
1. Convenience: the client devs can just use the ABIs from the repo without compiling.
2. Versioning: to make sure we are using the consistent Vyper version.

deposit_contract/contracts/validator_registration.json
hwwhww: I see the benefits of having ABI in the repo are:
1. Convenience: the client devs can just use the ABIs from the repo without compiling.
2. Versioning: to make sure we are using the consistent Vyper version.

Added the command in `Makefile`.

.circleci/config.yml
hwwhww: To avoid the race condition of `venv` :'(

I will think more about how to make it in parallel...

Makefile
hwwhww: As I mentioned in the PR description, if we remove this line, the CI would throw:
```sh
==================================== ERRORS ====================================
_______________ ERROR collecting tests/contracts/test_deposit.py _______________
ImportError while importing test module '/home/circleci/specs-repo/deposit_contract/tests/contracts/test_deposit.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/contracts/test_deposit.py:20: in <module>
    from eth2spec.phase0.spec import (
E   ModuleNotFoundError: No module named 'eth2spec.phase0.spec'
!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
=========================== 1 error in 1.15 seconds ============================
Makefile:56: recipe for target 'test_deposit_contract' failed
make: *** [test_deposit_contract] Error 2
Exited with code 2
```

And if we added it, when executing this line, basically most packages will be "Requirement already satisfied" instead of `pyspec` (see: https://circleci.com/gh/ethereum/eth2.0-specs/2681?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

I guess between the different CircleCI jobs, the local requirement would be lost. Right now having this line doesn't increase too much overhead since all other packages are detected to "installed".

djrtwo:
Do we want to get this merged or is there anything else you want to handle in this PR?

protolambda:
> @protolambda do you have any insight on it? How would test_generators deal with it in CI?

The CI caching problems may be because of the formatting of the cache string. (two checksums to pop through when finding the cache). It's tempting to just copy over the dependencies from the basic requirements file tot the requirements-testing, instead of importing it from within. Then we can just use the standard single-checksum cache key. If you like we can try it in this PR.

Test-generators don't run in CI. Mainnet tests are big, and writing half a GB worth of yaml is not worth it every push to CI. Instead, state-transition test generation is unified with testing; when testing, it runs in a pytest context, and doesn't output yaml.

hwwhww:
@protolambda 
> The CI caching problems may be because of the formatting of the cache string. (two checksums to pop through when finding the cache).

Do you mean in `reqs_checksum` (I just rewrote it a little bit in https://github.com/ethereum/eth2.0-specs/pull/1127/commits/6617f891462c0d0ed674585f8402d2e06778f56a )? (3 checksums)
```sh
reqs_checksum: cache-v1-{{ checksum "test_libs/pyspec/requirements.txt" }}-{{ checksum "test_libs/pyspec/requirements-testing.txt" }}-{{ checksum "deposit_contract/requirements-testing.txt" }}
```

Sorry I'm not sure why it causes the problem since they are all static txt files, and what do you mean by "just copy over the dependencies from the basic requirements file"? Could you give an example? 😃

And @djrtwo @protolambda 
I'd say we can merge this PR first since it's blocking #1129, and then make the CI improvements in other PRs.

protolambda:
@hwwhww I meant the opposite. Simplification.

That is, change the current `requirements-testing.txt`:
from:
```yaml
-r requirements.txt
pytest>=3.6,<3.7
../config_helpers
```
to:
```yaml
eth-utils>=1.3.0,<2
eth-typing>=2.1.0,<3.0.0
pycryptodome==3.7.3
py_ecc>=1.6.0
typing_inspect==0.4.0
pytest>=3.6,<3.7
../config_helpers
```
and have one less checksum to worry about, as caching problems seem to be worse than copying over a few dependencies rarely.

> `reqs_checksum: cache-v1-{{ checksum "test_libs/pyspec/requirements.txt" }}-{{ checksum "test_libs/pyspec/requirements-testing.txt" }}-{{ checksum "deposit_contract/requirements-testing.txt" }}`

If it's a different venv, we should NOT include its checksum in the cache key. We're effectively taking the hash of the version numbers to version the cache of the downloaded dependencies of those version numbers. If the deposit contract requirements are used in a separate venv, then they should be cached separately.

Also, when incrementing a cache version number, it would look like:
`v2-pyspec`, not `v1-pyspec-05` (why two versions?). ([circle CI docs here](https://circleci.com/docs/2.0/caching/#restoring-cache))

Also note that the key is prefix based, so incrementing to `v2` forces the cache out. While changing a later part, tries to retrieve a cache keyed with an equal prefix (I think this is the source of our caching problems. Not 100% sure).

.circleci/config.yml
```diff
@@ -59,26 +72,19 @@ jobs:
     steps:
       - restore_cache:
           key: v1-specs-repo-{{ .Branch }}-{{ .Revision }}
-      - restore_cached_venv:
-          venv_name: v1-pyspec-03
-          reqs_checksum: '{{ checksum "test_libs/pyspec/requirements.txt" }}-{{ checksum "test_libs/pyspec/requirements-testing.txt" }}'
+      - restore_default_cached_venv
```
protolambda: 1. Let's not mingle caches of different venvs, there's no "default"
2. Let's reduce the checksum complexity to 1 hash, and follow the standard
3. Let's don't use double version tickers in the caching key
4. Multiple parts split by a `-` are likely the root of the cache-key problems. See point 2.

Caching is a nightmare sometimes, but we can get this right :)
Marking this PR is blocked for now, caching is not quite ready to merge currently.

hwwhww:
@protolambda 

Agreed and disagreed some points. 😄🤓

### tl;dr
1. Making it in single checksum doesn't seem to solve the original issue.
2. We can consider using Pipenv to manage the packages instead of `-r requirements.txt`. (but not in this PR)
3. Open to discuss if we want deposit contract has its own `venv` or not.
4. Can I just (i) rename `restore_default_cached_venv` to something else (ii) remove `cache-v1-` from `reqs_checksum` and then we can merge this PR?

---

> and have one less checksum to worry about, as caching problems seem to be worse than copying over a few dependencies rarely.

1. But in the current cache key, you already include (i) `checksum "test_libs/pyspec/requirements.txt"` and (ii) `checksum "test_libs/pyspec/requirements-testing.txt"`. Sorry, could you point out *why* two checksums of the static files would cause problem? I don’t understand what’s the fundamental difference between your proposal and the current `dev` branch setting.
2. I tried to test with the single checksum by simply turning off other jobs: https://github.com/hwwhww/eth2.0-specs/commit/a0e13cd66f005491ff8475e70ad20fad2ea98319
The issue is still as same as https://github.com/ethereum/eth2.0-specs/pull/1127/commits/3e6f7a2e961183b0f1d65e4bf1a7f694f24574e9: https://circleci.com/gh/hwwhww/eth2.0-specs/191?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
3. By the way, I think `../config_helpers` itself might cause some problem here since we are not aware of if `../config_helpers` were changed.

> If it’s a different venv, we should NOT include its checksum in the cache key. We’re effectively taking the hash of the version numbers to version the cache of the downloaded dependencies of those version numbers. If the deposit contract requirements are used in a separate venv, then they should be cached separately.

`deposit_contract` is sharing the same `venv` in this PR.

The ideal CI workflow in my mind would be:
```
* checkout_specs
    * install pyspec env
        * test pyspec
        * install deposit contract env
            * test deposit contract
    * install lint env
        * linter check
```

But unfortunately, due to the race condition (#1081), all jobs are executed sequentially… :’(

So that’s what we have now:
```
* checkout_specs
    * install all env (pyspec, linter, deposit contract)
        * test pyspec
            * linter check
                * test deposit contract
```

If we want to make deposit_contract in another `venv`, the CI workflow would be like:
```
* checkout_specs
    * install all env (pyspec, linter)
        * test pyspec
            * linter check
* checkout_specs
    * install pyspec (requirements.txt only)
        * install deposit contract env
            * test deposit contract
```
^^^^ IMO That would increase some redundant jobs and make CircleCI config more complicated.

Do you feel strongly that it should be in separate `venv`? I'm in inclined to use single `venv` unless the race condition issue has been solved.  But open to discuss. 😄

> 1. Let’s not mingle caches of different venvs, there’s no “default”

Ohh the `*_default_*` commands are just for making the keywords in one place as we only have one `venv`. Assume if we keep only one `venv`. Do you feel okay if I rename it to `restore_configurated_cached_venv`?

> 2. Let’s reduce the checksum complexity to 1 hash, and follow the standard

I can see how “just copy over the dependencies from the basic requirements file” make it in one checksum (although I think it doesn't solve the problem above 😄). But anyway, if we want to keep deposit contract in the same `venv`, there will still be some more dependencies.

Here are some recommendations from CircleCI doc:
1. https://circleci.com/docs/2.0/caching/#using-a-lock-file: 
An alternative is to do `ls -laR your-deps-dir > deps_checksum` and reference it with `{{ checksum “deps_checksum” }}`. For example, in Python, to get a more specific cache than the checksum of your `requirements.txt` file you could install the dependencies within a virtualenv in the project root `venv` and then do `ls -laR venv > python_deps_checksum`.
    * I found it’s weird that the result includes timestamps, not sure if it can be used in our case, but I can see the point of "dumping the dependencies in one file and calculate the checksum of it".

2. Or using [Pipenv](https://circleci.com/docs/2.0/caching/#pip-python)  to manage the package. I tried a little bit, it can install the with local dependency. For example:
```
[[source]]
name = "pypi"
url = "https://pypi.org/simple"
verify_ssl = true

[dev-packages]
pytest = ">=3.6,<3.7"
config_helpers = {editable = true, path = "./../config_helpers"}

[packages]
eth-utils = "<2,>=1.3.0"
eth-typing = "<3.0.0,>=2.1.0"
pycryptodome = "==3.7.3"
py-ecc = ">=1.6.0"

[requires]
python_version = "3.6"
```

`pipenv` looks nice and also may be a compromise between `requirements.txt` files and `setup.py`! It can easily make single checksum of one package; and if we want to serve multiple packages in one checksum, we can still dump multiple `Pipfile.lock` files into one text file, and calculate the checksum of it.

> 3. Let’s don’t use double version tickers in the caching key

Agreed! 👍

> 4. Multiple parts split by a `-` are likely the root of the cache-key problems.

It looks like as same as your point 2? Also not sure why that’s a problem. Any link? :)

protolambda:
> Sorry, could you point out why two checksums of the static files would cause problem

I'm not sure about the issues in CI aside from linting running on the wrong version. But hyphens cause the CI to fall back on keys with the same prefix (up to any hyphen), which may be the reason / co-factor of dependencies of the 2nd checksum not being installed as expected. Circle CI docs here: https://circleci.com/docs/2.0/caching/#restoring-cache, note the "**partial cache restore**", we should re-run the dependcies install on top of the restored cache contents.

> Ohh the *_default_* commands ...

The previous CI helper functions where there to make it easy to implement storing + restoring of additional venvs. The generators never made it due to speed & parallellization issues in the CI environment, but we may have others in the future. (contract testing?). Hence the preference to not call anything "default", and just make things explicit.

> Or using Pipenv to manage the package

Looks good! The current reason to stick with `requirements.txt` is that it is much less verbose/complex than setup.py, and easy to checksum. But the generators need to be able to require the `pyspec`, `config_helpers` and `gen_helpers` modules, and other modules may just need a few of them. Hence a `setup.py` for those modules :(. Changing to just a setup.py never made it because of local editable dependency issues (yes, I dislike it too, but things have to work somehow, and options/time were limited, without breaking encapsulation of modules). A POC with pipenv would be nice :+1: 

------

I'm going to try and fix the linting dependency blocking #1077 by moving the linting install out of the makefile, into the testing requirements. (We could make a 3rd requirements file for non-runtime deps, or another bigger change, but really just need to unblock the PR).

hwwhww:
@protolambda @djrtwo Sorry for the delay, I was traveling yesterday.

(Decided to push commit into this branch to get review easier 😅) here is the new config:
https://github.com/ethereum/eth2.0-specs/pull/1127/commits/96237c74e1a7f687ba3fc2481ac670e3156b9e67

And the new workflow:
https://circleci.com/workflow-run/1b479336-9951-4d27-bc06-81ecf7c7cd95

image

Ready for another look.

Makefile
protolambda: > Right now, install_test and test are independent.

People run tests locally repeatedly, but testing requirements only have to be installed once, running those over again (even if cached) just takes extra time, but that slowdown is nice to avoid.

protolambda:
> cd $(PY_SPEC_DIR); python3 -m venv venv; . venv/bin/activate; \
	cd ../..; cd $(DEPOSIT_CONTRACT_DIR); \

Why not just create a separate venv? It's much simpler and cleaner (no caching of separate cache of same venv)

Say, the location would be `./deposit_contract/venv` (nicely with the code that needs the venv)

in circle ci config:

```yaml
venv_path: ./deposit_contract/venv
```

Then we have a clear consistent reproducible venv locally (i.e. no need to clear and refresh cache of pyspec when you're working on both and switching feature branches)

Makefile
hwwhww: I finally figured out how to fix it: https://github.com/ethereum/eth2.0-specs/pull/1127/commits/e52c4a5526866deba675ef11ac6c4cf8499dba65 😅

hwwhww:
@protolambda 

> Why not just create a separate venv?

I was thinking about efficiency and the use case of testing deposit contract is rare. But after I created a new venv as you said, `Makefile` is much simpler so 👍! (PR feedback is applied: https://github.com/ethereum/eth2.0-specs/pull/1127/commits/5b8cca831458be07ec8e9a0ca580d60354bfd024)

Ready for another look. :)

protolambda:
Regarding that last 1-line addition: I declared the new targets (and 1 missing one: `lint`) as non-file make targets. See https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

PR #1130, opened 2019-05-27
Release V06x into master

djrtwo:
Detailed changelog to accompany the release.

Mostly testing and test generator enhancements


specs/test_formats/epoch_processing/README.md
```diff
@@ -0,0 +1,34 @@
...
+description: string    -- description of test case, purely for debugging purposes
+bls_required: bool     -- optional, true if the test validity is strictly dependent on BLS being ON. False otherwise.
+bls_ignored: bool      -- optional, true if the test validity is strictly dependent on BLS being OFF. False otherwise.
+pre: BeaconState       -- state before running the sub-transition
+post: BeaconState      -- state after applying the epoch sub-transition.
+```
+
+Note: if both `bls_required` and `bls_ignored` are false (or simply not included),
+ then the test consumer can freely choose to run with BLS ON or OFF.
+One may choose for OFF for performance reasons during repeated testing. Otherwise it is recommended to run with BLS ON.
```
hwwhww: Suggestion: the combination of `bls_required` and `bls_ignored` can be simplified to using only one required parameter `bls_setting: int` to indicate all the cases.
```
bls_setting=0: restrict to off
bls_setting=1: restrict to on
bls_setting=2: other cases, arbitrary
```

protolambda: See 6ffd416

specs/test_formats/README.md
```diff
@@ -176,6 +176,18 @@ To prevent parsing of hundreds of different YAML files to test a specific test t
 ...                           <--- more test types
 ```
 
+## Common test-case properties
+
+Some test-case formats share some common key-value pair patterns, and these are documented here:
+
+```
+bls_setting: int     -- optional, can have 3 different values:
```
hwwhww: `description` should also be in the "Common test-case properties" if it's required in every test format.

specs/test_formats/epoch_processing/README.md
```diff
@@ -0,0 +1,33 @@
...
+```yaml
+description: string    -- description of test case, purely for debugging purposes
+bls_setting: int       -- see general test-format spec.
+pre: BeaconState       -- state before running the sub-transition
+post: BeaconState      -- state after applying the epoch sub-transition.
+```
+
+Note: if both `bls_required` and `bls_ignored` are false (or simply not included),
+ then the test consumer can freely choose to run with BLS ON or OFF.
+One may choose for OFF for performance reasons during repeated testing. Otherwise it is recommended to run with BLS ON.
```
hwwhww: These two lines look like leftover

specs/test_formats/operations/README.md
```diff
@@ -2,9 +2,38 @@
...
+description: string                    -- description of test case, purely for debugging purposes
+bls_setting: int                       -- see general test-format spec.
+pre: BeaconState                       -- state before applying the operation
+<operation-name>: <operation-object>   -- the YAML encoded operation, e.g. a "ProposerSlashing", or "Deposit".
+post: BeaconState                      -- state after applying the operation. No value if operation processing is aborted.
+```
 
+Note: if both `bls_required` and `bls_ignored` are false (or simply not included),
+ then the test consumer can freely choose to run with BLS ON or OFF.
+One may choose for OFF for performance reasons during repeated testing. Otherwise it is recommended to run with BLS ON.
```
hwwhww: ditto

specs/test_formats/README.md
protolambda: Yes, but it's formatting of the test-format docs, not the format itself. So I'd rather not hotfix unnecessarily, and merge it with other test-format improvements later on (I hope to get & implement feedback from implementers on it as well).

protolambda:
For v0.6.2 changelog:
```
- `ssz_static` test-case format is changed to facilitate more efficient parsing. The SSZ type is specified in the key now.
```

PR #1132, opened 2019-05-27
fix dev branch build script, missing quotes

protolambda:
SSZ-static runner needs the names to be quoted (as they were before an older change). Fix the erroneous formatting.


protolambda:
> Blocked due to plan-no-credits-available Please purchase a new credit block then push a new commit or call the API to run a new build.

2 character change, not affecting anything important, and tested locally. Too bad the CI is out of credits. Merging this.

PR #1133, opened 2019-05-29
Test deposit top-up with inconsistent withdrawal credentials

michaelsproul:
Lighthouse had a check on Deposit withdrawal credentials that was left over from v0.4, which is no longer required (and would break consensus). I wrote this test to confirm the spec's behaviour, and figured it might be useful for other client implementers too.

Not sure exactly which base branch to target, if you'd prefer this merged into `dev`, I'll update the PR

CC @protolambda

### DIFF
--- a/scripts/phase0/build_spec.py
+++ b/scripts/phase0/build_spec.py
@@ -13,9 +13,17 @@ from typing import (
     NewType,
     Tuple,
 )
-from eth2spec.utils.minimal_ssz import *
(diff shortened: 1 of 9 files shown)


djrtwo:
awesome! We could:

1. do another minor version release
2. make a quick patch to the v0.6.2 release and rerun tests
3. just merge into `dev` and release with v0.7.0 next week

No strong preference here

michaelsproul:
I'm happy with (3)

PR #1141, opened 2019-06-05
Change data structure order to mirror beacon state property order

0xKiwi:
I'm not sure what the logic behind the sorting of the `Misc dependencies` data structures is but this PR changes the order to be more inline with the order they're used in `Beacon State`. 

I'm always confused on the order whenever I check the spec so hopefully this makes it a tad more consistent.

If something like this is not needed feel free to let me know :)


dankrad:
>  Due to the way we build the spec, the datastructures need to be sorted topologically (any dependencies coming earlier). That's why we have this crappy "Misc" category.

I think Carl fixed this in #1061. :)

PR #1145, opened 2019-06-07
Add Condition When Iterating Over Variable Parts

nisdas:
This makes it explicit to only add to the variable length list if the length of the part is non-zero

### DIFF
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -107,7 +107,7 @@ variable_parts = [serialize(element) if is_variable_size(element) else b"" for e
 
 # Compute and check lengths
 fixed_lengths = [len(part) if part != None else BYTES_PER_LENGTH_OFFSET for part in fixed_parts]
-variable_lengths = [len(part) for part in variable_parts]
+variable_lengths = [len(part) if len(part) != 0 for part in variable_parts]
 assert sum(fixed_lengths + variable_lengths) < 2**(BYTES_PER_LENGTH_OFFSET * BITS_PER_BYTE)
 
 # Interleave offsets of variable-size parts with fixed-size parts


terencechain:
Ops, I accidentally clicked approved. Please ignore that

specs/simple-serialize.md
```diff
@@ -107,7 +107,7 @@ variable_parts = [serialize(element) if is_variable_size(element) else b"" for e
 
 # Compute and check lengths
 fixed_lengths = [len(part) if part != None else BYTES_PER_LENGTH_OFFSET for part in fixed_parts]
-variable_lengths = [len(part) for part in variable_parts]
+variable_lengths = [len(part) if len(part) != 0 for part in variable_parts]
```
protolambda: Have you tested this? It looks like it breaks the variable_offsets calculation (down a few lines):

```python
variable_offsets = [serialize(sum(fixed_lengths + variable_lengths[:i])) for i in range(len(value))]
fixed_parts = [part if part != None else variable_offsets[i] for i, part in enumerate(fixed_parts)]
```

There needs to be a length for every item, even if 0.

protolambda:
```python
value = [b"foo", b"bar", b"test", b"\x42\x42", b"eth2", b"very", b"wow"]

def serialize(v): return v if isinstance(v, bytes) else v.to_bytes(4, 'little')

def is_variable_size(v): return b"t" in v

BYTES_PER_LENGTH_OFFSET = 4
BITS_PER_BYTE = 8


# Recursively serialize
fixed_parts = [serialize(element) if not is_variable_size(element) else None for element in value]
variable_parts = [serialize(element) if is_variable_size(element) else b"" for element in value]

# Compute and check lengths
fixed_lengths = [len(part) if part != None else BYTES_PER_LENGTH_OFFSET for part in fixed_parts]

variable_lengths = [len(part) for part in variable_parts if len(part) != 0]
print('pr: ', variable_lengths)
variable_lengths = [len(part) for part in variable_parts]
print('now: ', variable_lengths)

assert sum(fixed_lengths + variable_lengths) < 2**(BYTES_PER_LENGTH_OFFSET * BITS_PER_BYTE)
# Interleave offsets of variable-size parts with fixed-size parts
variable_offsets = [serialize(sum(fixed_lengths + variable_lengths[:i])) for i in range(len(value))]
fixed_parts = [part if part != None else variable_offsets[i] for i, part in enumerate(fixed_parts)]

print("fixed parts", fixed_parts)
print("var parts  ", variable_parts)

# Return the concatenation of the fixed-size parts (offsets interleaved) with the variable-size parts
x = b"".join(fixed_parts + variable_parts)

print(x.hex())
print(str(x))
print(len(x))

# slice out of bounds: valid in python, obfuscates bug in PR.
print(variable_lengths[:1000000])
```
Output:
```txt
pr:  [4, 4]
now:  [0, 0, 4, 0, 4, 0, 0]
fixed parts [b'foo', b'bar', b'\x17\x00\x00\x00', b'BB', b'\x1b\x00\x00\x00', b'very', b'wow']
var parts   [b'', b'', b'test', b'', b'eth2', b'', b'']
666f6f6261721700000042421b00000076657279776f777465737465746832
b'foobar\x17\x00\x00\x00BB\x1b\x00\x00\x00verywowtesteth2'
31
[0, 0, 4, 0, 4, 0, 0]
```

If you're looking for a more realistic way of encoding offsets, take a look at the pyspec implementation: 
 https://github.com/ethereum/eth2.0-specs/blob/6f82480df2acd5627fee600aeccf0b975c8a1022/test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py#L72
Or, alternatively:
PySSZ (MIT license): https://github.com/ethereum/py-ssz/blob/master/ssz/sedes/base.py
ZSSZ (disclaimer: my code, experimental, MIT license): https://github.com/protolambda/zssz/blob/77bb42f406abe74d9da0cad199dbb0859d161484/types/elem_var_size_series.go#L12

protolambda:
Closing this PR:
- introduces a bug in a 1 line change
- syntax is off (placement of `if` in list comprehension)
- don't see the alternative interpretation of `if len(0) != 0 else 0` helping either. As `else` can't be returning `None`, since we also sum it.

But thanks for looking into improving the ssz code :+1:

Issue #1129, opened 2019-05-27
Deposit contract checklist


JustinDrake:
- [x] 1. **Unify repos**: Move deposit contract to `eth2.0-specs` repo. See #1127.
- [x] 2. **Remove `Eth2Genesis`**: Rely on a more flexible genesis trigger, external to deposit contract. (Also reduces line count from 140 to 120.) See also #1048.
- [x] 3. **Remove `merkle_tree_index`**: See #1122.
- [x] 4. **Cleanups**: Consistent comment formatting, linting, more comments, etc.
- [x] 5. **Gas profiling**: Understand where gas is spent and check there is no significant wastage. See #1150.
- [x] 6. (*post-freeze discussion*) **Formal verification**: Opcode-level formal verification to start in June by [Runtime Verification](https://runtimeverification.com/).
- [x] 7. (*post-freeze discussion*) **Wait for BLS standardisation**: Various details (G1/G2 point serialisation ambiguity, cipher-suite bits, proof of possession standard, etc.) need to figured out before deployment.
- [ ] 8. (*post-freeze discussion*) **Deployment address**: Agree on mechanism to generate deployment address (e.g. use CREATE2 or not), and clearly communicate to public. (Want to avoid scam/fake deposit contracts.)
- [ ] 9. (*post-freeze discussion*) **Deploy early**: Deploy ahead of target Eth2 genesis, even if less than two clients are production ready. Suggested deposit contract deployment target: end of Q3 2019 (30 September). Suggested Eth2 genesis target: 3 January 2020 at 00:00 UTC (Bitcoin's 11'th anniversary).

hwwhww:
Link [the formal correctness proof of `get_deposit_root` by Runtime Verification](https://github.com/runtimeverification/verified-smart-contracts/blob/master/deposit/formal-incremental-merkle-tree-algorithm.pdf) (https://github.com/ethereum/deposit_contract/issues/44#issuecomment-499625155)

Thank you @daejunpark!

JustinDrake:
> array access for bytes[] not currently supported in vyper

We could try asking the Vyper people to add this feature :)

djrtwo:
The pressing items here have been handled. Still need to deploy and publicize

closing

axic:
Points 8) and 9) don't seem to have been covered here. Are there new issues for tracking them?

q9f:
[...]

Issue #1151, opened 2019-05-06
Reduce accessibility modifiers as much as possible


ralexstokes:
Some functions in the deposit contract are marked `@public` when they have no reason to be (other than implementation details of the vyper stack, i.e. they are easier to test when marked `@public`).

I suggest in the spirit of keeping this contract as simple and small as possible, we reduce the size of the public API by remove the `@public` decorators where we can.

At the moment that is:

```python
@public
@constant
def to_little_endian_64(value: uint256) -> bytes[8]:
```

assuming we also address #34.

JustinDrake:
As you mention, removing `@public` on `to_little_endian_64` breaks the tests.

Please open a PR ASAP if you feel strongly about this :) It will be harder to make changes after spec freeze, and after formal verification of the deposit contract.

JustinDrake:
Fixed in #1152 :)

PR #1152, opened 2019-06-08
Custom genesis trigger, and deposit contract cleanups

JustinDrake:
Address items 2, 3, 4 in #1129.

Substantive changes:

* Remove `Eth2genesis` log in favour of a custom `is_genesis_trigger` function
* Reorder `block_hash` field within `Eth1Data`
* Remove `merkle_tree_index` from `Deposit` log


specs/core/0_beacon-chain.md
```diff
@@ -1156,20 +1156,45 @@ def slash_validator(state: BeaconState,
...
 
-### `Eth2Genesis`
+### Genesis trigger
+
+Whenever the deposit contract emits a `Deposit` log call the function `is_genesis_trigger(deposits: List[Deposit], timestamp: uint64) -> bool` where:
+
+* `deposits` is the list of all deposits, ordered chronologically, up to and including the deposit triggering the latest `Deposit` log
+* `timestamp` is the Unix timestamp in the Ethereum 1.0 block that emitted the latest `Deposit` log
+
+When `is_genesis_trigger(deposits, timestamp) == True` for the first time let:
```
hwwhww: super-nitpick
```suggestion
When `is_genesis_trigger(deposits, timestamp) is True` for the first time let:
```

specs/core/0_beacon-chain.md
```diff
@@ -370,12 +370,12 @@ class PendingAttestation(Container):
...
 ```python
 class Eth1Data(Container):
+    # Block hash
+    block_hash: Bytes32
     # Root of the deposit tree
     deposit_root: Bytes32
     # Total number of deposits
     deposit_count: uint64
-    # Block hash
-    block_hash: Bytes32
```
hwwhww: It would be nice to note this change in the changelog.

specs/core/0_deposit-contract.md
```diff
@@ -62,24 +53,13 @@ The private key corresponding to `withdrawal_pubkey` will be required to initiat
 #### Amount
 
 * A valid deposit amount should be at least `MIN_DEPOSIT_AMOUNT` in Gwei.
```
hwwhww: Since it's only one bullet left:
```suggestion
A valid deposit amount should be at least `MIN_DEPOSIT_AMOUNT` in Gwei.
```

deposit_contract/contracts/validator_registration.v.py
```diff
@@ -47,7 +40,6 @@ def to_little_endian_64(value: uint256) -> bytes[8]:
 
     return slice(convert(y, bytes32), start=24, len=8)
 
-
```
hwwhww: kinda PEP8 rule: two lines between functions.

specs/core/0_beacon-chain.md
```diff
@@ -1156,20 +1156,45 @@ def slash_validator(state: BeaconState,
...
+
+Whenever the deposit contract emits a `Deposit` log call the function `is_genesis_trigger(deposits: List[Deposit], timestamp: uint64) -> bool` where:
+
+* `deposits` is the list of all deposits, ordered chronologically, up to and including the deposit triggering the latest `Deposit` log
+* `timestamp` is the Unix timestamp in the Ethereum 1.0 block that emitted the latest `Deposit` log
+
+When `is_genesis_trigger(deposits, timestamp) == True` for the first time let:
+
+* `genesis_deposits = deposits`
+* `genesis_time = timestamp % SECONDS_PER_DAY + 2 * SECONDS_PER_DAY` where `SECONDS_PER_DAY = 86400`
```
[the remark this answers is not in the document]
JustinDrake: I did not put `SECONDS_PER_DAY ` in the constants because the whole "Genesis trigger" section should probably go in the `0_fork-choice.md` file. This is an intermediate step. I will migrate everything (including the `SECONDS_PER_DAY` constant, putting it in a table) once #1068 is merged.

deposit_contract/tests/contracts/conftest.py
```diff
@@ -63,28 +62,14 @@ def registration_contract(w3, tester):
     return registration_deployed
 
 
-@pytest.fixture(scope="session")
-def chain_start_full_deposit_thresholds():
-    return [randint(1, 5), randint(6, 10), randint(11, 15)]
-
-
 @pytest.fixture(params=[0, 1, 2])
 def modified_registration_contract(
```
NIC619: This fixture can be safely removed. It was used to make testing chain start event easier by modifying `CHAIN_START_FULL_DEPOSIT_THRESHOLD`.

deposit_contract/contracts/validator_registration.v.py
```diff
@@ -1,140 +1,99 @@
...
 @public
 def deposit(pubkey: bytes[PUBKEY_LENGTH],
             withdrawal_credentials: bytes[WITHDRAWAL_CREDENTIALS_LENGTH],
             signature: bytes[SIGNATURE_LENGTH]):
-    # Prevent edge case in computing `self.branch` when `self.deposit_count == MAX_DEPOSIT_COUNT`
-    # NOTE: reaching this point with the constants as currently defined is impossible due to the
-    # uni-directional nature of transfers from eth1 to eth2 and the total ether supply (< 130M).
-    assert self.deposit_count < MAX_DEPOSIT_COUNT
+    # Avoid overflowing the Merkle tree
+    assert self.deposit_count < 2**DEPOSIT_CONTRACT_TREE_DEPTH - 1
```
djrtwo: At the time of creating `MAX_DEPOSIT_COUNT`, having this constant calculation here was not optimized by the compiler and we took a noticeable (but probably still relatively small) gas hit to calculate dynamically.

djrtwo: This actually prevents not only overflowing the merkle tree but prevents the last write which due to the incremental merkle root alg creates a root in error.

I think the comment should reflect this fact

deposit_contract/tests/contracts/test_deposit.py
```diff
@@ -56,7 +56,7 @@ def deposit_input():
         (10, True),
         (55555, True),
         (2**64 - 1, True),
-        (2**64, False),
+        # (2**64, True),  # Note that all calls to `to_little_endian_64` have an input less than 2**64
```
djrtwo: I'm not certain why you took the `assert` out and this test out. The method is exposed publicly and should thus have the assert blocker. The only way the assert should be removed is if the method is private, and the comment on max size passed in should stay within the method.

specs/core/0_beacon-chain.md
```diff
@@ -1156,20 +1156,45 @@ def slash_validator(state: BeaconState,
 
 ## Genesis
 
-### `Eth2Genesis`
+### Genesis trigger
+
+Whenever the deposit contract emits a `Deposit` log call the function `is_genesis_trigger(deposits: List[Deposit], timestamp: uint64) -> bool` where:
```
djrtwo: I would say "Before genesis has occurred"

specs/core/0_beacon-chain.md
```diff
@@ -1156,20 +1156,45 @@ def slash_validator(state: BeaconState,
...
+
+* `deposits` is the list of all deposits, ordered chronologically, up to and including the deposit triggering the latest `Deposit` log
+* `timestamp` is the Unix timestamp in the Ethereum 1.0 block that emitted the latest `Deposit` log
+
+When `is_genesis_trigger(deposits, timestamp) is True` for the first time let:
+
+* `genesis_deposits = deposits`
+* `genesis_time = timestamp - timestamp % SECONDS_PER_DAY + 2 * SECONDS_PER_DAY` where `SECONDS_PER_DAY = 86400`
+* `genesis_eth1_data` be the object of type `Eth1Data` where:
+    * `genesis_eth1_data.block_hash` is the block hash for the last deposit in `deposits`
```
```suggestion
    * `genesis_eth1_data.block_hash` is the eth1 block hash that emitted the log for the last deposit in `deposits`
```

specs/core/0_deposit-contract.md
```diff
@@ -28,66 +25,40 @@ This document represents the specification for the beacon chain deposit contract
...
 
-### Arguments
+### `deposit` function
+
+The deposit contract has a public `deposit` function to make deposits. It takes as arguments `pubkey: bytes[48], withdrawal_credentials: bytes[32], signature: bytes[96]` corresponding to a `DepositData` object.
 
-The deposit contract has a `deposit` function which takes the amount in Ethereum 1.0 transaction, and arguments `pubkey: bytes[48], withdrawal_credentials: bytes[32], signature: bytes[96]` corresponding to `DepositData`.
+#### Deposit amount
+
+The ETH sent to the deposit contract, i.e. the deposit amount, is burnt on Ethereum 1.0. The minimum deposit amount is `MIN_DEPOSIT_AMOUNT` Gwei.
```
djrtwo: I'd rather not use the term "burnt".

deposit_contract/tests/contracts/test_deposit.py
JustinDrake: > The only way the assert should be removed is if the method is private

This is another argument in favour of #1151.

> the comment on max size passed in should stay within the method

:+1:

specs/core/0_deposit-contract.md
JustinDrake: Replaced by "no longer usable on Ethereum 1.0" 😂

deposit_contract/tests/contracts/test_deposit.py
djrtwo: I'm in favor of it. The method was originally exposed publicly to be able to write tests directly against it because it was using weird vyper magic. I think it reasonable to make it private and have it tested implicitly by all the other tests that depend on it.

deposit_contract/contracts/validator_registration.v.py
djrtwo: How about my second comment here

djrtwo:
I'm not seeing any changes wrt any of my comments. Did you not push something? For example "burnt" still in there

JustinDrake:
@djrtwo Sorry my `git push` failed and I walked away from keyboard.

deposit_contract/tests/contracts/test_deposit.py
JustinDrake: @NIC619 or @hwwhww Can you help make `to_little_endian_64` private while keeping the tests working?

djrtwo: The thing is we _won't_ be able to run tests directly against the method any more (unless we did weird dynamic surgery to the contract for testing.. no advised).

We just need to ensure that the methods we are testing that use the helper method produce the correct results. So making sure `get_deposit_count` and the merkle root (that calls the method on `amount`) work correctly tests the method enough.

djrtwo: @JustinDrake Do you want to resolve this issue in this PR?

JustinDrake:
I think all issues are addressed. (I also submitted a Vyper issue [here](https://github.com/ethereum/vyper/issues/1476)—it seems cleaner to emit the `Deposit` log after validating the deposit data but tests fails when I do that.)

PR #1159, opened 2019-06-10
Tweak inclusion delay rewards and set BASE_REWARD_FACTOR

JustinDrake:
Substantive changes:

1) Split the inclusion delay reward between attester and proposer to add up to at most one base reward. This is analogous to the reward logic in `slash_validator`, and makes the `BASE_REWARDS_PER_EPOCH` constant include proposer rewards.
2) Double `BASE_REWARD_FACTOR` to 2^6 (addressing item 4 in #1054). When the total effective balance is 2^17 ETH then maximum annual issuance is a bit below 2^21 ETH. Maximum annual issuance happens when a) all validators make perfect attestations (matching source, target, head, as well as consistent crosslink data), b) all attestations are included as fast as possible (in particular, no skip blocks), and c) there are no slashings.

```python
BASE_REWARD_FACTOR = 2**6
SLOTS_PER_EPOCH = 2**6
SECONDS_PER_SLOT = 6
BASE_REWARDS_PER_EPOCH = 5
GWEI_PER_ETH = 10**9
MAX_TOTAL_EFFECTIVE_BALANCE = 2**27 * GWEI_PER_ETH
TARGET_MAX_ISSUANCE = 2**21 * GWEI_PER_ETH

def integer_squareroot(n: int) -> int:
    """
    The largest integer ``x`` such that ``x**2`` is less than or equal to ``n``.
    """
    assert n >= 0
    x = n
    y = (x + 1) // 2
    while y < x:
        x = y
        y = (x + n // x) // 2
    return x

MAX_REWARDS_PER_EPOCH = MAX_TOTAL_EFFECTIVE_BALANCE * BASE_REWARD_FACTOR // integer_squareroot(MAX_TOTAL_EFFECTIVE_BALANCE) // BASE_REWARDS_PER_EPOCH * BASE_REWARDS_PER_EPOCH
EPOCHS_PER_YEAR = 365.25*24*60*60 / (SECONDS_PER_SLOT * SLOTS_PER_EPOCH)
MAX_REWARDS_PER_YEAR = EPOCHS_PER_YEAR * MAX_REWARDS_PER_EPOCH

print(MAX_REWARDS_PER_YEAR / TARGET_MAX_ISSUANCE)
```


djrtwo:
This seems reasonable... The counter argument is it's just as simple to bound issuance with as `BASE_REWARDS_PER_EPOCH + 1 / PROPOSER_REWARD_QUOTIENT`, and we want to insure that we incentivize validators to not delay in creating their attestation.

Also need to remember there will be participation rewards on shard chains.

This also might not be the final adjustment for [2] until we lock down slot time

JustinDrake:
> This also might not be the final adjustment for [2] until we lock down slot time

I suggest keeping the slot time at 6 seconds for the spec freeze to "stress test" the network. So it could be the last adjustment before freeze (though possibly not before finalization).

specs/core/0_beacon-chain.md
```diff
@@ -224,14 +224,13 @@ These configurations are updated for releases, but may be out of sync during `de
...
-* The `INACTIVITY_PENALTY_QUOTIENT` equals `INVERSE_SQRT_E_DROP_TIME**2` where `INVERSE_SQRT_E_DROP_TIME := 2**12 epochs` (~18 days) is the time it takes the inactivity penalty to reduce the balance of non-participating [validators](#dfn-validator) to about `1/sqrt(e) ~= 60.6%`. Indeed, the balance retained by offline [validators](#dfn-validator) after `n` epochs is about `(1 - 1/INACTIVITY_PENALTY_QUOTIENT)**(n**2/2)` so after `INVERSE_SQRT_E_DROP_TIME` epochs it is roughly `(1 - 1/INACTIVITY_PENALTY_QUOTIENT)**(INACTIVITY_PENALTY_QUOTIENT/2) ~= 1/sqrt(e)`.
+* The `INACTIVITY_PENALTY_QUOTIENT` equals `INVERSE_SQRT_E_DROP_TIME**2` where `INVERSE_SQRT_E_DROP_TIME := 2**12 epochs` (about 18 days) is the time it takes the inactivity penalty to reduce the balance of non-participating [validators](#dfn-validator) to about `1/sqrt(e) ~= 60.6%`. Indeed, the balance retained by offline [validators](#dfn-validator) after `n` epochs is about `(1 - 1/INACTIVITY_PENALTY_QUOTIENT)**(n**2/2)` so after `INVERSE_SQRT_E_DROP_TIME` epochs it is roughly `(1 - 1/INACTIVITY_PENALTY_QUOTIENT)**(INACTIVITY_PENALTY_QUOTIENT/2) ~= 1/sqrt(e)`.
```
djrtwo: Should we remove this warning? The base reward factor is not final until some of the other factors are final

JustinDrake: It's likely that the `BASE_REWARD_FACTOR` in this PR is good enough for freeze. The factor can also be adjusted (like any other part of the protocol) during post-freeze finalisation. I guess the warning "Notice: This document is a work-in-progress for researchers and implementers." should *not* be removed during the freeze, only after post-freeze finalisation.

djrtwo: cool with me.

cryptoneobancoin:
yes 6 seconds

Issue #1091, opened 2019-05-17
Explicit unhandled exceptions


sorpaas:
The spec specifies that if the runtime "panics", the block is considered invalid. However, in real life, we need to properly return errors. In the specs, `assert` can be easily turned into errors. However, there're several cases where addition/subtraction can overflow (such as deposit counting), and array access can be out of bound (such as merkle proof validation) for invalid states/blocks. It would be great to add additional `assert` in the spec to clearly document them.

JustinDrake:
I wouldn't be opposed to *all* unhandled exceptions being thrown by an explicit `assert` :) Fuzz testing can help detect lingering implicit exceptions.

guidovranken:
I'll list IndexError exceptions found by my fuzzer here:

```process_proposer_slashing```:

```python
proposer = state.validator_registry[proposer_slashing.proposer_index]
```

```process_voluntary_exit```:

```python
validator = state.validator_registry[exit.validator_index]
```

```process_attestation```:

```
  File "/home/jhg/eth-2019/x/eth2.0-fuzzing/files/fuzzers/block/pythonlib/eth2spec/phase0/spec.py", line 1211, in process_attestation
    parent_crosslink = state.current_crosslinks[data.crosslink.shard]
  File "/home/jhg/eth-2019/x/eth2.0-fuzzing/files/fuzzers/block/pythonlib/eth2spec/utils/ssz/ssz_typing.py", line 269, in __getitem__
    return self.items[key]
IndexError: cannot fit 'int' into an index-sized integer
```

djrtwo:
Closing. These additional asserts won't make it into the near-frozen spec.

pyspec/SSZ-typing do throw for under and overflows so at least this isn't blindly happening in the spec anymore

PR #1171, opened 2019-06-13
attestations: check shard, and check epoch earlier. Open question about another attestation check.

protolambda:
Few minor bugfixes in attestation processing:
- out of bounds shards results in invalid lookups/calculations, catch it. (avoid panic in Go)
- target epoch should be checked when we can, not after using it.

Remaining question, for @JustinDrake

~~When there are two bitfields (Aggregation, and custody1), and we derive the indices for custody0 (in `convert_to_indexed`), should we check len(participants) < len(custody1)? E.g. the bitfields could be verified to be committee size, yet the custody bits could be more full than the actual participants, resulting in extra custody-1 indices. Or is this prevented elsewhere?~~ Moved to #1173

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1722,6 +1722,10 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
     Process ``Attestation`` operation.
     """
     data = attestation.data
+
+    assert data.crosslink.shard < SHARD_COUNT
+    assert data.target_epoch in (get_previous_epoch(state), get_current_epoch(state))
+
     attestation_slot = get_attestation_data_slot(state, data)
     assert attestation_slot + MIN_ATTESTATION_INCLUSION_DELAY <= state.slot <= attestation_slot + SLOTS_PER_EPOCH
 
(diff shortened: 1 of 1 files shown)


JustinDrake:
> When there are two bitfields (Aggregation, and custody1), and we derive the indices for custody0 (in convert_to_indexed), should we check len(participants) < len(custody1)?

Oh, I think you're right that something is fishy in `convert_to_indexed`. Specifically, we want to check that every bit in `custody_bitfield` is zero whenever the corresponding bit in `aggregation_bitfield` is zero. It's possible this invariant was lost with the [refactor to `IndexedAttestation`](https://github.com/ethereum/eth2.0-specs/pull/831). cc @djrtwo

JustinDrake:
> Few minor bugfixes

These are arguably not bug fixes in a substantive sense, i.e. the fixes don't change the result of the state transition function (exceptions are thrown regardless). Still good to cleanup from a cosmetic standpoint in the context of #1091.

protolambda:
@JustinDrake Made issue #1173 for the bitfield checking question, so we can move on with this PR. LGTM?

djrtwo:
will check out #1091

PR #1172, opened 2019-06-13
Update genesis trigger

terencechain:
`timestamp` was never used in `is_genesis_trigger`, not sure if it was meant to be a placeholder for the later updates


djrtwo:
`timestamp` could very reasonably be used  in `is_genesis_trigger` (for example blocking before Jan 3) so I think it reasonable to keep in for now.

Issue #1169, opened 2019-06-13
RANDAO is not procssed in block sanity test


sorpaas:
In block sanity tests, `latest_randao_mixes` is always `ZERO_HASH` no matter how many blocks are processed.

djrtwo:
weird. investigating

djrtwo:
addressed in #1175 

Will release a `v0.7.1` with updated tests by the weekend

Issue #1146, opened 2019-06-07
Invalid sanity test block state root


sorpaas:
For `blocks` sanity test. Block state root is always zero, which makes it fail the final state root check required in the spec.

JustinDrake:
I'm sorry, I don't understand the issue. Would you mind expanding?

sorpaas:
@JustinDrake blocks provided in the sanity test have its block state root field set to ZERO_HASH, but that’s clearly a wrong value.

djrtwo:
looking into this

protolambda:
Possible solution for tests:
- transition up to some testing scenario (optional)
- sign block
- output pre
- transition block, verify-state-root is off
- update block state root
- sign block again
- output block
- output post

djrtwo:
addressed in #1177.
to be released in `v0.7.1`

PR #1176, opened 2019-06-13
release `v0.7.1`

djrtwo:
A few critical fixes:
* Fix for randao processing (#1169)
* Fix for properly handling `state_root` in blocks in tests (#1146)
* Fix for G2 padding in bls tests (#1179)

Additional changes:
* fuzzing tool enhancement (#1178)
* add test code coverage reports (#1182)

### DIFF
--- a/test_libs/pyspec/eth2spec/test/helpers/state.py
+++ b/test_libs/pyspec/eth2spec/test/helpers/state.py
@@ -1,3 +1,6 @@
+from eth2spec.test.helpers.block import sign_block
+
+
 def get_balance(state, index):
     return state.balances[index]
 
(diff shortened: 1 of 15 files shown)


test_libs/pyspec/eth2spec/test/helpers/state.py
```diff
@@ -23,3 +26,13 @@ def get_state_root(spec, state, slot) -> bytes:
     """
     assert slot < state.slot <= slot + spec.SLOTS_PER_HISTORICAL_ROOT
     return state.latest_state_roots[slot % spec.SLOTS_PER_HISTORICAL_ROOT]
+
+
+def state_transition_and_sign_block(spec, state, block):
```
protolambda: Verified we are not outputting any blocks in the tests not produced by this function. `appyl_empty_block` is not used for output anywhere. But we may want to remove its return in the future, so we don't accidentally get a block with faulty state root in there.

test_libs/pyspec/requirements-testing.txt
```diff
@@ -2,3 +2,4 @@
 pytest>=3.6,<3.7
 ../config_helpers
 flake8==3.7.7
+pytest-cov
```
protolambda: yay for coverage reports. Don't forget to `make install_test` to enable it. Good for changelog.

PR #1178, opened 2019-06-14
Fuzzing utilities package / SSZ decoding for spec

protolambda:
SSZ decoding through pyssz for fuzzing. Intended to use with fuzzing. But maybe it's better to make a generic toolbelt package?

Implemented:
- translate a pyspec type to a py-ssz sedes (serialization description?)
- translate a py-ssz decoded value to a pyspec value

Usage:
```python

from eth2spec.phase0 import spec
from preset_loader import loader
from eth2spec.utils.ssz import ssz_impl as spec_ssz_impl

configs_path = "../../configs/"
config_name = "minimal"
presets = loader.load_presets(configs_path, config_name)
spec.apply_constants_preset(presets)

typ = spec.BeaconBlock
v = typ()
v.slot = 123

data = spec_ssz_impl.serialize(v)
print('data', data.hex())

block_sedes = translate_typ(typ)
print('sedes', block_sedes)

raw_value = block_sedes.deserialize(data)
print('raw_value', raw_value)

assert block_sedes.serialize(raw_value) == data

block = translate_value(raw_value, typ)
print('res', block)

assert spec_ssz_impl.hash_tree_root(v) == spec_ssz_impl.hash_tree_root(block)

```

TODO:
- [x] decide on packaging ~~(fuzzing? toolbelt?)~~ Moved into pyspec, since it closely works with py-ssz, and needs to be readily accessible for Guido to start fuzzing
-  ~~add setup.py so it can be imported into other modules~~ (see above)
- [x] **update py-ssz dependency version, this version does not use offsets, but it looks like the newest unreleased code on github does**

### DIFF
--- a/test_libs/pyspec/eth2spec/fuzzing/decoder.py
+++ b/test_libs/pyspec/eth2spec/fuzzing/decoder.py
@@ -72,7 +72,7 @@ def translate_value(value, typ):
         return value
     elif spec_ssz.is_vector_type(typ):
         elem_typ = spec_ssz.read_elem_type(typ)
-        return typ(translate_value(elem, elem_typ) for elem in value)
+        return typ(*(translate_value(elem, elem_typ) for elem in value))
     elif spec_ssz.is_bytesn_type(typ):
         return typ(value)
     elif spec_ssz.is_bytes_type(typ):
--- a/test_libs/pyspec/eth2spec/fuzzing/decoder.py
+++ b/test_libs/pyspec/eth2spec/fuzzing/decoder.py
@@ -10,7 +10,7 @@ def translate_typ(typ) -> ssz.BaseSedes:
     """
     if spec_ssz.is_container_type(typ):
         return ssz.Container(
-            [(field_name, translate_typ(field_typ)) for (field_name, field_typ) in typ.get_fields()])
+            [translate_typ(field_typ) for (field_name, field_typ) in typ.get_fields()])
     elif spec_ssz.is_bytesn_type(typ):
         return ssz.ByteVector(typ.length)
     elif spec_ssz.is_bytes_type(typ):
@@ -38,7 +38,7 @@ def translate_typ(typ) -> ssz.BaseSedes:
         else:
             raise TypeError("invalid uint size")
     else:
-        raise Exception("Type not supported: {}".format(typ))
+        raise TypeError("Type not supported: {}".format(typ))
 
 
 def translate_value(value, typ):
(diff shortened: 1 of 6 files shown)


protolambda:
@hwwhww or @carver can you help me out with a new py-ssz release to read offsets in decoding? :pray:

carver:
I'm not deeply familiar with py-ssz, I seem to be a pypi maintainer only as emergency backup. Maybe @pipermerriam could help here with the question about read offsets (and a release)?

test_libs/pyspec/eth2spec/fuzzing/decoder.py
```diff
@@ -0,0 +1,84 @@
+from eth2spec.utils.ssz import ssz_typing as spec_ssz
+import ssz
+
+
+def translate_typ(typ) -> ssz.BaseSedes:
```
djrtwo: Do we want to add some tests here?
Because we don't use this anywhere in the spec-dev, CI, or generators, the chances of missing a corresponding update here when we change SSZ is high. For example, adding max length to lists.

test_libs/pyspec/eth2spec/fuzzing/decoder.py
```diff
@@ -0,0 +1,84 @@
...
+    elif spec_ssz.is_vector_type(typ):
+        elem_typ = spec_ssz.read_elem_type(typ)
+        return typ(translate_value(elem, elem_typ) for elem in value)
+    elif spec_ssz.is_bytesn_type(typ):
+        return typ(value)
+    elif spec_ssz.is_bytes_type(typ):
+        return value
+    elif spec_ssz.is_container_type(typ):
+        return typ(**{f_name: translate_value(f_val, f_typ) for (f_name, f_val, f_typ)
+                      in zip(typ.get_field_names(), value, typ.get_field_types())})
```
djrtwo: is `value` just a list here or do we need to do something like `value.get_field_values()`?

protolambda: value is a tuple of values, one element per field, in order.

test_libs/pyspec/eth2spec/fuzzing/decoder.py
protolambda: Agree, will transform the example usage in a test that goes over all phase-0 types.

protolambda:
Blocked by https://github.com/ethereum/py-ssz/pull/74

test_libs/pyspec/eth2spec/fuzzing/decoder.py
protolambda: Included a test but not sure where I like the test to be located. Kept it near the code for now, since it's not really a spec test.

protolambda:
This PR is ready for review

test_libs/pyspec/eth2spec/fuzzing/decoder.py
djrtwo: seems appropriate for now 👍

PR #1180, opened 2019-06-14
pyspec-SSZ: lists-rework (enable static generalized indices) + fully python class based now.

vbuterin:
Not yet done:

* Making it work with tests
* Any SSZ partial related things
* Modifying the phase 0 spec to add explicit lists


test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -208,221 +138,115 @@ def __new__(cls, class_name, parents, attrs):
...
-    def __eq__(self, other):
-        return _is_equal_vector_type(self, other)
-
-    def __ne__(self, other):
-        return not _is_equal_vector_type(self, other)
-
-    def __hash__(self):
-        return hash(self.__class__)
-
+    def __instancecheck__(self, obj):
```
protolambda: not sure about these instance checks. If we're refactoring it again, it may be nice to find a better solution for the `issubclass(Vector[T, N], Vector)` and alike type checks.

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -208,221 +138,115 @@ def __new__(cls, class_name, parents, attrs):
...
 
     def __iter__(self):
         return iter(self.items)
 
-    def __len__(self):
-        return len(self.items)
-
     def __eq__(self, other):
-        return self.hash_tree_root() == other.hash_tree_root()
+        return self.items == other.items
```
protolambda: :+1:

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -208,221 +138,115 @@ def __new__(cls, class_name, parents, attrs):
...
-    def __getitem__(self, n):
-        return self.__class__(self.__name__, (BytesN,), {'length': n})
+class Vector(AbstractList, metaclass=AbstractListMeta):
+    def value_check(self, value):
+        return len(value) == self.__class__.length and super().value_check(value)
 
-    def __subclasscheck__(self, sub):
-        return _is_bytes_n_instance_of(self, sub)
+    def default(self):
+        return [get_zero_value(self.__class__.elem_type) for _ in range(self.__class__.length)]
```
protolambda: you are removing `get_zero_value` right? Shouldn't we call something like `self.__class__.elem_type.default()` instead now?

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -502,22 +325,5 @@ def is_container_type(typ):
...
-    if list_typ.__args__ is None or len(list_typ.__args__) != 1:
-        raise TypeError("Supplied list-type is invalid, no element type found.")
-    return list_typ.__args__[0]
-
-
-def read_vector_elem_type(vector_typ: Type[Vector[T, L]]) -> T:
-    return vector_typ.elem_type
-
-
 def read_elem_type(typ):
```
protolambda: we may as well remove this now, yay :+1:

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -452,15 +276,14 @@ def is_list_type(typ):
...
+    return isinstance(typ, type) and issubclass(typ, List)
 
 
 def is_bytes_type(typ):
     """
     Check if the given type is a ``bytes``.
     """
-    # Do not accept subclasses of bytes here, to avoid confusion with BytesN
-    return typ == bytes
+    return isinstance(typ, type) and issubclass(typ, Bytes)
```
protolambda: now that everything is a type, can we not do `issubclass(typ, Bytes)` directly in the implementation? (or alias that subclass check, if you find it ugly for readability. I think that's better than maintaining 10 `is_foobar_type` methods. Yay for consistent typing :+1:

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -208,221 +138,115 @@ def __new__(cls, class_name, parents, attrs):
...
+            return False
+        if hasattr(self, 'length') and obj.__class__.length != self.length:
+            return False
+        return True
 
-class Vector(metaclass=VectorMeta):
+class ValueCheckError(Exception):
+    pass
+
+class AbstractList(metaclass=AbstractListMeta):
```
protolambda: If we're not using `AbstractList` directly, we shouldn't need to create an `AbstractListMeta` for it, we can leave it out. (or maybe leave it out in the subclass declarations, but unsure if that breaks things)

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
vbuterin: Back when we had multiple meta classes, I got it to actually work without relying on __name__ in this hacky way. Haven't figured out how to do it with a single meta class though.

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
vbuterin: I suppose we could. I was originally designing the code so that we could put values at the bottom that are not SSZ types, but if that's not an objective we can definitely get rid of `get_zero_value` and just use `typ.default()`

protolambda:
```python

print(Vector[uint8, 7]())
print(Vector[uint8, 7](uint8(i) for i in range(7)))
print(List[uint8, 128](uint8(1), uint8(2), uint8(3)))

BigVector = Vector[uint8, 64]
foo = BigVector()
foo[3] = uint8(123)
print(foo)


class Foobar(Container):
    a: uint8
    b: uint32


x = Foobar()
print(x)

y = Bytes[123](b"foobar")
print(y)

print(BytesN[3](b"yay"))
print(BytesN[8](i ^ 0xa0 for i in range(8)))
```

out:
```

print(Vector[uint8, 7]())
print(Vector[uint8, 7](uint8(i) for i in range(7)))
print(List[uint8, 128](uint8(1), uint8(2), uint8(3)))

BigVector = Vector[uint8, 64]
foo = BigVector()
foo[3] = uint8(123)
print(foo)


class Foobar(Container):
    a: uint8
    b: uint32


x = Foobar()
print(x)

y = Bytes[123](b"foobar")
print(y)

print(BytesN[3](b"yay"))
print(BytesN[8](i ^ 0xa0 for i in range(8)))
```

Also:
- can't instantiate a type with wrong amount of params or none at all
- `typ.default()` is a thing now :)
- everything can be checked with `issubclass`
- easily declare parametrized type properties (just 3 clean lines to define/override a metaclass)

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -159,365 +119,173 @@ def get_field_types(cls):
...
+            return bytes(args)
 
-def is_vector_kind(typ):
-    """
-    Check if the given type is a kind of vector. Can be BytesN.
-    """
-    return is_vector_type(typ) or is_bytesn_type(typ)
+    @classmethod
+    def value_check(cls, value):
+        return len(value) == cls.length and isinstance(value, bytes)
```
vbuterin: Any reason why this goes here and not in `Bytes`? Seems like the latter would be more intuitive due to more symmetry? Or are we afraid of the parent class not having a `value_check` method for some reason?

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -159,365 +119,173 @@ def get_field_types(cls):
...
-        return uint64
-    elif isinstance(obj, list):
-        return List[infer_type(obj[0])]
-    elif isinstance(obj, (Vector, Container, bool, BytesN, bytes)):
-        return obj.__class__
-    else:
-        raise Exception("Unknown type for {}".format(obj))
-
-
-def infer_input_type(fn):
```
vbuterin: Wait, where do `infer_input_type` and `infer_type` live now?

protolambda:
test output for new merkleization:
```python

def hash(v):
    return f"h({v})"

zerohashes = [f"Z({i})" for i in range(0, 32)]

... merkleization


for i in range(20):
    print(merkleize_chunks([f"{v}" for v in range(i)], 32))
```
```
h(h(h(h(h(Z(0)Z(0))Z(1))Z(2))Z(3))Z(4))
h(h(h(h(h(0Z(0))Z(1))Z(2))Z(3))Z(4))
h(h(h(h(h(01)Z(1))Z(2))Z(3))Z(4))
h(h(h(h(h(01)h(2Z(0)))Z(2))Z(3))Z(4))
h(h(h(h(h(01)h(23))Z(2))Z(3))Z(4))
h(h(h(h(h(01)h(23))h(h(4Z(0))Z(1)))Z(3))Z(4))
h(h(h(h(h(01)h(23))h(h(45)h(Z(0)Z(0))))Z(3))Z(4))
h(h(h(h(h(01)h(23))h(h(45)h(6Z(0))))Z(3))Z(4))
h(h(h(h(h(01)h(23))h(h(45)h(67)))Z(3))Z(4))
h(h(h(h(h(01)h(23))h(h(45)h(67)))h(h(h(8Z(0))Z(1))Z(2)))Z(4))
h(h(h(h(h(01)h(23))h(h(45)h(67)))h(h(h(89)h(Z(0)Z(0)))Z(2)))Z(4))
h(h(h(h(h(01)h(23))h(h(45)h(67)))h(h(h(89)h(10Z(0)))Z(2)))Z(4))
h(h(h(h(h(01)h(23))h(h(45)h(67)))h(h(h(89)h(1011))h(h(Z(0)Z(0))Z(1))))Z(4))
h(h(h(h(h(01)h(23))h(h(45)h(67)))h(h(h(89)h(1011))h(h(12Z(0))Z(1))))Z(4))
h(h(h(h(h(01)h(23))h(h(45)h(67)))h(h(h(89)h(1011))h(h(1213)h(Z(0)Z(0)))))Z(4))
h(h(h(h(h(01)h(23))h(h(45)h(67)))h(h(h(89)h(1011))h(h(1213)h(14Z(0)))))Z(4))
h(h(h(h(h(01)h(23))h(h(45)h(67)))h(h(h(89)h(1011))h(h(1213)h(1415))))Z(4))
h(h(h(h(h(01)h(23))h(h(45)h(67)))h(h(h(89)h(1011))h(h(1213)h(1415))))h(h(h(h(16Z(0))Z(1))Z(2))Z(3)))
h(h(h(h(h(01)h(23))h(h(45)h(67)))h(h(h(89)h(1011))h(h(1213)h(1415))))h(h(h(h(1617)h(Z(0)Z(0)))Z(2))Z(3)))
h(h(h(h(h(01)h(23))h(h(45)h(67)))h(h(h(89)h(1011))h(h(1213)h(1415))))h(h(h(h(1617)h(18Z(0)))Z(2))Z(3)))
```

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
protolambda: not sure, catching cases where we don't type inputs properly is good imho.
And not sure about wrapping `bool` yet, made it an edge-case for now (see new commits)

protolambda: For now we have:
`TypeWithDefault = Union[DefaultingTypeMeta, bool]`
- parameter type checking in `ParamsMeta` accounts for this, and checks if it's either one of the two options (bool/DefaultingTypeMeta)
- `get_zero_value` checks if the type is a bool, otherwise it calls `typ.default()`.

We could wrap all bools. But unsure if it's worth the type purity (everything having a default, being able to check if something is a "SSZ" type), or if it just bothers us with `Bool(True)`-like problems.

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
protolambda: I think d83017d fixes this. BytesLike checks if it is a `bytes`, `AbstractList` goes through each element to check the type. `BytesN` and `Vector` check the length, then call the checking function of their super class.

protolambda:
@vbuterin Can I get a final review before updating the usages of these ssz types everywhere? (and fixing whatever small bugs I can find along the way)

test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
```diff
@@ -126,36 +133,41 @@ def mix_in_length(root, length):
...
-@infer_input_type
-def hash_tree_root(obj, typ=None):
+def item_length(typ):
+    if typ == bool:
+        return 1
+    elif issubclass(typ, uint):
+        return typ.byte_len
+    else:
+        return 32
+
```
vbuterin:
```suggestion
def chunk_count(typ):
    if is_basic_type(typ):
       return 1
   elif is_list_kind(typ) or is_vector_kind(typ):
       return (typ.length * item_length(typ.elem_type) + 31) // 32
   else:
       return len(typ.get_fields())
```

test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
```diff
@@ -126,36 +133,41 @@ def mix_in_length(root, length):
...
     if is_bottom_layer_kind(typ):
-        data = serialize_basic(obj, typ) if is_basic_type(typ) else pack(obj, read_elem_type(typ))
+        data = serialize_basic(obj, typ) if is_basic_type(typ) else pack(obj, typ.elem_type)
         leaves = chunkify(data)
     else:
         fields = get_typed_values(obj, typ=typ)
         leaves = [hash_tree_root(field_value, typ=field_typ) for field_value, field_typ in fields]
     if is_list_kind(typ):
-        return mix_in_length(merkleize_chunks(leaves), len(obj))
+        return mix_in_length(merkleize_chunks(leaves, pad_to=typ.length), len(obj))
```
vbuterin:
```suggestion
        return mix_in_length(merkleize_chunks(leaves, pad_to=chunk_count(typ)), len(obj))
```

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
protolambda: got rid of the unnecessary argument, it directly using the value type seems more sensible, and less error prone (you can't have a different type than the value actually is). And it is all nicely checking / type-hinting :)

protolambda:
- updated spec builder to handle container field dependencies of new style types
- implemented tests for typing + ssz implementation, new ssz seems to be working
- blocker: `NewType` doesn't type check. Need proper classes to declare element types and field types. Since a NewType around an int isn't a subclass of `SSZType`, or has any default value. I know @hwwhww is implementing replacement of newtypes with classes. If we get that merged, we can continue this PR using those class based declarations.

protolambda:
- [x] Spec builder updated to deal with new types, cleanly handle dependencies
- [x] merged in dev, no conflicts anymore (thx @hwwhww for mypy PR)
- [x] conforms fully to mypy and flake8 linting
- [x] phase-0 compiles (tests start, crash on phase-0)
- [ ] phase-1 compiles. Todo: need to assign lengths to list.
- [ ] run tests, see if test code needs to adapt new types somewhere
- [ ] double check ssz serialization behavior by comparing `ssz-static` generator output with `ssz-static` output on Dev. (disabling hash-tree-root, since that behavior changed)
- [ ] merge

specs/core/0_beacon-chain.md
```diff
@@ -1370,17 +1396,17 @@ def process_crosslinks(state: BeaconState) -> None:
 def get_base_reward(state: BeaconState, index: ValidatorIndex) -> Gwei:
     total_balance = get_total_active_balance(state)
     effective_balance = state.validators[index].effective_balance
-    return effective_balance * BASE_REWARD_FACTOR // integer_squareroot(total_balance) // BASE_REWARDS_PER_EPOCH
+    return Gwei(effective_balance * BASE_REWARD_FACTOR // integer_squareroot(total_balance) // BASE_REWARDS_PER_EPOCH)
 ```
 
 ```python
-def get_attestation_deltas(state: BeaconState) -> Tuple[List[Gwei], List[Gwei]]:
+def get_attestation_deltas(state: BeaconState) -> Tuple[Deltas, Deltas]:
```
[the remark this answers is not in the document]
protolambda: yes, not too happy with it yet. I'm going to change SSZ `List` to be a subclass of `list`, and if we make the limit-param optional, we may be able to just use SSZ `List` everywhere after all. But time and iteration, classes + hinting types mixed in Python are no fun.

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -1,137 +1,166 @@
...
+    byte_len = 1
+
+    def __new__(cls, value, *args, **kwargs):
+        if value < 0 or value > 1:
+            raise ValueError(f"value {value} out of bounds for bit")
+        return super().__new__(cls, value)
+
+    @classmethod
+    def default(cls):
+        return cls(False)
```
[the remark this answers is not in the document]
protolambda: fixed in f157745

test_libs/pyspec/eth2spec/utils/ssz/ssz_partials.py
```diff
@@ -0,0 +1,435 @@
...
+    """
+    Given a {generalized_index: chunk} map containing only leaves, fills intermediate
+    chunks as far as possible.
+    """
+    objects = {k: v for k, v in objects.items()}
+    keys = sorted(objects.keys())[::-1]
+    pos = 0
+    while pos < len(keys):
+        k = keys[pos]
+        if k in objects and k ^ 1 in objects and k // 2 not in objects:
```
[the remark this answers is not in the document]
protolambda: partials are out with the rebase update now

specs/core/0_beacon-chain.md
```diff
@@ -1247,21 +1273,21 @@ def get_total_active_balance(state: BeaconState) -> Gwei:
...
+def get_matching_source_attestations(state: BeaconState, epoch: Epoch) -> Iterable[PendingAttestation]:
     assert epoch in (get_current_epoch(state), get_previous_epoch(state))
     return state.current_epoch_attestations if epoch == get_current_epoch(state) else state.previous_epoch_attestations
 ```
 
 ```python
-def get_matching_target_attestations(state: BeaconState, epoch: Epoch) -> List[PendingAttestation]:
-    return [
+def get_matching_target_attestations(state: BeaconState, epoch: Epoch) -> Iterable[PendingAttestation]:
+    return [
```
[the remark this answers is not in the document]
protolambda: fixed in f157745

protolambda:
Rebased on `dev` branch to minimize change-set. Still a lot, here's an overview:

- Lists have a "length" now. But artificial, serialization is still the same. It's a limit imposed on lists, and padded to (virtually) when doing hash-tree-root. The result is that we have simple balanced binary trees with fixed depth everywhere: we can index everything we like with static generalized indices! This should enable the SSZ partials to be simple and consistent.
- **All (in-use) SSZ types are defined by pyspec now**. No edge-case prone half-python half-ssz. No `typing.NewType` or `typing.List` anymore, or double types (values having a class, and types being passed around). It's very well-defined now:
  - Types in SSZ are instances of `SSZType` (python metaclass). Such types are subclasses of `SSZValue` (also an instance of `SSZType`)
  - Basic types in SSZ are instances of `BasicType`, which is a subclass of SSZType.
    - Basic types: `uint, uint8/16/32/64/128/256, Bit`. And `uint8` has a `byte` alias.
  - Non-basic types are a `Series`, you can get an iterator to go through their data. Simplifies serialization.
  - For all SSZ types, you can simply call `.fixed_length()` to check if it's fixed length.
  - `Vector` and `List` are subclasses of `BaseList`, which is both a subclass of `list` and `Elements`
  - `BytesN` and `Bytes` are subclasses of `BaseBytes`, which is both a subclass of `bytes` and `Elements`
  - All these `Elements` subclasses share the unified type information from their type `ElementsType`. This enables you to get SSZ information about such type easily: `typ.elem_type` and `typ.length`. For `Bytes`/`BytesN`, `elem_type` is always `byte`.
  - These element subclasses can be parametrized, to fill these `ElementsType` properties in order of declaration. These params are type checked. So one can't create a `Vector[NonSSZClass, 123]`. The element type has to be an instance of `SSZType`.
    - If we like later, the parametrization code is easily re-used through inheritance of `ParamsBase` (makes instantiation of bare unparametrized types raise an exception) and `ParamsType` (fills the whatever non-empty attributes a type has declared in annotations with a parameter).
  - SSZ defaults are a lot cleaner now. Just call `typ.default()` to get a default instance of a SSZ type.
  - There is coercion of integers, bytes, etc. But only when:
    - same size `uint`
    - from default python types to whatever type it expects. (Types will check bounds of values!)
  - For safety (and non-python readers), negative index lookups are disabled, but there's a `.last()` for the most common use case.
  - You can subclass or alias `BytesN[100]` as `Bytes100`. Common ones are already provided: `4`, `32`, `48`, `96`.
  - Recently introduced spec types like `Hash` are simply a subclass of `Bytes32`. Fully compatible with mypy and the SSZ typing.


Other changes:
- spec builder:
  - cleans up re-initialization code of SSZ, to not make linting difficult to debug (remove comments to avoid line-length problems with higher than expected indentation)
  - clean up dependency checking in type-ordering
  - remove (now) unused parts
- Add some tests for SSZ types and implementation. Still need to go through comparison of `ssz_static` output with `dev` branch
- **Merkleization is implemented differently,  but same effective behavior**. To do padding of giant-limit lists effectively, merkleization needs to handle the zeroes as virtual, and just merge in whatever order zero-hash where needed. Also more compatible with future sparse-merkleization behavior
- Updated SSZ usages:
  - randomization of SSZ values (used by ssz-static generator to create instances with random contents)
  - decoding/encoding of SSZ values (used for spec-test output/generators/debugging)
  - Fuzzing SSZ util: need to translate between py-ssz type and pyspec. (we use their decoder, since that SSZ direction is not implemented in the pyspec and more efficient for fuzzing)

protolambda:
Really close to passing all tests now. Also found two data-type issues with the new ssz: signature being used as bytes32  in a custody test, and the deposit-index test being invalid now that DepositData has no index (is that being re-introduced?)

protolambda:
Note that I added tests for SSZ typing, hence netto more than 300 new lines. Non-testing code decreased however :tada:

protolambda:
Yay, all tests pass with new SSZ. I'll double check the `ssz_static` behavior, and then we can technically merge.
Few points to address with help from others:
- Phase 1 spec needs list length limits, they are `PLACEHOLDER` now.
- Phase 1 test helpers has a 32-byte mask that is filled with a signature (96 bytes). I just sliced it. @CarlBeek Can you check, and fix if necessary?
- Deposit-index test is commented for now, @JustinDrake Are we putting deposit-index back in or not?
- Both specs could improve type checking a bit. We have a `typing.List` `ssz_typing.List` name collision. However, using tuples and sets is both more safe (tuple immutability) and more documenting (unordered output with sets). @djrtwo Can you do some final type hint polishing, based on your previous review?

JustinDrake:
> Are we putting deposit-index back in or not?

No deposit index in `DepositData`. (It only helps to have it in the `Deposit` log.)

protolambda:
Quick note before merging: this PR changes hash-tree-root for lists. So that means clients will need to update their SSZ implementation for new hash-tree-root (lists virtually padded to their limit during merkleization). Do we want that in spec-freeze? I suppose we do, but maybe we want to do one final release of test-vectors (0.7.2/3) with the old SSZ, for clients to work on later?

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -1,142 +1,171 @@
...
 
+    def is_fixed_size(cls):
+        return True
+
+
+class BasicValue(int, SSZValue, metaclass=BasicType):
+    pass
+
+
+class Bit(BasicValue):  # can't subclass bool.
```
hwwhww: In https://github.com/ethereum/eth2.0-specs/blob/dev/specs/simple-serialize.md#basic-types, it's still be known as `bool`. I think either (i) adding some description of `Bit` in the phase 0 spec, or (ii) updating the markdown spec.

I guess (i) would be less overhead for the implementers.

specs/core/1_shard-data-chains.md
```diff
@@ -282,9 +288,9 @@ Let:
 * `candidate` be a candidate `ShardBlock` for which validity is to be determined by running `is_valid_shard_block`
 
 ```python
-def is_valid_shard_block(beacon_blocks: List[BeaconBlock],
+def is_valid_shard_block(beacon_blocks: TypingList[BeaconBlock],
                          beacon_state: BeaconState,
-                         valid_shard_blocks: List[ShardBlock],
+                         valid_shard_blocks: Iterable[ShardBlock],
```
hwwhww: Would `valid_shard_blocks` be `Tuple/List`? If so, recommend using `Sequence`. 
```suggestion
                         valid_shard_blocks: Sequence[ShardBlock],
```

specs/core/1_shard-data-chains.md
```diff
@@ -348,7 +354,7 @@ Let:
 * `candidate` be a candidate `ShardAttestation` for which validity is to be determined by running `is_valid_shard_attestation`
 
 ```python
-def is_valid_shard_attestation(valid_shard_blocks: List[ShardBlock],
+def is_valid_shard_attestation(valid_shard_blocks: Iterable[ShardBlock],
```
hwwhww:
```suggestion
def is_valid_shard_attestation(valid_shard_blocks: Sequence[ShardBlock],
```

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -1,142 +1,171 @@
+from typing import Dict, Iterator
 from types import GeneratorType
-from typing import List, Iterable, TypeVar, Type, NewType
-from typing import Union
-from typing_inspect import get_origin
```
hwwhww: Could remove `typing_inspect` from `requirements.txt` and `setup.py`.

hwwhww:
Actually I approved too fast. Would be great to have some tests for merkle minimal (`tests_merkle_minimal.py`?)

protolambda:
Small underflow problem (incorrectly trying to attest pre-genesis slot) in finality tests. `slot_to_attest = post_state.slot - spec.MIN_ATTESTATION_INCLUSION_DELAY + 1` CI is otherwise fine. Will resolve later.

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
protolambda: See d1fa3ac

CarlBeek:
>  Phase 1 test helpers has a 32-byte mask that is filled with a signature (96 bytes). I just sliced it. @CarlBeek Can you check, and fix if necessary?

Looks good to me! The purpose of the mask is to hide the hide the secret and to do this, it only needs to be 32 random bytes.

specs/core/0_beacon-chain.md
```diff
@@ -810,7 +819,7 @@ def get_beacon_proposer_index(state: BeaconState) -> ValidatorIndex:
 ### `verify_merkle_branch`
 
 ```python
-def verify_merkle_branch(leaf: Hash, proof: List[Hash], depth: int, index: int, root: Hash) -> bool:
+def verify_merkle_branch(leaf: Hash, proof: Tuple[Hash, ...], depth: int, index: int, root: Hash) -> bool:
```
vbuterin: This is especially a decrease in understandability imo. It implies to a casual reader that it's a tuple where only the first item needs to be a hash.

specs/core/0_beacon-chain.md
```diff
@@ -629,11 +649,13 @@ def is_slashable_validator(validator: Validator, epoch: Epoch) -> bool:
 ### `get_active_validator_indices`
 
 ```python
-def get_active_validator_indices(state: BeaconState, epoch: Epoch) -> List[ValidatorIndex]:
+def get_active_validator_indices(state: BeaconState, epoch: Epoch) -> List[ValidatorIndex, VALIDATOR_REGISTRY_SIZE]:
     """
     Get active validator indices at ``epoch``.
     """
-    return [i for i, v in enumerate(state.validators) if is_active_validator(v, epoch)]
+    return List[ValidatorIndex, VALIDATOR_REGISTRY_SIZE](
```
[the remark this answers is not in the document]
vbuterin: The only thing I could come up with is to go all the way and make sure _every_ variable we work with is an SSZ of some kind. But that feels overkill, and would require explicit SSZ classes (don't need to be serializable or hashable, but do need to be implemented) for dicts and sets.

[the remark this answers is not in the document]
vbuterin: Using a synonym for "list" seems least bad if we can come up with one.

specs/core/0_beacon-chain.md
```diff
@@ -1288,22 +1298,22 @@ def get_matching_head_attestations(state: BeaconState, epoch: Epoch) -> List[Pen
 
 ```python
 def get_unslashed_attesting_indices(state: BeaconState,
-                                    attestations: List[PendingAttestation]) -> List[ValidatorIndex]:
+                                    attestations: Iterable[PendingAttestation]) -> Iterable[ValidatorIndex]:
```
vbuterin: Is it possible to use `Set`?

I think we need to remember that we're writing this spec for humans' benefit first and foremost, and the mypy checker second, and `Iterable` is not really understandable to the average reader.

test_libs/pyspec/eth2spec/test/helpers/custody.py
```diff
@@ -20,14 +20,14 @@ def get_valid_early_derived_secret_reveal(spec, state, epoch=None):
...
-        message_hash=spec.hash_tree_root(epoch),
+        message_hash=spec.hash_tree_root(spec.Epoch(epoch)),
         privkey=privkeys[masker_index],
         domain=spec.get_domain(
             state=state,
             domain_type=spec.DOMAIN_RANDAO,
             message_epoch=epoch,
         ),
-    )
+    )[:32]  # TODO(Carl): mask is 32 bytes, and signature is 96? Correct to slice the first 32 out?
```
vbuterin: Is the point of this to be more MPC friendly than hashing the signature?

specs/core/0_beacon-chain.md
hwwhww: There's `Sequence[<type>]` for `Union[List[<type>], Tuple[<type>, ...]] case:
```suggestion
                                    attestations: Iterable[PendingAttestation]) -> Sequence[ValidatorIndex]:
```

Might be more natural than `Iterable` and `Set`? 😄

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
protolambda: keeping this here the old way for now with `Sequence`, see full explanation of types in comment.

specs/core/0_beacon-chain.md
```diff
@@ -1247,21 +1273,21 @@ def get_total_active_balance(state: BeaconState) -> Gwei:
...
-def get_matching_source_attestations(state: BeaconState, epoch: Epoch) -> List[PendingAttestation]:
+def get_matching_source_attestations(state: BeaconState, epoch: Epoch) -> Iterable[PendingAttestation]:
     assert epoch in (get_current_epoch(state), get_previous_epoch(state))
     return state.current_epoch_attestations if epoch == get_current_epoch(state) else state.previous_epoch_attestations
 ```
 
 ```python
-def get_matching_target_attestations(state: BeaconState, epoch: Epoch) -> List[PendingAttestation]:
-    return [
+def get_matching_target_attestations(state: BeaconState, epoch: Epoch) -> Iterable[PendingAttestation]:
```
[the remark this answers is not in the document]
protolambda: Made it `Sequence`, see explanation below.

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
protolambda: It's `Tuple[Sequence[Gwei], Sequence[Gwei]]` now, see explanation.

specs/core/1_shard-data-chains.md
protolambda: Put sequece in more places now, thanks for suggestion.

specs/core/0_beacon-chain.md
protolambda: It's `Set`, since it's only used for iteration/inclusion.

protolambda: *non-order-specific iteration

specs/core/0_beacon-chain.md
protolambda: Fixed, it's `Sequence` now, see explanation below.

test_libs/pyspec/eth2spec/test/helpers/custody.py
protolambda: See older status update and comment by Carl. It's good enough for now, but (as I also suggested to him), a hash may be used too. It's something for a later PR however, as it's mostly just a problem uncovered by the typing we have now.

protolambda:
Rationale for `Set` + `Sequence`:

- We make `List` and `Vector` as *SSZ-ONLY*
- We make `Sequence` and `Set` *LOGIC-ONLY*.
  - Very minimal on data types, with two opposing ends:
  - `Sequence` encompasses both `List` and `Vector`, and `list` (python). We use this to express that an input/output has an order that should be maintained for spec conformance. It supports lookups by index, and getting the length.
  - `Set` is a special case, where we describe that elements can be any order, as long as they are unique in the collection.
- We may want to introduce a custom `OrderedSet`, to clean up typing:
  - In many places, we want to express we expect unique elements
  - But we also like to give them an ordered iteration; keep indices in their natural order.
  - For now we use `Sequence` instead, and don't express the uniqueness, since it is guaranteed anyway, because these ordered sets of indices are derived from the state registry.

protolambda:
Ok, PR mostly ready, summary of last TODOs:
- [ ] Review naming of "Bit", possibly adapt SSZ, or make it clear in spec that a "Bit" is a SSZ bool. See https://github.com/ethereum/eth2.0-specs/pull/1180#discussion_r296321754
- [ ] [As per suggestion from Hww](https://github.com/ethereum/eth2.0-specs/pull/1180#issuecomment-504509316) add tests for new merkleization. Already verified it roughly, but tests would be good: https://github.com/ethereum/eth2.0-specs/pull/1180#issuecomment-502373800
- [ ] Verify ssz behavior by comparing `ssz_static` generator output, for serialization at least (hash-tree-root-changes).

protolambda:
Verified that:
- objects of fixed-size have same hash-tree-root and serialization as in the dev branch
- objects of variable size:
  - have same serialization (up to point where RNG changes due to var. length limiting, see below)
  - different hash-tree-root due to limit on lists
The RNG diverges when the `ssz_static` generator passes `transfers` (limited to 0); it generates different values, since bigger are simply not allowed within the list-limit. So I couldn't compare the full data, but what I could compare matched expectations.

And we've tests for the new merkleization now, see `test_libs/pyspec/eth2spec/utils/test_merkle_minimal.py`

protolambda:
Also, this PR introduces 230+59+119=408 lines of ssz/merkleization test-code, and includes a bunch of fixes + updated tests for finalization (thanks @djrtwo). I'm happy we are actually reducing lines in the spec util code, while improving in style and tests :tada:

test_libs/pyspec/eth2spec/test/phase_0/block_processing/test_process_deposit.py
```diff
@@ -114,21 +114,6 @@ def test_invalid_withdrawal_credentials_top_up(spec, state):
     yield from run_deposit_processing(spec, state, deposit, validator_index, valid=True, effective=True)
 
 
-@with_all_phases
-@spec_state_test
```
djrtwo: Why is this test wiped?

test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
```diff
@@ -14,68 +10,47 @@
...
             return b'\x00'
     else:
-        raise Exception("Type not supported: {}".format(typ))
+        raise Exception(f"Type not supported: {type(value)}")
 
 
-def deserialize_basic(value, typ):
-    if is_uint_type(typ):
+def deserialize_basic(value, typ: BasicType):
+    if issubclass(typ, uint):
```
djrtwo: why use `issubclass` here but `isinstance` in `serialize_basic`?

test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
```diff
@@ -14,68 +10,47 @@
...
         return typ(int.from_bytes(value, 'little'))
-    elif is_bool_type(typ):
+    elif issubclass(typ, Bit):
```
djrtwo: ditto

test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
```diff
@@ -14,68 +10,47 @@
...
         assert value in (b'\x00', b'\x01')
-        return True if value == b'\x01' else False
+        return Bit(value == b'\x01')
```
djrtwo: Can use `typ` instead of `Bit` here, right?
Is there a reason to not be consistent with `uint` by using `typ`?

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -1,149 +1,178 @@
...
-    if hasattr(typ, '__supertype__'):
-        # get the type that the NewType is wrapping
-        typ = typ.__supertype__
 
-    return isinstance(typ, type) and issubclass(typ, int) and not issubclass(typ, bool)
+def coerce_type_maybe(v, typ: SSZType, strict: bool = False):
+    v_typ = type(v)
+    # shortcut if it's already the type we are looking for
+    if v_typ == typ:
+        return v
```
djrtwo: Should this return a new instance of the value rather than the same to avoid mutability issues?

specs/core/0_beacon-chain.md
```diff
@@ -310,7 +317,7 @@ class Validator(Container):
     pubkey: BLSPubkey
     withdrawal_credentials: Hash  # Commitment to pubkey for withdrawals and transfers
     effective_balance: Gwei  # Balance at stake
-    slashed: bool
+    slashed: Bit
```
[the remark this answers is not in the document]
protolambda: Agree there. But using `bool` (lowercase, the python bool) means:
- handling it as an edge-case of types, everywhere. `bool` is not a subclass of `SSZValue`, or a type of `SSZType, or a subclass of `BasciValue` etc. Typing generally breaks. This is not typescript, where we just say `SSZValue = OurSSZValue | bool`, nor can we (or should we) really attach anything to the python `bool` class.
- Missing `.default()`, `.is_fixed_size()`, and not being able to add additional things into its class in the future (caching, partials, building gen. indices etc.)

We can also make `Bit` a `Bool`, and subclass `Bit` from it for style where we do like it (similar with `uint8` <> `byte`)

test_libs/pyspec/eth2spec/test/phase_0/block_processing/test_process_deposit.py
protolambda: Side effect of discovering (with strong typing) that `index` is a non-existent field now, it was removed. See https://github.com/ethereum/eth2.0-specs/pull/1180#issuecomment-504218583

test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
protolambda: Values are instances of classes. types are subclasses of classes. Types are instances of type-subclasses (meta classes). Value here is a a python value, hence the type is added as an argument here, to describe what to handle the value like.

test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
protolambda: ditto. (https://github.com/ethereum/eth2.0-specs/pull/1180/files/1972ccaaf3a8eccc2f606e1b82bf50d4849a4621#r296918423)

test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
protolambda: good spot. `typ` is generally better, as it creates the proper subclass (e.g. byte, instead of uint8). But with coercion, it's eventually put into the right type. Will fix this though.

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
protolambda: No, if we set a value/member, we want to get the proper python object into that spot. Only if it's not the same type, then we coerce it. But for lists etc, this only really means that the elements get a new container. And for basic types (uint, bool), it doesn't matter if we copy or not.

specs/core/0_beacon-chain.md
```diff
@@ -284,6 +289,8 @@ We define the following Python custom types for type hinting and readability:
 | `BLSPubkey` | `Bytes48` | a BLS12-381 public key |
 | `BLSSignature` | `Bytes96` | a BLS12-381 signature |
 
+`Deltas` is a non-SSZ type, a series of changes applied to balances, optimized by clients.
+
```
```suggestion
```

specs/core/0_beacon-chain.md
```diff
@@ -235,6 +236,10 @@ The following values are (non-configurable) constants used throughout the specif
 | - | - | :-: | :-: |
 | `EPOCHS_PER_HISTORICAL_VECTOR` | `2**16` (= 65,536) | epochs | ~0.8 years |
 | `EPOCHS_PER_SLASHED_BALANCES_VECTOR` | `2**13` (= 8,192) | epochs | ~36 days |
+| `RANDAO_MIXES_LENGTH` | `2**13` (= 8,192) | epochs | ~36 days |
```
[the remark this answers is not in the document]
protolambda: See 82ae180

specs/core/0_beacon-chain.md
```diff
@@ -222,6 +222,7 @@ The following values are (non-configurable) constants used throughout the specif
 | `ACTIVATION_EXIT_DELAY` | `2**2` (= 4) | epochs | 25.6 minutes |
 | `SLOTS_PER_ETH1_VOTING_PERIOD` | `2**10` (= 1,024) | slots | ~1.7 hours |
 | `SLOTS_PER_HISTORICAL_ROOT` | `2**13` (= 8,192) | slots | ~13 hours |
+| `HISTORICAL_ROOTS_LENGTH` | `2**24` (= 16,777,216) | historical roots | ~26,131 years |
```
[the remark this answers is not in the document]
protolambda: See 82ae180

specs/core/0_beacon-chain.md
```diff
@@ -235,6 +236,10 @@ The following values are (non-configurable) constants used throughout the specif
 | - | - | :-: | :-: |
```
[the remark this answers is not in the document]
protolambda: See 82ae180

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
protolambda: See 5989e5c, executed the `Bool` idea

specs/core/0_beacon-chain.md
```diff
@@ -235,6 +236,10 @@ The following values are (non-configurable) constants used throughout the specif
...
 | `EPOCHS_PER_HISTORICAL_VECTOR` | `2**16` (= 65,536) | epochs | ~0.8 years |
 | `EPOCHS_PER_SLASHED_BALANCES_VECTOR` | `2**13` (= 8,192) | epochs | ~36 days |
+| `RANDAO_MIXES_LENGTH` | `2**13` (= 8,192) | epochs | ~36 days |
+| `ACTIVE_INDEX_ROOTS_LENGTH` | `2**13` (= 8,192) | epochs | ~36 days |
+| `SLASHED_EXIT_LENGTH` | `2**13` (= 8,192) | epochs | ~36 days |
+| `VALIDATOR_REGISTRY_SIZE` | `2**40` (= 1,099,511,627,776) | | |
```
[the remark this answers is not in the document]
protolambda: Mostly old constants anyway, see 82ae180, cleaned it up.

specs/core/0_beacon-chain.md
```diff
@@ -1185,8 +1193,10 @@ def get_genesis_beacon_state(deposits: List[Deposit], genesis_time: int, genesis
             validator.activation_eligibility_epoch = GENESIS_EPOCH
             validator.activation_epoch = GENESIS_EPOCH
 
-    # Populate active_index_roots
-    genesis_active_index_root = hash_tree_root(get_active_validator_indices(state, GENESIS_EPOCH))
+    # Populate active_index_roots    
+    genesis_active_index_root = hash_tree_root(
+        List[ValidatorIndex, VALIDATOR_REGISTRY_SIZE](get_active_validator_indices(state, GENESIS_EPOCH))
```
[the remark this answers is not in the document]
protolambda: I'd be up for that. I know I'm forced to in Go to add the additional limit-type-data. But also like simplicity, this is the only use in the spec, aside from the definition. Maybe we can add it later after feedback, it's a cosmetic change anyway.

specs/core/0_beacon-chain.md
```diff
@@ -1261,25 +1271,25 @@ def process_epoch(state: BeaconState) -> None:
 
 ```python
 def get_total_active_balance(state: BeaconState) -> Gwei:
-    return get_total_balance(state, get_active_validator_indices(state, get_current_epoch(state)))
+    return get_total_balance(state, set(get_active_validator_indices(state, get_current_epoch(state))))
```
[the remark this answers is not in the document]
protolambda: See 9fb5806

specs/core/1_shard-data-chains.md
```diff
@@ -377,9 +383,9 @@ Let:
 
 ```python
 def is_valid_beacon_attestation(shard: Shard,
-                                shard_blocks: List[ShardBlock],
+                                shard_blocks: Sequence[ShardBlock],
                                 beacon_state: BeaconState,
-                                valid_attestations: List[Attestation],
+                                valid_attestations: Set[Attestation],
```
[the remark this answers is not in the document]
protolambda: see https://github.com/ethereum/eth2.0-specs/commit/9fb58067645da1aa02dcfa995cb61fb86a2dff16

test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
protolambda: see c73417b

specs/core/0_beacon-chain.md
```diff
@@ -903,7 +911,7 @@ def bytes_to_int(data: bytes) -> int:
 ### `get_total_balance`
 
 ```python
-def get_total_balance(state: BeaconState, indices: List[ValidatorIndex]) -> Gwei:
+def get_total_balance(state: BeaconState, indices: Iterable[ValidatorIndex]) -> Gwei:
```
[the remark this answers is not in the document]
protolambda: Outdated, we're limiting it to `Sequence` and `Set` now. And for total-balance `Set` makes more sense (don't want to count things twice).

scripts/build_spec.py
```diff
@@ -12,12 +12,7 @@
 
 
 PHASE0_IMPORTS = '''from typing import (
-    Any,
-    Callable,
-    Dict,
-    List,
-    Set,
-    Tuple,
+    Any, Callable, Iterable, Dict, Set, Sequence, Tuple,
```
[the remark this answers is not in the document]
protolambda: ditto, Iterable is already gone :)

scripts/build_spec.py
```diff
@@ -30,27 +25,20 @@
...
 '''
 PHASE1_IMPORTS = '''from typing import (
-    Any,
-    Callable,
-    Dict,
-    List,
-    Optional,
-    Set,
-    Tuple,
+    Any, Callable, Dict, Iterable, Optional, Set, Sequence, Tuple,
```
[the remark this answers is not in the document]
protolambda: ditto

specs/core/1_shard-data-chains.md
```diff
@@ -242,11 +248,11 @@ def verify_shard_attestation_signature(state: BeaconState,
 ### `compute_crosslink_data_root`
 
 ```python
-def compute_crosslink_data_root(blocks: List[ShardBlock]) -> Bytes32:
+def compute_crosslink_data_root(blocks: Sequence[ShardBlock]) -> Bytes32:
     def is_power_of_two(value: int) -> bool:
         return (value > 0) and (value & (value - 1) == 0)
 
-    def pad_to_power_of_2(values: List[bytes]) -> List[bytes]:
+    def pad_to_power_of_2(values: TypingList[bytes]) -> TypingList[bytes]:
```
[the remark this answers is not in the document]
protolambda: Changed it to `MutableSequence` for now, along with another `TypingList` usage. Phase-1 is still a lot of dead (untested) code, e.g. the application of a list value to hash-tree-root a few lines down is outdated. Added suggestions in a comment to fix later, but do not want to get into phase-1 changes. This PR is more than big enough already.

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
protolambda: This was resolved a bit ago, it's `Bool` now, to keep it within the typing support of ssz. And we've `Bit` as alias (subclass of `Bool`).

PR #1181, opened 2019-06-14
skip transfer tests and enable the commented tests

hwwhww:
* Skip `test_transfer`
* Enable `test_empty_epoch_transition_not_finalizing` and `test_eth1_data_votes`: need to check with @protolambda @djrtwo, why they were commented out? I can add `skip` to them if they should be disabled.


protolambda:
The pytest import isn't working in generators, so it breaks test-vector generation. Either wrap it in a custom `skip` flag, or comment it out... We'll introduce it back at some point, but it's really testing unused (and possibly to be removed from spec for hard-fork test idea) functionality.

hwwhww:
@protolambda oh dear, got it now. 👍 

Besides `pytest`, are `test_empty_epoch_transition_not_finalizing` and `test_eth1_data_votes` ready to be included in the tests now?

protolambda:
kind of. I think it's was a mix of speed and update issues. Some of these tests in mainnet config are unbearably slow. Agree that commenting them out is not the solution. But at what running duration do you start ignoring a test? 10 minutes? 1 hour? If we're not running them, we may as well not have them.

hwwhww:
@protolambda 
running `make ./eth2.0-spec-tests/tests/sanity` on my laptop:
before this PR: 15m 38s
+these two test cases: 23m 20s

djrtwo:
I'm working on getting the v0.7.1 release ready. Looks like this is not going to make it in, but we should still discuss a better solution. Correct?

hwwhww:
Targeting on `dev` branch now!

protolambda:
- `test_transfer` is not to be re-enabled now. It's hacking the max-transfer constant. Which is a data type, also requiring the list limit to change now with new SSZ. We should properly run it with a config that declares the change. So that clients can actually also run it.
- `test_eth1_data_votes` is re-enabled for minimal configuration, and expanded to cover the two cases: consensus, and no consensus about a new eth1 data. It's in PR #1206 
- `test_empty_epoch_transition_not_finalizing` , updated it to use the proper tools, sign the block properly, and run in minimal configuration only (same effect of calling epoch-transition a few times anyway, but mainnet takes a very, very long time). Also in PR #1206

protolambda:
Closing this issue now that the bulk is solved. We'll test max-transfers when the time comes (when we have any configuration that enables transfers). We can't really test it like previously now, *since the data itself is not valid anymore by ssz limit*.

PR #1182, opened 2019-06-14
generate coverage reports in make test, open as html site

protolambda:
`make test` to run tests, and generate the coverage report along with it.
`make open_cov` to open the report in your browser (using `open` (OSX) or `xdg-open` (Linux)). If you like to open it manually, just open `test_libs/pyspec/.htmlcov/index.html`, and book-mark it for later.

Note: we can't really export it to a service like codecov unfortunately, since it's generated code that is being tests, which can't be linked back to the repo. One hacky solution would be to export the source spec.py code and coverage to an external repo, and run the codecov upload from there. But local coverage checking is far less hacky, and already a good step forward.

preview:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,10 @@ eth2.0-spec-tests/
 # Dynamically built from Markdown spec
 test_libs/pyspec/eth2spec/phase0/spec.py
 test_libs/pyspec/eth2spec/phase1/spec.py
+
+# coverage reports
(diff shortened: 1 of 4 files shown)


JustinDrake:
`make test` gives me an error

Screenshot 2019-06-15 at 16 16 37

protolambda:
@JustinDrake `make install_test` for new coverage support

PR #1186, opened 2019-06-17
Updates to SSZ Merkle proofs phase 0 doc

vbuterin:


hwwhww:
`specs/light_client/merkle_proofs.md` is coupled with SSZ partial implementation. IMO it seems more reasonable to move the code part into the SSZ module. :'(

vbuterin:
Where's the "let's discuss on the next call" emoji where you need one :smile:

hwwhww:
fyi: the `is_*_kind` and `is_*_type` APIs are removed in #1180.

djrtwo:
What's the status here @vbuterin ?

specs/light_client/merkle_proofs.md
```diff
@@ -116,72 +180,98 @@ x x . . . . x *
...
+def extract_value_at_path(chunks: Dict[int, Bytes32], typ: Type, path: List[Union[int, str]]) -> Any:
+    """
+    Provides the value of the element in the object represented by the given encoded SSZ partial at
+    the given path. Returns a KeyError if that path is not covered by this SSZ partial.
+    """
+    root = 1
+    for p in path:
+        if p == '__len__':
+            return deserialize_basic(chunks[root * 2 + 1][:8], uint64)
+        if iissubclass(typ, (List, Bytes)):
```
djrtwo:
```suggestion
        if issubclass(typ, (List, Bytes)):
```

specs/light_client/merkle_proofs.md
```diff
@@ -64,43 +64,107 @@ y_data_root  len(y)
...
+
+def get_generalized_index(typ: Type, path: List[Union[int, str]]) -> GeneralizedIndex:
+    """
+    Converts a path (eg. `[7, "foo", 3]` for `x[7].foo[3]`, `[12, "bar", "__len__"]` for
+    `len(x[12].bar)`) into the generalized index representing its position in the Merkle tree.
+    """
+    for p in path:
+        assert not is_basic_type(typ)  # If we descend to a basic type, the path cannot continue further
+        if p == '__len__':
+            typ, root = uint256, root * 2 + 1 if issubclass(typ, (List, Bytes)) else None
```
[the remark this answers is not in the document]
vbuterin: I think using uint64 is fine!

specs/light_client/merkle_proofs.md
```diff
@@ -64,43 +64,107 @@ y_data_root  len(y)
...
+        return typ.get_field_names().index(index), 0, item_length(get_elem_type(typ, index))
+    else:
+        raise Exception("Only lists/vectors/containers supported")
+
+
+def get_generalized_index(typ: Type, path: List[Union[int, str]]) -> GeneralizedIndex:
+    """
+    Converts a path (eg. `[7, "foo", 3]` for `x[7].foo[3]`, `[12, "bar", "__len__"]` for
+    `len(x[12].bar)`) into the generalized index representing its position in the Merkle tree.
+    """
```
vbuterin:
```suggestion
    """
    root = 1
```

specs/light_client/merkle_proofs.md
```diff
@@ -64,43 +64,108 @@ y_data_root  len(y)
...
+    root = 1
+    for p in path:
+        assert not is_basic_type(typ)  # If we descend to a basic type, the path cannot continue further
+        if p == '__len__':
+            typ, root = uint256, root * 2 + 1 if issubclass(typ, (List, Bytes)) else None
```
vbuterin:
```suggestion
            typ, root = uint64, root * 2 + 1 if issubclass(typ, (List, Bytes)) else None
```

specs/light_client/merkle_proofs.md
```diff
@@ -64,43 +59,109 @@ y_data_root  len(y)
...
+        return 32
+        
+        
+def get_elem_type(typ: ComplexType, index: int) -> Type:
+    """
+    Returns the type of the element of an object of the given type with the given index
+    or member variable name (eg. `7` for `x[7]`, `"foo"` for `x.foo`)
+    """
+    return typ.get_fields()[key] if issubclass(typ, Container) else typ.elem_type        
+
```
djrtwo:
```suggestion


```

specs/light_client/merkle_proofs.md
```diff
@@ -64,43 +59,109 @@ y_data_root  len(y)
...
+    else:
+        return 32
+        
+        
+def get_elem_type(typ: ComplexType, index: int) -> Type:
+    """
+    Returns the type of the element of an object of the given type with the given index
+    or member variable name (eg. `7` for `x[7]`, `"foo"` for `x.foo`)
+    """
+    return typ.get_fields()[key] if issubclass(typ, Container) else typ.elem_type
```
djrtwo: `key` is not defined

specs/light_client/merkle_proofs.md
```diff
@@ -116,72 +177,98 @@ x x . . . . x *
...
-                if (index ^ 1) not in redundant_indices:
-                    break
-                index //= 2
-    return [i for i in proof if i not in tree_indices]
-```
-
-Generating a proof is simply a matter of taking the node of the SSZ hash tree with the union of the given generalized indices for each index given by `get_proof_indices`, and outputting the list of nodes in the same order.
-
-Here is the verification function:
+```
```
djrtwo:
```suggestion
```python
```

specs/light_client/merkle_proofs.md
```diff
@@ -64,43 +59,131 @@ y_data_root  len(y)
...
+    else:
+        return 32
+        
+        
+def get_elem_type(typ: ComplexType, index: int) -> Type:
+    """
+    Returns the type of the element of an object of the given type with the given index
+    or member variable name (eg. `7` for `x[7]`, `"foo"` for `x.foo`)
+    """
+    return typ.get_fields()[key] if issubclass(typ, Container) else typ.elem_type
```
vbuterin:
```suggestion
    return typ.get_fields()[index] if issubclass(typ, Container) else typ.elem_type        
```

specs/light_client/merkle_proofs.md
```diff
@@ -64,43 +59,131 @@ y_data_root  len(y)
...
+    """
+    Returns the number of bytes in a basic type, or 32 (a full hash) for compound types.
+    """
+    if issubclass(typ, BasicValue):
+        return typ.byte_len
+    else:
+        return 32
+        
+        
+def get_elem_type(typ: ComplexType, index: int) -> Type:
```
vbuterin:
```suggestion
def get_elem_type(typ: ComplexType, index: Union[int, str]) -> Type:
```

specs/light_client/merkle_proofs.md
```diff
@@ -64,43 +59,131 @@ y_data_root  len(y)
...
-    else:
-        raise Exception("Unknown type / path")
+def concat_generalized_indices(*indices: Sequence[GeneralizedIndex]) -> GeneralizedIndex:
+    """
+    Given generalized indices i1 for A -> B, i2 for B -> C .... i_n for Y -> Z, returns
+    the generalized index for A -> Z.
+    """
+    o = GeneralizedIndex(1)
+    for i in indices:
+        o = o * get_previous_power_of_2(i) + i
```
protolambda: Do I read this right? We do padding by the length that does not include the leading 1 bit, but not actually remove the bit from the index before appending the bits to the output?

Example of what it looks like is happening here (assuming `get_previous_power_of_2(i)` just returns the closest power of 2 at `i` or lower):
```txt
indices = [ # (in binary for example purposes)
0b10001
0b1011
0b110001
]
o = 1
o = 1 0000 # pad
o = 1 0 0001 # add
o =    1 0001 000 # pad
o =    1 0010 011 # add
o =    1 0001 011 00000 # pad
o =    1 0001 100 10001 # add
```
What I expected:
```txt
o = 1
o = 1 0000 # pad
o = 1 0001 # add
o = 1 0001 000 # pad
o = 1 0001 011 # add
o = 1 0001 011 00000 # pad
o = 1 0001 011 10001 # add
```

specs/light_client/merkle_proofs.md
```diff
@@ -64,43 +59,131 @@ y_data_root  len(y)
...
+```
+
+#### `get_generalized_index_length`
+
+```python
+def get_generalized_index_length(index: GeneralizedIndex) -> int:
+   """
+   Returns the length of a path represented by a generalized index.
+   """
+   return log(index)
```
protolambda: Maybe use `bit_length()`, or at least `log2()`? And maybe explicitly rounded to an int, if choosing the log function?

specs/light_client/merkle_proofs.md
```diff
@@ -116,72 +199,69 @@ x x . . . . x *
...
+    objects = {
+        **{index:node for index, node in zip(indices, leaves)},
+        **{index:node for index, node in zip(helper_indices, proof)}
+    }
+    keys = sorted(objects.keys())[::-1]
+    pos = 0
+    while pos < len(keys):
+        k = keys[pos]
+        if k in objects and k ^ 1 in objects and k // 2 not in objects:
+            objects[k // 2] = hash(objects[k & -2] + objects[k | 1])
```
protolambda:
```suggestion
            objects[k // 2] = hash(objects[(k | 1) ^ 1] + objects[k | 1])
```
Not sure about this, python doesn't seem to have a `&^` operator, and negative integers for bit-masks feel a bit much for a spec (although I enjoy bit-magic). This makes it clear that they are sibling indices at least.

specs/light_client/merkle_proofs.md
vbuterin: Fixed! For inputs `(10001, 1011, 110001)` the output should indeed be `1 0001 011 10001` now.

specs/light_client/merkle_proofs.md
vbuterin: Fixed to `log2`. I was assuming we have `log2` explicitly defined elsewhere; if not happy to do `bit_length()` instead.

specs/light_client/merkle_proofs.md
```diff
@@ -64,43 +59,132 @@ y_data_root  len(y)
...
+            typ, root = uint64, root * 2 + 1 if issubclass(typ, (List, Bytes)) else None
+        else:
+            pos, _, _ = get_item_position(typ, p)
+            root = root * (2 if issubclass(typ, (List, Bytes)) else 1) * next_power_of_two(chunk_count(typ)) + pos
+            typ = get_elem_type(typ, p)
+    return root
 ```
 
-We can now define a function `get_generalized_indices(object: Any, path: List[int], root: int=1) -> List[int]` that converts an object and a path to a set of generalized indices (note that for constant-sized objects, there is only one generalized index and it only depends on the path, but for dynamically sized objects the indices may depend on the object itself too). For dynamically-sized objects, the set of indices will have more than one member because of the need to access an array's length to determine the correct generalized index for some array access.
+### Helpers for generalized indices
```
vbuterin:
```suggestion
### Helpers for generalized indices

_Usage note: functions outside this section should manipulate generalized indices using only functions inside this section. This is to make it easier for developers to implement generalized indices with underlying representations other than bigints._
```

specs/light_client/merkle_proofs.md
```diff
@@ -64,43 +59,134 @@ y_data_root  len(y)
...
+    elif issubclass(typ, Elements):
+        return (typ.length * item_length(typ.elem_type) + 31) // 32
+    elif issubclass(typ, Container):
+        return len(typ.get_fields())
     else:
-        raise Exception("Unknown type / path")
+        raise Exception(f"Type not supported: {typ}")
+
+
+def get_item_position(typ: SSZType, index: Union[int, str]) -> Tuple[int, int, int]:
```
protolambda: Worried about the `[int, str]` a bit, since they are exclusively (from each other) used based on `typ`. So e.g. Container with int is invalid. It is likely something to revisit with partials/other uses, but is ok for now.

specs/light_client/merkle_proofs.md
```diff
@@ -116,72 +202,69 @@ x x . . . . x *
...
+    """
+    all_indices = set()
+    for index in indices:
+        all_indices = all_indices.union(set(get_branch_indices(index) + [index]))
+    
+    return sorted([
+        x for x in all_indices if not
+        (generalized_index_child(x, 0) in all_indices and generalized_index_child(x, 1) in all_indices) and not
+        (x in indices)
+    ])[::-1]
```
protolambda: Reversal of the list may be a bit opaque here for non-python readers. Instead we can just make the `sorted` descending order.
```suggestion
    ], reverse=True)
```

specs/light_client/merkle_proofs.md
```diff
@@ -116,72 +202,69 @@ x x . . . . x *
...
+```python
+def verify_merkle_multiproof(leaves: Sequence[Hash], proof: Sequence[Hash], indices: Sequence[GeneralizedIndex], root: Hash) -> bool:
+    assert len(leaves) == len(indices)
+    helper_indices = get_helper_indices(indices)
+    assert len(proof) == len(helper_indices)
+    objects = {
+        **{index:node for index, node in zip(indices, leaves)},
+        **{index:node for index, node in zip(helper_indices, proof)}
+    }
+    keys = sorted(objects.keys())[::-1]
```
protolambda:
```suggestion
    keys = sorted(objects.keys(), reverse=True)
```

specs/light_client/merkle_proofs.md
vbuterin: I think in all cases where we expect to use paths, the path would not be serialized, and it would be explicitly constructed directly as an input to `get_generalized_index` or similar methods (ie. never saved as a variable). So I think it's fine for that reason.

protolambda: Yes agree there, but we can maybe make it less error prone. We have others that are unfamiliar with the interface and need to use it as well.

Issue #1187, opened 2019-06-17
Distinguish true constants from network-configurable parameters


wemeetagain:
Currently, the constants listed in the beacon chain spec are a combination of constants that will never change, eg: `ZERO_HASH`, and "constants" that may change for testing or other purposes, eg: `SHARD_COUNT`.

It could be helpful to separate constants from configuration to help implementers determine which variables can be set at compile time vs runtime, how much flexibility to give in editing certain variables, etc.

djrtwo:
I'm on this

djrtwo:
addressed in #1191

PR #1191, opened 2019-06-17
split constants vs configuration

djrtwo:
address #1187


protolambda:
`BASE_REWARDS_PER_EPOCH` is another candidate for the constants section

PR #1192, opened 2019-06-17
update note about when proposer assignment valid

djrtwo:
There was a minor bug in the language here. The validator guide claimed that proposers are stable within the current epoch, but due to the ability for `slash_validator` to modify balances on a per block basis, this is not actually true.


protolambda:
Needs a review, `balances[i] <> validator(i).effective_balance`.

djrtwo:
Thanks @protolambda for the reminder. `effective_balance` only changes at the epoch boundaries. `increase_` and `decrease_balance` only adjust the `balances` array.

All good. false alarm!

PR #1193, opened 2019-06-17
Update spec for new SSZ with list max length

vbuterin:


specs/core/0_beacon-chain.md
```diff
@@ -964,8 +968,6 @@ def validate_indexed_attestation(state: BeaconState, indexed_attestation: Indexe
 
     # Verify no index has custody bit equal to 1 [to be removed in phase 1]
     assert len(bit_1_indices) == 0
-    # Verify max number of indices
-    assert len(bit_0_indices) + len(bit_1_indices) <= MAX_INDICES_PER_ATTESTATION
```
JustinDrake: Maybe worth keeping this line because `len(bit_0_indices) + len(bit_1_indices)` can be greater than `MAX_INDICES_PER_ATTESTATION`.

specs/core/0_beacon-chain.md
```diff
@@ -219,6 +220,9 @@ These configurations are updated for releases, but may be out of sync during `de
 | `RANDAO_MIXES_LENGTH` | `2**13` (= 8,192) | epochs | ~36 days |
 | `ACTIVE_INDEX_ROOTS_LENGTH` | `2**13` (= 8,192) | epochs | ~36 days |
 | `SLASHED_EXIT_LENGTH` | `2**13` (= 8,192) | epochs | ~36 days |
+| `VALIDATOR_REGISTRY_SIZE` | `2**40 (= 1,099,511,627,776)` | | |
+
+* Assuming the maximum 16 deposits per slot, the validator registry will last for at least 13,065 years (but likely much much longer)
```
JustinDrake: I'm tempted to remove this comment for a couple reasons:

1) In phase 2 it may be possible to create more than `MAX_DEPOSITS = 16` new validators per slot.
2) We don't usually put design rationales in the spec. (A separate design rationale doc needs to be created.)

djrtwo:
This depends no #1180, right?

protolambda:
merged this into #1180, as that doesn't build without proper type declarations (lengths being defined).

djrtwo:
closing in favor of #1180

PR #1197, opened 2019-06-19
Fix sig verification in transfer

terencechain:
missing `epoch` input for `get_domain` in `process_transfer`


djrtwo:
`get_domain` has an optional third parameter. If no `epoch` is passed in, it uses `current_epoch`.

Should be good as is. Let me know if I'm missing something.

Issue #768, opened 2019-03-13
Justifed state in LMD GHOST


djrtwo:
# Issue
`justified_head_state` which is used to get the active validator set for `lmd_ghost` is currently defined as "the resulting `BeaconState` object from processing the chain up to the `justified_head`".

This is correct in the "happy" case in which no slots were skipped between the `justified_head` and the epoch boundary that was justified, but in the case that slots were skipped between the head and epoch boundary slot (`% epoch_length == 0`), the justified head is actually the block from a prior slot but at the state at the epoch boundary slot. For example, if a block at slot 63 (epoch 0) is justified for epoch 1 (epoch boundary @ slot 64) in the case that there was no block for slot 64, then the state associated with this `(justified_block, justified_epoch)` tuple is actually the justified block from slot 63 transitioned through an additional slot to slot 64.

# Solution
Define `justified_head_state` as "the resulting `BeaconState` object from processing the chain up to the `justified_head` through the start slot of `justified_epoch`"

JustinDrake:
We probably also want to make the fork choice rule spec executable.

mratsim:
That would be great to produce test vectors.

The only existing reference code in Python is unfortunately outdated:
  - https://github.com/ethereum/research/blob/master/clock_disparity/lmd_node.py (uses notaries instead of validators and heights instead of slots)
  - https://github.com/ethereum/research/blob/master/ghost/ghost.py seems to be bitwise LMD Ghost

JustinDrake:
This was addressed in 1198

PR #1198, opened 2019-06-20
modify fork choice to utilize epochs properly

djrtwo:
This is a full adaptation of the issue discussed in #768 and is a crucial component of how state must be utilized in a slot/epoch based fork choice. note that this is a PR to the current fork choice PR #1185 

When justifying and finalizing a block, we are actually justifying/finalize a `(block, epoch)` tuple (`Checkpoint`). This represents the block _at_ the 0th slot of the epoch and we must use the associated state from here. In the normal case, this is at parity but in the case of skipped slots, the finalized block must be transitions up to the start of the epoch it was finalized in.


specs/core/0_fork-choice.md
```diff
@@ -63,17 +63,30 @@ class Target(object):
     root: Hash
 ```
 
+#### `RootSlot`
+
+```python
+@dataclass
+class RootSlot(object):
+    slot: Slot
+    root: Hash
```
hwwhww: Wow it actually needs to reorder 😂
```python
    root: Hash
    slot: Slot
```

djrtwo: Ha!

specs/core/0_fork-choice.md
```diff
@@ -63,17 +63,30 @@ class Target(object):
     root: Hash
 ```
 
+#### `RootSlot`
+
+```python
+@dataclass
+class RootSlot(object):
+    slot: Slot
```
JustinDrake: What is the reason for this field being a `Slot` (as opposed to an `Epoch`)? If we can go with `Epoch` then we could define a `Checkpoint` type to replace both `Target` and `RootSlot`.

specs/core/0_fork-choice.md
```diff
@@ -63,17 +63,30 @@ class Target(object):
...
 ```python
 @dataclass
 class Store(object):
     blocks: Dict[Hash, BeaconBlock] = field(default_factory=dict)
-    states: Dict[Hash, BeaconState] = field(default_factory=dict)
+    states: Dict[RootSlot, BeaconState] = field(default_factory=dict)
     time: int = 0
     latest_targets: Dict[ValidatorIndex, Target] = field(default_factory=dict)
     justified_root: Hash = ZERO_HASH
+    justified_epoch: Epoch = 0
```
JustinDrake: `justified_root` and `justified_epoch` can be combined into `justified_checkpoint`. Same for `finalized_root` and `finalized_epoch`.

specs/core/0_fork-choice.md
```diff
@@ -63,17 +63,30 @@ class Target(object):
...
     finalized_root: Hash = ZERO_HASH
+    finalized_epoch: Epoch = 0
```
JustinDrake: Prefer `GENESIS_EPOCH` over the constant `0`. Even better, make use of SSZ defaults.

specs/core/0_fork-choice.md
```diff
@@ -63,17 +63,30 @@ class Target(object):
...
+
+```python
+@dataclass
+class RootSlot(object):
+    slot: Slot
+    root: Hash
+```
+
+
+
```
JustinDrake: One empty line (as opposed to three) is sufficient :)

specs/core/0_fork-choice.md
```diff
@@ -137,30 +154,59 @@ def on_block(store: Store, block: BeaconBlock) -> None:
...
+            base_state = store.states[RootSlot(store.finalized_root, store.blocks[store.finalized_root].slot)]
+            store.states[finalized_root_slot] = process_slots(deepcopy(base_state), finalized_slot)
 ```
 
 #### `on_attestation`
 
 ```python
-    state = store.states[get_head(store)]
+    # cannot calculate the current shuffling if have not seen the target
```
JustinDrake: Capitalise comment first letter (for consistency).

specs/core/0_fork-choice.md
```diff
@@ -137,30 +154,59 @@ def on_block(store: Store, block: BeaconBlock) -> None:
...
+
+    # Update finalized block root and epoch, and store finalized state
+    if state.finalized_epoch > state.finalized_epoch:
         store.finalized_root = state.finalized_root
+        store.finalized_epoch = state.finalized_epoch
+        finalized_slot = get_epoch_start_slot(store.finalized_epoch)
+        finalized_root_slot = RootSlot(store.finalized_root, finalized_slot)
+        if finalized_root_slot not in store.states:
+            base_state = store.states[RootSlot(store.finalized_root, store.blocks[store.finalized_root].slot)]
+            store.states[finalized_root_slot] = process_slots(deepcopy(base_state), finalized_slot)
```
JustinDrake: `store.states` is not keyed by `Slot` (should be keyed by `Checkpoint`)

specs/core/0_fork-choice.md
djrtwo: We can do that, and agreed it's cleaner.

In some places, we need to know the slot because of the the state lookup and the restriction of the consideration of blocks only prior to the 0th slot in the epoch (rather than just prior to the current epoch). In these cases, we'll do the `get_epoch_start_slot` conversion.

djrtwo: Ah, the reason is that states are stored for various slots (some checkpoints, some just block states). Thinking about how to clean this up properly..

specs/core/0_fork-choice.md
```diff
@@ -141,35 +149,53 @@ def on_tick(store: Store, time: int) -> None:
...
-        store.justified_root = state.previous_justified_root
-    # Update finalized block root
-    if state.finalized_epoch > slot_to_epoch(store.blocks[store.finalized_root].slot):
-        store.finalized_root = state.finalized_root
+    # Add new state for this block to the store
+    store.block_states[signing_root(block)] = state
+
+    # Update justified checkpoint
+    if state.current_justified_epoch > store.justified_checkpoint.epoch:
+        store.justified_checkpoint = Checkpoint(state.current_justified_epoch, state.current_justified_root)
```
JustinDrake: It's very tempting to define the `Checkpoint` object in `0_beacon-chain.md` and rework the `state` so that we can simply do `store.justified_checkpoint = state.current_justified_checkpoint`. It would also clean up `state`.

djrtwo: yeah, I was thinking the same thing...

djrtwo: How about I do that in separate PR?

JustinDrake: Separate PR works 👍

specs/core/0_fork-choice.md
```diff
@@ -141,35 +149,53 @@ def on_tick(store: Store, time: int) -> None:
...
+    pre_state = store.block_states[parent_block.root].copy()
     # Blocks cannot be in the future. If they are, their consideration must be delayed until the are in the past.
     assert store.time >= pre_state.genesis_time + block.slot * SECONDS_PER_SLOT
     # Add new block to the store
     store.blocks[signing_root(block)] = block
     # Check block is a descendant of the finalized block
-    assert get_ancestor(store, signing_root(block), store.blocks[store.finalized_root].slot) == store.finalized_root
+    assert get_ancestor(store, signing_root(block), store.blocks[get_epoch_start_slot(store.finalized_checkpoint)].slot) == store.finalized_checkpoint.root
+    # Check that block is later than the finalized epoch slot
+    assert block.slot > get_epoch_start_slot(store.finalized_checkpoint.epoch)
```
JustinDrake: Would it make sense to pass a `Checkpoint` to `get_ancestor` (thinking out loud—going to sleep)?

specs/core/0_fork-choice.md
```diff
@@ -141,35 +149,53 @@ def on_tick(store: Store, time: int) -> None:
...
-    state = store.states[get_head(store)]
+    target_checkpoint = Checkpoint(attestation.data.target_epoch, attestation.data, target_root)
+
+    # Cannot calculate the current shuffling if have not seen the target
+    assert target_checkpoint.root in store.blocks
+
+    # Store target checkpoint state if not yet seen
+    if target_checkpoint not in store.checkpoint_states:
+        base_state = store.block_states[target_checkpoint.root].copy()
+        store.checkpoint_states[target_checkpoint] = process_slots(base_state, get_epoch_start_slot(target_checkpoint.epoch))
```
JustinDrake: `target_checkpoint.epoch` can be chosen adversarially. We should maybe validate it against `block.slot` and `store.time`.

specs/core/0_fork-choice.md
```diff
@@ -70,11 +70,12 @@ class Target(object):
 @dataclass
 class Store(object):
     blocks: Dict[Hash, BeaconBlock] = field(default_factory=dict)
-    states: Dict[Hash, BeaconState] = field(default_factory=dict)
-    time: int = 0
-    latest_targets: Dict[ValidatorIndex, Target] = field(default_factory=dict)
-    justified_root: Hash = ZERO_HASH
-    finalized_root: Hash = ZERO_HASH
+    block_states: Dict[Hash, BeaconState] = field(default_factory=dict)
+    checkpoint_states: Dict[Checkpoint, BeaconState] = field(default_factory=dict)
```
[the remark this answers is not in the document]
djrtwo: One is looked up by `(block)` (or the state at the slot the block was created). The other is looked up by checkpoints `(block, epoch)`.  Coercing one into the other doesn't really work nicely so we can't just naively squash, but open to ideas.

specs/core/0_fork-choice.md
djrtwo: In this case, we are looking for a block at a precise (rather than the epoch granularity of a checkpoint)

specs/core/0_fork-choice.md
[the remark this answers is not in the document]
JustinDrake: I see, we can't use `checkpoint_states` everywhere because of `assert block.parent_root in store.block_states`. Would it make sense replace `parent_root` with `parent_checkpoint`?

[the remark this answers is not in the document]
djrtwo: No, because when we get a new block, we store the state at the block/slot. Checkpoints are state at a block/epoch-boundary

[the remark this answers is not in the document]
JustinDrake: What about adding `parent_slot`?

specs/core/0_fork-choice.md
```diff
@@ -55,11 +55,11 @@ The head block root associated with a `store` is defined as `get_head(store)`. A
 
 ### Helpers
 
-#### `Target`
+#### `Checkpoint`
 
 ```python
-@dataclass
-class Target(object):
+@dataclass(eq=True, frozen=True)
```
djrtwo: This is temporary to make `Checkpoint` usable as a lookup key in a dictionary.
In #1210, we elevate `Checkpoint` to an SSZ type in beacon chain spec and this is removed

specs/core/0_fork-choice.md
```diff
@@ -118,9 +133,13 @@ def get_latest_attesting_balance(store: Store, root: Hash) -> Gwei:
...
 def get_head(store: Store) -> Hash:
     # Execute the LMD-GHOST fork choice
-    head = store.justified_root
+    head = store.justified_checkpoint.root
+    justified_slot = get_epoch_start_slot(store.justified_checkpoint.epoch)
     while True:
-        children = [root for root in store.blocks.keys() if store.blocks[root].parent_root == head]
+        children = [
+            root for root in store.blocks.keys()
+            if store.blocks[root].parent_root == head and store.blocks[root].slot > justified_slot
```
protolambda: Just checking `parent_root` should be enough. But if we are doing both, maybe swap the slot check to be first (much faster)?

djrtwo: It isn't actually enough. There is a case where the block is in the block tree since latest justified block but is from an epoch prior to latest justified epoch.

Example:
* Block A at slot 60
* Block B (parent A) at slot 61
* Block A is justified for epoch 1 (slot 64)
* The block tree under consideration is now descendants of Block A but exclusively after the epoch boundary slot of epoch 1.
* Thus Block B should not be in consideration.

That said, we can flip the two bool checks for speed

PR #1202, opened 2019-06-21
Add `test_genesis.py` and fix `is_genesis_trigger`

### LINKED ISSUE #1199: is_genesis_trigger fails to verify deposit merkle branch
## Issue
Deposits are processed against empty state with undefined `eth1_data` while `genesis_eth1_data` should be in place.

hwwhww:
1. Fix #1199
    - `is_genesis_trigger`
    - Fix the incorrect variable name
    - Use `state.eth1_data` and `state.eth1_deposit_index` as the temporary container for calling `process_deposit()`
    - *Incrementally* generate the merkle proof against the *each* `hash_tree_root(List[DepositData, 2**32](*deposit_data_list))` and add `deposit_data` to `deposit_data_list` during creating the given `deposits: Seqence[Deposit]` input. 
2. Fix `get_genesis_state` as well
2. Add `test_libs/pyspec/eth2spec/test/genesis/` tests 
3. Update deposit contract (per #1229)


specs/core/0_beacon-chain.md
```diff
@@ -1134,32 +1135,32 @@ Before genesis has been triggered and whenever the deposit contract emits a `Dep
...
-    * `genesis_eth1_data.block_hash` is the Ethereum 1.0 block hash that emitted the log for the last deposit in `deposits`
     * `genesis_eth1_data.deposit_root` is the deposit root for the last deposit in `deposits`
     * `genesis_eth1_data.deposit_count = len(genesis_deposits)`
+    * `genesis_eth1_data.block_hash` is the Ethereum 1.0 block hash that emitted the log for the last deposit in `deposits`
 
 *Note*: The function `is_genesis_trigger` has yet to be agreed by the community, and can be updated as necessary. We define the following testing placeholder:
 
 ```python
-def is_genesis_trigger(deposits: List[Deposit], timestamp: uint64) -> bool:
+def is_genesis_trigger(deposits: List[Deposit], timestamp: uint64, deposit_root: Hash) -> bool:
```
JustinDrake: There's no need to add `deposit_root`. It's redundant information (can be derived from `deposits`).

specs/core/0_beacon-chain.md
```diff
@@ -1134,32 +1135,32 @@ Before genesis has been triggered and whenever the deposit contract emits a `Dep
...
-        process_deposit(state, deposit)
+    for index, deposit in enumerate(deposits):
+        process_deposit(state, deposit, deposit_index=index, deposit_root=deposit_root)
 
     # Count active validators at genesis
     active_validator_count = 0
-    for validator in state.validator_registry:
-        if validator.effective_balance == MAX_EFFECTIVE_BALANCE:
+    for validator in state.validators:
+        if validator.effective_balance >= MAX_EFFECTIVE_BALANCE:
```
JustinDrake: `validator.effective_balance > MAX_EFFECTIVE_BALANCE` is not possible. Seems semantically cleaner to write `validator.effective_balance == MAX_EFFECTIVE_BALANCE`

specs/core/0_beacon-chain.md
```diff
@@ -1726,17 +1727,24 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
 ##### Deposits
 
 ```python
-def process_deposit(state: BeaconState, deposit: Deposit) -> None:
+def process_deposit(state: BeaconState,
+                    deposit: Deposit,
+                    deposit_index: Optional[uint64]=None,
+                    deposit_root: Optional[Hash]=None) -> None:
```
JustinDrake: No need to add `deposit_index` and `deposit_root` :) It suffices to properly initialise `state.eth1_data.deposit_root` corresponding to the empty sparse Merkle tree after `state = BeaconState()` in `is_genesis_trigger`.

specs/core/0_beacon-chain.md
hwwhww: Oh but there's also:
```python
# Process genesis activations
    for validator in state.validators:
        if validator.effective_balance >= MAX_EFFECTIVE_BALANCE:
            validator.activation_eligibility_epoch = GENESIS_EPOCH
            validator.activation_epoch = GENESIS_EPOCH
```
in `get_genesis_beacon_state`. Should they both be `==`?

JustinDrake: Aha, I see. They are equivalent and I suggest `validator.effective_balance == MAX_EFFECTIVE_BALANCE` in both cases :)

specs/core/0_beacon-chain.md
JustinDrake: A mathematically clean definition would be `state.eth1_data.deposit_root = hash_tree_root(Vector[Hash, 2** DEPOSIT_CONTRACT_TREE_DEPTH])` but that would stall any implementation that doesn't implement the sparse Merkle tree optimisations.

hwwhww: 1. Right! But that means we have to initialize and update `Eth1Data` for *each* genesis deposit in both `is_genesis_trigger` and `get_genesis_beacon_state`... but I agreed that making `process_deposit` clean is a win for the long term!
2. Also note that there will be so many different ways to optimize `is_genesis_trigger` in client side. 😄

specs/core/0_beacon-chain.md
hwwhww: @JustinDrake to confirm, that means we have add [prepare_genesis_deposits](https://github.com/ethereum/eth2.0-specs/blob/6aef6c5634acc0c378c1f03b2cefd32ba3501954/test_libs/pyspec/eth2spec/test/helpers/deposits.py#L63-L90) logic to `is_genesis_trigger`, including calling [`calc_merkle_tree_from_leaves`, `get_merkle_root`, and `get_merkle_proof` (in `merkle_minimal.py`)](https://github.com/ethereum/eth2.0-specs/blob/e8b4c4c57f1f07a8eae6c88249bad6c19335017f/test_libs/pyspec/eth2spec/utils/merkle_minimal.py#L12-L24).

I think it's good to move the `merkle_minimal.py` to the spec, but the `zerohashes` trick might not fit the spec format well...

JustinDrake: I'm happy to rework the PR :) I think we can just roll back to what it was, and add the one line that initialises `deposit_root` to the correct thing (as opposed to all zeros).

JustinDrake:
@hwwhww What do you think of https://github.com/ethereum/eth2.0-specs/pull/1202/commits/a3f8f50beed9f66801ce266b192b234c2dd11299?

hwwhww:
@JustinDrake 

In practice, a client, they would probably:
1. Build `deposit_data_list: List[DepositData]` from deposit contract log.
2. Generate `proof` against the latest `deposit_root` (can be gotten from the contract, or just merkleize the `deposit_data_list`).
3. Build the `deposits: List[Deposit]` with these `proof`.
4. Check the if it's triggered the genesis starting point.

So it's unlikely they would need to regenerate `deposit_root` after they have created the `Deposit`s, in practice. That's why I'd say we don't need to regenerate the `deposit_root` again.

Another way to describe the logic is changing the function to `def is_genesis_trigger(deposit_data_list: List[DepositData], timestamp: uint64) -> bool`, and generate the proof in this function. I hope to balance between (i) making the spec describe the protocol in details and (ii) making it too far from practice. 😄

What do you think about it? 😸

JustinDrake:
Right. In practice I expect implementers to have a cache (e.g. a `Store` mapping deposits to beacon states) and an incremental version of `is_genesis_trigger`. This would avoid recomputing things from genesis, and handle Eth1 reorgs. I'd favour spec readability over efficiency, but don't feel too strongly either way :)

protolambda:
- fixed the merge from Justin, it was broken:
  - introduced back an invalid deposit test, we don't have deposit.index anymore
  - syntax problem, decorator was over empty line
  - confusing big merge, few merge conflicts, but know not to pick the old(!!!) diff parts, when they changed intentionally...
- Moved genesis tests out of sanity. The idea is that the shape of each test is the same within one file.
- Annotated the yielded data, so that it actually gets into a dict of key-value for a future test generator
- Fixed the genesis-trigger-true tests partially. Date was the same as the false case...
- Introduced local `SECONDS_PER_DAY` constant. You can make it a real global, but if it's for an "example" part of the code, we shouldn't include it in the main constants imho.
- Fixed problems with python syntax, `map` was used wrong, and generator (`... for ... in ...`) results were used as if it were lists (they are `types.GeneratorType`)
- The Deposit-root for the truee-trigger is still wrong somehow, but something for you to figure out, may be the expected root that is wrong, or some other funky reason more familiar to the PR authors

test_libs/pyspec/eth2spec/test/helpers/deposits.py
```diff
@@ -4,25 +4,31 @@
...
-def build_deposit_data(spec, state, pubkey, privkey, amount, withdrawal_credentials, signed=False):
+def build_deposit_data(spec, pubkey, privkey, amount, withdrawal_credentials, state=None, signed=False):
     deposit_data = spec.DepositData(
         pubkey=pubkey,
         withdrawal_credentials=withdrawal_credentials,
         amount=amount,
     )
     if signed:
-        sign_deposit_data(spec, state, deposit_data, privkey)
+        sign_deposit_data(spec, deposit_data, privkey, state)
```
protolambda: Looks unnecessary, just hold the `spec, state, args...` order (as introduced by Carl in phase-1 PR) everywhere for consistency please.

test_libs/pyspec/eth2spec/test/phase_0/block_processing/test_process_deposit.py
```diff
@@ -167,6 +167,6 @@ def test_bad_merkle_proof(spec, state):
     # mess up merkle branch
     deposit.proof[5] = spec.ZERO_HASH
 
-    sign_deposit_data(spec, state, deposit.data, privkeys[validator_index])
+    sign_deposit_data(spec, deposit.data, privkeys[validator_index], state=state)
```
protolambda: ditto

test_libs/pyspec/eth2spec/test/helpers/deposits.py
hwwhww: because I changed `state` to optional. (`def sign_deposit_data(spec, deposit_data, privkey, state=None)`) It should be `sign_deposit_data(spec, deposit_data, privkey, state=state)`.

specs/core/0_beacon-chain.md
```diff
@@ -1117,26 +1119,35 @@ def slash_validator(state: BeaconState,
...
+def is_genesis_trigger(deposits: List[Deposit, 2**DEPOSIT_CONTRACT_TREE_DEPTH], time: uint64) -> bool:
+    SECONDS_PER_DAY = 86400
+    # Do not deploy too early
+    if time - time % SECONDS_PER_DAY + 2 * SECONDS_PER_DAY < MIN_GENESIS_TIME:
+        return False
+
+    # Initialize deposit root
     state = BeaconState()
+    state.eth1_data.deposit_root = hash_tree_root(
+        Vector[DepositData, len(deposits)](list(map(lambda deposit: deposit.data, deposits)))
```
hwwhww: Ahhhh, in the telegram chat discussion, I said that "I just understand that the `Deposit.proof` you mean is targeting on the `deposit_root` after the given DepositEvent is triggered… 🙈 (I was thinking to regenerate the proof with the latest deposit_root when triggering the genesis as what the dev branch is, but that’s much more overhead)" but it turns out you *did* mean using the same `deposit_root` for all the `proof`s! 😅 

We expect the implementers would optimize it with their own approach. However, this function is part of the consensus tests and the implementers can use this function to testify the optimized version.

Generating `Deposit.proof` with the same `deposit_root` disallows the implementation to build the tree `DepositData` by `DepositData`. So I now think using `deposit_root` per `DepositEvent` might be more realistic and providing the same testing input as the clients would do.

/cc @terencechain since I think Prysm already has the deposit contract handler? Any thought? :)

JustinDrake: I don't mind either way. Changed it to incrementally compute `deposit_root`.

djrtwo:
Where do we stand on this PR? One test fails. Are any changes and/or this test being worked on still?

JustinDrake:
@hwwhww is working on it. She's been liaising with me and I'm expecting a push soon.

protolambda:
Reviewed and debugged it.
Findings:

1) proofs are constructed for the whole tree, not up to the given deposit.
2) but eth1 deposit root changes per deposit
3) the full vector hash-tree-root doesn't work, since input is not padded, hence invalid.

1 is solved by adding a special `create_merkle_proof_up_to` that creates the proof based on part of the tree.
2 is ok, but slow, and coincides with the solution for 3. I recommend changing it to:
```python
    leaves = list(map(lambda deposit: hash_tree_root(deposit.data), deposits))
    for i, deposit in enumerate(deposits):
        state.eth1_data.deposit_root = merkleize_chunks(leaves[:i], pad_to=2**DEPOSIT_CONTRACT_TREE_DEPTH)
        process_deposit(state, deposit)
```

Alternatively we do a full root, don't change it every deposit, and construct proofs for each deposit by the full tree. But this is not very accumulator like. But valid enough for genesis.

JustinDrake:
> In the end, `get_merkle_root` is back

I don't understand. Isn't this unnecessary complexity, and go against what we discussed?

hwwhww:
Thanks @protolambda!
@JustinDrake as @protolambda mentioned, both SSZ `List` and `Vector` are not the normal merkleization that we can use directly. The `Vector` will initialize the sparse tree with `DepostData()` instead of `ZERO_HASH`. And to use `List`, we might need an extra "flag" in `hash_tree_root` to escape from `mix_in_length`, which I don't think it's idea too...


For the merklization, I think it's reasonable to migrate most of these `merkle_minimal.py` logic with https://github.com/ethereum/eth2.0-specs/blob/dev/specs/light_client/merkle_proofs.md someday when we make light client protocol executable.

hwwhww:
Sorry for the delay of this PR!!! I will have time to clean up it at the 29th afternoon (Asia).

TODO thoughts:
1. `is_genesis_trigger` and `get_genesis_state` are really similar. I'm thinking either (i) using one function to check and get the genesis state altogether or (ii) using the result from `is_genesis_trigger()` as the input of  `get_genesis_state()`.
2. Clean up `merkle_minimal.py` code

test_libs/pyspec/eth2spec/test/phase_0/block_processing/test_process_deposit.py
hwwhww: Ditto, optional parameter here.

specs/core/0_beacon-chain.md
```diff
@@ -1100,18 +1101,24 @@ When `is_genesis_trigger(deposits, timestamp) is True` for the first time, let:
 
 * `genesis_deposits = deposits`
```
djrtwo: > Before genesis has been triggered and whenever the deposit contract emits a `Deposit` log, call the function `is_genesis_trigger(deposits: Sequence[Deposit], timestamp: uint64) -> bool` where:

Update this to say to run on each `block` rather than each log

specs/core/0_beacon-chain.md
```diff
@@ -1121,29 +1128,34 @@ def is_genesis_trigger(deposits: Sequence[Deposit], timestamp: uint64) -> bool:
             active_validator_count += 1
 
     # Check effective balance to trigger genesis
```
djrtwo: Comment is misleading

specs/core/0_beacon-chain.md
```diff
@@ -1089,63 +1090,41 @@ def slash_validator(state: BeaconState,
...
         latest_block_header=BeaconBlockHeader(body_root=hash_tree_root(BeaconBlockBody())),
     )
+    assert state.genesis_time >= MIN_GENESIS_TIME
 
-    # Process genesis deposits
-    for deposit in deposits:
+    # Process deposits
+    leaves = list(map(lambda deposit: deposit.data, deposits))
+    for index, deposit in enumerate(deposits):
+        state.eth1_data.deposit_root = hash_tree_root(
```
djrtwo: Why are the deposits passed in with proofs against each subsequent root rather than the end root?
Having all proofs against the most recent deposit is more in line with how we process deposits in blocks.

```python
leaves = ...
state.eth1_data.deposit_root = hash_tree_root(leaves)
for index, deposit in enumerate(deposits):
    process_deposit(state, deposit)
```

djrtwo: Ah, is it to eliminate having to recompute proofs each block?
Instead you have to do this `hash_tree_root` `deposit_count` times...

specs/core/0_beacon-chain.md
```diff
@@ -1089,63 +1090,41 @@ def slash_validator(state: BeaconState,
...
+            List[DepositData, 2**DEPOSIT_CONTRACT_TREE_DEPTH](*leaves[:index + 1])
+        )
         process_deposit(state, deposit)
 
-    # Process genesis activations
-    for validator in state.validators:
-        if validator.effective_balance >= MAX_EFFECTIVE_BALANCE:
+    # Process activations
+    for index, validator in enumerate(state.validators):
+        if state.balances[index] >= MAX_EFFECTIVE_BALANCE:
```
djrtwo: Why `balances` and `>=` here rather than the `effective_balance` and `==`?

specs/core/0_beacon-chain.md
```diff
@@ -1682,7 +1661,7 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
     assert verify_merkle_branch(
         leaf=hash_tree_root(deposit.data),
         proof=deposit.proof,
-        depth=DEPOSIT_CONTRACT_TREE_DEPTH,
+        depth=DEPOSIT_CONTRACT_TREE_DEPTH + 1,
```
djrtwo: `+1` due to the length mixin? Does that deserve a comment?

specs/core/0_beacon-chain.md
JustinDrake: I think there was a bug. The problem is that `effective_balance` is only set on the first deposit. I intend to do a review of the `effective_balance` logic tomorrow.

specs/core/0_beacon-chain.md
hwwhww: Yeah, it was easier to incrementally update the normal Merkle tree... but I think for the client side, they can still optimize it with global SSZ Merkle tree structure?

specs/core/0_beacon-chain.md
hwwhww: I have the same question here! Ping @JustinDrake

JustinDrake: Answered

specs/core/0_beacon-chain.md
JustinDrake: Yes, length mixin

specs/core/0_beacon-chain.md
protolambda: I think 8d2cbc9 provides a warranted description on what one can (and should) do instead of quadratic hash-tree-root. Open for text improvements, but otherwise want to ship for spec freeze.

specs/core/0_beacon-chain.md
protolambda: It's genesis, so it's fine for activation. Going through the usages of `effective_balance` sounds like a good last action point. I'll remind you. Looking to ship this PR now.

specs/core/0_beacon-chain.md
protolambda: Added a small comment for it in 8d2cbc9

protolambda:
- Added `signed=True` args to make it work with BLS
- Added a few remarks to avoid confusion in genesis/deposit verification

Ready to merge imho. We can add a few additional tests after spec-freeze. Possibly even improve the efficiency of genesis construction. As clients are (very) likely not implementing it like this anyway, and state output will be the same. This description and pseudocode will at least set a baseline, and set the interface to be compatible with SSZ partials in the future.

PR #1208, opened 2019-06-22
Add note on default zero-values, and be explicit with state-root

protolambda:
Small change to clear up confusion, and be explicit in the most confusing part. The difference in this specific case compared to other use of defaulting is that we are overwriting the `state_root` field, without naming the field.


protolambda:
@JustinDrake Thanks for improving the wording, I will check things next time. :+1:

PR #1209, opened 2019-06-22
Resolves issue in early_secret_reveal() testing

CarlBeek:
Because no BLS decorators were added to the early derived secret reveal tests, the function was not fully being tested. Specifically, the mask and the message being signed by the masker did not line up and moreover, the masker's signature was not even being combined with the revealer's.


test_libs/pyspec/eth2spec/test/helpers/custody.py
```diff
@@ -19,20 +20,24 @@ def get_valid_early_derived_secret_reveal(spec, state, epoch=None):
             message_epoch=epoch,
         ),
     )
-    mask = bls_sign(
-        message_hash=spec.hash_tree_root(epoch),
+    # Generate the mask (any random 32 bytes will do)
+    mask = reveal[:32]
```
dankrad:
```suggestion
    mask = hash(reveal)
```

For the test it does not matter, but in reality it would be a bad idea to use `reveal[:32]` (it may be possible to brute force the full reveal from this, which defeats the purpose of masking). So let's put something here that people can use safely in practice, in case someone happens to use this code as a reference.

dankrad:
Good catch :)

test_libs/pyspec/eth2spec/test/helpers/custody.py
CarlBeek: I had this debate with myself and opted to just use the first 32 bytes for the sake of speed. That said, considering how marginal the cost is, I'm happy to switch over to `hash(x)` for signaling's sake.

cryptoneobancoin:
good one

PR #1210, opened 2019-06-22
Checkpoints in beacon spec

djrtwo:
Ready for review.

Is based on #1198. Let's get that one merge first and then will switch this to merge to `dev`


specs/core/0_beacon-chain.md
```diff
@@ -1705,19 +1698,19 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
...
     else:
-        ffg_data = (state.previous_justified_epoch, state.previous_justified_root, get_previous_epoch(state))
+        ffg_source = state.previous_justified_checkpoint
         parent_crosslink = state.previous_crosslinks[data.crosslink.shard]
         state.previous_epoch_attestations.append(pending_attestation)
 
-    # Check FFG data, crosslink data, and signature
-    assert ffg_data == (data.source_epoch, data.source_root, data.target_epoch)
+    # Check FFG source, crosslink data, and signature
+    assert ffg_source == data.source
```
JustinDrake: Do we also need to check that `data.target_epoch` matches?

JustinDrake:
I very much like this :)

specs/core/0_beacon-chain.md
djrtwo: Do to the initial assert that `target_epoch in prev or cur` and the conditional statement above that filters the proper source by the attestation's target_epoch, I believe this assertion actually covers it

djrtwo:
nice cleanup `AttestationData.target_checkpoint` -> `AttestationData.target`

djrtwo:
Thoughts on changing `state.current_justified_checkpoint` to `state.current_justified`?

changes accessors to `state.current_justified.epoch` and `state.current_justified.root` which is nice, but `state.current_justified` loses a bit of semantic meaning on it's own.

JustinDrake:
> Thoughts on changing `state.current_justified_checkpoint` to `state.current_justified`?

Would be nice to shorten :) Brainstorming:

* `state.currently_justified` (reads a bit better as a standalone; also distinguishes semantically from `previous_` in `previous_epoch_attestations`)
* `state.current_justification` (reads a bit better as a standalone)

PR #1039, opened 2019-05-03
Doc standardization

JSON:
Some standardization edits, given diff writers between docs.

Primary updates:

- Abbreviations Eth 2.0/Eth2.0/eth2.0/ETH 2.0 standardized to “Eth 2.0” (ETH when referring to ether is left as is; Ethereum→Eth, ether→ETH) (Ethereum 2.0 when spelled out, left as is; any code like `Eth2Genesis` or the spec name eth2.0-specs, ofc also left as is)
- If a page has a ToC, parallel formatting as the specs/core docs’ ToCs 
- ie/ie./i.e./i.e., + eg/eg./e.g./e.g., standardized to i.e.,/e.g., (chicago manual of style/ap stylebook use both periods + a comma) (didn’t touch the ones in code snippets)
- Replacing “:” with “-” in 1_custody-game.md’s Terminology section to match 0_beacon-chain.md’s
- Capitalization when referring to a specific phase (was about half and half beforehand)

Misc.:

- Adding the title to the “Merkle proof formats” doc; was missing
- Adding the RFC 2119 link when referenced
- Adding the C++ SSZ maintainer
- Updating links to the now deprecated ethereum/eth2.0-tests repo to the new ethereum/eth2.0-spec-tests repo
- Grammar

PR #1212, opened 2019-06-24
phase 0 doc standardization b4 spec freeze

JSON:
Following up on #1039 for documents pertaining to Phase 0, in particular.

Summary:

Dead links, stay-on-the-same-spec links, capitalization for the first word after "Note:" & when referring to a specific phase, abbreviations spelled out when the full version is used more often in the spec and the abbreviated version appears rarely, grammatical commas, standardized syntax for describing where 'Helper functions' on 0_beacon-chain.md are defined, British spelling 'utilised' --> "utilized", etc.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,18 +1,18 @@
 # Building pyspecs from specs.md
 
-The benefit of the particular spec design is that the given markdown files can be converted to a `spec.py` file for the purposes of testing and linting. The result of this is that bugs are discovered and patched more quickly.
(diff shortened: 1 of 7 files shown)


JustinDrake:
Thanks for this :)

(As a side note, the freeze is only for substantive changes. Cosmetics things such as these are definitely welcome during and after the freeze.)

djrtwo:
* I count var names as substantive 😬 (unless _realllllly_ good reason)

JSON:
> * I count var names as substantive 😬 (unless _realllllly_ good reason)

@djrtwo if i'm understanding correctly, i don't think i touched any var names

djrtwo:
@JSON You didn't! That was a comment for @JustinDrake ...

PR #1217, opened 2019-06-26
Slashing penalty calculation change

vbuterin:
If the exit queue is very long, then a validator may take many months to exit. With the code as currently written, however, self-slashing is a potentially lucrative route to get one's money out faster, because one can exit in 36 days.

This PR changes it so that slashing can only extend your withdrawal time, not contract it. Also, instead of the slashed balances used to calculate one's slashing penalty being those in `[withdrawal - 54 days ... withdrawal - 18 days]`, we now run the penalization algorithm once every 36 days that a validator is slashed but not withdrawn, so that it covers the 36-day period where the validator was actually slashed.  It also moves the minimum slashing penalty to the `slash_validator` function so that it is only applied once.


djrtwo:
I'm fixing tests as I review

JustinDrake:
I've revamped and simplified the slashing logic:

1) Avoid the ever-growing sum in `slashed_balances`. Instead, simply store the per-epoch sum. This is less confusing, allows for simpler logic (simply do `sum(slashed_balances)`), and removes the long-term risk of `slashed_balances` overflow.
2) Rename `slashed_balances` to `slashings`. Shorter, consistent with `BeaconState` group label, just as descriptive.
3) `whistleblowing_reward => whistleblower_reward` (for consistency with `proposer_reward`)

specs/core/0_beacon-chain.md
```diff
@@ -1091,26 +1091,27 @@ def initiate_validator_exit(state: BeaconState, index: ValidatorIndex) -> None:
...
-    state.validators[slashed_index].withdrawable_epoch = Epoch(current_epoch + EPOCHS_PER_SLASHED_BALANCES_VECTOR)
-    slashed_balance = state.validators[slashed_index].effective_balance
-    state.slashed_balances[current_epoch % EPOCHS_PER_SLASHED_BALANCES_VECTOR] += slashed_balance
-
+    epoch = get_current_epoch(state)
+    initiate_validator_exit(state, slashee_index)
+    slashee = state.validators[slashee_index]
+    slashee.slashed = True
+    slashee.withdrawable_epoch = max(slashee.withdrawable_epoch, Epoch(epoch + EPOCHS_PER_SLASHINGS_VECTOR))
+    decrease_balance(state, slashee_index, slashee.effective_balance // MIN_SLASHING_PENALTY_QUOTIENT)
```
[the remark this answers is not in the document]
JustinDrake: Good point about the `effective_balance` confusion. I'm putting `decrease_balance` *after* `state.slashings[epoch % EPOCHS_PER_SLASHINGS_VECTOR] += slashee.effective_balance` and hoping it's an acceptable compromise :)

specs/core/0_beacon-chain.md
```diff
@@ -1498,18 +1499,9 @@ def process_registry_updates(state: BeaconState) -> None:
...
-    total_at_end = state.slashed_balances[epoch % EPOCHS_PER_SLASHED_BALANCES_VECTOR]
-    total_penalties = total_at_end - total_at_start
-
     for index, validator in enumerate(state.validators):
-        if validator.slashed and epoch + EPOCHS_PER_SLASHED_BALANCES_VECTOR // 2 == validator.withdrawable_epoch:
-            penalty = max(
-                validator.effective_balance * min(total_penalties * 3, total_balance) // total_balance,
-                validator.effective_balance // MIN_SLASHING_PENALTY_QUOTIENT
-            )
+        if validator.slashed and epoch + EPOCHS_PER_SLASHINGS_VECTOR // 2 == validator.withdrawable_epoch:
```
JustinDrake: We need to make sure a slashed validator cannot have it's `withdrawable_epoch` pushed back (e.g. in phase 1 custody challenges). Otherwise it may be possible for `process_slashings` to apply the slashing penalty to the same validator multiple times.

djrtwo: The calling functions ensure this

JustinDrake: Right. We just need to remember this invariant for future reference 👍

Issue #918, opened 2019-04-14
Group helper functions by type


JustinDrake:
Suggest grouping helpers in four categories:

1) Math and crypto
2) Non-boolean pure functions (`get_...`)
3) Boolean pure functions (`is_...`)
4) State-modifying functions

I have done the following renamings:

* get_epoch_start_slot => get_start_slot
* get_beacon_proposer_index => get_proposer_index
* bls_domain => get_bls_domain
* generate_seed => get_seed
* compute_committee => get_committee
* convert_to_indexed => get_indexed_attestation
* slot_to_epoch => get_epoch_from_slot
* verify_bitfield => is_valid_bitfield
* validate_indexed_attestation => is_valid_indexed_attestation
* get_shuffled_index => shuffle_index

### Math

int_to_bytes
bytes_to_int
integer_squareroot
xor

### Crypto

hash
hash_tree_root
signing_root
verify_merkle_branch
bls_verify
bls_verify_multiple
bls_aggregate_pubkeys

### Non-boolean pure functions (`get_...`)

get_previous_epoch
get_current_epoch
get_start_slot
get_start_shard
get_epoch_from_slot
get_active_validator_indices
get_epoch_committee_count
get_shard_delta
get_crosslink_committee
get_block_root
get_randao_mix
get_seed
get_proposer_index
get_total_balance
get_churn_limit
get_delayed_activation_exit_epoch
get_crosslink_committee_for_attestation
get_attestation_participants
get_domain

### Boolean pure functions (`is_...`)

is_active_validator
is_slashable_validator
is_double_vote
is_surround_vote
is_valid_bitfield
is_valid_indexed_attestation

### State-modifying functions

activate_validator
initiate_validator_exit
slash_validator
set_balance
increase_balance
decrease_balance

djrtwo:
I think this is reasonable.

ralexstokes:
my two dai:

1. a nicer name for `boolean pure functions` could be `predicates`
2. consider the cost of losing semantic intent while pursuing uniformity, cf. `compute_committee` => `get_committee`, `generate_seed` => `get_seed`

i could easily be out-weighed here but ime `get_foo` has the connotation that you are just "chasing pointers" to get to some other data, with less, rather than more, computation along the way

JustinDrake:
Mostly addressed in #1237 🎉

PR #1219, opened 2019-06-26
Compact committee roots

JustinDrake:
Replacement of `active_index_roots` with `compact_committee_roots` to allow for significantly more efficient light clients (see item 17 in #1054). Key changes:

* Validator information (including indices) is pre-shuffled and grouped by crosslink committee.
* Validator information relevant to light clients (`validator.effective_balance` and `validator.slashed`) is kept "close" to the validator index to avoid accessing `state.validators`.
* Validator information (`index`, `effective_balance`, `slashed`) is compactified into a `uint64` which is friendly to SSZ packing.
* Shuffled pubkeys are available but kept separate from compact validator information. Light clients that have pubkeys cached do not need to re-download and re-authenticate them.

The plan is to also have persistent committee roots in phase 1.


specs/core/0_beacon-chain.md
```diff
@@ -742,17 +753,23 @@ def get_randao_mix(state: BeaconState,
...
     """
-    Return the index root at a recent ``epoch``.
-    ``epoch`` expected to be between
-    (current_epoch - EPOCHS_PER_HISTORICAL_VECTOR + ACTIVATION_EXIT_DELAY, current_epoch + ACTIVATION_EXIT_DELAY].
+    Return the compact committee root for the current epoch.
     """
-    return state.active_index_roots[epoch % EPOCHS_PER_HISTORICAL_VECTOR]
+    committee_data = CompactCommitteees().data
+    for committee_number in range(get_epoch_committee_count(state, epoch)):
+        shard = (get_epoch_start_shard(state, epoch) + committee_number) % SHARD_COUNT
```
djrtwo: seems reasonable to assign `epoch_start_shard` before the loop

```python
epoch_start_shard = get_epoch_start_shard(state, epoch)
for ....
    shard = (epoch_start_shard + committee_number) % SHARD_COUNT
    ...

File: specs/core/0_beacon-chain.md [hunk 2]
```diff
@@ -742,17 +753,23 @@ def get_randao_mix(state: BeaconState,
...
+        for index in get_crosslink_committee(state, epoch, shard):
+            validator = validators[index]
+            committee_data[shard].pubkeys.append(validator.pubkey)
+            # `index` (top 7 bytes) + `slashed` (8th bit) + `effective_balance` (bottom 7 bits)
+            compact_validator = index << 8 + validator.slashed << 7 + validator.effective_balance // GWEI_PER_ETH
```
djrtwo: Does this break if validator balance `> 127`? (`2**7`)
Suggest we allocate more bits here

djrtwo: Think this might need a `Uint64` cast

JustinDrake: The effective balance is capped at 32 ETH.

JustinDrake:
@protolambda Testing goes a little crazy on this branch. Not sure what to do.

Screenshot 2019-06-27 at 11 06 43

protolambda:
I'm seeing the same on this branch, here's my test output:
https://gist.github.com/protolambda/f60b5eecba89a4f29367ae9afc76c725

specs/core/0_beacon-chain.md
djrtwo: Right

My main concern with this section is how specifically it is tuned to our section of constants

djrtwo:
Python crashes are often recursion bugs

djrtwo:
taking a look at this

djrtwo:
You have a circular dependency

`get_compact_committees_root` -> `get_crosslink_committee` -> `generate_seed` -> `get_compact_committees_root`

specs/core/0_beacon-chain.md
JustinDrake: Made the `compact_balance` 15 bits :)

djrtwo:
I fixed the SSZ typing error, but there is a more substantive error now. We are trying to set `compact_committees` for an epoch `ACTIVATION_EXIT_EPOCH`s in the future. We can;t actually reliably do this because we have `MIN_SEED_LOOKAHEAD`.

It seems more appropriate to just set for `MIN_SEED_LOOKAHEAD` epochs in the future because that is the most immediate epoch available that we know both the v-set and the seed (requirements for computing the committees).

Setting it to this value actually triggers an issue in `get_epoch_start_shard` which needs to be adjusted to work for epochs more than just the next. Need to think this over a bit.

motes.md
```diff
@@ -0,0 +1,42 @@
+* `BLS_WITHDRAWAL_PREFIX`
```
JustinDrake: Loving your motes :) Will comment on the various points later today 😂

motes.md
```diff
@@ -0,0 +1,42 @@
...
+  * Is this actually tunable?
+  * If so, what are the reprecussions?
+* `ACTIVATION_EXIT_DELAY`
+  * Reaquaint with purpose
+* AttesterSlashings
+  * `MAX_ATTESTER_SLASHINGS` is 1.
+  * Are there scenarios in which validators can create more effective slashable
+    messages than can be included on chain? For example, Validators split up to
+    create double attestations for checkpoints but different (junk) crosslink
+    data to prevent them from being aggregatable to the fullest
```
JustinDrake: I remember thinking about this. One way to create more effective slashable messages is as follows. Let's say you control `n` validators, and you want to make bad attestations `a_1`, ..., `a_n` for each of these validators. The attacker can broadcast aggregated attestations `(a_1, a_i)` for `i > 1` and never reveal the individual `a_i`. This is an argument for generalising the bitfield to a multibitfield.

motes.md
```diff
@@ -0,0 +1,42 @@
...
+  * Max is for block size, no?
```
JustinDrake: `MAX_ATTESTER_SLASHINGS` is to cap the maximum block size, yes.

motes.md
```diff
@@ -0,0 +1,42 @@
...
+* Signature domains
+  * Max 4byte ints
```
JustinDrake: That's a lot of signature domains :)

motes.md
```diff
@@ -0,0 +1,42 @@
...
+* `Version` not defined in one of the lists of custom types (2!!). ensure in spec
```
JustinDrake: To be cleaned up

motes.md
```diff
@@ -0,0 +1,42 @@
...
+* `PendingAttestation`
+  * Don't think `proposer_index` is actually necessary here because effective
+    balance is stable until end of epoch so can do dynamic lookups
```
JustinDrake: Oh, nice find! We only use `proposer_index` once so this is a nice cleanup I guess.

motes.md
```diff
@@ -0,0 +1,42 @@
...
+* is_genesis_trigger
+  * only run at ends of blocks to preserve invariant that eth1data.deposit_root
+    is the deposit root at the _end_ of an eth1 block
```
JustinDrake: Done

motes.md
```diff
@@ -0,0 +1,42 @@
...
+* `Attestation` 
+  * why bitfields not together?
```
JustinDrake: What do you mean by "together"? Neither concatenation nor interleaving is nice. A mathematically clean solution would be (ternary) "trits".

motes.md
```diff
@@ -0,0 +1,42 @@
...
+* `Transfer`
+  * replay mechanism... say the slot gets missed and you sign another transfer
+  * in a fork you could include both transfers
```
JustinDrake: I discussed this with Vitalik at Devcon. The current replay mechanism is quite bad and user unfriendly. The latest suggestion is to have a buffer of the latest `n` (where `n` is sufficiently large) transfer roots, and use that as replay protection.

Since we know we want to change the current transfer logic, can we please remove transfers from phase 0? Including the current logic is now actively misleading.

motes.md
```diff
@@ -0,0 +1,42 @@
...
+* `get_previous_epoch`
+  * do a once over on the genesis stuff
```
JustinDrake: Not sure I understand

motes.md
```diff
@@ -0,0 +1,42 @@
...
+* `get_epoch_start_shard`
+  * checking next hinges upon the fact that the validator set for the next
+    epoch is 100% known at the current epoch. Ensure this is the case
```
JustinDrake: Don't understand

motes.md
```diff
@@ -0,0 +1,42 @@
...
+* `get_block_root_at_slot` .. `generate_seed` can be bade into one line
+  function signatures
```
JustinDrake: Yay for one-line function signatures. This is part of the cleanup planned in #918.

motes.md
```diff
@@ -0,0 +1,42 @@
...
+* `get_shuffled_index`
+  * I think it should be maybe `assert index_count <= VALIDATOR_REGISTRY_LIMIT`
+  * is the `2**40` special for security of alg? probably.
+
+
+pubkey/privkey g1 vs g2
```
JustinDrake: pubkey should be in `g1` since we're adding a bunch of them when aggregating.

motes.md
```diff
@@ -0,0 +1,42 @@
...
+  * in a fork you could include both transfers
+* `get_previous_epoch`
+  * do a once over on the genesis stuff
+* `get_epoch_start_shard`
+  * checking next hinges upon the fact that the validator set for the next
+    epoch is 100% known at the current epoch. Ensure this is the case
+* `get_block_root_at_slot` .. `generate_seed` can be bade into one line
+  function signatures
+* `get_shuffled_index`
+  * I think it should be maybe `assert index_count <= VALIDATOR_REGISTRY_LIMIT`
```
JustinDrake: Right

motes.md
```diff
@@ -0,0 +1,42 @@
...
+  * is the `2**40` special for security of alg? probably.
```
JustinDrake: Hum, not sure.

motes.md
```diff
@@ -0,0 +1,42 @@
+* `BLS_WITHDRAWAL_PREFIX`
+  * Why int rather than bytes?
+* `MIN_SEED_LOOKAHEAD`
+  * Is this actually tunable?
+  * If so, what are the reprecussions?
+* `ACTIVATION_EXIT_DELAY`
+  * Reaquaint with purpose
```
JustinDrake: The problem is that attackers can know the randomness beyond `MIN_SEED_LOOKAHEAD`, and hence bias the shuffling by strategically registering/deregistering validators in the window between `MIN_SEED_LOOKAHEAD` and `MIN_SEED_LOOKAHEAD + 3` to "grind" the shuffling. (The number 3 is large enough to guarantee that, with high probability, an attacker's lookahead cannot go beyond `MIN_SEED_LOOKAHEAD + 3`.)

motes.md
djrtwo: Oh no! I accidentally pushed. Haaa

motes.md
djrtwo: I just meant in the field ordering

motes.md
djrtwo: Was personal note to dig into how `previous_epoch` performs near genesis epoch. Generally, we are just skipping things in 0th or 1st epoch

motes.md
djrtwo: Ensure that there are no ways that during epoch N, we don't 100% know the active v-set during epoch N+1. In general is our assumption but want to go through and check.

specs/core/0_beacon-chain.md
```diff
@@ -745,17 +757,25 @@ def get_randao_mix(state: BeaconState,
...
-    return state.active_index_roots[epoch % EPOCHS_PER_HISTORICAL_VECTOR]
+    committees = [CompactCommittee() for _ in range(SHARD_COUNT)]
+    start_shard = get_epoch_start_shard(state, epoch)
+    for committee_number in range(get_epoch_committee_count(state, epoch)):
+        shard = Shard((start_shard + committee_number) % SHARD_COUNT)
+        for index in get_crosslink_committee(state, epoch, shard):
+            validator = state.validators[index]
+            committees[shard].pubkeys.append(validator.pubkey)
+            compact_balance = validator.effective_balance // EFFECTIVE_BALANCE_INCREMENT
+            # `index` (top 6 bytes) + `slashed` (16th bit) + `compact_balance` (bottom 15 bits)
```
CarlBeek: Is it worth adding to this comment something about how `compact_balance` is always 6 bits or fewer due to the nature of `effective_balance`? As it stands, this looks (although it isn't) like a prime place for overflow errors.

specs/light_client/sync_protocol.md
```diff
@@ -40,7 +40,7 @@ Note that there is now a new way to compute `get_active_validator_indices`:
 
 ```python
 def get_active_validator_indices(state: ExtendedBeaconState, epoch: Epoch) -> List[ValidatorIndex]:
-    return state.active_indices[epoch % ACTIVE_INDEX_ROOTS_LENGTH]
+    return state.active_indices[epoch % compact_committees_rootS_LENGTH]
```
ralexstokes:
```suggestion
    return state.active_indices[epoch % EPOCHS_PER_HISTORICAL_VECTOR]
```

djrtwo: thanks!

specs/core/0_beacon-chain.md
djrtwo: making note of it in the secondary doc for now.

specs/core/0_beacon-chain.md
```diff
@@ -1475,7 +1492,7 @@ def process_slashings(state: BeaconState) -> None:
 ```python
 def process_final_updates(state: BeaconState) -> None:
-    next_epoch = current_epoch + 1
+    next_epoch = Shard(current_epoch + 1)
```
NIC619: typo?

protolambda: See https://github.com/ethereum/eth2.0-specs/commit/6cfd3b5047ec54e0a4250a2b262d71a87e99830c

PR #1220, opened 2019-06-26
Critical fix: introduce back total-value check

protolambda:
This was dropped in a376b6607fe5e6406371f44254960e891ee5ee8d, as improvement in dust checking.
Now that dust-checking is done, we still need to check if the sender has the minimum value, as decrease balance just clips to 0.
See be86f966f87958856584b3f20c095abf910a3d0c for older dust-creation problem work around, which was dropped in the above.

The bug enabled you to transfer your full balance to someone else, and pay the same amount in fee, possibly to a puppet proposer to collect back funds.
Effectively enabling printing of money. Silly bug, good to fix and introduce tests for.


protolambda:
Added more tests, and tidied formatting a bit. This should cover the edge cases in the code from client implementations better.

PR #1221, opened 2019-06-26
Clarify `get_randao_mix` accessor

ralexstokes:
We avoid a genesis underflow by taking the randao epoch in `generate_seed` to be `+ EPOCHS_PER_HISTORICAL_VECTOR`.

This conflicts with the expected epoch bounds noted  in `get_randao_mix` and this PR attempts to clarify the situation by leaving a note.


JustinDrake:
IMO both comments (the old one, and the new one) are confusing. Less is more.

ralexstokes:
I agree less is more :)

I can say that we had a validation condition in Trinity based on an earlier instantiation of this comment and it did help catch a bug. So providing the guidance is helpful...

I guess it is a matter of style as to where this knowledge lives, if anywhere... Feel free to close the PR if there is consensus this isn't a helpful change

PR #1223, opened 2019-06-27
Ignore .DS_Store

JustinDrake:


djrtwo:
This is usually more appropriate in your global gitignore https://stackoverflow.com/a/22885996

PR #1224, opened 2019-06-27
Add Bitlist and Bitvector

dankrad:
Replaces #1096 for Bitfields only

- Adding `Bitvector` (static length Bitfield) and `Bitlist` (dynamic length Bitfield)

Further fixes:
- Update simple-serialize.md with the changes that were made to the SSZ python implementation in #1180 
- Rename `Bool`, `Bit` -> `boolean`, `bit`. This means that all basic SSZ types are now lowercase (`bool` is not possible due to python type clash)


scripts/build_spec.py
```diff
@@ -25,7 +25,7 @@
     signing_root,
 )
 from eth2spec.utils.ssz.ssz_typing import (
-    Bit, Bool, Container, List, Vector, Bytes, uint64,
+    bit, boolean, Container, List, Vector, Bytes, uint64,
```
JustinDrake: Why not be consistent with casing, i.e. `bit` => `Bit` and `boolean` => `Boolean`? Also why `Boolean` instead of the "standard" `Bool`?

JustinDrake: My guess would be lower-case for basic types, and upper-case for non-basic types. In which case this makes sense :)

dankrad: Yes, I think this is consistent with typing, just like uintN are lowercase. So I'm using "boolean" because bool is already a type in python.

specs/core/1_custody-game.md
```diff
@@ -272,6 +272,16 @@ def get_custody_chunk_count(crosslink: Crosslink) -> int:
     return crosslink_length * chunks_per_epoch
 ```
 
+### `get_bitfield_bit`
```
JustinDrake: We don't need this function, right?

dankrad: I did not change this logic at the moment because that will likely change quite a bit anyway. I think in the future, we will not need it, but I wanted to keep the old implementation alive for now.

dankrad:
So the test that is currently failing is for the bridge to py-ssz. I think we probably can't get that to work until py-ssz has the new Bitvector and Bitlist classes. (Also, I noticed that at the moment, py-ssz does not seem to support the new max-length lists)

specs/core/0_beacon-chain.md
```diff
@@ -1322,38 +1291,38 @@ def process_justification_and_finalization(state: BeaconState) -> None:
...
-    state.justification_bitfield = (state.justification_bitfield << 1) % 2**64
+    state.justification_bitfield = Bitvector[4](*([0] + state.justification_bitfield[0:3]))
     previous_epoch_matching_target_balance = get_attesting_balance(
         state, get_matching_target_attestations(state, previous_epoch)
     )
     if previous_epoch_matching_target_balance * 3 >= get_total_active_balance(state) * 2:
         state.current_justified_epoch = previous_epoch
         state.current_justified_root = get_block_root(state, state.current_justified_epoch)
-        state.justification_bitfield |= (1 << 1)
+        state.justification_bitfield[1] = True
```
JustinDrake: Semantically cleaner to use `0b1` for bits (as opposed to `True`).

specs/core/0_beacon-chain.md
```diff
@@ -1322,38 +1291,38 @@ def process_justification_and_finalization(state: BeaconState) -> None:
     # Process justifications
     state.previous_justified_epoch = state.current_justified_epoch
     state.previous_justified_root = state.current_justified_root
-    state.justification_bitfield = (state.justification_bitfield << 1) % 2**64
+    state.justification_bitfield = Bitvector[4](*([0] + state.justification_bitfield[0:3]))
```
JustinDrake: Semantically cleaner to use `0b0` for the zero bit (as opposed to `0`).

specs/core/0_beacon-chain.md
```diff
@@ -525,7 +523,7 @@ class BeaconState(Container):
     previous_justified_root: Hash  # Previous epoch snapshot
     current_justified_epoch: Epoch
     current_justified_root: Hash
-    justification_bitfield: uint64  # Bit set for every recent justified epoch
+    justification_bitfield: Bitvector[4]  # Bit set for every recent justified epoch
```
JustinDrake: May be worth adding a configuration/constant for `4`.

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
```diff
@@ -281,6 +281,10 @@ class ElementsType(ParamsMeta):
     length: int
 
 
+class BitElementsType(ElementsType):
+    elem_type = boolean
```
vbuterin: Shouldn't this have more of an implementation? Even `def __getitem__(self, i): return (self.items[i // 8] >> (i % 8)) % 2` and the equivalent for setitem?

vbuterin: Also, in the phase 0 spec, there is one place where we left-shift a bitfield by 1. The simplest way to left-shift a bitfield would probably be a slice; we'd need code to do a slice.

dankrad: I am not using the integer representation internally, they are just lists of bools. Only at the time of serialization/merkleization the compact representation is used (makes the implementation much cleaner and probably more performant at the same time)

protolambda:
See a5154da1ff76e8b768e64987471820f958c391d1 on the `bitfield-suggestion` branch.
Changed things:
- made the base class "Bitfield"
- made the metaclass consistent with the metaclass for bytes
- made it a subclass of elements-base, so that it's a list, *and slicing works* (slice of `boolean` as slice result`, can be coerced back into a bitfield
- some updates to the ssz-implementation; handle chunk count as special case for bitfields, since it's not like the other subclasses of `Elements` where we just assume a basic type is at least 1 byte.

specs/core/0_beacon-chain.md
```diff
@@ -1322,38 +1292,38 @@ def process_justification_and_finalization(state: BeaconState) -> None:
     # Process justifications
     state.previous_justified_epoch = state.current_justified_epoch
     state.previous_justified_root = state.current_justified_root
-    state.justification_bitfield = (state.justification_bitfield << 1) % 2**64
+    state.justification_bitfield = Bitvector[4](*([0b0] + state.justification_bitfield[0:JUSTIFICATION_BITVECTOR_LENGTH - 1]))
```
JustinDrake: The alternative to bitshift is in-place writes, similar to how `randao_mixes` is handled. It would avoid the `Bitvector[4](*(` magic, and be more consistent with the rest of the spec.

dankrad: You mean the modulo logic? I think in this case it would make the following code (that checks recently justified epochs and handles finalization) harder to read, so I would prefer the current state for readability if you don't have any other suggestions.

specs/core/0_beacon-chain.md
```diff
@@ -1322,38 +1292,39 @@ def process_justification_and_finalization(state: BeaconState) -> None:
     # Process justifications
     state.previous_justified_epoch = state.current_justified_epoch
     state.previous_justified_root = state.current_justified_root
-    state.justification_bitfield = (state.justification_bitfield << 1) % 2**64
+    state.justification_bitfield = Bitvector[JUSTIFICATION_BITVECTOR_LENGTH](
```
djrtwo: I'm not sure using a `Bitvector` for `justification_bitfield` actually aids in readability and am inclined to leave as is.

Just want to gauge the temperature and see if I'm alone in thinking this

protolambda: Technically we can implement the `<<` operator. Or base bitfield off of `int` instead of `list`. I feel like `byte` (or the old `uint64` if we don't care about size) here would be a much simpler solution, and less error prone (we have to test all of this!).

Also, I still like to see bitfields, but for places where we actually can't fit it into a standard type. Which is phase 1+. This would lift the pressure of SSZ changes now in phase-0 freeze, and we can design towards phase-1 needs.
Another worry is whether or not bitfields should be mutable in the spec. A `list` is. An `int` is not. Here we just overwrite it, so I'm inclined to say we don't need mutability. Maybe in phase 1+.

Last thing, we've coercion of container fields, so I think it's fine to remove the `Bitvector[JUSTIFICATION_BITVECTOR_LENGTH](` wrapper, as the container field-setter will force it into that type.

protolambda: And *personally* (willing to help out here, but carefully), I prefer `bytes`/`bytesN`, enhanced with a set of bitfield operations, for phase 1+, and no bitfields in phase0. Simplicity and an idiomatic preference of bytes (good support everywhere) over custom bitfield logic.

dankrad: I can see that the Bitvector is not strictly necessary for the justification_bitfield. I think it could work perfectly well with Vector[boolean, 4]. The Bitvector is not necessary as an optimisation. However, I would still argue that either of those are better than integer notation:
- for people who are not used to bit manipulation, array notation is much clearer
- bit manipulation depends on more conventions (LSB or MSB first, etc.) than array notation; so even a reader familiar with bit manipulation has to do more to understand what is happening
- with vectors, we can make it very clear that we only ever use the last 4 epochs to estimate finalization
I understand bit manipulation well, and so do both of you obviously, but that does not at all mean that it is a natural way to describe things! An array is much more natural object here. If we want to implement an array shift notation, I fully support that, but I think it will not increase readability (as again, the reader would have to understand how the array shift is defined).


> And _personally_ (willing to help out here, but carefully), I prefer `bytes`/`bytesN`, enhanced with a set of bitfield operations, for phase 1+, and no bitfields in phase0. Simplicity and an idiomatic preference of bytes (good support everywhere) over custom bitfield logic.

Now, on this more general point, I do think that the spec benefits from having the Bitlist at phase 0. That is simply because for a reader of the spec, the Bitlist behaves exactly like a List[boolean, N] as far as the interface is concerned. There is no need for "get_bitfield_bit" and "verify_bitfield"-like operations, which should be handled by types rather than by the spec itself. And for the attestation/custody bitfields, we do want the more efficient storage properties. And these are both already in phase 0 (though we could probably change the custody bitfield).

These are my opinions, of course if the consensus sways against bitfields we should stop here and leave phase 0 as it is (that will however likely mean no bitfields ever as we will be baking the attestation logic in now).

protolambda: Convinced :+1: But want bitfields to serialize/hash well (tests! I can help if needed), and hope the readability improves (avoid wrapping with coercion, and prefer no unpacking magic in the spec doc itself).

test_libs/pyspec/eth2spec/utils/ssz/test_ssz_impl.py
```diff
@@ -85,6 +85,16 @@ def merge(a: str, branch: Iterable[str]) -> str:
      "c5c2", chunk("c5c2")),
     ("bitlist TFTFFFTTFTFFFFTT", Bitlist[16](1, 0, 1, 0, 0, 0, 1, 1, 0, 1, 0, 0, 0, 0, 1, 1),
      "c5c201", h(chunk("c5c2"), chunk("10"))),
+    ("long bitvector", Bitvector[512](1 for i in range(512)),
```
protolambda: A test for bitlists/vectors of byte size, and smaller, would be nice to have.

specs/core/0_beacon-chain.md
dankrad: > avoid wrapping with coercion, and prefer no unpacking magic in the spec doc itself

So I replaced that with a slice notation, have a look if you like it better :)

test_libs/pyspec/eth2spec/utils/ssz/test_ssz_impl.py
dankrad: OK, done :)

test_libs/pyspec/eth2spec/utils/ssz/ssz_typing.py
vbuterin: Aaah I see. Got it.

specs/core/0_beacon-chain.md
vbuterin: I felt that the whole point of adding bitfield classes is to remove the need to do integer manipulation in the user-facing spec code. So definitely oppose keeping that justification bitfield as uint64 for that reason. Bitvector seems fine to me.

JustinDrake:
Cleaned up naming:

1) "bitfield" => "bits" (shorter; natural pluralisation, e.g. `custody_bit` => `custody_bits`; avoids introducing a fourth concept beyond "bit", "bitlist" and "bitvector")
2) `JUSTIFICATION_BITVECTOR_LENGTH` => `JUSTIFICATION_BITS_LENGTH` for consistency with `justification_bits`

JustinDrake:
@protolambda: Can we avoid throwing an error with `Checkpoint(previous_epoch, get_block_root(state, previous_epoch))`?

```
>           state.current_justified_checkpoint = Checkpoint(current_epoch, get_block_root(state, current_epoch))
E           TypeError: __init__() takes 1 positional argument but 3 were given
```

protolambda:
@JustinDrake Containers expect named arguments, not bare arguments. I prefer named arguments, as it enables you to shorten the names for input variables, without being less descriptive where it matters. And it enables you to leave out unused arguments. If we really want, I can add support for unnamed args, but it feels error prone (things change, and phase 1 will keep changing for a while).

protolambda:
@dankrad @JustinDrake Can you take a look at the more explicit wording for list-limits? It's important, since it is part of consensus (max operations in a block), and merkleization with padding is a subtle combined thing, possibly not familiar to some readers.

protolambda:
Resolving the merge conflict now

specs/core/0_beacon-chain.md
```diff
@@ -1323,34 +1295,32 @@ def process_justification_and_finalization(state: BeaconState) -> None:
 
     # Process justifications
     state.previous_justified_checkpoint = state.current_justified_checkpoint
-    state.justification_bitfield = (state.justification_bitfield << 1) % 2**64
-    previous_epoch_matching_target_balance = get_attesting_balance(
-        state, get_matching_target_attestations(state, previous_epoch)
-    )
-    if previous_epoch_matching_target_balance * 3 >= get_total_active_balance(state) * 2:
+    state.justification_bits[1:] = state.justification_bits[:JUSTIFICATION_BITS_LENGTH - 1]
```
djrtwo: If we are comfortable using python array slices here, I'd say we go all the way and use the `-1` notation.
```suggestion
    state.justification_bits[1:] = state.justification_bits[:-1]
```

djrtwo: could also do

```suggestion
    state.justification_bits = [0b0] + state.justification_bits[:-1]
```

specs/simple-serialize.md
```diff
@@ -37,71 +37,96 @@
...
+class ContainerExample(Container):
+    foo: uint64
+    bar: boolean
+```
+* **vector**: ordered fixed-length homogeneous collection, with `N` values
+    * notation `Vector[type, N]`, e.g. `Vector[uint64, N]`
+* **list**: ordered variable-length homogeneous collection, limited to `N` values
+    * notation `List[type, N]`, e.g. `List[uint64, N]`
+* **bitvector**: ordered fixed-length collection of `boolean` values, with `N` bits
+    * notation `Bitvector[N]`
```
djrtwo: `BitVector` and `BitList` (capitalize first letter of second word) are more pythonic names for types. 
What do you think about changing this?

specs/validator/0_beacon-chain-validator.md
```diff
@@ -329,19 +329,19 @@ Next, the validator creates `attestation`, an [`Attestation`](../core/0_beacon-c
 
 Set `attestation.data = attestation_data` where `attestation_data` is the `AttestationData` object defined in the previous section, [attestation data](#attestation-data).
 
-##### Aggregation bitfield
+##### Aggregation bits
 
-* Let `aggregation_bitfield` be a byte array filled with zeros of length `(len(committee) + 7) // 8`.
+* Let `aggregation_bits` be a byte array filled with zeros of length `(len(committee) + 7) // 8`.
```
djrtwo: Suggest we use a `Bitlist` here and use the `0b1` notation and simple array index for setting the bit

specs/validator/0_beacon-chain-validator.md
```diff
@@ -329,19 +329,19 @@ Next, the validator creates `attestation`, an [`Attestation`](../core/0_beacon-c
...
 
-*Note*: Calling `get_attesting_indices(state, attestation.data, attestation.aggregation_bitfield)` should return a list of length equal to 1, containing `validator_index`.
+*Note*: Calling `get_attesting_indices(state, attestation.data, attestation.aggregation_bits)` should return a list of length equal to 1, containing `validator_index`.
 
-##### Custody bitfield
+##### Custody bits
 
-* Let `custody_bitfield` be a byte array filled with zeros of length `(len(committee) + 7) // 8`.
-* Set `attestation.custody_bitfield = custody_bitfield`.
+* Let `custody_bits` be a byte array filled with zeros of length `(len(committee) + 7) // 8`.
```
djrtwo: Describe this as a `Bitlist` for simplicity

test_libs/pyspec/eth2spec/test/phase_0/block_processing/test_process_attestation.py
```diff
@@ -274,11 +275,14 @@ def test_bad_crosslink_end_epoch(spec, state):
 
 @with_all_phases
 @spec_state_test
-def test_inconsistent_bitfields(spec, state):
+def test_inconsistent_bits(spec, state):
     attestation = get_valid_attestation(spec, state)
     state.slot += spec.MIN_ATTESTATION_INCLUSION_DELAY
 
-    attestation.custody_bitfield = deepcopy(attestation.aggregation_bitfield) + b'\x00'
+    custody_bits = deepcopy(attestation.aggregation_bits)
```
djrtwo: pythonic method to copy list
```suggestion
    custody_bits = attestation.aggregation_bits[:]
```

test_libs/pyspec/eth2spec/test/phase_0/block_processing/test_process_attestation.py
```diff
@@ -287,11 +291,11 @@ def test_inconsistent_bitfields(spec, state):
 
 @with_phases(['phase0'])
 @spec_state_test
-def test_non_empty_custody_bitfield(spec, state):
+def test_non_empty_custody_bits(spec, state):
     attestation = get_valid_attestation(spec, state)
     state.slot += spec.MIN_ATTESTATION_INCLUSION_DELAY
 
-    attestation.custody_bitfield = deepcopy(attestation.aggregation_bitfield)
+    attestation.custody_bits = deepcopy(attestation.aggregation_bits)
```
djrtwo: ditto
```suggestion
    attestation.custody_bits = attestation.aggregation_bits[:]
```

test_libs/pyspec/eth2spec/test/phase_0/block_processing/test_process_attestation.py
```diff
@@ -300,11 +304,12 @@ def test_non_empty_custody_bitfield(spec, state):
 
 @with_all_phases
 @spec_state_test
-def test_empty_aggregation_bitfield(spec, state):
+def test_empty_aggregation_bits(spec, state):
     attestation = get_valid_attestation(spec, state)
     state.slot += spec.MIN_ATTESTATION_INCLUSION_DELAY
 
-    attestation.aggregation_bitfield = b'\x00' * len(attestation.aggregation_bitfield)
+    attestation.aggregation_bits = Bitlist[spec.MAX_INDICES_PER_ATTESTATION](
```
djrtwo: Doesn't the ssz type handle the casting here?
```suggestion
    attestation.aggregation_bits = [0b0] * len(attestation.aggregation_bits)
```

djrtwo:
Great work @dankrad! 
I have a number of minor comments but very close to merge time :)

specs/core/0_beacon-chain.md
JustinDrake: Did that and @dankrad mentioned this makes `justification_bits` a list (as opposed to a bitvector). The tests didn't complain though.

djrtwo: I think the assignment can handle the cast

specs/simple-serialize.md
```diff
@@ -136,23 +161,47 @@ return serialized_type_index + serialized_bytes
...
+    - `1+` chunks: pad to the next power of 2, merkleize as binary tree.
+    - with `pad_to` set: pad the `data` with zeroed chunks to this power of two (virtually for memory efficiency).
 * `mix_in_length`: Given a Merkle root `root` and a length `length` (`"uint256"` little-endian serialization) return `hash(root + length)`.
 * `mix_in_type`: Given a Merkle root `root` and a type_index `type_index` (`"uint256"` little-endian serialization) return `hash(root + type_index)`.
 
 We now define Merkleization `hash_tree_root(value)` of an object `value` recursively:
 
 * `merkleize(pack(value))` if `value` is a basic object or a vector of basic objects
-* `mix_in_length(merkleize(pack(value)), len(value))` if `value` is a list of basic objects
+* `mix_in_length(merkleize(pack(value), pad_to=N), len(value))` if `value` is a list of basic objects. `N` is the amount of chunks needed for the list limit (*this depends on element-type size*).
```
naterush: To confirm, `N` is the max length * element length (in bytes) //  BYTES_PER_CHUNK? It might be nice to define this explicitly.

specs/simple-serialize.md
```diff
@@ -136,23 +161,47 @@ return serialized_type_index + serialized_bytes
...
 We first define helper functions:
 
 * `pack`: Given ordered objects of the same basic type, serialize them, pack them into `BYTES_PER_CHUNK`-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
-* `merkleize`: Given ordered `BYTES_PER_CHUNK`-byte chunks, if necessary append zero chunks so that the number of chunks is a power of two, Merkleize the chunks, and return the root. Note that `merkleize` on a single chunk is simply that chunk, i.e. the identity when the number of chunks is one.
+* `merkleize(data, pad_to)`: Given ordered `BYTES_PER_CHUNK`-byte chunks, if necessary append zero chunks so that the number of chunks is a power of two, Merkleize the chunks, and return the root.
+  There merkleization depends on the input length:
+    - `0` chunks: A chunk filled with zeroes.
+    - `1` chunk: A single chunk is simply that chunk, i.e. the identity when the number of chunks is one.
+    - `1+` chunks: pad to the next power of 2, merkleize as binary tree.
+    - with `pad_to` set: pad the `data` with zeroed chunks to this power of two (virtually for memory efficiency).
```
naterush: What if `pad_to` is set to a non-power of two (like, `3`)? Are we insisting that max length must be a power of two? Or should `pad_to` be rounded up with the above rules?



E.g. consider a List[uint128, ]

specs/core/0_beacon-chain.md
dankrad: We could probably overload the `1` operator to do it but currently I think it doesn't:

```python
>>> a = ssz_typing.Bitvector[2](0, 0)
>>> b = [0] + a
>>> type(b)
<class 'list'>
```

test_libs/pyspec/eth2spec/test/phase_0/block_processing/test_process_attestation.py
dankrad: I don't think that python does any typecasting on assignment, that would be very un-pythonic (they are only type annotations, after all). The only exceptions is slice assignments (which aren't really assignments)

specs/simple-serialize.md
protolambda: Yes, that should be it, different `N` than in typing. My bad. Will fix it.

specs/simple-serialize.md
protolambda: rounded up for padding. But limit should be strictly enforced. We really only have power of 2 limits currently, but can imagine it's good to define behavior for non-power of 2 limits.

protolambda:
Fixed the last suggestion: removed unused import for code linter

test_libs/pyspec/eth2spec/test/phase_0/block_processing/test_process_attestation.py
djrtwo: you're right. don't know what I was thinking!

protolambda: It's unpythonic. But support is there to do implicit type casting for integers, and other types, where we really wouldn't want to for readability.

specs/simple-serialize.md
protolambda: updated it

specs/simple-serialize.md
protolambda: updated it

PR #1225, opened 2019-06-27
get_attesting_indices set instead of sorted

djrtwo:
It came to my attention that `get_attesting_indices` was still unnecessarily sorting instead of returning a `set`. The only case that needs sorting is when it is used in `convert_to_indexed` rather than in all of the attester accessors used in rewards processing during epoch processing.

This actually let me to uncover a small bug in phase 1 that has been resolved.

Thanks @prestonvanloon for bringing this to my attention. I hope this relieves that hotspot you found in your epoch processing.


specs/core/1_custody-game.md
```diff
@@ -566,7 +569,8 @@ def process_bit_challenge(state: BeaconState,
     chunk_count = get_custody_chunk_count(attestation.data.crosslink)
     assert verify_bitfield(challenge.chunk_bits, chunk_count)
     # Verify the first bit of the hash of the chunk bits does not equal the custody bit
-    custody_bit = get_bitfield_bit(attestation.custody_bitfield, attesters.index(challenge.responder_index))
```
djrtwo: _Bugfix_: The previous accessor here for the responder bit just from the list of attestors would likely correspond to the incorrect bit in the bitfield. Instead it should be the index of the responder within the entire committee.

specs/core/0_beacon-chain.md
```diff
@@ -950,11 +950,11 @@ def convert_to_indexed(state: BeaconState, attestation: Attestation) -> IndexedA
     attesting_indices = get_attesting_indices(state, attestation.data, attestation.aggregation_bitfield)
     custody_bit_1_indices = get_attesting_indices(state, attestation.data, attestation.custody_bitfield)
     assert set(custody_bit_1_indices).issubset(attesting_indices)
-    custody_bit_0_indices = [index for index in attesting_indices if index not in custody_bit_1_indices]
+    custody_bit_0_indices = attesting_indices.difference(custody_bit_1_indices)
 
     return IndexedAttestation(
-        custody_bit_0_indices=custody_bit_0_indices,
-        custody_bit_1_indices=custody_bit_1_indices,
+        custody_bit_0_indices=sorted(custody_bit_0_indices),
```
prestonvanloon: It would be helpful to explain why this has to be sorted. Either here or in the definition of IndexedAttestation. 

Is it an implementation detail or absolute requirement for spec conformance?

specs/core/0_beacon-chain.md
```diff
@@ -865,13 +865,13 @@ def get_crosslink_committee(state: BeaconState, epoch: Epoch, shard: Shard) -> S
 ```python
 def get_attesting_indices(state: BeaconState,
                           attestation_data: AttestationData,
-                          bitfield: bytes) -> Sequence[ValidatorIndex]:
+                          bitfield: bytes) -> Set[ValidatorIndex]:
     """
     Return the sorted attesting indices corresponding to ``attestation_data`` and ``bitfield``.
```
prestonvanloon:
```suggestion
    Return the set of attesting indices corresponding to ``attestation_data`` and ``bitfield``.
```

prestonvanloon:
Thanks for putting this together so quickly.

specs/core/0_beacon-chain.md
JustinDrake: > absolute requirement for spec conformance?

Yes.

> explain why this has to be sorted

One reason is that it will allow for better network-layer compression of indices. The rationale for design decisions will come in a separate document which is definitely currently lacking.

prestonvanloon: So `convert_to_indexed` is called on `process_attestation` so this sorting is still going to be a huge bottleneck. 

From what I am hearing, sorting only serves for some auxiliary benefits but is not actually required for consensus. 
Is that the case?

JustinDrake: > `validate_indexed_attestation` is also called on `process_attestation` so we are sorting on every processing.

Notice that sort*ing* the indices is not required, merely checking that the indices are sort*ed*. This can be done in linear time with a single-pass of the indices:

```python
previous_index = 0
for index in indices:
    assert previous_index <= index
    previous_index = index
```

protolambda: Already posted this in the gitter, before the issue opened :)

protolambda: But there is a different place where it could matter (attester slashing converting to sorted), but that shouldn't be a bottleneck. And if it is, you can use precomputed sorted committees (precompute is O(n) work, with sufficient memory, since unshuffled indices are sequential-ish (fit in a simple array for fast lookup/writes)).

prestonvanloon: I understand that bit, but if the indices are not sorted (as they usually are), we have to sort them.
Therefore sorting is required. 

All of this sounds still sounds like implementation detail and not a consensus requirement. 
Can we get confirmation? Is the sorting requirement for consensus?

JustinDrake: > Can we get confirmation? Is the sorting requirement for consensus?

Confirmed. Sor**ing** is not required for consensus. Checking indices are sort**ed** in `process_attestation` is required for consensus.

PR #1226, opened 2019-06-28
Slashing invariants for validator without message history

CarlBeek:
Adds to the V-guide advice on how long to wait before making messages after coming online without knowledge of your message history.

Ready for review & merge. (Ideally before spec freeze)

### DIFF
--- a/specs/validator/0_beacon-chain-validator.md
+++ b/specs/validator/0_beacon-chain-validator.md
@@ -50,6 +50,7 @@
     - [How to avoid slashing](#how-to-avoid-slashing)
         - [Proposer slashing](#proposer-slashing)
         - [Attester slashing](#attester-slashing)
+        - [Unavailability of message history](#unavailability-of-message-history)
 
 <!-- /TOC -->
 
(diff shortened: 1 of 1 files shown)


vbuterin:
Do we assume that clients have reasonably working local clocks? If so, isn't just waiting one epoch, and only signing votes whose source is at least (epoch you come online + 1) sufficient?

CarlBeek:
> Do we assume that clients have reasonably working local clocks? If so, isn't just waiting one epoch, and only signing votes whose source is at least (epoch you come online + 1) sufficient?

Basically, yes. The downside of taking this approach is that a validator needs special logic to handle making attestations differently when it comes online. What I am going for here is: "what is the minimum delay a validator has to wait for before it can just make votes "normally" according to the V-guide."

In the happy case the difference is small, but when the chain is failing to justify it gets more interesting.

@vbuterin, Thoughts?

CarlBeek:
I am closing this as I don't think it is vital for implementers. They can figure this out on their own and choose between the two systems discussed above based on their own preferences.

PR #1228, opened 2019-06-29
Some fixes to finality tests

djrtwo:
Fix finality_12. This should be a good base to fix the others.

* need to set the associated epoch boundary block roots to ensure that `get_matching_target_` works
* justification bit 0 should be set at first so it gets shift to position 1. Then 0 gets justified and you have 0 and 1 for finality.
* `aggregation_bitfield` -> `aggregation_bits` in `add_mock_attestations` (use bool array for init)
* `add_mock_attestations` had bug, setting `target=source`. fixed
* `add_mock_attestations` sets `shard` in `attestation.data.crosslink`. This piece of data is required to figure out what committee is attesting. Previous, all were defaulting to zero so each committee added looked the same.

### DIFF
--- a/test_libs/pyspec/eth2spec/test/phase_0/epoch_processing/test_process_justification_and_finalization.py
+++ b/test_libs/pyspec/eth2spec/test/phase_0/epoch_processing/test_process_justification_and_finalization.py
@@ -16,14 +16,6 @@ def get_shards_for_slot(spec, state, slot):
     return [shard + i for i in range(committees_per_slot)]
 
 
-def get_committee_size(spec, epoch_start_shard, shard, committee_count, indices):
-    committee_index = (shard + spec.SHARD_COUNT - epoch_start_shard) % spec.SHARD_COUNT
-    start = (len(indices) * committee_index) // committee_count
-    end = (len(indices) * (committee_index + 1)) // committee_count
-    size = end - start
-    return size
-
-
 def add_mock_attestations(spec, state, epoch, source, target, sufficient_support=False):
     # we must be at the end of the epoch
     assert (state.slot + 1) % spec.SLOTS_PER_EPOCH == 0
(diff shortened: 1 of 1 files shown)


test_libs/pyspec/eth2spec/test/phase_0/epoch_processing/test_process_justification_and_finalization.py
```diff
@@ -147,16 +149,19 @@ def finalize_on_123(spec, state, epoch, support):
...
     c3 = spec.Checkpoint(epoch=epoch - 3, root=b'\xaa' * 32)
     c2 = spec.Checkpoint(epoch=epoch - 2, root=b'\xbb' * 32)
     c1 = spec.Checkpoint(epoch=epoch - 1, root=b'\xcc' * 32)
+    state.block_roots[spec.get_epoch_start_slot(c3.epoch) % spec.SLOTS_PER_HISTORICAL_ROOT] = c3.root
+    state.block_roots[spec.get_epoch_start_slot(c2.epoch) % spec.SLOTS_PER_HISTORICAL_ROOT] = c2.root
+    state.block_roots[spec.get_epoch_start_slot(c1.epoch) % spec.SLOTS_PER_HISTORICAL_ROOT] = c1.root
 
     old_finalized = state.finalized_checkpoint
-    state.previous_justified_checkpoint = c3
+    state.previous_justified_checkpoint = c4
```
CarlBeek: It feels strange to do this as it is not the behaviour one would expect from this test. The problem, however, is that otherwise, a new checkpoint is finalised as the condition of "2nd/3rd most recent epochs are justified, the 2nd using the 3rd as source" is met. This means that in the case where there is insufficient support, `assert state.finalized_checkpoint == old_finalized  # no new finalized` fails. @djrtwo, please can you check me here (and propose an alternative solution if you have one.)

TLDR: this is required because of an artifact of how the tests are constructed by manipulating what is and isn't justified in the state.

CarlBeek:
I extended these finalisation fixes to all of the finalisation cases. Fixes include:
* all of the `boundry_epochs` are now set (needed for the correct calculation of attestations).
* Many list-slicing bugs
* correction of what is justified in each of the tests.

protolambda:
Fixed the attestation creation, since it was broken on mainnet config (multi committees per slot). And also reworded the bitfield indices and justification, it's clear now.

PR #1229, opened 2019-06-29
Make deposit root the root of an SSZ list

JustinDrake:
Make the `deposit_root` the `hash_tree_root` of an SSZ `List`. Benefits:

1) Semantically natural integration of the deposit tree within the SSZ framework.
2) Cleaner and simpler (about 30 lines reduced) `is_genesis_trigger` helper code (see #1202).
3) Ability to refer to the `i`th deposit `state.eth1_data.deposits[i]` using SSZ partials.

Notice that this change to the deposit contract imposes zero gas overhead. Indeed, only `get_deposit_root()` is changed which is called offchain (as opposed to `deposit()` which is called onchain).


protolambda:
Hmm, interesting, that list-pass in constructor seemed to work fine locally. Changing it back to unpack operator, without tuple wrap should work though.

protolambda:
Is the change in how we deal with `deposit.proof` going to be part of this PR, or the genesis-PR (#1202)? If the latter, I think this PR is ready to merge, and we can then finish it as part of the genesis PR, which largely deals with the beacon-spec side of the logic.

JustinDrake:
> Is the change in how we deal with `deposit.proof` going to be part of this PR, or the genesis-PR (#1202)?

I suggest putting everything in this PR, and closing #1202 in favour of this one :) Will deal with `deposit.proof` shortly.

hwwhww:
It seems the PR title is misleading, you're adding `mix_in_length` to the `get_deposit_root()` function and using SSZ `List` instead of changing the leaves `DepositData()` + using `Vector`.

JustinDrake:
> It seems the PR title is misleading

Fixed, thanks 👍

djrtwo:
This PR is now in 1202 so can be closed?

PR #1232, opened 2019-06-29
Explicit assert when accessing state.validators out of bounds

JustinDrake:
Address most of #1091.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -63,6 +63,7 @@
         - [`get_active_validator_indices`](#get_active_validator_indices)
         - [`increase_balance`](#increase_balance)
         - [`decrease_balance`](#decrease_balance)
+        - [`get_validator`](#get_validator)
         - [`get_epoch_committee_count`](#get_epoch_committee_count)
         - [`get_shard_delta`](#get_shard_delta)
         - [`get_epoch_start_shard`](#get_epoch_start_shard)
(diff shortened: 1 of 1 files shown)


protolambda:
Although I do fully support the idea of checking inputs for fuzzing, this PR takes it too extreme:
- IndexError should be like a panic. Not an Assertion. This defeats the whole purpose of input checking. We want to check inputs where we think they are untrusted. Then catch unexpected errors with fuzzing, like Go panics, (IndexError here).
- This pollutes the epoch-processing execution path with asserts. We don't have, or need, any asserts for trusted data.
- This unfortunately breaks a few readability things (typing becomes stronger in a few places, but I dislike changing that in the spec just before freeze)

I propose we introduce `check_index` function which maps an `int` to a `ValidatorIndex`, asserting it is within range, and only apply it on untrusted block inputs. A presentation thing, for after freeze.

JustinDrake:
> IndexError should be like a panic. Not an Assertion.

Disagree for user-inputs which are not trusted.

protolambda:
You're misunderstanding me there. Untrusted inputs can be checked with an assertion instead. This will make those invalid lookups throw an assertion, as part of the regular assertion based validation. Then IndexError can then start to be regarded as a panic, something that only occurs when logic or assumptions are wrong. We can then catch those cases, and separately so from the noise created by correct (i.e. reject block because this/that) assertion errors during fuzzing.
This would effectively require only a check function (or getter with assertion) in a select few spots where we parse validator indices from untrusted block inputs.

JustinDrake:
> in a select few spots 

So I tried going that route to start with and there's a bunch of spots. Don't care too much either way.

PR #1233, opened 2019-06-30
rebased old BLS rework. For discussion purposes.

protolambda:
Honestly, the BLS interface as-is is ok\*. But if we still want to separate out application layer from the inner BLS layer, we may want to consider something alike to this PR; a rebased version of the initial work by @JustinDrake

I propose we discuss if we move forward with it, making it a post spec-freeze non-breaking change, i.e. fully backwards compatible for signatures during testnet. And keep this apart from the BLS spec-update itself, as a presentation thing.

\*: ok, test-vectors will adopt the spec status to alleviate the typing inconsistency highlighted in #964

### DIFF
--- a/specs/bls_signature.md
+++ b/specs/bls_signature.md
@@ -71,10 +71,10 @@ We require:
 G2_cofactor = 305502333931268344200999753193121504214466019254188142667664032982267604182971884026507427359259977847832272839041616661285803823378372096355777062779109
 q = 4002409555221667393417789825735904156556882819939007885332058136124031650490837864442687629129015664037894272559787
 
-def hash_to_G2(message_hash: Bytes32, domain: uint64) -> Tuple[uint384, uint384]:
+def hash_to_G2(message: bytes) -> Tuple[uint384, uint384]:
     # Initial candidate x coordinate
-    x_re = int.from_bytes(hash(message_hash + bytes8(domain) + b'\x01'), 'big')
-    x_im = int.from_bytes(hash(message_hash + bytes8(domain) + b'\x02'), 'big')
+    x_re = int.from_bytes(hash(message + b'\x01'), 'big')
+    x_im = int.from_bytes(hash(message + b'\x02'), 'big')
     x_coordinate = Fq2([x_re, x_im])  # x = x_re + i * x_im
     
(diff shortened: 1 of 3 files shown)


specs/bls_signature.md
```diff
@@ -130,17 +134,43 @@ g = Fq2([g_x, g_y])
...
+| `BLSDomain` | `Bytes8` | a BLS12-381 domain |
+| `BLSPubkey` | `Bytes48` | a BLS12-381 public key |
+| `BLSSignature` | `Bytes96` | a BLS12-381 signature |
+
+### `bls_verify`
+
+```python
+
+def bls_verify(pubkey: BLSPubkey, self_signed_object: Container, domain: BLSDomain) -> bool:
+    return bls.bls_verify(pubkey, domain + signed_root(self_signed_object), self_signed_object.signature)
```
vbuterin: If we go this way (which I'm really not sold on!! I think there's a lot of value in reducing the depth of the abstraction tower that readers have to waddle through, and the `verify(pub, msg, sig)` format has a lot going for it given it's simple and familiar, and also symmetry with `bls_verify_multiple` is valuable), can we at least call it `bls_verify_signed_container`? Would make it clearer to readers what's going on imo.

JustinDrake:
I suggest we stick with current interface and go ahead with the misc cleanups (e.g. add `BLSDomain` type).

protolambda:
@JustinDrake You can go ahead with the misc. cleanups in the helper cleanup PR. As expressed in the closed issue, I generally think the primary part of this PR should not be considered for spec-freeze. So also dislike labeling this as "spec-freeze", for just a `BLSDomain` alias. 

Instead I want to keep the discussion alive to some time after freeze, where there is less time pressure, and we can find some sweet spot (not necessarily with self-signed objects in interface, but possible) in separation of BLS and spec.

protolambda:
Unlabeling "spec freeze", in favor of #1246 which implements Domain and DomainType, but leaves interface as is. A simple and clean solution for freeze. We can experiment more with this PR later, or not, see what the cons/pros are here.

JustinDrake:
I'd say close this PR and revisit when we merge in the BLS standardisation post-freeze.

PR #1237, opened 2019-06-30
Helpers cleanup

JustinDrake:
Addresses #918. It includes:

1) Grouping helpers by category
2) Renamings (for consistency, length, clarity)
3) Add top-level function comments where missing, and make them consistent
4) Remove neglected and sometimes misleading terminology section (to be readded post-freeze)
5) Various other cosmetic cleanups


specs/core/0_beacon-chain.md
```diff
@@ -547,385 +534,469 @@ class BeaconState(Container):
...
+    end = (len(indices) * (index + 1)) // count
+    return [indices[shuffle_index(ValidatorIndex(i), len(indices), seed)] for i in range(start, end)]
 ```
 
-### `is_slashable_validator`
+#### `validate_indexed_attestation`
 
 ```python
-def is_slashable_validator(validator: Validator, epoch: Epoch) -> bool:
+def validate_indexed_attestation(state: BeaconState, indexed_attestation: IndexedAttestation) -> None:
```
CarlBeek: I think `verify` is a better term to be using here. I did a bunch of searching to try back up my intuition here, best I can come up with is this from an (adopted) IEEE standard:
* Validation: The assurance that a product, service, or system meets the needs of the customer and other identified stakeholders.
* Verification. The evaluation of whether or not a product, service, or system complies with a regulation, requirement, specification, or imposed condition.

specs/core/0_beacon-chain.md
```diff
@@ -547,385 +534,469 @@ class BeaconState(Container):
...
-    state.balances[index] += delta
+    return Slot(epoch * SLOTS_PER_EPOCH)
 ```
 
-### `decrease_balance`
+#### `delayed_activation_exit_epoch`
 
 ```python
-def decrease_balance(state: BeaconState, index: ValidatorIndex, delta: Gwei) -> None:
+def delayed_activation_exit_epoch(epoch: Epoch) -> Epoch:
```
CarlBeek: This sounds like the property of an object to me as it is a common noun. I think `get_delayed_activation_exit_epoch` is better.

specs/core/0_beacon-chain.md
```diff
@@ -547,385 +534,469 @@ class BeaconState(Container):
...
     """
-    Get active validator indices at ``epoch``.
+    Return the epoch number of ``slot``.
     """
-    return [ValidatorIndex(i) for i, v in enumerate(state.validators) if is_active_validator(v, epoch)]
+    return Epoch(slot // SLOTS_PER_EPOCH)
 ```
 
-### `increase_balance`
+#### `epoch_first_slot`
```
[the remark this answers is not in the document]
CarlBeek: `compute` works as an alternative to `get` if people are more amenable to it. `compute` is already used in `compute_committee`

specs/core/0_beacon-chain.md
```diff
@@ -171,7 +158,6 @@ The following values are (non-configurable) constants used throughout the specif
 | Name | Value |
 | - | - |
 | `FAR_FUTURE_EPOCH` | `Epoch(2**64 - 1)` |
-| `ZERO_HASH` | `Hash(b'\x00' * 32)` |
```
CarlBeek: I really link the removal of this, it's much cleaner and also serves as a reminder to readers that SSZ defaults to `0b0`.

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
JustinDrake: Danny is changing it to `epoch_start_slot`.

specs/core/0_beacon-chain.md
JustinDrake: Agreed it's more descriptive locally. I'm trying to reserve the `get_` prefix for state accessors.

specs/core/0_beacon-chain.md
JustinDrake: I think I want to do the same thing with slots and epochs, i.e. nuke `GENESIS_SLOT` in favour of `Slot()` (and same for epochs).

specs/core/0_beacon-chain.md
CarlBeek: I thought as much and I really like that idea. I still think we should find an alternate prefix to use within the `Misc` section. Ideas I've had thus far are:
* Compute
* Calculate
* Determine
* Obtain
* Acquire
* Compute

specs/core/0_beacon-chain.md
djrtwo: I think it reasonable to keep `GENESIS_SLOT` configurable. That's why I didn't put it in the "constants" section.

specs/core/0_beacon-chain.md
```diff
@@ -541,7 +541,7 @@ class BeaconState(Container):
 #### `integer_squareroot`
 
 ```python
-def integer_squareroot(n: uint64) -> uint64:
+def integer_squareroot(n: int) -> int:
```
JustinDrake: Using `uint64` allows us to remove the untriggerable `assert n >= 0`.

specs/core/0_beacon-chain.md
djrtwo: We were distinguishing `verify` and `validate` as -- 
* `verify` was returning a bool
* `validate` has no return value and instead throws if bad input

I'm fine leaving as is..

CarlBeek: Fine. As long as it is consistent (and I know the rule) then I'm happy. :)

specs/core/0_beacon-chain.md
djrtwo: It is a generic helper that has nothing to do with the SSZ types and works on all non-negative ints. Similar to `int_to_bytes` in typing.

I think we add in a comment about the value expected to be non-negative if we don't want th assert in there.

JustinDrake: I'm inclined to signal to implementers that we don't use unsigned integers anywhere. It's a nice conceptual simplification. Tempted to change the various `int` to `uint64`.

djrtwo: reasonable. handling it

protolambda: You could make it `uint` if you want to. Just include it in the spec builder in imports. But prefer not to break anything like that, with the amount of time for freeze. We can do such super minor presentation things post-freeze.

CarlBeek: The only place I can think of having `uint64`s fail is for BLS privkeys. Are there any other places you can think of @protolambda? I am not saying I don't favour the conservative route too.

JustinDrake:
I tried making `BLS_WITHDRAWAL_PREFIX` a `Bytes1` but broke something. (I think the spec builder is confused somehow.)

djrtwo:
I think we should leave as is instead of the "default" value here. It is a configurable constant that might be changed in the future (or in different deployments)

djrtwo:
I'd prefer to revert to my last commit and get this merged. We have other things to handle that are waiting on this PR

JustinDrake:
> I think we should leave as is instead of the "default" value here.

That's fine :) (Tried moving it to see if spec builder would be happier.) Making it `Bytes1()` is still probably the way forward.

specs/core/0_beacon-chain.md
djrtwo: went with `uint64`. It's our default int type and fine for these purposes.
We can change after if nedded as it is almost entirely cosmetic

specs/core/0_beacon-chain.md
djrtwo: `compute_activation_exit_epoch` ?

JustinDrake: Agreed that "delayed" doesn't add much value here

CarlBeek: I like that! I would like to change all of them in the `Misc` section, but let's merge this first.

PR #1238, opened 2019-06-30
Update test vector generators

protolambda:
Update the test generators to output the new tests.

- Fixed genesis tests:
  - one was testing balances. But that's not a condition in the beacon doc atm. And it erroneously passed, because we were not asserting the output. Commented it.
  - one was simply expecting the wrong outcome (name is very clear on it though). Fixed that
  - outcomes are checked now, and yielded data is properly labeled.
- Updated test format docs
- Write genesis test generator (small piece of code to make the genesis tests run)
- Update epoch-processing generator to include new test files.


JustinDrake:
Suggest merging after #1237 to minimise conflicts.

protolambda:
@JustinDrake Ok, please check if you covered all renamed variables then. I'll check if any processing function changed name, and update test docs accordingly.

protolambda:
Decided to disable transfer test vector output, but only for mainnet config. The output size of tests blows up due to the inefficiency in yaml encoding of full mainnet states. It got to ~20% of the total test vectors size. Looking for better alternatives in the future. Yaml is feels more and more as a legacy test format, when test-amount or state-size issues were not a thing. Something with compression, or another way to reduce duplicate data nodes would be great. (there are a lot of the same pre-states for the atomic tests).

The test-generators are otherwise all running and confirmed to produce vectors, so we are ready for freeze wrt testing. :tada:

specs/test_formats/epoch_processing/README.md
```diff
@@ -17,13 +17,17 @@ post: BeaconState      -- state after applying the epoch sub-transition.
...
 
 Sub-transitions:
 
-| *`sub-transition-name`* | *`processing call`*               |
-|-------------------------|-----------------------------------|
-| `crosslinks`            | `process_crosslinks(state)`       |
-| `registry_updates`      | `process_registry_updates(state)` |
+- `justification_and_finalization`
+- `crosslinks`
+- *`justification_and_finalization` - planned testing extension*
```
djrtwo:
```suggestion
- *`rewards_and_penalties` - planned testing extension*
```

JustinDrake:
@protolambda #1237 has been merged now :)

test_libs/pyspec/eth2spec/test/genesis/test_validity.py
```diff
@@ -51,13 +51,14 @@ def test_is_valid_genesis_state_true_more_balance(spec):
...
-    state = create_valid_beacon_state(spec)
-    state.validators[0].effective_balance = spec.MAX_EFFECTIVE_BALANCE - 1
-
-    yield from run_is_valid_genesis_state(spec, state, valid=False)
+# TODO: not part of the genesis function yet. Erroneously merged.
+# @with_phases(['phase0'])
+# @spectest_with_bls_switch
+# def test_is_valid_genesis_state_false_not_enough_balance(spec):
+#     state = create_valid_beacon_state(spec)
+#     state.validators[0].effective_balance = spec.MAX_EFFECTIVE_BALANCE - 1
```
djrtwo: Does this test pass?

protolambda: No, see https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#genesis-state
No balance condition is present. So there's no such thing as "not enough balance"

PR #1241, opened 2019-06-30
Renaming of functions in the `Misc` section

CarlBeek:
As per my discussion points in #1237, I think functions that return a value based on other values should have a name that indicates their function. For example I think `compute_shuffle_index` & `compute_epoch_start_slot` are much clearer than `shuffle_index` & `epoch_start_slot`. This implements that.


specs/core/0_beacon-chain.md
```diff
@@ -66,13 +66,13 @@
             - [`is_slashable_attestation_data`](#is_slashable_attestation_data)
             - [`is_valid_merkle_branch`](#is_valid_merkle_branch)
         - [Misc](#misc)
-            - [`shuffle_index`](#shuffle_index)
+            - [`compute_shuffle_index`](#compute_shuffle_index)
```
JustinDrake: What about `compute_shuffled_index` or `compute_index_shuffle`?

specs/core/0_beacon-chain.md
```diff
@@ -1677,8 +1677,9 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
     if pubkey not in validator_pubkeys:
         # Verify the deposit signature (proof of possession) for new validators.
         # Note: The deposit contract does not check signatures.
-        # Note: Deposits are valid across forks, hence the deposit domain is retrieved directly from `bls_domain`
-        if not bls_verify(pubkey, signing_root(deposit.data), deposit.data.signature, bls_domain(DOMAIN_DEPOSIT)):
+        # Note: Deposits are valid across forks. thus the deposit domain is retrieved directly from `compute_bls_domain`
```
JustinDrake: You changed `,` to `.`

specs/core/0_beacon-chain.md
```diff
@@ -1677,8 +1677,9 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
...
+        if not bls_verify(pubkey, signing_root(deposit.data),
+                          deposit.data.signature, compute_bls_domain(DOMAIN_DEPOSIT)):
```
JustinDrake: add `domain` variable and split into two

specs/core/0_beacon-chain.md
CarlBeek: I like `compute_shuffled_index`. :)

specs/core/0_beacon-chain.md
CarlBeek: My bad. I intended just to change 'hence' to 'thus' as is brings it under 120 chars agin.

specs/core/0_beacon-chain.md
CarlBeek: Will do

specs/core/0_beacon-chain.md
```diff
@@ -764,16 +731,49 @@ def compute_activation_exit_epoch(epoch: Epoch) -> Epoch:
...
 ```python
-def bls_domain(domain_type: uint64, fork_version: bytes=b'\x00' * 4) -> int:
+def compute_bls_domain(domain_type: uint64, fork_version: bytes=b'\x00' * 4) -> int:
     """
     Return the BLS domain for the ``domain_type`` and ``fork_version``.
     """
     return bytes_to_int(int_to_bytes(domain_type, length=4) + fork_version)
 ```
 
+#### `validate_indexed_attestation`
```
JustinDrake: Really `validate_indexed_attestation` should refactored to be `is_valid_indexed_attestation` and put under "Predicates".

specs/core/0_beacon-chain.md
```diff
@@ -644,6 +644,39 @@ def is_slashable_attestation_data(data_1: AttestationData, data_2: AttestationDa
     )
 ```
 
+#### `is_valid_indexed_attestation`
+
+```python
+def is_valid_indexed_attestation(state: BeaconState, indexed_attestation: IndexedAttestation) -> None:
```
JustinDrake: Now we need to make it return `-> bool`, and refactor to use `return`. My favourite is something like

```python
return (
    len(bit_1_indices) == 0 and
    len(bit_0_indices) + len(bit_1_indices) <= MAX_VALIDATORS_PER_COMMITTEE and
    len(set(bit_0_indices).intersection(bit_1_indices)) == 0 and
    bit_0_indices == sorted(bit_0_indices) and bit_1_indices == sorted(bit_1_indices) and
    bls_verify_multiple(
        pubkeys=[
            bls_aggregate_pubkeys([state.validators[i].pubkey for i in bit_0_indices]),
            bls_aggregate_pubkeys([state.validators[i].pubkey for i in bit_1_indices]),
        ],
        message_hashes=[
            hash_tree_root(AttestationDataAndCustodyBit(data=indexed_attestation.data, custody_bit=0b0)),
            hash_tree_root(AttestationDataAndCustodyBit(data=indexed_attestation.data, custody_bit=0b1)),
        ],
        signature=indexed_attestation.signature,
        domain=get_domain(state, DOMAIN_ATTESTATION, indexed_attestation.data.target.epoch),
    )
)
```

JustinDrake: Alternatively we could just put `return True` at the end and call it a day

CarlBeek: On it

djrtwo: We made this into a `validate` with separated statements to allow for ease in debugging. After surveying implementers, they greatly preferred _not _ grouping this all into one boolean.

Prefer sticking with `validate`.

specs/core/0_beacon-chain.md
djrtwo: disagree! see other comment

specs/core/0_beacon-chain.md
CarlBeek: Lol.

CarlBeek: What about Justin's many-asserts-followed-by-a-return option? It is less pretty from a code standpoint, but conceptually I think this function should be a predicate.

djrtwo: Not seeing the issue with a `validate_` function handling asserts.

The alternative is either (1) overly verbose but having if/return statements for each condition or (2) semantically  strange if we have asserts and then a return True.

djrtwo: We debated this a month ago and was the compromise we (including @vbuterin) agreed upon

specs/core/0_beacon-chain.md
```diff
@@ -698,56 +734,23 @@ def compute_committee(indices: Sequence[ValidatorIndex],
...
-    )
+    return [indices[compute_shuffled_index(ValidatorIndex(i), len(indices), seed)] for i in range(start, end)]
 ```
 
-#### `slot_to_epoch`
+#### `compute_slot_epoch`
 
 ```python
-def slot_to_epoch(slot: Slot) -> Epoch:
+def compute_slot_epoch(slot: Slot) -> Epoch:
```
djrtwo: Don't think this is very readable in the new form.

How do any of the following sound?
`compute_epoch_from_slot`, `compute_epoch_of_slot`, `compute_epoch(slot)`

specs/core/0_beacon-chain.md
djrtwo: cc: @JustinDrake

specs/core/0_beacon-chain.md
CarlBeek: Agreed. Ideally it would be compute_slot's_epoch, but since that's not happening, I like `compute_epoch_of_slot` best.

specs/core/0_beacon-chain.md
CarlBeek: > overly verbose but having if/return statements for each condition 

I considered this momentarily before proposing going back to the many asserts then return option.

My opinions aren't really that strong about this. If there was a discussion on this topic and the conclusion was a preference for the original, I will revert.

> 2) semantically strange if we have asserts and then a return True.

Agreed this is strange. It is a lot of ugliness just to so a function name is slightly more aligned with others in the spec.

specs/core/0_beacon-chain.md
djrtwo: works for me

CarlBeek: In a similar vein, I am changing `compute_epoch_start_slot` to `compute_start_slot_of_epoch`

specs/core/0_beacon-chain.md
JustinDrake: > We debated this a month ago and was the compromise we (including @vbuterin) agreed upon

Yes this was debated. I guess there are new circumstances now, given how `is_valid_indexed_attestation` makes all the helpers fall into nice clean categories.

djrtwo: I'm against collapsing into one boolean (esp as someone who writes and maintains a bunch of tests). 

The `if not (condition): return False` is verbose but seems like where the compromise currently lies

JustinDrake: I'm happy with the compromise :)

PR #1246, opened 2019-06-30
Fix bls domain type

protolambda:
cc @JustinDrake 

Executed idea to make BLS domain and domain typed,  byte4/8 types with their own alias.


specs/core/0_beacon-chain.md
```diff
@@ -147,6 +147,8 @@ We define the following Python custom types for type hinting and readability:
 | `ValidatorIndex` | `uint64` | a validator registry index |
 | `Gwei` | `uint64` | an amount in Gwei |
 | `Version` | `Bytes4` | a fork version number |
+| `DomainType` | `Bytes4` | a signature domain type |
+| `Domain` | `Bytes8` | a signature domain |
 | `Hash` | `Bytes32` | a hash |
```
JustinDrake: Maybe put `Hash` before `DomainType` so that the signature types are all together.

specs/core/0_beacon-chain.md
```diff
@@ -767,11 +771,11 @@ def compute_activation_exit_epoch(epoch: Epoch) -> Epoch:
 #### `bls_domain`
 
 ```python
-def bls_domain(domain_type: uint64, fork_version: bytes=b'\x00' * 4) -> int:
+def bls_domain(domain_type: DomainType, fork_version: bytes=b'\x00' * 4) -> Domain:
```
JustinDrake: I addressed this in another PR but might as well do it here to avoid the conflict. `fork_version: bytes=b'\x00' * 4` should be `fork_version: Version=Version()`

configs/constant_presets/minimal.yaml
```diff
@@ -118,9 +118,9 @@ MAX_TRANSFERS: 0
 
 # Signature domains
 # ---------------------------------------------------------------
-DOMAIN_BEACON_PROPOSER: 0
-DOMAIN_RANDAO: 1
-DOMAIN_ATTESTATION: 2
-DOMAIN_DEPOSIT: 3
-DOMAIN_VOLUNTARY_EXIT: 4
-DOMAIN_TRANSFER: 5
+DOMAIN_BEACON_PROPOSER: 0x00000000
```
JustinDrake: Strictly these should all be `DOMAIN_TYPE_...` (as opposed to `DOMAIN_...`)

protolambda: It would push quite some lines over the line limit. I think it's ok like this. The type tells enough.

specs/core/0_beacon-chain.md
protolambda: fixed

specs/core/0_beacon-chain.md
protolambda: fixed

prestonvanloon:
Why bytes4/8 instead of uint32/uint64?

JustinDrake:
> Why bytes4/8 instead of uint32/uint64?

We use `uint` when we want to do integer addition/multiplication. Here's we're just doing byte concatenation. Also want to use `uint64` exclusively for integers.

PR #1248, opened 2019-06-30
Fix genesis effective balance

JustinDrake:
The problem was that `effective_balance` was not set properly if a validator made multiple small deposits.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1139,7 +1139,9 @@ def initialize_beacon_state_from_eth1(eth1_block_hash: Hash,
 
     # Process activations
     for index, validator in enumerate(state.validators):
-        if state.balances[index] >= MAX_EFFECTIVE_BALANCE:
+        balance = state.balances[index]
+        validator.effective_balance = min(balance - balance % EFFECTIVE_BALANCE_INCREMENT, MAX_EFFECTIVE_BALANCE)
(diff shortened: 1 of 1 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1689,7 +1691,6 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
             activation_epoch=FAR_FUTURE_EPOCH,
             exit_epoch=FAR_FUTURE_EPOCH,
             withdrawable_epoch=FAR_FUTURE_EPOCH,
-            effective_balance=min(amount - amount % EFFECTIVE_BALANCE_INCREMENT, MAX_EFFECTIVE_BALANCE)
```
djrtwo: Should leave this line imo.
When a validator as an initial deposit, it makes sense to set their effective immediately.
Also keeps the tests passing.

protolambda: I think it's a nice property to know stable low-resolution balances (i.e. only changing stepwise once per epoch at most) for non-activated validators. It helps people monitoring for other reasons. Besides, de-activated validators get a effective balance too, and that is updated as well. It's much more consistent to keep it as-is.

JustinDrake: Added back (with the missing comma added). Not fully satisfied but can't think of a better solution right now.

PR #1249, opened 2019-06-30
Minor python style tweaks

CarlBeek:
This implements the following minor style tweaks that I'd rather have in before the freeze:

- Binary operators that are across multiple lines lead with the operator on the new line
- Consistency in how long function arguments are broken up


djrtwo:
lint error, otherwise looks good

PR #1252, opened 2019-07-01
remove bounds check in `compute_shuffled_index` and Fix ToC

hwwhww:
Bounds check enforced by `state.validators` SSZ `List` limit


hwwhww:
Ahh sorry I thought we set the `VALIDATOR_REGISTRY_LIMIT` to `2**40` *because of* the limitation of shuffling function. 😅

PR #1253, opened 2019-07-01
Update v-guide wrt recent spec updates

hwwhww:
Fix some known bugs:
1. Fix outdated `fork.slot` and `attestation_data.slot`
2. Use `Checkpoint`


cryptoneobancoin:
very important

PR #1257, opened 2019-07-01
scale inclusion reward properly

djrtwo:
Inclusion reward was too aggressive scaling down. This happened in a "clean up" PR a long time ago.


protolambda:
Do we need to adjust epoch slot down, to make 63/63 a thing? I.e. 100%

protolambda:
This 64/64 works too. But results in 1/64 minimum

djrtwo:
The `1/64` minimum is good because they _did_ manage to get their attestation on chain so some reward for inclusion. imo

PR #1184, opened 2019-06-16
Work-in-progress on SSZ partials

vbuterin:
Now updated to the latest spec.

PR #1261, opened 2019-07-01
Partials v2

protolambda:
Credits to @vbuterin for exploring partials early on. I rebased his branch on the dev branch, since the older PR (#1184) was based off of an intermediate SSZ implementation, with missing features/bugfixes/types/consistency. Now we are building towards fully SSZ-compatible partials again, incl. use with spec state transition functions.

Status:
- initial partials code by @vbuterin refactored (WIP) to be more class-based, to fit seamlessly in with the other SSZ types.
- logistical functions mostly updated to use new types
- need to split some of the helper functions, move their functionality into the partial types. Problem was that every helper func needed a typ param, and did a big switch-like statement on SSZ types. Instead we can generalize this nicely, make type information easily accessible, and move it into the right part of the type inheritance tree.

Summary of design:

*note the difference between abstraction levels in the inheritance structure*:
- object: instance of a `SSZValue`
- ssz class: class for an object, instance of `SSZType`
- ssz class type: form of ssz type, subclass of `SSZType`. E.g. some ssz classes are a BasicType, some not. Some are `ElementsType`: they have an `elem_type` and `length` property.

Due to limitations in python, `Container` has no `ContainerType` metaclass. This means that we use the `__annotations__` attribute to get the fields and their types from a `Container` subclass. These are accessible with `classmethod`-annotated functions, to allow for `typ.get_fields()`: retrieve class information without creating an object (but it does create the class)

Partials can be modeled as subclasses of the ssz classes. With a `Partials` class mixed in to provide partials functionality (pre-refactor). This enables a partial of a given type to be passed into any function that accepts that regular type.
In this refactor, I'm splitting the Partials class, so that the different variations can subclass ssz-classes, which lets us get rid of the `typ` argument, and fetch type data like fields, length, element type etc. from the super type. Along with the values in the super class. 
Get/Set-item/attribute is overridden to inject the partials logic: retrieve elements/fields from included data, or error when the data is unknown (i.e. not included in the partial)

I'll post a visual overview of the OOP structure after refactoring some more.

I think the final idea is to support something like:

```python
state_partial = Partial[BeaconState](data objects, root index)
v = state_partial.validators[my_index]
and...
get_important_value(state_partial)# and any other spec function that accepts BeaconState
etc...
```
And then we can start building POC light clients on top of SSZ.

**this is still very experimental, work in progress.**

#### Misc ideas

`SSZValue` (and any subclass) as path root, give the class a `/` operator overload, and then output a `SSZPath`, carrying the generalized index and the type of the path node.
Then also define `/` for the `SSZPath`, and we can use paths to abstract away any light client logic.
*Basic* example:
```python
path = BeaconState / "validators" / 123 / "slashed"
# see if slashed, in generic way. Could be proving more at the same time.
value = some_partial.proof(path.gen_index, False)
```


test_libs/pyspec/eth2spec/utils/ssz/test_ssz_partials.py
```diff
@@ -0,0 +1,112 @@
...
+    Person(is_male=True, age=uint64(51), name=Bytes[32](b"Danny")),
+    Person(is_male=True, age=uint64(53), name=Bytes[32](b"Evan")),
+    Person(is_male=False, age=uint64(55), name=Bytes[32](b"Fae")),
+    Person(is_male=False, age=uint64(57), name=Bytes[32](b"Ginny")),
+    Person(is_male=False, age=uint64(59), name=Bytes[32](b"Heather")),
+    Person(is_male=False, age=uint64(61), name=Bytes[32](b"Ingrid")),
+    Person(is_male=False, age=uint64(63), name=Bytes[32](b"Kane")),
+)
+
+city = City(coords=Vector[uint64, 2](uint64(45), uint64(90)), people=people)
```
vbuterin: Can we do `city = City(coords=[45, 90], people=people)` now due to auto-type-coercion?

djrtwo:
What's the status here @protolambda ?

protolambda:
Put everything together, incl. multi-proofs code from latest spec, and fixed types/usage, along with improvements in type structure (ComplexValue/Type, and length<>limit separation). Plan is now to implement a `Partial[...InnerType...]` notation, to have static type-checked partials available.
E.g.:
```
# subset of the fields (*random* example)
class LightState(Partial[BeaconState]):
    eth1_data: Eth1Data
    finalized_checkpoint: ExamplePartCheckPoint
    some_list: Partial[List[128, Foo], [3,4,100]]

class ExamplePartCheckPoint(Partial[Checkpoint]):
   root: Hash
   # epoch is ignored but proof data is supported by Partial wrapper
```

And then defer proof construction to when we need it, instead of tracking the branches and leaf chunks inside the Parital instance. This should be cheaper and more realistic. And then on Partial instantiation, we can parse the multi-proof, verify correctness (both merkleization as typing), and have the thing available wherever we need it. (As the wrapper produces a subclass instance of the wrapped type).

protolambda:
Update: there is a lot of trial and error in this PR, along with the previous partials PR. But made some great progress in a long IRL discussion with Matt to decide on a new clean design, which also works as a pattern for use in an ewasm setting. Me and Matt went over this in the python call, and are planning to put out a new updated doc some time this week, after getting feedback from Vitalik here in Berlin, and more iteration on this new approach.

protolambda:
See https://github.com/ethereum/eth2.0-specs/issues/1303#issuecomment-526366715

Current thoughts are that a multi-proof behind an interface is sufficient (and efficient for our needs) for a partial, and that we may want to standardize offsets or other data to encode the structure of the proof with the packed data efficiently. Leaving this for now though, as phase 0 interop work is more pressing. Let me know if someone likes to pick this up in the meantime.

protolambda:
The latest idea is to separate and isolate complexity better, and layer it as:
- proof-backings, support multiple options
- multi-proofs: interface to work with bare proof backings
- partials: typing around the multi-proof interface

This PR is largely outdated, as it still mixes these layers.

During devcon discussions about ssz/merkleization/standards it became clear that SSZ is outgrowing this specs repo, and that partials and other advanced features need a place to live with an "experimental" standardization stage. I started a SSZ specs repo draft [here](https://github.com/protolambda/eth2.0-ssz/), and plan to discuss on the implementers call of 24 Oct.

Closing this PR after the call, and will keep track of the SSZ repo transition progress in some new issue/PR.

protolambda:
See https://github.com/protolambda/eth-merkle-trees/blob/master/typing_partials.md for a write up of how typing over binary tree backings would work. I've [an open python project in progress](https://github.com/protolambda/pymerkles) that implements this, but will keep it separate from the spec for now.

PR #1263, opened 2019-07-02
Changes `Bytes48` to `BLSPubkey`

decanus:


djrtwo:
will make a `v08x` branch for minor fixes like this. hold off on merge for a bit

PR #1264, opened 2019-07-02
Fix #1262

### LINKED ISSUE #1262: Check attestation bit list length against committee size
In `get_attesting_indices`, we used to `verify_bitfield` to ensure the attestation bitfield length was the size the committee.
Now we've moved the verification of _a_ length to the ssz deserialization, but removed the committee size check.

JustinDrake:
Check length of (aggregation and custody) bits against the committee size.

Todo: Add tests

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1044,6 +1044,7 @@ def get_attesting_indices(state: BeaconState,
     Return the set of attesting indices corresponding to ``data`` and ``bits``.
     """
     committee = get_crosslink_committee(state, data.target.epoch, data.crosslink.shard)
+    assert len(committee) == len(bits)
     return set(index for i, index in enumerate(committee) if bits[i])
 ```
 


JustinDrake:
>  djrtwo changed the base branch from `dev` to `v08x`

Is this a bug to fix in `v08x`? My assumption was that, for the duration of the freeze, we would only push non-substantive changes as well as bugs that prevent the proper formation of cross-client testnets. This bug arguably does not impede cross-client testnets.

JustinDrake:
> My inclination is to put the check in `process_attestation` and to check explicitly against both aggregation and custody bits

Yes, that is clearer :)

djrtwo:
cleaning this up

dankrad:
> My inclination is to put the check in `process_attestation` and to check explicitly against both aggregation and custody bits.

Just independently came across this and put this fix in #1305  in `process_attestation`. Should we move the implementation here: https://github.com/ethereum/eth2.0-specs/pull/1305/files#diff-51a43328a58414e132a744f3771f018c

djrtwo:
yes, That's where the fix should go. Have a separate PR with tests coming

djrtwo:
close in favor of #1317

PR #1267, opened 2019-07-03
Define Bitlist/Bitvector serialization using bytes, not bigints

dankrad:
This does not change any logic and the serialization and Merkleization is exactly as before, but now defined using byte operations instead of bigints.

### DIFF
--- a/test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
+++ b/test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
@@ -41,15 +41,13 @@ def serialize(obj: SSZValue):
     if isinstance(obj, BasicValue):
         return serialize_basic(obj)
     elif isinstance(obj, Bitvector):
-        as_bytearray = [0] * ((len(obj) + 7) // 8)
-        for i in range(len(obj)):
-            as_bytearray[i // 8] |= obj[i] << (i % 8)
-        return bytes(as_bytearray)
+        return obj.as_bytes()
     elif isinstance(obj, Bitlist):
-        as_bytearray = [0] * (len(obj) // 8 + 1)
-        for i in range(len(obj)):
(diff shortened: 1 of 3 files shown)


test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
```diff
@@ -92,12 +97,11 @@ def encode_series(values: Series):
 def pack(values: Series):
     if isinstance(values, bytes):  # Bytes and BytesN are already packed
         return values
-    elif isinstance(values, Bitvector):
-        as_integer = sum([values[i] << i for i in range(len(values))])
-        return as_integer.to_bytes((values.length + 7) // 8, "little")
-    elif isinstance(values, Bitlist):
-        as_integer = sum([values[i] << i for i in range(len(values))])
-        return as_integer.to_bytes((values.length + 7) // 8, "little")
+    elif isinstance(values, Bitvector) or isinstance(values, Bitlist):
```
protolambda:
```suggestion
    elif isinstance(values, (Bitvector, Bitlist)):
```

And can you confirm that we are not prefixing the bit here, since we already mix in the length for hashing?

dankrad: Correct, the Merkleization does not mix in the bit delimiter.

hwwhww:
@dankrad 
also need to update the SSZ spec doc as well. :)

protolambda:
It looked like this PR was stuck, so I updated it.
- reduced the code duplication with an `.as_bytes()` in the pyspec implementation
- updated the ssz spec doc to also directly go from bits to bytes (instead of intermediate as_integer)

@dankrad a review back is welcome :)

protolambda:
Also: the merkleization part is cleaned up in #1292, along with other merkleization wording/implementation fixes

PR #1270, opened 2019-07-04
Some cosmetic/type hinting/configuration-relocation

hwwhww:
### Configuration Relocation
1. `PERSISTENT_COMMITTEE_PERIOD` has been defined in phase 0, so remove it from phase 1 spec
2. When reading `0_beacon_chain.md`, it's odd to see "`MIN_ATTESTATION_INCLUSION_DELAY` is one slot and the duration is 6 seconds" out of nowhere. This PR moves `SECONDS_PER_SLOT` back to `0_beacon_chain` spec.

### Phase 1
1. Update some missing type hinting of phase 1
2. Add phase 1 domain to constant_presets files


specs/core/0_beacon-chain.md
```diff
@@ -207,6 +207,7 @@ The following values are (non-configurable) constants used throughout the specif
 
 | Name | Value | Unit | Duration |
 | - | - | :-: | :-: |
+| `SECONDS_PER_SLOT` | `6` | seconds | 6 seconds |
```
JustinDrake: `SECONDS_PER_SLOT` is not used by the state transition function

hwwhww: @JustinDrake The reason is about spec file dependencies.
* `0_beacon_chain.md` is the kind of the "main spec file" that `0_fork-choice.md` imports.
* However, in the duration note: "`SLOTS_PER_EPOCH`: 64 slots = 6.4 minutes" implies that the fact `SECONDS_PER_SLOT` is `6`.

Do you think it makes sense?

djrtwo: I agree. 6 seconds and a slot time are implied all over this spec and it seems incomplete to hide it in another file. The validator doc also uses it and it seems strange to have it inherit the var from fork choice.

PR #1271, opened 2019-07-04
Cosmetic SSZ spec + add `pack()` in `Bitvector` and `Bitlist` Merkleization

hwwhww:
1. Clarify optional parameter `pad_for=1`.
2. To confirm with @dankrad @protolambda, it looks like there are some missing `pack()`s in `Bitvector` and `Bitlist` Merkleization? related to #1267
3. Some cosmetic.


protolambda:
> To confirm with @dankrad @protolambda, it looks like there are some missing pack()s in Bitvector and Bitlist Merkleization?

Yes. Checked the exec-spec ssz-implementation again too. "Bitlist" and "Bitvector" are recognized as some kind of an "elements" container, with "Bit" as element-type. Hence detected as bottom-type, thus packed and chunkified. So all good there.

PR #1276, opened 2019-07-06
Refactoring shard committees and attestations

vbuterin:
* Removed the attestation list, replaced with one attestation
* Reformed how committees are computed, to a structure where a 64-sized committee (due to be 128 if we increase shard slots to 2 per one beacon slot) is created per epoch, and then a proposer from that committee is chosen per slot, and the attesters of a slot are the last N (here N=32) proposers
* Added shard slot <-> beacon slot decoupling and more precise rules for which beacon slot to use when processing a shard slot

I welcome other concrete proposals; there are some difficult tradeoffs here to make. As I can see, the goals are:

* Efficiency: reduce the number of validators in a persistent committee that have to verify any given block
* Light client friendliness: give light clients the ability to verify any block with at least some security guarantee
* Unpredictability of who proposes blocks
* Probability of being selected proportional to balance
* You get kicked out if slashed

Does NOT include:

* The beacon chain committing to the persistent committees

### DIFF
--- a/specs/core/1_shard-data-chains.md
+++ b/specs/core/1_shard-data-chains.md
@@ -47,7 +47,7 @@ This document describes the shard data layer and the shard fork choice rule in P
 | - | - |
 | `BYTES_PER_SHARD_BLOCK_BODY` | `2**14` (= 16,384) |
 | `MAX_SHARD_ATTESTIONS` | `2**4` (= 16) |
-| `SHARD_SLOTS_PER_BEACON_SLOT` | `2**0` (= 1) |
+| `SHARD_SLOTS_PER_BEACON_SLOT` | `2**1` (= 2) |
 | `SHARD_SLOT_COMMITTEE_SIZE` | `2**5` (= 32) |
 
 ### Initial values
@@ -92,7 +92,7 @@ class ShardBlockBody(Container):
 
 ```python
 class ShardBlock(Container):
-    slot: Slot
+    slot: ShardSlot
     shard: Shard
     beacon_chain_root: Bytes32
     parent_root: Bytes32
@@ -106,7 +106,7 @@ class ShardBlock(Container):
 
 ```python
 class ShardBlockHeader(Container):
-    slot: Slot
+    slot: ShardSlot
     shard: Shard
     beacon_chain_root: Bytes32
     parent_root: Bytes32
@@ -117,6 +117,13 @@ class ShardBlockHeader(Container):
 ```
 
 ## Helper functions
+
+### `compute_epoch_of_shard_slot`
+
+```python
+def compute_epoch_of_shard_slot(slot: ShardSlot) -> Epoch:
+    return slot // SHARD_SLOTS_PER_BEACON_SLOT // SLOTS_PER_EPOCH
+```
 
 ### `get_period_committee`
 
@@ -192,8 +199,8 @@ def get_shard_epoch_committee(state: BeaconState,
 ```python
 def get_shard_block_proposer_index(state: BeaconState,
                                    shard: Shard,
-                                   slot: Slot) -> Optional[ValidatorIndex]:
-    epoch_committee = get_shard_epoch_committee(state, shard, compute_epoch_of_slot(slot))
+                                   slot: ShardSlot) -> Optional[ValidatorIndex]:
+    epoch_committee = get_shard_epoch_committee(state, shard, compute_epoch_of_shard_slot(slot))
     if len(epoch_committee) == 0:
         return None
     else:
@@ -207,10 +214,10 @@ def get_shard_block_attester_committee(state: BeaconState,
                                        shard: Shard,
                                        slot: Slot) -> Sequence[Optional[ValidatorIndex]]:
     committee_size = min(
-        len(get_shard_epoch_committee(state, shard, compute_epoch_of_slot(slot))),
+        len(get_shard_epoch_committee(state, shard, compute_epoch_of_shard_slot(slot))),
         SHARD_SLOT_COMMITTEE_SIZE,
     )
-    return [get_shard_block_proposer_index(state, shard, Slot(slot - i)) for i in range(committee_size)]
+    return [get_shard_block_proposer_index(state, shard, ShardSlot(slot - i)) for i in range(committee_size)]
 ```
 
 ### `get_shard_header`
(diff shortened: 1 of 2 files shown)


specs/core/1_shard-data-chains.md
```diff
@@ -157,57 +161,73 @@ def get_switchover_epoch(state: BeaconState, epoch: Epoch, index: ValidatorIndex
...
 ```
 
-### `get_persistent_committee`
+### `get_shard_epoch_committee`
 
 ```python
-def get_persistent_committee(state: BeaconState,
-                             shard: Shard,
-                             slot: Slot) -> Sequence[ValidatorIndex]:
+def get_shard_epoch_committee(state: BeaconState,
```
djrtwo: I think it makes sense to have two functions `get_persistent_committee` as well as `get_shard_epoch_committee`. The latter calls the former. keeping `get_persistent_committee` exposed will likely be of use to light clients.

specs/core/1_shard-data-chains.md
```diff
@@ -157,57 +161,73 @@ def get_switchover_epoch(state: BeaconState, epoch: Epoch, index: ValidatorIndex
...
-    # No block can be proposed if no validator is active
-    return None
+def get_shard_block_proposer_index(state: BeaconState,
+                                   shard: Shard,
+                                   slot: ShardSlot) -> Optional[ValidatorIndex]:
+    epoch_committee = get_shard_epoch_committee(state, shard, compute_epoch_of_shard_slot(slot))
+    if len(epoch_committee) == 0:
+        return None
+    else:
+        return epoch_committee[slot % len(epoch_committee)]
```
djrtwo: So predictable with min 1 epoch lookahead?

specs/core/1_shard-data-chains.md
```diff
@@ -157,57 +161,73 @@ def get_switchover_epoch(state: BeaconState, epoch: Epoch, index: ValidatorIndex
...
+```
+
+### `get_shard_block_attester_committee`
+
+```python
+def get_shard_block_attester_committee(state: BeaconState,
+                                       shard: Shard,
+                                       slot: ShardSlot) -> Sequence[Optional[ValidatorIndex]]:
+    committee_size = min(
```
djrtwo: Comment saying the previous `committee_size` proposers attest to `slot` (inclusive of current slot) would be valuable

specs/core/1_shard-data-chains.md
```diff
@@ -221,7 +241,7 @@ def get_shard_header(block: ShardBlock) -> ShardBlockHeader:
         parent_root=block.parent_root,
         body_root=hash_tree_root(block.body),
         state_root=block.state_root,
-        attestations=block.attestations,
+        attester_bitfield=block.attester_bitfield,
```
djrtwo: Going to need to add the `attester_signature` field here

specs/core/1_shard-data-chains.md
```diff
@@ -230,20 +250,16 @@ def get_shard_header(block: ShardBlock) -> ShardBlockHeader:
...
-        domain=get_domain(state, DOMAIN_SHARD_ATTESTER, compute_epoch_of_slot(data.slot))
+                                       block: ShardBlock) -> bool:
+    attester_committee = get_shard_block_attester_committee(state, block.shard, block.slot)
+    attesters = [v for i, v in enumerate(attester_committee) if i is not None and block.attester_bitfield[i]]
+    if get_shard_block_proposer_index(state, block.shard, block.slot) not in attesters:
+        return False
+    return bls_verify(
+        pubkey=bls_aggregate_pubkeys([state.validators[v].pubkey for v in attesters]),
+        message_hash=signing_root(block),
+        signature=block.signature,
```
djrtwo: This needs to change to the `block.attester_signature`

specs/core/1_shard-data-chains.md
```diff
@@ -230,20 +250,16 @@ def get_shard_header(block: ShardBlock) -> ShardBlockHeader:
...
-            pubkeys.append(validator.pubkey)
-    assert bls_verify(
-        pubkey=bls_aggregate_pubkeys(pubkeys),
-        message_hash=data.shard_block_root,
-        signature=attestation.aggregate_signature,
-        domain=get_domain(state, DOMAIN_SHARD_ATTESTER, compute_epoch_of_slot(data.slot))
+                                       block: ShardBlock) -> bool:
+    attester_committee = get_shard_block_attester_committee(state, block.shard, block.slot)
+    attesters = [v for i, v in enumerate(attester_committee) if i is not None and block.attester_bitfield[i]]
+    if get_shard_block_proposer_index(state, block.shard, block.slot) not in attesters:
```
djrtwo: I think the logic is off here. The attesters included in block at `N` would be those that attested to block at `N-1`. Only the proposer knows this block to sign when it's created. There is also currently a circularity issue with the signature being of `block` but the block including who signed it in the bitfield.

I think we need two signatures -- `block.signature` (proposer signature) and `block.attester_signature` (aggregate attestation signature of slot `block.slot - 1`

djrtwo:
In general, this looks like a good path at least as a start

specs/core/1_shard-data-chains.md
```diff
@@ -87,64 +98,57 @@ class ShardBlockBody(Container):
...
 class ShardBlock(Container):
-    slot: Slot
+    slot: ShardSlot
     shard: Shard
     beacon_chain_root: Bytes32
     parent_root: Bytes32
     data: ShardBlockBody
     state_root: Bytes32
-    attestations: List[ShardAttestation, PLACEHOLDER]
+    attester_bitfield: Bitvector[SHARD_SLOT_COMMITTEE_SIZE]
```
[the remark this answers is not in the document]
vbuterin: Ah I'm deliberately merging the two. The proposer is just one of the attesters.

specs/core/1_shard-data-chains.md
vbuterin: Ah, I'm doing things differently here. The attestations included in block N are attestations _of_ block N. So a block would be broadcasted with just one signature (the proposer) and others would add their signatures in.

I'm happy to go back to the other way, though it is technically ~96 bytes and 1 pairing less efficient :)

specs/core/1_shard-data-chains.md
vbuterin: Yep!

djrtwo:
Seems like this is all being handled in pieces in other PRs.
Can we close @vbuterin ?

PR #1278, opened 2019-07-08
Alternative crosslink data construction, take 1

vbuterin:
Properties:

* Optimal data packing
* Fixed depth
* All data is encoded; makes fraud proofs not-too-difficult

Weaknesses:

* Uses SSZ serialization.
* Positions of blocks dynamic, so accesses require an extra Merkle branch

### DIFF
--- a/specs/core/1_shard-data-chains.md
+++ b/specs/core/1_shard-data-chains.md
@@ -253,7 +253,7 @@ def verify_shard_attestation_signature(state: BeaconState,
 ### `flatten_block`
 
 ```python
-def flatten_block(block: ShardBlock) -> Bytes:
+def flatten_block(block: ShardBlock) -> bytes:
     return zpad(serialize(get_shard_header(block)), BYTES_PER_SHARD_BLOCK_HEADER) + serialize(block.body)
 ```
 
--- a/specs/core/1_shard-data-chains.md
+++ b/specs/core/1_shard-data-chains.md
@@ -27,6 +27,8 @@
         - [`get_shard_proposer_index`](#get_shard_proposer_index)
         - [`get_shard_header`](#get_shard_header)
         - [`verify_shard_attestation_signature`](#verify_shard_attestation_signature)
+        - [`flatten_block`](#flatten_block)
+        - [`compute_crosslink_data`](#compute_crosslink_data)
         - [`compute_crosslink_data_root`](#compute_crosslink_data_root)
     - [Object validity](#object-validity)
         - [Shard blocks](#shard-blocks)
@@ -46,6 +48,7 @@ This document describes the shard data layer and the shard fork choice rule in P
 
 | Name | Value |
 | - | - |
+| `BYTES_PER_SHARD_BLOCK_HEADER` | `2**9` (= 512) |
 | `BYTES_PER_SHARD_BLOCK_BODY` | `2**14` (= 16,384) |
 | `MAX_SHARD_ATTESTIONS` | `2**4` (= 16) |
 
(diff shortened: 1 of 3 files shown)


specs/core/1_shard-data-chains.md
```diff
@@ -76,7 +79,7 @@ The following types are defined, mapping into `DomainType` (little endian):
 
 | Name | Value |
 | - | - |
-| `PLACEHOLDER` | `2**32` |
+| `PLACEHOLDER` | `2**3` |
```
hwwhww: I changed it to `2**3` for testing.

specs/core/1_shard-data-chains.md
```diff
@@ -107,7 +110,7 @@ class ShardBlock(Container):
     shard: Shard
     beacon_chain_root: Bytes32
     parent_root: Bytes32
-    data: ShardBlockBody
+    body: ShardBlockBody
```
hwwhww: Also hotfix in #1298.

specs/core/1_shard-data-chains.md
```diff
@@ -247,36 +250,37 @@ def verify_shard_attestation_signature(state: BeaconState,
...
+    return zpad(serialize(get_shard_header(block)), BYTES_PER_SHARD_BLOCK_HEADER) + serialize(block.body)
+```
+
+### `compute_crosslink_data`
+
+```python
+def compute_crosslink_data(blocks: Sequence[ShardBlock]) -> Bytes32:
+    """
+    Flattens a series of blocks. Designed to be equivalent to SSZ serializing a list of
+    flattened blocks with one empty block at the end as a termination marker.
```
hwwhww: Could you elaborate what do you mean by "one empty block at the end" here?

specs/core/1_shard-data-chains.md
```diff
@@ -247,36 +250,37 @@ def verify_shard_attestation_signature(state: BeaconState,
     )
 ```
 
+### `flatten_block`
+
+```python
+def flatten_block(block: ShardBlock) -> Bytes:
+    return zpad(serialize(get_shard_header(block)), BYTES_PER_SHARD_BLOCK_HEADER) + serialize(block.body)
```
hwwhww: I added `serialize()` here, correct?

specs/core/1_shard-data-chains.md
hwwhww: Could you elaborate what's the main advantage of this one over just using SSZ `serialize` like:
```python
class CrosslinkingBlocks:
    data: List[ShardBlock, SHARD_SLOTS_PER_EPOCH * MAX_EPOCHS_PER_CROSSLINK]

assert len(blocks) <= SHARD_SLOTS_PER_EPOCH * MAX_EPOCHS_PER_CROSSLINK
crosslinking_blocks = CrosslinkingBlocks(data=*(blocks))
return serialize(crosslinking_blocks)
```
?

specs/core/1_shard-data-chains.md
```diff
@@ -247,36 +250,37 @@ def verify_shard_attestation_signature(state: BeaconState,
...
+```python
+def compute_crosslink_data(blocks: Sequence[ShardBlock]) -> bytes:
+    """
+    Flattens a series of blocks. Designed to be equivalent to SSZ serializing a list of
+    flattened blocks with one empty block at the end as a termination marker.
+    """
+    positions = [4 * len(blocks)]
+    bodies = []
+    for block in blocks:
+        bodies.append(flatten_block(block))
```
djrtwo: Doesn't flattening the block here reduce our ability to construct proofs about components of a block through the crosslink root? Still possible but not as clean due to flattening to bytes

dankrad:
> Uses SSZ serialization.

I wouldn't see that as a weakness, if it is the right tool for the job, I think we should be able to bring it into consensus.

djrtwo:
> I wouldn't see that as a weakness, if it is the right tool for the job, I think we should be able to bring it into consensus.

agreed.

djrtwo:
@vbuterin This can be closed as it was superseded by #1294, right?

djrtwo:
Closing -- superseded by #1294

PR #1279, opened 2019-07-08
Rename get_hash_tree_root to get_deposit_root

JustinDrake:
Improve naming

### DIFF
--- a/deposit_contract/contracts/validator_registration.v.py
+++ b/deposit_contract/contracts/validator_registration.v.py
@@ -42,7 +42,7 @@ def to_little_endian_64(value: uint256) -> bytes[8]:
 
 @public
 @constant
-def get_hash_tree_root() -> bytes32:
+def get_deposit_root() -> bytes32:
     zero_bytes32: bytes32 = 0x0000000000000000000000000000000000000000000000000000000000000000
     node: bytes32 = zero_bytes32
     size: uint256 = self.deposit_count


djrtwo:
The deposit contract is already under bytecode formalization and I believe changing a name changes it's internal identified representation. Would rather just close this

JustinDrake:
> I believe changing a name changes it's internal identified representation

Do you mean the bytecode changes?

ralexstokes:
the function selector bytes will change if the name changes -- function is identified by first four bytes of `keccack(fn-sig)` where the `fn-sig` is the `function-name(function-args*)`... vyper inherited this dispatch machinery from solidity. it shouldn't affect the semantics of the bytecode but _does_ constitute a bytecode change...

JustinDrake:
I guess we can put in this cleanup if we do end up changing the bytecode for whatever reason.

JustinDrake:
Reopening to bundle with #1341 :joy:

daejunpark:
In terms of the bytecode formal verification we are currently working on, the function name change wouldn't be a big problem, and I think we can manage to adjust our work to such a simple change.

But this change will result in having two functions with similar names and interfaces: `get_deposit_count` and `get_deposit_root`, which may lead to cases where clients make mistakes confusing their name.

JustinDrake:
> the function name change wouldn't be a big problem, and I think we can manage to adjust our work to such a simple change

❤️

> this change will result in having two functions with similar names and interfaces: `get_deposit_count` and `get_deposit_root`

The [`Eth1Data` object](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#eth1data) also has two corresponding properties with the same `deposit_` prefix:

```python
class Eth1Data(Container):
    deposit_root: Hash  # Populated with `get_deposit_root` from the deposit contract
    deposit_count: uint64  # Populated with `get_deposit_count` from the deposit contract
    block_hash: Hash
```

JustinDrake:
Closing in favour of #1362

PR #1281, opened 2019-07-10
Libp2p Standardization Update

AgeManning:
# Overview

Currently there are a few suggestions about how Ethereum 2 clients should implement the RPC in the context of libp2p as well as various encoding/decoding and formats. Specifically,  #1158 proposes segregating protocol Ids for each RPC request. 

Furthermore, the current RPC specification has implementation issues (some of which are discussed in #1158) and does not primarily focus on libp2p implementations. 

This PR consolidates #1158, #1100 and the RPC specification into a libp2p-specific context into `libp2p_standardization.md`. This document is aimed for client implementations to discuss and come to consensus on the final libp2p framework used in all implementations.

I'm also happy to help maintain this iterative specification.


zah:
Thank you for continuing the work on this, Age.

One small nitpick: I see the use of varint length prefixes as something specific to the negotiated encoding. The SSZ encoding requires it, but other serialization schemes may have a built-in record length information. I would create a dedicated section within the document detailing the SSZ encoding with examples of exchanged messages.

AgeManning:
@zah - Thanks!
I agree. I'll add a dedicated example for SSZ encoding. 

I've extracted the varint from the encoding as I envisaged it as part of the libp2p framework. By this, I mean that a requester always reads the varint, regardless of encoding, then proceeds to read the exact amount of bytes specified by the length. Once the bytes are read, the stream is closed and the arbitrary encoding is used to decode the received content. In retrospect, I think this train of thought is primarily based on how my implementation works, and if an encoding already defines a length-prefix, it may be duplicated in this scenario. 

In the case of SSZ, SSZ-encoded objects do not have an overall length-prefix. Given the initial part of an SSZ-encoded object, a receiver does not have a way of telling how many more bytes will be sent by the sender. If we know the schema, and the object only has fixed-length fields, we could know the length of the payload, but this is in general not true. By length-prefixing the payload, a receiver can always know how many bytes need to read, and can reject the stream if the length is too large.

To be clear, is your suggestion, that we only allow encodings that apply some length prefix, and for each encoding implement in the libp2p logic to read that length and control the stream based on that encoding?

specs/networking/libp2p-standardization.md
```diff
@@ -141,18 +153,327 @@ The `data` field of a Gossipsub `Message` is an SSZ-encoded object. For the `bea
...
+
+This protocol segregation allows libp2p `multistream-select` to handle the RPC-type, version and encoding negotiation before establishing the underlying substreams.
+
+
+#### Requests and Responses
+
+##### Request
+
+A request is formed by initiating a connection with the protocol id matching
+the desired request type, encoding and version. Once a successful substream is
```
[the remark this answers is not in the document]
AgeManning: Yes. I should be more careful with my terminology. In this case I'm referring to a stream (multiplexed on a libp2p connection) that has been negotiated with `multistream-select` for a specific protocol. In this case, the protocol id associated with the request.

specs/networking/libp2p-standardization.md
```diff
@@ -141,18 +153,327 @@ The `data` field of a Gossipsub `Message` is an SSZ-encoded object. For the `bea
...
++--------+--------+--------+--------+
+| unsigned_varint | encoded_request |
++--------+--------+--------+--------+
+```
+
+Once a stream is negotiated, the requester SHOULD send the request within **3
+seconds**.
+
+The requester SHOULD then wait for a response on the negotiated stream for at
+most **10 seconds**, before dropping the stream. A requester SHOULD decode the
```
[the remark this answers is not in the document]
AgeManning: Good question! I had intended on the requester to drop their end of the stream, but in this case `Stream.reset()` would be more reasonable. I'm working with my knowledge of rust-libp2p which I don't recall seeing the `reset()` functionality as is likely in go. 
Let me dig a little deeper, perhaps we should update this to `reset()` for clarity.

specs/networking/libp2p-standardization.md
```diff
@@ -141,18 +153,327 @@ The `data` field of a Gossipsub `Message` is an SSZ-encoded object. For the `bea
...
+length-prefix and wait for the exact number of bytes to sent. Once received the
+requester drops the stream.
```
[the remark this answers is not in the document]
AgeManning: Yes, correct. I didn't distinguish the two due to my comment above. 
I will update it to do so. 
Thanks!

zah:
The document doesn't need to use too abstract terms at the moment, but the encoding is generally responsible for implementing a framing scheme that divides the stream into individual messages. 

With SSZ in particular, we can say that the records are wrapped in a SSZ envelope which consist of adding a varint length prefix to the serialized bytes. I guess we need some feedback from the community in order to decide whether we want anything fancier such as including the length prefix only for variable-sized records. In our own implementation, it would be relatively easy to implement this.

specs/networking/libp2p-standardization.md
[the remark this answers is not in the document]
mhchia: No problem. I'm not 100% sure if we should use reset here though. IMO specifying the stream close/reset will be clearer when we implement the stream handlers. I will also look into it. Thank you. 🙏

specs/networking/libp2p-standardization.md
```diff
@@ -141,18 +153,329 @@ The `data` field of a Gossipsub `Message` is an SSZ-encoded object. For the `bea
...
+
+### RequestBeaconBlockRoots
+
+**Protocol ID:** `/eth/serenity/rpc/beacon_block_roots/1.0.0/<Encoding>`
+
+**Request Content**
+
+```
+(
+	start_slot: uint64
```
djrtwo:
```suggestion
    start_slot: uint64
```

specs/networking/libp2p-standardization.md
```diff
@@ -141,18 +153,329 @@ The `data` field of a Gossipsub `Message` is an SSZ-encoded object. For the `bea
...
+	count: uint64
```
djrtwo:
```suggestion
    count: uint64
```

specs/networking/libp2p-standardization.md
```diff
@@ -141,18 +153,329 @@ The `data` field of a Gossipsub `Message` is an SSZ-encoded object. For the `bea
...
+)
+
+(
+    roots: []BlockRootSlot
+)
+```
+
+Requests a list of block roots and slots from the peer. The `count` parameter
+MUST be less than or equal to `32768`. The slots MUST be returned in ascending
+pslot order.
```
djrtwo:
```suggestion
slot order.
```

specs/networking/libp2p-standardization.md
```diff
@@ -39,9 +47,10 @@ Current defaults are:
 - Digests: `SHA256` (also supports `SHA512`)
 
 *Note: Secio is being deprecated in favour of [TLS
-1.3](https://github.com/libp2p/specs/blob/master/tls/tls.md). It is our
-intention to transition to use TLS 1.3 for encryption between nodes, rather
-than Secio.*
+1.3](https://github.com/libp2p/specs/blob/master/tls/tls.md) or
```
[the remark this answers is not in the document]
AgeManning: Yep agreed. Perhaps we raise this in the call.

AgeManning:
@zah - Yep agree with everything you're saying. 

My suggestion was to keep the framing static across all encodings. The reason for this, is that in the RPC implementation we are opening a stream sending a request and receiving a response then closing the stream. I figured a standard single frame which allows for all encodings to provide a max-size to be sent per stream would be applicable. 

I'm happy to change this also and leave it up to the encoding and ensure all encodings allow us to size a single-framed response.

AgeManning:
@zah - Agree with you and have updated to shift the length-prefix in the encoding. 
I've added a small example with SSZ, let me know if you think it needs further details. 

Things still to consider:
- `RPC_MAX_SIZE` - what should be the maximum size of RPC payloads a client should accept
- Gossipsub encoding - Gossipsub sends data via protobuf - but the data within this is up to us. Should we add an `<Encoding>` string to the gossipsub protocol id or specify a specific compression algorithm. I expect we want to compress the bytes rather than just send ssz-encoded objects over gossip

specs/networking/libp2p-standardization.md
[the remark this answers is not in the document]
AgeManning: I have updated to include a terminology section. Feel free to re-word/add parts that are unclear or misleading as you see fit. :)

specs/networking/libp2p-standardization.md
```diff
@@ -183,42 +197,53 @@ Each RPC message is segregated into it's own libp2p protocol id, which is a stri
...
-* **MessageName** -- each RPC request is identified by a name consisting of English letters, digits and underscores (_).
-* **SchemaVersion** -- a semantic version consisting of one or more numbers separated by dots (.). Each schema is versioned to facilitate backward and forward-compatibility when possible.
-* **Encoding** -- while the schema defines the data types in more abstract terms, the encoding describes a specific representation of bytes that will be transmitted over the wire. See the [Encodings](#encodings) section, for further details.
-
-This protocol segregation allows libp2p `multistream-select` to handle the RPC-type, version and encoding negotiation before establishing the underlying substreams.
+* **ProtocolPrefix** -- the RPC messages are grouped into families identified
+  by a shared LibP2P protocol name prefix. A conforming implementation is
+  expected to support either all messages appearing within a family or none of
+  them. In this case, we use `/eth/serenity/rpc`.
```
zah: I have no strong objections to using this prefix, but I'll share my reasoning for why I used `/ETH/BeaconChain/` in my own proposal. 

The current RPC procs describe everything necessary to sync and propagate changes to the beacon chain. As soon as we add shards to the mix, we'll need new RPC procs for syncing the state of shards. Custom execution environments may also need their own specialized protocols for communicating with the relayers in the network.

"Serenity" is a term that encompasses all of these future developments and the wording in the spec suggests that a node should support either all of the protocols or none of them, but I see some potential for more specialized software that syncs only with the beacon chain for example. 

Thus, I tried to come up with the most unambiguous prefix name that is unlikely to clash with anything else in the future. "/ETH/BeaconChain/" seemed to fit that bill. The choice of PascalCase was another silly concern trying to win 1 byte in the multiselect handshakes :)

specs/networking/libp2p-standardization.md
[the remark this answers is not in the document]
raulk: This didn't come up in the call. But indeed, TLS 1.3 language support is not universal for now. For example, nodejs lacks such support. Protocol Labs is funding @jasnell's work on adding QUIC support to nodejs (https://github.com/nodejs/quic) + TLS 1.3, because it's a prerequisite.

The libp2p team is very favourable to Noise. I believe rust-libp2p introduced early experimental support (but hasn't submitted a spec to libp2p/specs yet), and we've toyed with the idea of adding support for go-libp2p. Here's a spec: https://github.com/libp2p/go-libp2p/issues/631

I think we should prioritise these efforts. I want to support the IK, IX and XX handshakes.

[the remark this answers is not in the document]
raulk: An issue with Noise is the compatibility with QUIC. There's [nQUIC](https://eprint.iacr.org/2019/028.pdf), and I've asked @marten-seemann (our QUIC expert) to look into it this quarter.

specs/networking/libp2p-standardization.md
AgeManning: I am also not strongly opinionated on this, but would opt for whatever we go for, to be somewhat consistent with all protocols. 
Hopefully there will be more input by someone and I'm happy to go with whatever is the consensus.

specs/networking/libp2p-standardization.md
```diff
@@ -1,29 +1,49 @@
...
 
 This section details the libp2p transport layer that underlies the
 [protocols](#protocols) that are listed in this document.
 
-Libp2p allows composition of multiple transports. Eth2.0 clients should support
-TCP/IP and optionally websockets. Websockets are useful for implementations
-running in the browser and therefore native clients would ideally support these implementations 
-by supporting websockets.
+Libp2p allows composition of multiple transports. Eth2.0 clients MUST support
+TCP/IP and MAY support websockets. Websockets are useful for implementations
```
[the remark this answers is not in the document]
AgeManning: Yep, will add

specs/networking/libp2p-standardization.md
```diff
@@ -1,29 +1,49 @@
...
+
+# Terminology
+
+* **Stream** - A two-way connection to a peer that has optionally negotiated
+	stream-multiplexing with either Mplex or Yamux.
+* **SubStream** - A two-way connection to peer that is dedicated to a specific
+	protocol. 
+* **Protocol Id** - A byte string used in the libp2p framework to negotiate
+	substreams for specific protocols. 
+* ** Close a (sub)stream** - Close the local end of a stream. I.e `stream.close()`.
```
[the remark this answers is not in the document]
AgeManning: Thanks for raising this. I'm not familiar with the intricacies of these. To my knowledge, rust-libp2p (which I'm most familiar with) only allows for closing the local end of streams. I'm not sure if this is general across all implementations.
Perhaps in this specification, we don't distinguish between the two? What are your thoughts?

specs/networking/libp2p-standardization.md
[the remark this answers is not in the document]
djrtwo: Considering QUIC is also not yet widely supported, I'm okay with Noise without QUIC for now with the intention to move to TLS 1.3 + QUIC as support becomes available.

I'd say getting a Noise spec written in libp2p/specs is a top priority at this point.

specs/networking/libp2p-standardization.md
djrtwo: If we are to go with `/eth/beacon_chain`, that would imply what other protocols...

* `/eth/shard` -- specific requests to shard chains, likely most calls would have 1st param as `shard_id`
* `/eth/network` -- akin to `net_` methods in eth1 rpc

Any others?

Another question is should the initial component of the prefix be `/eth2` -- potentially allows for better segregation if components of eth1 end up using the libp2p for any part of the stack

specs/networking/libp2p-standardization.md
```diff
@@ -117,11 +141,13 @@ There are two main topics used to propagate attestations and beacon blocks to
...
 - The `beacon_attestation` topic - This topic is used to propagate
-	aggregated attestations to subscribing nodes (typically block proposers) to
-	be included into future blocks. Attestations are aggregated in their
-	respective subnets before publishing on this topic.
+	aggregated attestations (in their entirety) to subscribing nodes (typically
+	block proposers) to be included into future blocks. Attestations are
+	aggregated in their respective subnets before publishing on this topic.
 
 Shards are grouped into their own subnets (defined by a shard topic). The
 number of shard subnets is defined via `SHARD_SUBNET_COUNT` and the shard
```
djrtwo: `shard{shard_number % SHARD_SUBNET_COUNT}_attestation` topic is going to have a potential name collision in phase 1 because `ShardAttestation`s will also need to be propagated in addition to the beacon `Attestation` defined in phase 0.

Maybe we should go with `shard{shard_number % SHARD_SUBNET_COUNT}_beacon_attestation` to clearly distinguish between the two

djrtwo:
> Gossipsub encoding - Gossipsub sends data via protobuf - but the data within this is up to us. Should we add an <Encoding> string to the gossipsub protocol id or specify a specific compression algorithm. I expect we want to compress the bytes rather than just send ssz-encoded objects over gossip

We should certainly compress in gossipsub.

Options:

* specify encoding via the gossip sub protocol id. gives us flexibility to add/change
* specify it across the board singularly for gossipsub without using the protocol id

I'm not certain we should support a non-compressed encoding here, but specifying the compressed encoding via the protocol ID seems to be in the spirit of libp2p and gives us flexiblity in upgrading down the line if needed

specs/networking/libp2p-standardization.md
jannikluhn: > Any others?

I guess everything phase 2 related -- propagating transactions, syncing state, requesting light client proofs, etc., and everything replicated for the different execution environments.

> Another question is should the initial component of the prefix be /eth2 -- potentially allows for better segregation if components of eth1 end up using the libp2p for any part of the stack

Originally, I preferred `eth2`, mostly because everything else is called `eth2` as well. But one could think of cases in which `eth1` and `eth2` clients would want to speak with each other and then having a different prefix might be awkward (thinking about this idea of storing raw block data in Swarm or another sort of persistence layer). But even then I guess we could have an entirely different prefix for those cases?

specs/networking/libp2p-standardization.md
```diff
@@ -135,24 +161,369 @@ number of shard subnets is defined via `SHARD_SUBNET_COUNT` and the shard
...
+
+After a request has been received on a negotiated stream, the responder SHOULD
+send a response within **10 seconds**.  A response consists of a **single-byte**
+response code which determines the contents of the response.
+
+It can have one of the following values
+
+* 0: **Success** -- a normal response with contents matching the expected
+  message schema and encoding specified in the request.
+* 1: **EncodingError** -- the receiver was not able to decode and deserialize
```
jannikluhn: Why not merge this with `InvalidRequest`? I think this would be nice from a conceptual point of view: Either you did something wrong, I did something wrong, or everything is ok. Also, `EncodingError`s should never happen anyways.

specs/networking/libp2p-standardization.md
```diff
@@ -117,11 +141,13 @@ There are two main topics used to propagate attestations and beacon blocks to
 all nodes on the network.
 
 - The `beacon_block` topic - This topic is used solely for propagating new
-	beacon blocks to all nodes on the networks.
+	beacon blocks to all nodes on the networks. Blocks are sent in their
+	entirety. Clients who receive blocks on a topic SHOULD validate the block
+	before forwarding it across the network.
```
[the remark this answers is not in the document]
djrtwo: Yes, good catch

specs/networking/libp2p-standardization.md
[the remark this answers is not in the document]
AgeManning: I agree with this and will update the wording to reflect this above, unless there are any objections?

specs/networking/libp2p-standardization.md
AgeManning: To make this discussion more concrete, there is only really two protocol Id's we currently need to specify.  Gossipsub and the RPC. The topics within gossipsub can be set independently and without a prefix. 
Is the consensus:
`gossipsub`: `/eth2/beacon_chain/gossipsub/<version>/<encoding>`
`rpc`: `/eth2/beacon_chain/rpc/<version>/<encoding>` 
?

specs/networking/libp2p-standardization.md
AgeManning: Yep sure. I guess that `ShardAttestation`'s will have their own set of gossipsub subnets right?

specs/networking/libp2p-standardization.md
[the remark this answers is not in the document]
AgeManning: Yep, this is definitely the intention. I will update to clarify

specs/networking/libp2p-standardization.md
```diff
@@ -135,24 +161,369 @@ number of shard subnets is defined via `SHARD_SUBNET_COUNT` and the shard
 Each Gossipsub
 [Message](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/rpc.proto#L17-L24)
 has a maximum size of 512KB (estimated from expected largest uncompressed block
-size).
+size). Clients SHOULD reject messages that are over this size limit.
```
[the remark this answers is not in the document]
AgeManning: Yep I'll add this. I originally didn't because I was worried that with `and MUST NOT send` implementers may then assume no messages over the size limit will get sent and then not implement protective measures in the case that some do.

specs/networking/libp2p-standardization.md
```diff
@@ -135,24 +161,369 @@ number of shard subnets is defined via `SHARD_SUBNET_COUNT` and the shard
...
 ([discv5](https://github.com/ethereum/devp2p/blob/master/discv5/discv5.md))
-will be used for discovery. This protocol uses a UDP transport and specifies
-its own encryption, ip-discovery and topic advertisement. Therefore, it has no
+is used for discovery. This protocol uses a UDP transport and specifies
+its own encryption, IP-discovery and topic advertisement. Therefore, it has no
 need to establish streams through `multistream-select`, rather, act
 as a standalone implementation that feeds discovered peers/topics (ENR-records) as
-`multiaddrs` into the libp2p service.
+`multiaddrs` into the libp2p service. The libp2p service subsequently forms
+connections and streams with discovered peers.
```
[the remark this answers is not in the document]
AgeManning: Yep, good point! I'll add in the bare necessity for discovery, and leave it open for added fields.

specs/networking/libp2p-standardization.md
```diff
@@ -135,24 +161,369 @@ number of shard subnets is defined via `SHARD_SUBNET_COUNT` and the shard
...
+With:
+* **ProtocolPrefix** -- the RPC messages are grouped into families identified
+  by a shared LibP2P protocol name prefix. A conforming implementation is
+  expected to support either all messages appearing within a family or none of
+  them. In this case, we use `/eth/serenity/rpc`.
+* **MessageName** -- each RPC request is identified by a name consisting of
+English letters, digits and underscores (_).
+* **SchemaVersion** -- a semantic version consisting of one or more numbers
+  separated by dots (.). Each schema is versioned to facilitate backward and
+  forward-compatibility when possible.
```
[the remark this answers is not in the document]
AgeManning: I agree entirely. This is an important discussion. We get some benefits when pushing things into multi-stream select, such as not requiring a method id in the RPC anymore. There is some discussion on the versioning here: https://github.com/ethereum/eth2.0-specs/pull/1158
I'm interested to know your thoughts on the above discussion.

specs/networking/libp2p-standardization.md
AgeManning: Yep I agree. I had headaches with implementing this anyway. The code will also be cleaner.

AgeManning:
A gossipsub encoding tag on the end of the protocol-id does seem in the spirit of libp2p. 
The implementation could be tricky when supporting multiple encodings, depending on logic. 
If Node1 supports encoding A, Node 2 supports A, B and node 3 supports B. I think there will be two subnets for A, and B and then should Node 2 route packets from the B subnet to A?
Whereas if we specify the encoding as some prefix in the data, all nodes would be on the same subnet but the message propagation properties of gossipsub would be modified depending on the percentage of nodes supporting the same protocol (as nodes that can't decode a block won't propagate it). 
@raulk - Do you have thoughts on the best way to handle multiple encodings/compression for gossipsub?
I guess we just keep all nodes segregated on their own subnets based on the encodings they support and therefore a protocol id including encoding/compression makes sense?

specs/networking/libp2p-standardization.md
zah: I agree, it's reasonable to use `InvalidRequest` in place of `EncodingError`.

zah:
Initially, I though that it would be a problem to specify the encoding in the Gossipsub topic name, but now I think it's reasonable. Each client will broadcast on a single topic using the preferred (e.g. latest) encoding. The client will also subscribe to all other topics/encodings that it can support. 

Thus, when we upgrade the network, newer clients will see the traffic from older clients, but the older clients will quickly become unusable. This is only slightly different from having a single pre-determined topic where the encoding is specified in the message content. The pros and cons go like this:

* With multiple topics, the clients will always get only messages relevant to them. 
* With a single topic, the client will know that it's out of date and it can notify the client operator somehow. This seems like a minor benefit, because there are other possible heuristics to detect this and the operator is not that likely to miss such major news such as an upcoming network upgrade.

specs/networking/libp2p-standardization.md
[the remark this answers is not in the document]
raulk: Sounds good. Note that clients supporting QUIC will advertise so in their multiaddrs, so other clients/languages supporting that combo should immediately benefit from that enhanced transport. Clients/languages not supporting QUIC will simply not dial those addresses, and will instead fallback to TCP automatically.

raulk:
@AgeManning 

> The reason for this, is that in the RPC implementation we are opening a stream sending a request and receiving a response then closing the stream.

Note that this is currently a bit costly in libp2p due to multistream 1.0 being chatty and stateless. multistream 2.0 will introduce 0-RTT negotiation by way of shared protocol tables and session resumption. So you'll be able to open streams with zero negotiation cost, as long as that protocol has been negotiated in that session/in the recent past with that peer.

I acknowledge that the Multistream 2.0 conversation in https://github.com/libp2p/specs/pull/95#discussion_r250016821 is arguably hard to follow, so @marten-seemann and I are putting together a document summarising the current thinking and technical design.

raulk:
@AgeManning 

> @raulk - Do you have thoughts on the best way to handle multiple encodings/compression for gossipsub?

Let me think about this and discuss this with a bunch of people. Will get back to you soon.

specs/networking/libp2p-standardization.md
djrtwo: yes

raulk:
@AgeManning 

> @raulk - Do you have thoughts on the best way to handle multiple encodings/compression for gossipsub?

A priori, the suggested approach of embedding the encoding and compression in the topic name appears reasonable. It enables selective consumption based on one's own capabilities. A few questions:

* Are you intending for nodes in the network to "translate" between serialisation formats and compression algorithms?
* How do you guarantee that a piece of data emitted by peer P in a specific encoding/compression combination can be effectively disseminated and understood by the entire network, despite the heterogeneity?
* Or is the setup such that nodes can safely ignore messages they don't understand, yet the network can come to consensus?

Suggestion: may be worthwhile to develop a small-scale PoC (e.g. 100 nodes) and capture metrics on write amplification, memory usage, network bandwidth, and control message piggybacking efficiency. The libp2p core team can help with that.

specs/networking/libp2p-standardization.md
```diff
@@ -135,24 +161,369 @@ number of shard subnets is defined via `SHARD_SUBNET_COUNT` and the shard
...
+  transmitted over the wire. See the [Encodings](#encodings) section, for
+  further details.
+
+
+This protocol segregation allows libp2p `multistream-select` to handle the
+RPC-type, version and encoding negotiation before establishing the underlying
+substreams.
+
+
+#### Requests and Responses
```
Nashatyrev: If I get it right, it is suggested that each request-response would need to open a new dedicated stream. 
Wouldn't this cause a significant overhead in terms of latency, traffic and client computational resources?
Opening a stream requires additional negotiation roundtrip when negotiating a protocol and may required more with multiple protocol versions. 
I'm also not sure how 'lightweight' is stream creation in different libp2p implementations.

raulk: Yes, in terms of latency; little in terms of traffic and computational resources. I pointed this out here: https://github.com/ethereum/eth2.0-specs/pull/1281#issuecomment-512212450.

AgeManning: This was also briefly discussed in #1158. As @raulk mentions there will be overheads in multistream select 1 and potential versioning issues/features where each RPC has it's own version/protocol-id.
We opted for a forward-looking design, where we assume that multistream-select2 is adopted and the overhead is minimized. This doesn't mitigate a potential versioning issue, but we gain some benefits, such as not sending the method_id and encodings in each RPC request. This is of course still up for debate.

specs/networking/libp2p-standardization.md
```diff
@@ -39,9 +59,10 @@ Current defaults are:
 - Digests: `SHA256` (also supports `SHA512`)
 
 *Note: Secio is being deprecated in favour of [TLS
-1.3](https://github.com/libp2p/specs/blob/master/tls/tls.md). It is our
-intention to transition to use TLS 1.3 for encryption between nodes, rather
-than Secio.*
+1.3](https://github.com/libp2p/specs/blob/master/tls/tls.md) or
+[Noise](https://perlin-network.github.io/noise/). It is our intention to
```
ralexstokes: pretty sure we are talking about this Noise: https://noiseprotocol.org/noise.html

not the linked noise

AgeManning:
@zah - I was originally considering an encoding in the protocol-id, but I agree it seems to make more sense having multiple topics segregating the encodings and a single protocol id. 

@raulk 
> Are you intending for nodes in the network to "translate" between serialisation formats and compression algorithms?

This is of course a possibility, but in my opinion I don't think we should add this complexity. As @zah mentions, if the network upgrades to a new format, perhaps we have a mechanism that newer nodes can notify other nodes that their format is out of date. We may introduce nodes that do translation on older topics (so older clients can still receive blocks/attestations, albeit slower), but I don't think we should include it in this document. 

> How do you guarantee that a piece of data emitted by peer P in a specific encoding/compression combination can be effectively disseminated and understood by the entire network, despite the heterogeneity?

I guess we don't. In such an upgrade perhaps we stress that clients should all be updated. In reality, it is probably reasonable to have a few clients do the translation from topic/encoding to another. As long as one client does this translation, the data should propagate on both topics. We would have to test how quickly this works for consensus however. Perhaps others have more thoughts on this, @djrtwo?

> Or is the setup such that nodes can safely ignore messages they don't understand, yet the network can come to consensus?

We are using gossipsub to propagate new blocks/attestations. It would not be good to lose these, consensus should still work depending on the % of validators split across topics. As above, technically this is solveable using translating nodes, whether we do this or require all clients to update is a separate question imo. But I don't see the design choice of using topics to separate encodings to be a prohibitive design choice when updating encodings.

> Suggestion: may be worthwhile to develop a small-scale PoC (e.g. 100 nodes) and capture metrics on write amplification, memory usage, network bandwidth, and control message piggybacking efficiency. The libp2p core team can help with that.

Absolutely. This is something I'm interested in, and we (lighthouse) will be doing version's of this with our client, but more granular tests for specific libp2p protocols should also be done (we would like to help when/where we can, just may not have the time/manpower). Perhaps other parties are interested in this also?

specs/networking/libp2p-standardization.md
AgeManning: Oh yep. Nice catch. Will update. Thanks

arnetheduck:
> We should certainly compress in gossipsub.

has anyone looked into the gains here, and algorithms? hashes compress poorly, not sure about the rest.

> multiple topics segregating the encodings

+1 for multiple topics I think :) during transition periods it's also easy for clients to publish to multiple topics and keeping it separate at this level should allow the topic listening mechanisms to adapt to new-vs-old client ratios so there's a natural ramp-up/down as people upgrade. One thing necessary for this to work well: that topic subscription is used as a propagation filter meaning that not everyone receives all data on all topics - is this the case with libp2p? Also in this scenario, how is peer connectivity handled? If I'm not listening to a topic, can I publish to it?

ping @raulk

arnetheduck:
> We opted for a forward-looking design, where we assume that multistream-select2 is adopted and the overhead is minimized. 

as a thought experiment, let's say we opt for an optimal solution given the capabilities of today and use a single stream to not burden multiselect 1. What and how difficult would the steps be to change the protocol to use multiple streams later?

AgeManning:
> that topic subscription is used as a propagation filter meaning that not everyone receives all data on all topics - is this the case with libp2p?

Yes. Gossipsub/floodsub will only publish/propagate to node subscribed to the topic.

> Also in this scenario, how is peer connectivity handled? If I'm not listening to a topic, can I publish to it?

Yes. This is the `fanout` peers in gossipsub.

> as a thought experiment, let's say we opt for an optimal solution given the capabilities of today and use a single stream to not burden multiselect 1. What and how difficult would the steps be to change the protocol to use multiple streams later?

I think this may be implementation specific. I understand this is not too difficult for nimbus to change to. It's not too difficult for us (Lighthouse) to support either. In a single stream scenario we add back complications such as tracking of method_ids and request/response id's (unless we assume ordering).
I'm not opposed to this btw.

zah:
> as a thought experiment, let's say we opt for an optimal solution given the capabilities of today and use a single stream to not burden multiselect 1. What and how difficult would the steps be to change the protocol to use multiple streams later?

> I think this may be implementation specific. I understand this is not too difficult for nimbus to change to. It's not too difficult for us (Lighthouse) to support either. In a single stream scenario we add back complications such as tracking of method_ids and request/response id's (unless we assume ordering).
I'm not opposed to this btw.

From my experience, using a single stream requires quite a lot of additional code for tracking the requests that are in flight. In Nimbus, our codebase is structured in a way such that the low-level details of the protocol are abstracted away and there is shared high-level code that holds only the application logic needed for responding to a particular request. Here is an example:

https://github.com/status-im/nim-beacon-chain/blob/0a2d09e7fe3ff556201b569f19c11dd1bcf337df/beacon_chain/sync_protocol.nim#L270-L282

I implemented a single-stream back-end for this code only because the older spec required it. The multi-stream back-end is quite a bit simpler.

arnetheduck:
Leaving aside for a moment the code implications, how would the upgrade look from a network perspective of running clients? How.. difficult.. would a negotiated upgrade be?

From a code perspective, I read that as going from ms1 to ms2 is a simplification, thus trivial.

zah:
Well, it will be trivial only if you have planned for it. I shared our implementation approach as a tip.
Otherwise, migrating to a new protocol is not hard. The clients will just negotiate and select different stream types in libp2p.

djrtwo:
> We would have to test how quickly this works for consensus however. Perhaps others have more thoughts on this, @djrtwo?

I don't expect to have much mismatch on gossiping the high throughput items on validating nodes. Coordinating network upgrade (even if outside of hardfork) is certainly likely. To force everyone's hand, can just couple network upgrades at same time as hard fork so there isn't much option to not upgrade the software.

djrtwo:
Should be noted that the additional latency in negotiating streams in ms1 for RPC calls will likely not hinder the protocol too much, especially in early deployments. The thing we would be most concerned about is latency wrt pubsub gossiping algs.

specs/networking/libp2p-standardization.md
```diff
@@ -135,24 +168,395 @@ number of shard subnets is defined via `SHARD_SUBNET_COUNT` and the shard
...
+Topics are post-fixed with an encoding. Encodings define how the `data` field
+of a Gossipsub `Message` is encoded. Implementations SHOULD support the
+following encodings:
+
+- `ssz` -  This is a temporary encoding used only for initial interop and
+	testing. All objects are SSZ-encoded. Example: The beacon block topic
+	string is:
+	`beacon_block/ssz` and the `data` field of a Gossipsub `Message` is an
+	ssz-encoded `BeaconBlock`.
+- `ssz_snappy` - All objects are ssz-encoded and then compressed with `snappy`.
```
Nashatyrev: Wouldn't it be more efficient to compress the stream *before* multiplexer (i.e. compress the whole transport stream)? May be libp2p has a suitable compressor out-of-the-box ?
@raulk ?

specs/networking/libp2p-standardization.md
Nashatyrev: @raulk Will multistream-2 be able to handle nested protocols in one round like `/secio/1.0.0/mplex/2.6.0` ?

specs/networking/libp2p-standardization.md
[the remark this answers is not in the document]
arnetheduck: > It's unnecessary granularity: If we want, we can just increment the whole protocol's version, even if we just update a single message.

The granularity helps manage upgrades such that when specialized clients use only a part of the protocol, they need not be upgraded in some scenarios. I think this is a win, specially with the broadcast parts.

> I suspect that often the versions will not be independent. 

any such dependencies can be specified in the message that requires them.

I'm concerned about the use of semver here - it is nice for humans selecting a collection of versions to test together before using in application, but automating message selection in-protocol based on it seems like a can of worms in terms of subtle issues. This ties in with using SSZ for encoding - SSZ has no concept of optional fields really - no way of introducing purely additive changes - meaning that every time a request or response is changed, we effectively need to create a new incompatible protocol version.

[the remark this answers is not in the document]
zah: > no way of introducing purely additive changes

There is one type of upgrade that is supported by SSZ - it's adding a field at the end of a struct.

Otherwise, I agree that using a single number as a version is reasonable enough for both humans and machines.

specs/networking/libp2p-standardization.md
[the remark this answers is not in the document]
dryajov: It's worth noting that libp2p streams are half-open, meaning that in order for the stream to be completely closed, both ends should call `close`, otherwise, the stream remains readable by the sender and writable by the receiver. Not sure about all other implementations but js and go should implement `reset` in the `mplex` muxer.

specs/networking/libp2p-standardization.md
```diff
@@ -71,14 +98,14 @@ optionally [yamux](https://github.com/hashicorp/yamux/blob/master/spec.md)
 
 ## Gossipsub
 
-#### Protocol id: `/eth/serenity/gossipsub/1.0.0`
+#### Protocol id: `/eth2/beacon_chain/gossipsub/1.0.0/`
```
mkalinin: This protocol uses libp2p gossipsub as a pubsub provider. How would a handler for this protocol look like? Would it use gossipsub handler as a delegate?

We might want to explicitly mention gossipsub in the spec, e.g. `/meshsub/1.0.0`. For the sake of conformance with `rpc` I'd suggest to specify beacon chain pubsub in a following way:
- `/eth2/beacon_chain/pubsub/<SchemaVersion>/<Encoding>`
- `/meshsub/1.0.0` is used as underlying protocol

Schema version would handle changes in message format if e.g. an update in aggregation algorithm leads to message patching. If encoding becomes a part of protocol id then topic id could get rid of it which makes forward compatibility less complicated.

There is a question here. What if we want to upgrade to `/meshsub/2.0.0`? What do we do then? But I guess the same question could be addressed to `/mplex` and the others.

mkalinin:
Forward compatibility principle described in [EIP-8](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-8.md) is a very nice feature. It preserves network liveness during various updates by keeping ability of old peers to speak with upgraded ones.

In order to be forward compatible devp2p has a negotiation process that includes protocol version exchange. Peers are sending highest supported version to each other and starts communication with minimal version supported by both ends.

We could leverage Discovery v5 ENRs to get a list of capabilities supported by remote peer _before_ establishing a connection with it. To get to that in our case nodes will need to advertise a couple of lists:
- supported schema versions
- supported encodings

Using this data node could build protocol id _before_ it starts to communicate with remote peer. Node picks schema version and encoding according to its own preferences and remote peer capabilities.

For example, if encoding would be updated to e.g. `xml` we would have following scenario:
- upgraded nodes start sending new ENRs with `eth2_codecs: ssz, xml` and `xml` takes preference over `ssz`
- if remote peer supports `xml` then to establish connection upgraded node builds protocol id with new encoding `/eth2/beacon_chain/rpc/1.0.0/hello/xml`
- otherwise, it uses old one `/eth2/beacon_chain/rpc/1.0.0/hello/ssz`

Schema version could be updated in the same fashion.

This approach has a drawback. There will be time gap between restarting a node and spreading updated ENRs across the network. In particular it means that two nodes will have to be restarted twice in order to start communicate with updated capabilities. 
Possible workarounds to that:
- exchange capabilities during negotiation process (Eth1 approach)
- send updated ENR before establishing connection with remote peer

What others think about forward compatibility and this particular approach?

specs/networking/libp2p-standardization.md
```diff
@@ -112,21 +139,27 @@ propagation times and message duplication. Current network-related defaults are:
...
+- `beacon_attestation`  - This topic is used to propagate
+	aggregated attestations (in their entirety) to subscribing nodes (typically
+	block proposers) to be included into future blocks. Attestations are
+	aggregated in their respective subnets before publishing on this topic.
 
 Shards are grouped into their own subnets (defined by a shard topic). The
 number of shard subnets is defined via `SHARD_SUBNET_COUNT` and the shard
 `shard_number % SHARD_SUBNET_COUNT` is assigned to the topic:
-`shard{shard_number % SHARD_SUBNET_COUNT}_attestation`.
+`shard{shard_number % SHARD_SUBNET_COUNT}_beacon_attestation`.
```
mkalinin: How will node advertise a shard topic that it is assigned to? A straightforward solution would be to use ENR records for this purpose. If this is what we want then maybe it worth listing under ENR records of Discovery section?

specs/networking/libp2p-standardization.md
AgeManning: The current train of thought for putting the encoding inside the topics is so that peers that support different encodings / schemas still connect on the protocol. Peers then in principle could subscribe to two different topics/encodings and act as a relay without having to support two protocols. 

In the event of an upgrade, some nodes could identify other nodes by subscribing to older topics and informing them an upgrade is required. 

In regards to the protocol id. The consensus has been to change the protocol ids to `/eth/..` to avoid accidental connection with ipfs and other non eth-specific nodes. 
If we want to upgrade to a new version of meshsub, we would have to support the new protocol-id. I imagine we would rename the protocol-id to `/eth2/beacon_chain/gossipsub/2.0.0`

In regards to forwards compatibility, I believe the idea would be to coordinate a hard fork such that everyone moves to a new network, either by incrementing the protocol-id version, or in the change of an encoding, a topic prefix. Currently, there are plans to support two encodings. An empty postfix, which has no compression (for interop) then a `snappy` encoding, likely for mainnet which will be realised via two topics, i.e `beacon_block_snappy`.

AgeManning:
> Using this data node could build protocol id before it starts to communicate with remote peer. Node picks schema version and encoding according to its own preferences and remote peer capabilities.

This is what multistream-select does for us. Given a set of encodings/versions of an RPC message, we can locally order these in preference when multistream negotiates which protocol to use. 
For upgrading, we would preference newer protocols over older protocols.

AgeManning:
I'm currently working with Raul and others to incorporate feedback given here and will be updating this in the coming days.

specs/networking/libp2p-standardization.md
AgeManning: I may be missing something. If a validator is assigned to attest to a shard, it should subscribe and publish on that shard topic. 

If in phase 1 a node wants to communicate blocks and messages dedicated to that shard, I imagine we will have a specific shard topic for that. This scenario is not included in this phase 0 spec however.

mkalinin: From [gossipsub](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub#topic-membership) spec:
> Also note that as part of the pubsub api, the peer emits `SUBSCRIBE` and `UNSUBSCRIBE` control messages to all its peers whenever it joins or leaves a topic. This is provided by the the ambient peer discovery mechanism and nominally not part of the router. A standalone implementation would have to implement those control messages.

We need a mechanism that spreads node's subscription updates all over the network. From my understanding Discovery v5 was designed bearing capability updates in mind.

> This scenario is not included in this phase 0 spec however.

Depending on aggregation strategy and number of peers in the network subnets can become a unit of aggregation process in phase 0.

specs/networking/libp2p-standardization.md
mkalinin: > Peers then in principle could subscribe to two different topics/encodings and act as a relay without having to support two protocols.

IMO, acting as a relay makes things more complicated than supporting two protocols. A relay will have to match different topic ids to a single logical topic and support two encodings. While supporting two protocols only means supporting two encodings. Please, let me know if I am missing something.

A number of topic ids matching to one topic turns into a number of net graphs linked with each other via relays. It requires to have reasonably high number of relays in order to keep not yet upgraded peers subscribed to that topic.

djrtwo:
Closing in favor of #1328

Issue #1274, opened 2019-07-05
Improved transfer replay protection, and removal of transfers in phase 0


JustinDrake:
Replace the current slot-based transfer replay protection mechanism by adding `state.transfer_roots: Vector[Hash, MAX_TRANSFERS * SLOTS_IN_REPLAY_PROTECTION_PERIOD]`. We can also completely move transfers from phase 0 to phase 1.

vbuterin:
So the only substantive thing here is removing `transfers` from the block structure, right?

vbuterin:
It's also worth noting that this mechanism could be reused in phase 2 to prevent replay of withdrawals from execution environments into other execution environments.

djrtwo:
Do you still have to specify a slot or epoch?

I'm not certain what problem this is actually solving. If a validator has a Transfer for a `slot` and it is not included in a chain and then broadcasts a repeat for `slot + N`, then both of these can still be replayed on a reorged chain even though the initial transfer was not included in the pre-reorged chain.

djrtwo:
Or are you suggesting removing `slot` from `Transfer`? If this is the case, then I can just replay a transfer _after_ the replay protection period.

JustinDrake:
> Do you still have to specify a slot or epoch?

Yes.

JustinDrake:
> Or are you suggesting removing `slot` from `Transfer`?

As discussed on our call, `slot` stays there (or can be replaced with `epoch`).

djrtwo:
So is the only thing this solving the ability to expand it to `epoch` instead of `slot`?

JustinDrake:
> If a validator has a Transfer for a `slot` and it is not included in a chain and then broadcasts a repeat for `slot + N`

The point is that the re-broadcast would *not* have to modify `slot + N`, hence removing the risk of having two transfers accidentally go through.

JustinDrake:
> So is the only thing this solving the ability to expand it to epoch instead of slot?

Nope. It expands slot to `SLOTS_IN_REPLAY_PROTECTION_PERIOD`. It's possible that `SLOTS_IN_REPLAY_PROTECTION_PERIOD >> SLOTS_PER_EPOCH`.

protolambda:
> We can also completely move transfers from phase 0 to phase 1.

Then we move it to a new phase-1 beacon-chain updates document? Or will it live with the updated beacon-block body definition in the custody game doc?

JustinDrake:
[...]

PR #1283, opened 2019-07-10
Clarify the illegal SSZ types

hwwhww:
It's pointed out by @jannikluhn and offline discussed with @protolambda and @dankrad. This PR clarifies that for legal composite type, `N` should be greater than `0`.

It doesn't change the logic of core specs, just clarify the edge cases for common `ssz` testing.


dankrad:
> `List[subtype, N]` is currently used in the spec; we have 0 as transfer limit. It's perfectly ok to define 0-limit objects in my opinion. As a means to define lists ahead of time, and change the limit constant to enable functionality (e.g. a fork enabling transfers).
> 0-length vectors are arguably illegal, since they do not serve a real purpose. I still think we should support them, for flexibility (why not? If a programmer wants to disable/dynamically define a vector, we need it to support zero. Also just a sane default to have in many contexts where the type is uninitialized/uncomplete).

0-length vectors could be used for the same thing (as in, temporarily setting the number of objects we want to support at a later stage to zero). I actually agree with the reasoning here, and would actually also be happy to allow empty container types (as "placeholders" to be filled in in a later phase, for example). But would prefer it to be consistent -- either we allow all empty containers or none of them.

hwwhww:
@protolambda 
Thanks, agreed with `MAX_TRANSFERS=0` case. I updated the file while `Vector` and `Bitvector` are arguable.

@dankrad @protolambda 
I'm okay with allowing them too. Just want to make sure the edge cases are clarified and handled in the spec correctly. :)

jannikluhn:
The problem is that empty vectors and containers would serialize to an empty string. So if you'd put them into a list you run into problems because you can't properly deserialize it anymore (you can't infer the number of elements from the size, it's a division by zero).

dankrad:
> The problem is that empty vectors and containers would serialize to an empty string. So if you'd put them into a list you run into problems because you can't properly deserialize it anymore (you can't infer the number of elements from the size, it's a division by zero).

OK, looks like it's impossible to allow any zero-length *fixed size* containers. Variable size containers of zero length are, however, ok. Not very pretty, but seems the only possible solution if we want to allow the `MAX_TRANSFERS=0`.

Overall, however, I think it would have been cleaner not to allow any 0-sized containers. Probably better to go with the currently suggested solution after the spec freeze, though.

protolambda:
@jannikluhn To better illustrate the problem and context:

- basic vectors: check `scope == N * elem_size`
- basic and fixed element lists: `N = scope / elem_size`, and check if `scope % elem_size == 0` and `N < limit`
  - the culprit. `elem_size` can be zero here if we have a fixed-length element vector with `N` set to 0.
  - can we just default to `N = 0` in this case? Or should we make it illegal?
- variable-size element vectors and lists, steps:
  1. if `scope == 0`: no offsets, no elements (`N = 0`)
  2. if `0 < scope < BYTES_PER_OFFSET`: not enough data for offsets, invalid encoding.
  3. else `N = first_offset / BYTES_PER_OFFSET`
  4. Check if computed `N` matches expected `N` if it's a vector.
  5. Check if `N < limit` if it's a list

JustinDrake:
> if we want to allow the `MAX_TRANSFERS=0`

I'd say `MAX_TRANSFERS=0` is likely to go away. See #1274.

protolambda:
For lists, it's perfectly fine to support N = 0. Flexible, and simple (lists of fixed-size non-basic elements always have offsets, so we can know how many elements are there, regardless of element size).

It's really just fixed-size non-basic element vectors that are a bit inconsistent. But we can make (i.e. keep status quo) that as illegal. Then the discussion would be whether or not to start allowing zero-length basic element vectors.

dankrad:
> I'd say `MAX_TRANSFERS=0` is likely to go away. See #1274.

I think this would be my preferable solution.



> For lists, it's perfectly fine to support N = 0. Flexible, and simple (lists of fixed-size non-basic elements always have offsets, so we can know how many elements are there, regardless of element size).
> 
> It's really just fixed-size non-basic element vectors that are a bit inconsistent. But we can make (i.e. keep status quo) that as illegal. Then the discussion would be whether or not to start allowing zero-length basic element vectors.
I think it's the other way round (0-length of non-basic would be fine, basic would not).

I think ideally the SSZ spec should make sense even if you have no idea how serialization works under the hood (previously, we just had the length in elements in the serialization, which did not have this problem; switching to byte offsets exposes it, which is really an implementation detail I do not think every developer should have to think about). So making this consistent by either allowing all 0-length types or none of them would be my preferred solution.

protolambda:
If we go the 0 zero length route:

I think that the best option would be to allow zero-length vectors, and make it illegal to make a non-empty list of fixed-size zero-length elements. This would allow nested 0-length structures wherever we like, and make ambiguity in element count not possible.

- For fixed-size element lists, `if scope == 0 then N = 0`, a sane default
- when constructing a list type, if the element type is fixed-length, and has a 0 length, then the list limit MUST be 0. Compile time checking, yay.

dankrad:
> I think that the best option would be to allow zero-length vectors, and make it illegal to make a non-empty list of fixed-size zero-length elements.

😱

protolambda:
>  fixed-size zero-length elements.

I.e. elements of 0 size shouldn't be in a list without length/offsets regardless. It's the bare minimum to support 0-length fixed-size elements in list at all. Open for alternatives

dankrad:
Yeah, I understand what you mean, but that sounds very frightening and not very transparent to new developers at all.

protolambda:
What's the alternative? Just simply disallow things? Do we want it to be opinionated?
It's a complexity <> ability trade-off. And I like more expressive ability when the extra complexity that is introduced consists of just reasonable defaults (zero everywhere, zeroes result in zeroes).

I will change my mind if you come with a counter-proposal for vectors (consistent with lists) that enables limits to be 0. Being able to disable data by simply saying "limit = 0, none of this please", and having a safe default when determining a limit dynamically, is preferable imo.

Status quo is good enough, although a bit inconsistent when you compare lists to vectors.

For a completely illegal-0-limit-list and illiegal-0-length-vector approach, we would need to change the spec as well, and muck around with state fields, while in reality we are really just changing a few numbers (e.g. maximum transfers). Open for it, but I'd rather not mix application layer refactoring, post-freeze, into a working but imperfect SSZ edge case.

jannikluhn:
Status quo is broken. The SSZ spec states "serialization is an injective function" which is not true right now. We have to disallow either

1) zero-length fixed sized objects or
2) lists of zero-length fixed sized objects.

If I haven't missed anything, zero-length fixed sized objects would be vectors of size 0, empty containers, and containers recursively containing (only) zero-length fixed sized objects.

Looking only at the SSZ spec, option 1 would in my opinion be slightly cleaner, easier to implement (you can get around the recursive check), and less surprising. SSZ is a serialization format and serializing empty data structures doesn't make sense anyway (they can only ever have one value, so the serialized string contains no information).

I can't speak for the spec though, if we'd really have to change a non-trivial amount then going with 2 is probably preferable.

dankrad:
> I can't speak for the spec though, if we'd really have to change a non-trivial amount then going with 2 is probably preferable.

I don't think that we are using zero-length fixed structures at the moment, so 1 should be ok. it's also my preference as I find 2 harder to communicate. Also the recursive check seems unelegant.

specs/simple-serialize.md
```diff
@@ -89,7 +89,7 @@ An SSZ object is called empty (and thus, `is_empty(object)` returns true) if it
 
 ### Illegal types
 
-Empty vector types (i.e. `[subtype, 0]` for some `subtype`) are not legal. The `null` type is only legal as the first type in a union subtype (i.e. with type index zero).
+The empty `Vector[subtype, N]` and `Bitvector[N]` types, where `N == 0` are not legal. The `null` type is only legal as the first type in a union subtype (i.e. with type index zero).
```
protolambda:
```suggestion
- Empty vector types (Vector[T, 0], Vector[byte, 0], bitvector)
- Containers with no fields
- The `null` type is only legal as the first type in a union subtype (i.e. with type index zero).
```

protolambda: This is the other end of the proposal, keep it stupid simple:
- no empty vectors, we do not even check if the element type is variable size (which would result in allowable `Vector[T, 0]`). Instead, just no empty vectors
- no empty containers. (if they are fixed-size, they result in the same 0-length non-offset element type problem)

djrtwo:
Where did we stand on this decision @hwwhww ?

specs/simple-serialize.md
hwwhww: Yeah, I'm inclined to this (disable empty vector and container). 
Modified your suggestion in 62138fe.

hwwhww:
f336e7f renames `is_empty` to `is_zero`, solves https://github.com/ethereum/eth2.0-specs/issues/1282#issuecomment-509991996 (cc @JustinDrake) since I'd like to distinguish the definition of "empty" object here. :)

@protolambda @dankrad @jannikluhn could you review it again?

protolambda:
Still not 100% convinced personally, but making the empty vector and container types (but not list) illegal will work. And it is the most easy thing to learn as a new user, even though it limits things more than we can push it if we wanted to. So OK with this. Thanks for updating the PR :+1:

dankrad:
Looks good to me.

@protolambda To alleviate your concerns, should we see that your definition is more useful in the long run, it would be an easy (and backward-compatible) fix to allow for these.

PR #1291, opened 2019-07-11
phase 2 wiki added to readme

djrtwo:


README.md
```diff
@@ -21,6 +21,12 @@ Core specifications for Eth 2.0 client validation can be found in [specs/core](s
 * [Custody Game](specs/core/1_custody-game.md)
 * [Shard Data Chains](specs/core/1_shard-data-chains.md)
 
+### Phase 2
+
+Phase 2 is still heavily in R&D and does not yet have any formal specifications.
```
hwwhww:
```suggestion
Phase 2 is still actively in R&D and does not yet have any formal specifications.
```

README.md
```diff
@@ -21,6 +21,12 @@ Core specifications for Eth 2.0 client validation can be found in [specs/core](s
...
+
+See the [Eth 2.0 Phase 2 Wiki](https://hackmd.io/UzysWse1Th240HELswKqVA?view) for current progress, discussions, and definitions regarding this work.
```
hwwhww: My 2 cents: the current hackmd permission makes it hard to collaborate on it. @villanuevawill have you considered either changing permission or moving to another platform/[Ethereum Wiki repo](https://github.com/ethereum/wiki/wiki)/[eth.wiki](http://eth.wiki/)? :)

Mikerah: Instead of linking to the hackmd, we can just add another spec folder for phase 2 related stuff like we do for phase 1 and phase 0.

villanuevawill: I will update permissions to enable commenting. Agreed in the future we should have a separate spec folder, but things are changing pretty quickly right now so it may just spam things. Wait for a bit more stability?

villanuevawill: OK - comments are supported. Also - happy to give permissions to anyone who wants to make updates. I'll add a note in the doc.

djrtwo: Agreed, prefer minimal for the time being. dont want to induce PR overhead yet

PR #1292, opened 2019-07-12
Correct merkleization

protolambda:
This PR fixes some bugs/problems in SSZ merkleization, some brought up by @zilm13 on gitter (thanks!):
- Spec is unclear about the enforcing of the limit in the merkleization.
   - [x] Changed: `pad_for` argument is a `limit` argument now. It can be `None` to not set a limit. But if set, it ensures the count does not exceed the limit before merkleizing. And then pads the chunks to the next power of two of the limit. (So we get the same behavior as before, pushing leafs to the bottom of the virtual full tree, for stable-generalized-indices).
- Padding under the count happens when a limit is exceptionally low, and we incorrectly put too much in a list. In this case it is more correct to enforce the padding, than to take the next power of 2 of the count.
   - [x] Changed: enforce padding limit by limiting count to pad_for length during merkleization.
- zilm13: 
  - Padding to 0 happens when a list limit is 0. But merkleization code was not handling it well, and instead padded too much.
     - [x] Changed: return `0x00 * 32` for the 0 case.
  - Spec is unclear about bitlist padding. The `merkleize(as_integer.to_bytes((N + 7) // 8, "little"))` is hard to read as padding.
     - [x] Get rid of separate bitlist/bitvector merkleization. Unify it with the other description, properly define the limit and mix-in, and use `chunk_count`. (definition will be useful for merkleization of ssz partials later on as well).
- [x] update merkleization tests

This PR affects generated test-vector output, as pyspec behavior is fixed. But does not change the described spec behavior (although it will make things more clear/explicit).

Old behavior:
```
[] pad to 0                     f5a5fd42d16a20302798ef6ed309979b43003d2320d9f0e8ea9831a92759fb4b
[] pad to 1                     0000000000000000000000000000000000000000000000000000000000000000
[] pad to 2                     f5a5fd42d16a20302798ef6ed309979b43003d2320d9f0e8ea9831a92759fb4b
[0x22 * 32] pad to 0            15f7aab5653fa38aa664b0a2271a15d05c703ef2e5cf51e8330d20180a17fb7f
[0x22 * 32] pad to 1            2222222222222222222222222222222222222222222222222222222222222222
[0x22 * 32] pad to 2            15f7aab5653fa38aa664b0a2271a15d05c703ef2e5cf51e8330d20180a17fb7f
[0x22 * 32, 0x33 * 32] pad to 0 c8e5ff41fdd4ca636d8fa244e9600532589f5023c7c17c3e02f7e0fa743e7fe1
[0x22 * 32, 0x33 * 32] pad to 1 c8e5ff41fdd4ca636d8fa244e9600532589f5023c7c17c3e02f7e0fa743e7fe1
[0x22 * 32, 0x33 * 32] pad to 2 c8e5ff41fdd4ca636d8fa244e9600532589f5023c7c17c3e02f7e0fa743e7fe1
```

New, correct behavior:
```
[] limit = 0                        0000000000000000000000000000000000000000000000000000000000000000
[] limit = 1                        0000000000000000000000000000000000000000000000000000000000000000
[] limit = 2                        f5a5fd42d16a20302798ef6ed309979b43003d2320d9f0e8ea9831a92759fb4b
[0x22 * 32] limit = 0               invalid
[0x22 * 32] limit = None            2222222222222222222222222222222222222222222222222222222222222222
[0x22 * 32] limit = 1               2222222222222222222222222222222222222222222222222222222222222222
[0x22 * 32] limit = 2               15f7aab5653fa38aa664b0a2271a15d05c703ef2e5cf51e8330d20180a17fb7f  h(0x222... ++ 0x000...)
[0x22 * 32, 0x33 * 32] limit = 0    invalid
[0x22 * 32, 0x33 * 32] limit = 1    invalid
[0x22 * 32, 0x33 * 32] limit = None c8e5ff41fdd4ca636d8fa244e9600532589f5023c7c17c3e02f7e0fa743e7fe1  h(0x222... ++ 0x333...)
[0x22 * 32, 0x33 * 32] limit = 2    c8e5ff41fdd4ca636d8fa244e9600532589f5023c7c17c3e02f7e0fa743e7fe1
```

### DIFF
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -25,8 +25,6 @@
         - [Vectors, containers, lists, unions](#vectors-containers-lists-unions)
     - [Deserialization](#deserialization)
     - [Merkleization](#merkleization)
-        - [`Bitvector[N]`](#bitvectorn-1)
-        - [`Bitlist[N]`](#bitlistn-1)
     - [Self-signed containers](#self-signed-containers)
     - [Implementations](#implementations)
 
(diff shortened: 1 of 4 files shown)


specs/simple-serialize.md
```diff
@@ -177,38 +175,36 @@ Note that deserialization requires hardening against invalid inputs. A non-exhau
 
 We first define helper functions:
 
+* `chunk_count(type)`: calculate the amount of leafs for merkleization of the type.
+   * all basic types: `1`
+   * bitlists and bitvectors: `(N + 255) // 256` (dividing by chunk size, rounding up)
+   * lists and vectors of basic types: `N * item_length(elem_type) + 31) // 32` (dividing by chunk size, rounding up)
```
djrtwo: `item_length` is not defined in the spec and there looks to be an un-opened trailing paren after `31`

specs/simple-serialize.md
```diff
@@ -177,38 +175,36 @@ Note that deserialization requires hardening against invalid inputs. A non-exhau
 
 We first define helper functions:
 
+* `chunk_count(type)`: calculate the amount of leafs for merkleization of the type.
+   * all basic types: `1`
+   * bitlists and bitvectors: `(N + 255) // 256` (dividing by chunk size, rounding up)
```
djrtwo: `chunk_count` only takes `type` as input so `N` ends up being kind of implicitly available. I suppose that's okay, thoughts?

specs/simple-serialize.md
protolambda: defined `size_of` in 1 line, for basic types. See d9fd1d3

specs/simple-serialize.md
protolambda: replaced type names with type descriptions, to source the `N` and element type from. See d9fd1d3

PR #1294, opened 2019-07-13
Attestation changes + persistent committee changes

vbuterin:
* Shard slots
* Attestation list -> single attestation
* One-day periods
* Committee simplification

### DIFF
--- a/specs/core/1_shard-data-chains.md
+++ b/specs/core/1_shard-data-chains.md
@@ -17,18 +17,21 @@
         - [Signature domain types](#signature-domain-types)
         - [TODO PLACEHOLDER](#todo-placeholder)
     - [Data structures](#data-structures)
-        - [`ShardBlockBody`](#shardblockbody)
-        - [`ShardBlock`](#shardblock)
         - [`ShardBlockHeader`](#shardblockheader)
+        - [`ShardBlock`](#shardblock)
+        - [`ShardBlockSignatures`](#shardblocksignatures)
+        - [`ShardBlockCore`](#shardblockcore)
+        - [`ExtendedShardBlockCore`](#extendedshardblockcore)
     - [Helper functions](#helper-functions)
         - [`compute_epoch_of_shard_slot`](#compute_epoch_of_shard_slot)
         - [`get_period_start_epoch`](#get_period_start_epoch)
         - [`get_period_committee`](#get_period_committee)
-        - [`get_switchover_epoch`](#get_switchover_epoch)
         - [`get_persistent_committee`](#get_persistent_committee)
         - [`get_shard_block_proposer_index`](#get_shard_block_proposer_index)
         - [`get_shard_block_attester_committee`](#get_shard_block_attester_committee)
         - [`get_shard_header`](#get_shard_header)
+        - [`pad`](#pad)
+        - [`flatten_shard_header`](#flatten_shard_header)
         - [`compute_crosslink_data_root`](#compute_crosslink_data_root)
     - [Object validity](#object-validity)
         - [Shard blocks](#shard-blocks)
@@ -55,7 +58,8 @@ We define the following Python custom types for type hinting and readability:
 
 | Name | Value |
 | - | - |
-| `BYTES_PER_SHARD_BLOCK_BODY` | `2**14` (= 16,384) |
+| `SHARD_HEADER_SIZE` | `2**9` (= 512) |
+| `SHARD_BLOCK_SIZE_LIMIT` | `2**16` (= 65,536) |
 | `MAX_SHARD_ATTESTIONS` | `2**4` (= 16) |
 | `SHARD_SLOTS_PER_BEACON_SLOT` | `2**1` (= 2) |
 | `MAX_PERSISTENT_COMMITTEE_SIZE` | `2**7` (= 128) |
@@ -72,7 +76,7 @@ We define the following Python custom types for type hinting and readability:
 | Name | Value | Unit | Duration |
 | - | - | :-: | :-: |
 | `CROSSLINK_LOOKBACK` | `2**0` (= 1) | epochs | 6.4 minutes |
-| `PERSISTENT_COMMITTEE_PERIOD` | `2**8` (= 256) | epochs | ~27 hours |
+| `EPOCHS_PER_SHARD_PERIOD` | `2**8` (= 256) | epochs | ~27 hours |
 
 ### Signature domain types
 
@@ -91,41 +95,54 @@ The following types are defined, mapping into `DomainType` (little endian):
 
 ## Data structures
 
-### `ShardBlockBody`
+_Note: the shard block header structure is carefully designed so that all of the values have the same depth in a hash tree implementation, so `hash_tree_root(SSZ_partial(x)) == hash_tree_root(x)` (using the "left-to-right leaves" scheme [here](https://github.com/ethereum/eth2.0-specs/issues/1303)), which allows shard block headers to look like an SSZ object when in the crosslink structure. This is done by balancing it so that 7 or 8 items are on the left side (the "core") and two 96-byte (ie. 3*2 = 6 chunk) items are on the right side. Change with care._
+
+### `ShardBlockHeader`
 
 ```python
-class ShardBlockBody(Container):
-    data: Vector[Bytes[PLACEHOLDER], BYTES_PER_SHARD_BLOCK_BODY]
+class ShardBlockHeader(Container):
+    core: ShardBlockCore
+    signatures: ShardBlockSignatures
 ```
 
 ### `ShardBlock`
 
 ```python
 class ShardBlock(Container):
-    slot: ShardSlot
-    shard: Shard
-    beacon_chain_root: Hash
-    parent_root: Hash
-    data: ShardBlockBody
-    state_root: Hash
-    attester_bitfield: Bitvector[MAX_PERSISTENT_COMMITTEE_SIZE * 2]
-    attestation_signature: BLSSignature
-    signature: BLSSignature
+    core: ExtendedShardBlockCore
+    signatures: ShardBlockSignatures
 ```
 
-### `ShardBlockHeader`
+### `ShardBlockSignatures`
 
 ```python
-class ShardBlockHeader(Container):
+class ShardBlockSignatures(Container):
+    attestation_signature: BLSSignature
+    proposer_signature: BLSSignature
+```
+
+### `ShardBlockCore`
+
+```python
     slot: ShardSlot
-    shard: Shard
     beacon_chain_root: Hash
     parent_root: Hash
-    body_root: Hash
+    data_root: Hash
     state_root: Hash
+    total_bytes: uint64
     attester_bitfield: Bitvector[MAX_PERSISTENT_COMMITTEE_SIZE * 2]
-    attestation_signature: BLSSignature
-    signature: BLSSignature
+```
+
+### `ExtendedShardBlockCore`
+
+```python
+    slot: ShardSlot
+    beacon_chain_root: Hash
+    parent_root: Hash
+    data_root: Bytes[SHARD_BLOCK_SIZE_LIMIT - SHARD_HEADER_SIZE]
+    state_root: Hash
+    total_bytes: uint64
+    attester_bitfield: Bitvector[MAX_PERSISTENT_COMMITTEE_SIZE * 2]
 ```
 
 ## Helper functions
@@ -137,10 +154,10 @@ def compute_epoch_of_shard_slot(slot: ShardSlot) -> Epoch:
     return Epoch(slot // SHARD_SLOTS_PER_BEACON_SLOT // SLOTS_PER_EPOCH)
 ```
 
-### `get_period_start_epoch`
+### `get_shard_period_start_epoch`
 
 ```python
-def get_period_start_epoch(epoch: Epoch, lookback:Epoch = 0) -> Epoch:
+def get_shard_period_start_epoch(epoch: Epoch, lookback:Epoch = 0) -> Epoch:
     return epoch - (epoch % PERSISTENT_COMMITTEE_PERIOD) - lookback * PERSISTENT_COMMITTEE_PERIOD
 ```
 
@@ -172,15 +189,6 @@ def get_period_committee(state: BeaconState,
     return [i for i in full_committee if active_and_balance_filter(i)][:MAX_PERSISTENT_COMMITTEE_SIZE]
 ```
 
-### `get_switchover_epoch`
-
-```python
-def get_switchover_epoch(state: BeaconState, epoch: Epoch, index: ValidatorIndex) -> int:
-    earlier_start_epoch = get_period_start_epoch(epoch, lookback=2)
-    return (bytes_to_int(hash(get_seed(state, earlier_start_epoch) + int_to_bytes(index, length=3)[0:8]))
-            % PERSISTENT_COMMITTEE_PERIOD)
-```
-
 ### `get_persistent_committee`
 
 ```python
@@ -198,8 +206,8 @@ def get_persistent_committee(state: BeaconState,
     # Take not-yet-cycled-out validators from earlier committee and already-cycled-in validators from
     # later committee; return a sorted list of the union of the two, deduplicated
     return sorted(set(
-        [i for i in earlier_committee if epoch % PERSISTENT_COMMITTEE_PERIOD < get_switchover_epoch(state, epoch, i)]
-        + [i for i in later_committee if epoch % PERSISTENT_COMMITTEE_PERIOD >= get_switchover_epoch(state, epoch, i)]
+        [i for i in earlier_committee if epoch % EPOCHS_PER_SHARD_PERIOD < i % EPOCHS_PER_SHARD_PERIOD]
+        + [i for i in later_committee if epoch % PERSISTENT_COMMITTEE_PERIOD >= i % EPOCHS_PER_SHARD_PERIOD]
     ))
 ```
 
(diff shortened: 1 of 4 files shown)


specs/core/1_shard-data-chains.md
```diff
@@ -327,53 +314,33 @@ def is_valid_shard_block(beacon_blocks: Sequence[BeaconBlock],
...
-        assert max(GENESIS_SHARD_SLOT, candidate.slot - SLOTS_PER_EPOCH) <= attestation.data.slot
-        assert attestation.data.slot <= candidate.slot - MIN_ATTESTATION_INCLUSION_DELAY
-        assert attestation.data.crosslink.shard == candidate.shard
-        verify_shard_attestation_signature(beacon_state, attestation)
-
-    # Check signature
-    proposer_index = get_shard_proposer_index(beacon_state, candidate.shard, candidate.slot)
+    persistent_committee = get_persistent_committee(beacon_state, block.shard, block.slot)
+    pubkeys = []
+    for i, index in enumerate(persistent_committee):
```
djrtwo: This is pulling from the entire persistent committee but the `attester_bitfield` is of size `SHARD_SLOT_COMMITTEE_SIZE`. 

This needs to be patched with some of the changes from #1276 to be complete/correct (specifically need `get_shard_block_attester_committee`)

specs/core/1_shard-data-chains.md
```diff
@@ -48,6 +57,8 @@ This document describes the shard data layer and the shard fork choice rule in P
 | - | - |
 | `BYTES_PER_SHARD_BLOCK_BODY` | `2**14` (= 16,384) |
 | `MAX_SHARD_ATTESTIONS` | `2**4` (= 16) |
+| `SHARD_SLOTS_PER_BEACON_SLOT` | `2**1` (= 2) |
+| `MAX_PERSISTENT_COMMITTEE_SIZE` | `2**5` (= 128) |
```
vbuterin:
```suggestion
| `MAX_PERSISTENT_COMMITTEE_SIZE` | `2**7` (= 128) |
```

djrtwo:
@vbuterin Should this be reviewed, tests fixed, and merged?

protolambda:
If we change `PLACEHOLDER`, we should also change the other `PLACEHOLDER` in custody spec, or rename this one to something more specific.

dankrad:
Is there a discussion on why we want to change it to one day? Isn't that worse for light clients?

vbuterin:
> Is there a discussion on why we want to change it to one day? Isn't that worse for light clients?

See the "implementation alternative" section here https://github.com/ethereum/eth2.0-specs/issues/1293

The shard period (I'm proposing to rename "persistent committee period" to "shard period" because the former is waaaay too long) goes down to 1 day, but the committee size drops to 128, so on net it's still a savings for light clients.

specs/core/1_shard-data-chains.md
```diff
@@ -75,207 +91,193 @@ The following types are defined, mapping into `DomainType` (little endian):
...
-            ) for block in blocks
-        ]))
-        + hash_tree_root(pad_to_power_of_2([
-            hash_tree_root_of_bytes(block.body) for block in blocks
-        ]))
-    )
+    header = b''.join([flatten_shard_header(get_shard_header(block)) for block in blocks])
+    footer = b''.join([block.core.data for block in blocks])
+    MAX_SIZE = SHARD_BLOCK_SIZE_LIMIT * SHARD_SLOTS_PER_BEACON_SLOT * SLOTS_PER_EPOCH * MAX_EPOCHS_PER_CROSSLINK
+    return hash_tree_root(pad(header + footer, MAX_SIZE))
```
vbuterin:
```suggestion
    return hash_tree_root(BytesN[MAX_SIZE](pad(header + footer, MAX_SIZE)))
```

specs/core/1_shard-data-chains.md
```diff
@@ -75,207 +91,193 @@ The following types are defined, mapping into `DomainType` (little endian):
...
 
     # Take not-yet-cycled-out validators from earlier committee and already-cycled-in validators from
     # later committee; return a sorted list of the union of the two, deduplicated
-    return sorted(list(set(
-        [i for i in earlier_committee if epoch % PERSISTENT_COMMITTEE_PERIOD < get_switchover_epoch(state, epoch, i)]
-        + [i for i in later_committee if epoch % PERSISTENT_COMMITTEE_PERIOD >= get_switchover_epoch(state, epoch, i)]
-    )))
+    return sorted(set(
+        [i for i in earlier_committee if epoch % EPOCHS_PER_SHARD_PERIOD < i % EPOCHS_PER_SHARD_PERIOD]
+        + [i for i in later_committee if epoch % PERSISTENT_COMMITTEE_PERIOD >= i % EPOCHS_PER_SHARD_PERIOD]
```
vbuterin:
```suggestion
        + [i for i in later_committee if epoch % EPOCHS_PER_SHARD_PERIOD >= i % EPOCHS_PER_SHARD_PERIOD]
```

specs/core/1_shard-data-chains.md
```diff
@@ -300,76 +302,52 @@ def is_valid_shard_block(beacon_blocks: Sequence[BeaconBlock],
...
-    beacon_block = beacon_blocks[candidate.slot]
-    assert candidate.beacon_block_root == signing_root(beacon_block)
-    assert beacon_block.slot <= candidate.slot
+    beacon_block = beacon_blocks[candidate.core.slot - candidate.core.slot % (SHARD_SLOTS_PER_BEACON_SLOT * SLOTS_PER_EPOCH)]
+    assert candidate.core.beacon_block_root == signing_root(beacon_block)
+    assert beacon_block.slot <= candidate.core.slot
 
     # Check state root
-    assert candidate.state_root == Hash()  # [to be removed in phase 2]
+    assert candidate.core.state_root == Hash()  # [to be removed in phase 2]
```
djrtwo: `state_root` represents the shard system state which will be around in phase 1.
We should remove this line

specs/core/1_shard-data-chains.md
```diff
@@ -300,76 +302,52 @@ def is_valid_shard_block(beacon_blocks: Sequence[BeaconBlock],
...
 
     # Check parent block
-    if candidate.slot == PHASE_1_FORK_SLOT:
-        assert candidate.parent_root == Hash()
+    if candidate.core.slot == PHASE_1_FORK_SLOT:
```
```suggestion
    if (
        candidate.core.slot % SHARD_SLOTS_PER_BEACON_SLOT == 0
        and compute_slot_from_shard_slot(candidate.core.slot) == PHASE_1_FORK_SLOT:
    )
```

djrtwo: Not sure this is actually correct. If the 0th slot of a shard chain is skipped, we need a way to recover

specs/core/1_shard-data-chains.md
```diff
@@ -300,76 +302,52 @@ def is_valid_shard_block(beacon_blocks: Sequence[BeaconBlock],
...
+        assert candidate.core.parent_root == Hash()
     else:
         parent_block = next(
-            (block for block in valid_shard_blocks if signing_root(block) == candidate.parent_root),
+            (block for block in valid_shard_blocks if hash_tree_root(block.core) == candidate.core.parent_root),
```
djrtwo: I think `signing_root(block)` should work just fine here

specs/core/1_shard-data-chains.md
djrtwo: Seems more like the conditional should be `if candidate.core.parent_root == Hash(): do nothing` rather than checking the slot

specs/core/1_shard-data-chains.md
```diff
@@ -300,76 +302,52 @@ def is_valid_shard_block(beacon_blocks: Sequence[BeaconBlock],
...
             None
         )
         assert parent_block is not None
-        assert parent_block.shard == candidate.shard
-        assert parent_block.slot < candidate.slot
-        assert signing_root(beacon_blocks[parent_block.slot]) == parent_block.beacon_chain_root
+        assert parent_block.core.slot < candidate.core.slot
+        assert signing_root(beacon_blocks[parent_block.core.slot]) == parent_block.core.beacon_chain_root
```
```suggestion
        assert signing_root(beacon_blocks[compute_slot_from_shard_slot(parent_block.core.slot)]) == parent_block.core.beacon_chain_root
```

djrtwo: This conditional statement should also be for the initial block in the shard chain but is skipped currently

specs/core/1_shard-data-chains.md
```diff
@@ -75,207 +91,191 @@ The following types are defined, mapping into `DomainType` (little endian):
 
 | Name | Value |
 | - | - |
-| `PLACEHOLDER` | `2**32` |
+| `PLACEHOLDER` | `2**3` |
```
protolambda:
```suggestion
```

specs/core/1_shard-data-chains.md
```diff
@@ -75,207 +91,191 @@ The following types are defined, mapping into `DomainType` (little endian):
...
+    slot: ShardSlot
     beacon_chain_root: Hash
     parent_root: Hash
-    body_root: Hash
+    data_root: Bytes[SHARD_BLOCK_SIZE_LIMIT - SHARD_HEADER_SIZE]
     state_root: Hash
-    attestations: List[ShardAttestation, PLACEHOLDER]
-    signature: BLSSignature
+    total_bytes: uint64
+    attester_bitfield: Bitvector[MAX_PERSISTENT_COMMITTEE_SIZE * 2]
```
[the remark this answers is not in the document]
vbuterin: * 2 is because a persistent committee is taken from parts of two adjacent period committees, and in a hypothetical worst case it ends up being twice the size of a period committee.

specs/core/1_shard-data-chains.md
```diff
@@ -75,207 +91,191 @@ The following types are defined, mapping into `DomainType` (little endian):
...
 ```
 
-### `ShardBlockHeader`
+### `ExtendedShardBlockCore`
 
 ```python
-class ShardBlockHeader(Container):
-    slot: Slot
-    shard: Shard
```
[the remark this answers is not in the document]
vbuterin: I can see that; I'd be ok with putting it back in (though we'd have to take it back out if we decide we want to add another slot as then we would exceed 8; or we could add slot offsets for different shards :smile:)

specs/core/1_shard-data-chains.md
```diff
@@ -75,207 +91,191 @@ The following types are defined, mapping into `DomainType` (little endian):
...
+    later_committee = get_period_committee(state, get_period_start_epoch(epoch, lookback=1), shard)
 
     # Take not-yet-cycled-out validators from earlier committee and already-cycled-in validators from
     # later committee; return a sorted list of the union of the two, deduplicated
-    return sorted(list(set(
-        [i for i in earlier_committee if epoch % PERSISTENT_COMMITTEE_PERIOD < get_switchover_epoch(state, epoch, i)]
-        + [i for i in later_committee if epoch % PERSISTENT_COMMITTEE_PERIOD >= get_switchover_epoch(state, epoch, i)]
-    )))
+    return sorted(set(
+        [i for i in earlier_committee if epoch % EPOCHS_PER_SHARD_PERIOD < i % EPOCHS_PER_SHARD_PERIOD]
```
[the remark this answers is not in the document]
vbuterin: The point is specifically to use the validator index. It mirrors how staggering currently works for the proof of custody, where transition epoch for a validator equals validator index mod committee length.

specs/core/1_shard-data-chains.md
```diff
@@ -75,207 +90,193 @@ The following types are defined, mapping into `DomainType` (little endian):
...
-
-    # No block can be proposed if no validator is active
-    return None
+    i = 0
+    while True:
+        candidate_index = persistent_committee[(slot + i) % len(persistent_committee)]
+        random_byte = hash(seed + int_to_bytes(i // 32, length=8))[i % 32]
+        effective_balance = state.validators[candidate_index].effective_balance
+        if effective_balance * MAX_RANDOM_BYTE >= MAX_EFFECTIVE_BALANCE * random_byte:
+            return ValidatorIndex(candidate_index)
```
vbuterin:
```suggestion
            if is_active_validator(state.validators[candidate_index], shard_slot_to_epoch(slot)):
                return ValidatorIndex(candidate_index)
```

specs/core/1_shard-data-chains.md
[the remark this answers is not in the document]
djrtwo: We can do a select from each that prevents ever hitting greater than `MAX_PERSISTENT_COMMITTEE_SIZE`

specs/core/1_shard-data-chains.md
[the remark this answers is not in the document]
djrtwo: Doesn't it make more since to take a simple slice of each to ensure that we actually have the property of committee being less than or equal to `MAX_PERSISTENT_COMMITTEE_SIZE`?

specs/core/1_shard-data-chains.md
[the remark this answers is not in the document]
vbuterin: The simplest way to do that would probably be to just do a simple crop (if the committee exceeds that size, chop off the last few members). I could see it being reasonable to do that.

(The approach you suggested would I think break the invariant that a validator cannot be a member of multiple shard committees at the same time)

specs/core/1_shard-data-chains.md
vbuterin: Agree! But I'm deliberately leaving this as is so that we can add the system state in a separate PR.

specs/core/1_shard-data-chains.md
vbuterin: True, but `hash_tree_root(block.core)` makes it clearer what's going on :)

specs/core/1_shard-data-chains.md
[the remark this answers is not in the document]
djrtwo: I see now. We can't have a simple 1-in/1-out from the `earlier` and `later` committee because we might have someone assigned to multiple committees at once. Instead need a private switchover epoch per validator index.

Thinking about alternatives..

specs/core/1_shard-data-chains.md
```diff
@@ -75,207 +91,204 @@ The following types are defined, mapping into `DomainType` (little endian):
...
 ```python
-class ShardBlockHeader(Container):
-    slot: Slot
-    shard: Shard
+class ExtendedShardBlockCore(Container):
+    slot: ShardSlot
     beacon_chain_root: Hash
     parent_root: Hash
-    body_root: Hash
+    data_root: Bytes[SHARD_BLOCK_SIZE_LIMIT - SHARD_HEADER_SIZE]
```
hwwhww:
```python
    data: Bytes[SHARD_BLOCK_SIZE_LIMIT - SHARD_HEADER_SIZE]
```

specs/core/1_shard-data-chains.md
```diff
@@ -75,207 +91,204 @@ The following types are defined, mapping into `DomainType` (little endian):
...
-        signature=attestation.aggregate_signature,
-        domain=get_domain(state, DOMAIN_SHARD_ATTESTER, compute_epoch_of_slot(data.slot))
+def pad(x: bytes, length: int) -> bytes:
+    assert len(x) <= length
+    return x + b'\x00' * (length - len(x))
+```
+
+### `flatten_shard_header`
+
+```pythonpad_
```
hwwhww:
```suggestion
```python
```

specs/core/1_shard-data-chains.md
```diff
@@ -75,207 +91,204 @@ The following types are defined, mapping into `DomainType` (little endian):
...
-    return None
+    i = 0
+    while True:
+        candidate_index = persistent_committee[(slot + i) % len(persistent_committee)]
+        random_byte = hash(seed + int_to_bytes(i // 32, length=8))[i % 32]
+        effective_balance = state.validators[candidate_index].effective_balance
+        if effective_balance * MAX_RANDOM_BYTE >= MAX_EFFECTIVE_BALANCE * random_byte:
+            if is_active_validator(state.validators[candidate_index], compute_epoch_of_shard_slot(slot)):
+                return ValidatorIndex(candidate_index)
+        i += 1
```
hwwhww: Edge case: Is it possible that none of the candidates is active? (It was handled previously)

specs/core/1_shard-data-chains.md
hwwhww: And the whole `### TODO PLACEHOLDER` section.

specs/core/1_shard-data-chains.md
vbuterin:
```suggestion
    data: Bytes[SHARD_BLOCK_SIZE_LIMIT - SHARD_HEADER_SIZE]
```

specs/core/1_shard-data-chains.md
vbuterin: So replace `while True:` with `for i in range(len(persistent_committee)):`, and then add a `return None` at the end?

hwwhww: @vbuterin sounds reasonable. :)

djrtwo: I'll handle this

djrtwo: opted for an exit condition prior to the loop. We want the loop to have the quality that if balances are low it can take multiple passes through the committee to eventually probabilistically choose one.

hwwhww: @djrtwo but ad30945b9b0cb63d00b0ea09fb0ccba9c6e35957 has an issue: it may return inactive validator. I think @vbuterin's suggestion is correct, and the client can *choose* to optimize it by checking if all the possible candidates are inactive before the while loop as you put.

specs/core/1_shard-data-chains.md
```diff
@@ -286,12 +286,14 @@ Let:
 
 - `beacon_blocks` be the `BeaconBlock` list such that `beacon_blocks[slot]` is the canonical `BeaconBlock` at slot `slot`
 - `beacon_state` be the canonical `BeaconState` after processing `beacon_blocks[-1]`
```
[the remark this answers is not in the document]
vbuterin: I'm going to recommend we not bother fixing this here, because this whole section will be reworked when we add shard state anyway.

specs/core/1_shard-data-chains.md
djrtwo: So we should exit if no active, and filter the validators for only active.
We want to be able to loop through the active validators more than once which @vbuterin's proposed solution doesn't give us

djrtwo: I think this is what we want -- https://github.com/ethereum/eth2.0-specs/pull/1294/commits/21f4e0334b60e217c6f1ac7948265448ab8665c0

vbuterin: Yep, looks good to me!

Issue #1269, opened 2019-07-04
Inconsistency between MIN_SEED_LOOKAHEAD and validator spec


michaelsproul:
The [validator spec](https://github.com/ethereum/eth2.0-specs/blob/v0.8.0/specs/validator/0_beacon-chain-validator.md#lookahead) states:

> The beacon chain shufflings are designed to provide a minimum of 1 epoch lookahead on the validator's upcoming committee assignments for attesting dictated by the shuffling and slot. Note that this lookahead does not apply to proposing, which must be checked during the epoch in question.

However, the shuffling for `current_epoch + 1` depends on the RANDAO mix for `current_epoch + 1 - MIN_SEED_LOOKAHEAD`, i.e. `current_epoch` with the lookahead set to 1 as it currently is. But this RANDAO mix for the current epoch changes every time a block is processed (as per [process_randao](https://github.com/ethereum/eth2.0-specs/blob/v0.8.0/specs/core/0_beacon-chain.md#randao)). Put another way: once a validator has transitioned to some epoch, they use the RANDAO mix from the previous epoch to determine their shuffling, and this RANDAO mix is changing right up until the last slot of the previous epoch.

In Lighthouse we were contemplating deleting our cache of the shuffling for the next epoch for this reason! See: https://github.com/sigp/lighthouse/issues/407

It would be good to have some clarity on whether the behaviour of the validator spec or of the beacon chain spec is what we really want. If the lookahead behaviour _is_ desired (which seems reasonable to me), I think we should update the beacon chain spec to account for the RANDAO mix changing in the current epoch.

djrtwo:
This looks like an off-by-one error. 

We should be getting the mix from `get_randao_mix(state, Epoch(epoch + EPOCHS_PER_HISTORICAL_VECTOR - 1 - MIN_SEED_LOOKAHEAD))` in `get_seed` (additional - 1) as the randao lookup. 

Need to either adjust the constant to `2` as you noted or add the `-1`. Need to think about the semantics a bit.

Off today through the weekend. Will pick this up monday

michaelsproul:
Yeah I think the additional `-1` is clearer than setting `MIN_SEED_LOOKAHEAD` to 2. Unless we change the name of the variable to make sense with the 2 value meaning "look at the randao from 2 epochs back".

djrtwo:
close via #1296

PR #1296, opened 2019-07-14
Lookahead fix

djrtwo:
address #1269

`current_epoch - 1` for randomness source doesn’t give us lookahead. It is just in time and it’s the bare minimum stable source of randao for any epoch. `MIN_SEED_LOOKAHEAD` of `0` should equate to the `current_epoch - 1` case. `MIN_SEED_LOOKAHEAD > 0` should have a base of the `-1`

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -865,7 +865,7 @@ def get_seed(state: BeaconState, epoch: Epoch) -> Hash:
     """
     Return the seed at ``epoch``.
     """
-    mix = get_randao_mix(state, Epoch(epoch + EPOCHS_PER_HISTORICAL_VECTOR - MIN_SEED_LOOKAHEAD))  # Avoid underflow
+    mix = get_randao_mix(state, Epoch(epoch + EPOCHS_PER_HISTORICAL_VECTOR - MIN_SEED_LOOKAHEAD - 1))  # Avoid underflow
     active_index_root = state.active_index_roots[epoch % EPOCHS_PER_HISTORICAL_VECTOR]
     return hash(mix + active_index_root + int_to_bytes(epoch, length=32))
 ```


protolambda:
Just because it is possibly the most important off-by-1 prone math in the spec, here's my take (and thanks @djrtwo for [rubber duck debugging](https://en.wikipedia.org/wiki/Rubber_duck_debugging) :joy: ):


- `process_slots`, from start `state.slot == x*SLOTS_PER_EPOCH - 1`, i.e. `current_epoch == x - 1`
- `(state.slot + 1) % SLOTS_PER_EPOCH == 0` -> epoch boundary detected, process epoch.
- `process_final_updates()` as last epoch sub transition
- `state.randao_mixes[next_epoch % EPOCHS_PER_HISTORICAL_VECTOR] = get_randao_mix(state, current_epoch)`, a.k.a. `state.randao_mixes[x % EPOCHS_PER_HISTORICAL_VECTOR] = get_randao_mix(state, x - 1)`
- epoch transition finishes
- slot transition finishes, slot is set to `x*SLOTS_PER_EPOCH`, now `current_epoch == x`

So transition looks ok at first: during an epoch `randao_mixes` is available up to the current epoch *as index*: `state.randao_mixes[current_epoch % EPOCHS_PER_HISTORICAL_VECTOR]` get the very latest known randao mix. __*But it changes with `process_randao` during the current_epoch.*__

`MIN_SEED_LOOKAHEAD = 1`

Without this PR:
- beacon proposer indices are looked up with `get_seed(state, current_epoch)`. This requires `state.randao_mixes[(current_epoch + EPOCHS_PER_HISTORICAL_VECTOR - MIN_SEED_LOOKAHEAD) % EPOCHS_PER_HISTORICAL_VECTOR]`, equivalent to `state.randao_mixes[previous_epoch % EPOCHS_PER_HISTORICAL_VECTOR]`. Which was initially set during the transition from `previous_epoch - 1` to `previous_epoch`. And is then modified during `previous_epoch`. Unpredictable till the very end of `previous_epoch`. But known during all of `current_epoch`.
- beacon proposer indices for the next epoch looked up with `get_seed(state, current_epoch + 1)`. This requires `state.randao_mixes[((current_epoch + 1) + EPOCHS_PER_HISTORICAL_VECTOR - MIN_SEED_LOOKAHEAD) % EPOCHS_PER_HISTORICAL_VECTOR]`, equivalent to `state.randao_mixes[current_epoch % EPOCHS_PER_HISTORICAL_VECTOR]`. Which was initially set during the transition from `current_epoch - 1` to `current_epoch`. And is then modified during `current_epoch`. Unpredictable till the very end of `current_epoch`. But known during all of `current_epoch + 1`.

With this PR:
offset lookups down by 1.

Next epoch proposer index lookup is now requiring: `state.randao_mixes[((current_epoch + 1) + EPOCHS_PER_HISTORICAL_VECTOR - MIN_SEED_LOOKAHEAD - 1) % EPOCHS_PER_HISTORICAL_VECTOR]`, a.k.a. `state.randao_mixes[previous_epoch % EPOCHS_PER_HISTORICAL_VECTOR]`. Which was initially set during the transition from `previous_epoch - 1` to `previous_epoch`. And is then modified during `previous_epoch`. Unpredictable till the very end of `previous_epoch`. But known during all of `current_epoch`.

So yes, we need this PR to make the next epoch predictable during all of the current epoch.

PR #1298, opened 2019-07-15
Fix the `ShardBlock` fields, `data` -> `body` and  `Bytes[PLACEHOLDER]` -> `BytesN[PLACEHOLDER]`

hwwhww:


specs/core/1_shard-data-chains.md
```diff
@@ -106,7 +106,7 @@ class ShardBlock(Container):
     shard: Shard
     beacon_chain_root: Hash
     parent_root: Hash
-    data: ShardBlockBody
+    body: ShardBlockBody
```
vbuterin: We should discuss just removing the one-item `ShardBlockBody` struct and making it directly be a `Bytes` during the call :smile:

djrtwo: probably reasonable..

djrtwo:
@hwwhww Did we make a call on the `ShardBlockBody` struct?

hwwhww:
The shard block rework will be included in #1294.

PR #1300, opened 2019-07-16
[hotfix] bump `py_ecc==1.7.1`

hwwhww:
### Issue
The new release of `py_ecc` includes #1246 changes and its not backward compatible, that's why pyspec is broken with the current dependency `py_ecc>=1.6.0`.

### Proposing solution
1. Pin `py_ecc==1.7.1`
2. Update both pyspec and BLS test generator's `domain` type.
3. Minor refactoring of BLS test generator

@djrtwo it needs to backport to `dev` branch soon.

PR #1301, opened 2019-07-16
[hothotfix] Pin `py_ecc==1.7.0`

hwwhww:
While #1300 is in reviewing, this PR simply pins to the earlier version of py_ecc.

p.s. CI looks fine due to cache... 😬


djrtwo:
closed in favor of #1300 and #1302

PR #1305, opened 2019-07-19
Legendre custody construction

dankrad:
This implements the Legendre proof of custody construction according to the attached document. [legendre_proof_of_custody_minimal_protected.pdf](https://github.com/ethereum/eth2.0-specs/files/3413092/legendre_proof_of_custody_minimal_protected.pdf)

Also fixes some issues in the custody game spec, and adds some initial tests.


specs/core/0_beacon-chain.md
```diff
@@ -1661,6 +1661,11 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
         proposer_index=get_beacon_proposer_index(state),
     )
 
+    # Check bitlist lengths
+    committee_size = get_committee_count(state, attestation.data.target.epoch)
+    assert len(attestation.aggregation_bits) == committee_size
+    assert len(attestation.custody_bits) == committee_size
```
dankrad: I think these lines were missing. While there is a test that was purportedly testing inconsistent bit tests (`test_inconsistent_bits` in `test_process_attestation.py`), it actually also sets some of the custody bits to one, so it actually triggered the `if not len(bit_1_indices) == 0:` check that should not occur in phase 1. So I think these checks should be added.

specs/core/1_custody-game.md
```diff
@@ -263,45 +274,99 @@ class BeaconBlockBody(Container):
...
 ```python
 def ceillog2(x: uint64) -> int:
-    return x.bit_length()
+    return (x - 1).bit_length()
+```
+
+### `is_valid_merkle_branch_with_mixin`
+
+```python
+def is_valid_merkle_branch_with_mixin(leaf: Hash,
```
dankrad: I noticed we need this function to check the Merkle proofs for Bitlists (and other variable types). Wasn't sure whether to put this as a modification in phase 0 (I think that would also be valid as it is such a small change and won't affect the phase 0 logic).

specs/core/1_custody-game.md
```diff
@@ -533,7 +598,10 @@ def process_bit_challenge(state: BeaconState, challenge: CustodyBitChallenge) ->
     assert is_valid_indexed_attestation(state, get_indexed_attestation(state, attestation))
     # Verify attestation is eligible for challenging
     responder = state.validators[challenge.responder_index]
-    assert epoch + responder.max_reveal_lateness <= get_reveal_period(state, challenge.responder_index)
+    assert get_current_epoch(state) <= get_randao_epoch_for_custody_period(
+        get_reveal_period(state, challenge.responder_index, epoch) + 1, 
+        challenge.responder_index
+    ) + responder.max_reveal_lateness
```
dankrad: The original assertion was obviously broke -- there is an epoch on the LHS and a period on the RHS. I tried to convert it into something that makes sense, but nor sure if this is what we want. The new assertion will allow challenges for one full custody period after the current one plus the max reveal lateness.

specs/core/1_custody-game.md
```diff
@@ -263,45 +275,99 @@ class BeaconBlockBody(Container):
...
+Given one proof of custody chunk, returns the proof of custody subchunks of the correct sizes.
+
+```python
+def custody_subchunkify(x: bytes) -> list:
+    subchunks = []
+    start = 0
+    for size in BYTES_PER_CUSTODY_SUBCHUNK:
+        subchunks.append(x[start:start + size])
+        start += size
+    return subchunks
```
vbuterin: Why not just zero-byte-pad the last chunk so that outputs a clean `Sequence[Bytes48]`?

specs/core/1_custody-game.md
```diff
@@ -263,45 +275,99 @@ class BeaconBlockBody(Container):
...
 ```python
 def get_custody_chunk_bit(key: BLSSignature, chunk: bytes) -> bool:
-    # TODO: Replace with something MPC-friendly, e.g. the Legendre symbol
-    return bool(get_bit(hash(key + chunk), 0))
+    full_G2_element = bls_signature_to_G2(key)
+    s = full_G2_element[0].coeffs
+    bits = [legendre_bit((i + 1) * s[i % 2] + int.from_bytes(subchunk, "little"), BLS12_381_Q)
+            for i, subchunk in enumerate(custody_subchunkify(chunk))]
+
+    return bool(sum(bits) % 2)
```
vbuterin: Cryptography looks good here. For a moment I thought it was broken because you could "Factor out" the (i+1) and (s[i+2]) and/or combine all the subchunks together, but I saw that you're adding the chunk and the key rather than multiplying, and I don't think there's additive properties of Legendre that could be attacked here.

dankrad: So you can actually multiply all the subchunks together, which I consider a feature -- doing that improves computation time by a factor of 3 in an MPC (and I think actually a similar amount in a cleartext computation). I think it's not a problem because for oursourcing, the outsourcer would still have to give away a multivariate polynomial of which the roots are his secrets `s[0]` and `s[1]`.

specs/core/1_custody-game.md
dankrad: Yeah, that sounds more elegant! I'm going to change it.

specs/core/0_beacon-chain.md
dankrad: (just noticed there is an issue about this #1262, might be better to fix this in a separate PR)

specs/core/1_custody-game.md
```diff
@@ -263,45 +275,99 @@ class BeaconBlockBody(Container):
...
-    aggregated_bits = bytearray([0] * 32)
-    for i in range(0, len(chunk_bits), 32):
-        for j in range(32):
-            aggregated_bits[j] ^= chunk_bits[i + j]
-    return hash(aggregated_bits)
+def get_chunk_bits_root(chunk_bits: Bitlist[MAX_CUSTODY_CHUNKS]) -> bit:
+    aggregated_bits = 0
+    for i, b in enumerate(chunk_bits):
+        aggregated_bits += 2**i * b
+    return legendre_bit(aggregated_bits, BLS12_381_Q)
```
vbuterin: This function uses bigints (and even unlimited-size ones), which have at times been viewed as a no-no. Is this a problem?

specs/core/1_custody-game.md
vbuterin: I think we were thinking about modifying the proof of custody data to be a vector rather than a list, so the mixin is unnecessary. CC @JustinDrake. Really nailing down the shard chain details is a dependency of this.

specs/core/1_custody-game.md
dankrad: True. Since all the computations are mod `q`, 48 bytes/384 bits would be enough, so we could add that size (or 512 bits) as an allowed int size. The Legendre bit computation function obviously needs it as well. I think rewriting this logic using only 256 bit integers would be a very unelegant solution to this problem.

specs/core/1_custody-game.md
dankrad: OK, so I will leave this as is for now and then if we are not going to need any length mixins, we can remove it.

specs/core/0_beacon-chain.md
djrtwo: I have a PR coming that handles this in pretty much the same way and had some additional tests. Would like to get it merged as a separate PR in `v08x`

dankrad: Yes, sounds sensible. I'll keep the lines here until that's merged because otherwise it'll fail the tests :D

djrtwo: addressed in #1317

specs/core/1_custody-game.md
dankrad: Actually, given that the modulus is 381 bits, it needs 768 bit integers to store intermediate (pre-reduction) values. I think the case for Bigints is actually strong here (and I would suggest all implementers use an actual Bignum library like GMP).

PR #1310, opened 2019-07-22
simplifies `is_valid_indexed_attestation` a little

decanus:


djrtwo:
A bit torn, but I think I prefer to stay with the `if not condition: return False` pattern through the entire function

PR #1313, opened 2019-07-24
Validate state root for fork choice on_block

terencechain:
I think we do want to validate state root for `state_transition` in `on_block`. The default argument for `validate_state_root` is false

### DIFF
--- a/specs/core/0_fork-choice.md
+++ b/specs/core/0_fork-choice.md
@@ -154,7 +154,7 @@ def on_block(store: Store, block: BeaconBlock) -> None:
     # Check that block is later than the finalized epoch slot
     assert block.slot > compute_start_slot_of_epoch(store.finalized_checkpoint.epoch)
     # Check the block is valid and compute the post-state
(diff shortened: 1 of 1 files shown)


specs/core/0_fork-choice.md
```diff
@@ -154,7 +154,7 @@ def on_block(store: Store, block: BeaconBlock) -> None:
     # Check that block is later than the finalized epoch slot
     assert block.slot > compute_start_slot_of_epoch(store.finalized_checkpoint.epoch)
     # Check the block is valid and compute the post-state
-    state = state_transition(pre_state, block)
+    state = state_transition(pre_state, block, true)
```
djrtwo:
```suggestion
    state = state_transition(pre_state, block, True)
```

terencechain:
Darn it, test failed at validating state root...

djrtwo:
hmm, probably not generating the blocks fully  with correct state root in testing. will take a look

PR #1317, opened 2019-07-25
Check attestation bit lengths

### LINKED ISSUE #1262: Check attestation bit list length against committee size

djrtwo:
resolves #1262 
replaces #1264 

Note that the bitlists with too _few_ bits had to be particular in it's signed and then modified to keep the signing helper from complaining

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1655,8 +1655,7 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
     assert attestation_slot + MIN_ATTESTATION_INCLUSION_DELAY <= state.slot <= attestation_slot + SLOTS_PER_EPOCH
 
     committee = get_crosslink_committee(state, data.target.epoch, data.crosslink.shard)
-    assert len(attestation.aggregation_bits) == len(committee)
-    assert len(attestation.custody_bits) == len(committee)
+    assert len(attestation.aggregation_bits) == len(attestation.custody_bits) == len(committee)
 
     pending_attestation = PendingAttestation(
         data=data,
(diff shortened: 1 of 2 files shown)


djrtwo:
I find two lines more readable and more directly translatable to languages without magic syntax

specs/core/0_beacon-chain.md
```diff
@@ -1654,6 +1654,9 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
     attestation_slot = get_attestation_data_slot(state, data)
     assert attestation_slot + MIN_ATTESTATION_INCLUSION_DELAY <= state.slot <= attestation_slot + SLOTS_PER_EPOCH
 
+    committee = get_crosslink_committee(state, data.target.epoch, data.crosslink.shard)
```
protolambda: Looks good, too bad there is no committee size function. We do not actually need the contents, and the committee size can be derived from total active indices count together with `data.crosslink.shard` (for rounding in committee size).

test_libs/pyspec/eth2spec/test/phase_0/block_processing/test_process_attestation.py
```diff
@@ -398,3 +398,61 @@ def test_empty_aggregation_bits(spec, state):
...
+
+@with_all_phases
+@spec_state_test
+def test_too_many_aggregation_bits(spec, state):
+    attestation = get_valid_attestation(spec, state, signed=True)
+    state.slot += spec.MIN_ATTESTATION_INCLUSION_DELAY
+
+    # one too many bits
+    attestation.aggregation_bits.append(0b0)
+
```
protolambda: Good to see tests. I was wondering about the signature correctness at my first look, because of the change in data after signing, but the aggregation bits don't influence it, so it is all fine.

specs/core/0_beacon-chain.md
djrtwo: indeed... hrmmm probably not worth adding for sake of freezy-ness

PR #1318, opened 2019-07-25
Proposed refactor of compact committees

vbuterin:
NOT substantive.

Goal is to abstract out the committee to compact committee transformation to allow it to be used for persistent committees as well.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -88,6 +88,7 @@
             - [`get_shard_delta`](#get_shard_delta)
             - [`get_beacon_proposer_index`](#get_beacon_proposer_index)
             - [`get_attestation_data_slot`](#get_attestation_data_slot)
+            - [`committee_to_compact_committee`](#committee_to_compact_committee)
             - [`get_compact_committees_root`](#get_compact_committees_root)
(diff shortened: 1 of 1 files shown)


JustinDrake:
Did we come to the conclusion that crosslink committee roots (as well as active validator indices) are not required? Lots of saved hashing.

vbuterin:
I think so! In which case the right thing to do would be to remove the compact committee logic entirely and move it into the phase 1 spec?

Also, what's the appropriate place to put beacon chain changes in the phase 1 spec that are NOT proof of custody related? (as eg. the persistent committee roots would be)

JustinDrake:
> the right thing to do would be to remove the compact committee logic entirely and move it into the phase 1 spec

I think so. And making light clients clearly fall into phase 1 is a nice conceptual cleanup :)

> what's the appropriate place to put beacon chain changes in the phase 1 spec that are NOT proof of custody related?

As suggested by @protolambda [here](https://github.com/ethereum/eth2.0-specs/issues/1274#issuecomment-510012929) we can have `1_misc-beacon-updates.md` complementing `1_custody-game.md`. This document would add light client infrastructure, transfers, and make miscellaneous changes like removing `assert data.crosslink.data_root == Bytes32()`.

PR #1325, opened 2019-07-29
Cap size of persistent committee to MAX_PERSISTENT_COMMITTEE_SIZE

vbuterin:
Note: does NOT change the lines

```
    for i in range(len(attester_committee), MAX_PERSISTENT_COMMITTEE_SIZE * 2):
        assert block.attester_bitfield[i] is False
```

As another PR will replace that section with a state transition function anyway.

### DIFF
--- a/specs/core/1_shard-data-chains.md
+++ b/specs/core/1_shard-data-chains.md
@@ -132,7 +132,7 @@ class ShardBlockCore(Container):
     data_root: Hash
     state_root: Hash
     total_bytes: uint64
-    attester_bitfield: Bitvector[MAX_PERSISTENT_COMMITTEE_SIZE * 2]
+    attester_bitfield: Bitvector[MAX_PERSISTENT_COMMITTEE_SIZE]
 ```
 
 ### `ExtendedShardBlockCore`
@@ -145,7 +145,7 @@ class ExtendedShardBlockCore(Container):
     data: Bytes[SHARD_BLOCK_SIZE_LIMIT - SHARD_HEADER_SIZE]
     state_root: Hash
     total_bytes: uint64
-    attester_bitfield: Bitvector[MAX_PERSISTENT_COMMITTEE_SIZE * 2]
+    attester_bitfield: Bitvector[MAX_PERSISTENT_COMMITTEE_SIZE]
 ```
 
 ## Helper functions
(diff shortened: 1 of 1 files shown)


specs/core/1_shard-data-chains.md
```diff
@@ -205,11 +205,15 @@ def get_persistent_committee(state: BeaconState,
...
-        [i for i in earlier_committee if epoch % EPOCHS_PER_SHARD_PERIOD < i % EPOCHS_PER_SHARD_PERIOD]
-        + [i for i in later_committee if epoch % EPOCHS_PER_SHARD_PERIOD >= i % EPOCHS_PER_SHARD_PERIOD]
-    ))
+    # later committee; return a sorted list of the union of the two, deduplicated. If the combined size is
+    # too large, take out the validators with the earlier induction epochs (ie. delay their induction)
+    prev_subset = [i for i in earlier_committee if epoch % EPOCHS_PER_SHARD_PERIOD < i % EPOCHS_PER_SHARD_PERIOD]
+
+    validators_from_post = MAX_PERSISTENT_COMMITTEE_SIZE - len(prev_subset)
+    post_subset = [i for i in later_committee if epoch % EPOCHS_PER_SHARD_PERIOD >= i % EPOCHS_PER_SHARD_PERIOD]
+    post_subset = sorted(post_subset, key=lambda i: -(i % EPOCHS_PER_SHARD_PERIOD))[:validators_from_post]
```
djrtwo: This seems backwards and might break the continuity in validating a shard requiremnt.

djrtwo: I'm also a bit concerned that validators with certain values wrt `% EPOCHS_PER_SHARD_PERIOD` might be more or less likely to be trimmed from the `post_subset` and have asymmetries over time in total validation time participation.

vbuterin: I think the simplest approach may well be not bothering capping the persistent committee size and letting the theoretical max size be `2 * MAX_PERSISTENT_COMMITTEE_SIZE`.

djrtwo: Fair. Thoughts on the name? 
Maybe `TARGET_PERSISTENT_COMMITTEE_SIZE`?

vbuterin: Sure, sounds good to me!

djrtwo:
close in favor of #1358

PR #1327, opened 2019-07-30
Add attestation test with extra aggregation bits

michaelsproul:
The spec allows the aggregation bits of an attestation to contain true bits beyond the size of the committee, as this test confirms.

We previously had code in Lighthouse to forbid this case, so I wrote this test to confirm the spec's behaviour and figured it was an edge case worth testing.


djrtwo:
We actually added back the bounds conditions on attestations in branch `v08x` to be released with a couple of other minor changes.

https://github.com/ethereum/eth2.0-specs/commit/bd69dc7e3cb3db2ba74b9b39afb5f572b30f0d73

michaelsproul:
Ok, sweet!

PR #1328, opened 2019-08-01
Ethereum 2.0 Networking Specification

arnetheduck:
This PR consolidates the work done by various teams and community towards a unified Ethereum 2.0 networking specification. In particular, this it consolidates the `RPC-Interface.md` and the `libp2p-standardization.md` documents into a single unified libp2p-based network specification.

Highlights of the PR include:

* Concrete libp2p protocol selection for interop and mainnet, including transports, protocol negotiation and payload encoding
* Broadcast for beacon blocks, attestations as well as a simplified request/response protocol for block recovery
* Design rationale section summarising past discussions

This PR is the result of a collaborative effort between the Lighthouse, Nimbus and Protocol Labs teams.

It is a continuation of, and replaces #1281.


specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,712 @@
+# Overview
+
+This document contains the network specification for Ethereum 2.0 clients.
+
+It consists of four main sections:
+
+1. A specification of the network fundamentals detailing the two network configurations: interoperability test network, and mainnet launch.
+2. A specification of the three network interaction _domains_ of ETH2.0: (a) the gossip domain, (b) the discovery domain, \(c\) the Req/Resp domain.
```
Mikerah: Instead of Req/Resp domain, maybe change it to RPC domain?

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,712 @@
...
+- `D_high` (topic stable mesh high watermark): 12
+- `D_lazy` (gossip target): 6
+- `fanout_ttl` (ttl for fanout maps for topics we are not subscribed to but have published to, seconds): 60
+- `gossip_advertise` (number of windows to gossip about): 3
+- `gossip_history` (number of heartbeat intervals to retain message IDs): 5
+- `heartbeat_interval` (frequency of heartbeat, seconds): 1
+
+### Topics
+
+Topics are plain UTF-8 strings, and are encoded on the wire as determined by protobuf (gossipsub messages are enveloped in protobuf messages).
```
Mikerah: Is there a rationale for storing topics as plain UTF-8 strings instead of topic hashes?

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,712 @@
...
+
+- `ssz_snappy` - All objects are ssz-encoded and then compressed with snappy. Example: The beacon attestation topic string is: `/beacon_attestation/ssz_snappy` and the data field of a gossipsub message is an `Attestation` that has been ssz-encoded then compressed with snappy.
+
+Implementations MUST use a single encoding. Changing an encoding will require coordination between participating implementations.
+
+## The discovery domain: discv5
+
+Discovery Version 5 ([discv5](https://github.com/ethereum/devp2p/blob/master/discv5/discv5.md)) is used for peer discovery, both in the interoperability testnet and mainnet.
+
+`discv5` is a standalone protocol, running on UDP on a dedicated port, meant for peer discovery only. `discv5` supports self-certified, flexible peer records (ENRs) and topic-based advertisement, both of which are (or will be) requirements in this context.
```
Mikerah: We should define what this dedicated port is ahead of time. Maybe have different ports for interop and mainnet

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,712 @@
...
+
+- `ssz` - All objects are SSZ-encoded. Example: The beacon block topic string is: `/beacon_block/ssz` and the data field of a gossipsub message is an ssz-encoded `BeaconBlock`.
+
+#### Mainnet
+
+- `ssz_snappy` - All objects are ssz-encoded and then compressed with snappy. Example: The beacon attestation topic string is: `/beacon_attestation/ssz_snappy` and the data field of a gossipsub message is an `Attestation` that has been ssz-encoded then compressed with snappy.
+
+Implementations MUST use a single encoding. Changing an encoding will require coordination between participating implementations.
+
+## The discovery domain: discv5
```
Mikerah: I know this has been discussed in the past but just to try to get a definitive answer: the current discv5 spec uses RLP and the current eth2 networking spec depends on SSZ. Should we simply choose 1 serialization format for all networking or keep both?

raulk: Alternatively, the ENR spec can be extended to support multiple serialisation formats, e.g. by pretending a prefix that identifies the encoding.

specs/networking/p2p-interface.md
raulk: Not necessary to define fixed ports. You only need to know the details of the bootstrappers. Past that, nodes announce their UDP ports in their self-certified ENR.

specs/networking/p2p-interface.md
raulk: Good point. Yes. I will add the response to the rationale section and crosslink the commit here.

specs/networking/p2p-interface.md
Mikerah: Ahh right! ENRs have a field for both TCP and UDP ports.

specs/networking/p2p-interface.md
raulk: Actually, it was already there. But I took the opportunity to enhance the answer anyway: https://github.com/status-im/eth2.0-specs/commit/1ba03b4c33a6ade51dbd60dae9a6a15acd7be531

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+#### Interop
+
+All implementations MUST support the TCP libp2p transport, and it MUST be enabled for both dialing and listening (i.e. outbound and inbound connections).
+
+The libp2p TCP transport supports listening on IPv4 and IPv6 addresses (and on multiple simultaneously). Clients SHOULD allow the operator to configure the listen IP addresses and ports, including the addressing schemes (IPv4, IPv6).
+
+To facilitate connectivity, and avert possible IPv6 routability/support issues, clients participating in the interoperability testnet MUST expose at least ONE IPv4 endpoint.
+
+All listening endpoints must be publicly dialable, and thus not rely on libp2p circuit relay, AutoNAT or AutoRelay facilities.
```
prestonvanloon: Why not? Is there some major blocker to dialing a node via circuit relay? 

We utilize circuit relay for our testnet orchestration, but we can hardcode inbound routes to meet this requirement if needed.

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+Clients MUST support [mplex](https://github.com/libp2p/specs/tree/master/mplex) and MAY support [yamux](https://github.com/hashicorp/yamux/blob/master/spec.md). If both are supported by the client, yamux must take precedence during negotiation. See the Rationale section of this document for tradeoffs.
+
+# ETH2 network interaction domains
+
+## Constants
+
+This section outlines constants that are used in this spec.
+
+- `RQRP_MAX_SIZE`: The max size of uncompressed req/resp messages that clients will allow.
```
prestonvanloon: nit: what is RQRP? Can this be MSG_MAX_SIZE?

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+**Protocol ID:** `/meshsub/1.0.0`
+
+**Gossipsub Parameters**
+
+*Note: Parameters listed here are subject to a large-scale network feasibility study.*
+
+The following gossipsub parameters will be used:
+
+- `D` (topic stable mesh target count): 6
```
prestonvanloon: Why is this called `D`? Other parameters have well formed names, can these have full names as well?

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+Additional topics are used to propagate lower frequency validator messages. Their `TopicName`’s are:
+
+- `voluntary_exit` - This topic is used solely for propagating voluntary validator exits to proposers on the network. Voluntary exits are sent in their entirety. Clients who receive a voluntary exit on this topic MUST validate the conditions within `process_voluntary_exit` before forwarding it across the network.
+- `proposer_slashing` - This topic is used solely for propagating proposer slashings to proposers on the network. Proposer slashings are sent in their entirety. Clients who receive a proposer slashing on this topic MUST validate the conditions within `process_proposer_slashing` before forwarding it across the network.
+- `attester_slashing` - This topic is used solely for propagating attester slashings to proposers on the network. Attester slashings are sent in their entirety. Clients who receive an attester slashing on this topic MUST validate the conditions within `process_attester_slashing` before forwarding it across the network.
+
+#### Interop
+
+Unaggregated attestations from all shards are sent to the `beacon_attestation` topic.
```
prestonvanloon: What about aggregated attestations? They also go to `beacon_attestation` topic in interop config?

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+#### Interop
+
+- `ssz` - All objects are SSZ-encoded. Example: The beacon block topic string is: `/beacon_block/ssz` and the data field of a gossipsub message is an ssz-encoded `BeaconBlock`.
+
+#### Mainnet
+
+- `ssz_snappy` - All objects are ssz-encoded and then compressed with snappy. Example: The beacon attestation topic string is: `/beacon_attestation/ssz_snappy` and the data field of a gossipsub message is an `Attestation` that has been ssz-encoded then compressed with snappy.
+
+Implementations MUST use a single encoding. Changing an encoding will require coordination between participating implementations.
```
prestonvanloon: Can't we handle encoding support in protocol negotiation? Seems like it would be a good idea to require clients to support ssz snappy, but allow them to support other encoding if they choose.

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+Two multiplexers are commonplace in libp2p implementations: [mplex](https://github.com/libp2p/specs/tree/master/mplex) and [yamux](https://github.com/hashicorp/yamux/blob/master/spec.md). Their protocol IDs are, respectively: `/mplex/6.7.0` and `/yamux/1.0.0`.
+
+Clients MUST support [mplex](https://github.com/libp2p/specs/tree/master/mplex) and MAY support [yamux](https://github.com/hashicorp/yamux/blob/master/spec.md). If both are supported by the client, yamux must take precedence during negotiation. See the Rationale section of this document for tradeoffs.
+
+# ETH2 network interaction domains
+
+## Constants
+
+This section outlines constants that are used in this spec.
```
prestonvanloon: Please add timeout constants in this section as well. TTFB timeout and response timeout.

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+If steps (1), (2) or (3) fail due to invalid, malformed or inconsistent data, the responder MUST respond in error. Clients tracking peer reputation MAY record such failures, as well as unexpected events, e.g. early stream resets.
+
+The entire request should be read in no more than **5 seconds**. Upon a timeout, the responder SHOULD reset the stream.
+
+The responder SHOULD send a response promptly, starting with a **single-byte** response code which determines the contents of the response (`result` particle in the BNF grammar above).
+
+It can have one of the following values:
+
+-  0: **Success** -- a normal response follows, with contents matching the expected message schema and encoding specified in the request.
```
prestonvanloon: Zero is a default value in some languages, can we set 0 to be UNKNOWN or forbid 0?

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+2. If there are any length assertions for length `N`, it should read exactly `N` bytes from the stream, at which point an EOF should arise (no more bytes). Should this is not the case, it should be treated as a failure.
+3. Deserialize the expected type, and process the request.
+4. Write the response (result, optional header, payload).
+5. Close their write side of the stream. At this point, the stream will be fully closed.
+
+If steps (1), (2) or (3) fail due to invalid, malformed or inconsistent data, the responder MUST respond in error. Clients tracking peer reputation MAY record such failures, as well as unexpected events, e.g. early stream resets.
+
+The entire request should be read in no more than **5 seconds**. Upon a timeout, the responder SHOULD reset the stream.
+
+The responder SHOULD send a response promptly, starting with a **single-byte** response code which determines the contents of the response (`result` particle in the BNF grammar above).
```
prestonvanloon: We can assume this byte represents an unsigned integer, but can you clarify it explicitly?

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+```
+
+Response Content:
+```
+(
+  blocks: []BeaconBlock
+)
+```
+
+Requests count beacon blocks from the peer starting from `start_slot` on the chain defined by `head_block_root`. The response MUST contain no more than count blocks. step defines the slot increment between blocks. For example, requesting blocks starting at `start_slot` 2 with a step value of 2 would return the blocks at [2, 4, 6, …]. In cases where a slot is empty for a given slot number, no block is returned. For example, if slot 4 were empty in the previous example, the returned array would contain [2, 6, …]. A step value of 1 returns all blocks on the range `[start_slot, start_slot + count)`.
```
prestonvanloon:
```suggestion
Requests count beacon blocks from the peer starting from `start_slot` on the chain defined by `head_block_root`. The response MUST contain no more than count blocks. `step` defines the slot increment between blocks. For example, requesting blocks starting at `start_slot` 2 with a step value of 2 would return the blocks at [2, 4, 6, …]. In cases where a slot is empty for a given slot number, no block is returned. For example, if slot 4 were empty in the previous example, the returned array would contain [2, 6, …]. A step value of 1 returns all blocks on the range `[start_slot, start_slot + count)`.
```

Added back ticks around `step`

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+  blocks: []BeaconBlock
+)
+```
+
+Requests blocks by their block roots. The response is a list of `BeaconBlock` with the same length as the request. Blocks are returned in order of the request and any missing/unknown blocks are left empty (SSZ null `BeaconBlock`).
+
+`RecentBeaconBlocks` is primarily used to recover recent blocks, for example when receiving a block or attestation whose parent is unknown.
+
+Clients MUST support requesting blocks since the latest finalized epoch.
+
```
wemeetagain: Do we need/want to request recent beacon state(s)? to eg: fetch a latest finalized state for syncing initially

specs/networking/p2p-interface.md
Mikerah: I think `D` is the degree of the gossipsub overlay network.

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+## Req/Resp
+
+### Why segregate requests into dedicated protocol IDs?
+
+Requests are segregated by protocol ID to:
+
+1. Leverage protocol routing in libp2p, such that the libp2p stack will route the incoming stream to the appropriate handler. This allows each the handler function for each request type to be self-contained. For an analogy, think about how you attach HTTP handlers to a REST API server.
+2. Version requests independently. In a coarser-grained umbrella protocol, the entire protocol would have to be versioned even if just one field in a single message changed.
+3. Enable clients to select the individual requests/versions they support. It would no longer be a strict requirement to support all requests, and clients, in principle, could support a subset of equests and variety of versions.
```
dryajov: typo: `equests` should be `requests`

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+4. Enable flexibility and agility for clients adopting spec changes that impact the request, by signalling to peers exactly which subset of new/old requests they support.
+5. Enable clients to explicitly choose backwards compatibility at the request granularity. Without this, clients would be forced to support entire versions of the coarser request protocol.
+6. Parallelise RFCs (or ETH2 EIPs). By decoupling requests from one another, each RFC that affects the request protocol can be deployed/tested/debated independently without relying on a synchronisation point to version the general top-level protocol.
+  1. This has the benefit that clients can explicitly choose which RFCs to deploy without buying into all other RFCs that may be included in that top-level version.
+  2. Affording this level of granularity with a top-level protocol would imply creating as many variants (e.g. /protocol/43-{a,b,c,d,...}) as the cartesian product of RFCs inflight, O(n^2).
+7. Allow us to simplify the payload of requests. Request-id’s and method-ids no longer need to be sent. The encoding/request type and version can all be handled by the framework.
```
dryajov: see first item for counterpoint.

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+In the interoperability testnet, all peers will be subscribed to all global beacon chain topics, so discovering peers in specific shard topics will be unnecessary.
+
+## Req/Resp
+
+### Why segregate requests into dedicated protocol IDs?
+
+Requests are segregated by protocol ID to:
+
+1. Leverage protocol routing in libp2p, such that the libp2p stack will route the incoming stream to the appropriate handler. This allows each the handler function for each request type to be self-contained. For an analogy, think about how you attach HTTP handlers to a REST API server.
+2. Version requests independently. In a coarser-grained umbrella protocol, the entire protocol would have to be versioned even if just one field in a single message changed.
```
dryajov: counterpoint, how to reliably handle interdependent endpoints (are there any?). If endpoint `A` also needs endpoint `B` upgraded, there is no mechanism to do that reliably with this approach.

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+## Transport
+
+Even though libp2p is a multi-transport stack (designed to listen on multiple simultaneous transports and endpoints transparently), we hereby define a profile for basic interoperability.
+
+#### Interop
+
+All implementations MUST support the TCP libp2p transport, and it MUST be enabled for both dialing and listening (i.e. outbound and inbound connections).
+
+The libp2p TCP transport supports listening on IPv4 and IPv6 addresses (and on multiple simultaneously). Clients SHOULD allow the operator to configure the listen IP addresses and ports, including the addressing schemes (IPv4, IPv6).
```
terencechain: Why `SHOULD`? the ability to configure listen IP/ports sounds like a `MUST` to me

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+### Topics
+
+Topics are plain UTF-8 strings, and are encoded on the wire as determined by protobuf (gossipsub messages are enveloped in protobuf messages).
+
+Topic strings have form: `/eth2/TopicName/TopicEncoding`. This defines both the type of data being sent on the topic and how the data field of the message is encoded. (Further details can be found in [Messages](#Messages)).
+
+There are two main topics used to propagate attestations and beacon blocks to all nodes on the network. Their `TopicName`'s are:
+
+- `beacon_block` - This topic is used solely for propagating new beacon blocks to all nodes on the networks. Blocks are sent in their entirety. Clients who receive a block on this topic MUST validate the block proposer signature before forwarding it across the network.
+- `beacon_attestation` - This topic is used to propagate aggregated attestations (in their entirety) to subscribing nodes (typically block proposers) to be included in future blocks. Similarly to beacon blocks, clients will be expected to perform some sort of validation before forwarding, but the precise mechanism is still TBD.
```
terencechain: what validation to be specific? signature? slot?

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+### Topics
+
+Topics are plain UTF-8 strings, and are encoded on the wire as determined by protobuf (gossipsub messages are enveloped in protobuf messages).
+
+Topic strings have form: `/eth2/TopicName/TopicEncoding`. This defines both the type of data being sent on the topic and how the data field of the message is encoded. (Further details can be found in [Messages](#Messages)).
+
+There are two main topics used to propagate attestations and beacon blocks to all nodes on the network. Their `TopicName`'s are:
+
+- `beacon_block` - This topic is used solely for propagating new beacon blocks to all nodes on the networks. Blocks are sent in their entirety. Clients who receive a block on this topic MUST validate the block proposer signature before forwarding it across the network.
```
terencechain: we probably should also validate the block is not older than certain slot (i.e. current finalized slot)

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+- 1: Client shut down.
+- 2: Irrelevant network.
+- 3: Fault/error.
+
+Clients MAY use reason codes above `128` to indicate alternative, erroneous request-specific responses.
+
+The range `[4, 127]` is RESERVED for future usage.
+
+#### BeaconBlocks
```
terencechain: I think we need to add one for `BeaconState` as well, to request the last finalized state to start syncing

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+### How are we going to discover peers in a gossipsub topic?
+
+Via discv5 topics. ENRs should not be used for this purpose, as they store identity, location and capability info, not volatile advertisements.
+
+In the interoperability testnet, all peers will be subscribed to all global beacon chain topics, so discovering peers in specific shard topics will be unnecessary.
+
+## Req/Resp
+
+### Why segregate requests into dedicated protocol IDs?
```
dryajov: In theory, I like the elegance of `REST` like endpoints, but in practice they might be as ambiguous as `semver`.

For example, how does the client know which protocol the other peer speaks? This isn't easy to do if every endpoint is versioned independently and the client relies on libp2p internals such as `identify` and `multistream/multiselect` to signal which versions the other end supports. Which right now (AFAIR), isn't possible without probing each endpoint explicitly.

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+-  Peer A supporting v1.1.1 of protocol X.
+-  Peer B supporting v1.1.2 of protocol X.
+
+These two peers should never speak to each other because the results can be unpredictable. This is an oversimplification: imagine the same problem with a set of 10 possible versions. We now have 10^2 (100) possible outcomes that peers need to model for. The resulting complexity is unwieldy.
+
+For this reason, we rely on negotiation of explicit, verbatim protocols. In the above case, peer B would provide backwards compatibility by supporting and advertising both v1.1.1 and v1.1.2 of the protocol.
+
+Therefore, semver would be relegated to convey expectations at the human level, and it wouldn't do a good job there either, because it's unclear if "backwards-compatibility" and "breaking change" apply only to wire schema level, to behaviour, etc.
+
+For this reason, we remove semver out of the picture and replace it with ordinals that require explicit agreement and do not mandate a specific policy for changes.
```
dryajov: It's unclear to me what the versioning policy is - is it strict equality (`==`) or are protocols by default backwards compatible (`<=`)? IMO, this should be explicit in the spec.

specs/networking/p2p-interface.md
dryajov: IMO, relays add complexity at this stage, but should be eventually incorporated as they play an important role in standing up well interconnected network. 

Here are some questions I'd ask:

- how are relays discovered?
- what are the security implications?
- are regular peers supposed to provide this functionality or is it backed by standalone infra, or both?

specs/networking/p2p-interface.md
ralexstokes: looks like an abbreviation for ReQuestResPonse

specs/networking/p2p-interface.md
AgeManning: It was to be consistent with the specification. Didn't want to name it something different to the spec as it may add confusion. See here: https://github.com/libp2p/specs/tree/master/pubsub/gossipsub#meshsub-an-overlay-mesh-router
And here: https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L22

specs/networking/p2p-interface.md
AgeManning: For lighthouse, during initial interop, we were planning on not doing aggregation (just to simplify initial compatibility). As interop progresses I think yes, we should be sending aggregated attestations to `beacon_attestation`.

specs/networking/p2p-interface.md
AgeManning: The rationale for this was that segregating topics is similar to segregating nodes based on protocol id. Currently, nodes should not support two encodings. If they did, they would form two segregated networks and blocks propagating from one network to the other would be dependent on relay nodes (nodes that support both) and this can cause all kinds of headaches. 
For now, I think its best to avoid having relay/translation nodes and disjoint networks and just enforce everyone agrees on one compression. If the compression changes, we coordinate a hard fork.

specs/networking/p2p-interface.md
AgeManning: It's strict equality. 
Support for previous version require the local node to explicitly support a version. You can order which is attempted to be negotiated first, i.e higher version before lower ones.

specs/networking/p2p-interface.md
djrtwo: State sync is something we certainly should add, but we need to examine a set of tradeoffs along with the current state sync upgrade proposals currently being considered in eth1. 

Fortunately for interop and even for mainnet for a while, we will be fine with a pure block sync. That said, I think we should have something in place or at least a proposal in consideration before mainnet launch

specs/networking/p2p-interface.md
djrtwo: Im thinking a signature validation but only require nodes to randomly perform to reduce overhead but still provide heard immunity from bad attestations. Need to consider this more before making a concrete proposal

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+#### Interop
+
+[SecIO](https://github.com/libp2p/specs/tree/master/secio) with `secp256k1` identities will be used for initial interoperability testing.
+
+The following SecIO parameters MUST be supported by all stacks:
+
+-  Key agreement: ECDH-P256.
+-  Cipher: AES-128.
+-  Digest: SHA256.
```
hwwhww: nitpick: using name `SHA-256` in this spec.
```suggestion
-  Digest: SHA-256.
```

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+#### Mainnet
+
+[Noise Framework](http://www.noiseprotocol.org/) handshakes will be used for mainnet. libp2p Noise support [is in the process of being standardised](https://github.com/libp2p/specs/issues/195) in the libp2p project.
+
+Noise support will presumably include IX, IK and XX handshake patterns, and may rely on Curve25519 keys, ChaCha20 and Poly1305 ciphers, and SHA256 as a hash function. These aspects are being actively debated in the referenced issue [ETH 2.0 implementers are welcome to comment and contribute to the discussion.]
```
hwwhww: ditto
```suggestion
Noise support will presumably include IX, IK and XX handshake patterns, and may rely on Curve25519 keys, ChaCha20 and Poly1305 ciphers, and SHA-256 as a hash function. These aspects are being actively debated in the referenced issue [ETH 2.0 implementers are welcome to comment and contribute to the discussion.]
```

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+Topics are post-fixed with an encoding. Encodings define how the payload of a gossipsub message is encoded.
+
+#### Interop
+
+- `ssz` - All objects are SSZ-encoded. Example: The beacon block topic string is: `/beacon_block/ssz` and the data field of a gossipsub message is an ssz-encoded `BeaconBlock`.
+
+#### Mainnet
+
+- `ssz_snappy` - All objects are ssz-encoded and then compressed with snappy. Example: The beacon attestation topic string is: `/beacon_attestation/ssz_snappy` and the data field of a gossipsub message is an `Attestation` that has been ssz-encoded then compressed with snappy.
```
hwwhww: Since topics are plain UTF-8 strings, does it make sense to have a map of `code -> encoding`, so that we can save some space in these topics? e.g., `ssz` is `1` and `ssz_snappy` is `2`, so the topics are `/beacon_attestation/1` and `/beacon_attestation/2` .

specs/networking/p2p-interface.md
raulk: Guilty as charged.`MSG` is ambiguous though; other layers also deal with messages.

specs/networking/p2p-interface.md
raulk: Yeah. Multiple encodings cause fragmentation, split brain, chimera; essentially diverging states. It may be a desirable property, but implementing it this way opens a can of worms, which can be addressed, but the cost/benefit isn’t clear to me at this point.

specs/networking/p2p-interface.md
raulk: Yes.

specs/networking/p2p-interface.md
raulk: Indeed. I’ll clarify/reword this in the spec.

specs/networking/p2p-interface.md
arnetheduck: multiple serialization formats means multiple signatures, and having everyone that publishes enr records understand all formats anyway, if I'm not mistaken. no real win. 

what perhaps would make more sense would be to use protobuf or some other established extensible format for enr's given that the requirements are significantly different from blockchain data, but that's probably unpalatable for eth1. regardless, this seems off topic for this thread and belongs in the discv5 forum, as rlp has been deemed unworkable for eth2.

specs/networking/p2p-interface.md
arnetheduck: for testing purposes, and for those that don't have `auto*` features implemented, it's easier to require a straightforward networking setup to start with - these exotic extras are essentially workarounds for when networking is difficult - not a complication we necessarily want to have from the start, when testing things.

specs/networking/p2p-interface.md
arnetheduck: I'm not too happy with Req/Resp in general, it's quite the mouthful, happy to hear alternatives.. `DM_MAX_SIZE` is one for direct messaging between peers, as opposed to gossip/broadcast.

specs/networking/p2p-interface.md
arnetheduck: Indeed, the aggregation strategy has yet to be decided. An open question is whether to allow aggregated attestations here explicitly - it's fine for nimbus, but would require that all testnet participants can at least deal with reading aggregated attestations. Shout if this is a problem, or we'll allow aggregated also.

specs/networking/p2p-interface.md
raulk: For keeping the comment reel tidy, this thread can be marked as resolved @arnetheduck (I can't do it).

raulk:
# Open points 💬

To keep things tidy, this comment will canonically track open points that (1) emerge from the discussion AND (2) are not immediately solvable.

We specify a follow-up action for each, in order to make them actionable.

- [ ] ETH1 and ETH2 cross-compatible encoding of ENRs (RLP vs SSZ): https://github.com/ethereum/devp2p/issues/107

specs/networking/p2p-interface.md
raulk: Opened a discussion: https://github.com/ethereum/devp2p/issues/107, and added to the open points tracker comment here: https://github.com/ethereum/eth2.0-specs/pull/1328#issuecomment-517664569.

specs/networking/p2p-interface.md
raulk: Agree with the above. Let's add network topology/connectivity improvements/complexity incrementally.

raulk: By phasing these aspects in, the barrier for WIP libp2p implementations to become _feature complete_ also becomes more lower and more attainable.

specs/networking/p2p-interface.md
raulk: Yeah. That said, I think the gossipsub spec could do a better job at variable naming. Especially now that it's being cross-referenced from the interwebs. I'll file an issue there.

specs/networking/p2p-interface.md
raulk: Naming things is hard 😛 How about we think outside the box of messaging patterns, and describe this domain functionally:

1. "the direct domain" / `DIRECT_MAX_SIZE`
2. "the service domain" / `SVC_MAX_SIZE`
3. "the operation domain" / `OP_MAX_SIZE`

Personally, I like _the service domain_, as it accurately describes what functionality this offers: peers offer services to one another, where one of those services is the synchronisation facility.

WDYT?

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+Such upgrades lead to fragmentation, so they’ll need to be carried out in a coordinated manner most likely during a hard fork.
+
+### Why are the topics strings and not hashes?
+
+Topics names have a hierarchical structure. In the future, gossipsub may support wildcard subscriptions (e.g. subscribe to all children topics under a root prefix) by way of prefix matching. Enforcing hashes for topic names would preclude us from leveraging such features going forward.
+
+No security or privacy guarantees are lost as a result of choosing plaintext topic names, since the domain is finite anyway, and calculating a digest's preimage would be trivial.
+
+Furthermore, the ETH2 topic names are shorter their digest equivalents (asuming SHA-256 hash), so hashing topics would bloat messages unnecessarily.
```
raulk:
```suggestion
Furthermore, the ETH2 topic names are shorter than their digest equivalents (asuming SHA-256 hash), so hashing topics would bloat messages unnecessarily.
```

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+## Req/Resp
+
+### Why segregate requests into dedicated protocol IDs?
+
+Requests are segregated by protocol ID to:
+
+1. Leverage protocol routing in libp2p, such that the libp2p stack will route the incoming stream to the appropriate handler. This allows each the handler function for each request type to be self-contained. For an analogy, think about how you attach HTTP handlers to a REST API server.
+2. Version requests independently. In a coarser-grained umbrella protocol, the entire protocol would have to be versioned even if just one field in a single message changed.
+3. Enable clients to select the individual requests/versions they support. It would no longer be a strict requirement to support all requests, and clients, in principle, could support a subset of equests and variety of versions.
```
raulk:
```suggestion
3. Enable clients to select the individual requests/versions they support. It would no longer be a strict requirement to support all requests, and clients, in principle, could support a subset of requests and variety of versions.
```

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+Such upgrades lead to fragmentation, so they’ll need to be carried out in a coordinated manner most likely during a hard fork.
+
+### Why are the topics strings and not hashes?
+
+Topics names have a hierarchical structure. In the future, gossipsub may support wildcard subscriptions (e.g. subscribe to all children topics under a root prefix) by way of prefix matching. Enforcing hashes for topic names would preclude us from leveraging such features going forward.
+
+No security or privacy guarantees are lost as a result of choosing plaintext topic names, since the domain is finite anyway, and calculating a digest's preimage would be trivial.
+
+Furthermore, the ETH2 topic names are shorter their digest equivalents (assuming SHA-256 hash), so hashing topics would bloat messages unnecessarily.
```
raulk:
```suggestion
Furthermore, the ETH2 topic names are shorter than their digest equivalents (assuming SHA-256 hash), so hashing topics would bloat messages unnecessarily.
```

specs/networking/p2p-interface.md
raulk: Yes, adding such a validation rule reduces DoS vectors.

specs/networking/p2p-interface.md
arnetheduck: breathe in, breathe out :)

specs/networking/p2p-interface.md
raulk: Fixed. @arnetheduck could you mark as resolved?

specs/networking/p2p-interface.md
arnetheduck: I'm sticking a link to that spec in there in the meantime

specs/networking/p2p-interface.md
raulk: > For example, how does the client know which protocol the other peer speaks?

We have planned to address that in the discovery layer, but we're deferring defining the exact mechanism because it's irrelevant until mainnet. There's an idea here though: https://github.com/ethereum/eth2.0-specs/pull/1328/files#diff-9e07d955515655364c9a5f28c94f5627R236

specs/networking/p2p-interface.md
arnetheduck: We can still allow clients to publish multiple encodings as long as they at least publish the mandatory one - need a phrasing for that.

specs/networking/p2p-interface.md
arnetheduck: Did another round of thinking here, and I'm leaning towards `REQ_MAX_SIZE` for the req/resp domain - services sounds like web services or the like which are quite different in style from this - in particular web services tend to be unidirectional.

We have the pub/sub duality and I suspect the req/resp is another example of the same, leaning on the terminology of p2p message buses like 0mq.

specs/networking/p2p-interface.md
arnetheduck: 0 = "no error" - seems appropriate, given that the other codes indicate some form of error - else you have to pick some arbitrary value for "success" in a protocol like this - better go with something that's commonly used this way already. Compare protobuf for example, you'd have an "error" enum in there, you'd set 0 as success because that's what the generated code gives you when the error field lacks a value.

specs/networking/p2p-interface.md
raulk: `REQ_` => my understanding was that this limitation applied to responses too.

specs/networking/p2p-interface.md
arnetheduck: that's what a `byte` generally is though, as opposed to an int8 for example - an octet of bits with no sign really. how should it be clarified? 

https://en.wikipedia.org/wiki/Byte:

> The modern de facto standard of eight bits, as documented in ISO/IEC 2382-1:1993, is a convenient power of two permitting the binary-encoded values 0 through 255 for one byte (2 to the power 8 is 256).

specs/networking/p2p-interface.md
mkalinin: Pubsub topics should be related to a type of content that is disseminating. In our case message encoding is not a part of the content itself, it rather a transport specific thing that matters between a pair of peers speaking to each other. Therefore, specifying encoding in topic name looks odd and puts additional complexity to the process of its update. It would require either multiple topic distribution from at least _some_ peers in the network or a hard fork coordination which could be quite a tricky thing. We might not want to have a deal with one or another.

Is there a rationale behind _not_ including encoding into a protocol?

specs/networking/p2p-interface.md
arnetheduck: agreed, post-interop at least.

specs/networking/p2p-interface.md
arnetheduck: a reasonable start is not to propagate orphans (missing parent block) - once you're there, it seems pretty reasonable to require sig validation also. the concern is that constructing the state for the attestation is too costly?

specs/networking/p2p-interface.md
arnetheduck: that is already implied - once finalization, all blocks not on the canonical chain before it become invalid by definition.

specs/networking/p2p-interface.md
arnetheduck: for interop, we start with full sync clients from genesis - syncing state should probably be landed together with light clients and a strategy for deciding _which_ states should be available for download (all, certain epochs, one per month, etc)

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+In the interoperability testnet, all peers will be subscribed to all global beacon chain topics, so discovering peers in specific shard topics will be unnecessary.
+
+## Req/Resp
+
+### Why segregate requests into dedicated protocol IDs?
+
+Requests are segregated by protocol ID to:
+
+1. Leverage protocol routing in libp2p, such that the libp2p stack will route the incoming stream to the appropriate handler. This allows each the handler function for each request type to be self-contained. For an analogy, think about how you attach HTTP handlers to a REST API server.
```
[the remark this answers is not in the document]
arnetheduck: this has been discussed at length in previous pr's, and landed here finally.

there's a half-baked plan to create "bundles" of features for advertisement purposes to manage version complexity, should this become a problem - so that clients can rely on there being a suite of protocols that work well together.

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+That said, in our case, streams are single-use. libp2p streams are full-duplex, and each party is responsible for closing their write side (like in TCP). We therefore use stream closure to mark the end of a request.
+
+Nevertheless, messages are still length-prefixed to prevent DOS attacks where malicious actors send large amounts of data disguised as a request. A length prefix allows clients to set a maximum limit, and once that limit is read, the client can cease reading and disconnect the stream. This allows a client to determine the exact length of the packet being sent, and it capacitates it to reset the stream early if the other party expresses they intend to send too much data.
```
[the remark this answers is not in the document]
arnetheduck: This is a good point - if somebody is instigating a DoS attack, would a length-prefix hinder them? It seems that they would spoof it anyway.

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+Entire objects should be sent to get the greatest propagation speeds. If only hashes are sent, then block and attestation propagation is dependent on recursive requests from each peer. In a hash-only scenario, peers could receive hashes without knowing who to download the actual contents from. Sending entire objects ensures that they get propagated through the entire network.
+
+### Should clients gossip blocks if they *cannot* validate the proposer signature due to not yet being synced, not knowing the head block, etc?
+
+The prohibition of unverified-block-gossiping extends to nodes that cannot verify a signature due to not being fully synced to ensure that such (amplified) DOS attacks are not possible.
+
+### How are we going to discover peers in a gossipsub topic?
+
+Via discv5 topics. ENRs should not be used for this purpose, as they store identity, location and capability info, not volatile advertisements.
```
mkalinin: It would be useful to add a link to Topic Advertisement here.

specs/networking/p2p-interface.md
arnetheduck: yeah, but it's the request domain which implies a response. not perfect, agree.

specs/networking/p2p-interface.md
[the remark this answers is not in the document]
dryajov: It makes sense when having to manually frame chunks, but in this case we're relying on dedicated streams, so it doesn't seem to add much of anything.

specs/networking/p2p-interface.md
arnetheduck: it's in the topic name because we're describing about the payload. 

broadcasts are generally not negotiated dynamically between two peers because that would require relaying/translation between different encodings and breaking gossip message identity (which determines if a message is retransmitted to neighbours or not). instead we're opting for a network-wide mandated encoding which will be upgraded only in a coordinated way (for example together with a hard fork). _insert radio station analogy_

there's a separate discussion to solve compression at the transport layer, but for the purposes of this spec, we're not there yet.

specs/networking/p2p-interface.md
dryajov: I think it makes sense to have discv5 section here that outlines how eth2.0 will use it specifically.

specs/networking/p2p-interface.md
mkalinin: > and breaking gossip message identity

Looks like I'm starting to get into this. We would have to have a deal with multiple identities for the same message in order to be able to gossip them across peers with different encodings. That is the point, right?

specs/networking/p2p-interface.md
arnetheduck: actually, I think we can simply remove that sentence. what configuration clients offer is domain-specific to that client.

specs/networking/p2p-interface.md
arnetheduck: well, different encodings on different topics means the message would have to be re-encoded by relays to reach the peers that listen only on the "other" encoding - this is the fragmentation that @raulk mentions above. in the process, you get differences in identity, so to know if a message has been published on the "other" encoding you have to reencode it first and check if you've seen it already. 

this is quite costly and non-trivial, so instead we're opting for mandating that all gossip happens on one topic with one encoding and upgrades are done "network-wide". 

also, gossip protocols typically remember what messages they've seen for a finite period of time - if you publish the same message again after that time has passed, it will be re-broadcast - adding a relay delay makes this scenario more likely (the identity problem, from a different angle).

one can imagine that in some complicated upgrade scenario, we might have peers publishing the same message on two topics/encodings, but the price here is pretty high in terms of overhead - both computational and networking, so we'd rather avoid that.

specs/networking/p2p-interface.md
djrtwo: >  it's fine for nimbus, but would require that all testnet participants can at least deal with reading aggregated attestations

these are the exact some datastructure so I think readability should be fine. I lean toward being able to broadcast aggregates on this same channel for interop

> the aggregation strategy has yet to be decided.

I need to get the "naive" strategy in the validator (or network) spec asap. (subset of committee aggregates and broadcasts to `beacon_`  channel)

specs/networking/p2p-interface.md
djrtwo: > a reasonable start is not to propagate orphans (missing parent block)

true. worth adding this IMO with a note on further validation.

> the concern is that constructing the state for the attestation is too costly

that plus the signature validation on all of the unaggregated (and aggregated) attestations might reduce our ability to gossip as quickly as necessary. Thinking the random validation with heard immunity is the way to go but need to do some modeling.

specs/networking/p2p-interface.md
prestonvanloon: it's a small addition to increase clarity, maybe it is intuitive to most people but it wasn't clear to me.

specs/networking/p2p-interface.md
prestonvanloon: Sounds good to me and I agree with that perspective. Thanks!

specs/networking/p2p-interface.md
prestonvanloon: Sounds good to me. We'll want relay support soon, but can work without it for interop testing.

specs/networking/p2p-interface.md
arnetheduck: a problem with random validations is that you can no longer disconnect peers that send you junk / police that they're doing their job

specs/networking/p2p-interface.md
mkalinin: > this is quite costly and non-trivial, so instead we're opting for mandating that all gossip happens on one topic with one encoding and upgrades are done "network-wide".

Would you mind to add a couple of sentences with this detail to the rationale section relating to topic encodings to prevent arising such questions from future readers?

specs/networking/p2p-interface.md
arnetheduck: if we do this, we have a higher-value target in `beacon_attestation` - ie one could imagine having a fourcc here where the first three chars are topic and the last is encoding or something like this. that said, if we get stream compression, that should take care of it.

Also, we're sending SSZ - it's not the most bandwidth-friendly format out there already.. not sure this will make a practical difference.

all that said, if there's strong support, I'm fine with indices too.

specs/networking/p2p-interface.md
ralexstokes: so whats wrong w/ `REQ_RESP_MAX_SIZE`? I think the stumbling block is less a perfect name and more that `RQRP` is not a commonly used abbreviation (as far as i have seen)

raulk: > in particular web services tend to be unidirectional

That isn't true. Both REST and SOAP are inherently bidirectional; with or without payloads, they're synchronous.

**Service domain** describes _what_ this domain does, **Req/resp domain** describes _how_ it does it in terms of messaging patterns.

We name other domains with the _what_, not with the _how_, so applying that consistency here would be fair.

Anyway, I'm fine whichever way; I do confess I never really liked req/resp to be honest 😛

arnetheduck:
> * Since we are already taking into account Noise, we should also take into account onion routing.

onion routing generally adds significant overhead and latency. this would need consideration before adding - numbers?

> 
> * If we move to QUIC, would we still need two separate protocols for peer discovery (UDP-based Discv5) and peer communication (TCP)?

clients will likely implement alternative peer discovery mechanisms regardless as discv5 is explicitly limited to UDP - a no-go in many networks (and browsers). 

that said, for core validating nodes in the network, it is reasonable to expect a stable networking setup with UDP access and there discv5 will do the job. 

It's going to take a while for QUIC to be deployed and generally accepted / implemented - in this spec and for this reason, we explicitly chose to not stray into QUIC territory (the door remains open for clients to experiment with it)

specs/networking/p2p-interface.md
arnetheduck: at this point, I guess it's less about choosing a good name and more about choosing one that is not too bad :) `REQ_RESP_MAX_SIZE` gets bonus points for staying true to eth2's penchant for excessively long names.

I hold commit rights and take bribes.

specs/networking/p2p-interface.md
arnetheduck: > I need to get the "naive" strategy in the validator (or network) spec asap.

It's not a blocker for interop, but indeed a high-priority concern so we can start investigating performance. 

https://github.com/ethereum/eth2.0-specs/issues/1331

specs/networking/p2p-interface.md
arnetheduck: https://github.com/ethereum/eth2.0-specs/issues/1332

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
+# Overview
+
+This document contains the network specification for Ethereum 2.0 clients.
+
+It consists of four main sections:
+
+1. A specification of the network fundamentals detailing the two network configurations: interoperability test network, and mainnet launch.
+2. A specification of the three network interaction _domains_ of ETH2.0: (a) the gossip domain, (b) the discovery domain, \(c\) the Req/Resp domain.
+3. The rationale and further explanation for the design choices made in the previous two sections.
```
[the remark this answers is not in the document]
arnetheduck: as this lists sections, it reads just fine without the `are` I think

specs/networking/p2p-interface.md
arnetheduck: link added, marked as under construction as discv5 specifics are still being worked out

514ff5814f233836eba8b45d9bb5d94b97d7a60e

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,743 @@
...
+
+During connection bootstrapping, libp2p dynamically negotiates a mutually supported multiplexing method to conduct parallel conversations. This applies to transports that are natively incapable of multiplexing (e.g. TCP, WebSockets, WebRTC), and is omitted for capable transports (e.g. QUIC).
+
+Two multiplexers are commonplace in libp2p implementations: [mplex](https://github.com/libp2p/specs/tree/master/mplex) and [yamux](https://github.com/hashicorp/yamux/blob/master/spec.md). Their protocol IDs are, respectively: `/mplex/6.7.0` and `/yamux/1.0.0`.
+
+Clients MUST support [mplex](https://github.com/libp2p/specs/tree/master/mplex) and MAY support [yamux](https://github.com/hashicorp/yamux/blob/master/spec.md). If both are supported by the client, yamux must take precedence during negotiation. See the Rationale section of this document for tradeoffs.
+
+# ETH2 network interaction domains
+
+## Constants
```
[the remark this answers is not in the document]
arnetheduck: 514ff5814f233836eba8b45d9bb5d94b97d7a60e

specs/networking/p2p-interface.md
arnetheduck: 514ff5814f233836eba8b45d9bb5d94b97d7a60e

specs/networking/p2p-interface.md
arnetheduck: 514ff5814f233836eba8b45d9bb5d94b97d7a60e

specs/networking/p2p-interface.md
arnetheduck: 514ff5814f233836eba8b45d9bb5d94b97d7a60e

specs/networking/p2p-interface.md
arnetheduck: 514ff5814f233836eba8b45d9bb5d94b97d7a60e

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,760 @@
...
+
+At this stage, clients are licensed to drop IPv4 support if they wish to do so, cognizant of the potential disadvantages in terms of Internet-wide routability/support. Clients MAY choose to listen only on IPv6, but MUST retain capability to dial both IPv4 and IPv6 addresses.
+
+Usage of circuit relay, AutoNAT or AutoRelay will be specifically re-examined closer to the time.
+
+## Encryption and identification
+
+#### Interop
+
+[SecIO](https://github.com/libp2p/specs/tree/master/secio) with `secp256k1` identities will be used for initial interoperability testing.
```
ralexstokes: where does `secp256k1` come into play here? if you are referring to the key exchange, i believe the P-256 curve (cf. the Go std lib `crypto/elliptic`) is not the same as the `secp256k1` curve

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,760 @@
...
+
+-  Key agreement: ECDH-P256.
+-  Cipher: AES-128.
+-  Digest: SHA-256.
+
+#### Mainnet
+
+[Noise Framework](http://www.noiseprotocol.org/) handshakes will be used for mainnet. libp2p Noise support [is in the process of being standardised](https://github.com/libp2p/specs/issues/195) in the libp2p project.
+
+Noise support will presumably include IX, IK and XX handshake patterns, and may rely on Curve25519 keys, ChaCha20 and Poly1305 ciphers, and SHA-256 as a hash function. These aspects are being actively debated in the referenced issue [ETH 2.0 implementers are welcome to comment and contribute to the discussion.]
```
zmanian: How would this interact with QUIC where the standard defines it's own authentication and confidentiality layer?

specs/networking/p2p-interface.md
```diff
@@ -0,0 +1,760 @@
...
+
+The following SecIO parameters MUST be supported by all stacks:
+
+-  Key agreement: ECDH-P256.
+-  Cipher: AES-128.
+-  Digest: SHA-256.
+
+#### Mainnet
+
+[Noise Framework](http://www.noiseprotocol.org/) handshakes will be used for mainnet. libp2p Noise support [is in the process of being standardised](https://github.com/libp2p/specs/issues/195) in the libp2p project.
```
zmanian: I think you should specify that mainnet clients MUST not connect via secio

specs/networking/p2p-interface.md
zmanian: Also it seems weird to me to support more than one noise handshake pattern.

zmanian: Also if you want to mandate an authenticated handshake like `XX`, how does this interact with `secp256k1` identity key?

zmanian: There is an extension to the noise protocol that supports signatures that may be relevant but to my knowledge none of the noise implementations support it.

https://github.com/noiseprotocol/noise_sig_spec/blob/master/output/noise_sig.pdf

arnetheduck: These are excellent points but belong in https://github.com/libp2p/specs/issues/195

Re QUIC, libp2p will understand the features that quic offers and disable its own corresponding protocols accordingly

specs/networking/p2p-interface.md
arnetheduck: We ensure that peers at least support the Noise handshake - naturally clients will gravitate towards it as it will have the most peers available. That said, we want to allow alternate transports and protocols for environments where the default is not feasible.

The expectation is that libp2p clients will gradually be deprioritizing connections through secio naturally as better alternatives take its place (fewer round trips, better security) - there's no need to explicitly forbid it.

specs/networking/p2p-interface.md
cheatfate: @ralexstokes of course for ECDHE, P-256 curve are used. But for peer identification libp2p supports many algorithms (RSA, ED25519, ECDSA P-256/384/512 and secp256k1). Peer identification used to build `PeerID` and to sign/verify packets in gossipsub protocol.

ralexstokes: ok thanks -- i thought peer IDs were only RSA, but that is not correct :)

specs/networking/p2p-interface.md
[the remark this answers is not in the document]
arnetheduck: d09d56bec8cec98e3d77a286c48972521dd2be96

specs/networking/p2p-interface.md
raulk: The plan is to group `A` and `B` in the discovery layer under protocol "families/suites/manifests" -- which represent aggregations of finer-grained protocols under one logical unit. that way we get the best of both worlds: upgraded peers can advertise/negotiate newer versions of a given operation, while continuing to serve older peers by supporting the former versions too.

while we could consider a mechanism to expose machine-readable metadata that specifies that `A` depends on `B`, I believe that's beyond scope. in practice, it should be enough to:

1. specify that in the RFC/EIP that introduces `A`.
2. potentially grouping `A` and `B` under a protocol family that can be atomically advertised.
3. allowing peers to incur in a runtime error when they encounter a peer that supports `A` but not `B`.

raulk: Cross-referenced from https://github.com/ethereum/eth2.0-specs/issues/1335.

specs/networking/p2p-interface.md
raulk: Opened an issue to track this: https://github.com/ethereum/eth2.0-specs/issues/1335, and added a checklist item to the Open Points comment below: https://github.com/ethereum/eth2.0-specs/pull/1328#issuecomment-517664569.

specs/networking/p2p-interface.md
raulk: @zmanian this should cover most questions: https://github.com/libp2p/specs/issues/195#issuecomment-517353941

Offering ?K and ?X variants is useful because in some cases you will have the peer’s public static key (e.g. bootstrappers, prev connection). And, as the initiator, you’ll want to leverage it to encrypt your 0-RTT early data.

re: QUIC. Transports offering native encryption/authentication or multiplexing don’t overlay such features.

PR #1316, opened 2019-07-25
Updated light client sync for newer committees

vbuterin:

PR #1329, opened 2019-08-01
Remove light client infrastructure from phase 0

JustinDrake:
Remove light client infrastructure (active index roots and crosslink committee roots) from phase 0. Alternative infrastructure (namely, persistent committee roots) will be added in phase 1 for light clients (following an improved design). Benefits:

* Significantly less hashing for the beacon chain (persistent committee roots are relatively rarely updated)
* Code simplification (40 lines removed)
* Conceptual cleanup with all light client infrastructure in phase 1


vbuterin:
Does it? AFAIK the light client PR only uses the added logic in https://github.com/ethereum/eth2.0-specs/pull/1323.

hwwhww:
@vbuterin we will have to copy the definition of `CompactCommittee` to #1316 if #1329 is merged.

```python
class CompactCommittee(Container):
    pubkeys: List[BLSPubkey, MAX_VALIDATORS_PER_COMMITTEE]
    compact_validators: List[uint64, MAX_VALIDATORS_PER_COMMITTEE]
```

PR #1350, opened 2019-08-09
Minor corrections and clarifications to the network specification

AgeManning:
It is not obvious in some of the requests/responses how the objects should be encoded. I.e as a container or not. 

Single field requests/responses should not be treated as a container as we add unnecessary bytes during the encoding. 

This PR specifies the above, adds a new undefined constant `SSZ_MAX_LIST_SIZE` (as we need to have this for dynamic SSZ lists), and corrects some small mistakes.


specs/networking/p2p-interface.md
```diff
@@ -286,7 +287,7 @@ The `ErrorMessage` schema is:
 )
 ```
 
-*Note that the String type is encoded as UTF-8 bytes without NULL terminator when SSZ-encoded.*
+*Note that the String type is encoded as UTF-8 bytes without NULL terminator when SSZ-encoded. As the `ErrorMessage` is not an SSZ-container, only the UTF-8 bytes will be sent when SSZ-encoded.*
```
mhchia: Good catch!

specs/networking/p2p-interface.md
```diff
@@ -300,7 +301,8 @@ Here `result` represents the 1-byte response code.
 
 The token of the negotiated protocol ID specifies the type of encoding to be used for the req/resp interaction. Two values are possible at this time:
 
--  `ssz`: the contents are [SSZ](https://github.com/ethereum/eth2.0-specs/blob/192442be51a8a6907d6401dffbf5c73cb220b760/specs/networking/libp2p-standardization.md#ssz-encoding) encoded. This encoding type MUST be supported by all clients.
+-  `ssz`: the contents are [SSZ](https://github.com/ethereum/eth2.0-specs/blob/192442be51a8a6907d6401dffbf5c73cb220b760/specs/networking/libp2p-standardization.md#ssz-encoding) encoded. This encoding type MUST be supported by all clients.
```
djrtwo:
```suggestion
-  `ssz`: the contents are [SSZ](#ssz-encoding) encoded. This encoding type MUST be supported by all clients. 
```

djrtwo: Looks like we should make the "SimpleSerialize (SSZ) specification" link below relative too

AgeManning: Yep, good idea

djrtwo:
Can you handle the merge conflict? down to merge after that

AgeManning:
Corrected the merge conflicts.

PR #1323, opened 2019-07-27
Starting on phase 1 misc beacon changes

vbuterin:

PR #1363, opened 2019-08-15
Executable Merkle proofs

hwwhww:
### Issue 
#1323 needs some shared logic in `specs/light_client/merkle_proofs.md`. And also, we can use the helper functions in `merkle_proofs.md` to generate the test cases for #1323.

### Description
This PR is making `merkle_proofs.md` executable.

### TODOs: 
- [x] Fix all linter errors
- [x] Add basic tests


specs/light_client/merkle_proofs.md
```diff
@@ -96,36 +133,43 @@ def chunk_count(typ: SSZType) -> int:
...
     """
     Converts a path (eg. `[7, "foo", 3]` for `x[7].foo[3]`, `[12, "bar", "__len__"]` for
     `len(x[12].bar)`) into the generalized index representing its position in the Merkle tree.
     """
-    root = 1
+    root: Optional[GeneralizedIndex] = GeneralizedIndex(1)
     for p in path:
         assert not issubclass(typ, BasicValue)  # If we descend to a basic type, the path cannot continue further
         if p == '__len__':
             typ, root = uint64, root * 2 + 1 if issubclass(typ, (List, Bytes)) else None
```
hwwhww: @vbuterin could you elaborate on why return `None`? Or should it just raise an exception?

specs/light_client/merkle_proofs.md
```diff
@@ -96,36 +133,43 @@ def chunk_count(typ: SSZType) -> int:
         return len(typ.get_fields())
     else:
         raise Exception(f"Type not supported: {typ}")
+```
 
-
-def get_item_position(typ: SSZType, index: Union[int, str]) -> Tuple[int, int, int]:
+```python
+def get_item_position(typ: SSZType, index_or_variable_name: Union[int, SSZVariableName]) -> Tuple[int, int, int]:
```
hwwhww: @protolambda @vbuterin 
addressing https://github.com/ethereum/eth2.0-specs/pull/1186#discussion_r314088781 in this function with `index_or_variable_name: Union[int, SSZVariableName]`. What do you think about it?

specs/light_client/merkle_proofs.md
```diff
@@ -61,25 +96,27 @@ We can now define a concept of a "path", a way of describing a function that tak
...
         return typ.byte_len
     else:
         return 32
-        
-        
-def get_elem_type(typ: ComplexType, index: Union[int, str]) -> Type:
+```
+
+```python
+def get_elem_type(typ: Union[BaseList, Container], index_or_variable_name: Union[int, SSZVariableName]) -> SSZType:
```
hwwhww: @vbuterin @protolambda 
`ComplexType` has not been defined yet. I use `Union[BaseList, Container]` to replace it now, could you confirm it's what you expected?

specs/light_client/merkle_proofs.md
```diff
@@ -6,17 +6,52 @@
...
+def get_next_power_of_two(x: int) -> int:
+    """
+    Get next power of 2 >= the input.
+    """
+    if x <= 2:
+        return x
+    elif x % 2 == 0:
+        return 2 * get_next_power_of_two(x // 2)
+    else:
+        return 2 * get_next_power_of_two((x + 1) // 2)
```
vbuterin: Why not just replace both of the last two sections with `return 2 * get_next_power_of_two((x + 1) // 2)`?

specs/light_client/merkle_proofs.md
```diff
@@ -6,17 +6,52 @@
...
+```
+
+```python
+def get_previous_power_of_two(x: int) -> int:
+    """
+    Get the previous power of 2 >= the input.
+    """
+    return x if x <= 2 else 2 * get_previous_power_of_two(x // 2)
```
vbuterin: The format (`return a if t else b` vs multiline if statement) should be consistent between this function and the last. I prefer single-line over multi-line here, but if multi-line is better then it should be used for both.

specs/light_client/merkle_proofs.md
vbuterin: It must be one of `Bytes`, `BytesN`, `List`, `Vector`, `Container`. Does `BaseList` cover all of the former four?

specs/light_client/merkle_proofs.md
vbuterin: The one issue is that I think the code will look terrible if we have to write `get_generalized_index(FooType, [SSZVariableName("bar"), 9, SSZVariableName("baz")])` instead of `get_generalized_index(FooType, ["bar", 9, "baz"])`. If we can still do the latter then I think it's good.

specs/light_client/merkle_proofs.md
vbuterin: The idea here is that if you ask for the length of a fixed-length type, then that does not require knowing anything about the data, so there's no GeneralizedIndex that it makes sense to return.

specs/light_client/merkle_proofs.md
hwwhww: Good point!

I changed the definition of `SSZVariableName` to just an alias of `str` (`SSZVariableName = str`). So we can still write `get_generalized_index(FooType, ["bar", 9, "baz"])`.

specs/light_client/merkle_proofs.md
hwwhww: Thank you for the elaboration! `BaseList` only covers `List` and `Vector`. I added `BaseList` to cover `Bytes` and `BytesN` in https://github.com/ethereum/eth2.0-specs/pull/1363/commits/a1061f0b2f777b8c2e55d8134edf95885a5607f6.

specs/light_client/merkle_proofs.md
```diff
@@ -6,20 +6,53 @@
...
+    - [Merkle multiproofs](#merkle-multiproofs)
 
 <!-- /TOC -->
 
+## Custom types
+
+We define the following Python custom types for type hinting and readability:
+
+| - | - | - |
+| `GeneralizedIndex` | `uint64` | the index of a node in a binary Merkle tree |
```
vbuterin: A GeneralizedIndex is NOT necessarily a uint64. It could be bigger. My idea was to represent it in the python executable as a bigint, but then clarify that representing it as a bitfield/bitlist is ok too.

hwwhww: Fixed in https://github.com/ethereum/eth2.0-specs/pull/1363/commits/bbaa238742a93e2aa0524baaf6fd9049d2943d59.

PR #1366, opened 2019-08-16
remove Val API and add links to outside spec/standards repos

djrtwo:
* removed Validator API from specs repo. This API has been migrated to https://github.com/ethereum/eth2.0-apis and will serve as a central location for additional APIs as created and standardized.
* add links for api, metrics, and interop standards repos to  README


README.md
```diff
@@ -37,6 +37,14 @@ See the [Eth 2.0 Phase 2 Wiki](https://hackmd.io/UzysWse1Th240HELswKqVA?view) fo
 * [Beacon node API for validator](specs/validator/0_beacon-node-validator-api.md)
 
 
+## Additional specifications for client implementers
+
+Additional specifications and standards outside of core consensus can be found in the following repos:
```
hwwhww: Hmm, networking spec is also "outside of core consensus" but we leave them here for more visibility/engagements, right? Could you elaborate why only moving restful APIs away? Is it for letting the clients take `eth2.0-apis` 
 as a submodule?

djrtwo: When I said outside of core consensus, I meant more broadly, that a client doesn't really have to implement any of these APIs. This only aids in interoperability and standards for dev tooling. The cycle on which these might emerge and the standards around them is likely to be pretty independent of the core specs. I also think we might want to write tests against these APIs which would even more so clutter this repo. 

I mispoke with "core consensus". I meant more broadly, the things that a client _must_ conform to.

JustinDrake:
What will happen to non-API standards (light client, test formats, networking, key store, HD wallets, checksums, ...)? Would it make sense to migrate standards to `ethereum/eth2.0-standards` (as opposed to `eth2.0-apis` which may be too "narrow")?

protolambda:
Generally I like tests, generation and formats to live together. But this doesn't mean we can't split it up. I would keep spec-tests, spec-tests-format, spec-test-generation, and welll...  the spec, in this specs repository. But then move apis, apis-test-format, apis-test-generation etc. to a new apis repository. Same thing for the other eth2 components.

djrtwo:
Right now some standards are being placed here and some in eth2.0-pm. Maybe makes sense for a new repo.

As for the light client, it is connected enough to the core spec that testing might be easier at lesat initially done here, but it too might evolve to an independent repo at some point

PR #1375, opened 2019-08-22
Release v0.8.3 to master

djrtwo:
Detailed changelog to accompany release


protolambda:
Invalid signature on cosmetic commit by @lightclient, reason as stated by Matt:

> I saw that 😔 the issue was that I signed that commit when my username was c-o-l-o-r and I use the private github email for commits to avoid leaking my personal email — problem is, when you change your username a new private email address is generated for the account, so that broke the relation to that signature and my account email

Rebased and signed it.

PR #1379, opened 2019-08-24
Fixed shard historic accumulator write to only highest matching index

villanuevawill:


villanuevawill:
@djrtwo this makes sense - you're right I looked at this wrong since technically the root is before the state transition :). Maybe a clarifying comment would be helpful here, but this makes a lot of sense. Thanks for describing this further.

PR #1383, opened 2019-08-27
Revamp 1_shard-data-chains.md

JustinDrake:
**TODO**:

* Add function descriptions.

**PR dependencies**:

1) `compute_proposer_index` from https://github.com/ethereum/eth2.0-specs/pull/1371

**Renamings**:

* persistent committee -> shard committee
* phase 1 fork -> shard genesis
* block data -> block body
* attester bitfield -> aggregation bits
* basefee -> block body price
* total bytes -> block size sum
* earlier committee -> older committee
* later committee -> newer committee
* most recent block core -> latest block header
* target persistent committee size -> max period committee size

**Substantive changes**:

* Use `FlatContainer` for shard blocks and state
* Use `Vector`s instead of `List`s in shard state to be friendly to flattening
* Make shard attestations sign `ShardCheckpoint`s (as opposed to `parent_root`s)
* Make block bodies be multiples of `SHARD_HEADER_SIZE` (simpler and partially cachable Merkleisation)
* Shuffle indices in `get_shard_proposer_index` (also make it consistent with phase 0—see #1371)
* Remove `receipt_root` (instead use `older_committee_rewards` and compute indices on the beacon chain)
* Reorder container fields into semantic groups
* Make fee and rewards logic use `Gwei` everywhere (and reuse `get_base_reward` from phase 0)
* Unify fees and rewards under `_deltas`
* Make body of type `List[byte, SHARD_BLOCK_SIZE_LIMIT - SHARD_HEADER_SIZE]` (as opposed to `Bytes[SHARD_BLOCK_SIZE_LIMIT - SHARD_HEADER_SIZE]`)
* Added missing `2 *` in `range(len(persistent_committee), 2 * MAX_PERSISTENT_COMMITTEE_SIZE)`
* Do not needlessly deduplicate shard committees (by construction there are no duplicates) do not sort validator indices in `get_shard_committee`
* Make `MAX_BLOCK_SIZE_PRICE` less conservative (32 ETH as opposed to 1 ETH)—fees are split across the shard committee anyway
* Added `shard` check in `process_shard_block_header`
* Added `beacon_chain_root` check in `process_shard_block_header`
* Change `len(data.body)` to `len(data.body) + SHARD_HEADER_SIZE` in `process_shard_block` (as for `block_size_sum`)
* Fix bug in `process_shard_block_size_fee` when computing `MAX_BLOCK_SIZE_PRICE` (should have been `// SLOTS_PER_EPOCH` as opposed to `* SLOTS_PER_EPOCH`)
* Do not allow `block.parent_root == Bytes32()` to bypass parent root check
* Possibly fixed a few bugs not listed above; most certainly added a few bugs of my own

**Misc cosmetic changes**:

* Flesh out table of contents
* Order containers topologically
* Add various code comments
* Remove "Beacon attestations" section and associated helpers (to go in `1_beacon-chain-misc.md`—see #1323)
* Add `MIN_BLOCK_SIZE_PRICE` parameter
* Remove `PHASE_1_FORK_SLOT`, `CROSSLINK_LOOKBACK`, `PLACEHOLDER` parameters
* Remove `SHARD_SLOTS_PER_BEACON_SLOT` in favour of `SHARD_SLOTS_PER_EPOCH`
* Avoid `raise Exception` which is not used in phase 0
* Split monolithic `shard_block_transition` into three functions (header, attestations, block size fee)
* Make function signatures consistent (e.g. `BeaconState` parameter, then `ShardState`, then `Shard`)
* Make function parameter names consistent (e.g. use `state` everywhere for `BeaconState`, and `shard_state` for `ShardState`)
* Remove unused `compute_slot_of_shard_slot`


specs/core/1_shard-data-chains.md
```diff
@@ -195,34 +176,67 @@ class ShardReceiptDelta(Container):
...
 
-### `get_period_committee`
+#### `compute_flat_shard_header`
 
 ```python
-def get_period_committee(state: BeaconState, epoch: Epoch, shard: Shard) -> Sequence[ValidatorIndex]:
+def compute_flat_shard_header(block: ShardBlock) -> Bytes[SHARD_HEADER_SIZE]:
     """
-    Return committee for a period. Used to construct persistent committees.
+    Return a flat serialisation of the ``block`` header which preserves hash tree root.
```
vbuterin: >      Return a flat serialisation of the ``block`` header which preserves hash tree root. 

I don't think this rework accomplishes this objective. The SSZ tree structure that the new PR creates is as follows:

```python
        r
      /   \
    / \  / \
s bcr pr sr agg totb br pad sigr
                            / \
                         sig0 sig1
                         /\     /\
                       /\  /\ /\  /\
                      d d d 0 d d d 0
```

Whereas the old one was

```python
         r
     /         \
   body       sigs
   / \         / \
            sig0 sig1
             /\    /\
           /\  /\ /\ /\
......... d d d 0 d d d 0
```

The structuring of "core" and "signatures" was there to precisely create two trees of equal size (core has 7 elements that gets rounded up to 8, and signatures have two elements of 3 chunks each, which get rounded up to 2 * 4 = 8). Signatures being an object under the header makes the tree "imbalanced" with signatures being in a lower position.

JustinDrake: What about

````python
class ShardBlock(Container):
    slot: ShardSlot
    beacon_chain_root: Hash
    parent_root: Hash
    state_root: Hash
    aggregation_bits: Bitvector[TARGET_PERSISTENT_COMMITTEE_SIZE * 2]
    total_bytes: uint64
    body: Bytes[SHARD_BLOCK_SIZE_LIMIT - SHARD_HEADER_SIZE]
    padding: Bytes[32]
    attester_signature: BLSSignature
    proposer_signature: BLSSignature
```

specs/core/1_shard-data-chains.md
```diff
@@ -359,203 +300,214 @@ def get_default_shard_state(beacon_state: BeaconState, shard: Shard) -> ShardSta
...
 
-Note that these acceptance conditions depend on the canonical beacon chain; when the canonical beacon chain reorganizes, the eligibility of shard blocks should be re-evaluated.
+### State mutators
 
-### Shard state transition function helpers
+#### `add_reward`
 
 ```python
-def add_reward(state: ShardState, beacon_state: BeaconState, index: ValidatorIndex, delta: int) -> None:
+def add_reward(state: ShardState, beacon_state: BeaconState, index: ValidatorIndex, delta: Gwei) -> None:
```
hwwhww: As [`state.earlier_committee_rewards` is not in Gwei](https://github.com/ethereum/eth2.0-specs/pull/1326#discussion_r309492535), `delta` and `base_reward` are also not in Gwei.

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +98,403 @@ We define the following Python custom types for type hinting and readability:
...
-    MAX_SIZE = SHARD_BLOCK_SIZE_LIMIT * SHARD_SLOTS_PER_BEACON_SLOT * SLOTS_PER_EPOCH * MAX_EPOCHS_PER_CROSSLINK
-    return hash_tree_root(BytesN[MAX_SIZE](pad(header + footer, MAX_SIZE)))
+def add_fee(state: BeaconState, shard_state: ShardState, index: ValidatorIndex, delta: Gwei) -> None:
+    epoch = compute_epoch_of_shard_slot(state.slot)
+    older_committee = get_period_committee(state, shard_state.shard, compute_shard_period_start_epoch(epoch, 2))
+    newer_committee = get_period_committee(state, shard_state.shard, compute_shard_period_start_epoch(epoch, 1))
+    if index in older_committee:
+        shard_state.older_committee_fees[older_committee.index(index)] += delta
+    elif index in newer_committee:
+        shard_state.newer_committee_fees[newer_committee.index(index)] += delta
```
vbuterin: `add_reward` and `add_fee` seem very mergeable. But this just brings us into the larger question of whether or not we can merge the fee and reward entirely. The most clean proposal seems to be to either just introduce signed integers as a class, or add some offset (eg. 2**63) to the combined variable.

specs/core/1_shard-data-chains.md
```diff
@@ -281,68 +283,7 @@ def get_shard_block_proposer_index(state: BeaconState,
         i += 1
```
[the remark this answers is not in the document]
JustinDrake: > What if all validators in a committee have zero effective balance

I think it's still OK because `effective_balance * MAX_RANDOM_BYTE >= MAX_EFFECTIVE_BALANCE * random_byte` has a greater-than-or-equal sign, and 1 in 32 times `random_byte` is zero.

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +94,346 @@ We define the following Python custom types for type hinting and readability:
...
-            data_root=hash_tree_root(block.core.data),
-            state_root=block.core.state_root,
-            total_bytes=block.core.total_bytes,
-            attester_bitfield=block.core.attester_bitfield,
-        ),
-        signatures=block.signatures,
-    )
+def get_shard_proposer_index(state: BeaconState, shard: Shard, slot: ShardSlot) -> ValidatorIndex:
+    epoch = get_current_epoch(state)
+    active_indices = [i for i in get_shard_committee(state, shard, epoch) if is_active_validator(state.validators[i], epoch)]
```
hwwhww: fix lint error

```suggestion
    active_indices = [
        i for i in get_shard_committee(state, shard, epoch) if is_active_validator(state.validators[i], epoch)
    ]
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +94,346 @@ We define the following Python custom types for type hinting and readability:
...
-    for i in range(len(attester_committee), TARGET_PERSISTENT_COMMITTEE_SIZE):
-        assert block.core.attester_bitfield[i] is False or block.core.attester_bitfield[i] == 0  # TODO: FIX Bitvector
-
-    assert bls_verify(
-        pubkey=bls_aggregate_pubkeys(pubkeys),
-        message_hash=block.core.parent_root,
-        signature=block.signatures.attestation_signature,
-        domain=get_domain(beacon_state, DOMAIN_SHARD_ATTESTER, compute_epoch_of_shard_slot(block.core.slot))
-    )
+def process_shard_period(shard_state: ShardState, state: BeaconState) -> None:
```
hwwhww: wrong order
```suggestion
def process_shard_period(state: BeaconState, shard_state: ShardState) -> None:
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +94,346 @@ We define the following Python custom types for type hinting and readability:
...
+    ))
 ```
 
-### Shard state transition function
 
 ```python
-def shard_state_transition(state: ShardState,
-                           beacon_state: BeaconState,
+def shard_state_transition(state: BeaconState,
```
hwwhww: IMO it's more clear to use the explicit name `beacon_state` in phase 1 specs, just like why we use `beacon_block` in the spec. :)

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +94,346 @@ We define the following Python custom types for type hinting and readability:
...
-            total_bytes=block.core.total_bytes,
-            attester_bitfield=block.core.attester_bitfield,
-        ),
-        signatures=block.signatures,
-    )
+def get_shard_proposer_index(state: BeaconState, shard: Shard, slot: ShardSlot) -> ValidatorIndex:
+    epoch = get_current_epoch(state)
+    active_indices = [i for i in get_shard_committee(state, shard, epoch) if is_active_validator(state.validators[i], epoch)]
+    seed = hash(get_seed(state, epoch) + int_to_bytes(slot, length=8) + int_to_bytes(shard, length=8))
+    compute_proposer_index(state, active_indices, seed)
```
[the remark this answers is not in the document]
JustinDrake: See https://github.com/ethereum/eth2.0-specs/pull/1371 :)

I put this logic in phase 0 to unify beacon and shard proposers.

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +94,346 @@ We define the following Python custom types for type hinting and readability:
...
-```
+### Misc
 
-### `compute_epoch_of_shard_slot`
+#### `compute_epoch_of_shard_slot`
 
 ```python
 def compute_epoch_of_shard_slot(slot: ShardSlot) -> Epoch:
-    return Epoch(slot // SHARD_SLOTS_PER_BEACON_SLOT // SLOTS_PER_EPOCH)
+    return compute_epoch_of_slot(slot // SHARD_SLOTS_PER_EPOCH)
```
terencechain:
```suggestion
    return Epoch(slot // SHARD_SLOTS_PER_EPOCH)
```

djrtwo: `compute_epoch_of_slot` is correct here. Might need a cast of `slot // SHARD_SLOTS_PER_EPOCH` to `Slot` before being passed into `compute_epoch_of_slot`

specs/core/1_shard-data-chains.md
terencechain: Ahh I see it now. Tricky...

specs/core/1_shard-data-chains.md
terencechain: Ahh I see it now, tricky...

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +94,346 @@ We define the following Python custom types for type hinting and readability:
...
-    later_committee = get_period_committee(beacon_state, get_shard_period_start_epoch(epoch, lookback=1), state.shard)
-    if index in earlier_committee:
-        state.earlier_committee_fees[earlier_committee.index(index)] += delta
-    elif index in later_committee:
-        state.later_committee_fees[later_committee.index(index)] += delta
-    else:
-        raise Exception("Should never be here")
+def get_genesis_shard_block(state: BeaconState, shard: Shard) -> ShardBlock:
+    return ShardBlock(data=ShardBlockData(
+        shard=shard,
```
terencechain: `ShardBlockData` doesn't have `shard` : )

specs/core/1_shard-data-chains.md
vbuterin: That sounds like it's dividing by epoch length squared....

Maybe `return compute_epoch_of_slot(slot // SHARD_SLOTS_PER_BEACON_SLOT)` ?

specs/core/1_shard-data-chains.md
JustinDrake: Fixed, thanks :)

specs/core/1_shard-data-chains.md
JustinDrake: I think @terencechain is right. Fixed it to `return Epoch(slot // SHARD_SLOTS_PER_EPOCH)`.

specs/core/1_shard-data-chains.md
JustinDrake: Fixed :+1:

specs/core/1_shard-data-chains.md
JustinDrake: I'm a bit torn because if we stick with `state: BeaconState` then it's consistent with phase 0.

specs/core/1_shard-data-chains.md
JustinDrake: Bit the bullet and went with `int64` :)

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
-        state.later_committee_fees = [Gwei(0) for _ in range(len(later_committee))],
-    else:
-        state.receipt_root = Hash()
-    state.slot += ShardSlot(1)
+### Period processing
+
+```python
+def process_shard_period(state: BeaconState, shard_state: ShardState) -> None:
+    # Rotate rewards and fees
+    state.older_committee_deltas = state.newer_committee_deltas
```
terencechain:
```suggestion
    shard_state.older_committee_deltas = shard_state.newer_committee_deltas
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
+    state.newer_committee_deltas = [GweiDelta(0) for _ in range(MAX_PERIOD_COMMITTEE_SIZE)]
```
terencechain:
```suggestion
    shard_state.newer_committee_deltas = [GweiDelta(0) for _ in range(MAX_PERIOD_COMMITTEE_SIZE)]
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
-        state.earlier_committee_fees = state.later_committee_fees
-        state.later_committee_rewards = [REWARD_COEFFICIENT_BASE for _ in range(len(later_committee))],
-        state.later_committee_fees = [Gwei(0) for _ in range(len(later_committee))],
-    else:
-        state.receipt_root = Hash()
-    state.slot += ShardSlot(1)
+### Period processing
+
+```python
+def process_shard_period(state: BeaconState, shard_state: ShardState) -> None:
```
terencechain:
```suggestion
def process_shard_period(shard_state: ShardState) -> None:
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
-        beacon_chain_root=block.core.beacon_chain_root,
-        parent_root=block.core.parent_root,
-        data_root=hash_tree_root(block.core.data),
-        state_root=Hash(),
-        total_bytes=block.core.total_bytes,
-        attester_bitfield=block.core.attester_bitfield,
+```python
+def process_shard_block_header(state: BeaconState, shard_state: ShardState, block: ShardBlock) -> None:
+    # Verify the shard number
+    assert block.shard == state.shard
```
terencechain:
```suggestion
    assert block.shard == shard_state.shard
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
+    # Verify the slot number
+    assert block.slot == state.slot
```
terencechain:
```suggestion
    assert block.slot == shard_state.slot
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
+    # Verify the beacon chain root
+    parent_epoch = compute_epoch_of_shard_slot(state.latest_block_header.slot)
```
terencechain:
```suggestion
    parent_epoch = compute_epoch_of_shard_slot(shard_state.latest_block_header.slot)
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
+    assert block.beacon_block_root == get_block_root(state, parent_epoch)
+    # Verify the parent root
+    assert block.parent_root == hash_tree_root(state.latest_block_header)
```
terencechain:
```suggestion
    assert block.parent_root == hash_tree_root(shard_state.latest_block_header)
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
+        aggregation_bits=block.aggregation_bits,
+        attestations=block.attestations,
+        # `signature` is zeroed
     )
-
-    # Check state root
-    if validate_state_root:
-        assert block.core.state_root == hash_tree_root(state)
+    # Verify the sum of the block sizes since genesis
+    state.block_size_sum += SHARD_HEADER_SIZE + len(block.body)
```
terencechain:
```suggestion
   shard_state.block_size_sum += SHARD_HEADER_SIZE + len(block.body)
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
+    assert block.shard == state.shard
+    # Verify the slot number
+    assert block.slot == state.slot
+    # Verify the beacon chain root
+    parent_epoch = compute_epoch_of_shard_slot(state.latest_block_header.slot)
+    assert block.beacon_block_root == get_block_root(state, parent_epoch)
+    # Verify the parent root
+    assert block.parent_root == hash_tree_root(state.latest_block_header)
+    # Save current block as the new latest block
+    state.latest_block_header = ShardBlockHeader(
```
terencechain:
```suggestion
    shard_state.latest_block_header = ShardBlockHeader(
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
     )
-
-    # Check state root
-    if validate_state_root:
-        assert block.core.state_root == hash_tree_root(state)
+    # Verify the sum of the block sizes since genesis
+    state.block_size_sum += SHARD_HEADER_SIZE + len(block.body)
+    assert block.block_size_sum == state.block_size_sum
+    # Verify proposer signature
+    proposer_index = get_shard_proposer_index(state, state.shard, block.slot)
```
terencechain:
```suggestion
    proposer_index = get_shard_proposer_index(state, shard_state.shard, block.slot)
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
 
-### Beacon attestations
+#### Attestations
 
-Let:
+```python
+def process_shard_attestations(state: BeaconState, shard_state: ShardState, block: ShardBlock) -> None:
+    pubkeys = []
+    attestation_count = 0
+    shard_committee = get_shard_committee(state, state.shard, block.slot)
```
terencechain:
```suggestion
    shard_committee = get_shard_committee(state, shard_state.shard, block.slot)
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
+            attestation_count += 1
+    # Verify there are no extraneous bits set beyond the shard committee
+    for i in range(len(shard_committee), 2 * MAX_PERIOD_COMMITTEE_SIZE):
+        assert block.aggregation_bits[i] == 0b0
+    # Verify attester aggregate signature
+    domain = get_domain(state, DOMAIN_SHARD_ATTESTER, compute_epoch_of_shard_slot(block.slot))
+    message = hash_tree_root(ShardCheckpoint(shard_state.slot, block.parent_root))
+    assert bls_verify(bls_aggregate_pubkeys(pubkeys), message, block.attestations, domain)
+    # Proposer micro-reward
+    proposer_index = get_shard_proposer_index(state, state.shard, block.slot)
```
terencechain:
```suggestion
    proposer_index = get_shard_proposer_index(state, shard_state.shard, block.slot)
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
-            return True
-
-    # Check previous attestation
-    if candidate.data.previous_crosslink.epoch <= PHASE_1_FORK_EPOCH:
-        assert candidate.data.previous_crosslink.data_root == Hash()
+def process_shard_block_body(state: BeaconState, shard_state: ShardState, block: ShardBlock) -> None:
+    # Verify block body size is a multiple of the header size
+    assert len(block.body) % SHARD_HEADER_SIZE == 0
+    # Apply proposer block body fee
+    proposer_index = get_shard_proposer_index(state, state.shard, block.slot)
```
terencechain:
```suggestion
    proposer_index = get_shard_proposer_index(state, shard_state.shard, block.slot)
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
+    block_body_fee = state.block_body_price * len(block.body) // MAX_SHARD_BLOCK_SIZE
+    process_delta(state, shard_state, proposer_index, -block_body_fee)  # Burn
+    process_delta(state, shard_state, proposer_index, block_body_fee // PROPOSER_REWARD_QUOTIENT)  # Reward
+    # Calculate new block body price
+    block_size = SHARD_HEADER_SIZE + len(block.body)
+    QUOTIENT = MAX_SHARD_BLOCK_SIZE * BLOCK_BODY_PRICE_QUOTIENT
+    price_delta = GweiDelta(state.block_body_price * (block_size - SHARD_BLOCK_SIZE_TARGET) // QUOTIENT)
```
terencechain:
```suggestion
    price_delta = GweiDelta(shard_state.block_body_price * (block_size - SHARD_BLOCK_SIZE_TARGET) // QUOTIENT)
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
+    if price_delta > 0:
+        # The maximum block body price caps the amount burnt on fees within a period
+        MAX_BLOCK_BODY_PRICE = MAX_EFFECTIVE_BALANCE // EPOCHS_PER_SHARD_PERIOD // SHARD_SLOTS_PER_EPOCH
+        state.block_body_price = Gwei(min(MAX_BLOCK_BODY_PRICE, state.block_body_price + price_delta))
```
terencechain:
```suggestion
        shard_state.block_body_price = Gwei(min(MAX_BLOCK_BODY_PRICE, shard_state.block_body_price + price_delta))
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
-                    slot=slot,
-                    state_root=shard_blocks_or_state_roots[slot],
-                    total_bytes=pre_state.total_bytes,
-                ),
-                signatures=ShardBlockSignatures(),
-            ))
-    assert candidate.data.crosslink.data_root == compute_crosslink_data_root(blocks)
-
-    return True
+        state.block_body_price = Gwei(max(MIN_BLOCK_BODY_PRICE, state.block_body_price + price_delta))
```
terencechain:
```suggestion
        shard_state.block_body_price = Gwei(max(MIN_BLOCK_BODY_PRICE, shard_state.block_body_price + price_delta))
```

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
+        if (shard_state.slot + 1) % (SHARD_SLOTS_PER_EPOCH * EPOCHS_PER_SHARD_PERIOD) == 0:
+            process_shard_period(state, shard_state)
+        shard_state.slot += ShardSlot(1)
+```
 
-    # Save states in history accumulator
+```python
+def process_shard_slot(state: BeaconState, shard_state: ShardState) -> None:
+    # Cache state root
+    previous_state_root = hash_tree_root(state)
```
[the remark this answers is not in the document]
JustinDrake: Thanks for all these suggestions :) Everything is now `beacon_state` or `shard_state`.

specs/core/1_shard-data-chains.md
JustinDrake: Ok, I bit the bullet! Everything is now `beacon_state` or `shard_state` :)

specs/core/1_shard-data-chains.md
```diff
@@ -95,517 +91,307 @@ We define the following Python custom types for type hinting and readability:
...
-class ShardBlockSignatures(Container):
-    attestation_signature: BLSSignature
-    proposer_signature: BLSSignature
-```
-
-### `ShardBlockCore`
-
-```python
-class ShardBlockCore(Container):
+class ShardBlock(FlatContainer):
```
vbuterin: This is actually a (big) problem with this approach. With a flat container, you no longer have this nice property that an object with an entire subtree substituted with its root necessarily has the same hash tree root as the original object, so you can't have shard blocks and shard block headers in the same way. Not sure how to get around this within the FlatContainer model.

This is arguably one reason why not doing flat containers and instead taking Danny's suggestion and having "embedded vectors (and containers)" would be better. Basically, the idea would be that if you have a container as follows:

```python
class Foo(Container):
    bar: uint64
    baz: Embedded[Vector[uint64, 4]]
    bav: Bytes32
```

It would be treated identically to:

```python
class Foo(Container):
    bar: uint64
    baz_0: uint64
    baz_1: uint64
    baz_2: uint64
    baz_3: uint64
    bav: Bytes32
```

This way you avoid introducing a separate type of container and instead have flattening as a feature that you can use or not use for specific elements at will.

djrtwo:
> With a flat container, you no longer have this nice property that an object with an entire subtree substituted with its root necessarily has the same hash tree root as the original object, so you can't have shard blocks and shard block headers in the same way. Not sure how to get around this within the FlatContainer model.

Was just about to question if we indeed get this property anymore. The embedded looks like a reasonable approach

specs/core/1_shard-data-chains.md
```diff
@@ -17,595 +17,380 @@
...
     )
-
-    # Check state root
-    if validate_state_root:
-        assert block.core.state_root == hash_tree_root(state)
+    # Verify the sum of the block sizes since genesis
+    shard_state.block_size_sum += SHARD_HEADER_SIZE + len(block.body)
+    assert block.block_size_sum == shard_state.block_size_sum
+    # Verify proposer signature
+    proposer_index = get_shard_proposer_index(beacon_state, shard_state.shard, block.slot)
```
djrtwo: Proposer is not guaranteed to be not slashed. In beacon blocks, we check for if slashed and prevent from proposing. Maybe we should filter in `get_shard_proposer_index` by active _and_ not slashed.

Also, maybe we should consider doing the same filtering in `get_beacon_proposer_index`

specs/core/1_shard-data-chains.md
djrtwo: Was just about to question if we indeed get this property anymore. The embedded looks like a reasonable approach

specs/core/1_shard-data-chains.md
```diff
@@ -17,595 +17,380 @@
...
-    beacon_chain_root: Hash
+    beacon_block_root: Hash
     parent_root: Hash
-    data_root: Hash
     state_root: Hash
-    total_bytes: uint64
-    attester_bitfield: Bitvector[TARGET_PERSISTENT_COMMITTEE_SIZE * 2]
+    body: List[byte, MAX_SHARD_BLOCK_SIZE - SHARD_HEADER_SIZE]
+    block_size_sum: uint64
+    aggregation_bits: Bitvector[2 * MAX_PERIOD_COMMITTEE_SIZE]
```
djrtwo: I'm a bit torn on the `ShardCheckpoint` and bits/signature construction currently. 

1. `ShardCheckpoint` isn't actually a checkpoint so we should consider renaming. Maybe `ShardAttestationData`
2. We have no direct datastructure mapped to what would be sent on the network. For example, in Phase 0, an `Attestation` is nice because it is the precise construction of the datastructure sent on the network. I'm considering defining a `ShardAttestation` for this purpose. Considering defining here _even if_ it is not actually used as a consensus object in the state transition. When we define the phase 1 network spec, if it isn't used in consensus, we can pull it out then.

specs/core/1_shard-data-chains.md
```diff
@@ -17,595 +17,380 @@
...
 
-### `ShardReceiptDelta`
+### `ShardCheckpoint`
 
 ```python
-class ShardReceiptDelta(Container):
-    index: ValidatorIndex
-    reward_coefficient: uint64
-    block_fee: Gwei
+class ShardCheckpoint(Container):
```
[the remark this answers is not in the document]
JustinDrake: Happy to rename to `ShardAttestationData`.

specs/core/1_shard-data-chains.md
```diff
@@ -17,595 +17,380 @@
...
-    i = 0
-    while True:
-        candidate_index = active_indices[(slot + i) % len(active_indices)]
-        random_byte = hash(seed + int_to_bytes(i // 32, length=8))[i % 32]
-        effective_balance = state.validators[candidate_index].effective_balance
-        if effective_balance * MAX_RANDOM_BYTE >= MAX_EFFECTIVE_BALANCE * random_byte:
-            return ValidatorIndex(candidate_index)
-        i += 1
+def get_shard_committee(beacon_state: BeaconState, shard: Shard, epoch: Epoch) -> Sequence[ValidatorIndex]:
+    older_committee = get_period_committee(beacon_state, shard, compute_shard_period_start_epoch(epoch, 2))
```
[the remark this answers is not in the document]
JustinDrake: And it's nice that `len("older") == len("newer")`

specs/core/1_shard-data-chains.md
```diff
@@ -17,595 +17,380 @@
...
-    seed = hash(get_seed(state, current_epoch) + int_to_bytes(shard, length=8) + int_to_bytes(slot, length=8))
-    i = 0
-    while True:
-        candidate_index = active_indices[(slot + i) % len(active_indices)]
-        random_byte = hash(seed + int_to_bytes(i // 32, length=8))[i % 32]
-        effective_balance = state.validators[candidate_index].effective_balance
-        if effective_balance * MAX_RANDOM_BYTE >= MAX_EFFECTIVE_BALANCE * random_byte:
-            return ValidatorIndex(candidate_index)
-        i += 1
+def get_shard_committee(beacon_state: BeaconState, shard: Shard, epoch: Epoch) -> Sequence[ValidatorIndex]:
```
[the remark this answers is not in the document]
JustinDrake: Adding header comments to all functions (similar to phase 0) is on my todo list :) I'll make the todo explicit and resolve all the "add header comment" for now.

specs/core/1_shard-data-chains.md
JustinDrake: Agreed on renaming to `ShardAttestationData`. Defining `ShardAttestation` in the networking and/or validator doc makes sense to me.

specs/core/1_shard-data-chains.md
```diff
@@ -17,595 +17,380 @@
...
+    shard: Shard
     slot: ShardSlot
-    beacon_chain_root: Hash
+    beacon_block_root: Hash
     parent_root: Hash
-    data_root: Hash
     state_root: Hash
-    total_bytes: uint64
-    attester_bitfield: Bitvector[TARGET_PERSISTENT_COMMITTEE_SIZE * 2]
+    body: List[byte, MAX_SHARD_BLOCK_SIZE - SHARD_HEADER_SIZE]
```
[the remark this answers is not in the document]
JustinDrake: I've added `shard: Shard` so now `flatten(ShardBlockHeader)` can be `SHARD_HEADER_SIZE ` exactly without padding.

> Will the object be padded out in the PoC construction?

If wanted, we can have `flatten(Vector[ShardBlockHeader, ...])` add padding.

> And is this intended for some allowance to grow?

Doubling `SHARD_HEADER_SIZE` may be wise given we're at the limit already.

specs/core/1_shard-data-chains.md
```diff
@@ -17,595 +17,380 @@
...
-
-    # Take not-yet-cycled-out validators from earlier committee and already-cycled-in validators from
-    # later committee; return a sorted list of the union of the two, deduplicated
-    return sorted(set(
-        [i for i in earlier_committee if epoch % EPOCHS_PER_SHARD_PERIOD < i % EPOCHS_PER_SHARD_PERIOD]
-        + [i for i in later_committee if epoch % EPOCHS_PER_SHARD_PERIOD >= i % EPOCHS_PER_SHARD_PERIOD]
-    ))
+def get_period_committee(beacon_state: BeaconState, shard: Shard, epoch: Epoch) -> Sequence[ValidatorIndex]:
+    active_validator_indices = get_active_validator_indices(beacon_state, epoch)
+    seed = get_seed(beacon_state, epoch)
```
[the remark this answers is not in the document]
JustinDrake: A clean way to do this is with "seeding domains", similar to signature domains. In hindsight, not having them in phase 0 feels like an oversight.

specs/core/1_shard-data-chains.md
JustinDrake: With hindsight filtering for slashing in `get_beacon_proposer_index` makes more sense :)

specs/core/1_shard-data-chains.md
djrtwo: yeah yeah. Validator doc might be the right place

specs/core/1_shard-data-chains.md
vbuterin: So are we going to replace FlatContainer with Embedded here? I guess the `attestations` and `signature` would need to be part of a 2-item Embedded container where the items are themselves Embedded to get flatness while gaining the ability to use the standard signature root method?

specs/core/1_shard-data-chains.md
```diff
@@ -17,596 +17,384 @@
...
-    # Randomly shift persistent committee
-    persistent_committee = list(get_persistent_committee(state, shard, slot))
-    current_epoch = get_current_epoch(state)
-
-    active_indices = [i for i in persistent_committee if is_active_validator(state.validators[i], current_epoch)]
+def get_shard_proposer_index(beacon_state: BeaconState, shard: Shard, slot: ShardSlot) -> Optional[ValidatorIndex]:
+    epoch = get_current_epoch(beacon_state)
+    shard_committee = get_shard_committee(beacon_state, shard, epoch)
+    active_indices = [i for i in shard_committee if is_active_validator(beacon_state.validators[i], epoch)]
     if not any(active_indices):
```
djrtwo: handled merge @JustinDrake.

I added this `if not any` clause back in in the spirit of being able to have extremely low validator count. `shard_committee` length is not guaranteed to be non-zero and if it is empty, `compute_proposer_index` will never terminate.

specs/core/1_shard-data-chains.md
JustinDrake: I've added the restriction that the proposer must not be slashed. We need to be very explicit about what `BeaconState` is when calling `shard_state_transition` because `BeaconState` changes on a slot-by-slot basis but we want the `BeaconState` at the last epoch boundary.

specs/core/1_shard-data-chains.md
```diff
@@ -17,596 +17,400 @@
...
-        b'\x00' * 32 +
-        pad(header.signatures.attestation_signature, 128) +
-        pad(header.signatures.proposer_signature, 128)
-    )
+def process_delta(beacon_state: BeaconState,
+                  shard_state: ShardState,
+                  index: ValidatorIndex,
+                  delta: Gwei,
+                  positive: bool=True) -> None:
+    epoch = compute_epoch_of_shard_slot(beacon_state.slot)
```
```suggestion
    epoch = compute_epoch_of_shard_slot(shard_state.slot)
```

specs/core/1_shard-data-chains.md
```diff
@@ -17,596 +17,400 @@
...
-        total_bytes=block.core.total_bytes,
-        attester_bitfield=block.core.attester_bitfield,
+```python
+def process_shard_block_header(beacon_state: BeaconState, shard_state: ShardState, block: ShardBlock) -> None:
+    # Verify the shard number
+    assert block.shard == shard_state.shard
+    # Verify the slot number
+    assert block.slot == shard_state.slot
+    # Verify the beacon chain root
+    parent_epoch = compute_epoch_of_shard_slot(shard_state.latest_block_header.slot)
```
villanuevawill: What is the rationale for verifying the slot from the latest block header? After a significant # of skip blocks you get an old picture. It also feels like this would be out of sync with the rest coming from the beacon state at the epoch boundaries - and then this by default making the fork choice also lag behind? Maybe I misunderstand the fork choice rule... I took it to mean you start from the latest winning/accepted crosslink root which has been calculated from the boundary of the epoch represented in the beacon block root.

villanuevawill: I suppose this does avoid problems right at the boundary though... which is why I assume this was there i the first place (IE at the first slot this will make you go back one epoch).

villanuevawill: Although this wouldn't always protect you on the 2nd shard slot of the epoch (assuming 2 shard slots per beacon slot)

JustinDrake: This does look wrong, thanks! 👍

specs/core/1_shard-data-chains.md
```diff
@@ -17,596 +17,400 @@
...
+    beacon_block_root: Hash
     parent_root: Hash
-    data: Bytes[SHARD_BLOCK_SIZE_LIMIT - SHARD_HEADER_SIZE]
     state_root: Hash
-    total_bytes: uint64
-    attester_bitfield: Bitvector[TARGET_PERSISTENT_COMMITTEE_SIZE * 2]
+    body_root: Hash
+    block_size_sum: uint64
+    aggregation_bits: Bitvector[2 * MAX_PERIOD_COMMITTEE_SIZE]
+    attestations: BLSSignature
```
terencechain: Any reasons we not using `attestation_signatures`? I find `attestations` a bit miss-leading

specs/core/1_shard-data-chains.md
```diff
@@ -17,596 +17,400 @@
...
+    # Process slots (including those with no blocks) since block
+    process_shard_slots(shard_state, block.slot)
+    # Process block
+    process_shard_block(beacon_state, shard_state, block)
+    # Validate state root (`validate_state_root == True` in production)
+    if validate_state_root:
+        assert block.state_root == hash_tree_root(shard_state)
+    # Return post-state
+    return shard_state
 ```
```
terencechain:
```suggestion
### Slot processing
```

PR #1384, opened 2019-08-27
Remove bug bitlist length checks in process_attestation

djrtwo:
Some bug bitlist length checks were accidentally added to `process_attestation` in #1305. They were not just duplicate, but used the wrong length function (`get_committee_count` instead of the actual length of the committee).

Although this was a bug introduced into `dev` it was not caught by our tests because in `minimal` config `get_committee_count() == len(committee)` in most cases. Changed the total number of validators to be `SLOTS_PER_EPOCH * 10` to avoid this coincidence.

When adjusting the number of validators a small bug in custody testing was revealed. It happened to work before due to the random validator index that was selected from the committee, but this validator index changed when adjusting the number of total validators. Made the fix to use the proper `epoch` when calculating latest epoch to respond.

Thank you @0xKiwi for pointing out the strange duplicate code in `process_attestation`!

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1615,11 +1615,6 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
         proposer_index=get_beacon_proposer_index(state),
     )
 
-    # Check bitlist lengths
-    committee_size = get_committee_count(state, attestation.data.target.epoch)
-    assert len(attestation.aggregation_bits) == committee_size
-    assert len(attestation.custody_bits) == committee_size
-
     if data.target.epoch == get_current_epoch(state):
         assert data.source == state.current_justified_checkpoint
         parent_crosslink = state.current_crosslinks[data.crosslink.shard]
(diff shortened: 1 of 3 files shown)


JustinDrake:
Somewhat surprised this bug was not caught earlier. In any case, this may be the first report eligible for a [phase 0 bounty](https://github.com/ethereum/eth2.0-specs/issues/1345) :)

protolambda:
When running the mainnet tests, the bug is spotted in current `dev` (as the numbers don't coincidentally match). And not just one test, quite a lot seem to fail when running with mainnet `--config mainnet`.

Example test output snippet:
```
 # Check bitlist lengths
        committee_size = get_committee_count(state, attestation.data.target.epoch)
>       assert len(attestation.aggregation_bits) == committee_size
E       AssertionError

eth2spec/phase1/spec.py:1355: AssertionError
```

So this would have been catched when running mainnet tests as part of test-generation before release merge (dev -> master). If no one would care to run it sooner. The mainnet tests are horrendously slow due to spec constraints however, so I understand that running them is often skipped. Maybe we can set up some other check?

dankrad:
Yes, this was my mistake, these lines should have been removed when the proper fix was applied in #1317. This looks good, and great to have a test for it.

JustinDrake:
Unfortunately it looks like this bug is not eligible for a bounty after all :(

Screenshot 2019-08-28 at 15 45 33

PR #1389, opened 2019-08-29
Suspend withdrawability if there are open custody challenges or not all keys have been revealed

### LINKED ISSUE #1348: Prevent validators from exiting when there are open bit or chunk challenges or when not all custody keys were revealed
We need to prevent validators from exiting voluntarily if any of these conditions are fulfilled:

- Open chunk/bit challenges
- Custody keys not revealed

This was originally implemented by the function below, but the function and its attachment point was lost over some refactoring of the exit queue:

```python
def eligible(state: BeaconState, index: ValidatorIndex) -> bool:
    # Cannot exit if there are still open chunk challenges
    if len([record for record in state.custody_chunk_challenge_records if record.responder_index == index]) > 0:
        return False
    # Cannot exit if there are still open bit challenges
    if len([record for record in state.custody_bit_challenge_records if record.responder_index == index]) > 0:
        return False
    # Cannot exit if you have not revealed all of your custody keys
    elif validator.next_custody_reveal_period <= get_validators_custody_reveal_period(state, index, validator.exit_epoch):
        return False
    # Cannot exit if you already have
    elif validator.withdrawable_epoch < FAR_FUTURE_EPOCH:
        return False
    # Return minimum time
    else:
        return current_epoch >= validator.exit_epoch + MIN_VALIDATOR_WITHDRAWAL_EPOCHS
```

dankrad:
Attempt to fix #1348

### DIFF
--- a/specs/core/1_custody-game.md
+++ b/specs/core/1_custody-game.md
@@ -563,7 +563,9 @@ def process_chunk_challenge(state: BeaconState, challenge: CustodyChunkChallenge
             >= get_current_epoch(state))
     responder = state.validators[challenge.responder_index]
     assert (responder.exit_epoch == FAR_FUTURE_EPOCH 
-            or + MAX_CHUNK_CHALLENGE_DELAY >= get_current_epoch(state))
+            or responder.exit_epoch + MAX_CHUNK_CHALLENGE_DELAY >= get_current_epoch(state))
+    # Verify responder is slashable
+    assert is_slashable_validator(responder, get_current_epoch(state))
     # Verify the responder participated in the attestation
     attesters = get_attesting_indices(state, challenge.attestation.data, challenge.attestation.aggregation_bits)
     assert challenge.responder_index in attesters
@@ -620,6 +622,8 @@ def process_bit_challenge(state: BeaconState, challenge: CustodyBitChallenge) ->
         challenge.responder_index
     ) + 2 * EPOCHS_PER_CUSTODY_PERIOD + responder.max_reveal_lateness
 
+    # Verify responder is slashable
+    assert is_slashable_validator(responder, get_current_epoch(state))
     # Verify the responder participated in the attestation
     attesters = get_attesting_indices(state, attestation.data, attestation.aggregation_bits)
     assert challenge.responder_index in attesters
--- a/specs/core/1_custody-game.md
+++ b/specs/core/1_custody-game.md
@@ -770,7 +770,7 @@ def process_challenge_deadlines(state: BeaconState) -> None:
     for custody_chunk_challenge in state.custody_chunk_challenge_records:
         if get_current_epoch(state) > custody_chunk_challenge.inclusion_epoch + CUSTODY_RESPONSE_DEADLINE:
             slash_validator(state, custody_chunk_challenge.responder_index, custody_chunk_challenge.challenger_index)
-            records = state.custody_chunk_challenge
+            records = state.custody_chunk_challenge_records
             records[records.index(custody_chunk_challenge)] = CustodyChunkChallengeRecord()
 
     for custody_bit_challenge in state.custody_bit_challenge_records:
--- a/specs/core/1_custody-game.md
+++ b/specs/core/1_custody-game.md
@@ -105,6 +105,7 @@ This document details the beacon chain additions and changes in Phase 1 of Ether
 | `EPOCHS_PER_CUSTODY_PERIOD` | `2**11` (= 2,048) | epochs | ~9 days |
 | `CUSTODY_PERIOD_TO_RANDAO_PADDING` | `2**11` (= 2,048) | epochs | ~9 days |
 | `MAX_REVEAL_LATENESS_DECREMENT` | `2**7` (= 128) | epochs | ~14 hours |
+| `MAX_SEED_LOOKAHEAD` | `2**2` (= 4) | epochs | ~25.6 minutes |
 
 ### Max operations per block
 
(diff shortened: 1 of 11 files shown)


dankrad:
WIP; need to add some tests here to stop it from happening again :)

dankrad:
One change I noticed is necessary (unless someone has a better idea) is a new stage in the exit process to give validators the time to reveal their custody secret, but also giving the time for them to be challenged (which can only be happened after they have revealed it).

So the full process now looks like this:
 - validator requests voluntary exit
 - exit in custody period N
 - withdrawability is suspened (`withrawable_epoch==FAR_FUTURE_EPOCH` until custody secret N is revealed)
 - when custody secret N is revealed, `all_custody_secrets_revealed_epoch` is set to the current epoch
 - `withdrawable_epoch` is now `all_custody_secrets_revealed_epoch` if no open challenges are present and `FAR_FUTURE_EPOCH` otherwise

We can only let the validator reveal their custody secret after they have exited, as otherwise they would be liable for slashing. But we have to delay `withdrawable_epoch` from the previous `exit_epoch + MIN_VALIDATOR_WITHDRAWABILITY_DELAY` to `all_custody_secrets_revealed_epoch + MIN_VALIDATOR_WITHDRAWABILITY_DELAY` so that it is not possible to escape being challlenged for custody bits by delaying custody reveal until the last possible epoch.

specs/core/1_custody-game.md
```diff
@@ -477,7 +491,8 @@ def process_early_derived_secret_reveal(state: BeaconState, reveal: EarlyDerived
 
     assert reveal.epoch >= get_current_epoch(state) + RANDAO_PENALTY_EPOCHS
     assert reveal.epoch < get_current_epoch(state) + EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS
-    assert not revealed_validator.slashed
+    assert is_slashable_validator(revealed_validator, get_current_epoch(state))
+    assert get_current_epoch(state) < revealed_validator.exit_epoch
```
dankrad: This is a change: We need to give the validator the opportunity to reveal their current custody secret after exit_epoch, so that they can get into withdrawable state. So from exit_epoch it is not an early reveal anymore.

specs/core/1_custody-game.md
```diff
@@ -545,10 +560,13 @@ def process_chunk_challenge(state: BeaconState, challenge: CustodyChunkChallenge
...
-    assert (compute_epoch_of_slot(challenge.attestation.data.slot)
-            >= get_current_epoch(state) - MAX_CHUNK_CHALLENGE_DELAY)
+    assert (challenge.attestation.data.target.epoch + MAX_CHUNK_CHALLENGE_DELAY
+            >= get_current_epoch(state))
     responder = state.validators[challenge.responder_index]
-    assert responder.exit_epoch >= get_current_epoch(state) - MAX_CHUNK_CHALLENGE_DELAY
+    assert (responder.exit_epoch == FAR_FUTURE_EPOCH 
+            or responder.exit_epoch + MAX_CHUNK_CHALLENGE_DELAY >= get_current_epoch(state))
+    # Verify responder is slashable
+    assert is_slashable_validator(responder, get_current_epoch(state))
```
dankrad: This was missing, which I assume is an oversight

specs/core/1_custody-game.md
```diff
@@ -605,6 +623,8 @@ def process_bit_challenge(state: BeaconState, challenge: CustodyBitChallenge) ->
         challenge.responder_index
     ) + 2 * EPOCHS_PER_CUSTODY_PERIOD + responder.max_reveal_lateness
 
+    # Verify responder is slashable
+    assert is_slashable_validator(responder, get_current_epoch(state))
```
dankrad: This was missing, which I assume is an oversight

specs/core/1_custody-game.md
```diff
@@ -670,9 +690,7 @@ def process_chunk_challenge_response(state: BeaconState,
     # Verify chunk index
     assert response.chunk_index == challenge.chunk_index
     # Verify bit challenge data is null
-    assert response.chunk_bits_branch == [] and response.chunk_bits_leaf == Hash()
-    # Verify minimum delay
-    assert get_current_epoch(state) >= challenge.inclusion_epoch + ACTIVATION_EXIT_DELAY
```
dankrad: I do not understand the logic behind the minimum delay, so I deleted it. @JustinDrake @vbuterin do you know why this was here?

specs/core/1_custody-game.md
```diff
@@ -778,7 +796,26 @@ def after_process_final_updates(state: BeaconState) -> None:
...
+                # custody secrets revealed
+                validator.withdrawable_epoch = FAR_FUTURE_EPOCH
+            else:
+                # Reset withdrawable epochs if challenge records are empty
+                if validator.withdrawable_epoch == FAR_FUTURE_EPOCH:
+                    validator.withdrawable_epoch = Epoch(validator.all_custody_secrets_revealed_epoch
+                                                         + MIN_VALIDATOR_WITHDRAWABILITY_DELAY)
 ```
+
+## Additional validator initialization parameters
```
dankrad: We need to add these to the validator initialization. We also need to think how we are going to implement this change in the phase 0 to phase 1 transition; this code needs to be run for each validator when the transition happens.

specs/core/1_custody-game.md
dankrad: @vbuterin So this was the line we talked about in the call. I think with your commend the idea behind it was that you should not be able to challenge yourself and respond in the same block to get the rewards?
Why can it not just be 1 slot difference then? As in make sure that you can't respond within the same block?

specs/core/1_custody-game.md
dankrad: @djrtwo Do you have a good idea for where we should keep code for phase 0 -> phase 1 transition?

djrtwo: I suppose we should have it in misc-beacon-changes, under a section for "fork transition scripts" or soemthing similar that makes changes to the beaconstate at transition

djrtwo:
@dankrad should we close this due to the impending phase 1 shard chain rewrite?

dankrad:
Yes, however we would still have to implement the key reveal part.

djrtwo:
closing. will address this in phase 1 redesign

PR #1390, opened 2019-08-31
p2p-interface: specify arrays as ssz-encoded lists

wemeetagain:
Based on conversation in discord, clarify that array values are encoded as lists when using ssz encoding.


djrtwo:
Can you point this to `master`?

wemeetagain:
fixed by #1404

PR #1392, opened 2019-09-02
Add FlatContainer definition to SSZ

vbuterin:


specs/simple-serialize.md
```diff
@@ -214,6 +214,25 @@ We now define Merkleization `hash_tree_root(value)` of an object `value` recursi
...
+* `pad_to_power_of_2(bitfield_bytes(value))` if `value` is a bitvector.
+* If the object is a vector of composite objects or a FlatContainer, use the algorithm below, where `elements` is the list of elements if the object is a vector, and an ordered list of its field values if it is a FlatContainer:
+
+```python
+output = b''
+for element in elements:
+    child = flatten(element)
+    # Pad output to a multiple of the length of the child, then add the child
+    output += b'\x00' * (-len(output) % len(child)) + child
+return output
```
vbuterin:
```suggestion
return pad_to_power_of_2(output)
```

specs/simple-serialize.md
```diff
@@ -214,6 +214,25 @@ We now define Merkleization `hash_tree_root(value)` of an object `value` recursi
...
+* `pad_to_power_of_2(pack(value))` is `value` is a basic object or a vector of basic objects.
+* `pad_to_power_of_2(bitfield_bytes(value))` if `value` is a bitvector.
+* If the object is a vector of composite objects or a FlatContainer, use the algorithm below, where `elements` is the list of elements if the object is a vector, and an ordered list of its field values if it is a FlatContainer:
+
+```python
+output = b''
+for element in elements:
+    child = flatten(element)
+    # Pad output to a multiple of the length of the child, then add the child
+    output += b'\x00' * (-len(output) % len(child)) + child
```
djrtwo: Let's avoid negative number modulo arithmetic. This isn't well defined across all languages and we have removed such math from the spec for this reason in the past

vbuterin:
```suggestion
    if len(output) % len(child) != 0:
        output += b'\x00' * (len(child) - len(output) % len(child))
    output += child
```

vbuterin: @djrtwo does this look good?

djrtwo: yes

PR #1397, opened 2019-09-04
Remove `unpack_compact_validator` from sync spec

terencechain:
Change list:
- Removed `unpack_compact_validator`, it's defined in [1_beacon-chain-misc](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/1_beacon-chain-misc.md#unpack_compact_validator)
- ~Moved `LightClientMemory` from `Helpers` section to `Containers` section~

### DIFF
--- a/specs/light_client/sync_protocol.md
+++ b/specs/light_client/sync_protocol.md
@@ -13,9 +13,8 @@
     - [Constants](#constants)
     - [Containers](#containers)
         - [`LightClientUpdate`](#lightclientupdate)
-    - [Helpers](#helpers)
         - [`LightClientMemory`](#lightclientmemory)
-        - [`unpack_compact_validator`](#unpack_compact_validator)
+    - [Helpers](#helpers)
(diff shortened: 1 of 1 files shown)


specs/light_client/sync_protocol.md
```diff
@@ -62,12 +61,9 @@ class LightClientUpdate(container):
     committee_branch: Vector[Hash, PERSISTENT_COMMITTEE_ROOT_IN_BEACON_STATE_DEPTH + log_2(SHARD_COUNT)]
 ```
 
-## Helpers
-
 ### `LightClientMemory`
 
 ```python
-@dataclass
 class LightClientMemory(object):
```
hwwhww: see https://github.com/ethereum/eth2.0-specs/pull/1316#issuecomment-521697986
it should not be SSZ container class. 😄

terencechain: Ahh ok. I'll revert the change. Thank you!

PR #1398, opened 2019-09-05
Update hash_to_g2 in BLS

kirk-baird:
The purpose of this PR is to update the `hash_to_g2` function to match those specified in the [BLS Standards](https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve). Riad has kindly already prepared the ciphersuite `BLS12381G2-SHA256-SSWU-RO`. It can be found in section 8.7 of the hash-to-curve standard. I've written a few extra notes in the hopes of assisting anyone implementing it.

Things worth noting:
- The links are currently pointing at the draft of the standard which is changing and hence they will need to be updated once it is finalised.
- Pseudo Code for `clear_cofactor` is currently being written and when it's complete we should point to that instead of the paper currently being pointed too.
- The standard states that serialisation should be done the same as [Zcash](https://github.com/zkcrypto/pairing/blob/master/src/bls12_381/README.md#serialization) which ours is slightly different.

Let me know if you would like me to expand into more detail on any parts of `hash_to_curve`


specs/bls_signature.md
```diff
@@ -68,45 +68,30 @@ We require:
...
-        x_coordinate += Fq2([1, 0])  # Add 1 and try again
+def hash_to_G2(message_hash: Bytes32) -> Tuple[uint384, uint384]:
+    hash_to_curve(message_hash)
 ```
 
-### `modular_squareroot`
+`hash_to_curve` is found in the [BLS standard](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve) and uses the ciphersuite `BLS12381G2-SHA256-SSWU-RO` found in section 8.7. It consists of three parts:
 
-`modular_squareroot(x)` returns a solution `y` to `y**2 % q == x`, and `None` if none exists. If there are two solutions, the one with higher imaginary component is favored; if both solutions have equal imaginary component, the one with higher real component is favored (note that this is equivalent to saying that the single solution with either imaginary component > p/2 or imaginary component zero and real component > p/2 is favored).
+* `hash_to_base` - Converting a message from bytes to a field point. The required constant parameters are: Security `k = 128` bits, Field Degree `m = 2` (i.e. Fp2), Length of HKDF `L = 64`, `H = SHA256`, Domain Separation Tag `DST = BLS12381G2-SHA256-SSWU-RO`.
```
hwwhww: Does `HKDF-Extract` function take parameter `salt` in bytes? Should we note that which encoding format for the `DST` value?

specs/bls_signature.md
```diff
@@ -68,45 +68,30 @@ We require:
...
-    
-    # Test candidate y coordinates until a one is found
-    while 1:
-        y_coordinate_squared = x_coordinate ** 3 + Fq2([4, 4])  # The curve is y^2 = x^3 + 4(i + 1)
-        y_coordinate = modular_squareroot(y_coordinate_squared)
-        if y_coordinate is not None:  # Check if quadratic residue found
-            return multiply_in_G2((x_coordinate, y_coordinate), G2_cofactor)
-        x_coordinate += Fq2([1, 0])  # Add 1 and try again
+def hash_to_G2(message_hash: Bytes32) -> Tuple[uint384, uint384]:
+    hash_to_curve(message_hash)
```
hwwhww: Can we just rename `hash_to_G2 ` to `hash_to_curve`?

specs/bls_signature.md
```diff
@@ -68,45 +68,30 @@ We require:
...
+   u0 = hash_to_base(alpha, 0)
+   u1 = hash_to_base(alpha, 1)
+   Q0 = map_to_curve(u0)
+   Q1 = map_to_curve(u1)
+   R = Q0 + Q1 # Point Addition
+   P = clear_cofactor(R)
+   return P
+ ```
+
+ An implementation of `hash_to_curve` can be found [here](https://github.com/kwantam/bls_sigs_ref/blob/master/python-impl/opt_swu_g2.py#L130).
```
hwwhww: Maybe use the permalink here?
```suggestion
 An implementation of `hash_to_curve` can be found [here](https://github.com/kwantam/bls_sigs_ref/blob/d82335835cfddd9b9e7f30b99d2dab653d2c3a14/python-impl/opt_swu_g2.py#L130).
```

specs/bls_signature.md
kirk-baird: Oh I've been spending too much time in rust land I've forgotten my return statements. I think it should be `return hash_to_curve(message_hash)`

I was very tempted to rename `hash_to_g2` however as `hash_to_g2` takes `Bytes32` whereas `hash_to_curve` takes `Bytes` I decided to make the abstraction. What are your thoughts on this?

specs/bls_signature.md
kirk-baird: Yep the `salt` for `HKDF-Extract` is `DST`. The standard has the type of `salt` as a `string` hence I left it as a string here. However my implementation (and I'm sure others will too) takes bytes so I would not be against also displaying it as bytes.

benjaminion:
The current signature of `hash_to_G2` is `hash_to_G2(message_hash: Bytes32, domain: Bytes8)`. This draft has `hash_to_G2(message_hash: Bytes32)`. How should we handle the domain parameter?

kirk-baird:
> The current signature of `hash_to_G2` is `hash_to_G2(message_hash: Bytes32, domain: Bytes8)`. This draft has `hash_to_G2(message_hash: Bytes32)`. How should we handle the domain parameter?

At this point it looks like there will be a wrapper class around the message and domain which will do `hash_tree_root` and the hash of that wrapper object will be passed to BLS.

You can see [here](https://github.com/ethereum/eth2.0-specs/issues/1387#issuecomment-526848605) for the discussion about the wrapper object.

specs/bls_signature.md
```diff
@@ -68,45 +68,30 @@ We require:
...
-    check = candidate_squareroot ** 2 / value
-    if check in eighth_roots_of_unity[::2]:
-        x1 = candidate_squareroot / eighth_roots_of_unity[eighth_roots_of_unity.index(check) // 2]
-        x2 = -x1
-        x1_re, x1_im = coerce_to_int(x1.coeffs[0]), coerce_to_int(x1.coeffs[1])
-        x2_re, x2_im = coerce_to_int(x2.coeffs[0]), coerce_to_int(x2.coeffs[1])
-        return x1 if (x1_im > x2_im or (x1_im == x2_im and x1_re > x2_re)) else x2
-    return None
-```
+def hash_to_curve(alpha: Bytes) -> Tuple[unit384, uint384]:
```
hermanjunge: `uint384` the first one :grimacing:

kirk-baird: Opps so close!

kirk-baird:
This has been updated to version 5 of the hash to curve draft and is ready for review :)

Let me know if you want more of less details on any sections.

p.s. I haven't made updates based on the [bls signatures](https://tools.ietf.org/html/draft-boneh-bls-signature-00) draft just the [hash to curve](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-05) draft.

specs/bls_signature.md
```diff
@@ -33,6 +30,18 @@ The BLS12-381 curve parameters are defined [here](https://z.cash/blog/new-snark-
 
 We represent points in the groups G1 and G2 following [zkcrypto/pairing](https://github.com/zkcrypto/pairing/tree/master/src/bls12_381). We denote by `q` the field modulus and by `i` the imaginary unit.
 
+## Ciphersuite
```
MrChico: Header missing in TOC

kirk-baird: I'll add that now :)

specs/bls_signature.md
```diff
@@ -33,6 +31,18 @@ The BLS12-381 curve parameters are defined [here](https://z.cash/blog/new-snark-
...
+
+We use the following Ciphersuite `BLS_SIG_BLS12381G2-SHA256-SSWU-RO_POP_`.
+
+Where:
+* `POP` refers to proof of possession.
+* `G2` refers to signatures on G2.
+* `SHA256` is the hash function used.
+* `SSWU` refers to the Simplified SWU Map from a finite field element to an elliptic curve point.
+* `RO` refers to `hash-to-curve` function rather than `encode-to-curve`.
```
JustinDrake: Could mention that "RO" stands for "random oracle"

specs/bls_signature.md
```diff
@@ -33,6 +31,18 @@ The BLS12-381 curve parameters are defined [here](https://z.cash/blog/new-snark-
 
 We represent points in the groups G1 and G2 following [zkcrypto/pairing](https://github.com/zkcrypto/pairing/tree/master/src/bls12_381). We denote by `q` the field modulus and by `i` the imaginary unit.
 
+## Ciphersuite
+
+We use the following Ciphersuite `BLS_SIG_BLS12381G2-SHA256-SSWU-RO_POP_`.
```
JustinDrake: Should this ciphersuite only be used for the proof of possession (`deposit_data.signature`)?

JustinDrake: Should the ciphersuite be `BLS_SIG_BLS12381G2-SHA256-SSWU-RO-_POP_` (notice the extra `-` after `RO-`)?

CarlBeek: I don't think so, the IETF spec has a `hash_to_point` definition for messages of this ciphersuite which is distinct from `hash_pubkey_to_point`. My understanding is that if you use the [PoP defined by the IETF spec](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-00#section-4.2.3), you should use `BLS_SIG_BLS12381G2-SHA256-SSWU-RO_POP_` or `BLS_SIG_BLS12381G2-SHA256-SSWU-RO-_NUL_` if their PoP is not used.

Considering that we do not use the exact PoP defined by the IETF spec, I think we should not claim to use the PoP ciphersuite anywhere.

```suggestion
We use the following Ciphersuite `BLS_SIG_BLS12381G2-SHA256-SSWU-RO-_NUL_`.
```

kirk-baird: I believe we should just use one ciphersuite for both PoP and regular signature validation.
The extra `-` after `RO-` is deliberate as that's how the examples were given in the IETF spec.

I agree with Carl, if we don't use the specified PoP we should not use that in our ciphersuite tag. However, I then ask why do we not use that standardized PoP?

JustinDrake: > The extra - after RO- is deliberate as that's how the examples were given in the IETF spec.

Right :) The extra `-` were missing and added them in.

> why do we not use that standardized PoP?

What is the difference between the standardized PoP and the currently specified PoP?

kirk-baird: Good question,
The standardize PoP signs the message:
`msg = bytes(public_key)`

Where as we sign the message:
`msg =  signing_root(deposit.data)`
Which contains `withdrawal_credentials` and `amount` along with the `public_key`.

CarlBeek: > The standardize PoP signs the message:
> `msg = bytes(public_key)`

Note that there is a specific domain separation tags just for signing pubkeys. (A third tag not mentioned here yet.)

kirk-baird: You're right sorry, I didn't notice the tag was slightly different between regular signatures and PoP.
PoP: `BLS_POP_BLS12381G2-SHA256-SSWU-RO-_POP_`
Regular Signature: `BLS_SIG_BLS12381G2-SHA256-SSWU-RO-_POP_`

With the option of changing the last `_POP_` to `_NUL_` if required.

kirk-baird:
Closed in favour of #1532

PR #1400, opened 2019-09-07
Add embedded values to containers

vbuterin:
Alternative to #1392 

See https://github.com/ethereum/eth2.0-specs/pull/1383#discussion_r321320585 for original rationale, though note that this proposal deviates from the original proposal in that it adds padding to ensure that embedded values get an aligned subtree, so one can still meaningfully speak of eg. `get_generalized_index(Container, "baz")` if `baz` is the embedded vector in the example given in this PR.

### DIFF
--- a/specs/simple-serialize.md
+++ b/specs/simple-serialize.md
@@ -203,6 +203,27 @@ We first define helper functions:
         - If `> 1` chunks: merkleize as binary tree.
 * `mix_in_length`: Given a Merkle root `root` and a length `length` (`"uint256"` little-endian serialization) return `hash(root + length)`.
 * `mix_in_type`: Given a Merkle root `root` and a type_index `type_index` (`"uint256"` little-endian serialization) return `hash(root + type_index)`.
+* `get_elements`: Given a `Vector` object, returns the object itself. Given a `Container` object, returns the result of the following code:
+
+```python
+def get_elements(value: Container):
+    o = []
+    for element in value:
+        # For normal elements, just append each element to the list
+        if not isinstance(element, Embedded):
+            o.append(element)
+        # For embedded elements, attach the full list consisting of recursively
+        # running `get_elements` on the element to the top-level list, with padding
+        # so that the child element gets an aligned subtree of the hash tree.
+        else:
+            list_to_embed = get_elements(element)
+            while len(o) % next_pow_of_two(len(list_to_embed)) != 0:
+                o.append(Bytes32())
+            o.extend(list_to_embed) + [Bytes32() for i in range(next_pow_of_two(len(list_to_embed)) - len(list_to_embed)]
+    return o
+```
+
+For example a class with three members `(bar: Bytes32, baz: Vector[uint64, 3], bak: bool)`, `get_elements` would treat it the same as a class with elements `(bar: Bytes32, padding_0: Bytes32, padding_1: Bytes32, padding_2: Bytes32, baz_0: uint64, baz_1: uint64, baz_2: uint64, padding_3: Bytes32, bak: bool)` where the `padding_*` elements are all filled with the default zero value.
 
 We now define Merkleization `hash_tree_root(value)` of an object `value` recursively:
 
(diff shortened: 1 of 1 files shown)


vbuterin:
We already have lots of inefficient padding, eg. uint64 and any smaller values get blown up to 32 bytes. So if we want to reduce padding bloat, the thing that would provide _more_ value arguably is to take the "pad to a multiple of the size of the next thing" rule that I introduce here and extend it to sizes below 32 bytes, so that in the happy case up to 4 uint64's would be packed into the same slot. Though arguably it's too late to do that large a change in any case; my point is just that we make sacrifices for simplicity already.

I've looked at the multiproof idea; I like its compactness in principle, my main concern with it is that it is _considerably_ more complex than dumb multiproofs (and also it doesn't have a base case that equals existing Merkle proofs). So I don't think it's a good idea to introduce it anywhere at consensus layer; rather, it should be a standard WASM library that can be used by execution environments.

protolambda:
Changed my mind, agree on simplicity if the padding is small enough.

The Embedding feature for SSZ is useful for other things too: it's much easier to remix an existing struct (or multiple) into a new extended struct without repetition. It may even be a nice way to extend phase-0 defintions without inheritance (no overriding or OOP in spec). E.g.
```
# phase1/spec.py
from phase0 import spec as phase0

BeaconState(Container):
  _: Embedded[phase0.BeaconState]
  additional_field: FooBar
  ...
```

----

RE multi-proof structure:

I think leaves and proof nodes should be interchangeable. Aggregation, transparency (body vs. body-root) and shared proofs are something highly desired, yet we make reasoning about such multi-proofs much more difficult by splitting up the nodes into two lists.

You are right on not introducing tree-offsets as canonical though. There are other ways of doing it that are more efficient in specific scenarios (and it will be more future proof when introducing some form of sparse trees).

Can we maybe go for a compromise where we split multi-proofs from encoding, something like this:
- **Multi-proofs** are simply defined as trees of nodes that hash to a root, matching the hash-tree-root of a given type. They are the minimum raw interface to do proofs with, with no additional complexity introduced (no offsets, no helper indices, just merging to a root).
- **Proof backings** are where we can make opinionated choices in encoding, and I propose the following list to start with:
  - Basic object: `Dict[generalized index, Bytes32]`.
  - Classic: `leaves: Sequence[Bytes32], proof: Sequence[Bytes32]`. Backwards compatible with the old single-value proof, and with an optional `Sequence[Generalized index]` for the leaves.
  - Tree-offsets: `offsets: Sequence[TreeOffset], chunks: Sequence[Bytes32]` for arbitrary but non-sparse proofs.
  - Room for more encodings, incl those that optimize for describing sparse-trees or partially known structures / specializations (e.g. a list of transactions). I am very sure we have not seen the end of optimizations and special use cases.
- **Paths** are defined as static singular selectors, and act like a human readable form of generalized indices. There can be different interactions with paths:
  - Type path: `path = BeaconState / validators / 123 / effective_balance`. `path.type -> Gwei`. This can be translated statically to a generalized index: `path.generalized_index()`, and parsed as one: `path = from_generalized_index(BeaconState, g_index)`
  - Use of paths for values: `eff_bal = state / path`
  - Use of path with multi-proofs (again, just an interface to access a tree): `subproof = mproof / path`
  - Use of paths with partials: `eff_bal = state_partial / path`
- **Partials** are defined as **static selectors**: an interface to read a subset of a SSZ type structure from a multi-proof. Emphasis on "selector", this is close to what IPLD (I had a call with them, Consensys, and Quilt to discuss mutlti-proofs and partials) [describes here](https://github.com/ipld/specs/blob/master/selectors/selectors.md), but a more minimal subset ("static" selection). The typed interface idea works much better in the EWASM context, as we minimize writes/reads and share a multi-proof backing easily for multiple transactions. Partials can be applied to any proof backing.
- **Dynamic selectors**: like partials, but with higher customizability for selection of data. Something to not introduce in the spec until we learn from the pitfalls from IPLD, and try different things when we have use cases for them.

djrtwo:
Looks good. Makes me think we should pull out the SSZ spec to a separate repo and version it separate from the core 2.0 specs

protolambda:
Separate repo draft can be found here: https://github.com/protolambda/eth2.0-ssz/
I am not sure if we still like to move on with this feature, as the newest shard proposal doesn't need it anymore (afaik).
Anyway, we can document this feature, and mark it as "proposal stage" in the new repo.

Marking this PR as do-not-merge until that is done.

djrtwo:
This should be closed. One of the benefits of the new sharding proposal is that we don't have to make this type of construction. Not sure we should even stage it in the new repo at this point

protolambda:
Signatures are being split off, and the custody game will not need this feature anymore, so we'll close this. If necessary we can bring it back, but no need to complicate ssz for nothing right now.

Issue #1395, opened 2019-09-04
Specify "to_bytes" and "from_bytes"


mratsim:
In https://github.com/ethereum/eth2.0-specs/blob/v0.8.3/specs/core/0_beacon-chain.md#int_to_bytes

The int_to_bytes and bytes_to_int depend on Python implementation of to_bytes

```python
def int_to_bytes(n: uint64, length: uint64) -> bytes:
    """
    Return the ``length``-byte serialization of ``n``.
    """
    return n.to_bytes(length, ENDIANNESS)
```

A description of what this method does should be added so that we don't have to look into the Python spec.

djrtwo:
I'll handle this in the `dev` branch. Seems reasonable to add another line to the header comment to clarify

protolambda:
Fixed in #1401

PR #1401, opened 2019-09-07
add explicit comments for int_to_bytes and bytes_to_int

djrtwo:
Addresses #1395.
@mratsim is this what you meant? It seems a bit verbose as the initial header comment generally says the same thing. Can you remind me what exactly your issue was?

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -560,6 +560,8 @@ def xor(bytes_1: Bytes32, bytes_2: Bytes32) -> Bytes32:
 def int_to_bytes(n: uint64, length: uint64) -> bytes:
     """
     Return the ``length``-byte serialization of ``n``.
+    ``n.to_bytes(length, endianness)`` returns ``n`` in endianness-endian bytes padded out to length bytes.
+
     """
     return n.to_bytes(length, ENDIANNESS)
 ```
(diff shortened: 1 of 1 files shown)


JustinDrake:
> the initial header comment generally says the same thing

Agreed 😂

mratsim:
I think the best would be to give an example, for example, int_to_bytes(1, little_endian) returns `0x01000000...00000000`

dankrad:
I also feel like the naming of the functions and parameters makes their behaviour very obvious. And I am usually in favour of more verbosity.

> I think the best would be to give an example, for example, int_to_bytes(1, little_endian) returns `0x01000000...00000000`

I don't understand this, it's not a padding problem. Little endian literally means that the least significant byte comes first, so how was it confusing?

djrtwo:
adjusted the comment to mention the endian-ness, but moved back to a 1-liner

specs/core/0_beacon-chain.md
```diff
@@ -569,7 +569,7 @@ def int_to_bytes(n: uint64, length: uint64) -> bytes:
 ```python
 def bytes_to_int(data: bytes) -> uint64:
     """
-    Return the integer deserialization of ``data``.
+    Return the integer deserialization of ``data`` intepretted as ``ENDIANNESS``-endian.
```
```suggestion
    Return the integer deserialization of ``data`` intepreted as ``ENDIANNESS``-endian.
```

PR #1404, opened 2019-09-08
Network specification update

AgeManning:
Provides updates to the networking specification. 

Specifically:
- Fixes the `REQ_RESP_MAX_SIZE`
- Renames the previously similar `BeaconBlocks` and `RecentBeaconBlocks` to `BeaconBlocksByRange` and `BeaconBlocksByRoot` respectively
- Removes the 1:1 mapping in `BeaconBlocksByRoot`, a responder may now return less blocks than requested
- The responder to a `BeaconBlocksByRange` request should also limit their response by the `REQ_RESP_MAX_SIZE` or `SSZ_MAX_LIST_SIZE`
- Adds the notion of a `response_chunk`. Responses that consist of a single SSZ-list (`BeaconBlocksByRange`, `BeaconBlocksByRoot`) are now sent back over the stream in individual `response_chunk`. 
- Adds clarification around the SSZ-encoding of the request/response types
- Adds clarification to the RPC requests

This extends on the improvements in #1390

### DIFF
--- a/specs/networking/p2p-interface.md
+++ b/specs/networking/p2p-interface.md
@@ -114,7 +114,7 @@ This section outlines constants that are used in this spec.
 | Name | Value | Description |
 |---|---|---|
 | `GOSSIP_MAX_SIZE` | `2**20` (= 1048576, 1 MiB) | The maximum allowed size of uncompressed gossip messages. |
-| `REQ_RESP_MAX_SIZE` | `2**20` (1048576, 1 MiB) | The maximum allowed size of uncompressed req/resp chunked responses. |
+| `MAX_CHUNK_SIZE` | `2**20` (1048576, 1 MiB) | The maximum allowed size of uncompressed req/resp chunked responses. |
 | `SHARD_SUBNET_COUNT` | `TODO` | The number of shard subnets used in the gossipsub protocol. |
 | `TTFB_TIMEOUT` | `5s` | The maximum time to wait for first byte of request response (time-to-first-byte). |
 | `RESP_TIMEOUT` | `10s` | The maximum time for complete response transfer. |
@@ -237,9 +237,9 @@ result    ::= “0” | “1” | “2” | [“128” ... ”255”]
 
 The encoding-dependent header may carry metadata or assertions such as the encoded payload length, for integrity and attack proofing purposes. Because req/resp streams are single-use and stream closures implicitly delimit the boundaries, it is not strictly necessary to length-prefix payloads; however, certain encodings like SSZ do, for added security.
 
-A `response` is formed by one or more `response_chunk`s. The exact request determines whether a response consists of a single `response_chunk` or possibly many. Responses that consist of a single SSZ-list (such as `BlocksByRange` and `BlocksByRoot`) send each list item as a `response_chunk`. All other response types (non-Lists) send a single `response_chunk`. The encoded-payload of a `response_chunk` has a maximum uncompressed byte size of `REQ_RESP_MAX_SIZE`.
+A `response` is formed by one or more `response_chunk`s. The exact request determines whether a response consists of a single `response_chunk` or possibly many. Responses that consist of a single SSZ-list (such as `BlocksByRange` and `BlocksByRoot`) send each list item as a `response_chunk`. All other response types (non-Lists) send a single `response_chunk`. The encoded-payload of a `response_chunk` has a maximum uncompressed byte size of `MAX_CHUNK_SIZE`.
 
-Clients MUST ensure the each encoded payload of a `response_chunk` is less than or equal to `REQ_RESP_MAX_SIZE`; if not, they SHOULD reset the stream immediately. Clients tracking peer reputation MAY decrement the score of the misbehaving peer under this circumstance.
+Clients MUST ensure the each encoded payload of a `response_chunk` is less than or equal to `MAX_CHUNK_SIZE`; if not, they SHOULD reset the stream immediately. Clients tracking peer reputation MAY decrement the score of the misbehaving peer under this circumstance.
 
 #### Requesting side
 
@@ -247,7 +247,7 @@ Once a new stream with the protocol ID for the request type has been negotiated,
 
 The requester MUST close the write side of the stream once it finishes writing the request message. At this point, the stream will be half-closed.
 
-The requester MUST wait a maximum of `TTFB_TIMEOUT` for the first response byte to arrive (time to first byte—or TTFB—timeout). On that happening, the requester allows a further `RESP_TIMEOUT` for each subsequent `response_chunk` received. For responses consisting of potentially many `response_chunk`s (an SSZ-list) the requester SHOULD read from the stream until either; a) An error result is received in one of the chunks, b) The responder closes the stream,  c) More than `REQ_RESP_MAX_SIZE` bytes have been read for a single `response_chunk` payload or d) More than the maximum number of requested chunks are read. For requests consisting of a single `response_chunk` and a length-prefix, the requester should read the exact number of bytes defined by the length-prefix before closing the stream.
+The requester MUST wait a maximum of `TTFB_TIMEOUT` for the first response byte to arrive (time to first byte—or TTFB—timeout). On that happening, the requester allows a further `RESP_TIMEOUT` for each subsequent `response_chunk` received. For responses consisting of potentially many `response_chunk`s (an SSZ-list) the requester SHOULD read from the stream until either; a) An error result is received in one of the chunks, b) The responder closes the stream,  c) More than `MAX_CHUNK_SIZE` bytes have been read for a single `response_chunk` payload or d) More than the maximum number of requested chunks are read. For requests consisting of a single `response_chunk` and a length-prefix, the requester should read the exact number of bytes defined by the length-prefix before closing the stream.
 
 If any of these timeouts fire, the requester SHOULD reset the stream and deem the req/resp operation to have failed.
 
@@ -267,7 +267,7 @@ If steps (1), (2), or (3) fail due to invalid, malformed, or inconsistent data, 
 
 The entire request should be read in no more than `RESP_TIMEOUT`. Upon a timeout, the responder SHOULD reset the stream.
 
-The responder SHOULD send a `response_chunk` promptly. Chunks start with a **single-byte** response code which determines the contents of the `response_chunk` (`result` particle in the BNF grammar above).
+The responder SHOULD send a `response_chunk` promptly. Chunks start with a **single-byte** response code which determines the contents of the `response_chunk` (`result` particle in the BNF grammar above). For multiple chunks, only the last chunk is allowed to have a non-zero error code (i.e. The chunk stream is terminated once an error occurs).
 
 The response code can have one of the following values, encoded as a single unsigned byte:
 
@@ -708,7 +708,7 @@ Disadvantages include:
 * Harder to stream as length must be known up-front
 * Additional code path required to verify length
 
-In some protocols, adding a length prefix serves as a form of DoS protection against very long messages, allowing the client to abort if an overlong message is about to be sent. In this protocol, we are globally limiting message sizes using `REQ_RESP_MAX_SIZE`, thus the length prefix does not afford any additional protection.
+In some protocols, adding a length prefix serves as a form of DoS protection against very long messages, allowing the client to abort if an overlong message is about to be sent. In this protocol, we are globally limiting message sizes using `MAX_CHUNK_SIZE`, thus the length prefix does not afford any additional protection.
 
 [Protobuf varint](https://developers.google.com/protocol-buffers/docs/encoding#varints) is an efficient technique to encode variable-length ints. Instead of reserving a fixed-size field of as many bytes as necessary to convey the maximum possible value, this field is elastic in exchange for 1-bit overhead per byte.
 
--- a/specs/networking/p2p-interface.md
+++ b/specs/networking/p2p-interface.md
@@ -114,8 +114,9 @@ This section outlines constants that are used in this spec.
 | Name | Value | Description |
 |---|---|---|
 | `GOSSIP_MAX_SIZE` | `2**20` (= 1048576, 1 MiB) | The maximum allowed size of uncompressed gossip messages. |
-| `REQ_RESP_MAX_SIZE` | `2**20` (1048576, 1 MiB) | The maximum allowed size of uncompressed req/resp messages. |
-| `SSZ_MAX_LIST_SIZE` | `TODO` | The maximum size of SSZ-encoded variable lists. |
+| `REQ_RESP_MAX_SIZE` | `2**20` (1048576, 1 MiB) | The maximum allowed size of uncompressed req/resp chunked responses. |
+| `MAX_REQUESTED_BLOCKS` | `20` | The maximum allowed number of blocks that can
+be requested. |
 | `SHARD_SUBNET_COUNT` | `TODO` | The number of shard subnets used in the gossipsub protocol. |
 | `TTFB_TIMEOUT` | `5s` | The maximum time to wait for first byte of request response (time-to-first-byte). |
 | `RESP_TIMEOUT` | `10s` | The maximum time for complete response transfer. |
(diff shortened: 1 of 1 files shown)


specs/networking/p2p-interface.md
```diff
@@ -113,7 +113,7 @@ This section outlines constants that are used in this spec.
 
 | Name | Value | Description |
 |---|---|---|
-| `REQ_RESP_MAX_SIZE` | `TODO` | The maximum size of uncompressed req/resp messages that clients will allow. |
+| `REQ_RESP_MAX_SIZE` | `2**22` (4194304, 4 MiB) | The maximum size of uncompressed req/resp messages that clients will allow. |
```
djrtwo: After discussions about bandwidth requirements, considering dropping this down to 1MB

specs/networking/p2p-interface.md
```diff
@@ -330,26 +330,28 @@ The [SimpleSerialize (SSZ) specification](../simple-serialize.md) outlines how o
...
+- `head_fork_version`: The beacon_state `Fork` version.
 - `finalized_root`: The latest finalized root the node knows about.
 - `finalized_epoch`: The latest finalized epoch the node knows about.
 - `head_root`: The block hash tree root corresponding to the head of the chain as seen by the sending node.
 - `head_slot`: The slot corresponding to the `head_root`.
 
-Clients exchange hello messages upon connection, forming a two-phase handshake. The first message the initiating client sends MUST be the hello message. In response, the receiving client MUST respond with its own hello message.
+The dialing client MUST send a `Hello` request upon connection.
+
+This should be encoded as an SSZ-container.
```
djrtwo:
```suggestion
This MUST be encoded as an SSZ-container.
```

specs/networking/p2p-interface.md
```diff
@@ -365,11 +367,13 @@ Clients MAY use reason codes above `128` to indicate alternative, erroneous requ
 
 The range `[4, 127]` is RESERVED for future usage.
 
-#### BeaconBlocks
+This should not be encoded as an SSZ-container.
```
djrtwo: MUST not?
Maybe we should change this to 
```suggestion
This MUST be encoded as a single SSZ-field.
```

specs/networking/p2p-interface.md
```diff
@@ -388,15 +392,24 @@ Response Content:
 
 Requests count beacon blocks from the peer starting from `start_slot` on the chain defined by `head_block_root`. The response MUST contain no more than count blocks. `step` defines the slot increment between blocks. For example, requesting blocks starting at `start_slot` 2 with a step value of 2 would return the blocks at [2, 4, 6, …]. In cases where a slot is empty for a given slot number, no block is returned. For example, if slot 4 were empty in the previous example, the returned array would contain [2, 6, …]. A step value of 1 returns all blocks on the range `[start_slot, start_slot + count)`.
 
-`BeaconBlocks` is primarily used to sync historical blocks.
+The request is encoded as an SSZ-container, the response is not encoded as an
```
djrtwo:
```suggestion
The request MUST be encoded as an SSZ-container, the response MUST be encoded as a single SSZ-field.
```

specs/networking/p2p-interface.md
```diff
@@ -388,15 +392,24 @@ Response Content:
...
+SSZ container.
```
djrtwo:
```suggestion
```

specs/networking/p2p-interface.md
```diff
@@ -414,12 +427,18 @@ Response Content:
 )
 ```
 
-Requests blocks by their block roots. The response is a list of `BeaconBlock` with the same length as the request. Blocks are returned in order of the request and any missing/unknown blocks are left empty (SSZ null `BeaconBlock`).
+Requests blocks by their block roots. The response is a list of `BeaconBlock` whose length is less or equal to the number of requested blocks. It may be less in the case that the responding peer is missing blocks.
+
+`BeaconBlocksByRoot` is primarily used to recover recent blocks (ex. when receiving a block or attestation whose parent is unknown).
 
-`RecentBeaconBlocks` is primarily used to recover recent blocks (ex. when receiving a block or attestation whose parent is unknown).
+Both the request and the response should not be encoded as an SSZ-container.
```
djrtwo:
```suggestion
Both the request and the response MUST be encoded as a single SSZ-field.
```

specs/networking/p2p-interface.md
```diff
@@ -260,14 +261,14 @@ The responder MUST:
...
-4. Write the response (result, optional header, payload).
+4. Write the response which may consist of one or many `response_chunks` (result, optional header, payload).
 5. Close their write side of the stream. At this point, the stream will be fully closed.
 
 If steps (1), (2), or (3) fail due to invalid, malformed, or inconsistent data, the responder MUST respond in error. Clients tracking peer reputation MAY record such failures, as well as unexpected events, e.g. early stream resets.
 
 The entire request should be read in no more than `RESP_TIMEOUT`. Upon a timeout, the responder SHOULD reset the stream.
 
-The responder SHOULD send a response promptly, starting with a **single-byte** response code which determines the contents of the response (`result` particle in the BNF grammar above).
+The responder SHOULD send a response promptly, starting with a **single-byte** response code which determines the contents of the `response_chunk` (`result` particle in the BNF grammar above).
```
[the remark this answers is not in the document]
AgeManning: It should be with respect to the `response_chunk`. We have a TTFB (time to first byte) which starts once the stream is open. This should probably worded more like:
```
- The responder SHOULD send a response_chunk promptly. Chunks start with a **single-byte** ...
```

zah:
If the `hello` message is now valid at any time, shall we rename it to `status`? A lot of people will associate the term "hello" with something that is appropriate only at the start of the session.

AgeManning:
I have updated this PR to accommodate suggestions. 
Specifically, we have changed the `REQ_RESP_MAX_SIZE` to apply to the encoded payload of the request/responses. This leads us to introduce `MAX_REQUESTS_BLOCKS` which is set to 20, which specifies the maximum number of blocks we can request.

specs/networking/p2p-interface.md
```diff
@@ -231,23 +230,24 @@ Request/response messages MUST adhere to the encoding specified in the protocol
 
 ```
 request   ::= <encoding-dependent-header> | <encoded-payload>
-response  ::= <result> | <encoding-dependent-header> | <encoded-payload>
+response  ::= <response_chunk>+
+response_chunk  ::= <result> | <encoding-dependent-header> | <encoded-payload>
```
raulk: Is the result-per-chunk pattern deliberate? Would the result be a global result per req/rep?

specs/networking/p2p-interface.md
```diff
@@ -113,9 +113,8 @@ This section outlines constants that are used in this spec.
 
 | Name | Value | Description |
 |---|---|---|
-| `REQ_RESP_MAX_SIZE` | `TODO` | The maximum size of uncompressed req/resp messages that clients will allow. |
-| `SSZ_MAX_LIST_SIZE` | `TODO` | The maximum size of SSZ-encoded variable lists. |
-| `GOSSIP_MAX_SIZE` | `2**20` (= 1048576, 1 MiB) | The maximum size of uncompressed gossip messages. |
+| `GOSSIP_MAX_SIZE` | `2**20` (= 1048576, 1 MiB) | The maximum allowed size of uncompressed gossip messages. |
+| `REQ_RESP_MAX_SIZE` | `2**20` (1048576, 1 MiB) | The maximum allowed size of uncompressed req/resp chunked responses. |
```
raulk: Should this be renamed to `CHUNK_MAX_SIZE`?

specs/networking/p2p-interface.md
```diff
@@ -231,23 +230,24 @@ Request/response messages MUST adhere to the encoding specified in the protocol
...
-response  ::= <result> | <encoding-dependent-header> | <encoded-payload>
+response  ::= <response_chunk>+
+response_chunk  ::= <result> | <encoding-dependent-header> | <encoded-payload>
 result    ::= “0” | “1” | “2” | [“128” ... ”255”]
 ```
 
 The encoding-dependent header may carry metadata or assertions such as the encoded payload length, for integrity and attack proofing purposes. Because req/resp streams are single-use and stream closures implicitly delimit the boundaries, it is not strictly necessary to length-prefix payloads; however, certain encodings like SSZ do, for added security.
 
-`encoded-payload` has a maximum byte size of `REQ_RESP_MAX_SIZE`.
+A `response` is formed by one or more `response_chunk`s. The exact request determines whether a response consists of a single `response_chunk` or possibly many. Responses that consist of a single SSZ-list (such as `BlocksByRange` and `BlocksByRoot`) send each list item as a `response_chunk`. All other response types (non-Lists) send a single `response_chunk`. The encoded-payload of a `response_chunk` has a maximum uncompressed byte size of `REQ_RESP_MAX_SIZE`.
```
raulk: This can incur in excessive fragmentation/overhead/underutilisation if list items are small, as you're effectively defining a 1:1 mapping between chunk and list element.

I'd recommend to make the boundaries of chunks full list elements, allowing a chunk to contain multiple _complete_ list elements, and disallowing partial elements or bleeding over between chunks.

It may be worth formally introducing the term `chunkable`, so you can label message fields that satisfy this property in the schemas below.

specs/networking/p2p-interface.md
```diff
@@ -231,23 +230,24 @@ Request/response messages MUST adhere to the encoding specified in the protocol
...
 
-Clients MUST ensure the payload size is less than or equal to `REQ_RESP_MAX_SIZE`; if not, they SHOULD reset the stream immediately. Clients tracking peer reputation MAY decrement the score of the misbehaving peer under this circumstance.
+Clients MUST ensure the each encoded payload of a `response_chunk` is less than or equal to `REQ_RESP_MAX_SIZE`; if not, they SHOULD reset the stream immediately. Clients tracking peer reputation MAY decrement the score of the misbehaving peer under this circumstance.
```
raulk: Rename constant to `CHUNK_MAX_SIZE` for clarity?

specs/networking/p2p-interface.md
```diff
@@ -231,23 +230,24 @@ Request/response messages MUST adhere to the encoding specified in the protocol
...
+The requester MUST close the write side of the stream once it finishes writing the request message. At this point, the stream will be half-closed.
 
-The requester MUST wait a maximum of `TTFB_TIMEOUT` for the first response byte to arrive (time to first byte—or TTFB—timeout). On that happening, the requester will allow further `RESP_TIMEOUT` to receive the full response.
+The requester MUST wait a maximum of `TTFB_TIMEOUT` for the first response byte to arrive (time to first byte—or TTFB—timeout). On that happening, the requester allows a further `RESP_TIMEOUT` for each subsequent `response_chunk` received. For responses consisting of potentially many `response_chunk`s (an SSZ-list) the requester SHOULD read from the stream until either; a) An error result is received in one of the chunks, b) The responder closes the stream,  c) More than `REQ_RESP_MAX_SIZE` bytes have been read for a single `response_chunk` payload or d) More than the maximum number of requested chunks are read. For requests consisting of a single `response_chunk` and a length-prefix, the requester should read the exact number of bytes defined by the length-prefix before closing the stream.
```
raulk: Beware of a starvation DoS attack where attackers could deliberately trickle their chunks slowly so as to keep that socket/file descriptor busy as long as possible.

specs/networking/p2p-interface.md
zah: Please note that adding chunks does not increase the size of the overall payload, because we are replacing the leading 4-byte offsets that would appear in the SSZ lists before with response codes and varint lengths which will often amount to just 2 bytes per chunk.

specs/networking/p2p-interface.md
```diff
@@ -260,16 +260,16 @@ The responder MUST:
...
 
 If steps (1), (2), or (3) fail due to invalid, malformed, or inconsistent data, the responder MUST respond in error. Clients tracking peer reputation MAY record such failures, as well as unexpected events, e.g. early stream resets.
 
 The entire request should be read in no more than `RESP_TIMEOUT`. Upon a timeout, the responder SHOULD reset the stream.
 
-The responder SHOULD send a response promptly, starting with a **single-byte** response code which determines the contents of the response (`result` particle in the BNF grammar above).
+The responder SHOULD send a `response_chunk` promptly. Chunks start with a **single-byte** response code which determines the contents of the `response_chunk` (`result` particle in the BNF grammar above).
 
-It can have one of the following values, encoded as a single unsigned byte:
+The response code can have one of the following values, encoded as a single unsigned byte:
```
arnetheduck: We should highlight here that only the last chunk may have a value different from `success`

specs/networking/p2p-interface.md
```diff
@@ -330,29 +334,33 @@ The [SimpleSerialize (SSZ) specification](../simple-serialize.md) outlines how o
...
 
-- `fork_version`: The beacon_state `Fork` version.
+- `head_fork_version`: The beacon_state `Fork` version.
 - `finalized_root`: The latest finalized root the node knows about.
 - `finalized_epoch`: The latest finalized epoch the node knows about.
 - `head_root`: The block hash tree root corresponding to the head of the chain as seen by the sending node.
 - `head_slot`: The slot corresponding to the `head_root`.
 
-Clients exchange hello messages upon connection, forming a two-phase handshake. The first message the initiating client sends MUST be the hello message. In response, the receiving client MUST respond with its own hello message.
+The dialing client MUST send a `Status` request upon connection.
```
prestonvanloon: Any recommendation for the scenario that the dialing client doesn't send status? Disconnect after 5 seconds? Send the dialing client a `Status`?

specs/networking/p2p-interface.md
```diff
@@ -414,12 +434,18 @@ Response Content:
 )
 ```
 
-Requests blocks by their block roots. The response is a list of `BeaconBlock` with the same length as the request. Blocks are returned in order of the request and any missing/unknown blocks are left empty (SSZ null `BeaconBlock`).
+Requests blocks by their block roots. The response is a list of `BeaconBlock` whose length is less than or equal to the number of requested blocks. It may be less in the case that the responding peer is missing blocks.
 
-`RecentBeaconBlocks` is primarily used to recover recent blocks (ex. when receiving a block or attestation whose parent is unknown).
+`BeaconBlocksByRoot` is primarily used to recover recent blocks (e.g. when receiving a block or attestation whose parent is unknown).
+
+The request MUST be encoded as an SSZ-field.
```
prestonvanloon: Is this still correct? The chunks are ssz containers now, `BeaconBlock`. I.e. it's a series of streamed containers rather than a ssz-field `[]BeaconBlock`.

prestonvanloon: Oops, i misread request/response. Please disregard

arnetheduck:
> LGTM. I would like a recommendation for the scenario when a dialing peer does not send a `Status` immediately.

timeout/disconnect - no point spending resources if they're not following the protocol

specs/networking/p2p-interface.md
arnetheduck: timeout/disconnect (see belo)

specs/networking/p2p-interface.md
```diff
@@ -382,21 +394,29 @@ Request Content
...
+Clients MUST respond with at least one block, if they have it.
+
+Clients MUST order blocks by increasing slot number.
+
+#### BeaconBlocksByRoot
 
-**Protocol ID:** `/eth2/beacon_chain/req/recent_beacon_blocks/1/`
+**Protocol ID:** `/eth2/beacon_chain/req/beacon_blocks_by_root/1/`
 
 Request Content:
```
nisdas: shouldn't both the request and response be a list of roots/blocks rather than a container ? 
Request Content:
```

(
  []HashTreeRoot
)
```
Response Content:
```
(
 []BeaconBlock
)
```

djrtwo: Looks like it

specs/networking/p2p-interface.md
```diff
@@ -231,23 +230,24 @@ Request/response messages MUST adhere to the encoding specified in the protocol
...
+The requester MUST close the write side of the stream once it finishes writing the request message. At this point, the stream will be half-closed.
 
-The requester MUST wait a maximum of `TTFB_TIMEOUT` for the first response byte to arrive (time to first byte—or TTFB—timeout). On that happening, the requester will allow further `RESP_TIMEOUT` to receive the full response.
+The requester MUST wait a maximum of `TTFB_TIMEOUT` for the first response byte to arrive (time to first byte—or TTFB—timeout). On that happening, the requester allows a further `RESP_TIMEOUT` for each subsequent `response_chunk` received. For responses consisting of potentially many `response_chunk`s (an SSZ-list) the requester SHOULD read from the stream until either; a) An error result is received in one of the chunks, b) The responder closes the stream,  c) More than `MAX_CHUNK_SIZE` bytes have been read for a single `response_chunk` payload or d) More than the maximum number of requested chunks are read. For requests consisting of a single `response_chunk` and a length-prefix, the requester should read the exact number of bytes defined by the length-prefix before closing the stream.
```
ChihChengLiang: > d) More than the maximum number of requested chunks are read.

@AgeManning Should this number be specified?

AgeManning: The two types that can request chunks are `BeaconBlocksByRange` and `BeaconBlocksByRoot`.  Each specify in the request the maximum number of blocks/chunks that should be returned. 

For `BeaconBlocksByRange` this is the `count` parameter, for `BeaconBlocksByRoot` this is the number of hashes requested.

Let me know if this doesn't answer your question :)

ChihChengLiang: That makes perfect sense. Thank you for the clarification ❤️.

PR #1407, opened 2019-09-15
Custody game cleanup

GregTheGreek:
Fix table


specs/core/1_custody-game.md
```diff
@@ -339,7 +340,7 @@ def legendre_bit(a: int, q: int) -> int:
 Given one proof of custody chunk, returns the proof of custody subchunks of the correct sizes.
 
 ```python
-def custody_subchunkify(bytez: bytes) -> list:
+def custody_subchunkify(bytez: bytes) -> list[int]:
```
djrtwo: This should be a `list[bytes]`

```suggestion
def custody_subchunkify(bytez: bytes) -> list[bytes]:
```

PR #1410, opened 2019-09-20
sync protocol: clarify committee type

wemeetagain:
Updates at period boundaries are period committee updates, not persistent committee updates.


djrtwo:
depends on #1383

PR #1414, opened 2019-09-23
Add seed domains

djrtwo:
This PR reuses domain types for seeds (in addition to signatures) to isolate sources randomness for different functions. The motivating factor was to cleanly prevent phase 1 persistent committees from colliding with phase 0 crosslink committees once per `PERSISTENT_COMMITTEE_PERIOD`.


djrtwo:
closing in favor of #1415

PR #1415, opened 2019-09-23
Add seed domains

djrtwo:
This PR reuses domain types for seeds (in addition to signatures) to isolate the source of randomness for different functions. The motivating factor was to cleanly prevent phase 1 persistent committees from colliding with phase 0 crosslink committees once per `PERSISTENT_COMMITTEE_PERIOD`.

`ACTIVATION_EXIT_DELAY` was also changed to `MAX_SEED_LOOKAHEAD` to be more generic and descriptive in its phase 1 usage. Also note that `MAX_SEED_LOOKAHEAD` will increase with VDF-based randomness.


specs/core/0_beacon-chain.md
```diff
@@ -250,15 +250,15 @@ The following values are (non-configurable) constants used throughout the specif
 | `MAX_VOLUNTARY_EXITS` | `2**4` (= 16) |
 | `MAX_TRANSFERS` | `0` |
 
-### Signature domain types
+### Domain types
```
djrtwo: I'm a little bit unsure about conflating the type of signature and seed domain. I understand the intuition, but the usage of `DOMAIN_BEACON_ATTESTER` for both an attestation signature and committee shuffling seed domain is a bit dirty.

Thoughts on having both "Signature" and "Seed Domains"?

djrtwo: The only reason I am for the singular domain type is to be minimally invasive to the phase 0 spec..

specs/core/1_shard-data-chains.md
```diff
@@ -270,7 +270,8 @@ def get_shard_block_proposer_index(state: BeaconState,
         return None
 
     MAX_RANDOM_BYTE = 2**8 - 1
-    seed = hash(get_seed(state, current_epoch) + int_to_bytes(shard, length=8) + int_to_bytes(slot, length=8))
+    seed = get_seed(state, current_epoch, DOMAIN_SHARD_PROPOSER)
```
djrtwo: Suggest calling this `epoch_seed` or something to distinguish it from the next assignment.
```suggestion
    epoch_seed = get_seed(state, current_epoch, DOMAIN_SHARD_PROPOSER)
```

An alternative would be something like:

```python
seed = hash(get_seed(state, current_epoch, DOMAIN_SHARD_PROPOSER)
            + int_to_bytes(shard, length=8)
            + int_to_bytes(slot, length=8))

djrtwo:
@JustinDrake See review

specs/core/1_shard-data-chains.md
JustinDrake: Went with `epoch_seed`. This will all get cleaned up shortly anyway with #1383.

specs/core/0_beacon-chain.md
JustinDrake: I'd say merging the two is fine. Semantically, the "lifecycle" of a validator role (beacon/shard attester/proposer) is to first get sampled for that role and then sign a unique message under that specific role. So semantically it makes sense to merge the seed and signature domains.

Avoiding two types (`SeedDomainType` and `SignatureDomainType`), four extra constants, as well as verbosity (e.g. `SIGNATURE_DOMAIN_BEACON_ATTESTER` as opposed to just `DOMAIN_BEACON_ATTESTER`) is also nice.

PR #1416, opened 2019-09-27
Shard chain CI cleanup

djrtwo:
Making CI happy for #1383

* Cleaned up some constants in `1_beacon-chain-misc`
* fix minor bug in `process_delta` -- `beacon_state.slot` --> `shard_state.slot`
* fix minor bug in `update_period_committee` -- order of arguments was incorrect
* assert `proposer_index is not None` when getting proposer index to make MyPy happy (because of the `Optional` return type
* add some missing `Gwei` type castings

Note:

* This does _not_ cleanup `shard_receipt_proof`s wrt to the updates in #1383. My plan of attack is to use this PR for minor fixes and use a separate PR for the substantive change to shard rewards processing into the beacon chain
* This does _not_ contain additional tests, it just gets the current ones passing so that #1383 can move forward

### DIFF
--- a/specs/core/1_beacon-chain-misc.md
+++ b/specs/core/1_beacon-chain-misc.md
@@ -226,16 +226,18 @@ def update_period_committee(state: BeaconState) -> None:
     """
     Updates period committee roots at boundary blocks.
     """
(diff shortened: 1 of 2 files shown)


djrtwo:
Closing in favor of #1417

PR #1417, opened 2019-09-27
Shard chain CI cleanup

djrtwo:
Making CI happy for #1383

* Cleaned up some constants in `1_beacon-chain-misc`
* fix minor bug in `process_delta` -- `beacon_state.slot` --> `shard_state.slot`
* fix minor bug in `update_period_committee` -- order of arguments was incorrect
* assert `proposer_index is not None` when getting proposer index to make MyPy happy (because of the `Optional` return type
* add some missing `Gwei` type castings

Note:

* This does _not_ cleanup `shard_receipt_proof`s wrt to the updates in #1383, and still utilizes the old `receipt_root`. My plan of attack is to use this PR for minor fixes and use a separate PR for the substantive change to shard rewards processing into the beacon chain
* This does _not_ contain additional tests, it just gets the current ones passing so that #1383 can move forward

### DIFF
--- a/specs/core/1_shard-data-chains.md
+++ b/specs/core/1_shard-data-chains.md
@@ -356,6 +356,8 @@ def process_shard_block_header(beacon_state: BeaconState, shard_state: ShardStat
     assert block.block_size_sum == shard_state.block_size_sum
     # Verify proposer signature
     proposer_index = get_shard_proposer_index(beacon_state, shard_state.shard, block.slot)
+    assert proposer_index is not None
+
     domain = get_domain(beacon_state, DOMAIN_SHARD_PROPOSER, compute_epoch_of_shard_slot(block.slot))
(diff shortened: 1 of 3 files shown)


specs/core/1_shard-data-chains.md
```diff
@@ -383,8 +385,9 @@ def process_shard_attestations(beacon_state: BeaconState, shard_state: ShardStat
     assert bls_verify(bls_aggregate_pubkeys(pubkeys), message, block.attestations, domain)
     # Proposer micro-reward
     proposer_index = get_shard_proposer_index(beacon_state, shard_state.shard, block.slot)
+    assert proposer_index is not None
```
JustinDrake: The various `assert proposer_index is not None` can be subsumed into a single one `assert` in `get_shard_proposer_index`.

PR #1418, opened 2019-09-27
[WIP] Shard chain sanity test

djrtwo:
_Work in Progress_
## content

Adding basic shard chain sanity tests to pass through the common flows and build out the basic test tooling/helpers to build such tests.

## notes
Contains commits found in #1417 and should be merged after. Currently is open against #1383 but will swap when that gets merged.

As discussed with @protolambda, I realized that the `@always_bls` decorator is order dependent with `@spec_state_test` and the basic shard block tests were being skipped due to this. This deserves an investigation to (1) see if any other tests are affected by this and (2) check into removing this order dependency (or at least throw a warning if done incorrectly)

### DIFF
--- a/specs/core/1_shard-data-chains.md
+++ b/specs/core/1_shard-data-chains.md
@@ -356,6 +356,8 @@ def process_shard_block_header(beacon_state: BeaconState, shard_state: ShardStat
     assert block.block_size_sum == shard_state.block_size_sum
     # Verify proposer signature
     proposer_index = get_shard_proposer_index(beacon_state, shard_state.shard, block.slot)
+    assert proposer_index is not None
+
     domain = get_domain(beacon_state, DOMAIN_SHARD_PROPOSER, compute_epoch_of_shard_slot(block.slot))
     assert bls_verify(beacon_state.validators[proposer_index].pubkey, signing_root(block), block.signature, domain)
 ```
@@ -381,8 +383,9 @@ def process_shard_attestations(beacon_state: BeaconState, shard_state: ShardStat
     assert bls_verify(bls_aggregate_pubkeys(pubkeys), message, block.attestations, domain)
     # Proposer micro-reward
     proposer_index = get_shard_proposer_index(beacon_state, shard_state.shard, block.slot)
+    assert proposer_index is not None
     reward = attestation_count * get_base_reward(beacon_state, proposer_index) // PROPOSER_REWARD_QUOTIENT
-    process_delta(beacon_state, shard_state, proposer_index, reward)
+    process_delta(beacon_state, shard_state, proposer_index, Gwei(reward))
 ```
 
 #### Block body
@@ -394,8 +397,9 @@ def process_shard_block_body(beacon_state: BeaconState, shard_state: ShardState,
     # Apply proposer block body fee
     block_body_fee = shard_state.block_body_price * len(block.body) // MAX_SHARD_BLOCK_SIZE
     proposer_index = get_shard_proposer_index(beacon_state, shard_state.shard, block.slot)
-    process_delta(beacon_state, shard_state, proposer_index, block_body_fee, positive=False)  # Burn
-    process_delta(beacon_state, shard_state, proposer_index, block_body_fee // PROPOSER_REWARD_QUOTIENT)  # Reward
+    assert proposer_index is not None
+    process_delta(beacon_state, shard_state, proposer_index, Gwei(block_body_fee), positive=False)  # Burn
+    process_delta(beacon_state, shard_state, proposer_index, Gwei(block_body_fee // PROPOSER_REWARD_QUOTIENT))  # Reward
     # Calculate new block body price
     block_size = SHARD_HEADER_SIZE + len(block.body)
     QUOTIENT = MAX_SHARD_BLOCK_SIZE * BLOCK_BODY_PRICE_QUOTIENT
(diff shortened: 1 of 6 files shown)


djrtwo:
@JustinDrake, I specifically want input on how to handle `beacon_block_root` wrt the shard slot

specs/core/1_shard-data-chains.md
```diff
@@ -374,12 +384,17 @@ def process_shard_attestations(beacon_state: BeaconState, shard_state: ShardStat
             pubkeys.append(beacon_state.validators[validator_index].pubkey)
             process_delta(beacon_state, shard_state, validator_index, get_base_reward(beacon_state, validator_index))
             attestation_count += 1
+    # Exit early if no participants
```
JustinDrake: What is the rationale behind this?

specs/core/1_shard-data-chains.md
```diff
@@ -333,10 +338,15 @@ def process_shard_block_header(beacon_state: BeaconState, shard_state: ShardStat
     # Verify the slot number
     assert block.slot == shard_state.slot
     # Verify the beacon chain root
-    parent_epoch = compute_epoch_of_shard_slot(shard_state.latest_block_header.slot)
-    assert block.beacon_block_root == get_block_root(beacon_state, parent_epoch)
+    epoch = compute_epoch_of_shard_slot(shard_state.slot)
+    assert epoch == compute_epoch_of_slot(beacon_state.slot)
+    if epoch * SLOTS_PER_EPOCH == beacon_state.slot:
```
JustinDrake: I'm tempted to force the beacon state to be at an epoch transition, i.e. `assert epoch * SLOTS_PER_EPOCH == beacon_state.slot`.

djrtwo: This is already the case. What this code is doing is retrieving the block root when `slot % SLOTS_PER_EPOCH == 0`. At this type of slot, the block root is not yet in the state so you can't get it from `get_block_root`.  You have to get it from the `latest_block_header`.

actually, now that I think of it, you would need to be even more verbose and add the state root into the block header

```
header = deepcopy(beacon_state.latest_block_header)
header.state_root = hash_tree_root(beacon_state)
beacon_block_root = signing_root(header)
```

(can do it without the `deepcopy`. just for illustration)

specs/core/1_shard-data-chains.md
djrtwo: A block proposer might presumably not have any attestations from the previous slot to include. Need to handle that case explicitly.

This was never the case for phase 0 attestations, because there is no value in including an empty attestation. You can just have an empty list of attestations if there were actually not attestations to include. Here, though, you could easily have a case where there were not attestatiosn from the previous slot but a block still needs to be able to be built

specs/core/1_shard-data-chains.md
djrtwo: edit:
> I'm tempted to force the beacon state to be at an epoch transition, i.e. assert epoch * SLOTS_PER_EPOCH == beacon_state.slot.

it's the case because of the fact that `get_block_root` takes in an epoch and gives the start block of the epoch so the assertion that `block.beacon_block_root == beacon_block_root`

specs/core/1_shard-data-chains.md
JustinDrake: > This was never the case for phase 0 attestations

I believe phase 0 attestations can be empty, i.e. `attestation.aggregation_bits` can be all zero bits. My guess is we don't need special logic in phase 1, similar to how we don't have special logic in phase 0.

specs/core/1_shard-data-chains.md
JustinDrake: While it is the case that the `get_block_root` is invariant within an epoch it's not necessarily the case for other pieces of beacon state. Just to be on the safe side I would force `shard_state_transition` to only accept the `BeaconState` if `beacon_state.slot % SLOTS_PER_EPOCH == 0`. It would also simplify the logic a bit in this specific instance (no need for `else` clause).

specs/core/1_shard-data-chains.md
djrtwo: What is the valid signature for no participants?
If there isn't one, then we need this special case

JustinDrake: > What is the valid signature for no participants?

It's the G2 point at infinity (i.e. `0 * g2` where `g2` is the `G2` generator). The reason is that the "empty" pubkey is the G1 point at infinity (i.e. `0 * g1` where `g1` is the `G1` generator) and `e(g1, sig) = e(0 * g1, H(m))` holds iff `sig` equals `0 * g2`.

> If there isn't one, then we need this special case

There is one (and it's unique!)—no need for a special case :)

specs/core/1_shard-data-chains.md
djrtwo: ah, I see now. Will change it up

djrtwo: Did this. You still have to do a conditional check on the `state_root == Bytes32` and fill in if necessary. (can't predict it ahead of time because there may or may not have been at block at the beacon slot in question)

djrtwo:
> current method is also not fully correct. It assumed at slot 0 of an epoch that the shard block proposer already has the beacon block from slot 0 of that same epoch which is not a reasonable assumption as the shard proposer is supposed to create a block at the start of the slot (before the beacon block could propogate). We could consider the beacon state root epoch to come from shard_state.slot - 1. I think this works, but is a bit hacky. Open to discussion/alternatives

Need to discuss this note I left in the header comment. Still not resolved as is

djrtwo:
I'm of the mind to get this and #1383 merged asap. Then continue conversations about the shape of phase 1 offline

PR #1420, opened 2019-09-28
K coverage tests

denis-bogdanas:
2 tests for code not covered by K coverage analysis.
Test 1: `get_matching_target_attestations`, line 842, case false.
Test 2: `get_beacon_proposer_index`, case when `while` loop has more than one iteration.

### DIFF
--- a/test_generators/epoch_processing/main.py
+++ b/test_generators/epoch_processing/main.py
@@ -7,6 +7,7 @@ from eth2spec.test.phase_0.epoch_processing import (
     test_process_final_updates,
     test_process_justification_and_finalization,
     test_process_registry_updates,
+    test_process_rewards_and_penalties,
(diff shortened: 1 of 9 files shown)


protolambda:
Hey, thank you for your patience, we have been busy around Devcon. Looking into this now :+1:

protolambda:
There is a bunch of code duplication (understandable, customization of genesis state wasn't there), but I cleaned it up, and added a custom-state context decorator.

Have a look at https://github.com/ethereum/eth2.0-specs/tree/k-cov-improv, I think we can merge with that (open for suggestions). Thank you for looking into test coverage!

denis-bogdanas:
@protolambda Thanks for cleanup. Looks good from partial inspection.

PR #1427, opened 2019-10-12
[WIP] New shard proposal

vbuterin:
Still very incomplete and being worked on.

Known TODOs:

* Flesh out fraud proofs into a complete specification


Good2no1:
Have any tips for a project that is working to combine a custom p2p network protocol(similar to torrenting) with multi layer nodes, a PoS model and consensus?The purpose is for a new and unique file sharing network by building most from scratch to ensure an optimum user experience.

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+        # Slot the attestation starts counting from
+        start_slot = state.shard_next_slots[shard]
+        # Correct data root count
+        offset_slots = [start_slot + x for x in SHARD_BLOCK_OFFSETS if start_slot + x < state.slot]
+        assert len(attestation.custody_bits) == len(offset_slots)
+        # Correct parent block root
+        assert data.beacon_block_root == get_block_root_at_slot(state, state.slot - 1)
+        # Apply
+        online_indices = get_online_indices(state)
+        if get_total_balance(state, online_indices.intersection(attesting_indices)) * 3 >= get_total_balance(state, online_indices) * 2:
```
djrtwo: Interesting implication of this:

In the previous design, attestations that were for the same crosslink data but were not aggregatable due to some amount of overlapping indices could both be included and count towards crosslinks and finality.

In this design, they _must_ be aggregated to both count toward the crosslinking, whereas non-aggregated can still both go toward finality.

djrtwo: This seems fine, but puts a larger reliance on good, non-overlapping aggregation

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+            # Check correct formatting of shard transition data
```
djrtwo: Need to consider what happens if a fully duplicate attestation (or a not fully duplicate attestation, but both have 2/3 voting on a crosslink) is included. This would create two `PendingAttestations` with `crosslink_success` as `True`

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+    committee = get_crosslink_committee(state, get_current_epoch(state), shard)
+    for bits in attestation.custody_bits + [attestation.aggregation_bits]:
+        assert len(bits) == len(committee)
+    # Check signature
+    assert is_valid_indexed_attestation(state, get_indexed_attestation(state, attestation))
+    # Get attesting indices
+    attesting_indices = get_attesting_indices(state, attestation.data, attestation.aggregation_bits)
+
+    # Prepare pending attestation object
+    pending_attestation = PendingAttestation(
```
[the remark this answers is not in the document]
djrtwo: pending_attestation already has `data` as a field so no need to add `slot` or `shard` explicitly

specs/core/1_new_shards.md
djrtwo: Depending on how rewards are done, it is likely fine. Just need to get the union of the successful participants at `slot/index`.

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+            state.shard_states[shard].slot = state.slot - 1
+
+            # Save success (for end-of-epoch rewarding)
+            pending_attestation.crosslink_success = True
+
+            # Apply proposer reward and cost
+            estimated_attester_reward = sum([get_base_reward(state, attester) for attester in attesting_indices])
+            increase_balance(state, proposer, estimated_attester_reward // PROPOSER_REWARD_COEFFICIENT)
+            for state, length in zip(transition.shard_states, transition.shard_block_lengths):
+                decrease_balance(state, proposer, state.gasprice * length)
```
djrtwo: Why is there a cost for including attestations? We don't currently have a `state.gasprice`.
Should this instead be a cost to the shard proposers?

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+            # Save updated state
+            state.shard_states[shard] = transition.shard_states[-1]
+            state.shard_states[shard].slot = state.slot - 1
+
+            # Save success (for end-of-epoch rewarding)
+            pending_attestation.crosslink_success = True
+
+            # Apply proposer reward and cost
+            estimated_attester_reward = sum([get_base_reward(state, attester) for attester in attesting_indices])
+            increase_balance(state, proposer, estimated_attester_reward // PROPOSER_REWARD_COEFFICIENT)
```
djrtwo: Why not do this reward during epoch transition?

djrtwo: Is this an _additional_ reward for successful crosslink inclusion?
Rather than the attestation inclusion reward that already exists in phase 0?

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+            for state, length in zip(transition.shard_states, transition.shard_block_lengths):
+                decrease_balance(state, proposer, state.gasprice * length)
+            
+    # Type 2: delayed attestations
+    else:
+        assert slot_to_epoch(data.slot) in (get_current_epoch(state), get_previous_epoch(state))
```
djrtwo: Should this be within the last `SLOTS_PER_EPOCH` slots? That is how it is in phase 0 so there aren't asymmetries depending on which slot you are assigned to

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+        if state.shard_states[shard].slot != state.slot - 1:
+            assert block.shard_transition[shard] == ShardTransition()
+```
+
+### Light client processing
+
+```python
+def verify_light_client_signatures(state: BeaconState, block: BeaconBlock):
+    period_start = get_current_epoch(state) - get_current_epoch(state) % LIGHT_CLIENT_COMMITTEE_PERIOD
+    committee = get_light_client_committee(state, period_start - min(period_start, LIGHT_CLIENT_COMMITTEE_PERIOD))
```
djrtwo: Why the `min` here?

Shouldn't it just be:

```suggestion
    committee = get_light_client_committee(state, period_start)
```

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+    signer_keys = []
```
djrtwo: Should add an `assert` that the `light_client_signature_bitfield` is of the expected length-- `len(committee)`

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+    for i, bit in enumerate(block.light_client_signature_bitfield):
+        if bit:
+            signer_keys.append(state.validators[committee[i]].pubkey)
+            increase_balance(state, committee[i], get_base_reward(state, committee[i]))
```
djrtwo: Seems like we need a proposer reward here

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+
+* `custody_bits[i][j] != generate_custody_bit(subkey, block_contents)` for any `j`
+* `execute_state_transition(shard, slot, attestation.shard_state_roots[i-1], hash_tree_root(parent), get_shard_proposer(state, shard, slot), block_contents) != shard_state_roots[i]` (if `i=0` then instead use `parent.shard_state_roots[s][-1]`)
+
+## Shard state transition function
+
+```python
+def shard_state_transition(shard: Shard, slot: Slot, pre_state: Hash, previous_beacon_root: Hash, proposer_pubkey: BLSPubkey, block_data: BytesN[MAX_SHARD_BLOCK_CHUNKS * SHARD_BLOCK_CHUNK_SIZE]) -> Hash:
+    # Beginning of block data is the previous block hash
+    assert block_data[:32] == pre_state.latest_block_hash
```
djrtwo: Why not a more structured shard block with header fields and a data field?

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+    assert block_data[32:64] == int_to_bytes8(slot) + b'\x00' * 24
+    # Signature check
+    assert len(block_data) >= 160
+    assert bls_verify(
+        pubkey=proposer_pubkey,
+        message_hash=hash_tree_root(block_data[:-96]),
+        signature=block_data[-96:],
```
djrtwo: again, why the unstructured-ness?

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+        domain=DOMAIN_SHARD_PROPOSER
+    )
+    # We will add something more substantive in phase 2
+    length = len(block.data.rstrip(b'\x00'))
```
djrtwo: `block` is not defined

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+    return ShardState(slot=slot, root=hash(pre_state + hash_tree_root(block_data)), gasprice=update_gasprice(pre_state, length), latest_block_hash = hash(block_data))
+```
+
+## Honest committee member behavior
+
+Suppose you are a committee member on shard `shard` at slot `current_slot`. Let `state` be the head beacon state you are building on. Three seconds into slot `slot`, run the following procedure:
```
djrtwo: Dividing the slot into 4

- beacon block proposal
- shard block proposal
- attestation broadcast
- aggregate broadcast

```suggestion
Suppose you are a committee member on shard `shard` at slot `current_slot`. Let `state` be the head beacon state you are building on. `2 * SECONDS_PER_SLOT // 4` seconds into slot `slot`, run the following procedure:
```

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+    return CompactCommittee(pubkeys=pubkeys, compact_validators=compact_validators)
+```
+
+### `get_light_client_committee`
+
+```python
+def get_light_client_committee(beacon_state: BeaconState, epoch: Epoch) -> Sequence[ValidatorIndex]:
+    assert epoch % LIGHT_CLIENT_COMMITTEE_PERIOD == 0
+    active_validator_indices = get_active_validator_indices(beacon_state, epoch)
+    seed = get_seed(beacon_state, epoch, DOMAIN_SHARD_LIGHT_CLIENT)
```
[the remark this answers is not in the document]
vbuterin: That's true though we have expressed desire for more domain separation before. I'd be ok either way.

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+```
+
+### `is_valid_indexed_attestation`
+
+```python
+def is_valid_indexed_attestation(state: BeaconState, indexed_attestation: IndexedAttestation) -> bool:
+    """
+    Check if ``indexed_attestation`` has valid indices and signature.
+    """
+
```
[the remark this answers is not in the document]
vbuterin: The length verification is already done by the SSZ datatype.

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+    # Verify indices are sorted
+    if indexed_attestation.participants != sorted(indexed_attestation.participants):
+        return False
+    
+    # Verify aggregate signature
+    all_pubkeys = []
+    all_message_hashes = []
+    for participant, custody_bits in zip(participants, indexed_attestation.custody_bits):
```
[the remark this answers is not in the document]
vbuterin: The problem is that in the normal case, the number of blocks in the attestation is ~1-4, so making a list of length 1-4 bitlists would be a huge amount of hashing overhead. But I fixed the loops to do things in the right order, good catch.

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,440 @@
...
+    )
+```
+
+## Beacon Chain Changes
+
+### New state variables
+
+```python
+    shard_states: Vector[ShardState, MAX_SHARDS]
+    online_countdown: Bytes[VALIDATOR_REGISTRY_LIMIT]
```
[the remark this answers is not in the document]
vbuterin: I was keeping it as a list of bytes because that way the hashing is lower. Hashing over a uint64 for every validator is currently only done in the validator balances list and is a primary component of epoch transition overhead; adding yet more hashing of this type would significantly add overhead.

If we think using bytes/uint8 is too impure, I have another suggestion: have a bitlist "online during previous 4-epoch period" and "online during current 4-epoch period". Validators would be considered online if their bit if flipped on for either one. This would have even less hashing overhead and would use existing types.

specs/core/1_new_shards.md
vbuterin: Interesting! We should discuss this; definitely open to considering going back to the "loop through attestations first, then finish in a second round" option.

specs/core/1_new_shards.md
vbuterin: I suppose we could, though if we do that we would also need to thread through the per-byte block fee.

specs/core/1_new_shards.md
vbuterin: We need to have logic somewhere to say "the light client committee used at slot N takes data from slot N - N % PERIOD - PERIOD (with one special case where N < PERIOD to avoid underflow). I just fixed; moved all that logic into `get_light_client_committee` itself.

specs/core/1_new_shards.md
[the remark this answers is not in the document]
djrtwo: Separate does feel more in line with past decisions. Keep as is

specs/core/1_new_shards.md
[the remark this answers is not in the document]
djrtwo: right

specs/core/1_new_shards.md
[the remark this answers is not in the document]
djrtwo: gotcha. looks good

specs/core/1_new_shards.md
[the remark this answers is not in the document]
djrtwo: Separate does feel more in line with past decisions. Keep as is

specs/core/1_new_shards.md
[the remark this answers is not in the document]
dankrad: I think List[uint8, ...] is clearer for this purpose than Bytes[...]

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,546 @@
...
+                participating_attestations.append(attestation)
+                if attestation.data.shard_transition_root == shard_transition_root:
+                    all_participants = all_participants.union(get_attesting_indices(state, attestation.data, attestation.aggregation_bits))
+            if (
+                get_total_balance(state, online_indices.intersection(all_participants)) * 3 >=
+                get_total_balance(state, online_indices.intersection(this_shard_committee)) * 2
+                and success is False
+            ):
+                assert shard_transition_root == hash_tree_root(block.shard_transition)
+                process_crosslink(state, shard, block.shard_transition)
```
vbuterin:
```suggestion
                apply_shard_transition(state, shard, block.shard_transition)
```

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,546 @@
...
+```python
+class ShardTransition(Container):
+    # Starting from slot
+    start_slot: Slot
+    # Shard block lengths
+    shard_block_lengths: List[uint64, MAX_SHARD_BLOCKS_PER_ATTESTATION]
+    # Shard data roots
+    shard_data_roots: List[List[Hash, MAX_SHARD_BLOCK_CHUNKS], MAX_SHARD_BLOCKS_PER_ATTESTATION]
+    # Intermediate state roots
+    shard_state_roots: List[ShardState, MAX_SHARD_BLOCKS_PER_ATTESTATION]
```
hwwhww: Should it be renamed to `shard_states`?
```suggestion
    shard_states: List[ShardState, MAX_SHARD_BLOCKS_PER_ATTESTATION]
```

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,546 @@
...
+    slot: Slot
+    body_root: Hash
+```
+
+### `ShardState`
+
+```python
+class ShardState(Container):
+    slot: Slot
+    gasprice: Gwei
```
hwwhww: Need to set initial `gasprice` value.

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,546 @@
...
+
+```python
+class ShardTransition(Container):
+    # Starting from slot
+    start_slot: Slot
+    # Shard block lengths
+    shard_block_lengths: List[uint64, MAX_SHARD_BLOCKS_PER_ATTESTATION]
+    # Shard data roots
+    shard_data_roots: List[List[Hash, MAX_SHARD_BLOCK_CHUNKS], MAX_SHARD_BLOCKS_PER_ATTESTATION]
+    # Intermediate state roots
```
hwwhww:
```suggestion
    # Intermediate shard states
```

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,546 @@
...
+    beacon_parent_root: Hash
+    slot: Slot
+    body: BytesN[SHARD_BLOCK_CHUNK_SIZE]
+    signature: BLSSignature
+```
+
+### `ShardSignedHeader`
+
+```python
+class ShardSignedHeader(Container):
```
hwwhww: Could you elaborate on what does "signed" mean here?

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,546 @@
...
+| `GASPRICE_ADJUSTMENT_COEFFICIENT` | `2**3` (= 8) | |
+| `DOMAIN_SHARD_LIGHT_CLIENT` | `192` | |
+| `DOMAIN_SHARD_PROPOSAL` | `193` | |
+
+## Containers
+
+### `ShardBlockWrapper`
+
+```python
+class ShardBlockWrapper(Container):
```
hwwhww: I suppose there will also have `ShardBlock`?

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,546 @@
...
+    custody_bits: List[Bitlist[MAX_VALIDATORS_PER_COMMITTEE], MAX_SHARD_BLOCKS_PER_ATTESTATION]
+    signature: BLSSignature
+```
+
+### `IndexedAttestation`
+
+```python
+class IndexedAttestation(Container):
+    committee: List[ValidatorIndex, MAX_VALIDATORS_PER_COMMITTEE]
+    attestation: Attestation
```
hwwhww: It seems very different from [v09x `IndexedAttestation`](https://github.com/ethereum/eth2.0-specs/blob/v09x/specs/core/0_beacon-chain.md#indexedattestation), maybe give it a new name?

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,546 @@
...
+        # Correct data root count
+        start_slot = state.shard_next_slots[shard]
+        offset_slots = [start_slot + x for x in SHARD_BLOCK_OFFSETS if start_slot + x < state.slot]
+        assert len(attestation.custody_bits) == len(offset_slots)
+        # Correct parent block root
+        assert data.beacon_block_root == get_block_root_at_slot(state, state.slot - 1)
+    # Type 2: delayed attestations
+    else:
+        assert state.slot - slot_to_epoch(data.slot) < EPOCH_LENGTH
+        assert data.shard_transition_root == Hash()
```
hwwhww: - Is it a kind of reduced form of attestation data?
- If so, does that mean the late attesters have to rebuild `AttestationData` of the old slot with empty `custody_bits` and `shard_transition_root`, and broadcast it?
- It seems the proposer can help transform ReducedAttestation in the [previous version](https://github.com/ethereum/eth2.0-specs/commit/0469d57ea26c848ffc2864741cc12d7dba2986c4)?

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,546 @@
...
+    pubkeys: List[BLSPubkey, MAX_VALIDATORS_PER_COMMITTEE]
+    compact_validators: List[uint64, MAX_VALIDATORS_PER_COMMITTEE]
+```
+
+### `AttestationCustodyBitWrapper`
+
+```python
+class AttestationCustodyBitWrapper(Container):
+    attestation_root: Hash
+    block_index: uint64
```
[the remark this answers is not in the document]
vbuterin: It's needed to distinguish the 0/1 values for each position, otherwise you'd just be adding up 1 bits without regard for what position they come in.

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,551 @@
...
+```
+
+### `ShardState`
+
+```python
+class ShardState(Container):
+    slot: Slot
+    gasprice: Gwei
+    data: Hash
+    latest_block_hash: Hash
```
terencechain: `block_root` instead of `block_hash`?

specs/core/1_new_shards.md
vbuterin: Yes, it's a reduced form, and late attesters do have to rebroadcast.

specs/core/1_new_shards.md
vbuterin: I think no need. It's zero by default, but then after one slot `get_updated_gasprice` pushes it up to `MIN_GASPRICE`.

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,583 @@
...
+```
+
+## Helpers
+
+### `get_online_validators`
+
+```python
+def get_online_indices(state: BeaconState) -> Set[ValidatorIndex]:
+    active_validators = get_active_validator_indices(state, get_current_epoch(state))
+    return set([i for i in active_validators if state.online_countdown[i] != 0])
```
terencechain: Is `set` really necessary?  Trying to optimize from client implementation's perspective

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,583 @@
...
+| `LIGHT_CLIENT_COMMITTEE_SIZE` | `2**7` (= 128) |
+| `LIGHT_CLIENT_COMMITTEE_PERIOD` | `2**8` (= 256) | epochs | ~27 hours |
+| `SHARD_COMMITTEE_PERIOD` | `2**8` (= 256) | epochs | ~27 hours |
+| `SHARD_BLOCK_CHUNK_SIZE` | `2**18` (= 262,144) | |
+| `MAX_SHARD_BLOCK_CHUNKS` | `2**2` (= 4) | |
+| `BLOCK_SIZE_TARGET` | `3 * 2**16` (= 196,608) | |
+| `SHARD_BLOCK_OFFSETS` | `[1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233]` | |
+| `MAX_SHARD_BLOCKS_PER_ATTESTATION` | `len(SHARD_BLOCK_OFFSETS)` | |
+| `EMPTY_CHUNK_ROOT` | `hash_tree_root(BytesN[SHARD_BLOCK_CHUNK_SIZE]())` | |
+| `MAX_GASPRICE` | `2**14` (= 16,384) | Gwei | |
```
terencechain: `gasprice` or `gas_price`? not sure which is more semantic 🤔

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,583 @@
...
+```
+
+### `committee_to_compact_committee`
+
+```python
+def committee_to_compact_committee(state: BeaconState, committee: Sequence[ValidatorIndex]) -> CompactCommittee:
+    """
+    Given a state and a list of validator indices, outputs the CompactCommittee representing them.
+    """
+    validators = [state.validators[i] for i in committee]
```
terencechain: Can simplify to one loop without sacrificing much readability. Example:
```go
	compactValidators := make([]uint64, len(committee))
	pubKeys := make([][]byte, len(committee))

	for i := 0; i < len(committee); i++ {
		v := state.Validators[committee[i]]
		compactValidators[i] = PackCompactValidator(committee[i], v.Slashed, v.EffectiveBalance/params.BeaconConfig().EffectiveBalanceIncrement)
		pubKeys[i] = v.PublicKey
	}
```

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,583 @@
...
+    pubkeys = [v.pubkey for v in validators]
+    return CompactCommittee(pubkeys=pubkeys, compact_validators=compact_validators)
+```
+
+### `get_shard_committee`
+
+```python
+def get_shard_committee(beacon_state: BeaconState, epoch: Epoch, shard: Shard) -> Sequence[ValidatorIndex]:
+    source_epoch = epoch - epoch % SHARD_COMMITTEE_PERIOD 
+    if source_epoch > 0:
```
terencechain: This feels safer 

```suggestion
    if source_epoch >= SHARD_COMMITTEE_PERIOD:
```

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,583 @@
...
+        for i, v in zip(committee, validators)
+    ]
+    pubkeys = [v.pubkey for v in validators]
+    return CompactCommittee(pubkeys=pubkeys, compact_validators=compact_validators)
+```
+
+### `get_shard_committee`
+
+```python
+def get_shard_committee(beacon_state: BeaconState, epoch: Epoch, shard: Shard) -> Sequence[ValidatorIndex]:
```
terencechain: `shard: Shard` is not used

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,583 @@
...
+    active_validator_indices = get_active_validator_indices(beacon_state, source_epoch)
+    seed = get_seed(beacon_state, source_epoch, DOMAIN_SHARD_COMMITTEE)
+    return compute_committee(active_validator_indices, seed, 0, ACTIVE_SHARDS)
+```
+
+### `get_shard_proposer_index`
+
+```python
+def get_shard_proposer_index(beacon_state: BeaconState, slot: Slot, shard: Shard) -> ValidatorIndex:
+    committee = get_shard_committee(beacon_state, slot_to_epoch(slot), shard)
```
terencechain:
```suggestion
    committee = get_shard_committee(beacon_state, compute_epoch_at_slot(slot), shard)
```

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,583 @@
...
+        signature=indexed_attestation.signature,
+        domain=get_domain(state, DOMAIN_BEACON_ATTESTER, indexed_attestation.data.target.epoch),
+    )
+```
+
+### `get_shard`
+
+```python
+def get_shard(state: BeaconState, attestation: Attestation) -> Shard:
+    return Shard((attestation.data.index + get_start_shard(state, data.slot)) % ACTIVE_SHARDS)
```
terencechain: `get_start_shard` is no longer defined in beacon chain spec, I think we have to reintroduce it here

terencechain:
```suggestion
    return Shard((attestation.data.index + get_start_shard(state, attestation.data.slot)) % ACTIVE_SHARDS)
```

specs/core/1_new_shards.md
```diff
@@ -0,0 +1,583 @@
...
+    slot: Slot
+    gasprice: Gwei
+    data: Hash
+    latest_block_root: Hash
+```
+
+### `AttestationData`
+
+```python
+class AttestationData(Container):
```
terencechain: Would it make sense to add `shard` to the field?  can avoid `get_shard`

protolambda:
Took the latest work and started on the necessary changes to make it run as part of the pyspec, and just organize things (start with custody spec cleanup, integrate with the rest of the spec).
Also rebased on the latest dev, with v0.9.1 changes too. And then some minor misc fixes in this code (blockbody vs block, loop indentation, etc.).

Opened a PR here: https://github.com/ethereum/eth2.0-specs/pull/1483
I'll try to merge any continued commits here, but it would be great if we can coordinate to continue phase 1 changes on a fork of this new PR.

specs/core/1_new_shards.md
protolambda: It's not, but when things are unordered and unique we declare it as `Set` in the spec. Here it would be a free conversion for a client to a set object anyway (or any validator indices container).

djrtwo:
Closing in favor of #1483

specs/core/1_new_shards.md
djrtwo: Would still need to validate the shard against the committee index for the slot, but I see value in including it here explicitly (similar to including `slot` even though it used to be able to be calculated from the constituent parts)

PR #1428, opened 2019-10-12
Remove Shards/Crosslinks from Phase 0

djrtwo:
In light of discussions around the alternate phase 1 proposal, this PR removes the notion of Shard and Crosslink entirely from Phase 0 to allow for (1) a clean slate to design phase 1 on top of and (2) to enable client implementers to keep moving forward on phase 0 testnets.

* Removed all Shard and Crosslink mechanics (`Crosslink`, `AttestationData.crosslink`, `BeaconState.start_shard`, `process_crosslinks`, `get_crosslink_deltas`, etc)
* Added notion of `CrosslinkIndex`, `AttestationData.index`, and `MAX_COMMITTEES_PER_SLOT`. This indexes committees on a per-slot basis. The notion of Shard will map on top of this in phase 1.
* Add `slot` to `AttestationData`. Required in this new context because with the removal `shard`, we can't reverse engineer the slot. Also we've been planning on doing this anyway.
* Renamed "crosslink committee" to "beacon committee"
* Temporarily disabled phase 1 tests from CI to isolate phase 0 for discussion before fixing/removing things from phase 1
* constants changes
    * `SECONDS_PER_SLOT` from `6` to `12` -- more events will occur per slot in these modified proposals so 12 to be conservation
    * `SLOTS_PER_EPOCH` from `64` to `32` -- reduction of 64 to 32 here keeps epochs at 6 minutes and also help with crosslink committee sizes (wrt `TARGET_COMMITTEE_SIZE`)
    * the combination of the prior two changes keep the same number of epochs per year so the same reward rate
    * `BASE_REWARDS_PER_EPOCH` from `5` to `4` -- adjustment due to the removal of crosslinks. keeps rewards constant
    * `MAX_VALIDATORS_PER_COMMITTEE` to `2048` -- `TOTAL_ETH_SUPPLY // ETH_PER_VALIDATOR // SLOTS_PER_EPOCH // MAX_COMMITTEES_PER_SLOT` -- `2**27 // 32 // 32 // 64`


specs/core/0_beacon-chain.md
```diff
@@ -174,6 +168,7 @@ The following values are (non-configurable) constants used throughout the specif
 
 | Name | Value |
 | - | - |
+| `COMMITTEES_PER_SLOT` | `2**5` (= 32) |
 | `SHARD_COUNT` | `2**10` (= 1,024) |
```
terencechain: Remove?

terencechain:
what's the timeline for this? do we plan to roll this out for the next release? : )

specs/core/0_beacon-chain.md
djrtwo: Will do. There are a few things still in there just beacuse phase 1 uses them and the build fails if I take them out...

djrtwo:
@terencechain, Made a few minor changes today. Still check out on your end?

> what's the timeline for this? do we plan to roll this out for the next release? : )

Discussing this now. Want to get this out asap, but also want to keep things moving forward smoothly on the client/testnet side of things

JustinDrake:
I'm loving the cleaner separation between phase 0 and phase 1 👍 

Minor comments:

1) The `Shard` custom type can go
2) Transfers can go
3) Consider renaming `get_committees_per_slot` to `get_committee_count_at_slot` ("at slot" is consistent with `get_block_root_at_slot`; "committee_count" is more correct than "committees")
4) `index` should go immediately after `slot` in `AttestationData` (remove "Committee Index" comment (with inconsistent capitalisation 😂))
5) Consider the shorter "a committee index at a slot" over "an index for a committee within a slot"
6) `MAX_COMMITTEES_PER_SLOT` => `MAX_COMMITTEE_COUNT_PER_SLOT`
7) Should `MAX_VALIDATORS_PER_COMMITTEE` now be `2**11`?
8) The "Shuffling" section comment in `BeaconState` is now more accurately "Randomness"
9) Rename `compute_epoch_of_slot` to `compute_epoch_at_slot` (consistent with `get_block_root_at_slot` and `get_committee_count_at_slot`)
10) If we remove the notion of persistent committee, do we need `compute_committee`?

specs/validator/0_beacon-chain-validator.md
```diff
@@ -278,7 +274,7 @@ Up to `MAX_VOLUNTARY_EXITS`, [`VoluntaryExit`](../core/0_beacon-chain.md#volunta
 
 ### Attestations
 
-A validator is expected to create, sign, and broadcast an attestation during each epoch. The `committee`, assigned `shard`, and assigned `slot` for which the validator performs this role during an epoch are defined by `get_committee_assignment(state, epoch, validator_index)`.
+A validator is expected to create, sign, and broadcast an attestation during each epoch. The `committee`, assigned `index`, and assigned `slot` for which the validator performs this role during an epoch are defined by `get_committee_assignment(state, epoch, validator_index)`.
 
 A validator should create and broadcast the attestation halfway through the `slot` during which the validator is assigned―that is, `SECONDS_PER_SLOT * 0.5` seconds after the start of `slot`.
```
protolambda: Maybe update "Attestation data" below to explicitly set the `slot`?

specs/core/0_beacon-chain.md
```diff
@@ -507,16 +490,12 @@ class BeaconState(Container):
     validators: List[Validator, VALIDATOR_REGISTRY_LIMIT]
     balances: List[Gwei, VALIDATOR_REGISTRY_LIMIT]
     # Shuffling
```
```suggestion
    # Randomness
```
Agree with Justin here. I tried keeping randomness separate from shuffling in ZRNT, and it was not pretty. This is great :)

specs/core/0_beacon-chain.md
```diff
@@ -143,6 +137,7 @@ We define the following Python custom types for type hinting and readability:
 | - | - | - |
 | `Slot` | `uint64` | a slot number |
 | `Epoch` | `uint64` | an epoch number |
+| `CommitteeIndex` | `uint64` | an index for a committee within a slot |
 | `Shard` | `uint64` | a shard number |
```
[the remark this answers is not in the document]
protolambda: It's used in `AttestationData`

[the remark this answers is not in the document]
hwwhww: But I meant `Shard`, not `CommitteeIndex`. 😄
No more `Crosslink` in `AttestationData`:
https://github.com/ethereum/eth2.0-specs/blob/283a8cbf0d67e00950d2a1800c9e8803b447643b/specs/core/0_beacon-chain.md#attestationdata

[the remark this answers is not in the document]
protolambda: Aha, sorry, was reading the highlighted line. I do think we should keep `Shard` though, but moved to the phase 1 definitions.

djrtwo:
@JustinDrake 

1) `Shard type` -- Will move to phase 1 (to be majorly reworked but the type being there keeps build fine for now
2) Transfers -- will open a separate PR to make this more digestible for implementers
3)  `get_committees_per_slot` -> `get_committee_count_at_slot` -- will do
4) `index` should go immediately after `slot` -- Was planning on it, but forgot. thanks!
5) Consider the shorter "a committee index at a slot" over "an index for a committee within a slot"
6)  `MAX_COMMITTEE_COUNT_PER_SLOT` -- a bit long and might push a bunch of lines over the 120 char limit. Will test it out
7)  `MAX_VALIDATORS_PER_COMMITTEE` now be `2**11` -- will be different for sure. will be defined in relation to the number of committees per slot but also the number of slots per epoch which is up for debate. I'll make the number work in relation to the current constants
8) "Randomness" -- willl do
9) `compute_epoch_of_slot` to `compute_epoch_at_slot` -- would rather reduce random cleanup changes, but.... okay
10) Think we should keep `compute_committee`. We are likely going to have a persistent committee upon which shard block proposers are selected and which serve as the p2p backbone for the shard

specs/core/0_beacon-chain.md
```diff
@@ -878,61 +855,39 @@ def get_seed(state: BeaconState, epoch: Epoch, domain_type: DomainType) -> Hash:
...
-def get_committee_count(state: BeaconState, epoch: Epoch) -> uint64:
+def get_committee_count_at_slot(state: BeaconState, slot: Slot) -> uint64:
     """
-    Return the number of committees at ``epoch``.
+    Return the number of committees at ``slot``.
     """
-    committees_per_slot = max(1, min(
-        SHARD_COUNT // SLOTS_PER_EPOCH,
+    epoch = compute_epoch_at_slot(slot)
+    return max(1, min(
```
terencechain: don't need `max(1,` since `MAX_COMMITTEES_PER_SLOT` will always be >1

djrtwo: but the min operation could result in `0`

Say active validator indices is 10, then it looks like 

```
max(1, min(64, 10 // 64 // 128))
```

`10 // 64 // 128 == 0` so

```
max(1, min(64, 0)) == max(1, 0)
```

terencechain: Ahh I see. That's right

specs/core/0_beacon-chain.md
```diff
@@ -878,61 +857,39 @@ def get_seed(state: BeaconState, epoch: Epoch, domain_type: DomainType) -> Hash:
...
 ```python
-def get_crosslink_committee(state: BeaconState, epoch: Epoch, shard: Shard) -> Sequence[ValidatorIndex]:
+def get_beacon_committee(state: BeaconState, slot: Slot, index: CommitteeIndex) -> Sequence[ValidatorIndex]:
     """
-    Return the crosslink committee at ``epoch`` for ``shard``.
+    Return the beacon committee at ``slot`` for ``index``.
     """
+    epoch = compute_epoch_of_slot(slot)
+    committees_per_slot = get_committees_per_slot(state, slot)
+    epoch_offset = index + (slot % SLOTS_PER_EPOCH) * committees_per_slot
```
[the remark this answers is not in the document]
djrtwo: It is the number of committee in the epoch out of the total committees.

Say there are 4 slots in the epoch and 10 committees per slot. There are a total of 40 committees from the shuffling and we want to figure out which committee out of the 40 it is to get the correct shuffling.  So if "committee index" of 5 in the last slot in the epoch, it would be the `(10*3 + 5)` (35) of the 40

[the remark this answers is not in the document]
djrtwo: Will work on the name or add a comment

configs/minimal.yaml
```diff
@@ -4,8 +4,8 @@
 # Misc
 # ---------------------------------------------------------------
 
-# [customized] Just 8 shards for testing purposes
-SHARD_COUNT: 8
+# [customized] Just 2 committees for slot for testing purposes
+MAX_COMMITTEES_PER_SLOT: 4
```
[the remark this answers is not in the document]
djrtwo: Changed to 4 to induce fewer than the max committees per slot in the minimal test config (along with the normal number of validators assigned in those tests)

specs/core/0_beacon-chain.md
```diff
@@ -1609,37 +1499,27 @@ def process_attester_slashing(state: BeaconState, attester_slashing: AttesterSla
 ```python
 def process_attestation(state: BeaconState, attestation: Attestation) -> None:
     data = attestation.data
-    assert data.crosslink.shard < SHARD_COUNT
+    assert data.index < get_committees_per_slot(state, data.slot)
     assert data.target.epoch in (get_previous_epoch(state), get_current_epoch(state))
+    assert data.slot + MIN_ATTESTATION_INCLUSION_DELAY <= state.slot <= data.slot + SLOTS_PER_EPOCH
```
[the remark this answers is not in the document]
djrtwo: probably.. but going to avoid making a new var here for now

djrtwo:
@JustinDrake @protolambda @hwwhww @terencechain 
Thank you for the feedback! Comments addressed

configs/mainnet.yaml
```diff
@@ -51,12 +51,12 @@ BLS_WITHDRAWAL_PREFIX: 0x00
...
 # ---------------------------------------------------------------
-# 6 seconds 6 seconds
-SECONDS_PER_SLOT: 6
+# 12 seconds
+SECONDS_PER_SLOT: 12
 # 2**0 (= 1) slots 6 seconds
 MIN_ATTESTATION_INCLUSION_DELAY: 1
-# 2**6 (= 64) slots 6.4 minutes
-SLOTS_PER_EPOCH: 64
+# 2**6 (= 32) slots 6.4 minutes
```
wemeetagain: minor comment change
```suggestion
# 2**5 (= 32) slots 6.4 minutes
```

specs/core/0_beacon-chain.md
```diff
@@ -878,61 +855,37 @@ def get_seed(state: BeaconState, epoch: Epoch, domain_type: DomainType) -> Hash:
     return hash(domain_type + int_to_bytes(epoch, length=8) + mix)
 ```
 
-#### `get_committee_count`
+#### `get_committee_count_at_slot`
```
JustinDrake: This function is not in the ToC

specs/core/0_beacon-chain.md
```diff
@@ -69,7 +68,7 @@
             - [`compute_shuffled_index`](#compute_shuffled_index)
             - [`compute_proposer_index`](#compute_proposer_index)
             - [`compute_committee`](#compute_committee)
-            - [`compute_epoch_of_slot`](#compute_epoch_of_slot)
+            - [`compute_epoch_at_slot`](#compute_epoch_at_slot)
             - [`compute_start_slot_of_epoch`](#compute_start_slot_of_epoch)
```
JustinDrake: should be `compute_start_slot_at_epoch` for consistency with `compute_epoch_at_slot`, `get_block_root_at_slot` and `get_committee_count_at_slot`.

specs/core/0_beacon-chain.md
djrtwo: ah, thanks. need to re-run ToC generator

terencechain:
`CustodyChunkChallenge` still uses crosslink from attestation. I'm guessing It'll be a separate effort to revamp the custody game spec?

djrtwo:
@terencechain, yes. I've currently disabled phase 1 CI as most of it needs to be revamped. Prioritizing getting the phase 0 things out and then we'll circle back to get the phase 1 spec in place

terencechain:
> @terencechain, yes. I've currently disabled phase 1 CI as most of it needs to be revamped. Prioritizing getting the phase 0 things out and then we'll circle back to get the phase 1 spec in place

Great. Thank you

PR #1432, opened 2019-10-16
[WIP] `start_index` (per slot)

djrtwo:


djrtwo:
got some feedback from @vbuterin. Removing the idea of rotating indices per slot. Keeping that for a phsae 1 concept.

Closing

PR #1437, opened 2019-10-18
[WIP] Remove custody bits

djrtwo:
Experimenting with removing custody bits from phase 0


djrtwo:
Not really pleased with this half removal of custody from phase 0. Closing and thinkig about it

PR #1440, opened 2019-10-23
Naive aggregation strategy

djrtwo:
Add the "naive" attestation aggregation strategy to phase 0.

* add mechanism for selecting which proposers are supposed to aggregate and broadcast
* add wrapper object `AggregateAndProof` for broadcasting aggregates that includes a "proof" that the aggregator was selected to aggregate
* add notes and validation conditions about aggregate and attestation subnet gossip channels
* rename some topics and vars to reflect new functionality
* add to and modify some FAQ


specs/networking/p2p-interface.md
```diff
@@ -147,10 +148,27 @@ Topics are plain UTF-8 strings and are encoded on the wire as determined by prot
...
+    - Clients MUST validate that the block being voted for (`aggregate_and_proof.aggregate.data.beacon_block_root`) passes validation.
+    - Clients MUST validate that `aggregate_and_proof.aggregate.data.slot` is
+      within the last `ATTESTATION_PROPAGATION_SLOT_RANGE` slots.
+    - Clients MUST validate that the validator index is within the aggregate's
+      committee -- i.e. `aggregate_and_proof.index in get_attesting_indices(state, aggregate_and_proof.aggregate.data, aggregate_and_proof.aggregate.aggregation_bits)`.
+    - Clients MUST validate that `aggregate_and_proof.selection_proof` selects
+      the validator as an aggregator for the slot -- i.e. `is_aggregator(state, aggregate_and_proof.aggregate.data.index, aggregate_and_proof.selection_proof)` returns `True`.
+    - Clients MUST validate that the `aggregate_and_proof.selection_proof` is a
+      valid signature of the `aggregate_and_proof.aggregate.data.slot` by the validator with index `aggregate_and_proof.index`.
+    - Clients MUST validate that the signature of `aggregate_and_proof.aggregate`.
```
protolambda: Note that this ensures bitlist correctness implicitly. And above checks can error unexpectedly because of attestation bitlist length issues. Maybe add a note that the attestation contents must be validated with the signature first?

specs/networking/p2p-interface.md
```diff
@@ -160,13 +178,15 @@ Additional topics are used to propagate lower frequency validator messages. Thei
 
 #### Interop
 
-Unaggregated and aggregated attestations from all shards are sent to the `beacon_attestation` topic. Clients are not required to publish aggregate attestations but must be able to process them. All validating clients SHOULD try to perform local attestation aggregation to prepare for block proposing.
+Unaggregated and aggregated attestations from all shards are sent as `Attestation`s to the `beacon_aggregate_and_proof` topic. Clients are not required to publish aggregate attestations but must be able to process them. All validating clients SHOULD try to perform local attestation aggregation to prepare for block proposing.
```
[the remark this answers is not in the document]
djrtwo: I'm a bit torn here. Think we should leave the interop spec as it was. I can add back in the `beacon_attestation` channel for global propagation of individual attestations and we can just deprecate it for mainnet configurations

specs/networking/p2p-interface.md
djrtwo: hmmm, but I generally want to do the cheap checks first before doing the expensive signature check.

Even if an attacker crafts an attestation with bitlength issues that show up prior to signature validation, what can they actually achieve?

protolambda: Nothing much, trying to find ways to bypass/mess with these conditions. And checking properties based on assumptions/side-effects of other checks is part of that.
But worst-case here would be an unexpected panic because of some indexing error during one of the above checks.

specs/networking/p2p-interface.md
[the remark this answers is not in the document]
protolambda: I think I like the `ATTESTATION_SUBNET_COUNT == 1` since it is close to mainnet, but super simple as there is only one subnet (so not really a subnet, can be implemented easily by everone). Let's ask on the call maybe?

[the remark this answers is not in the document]
djrtwo: thoughts on that?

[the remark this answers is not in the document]
arnetheduck: I suspect we should do an `interopv2` which is basically mainnet-parameters but testnet - this way we can leave interop as is for now and let people get the basics right  (we need this for noise and the other mainnet changes anyway)

it looks like it should possible that clients that support aggregation can start publishing to the aggregation topics for testing as soon as they're ready.

specs/validator/0_beacon-chain-validator.md
```diff
@@ -339,6 +350,81 @@ def get_signed_attestation_data(state: BeaconState, attestation: IndexedAttestat
...
+If the validator is selected to aggregate (`is_aggregator`), then they broadcast their best aggregate to the global aggregate channel (`beacon_aggregate_and_proof`) two-thirds of the way through the `slot`-that is, `SECONDS_PER_SLOT * 2 / 3` seconds after the start of `slot`.
+
+Aggregate attestations are broadcast as `AggregateAndProof` objects to prove to the gossip channel that the validator has been selected as an aggregator.
+
+##### `AggregateAndProof`
+
+```python
+class AggregateAndProof(Container):
+    index: ValidatorIndex
+    selection_proof: BLSSignature
```
arnetheduck: one downside of this scheme is that attestation aggregates are no longer idempotent meaning that one always has to pay for the aggregated attestations even if two aggregators come up with the same aggregation (which otherwise could simply be ignored / swallowed by gossip dissemination)

specs/validator/0_beacon-chain-validator.md
```diff
@@ -339,6 +350,81 @@ def get_signed_attestation_data(state: BeaconState, attestation: IndexedAttestat
...
+def slot_signature(state: BeaconState, slot: Slot, privkey: int) -> BLSSignature:
+    domain = get_domain(state, DOMAIN_BEACON_ATTESTER, compute_epoch_at_slot(slot))
+    return bls_sign(privkey, hash_tree_root(slot), domain)
+```
+
+```python
+def is_aggregator(state: BeaconState, slot: Slot, index: CommitteeIndex, slot_signature: BLSSignature) -> bool:
+    committee = get_beacon_committee(state, slot, index)
+    modulo = max(1, len(committee) // TARGET_AGGREGATORS_PER_COMMITTEE)
+    return bytes_to_int(hash(slot_signature)[0:8]) % modulo == 0
```
hwwhww: Is it possible that there's *no any aggregator* gets selected from the committee even the committee is not empty?
How about just naively using the "index of validator in committee" to select the aggregators? 

```python
def is_aggregator(state: BeaconState, slot: Slot, index: CommitteeIndex, slot_signature: BLSSignature, validator_index: ValidatorIndex) -> bool:
    committee = get_beacon_committee(state, slot, index)
    modulo = max(1, len(committee) // TARGET_AGGREGATORS_PER_COMMITTEE)
    validator_index_in_committee = committee.index(validator_index)
    return validator_index_in_committee % module == 0
```

specs/networking/p2p-interface.md
```diff
@@ -147,10 +148,27 @@ Topics are plain UTF-8 strings and are encoded on the wire as determined by prot
...
+- `beacon_aggregate_and_proof` - This topic is used to propagate aggregated attestations (as `AggregateAndProof`s) to subscribing nodes (typically validators) to be included in future blocks. The following validations MUST pass before forwarding the `aggregate_and_proof` on the network.
+    - Clients MUST validate that the block being voted for (`aggregate_and_proof.aggregate.data.beacon_block_root`) passes validation.
+    - Clients MUST validate that `aggregate_and_proof.aggregate.data.slot` is
+      within the last `ATTESTATION_PROPAGATION_SLOT_RANGE` slots.
+    - Clients MUST validate that the validator index is within the aggregate's
+      committee -- i.e. `aggregate_and_proof.index in get_attesting_indices(state, aggregate_and_proof.aggregate.data, aggregate_and_proof.aggregate.aggregation_bits)`.
```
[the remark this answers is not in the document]
djrtwo: This means that being selected as an aggregator only gives you the right to send your aggregate. If this weren't the case, then I could take your proof and send whatever aggregates I wanted.

That is, unless we added a third signature that signed over the aggregate+proof itself. Which is... overkill (?)

specs/networking/p2p-interface.md
```diff
@@ -147,10 +148,27 @@ Topics are plain UTF-8 strings and are encoded on the wire as determined by prot
...
+    - Clients MUST validate that `aggregate_and_proof.selection_proof` selects
+      the validator as an aggregator for the slot -- i.e. `is_aggregator(state, aggregate_and_proof.aggregate.data.index, aggregate_and_proof.selection_proof)` returns `True`.
+    - Clients MUST validate that the `aggregate_and_proof.selection_proof` is a
+      valid signature of the `aggregate_and_proof.aggregate.data.slot` by the validator with index `aggregate_and_proof.index`.
+    - Clients MUST validate that the signature of `aggregate_and_proof.aggregate`.
+
+Attestation subnets are used to propagate unaggregated attestations to subsections of the network. Their `TopicName`s are:
+
+- `index{index % ATTESTATION_SUBNET_COUNT}_beacon_attestation` - These topics are used to propagate unaggregated attestations to subsections of the network (typically beacon and persistent committees) to be aggregated before being gossiped to `beacon_aggregate_and_proof`. The following validations MUST pass before forwarding the `attestation` on the network.
```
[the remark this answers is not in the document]
djrtwo: it's supposed to read as something like `index10_beacon_attestation`. This will ultimately be replaced with shard specific attestation subnets, but the concept of shard doesn't exist in phase 0 and will be an overlay on top of `CommitteeIndex` which is defined in phase 0.

Maybe `committee_index{committee_index % ATTESTATION_SUBNET_COUNT}_beacon_attestation` would be clearer

specs/validator/0_beacon-chain-validator.md
djrtwo: >  pay for the aggregated attestations

Pay wrt network bandwidth? Yes, that is the case and a minor downside. At least at first glance if we go this route, it _doesn't_ seem worth imodifying gossipsub to be able to use a component of a message as the identified.

djrtwo: Actually, we can handle this by adding a validation condition, and probably should.

> validate that the Attestation defined by `hash_tree_root(aggregate)` has not already been seen (in pool or in a block locally).

specs/networking/p2p-interface.md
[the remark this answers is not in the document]
djrtwo: I added back in the `beacon_attestation` topic for interop with a note that it will be removed for mainnet. Agreed that this simple procedure should stick around a little longer

specs/networking/p2p-interface.md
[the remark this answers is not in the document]
djrtwo: renamed to `committee_indexN_beacon_attestation`

specs/validator/0_beacon-chain-validator.md
djrtwo: condition added

specs/networking/p2p-interface.md
```diff
@@ -147,10 +148,26 @@ Topics are plain UTF-8 strings and are encoded on the wire as determined by prot
 
 Topic strings have form: `/eth2/TopicName/TopicEncoding`. This defines both the type of data being sent on the topic and how the data field of the message is encoded. (Further details can be found in [Messages](#Messages)).
 
-There are two main topics used to propagate attestations and beacon blocks to all nodes on the network. Their `TopicName`s are:
+There are two main topics used to propagate aggregate attestations and beacon blocks to all nodes on the network. Their `TopicName`s are:
```
hwwhww: To clarify, this section is for global, right?
```suggestion
There are two main topics used to propagate aggregate attestations and beacon blocks to all nodes on the global network. Their `TopicName`s are:
```

specs/networking/p2p-interface.md
```diff
@@ -177,19 +196,22 @@ Clients MUST reject (fail validation) messages that are over this size limit. Li
...
-| attester_slashing            | AttesterSlashing  |
+| Topic                                  | Message Type      |
+|----------------------------------------|-------------------|
+| beacon_block                           | BeaconBlock       |
+| beacon_aggregate_and_proof             | AggregateAndProof |
+| beacon_attestation\*                   | Attestation       |
+| committee_index{N}\_beacon_attestation | Attestation       |
+| voluntary_exit                         | VoluntaryExit     |
+| proposer_slashing                      | ProposerSlashing  |
+| attester_slashing                      | AttesterSlashing  |
```
hwwhww: IMO the current "Topics" and "Messages" sections are difficult to look up.
It would be nice to (i) have a single table to explain the topics and (ii) move the `beacon_aggregate_and_proof` and `committee_index{N}\_beacon_attestation` validation logic explainer to its own section.

djrtwo: did a bit of a reorg. Think it is better now

specs/networking/p2p-interface.md
```diff
@@ -140,54 +141,72 @@ The following gossipsub [parameters](https://github.com/libp2p/specs/tree/master
...
+- `beacon_block` - This topic is used solely for propagating new beacon blocks to all nodes on the networks. Blocks are sent in their entirety. Clients MUST validate the block proposer signature before forwarding it across the network.
+- `beacon_aggregate_and_proof` - This topic is used to propagate aggregated attestations (as `AggregateAndProof`s) to subscribing nodes (typically validators) to be included in future blocks. The following validations MUST pass before forwarding the `aggregate_and_proof` on the network.
+    - The aggregate attestation defined by `hash_tree_root(aggregate_and_proof.aggregate)` has _not_ already been seen (via aggregate gossip, within a block, or through the creation of an equivalent aggregate locally).
+    - The block being voted for (`aggregate_and_proof.aggregate.data.beacon_block_root`) passes validation.
+    - `aggregate_and_proof.aggregate.data.slot` is within the last `ATTESTATION_PROPAGATION_SLOT_RANGE` slots.
```
hwwhww: nitpick: maybe define "within the last `ATTESTATION_PROPAGATION_SLOT_RANGE` slots" more precisely? Does that mean `aggregate_and_proof.aggregate.data.slot >= state.slot - ATTESTATION_PROPAGATION_SLOT_RANGE`?

specs/networking/p2p-interface.md
```diff
@@ -140,54 +141,72 @@ The following gossipsub [parameters](https://github.com/libp2p/specs/tree/master
...
+Attestation subnets are used to propagate unaggregated attestations to subsections of the network. Their `TopicName`s are:
 
-The payload is carried in the `data` field of a gossipsub message, and varies depending on the topic:
+- `committee_index{index % ATTESTATION_SUBNET_COUNT}_beacon_attestation` - These topics are used to propagate unaggregated attestations to subsections of the network (typically beacon and persistent committees) to be aggregated before being gossiped to `beacon_aggregate_and_proof`. The following validations MUST pass before forwarding the `attestation` on the network.
+    - The attestation's committee index (`attestation.data.index`) is for the correct subnet.
+    - The attestation is unaggregated -- that is, it has exactly one participating validator (`len([bit for bit in attestation.aggregation_bits if bit == 0b1]) == 1`).
+    - The block being voted for (`attestation.data.beacon_block_root`) passes validation.
+    - `attestation.data.slot` is within the last `ATTESTATION_PROPAGATION_SLOT_RANGE` slots.
```
hwwhww: Ditto. Does that mean `attestation.data.slot >= state.slot - ATTESTATION_PROPAGATION_SLOT_RANGE`?

specs/networking/p2p-interface.md
```diff
@@ -140,54 +141,72 @@ The following gossipsub [parameters](https://github.com/libp2p/specs/tree/master
...
 
-Each gossipsub [message](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/rpc.proto#L17-L24) has a maximum size of `GOSSIP_MAX_SIZE`.
+#### Attestation subnets
 
-Clients MUST reject (fail validation) messages that are over this size limit. Likewise, clients MUST NOT emit or propagate messages larger than this limit.
+Attestation subnets are used to propagate unaggregated attestations to subsections of the network. Their `TopicName`s are:
 
-The payload is carried in the `data` field of a gossipsub message, and varies depending on the topic:
+- `committee_index{index % ATTESTATION_SUBNET_COUNT}_beacon_attestation` - These topics are used to propagate unaggregated attestations to subsections of the network (typically beacon and persistent committees) to be aggregated before being gossiped to `beacon_aggregate_and_proof`. The following validations MUST pass before forwarding the `attestation` on the network.
```
hwwhww: The `index` is explained in the following "Mainnet" section. Maybe use `committee_index{subnet_id}_beacon_attestation` here?
```suggestion
- `committee_index{subnet_id}_beacon_attestation` - These topics are used to propagate unaggregated attestations to the subnet `subnet_id` (typically beacon and persistent committees) to be aggregated before being gossiped to `beacon_aggregate_and_proof`. The following validations MUST pass before forwarding the `attestation` on the subnet.
```

specs/networking/p2p-interface.md
```diff
@@ -140,54 +141,72 @@ The following gossipsub [parameters](https://github.com/libp2p/specs/tree/master
...
+The payload is carried in the `data` field of a gossipsub message, and varies depending on the topic:
 
-- `beacon_block` - This topic is used solely for propagating new beacon blocks to all nodes on the networks. Blocks are sent in their entirety. Clients MUST validate the block proposer signature before forwarding it across the network.
-- `beacon_attestation` - This topic is used to propagate aggregated attestations (in their entirety) to subscribing nodes (typically block proposers) to be included in future blocks. Clients MUST validate that the block being voted for passes validation before forwarding the attestation on the network (TODO: [additional validations](https://github.com/ethereum/eth2.0-specs/issues/1332)).
+| Topic                                  | Message Type      |
+|----------------------------------------|-------------------|
+| beacon_block                           | BeaconBlock       |
+| beacon_aggregate_and_proof             | AggregateAndProof |
+| beacon_attestation\*                   | Attestation       |
+| committee_index{N}\_beacon_attestation | Attestation       |
```
hwwhww:
```suggestion
| committee_index{subnet_id}\_beacon_attestation | Attestation       |
```

specs/networking/p2p-interface.md
```diff
@@ -140,54 +141,72 @@ The following gossipsub [parameters](https://github.com/libp2p/specs/tree/master
...
-| voluntary_exit               | VoluntaryExit     |
-| proposer_slashing            | ProposerSlashing  |
-| attester_slashing            | AttesterSlashing  |
+Unaggregated and aggregated attestations from all shards are sent as `Attestation`s to the `beacon_attestation` topic. Clients are not required to publish aggregate attestations but must be able to process them. All validating clients SHOULD try to perform local attestation aggregation to prepare for block proposing.
 
-Clients MUST reject (fail validation) messages containing an incorrect type, or invalid payload.
+#### Mainnet
 
-When processing incoming gossip, clients MAY descore or disconnect peers who fail to observe these constraints.
+Attestation broadcasting is grouped into subnets defined by a topic. The number of subnets is defined via `ATTESTATION_SUBNET_COUNT`. The `CommitteeIndex`, `index`, is assigned to the topic: `committee_index{index % ATTESTATION_SUBNET_COUNT}_beacon_attestation`.
```
hwwhww: If modifying to `committee_index{subnet_id}\_beacon_attestation` as above, suggesting:
```suggestion
Attestation broadcasting is grouped into subnets defined by a topic. The number of subnets is defined via `ATTESTATION_SUBNET_COUNT`. For the `committee_index{subnet_id}_beacon_attestation` topics, `subnet_id` is set to `index % ATTESTATION_SUBNET_COUNT`, where `index` is the `CommitteeIndex` of the given committee.
```

specs/validator/0_beacon-chain-validator.md
hwwhww: FWIW, I made a chart: https://docs.google.com/spreadsheets/d/1C7pBqEWJgzk3_jesLkqJoDTnjZOODnGTOJUrxUMdxMA/edit?usp=sharing

- If `TARGET_AGGREGATORS_PER_COMMITTEE = 16`, the probability is close to `1.06E-07` for the worst case.
- If `TARGET_AGGREGATORS_PER_COMMITTEE = 32`, the probability is close to `9.84E-15` for the worst case.

The probability looks fine, at least for now. We can adjust `TARGET_AGGREGATORS_PER_COMMITTEE` when committee size getting larger anyway.

djrtwo:
addressed your comments @hwwhww !

Issue #1446, opened 2019-10-24
Manipulating deposit contract to gain an early majority


paulhauner:
This issue describes a feasible way to gain a dishonest majority in the first two epochs.

### Assumptions
- The shuffling of the first two epochs is determined only by the number of active validators (`v`).
- `v` can only vary by at most `n`, where `n = eth1_gas_limit / gas_for_deposit` (I think this is ~100).

### Gaining an early dishonest majority

If an attacker can control their index in the deposit tree (this seems feasible), then it would take `n * (SLOTS_PER_EPOCH / 2)` validators to ensure they control 50% of the validators in either the first or second epochs.

#### Influencing future shuffling

An attacker with a block-producer majority gains the standard 1-bit of control by submitting/withholding a block.

Additionally, an attacker can also choose thier randao reveal (to some degree) by choosing their private key. Note that the attacker has a long time (now until genesis) to compute these private keys.

Some rough calculations show that gaining _full_ control of future shufflings is infeasible, but _some_ degree of control is possible.

#### Breaking eth1 linking

If used with above attack (influencing future shuffling), it may be feasible for the attacker to produce a majority of the blocks the `ETH1_VOTING_PERIOD` and break the eth1-linking.

We have not looked into the feasibility of this attack.

### Additional Comments

Setting `v` seems to be an instance of "eth1 transaction ordering attack" on smart contracts. Here are some ways to set `v`:

- Be an eth1 miner and supress transactions.
- Use known gas-war techniques to manipulate the ordering of deposits.
  - `1 ether` deposits can be used to cheapen this attack.

djrtwo:
closed via #1447

PR #1447, opened 2019-10-25
Implement solution for #1446, based on suggested use of eth1 hash

protolambda:
First of all, I am very aware that the use of block hashes does not provide great randomness.

However, the concern was that the selection of committees of the early epochs could realistically be biased by deposit ordering manipulation. Setting some initial randomness, even if biasable by a few bits, should help work around this attack vector.
This solution came out of a conversation between @paulhauner and @djrtwo in #1446, but I am open to any better alternatives.

Also note that I adapted it to:
1. Fill randao mixes based on the `LOOKAHEAD` constant
2. Use the eth1 hash ~~as seed, and set different randao mixes.~~

Alternatively we could also fill the whole mixes array with seeded hashes: no concern about exact `LOOKAHEAD` value, but much more work (mainnet history size = `2**16`). Or we do not care about different mixes, and copy over the eth1 hash.

Edit: not as seed, just plain copy, and into the full extent of the randao-mixes (less error prone, more minimal). Thanks to Justin for pointing out that `get_seed` already mixes in the epoch, for us to not care about duplicate randao mixes here, just what we need.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1112,10 +1112,8 @@ def initialize_beacon_state_from_eth1(eth1_block_hash: Hash,
         genesis_time=eth1_timestamp - eth1_timestamp % SECONDS_PER_DAY + 2 * SECONDS_PER_DAY,
         eth1_data=Eth1Data(block_hash=eth1_block_hash, deposit_count=len(deposits)),
         latest_block_header=BeaconBlockHeader(body_root=hash_tree_root(BeaconBlockBody())),
+        randao_mixes=[eth1_block_hash] * EPOCHS_PER_HISTORICAL_VECTOR,  # to limit deposit order bias in early epochs
     )
-    # Set the initial RANDAO mixes to hashes seeded by the Eth1 hash, to limit deposit ordering committee bias.
-    for i in range(MIN_SEED_LOOKAHEAD + 1):
-        state.randao_mixes[EPOCHS_PER_HISTORICAL_VECTOR - i - 1] = hash(eth1_block_hash + int_to_bytes(i, 8))
 
     # Process deposits
     leaves = list(map(lambda deposit: deposit.data, deposits))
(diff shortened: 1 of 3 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1113,6 +1113,9 @@ def initialize_beacon_state_from_eth1(eth1_block_hash: Hash,
         eth1_data=Eth1Data(block_hash=eth1_block_hash, deposit_count=len(deposits)),
         latest_block_header=BeaconBlockHeader(body_root=hash_tree_root(BeaconBlockBody())),
     )
+    # Set the initial RANDAO mixes to hashes seeded by the Eth1 hash, to limit deposit ordering committee bias.
+    for i in range(MIN_SEED_LOOKAHEAD + 1):
+        state.randao_mixes[EPOCHS_PER_HISTORICAL_VECTOR - i - 1] = hash(eth1_block_hash + int_to_bytes(i, 8))
```
JustinDrake: What about

```python
    for i in range(EPOCHS_PER_HISTORICAL_VECTOR):
        state.randao_mixes[i] = eth1_block_hash
```

It seems to have equivalent security and it's more obvious what is going on (no `+ 1`, `- i - 1`, `int_to_bytes`, `hash`, or `MIN_SEED_LOOKAHEAD`).

djrtwo: agreed here. no harm in backfilling the entire array. That is essentially what we had when they were all default value of ZERO

djrtwo: ah, just saw that you are hashing in the eth1_block_hash with an integer index. I see some marginal value in doing that, but I think just filling as `eth1_block_hash` is fine.

JustinDrake: > I see some marginal value in doing that

Why? `get_seed ` already mixes in the `epoch` with `hash(domain_type + int_to_bytes(epoch, length=8) + mix)`.

protolambda: Ah missed that, thought there would be some assumption somewhere that every mix should be unique. Just copying over and completely filling is better then.

PR #1458, opened 2019-10-29
Update link to current bls standard draft.

gnattishness:


gnattishness:
Negligible, but might as well :)

PR #1459, opened 2019-10-31
hash cleanups

arnetheduck:
* one more hash tree root gone for blocks - block hashes are always
signing roots!
* use simple serialize data types consistently

### DIFF
--- a/specs/networking/p2p-interface.md
+++ b/specs/networking/p2p-interface.md
@@ -320,7 +320,7 @@ Here, `result` represents the 1-byte response code.
 
 The token of the negotiated protocol ID specifies the type of encoding to be used for the req/resp interaction. Two values are possible at this time:
 
--  `ssz`: the contents are [SSZ-encoded](../simple-serialize.md). This encoding type MUST be supported by all clients. For objects containing a single field, only the field is SSZ-encoded not a container with a single field. For example, the `BeaconBlocksByRoot` request is an SSZ-encoded list of `Hash`'s.
+-  `ssz`: the contents are [SSZ-encoded](../simple-serialize.md). This encoding type MUST be supported by all clients. For objects containing a single field, only the field is SSZ-encoded not a container with a single field. For example, the `BeaconBlocksByRoot` request is an SSZ-encoded list of `Bytes32`'s.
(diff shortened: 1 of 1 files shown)


specs/networking/p2p-interface.md
```diff
@@ -344,20 +344,20 @@ constituents individually as `response_chunk`s. For example, the
...
-The fields are:
+The fields are, as seen by the client at the time of sending the message:
 
 - `head_fork_version`: The beacon_state `Fork` version.
-- `finalized_root`: The latest finalized root the node knows about.
-- `finalized_epoch`: The latest finalized epoch the node knows about.
-- `head_root`: The block hash tree root corresponding to the head of the chain as seen by the sending node.
-- `head_slot`: The slot corresponding to the `head_root`.
+- `finalized_root`: The signing root of the latest finalized block.
+- `finalized_epoch`: The epoch of the block corresponding to `finalized_root`.
```
djrtwo: This isn't exactly correct. The `finalized_epoch` and the `finalized_root` are best just grabbed from state.

There are scenarios in which `latest_finalized_epoch > slot_to_epoch(latest_finalized_block.slot)`

arnetheduck: Ah, might as well make it explicit, wdyt?

arnetheduck: ...though technically that's a bit more restrictive, and clients can/should "optimize" it based on when `finalized_checkpoint` is updated.

protolambda:
This reminds me, we should drop `Hash` in favor of `Bytes32` in the other parts of the spec. It's always confusing with `hash` (the function).

specs/networking/p2p-interface.md
djrtwo: Do you mean update early if they know something is finalized, but the epoch transition hasn't yet happened?

About to make a mod to the fork choice in a separate PR that only allows updating the finalized info on the fork choice in a restricted range to avoid a type of "bounce attack" -- https://ethresear.ch/t/prevention-of-bouncing-attack-on-ffg/6114

Might affect this language here. we'll see

arnetheduck: hm, I guess perhaps that's possible if enough votes accumulate early (?), but I was more thinking about blocks during non-epoch slots - ie the finalized state of the latest epoch transition block is enough, you don't need the head block technically

PR #1460, opened 2019-10-31
Specify inclusive range for genesis deposits

paulhauner:
I think the change is pretty self-explanatory :)

Keen to hear if my understanding is wrong (i.e., it should _not_ be an inclusive range).


djrtwo:
great! can you rebase to branch `v09x`?
I'll release this in the pending `v0.9.1`

paulhauner:
Done :)

PR #1463, opened 2019-11-04
period_tail in get_eth1_vote

paulhauner:
The change I have suggested targets [`get_eth1_vote()`](https://github.com/ethereum/eth2.0-specs/blob/v0.8.4/specs/validator/0_beacon-chain-validator.md#eth1-data) and is the result of the following 3 (potentially incorrect) observations:  

1. `slot` seems a misleading variable name for indexing `state.eth1_data_votes`.
1. `slot % SLOTS_PER_ETH1_VOTING_PERIOD` when `len(state.eth1_data_votes) <= SLOTS_PER_ETH1_VOTING_PERIOD` is a no-op.
1. I assume the `period_tail` is designed to validate _earlier_ `eth1_data_votes` if they are in `all_eth1_data` (but not in `new_eth1_data`). However, when `period_tail = slot >= integer_squareroot(...)`, only _later_ `eth1_data_votes` are tested against `all_eth1_data`.

### DIFF
--- a/specs/validator/0_beacon-chain-validator.md
+++ b/specs/validator/0_beacon-chain-validator.md
@@ -239,8 +239,8 @@ def get_eth1_vote(state: BeaconState, previous_eth1_distance: uint64) -> Eth1Dat
     all_eth1_data = [get_eth1_data(distance) for distance in range(ETH1_FOLLOW_DISTANCE, previous_eth1_distance)]
 
     valid_votes = []
-    for slot, vote in enumerate(state.eth1_data_votes):
-        period_tail = slot % SLOTS_PER_ETH1_VOTING_PERIOD >= integer_squareroot(SLOTS_PER_ETH1_VOTING_PERIOD)
+    for i, vote in enumerate(state.eth1_data_votes):
+        period_tail = i < integer_squareroot(SLOTS_PER_ETH1_VOTING_PERIOD)
         if vote in new_eth1_data or (period_tail and vote in all_eth1_data):
             valid_votes.append(vote)
 


paulhauner:
It's probably worth considering this alongside #1464

djrtwo:
1. agreed
2. agreed
3. You have this flipped. Essentially, for earlier in the period, we just want people to attempt to vote on new data, so in the beginning of the period, only pile on your vote to votes that are for newer data. But once we've entered the tail of the period, we just want consensus so we consider all votes

----------------------------------

The previous method was making the assumption that there was one eth1 vote per slot. I am considering that maybe it should just be the following to remove this assumption. 

```python
def get_eth1_vote(state: BeaconState, previous_eth1_distance: uint64) -> Eth1Data:
    new_eth1_data = [get_eth1_data(distance) for distance in range(ETH1_FOLLOW_DISTANCE, 2 * ETH1_FOLLOW_DISTANCE)]
    all_eth1_data = [get_eth1_data(distance) for distance in range(ETH1_FOLLOW_DISTANCE, previous_eth1_distance)]

    period_tail = state.slot % SLOTS_PER_ETH1_VOTING_PERIOD >= integer_squareroot(SLOTS_PER_ETH1_VOTING_PERIOD)
    if period_tail:
        votes_to_consider = all_eth1_data
    else:
        votes_to_consider = new_eth1_data

    valid_votes = [vote for vote in state.eth1_data_votes if vote in votes_to_consider]

    return max(
        valid_votes,
        key=lambda v: (valid_votes.count(v), -all_eth1_data.index(v)),  # Tiebreak by smallest distance
        default=get_eth1_data(ETH1_FOLLOW_DISTANCE),
    )
```

djrtwo:
closing in favor of #1468

PR #1465, opened 2019-11-05
Bounce attack resistance

djrtwo:
Only allow updates to the latest justified for fork choice if within the first `SAFE_SLOTS_TO_UPDATE_JUSTIFIED` slots of an epoch _or_ if the new justified does not conflict with the previous. If a new justified is learned about but doesn't satisfy these constraints, queue it up for consideration at next epoch boundary.

This addresses the bouncing attack discussed [here](https://ethresear.ch/t/prevention-of-bouncing-attack-on-ffg/6114).

Testing justification and finalization in the fork choice is painful and needs to be reconsidered

### DIFF
--- a/specs/core/0_fork-choice.md
+++ b/specs/core/0_fork-choice.md
@@ -47,7 +47,7 @@ The head block root associated with a `store` is defined as `get_head(store)`. A
 
 | Name | Value | Unit | Duration |
 | - | - | :-: | :-: |
-| `SAFE_SLOTS_TO_UPDATE_JUSTIFIED` | `2**3 (= 8)` | slots | 96 seconds |
+| `SAFE_SLOTS_TO_UPDATE_JUSTIFIED` | `2**3` (= 8) | slots | 96 seconds |
 
 ### Helpers
 
--- a/specs/core/0_fork-choice.md
+++ b/specs/core/0_fork-choice.md
@@ -148,8 +148,6 @@ def get_head(store: Store) -> Hash:
 
 ```python
 def should_update_justified_checkpoint(store: Store, justified_checkpoint: Checkpoint) -> bool:
-    current_epoch = compute_epoch_at_slot(get_current_slot(store))
-
     if get_current_slot(store) % SLOTS_PER_EPOCH < SAFE_SLOTS_TO_UPDATE_JUSTIFIED:
         return True
 
--- a/specs/core/0_fork-choice.md
+++ b/specs/core/0_fork-choice.md
@@ -96,9 +96,18 @@ def get_genesis_store(genesis_state: BeaconState) -> Store:
     )
 ```
 
+#### `get_current_slot`
+
 ```python
 def get_current_slot(store: Store) -> Slot:
     return Slot((store.time - store.genesis_time) // SECONDS_PER_SLOT)
+```
+
+#### `compute_slots_since_epoch_start`
+
+```python
+def compute_slots_since_epoch_start(slot: Slot) -> int:
+    return slot - compute_start_slot_at_epoch(compute_epoch_at_slot(slot))
 ```
 
 #### `get_ancestor`
@@ -149,7 +158,7 @@ def get_head(store: Store) -> Hash:
 
 ```python
 def should_update_justified_checkpoint(store: Store, new_justified_checkpoint: Checkpoint) -> bool:
-    if get_current_slot(store) % SLOTS_PER_EPOCH < SAFE_SLOTS_TO_UPDATE_JUSTIFIED:
+    if compute_slots_since_epoch_start(get_current_slot(store)) < SAFE_SLOTS_TO_UPDATE_JUSTIFIED:
         return True
 
     new_justified_block = store.blocks[new_justified_checkpoint.root]
--- a/specs/core/0_fork-choice.md
+++ b/specs/core/0_fork-choice.md
@@ -47,7 +47,7 @@ The head block root associated with a `store` is defined as `get_head(store)`. A
 
 | Name | Value | Unit | Duration |
 | - | - | :-: | :-: |
-| `SAFE_SLOTS_TO_UPDATE_JUSTIFIED` | `9` | slots | 108 seconds |
+| `SAFE_SLOTS_TO_UPDATE_JUSTIFIED` | `2**3 (= 8)` | slots | 96 seconds |
 
 ### Helpers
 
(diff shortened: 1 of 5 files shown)


specs/core/0_fork-choice.md
```diff
@@ -60,8 +66,10 @@ class LatestMessage(object):
 @dataclass
 class Store(object):
     time: uint64
+    genesis_time: uint64
     justified_checkpoint: Checkpoint
     finalized_checkpoint: Checkpoint
+    queued_justified_checkpoints: List[Checkpoint, 2**40] = field(default_factory=list)
```
protolambda: - `2**40` is a magic number
- List with a limit is unnecessary, it's not a SSZ container. `Sequence[Checkpoint]` works. 
- The entry with the highest epoch is only ever used, and then the list is flushed. Why not change it to a single checkpoint and update with `x = max(old.epoch, new.epoch)`

specs/core/0_fork-choice.md
```diff
@@ -130,13 +144,43 @@ def get_head(store: Store) -> Hash:
         head = max(children, key=lambda root: (get_latest_attesting_balance(store, root), root))
 ```
 
+#### `should_update_justified_checkpoint`
+
+```python
+def should_update_justified_checkpoint(store: Store, justified_checkpoint: Checkpoint) -> bool:
+    if get_current_slot(store) % SLOTS_PER_EPOCH < SAFE_SLOTS_TO_UPDATE_JUSTIFIED:
```
protolambda: We don't make the assumption that slots are epoch aligned anywhere else in the spec (afaik). Maybe we add a helper func for slots since epoch start? There exists a chance that epoch length changes in a fork, like with v0.9

specs/core/0_fork-choice.md
```diff
@@ -130,13 +144,43 @@ def get_head(store: Store) -> Hash:
...
+        return True
+
+    justified_block = store.blocks[justified_checkpoint.root]
+    if justified_block.slot <= compute_start_slot_at_epoch(store.justified_checkpoint.epoch):
```
protolambda: Quite confusing here with `justified_checkpoint` vs `store.justified_checkpoint`. Was thinking "can this even hit this branch" before I noticed. Maybe change the argument name?

specs/core/0_fork-choice.md
```diff
@@ -130,13 +144,43 @@ def get_head(store: Store) -> Hash:
...
 ```python
 def on_tick(store: Store, time: uint64) -> None:
+    previous_slot = get_current_slot(store)
+
+    # update store time
     store.time = time
+
+    current_slot = get_current_slot(store)
+    # Not a new epoch, return
+    if not (current_slot > previous_slot and current_slot % SLOTS_PER_EPOCH == 0):
```
protolambda: ditto

specs/core/0_fork-choice.md
```diff
@@ -130,13 +144,43 @@ def get_head(store: Store) -> Hash:
...
+        return
+    # If new epoch and there are queued_justified_checkpoints, update if any is better than the best in store
+    if any(store.queued_justified_checkpoints):
+        best_justified_checkpoint = max(store.queued_justified_checkpoints, key=lambda checkpoint: checkpoint.epoch)
+        if best_justified_checkpoint.epoch > store.justified_checkpoint.epoch:
+            store.justified_checkpoint = best_justified_checkpoint
+    store.queued_justified_checkpoints = []
```
protolambda: See other comment. We can just track the `best_justified_checkpoint` instead of maintaining a whole queue?

specs/core/0_fork-choice.md
djrtwo: Meant to address that before review. whoops

Can't use `Sequence` because it is immutable! 
I can import `typing` and use `typing.List`. It's a bit confusing because we overroad the built in type..

djrtwo: good call on just keeping one value. will do

specs/core/0_fork-choice.md
djrtwo: But we do make the assumption that epochs are `SLOTS_PER_EPOCH` within helpers, but I suppose the internals there can be easily swapped out.

Will add a helper

specs/core/0_fork-choice.md
```diff
@@ -130,13 +154,43 @@ def get_head(store: Store) -> Hash:
...
+    previous_slot = get_current_slot(store)
+
+    # update store time
     store.time = time
+
+    current_slot = get_current_slot(store)
+    # Not a new epoch, return
+    if not (current_slot > previous_slot and current_slot % SLOTS_PER_EPOCH == 0):
+        return
+    # Update store.justified_checkpoint if a better checkpoint is know
```
protolambda:
```suggestion
    # Update store.justified_checkpoint if a better checkpoint is known
```

specs/core/0_fork-choice.md
```diff
@@ -130,13 +154,43 @@ def get_head(store: Store) -> Hash:
         head = max(children, key=lambda root: (get_latest_attesting_balance(store, root), root))
 ```
 
+#### `should_update_justified_checkpoint`
+
+```python
+def should_update_justified_checkpoint(store: Store, new_justified_checkpoint: Checkpoint) -> bool:
```
protolambda: A reference to the bounce attack, or description of why we have the `best_justified_checkpoint` would be nice for the spec.

specs/core/0_fork-choice.md
```diff
@@ -130,13 +154,43 @@ def get_head(store: Store) -> Hash:
...
 ```python
 def on_tick(store: Store, time: uint64) -> None:
+    previous_slot = get_current_slot(store)
+
+    # update store time
     store.time = time
+
+    current_slot = get_current_slot(store)
+    # Not a new epoch, return
+    if not (current_slot > previous_slot and current_slot % SLOTS_PER_EPOCH == 0):
```
protolambda:
```suggestion
    if not (current_slot > previous_slot and compute_slots_since_epoch_start(current_slot) == 0):
```

protolambda:
```suggestion
    if not (current_slot > previous_slot and compute_slots_since_epoch_start(current_slot) == 0):
```

protolambda: Oops, duplicate comment due to reconnect.

PR #1466, opened 2019-11-05
Decoy flip flop resistance

djrtwo:
Integrate FMD GHOST into the fork choice rule to prevent "decoy flip flop attacks" discussed [here](https://ethresear.ch/t/decoy-flip-flop-attack-on-lmd-ghost/6001).

* Only consider an `attestation` in `on_attestation` if it is from the current or previous epoch (wrt wall clock time).
* add a few tests

Note: this currently has `bounce-attack` as base branch for easy review. Will swap to `v09x` once `bounce-attack` is merged

### DIFF
--- a/specs/core/0_fork-choice.md
+++ b/specs/core/0_fork-choice.md
@@ -224,6 +224,9 @@ def on_block(store: Store, block: BeaconBlock) -> None:
 def on_attestation(store: Store, attestation: Attestation) -> None:
     target = attestation.data.target
 
+    # Attestations must be from the current or previous epoch
+    current_epoch = compute_epoch_at_slot(get_current_slot(store))
+    assert target.epoch in [current_epoch, current_epoch - 1 if current_epoch > GENESIS_EPOCH else GENESIS_EPOCH]
     # Cannot calculate the current shuffling if have not seen the target
     assert target.root in store.blocks
 
--- a/specs/core/0_fork-choice.md
+++ b/specs/core/0_fork-choice.md
@@ -233,9 +233,11 @@ def on_block(store: Store, block: BeaconBlock) -> None:
 def on_attestation(store: Store, attestation: Attestation) -> None:
     target = attestation.data.target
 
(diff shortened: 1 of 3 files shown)


specs/core/0_fork-choice.md
```diff
@@ -224,6 +224,9 @@ def on_block(store: Store, block: BeaconBlock) -> None:
 def on_attestation(store: Store, attestation: Attestation) -> None:
     target = attestation.data.target
 
+    # Attestations must be from the current or previous epoch
+    current_epoch = compute_epoch_at_slot(get_current_slot(store))
+    assert target.epoch in [current_epoch, current_epoch - 1 if current_epoch > GENESIS_EPOCH else GENESIS_EPOCH]
```
protolambda: Just noticed that the comparison is different than `get_previous_epoch` but still valid. And we can't reuse that state helper func sadly.
I'm also wondering if the spec is saying something about a target epoch set equal to the `GENESIS_EPOCH`. Checked beacon and validator spec, but everything looks to implicitly support it. Maybe we should add a note about genesis epoch for completeness?

djrtwo: done

djrtwo: We skip over having to deal with the corner case for finality of source and target being equal during the first epoch in `process_justification_and_finalization` with the following.

```python
    if get_current_epoch(state) <= GENESIS_EPOCH + 1:
        return
```

As for the fork choice, we want epoch 0 attestations to be incorporated

djrtwo: ah, you meant a note about genesis epoch for validator attestations/fork choice. reasonable

protolambda: Yes, I think `GENESIS_EPOCH` deserves a mention in the validator and/or fork-choice spec on how to attest it. Even if not that different, reading that something isn't actually missing is still useful.

djrtwo: done

PR #1467, opened 2019-11-07
reduce genesis validator count for mainnet config

djrtwo:
Reduce genesis validator count for mainnet config to allow for a lower count and higher potential return in early Phase 0 validation if long lockup/participation is too discouraging for the previously specified number.

This number was previously increased to prevent "gatekeeper attacks" in which an early moving attacker quickly hits the minimum, starts the chain, and never allows in new deposits. This attack scenario is unlikely due to the additional time parameter (`MIN_GENESIS_TIME`) along with `MIN_GENESIS_ACTIVE_VALIDATOR_ACCOUNT`. The time parameter ensures that anyone that wants to get in on genesis has time to do so (in addition to the potential attacker).

Even if a "gatekeeper attack" occurs, it would be very easy to detect and give us a chance to exercise our "social coordination" tools to fork out an attacking balance.


JustinDrake:
Potential counter arguments:

1) 0.5M ETH is less than 0.5% of total supply, i.e. a "de minimis" amount in the grand scheme of things.
2) This puts more pressure on choosing an appropriate `MIN_GENESIS_TIME`. (Note that a majority attacker can do more than gatekeeping. And social coordination should not be taken lightly—I see it as a last resort, somewhat like the DAO fork.)
3) Edgeware (a project relatively unknown compared to Eth2) locked [1.2M ETH](https://commonwealth.im/stats/edgeware). I'd hope we can lock more ETH than Edgeware.
4) Long lockups being a deterrent for participation is largely speculation at this point. Note that over half of the Edgeware locks (641K ETH) have a lock duration of 12 months, signalling that hodlers are happy to hodl for the long term.

econoar:
@JustinDrake the one difference with Edgeware and phase 0 staking is with Edgeware your lockup time is known and returns start immediately (you can back into an interest rate from total supply created). An issue I'm really concerned about in phase 0 has to do with the fact that stakers will not get a return until the chain actually starts. This could create a standoff between stakers with a "no, you go first" mentality.

The pitch to early stakers before the chain starts is one of unknown time and therefore unknown returns. The counter to that is there are a lot of people willing to hold ETH for years and never cash it but then we are leaning on a bit of altruism.

I'm glad this is being thought out and discussed!

JustinDrake:
> This could create a standoff between stakers with a "no, you go first" mentality.

A standoff could still happen, even with `MIN_GENESIS_ACTIVE_VALIDATOR_ACCOUNT` reduced. One way to address standoff situations is to introduce a `MAX_GENESIS_TIME` (e.g. Jan 3, 2020) whereby phase 0 launches even if `MIN_GENESIS_ACTIVE_VALIDATOR_ACCOUNT` is not reached.

It would also be helpful if you could provide evidence suggesting that Ethereans are actually standoff-ish. I'd argue there's evidence that Ethereans are not overly standoff-ish. For example, look at the [ETH presale dynamics](https://medium.com/@hasufly/ethereum-presale-dynamics-revisited-c1b70ac38448). Of the 22,500 BTC raised in the first two weeks ~1/3 rushed in at the start, ~1/3 gradually came in, and ~1/3 rushed in at the first cut-off point. The "rational" presale behaviour was to standoff to maximise participation information, yet only 1/3 of the investment money was standoff-ish.

PR #1476, opened 2019-11-11
Phase 0 attestation topic directives

djrtwo:
* clarify when and how to join attestation pubsub topics in the validator guide
* dictate that validators maintain semi-persistent subscriptions to random attestation pubsub topics during phase 0

### DIFF
--- a/specs/validator/0_beacon-chain-validator.md
+++ b/specs/validator/0_beacon-chain-validator.md
@@ -183,9 +183,13 @@ def is_proposer(state: BeaconState,
 
 The beacon chain shufflings are designed to provide a minimum of 1 epoch lookahead on the validator's upcoming committee assignments for attesting dictated by the shuffling and slot. Note that this lookahead does not apply to proposing, which must be checked during the epoch in question.
 
-`get_committee_assignment` should be called at the start of each epoch to get the assignment for the next epoch (`current_epoch + 1`). A validator should plan for future assignments by noting at which future slot they will have to attest.
+`get_committee_assignment` should be called at the start of each epoch to get the assignment for the next epoch (`current_epoch + 1`). A validator should plan for future assignments by noting at which future slot they will have to attest and joining the committee index attestation subnet related to their committee assignment.
 
-Specifically, a validator should call `get_committee_assignment(state, next_epoch, validator_index)` when checking for next epoch assignments.
+Specifically a validator should:
+* Call `get_committee_assignment(state, next_epoch, validator_index)` when checking for next epoch assignments.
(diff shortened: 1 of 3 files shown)


specs/validator/0_beacon-chain-validator.md
```diff
@@ -79,6 +80,8 @@ All terminology, constants, functions, and protocol mechanics defined in the [Ph
 | - | - | :-: | :-: |
 | `ETH1_FOLLOW_DISTANCE` | `2**10` (= 1,024) | blocks | ~4 hours |
 | `TARGET_AGGREGATORS_PER_COMMITTEE` | `2**4` (= 16) | validators | |
+| `RANDOM_SUBNETS_PER_VALIDATOR` | `2**0` (= 1) | subnets | |
+| `EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION` | `2**8` (= 256) | epochs | ~27 hours |
```
hwwhww: Although they are not consensus constants, I think it would be easier if we set all these configs in the minimal, mainnet, and testnet configs, so that the client teams only need to look up the same configuration for testnet.

djrtwo: done

specs/validator/0_beacon-chain-validator.md
```diff
@@ -414,6 +421,10 @@ Where
 * `selection_proof` is the signature of the slot (`slot_signature()`).
 * `aggregate` is the `aggregate_attestation` constructed in the previous section.
 
+## Phase 0 attestation subnet stability
+
+Because Phase 0 does not have shards and thus does not have Shard Committees, there is no stable backbone to the attestation subnets (`committee_index{subnet_id}_beacon_attestation`). To provide this stability, each validator must randomly select and remain subscribed to `RANDOM_SUBNETS_PER_VALIDATOR` attestation subnets. The lifetime of each random subscription must be _at least_ `EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION`.
```
protolambda: Can we add a note that this does not mean "`epoch % EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION`"? Gradual changes (looking at the complete network) seem better here, we do not want everyone to search and switch on the exact same time.

djrtwo: yeah, I was worried about the same thing. Mass exodus. I'll put that the lifetime should be between [EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION, 2*EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION]`

djrtwo: updated

Issue #1406, opened 2019-09-11
Attestations for blocks from incorrect slots


djrtwo:
@paulhauner Brought up an interesting scenario in which an attestation votes for a beacon block root from a future slot.

With respect to the state transition, it doesn't really matter. It just looks like an incorrect head root and thus won't be rewarded for the head vote. This operates the same way as though the head voted on was from a different fork.

The fork choice on the other hand currently handles this in a degenerate manner. The only check currently when inducting an attestation into latest messages is that the target_epoch is greater than the previous latest message for the validator and that the slot of the attestation is not in the future. We make no checks wrt `beacon_block_root`, but we store the `beacon_block_root` into the `latest_messages` store for that validator.

This results in a case in which the validator can successfully influencing the fork choice of a head block later than their assigned slot and can even use an attestation from a past epoch to cast a vote on a block from a future (wrt to the attestation slot) epoch.

I haven't fully assessed the attack domain here, but the simple solution to simply remove this case from the attack surface is to make these attestations invalid wrt the fork choice.

djrtwo:
> This should be allowed because the sender of the attestation might have failed to receive the block from the slot?

Yep! wasn't thinking. good catch!

PR #1477, opened 2019-11-12
Forkchoice mods

djrtwo:
* Ensure `attestation.data.target.root` is known
* Ensure `attestation.data.beacon_block_root` is known
* Do not allow attestations to vote for blocks from future slots (#1406)

### DIFF
--- a/specs/core/0_fork-choice.md
+++ b/specs/core/0_fork-choice.md
@@ -238,6 +238,12 @@ def on_block(store: Store, block: BeaconBlock) -> None:
 
 ```python
 def on_attestation(store: Store, attestation: Attestation) -> None:
+    """
+    Run ``on_attestation`` upon receiving a new ``attestation`` from either within a block or directly on the wire.
+
+    An ``attestation`` that is asserted as invalid may be valid at a later time,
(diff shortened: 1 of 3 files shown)


specs/core/0_fork-choice.md
```diff
@@ -248,10 +248,17 @@ def on_attestation(store: Store, attestation: Attestation) -> None:
     # Cannot calculate the current shuffling if have not seen the target
     assert target.root in store.blocks
 
+    # Attestations target be for a known block. If target block is unknown, delay consideration until the block is found
```
protolambda: > delay consideration until the block is found

Considered returning the block root to await instead. But then there are slot/time based exit conditions as well, and we likely don't want more complicated scheduling in the spec. (Also since not all cases may be fulfilled later, e.g. a non-existing block root).

Agree that there is some freedom in processing here, and the delayed consideration shouldn't be part of the spec.
Same for delayed-consideration cases below.

A warning at the top of the function may be in place: "An `Attestation` that is asserted as invalid may be valid at a later time, consider scheduling it for later processing in such case"

PR #1478, opened 2019-11-12
Avoid Hash <> hash() type and func confusion

protolambda:
Why:
- `hash()` is the SHA-256 of empty input. `Hash()` is the default instantiation of the `Bytes32` alias, i.e. 32 zeroed bytes.
- `Hash(*input*)` casts the input to the `Bytes32` alias (not changing the contents!). Whereas `hash(*input*)` hashes the input to a `Hash` (with SHA-256).

Changelog (no functional changes):
- remove `Hash` alias of `Bytes32`
- introduce `Root` alias of `Bytes32`
- change roots to `Root`
- change non-roots to `Bytes32`

Difficult choices:
- a `Bytes32` inside a proof is technically also a combination of nodes. But not necessarily so for the very bottom sibling. So not a `Root` or `Hash`
  - Also, `Hash` is especially confusing, as a leaf item is packed into 32 bytes but not hashed.
- There are a bunch of hashes that are not SSZ roots that we start calling `Bytes32` now:
  - Eth1 block hash
  - Randao mix / seeds
  - `withdrawal_credentials`

I am opposed to making a separation between `SigningRoot` and `HashRoot`: the roots should be typed by the (summary) type they represent, not the function that is used.

`signing_root` already is a hack to make the same object definition work for multiple purposes. But really signatures shouldn't be part of the tree, unless intended so. A SSZ type simply cannot be represented by the same immutable cached (each node) tree, covering both signing-root and hash-tree-root, as the tree structure is dfifferent.
And since we are using `signing_root` and `hash_tree_root` pretty much exclusively, I would prefer to just do away with `signing_root`. Maybe in favor of a way to represent a container with a signature as a serialization-only thing.

Open for suggestions/comments.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -137,7 +137,7 @@ We define the following Python custom types for type hinting and readability:
 | `CommitteeIndex` | `uint64` | a committee index at a slot |
 | `ValidatorIndex` | `uint64` | a validator registry index |
 | `Gwei` | `uint64` | an amount in Gwei |
-| `Hash` | `Bytes32` | a hash |
+| `Root` | `Bytes32` | a Merkle root |
 | `Version` | `Bytes4` | a fork version number |
 | `DomainType` | `Bytes4` | a domain type |
 | `Domain` | `Bytes8` | a signature domain |
@@ -275,7 +275,7 @@ class Fork(Container):
 ```python
 class Checkpoint(Container):
     epoch: Epoch
-    root: Hash
+    root: Root
 ```
 
 #### `Validator`
@@ -283,7 +283,7 @@ class Checkpoint(Container):
 ```python
 class Validator(Container):
     pubkey: BLSPubkey
-    withdrawal_credentials: Hash  # Commitment to pubkey for withdrawals
+    withdrawal_credentials: Bytes32  # Commitment to pubkey for withdrawals
     effective_balance: Gwei  # Balance at stake
     slashed: boolean
     # Status epochs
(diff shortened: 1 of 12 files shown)


ralexstokes:
changes look sound @protolambda  -- can you elaborate on what you mean by "do away with `signing_root`"?

protolambda:
@ralexstokes As in completely remove it. With more type definitions it's technically not necessary. And it's difficult to fit one type that can be merkleized in two different forms onto a single binary tree.

test_libs/pyspec/eth2spec/test/sanity/test_blocks.py
```diff
@@ -63,7 +63,7 @@ def test_empty_block_transition(spec, state):
 
     assert len(state.eth1_data_votes) == pre_eth1_votes + 1
     assert spec.get_block_root_at_slot(state, pre_slot) == block.parent_root
-    assert spec.get_randao_mix(state, spec.get_current_epoch(state)) != spec.Hash()
+    assert spec.get_randao_mix(state, spec.get_current_epoch(state)) != spec.Root()
```
djrtwo:
```suggestion
    assert spec.get_randao_mix(state, spec.get_current_epoch(state)) != spec.Bytes32()
```

specs/core/0_beacon-chain.md
```diff
@@ -329,25 +329,25 @@ class PendingAttestation(Container):
 
 ```python
 class Eth1Data(Container):
-    deposit_root: Hash
+    deposit_root: Root
     deposit_count: uint64
-    block_hash: Hash
+    block_hash: Bytes32
```
djrtwo: This is the one field I was torn about, but this is correct

PR #1480, opened 2019-11-15
Rename BytesN to ByteVector, rename Bytes to ByteList

### LINKED ISSUE #1479: BytesN SSZ Alias is Misleading
Hi all, currently the SSZ spec [alias section](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/simple-serialize.md#aliases) mentions:

```
BytesN to Vector[byte, N] (this is not a basic type)
```
and we also have the type `Byte[N]` in some other places across the repo including the test generators for ssz generic containers [here](https://github.com/ethereum/eth2.0-specs/blob/40cb72ec112903a28cbfc9e310e14844680476e5/test_generators/ssz_generic/ssz_container.py#L35). This tricky nomenclature led us to some painful debugging and quite some time spent fixing broken tests in Prysm. The alias doesn't add much utility and we at @prylabs would just prefer to stick to the suggested naming conventions for lists vs. vectors:

```
List[byte, N]
Vector[byte, N]
```

protolambda:
People reading the pyspec SSZ util code got confused by `Bytes` vs `BytesN`. So this PR changes it to a more verbose but also more clear `ByteList` and `ByteVector`. No functional changes.

Fixes #1479

### DIFF
--- a/scripts/build_spec.py
+++ b/scripts/build_spec.py
@@ -56,7 +56,7 @@ from eth2spec.utils.ssz.ssz_impl import (
 )
 from eth2spec.utils.ssz.ssz_typing import (
     BasicValue, Elements, BaseBytes, BaseList, SSZType,
-    Container, List, Vector, Bytes, BytesN, Bitlist, Bitvector, Bits,
+    Container, List, Vector, ByteList, ByteVector, Bitlist, Bitvector, Bits,
     Bytes1, Bytes4, Bytes8, Bytes32, Bytes48, Bytes96,
     uint64, bit, boolean, byte,
 )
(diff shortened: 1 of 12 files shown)


djrtwo:
I support this

JustinDrake:
> `ByteList, ByteVector, Bitlist, Bitvector`

I also support this but I guess we need consistent capitalisation. Could rename `Bitlist` to `BitList`.

protolambda:
Agree with Justin, but there's something to mixed capitalisation too: `ByteVector[N]` is exactly a `Vector[Byte, N]`, whereas `Bitvector` is not a `Vector[Bit, N]`: it's like a bitfield instead, a full word without stops. Same thing for the list variant.

djrtwo:
Ah, you're right @protolambda. Unfortunately the subtlety in the difference in capitalization doesn't really send much of a signal to a dev imo.

That said, I think I lean toward keeping as is to reduce overhead in changing implementations and capturing the minor difference

djrtwo:
Ah, maybe should have moved to `v09x`. I suppose no need to get this out into a release soon

PR #1481, opened 2019-11-15
Pyspec SSZ; HTR caching of Vector.

protolambda:
Although I still like to see pyspec-ssz replaced with py-ssz or moved out to refactor more, it looks like py-ssz is not quite ready, or maybe not the right choice. And moving it out and refactoring brings different problems: keeping it sync, and readability / size (py-ssz pyrsistent looks great, but also ~10x more code to go through as a reader).

The two big pain points for mainnet ssz tree-hashing are:
- the large validator registry
- the large vectors of data

In tests the validator registry is not that big however, it is really the vector data that's slow without caching: 8192 roots to merkleize together.

So this PR split a vector into smaller vectors during hash-tree-root, and caches the results. On a modification, it removes the cache entry. And caching is only active for large enough vectors, of elements of an immutable type.

A quick bench shows a ~80 times improvement for a `Vector[Bytes32, 8192]`, when modifying elements in a rotation (like the historical vectors in the spec): https://gist.github.com/protolambda/4509db7f91d07b40a65ca3daf1e37685

Writing some tests and a bench of the `BeaconState` later.

Functionally this does not change SSZ or the spec. And although not too pretty, it helps to make mainnet test generation more bareable.

Note: base-branch on the other SSZ PR, which I would like to merge first, and then update the base.

### DIFF
--- a/test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
+++ b/test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py
@@ -1,7 +1,7 @@
 from ..merkle_minimal import merkleize_chunks
 from ..hash_function import hash
 from .ssz_typing import (
-    SSZValue, SSZType, BasicValue, BasicType, Series, Elements, Bits, boolean, Container, List, ByteList,
+    SSZValue, SSZType, BasicValue, BasicType, Series, Elements, Bits, boolean, Container, List, ByteList, Vector,
     Bitlist, Bitvector, uint,
 )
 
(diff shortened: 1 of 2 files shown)


protolambda:
Update: inclined to specialize the pyspec-ssz implementation for merkle proofs and caching by doing something like [I describe here](https://github.com/protolambda/eth-merkle-trees/blob/master/typing_partials.md) and [like in this POC](https://github.com/protolambda/eth-merkle-trees/blob/master/typing_partials.md), which would make this temporary caching hack unnecessary.

protolambda:
Update: I'm experimenting with a new python SSZ implementation build with binary trees as backings, thus caching every single hash in-place by default. See https://github.com/protolambda/pymerkles

So far it:
- supports every SSZ type (except Union...)
- binary tree backings (also partial trees!) work
- initial phase of implementing deserialization (required for running tests)
- compatible with spec, but needs more testing (see spec experiment file)

When I completete the serialize/deserialize part, and when it meets pyspec tests, we can swap the pyspec implemention and avoid caching-hacks like in this PR.

protolambda:
Closing this in favor of #1552 and future iterations of that.

PR #1483, opened 2019-11-15
Move new phase 1 changes into the spec

protolambda:
This PR continues the work of @vbuterin and #1427 contributors, rebases it to the latest branch dev branch (with v0.9.1 changes too), and refactors it for integration into the pyspec:
- Split proposal into:
   - "The Beacon Chain for Shards"
   - Shard transition / Fraud proof doc. (The WIP part)
- The old "misc beacon chain" and "shard data chain" docs are moved to `specs/old`. Some of it may still be relevant. To be removed from the spec once the last details are integrated well in the new docs.
    - [ ] Permanently remove old docs
- [x] Start pruning/updating the outdated parts of the custody game doc
  - From conversation with Vitalik:
    - Remove challenge records and responses, the game can be more simple and immediate now
    - "I provide a preimage of some root along with proof that you signed it, and your subkey, and on-chain the entire custody function gets executed; if the bit doesn't match you get slashed and I get a reward"
  - [x] TODO: the record creation and response processing needs to be merged
    - ~~And I'm probably missing part of the new picture still, this is not my domain.~~ More discussion helper. **implemented custody revamp with the (simplified/faster) custody slashings** :tada:
- WIP: refactor spec builder:
  - ~~Phase 0 and phase 1 are namespaced as modules~~ Same as previous
  - [x] Phase 1 copies phase 0 functions, and overrides everything with the same name
  - ~~Phase 1 type definitions overrule the phase 0 definitions. However, they inherit the phase0 types to make the new types compatible with the previous phase0 functions.~~ Override with spec-builder. Imports from more modules result in import-loops and inheritance issues. Not to mention config reload problems.
    - ~~WIP: make pyspec SSZ correctly pick up that the super-type (phase 0) field annotations should be ignored in favor of the sub-type (phase 1) fields~~ See above
   - [x] Change config system to load presets immediately before container declarations. Reload spec module to apply config changes. No code-gen re-init containers anymore.
      - [ ] TODO test new approach
- [ ] TODO: update configs with all the new constants.


specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,683 @@
...
+
+
+#### Operations
+
+```python
+def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
+    # Verify that outstanding deposits are processed up to the maximum number of deposits
+    assert len(body.deposits) == min(MAX_DEPOSITS, state.eth1_data.deposit_count - state.eth1_deposit_index)
+    
+    def process_operations(operations, fn):
```
vbuterin: This looks confusing. The outer and inner function should not have the same name.

specs/core/0_beacon-chain.md
```diff
@@ -1385,16 +1379,11 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
-        (body.attestations, process_attestation),
-        (body.deposits, process_deposit),
-        (body.voluntary_exits, process_voluntary_exit),
-        # @process_shard_receipt_proofs
-    ):
-        for operation in operations:
-            function(state, operation)
+    process_operations(body.proposer_slashings, process_proposer_slashing)
+    process_operations(body.attester_slashings, process_attester_slashing)
+    process_operations(body.attestations, process_attestations)
```
vbuterin: Are we sure that this works? Can `process_operations` just take a singular _or_ plural function like that?

specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,683 @@
...
+    withdrawal_credentials: Hash  # Commitment to pubkey for withdrawals
+    effective_balance: Gwei  # Balance at stake
+    slashed: boolean
+    # Status epochs
+    activation_eligibility_epoch: Epoch  # When criteria for activation were met
+    activation_epoch: Epoch
+    exit_epoch: Epoch
+    withdrawable_epoch: Epoch  # When validator can withdraw funds
+
+    # TODO: older pre-proposal custody field additions, keep this?
```
vbuterin: Yeah, we still need these two.

specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,683 @@
...
+    #
+    # next_custody_secret_to_reveal is initialised to the custody period
+    # (of the particular validator) in which the validator is activated
+    # = get_custody_period_for_validator(...)
+    next_custody_secret_to_reveal: uint64
+    max_reveal_lateness: Epoch
```
vbuterin: Yeah, we still need these two.

specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,683 @@
...
+```
+
+### Beacon state accessors
+
+#### `get_online_validators`
+
+```python
+def get_online_indices(state: BeaconState) -> Set[ValidatorIndex]:
+    active_validators = get_active_validator_indices(state, get_current_epoch(state))
+    return set([i for i in active_validators if state.online_countdown[i] != 0])
```
vbuterin: BTW did we decide that the uint8 list for `online_countdown` is okay? @djrtwo @JustinDrake @hwwhww 

The alternative that doesn't introduce a unique use of uint8 is my proposal to have two bitlists.

specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,683 @@
...
+```
+
+#### `get_shard_committee`
+
+```python
+def get_shard_committee(beacon_state: BeaconState, epoch: Epoch, shard: Shard) -> Sequence[ValidatorIndex]:
```
vbuterin: Consider a rename to `get_proposer_committee`. Also consider re-adding staggering.

specs/core/1_beacon-chain.md
djrtwo: The controversy here was to use a smaller bit-size (uint8) or else incur a high amount of hashing (uint64)? 

uint8 seems fine at least for now

specs/core/0_beacon-chain.md
djrtwo: `process_operations` is the name of the calling function here. As currently written, this isn't correct.

I see that there is an embedded function in the phase 1 def, but this function has no access to that

specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,683 @@
...
+    shard_transitions: Vector[ShardTransition, MAX_SHARDS]
+    # Light clients
+    light_client_signature_bitfield: Bitlist[LIGHT_CLIENT_COMMITTEE_SIZE]
+    light_client_signature: BLSSignature
+```
+
+### New extended `BeaconBlock`
+
+```python
+class BeaconBlock(phase0.BeaconBlock):
```
djrtwo: Why do you have to do an override here? because `BeaconBlockBody` was overridden, correct?

Might want to comment for clarity. Otherwise, the implementer/spec-ready would not immediately see why a new type-def

specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,683 @@
...
+    parent_root: Hash
+    state_root: Hash
+    body: BeaconBlockBody
+    signature: BLSSignature
+```
+
+### New extended `BeaconState`
+
+```python
+class BeaconState(phase0.BeaconState):
```
djrtwo: We might want some unified comment structure to identify modified and/or added fields in the "new extended" structures. Otherwise, it is very difficult to audit and understand the changes being introduced at a phase change

specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,683 @@
...
+def chunks_to_body_root(chunks):
+    return hash_tree_root(chunks + [EMPTY_CHUNK_ROOT] * (MAX_SHARD_BLOCK_CHUNKS - len(chunks)))
+```
+
+### Beacon state accessors
+
+#### `get_online_validators`
+
+```python
+def get_online_indices(state: BeaconState) -> Set[ValidatorIndex]:
```
djrtwo: Rename `get_online_validator_indices` to match naming with `get_active_validator_indices`

specs/core/1_beacon-chain.md
djrtwo: Might be reasonable to call it a `shard_committee` from which we sample proposers. We want a shard committee to provide some network stability/resources to the assigned shard and the naming implies that much

specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,683 @@
...
+    seed = get_seed(beacon_state, source_epoch, DOMAIN_SHARD_LIGHT_CLIENT)
+    return compute_committee(active_validator_indices, seed, 0, ACTIVE_SHARDS)[:TARGET_COMMITTEE_SIZE]
+```
+
+#### `get_indexed_attestation`
+
+```python
+def get_indexed_attestation(beacon_state: BeaconState, attestation: Attestation) -> AttestationAndCommittee:
+    committee = get_beacon_committee(beacon_state, attestation.data.slot, attestation.data.index)
+    return AttestationAndCommittee(committee, attestation)
```
djrtwo: The name of the function and return type here are confusing

specs/core/1_beacon-chain.md
hwwhww: I'm inclined to have `List[uint8, N]` since it's essentially just `List[byte, N]`. Two bitlists would add some confusion.

specs/core/0_beacon-chain.md
protolambda: Yes, messed up there, will add the embedded function and fix the name to not be a duplicate. Goal here is to not make a client stuck to processing lists of operations, list by list. Instead, a list like the phase 1 attestations should be processable as a whole, in between the other operation types. Open for suggestions, will fix this first.

specs/core/1_beacon-chain.md
protolambda: Added a note. That said, the override does indicate nicely that something has changed, even though it may not be obvious immediately from the field declarations.

specs/core/1_beacon-chain.md
protolambda: This is tricky: it replaces the phase 0 `get_indexed_attestation`, but it moves away from the `IndexedAttestation` in the process. Attester slashings will also have to be updated to cover the new attestation data, and call this function with the right data. Or we refactor this back, and override `IndexedAttestation` to contain custody indices, and a getter in phase 0 like `get_participants` to override in phase 1.

protolambda:
Added a doc to define the `phase 0 -> phase 1` state upgrade in, and any future instructions for the fork process.

specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,686 @@
...
+    proposer_index = get_beacon_proposer_index(state)
+
+    # Signature check
+    assert is_valid_indexed_attestation(state, get_indexed_attestation(state, attestation))
+    # Type 1: on-time attestations
+    if attestation.custody_bits != []:
+        # Correct slot
+        assert data.slot == state.slot
+        # Correct data root count
+        assert len(attestation.custody_bits) == len(get_offset_slots(state, state.shard_next_slots[shard]))
```
wemeetagain: `state.shard_next_slots` is not defined in the BeaconState definition above

protolambda:
Comments on custody spec / beaconchain shard management, after working with the experimental code and discussing with Vitalik:
- As mentioned earlier, the custody game is single-round (= immediate) now. Since we have an upper-bound of ~512 kB shard blocks. (Or 1MB or whatever the number is going to be)
  - This means completely removing the challenge/record/response/elapse stuff.
  - Merging the challenge and response to directly process custody data verification
  - New incentives necessary, since there is no more "challenger" and the slashing work is easily stolen by any proposer:
     - any validator can sign custody-slashing work to claim it is valid.
     - invalid claims get slashed
     - valid claims slash the custody bit malefactor, and reward the attestation committee
     - the committee is at least incentivised to produce signed custody slashings.
     - other actors can volunteer to sign if they care about the deflation and network health.
     - the proposer gets a small inclusion reward
  - Improvised to get custody bits for the new data. Because of the immediate processing, the old intermediate custody chunks don't help. No merkle branches either. Just compute as if it was one big custody chunk.
    - Haven't done much with the previous custody spec, so I probably made a mistake somewhere. I really just want to get this unstuck and moving, quick. We needed it yesterday to have implementers try the phase 1 spec. Review/improvements welcome, but if the scope is bigger we maybe should move it to a future PR.
    - I urge everyone to stop using the previous "challenge chunk" for subdivided custody data. Legendre bits only are constructed from a constant 48 bytes of data, which I call an "atom", since we don't subdivide it (classically...). The "chunk" is only confusing with the new chunking of shard blocks, which is different. See below.
  - This introduces a new `CustodySlashing`: provide the data to reconstruct the custody bit, along with the data to match it to an attestation, check the attestation, and identify the "malefactor": literally a validator who made a wrong custody bit.
    - The proof can be compressed more later; `attestation -> transition -> shard data` has extra unnecessary data along the way that we don't really need in the slashing.
- The work of V has a new interesting split of shard block data in the proposal:
  - Shards are tracked by shard-transitions, which in turn can contain multiple shard blocks to catch up missed slots.
  - Each of these shard blocks contains data split in chunks:
    - We do not want an mostly empty block (e.g. 23%) to cost as much space as a full one (100%). Instead, it is split in chunks to limit it to increments of `1 / MAX_SHARD_BLOCK_CHUNKS`. Current increments are `0/25/50/75/100 %`. These chunks can be concatenated and padded with zeroes to do the regular custody bit computation with.
      - Thought about `ByteList` here instead, but the goal is that all data can be concatenated for a single data availability root. And lists don't align well to subtrees to support that. "So the point is that you don't want the shard data to be a list because you want to be able to concat it all into a shared data root"
      - The CustodySlashing does not split data as intelligently yet. We can change the custody bit computation to deal with empty chunks. E.g. `void atom custody bit == 0`
      - I dislike it, as the padding is not natural. List-mixins of `ByteList` are the blocker to change things here.
     - Making shard blocks smaller and allowing a proposer to produce max. e.g. 4 consecutively in the same slot of them introduces more headers/network overhead, but making the shard block a fixed-size logistical unit sounds good. We have `ShardTransition` already to deal with multiple of them, and it would flatten this lists of lists thing (`transition -> blocks -> chunks` into `transition -> blocks`).
- Because of the multiple blocks per transition, and multiple chunks per block, the custody-bit for an attestation is complicated, and custody slashing may be even more data heavy to cover the data of block chunks.

protolambda:
Also, `shard_block_lengths` is not really enforced other than the right amount of chunks. A length-mixin as part of the shard data root could fix that.

dangerousfood:
In the method `get_shard_committee(beacon_state: BeaconState, epoch: Epoch, shard: Shard)` the parameter `shard` is not utilized within the method

protolambda:
@dangerousfood Thanks. It was pointed out in the earlier PR already though. I think it's related to the shard-rotation TODO that's left. Will look into it :+1:

test_libs/pyspec/eth2spec/test/context.py
```diff
@@ -8,6 +8,11 @@
 
 from typing import Any, Callable, Sequence
 
+from importlib import reload
+
+reload(spec_phase0)
```
djrtwo: Why `reload` here?

protolambda: A comment is probably a good idea, oops. Why: the presets are loaded into another module for the full test run. When spec phase 0 and 1 modules are initialized, they override any of the constants with these presets. And this ensures that the spec module is loaded after the correct configuration is loaded. Result: no re-initialization code in the spec code-gen, just reload and done.

specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,736 @@
...
+    aggregation_bits = indexed_attestation.attestation.aggregation_bits
+    assert len(aggregation_bits) == len(indexed_attestation.committee)
+    for i, custody_bits in enumerate(indexed_attestation.attestation.custody_bits):
+        assert len(custody_bits) == len(indexed_attestation.committee)
+        for participant, abit, cbit in zip(indexed_attestation.committee, aggregation_bits, custody_bits):
+            if abit:
+                all_pubkeys.append(state.validators[participant].pubkey)
+                # Note: only 2N distinct message hashes
+                all_message_hashes.append(hash_tree_root(
+                    AttestationCustodyBitWrapper(hash_tree_root(indexed_attestation.data), i, cbit)
```
dangerousfood: `indexed_attestation.data` does not retrieve `attestation.data` because `indexed_attestation` is encapsulating the `attestation`. I believe the correct implementation is as follows:
```
AttestationCustodyBitWrapper(hash_tree_root(indexed_attestation.attestation.data), i, cbit)

```

specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,736 @@
...
+                ))
+            else:
+                assert not cbit
+        
+    return bls_verify_multiple(
+        pubkeys=all_pubkeys,
+        message_hashes=all_message_hashes,
+        signature=indexed_attestation.signature,
+        domain=get_domain(state, DOMAIN_BEACON_ATTESTER, indexed_attestation.data.target.epoch),
```
dangerousfood: `DOMAIN_BEACON_ATTESTER` is utilized but never defined in the spec

specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,736 @@
...
+                all_message_hashes.append(hash_tree_root(
+                    AttestationCustodyBitWrapper(hash_tree_root(indexed_attestation.data), i, cbit)
+                ))
+            else:
+                assert not cbit
+        
+    return bls_verify_multiple(
+        pubkeys=all_pubkeys,
+        message_hashes=all_message_hashes,
+        signature=indexed_attestation.signature,
```
dangerousfood: Same comment as the previous one `indexed_attestation.attestation.signature` is probably the correct implementation since `indexed_attestation` is encapsulating the `attestation`

specs/core/1_beacon-chain.md
djrtwo: Removed `shard_next_slots` and added `get_next_slot_for_shard`. The value can be derived from `state.shard_transitions[shard].slot`.

specs/core/1_beacon-chain.md
djrtwo: fixed! thanks

specs/core/1_beacon-chain.md
djrtwo: thanks!

specs/core/1_beacon-chain.md
```diff
@@ -0,0 +1,791 @@
...
+    # Attesting
+    attestations: List[Attestation, MAX_ATTESTATIONS]
+    # Entry & exit
+    deposits: List[Deposit, MAX_DEPOSITS]
+    voluntary_exits: List[VoluntaryExit, MAX_VOLUNTARY_EXITS]
+    # Custody game
+    custody_slashings: List[CustodySlashing, MAX_CUSTODY_SLASHINGS]
+    custody_key_reveals: List[CustodyKeyReveal, MAX_CUSTODY_KEY_REVEALS]
+    early_derived_secret_reveals: List[EarlyDerivedSecretReveal, MAX_EARLY_DERIVED_SECRET_REVEALS]
+    # Shards
```
dangerousfood: `CustodySlashing`, `CustodyKeyReveal`, `EarlyDerivedSecretReveal` types are not currently defined within the spec.

protolambda: They are part of the custody spec document. Currently only referencing the epoch processing functions correctly. Will add references to type definitions and custody operations processing later.

djrtwo:
clsing in favor of #1504

PR #1491, opened 2019-11-22
rm signing root (fixes #1487)

### LINKED ISSUE #1487: Signing root problems
## Quick definition

- `hash_tree_root` merkleizes all fields
- `signing_root` merkleizes all fields except the last.

### Example:
```python
class BeaconBlock(Container):
    slot: Slot
    parent_root: Hash
    state_root: Hash
    body: BeaconBlockBody
    signature: BLSSignature
```

- `hash_tree_root`: `H(H(slot, parent_root), H(state_root, body))`
- `signing_root`: `H(H(H(slot, parent_root), H(state_root, body)), H(H(signature, Z[0]), Z[1]))`

## Why?

- Consensus only cares about `signing_root` on-chain in case of data types with a signature, since the signature is already verified (as part of block processing)
- `signing_root` provides the hash to sign and create `signature` with.

## Why not?

- Summaries/Expansions in SSZ: The state embeds a `block_root`, but expanding it into a block header or full block is not possible without knowing that it's a signing root.
    - Use separate type definition for the data, omitting the signature field. E.g. `some_msg_root: Root[FieldTypeWithoutSig]`
    - Mark the root as a signing root, and expand into the definition ignoring the signature field. E.g. `some_msg_root: SigRoot[FieldTypeWithSig]`, `some_signed_msg_root: HashRoot[FieldTypeWithSig]`. 
- Tree representation:
    - To represent a `BeaconBlock` as a tree you have `hash_tree_root` and `singing_root` trees.
        - State representation as a tree is useful for:
            - Caching: lazily hash child nodes `a` and `b`, cache `H(a,b)` at the tree node.
            - Fork a state into two trees with common subtrees: effectively only storing the difference to pre-state with each addition. Alike to a merkle datastore.
            - Quickly respond to proof requests by having the tree ready.
    - To navigate the tree, the expansion problem repeats: again a need to separate the two different merkle tree types.
- Inconsistency: Hash-tree-root is universal: everything has a hash-tree-root. But signing-root is not, only containers can have it. 
- Indexing: A signing-root db index is necessary for generally on-chain activity, and a hash-tree-root db index for off-chain (see if a block with specific signature is known)
    - Cannot convert the signing-root to hash-tree-root (except technically for `BeaconBlock`, but only as a by-product of current field count).

## Alternatives?

1. No `signing_root`:
    - Define `BeaconBlock` and `SignedBeaconBlock`.
        ```python
        class BeaconBlock(Container):
            slot: Slot
            parent_root: Hash
            state_root: Hash
            body: BeaconBlockBody
        
        class SignedBeaconBlock(Container):
            data: BeaconBlock
            signature: BLSSignature
        ```
        - Convenient: `hash_tree_root(SignedBeaconBlock) == H(hash_tree_root(BeaconBlock), signature)`
            - No need to provide the full branch to the signature to proof a hash-tree-root or convert a signing root into a hash-tree-root. Just provide the signature.
        - No two different type definitions, always one way to represent the type as a tree.
            - No `SigRoot`/`HashRoot` confusion or duplicate choices of representation.
        - No inconsistency with other types that are signed in different ways.
    - Optional: less repetitive definitions, add a `Signed[MsgType]` helper type to wrap any container with a signature
    - No confusion between two root types in block explorers/etc, the types have names now.
2. Type the output, work around for only some of the issues:
    - Backwards compatible, no functional change. `signing_root` stays.
    - Define `HashRoot` and `SigRoot`, both derived from `Bytes32`
    - Wherever a root needs to be expanded into a type, it can be done based on its `HashRoot` / `SigRoot`
    - Does not solve the represent-as-tree problem: `BeaconBlock` still has two different tree representations.
    - Does not solve the inconsistency problem: there will still be cases where a regular `SigRoot` does not fit due to different message / signature requirements.

## Edge cases

- `DepositData` in the spec: hash-tree-roots with signatures included are part of the deposits list root. But processing the signature is done with signing root.
    - `process_deposit`: We do not like to change the spec here: but removal of signing root doesn't require a change to the deposit contract. We just need to define a `DepositMsg` with the fields that were signed over, functionally producing the same old signing_root behavior.
- Networking: signatures may not be changed and the appearance of a block has to be checked by.
    - Things improve here with option 1: getting a root that includes the signature is as easy as hashing the signing root with the signature.

## Related

Expansions/summaries are difficult with two root variants and effectively different trees:
- https://github.com/ethereum/eth2.0-specs/issues/1436
    - The temporary "solution" is to think of it as an implementation problem. 
        - "Two options: define explicit types with/without signature yourself, or type the roots yourself"
- Light client desire for `Root[UnderlyingType]` to easily expand to `UnderlyingType` (and its further expansions). With two root types this gets difficult.
- The original introduction of "signed_root", with similar but slightly different definition, and mostly focused on unification of type definitions / line count reduction: https://github.com/ethereum/eth2.0-specs/pull/625

## Proposal

Since SSZ itself is not quite frozen, but the core functionality for Phase 0 was, so this proposal will go slower and through more review.

- Implement alternative 1 in a PR to the specs repo
- Personally expect that it will mostly be a welcome change for clients:
    - cost: defining a few simple containers, updating signing_root calls to hash_tree_root these signature-less containers instead.
    - benefit: less merkleization variation to worry about, enable caching/trees/more.
- Others involved with SSZ share their feedback / concerns / support.
- Decide next Eth2 implementers call (early Dec.) if it can go in.

protolambda:
The reasoning for the signing-root removal can be found in #1487

This is a proposal. So far there was reasonable support expressed, but more discussion/suggestions are welcome.

Changelog:
- **Introduce signed containers**:
   - DepositData already uses "data". And since we sign over a message, I think I like "msg" as general field name for the `(msg, signature)` type wrappers.
   - May generalize it to `SignedContainer[MsgType]` helper type construction. Just want to see it working first.
- **Update containers/state** to use new signed and bare container types.
  - No empty zeroed signature in the state anymore
  - Operations (and things like proposer slashings) use the new signed containers.
- **DepositData is an edge case**: I made a `DepositMsg` type to sign the contents over, replicating the exact old signature of signing root. I don't think this is much of a problem, as `DepositData` is not a message anyway, and within `Deposit` it is indeed more like data (with a proof) than a message (with a signature).
- **Improve transition function for proposers**: to verify the signature apart from the bare header contents (which don't contain a signature anymore). Blocks and headers are still transparent, but can now both be wrapped in `Signed` variants to (block/header into `msg` field, and add the `signature field`)
   - Validator API doc is improved with this now: a minimal working function to produce a valid block state root.
   - Split previous test case into an invalid-sig and zeroed sig case, just to be sure.
   - Changed the `validate_state_root` to just "`checked`": the signature signs over the state root. It makes no sense to ignore the state root, but then enforce it partially with the signature check (testing infra previously signed twice: one dummy signature for missing state root, then the real one once the state root could be computed).
     - Also note that `process_blockheader` is now so clean it could have been a `BeaconBlockHeader` as input type. But kept it as block, as scoping it to just the header data only requires the spec to convert from normal block to header. Clients can abstract/interface this nicely how they like it (or nearly keep it as is).
- **Updated validator and network docs** to account for this removal of signing root. Verbosely again to make it clear: `SignedBeaconBlock.msg` is a `BeaconBlock`, and the `hash_tree_root` of this `BeaconBlock` msg is signed over to create signature `signature`.
- **Testing infra improved**:
   - Use new signed container types
   - always clearly state "signed_block" when it's a `(msg, signature)` pair.
   - `hash_tree_root` everywhere
   - fix signature problems with fork-choice pyspec tests (Ran full test suite with BLS forced on to catch problems, and noticed these got in, since we don't output them as vectors, and CI didn't check the BLS part of these tests)
   - No more double signature hack or just-in-time randao signature creation. Clear bare data is passed around and when signed wrapped in the appropriate signed container type.
- **no `signingroot`** :tada: 

Note: left old Phase 1 use of signing root as-is, it's changing anyway with the new PR (see #1483 )

We may want to squash this PR when merging. Or I can rebase and split the test work from the spec changes.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -33,20 +33,21 @@
             - [`DepositMessage`](#depositmessage)
             - [`DepositData`](#depositdata)
             - [`BeaconBlockHeader`](#beaconblockheader)
-            - [`SignedBeaconBlockHeader`](#signedbeaconblockheader)
         - [Beacon operations](#beacon-operations)
             - [`ProposerSlashing`](#proposerslashing)
             - [`AttesterSlashing`](#attesterslashing)
             - [`Attestation`](#attestation)
             - [`Deposit`](#deposit)
             - [`VoluntaryExit`](#voluntaryexit)
-            - [`SignedVoluntaryExit`](#signedvoluntaryexit)
         - [Beacon blocks](#beacon-blocks)
             - [`BeaconBlockBody`](#beaconblockbody)
             - [`BeaconBlock`](#beaconblock)
-            - [`SignedBeaconBlock`](#signedbeaconblock)
         - [Beacon state](#beacon-state)
             - [`BeaconState`](#beaconstate)
+        - [Signed envelopes](#signed-envelopes)
+            - [`SignedVoluntaryExit`](#signedvoluntaryexit)
+            - [`SignedBeaconBlock`](#signedbeaconblock)
+            - [`SignedBeaconBlockHeader`](#signedbeaconblockheader)
     - [Helper functions](#helper-functions)
         - [Math](#math)
             - [`integer_squareroot`](#integer_squareroot)
@@ -374,14 +375,6 @@ class BeaconBlockHeader(Container):
     body_root: Root
 ```
 
-#### `SignedBeaconBlockHeader`
-
-```python
-class SignedBeaconBlockHeader(Container):
-    message: BeaconBlockHeader
-    signature: BLSSignature
-```
-
 ### Beacon operations
 
 #### `ProposerSlashing`
@@ -426,14 +419,6 @@ class VoluntaryExit(Container):
     validator_index: ValidatorIndex
 ```
 
-#### `SignedVoluntaryExit`
-
-```python
-class SignedVoluntaryExit(Container):
-    message: VoluntaryExit
-    signature: BLSSignature
-```
-
 ### Beacon blocks
 
 #### `BeaconBlockBody`
@@ -459,14 +444,6 @@ class BeaconBlock(Container):
     parent_root: Root
     state_root: Root
     body: BeaconBlockBody
-```
-
-#### `SignedBeaconBlock`
-
-```python
-class SignedBeaconBlock(Container):
-    message: BeaconBlock
-    signature: BLSSignature
 ```
 
 ### Beacon state
@@ -503,6 +480,34 @@ class BeaconState(Container):
     previous_justified_checkpoint: Checkpoint  # Previous epoch snapshot
     current_justified_checkpoint: Checkpoint
     finalized_checkpoint: Checkpoint
+```
+
+### Signed envelopes
+
+Some messages in the protocol are wrapped in an envelop to better facilitate adding/pruning the signature and to `hash_tree_root` the `message` separate from the signature.  
+
+#### `SignedVoluntaryExit`
+
+```python
+class SignedVoluntaryExit(Container):
+    message: VoluntaryExit
+    signature: BLSSignature
+```
+
+#### `SignedBeaconBlock`
+
+```python
+class SignedBeaconBlock(Container):
+    message: BeaconBlock
+    signature: BLSSignature
+```
+
+#### `SignedBeaconBlockHeader`
+
+```python
+class SignedBeaconBlockHeader(Container):
+    message: BeaconBlockHeader
+    signature: BLSSignature
 ```
 
 ## Helper functions
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -30,17 +30,21 @@
             - [`PendingAttestation`](#pendingattestation)
             - [`Eth1Data`](#eth1data)
             - [`HistoricalBatch`](#historicalbatch)
+            - [`DepositMessage`](#depositmessage)
             - [`DepositData`](#depositdata)
             - [`BeaconBlockHeader`](#beaconblockheader)
+            - [`SignedBeaconBlockHeader`](#signedbeaconblockheader)
         - [Beacon operations](#beacon-operations)
             - [`ProposerSlashing`](#proposerslashing)
             - [`AttesterSlashing`](#attesterslashing)
             - [`Attestation`](#attestation)
             - [`Deposit`](#deposit)
             - [`VoluntaryExit`](#voluntaryexit)
+            - [`SignedVoluntaryExit`](#signedvoluntaryexit)
         - [Beacon blocks](#beacon-blocks)
             - [`BeaconBlockBody`](#beaconblockbody)
             - [`BeaconBlock`](#beaconblock)
+            - [`SignedBeaconBlock`](#signedbeaconblock)
         - [Beacon state](#beacon-state)
             - [`BeaconState`](#beaconstate)
     - [Helper functions](#helper-functions)
@@ -52,7 +56,6 @@
         - [Crypto](#crypto)
             - [`hash`](#hash)
             - [`hash_tree_root`](#hash_tree_root)
-            - [`signing_root`](#signing_root)
             - [`bls_verify`](#bls_verify)
             - [`bls_aggregate_pubkeys`](#bls_aggregate_pubkeys)
         - [Predicates](#predicates)
@@ -342,6 +345,15 @@ class HistoricalBatch(Container):
     state_roots: Vector[Root, SLOTS_PER_HISTORICAL_ROOT]
 ```
 
+#### `DepositMessage`
+
+```python
+class DepositMessage(Container):
+    pubkey: BLSPubkey
+    withdrawal_credentials: Bytes32
+    amount: Gwei
+```
+
 #### `DepositData`
 
 ```python
@@ -349,7 +361,7 @@ class DepositData(Container):
     pubkey: BLSPubkey
     withdrawal_credentials: Bytes32
     amount: Gwei
-    signature: BLSSignature
+    signature: BLSSignature  # signing over DepositMessage
 ```
 
 #### `BeaconBlockHeader`
@@ -360,6 +372,13 @@ class BeaconBlockHeader(Container):
     parent_root: Root
     state_root: Root
     body_root: Root
+```
+
+#### `SignedBeaconBlockHeader`
+
+```python
+class SignedBeaconBlockHeader(Container):
+    message: BeaconBlockHeader
     signature: BLSSignature
 ```
 
@@ -370,8 +389,8 @@ class BeaconBlockHeader(Container):
 ```python
 class ProposerSlashing(Container):
     proposer_index: ValidatorIndex
-    header_1: BeaconBlockHeader
-    header_2: BeaconBlockHeader
+    signed_header_1: SignedBeaconBlockHeader
+    signed_header_2: SignedBeaconBlockHeader
 ```
 
 #### `AttesterSlashing`
@@ -405,6 +424,13 @@ class Deposit(Container):
 class VoluntaryExit(Container):
     epoch: Epoch  # Earliest epoch when voluntary exit can be processed
     validator_index: ValidatorIndex
+```
+
+#### `SignedVoluntaryExit`
+
+```python
+class SignedVoluntaryExit(Container):
+    message: VoluntaryExit
     signature: BLSSignature
 ```
 
@@ -422,7 +448,7 @@ class BeaconBlockBody(Container):
     attester_slashings: List[AttesterSlashing, MAX_ATTESTER_SLASHINGS]
     attestations: List[Attestation, MAX_ATTESTATIONS]
     deposits: List[Deposit, MAX_DEPOSITS]
-    voluntary_exits: List[VoluntaryExit, MAX_VOLUNTARY_EXITS]
+    voluntary_exits: List[SignedVoluntaryExit, MAX_VOLUNTARY_EXITS]
 ```
 
 #### `BeaconBlock`
@@ -433,6 +459,13 @@ class BeaconBlock(Container):
     parent_root: Root
     state_root: Root
     body: BeaconBlockBody
+```
+
+#### `SignedBeaconBlock`
+
+```python
+class SignedBeaconBlock(Container):
+    message: BeaconBlock
     signature: BLSSignature
 ```
 
@@ -532,10 +565,6 @@ def bytes_to_int(data: bytes) -> uint64:
 #### `hash_tree_root`
 
 `def hash_tree_root(object: SSZSerializable) -> Root` is a function for hashing objects into a single root by utilizing a hash tree structure, as defined in the [SSZ spec](../simple-serialize.md#merkleization).
-
-#### `signing_root`
-
-`def signing_root(object: Container) -> Root` is a function for computing signing messages, as defined in the [SSZ spec](../simple-serialize.md#self-signed-containers).
 
 #### `bls_verify`
 
@@ -1048,16 +1077,25 @@ Let `genesis_block = BeaconBlock(state_root=hash_tree_root(genesis_state))`.
 The post-state corresponding to a pre-state `state` and a block `block` is defined as `state_transition(state, block)`. State transitions that trigger an unhandled exception (e.g. a failed `assert` or an out-of-range list access) are considered invalid.
 
 ```python
-def state_transition(state: BeaconState, block: BeaconBlock, validate_state_root: bool=False) -> BeaconState:
+def state_transition(state: BeaconState, signed_block: SignedBeaconBlock, validate_result: bool=True) -> BeaconState:
     # Process slots (including those with no blocks) since block
-    process_slots(state, block.slot)
+    process_slots(state, signed_block.message.slot)
+    # Verify signature
+    if validate_result:
+        assert verify_block_signature(state, signed_block)
     # Process block
-    process_block(state, block)
-    # Validate state root (`validate_state_root == True` in production)
-    if validate_state_root:
-        assert block.state_root == hash_tree_root(state)
+    process_block(state, signed_block.message)
+    if validate_result:
+        assert signed_block.message.state_root == hash_tree_root(state)  # Validate state root
     # Return post-state
     return state
+```
+
+```python
+def verify_block_signature(state: BeaconState, signed_block: SignedBeaconBlock) -> bool:
+    proposer = state.validators[get_beacon_proposer_index(state)]
+    domain = get_domain(state, DOMAIN_BEACON_PROPOSER)
+    return bls_verify(proposer.pubkey, hash_tree_root(signed_block.message), signed_block.signature, domain)
 ```
 
 ```python
@@ -1080,7 +1118,7 @@ def process_slot(state: BeaconState) -> None:
     if state.latest_block_header.state_root == Bytes32():
         state.latest_block_header.state_root = previous_state_root
     # Cache block root
-    previous_block_root = signing_root(state.latest_block_header)
+    previous_block_root = hash_tree_root(state.latest_block_header)
     state.block_roots[state.slot % SLOTS_PER_HISTORICAL_ROOT] = previous_block_root
 ```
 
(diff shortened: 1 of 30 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1048,18 +1077,27 @@ Let `genesis_block = BeaconBlock(state_root=hash_tree_root(genesis_state))`.
 The post-state corresponding to a pre-state `state` and a block `block` is defined as `state_transition(state, block)`. State transitions that trigger an unhandled exception (e.g. a failed `assert` or an out-of-range list access) are considered invalid.
 
 ```python
-def state_transition(state: BeaconState, block: BeaconBlock, validate_state_root: bool=False) -> BeaconState:
+def state_transition(state: BeaconState, signed_block: SignedBeaconBlock, checked: bool=True) -> BeaconState:
```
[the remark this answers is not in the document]
protolambda: The signature check is dependent on the expected state root (part of the block, thus signed over). This is why it doesn't make much sense to have a way to disable the state-root check, like previously, without disabling the signature.
If we want to be more explicit, I think it makes more sense to call it `validate_result` or something like that. Naming things is difficult :sweat_smile:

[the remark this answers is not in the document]
hwwhww: Aah sorry I misread it!
IMO `checked` sounds like it has been checked *before* this function and `validate_result` is more accurate. :)

FYI there is no `checked=False` in this branch (including tests), but I think it's fine to keep it in case we end up needing it in the future...?

specs/core/0_fork-choice.md
```diff
@@ -32,7 +32,7 @@ This document is the beacon chain fork choice spec, part of Ethereum 2.0 Phase 0
 The head block root associated with a `store` is defined as `get_head(store)`. At genesis, let `store = get_genesis_store(genesis_state)` and update `store` by running:
 
 - `on_tick(time)` whenever `time > store.time` where `time` is the current Unix time
-- `on_block(block)` whenever a block `block` is received
+- `on_block(block)` whenever a block `block` (`SignedBeaconBlock`) is received
```
hwwhww: What do you can use the notation like this?
```suggestion
- `on_block(block)` whenever a block `block: SignedBeaconBlock` is received
```

specs/validator/0_beacon-chain-validator.md
```diff
@@ -111,7 +111,7 @@ To submit a deposit:
 - Pack the validator's [initialization parameters](#initialization) into `deposit_data`, a [`DepositData`](../core/0_beacon-chain.md#depositdata) SSZ object.
 - Let `amount` be the amount in Gwei to be deposited by the validator where `amount >= MIN_DEPOSIT_AMOUNT`.
 - Set `deposit_data.amount = amount`.
-- Let `signature` be the result of `bls_sign` of the `signing_root(deposit_data)` with `domain=compute_domain(DOMAIN_DEPOSIT)`. (Deposits are valid regardless of fork version, `compute_domain` will default to zeroes there).
+- Let `signature` be the result of `bls_sign` of the `hash_tree_root(DepositMsg())` with `domain=compute_domain(DOMAIN_DEPOSIT)`. (Deposits are valid regardless of fork version, `compute_domain` will default to zeroes there).
```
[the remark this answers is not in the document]
djrtwo: Can do a "Let `deposit_msg` be..." as a separate line above

protolambda:
@hwwhww renamed `checked` to `validate_result`. Two things to consider here:
- I think it should absolutely be True by default, even if overruled elsewhere. Verification of the block signature and state root are probably the two most important things to eth2 phase0...
- Added an unsigned-transition method in the testing machinery. Which is a clean call to `process_slots`, `process_blocks`, and done. Same thing in the validator spec for proposers to build their block. But for spec purposes, I think it's useful to highlight the option as part of the beacon chain spec itself, or alternatively we link to the validator spec which describes the transition for a proposer building its block.

specs/validator/0_beacon-chain-validator.md
```diff
@@ -214,13 +214,21 @@ Set `block.slot = slot` where `slot` is the current slot at which the validator
...
 
 Set `block.state_root = hash_tree_root(state)` of the resulting `state` of the `parent -> block` state transition.
 
-*Note*: To calculate `state_root`, the validator should first run the state transition function on an unsigned `block` containing a stub for the `state_root`. It is useful to be able to run a state transition function that does _not_ validate signatures or state root for this purpose.
+*Note*: To calculate `state_root`, the validator should first run the state transition function on an unsigned `block` containing a stub for the `state_root`.
+It is useful to be able to run a state transition function that does _not_ validate signatures or state root for this purpose:
+
+```python
+def compute_new_state_root(state: BeaconState, block: BeaconBlock) -> Root:
+    process_slots(state, block.slot)
```
djrtwo: shouldn't we instead use `state_transition` with the `validate_results` set to false here?

Also kind of sucks that these functions all mutate `state`, but that's not changing

specs/validator/0_beacon-chain-validator.md
```diff
@@ -262,12 +270,12 @@ def get_eth1_vote(state: BeaconState, previous_eth1_distance: uint64) -> Eth1Dat
 
 ##### Signature
 
-Set `header.signature = block_signature` where `block_signature` is obtained from:
+Set `signed_header.signature = block_signature` where `block_signature` is obtained from:
```
djrtwo: `header` or `signed_header` isn't previously defined, nor is it used subsequently. A valid signature here also depends on already creating the block body.

I suggest that we remove `Signature` section ehre and create a `Signed Beacon Block` section after `Block Body`. This section will construct the wrapper container and add in the signature. It will also include a note about how this is the message that is broadcast to the network

djrtwo: (This issue was generally here before you made edits. We should fix it none the less)

test_libs/pyspec/eth2spec/test/helpers/block.py
```diff
@@ -33,26 +37,45 @@ def sign_block(spec, state, block, proposer_index=None):
...
             spec.compute_epoch_at_slot(block.slot)))
 
 
+def sign_block(spec, state, block, proposer_index=None):
+    signed_block = spec.SignedBeaconBlock(msg=block)
+    apply_sig(spec, state, signed_block, proposer_index)
+    return signed_block
+
+
+def transition_unsigned_block(spec, state, block):
```
djrtwo: why this instead of using `state_trnasition` with the extra checks set to `False`?

specs/core/0_beacon-chain.md
```diff
@@ -1048,18 +1077,27 @@ Let `genesis_block = BeaconBlock(state_root=hash_tree_root(genesis_state))`.
...
     # Process slots (including those with no blocks) since block
-    process_slots(state, block.slot)
+    process_slots(state, signed_block.msg.slot)
+    # Verify signature
+    if checked:
```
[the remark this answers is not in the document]
protolambda: Keeping it as is now. Technically it doesn't change behavior. Invalid sig will be an invalid sig. And processed after slots, so proposer index data is available to check. (Declaring it in the block additionally is for another issue, if ever).

specs/validator/0_beacon-chain-validator.md
protolambda: `state_transition` uses a signed block as input, which makes no sense for a proposer that is in the process of producing it. This is the kind of thing Nimbus is struggling with, empty/placeholder signatures are bad. Alternatively we can make `state_transition` go on the unsinged block, and make signature verification with proposer a separate thing, but prefer not to go that far. Regardless, I think this `compute_new_state_root` is so minimal (just call two functions cleanly, without any placeholder info or transformed data), that it is a good design to promote as spec.

test_libs/pyspec/eth2spec/test/helpers/block.py
protolambda: Different input type. Just the bare `BeaconBlock`, no signature placeholder, and thus explicitly unsigned.

specs/core/0_beacon-chain.md
```diff
@@ -360,6 +372,13 @@ class BeaconBlockHeader(Container):
     parent_root: Root
     state_root: Root
     body_root: Root
+```
+
+#### `SignedBeaconBlockHeader`
+
+```python
+class SignedBeaconBlockHeader(Container):
+    msg: BeaconBlockHeader
```
JustinDrake: `msg` => `message` (the spec doesn't otherwise use abbreviations)

specs/core/0_beacon-chain.md
```diff
@@ -342,14 +345,23 @@ class HistoricalBatch(Container):
     state_roots: Vector[Root, SLOTS_PER_HISTORICAL_ROOT]
 ```
 
+#### `DepositMsg`
+
+```python
+class DepositMsg(Container):
```
JustinDrake: `DepositMsg` => `DepositMessage`

JustinDrake: Should it be `DepositData` and `SignedDepositData`?

specs/core/0_beacon-chain.md
JustinDrake: On second thought, maybe we want `signed_header.header`.

protolambda: I like `msg` because 9/10 times the field is used for `bls_verify(pubkey, msg, signature, domain)` and you have a standard name that is explicitly about the inner container, not confusing with the signed header. `message` works too and is more consistent indeed.

specs/core/0_beacon-chain.md
protolambda: Didn't want to change existing names too much. Note that genesis code already merkleizes deposit-data including signatures, and there are a lot of clients who call it "deposit_data" all over the place, to work-around collisions with "deposit". Calling the inner type the message condenses the amount of changes to the bare minimum: the signature step briefly uses the type, everything else stays as-is.

This brings me to another part of the reason: that `Data` already has meaning in respect to the `Deposit` that adds a merkle proof for the data. I like to keep names with only a single add-on like that to describe what is happening.
The more extreme option would be to do: `MerkledSignedDeposit`, `SignedDeposit`, `Deposit`. But it's probably way too late for that, and maybe too verbose.

And note that with blocks etc. I think it is better the other way around: the signature is a thing to be verified as one of the very first things when processing a block. After that it can be forgotten and ignored. All the existing code that is defined for a `BeaconBlock` type can stay as-is, allowing to process unsigned blocks for testing etc. (also hear complaints from Nimbus who previously had problems with empty/invalid placeholder signatures). And `SignedBeaconBlock` is only really used at the outer signature-checking layer, minimizing the amount of changes.

specs/core/0_beacon-chain.md
JustinDrake: > `message` works too and is more consistent indeed

Please make it `message` :)

specs/core/0_beacon-chain.md
djrtwo: Agreed that using the `Object`/`SignedObject` pattern in naming for Deposit should be avoided because it doesn't actually follow the pattern.

specs/core/0_beacon-chain.md
```diff
@@ -360,6 +372,13 @@ class BeaconBlockHeader(Container):
     parent_root: Root
     state_root: Root
     body_root: Root
+```
+
+#### `SignedBeaconBlockHeader`
```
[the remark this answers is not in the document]
djrtwo: bump, still prefer this method. The spec building fixes topological ordering so that's fine

protolambda:
Changed `msg -> message` and `DepositMsg -> DepositMessage`, and squashed the PR commits.

specs/core/0_beacon-chain.md
protolambda: Done

specs/core/0_beacon-chain.md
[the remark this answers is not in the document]
protolambda: See afb9a1d, you can change the wording if you like.

arnetheduck:
LGTM - generally, naming has been brought up as a potential area of improvement  and I'd generally agree that it would be good to do a holistic rename round once something like this is merged so that there is structure to the naming (`DepositMessage` for example looks a bit odd)

specs/core/0_beacon-chain.md
```diff
@@ -342,14 +346,23 @@ class HistoricalBatch(Container):
...
+
 #### `DepositData`
 
 ```python
 class DepositData(Container):
     pubkey: BLSPubkey
     withdrawal_credentials: Bytes32
     amount: Gwei
-    signature: BLSSignature
+    signature: BLSSignature  # signing over DepositMessage
```
prestonvanloon: Can you clarify this comment? What does `signing over DepositMessage` mean?

specs/core/0_beacon-chain.md
```diff
@@ -1340,20 +1383,17 @@ def process_block_header(state: BeaconState, block: BeaconBlock) -> None:
...
         parent_root=block.parent_root,
         # `state_root` is zeroed and overwritten in the next `process_slot` call
         body_root=hash_tree_root(block.body),
-        # `signature` is zeroed
     )
     # Verify proposer is not slashed
     proposer = state.validators[get_beacon_proposer_index(state)]
     assert not proposer.slashed
-    # Verify proposer signature
-    assert bls_verify(proposer.pubkey, signing_root(block), block.signature, get_domain(state, DOMAIN_BEACON_PROPOSER))
```
prestonvanloon: Just wondering, why is this assertion removed?

protolambda: The signature check is moved to a separate function, `verify_block_signature`, to separate the actual header processing from the signature check that clients would perform in their own optimal way. Now that we have the clear distinction between signed and unsigned header data, I think this is cleaner. And functionally it's still the same.

prestonvanloon: Sounds good. Just wanted to ask as a sanity check here since the diff is hard to follow. Thanks for the response.

specs/core/0_beacon-chain.md
protolambda: `DepositData` is a special case:
- The signature is included as part of the data in the deposit-datas hash-tree-root. And we can not completely separate it from the message as with the other containers, since that would not be compatible with the existing verified deposit contract code. The contract expects `H(H(p,w), H(a,s))`, not `H(H(H(p,w),H(a,0)), s)` (first characters of field names).
- The signature is not included when creating the message hash to sign (because the signature does not exist yet at that point). That is what `DepositMessage` is for: to replicate the old signing-root behavior of hash-tree-root over all fields except the signature.
Also see `Deposit`(merkle proof + signature + contents)/`DepositData`(signature + contents)/`DepositMessage`(contents) naming discussion above in this PR. 

It's not pretty, but deposits are not broken this way, and we can have all the benefits as outlined in #1487 to make up for it.

specs/core/0_beacon-chain.md
```diff
@@ -1048,18 +1077,27 @@ Let `genesis_block = BeaconBlock(state_root=hash_tree_root(genesis_state))`.
 The post-state corresponding to a pre-state `state` and a block `block` is defined as `state_transition(state, block)`. State transitions that trigger an unhandled exception (e.g. a failed `assert` or an out-of-range list access) are considered invalid.
 
 ```python
-def state_transition(state: BeaconState, block: BeaconBlock, validate_state_root: bool=False) -> BeaconState:
+def state_transition(state: BeaconState, signed_block: SignedBeaconBlock, validate_result: bool=True) -> BeaconState:
     # Process slots (including those with no blocks) since block
-    process_slots(state, block.slot)
+    process_slots(state, signed_block.msg.slot)
+    # Verify signature
+    if validate_result:
```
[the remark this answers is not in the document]
protolambda: I think it's better to not process the full block until having checked the signature, more alike to the real world processing, which clients are looking at implementing.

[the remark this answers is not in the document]
djrtwo: My counter to that is that in a real client, they would have already checked the signature when getting the block from the wire and would probably call `state_transition` in a way that bypassed the signature verification once ready to do the full transition.

Probably a moot point either way. resolving

PR #1494, opened 2019-11-25
Churn limit simplification

### LINKED ISSUE #1493: possible get_validator_churn_limit() issue
For v9.2 of the phase 0 beacon chain, I am assuming that we calculate the max number of validators as follows: 

`MAX_VALIDATORS_PER_COMMITTEE` * `MAX_COMMITTEES_PER_SLOT` = 2^11 * 2^6 = 2^17 

if so, then this function:

```python
def get_validator_churn_limit(state: BeaconState) -> uint64:
    """
    Return the validator churn limit for the current epoch.
    """
    active_validator_indices = get_active_validator_indices(state, get_current_epoch(state))
    return max(MIN_PER_EPOCH_CHURN_LIMIT, len(active_validator_indices) // CHURN_LIMIT_QUOTIENT)
```
will always be equal to `MIN_PER_EPOCH_CHURN_LIMIT`.

jrhea:
Removed `get_validator_churn_limit()` and `CHURN_LIMIT_QUOTIENT` in favor of `PER_EPOCH_CHURN_LIMIT`

resolves #1493


djrtwo:
This missed that the total number of validators is derived from `MAX_VALIDATORS_PER_COMMITTEE` * `MAX_COMMITTEES_PER_SLOT` * `SLOTS_PER_EPOCH` rather than just `MAX_VALIDATORS_PER_COMMITTEE` * `MAX_COMMITTEES_PER_SLOT`

PR #1495, opened 2019-11-25
filter viable branches in fork choice

djrtwo:
Filter for branches that see latest finalized/justified info as correct _on-chain_ before running `get_head`.

more info here https://notes.ethereum.org/Fj-gVkOSTpOyUx-zkWjuwg?view


specs/core/0_fork-choice.md
```diff
@@ -136,17 +136,69 @@ def get_latest_attesting_balance(store: Store, root: Root) -> Gwei:
...
+    block = store.blocks[block_root]
+    children = [
+        root for root in store.blocks.keys()
+        if store.blocks[root].parent_root == block_root
+    ]
+
+    # If any children branches contain expected finalized/justified checkpoints,
+    # add to filtered block-tree and signal viability to parent.
+    if any(children):
+        if True in [filter_block_tree(store, child, blocks) for child in children]:
```
protolambda:
```suggestion
        if any(filter_block_tree(store, child, blocks) for child in children):
```

specs/core/0_fork-choice.md
```diff
@@ -136,17 +136,69 @@ def get_latest_attesting_balance(store: Store, root: Root) -> Gwei:
...
+
+    # If leaf block, check finalized/justified checkpoints as matching latest.
+    # If matching, add to viable block-tree and signal viability to parent.
+    head_state = store.block_states[block_root]
+
+    # Handle base case where justified hasn't updated yet
+    if head_state.current_justified_checkpoint.epoch == GENESIS_EPOCH:
+        is_viable_branch = True
+    else:
+        is_viable_branch = (
```
protolambda: All this forkchoice PR really does is to limit the tree to blocks that have a finalized checkpoint equal to that of the store? So branches that have a later finalized epoch state, but still descendant of the finalized epoch state in the store, are not considered as head, and their complete branch is ignored? And thus it fails to sync new valid finalized branches? Or is the store finalized checkpoint assumed to always be as up to date with store states as can be? Can confirm that 130 and 144 in the example sketch you sent won't be part of the subtree though in the example as-is.

can we make this check if `head_state.finalized_checkpoint` is equal to, or descendant of, store.finalized_checkpoint instead? And same thing for justified?

djrtwo: Because we (potentially) update latest justified and finalized when adding any block to the block tree via `on_block`, the `store.finalized_checkpoint`/`store.justified_checkpoint` is always the highest possible value found in the block tree.

So I'm making the claim that there should never be a branch with a higher latest finalized/justified than that tracked in the store. If this is _not_ the case, then we would want to change it to be as you described.

specs/core/0_fork-choice.md
djrtwo: Tricky. This actually breaks things. If you pass an iterator into `any`, then the iterator stops running upon first item that is `True`. Because I am relying upon the side effect of `filter_block_tree` run on all children, then this can truncate valid branches

Example from python console
```
>>> def test(i):
...     print(i)
...     return i > 4
... 
>>> any(test(i) for i in range(10))
0
1
2
3
4
5
True
>>> 
```

djrtwo: But if you put the iterator inside a list def, it evaluates all

```
>>> any([test(i) for i in range(10)])
0
1
2
3
4
5
6
7
8
9
True
```

protolambda: Well, it's probably better to be more explicit about that side-effect then. Implementers looking to optimize things even more than this, and are likely to make the same mistake.

protolambda:
Do we merge rm-signing-root first? If so, we should change the signing-roots here to hash-tree-root.

hwwhww:
@djrtwo I think you should change the target branch to `v09x`?

specs/core/0_fork-choice.md
hwwhww: To be more explicit, maybe split into two lines?

```suggestion
    filter_block_tree_result = [filter_block_tree(store, child, blocks) for child in children]
    if any(filter_block_tree_result):
```

djrtwo:
yep nice catch @hwwhww !

PR #1499, opened 2019-12-02
BLS Specs rewrite for IETF clarity

CarlBeek:
This PR is an alternate proposal to @kirk-baird's #1398.

It is based on two ideas:

0. If something is defined in the IETF specs, it should not be redefined in the eth2-specs. This reduces the scope for errors and reflects the workflow of integrating with an IETF spec-compliant BLS implementation.

1. `tag` replaces `domain`. Confusion can arise due to the name collision of `domain` in the eth2 specifications and those of the IETF. `tag` is an alternative that implies also separation and is unique to the eth2-specs.

PS: The diff appears large here, but most of that is due to `tag` replacing `domain`. The substantive changes are (almost entirely) limited to `specs/bls_signature.md`


specs/bls_signature.md
```diff
@@ -1,146 +1,96 @@
...
-    candidate_squareroot = value ** ((Fq2_order + 8) // 16)
-    check = candidate_squareroot ** 2 / value
-    if check in eighth_roots_of_unity[::2]:
-        x1 = candidate_squareroot / eighth_roots_of_unity[eighth_roots_of_unity.index(check) // 2]
-        x2 = -x1
-        x1_re, x1_im = coerce_to_int(x1.coeffs[0]), coerce_to_int(x1.coeffs[1])
-        x2_re, x2_im = coerce_to_int(x2.coeffs[0]), coerce_to_int(x2.coeffs[1])
-        return x1 if (x1_im > x2_im or (x1_im == x2_im and x1_re > x2_re)) else x2
-    return None
+def bls_sign(privkey: int, message_hash: Hash, tag: Tag) -> BLSSignature:
```
djrtwo: We removed `Hash` to disambiguate from the function `hash()`. `Bytes32` is preferred here

```suggestion
def bls_sign(privkey: int, message_hash: Bytes32, tag: Tag) -> BLSSignature:
```

specs/bls_signature.md
```diff
@@ -1,146 +1,96 @@
...
+```python
+Sign(SK: int, message: Bytes) -> BLSSignature
+```
 
-### G2 points
+### `Aggregate`
 
-A point in G2 is represented as a pair of 384-bit integers `(z1, z2)`. We decompose `z1` as above into `x1`, `a_flag1`, `b_flag1`, `c_flag1` and `z2` into `x2`, `a_flag2`, `b_flag2`, `c_flag2`.
+```python
+Aggregate(signature_1: BLSSignature, ..., signature_n:BLSSignature) -> BLSSignature
```
djrtwo: missing a space

```suggestion
Aggregate(signature_1: BLSSignature, ..., signature_n: BLSSignature) -> BLSSignature
```

specs/bls_signature.md
```diff
@@ -1,146 +1,96 @@
...
 
 ### `bls_verify`
 
-Let `bls_verify(pubkey: Bytes48, message_hash: Bytes32, signature: Bytes96, domain: Bytes8) -> bool`:
-
-* Verify that `pubkey` is a valid G1 point.
-* Verify that `signature` is a valid G2 point.
-* Verify that `e(pubkey, hash_to_G2(message_hash, domain)) == e(g, signature)`.
+```python
+def bls_verify(pubkey: BLSPubkey, message_hash: Hash, signature: BLSSignature, tag: Tag) -> bool:
```
djrtwo:
```suggestion
def bls_verify(pubkey: BLSPubkey, message_hash: Bytes32, signature: BLSSignature, tag: Tag) -> bool:
```

specs/bls_signature.md
```diff
@@ -1,146 +1,96 @@
...
 ### `bls_verify_multiple`
 
-Let `bls_verify_multiple(pubkeys: List[Bytes48], message_hashes: List[Bytes32], signature: Bytes96, domain: Bytes8) -> bool`:
-
-* Verify that each `pubkey` in `pubkeys` is a valid G1 point.
-* Verify that `signature` is a valid G2 point.
-* Verify that `len(pubkeys)` equals `len(message_hashes)` and denote the length `L`.
-* Verify that `e(pubkeys[0], hash_to_G2(message_hashes[0], domain)) * ... * e(pubkeys[L-1], hash_to_G2(message_hashes[L-1], domain)) == e(g, signature)`.
+```python
+def bls_verify_multiple(pubkeys: List[BLSPubkey],  message_hash: Hash, signature: BLSSignature, tag: Tag) -> bool:
```
djrtwo:
```suggestion
def bls_verify_multiple(pubkeys: List[BLSPubkey], message_hash: Hash, signature: BLSSignature, tag: Tag) -> bool:
```

specs/bls_signature.md
```diff
@@ -1,146 +1,96 @@
...
+Eth2 treats all the functions defined by [`draft-irtf-cfrg-bls-signature-00`](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-00) as black boxes with the following function interfaces:
 
-* `x = z % 2**381`
-* `a_flag = (z % 2**382) // 2**381`
-* `b_flag = (z % 2**383) // 2**382`
-* `c_flag = (z % 2**384) // 2**383`
+### `VALID` and `INVALID`
 
-Respecting bit ordering, `z` is decomposed as `(c_flag, b_flag, a_flag, x)`.
+The values `VALID` and `INVALID` used by the  IETF BLS specification are mapped to `True` and `False` respectively. This is done for notational convenience here, but may not be the case for all BLS implementations.
```
djrtwo: remove extra space

```suggestion
The values `VALID` and `INVALID` used by the IETF BLS specification are mapped to `True` and `False` respectively. This is done for notational convenience here, but may not be the case for all BLS implementations.
```

specs/bls_signature.md
```diff
@@ -1,146 +1,96 @@
...
-    return None
+def bls_sign(privkey: int, message_hash: Hash, tag: Tag) -> BLSSignature:
+    return Sign(privkey, message_hash + tag)
 ```
 
-## Aggregation operations
-
-### `bls_aggregate_pubkeys`
-
-Let `bls_aggregate_pubkeys(pubkeys: List[Bytes48]) -> Bytes48` return `pubkeys[0] + .... + pubkeys[len(pubkeys)-1]`, where `+` is the elliptic curve addition operation over the G1 curve. (When `len(pubkeys) == 0` the empty sum is the G1 point at infinity.)
```
djrtwo: Is there no public interface in the BLS standard to aggregate pubkeys?
I can't immediately think of a need to aggregate it independently and have the resulting aggregate pubkey, but interesting none the less.

specs/core/0_beacon-chain.md
```diff
@@ -239,17 +239,17 @@ The following values are (non-configurable) constants used throughout the specif
 | `MAX_DEPOSITS` | `2**4` (= 16) |
 | `MAX_VOLUNTARY_EXITS` | `2**4` (= 16) |
 
-### Domain types
+### Tag types
```
djrtwo: Maybe we should call these "Signature Tags" or something a little bit more descriptive in the header.

If you want to leave the name as-is, then let's at least put a few words in this section to describe what we use them for.

djrtwo:
Oh shit, I just noticed the merge conflict here. I'll have a stab at it

specs/core/0_beacon-chain.md
```diff
@@ -239,17 +239,17 @@ The following values are (non-configurable) constants used throughout the specif
...
+The following types are defined, mapping into `TagType` (little endian):
 
 | Name | Value |
 | - | - |
-| `DOMAIN_BEACON_PROPOSER` | `0` |
-| `DOMAIN_BEACON_ATTESTER` | `1` |
-| `DOMAIN_RANDAO` | `2` |
-| `DOMAIN_DEPOSIT` | `3` |
-| `DOMAIN_VOLUNTARY_EXIT` | `4` |
+| `TAG_BEACON_PROPOSER` | `0` |
```
MrChico: I don't like the inconsistency in how these are presented here as numbers "mapping into bytes" as opposed to their presentation as literal bytestrings in the config files. 

Is there any reason to not display the literal bytestrings here as well?

djrtwo:
Merge conflicts resolved

kirk-baird:
I like the way you've presented it here and I agree that it's much better to not redefine things that are already defined in the IETFs. So I too think this is the way forward, I'm happy to close #1398 when this is merged :)

JustinDrake:
A few issues (discussed with Carl privately):

1) We need support for uniqueness roots to unify the slashing conditions. See https://github.com/ethereum/eth2.0-specs/issues/1387
2) Application-layer logic specific to Eth2 (namely, the `tag` as well as the uniqueness root) should not go in the BLS spec. Instead, it should go in the "core" Eth2 spec.
3) We should sign over 32-byte messages (hash outputs) as apposed to appending the tag with `+ tag`. The appendage is adhoc and not future proof (for example, does work well with uniqueness roots). The appendage is also ugly 😂

specs/bls_signature.md
CarlBeek: No there isn't. It is a strange absence.

specs/core/0_beacon-chain.md
CarlBeek: It is a function of how pyspec-SSZ works. It needs a bytes input, and python bytes look ugly.

However, there is a special case already in the pyspec builder for handling Tags, so I have modified that to work with raw bytes. Have a look at the update and let me know what you think of it. :)

MrChico: I like it!

djrtwo:
Are you suggesting moving the methods like `bls_sign` out of the bls spec?
Even if using a wrapper object, I think it makes sense to keep these functions in a separate file (like the bls spec) rather than cramming them in the phase 0 doc.

I think it natural to present the functions from the BLS standards, immediately followed by the set of wrapper functions used explicitly throughout the rest of the spec.

This allows for an auditor to quickly see the BLS standards and how we use them all in one place without having to dig through other docs

JustinDrake:
> Are you suggesting moving the methods like `bls_sign` out of the bls spec?

Yes, that was the suggestion in terms of presentation :) Notice that without that separation of concerns there would be a somewhat circular definition. For example, the Eth2 spec currently uses `bls_verify` defined the BLS spec, but `bls_verify` itself uses `hash` which is defined in the Eth2 spec. The situation becomes worse with the application-level details involved in wrapping (namely `Container`, `hash_tree_root`, `tag`). The BLS spec does not need to import the SSZ spec to know what a `Container` is.

JustinDrake:
@CarlBeek: I think we're missing `bls_aggregate_pubkeys`.

djrtwo:
closing in favor of #1532

PR #1504, opened 2019-12-05
Phase 1 rebase

djrtwo:
Rebase and replace #1483 

-------- original comment from previous PR ------------

This PR continues the work of @vbuterin and #1427 contributors, rebases it to the latest branch dev branch (with v0.9.1 changes too), and refactors it for integration into the pyspec:
- Split proposal into:
   - "The Beacon Chain for Shards"
   - Shard transition / Fraud proof doc. (The WIP part)
- The old "misc beacon chain" and "shard data chain" docs are moved to `specs/old`. Some of it may still be relevant. To be removed from the spec once the last details are integrated well in the new docs.
    - [x] Permanently remove old docs
- [x] Start pruning/updating the outdated parts of the custody game doc
  - From conversation with Vitalik:
    - Remove challenge records and responses, the game can be more simple and immediate now
    - "I provide a preimage of some root along with proof that you signed it, and your subkey, and on-chain the entire custody function gets executed; if the bit doesn't match you get slashed and I get a reward"
  - [x] the record creation and response processing needs to be merged
    - ~~And I'm probably missing part of the new picture still, this is not my domain.~~ More discussion helper. **implemented custody revamp with the (simplified/faster) custody slashings** :tada:
- WIP: refactor spec builder:
  - ~~Phase 0 and phase 1 are namespaced as modules~~ Same as previous
  - [x] Phase 1 copies phase 0 functions, and overrides everything with the same name
  - ~~Phase 1 type definitions overrule the phase 0 definitions. However, they inherit the phase0 types to make the new types compatible with the previous phase0 functions.~~ Override with spec-builder. Imports from more modules result in import-loops and inheritance issues. Not to mention config reload problems.
    - ~~WIP: make pyspec SSZ correctly pick up that the super-type (phase 0) field annotations should be ignored in favor of the sub-type (phase 1) fields~~ See above
   - [x] Change config system to load presets immediately before container declarations. Reload spec module to apply config changes. No code-gen re-init containers anymore.
      - [x] test new approach
- [x] update configs with all the new constants.


specs/core/1_custody-game.md
```diff
@@ -336,37 +172,27 @@ def legendre_bit(a: int, q: int) -> int:
...
-
-### `get_chunk_bits_root`
-
-```python
-def get_chunk_bits_root(chunk_bits: Bitlist[MAX_CUSTODY_CHUNKS]) -> bit:
-    aggregated_bits = 0
-    for i, b in enumerate(chunk_bits):
-        aggregated_bits += 2**i * b
-    return legendre_bit(aggregated_bits, BLS12_381_Q)
+    bits = [legendre_bit((i + 1) * s[i % 2] + int.from_bytes(atom, "little"), BLS12_381_Q)
```
dankrad: This is the old construction, broken by Dmitry. It needs to be replaced by legendre_bit(sum(s[i % 2]**i * int.from_bytes(atom, "little")), BLS12_381_Q)

specs/core/1_custody-game.md
```diff
@@ -536,252 +352,93 @@ def process_early_derived_secret_reveal(state: BeaconState, reveal: EarlyDerived
...
+        slash_validator(state, custody_slashing.malefactor_index)
+        others_count = len(committee) - 1
+        whistleblower_reward = Gwei(malefactor.effective_balance // WHISTLEBLOWER_REWARD_QUOTIENT // others_count)
+        for attester_index in attesters:
+            if attester_index != custody_slashing.malefactor_index:
+                increase_balance(state, attester_index, whistleblower_reward)
+        # No special whisteblower reward: it is expected to be an attester. Others are free to slash too however. 
+    else:
+        # The claim was false, the custody bit was correct. Slash the whistleblower that induced this work.
+        slash_validator(state, custody_slashing.whistleblower_index)
```
dankrad: Do we need this or can we also just declare the transaction as invalid if this is the case?

The only reason to include this line would be to protect against DOS attacks by flooding the network with invalid custody slashing messages. (No honest block producers would include them in a block, though)

specs/core/1_custody-game.md
```diff
@@ -536,252 +352,93 @@ def process_early_derived_secret_reveal(state: BeaconState, reveal: EarlyDerived
...
+    # Verify the claim
+    if claimed_custody_bit != computed_custody_bit:
+        # Slash the malefactor, reward the other committee members
+        slash_validator(state, custody_slashing.malefactor_index)
+        others_count = len(committee) - 1
+        whistleblower_reward = Gwei(malefactor.effective_balance // WHISTLEBLOWER_REWARD_QUOTIENT // others_count)
+        for attester_index in attesters:
+            if attester_index != custody_slashing.malefactor_index:
+                increase_balance(state, attester_index, whistleblower_reward)
+        # No special whisteblower reward: it is expected to be an attester. Others are free to slash too however.
```
dankrad: This makes sense as there's no way to protect against front-running. Strange side-effect though, if several attesters produced incorrect custody bits, then some of them could collect the whistleblower rewards before being slashed?

[Two other possible solutions to the front-running problem: 1. Include proper front-running protection like we did in the Legendre bounty contract (probably don't want the complexity); 2. Give a reward only to the block producer]

specs/core/1_custody-game.md
```diff
@@ -536,252 +352,93 @@ def process_early_derived_secret_reveal(state: BeaconState, reveal: EarlyDerived
...
 
-### Handling of custody-related deadlines
+### Handling of reveal deadlines
 
-Run `process_reveal_deadlines(state)` immediately after `process_registry_updates(state)`:
+Run `process_reveal_deadlines(state)` after `process_registry_updates(state)`:
 
 ```python
-# begin insert @process_reveal_deadlines
-    process_reveal_deadlines(state)
```
dankrad: Don't we still need this?

specs/core/1_custody-game.md
```diff
@@ -129,79 +81,36 @@ The following types are defined, mapping into `DomainType` (little endian):
...
 
 ```python
-class CustodyChunkChallenge(Container):
-    responder_index: ValidatorIndex
+class CustodySlashing(Container):
+    # Attestation.custody_bits[data_index][committee.index(malefactor_index)] is the target custody bit to check.
+    # (Attestation.data.shard_transition_root as ShardTransition).shard_data_roots[data_index] is the root of the data.
+    data_index: uint64
+    malefactor_index: ValidatorIndex
+    malefactor_key: BLSSignature
```
```suggestion
    malefactor_secret: BLSSignature
```

specs/core/1_custody-game.md
```diff
@@ -536,252 +352,93 @@ def process_early_derived_secret_reveal(state: BeaconState, reveal: EarlyDerived
...
-        index=response.chunk_index,
-        root=challenge.data_root,
-    )
-    # Clear the challenge
-    records = state.custody_chunk_challenge_records
-    records[records.index(challenge)] = CustodyChunkChallengeRecord()
-    # Reward the proposer
-    proposer_index = get_beacon_proposer_index(state)
-    increase_balance(state, proposer_index, Gwei(get_base_reward(state, proposer_index) // MINOR_REWARD_QUOTIENT))
+    assert bls_verify(malefactor.pubkey, hash_tree_root(epoch_to_sign), custody_slashing.malefactor_key, domain)
```
```suggestion
    assert bls_verify(malefactor.pubkey, hash_tree_root(epoch_to_sign), custody_slashing.malefactor_secret, domain)
```

specs/core/1_custody-game.md
djrtwo: The incentives of the policing of the game are much cleaner when the natural policemen can be directly rewarded. If we only reward the proposer, then the attester-policemen who likely found the issue have to make a decision to either (a) wait until they are a proposer to get the reward or (b) broadcast on gossip immediately and forgo the reward for the health of the network.

In the normal case in which the crosslink is actually available, (b) isn't so strong because the network is currently healthy and getting the slashing on chain only serves to dissuade unhealthy activity in the future.

If attesters get rewarded, they are incentivized to broadcast immediately to a public gossip channel such that they can get rewarded and whoever picks it up gets something too.

-------

Agree that some sort of commit reveal or time-based mechanism with multiple on-chain messages should likely be avoided due to complexity.

-------
> if several attesters produced incorrect custody bits, then some of them could collect the whistleblower rewards before being slashed?

This is a slightly weird side effect, but because of the `WHISTLEBLOWER_REWARD_QUOTIENT`, even if all attesters were slashed and received some marginal reward, they would all strictly be losing `1 - 1 / WHISTLEBLOWER_REWARD_QUOTIENT` fraction of balance which is the majority of their capital.

specs/core/1_custody-game.md
djrtwo: We don't use the previous comment-based code generation method anymore. It was brittle and difficult to follow for clients and other contributors.

Instead we override functions and objects explicitly and fully in subsequent phases. For example in phase 1, we fully redefine `process_epoch` explicitly writing the previous and new calls

https://github.com/ethereum/eth2.0-specs/pull/1504/files#diff-fea137c7b15f508b0dce4aeb83f9e98cR759-R768

protolambda:
Merged in the earlier experimental fix from @djrtwo for attestations, and fixed the timing of the configs loading.
Remaining tests are still mostly out-of-range errors, but for different reasons than the config: new lists like online-countdowns and shard-states in attestations need to be reviewed.

specs/core/1_custody-game.md
vbuterin: I think the goal is precisely to protect against network-level DoS.

protolambda:
Configration, tests handling multiple phases, phase1 state creation, different typing/spec inconsistencies fixed, and some small spec improvements. Not many tests (relatively) remaining to fix :rocket: . The reveal-deadlines processing is slashing too often though, and breaking most of the remaining tests as far as I can see.

specs/core/1_custody-game.md
```diff
@@ -115,9 +67,7 @@ This document details the beacon chain additions and changes in Phase 1 of Ether
 | - | - |
 | `MAX_CUSTODY_KEY_REVEALS` | `2**4` (= 16) |
```
vbuterin: I think we need more than 16 `MAX_CUSTODY_KEY_REVEALS`. Every validator reveals once per 2048-epoch period, so in the worst case that's 2**22 validators revealing once every 2**16 slots, or 64 per slot. So the max should probably be 256.

protolambda:
Python 3.8 use in the spec now works thanks to the deposit contract PR.

So CI passes, and re-introduces phase1 spec :tada:.

Note that it's not perfect yet, and likely has some bugs left. It's experimental code after all.
Regardless, I am happy that we can finally surface this work into `dev`. From here we can iterate on the design, and make use of the performance and usability improvements that happened in parallel.

PR #1506, opened 2019-12-06
Avoid potential front-running exploit

mcdee:
This tracks (validator pubkey, withdrawal credentials) tuples in the deposit contract to ensure that for any single validator pubkey the contract will only accept a single withdrawal credentials.  If a later transaction is sent with (validator pubkey, withdrawal credentials') it will be rejected.


MrChico:
I'm not sure what front-running exploit is trying to be avoided here, but here's one that is introduced by the additions suggested in this PR:

Since the deposit contract doesn't do any signature verification, a front-running agent can look at pending transactions to the deposit contract and create a transaction with the same public key but different withdrawal credentials. The contract will now reject the transaction with the real credentials since the public key is already present, effectively allowing malicious agents to block valid entries to the merkle tree.

This is not possible in the current bytecode, since such frontruns do not have any effect (their signatures are wrong and their entries will be rejected in the `process_deposit` function).

mcdee:
Details of the exploit are at https://hackmd.io/bCZjQnOyRBqxyHKxKnTz0Q?view

And yes there is the possibility of attacking this but it would cost the attacker 1 Ether, which they may or may not be able to get back but certainly would be locked in the short to medium term, and would still result in the valid deposit being rejected rather than stolen.  So it's a step in the right direction.

It looks like it should be possible to verify the signature in `deposit()`, which would avoid this issue.  Do you concur?

MrChico:
That sounds more like a way to shoot yourself in the foot rather than an exploit. Wouldn't it suffice to clearly communicate to validators that there is no way of changing the withdrawal credentials after they have been set?

mcdee:
If the validator and staker are separate (possibly non-trusting) entities then this can be exploited.  The general point, though, is that the following statement does not hold true for the validator contract without this change:

> If the deposit contract is given a deposit with specific withdrawal credentials it should result in either the deposited amount being withdrawable by those credentials or the deposit being rejected.

MrChico:
Making that statement hold would require BLS signature verification, as you noted before.

mcdee:
In what situation does this not hold, with the proposed patch?

djrtwo:
So you can actually avoid front-running without passing the _full_ private key without making changes to the deposit contract. It uses the a

* Sally the staker has a BLS private and public key
* Derrick the delegate has a BLS private and public key
* Derrick sends his public key to Sally. Sally combines it with her public key to create the public signing key for the validator.
* Sally sends the deposit data `D` (including the combined key) to Derrick.
* Derrick signs `D` with his half of the key and sends it back to Sally
* Sally signs `D` and creates an aggregate of the signature that can be verified by the combined key.
* Sally sends `D` along with the 32 eth to the deposit contract
* _After_ successful confirmation in the deposit contract, Sally sends her half of the private key to Derrick to do active validation

Only half of the private key was every transmit on the wire, so exposure of the full key to a monitoring attacker is not a risk

mcdee:
@djrtwo I would suggest that is a workaround rather than a fix.  The issue still exists within the contract, and anyone who is unaware of it would remain vulnerable.

djrtwo:
It is a _method_ to trustlessly ensure your withdrawal credentials are as you want them while at the same time fully delegating your signing key to a delegate without exposing the full signing privkey in transit.

The proposed change to the deposit contract still does not solve for the scenario in which a staker wants to trustlessly define their withdrawal credentials without sending a private key on the wire.

mcdee:
Sorry I'm unsure of why the proposed change does not meet the requirement, could you point out the specific issue?

djrtwo:
The deposit contract has been formally verified and is frozen unless a major security vulnerability is found.

I suggest using the [above protocol](https://github.com/ethereum/eth2.0-specs/pull/1506#issuecomment-564086671) to ensure that a user can trustlessly make the deposit and set their withdrawal credentials before handing off full signing rights to the custodian.

Issue #1496, opened 2019-11-26
Store.best_justified_checkpoint can be updated with an older checkpoint


ericsson49:
Recent update to the [fork choice](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_fork-choice.md) spec introduced `best_justified_checkpoint` field in the [`Store`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_fork-choice.md#store) class.
However, when it's updated in the [on_block](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_fork-choice.md#on_block) method, there is no check that the new value's epoch is newer than `store.best_justified_checkpoint`. The check only ensures that the new value's epoch is newer than the `store.justified_checkpoint`.
```
# Update justified checkpoint
if state.current_justified_checkpoint.epoch > store.justified_checkpoint.epoch:
    store.best_justified_checkpoint = state.current_justified_checkpoint
    if should_update_justified_checkpoint(store, state.current_justified_checkpoint):
        store.justified_checkpoint = state.current_justified_checkpoint
```

For example, if `store.justified_checkpoint.epoch == N` and there arrived a block which state has `current_justified_checkpoint == N+2`, `store.best_justified_checkpoint` becoming `N+2` and `current_justified_checkpoint` being still `N+2`, if the block arrived at the middle or at the end of `N+3` epoch.
If later another block arrives (still during the `N+3` epoch) which state has `current_justified_checkpoint == N+1`, then it will override the `N+2` value in the `store.best_justified_checkpoint`.

djrtwo:
Nice catch!

fixing this now for the coming release

djrtwo:
@ericsson49 A fix is in #1508 for review

djrtwo:
closed via #1508

PR #1508, opened 2019-12-09
ensure the best better-justification is stored in fork choice

djrtwo:
Addresses #1496 

Minor change to fork choice to ensure that the best better-justification is tracked.


ericsson49:
Looks good to me.

PR #1510, opened 2019-12-10
explicit use of GENESIS_SLOT in fork choice 

### LINKED ISSUE #1502: genesis_time in relation to GENESIS_EPOCH and GENESIS_SLOT
Beacon chain [specification](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#initial-values) defines `GENESIS_SLOT` and `GENESIS_EPOCH` constants., but doesn't specify how they are related to `genesis_time`.
Intuitively, `genesis_time` should be equal to the `GENESIS_SLOT` start time (and `GENESIS_EPOCH` should be the epoch of `GENESIS_SLOT`).
However, [fork choice](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_fork-choice.md) calculates time ignoring `GENESIS_SLOT`.
It could be possible though that [get_current_slot](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_fork-choice.md#get_current_slot) actually specifies the right way to convert time to a slot value (so that `genesis_time` means start time of `Slot(0)`).

There is no problem, if `GENESIS_SLOT` and `GENESIS_EPOCH` are zeros (and always will be), as they currently are. However, if it's admissible that `GENESIS_SLOT`/`GENESIS_EPOCH` change in future, then there can be problems with the beacon chain spec and/or with its implementations.
Overall, it's not clear now what is the right way to convert time to a slot number, i.e. should `GENESIS_SLOT` be added or not. While, it's trivial to implement, if 'GENESIS_SLOT` changes, it can be quite troublesome to verify that all the necessary updates are really performed.

Because of the above, we propose to clarify explicitly in the specification:
- whether `GENESIS_SLOT`/`GENESIS_EPOCH` are expected to change in future, or they are just symbolic names for `0` and never going to have a different value
- if it is admissible that `GENESIS_SLOT` can change for any reason, then the specification should clarify whether `genesis_time` corresponds to the start of `Slot(0)` or to the start of `GENESIS_SLOT`

Additionally, we propose to modify the [fork choice](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_fork-choice.md) spec, so that it uses `slot` instead of `time`. I.e. replace `Store.time` with `Store.slot` and `on_tick(store, time)` with `on_tick(store, slot)` (and other related changes). Basically, the `Store.time` is only used to detect whether a block/attestation is too early/too late or in-time, and current slot can be used for the same purpose.
This will get rid of the problem with `GENESIS_SLOT` entirely, in the context of the fork choice (sub)spec. But the problem remains for the purpose of time-to-slot conversion, of course. Additionally, it will simplify the on_block/on_attestation code and make it a bit more readable.
We can provide a Pull Request with the proposed fork choice modification, if needed.

djrtwo:
resolves #1502 

* Make use of `GENESIS_SLOT` explicit when calculating slot for fork choice
* use `get_current_slot(store)` when assessing if can include new attestation in `on_attestation`
* Note: the issue above recommends changing `time` to `slot` in `Store`. We did not take this approach because the granularity of time may very well be necessary in future forks if we, for example, change the `SECONDS_PER_SLOT` and need to do non-trivial slot calculations

### DIFF
--- a/specs/core/0_fork-choice.md
+++ b/specs/core/0_fork-choice.md
@@ -13,9 +13,15 @@
             - [`LatestMessage`](#latestmessage)
             - [`Store`](#store)
             - [`get_genesis_store`](#get_genesis_store)
+            - [`get_slots_since_genesis`](#get_slots_since_genesis)
+            - [`get_current_slot`](#get_current_slot)
+            - [`compute_slots_since_epoch_start`](#compute_slots_since_epoch_start)
             - [`get_ancestor`](#get_ancestor)
             - [`get_latest_attesting_balance`](#get_latest_attesting_balance)
+            - [`filter_block_tree`](#filter_block_tree)
+            - [`get_filtered_block_tree`](#get_filtered_block_tree)
             - [`get_head`](#get_head)
+            - [`should_update_justified_checkpoint`](#should_update_justified_checkpoint)
         - [Handlers](#handlers)
             - [`on_tick`](#on_tick)
             - [`on_block`](#on_block)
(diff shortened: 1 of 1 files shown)


mkalinin:
We might want to use `get_current_slot(store)` in a few other places:
- https://github.com/ethereum/eth2.0-specs/pull/1510/files#diff-65ff52fa41121e465ff23d3b53fd0518L315
- https://github.com/ethereum/eth2.0-specs/pull/1510/files#diff-65ff52fa41121e465ff23d3b53fd0518L267

djrtwo:
yep, thanks @mkalinin

Issue #1511, opened 2019-12-10
The dequeue process of activation queue is inefficient


hwwhww:
(Updated the table)

### Issue
By observing Lighthouse testnet status, I just realized that the validators dequeue rate is `MIN_PER_EPOCH_CHURN_LIMIT/2` per epoch instead of `MIN_PER_EPOCH_CHURN_LIMIT` per epoch.

It's because the current `process_registry_updates` function first selected the eligible 4 validators sorted by `activation_eligibility_epoch`, and then update the `validator.activation_epoch` (if `validator.activation_epoch == FAR_FUTURE_EPOCH`). However, it may select **the same validators** as the previous round.

The current code:
```python
    # Queue validators eligible for activation and not dequeued for activation prior to finalized epoch
    activation_queue = sorted([
        index for index, validator in enumerate(state.validators)
        if validator.activation_eligibility_epoch != FAR_FUTURE_EPOCH
        and validator.activation_epoch >= compute_activation_exit_epoch(state.finalized_checkpoint.epoch)
    ], key=lambda index: state.validators[index].activation_eligibility_epoch)
    # Dequeued validators for activation up to churn limit (without resetting activation epoch)
    for index in activation_queue[:get_validator_churn_limit(state)]:
        validator = state.validators[index]
        if validator.activation_epoch == FAR_FUTURE_EPOCH:
            validator.activation_epoch = compute_activation_exit_epoch(get_current_epoch(state))
```

To make it easier to understand, here is the scenario:
1. At epoch `816`:
    * 16 validators get set activation eligibility. `validator.activation_eligibility_epoch` = `816`.
    * 4 validators get selected to `activation_queue`.
    * These validators got set `validator.activation_epoch` to `821`
2. At epoch `817`:
    * Since we sort the queue by `key=lambda index: state.validators[index].activation_eligibility_epoch`, the tiebreaker key may be validator index. So **the same** `4` validators get selected to `activation_queue`. (Their `activation_epoch = 821` is greater than or equal to `compute_activation_exit_epoch(finalized_epoch) = 820`, matching the condition). 
    * Since `validator.activation_epoch` != `FAR_FUTURE_EPOCH`, **no validators' `activation_epoch` would be changed**.

![image](https://user-images.githubusercontent.com/9263930/70531976-05d0cf80-1b91-11ea-8811-d1e996ff3858.png)

[...]

PR #1513, opened 2019-12-10
fix queue efficiency bug

djrtwo:
addresses #1511 

After digging around, I am 95% sure that this captures the original intent of the queue and also resolves the issue found in #1511. My main concern was if activations should be delayed wrt finality conditions, but I (1) found no historic evidence of this in the discussion and (2) have done due diligence in reasoning about it and confirm that the behavior specified now is safe and intended.

The behavior in this PR actually represents the original coded behavior before a series of small tweaks were made.

For (1), the comment "not dequeued for activation prior to finalized epoch" appears to show that there is an intended behavior in rate limiting wrt finality, but this comment was actually added after-the-fact as descriptive of the behavior of the code, not as an original comment describing the intended behavior of the system.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -1311,17 +1311,16 @@ def process_registry_updates(state: BeaconState) -> None:
         if is_active_validator(validator, get_current_epoch(state)) and validator.effective_balance <= EJECTION_BALANCE:
             initiate_validator_exit(state, ValidatorIndex(index))
 
-    # Queue validators eligible for activation and not dequeued for activation prior to finalized epoch
+    # Queue validators eligible for activation and not yet dequeued for activation prior
     activation_queue = sorted([
         index for index, validator in enumerate(state.validators)
         if validator.activation_eligibility_epoch != FAR_FUTURE_EPOCH
-        and validator.activation_epoch >= compute_activation_exit_epoch(state.finalized_checkpoint.epoch)
+        and validator.activation_epoch == FAR_FUTURE_EPOCH
     ], key=lambda index: state.validators[index].activation_eligibility_epoch)
-    # Dequeued validators for activation up to churn limit (without resetting activation epoch)
+    # Dequeued validators for activation up to churn limit
     for index in activation_queue[:get_validator_churn_limit(state)]:
         validator = state.validators[index]
-        if validator.activation_epoch == FAR_FUTURE_EPOCH:
-            validator.activation_epoch = compute_activation_exit_epoch(get_current_epoch(state))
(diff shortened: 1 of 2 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -1311,17 +1311,16 @@ def process_registry_updates(state: BeaconState) -> None:
...
             initiate_validator_exit(state, ValidatorIndex(index))
 
-    # Queue validators eligible for activation and not dequeued for activation prior to finalized epoch
+    # Queue validators eligible for activation and not yet dequeued for activation
     activation_queue = sorted([
         index for index, validator in enumerate(state.validators)
         if validator.activation_eligibility_epoch != FAR_FUTURE_EPOCH
-        and validator.activation_epoch >= compute_activation_exit_epoch(state.finalized_checkpoint.epoch)
+        and validator.activation_epoch == FAR_FUTURE_EPOCH
     ], key=lambda index: state.validators[index].activation_eligibility_epoch)
```
hwwhww: Maybe explicitly point out the tiebreaker is `index` for the readers.

```suggestion
    # Order by the sequence of activation_eligibility_epoch setting and then index.
    ], key=lambda index: (state.validators[index].activation_eligibility_epoch, index))
```

shayzluf:
@djrtwo can it lead to a situation that a validator have different index or doesn't exist in one fork and exists in other. activation can also differ on different forks. slashing can become even harder if finality assumption on validator state is taken out. wdyt?

djrtwo:
>  slashing can become even harder if finality assumption on validator state is taken out

The finality assumption doesn't actually exist in the code as is today.

The current activation queue (before and after the fix in this PR) allows for activating validators even when no finality. In certain scenarios, this can lead to issues being able to slash validators across forks.

Example scenario:
* two forks — A and B
* fork A comes to consensus on new eth1data, processes some new deposits, and then inducts new validator M
* fork B does not
* attackers create slashable messages between A and B, but makes sure to only reveal aggregate attestations of A that include validator M in the aggregates (no singles propogated)
* fork B can’t process the slashable messages with validator M
* but fork A can!

This assumes that the likelihood of new eth1data on one chain and not the other is negligible. Which is maybe a reasonable assumption. It might have latency issues in punishing though.

IMO the question of finality affecting this queue is a bit orthogonal to the immediate fix here. I'm chewing on it a little more though before formalizing my opinion.

One option is to be able to include missing deposits in the slashing message, do the PoP verification and use the provided pubkey(s) in the deposit in addition to the validator registry pubkeys.

djrtwo:
Intention is to only activate upon finality to avoid all these issues with slashing across forks. Closing in favor of #1514

PR #1512, opened 2019-12-10
Check table of content of markdown files with CI

MrChico:
The out of sync table of content problems started bothering me enough to actually do something about it. Addresses part of https://github.com/ethereum/eth2.0-specs/issues/1369
~Not sure if `master` is the appropriate target for this PR, I'll rebase if needed.~
Target updated to `v09x`

The diff is large, but the actual changes are small:
- `check_toc` makefile rule, fails if TOCs are out of date
- add `check_toc` to CI ~(please check this)~ (seems to work as intended now)
- Page titles and `table of contents` headers are removed from TOC
- Consistent header level for all files (which adds plenty of headers to `p2p-interface.md`)
- Fix malformed TOCs (`compute_slots_since_epoch_start`, `should_update_justified_checkpoint` in `Fork_choice` added, `BLS signature verification` removed from `bls_signature.md`).

Besides CI usage of `check_toc`, it can also be run _twice_ locally to update TOC of all relevant markdown files (requires `npm install -g doctoc` before run).

PR #1514, opened 2019-12-11
Fix queue rate and only activate upon finality

djrtwo:
addresses #1511 as well as a previously uncaught bug allowing for activations when no finality

* ensures efficient processing of activation queue
* ensures activations only happen at finality
* adds a number of explicit tests for the activation queue
* enhances `process_registry_updates` testing in general

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -601,6 +601,21 @@ def is_slashable_validator(validator: Validator, epoch: Epoch) -> bool:
     return (not validator.slashed) and (validator.activation_epoch <= epoch < validator.withdrawable_epoch)
 ```
 
+#### `is_eligible_for_activation`
+
+```python
+def is_eligible_for_activation(state: BeaconState, validator: Validator) -> bool:
+    """
+    Check if ``validator`` is eligible for activation.
+    """
+    return (
+        # Was placed in activation queue prior to most recent finalized epoch
+        validator.activation_eligibility_epoch < state.finalized_checkpoint.epoch
+        # Has not yet been activated
+        and validator.activation_epoch == FAR_FUTURE_EPOCH
+    )
+```
+
 #### `is_slashable_attestation_data`
 
 ```python
@@ -1314,8 +1329,7 @@ def process_registry_updates(state: BeaconState) -> None:
     # Queue validators eligible for activation and not yet dequeued for activation
     activation_queue = sorted([
         index for index, validator in enumerate(state.validators)
-        if validator.activation_eligibility_epoch != FAR_FUTURE_EPOCH
-        and validator.activation_epoch == FAR_FUTURE_EPOCH
+        if is_eligible_for_activation(state, validator)
     ], key=lambda index: state.validators[index].activation_eligibility_epoch)
     # Dequeued validators for activation up to churn limit
     for index in activation_queue[:get_validator_churn_limit(state)]:
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -623,8 +623,8 @@ def is_eligible_for_activation(state: BeaconState, validator: Validator) -> bool
     Check if ``validator`` is eligible for activation.
     """
     return (
-        # Was placed in activation queue prior to most recent finalized epoch
-        validator.activation_eligibility_epoch < state.finalized_checkpoint.epoch
+        # Placement in queue is finalized
+        validator.activation_eligibility_epoch <= state.finalized_checkpoint.epoch
         # Has not yet been activated
         and validator.activation_epoch == FAR_FUTURE_EPOCH
     )
@@ -1332,7 +1332,7 @@ def process_registry_updates(state: BeaconState) -> None:
     # Process activation eligibility and ejections
     for index, validator in enumerate(state.validators):
         if is_eligible_for_activation_queue(validator):
-            validator.activation_eligibility_epoch = get_current_epoch(state)
+            validator.activation_eligibility_epoch = get_current_epoch(state) + 1
 
         if is_active_validator(validator, get_current_epoch(state)) and validator.effective_balance <= EJECTION_BALANCE:
             initiate_validator_exit(state, ValidatorIndex(index))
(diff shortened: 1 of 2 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -601,6 +601,21 @@ def is_slashable_validator(validator: Validator, epoch: Epoch) -> bool:
...
+#### `is_eligible_for_activation`
+
+```python
+def is_eligible_for_activation(state: BeaconState, validator: Validator) -> bool:
+    """
+    Check if ``validator`` is eligible for activation.
+    """
+    return (
+        # Was placed in activation queue prior to most recent finalized epoch
+        validator.activation_eligibility_epoch < state.finalized_checkpoint.epoch
```
protolambda: So instead of making a temporary activation that is overwritten when not within finality, we only consider eligible epochs directly? Doesn't this allow activations after the current finalized epoch to be scheduled, if finality takes longer than the activation delay?

specs/core/0_beacon-chain.md
```diff
@@ -1311,17 +1326,15 @@ def process_registry_updates(state: BeaconState) -> None:
...
             initiate_validator_exit(state, ValidatorIndex(index))
 
-    # Queue validators eligible for activation and not dequeued for activation prior to finalized epoch
+    # Queue validators eligible for activation and not yet dequeued for activation
     activation_queue = sorted([
         index for index, validator in enumerate(state.validators)
-        if validator.activation_eligibility_epoch != FAR_FUTURE_EPOCH
-        and validator.activation_epoch >= compute_activation_exit_epoch(state.finalized_checkpoint.epoch)
+        if is_eligible_for_activation(state, validator)
     ], key=lambda index: state.validators[index].activation_eligibility_epoch)
```
protolambda: :+1: Check that we are still sorting by eligibility epoch; the queue should be fair.

specs/core/0_beacon-chain.md
```diff
@@ -1311,17 +1326,15 @@ def process_registry_updates(state: BeaconState) -> None:
         if is_active_validator(validator, get_current_epoch(state)) and validator.effective_balance <= EJECTION_BALANCE:
             initiate_validator_exit(state, ValidatorIndex(index))
 
-    # Queue validators eligible for activation and not dequeued for activation prior to finalized epoch
+    # Queue validators eligible for activation and not yet dequeued for activation
     activation_queue = sorted([
         index for index, validator in enumerate(state.validators)
-        if validator.activation_eligibility_epoch != FAR_FUTURE_EPOCH
-        and validator.activation_epoch >= compute_activation_exit_epoch(state.finalized_checkpoint.epoch)
```
protolambda: :+1: `compute_activation_exit_epoch` is a pure function, looking for side effects of old code.

specs/core/0_beacon-chain.md
djrtwo: * We can move into the activation queue (set `activation_eligibility_epoch`) whenever the minimum balance threshold is hit
* We allow _removal_ from the queue and into activation, when the knowledge of being in the queue  is finalized
* When removed from activation queue, we set `activation_epoch` to be `compute_activation_exit_epoch(current_epoch)` (an epoch slightly in the future). This will happen regardless of subsequent finality. We are protected because the entrance into the queue was finalized and known

specs/core/0_beacon-chain.md
```diff
@@ -1302,26 +1331,21 @@ def process_rewards_and_penalties(state: BeaconState) -> None:
...
+        if is_eligible_for_activation(state, validator)
     ], key=lambda index: state.validators[index].activation_eligibility_epoch)
```
hwwhww: Repost: Maybe explicitly point out the tiebreaker is `index` for the readers, and for abstracting the sorting algorithm.

```suggestion
    # Order by the sequence of activation_eligibility_epoch setting and then index.
    ], key=lambda index: (state.validators[index].activation_eligibility_epoch, index))
```

specs/core/0_beacon-chain.md
```diff
@@ -601,6 +601,35 @@ def is_slashable_validator(validator: Validator, epoch: Epoch) -> bool:
...
+    Check if ``validator`` is eligible to be placed into the activation queue.
+    """
+    return (
+        validator.activation_eligibility_epoch == FAR_FUTURE_EPOCH
+        and validator.effective_balance == MAX_EFFECTIVE_BALANCE
+    )
+```
+
+
+#### `is_eligible_for_activation`
```
hwwhww: Need to update ToC, or clean up together in #1512.

specs/core/0_beacon-chain.md
hwwhww: 1. Ah `<` makes much more sense...! 👍 
2. What's wrong with the `validator.activation_eligibility_epoch == state.finalized_checkpoint.epoch` case here? Isn't the `finalized_checkpoint.epoch` itself safe for activation?

specs/core/0_beacon-chain.md
```diff
@@ -601,6 +601,35 @@ def is_slashable_validator(validator: Validator, epoch: Epoch) -> bool:
     return (not validator.slashed) and (validator.activation_epoch <= epoch < validator.withdrawable_epoch)
 ```
 
+#### `is_eligible_for_activation_queue`
```
hwwhww: Need to update ToC, or clean up together in #1512.

shayzluf:
thanks @djrtwo assuring finality in activation makes slashing simpler

specs/core/0_beacon-chain.md
djrtwo: When we finalize epoch N, we are finalizing a block/state at the 0th block of N. The latest activation eligibilities happen at the end of an epoch and are set to current_epoch then. So the finalized activations at N are really those that were at the end of N-1.

I'm realizing, it might just make more sense to `validator.activation_eligibility_epoch = get_current_epoch(state) + 1` instead of doing this indirect `<`

hwwhww: ahh you're right, it's from `compute_start_slot_at_epoch`.

`validator.activation_eligibility_epoch = get_current_epoch(state) + 1` would be more clear on what is happening here and fit for the naming. 👍

specs/core/0_beacon-chain.md

PR #1518, opened 2019-12-12
Two tests for proposer indices being off because of (in)active validators

### LINKED ISSUE #1515: Proposer index edge cases to test
So thanks to Prysm testnet, an under-tested proposer property was found: the proposer index and shuffling can both be correct, but if the active validators don't match regular indices there can still be a problem. And it only is off after the first inactive validator index in the system, so some tests could look like they test shuffling well (or indirectly as by rewards or otherwise), but just missed it.

TLDR: todo, add test cases for:
- `proposer index > len(active validators)` (already have proposals, but not this edge case)
- `proposer index > any slashed index` (already have slashings, but not this edge case).

Larger validator numbers move the proposer index to be more random and more likely to be bigger than those other numbers, while the current smaller validator sets in tests are generally only 512 validators. (and increasing that would increase the total test size by approx. the same factor).

protolambda:
fixes #1515

As described in the issue:
- test for a block with a proposer index higher than an inactive index
- test for a block with a proposer index higher than active validator count

Since we can't rely on the other tests, since transitions may not fully run and hit it, and the registry size is small enough that chances would be low for this kind of thing to happen.

### DIFF
--- a/test_libs/pyspec/eth2spec/test/sanity/test_blocks.py
+++ b/test_libs/pyspec/eth2spec/test/sanity/test_blocks.py
@@ -269,7 +269,7 @@ def test_proposer_after_inactive_index(spec, state):
             # found a proposer that has a higher index than a disabled validator
             yield 'pre', state
             # test if the proposer can be recognized correctly after the inactive validator
-            signed_block = sign_block(spec, state, build_empty_block(spec, state), proposer_index=proposer_index)
+            signed_block = state_transition_and_sign_block(spec, state, build_empty_block(spec, state))
             yield 'blocks', [signed_block]
             yield 'post', state
             break
@@ -295,7 +295,7 @@ def test_high_proposer_index(spec, state):
             # found a proposer that has a higher index than the active validator count
             yield 'pre', state
             # test if the proposer can be recognized correctly, even while it has a high index.
-            signed_block = sign_block(spec, state, build_empty_block(spec, state), proposer_index=proposer_index)
+            signed_block = state_transition_and_sign_block(spec, state, build_empty_block(spec, state))
             yield 'blocks', [signed_block]
             yield 'post', state
             break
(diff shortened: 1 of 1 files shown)


terencechain:
Thanks proto!

test_libs/pyspec/eth2spec/test/sanity/test_blocks.py
```diff
@@ -253,6 +253,56 @@ def test_attester_slashing(spec, state):
...
+    # skip forward, get brand new proposers
+    state.slot = spec.SLOTS_PER_EPOCH * 2
+
+    while True:
+        proposer_index = spec.get_beacon_proposer_index(state)
+        if proposer_index > inactive_index:
+            # found a proposer that has a higher index than a disabled validator
+            yield 'pre', state
+            # test if the proposer can be recognized correctly after the inactive validator
+            signed_block = sign_block(spec, state, build_empty_block(spec, state), proposer_index=proposer_index)
```
djrtwo: I think you want `state_transition_and_sign_block` here to make the `post` be the post state of the block transition

test_libs/pyspec/eth2spec/test/sanity/test_blocks.py
```diff
@@ -253,6 +253,56 @@ def test_attester_slashing(spec, state):
...
+    state.slot = spec.SLOTS_PER_EPOCH * 2
+
+    active_count = len(spec.get_active_validator_indices(state, current_epoch))
+    while True:
+        proposer_index = spec.get_beacon_proposer_index(state)
+        if proposer_index >= active_count:
+            # found a proposer that has a higher index than the active validator count
+            yield 'pre', state
+            # test if the proposer can be recognized correctly, even while it has a high index.
+            signed_block = sign_block(spec, state, build_empty_block(spec, state), proposer_index=proposer_index)
```
djrtwo: ditto

PR #1525, opened 2019-12-15
Minor cosmetic cleanups

JustinDrake:
* avoid inline comment (# Validate state root)
* simplify header inequality check in `process_proposer_slashing` (using uniqueness of BLS signatures)
* add `block = signed_block.message` helper variable for readability
* (typo) clarify that the state transition function consumes a signed block (as opposed to a block)
* generally make comments more consistent
* consistent formatting of container instantiation for `DepositMessage`
* avoid using three lines for `rewards[index] += Gwei(max_attester_reward // attestation.inclusion_delay)`
* introduce `effective_balance` helper variable for readability, and to avoid multi-line statement
* consistent ordering of `MIN_EPOCHS_TO_INACTIVITY_PENALTY` in the time parameters table
* (typo) "Dequeued validators for activation up to churn limit" => Dequeue validators
* "Save current block as the new latest block" => "Cache current block" (for consistent with `process_slot`)
* (typo) "Verify the validator has not yet exited" => "Verify exit has not been initiated"
* Use Pythonic `default=` for `max()` call in `initiate_validator_exit`


hwwhww:
@JustinDrake just to make sure, do you want to switch to targeting on `v09x` branch (gets release sooner)?

JustinDrake:
> do you want to switch to targeting on v09x branch (gets release sooner)?

I don't mind either way :) (It looks like the `v09x` branch is 14 commits behind `dev`.)

If you think it's the best way forward, do feel free to switch the target branch. (Me doing the switch will likely result in a Git mess.)

protolambda:
Nice cleanup, but I think v0.9.4 was the last 0.9 release. So removing the draft status is all that's necessary for review & merge.

JustinDrake:
> removing the draft status is all that's necessary for review & merge

Done :)

djrtwo:
`dev` is what will be released as `v0.10`. 
Merging this and getting it into the release this week

specs/core/0_beacon-chain.md
```diff
@@ -1038,7 +1036,7 @@ def initiate_validator_exit(state: BeaconState, index: ValidatorIndex) -> None:
 
     # Compute exit queue epoch
     exit_epochs = [v.exit_epoch for v in state.validators if v.exit_epoch != FAR_FUTURE_EPOCH]
-    exit_queue_epoch = max(exit_epochs + [compute_activation_exit_epoch(get_current_epoch(state))])
+    exit_queue_epoch = max(exit_epochs, default=compute_activation_exit_epoch(get_current_epoch(state)))
```
ChihChengLiang: Looks like this is a not cosmetic change, but a behavior one. 

Say we have `exit_epochs=[100, 101, 102]` and `compute_activation_exit_epoch(get_current_epoch(state)))` evaluates `105`.
- `max(exit_epochs + [compute_activation_exit_epoch(get_current_epoch(state))])` evaluates `105`, and
- `max(exit_epochs, default=compute_activation_exit_epoch(get_current_epoch(state)))` evaluates `102`

The old expression wants a min of max, while the new expression wants a default value when `exit_epochs` is empty. That results in these behaviors:

- The old one says "' Even' if no one is exiting soon, you can't exit immediately", and 
- The new one says "If no one is exiting soon, you can't exit immediately. But if someone is, you can exit with them at the same epoch."

I think the former is what we want.

protolambda: Good catch :+1: 
And also note that `exit_epochs` may contain already-exited validators, so `default` will effectively not happen after the first exit. So it looks like it was completely broken :(

Issue #1529, opened 2019-12-17
96 undefined names in ./deposit_contract/contracts/validator_registration.v.py


cclauss:
[flake8](http://flake8.pycqa.org) testing of https://github.com/ethereum/eth2.0-specs on Python 3.8.0

$ __flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics__
```
./deposit_contract/contracts/validator_registration.v.py:2:21: F821 undefined name 'constant'
MIN_DEPOSIT_AMOUNT: constant(uint256) = 1000000000  # Gwei
                    ^
./deposit_contract/contracts/validator_registration.v.py:2:30: F821 undefined name 'uint256'
MIN_DEPOSIT_AMOUNT: constant(uint256) = 1000000000  # Gwei
                             ^
./deposit_contract/contracts/validator_registration.v.py:3:30: F821 undefined name 'constant'
DEPOSIT_CONTRACT_TREE_DEPTH: constant(uint256) = 32
                             ^
./deposit_contract/contracts/validator_registration.v.py:3:39: F821 undefined name 'uint256'
DEPOSIT_CONTRACT_TREE_DEPTH: constant(uint256) = 32
                                      ^
./deposit_contract/contracts/validator_registration.v.py:4:20: F821 undefined name 'constant'
MAX_DEPOSIT_COUNT: constant(uint256) = 4294967295 # 2**DEPOSIT_CONTRACT_TREE_DEPTH - 1
                   ^
./deposit_contract/contracts/validator_registration.v.py:4:29: F821 undefined name 'uint256'
MAX_DEPOSIT_COUNT: constant(uint256) = 4294967295 # 2**DEPOSIT_CONTRACT_TREE_DEPTH - 1
                            ^
./deposit_contract/contracts/validator_registration.v.py:5:16: F821 undefined name 'constant'
PUBKEY_LENGTH: constant(uint256) = 48  # bytes
               ^
./deposit_contract/contracts/validator_registration.v.py:5:25: F821 undefined name 'uint256'
PUBKEY_LENGTH: constant(uint256) = 48  # bytes
                        ^
./deposit_contract/contracts/validator_registration.v.py:6:32: F821 undefined name 'constant'
WITHDRAWAL_CREDENTIALS_LENGTH: constant(uint256) = 32  # bytes
                               ^
./deposit_contract/contracts/validator_registration.v.py:6:41: F821 undefined name 'uint256'
[...]

PR #1530, opened 2019-12-17
validator_registration.v.py is a Vyper file, not a Python file

cclauss:
As discussed in #1529, validator_registration.v.py is a Vyper file, not a Python file.


djrtwo:
Thanks! Closing in favor of #1559

PR #1532, opened 2019-12-17
Yet another attempt at adopting IETF BLS Standards

CarlBeek:
This is an alternate approach to #1398 and #1499 with the goal of having complete separation between BLS libraries and the spec.

Initially I attempted to include @JustinDrake's requests in #1499, but it got very ugly very quickly. So this is a rewrite that deprecates #1499.

@JustinDrake, I believe this addresses your concerns from #1499, please let me know whether it does.


specs/core/1_custody-game.md
```diff
@@ -131,7 +131,7 @@ The following types are defined, mapping into `DomainType` (little endian):
 
 | Name | Value |
 | - | - |
-| `DOMAIN_CUSTODY_BIT_CHALLENGE` | `6` |
+| `DOMAIN_CUSTODY_BIT_CHALLENGE` | `0x05000000` |
```
protolambda: Careful here. 5 used to be the transfer domain type. Hope clients pick up on this change. Some may still have the unused transfer code around.

CarlBeek: I am aware, but I did the switch so as not to have an empty index. Happy to skip `5` if there are concerns about clients missing the change.

CarlBeek: An alternate idea would be to put the change from `uints` -> `bytes` in a separate PR which highlights the changes.

specs/core/0_beacon-chain.md
```diff
@@ -575,13 +582,17 @@ def bytes_to_int(data: bytes) -> uint64:
...
+Eth2 makes use of BLS signatures as specified in the [IETF draft BLS specification](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-00).
 
-#### `bls_aggregate_pubkeys`
+Specifically, eth2 uses the `BLS_SIG_BLS12381G2-SHA256-SSWU-RO-_POP_` ciphersuite where it makes use of the following functions:
 
-`bls_aggregate_pubkeys` is a function for aggregating multiple BLS public keys into a single aggregate key, as defined in the [BLS Signature spec](../bls_signature.md#bls_aggregate_pubkeys).
+* `def Sign(SK: int, message: Bytes) -> BLSSignature`
+* `def Verify(PK: BLSPubkey, message: Bytes, signature: BLSSignature) -> bool`
+* `def Aggregate(signatures: Sequence[BLSSignature]) -> BLSSignature`
+* `def bls_aggregate_pubkeys(PKs: Sequence[BLSPubkey]) -> BLSPubkey`
```
djrtwo: Looks like this function is unused in the pyspec (and not explicitly defined in the BLS standard)

specs/core/0_beacon-chain.md
```diff
@@ -789,6 +796,17 @@ def compute_domain(domain_type: DomainType, fork_version: Version=Version()) ->
     return Domain(domain_type + fork_version)
 ```
 
+### `compute_domain_wrapper_root`
+
+```python
+def compute_domain_wrapper_root(object: SSZObject, domain: Domain) -> Root:
```
djrtwo: `object` is a keyword in python. Probably best to not overwrite it here.

how about `ssz_object`?

specs/core/0_beacon-chain.md
```diff
@@ -1131,8 +1149,8 @@ def state_transition(state: BeaconState, signed_block: SignedBeaconBlock, valida
 ```python
 def verify_block_signature(state: BeaconState, signed_block: SignedBeaconBlock) -> bool:
-    domain = get_domain(state, DOMAIN_BEACON_PROPOSER)
-    return bls_verify(proposer.pubkey, hash_tree_root(signed_block.message), signed_block.signature, domain)
+    message = compute_domain_wrapper_root(signed_block.message, get_domain(state, DOMAIN_BEACON_PROPOSER))
```
djrtwo: I know `message` is the name of the param  in the context of `Verify`, but `message` here can easily be conflated with `message` we see in our signed containers (like `signed_block.message`). Maybe we call it `signed_root` or something to disambiguate

specs/core/0_beacon-chain.md
```diff
@@ -1431,7 +1449,8 @@ def process_randao(state: BeaconState, body: BeaconBlockBody) -> None:
     epoch = get_current_epoch(state)
     # Verify RANDAO reveal
     proposer = state.validators[get_beacon_proposer_index(state)]
-    assert bls_verify(proposer.pubkey, hash_tree_root(epoch), body.randao_reveal, get_domain(state, DOMAIN_RANDAO))
+    message = compute_domain_wrapper_root(epoch, get_domain(state, DOMAIN_RANDAO))
```
djrtwo: Ditto from last comment. But open to discussion. not 100% sure

specs/core/0_beacon-chain.md
djrtwo: Want to hear your opinion. I'm no longer sure 😅

specs/validator/0_beacon-chain-validator.md
```diff
@@ -311,7 +312,8 @@ def compute_new_state_root(state: BeaconState, block: BeaconBlock) -> Root:
 ```python
 def get_block_signature(state: BeaconState, header: BeaconBlockHeader, privkey: int) -> BLSSignature:
     domain = get_domain(state, DOMAIN_BEACON_PROPOSER, compute_epoch_at_slot(header.slot))
-    return bls_sign(privkey, hash_tree_root(header), domain)
+    message = compute_domain_wrapper_root(compute_epoch_at_slot(header), domain)
```
djrtwo:
```suggestion
    message = compute_domain_wrapper_root(header, domain)
```

test_libs/pyspec/eth2spec/utils/bls.py
```diff
@@ -23,31 +23,35 @@ def entry(*args, **kw):
...
     return bls.aggregate_signatures(signatures)
 
 
 @only_with_bls(alt_return=STUB_SIGNATURE)
-def bls_sign(message_hash, privkey, domain):
-    return bls.sign(message_hash=message_hash, privkey=privkey,
-                    domain=domain)
+def Sign(SK, message):
+    return bls.sign(message_hash=message, privkey=SK,
+                    domain=b'')
```
djrtwo: Let's make this one line

specs/core/0_beacon-chain.md
CarlBeek: Agreed I've removed it.

specs/core/0_beacon-chain.md
CarlBeek: `signing_root` is a fantastic name that I keep coming back to. 😅

I'm a little reticent about using `signed_root` a signed root is a signature in my mind.

One of the ideas I came up with is `wrapped_root`. In general, I am not 100% satisfied with the names I've implemented and am open to suggestions.

specs/core/1_custody-game.md
CarlBeek: I've reverted this change for the aforementioned clarity reasons. I had failed to make the changes in the configs.yaml which was further evidence for your point.

CarlBeek:
> How far are we from having a conforming py_ecc?

Pretty darn close. I opened [a big PR](https://github.com/ethereum/py_ecc/pull/83) this morning. It passes all tests etc.

Annother point I'd like to raise for clarity is whether we should treat BLS as a module. Is `bls.Verify()` clearer than `Verify`? In a similar vein, we could alias all bls functions: `bls_verify = Verify` etc.

specs/light_client/sync_protocol.md
```diff
@@ -135,9 +135,10 @@ def update_memory(memory: LightClientMemory, update: LightClientUpdate) -> None:
     assert 3 * sum(filter(lambda i: update.aggregation_bits[i], balances)) > 2 * sum(balances)
 
     # Verify shard attestations
-    pubkey = bls_aggregate_pubkeys(filter(lambda i: update.aggregation_bits[i], pubkeys))
+    pubkeys = filter(lambda i: update.aggregation_bits[i], pubkeys)
     domain = compute_domain(DOMAIN_SHARD_ATTESTER, update.fork_version)
-    assert bls_verify(pubkey, update.shard_block_root, update.signature, domain)
+    message = compute_domain_wrapper_root(update.shard_block_root, domain)
+    assert FastAggregateVerify(pubkey, message, update.signature)
```
djrtwo:
```suggestion
    assert FastAggregateVerify(pubkeys, message, update.signature)
```

djrtwo:
I like the idea of treating it like a module for readability. You can easily jump to a random part of the spec and see `bls.Verify` and know what is going on.

A quick note that they are used under the `bls` module in the "BLS Signatures" section for added clarification.

As for `bls.Verify` vs `bls_verify`... Keeping the names in line with the standard seems fine even though it breaks pythonic convention. I like the easy mapping to the standards spec. I find myself searching that doc from time to time and it's easy to see what to search for.

CarlBeek:
BLS is now a module and I think it is much clearer.

If we can reach agreement on what to call roots before they are signed/verified then we can get this thing merged. 🎉@djrtwo, can I ask for your input on our comment chain above re: `wrapped_root`.

specs/core/0_beacon-chain.md
djrtwo: I'm actually pretty fine with `message`. Maybe `message_root` but thats a bit long

djrtwo:
`wrapped_root` is reasonable but wrapped doesn't give too much description on what the wrapping is.

`message` is reasonable except that the signed containers use `message` for the entire object. `message_root` might work well because it is the `root` of the `message` from the signed container.

I'm okay with `message` or `message_root`

specs/core/0_beacon-chain.md
```diff
@@ -575,13 +582,16 @@ def bytes_to_int(data: bytes) -> uint64:
...
 `def hash_tree_root(object: SSZSerializable) -> Root` is a function for hashing objects into a single root by utilizing a hash tree structure, as defined in the [SSZ spec](../simple-serialize.md#merkleization).
 
-#### `bls_verify`
+#### BLS Signatures
 
-`bls_verify` is a function for verifying a BLS signature, as defined in the [BLS Signature spec](../bls_signature.md#bls_verify).
+Eth2 makes use of BLS signatures as specified in the [IETF draft BLS specification](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-00). Specifically, eth2 uses the `BLS_SIG_BLS12381G2-SHA256-SSWU-RO-_POP_` ciphersuite which implements the following interfaces:
 
-#### `bls_aggregate_pubkeys`
+* `def Sign(SK: int, message: Bytes) -> BLSSignature`
```
JustinDrake: Do we need `Sign`?

specs/core/0_beacon-chain.md
```diff
@@ -575,13 +582,16 @@ def bytes_to_int(data: bytes) -> uint64:
...
+* `def Verify(PK: BLSPubkey, message: Bytes, signature: BLSSignature) -> bool`
+* `def Aggregate(signatures: Sequence[BLSSignature]) -> BLSSignature`
```
JustinDrake: Do we need `Aggregate`?

specs/core/0_beacon-chain.md
```diff
@@ -789,6 +795,17 @@ def compute_domain(domain_type: DomainType, fork_version: Version=Version()) ->
     return Domain(domain_type + fork_version)
 ```
 
+### `compute_domain_wrapper_root`
```
JustinDrake: "domain wrapper" feels semantically wrong. It's not the domain that is being wrapped, it's the object root. Also the wrapping adds more than just the domain (in particular, it adds the uniqueness root).

What about `compute_signing_root`? (Also shorter name.)

specs/core/0_beacon-chain.md
djrtwo: We use `Sign` and `Aggregate` in the validator doc. For simplicity in specifying the BLS methods in one place, I'm in favor of putting them here

JustinDrake: There's a similar thing going on with `BLS_WITHDRAWAL_PREFIX`.

specs/core/0_beacon-chain.md
djrtwo: By uniqueness root, you mean the root of the object?

I would say `compute_signing_root` or something similar works.
Funny, this is intimately tied to the other conversation on what to call the output of this function

JustinDrake: > By uniqueness root, you mean the root of the object?

No, a different root where `uniqueness_root = hash_tree_root(uniqueness_data))` :) See https://github.com/ethereum/eth2.0-specs/issues/1387 which you suggested by handled in a separate PR for clarity.

CarlBeek:
The way domain-bytes were handled was a hack which I perpetuated with the change of domain into `bytes` in lieu of `ints` cast to `bytes`. 

`BLS_WITHDRAWAL_PREFIX` had a similar problem and so I fixed both in #1551

specs/core/0_beacon-chain.md
CarlBeek: > There's a similar thing going on with BLS_WITHDRAWAL_PREFIX.

I didn't want to touch this due to it not strictly being related to the IETF BLS standards and because it needs to be fixed with either a different hack in the spec builder or change the way the PySpec SSZ works. The latter option was taken in #1551 .

specs/core/0_beacon-chain.md
CarlBeek: So IMO, `signing_root` is the correct name here. The problem is the collision with the signing root removed from SSZ. My concern is that it may be confusing to implementors (particularly if they are implementing both changes at once.) Do people that this is a problem?

An alternate name is `domain_wrapped_root` which is semantically correct.

As @djrtwo points out, the naming problem exists across the whole PR. Which is what I was getting at when I mentioned that "In general, I am not 100% satisfied with the names I've implemented and am open to suggestions."

djrtwo: I'm okay with `signing_root` with a warning in the release. It is semantically correct and easy to discuss verbally.

If this were a phase 1 question after release of phase 0, I'd be a little bit more concerned.

CarlBeek: Alright, I've gone with `signing_root`. :) It is definitely the clearest solution.

CarlBeek:
Having replaced all the `domain_wrapper`s with `signing_root`s, I think this PR is semantically complete. 🎉 The only thing that should change first is the BLS API in the tests. As soon as https://github.com/ethereum/py_ecc/pull/83 is merged and a new release is cut, I can change the API in this PR and then merge.

djrtwo:
https://github.com/ethereum/py_ecc/pull/83 is merged!

JustinDrake:
@CarlBeek: What is the status of PoP separation domain? Is that nailed or still somewhat in flux relative to the standardisation effort?

> waiting on the py_ecc release before merge

Do we also want basic tests?

djrtwo:
> Do we also want basic tests

@JustinDrake once we integrate `py_ecc` here and all existing tests run, we'll have satisfied the testing requirement. [py_ecc independent tests and all of the existing eth2 test suite]

CarlBeek:
> * added version bump in a couple more places
> * bumped circlecli cache version to resolve CI failure
> * fixed merge conflicts

I was just trying to push these changes myself, but I see you've already done them. Looks good to me! Ready for merge.

PR #1534, opened 2019-12-17
Add ENR shard bitfield

djrtwo:
Make persistent shard subnet gossip subscriptions easily accessible via the ENR. This allows for simpler discovery of peers on shard subnets and for the removal of the "guess and check" method previously described in the validator doc.

The end goal is to still use discv5 topics as a superior mechanism in the long run, but this will do for Phase 0.

### DIFF
--- a/specs/networking/p2p-interface.md
+++ b/specs/networking/p2p-interface.md
@@ -53,11 +53,10 @@ It consists of four main sections:
   - [The discovery domain: discv5](#the-discovery-domain-discv5)
     - [Integration into libp2p stacks](#integration-into-libp2p-stacks)
     - [ENR structure](#enr-structure)
-      - [Shard bitfield](#shard-bitfield)
+      - [Attestation subnet bitfield](#attestation-subnet-bitfield)
       - [Interop](#interop-5)
       - [Mainnet](#mainnet-5)
     - [Topic advertisement](#topic-advertisement)
-      - [Interop](#interop-6)
       - [Mainnet](#mainnet-6)
 - [Design decision rationale](#design-decision-rationale)
   - [Transport](#transport-1)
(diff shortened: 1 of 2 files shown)


specs/validator/0_beacon-chain-validator.md
```diff
@@ -198,7 +198,7 @@ Specifically a validator should:
 * Call `get_committee_assignment(state, next_epoch, validator_index)` when checking for next epoch assignments.
 * Join the pubsub topic -- `committee_index{committee_index % ATTESTATION_SUBNET_COUNT}_beacon_attestation`.
     * If any current peers are subscribed to the topic, the validator simply sends `subscribe` messages for the new topic.
```
djrtwo: Considering changing the language here to "sufficient" number of peers, otherwise go find some more. "Any" implies that 1 is enough.. which it might be, but a validator might also want to go find some more to better disseminate their attestation.

specs/networking/p2p-interface.md
```diff
@@ -557,6 +558,14 @@ The Ethereum Node Record (ENR) for an Ethereum 2.0 client MUST contain the follo
 
 Specifications of these parameters can be found in the [ENR Specification](http://eips.ethereum.org/EIPS/eip-778).
 
+#### Shard bitfield
+
+The ENR MAY contain an entry signifying the shard subnet bitfield with the following form to more easily discover peers participating in particular shard gossip subnets.
```
protolambda: This is confusing. Subnets != shards, as per `ATTESTATION_SUBNET_COUNT`. Are we sticking long term to describe peers by their shard they are focusing on? Or the subnet they are on? Which may have more shards (possible if shard count increases), or only be part of a shard (unlikely). If just subnets, let's don't name it `shards`.

specs/validator/0_beacon-chain-validator.md
```diff
@@ -198,7 +198,7 @@ Specifically a validator should:
...
-    * If no current peers are subscribed to the topic, the validator must discover new peers on this topic. If "topic discovery" is available, use topic discovery to find peers that advertise subscription to the topic. If not, "guess and check" by connecting with a number of random new peers, persisting connections with peers subscribed to the topic and (potentially) dropping the new peers otherwise.
+    * If no current peers are subscribed to the topic, the validator must discover new peers on this topic. Find peers via the discovery protocol with an ENR containing the `shards` entry such that `ENR["shards"][committee_index % ATTESTATION_SUBNET_COUNT] == True`.
```
protolambda: Agree with the method, but if we look up a subnet, we should call it a subnet, and not `shards`.

protolambda:
Naming issue, but otherwise ok. And if intentional, `shards` is not too bad as identifier.

djrtwo:
updated to explicitly use attestation subnets

also made some updates to network spec FAQs and discussion of discv5 topics

specs/networking/p2p-interface.md
djrtwo: resolved

PR #1535, opened 2019-12-17
Content-address gossipsub messages

### LINKED ISSUE #1528: [Networking] Gossipsub content-addressed messages
Currently, gossipsub uses `source_peer_id + sequence_number` to address the messages sent across the gossipsub network. If a client re-publishes a seen message, this will look like a new message to all other peers. This can lead to duplicate messages on the network, which then needs to be filtered at the application layer. 

Rust and go (https://github.com/libp2p/go-libp2p-pubsub/pull/248) now have the ability to customise  the message id of gossipsub messages. I propose we set the gossipsub message id to:
`base64(sha256(data))` where `data` is the gossipsub protobuf `data` field which typically contains our ssz-encoded data or snappy-compressed data. 

This way, gossipsub will filter out duplicate messages before notifying the application layer. In principle, we then just need to verify the hash at the application layer to ensure duplicates aren't sent/received.

AgeManning:
## Description

Specifies that all gossipsub messages are identified as `base64(sha256(message.data))` which prevents duplicate messages (by content) from being received at the application layer. 

For our current purposes, there is no need to address messages based on source peer and by content-addressing we can filter unnecessary duplicates before hitting the application layer.

Some examples of where messages could be duplicated:
- A validator client connected to multiple beacon nodes publishing duplicate gossip messages
- Attestation aggregation strategy where clients partially aggregate attestations and propagate them. Partial aggregates could be duplicated.
- Clients re-publishing seen messages.

This closes #1528.

### DIFF
--- a/specs/networking/p2p-interface.md
+++ b/specs/networking/p2p-interface.md
@@ -212,6 +212,13 @@ Topics are plain UTF-8 strings and are encoded on the wire as determined by prot
 
 Each gossipsub [message](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/rpc.proto#L17-L24) has a maximum size of `GOSSIP_MAX_SIZE`. Clients MUST reject (fail validation) messages that are over this size limit. Likewise, clients MUST NOT emit or propagate messages larger than this limit.
 
+The message-id of a gossipsub message MUST be:
+
+```python
+   message-id: base64(SHA256(message.data))
+```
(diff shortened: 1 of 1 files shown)


djrtwo:
Can you add a quick reference to why in the FAQ (the PR comment looks about right)?
and can you rebase to `v09x` please?

I will release this asap along with the ENR bitfield in #1534

djrtwo:
closing in favor of #1538

PR #1549, opened 2019-12-20
Allow empty lists in streamed responses

arnetheduck:
It's possible that block request responses end up not containing any blocks, so we need a way to encode this.


protolambda:
Sadly this didn't make v0.9.4, but it looks like a good solution. I'm thinking we should open a networking-only branch for things that pop up during the testnets until v0.10 some time end Jan./mid Feb., and ensure consensus-compatibility with v0.9.3, as this also probably is not the last networking feature to improve testnets with. Sounds good?

djrtwo:
@arnetheduck Can you run `make check_toc` twice (first will find issues in table of contents, second time makes the edits) to fix CI?

arnetheduck:
oh my, twice?

> networking-only branch

or just do 0.9.5 - technically, nothing that worked before is broken with this branch - I don't think we're at the stage yet where we need to be overly strict with compat - this is a much smaller break than `signed_root` for example

djrtwo:
`dev` is fine. We will be releasing `v0.10.0` from `dev`

PR #1551, opened 2019-12-27
PySpec SSZ Bytes instantiated from hex string

CarlBeek:
This enables PySpec SSZ to instantiate Bytes from hex strings as input (both with and without leading `0x`). This removes some hackiness in the form of special cases from the spec builder. Bytes objects can then be created directly from octal strings without being exposed to python's ugly byte formatting.

### DIFF
--- a/specs/core/0_beacon-chain.md
+++ b/specs/core/0_beacon-chain.md
@@ -251,11 +251,11 @@ The following values are (non-configurable) constants used throughout the specif
 
 | Name | Value |
 | - | - |
-| `DOMAIN_BEACON_PROPOSER` | `Bytes4('0x00000000')` |
-| `DOMAIN_BEACON_ATTESTER` | `Bytes4('0x01000000')` |
-| `DOMAIN_RANDAO` | `Bytes4('0x02000000')` |
-| `DOMAIN_DEPOSIT` | `Bytes4('0x03000000')` |
-| `DOMAIN_VOLUNTARY_EXIT` | `Bytes4('0x04000000')` |
+| `DOMAIN_BEACON_PROPOSER` | `DomainType('0x00000000')` |
+| `DOMAIN_BEACON_ATTESTER` | `DomainType('0x01000000')` |
+| `DOMAIN_RANDAO` | `DomainType('0x02000000')` |
+| `DOMAIN_DEPOSIT` | `DomainType('0x03000000')` |
+| `DOMAIN_VOLUNTARY_EXIT` | `DomainType('0x04000000')` |
 
 ## Containers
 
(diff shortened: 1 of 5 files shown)


specs/core/0_beacon-chain.md
```diff
@@ -249,15 +249,13 @@ The following values are (non-configurable) constants used throughout the specif
...
-| `DOMAIN_BEACON_PROPOSER` | `0` |
-| `DOMAIN_BEACON_ATTESTER` | `1` |
-| `DOMAIN_RANDAO` | `2` |
-| `DOMAIN_DEPOSIT` | `3` |
-| `DOMAIN_VOLUNTARY_EXIT` | `4` |
+| `DOMAIN_BEACON_PROPOSER` | `Bytes4('0x00000000')` |
+| `DOMAIN_BEACON_ATTESTER` | `Bytes4('0x01000000')` |
+| `DOMAIN_RANDAO` | `Bytes4('0x02000000')` |
+| `DOMAIN_DEPOSIT` | `Bytes4('0x03000000')` |
+| `DOMAIN_VOLUNTARY_EXIT` | `Bytes4('0x04000000')` |
```
protolambda: Why not `DomainType` instead of `Bytes4`? Is this changing again in the BLS PR?

specs/core/1_custody-game.md
```diff
@@ -131,7 +131,7 @@ The following types are defined, mapping into `DomainType` (little endian):
 
 | Name | Value |
 | - | - |
-| `DOMAIN_CUSTODY_BIT_CHALLENGE` | `6` |
+| `DOMAIN_CUSTODY_BIT_CHALLENGE` | `Bytes4('0x06000000')` |
```
protolambda: ditto

specs/core/1_shard-data-chains.md
```diff
@@ -101,8 +101,8 @@ This document describes the shard transition function (data layer only) and the
 
 | Name | Value |
 | - | - |
-| `DOMAIN_SHARD_PROPOSER` | `128` |
-| `DOMAIN_SHARD_ATTESTER` | `129` |
+| `DOMAIN_SHARD_PROPOSER` | `Bytes4('0x80000000')` |
+| `DOMAIN_SHARD_ATTESTER` | `Bytes4('0x81000000')` |
```
protolambda: ditto

specs/core/0_beacon-chain.md
CarlBeek: Of course, `DomainType` is much better.

>  Is this changing again in the BLS PR?

I don't intend it to do so. I am hoping this cleans up #1532.

PR #1552, opened 2020-01-02
Tree based spec SSZ implementation - remerkleable

protolambda:
This mostly just deletes the old SSZ implementation from the spec.
And then if we link in this library: https://github.com/protolambda/remerkleable, it passes all spec tests.

The library is different from status quo in the following ways:
- data-sharing at the core: every type is a view over a binary merkle tree.
- supports typing over **partial backings**: just provide the branches of the tree that matter to the reads/writes on the object, and it works.
- hash-function doesn't even show up in initial profiling, everything is cached in place
- data accesses are more expensive, as it creates temporary views over subtrees. E.g. a validator view for each registry entry to get active validator indices.
  - Quadratic-complexity state accesses slow down things, even more than a minimal-configuration state hash-tree-root. Naively calling a function for every validator, that then calls a function for every validator, is a big slowdown. Normal eth2 clients wouldn't do this, but the spec does things like this in reward-computation.
- For normal use (not the quadratic problem as described above), mainnet config is much more manageable for testing, state-hashes are fully cached.
- Typing/functionality extensions and improvements over status quo spec implementation
- Deserialization now fully supported.
- Lots of other SSZ feature goodies from remerkleable to use in tooling.

And compared to py-ssz:
- More readable types, pretty much the same syntax as status-quo spec.
- Caching and data-sharing everything, with access to the binary tree (good for loading typed merkle proofs at 0 cost). As opposed to pyrsistent, the data-sharing library that py-ssz uses.
- (de)serialization should be slightly faster, passing an stream around to read/write from.

~~PS: `pymerkle` is already taken as a name, `pymerkles` was a working name. Suggestions for a good name are welcome. Something that covers ssz / merkle partials better.~~ Decided on a new name: remerkleable! A mix-up of "remarkable"/"re-merkle-able".

In addition, the new cached hash-tree-roots are very useful for memoization of spec functions, bring big performance improvements to the pyspec code.


protolambda:
Need Python 3.8 for remerkleable typing features (as well as changes in phase 1 PR). Deposit tests are failing with 3.8 though, since vyper fails parsing when using the 3.8 AST code. Need to backport the PR fromt the vyper repo that fixed that.

protolambda:
Hmm pypi db index is slow, CI can't find v0.1.9 I just published. Will re-run after giving it some time to catch up

protolambda:
This PR is ready for review.

Some final notes:
- I checked the `ssz_static` generator output against that of the previous ssz solution, and it matched, except for some files that had differently generated random values. The generator is not perfect.
- Although the memoization is a big performance gain, the memoization is backed by just `dict` now. I'll swap in a proper memoization cache in another PR.

specs/phase0/beacon-chain.md
```diff
@@ -1308,27 +1307,27 @@ def get_attestation_deltas(state: BeaconState) -> Tuple[Sequence[Gwei], Sequence
         attesting_balance = get_total_balance(state, unslashed_attesting_indices)
         for index in eligible_validator_indices:
             if index in unslashed_attesting_indices:
-                rewards[index] += get_base_reward(state, index) * attesting_balance // total_balance
+                rewards[index] += get_base_reward(state, total_balance, index) * attesting_balance // total_balance
```
djrtwo: Instead of changing the input args on this function, can we instead put a cache on `get_total_active_balance` to get a similar gain?

specs/phase1/beacon-chain.md
```diff
@@ -305,7 +297,7 @@ class ShardBlockWrapper(Container):
     shard_parent_root: Root
     beacon_parent_root: Root
     slot: Slot
-    body: ByteList[MAX_SHARD_BLOCK_CHUNKS * SHARD_BLOCK_CHUNK_SIZE]
+    body: List[byte, MAX_SHARD_BLOCK_CHUNKS * SHARD_BLOCK_CHUNK_SIZE]
```
djrtwo: Why removal of helper type `ByteList`?

tests/core/pyspec/eth2spec/debug/random_value.py
```diff
@@ -51,65 +51,68 @@ def get_random_ssz_object(rng: Random,
...
-        elif mode == RandomizationMode.mode_one_count:
-            return typ(get_random_bytes_list(rng, min(1, typ.length)))
-        elif mode == RandomizationMode.mode_zero:
-            return typ(b'\x00' * min(1, typ.length))
-        elif mode == RandomizationMode.mode_max:
-            return typ(b'\xff' * min(1, typ.length))
-        else:
-            return typ(get_random_bytes_list(rng, rng.randint(0, min(max_bytes_length, typ.length))))
-    elif issubclass(typ, ByteVector):
+    # if issubclass(typ, ByteList):
```
djrtwo: Why removal here?

tests/core/pyspec/eth2spec/debug/random_value.py
```diff
@@ -51,65 +51,68 @@ def get_random_ssz_object(rng: Random,
...
+    #         return typ(b'\x00' * min(1, typ.limit()))
+    #     elif mode == RandomizationMode.mode_max:
+    #         return typ(b'\xff' * min(1, typ.limit()))
+    #     else:
+    #         return typ(get_random_bytes_list(rng, rng.randint(0, min(max_bytes_length, typ.limit()))))
+    if issubclass(typ, ByteVector):
         # Sanity, don't generate absurdly big random values
         # If a client is aiming to performance-test, they should create a benchmark suite.
-        assert typ.length <= max_bytes_length
+        assert typ.type_byte_length() <= max_bytes_length
```
djrtwo: Is this an alias for `vector_length` on the `ByteVector` type?

tests/core/pyspec/eth2spec/test/helpers/custody.py
```diff
@@ -119,7 +119,7 @@ def get_valid_custody_response(spec, state, bit_challenge, custody_data, challen
     data_branch = get_merkle_proof(data_tree, chunk_index)
 
     bitlist_chunk_index = chunk_index // BYTES_PER_CHUNK
-    bitlist_chunks = chunkify(pack(bit_challenge.chunk_bits))
+    bitlist_chunks = None  # TODO chunkify(pack(bit_challenge.chunk_bits))
```
djrtwo: What's going on here? Is this resolved in recent Phase 1 mods?

tests/core/pyspec/eth2spec/test/utils.py
```diff
@@ -1,6 +1,6 @@
 from typing import Dict, Any
 from eth2spec.debug.encode import encode
-from eth2spec.utils.ssz.ssz_typing import SSZValue
+from eth2spec.utils.ssz.ssz_typing import View
```
djrtwo: What exactly is a `View`?

tests/core/pyspec/eth2spec/utils/ssz/ssz_impl.py
```diff
@@ -1,157 +1,10 @@
...
-    Bitlist, Bitvector, uint, Bytes32
-)
+from remerkleable.core import View
+from remerkleable.byte_vector import Bytes32
 
-# SSZ Serialization
-# -----------------------------
 
-BYTES_PER_LENGTH_OFFSET = 4
+def serialize(obj: View) -> bytes:
```
djrtwo: I'm hesitant to fully delete the existing minimal impl, especially because it is (likely) much easier to build and test for correctness than remerkable. 

When SSZ in own repo, do you see value in maintaining this minimal impl there?

specs/phase0/beacon-chain.md
protolambda: Yes we probably should, this idea of not recomputing the balance to avoid a lot of state iterations was put in place before the memoization was implemented.

specs/phase1/beacon-chain.md
protolambda: Since phase0 doesn't use it, remerkleable doesn't have this type, and `List[byte, N]` will behave the same. Once we get into more serious phase 1 testing I can optimize it again.

tests/core/pyspec/eth2spec/debug/random_value.py
protolambda: See above, ByteList is out for now, `List[byte, N]` suffices.

tests/core/pyspec/eth2spec/debug/random_value.py
protolambda: `type_byte_length()` is common to every single view, and computes the serialized size for fixed-width types (exception if it can't be determined, for e.g. a dynamic size. That's what min/max size getters are for instead). For a vector of bytes, it's the same as the vector length.

tests/core/pyspec/eth2spec/test/helpers/custody.py
protolambda: This was one of the more tricky parts, but unused currently. The bitlist chunkification doesn't make much sense when the bitlist is already a tree. I can probably write some function in remerkleable to return the leaf nodes as chunks. But it was more of a pointless blocker than a critical thing for this PR.

tests/core/pyspec/eth2spec/test/utils.py
protolambda: A `View` is a typed interpretation of a binary tree node/tree (a backing). Essentially a value, but referencing immutable data (for complex types at least, basic types are the backing themselves as well).

tests/core/pyspec/eth2spec/utils/ssz/ssz_impl.py
protolambda: Legacy code has its own costs. Also see https://github.com/protolambda/remerkleable/blob/master/remerkleable/test_impl.py for hand-crafted tests. Can't be more correct than that. 

Only value would be in comparing the two implementations, but I'd rather compare against the old test vectors, and move on to new testing (more handcrafted edge cases, and compare against clients, fuzzing, etc.) instead of maintaining legacy code.

specs/phase0/beacon-chain.md
protolambda: Fixed in 5c5b528
Probably a good idea to squash the changes in this PR. There's a lot of dev iteration on small things.

tests/core/pyspec/eth2spec/debug/random_value.py
djrtwo: Probably best to make sure this type still shows up in test generators whenever dev gets released next

protolambda:
Just implemented `ByteList`, testing and preparing a new remerkleable release to include here.

tests/core/pyspec/eth2spec/debug/random_value.py
protolambda: ByteList is back with a new implementation. But that said, it's not used in Phase0, so I am not too worried.

protolambda:
Reduced it all to one single commit :+1:

Issue #1537, opened 2019-12-18
Concerns about eth1 voting in the context of caching


paulhauner:
I think the current eth1 voting mechanism has some undesirable properties when we consider that staking eth2 clients should cache eth1 blocks. They should cache because:

1. Requiring a call to some external server during block production is risky. If the eth1 node is unavailable at that time, we miss a beacon slot.
2. The amount of data required to make the vote is >1,000 blocks. Doing this remote call during block production will add significant (likely infeasible) lag to block production.

First I will state two undesirable properties of the current system in the context of caching and then suggest a simpler system.

### Undesirable property 1: Clients must cache the eth1 chain all the way up to the head

Consider the first block proposer in the eth1 voting period (`slot % SLOTS_PER_ETH1_VOTING_PERIOD == 0`). In order to calculate `get_eth1_data(distance)` it needs to know the block number of the eth1 block at the start of the voting period (now). That block is the head of the eth1 chain.

This is the primary problem I have and it has two effects:
- The eth2 client caches must be concerned with eth1 re-orgs, whilst the core eth2 protocol is not.
- There are points in time (the start of the voting period) where an eth2 node's ability to vote is reduced by not having a cache that is _immediately_ up to date.

### Undesirable property 2: You need to cache all the way back to the current eth1_data

In order to cast a vote (i.e., not trigger an exception in the spec), a node must have in their cache all descendants of the block represented by `state.eth1_data` (at the start of the current voting period). So, the cache grows linearly with the time since a successful eth1 voting period.

Additionally, if a node wants their cache to be safe in the case of an _eth2_ re-org, they should cache all the way back to the eth1_data in the last finalized block. Therefore, the cache also grows linearly with the time since finalization.

## My proposal:

Below is some **rough** Python code that I think is minimal and viable. I'm not convinced this should be the final solution, but it's a starting point at least.

It has the following properties that the present solution does not:

- Clients can go for `ETH1_FOLLOW_DISTANCE * SECONDS_PER_SLOT` without contacting an eth1 node and still vote perfectly.
- The size of this cache is limited to the amount of eth1 blocks in `ETH1_FOLLOW_DISTANCE * SECONDS_PER_SLOT`.
- It is reasonable to assume that a re-org will never affect this cache.

```python
def voting_period_start(state: BeaconState):
    eth1_voting_period_start_slot = state.slot % SLOTS_PER_ETH1_VOTING_PERIOD
[...]

PR #1553, opened 2020-01-03
`get_eth1_data` uses timestamp instead of block height

djrtwo:
Address #1537


djrtwo:
cc @paulhauner

specs/validator/0_beacon-chain-validator.md
```diff
@@ -239,28 +242,50 @@ def get_epoch_signature(state: BeaconState, block: BeaconBlock, privkey: int) ->
...
+def voting_period_start_time(state: BeaconState) -> uint64:
+    eth1_voting_period_start_slot = state.slot % SLOTS_PER_ETH1_VOTING_PERIOD
+    return state.genesis_time + eth1_voting_period_start_slot * SECONDS_PER_SLOT
+```
+
+```python
+def is_candidate_block(block: Eth1Block, period_start: uint64) -> bool:
+    return (
+        block.timestamp <= period_start - SECONDS_PER_ETH1_BLOCK * ETH1_FOLLOW_DISTANCE
+        and block.timestamp >= period_start - SECONDS_PER_ETH1_BLOCK * ETH1_FOLLOW_DISTANCE * 2
```
protolambda: I like the simplicity, but aside from a deep re-org, but just leaving a note here that a very slow (deprecation) or dead (with a remaining tail of deposits) eth1 will also be an extra-protocol thing for eth2.

specs/validator/0_beacon-chain-validator.md
```diff
@@ -239,28 +242,50 @@ def get_epoch_signature(state: BeaconState, block: BeaconBlock, privkey: int) ->
...
+###### `get_eth1_data`
+
+Let `get_eth1_data(block: Eth1Block) -> Eth1Data` be the function that returns the Eth1 data for a given Eth1 block.
+
+An honest block proposer sets `block.body.eth1_data = get_eth1_vote(state)` where:
+
+```python
+def voting_period_start_time(state: BeaconState) -> uint64:
+    eth1_voting_period_start_slot = state.slot % SLOTS_PER_ETH1_VOTING_PERIOD
+    return state.genesis_time + eth1_voting_period_start_slot * SECONDS_PER_SLOT
```
protolambda: Similar to the slot/epoch conversions, helper functions for time - slot conversion would be useful too. But leaving this as-is, we can change it when there are more cases.

PR #1556, opened 2020-01-05
move GENESIS_SLOT/EPOCH to constants

djrtwo:
`GENESIS_SLOT` and `GENESIS_EPOCH` are not truly configurable as currently specified (ex: they aren't explicitly written into genesis block/state). Instead of making substantive changes to allow for this, instead just move them to "Constants" section.

Explicitness in helpers is still desired, so also clarify the use of `GENESIS_SLOT` in a validator helper.

Also fix a bug that was introduced in #1553. Thanks @protolambda !

### DIFF
--- a/specs/validator/0_beacon-chain-validator.md
+++ b/specs/validator/0_beacon-chain-validator.md
@@ -267,7 +267,7 @@ def compute_time_at_slot(state: BeaconState, slot: Slot) -> uint64:
 
 ```python
 def voting_period_start_time(state: BeaconState) -> uint64:
-    eth1_voting_period_start_slot = Slot(state.slot % SLOTS_PER_ETH1_VOTING_PERIOD)
+    eth1_voting_period_start_slot = Slot(state.slot - state.slot % SLOTS_PER_ETH1_VOTING_PERIOD)
     return compute_time_at_slot(state, eth1_voting_period_start_slot)
 ```
 
--- a/specs/validator/0_beacon-chain-validator.md
+++ b/specs/validator/0_beacon-chain-validator.md
@@ -263,7 +263,8 @@ An honest block proposer sets `block.body.eth1_data = get_eth1_vote(state)` wher
 ```python
 def voting_period_start_time(state: BeaconState) -> uint64:
     eth1_voting_period_start_slot = state.slot % SLOTS_PER_ETH1_VOTING_PERIOD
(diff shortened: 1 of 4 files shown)


specs/validator/0_beacon-chain-validator.md
```diff
@@ -263,7 +263,8 @@ An honest block proposer sets `block.body.eth1_data = get_eth1_vote(state)` wher
 ```python
 def voting_period_start_time(state: BeaconState) -> uint64:
     eth1_voting_period_start_slot = state.slot % SLOTS_PER_ETH1_VOTING_PERIOD
-    return state.genesis_time + eth1_voting_period_start_slot * SECONDS_PER_SLOT
+    time_since_genesis = (eth1_voting_period_start_slot - GENESIS_SLOT) * SECONDS_PER_SLOT
```
protolambda: I don't think substracting by 0 really adds much here. A slot -> time helper function would be nice though.

djrtwo: resovled

specs/validator/0_beacon-chain-validator.md
```diff
@@ -260,10 +260,15 @@ Let `get_eth1_data(block: Eth1Block) -> Eth1Data` be the function that returns t
...
+```python
+def compute_time_at_slot(state: BeaconState, slot: Slot) -> uint64:
+    return state.genesis_time + slot * SECONDS_PER_SLOT
+```
+
 ```python
 def voting_period_start_time(state: BeaconState) -> uint64:
-    eth1_voting_period_start_slot = state.slot % SLOTS_PER_ETH1_VOTING_PERIOD
-    return state.genesis_time + eth1_voting_period_start_slot * SECONDS_PER_SLOT
+    eth1_voting_period_start_slot = Slot(state.slot % SLOTS_PER_ETH1_VOTING_PERIOD)
```
protolambda:
```suggestion
    eth1_voting_period_start_slot = state.slot - Slot(state.slot % SLOTS_PER_ETH1_VOTING_PERIOD)
```
It's pulled to alignment with `SLOTS_PER_ETH1_VOTING_PERIOD` right? Not from 0.

djrtwo: nice catch. made it in the last PR.
Going to fix it up with using spec helpers

djrtwo: nevermind. don't have good helpers for non-epoch based periods

PR #1560, opened 2020-01-08
Remove shard block chunking

vbuterin:
Only store a 32 byte root for every shard block

Rationale: originally, I added shard block chunking (store 4 chunks for every shard block instead of one root) to facilitate construction of data availability roots. However, it turns out that there is an easier technique. Set the width of the data availability rectangle's rows to be 1/4 the max size of a shard block, so each block would fill multiple rows. Then, non-full blocks will generally create lots of zero rows. For example if the block bodies are `31415926535` and `897932` with a max size of 24 bytes, the rows might look like this:

```
31415926
53500000
00000000
89793200
00000000
00000000
```
Zero rows would extend rightward to complete zero rows, and when extending downward we can count the number of zero rows, and reduce the number of extra rows that we make, so we only make a new row for every nonzero row in the original data. This way we get only a close-to-optimal ~4-5x blowup in the data even if the data has zero rows in the middle.

### DIFF
--- a/scripts/build_spec.py
+++ b/scripts/build_spec.py
@@ -48,7 +48,7 @@ from dataclasses import (
 from eth2spec.utils.ssz.ssz_impl import hash_tree_root
 from eth2spec.utils.ssz.ssz_typing import (
     View, boolean, Container, List, Vector, uint64, uint8, bit,
-    ByteVector, ByteList, Bytes1, Bytes4, Bytes8, Bytes32, Bytes48, Bytes96, Bitlist, Bitvector,
+    ByteList, Bytes1, Bytes4, Bytes8, Bytes32, Bytes48, Bytes96, Bitlist, Bitvector,
 )
 from eth2spec.utils import bls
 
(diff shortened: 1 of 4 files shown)


specs/core/1_beacon-chain.md
```diff
@@ -705,7 +691,7 @@ def apply_shard_transition(state: BeaconState, shard: Shard, transition: ShardTr
 def process_crosslink_for_shard(state: BeaconState,
                                 shard: Shard,
                                 shard_transition: ShardTransition,
-                                attestations: Sequence[Attestation]) -> Root:
+                                attestations: Sequence[Attestation]) -> Root:US
```
JustinDrake: Is the extra `US` a typo here?

vbuterin:
```suggestion
                                attestations: Sequence[Attestation]) -> Root:
```

djrtwo:
This looks good @vbuterin

We are going to get #1504 merged in shortly, then I'll rebase this and get it merged

PR #1575, opened 2020-01-16
Fix broken test generators

djrtwo:
* Fix BLS test generator by utilizing new py_ecc functions, types, etc. Strip down test cases to only test the 5 BLS functions in the spec
* Fix ssz_generic test generator with a simple import fix (`Bytes` -> `ByteList`)

### DIFF
--- a/tests/generators/bls/main.py
+++ b/tests/generators/bls/main.py
@@ -115,7 +115,7 @@ def case02_message_hash_G2_compressed():
 
 
 def case03_private_to_public_key():
-    pubkeys = [bls.G2Basic.PrivToPub(privkey) for privkey in PRIVKEYS]
+    pubkeys = [bls. G2ProofOfPossession.PrivToPub(privkey) for privkey in PRIVKEYS]
     pubkeys_serial = ['0x' + pubkey.hex() for pubkey in pubkeys]
     for privkey, pubkey_serial in zip(PRIVKEYS, pubkeys_serial):
         yield f'priv_to_pub_{int_to_hex(privkey)}', {
--- a/tests/generators/bls/main.py
+++ b/tests/generators/bls/main.py
@@ -20,6 +20,7 @@ def hash(x):
 
 F2Q_COEFF_LEN = 48
 G2_COMPRESSED_Z_LEN = 48
+DST = b'BLS_SIG_BLS12381G2-SHA256-SSWU-RO_POP_'
 
 
 def int_to_hex(n: int, byte_length: int = None) -> str:
@@ -27,6 +28,13 @@ def int_to_hex(n: int, byte_length: int = None) -> str:
     if byte_length:
         byte_value = byte_value.rjust(byte_length, b'\x00')
     return encode_hex(byte_value)
+
+
+def int_to_bytes(n: int, byte_length: int = None) -> bytes:
+    byte_value = int_to_big_endian(n)
+    if byte_length:
+        byte_value = byte_value.rjust(byte_length, b'\x00')
+    return byte_value
 
 
 def hex_to_int(x: str) -> int:
@@ -57,13 +65,11 @@ PRIVKEYS = [
 ]
 
 
-def hash_message(msg: bytes,
-                 domain: bytes) -> Tuple[Tuple[str, str], Tuple[str, str], Tuple[str, str]]:
+def hash_message(msg: bytes) -> Tuple[Tuple[str, str], Tuple[str, str], Tuple[str, str]]:
     """
     Hash message
     Input:
         - Message as bytes32
-        - domain as bytes8
     Output:
         - Message hash as a G2 point
     """
@@ -72,49 +78,44 @@ def hash_message(msg: bytes,
             int_to_hex(fq2.coeffs[0], F2Q_COEFF_LEN),
             int_to_hex(fq2.coeffs[1], F2Q_COEFF_LEN),
         ]
-        for fq2 in bls.utils.hash_to_G2(msg, domain)
+        for fq2 in bls.hash_to_curve.hash_to_G2(msg, DST)
     ]
 
 
-def hash_message_compressed(msg: bytes, domain: bytes) -> Tuple[str, str]:
+def hash_message_compressed(msg: bytes) -> Tuple[str, str]:
     """
     Hash message
     Input:
         - Message as bytes32
-        - domain as bytes8
     Output:
         - Message hash as a compressed G2 point
     """
-    z1, z2 = bls.utils.compress_G2(bls.utils.hash_to_G2(msg, domain))
+    z1, z2 = bls.point_compression.compress_G2(bls.hash_to_curve.hash_to_G2(msg, DST))
     return [int_to_hex(z1, G2_COMPRESSED_Z_LEN), int_to_hex(z2, G2_COMPRESSED_Z_LEN)]
 
 
 def case01_message_hash_G2_uncompressed():
     for msg in MESSAGES:
-        for domain in DOMAINS:
-            yield f'uncom_g2_hash_{encode_hex(msg)}_{encode_hex(domain)}', {
-                'input': {
-                    'message': encode_hex(msg),
-                    'domain': encode_hex(domain),
-                },
-                'output': hash_message(msg, domain)
-            }
+        yield f'uncom_g2_hash_{encode_hex(msg)}', {
+            'input': {
+                'message': encode_hex(msg),
+            },
+            'output': hash_message(msg)
+        }
 
 
 def case02_message_hash_G2_compressed():
     for msg in MESSAGES:
-        for domain in DOMAINS:
-            yield f'com_g2_hash_{encode_hex(msg)}_{encode_hex(domain)}', {
-                'input': {
-                    'message': encode_hex(msg),
-                    'domain': encode_hex(domain),
-                },
-                'output': hash_message_compressed(msg, domain)
-            }
+        yield f'com_g2_hash_{encode_hex(msg)}', {
+            'input': {
+                'message': encode_hex(msg),
+            },
+            'output': hash_message_compressed(msg)
+        }
 
 
 def case03_private_to_public_key():
-    pubkeys = [bls.privtopub(privkey) for privkey in PRIVKEYS]
+    pubkeys = [bls.G2Basic.PrivToPub(privkey) for privkey in PRIVKEYS]
     pubkeys_serial = ['0x' + pubkey.hex() for pubkey in pubkeys]
     for privkey, pubkey_serial in zip(PRIVKEYS, pubkeys_serial):
         yield f'priv_to_pub_{int_to_hex(privkey)}', {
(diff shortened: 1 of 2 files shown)


djrtwo:
cc: @ChihChengLiang

ChihChengLiang:
I'd lean toward modifying the test case to test `AggregateVerify` and `FastAggregateVerify`. In short-run, client teams need to update their tests. But the tests should reflect on the specified functions.

tests/generators/bls/main.py
```diff
@@ -29,6 +30,13 @@ def int_to_hex(n: int, byte_length: int = None) -> str:
     return encode_hex(byte_value)
 
 
+def int_to_bytes(n: int, byte_length: int = None) -> bytes:
+    byte_value = int_to_big_endian(n)
+    if byte_length:
+        byte_value = byte_value.rjust(byte_length, b'\x00')
```
mratsim: I'm concerned about the new null-terminated strings.
I think a comment should be added that it creates null-terminated string
only for the BLS spec (https://github.com/ethereum/eth2.0-specs/issues/1576)

tests/generators/bls/main.py
```diff
@@ -20,6 +20,7 @@ def hash(x):
 
 F2Q_COEFF_LEN = 48
 G2_COMPRESSED_Z_LEN = 48
+DST = b'BLS_SIG_BLS12381G2-SHA256-SSWU-RO_POP_'
```
benjaminion: For consistency with the spec, this ought to be `BLS_SIG_BLS12381G2-SHA256-SSWU-RO-_POP_` (note the extra `-` before `_POP_`).

tests/generators/bls/main.py
benjaminion: Everything is octet strings, which are _not_ null terminated, including the DST.

The zero bytes here are just padding to length (iiuc they are added to the left in big-endian, so they are prepended, not appended, in any case).

djrtwo:
added the following test cases:

* `case05_verify_message` handling valid and invalid cases of `bls.Verify`
*  `case08_fast_aggregate_verify` handling valid and invalid cases of `bls.FastAggregateVerify`
* `case09_aggregate_verify` handling valid and invalid cases of `bls.AggregateVerify`

I also kept in `case07_aggregate_pubkeys` as it is an independent feature that might very well be used by a validator/tooling in some cases.

ready for additional review @ChihChengLiang

tests/generators/bls/main.py
```diff
@@ -72,49 +78,44 @@ def hash_message(msg: bytes,
...
-            }
+        yield f'com_g2_hash_{encode_hex(msg)}', {
+            'input': {
+                'message': encode_hex(msg),
+            },
+            'output': hash_message_compressed(msg)
+        }
 
 
 def case03_private_to_public_key():
```
CarlBeek: I am not convinced that we should be explicitly checking this. While I exposed this functionality in `py_ecc`, the BLS specs don't feature an implementation of this function. Maybe I should have made it a "private" method.

tests/generators/bls/main.py
```diff
@@ -72,49 +78,44 @@ def hash_message(msg: bytes,
...
-    pubkeys = [bls.privtopub(privkey) for privkey in PRIVKEYS]
+    pubkeys = [bls.G2Basic.PrivToPub(privkey) for privkey in PRIVKEYS]
```
CarlBeek:
```suggestion
    pubkeys = [bls. G2ProofOfPossession.PrivToPub(privkey) for privkey in PRIVKEYS]
```

Not that it matters from a functionality perspective, but I recon it should be `G2ProofOfPossession` everywhere as that is the specific ciphersuite implemented.

tests/generators/bls/main.py
```diff
@@ -72,49 +78,44 @@ def hash_message(msg: bytes,
...
-        - domain as bytes8
     Output:
         - Message hash as a compressed G2 point
     """
-    z1, z2 = bls.utils.compress_G2(bls.utils.hash_to_G2(msg, domain))
+    z1, z2 = bls.point_compression.compress_G2(bls.hash_to_curve.hash_to_G2(msg, DST))
     return [int_to_hex(z1, G2_COMPRESSED_Z_LEN), int_to_hex(z2, G2_COMPRESSED_Z_LEN)]
 
 
 def case01_message_hash_G2_uncompressed():
```
CarlBeek: Is this (and `hash_to_G2_compressed`) a meaningful test case. I imagine (and hope) that most BLS implementations won't expose the "raw" hash-to-curve functionality.

tests/generators/bls/main.py
djrtwo: Is this not an exceedingly common operation that we want to test? -- the determinism of moving from a secret integer to a pubkey?

I suppose I might be missing something

tests/generators/bls/main.py
djrtwo: I don't feel very strongly. These tests are really to ensure that you've integrated and configured your BLS library properly before getting lost in pyspec tests.

Curious to hear @benjaminion or @mratsim's opinion on this

benjaminion: On the whole, I think that the reference tests should test only things used directly in the spec. Implementation-specific unit tests should take care of anything lower-level. So I'd vote for omitting the aggregate pubkeys tests, as well as the hash-to-G2 tests. The remaining BLS reference tests implicitly test these things in any case.

benjaminion: I should add a concrete rationale for the above, which is that it's undesirable to have to expose the inner workings of the BLS stuff just to satisfy the reference tests. If we can avoid doing that then it's easier to keep everything nicely encapsulated.

djrtwo: agree all around. Only the 5 specified BLS functions are now directly tested in this suite

Would love a last set of review @ChihChengLiang  and/or @CarlBeek.
The latest commit is simply removing the non-api related tests

CarlBeek: My opinions are basically summarised by @benjaminion's comments above. Only test the high level API functionality that we use in the spec offered by a "standard" implementation of the specs.

djrtwo:
Sorry, accidentally didn't push my local commit with the changes. Thanks for the review

tests/generators/bls/main.py
```diff
@@ -72,101 +78,160 @@ def hash_message(msg: bytes,
             int_to_hex(fq2.coeffs[0], F2Q_COEFF_LEN),
             int_to_hex(fq2.coeffs[1], F2Q_COEFF_LEN),
         ]
-        for fq2 in bls.utils.hash_to_G2(msg, domain)
+        for fq2 in bls.hash_to_curve.hash_to_G2(msg, DST)
     ]
 
 
-def hash_message_compressed(msg: bytes, domain: bytes) -> Tuple[str, str]:
+def hash_message_compressed(msg: bytes) -> Tuple[str, str]:
```
CarlBeek: This function is no longer needed and can be deleted.

tests/generators/bls/main.py
```diff
@@ -72,101 +78,160 @@ def hash_message(msg: bytes,
...
-def case02_message_hash_G2_compressed():
-    for msg in MESSAGES:
-        for domain in DOMAINS:
-            yield f'com_g2_hash_{encode_hex(msg)}_{encode_hex(domain)}', {
+def case02_verify():
+    for i, privkey in enumerate(PRIVKEYS):
+        for message in MESSAGES:
+            # Valid signature
+            signature = bls.G2ProofOfPossession.Sign(privkey, message)
+            pubkey = bls.G2Basic.PrivToPub(privkey)
```
CarlBeek:
```suggestion
            pubkey = bls. G2ProofOfPossession.PrivToPub(privkey)
```

IMO, it's better to be consistent with the cipher suite we are using.

tests/generators/bls/main.py
```diff
@@ -72,101 +78,160 @@ def hash_message(msg: bytes,
...
+                    'pubkey': encode_hex(pubkey),
+                    'message': encode_hex(message),
+                    'signature': encode_hex(signature),
                 },
-                'output': hash_message_compressed(msg, domain)
+                'output': True,
             }
 
+            # Invalid signatures -- wrong pubkey
+            wrong_pubkey = bls.G2Basic.PrivToPub(PRIVKEYS[(i + 1) % len(PRIVKEYS)])
```
CarlBeek:
```suggestion
            wrong_pubkey = bls.G2ProofOfPossession.PrivToPub(PRIVKEYS[(i + 1) % len(PRIVKEYS)])
```
ditto

tests/generators/bls/main.py
```diff
@@ -72,101 +78,160 @@ def hash_message(msg: bytes,
...
 
 
-# TODO: case05_verify_messages: Verify messages signed in case04
-# It takes too long, empty for now
+def case04_fast_aggregate_verify():
+    for i, message in enumerate(MESSAGES):
+        privkeys = PRIVKEYS[:i + 1]
+        sigs = [bls.G2ProofOfPossession.Sign(privkey, message) for privkey in privkeys]
+        aggregate_signature = bls.G2ProofOfPossession.Aggregate(sigs)
+        pubkeys = [bls.G2Basic.PrivToPub(privkey) for privkey in privkeys]
```
CarlBeek:
```suggestion
        pubkeys = [bls.G2ProofOfPossession.PrivToPub(privkey) for privkey in privkeys]
```

tests/generators/bls/main.py
```diff
@@ -72,101 +78,160 @@ def hash_message(msg: bytes,
...
+            'input': {
+                'pubkeys': pubkeys_serial,
+                'message': encode_hex(message),
+                'signature': encode_hex(aggregate_signature),
+            },
+            'output': True,
+        }
 
+        # Invalid signature -- extra pubkey
+        pubkeys_extra = pubkeys + [bls.G2Basic.PrivToPub(PRIVKEYS[-1])]
```
CarlBeek:
```suggestion
        pubkeys_extra = pubkeys + [bls.G2ProofOfPossession.PrivToPub(PRIVKEYS[-1])]
```

tests/generators/bls/main.py
```diff
@@ -72,101 +78,160 @@ def hash_message(msg: bytes,
...
-    pubkeys_serial = [encode_hex(pubkey) for pubkey in pubkeys]
-    yield f'agg_pub_keys', {
-        'input': pubkeys_serial,
-        'output': encode_hex(bls.aggregate_pubkeys(pubkeys)),
+def case05_aggregate_verify():
+    pairs = []
+    sigs = []
+    for privkey, message in zip(PRIVKEYS, MESSAGES):
+        sig = bls.G2ProofOfPossession.Sign(privkey, message)
+        pubkey = bls.G2Basic.PrivToPub(privkey)
```
CarlBeek:
```suggestion
        pubkey = bls.G2ProofOfPossession.PrivToPub(privkey)
```

ChihChengLiang:
Oops, an hour late, never mind 😂

djrtwo:
Thanks! Fixed in `v010x`

PR #1577, opened 2020-01-19
Ratelimit

mcdee:
This adds a response code of "rate limited" for peers.

Prysm has already seen situations where a peer is swamped for requests for blocks, and has implemented a rate limiter in an attempt to alleviate this.  At current it sends back a "server error" response when replying, but this isn't really the case and causes knock-on effects such as the requestor deciding the peer is non-functional, rather than just busy.

By adding a "rate limited" response code peers can understand the difference between these two situations and act accordingly.

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -371,10 +371,11 @@ The response code can have one of the following values, encoded as a single unsi
 -  0: **Success** -- a normal response follows, with contents matching the expected message schema and encoding specified in the request.
 -  1: **InvalidRequest** -- the contents of the request are semantically invalid, or the payload is malformed, or could not be understood. The response payload adheres to the `ErrorMessage` schema (described below).
 -  2: **ServerError** -- the responder encountered an error while processing the request. The response payload adheres to the `ErrorMessage` schema (described below).
+-  3: **RateLimited** -- the responder has served too many requests or sent too much data recently, either overall or to the requestor specifically. The response payload adheres to the `ErrorMessage` schema (described below).
 
 Clients MAY use response codes above `128` to indicate alternative, erroneous request-specific responses.
 
-The range `[3, 127]` is RESERVED for future usages, and should be treated as error if not recognized expressly.
+The range `[4, 127]` is RESERVED for future usages, and should be treated as error if not recognized expressly.
 
 The `ErrorMessage` schema is:
 


prestonvanloon:
Prysm uses error code 0x1 (invalid request) with the status message of "rate limited". While this works, it would be ideal to have a specific error code like @mcdee mentioned. Perhaps something analogous to [429 Too Many Requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429)?

prestonvanloon:
Even better: additionally the serving peer can send a specific rate limit error that indicates when the client should retry the request.

AgeManning:
We are doing a similar thing with a custom error response. 

I think it's useful to add this in. I have no objection to making the internal value an integer that specifies the time to re-request, which could also be handy.

arnetheduck:
I'm curious to hear more about the scenario under which spamming happens. 

Generally, there seems to be little benefit for a non-malicious client to have more than 1-2 requests in-flight - when syncing, you can scale the requested range by starting small and going a little larger - this allows you to fully saturate whatever bandwidth you have with the serving peer without flooding it. The server can then rate-limit by trickling responses at an appropriate speed, and make sure it's not overwhelmed by monitoring response times against the protocol timeouts - if this is happening, perhaps it simply shouldn't allow the connection to happen?

As a server, the way to handle flooding then becomes to keep a per-peer bounded request queue - as soon as that request queue goes out of bounds, you're likely dealing with a malicious peer and can drop them without error message.

As a client, you already need to write code that works well with servers that don't give the correct response code - maliciously or not. 

The added error response comes with an expectation of more complex behaviour of peers to correctly implement it - on balance, I'd prefer to see it more strongly motivated before adding it to the protocol.

jannikluhn:
For debugging I think this might be helpful, but doesn't have much benefit to an error response with appropriate status message. In production, I feel we'd need to include some details about what the rate limit actually is to make this useful (e.g. messages/kB/blocks per second to request) so that the peer can react by limiting its requests properly. Since it's probably difficult for implementations to come up with hard numbers (and making them public might actually a bad idea), I don't think a separate rate limited response adds a lot of value.

mcdee:
> I'm curious to hear more about the scenario under which spamming happens.

Usually when a single node is serving multiple syncing clients, either a low number of requests for a large number of blocks each, or a larger number of requests for a small number of blocks each; the point is total blocks requested per second is high.

> As a server, the way to handle flooding then becomes to keep a per-peer bounded request queue - as soon as that request queue goes out of bounds, you're likely dealing with a malicious peer and can drop them without error message.

We don't find that to be the case in the prysm testnet; well-behaved clients can end up asking us for a lot of blocks if we're the only node they can find.

> The added error response comes with an expectation of more complex behaviour of peers to correctly implement it

The behaviour is already going in to implementations.  @AgeManning mentioned that they are using custom error responses, and prysm will do likewise.  Chances are these will be incompatible (if not immediately then they may diverge in future) so I'd rather see a standard error code and a "correct" to implement.

mcdee:
>  In production, I feel we'd need to include some details about what the rate limit actually is to make this useful

I don't think this is the case.  In its simplest form a client receiving a rate limit response can back off for whatever time they themselves choose.  The main point is that the client can differentiate between a peer that is "bad" in some form, and one that is just busy.

If people are worried about the idea of handling rate limiting specifically, would a "transient error" response be better _i.e._ a response of "we could serve this request but won't at the moment"?  It could cover more situations in future and still carry the same basic meaning and similar client reaction.

arnetheduck:
> the point is total blocks requested per second is high.

the server is free to serve as many responses as it likes - it can already use this to rate limit peers that are requesting too much - in particular, it is allowed to answer with fewer blocks than the peer requested if the request is excessive. this is not an error, and thus doesn't require an error response.

> The behaviour

I'm still curious to hear more about this behaviour, and how it is correct in the face of malicious servers and peers, and how the network benefits from reacting to a custom error message, keeping in mind that switching to another peer is a cheap and expected operation in a p2p network (ie if a peer is overloaded, you should go look for another peer).

mcdee:
> the server is free to serve as many responses as it likes - it can already use this to rate limit peers that are requesting too much - in particular, it is allowed to answer with fewer blocks than the peer requested if the request is excessive. this is not an error, and thus doesn't require an error response.

This is a great example of why an error response would be a good thing.  If I ask a peer for some blocks and it returns an empty array, how am I meant to know if it has done so because the slots didn't have blocks produced, or because I'm rate limited?  Having an explicit rate limited response would provide certainty.

> ...if a peer is overloaded...

and how would we know it was overloaded without a response telling us this?

The behaviour at current is to send the generic error message to peers that request too many blocks from us in too short a time and cut them off.  If the peer is well-behaved it may go bother someone else for blocks; great.  If it is not, or it is genuinely short of peers, it may ask us again for the same information, and get much the same response (eventually we'll either get fed up with it and ban it, or our limiting will decay to the point we'll return it the requested blocks).

If the rate limit response was ignored by a badly-behaved client we'd carry on much as we do now.  If the client did back off for a bit it would have less chance of being banned next time it asked us for blocks.

I don't know what a malicious server looks like in this situation compared to a malicious server without rate limiting, I'm afraid.

> ...switching to another peer is a cheap and expected operation...

I also think you underestimate the real-world difficulty of finding and maintaining peer connections when many peers will be running behind NATted networks and commonly firewalled off.  I spun up a new node to see how it would manage in terms of peers; after 10 minutes it has 7 potential peers with which to sync (not all of whom will actually have all the required blocks).  There are also 30 peers it has heard of through discovery but failed connect with.  Finding enough peers with which to sync is never as easy as one would hope, and this would give us more of a chance of keeping the ones we find.

arnetheduck:
> This is a great example of why an error response would be a good thing.

from a client perspective, you have to handle the case where a peer sends an incomplete response regardless of the reason - it may inform you that you're being rate limited but it may also not. thus it doesn't really matter which of the two cases it is - you can try again, or not, depending on what your peer table looks like for example by scoring the peer.

> provide certainty.

if a block exists in the range, the peer is required to respond with at least one. if later you notice that the peer was withholding blocks (by analysing parent of successor blocks), you know it's not reliable. this does not relieve you from implementing logic for handling gaps however (ie rerequesting) - there is no such thing as certainty with remote peers and varying network conditions. the requirement is there in the spec mainly because we could think of no useful cases for completely empty responses when there are blocks, and it's a nice enough heuristic to drop misbehaving clients.

> how would we know it was overloaded

the server will drop you and not accept your attempt at reconnecting.

> it may ask us again

it can do so regardless - if the server is unable to respond with 1 block / 10 seconds (or whatever the response timeout currently is), it's simply not useful and should be descored accordingly based on whatever block rate it's effectively giving you. if you're unable to send blocks to a peer at a reasonable rate (some multiple or fraction of block production rate), you should probably drop that peer because they're a burden on the network in general in that they'll never be able to sync - better give another peer a chance.

> maintaining peer connections

you have to consider why you want to maintain these  poor connections over time when blocks cannot be sent over at a reasonable rate - every connection is a cost that must be motivated by some benefit - if you can't run sync, is the peer worth the broadcast cost?

arnetheduck:
> Software bugs.

if software bugs cause spamming to happen, would it not be better to disconnect them? At that point they are unduly burdening the network with excessive requests.

> peer simply returns zero blocks

the only time this happens is when the requested range has no blocks (because of skipped slots) - if a block exists in the range, the server has to send it within the protocol timeout. If it can't, the request fails with timeout - at that point, the value of the connection is questionable because the block transfer is slower than the advancement of the chain.

> I'm struggling to understand the hesitation on adding more status codes

With the status codes in there, there's an expectation that all clients will implement it, which requires effort, so it seems reasonable to get some value out of it, thus it's good to at least document the problem it solves for future implementers.

raulk:
Just some musings, in case it helps — feel free to disregard:

1. Introducing an error code may help the requestor discriminate if we're returning insufficient/partial data deliberately (because you've exceeded your quota and we've decided to decrease our level of service to you), or if we really don't have the data you've requested. In this sense, it could be more of a WARN code instead of an ERROR code, and it could be generalised into a _partial data_ warning code which communicates that the supplier of that data acknowledges there _may_ be more.

2. Usually, a rate limit response without any details on the limit being enforced, sucks. A rate limit by definition is temporary (otherwise we'd be talking about banning or blacklisting), and the whole reason you'd want to communicate a rate limit is in force is to allow the other peer to back off gracefully, but return to you some other time in the future, if they really need their request fulfilled _by you_ and can't find an alternative in the p2p network. If you don't tell them _when_ they can return and reasonably expect that their request will be fulfilled, you've lost the opportunity to leverage the whole adaptiveness property of making rate limits explicit by modelling them in the protocol.

---

Note: I'm a [chicken](https://en.wikipedia.org/wiki/The_Chicken_and_the_Pig) in this discussion -- I have no stake here.

djrtwo:
This is stale and clients are currently handling sync without a ratelimit

will resurface after mainnet launch if demand

PR #1579, opened 2020-01-21
Properly handle skip slots in fork choice

djrtwo:
Due to the way epoch pairs are justified/finalized and how we handle skipped slots in the consensus, we have to be careful to ensure not only that the justified/finalized block is in the chain, but that it is in the chain at the precise justified/finalized epoch (accounting for skip slots in exceptional cases)

Previously I erroneously put in conditions such the following:
```python
assert block.slot > compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)
```

This just ensured that the block was in fact later than the finalized epoch, but not that the finalized root was in the chain (potentially at a skip slot) at that epoch.

To fix this, we modify `get_ancestor` to return the latest root at or before the queried slot in the chain in question. This mimics the behavior of how we save roots in state and how we vote on roots, and it very nicely fixes this corner case.

-------

Helps properly handle cases in which we justify/finalize blocks after skipped epochs like in the following:

Screen Shot 2020-01-20 at 6 02 21 PM

Say block 64 at epoch 2 is finalized. Blocks that build off of block 64 at epoch 1 should _not_ be considered in the fork choice. the modificaiton to `get_ancestor` protect against this case

### DIFF
--- a/specs/phase0/fork-choice.md
+++ b/specs/phase0/fork-choice.md
@@ -279,9 +279,13 @@ def on_block(store: Store, signed_block: SignedBeaconBlock) -> None:
     assert get_current_slot(store) >= block.slot
     # Add new block to the store
     store.blocks[hash_tree_root(block)] = block
-    # Check block is a descendant of the finalized block
+
+    # Check that block is later than the finalized epoch slot (optimization to reduce calls to get_ancestor)
     finalized_slot = compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)
+    assert block.slot > finalized_slot
+    # Check block is a descendant of the finalized block
     assert get_ancestor(store, hash_tree_root(block), finalized_slot) == store.finalized_checkpoint.root
+
     # Check the block is valid and compute the post-state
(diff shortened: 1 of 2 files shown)


specs/phase0/fork-choice.md
```diff
@@ -284,12 +280,8 @@ def on_block(store: Store, signed_block: SignedBeaconBlock) -> None:
     # Add new block to the store
     store.blocks[hash_tree_root(block)] = block
     # Check block is a descendant of the finalized block
-    assert (
-        get_ancestor(store, hash_tree_root(block), store.blocks[store.finalized_checkpoint.root].slot) ==
-        store.finalized_checkpoint.root
-    )
-    # Check that block is later than the finalized epoch slot
-    assert block.slot > compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)
```
protolambda: Is it worth keeping this line, as a way to reduce possible calls to `get_ancestor`, which is a more expensive operation?

djrtwo: added and noted it is an optimization

PR #1580, opened 2020-01-21
Fork choice fix issue 1574

### LINKED ISSUE #1574: Fork choice: justified_epoch <= finalized_epoch
In fork choice, I believe it's possible for `store.justified_checkpoint.epoch <= store.finalized_checkpoint.epoch`. As I understand it, this means that `filter_block_tree` will filter out all blocks.

Consider that we're syncing a fork and in the later `SLOTS_PER_EPOCH - SAFE_SLOTS_TO_UPDATE_JUSTIFIED` slots of the epoch. During this time we sync through several epochs, where each epoch justifies and finalizes the closest possible prior epochs (i.e., perfect validator participation).

As we call `on_block` we're delaying updating `store.current_justified_epoch` because we're in the later slots of the epoch. However, we _are_ updating `store.finalized_checkpoint`. It follows that we could now have a `store.finalized_epoch` that is from a block several epochs _later_ than `store.justified_epoch`.

djrtwo:
Fixes #1574 

Ensure justified checkpoint in store updates when finalized checkpoint updates if either of the following:
* new justified is higher epoch than previous in store
* previous justified checkpoint in store is not in the chain with the new finalized checkpoint

Built off of branch `epochs-in-fork-choice` to use the fixed `get_ancestor` logic. Will switch to `v010x` once #1579 is merged

### DIFF
--- a/specs/phase0/fork-choice.md
+++ b/specs/phase0/fork-choice.md
@@ -301,6 +301,15 @@ def on_block(store: Store, signed_block: SignedBeaconBlock) -> None:
     # Update finalized checkpoint
     if state.finalized_checkpoint.epoch > store.finalized_checkpoint.epoch:
         store.finalized_checkpoint = state.finalized_checkpoint
+        finalized_slot = compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)
+
+        # Update justified if new justified is later than store justified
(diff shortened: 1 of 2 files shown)


djrtwo:
cc @paulhauner

PR #1585, opened 2020-01-22
Per-committee anti-correlation penalties

vbuterin:
The basic rationale to have per-committee anti-correlation penalties is that in phases 1+, there is a lot of harm that can be caused by even a single slashed committee, and so the harm from multiple validators acting harmfully is best viewed as a percentage of the committees that those validators were part of, and not a percentage of the whole validator set.

This is a maximally simple proposal for accomplishing this.

Note that there may be better ways of doing this, for example identifying for every slashing that uses an attestation what committee that attestation was for, and scanning through that entire committee at that time, removing the need for an independent object. Also, this code should be moved into phase 1 after the merge of the phase 1 PR.

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -41,6 +41,7 @@
     - [`Attestation`](#attestation)
     - [`Deposit`](#deposit)
     - [`VoluntaryExit`](#voluntaryexit)
+    - [`MultiSlashingClaim`](#multislashingclaim)
   - [Beacon blocks](#beacon-blocks)
     - [`BeaconBlockBody`](#beaconblockbody)
     - [`BeaconBlock`](#beaconblock)
@@ -248,6 +249,7 @@ The following values are (non-configurable) constants used throughout the specif
 | `MAX_ATTESTATIONS` | `2**7` (= 128) |
 | `MAX_DEPOSITS` | `2**4` (= 16) |
 | `MAX_VOLUNTARY_EXITS` | `2**4` (= 16) |
+| `MAX_MULTI_SLASHING_CLAIMS` | `2**2` (= 4) |
 
 ### Domain types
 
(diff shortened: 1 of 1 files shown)


mcdee:
There is no time protection here for the slashed validators that I can see, which seems to lead to a couple of problems:

  -  If multiple validators have been slashed whilst in different committees, but at some point before exit find themselves in the same committee,  a `MultiSlashingClaim` can be submitted to punish the validators erroneously
  - There is also no time limit on `MultiSlashingClaim` so it would be possible to hunt back through committee assignments to find the "best" committee in terms of number of slashed validators and apply that punishment (and I *think* this would hold for all history, so if my validator was *ever* in a committee with validators that subsequently were slashed themselves my slashing penalty is forever higher)

Even with these issues fixed, this change increases the chance of the penalty for a zero-impact slashing being higher.  It would be nice to perhaps consider a quadratic curve for punishment to reduce slashings that have no impact on network health (given these are most likely to be user mistakes or software bugs).

Given this isn't an issue until phase  1 is it worth keeping things as they are for phase 0 and exploring models for per-committee slashings with a view to implement in phase 1?

dankrad:
>     * There is also no time limit on `MultiSlashingClaim` so it would be possible to hunt back through committee assignments to find the "best" committee in terms of number of slashed validators and apply that punishment (and I _think_ this would hold for all history, so if my validator was _ever_ in a committee with validators that subsequently were slashed themselves my slashing penalty is forever higher)

That does indeed seem like an important problem! It seems to me that you would have to check that they were actually slashed for an offence committed with regards to that committee. So you would probably have to provide the slashing records for these validators of the MultiSlashingClaim; alternatively store this information for each slashed validator in the state.

This may make it more attractive to do the counting per shard as I suggested :) We could make it a bit more granular by counting per committee_index.

Also, shouldn't we reward the inclusiong for a MultiSlashingClaim?

djrtwo:
Closing for now. If we introduce something like this it will need to be in a subsequent phase

PR #1587, opened 2020-01-22
Deposit contract testing cleanup, manage python versions.

protolambda:
- Pin contract compiler to 3.7
- Tester supports any (py 3.8 support finally)
- Update makefile commands
- Separate venvs
- Separate compiler and tester code
- Update circle ci config to run compiler tests in 3.7, and all other in 3.8


.circleci/config.yml
```diff
@@ -35,32 +35,45 @@ commands:
...
+      - restore_cached_venv:
+          venv_name: v15-deposit-contract-tester
+          reqs_checksum: cache-{{ checksum "tests/core/pyspec/requirements.txt" }}-{{ checksum "deposit_contract/tester/requirements.txt" }}
+  save_deposit_contract_tester_cached_venv:
+    description: "Save the venv to cache for later use of the deposit contract tester"
+    steps:
+      - save_cached_venv:
+          venv_name: v15-deposit-contract-tester
+          reqs_checksum: cache-{{ checksum "tests/core/pyspec/requirements.txt" }}-{{ checksum "deposit_contract/tester/requirements.txt" }}
+          venv_path: ./deposit_contract/tester/venv
```
hwwhww: Not due to this PR, but I recalled that sometimes we needed to kick the`deposit_contract_cached_venv` cache by updating `venv_name` version manually. It's probably because `deposit-contract-tester` depends on `pyspec`, but we couldn't get **the `pyspec` version** in the checksum.

Will we have an explicit ` pyspec` package version in #1584?

tests/core/pyspec/setup.py
```diff
@@ -11,5 +11,6 @@
         "py_ecc==2.0.0",
         "ssz==0.1.3",
         "dataclasses==0.6",
+        "pytest"
```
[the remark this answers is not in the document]
protolambda: Right, tests wouldn't run in CI somehow, and so I put this in quickly. New packaging PR changes it anyway.

.circleci/config.yml
protolambda: Although packaging works nicely, and it can be pulled in as git dependency with pip, I'm not sure about versioning of the pyspec itself yet. Putting in the spec version could work, but staying on 0.0.0, with git as versioning, serves all current use cases for now fine.

PR #1591, opened 2020-01-23
Correct human-readable time

mcdee:
Follow-on from #1588; the constant was renamed but the accompanying human-readable time was still incorrect as it had not been updated when `SECONDS_PER_SLOT` went from 6 to 12.


djrtwo:
Thanks! cherry picked the fix directly to `v010x` for the impending release. will be backported to `dev` today

PR #1606, opened 2020-02-04
RPC chunks with streaming SSZ decoding, snappy frames, and stricter DOS limits where possible.

protolambda:
This is a proposal to focus on length-encoding SSZ contents, enable streaming of chunk contents, and put stricter DOS limits in place.

**`ssz` without compression is still exactly the same, testnets can live on**.

So far I am aware that Prysm has non-streaming Snappy compression in place as option, and Lighthouse has a PR open that made a start on snappy support, but blocked by questions/ongoing network changes by Adrian. In other clients, such as Lodestar, Artemis and Nimbus, I could find the Snappy dependency, but no implementation of `ssz_snappy` for RPC.

The motivation here is that it is relatively quick to calculate the encoded length of an SSZ value, since fixed-length values are easily computed from the type, and list values often from multiplying a fixed-length element size with a list length.

Given a SSZ length, a SSZ decoder can directly read contents from the stream, avoiding the need for a temporary buffer. Additionally, snappy-frames avoid the need for more than a single frame of buffered bytes to decompress data (worst case `2**16=65536` bytes), avoiding the need to buffer the compressed bytes fully.
The writer just calculates the SSZ length, and can then stream-encode the contents, even if using compression. Instead of having to fully buffer the compressed data to get the compressed length.

Additionally, this puts better DOS protection in place, as we can calculate the maximum size of a SSZ object, and should use it to be stricter about inputs. And given the worst-case snappy encoded length, we can account for compressed bytes even if we don't know the exact number of compressed bytes in advance. We don't need to know the exact number, as we can just verify the decompressed bytes instead, while checking the bytes read from the stream against the limits we should be checking regardless.

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -414,15 +414,15 @@ Since snappy frame contents [have a maximum size of `65536` bytes](https://githu
 
 **Encoding-dependent header:** Req/Resp protocols using the `ssz` or `ssz_snappy` encoding strategies MUST encode the length of the raw SSZ bytes, encoded as an unsigned [protobuf varint](https://developers.google.com/protocol-buffers/docs/encoding#varints). 
 
-*Writing*: By first computing and writing the SSZ byte length the SSZ encoder can then directly write the chunk contents to the stream.
+*Writing*: By first computing and writing the SSZ byte length, the SSZ encoder can then directly write the chunk contents to the stream.
 If Snappy is applied, it can be passed through a buffered Snappy writer to compress frame by frame.
 
 *Reading*: After reading the expected SSZ byte length, the SSZ decoder can directly read the contents from the stream.
 If snappy is applied, it can be passed through a buffered Snappy reader to decompress frame by frame.
 
 A reader:
-- SHOULD not read more than `max_encoded_len(n)` bytes (`32 + n + n/6`) after reading the SSZ length prefix `n` from the header, [this is considered the worst-case compression result by Snappy](https://github.com/google/snappy/blob/537f4ad6240e586970fe554614542e9717df7902/snappy.cc#L98).
-- SHOULD not accept a SSZ length prefix that is bigger than the expected maximum length for the SSZ type (derived from SSZ type information such as vector lengths and list limits).
+- SHOULD NOT read more than `max_encoded_len(n)` bytes (`32 + n + n // 6`) after reading the SSZ length prefix `n` from the header, [this is considered the worst-case compression result by Snappy](https://github.com/google/snappy/blob/537f4ad6240e586970fe554614542e9717df7902/snappy.cc#L98).
+- SHOULD NOT accept an SSZ length prefix that is bigger than the expected maximum length for the SSZ type (derived from SSZ type information such as vector lengths and list limits).
 - MUST consider remaining bytes, after having read the `n` SSZ bytes, as an invalid input. An EOF is expected.
 - MUST consider an early EOF, before fully reading the declared length prefix worth of SSZ bytes, as an invalid input.
 
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -420,11 +420,20 @@ If Snappy is applied, it can be passed through a buffered Snappy writer to compr
 *Reading*: After reading the expected SSZ byte length, the SSZ decoder can directly read the contents from the stream.
 If snappy is applied, it can be passed through a buffered Snappy reader to decompress frame by frame.
 
-A reader:
-- SHOULD NOT read more than `max_encoded_len(n)` bytes (`32 + n + n // 6`) after reading the SSZ length prefix `n` from the header, [this is considered the worst-case compression result by Snappy](https://github.com/google/snappy/blob/537f4ad6240e586970fe554614542e9717df7902/snappy.cc#L98).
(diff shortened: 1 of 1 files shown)


protolambda:
@arnetheduck @AgeManning @nisdas Please have a look at this proposal. The PR description may not be that clear, the updated networking doc explains it best.

specs/phase0/p2p-interface.md
```diff
@@ -403,9 +403,28 @@ The token of the negotiated protocol ID specifies the type of encoding to be use
...
+*Writing*: By first computing and writing the SSZ byte length the SSZ encoder can then directly write the chunk contents to the stream.
+If Snappy is applied, it can be passed through a buffered Snappy writer to compress frame by frame.
+
+*Reading*: After reading the expected SSZ byte length, the SSZ decoder can directly read the contents from the stream.
+If snappy is applied, it can be passed through a buffered Snappy reader to decompress frame by frame.
+
+A reader:
+- SHOULD not read more than `max_encoded_len(n)` bytes (`32 + n + n/6`) after reading the SSZ length prefix `n` from the header, [this is considered the worst-case compression result by Snappy](https://github.com/google/snappy/blob/537f4ad6240e586970fe554614542e9717df7902/snappy.cc#L98).
+- SHOULD not accept a SSZ length prefix that is bigger than the expected maximum length for the SSZ type (derived from SSZ type information such as vector lengths and list limits).
+- MUST consider remaining bytes, after having read the `n` SSZ bytes, as an invalid input. An EOF is expected.
```
nisdas: This should be clarified, if we read more than `n` bytes, we should return an error on the request/response and ignore the data.

specs/phase0/p2p-interface.md
```diff
@@ -403,9 +403,28 @@ The token of the negotiated protocol ID specifies the type of encoding to be use
...
-The [SimpleSerialize (SSZ) specification](../../ssz/simple-serialize.md) outlines how objects are SSZ-encoded. If the Snappy variant is selected, we feed the serialized form to the Snappy compressor on encoding. The inverse happens on decoding.
+The [SimpleSerialize (SSZ) specification](../../ssz/simple-serialize.md) outlines how objects are SSZ-encoded.
 
-**Encoding-dependent header:** Req/Resp protocols using the `ssz` or `ssz_snappy` encoding strategies MUST prefix all encoded and compressed (if applicable) payloads with an unsigned [protobuf varint](https://developers.google.com/protocol-buffers/docs/encoding#varints).
+If the Snappy variant is selected, we feed the serialized form of the object to the Snappy compressor on encoding. The inverse happens on decoding.
+
+Snappy has two formats: "block" and "frames" (streaming). To support large requests and response chunks, snappy-framing is used.
+
+Since snappy frame contents [have a maximum size of `65536` bytes](https://github.com/google/snappy/blob/master/framing_format.txt#L104)
+ and frame headers are just `identifier (1) + checksum (4)` bytes, the expected buffering of a single frame is acceptable.
```
nisdas: So what I understand is now we have two length prefixes encoded into our data ? 
one is for snappy frames in the frame header and the other for our ssz objects in their decompressed form ?

protolambda: Possibly, 1 Snappy frame doesn't necessarily match 1 eth2 rpc chunk. With status quo it does match one snappy block, but it also can't stream within a chunk.

AgeManning:
I think it's reasonable to use the framed format of snappy over the wire. It may not be super important for us now given current objects but could be more useful for larger objects in the future. 

In principle, we could have a `snappy_framed` and `snappy_block` protocols and clients could take preference over either, but i'm happy with this current proposal.

djrtwo:
I am pro frames.
@protolambda is there anything we need to do here to get this merged?

specs/phase0/p2p-interface.md
```diff
@@ -403,9 +403,28 @@ The token of the negotiated protocol ID specifies the type of encoding to be use
...
+
+**Encoding-dependent header:** Req/Resp protocols using the `ssz` or `ssz_snappy` encoding strategies MUST encode the length of the raw SSZ bytes, encoded as an unsigned [protobuf varint](https://developers.google.com/protocol-buffers/docs/encoding#varints). 
+
+*Writing*: By first computing and writing the SSZ byte length the SSZ encoder can then directly write the chunk contents to the stream.
```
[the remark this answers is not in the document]
hwwhww: Could you elaborate what do you mean by "the SSZ encoder" here? The encoder that encodes the SSZ bytes to chunks?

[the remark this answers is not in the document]
hwwhww: Could you elaborate on what's "the SSZ encoder" here? From its name, it feels like it's just doing the SSZ serialization, not the stream writer.
Also saw "stream-decode SSZ" and "SSZ-encoded" in different sentences, it would be nice to clarify it.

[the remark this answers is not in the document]
djrtwo: I read it as whatever is encoding (serializing) the SSZ objects can do so directly to the stream instead of a separate buffer

@protolambda thoughts here? I'd like to get this PR cleaned up

specs/phase0/p2p-interface.md
djrtwo: @protolambda can you clean this one up too?

protolambda: Yes, looking into this now.

protolambda:
Clarified the invalid input handling, and cleaned up to sum the different cases better. Can I get a final review?

PR #1608, opened 2020-02-05
Minor fixes in BLS tests

ChihChengLiang:
1. Fix directory name. Before this fix, it currently looks like

Screen Shot 2020-02-05 at 1 05 48 PM

2. Move human-readable names out from the hash

PR #1610, opened 2020-02-06
Bls nitpicks

djrtwo:
Replaces #1608

------- original header ---------------

1. Fix directory name. Before this fix, it currently looks like

Screen Shot 2020-02-05 at 1 05 48 PM

2. Move human-readable names out from the hash

### DIFF
--- a/tests/generators/bls/main.py
+++ b/tests/generators/bls/main.py
@@ -69,8 +69,8 @@ def case02_verify():
             # Valid signature
             signature = bls.G2ProofOfPossession.Sign(privkey, message)
             pubkey = bls.G2ProofOfPossession.PrivToPub(privkey)
-            full_name = f'{encode_hex(pubkey)}_{encode_hex(message)}_valid'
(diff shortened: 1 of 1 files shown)


tests/generators/bls/main.py
```diff
@@ -69,8 +69,8 @@ def case02_verify():
             # Valid signature
             signature = bls.G2ProofOfPossession.Sign(privkey, message)
             pubkey = bls.G2ProofOfPossession.PrivToPub(privkey)
-            full_name = f'{encode_hex(pubkey)}_{encode_hex(message)}_valid'
-            yield f'verify_case_{(hash(bytes(full_name, "utf-8"))[:8]).hex()}', {
+            full_name = f'{encode_hex(pubkey)}_{encode_hex(message)}'
```
ChihChengLiang: nitpick: What do you think renaming all the `full_name`s to `identifier`? They are no longer the full name actually.

djrtwo: will do!

PR #1614, opened 2020-02-10
add eth2 key/value ENR to phase 0 p2p

djrtwo:
[UPDATED]

* Add genesis_validators_root to BeaconState and utilize for better signature domain separation as well as ENR/chain separation. This strengthens chain separation such that one must only be concerned about isolating fork version from a different chain when a contentious fork. Networks such as testnets and completely different chains with different genesis conditions can reuse fork versioning with no worry
* add `eth2` ENR entry that maps to 16-byte `ENRForkID` (`current_fork_digest`, `next_fork_version`, `next_fork_epoch`
* add note on how clients should use this information to make connections.
  * One concern here is if a peer from a long-past fork version is trying to sync earlier blocks. In such a case, they would not be able to (easily) find peers.
  * On the surface this looks concerning, but in practice, this node would not be able to sync to the head and follow consensus anyway because their consensus rules are so out of date.
  * I am comfortable leaving as is, but open to further input.
* add FAQ entry to disambiguate how fork `Version` is planned to be used when forking a network

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -604,6 +604,12 @@ Nonetheless, ENRs MUST carry a generic `eth2` key with nil value, denoting that 
 
 #### Mainnet
 
+ENRs MUST carry a generic `eth2` with a 4-byte value of the node's current fork version to ensure connections are made with peers on the intended eth2 network.
+
+| Key          | Value               |
+|:-------------|:--------------------|
+| `eth2`       | SSZ `Bytes4`        |
+
 On mainnet, ENRs MUST include a structure enumerating the capabilities offered by the peer in an efficient manner. The concrete solution is currently undefined. Proposals include using namespaced bloom filters mapping capabilities to specific protocol IDs supported under that capability.
 
 ### Topic advertisement
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -615,23 +615,37 @@ ENRs MUST carry a generic `eth2` key with an 16-byte value of the node's current
 |:-------------|:--------------------|
 | `eth2`       | SSZ `ENRForkID`        |
 
-Specifically, the value of the `eth2` key MUST be the following SSZ encoded object (`ENRForkID`), where
+First we define `current_fork` as the following SSZ encoded object
 
 ```
 (
-    current_fork_version: Fork
-    next_fork_version: Fork
+    current_fork_version: Version
+    genesis_validators_root: Root
+)
+```
+
+where
+
+* `current_fork_version` is the fork version at the node's current epoch defined by the wall-clock time (not necessarily the epoch to which the node is sync)
+* `genesis_validators_root` is the static `Root` found in `state.genesis_validators_root`
+
+Specifically, the value of the `eth2` key MUST be the following SSZ encoded object (`ENRForkID`)
+
+```
+(
+    current_fork_digest: Bytes4
+    next_fork_version: Version
     next_fork_epoch: Epoch
 )
 ```
 
 where the fields of `ENRForkID` are defined as
 
-* `current_fork_version` is the fork version at the node's current epoch defined by the wall-clock time (not necessarily the epoch to which the node is sync)
+* `current_fork_digest` is `hash_tree_root(current_fork)[:4]`
 * `next_fork_version` is the fork version corresponding to the next planned hard fork at a future epoch. If no future fork is planned, set `next_fork_version = current_fork_version` to signal this fact
 * `next_fork_epoch` is the epoch at which the next fork is planned and the `current_fork_version` will be updated. If no future fork is planned, set `next_fork_epoch = FAR_FUTURE_EPOCH` to signal this fact
 
-Clients SHOULD connect to peers with `current_fork_version`, `next_fork_version`, and `next_fork_epoch` that match local values.
+Clients SHOULD connect to peers with `current_fork_digest`, `next_fork_version`, and `next_fork_epoch` that match local values.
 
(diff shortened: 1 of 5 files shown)


djrtwo:
cc @arnetheduck @AgeManning @nisdas

AgeManning:
The actual PR is generic, so looks fine to me. 

1) Seems to make sense to me. As all upgraded nodes that match a `fork-version` should also agree on all subsequent forks and are therefore compatible. The client maintains the canonical `fork-version` which should match the state after sync. 

When doing status messages, which fork version should be sent (and which fork version is in the ENR), the one from the state or a client-maintained `fork-version` which may include `next-fork`, or am i missing the mechanics of this?

protolambda:
From discussion earlier: how do we handle the pre-genesis case, where the root is not yet known? Default to `0x0000...` and accept that too? Very stable ENRs long before genesis would be good to have. If there is one time to have lots of connections to bootnodes, and in advance to get connected and move away from the bootnode, then it is genesis.

nisdas:
The PR looks good to me , 1) seems to be the simplest and makes the most sense to me. 
> (1) is sufficient to avoid collisions unless a network does not bump their fork version on an upgrade and the previous version of the net keeps running.

This is a safe assumption, I do not know if there would ever be a situation where there is a network upgrade without updating the fork version.
> From discussion earlier: how do we handle the pre-genesis case, where the root is not yet known? Default to 0x0000... and accept that too? Very stable ENRs long before genesis would be good to have. If there is one time to have lots of connections to bootnodes, and in advance to get connected and move away from the bootnode, then it is genesis.

For pre-genesis it would make sense to default with `0x0000...` or something like
```
hash(0x000000...)[:4]
```
either one is fine with me

djrtwo:
Made updates re: our conversation today. It isnt exactly what we discussed, but as I worked through, it seemed clean. I'm not yet set on this. Open to feedback

cc @arnetheduck @AgeManning @nisdas @protolambda

specs/phase0/p2p-interface.md
```diff
@@ -604,6 +607,50 @@ Nonetheless, ENRs MUST carry a generic `eth2` key with nil value, denoting that
...
+```
+
+where the fields of `ENRForkID` are defined as
+
+* `current_fork_digest` is `hash_tree_root(current_fork)[:4]`
+* `next_fork_version` is the fork version corresponding to the next planned hard fork at a future epoch. If no future fork is planned, set `next_fork_version = current_fork_version` to signal this fact
+* `next_fork_epoch` is the epoch at which the next fork is planned and the `current_fork_version` will be updated. If no future fork is planned, set `next_fork_epoch = FAR_FUTURE_EPOCH` to signal this fact
+
+Clients SHOULD connect to peers with `current_fork_digest`, `next_fork_version`, and `next_fork_epoch` that match local values.
+
```
nisdas: Should we also add a note that clients `SHOULD` connect to peers that have previous fork digests as their current fork digests ? Ex: Peers syncing from genesis/past epochs

specs/phase0/p2p-interface.md
```diff
@@ -823,6 +870,14 @@ In Phase 0, peers for attestation subnets will be found using the `attnets` entr
 
 Although this method will be sufficient for early phases of Eth2, we aim to use the more appropriate discv5 topics for this and other similar tasks in the future. ENRs should ultimately not be used for this purpose. They are best suited to store identity, location, and capability information, rather than more volatile advertisements.
 
+### How should fork version be used in practice?
+
+Fork versions are to be manually updated (likely via incrementing or using the less collision-prone git spec hash) at each hard fork. This is to provide native domain separation for signatures as well as to aid in usefulness for identitying peers (via ENRs) and versioning network protocols (e.g. using fork version to naturally version gossipsub topics).
```
```suggestion
Fork versions are to be manually updated (likely via incrementing or hash of the spec version) at each hard fork. This is to provide native domain separation for signatures as well as to aid in usefulness for identitying peers (via ENRs) and versioning network protocols (e.g. using fork version to naturally version gossipsub topics).
```
Using the spec or git is limiting, and I wouldn't say it is less collision prone, as not all forks may have the same kind of spec. I think it is better to just give a bare "hash spec version" example.

specs/phase0/p2p-interface.md
```diff
@@ -604,6 +607,50 @@ Nonetheless, ENRs MUST carry a generic `eth2` key with nil value, denoting that
...
+(
+    current_fork_digest: Bytes4
+    next_fork_version: Version
+    next_fork_epoch: Epoch
+)
+```
+
+where the fields of `ENRForkID` are defined as
+
+* `current_fork_digest` is `hash_tree_root(current_fork)[:4]`
```
protolambda: Should we make a fork-digest data-type? To make it obviously different from `Version`, given that both are a bytes4? Can also do cosmetics later if necessary.

specs/phase0/p2p-interface.md
djrtwo: With the genesis domain separation, I'm actually now just considering going back to "increment" with a suggestion to flip a high order bit in the event of a contentious fork.

specs/phase0/p2p-interface.md
djrtwo: Maybe so. I'll see if I can quickly, cleanly add this in

specs/phase0/p2p-interface.md
djrtwo: I've been back and forth on this. I think the answer is no in the general case.

Having a different fork digest means that the current slot wrt wall time, you disagree on fundamental consensus rules (even if it is just the fork version). From the perspective of the ENR, I can't tell if the client wrt sync is in the past or up to the current slot. If they are in the past, I could presumable serve them some blocks to get them up to a certain point, but then we will fork. In such a case, it is better for them to find peers that are actually on their chain (fork digest) and sync all the way to the head rather than me connecting to them to see if I can help them part of the way.

djrtwo:
Feedback addressed @protolambda

PR #1624, opened 2020-02-17
Add fork version to topic

arnetheduck:
Gossipsub peers are separate from the ETH2 RPC protocol, and thus cannot
rely on the application-level `Status` negotiation to establish if
they're on the same network.

Segregating gossipsub topics by fork version decouples RPC from gossip
further and allows peers to more easily listen only to the traffic of
the network they're interested in, without further negotiation.

this is a work-in-progress complement to #1614 - in particular, it's an open question which fork version should be used during upgrades - most natural seems to be to use the head fork version - or more accurately perhaps, the fork version as it would be when time has advanced to where the fork has taken effect.

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -763,9 +763,9 @@ For future extensibility with almost zero overhead now (besides the extra bytes 
 
 ### How do we upgrade gossip channels (e.g. changes in encoding, compression)?
 
-Changing gossipsub/broadcasts requires a coordinated upgrade where all clients start publishing to the new topic together, for example during a hard fork.
+Changing gossipsub/broadcasts requires a coordinated upgrade where all clients start publishing to the new topic together, during a hard fork.
 
-One can envision a two-phase deployment as well where clients start listening to the new topic in the first phase then start publishing some time later, letting the traffic naturally move over to the new topic.
+When a node is preparing for upcoming tasks (e.g. validator duty lookahead) on a gossipsub topic, the node should join the topic of the future epoch in which the task is to occur in addition to listening to the topics for the current epoch.
 
 ### Why must all clients use the same gossip topic instead of one negotiated between each peer pair?
 
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -216,7 +216,17 @@ The following gossipsub [parameters](https://github.com/libp2p/specs/tree/master
 
 ### Topics and messages
 
-Topics are plain UTF-8 strings and are encoded on the wire as determined by protobuf (gossipsub messages are enveloped in protobuf messages). Topic strings have form: `/eth2/TopicName/TopicEncoding`. This defines both the type of data being sent on the topic and how the data field of the message is encoded.
+Topics are plain UTF-8 strings and are encoded on the wire as determined by protobuf (gossipsub messages are enveloped in protobuf messages). Topic strings have form: `/eth2/ForkVersion/Name/Encoding`. This defines both the type of data being sent on the topic and how the data field of the message is encoded.
+
+- `ForkVersion` - the hex-encoded bytes from `state.fork.current_version` of the head state of the client, as also seen in `Status.head_fork_version`.
+- `Name` - see table below
+- `Encoding` - the encoding strategy describes a specific representation of bytes that will be transmitted over the wire. See the [Encodings](#Encoding-strategies) section for further details.
+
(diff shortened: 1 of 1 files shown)


specs/phase0/p2p-interface.md
```diff
@@ -216,7 +216,17 @@ The following gossipsub [parameters](https://github.com/libp2p/specs/tree/master
...
+
+- `ForkVersion` - the hex-encoded bytes from `state.fork.current_version` of the head state of the client, as also seen in `Status.head_fork_version`.
+- `Name` - see table below
+- `Encoding` - the encoding strategy describes a specific representation of bytes that will be transmitted over the wire. See the [Encodings](#Encoding-strategies) section for further details.
+
+The fork version is hex-encoded using the following scheme:
+```python
+  ForkVersion = ''.join('{:02x}'.format(x) for x in state.fork.current_version)
+```
+For example, the fork version `[0, 1, 2, 10]` will be encoded as `0001020a`.
```
djrtwo: Recommend we use the format/type as in the rest of the specs

```suggestion
For example, the fork version `Version('0x0001020a')` will be encoded as `0001020a`.
```

specs/phase0/p2p-interface.md
```diff
@@ -216,7 +216,17 @@ The following gossipsub [parameters](https://github.com/libp2p/specs/tree/master
 
 ### Topics and messages
 
-Topics are plain UTF-8 strings and are encoded on the wire as determined by protobuf (gossipsub messages are enveloped in protobuf messages). Topic strings have form: `/eth2/TopicName/TopicEncoding`. This defines both the type of data being sent on the topic and how the data field of the message is encoded.
+Topics are plain UTF-8 strings and are encoded on the wire as determined by protobuf (gossipsub messages are enveloped in protobuf messages). Topic strings have form: `/eth2/ForkVersion/Name/Encoding`. This defines both the type of data being sent on the topic and how the data field of the message is encoded.
+
+- `ForkVersion` - the hex-encoded bytes from `state.fork.current_version` of the head state of the client, as also seen in `Status.head_fork_version`.
```
djrtwo: Might be worth a note on guidance on gossip subnet lookahead wrt upcoming forks. Maybe a good note to add in the FAQs

Something like: How are gossipsub topics expected to be handled during forks?

> When a node is preparing for upcoming tasks (e.g. validator duty lookahead) on a gossipsub topic, the node should join the topic with the `ForkVersion` of the future epoch in which the task is to occur.

arnetheduck: we kind of had one that could be adapted

djrtwo: Might make sense to rephrase this in terms of fork version of the epoch of the relevant message, rather than that of the head state.

Because we might prep for and listen to multiple fork version topics of the same message type at once, it isn't _strictly_ related to the head state.

djrtwo: Maybe something like:
```suggestion
- `ForkVersion` - the hex-encoded bytes of the `fork_version` of the epoch of the related messages to be sent on the topic
```

specs/phase0/p2p-interface.md
```diff
@@ -216,7 +216,17 @@ The following gossipsub [parameters](https://github.com/libp2p/specs/tree/master
...
+- `Name` - see table below
+- `Encoding` - the encoding strategy describes a specific representation of bytes that will be transmitted over the wire. See the [Encodings](#Encoding-strategies) section for further details.
+
+The fork version is hex-encoded using the following scheme:
+```python
+  ForkVersion = ''.join('{:02x}'.format(x) for x in state.fork.current_version)
```
djrtwo:
```suggestion
  ForkVersion = ''.join('{:02x}'.format(x) for x in fork_version)
```

specs/phase0/p2p-interface.md
```diff
@@ -229,7 +239,7 @@ where `base64` is the [URL-safe base64 alphabet](https://tools.ietf.org/html/rfc
 
 The payload is carried in the `data` field of a gossipsub message, and varies depending on the topic:
 
-| Topic                                          | Message Type            |
+| Name                                           | Message Type            |
```
hwwhww: It seems we are still using `TopicName` to describe the field in the following. (Section `Global topics
` and `Attestation subnets`)

djrtwo:
closing in favor of #1652

specs/phase0/p2p-interface.md
djrtwo: updated this in #1652

PR #1625, opened 2020-02-18
Measure eth1 voting period in epochs instead of slots

Nashatyrev:
While debugging beacon chain code hit this minor spec issue: `SLOTS_PER_ETH1_VOTING_PERIOD` is measured in slots while the `beaconState.eth1_data_votes` list is reset during epoch processing. 
The condition for reset is 
```python
if (state.slot + 1) % SLOTS_PER_ETH1_VOTING_PERIOD == 0:
```

This implies that the `SLOTS_PER_ETH1_VOTING_PERIOD ` constant should be multiple of `SLOTS_PER_EPOCH`. May be it makes sense to nominate this constant in epochs to avoid malformed values for a custom genesis?


Nashatyrev:
I will apply changes to tests and phase-1 spec if this PR has any positive feedback

Nashatyrev:
Another option is to change the reset condition to 
```python
if len(state.eth1_data_votes) + SLOTS_PER_EPOCH > SLOTS_PER_ETH1_VOTING_PERIOD 
```
this way `eth1_data_votes` would not overflow during the next epoch, but that doesn't prevent from setting `SLOTS_PER_ETH1_VOTING_PERIOD < SLOTS_PER_EPOCH` by mistake in a custom config

ralexstokes:
> Anyone feel strongly about getting this into the next release or not?

if we are going to change it then the sooner the better :)

Issue #1601, opened 2020-01-28
Add `proposer_id` to BeaconBlock


djrtwo:
I didn't realize #1339 had been closed! Creating this issue for some fresh eyes. This is something I'd definitely like to revisit and maybe have auditors look at this month.

See #1339 for the original description of the issue.

The tldr is that without a `proposer_id` and a way for a quick signature verification on an incoming block, an attacker can create blocks with some amount of skip slots that require database look ups and state transitions _to even validate the proposer signature_. In the current case, that attacker doesn't even need to be in the validator set.

By adding the `proposer_id` we at least reduce the set that can craft such blocks to the validators and limit them to a 1 per slot allowance (due to [gossip conditions](https://github.com/ethereum/eth2.0-specs/issues/1595) and slashing conditions). This rate limit combined with peer scoring (and dropping!) at least begins to bound this problem.

protolambda:
You have my support for this change. And prefer `proposer_index` over `proposer_id` to be more consistent with the other places where we use validator indices.

terencechain:
+1 on this change. This will result in better client code

protolambda:
Closing this issue, the proposer id (index) has been introduced in #1626, to be featured in upcoming v0.11.0 release.

PR #1626, opened 2020-02-18
add proposer index to BeaconBlock

djrtwo:
Addresses #1601 

* add `proposer_index` to `BeaconBlock` and `BeaconBlockHeader`
* remove `proposer_index` from `ProposerSlashing` and instead assert that both headers have same proposer index
* add and modify tests

---------------

Comment from @protolambda below copied here for visibility

> LGTM, happy to see this! It will make cases that trust block-roots, but are unaware of the state, aware of proposers. This is good for tooling/bridges/etc.
>
> Note for implementers though: careful, don't trust blocks blindly, follow spec and verify the proposer index.

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -377,6 +377,7 @@ class DepositData(Container):
 ```python
 class BeaconBlockHeader(Container):
     slot: Slot
+    proposer_index: ValidatorIndex
     parent_root: Root
     state_root: Root
     body_root: Root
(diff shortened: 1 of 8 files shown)


djrtwo:
Sounds good. Added a couple of additional sanity block tests to cover the combinations on how the index might be messed up

hwwhww:
```python
def process_randao(state: BeaconState, body: BeaconBlockBody) -> None:
    # Verify RANDAO reveal
    ...
```

could be refactored to
```python
def process_randao(state: BeaconState, block: BeaconBlock) -> None:
    # Verify RANDAO reveal
    proposer = block.proposer_index
    randao_reveal = block.body.randao_reveal
    ...
```

djrtwo:
@hwwhww I decided to make the changes minimally invasive and to only check that the specified proposer index in the block matches the expected in state once `process_block_header`. Otherwise to keep everything to same. 

One place that could also be refactored is `process_attestation`, but this would involve changing the function signature.

Issue #1609, opened 2020-02-05
Unexpected multi-deposit incentive


JustinDrake:
See context from the 3rd AMA [here](https://www.reddit.com/r/ethereum/comments/ez972u/ama_we_are_the_eth_20_research_team_pt_3/fglqkfi/). The problem is that if someone deposits 32 ETH and early on their balance goes down (even just a bit) then they will have to wait until their balance goes back up to 32.5 ETH (because of hysteresis) for the effective balance to become 32 ETH. This could incentivise some validators to make a second deposit which is not ideal.

vbuterin:
The main solutions I see are:

1. Change the increment/decrement thresholds to x-0.25 and x+0.75
2. Allow deposits up to 32.5 ETH instead of 32.

JustinDrake:
> Allow deposits up to 32.5 ETH instead of 32.

Deposits above 32 ETH are allowed. Any size above 1 ETH is allowed.

Buttaa:
> Deposits above 32 ETH are allowed. Any size above 1 ETH is allowed.

1 ETH seems too high, if someone runs 32 validators and 1 ETH is the minimum, this would be an additional validator.

Buttaa:
There are two cases:
1) you deposit and don't get any early punishments (=don't need additional ETH)
2) you get punished early, and then you will need to deposit >1ETH
not the best UX

mcdee:
Secondary deposits have rather limited use cases, due to the minimum deposit being 1 Ether.

They come in to play if your balance is in the range `(x+0.5Ξ,x+1Ξ)` for some integer value of `x` between 16 and 30, as you can carry out a top-up to get yourself above the latch point for the next effective balance increment.  For example, if you had a balance of 29.9Ξ (and hence effective balance of 29Ξ) then a top-up deposit of 1.61Ξ would result in a balance of 31.51Ξ (and hence effective balance of 31Ξ).  For any other situation I believe that the effective balance won't increase further than the the increase in balance.

In general, though, a staker that finds themself in this position (_i.e._ with a balance less than 31Ξ)is unlikely to be sophisticated enough to carry out this operation (unless there have been some serious finality issues).

prestonvanloon:
[...]

PR #1627, opened 2020-02-18
adjust hysteresis to avoid initial over-deposit incentive

djrtwo:
Address #1609 

Make hysteresis calculations configurable.
Change default increment/decrement thresholds to x-0.25 and x+1.25


protolambda:
From chat earlier, we need to consider that if we do not offset the actual effective balance, we:
- increase allowance for mistakes by 0.25
- increase punishment by requiring another 0.25 extra

```
Previously:
Downwards hysteresis width: 32-32 = 0
Upwards hysteresis width: 32.5-31 = 1.5
Now:
Downwards hysteresis width: 32-31.75 = 0.25
Upwards hysteresis width: 32.75 - 31 = 1.75
```

I think the system is closer to original intention if we change it to:
- A small margin of e.g. 0.05 for initially offline validators to not immediately lose 1 effective balance increment because of some small inactivity.
- The old 1.5 upwards width.

mcdee:
What is the rationale behind x+0.75 rather than x+0.25?  A +/-0.25 would retain the current gap of requiring 0.5 Ether in rewards (or additional deposit) prior to an uptick in effective balance, and solves the issue of newly on-boarded deposits not suffering an immediate decrease in effective balance if they miss their first attestation.

vbuterin:
Yeah, it should be -0.25 and +1.25, not +1.75.

> Downwards hysteresis width: 32-31.75 = 0.25
> Upwards hysteresis width: 32.75 - 31 = 1.75

The error here is that as implemeneted the "downwards hysteresis width" (wrong term arguably) is actually _negative_ 0.25, and not 0.25. So preserving the 1.5 width would require an upwards threshold of 1.25.

protolambda:
Yeah the term could have been better. I care more about the hysteresis numbers than the names though, the leap needed to get back to 32 is affected significantly.

djrtwo:
modded to be 0.25/1.25

Ready for review and merge @protolambda

protolambda:
With experimentation and review these numbers may change again, I am thinking we could just make it configurable, and avoid further spec changes.

djrtwo:
okay, now configurable @protolambda

configs/mainnet.yaml
```diff
@@ -21,6 +21,12 @@ SHUFFLE_ROUND_COUNT: 90
 MIN_GENESIS_ACTIVE_VALIDATOR_COUNT: 16384
 # Jan 3, 2020
 MIN_GENESIS_TIME: 1578009600
+# 4
+HYSTERESIS_QUOTIENT: 4
+# 1 (minus 0.25)
+HYSTERESIS_DOWNWARD_MULTIPLIER: 1
+# 5 (plus 1.25)
+HYSTERESIS_UPWARD_MULTIPLIER: 5
```
protolambda: So previous behavior is, for context:
```
HYSTERESIS_QUOTIENT: 2
HYSTERESIS_DOWNWARD_MULTIPLIER: 0
HYSTERESIS_UPWARD_MULTIPLIER: 3
```

Sintayew4:
#1627

PR #1638, opened 2020-03-05
add genesis_validators_root for domain/chain separation

djrtwo:
[based on PR #1614]

Add `genesis_validators_root` to `BeaconState` and utilize for better signature domain separation as well as ENR/chain separation. This strengthens chain separation such that one must only be concerned about isolating fork version from a different chain when a contentious fork. Networks such as testnets and completely different chains with different genesis conditions can reuse fork versioning with no worry

I currently favor this approach and suggest we adopt it for the coming `v0.11` release. The consensus level changes are very minimal for a solid gain in sigs and peer discovery

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -95,7 +95,7 @@ from dataclasses import (
 from eth2spec.utils.ssz.ssz_impl import hash_tree_root
 from eth2spec.utils.ssz.ssz_typing import (
     View, boolean, Container, List, Vector, uint64,
-    Bytes1, Bytes4, Bytes8, Bytes32, Bytes48, Bytes96, Bitlist, Bitvector,
+    Bytes1, Bytes4, Bytes32, Bytes48, Bytes96, Bitlist, Bitvector,
 )
 from eth2spec.utils import bls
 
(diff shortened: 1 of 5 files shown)


djrtwo:
cc: @arnetheduck @nisdas @AgeManning @protolambda

mkalinin:
> This strengthens chain separation such that one must only be concerned about isolating fork version from a different chain when a contentious fork.

One must be concerned about isolating _all_ subsequent fork versions after contentious fork. Do we want to take care about that?

An alternative solution could be as follows:
- set current fork version to `genesis_validator_root[:4]` in genesis (should be enough to distinguish various networks)
- mix a root of fork version tree into the domain, fork version tree leaf is tuple `(epoch, version)`

This solution is suitable if we want to take care of domain isolation after contentious fork from the very beginning. Otherwise, it merely makes things more complicated.

djrtwo:
Hm, I see isolation from a contentious fork as a much simpler problem than isolation from different chains in general. A contentious fork can just flip a high-order bit and increment versions as normal.

My gut is that your proposed solution might be overly complicated.
Are you suggesting maintaining all `(epoch, version)` as a list and using the root of the list for domain separation?

One of my goals here is to also minimize changes to the phase 0 consensus while getting relatively good properties

mkalinin:
> A contentious fork can just flip a high-order bit and increment versions as normal.

Truth. With this alternative we would not need to do an explicit separation. It seems to be the only gain that we get. In general contentious fork will be eager to isolate their fork versions.

> My gut is that your proposed solution might be overly complicated.
Are you suggesting maintaining all (epoch, version) as a list and using the root of the list for domain separation?

Agreed, let's keep it simple. Moreover, it's not that clear how to mix a list of fork versions that are likely placed in a config with genesis fork version from the state.

AgeManning:
This looks good to me!

PR #1647, opened 2020-03-09
bump version to v0.11.0 for coming release

djrtwo:


djrtwo:
So actually inside of master after the release is merged?

djrtwo:
then backport to some `dev` versioning?

protolambda:
I thought about it more, and I think that it doesn't actually matter much for `dev`, so updating ahead of release sounds ok. You would need to pull a specific branch anyway to get something else than the `master` version, so let's only use `dev` or other add-ons in the version if we make temporary builds for special cases.

Issue #1641, opened 2020-03-07
Storage of SignedBeaconBlock


sgryphon:
The P2P specification lists validations before forwarding signed_beacon_block, which implies you will forward it, but doesn't have a requirement to. 

Similarly, the request/response section implies nodes should record signed_beacon_block in order to respond with it, "if they have it". 

However, in phase 0 fork choice on_block only signed_block.message, the BeaconBlock is stored. 

Note: In the current dev branch, the definition of Store has been changed to store BeaconBlockHeader, instead of BeaconBlock, although the change has not yet been propagated to all functions.

The specification does talk about messages without parents or in the future being delayed, again implying storage. 

Should there be an explicit mention, even if just something like "clients SHOULD keep a record of messages received, including signature", if not something formal like Store?

**EDIT** for clarity.

The relevant requirements in the 0.10.1 spec are:

> Clients MUST support requesting blocks since the start of the weak subjectivity period and up to the given head_block_root.
> Clients MUST support head_block_root values since the latest finalized epoch.
> Clients MUST respond with at least one block, if they have it and it exists in the range. Clients MAY limit the number of blocks in the response.

The current dev spec removes the duplication:

> Clients MUST support requesting blocks since the start of the weak subjectivity period and up to the given head_block_root.
> Clients MUST respond with at least one block, if they have it and it exists in the range. Clients MAY limit the number of blocks in the response.

In spec 0.9.2 and earlier, the signature was in BeaconBlock, and fork choice Store kept a record of all BeaconBlocks. This means the "if they have it" criteria would apply for all blocks you have in Store.

However, from 0.9.3 onwards, SignedBeaconBlock was introduced, and not explicitly stored, so clients would support the request, but simply send an empty response (they have no signed blocks stored anywhere).

djrtwo:
The requirement of the fork choice is only validated the blocks/headers so that's why the naive spec stores them there, whereas the network spec needs signed blocks to be passed around for others to validate.

There are statements like
```
Clients MUST support requesting blocks since the latest finalized epoch.
```

I suppose we can change this to "signed blocks" to be more explicit

sgryphon:
That's probably fair enough; I was concentrating on the "if they have it and it exists in the range" exception, and interpreting the "MUST support requesting" as simply sure, I'll accept your request, but because I never stored any blocks I don't have them and so support your request by responding with nothing. (Before the signing root change, the Store contained full blocks, including signatures, so the clients would have the blocks.)

[...]

PR #1648, opened 2020-03-10
Put back in a requirement to store recent signed blocks 

sgryphon:
In spec 0.9.2 and earlier, the signature was in BeaconBlock, which was recorded in Store, so clients would generally have the blocks available to service requests (i.e. pass the "if they have it" criteria).

However since SignedBeaconBlock was introduced there has been no explicit storage of signed blocks received.

While the specification says "MUST support requesting" it is qualified by "respond .. if they have it".

This change removes any confusion by putting back in an explicit requirement to store blocks (although the mechanism is not specified, i.e. they are no longer necessarily in Store).

Further information is in issue #1641

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -548,7 +548,7 @@ The request MUST be encoded as an SSZ-container.
 
 The response MUST consist of zero or more `response_chunk`. Each _successful_ `response_chunk` MUST contain a single `SignedBeaconBlock` payload.
 
-Clients MUST keep a record of signed blocks seen since the since the start of the weak subjectivity period and MUST support requesting blocks up to the given `head_block_root`.
+Clients MUST keep a record of signed blocks seen since the since the start of the weak subjectivity period and MUST support serving requests of blocks up to their own `head_block_root`.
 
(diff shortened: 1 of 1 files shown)


specs/phase0/p2p-interface.md
```diff
@@ -547,7 +547,8 @@ Requests count beacon blocks from the peer starting from `start_slot`, leading u
 The request MUST be encoded as an SSZ-container.
 
 The response MUST consist of zero or more `response_chunk`. Each _successful_ `response_chunk` MUST contain a single `SignedBeaconBlock` payload.
-Clients MUST support requesting blocks since the start of the weak subjectivity period and up to the given `head_block_root`.
+
+Clients MUST keep a record of signed blocks seen since the since the start of the weak subjectivity period and MUST support requesting blocks up to the given `head_block_root`.
```
```suggestion
Clients MUST keep a record of signed blocks seen since the since the start of the weak subjectivity period and MUST support serving requests of blocks up to their own `head_block_root`.
```

Wording was strange here, and I think we should be defining what the responder must support? @djrtwo

djrtwo: looks good

PR #1650, opened 2020-03-10
Update latest messages according to attestations in blocks

cemozerr:


protolambda:
Fork-choice is based on *any* valid attestation, not just those in the blocks. However, you can definitely read attestations from blocks, just in case you haven't seen them in the gossip already.

The spec does this implicitly now, taken from the `on_attestation` comment:
```
Run ``on_attestation`` upon receiving a new ``attestation`` from either within a block or directly on the wire.
```

And I would keep it at as-is, since clients process/pool attestations in their own way.

djrtwo:
Agreed. Let's keep as is. The `on_attestation` routine is expected to be run anytime you receive new valid attestations from any source. If I FEDEX your node an attestation, you better include it.

cemozerr:
We're working on an optical character recognition software to read your FEDEX'ed attestations @djrtwo. Don't you worry

Issue #1475, opened 2019-11-10
Phase 0 to 1 upgradability: fork/version boundary deserialization issue


hwwhww:
Thanks to @dankrad for giving input of the proposed solution.

tldr: (i) The SSZ object versioning is based on protocol versions, and (ii) heterogeneous list/vector is *not* allowed in SSZ. So with the current phase 1 spec (https://github.com/ethereum/eth2.0-specs/pull/1427), when we cross the version boundary, we are not able to process the previous operations with the new scheme.

### Versioning in the networking layer
* For the individual operation gossip topics messages (e.g., `attestation`, `voluntary_exit`) are versioned by protocol IDs. So it's fine to receive messages of different versions at the single-operation level.
* The `Protocol ID` would map to the specific schema version of the SSZ object definition, so that we can deserialize it with the corresponding schema. So 👍

### Fork/version boundary operations
* However, when crossing the version boundary, the current `BeaconBlockBody.[operations]: List[Operation, N]` only hold for one version of the given `Operation`.
* The block proposer cannot include both the previous version and current version objects in the same SSZ List. For example:

#### Example 1: Attestation inclusion
* In the phase 1 protocol changes, we can forecast that at least `AttestationData` fields will be changed.
* For the attesters of the **last slot of the previous version**, if they create attestation with the *old* schema, their attestation will *not* be able to be included in the **block of the first slot of the current version**.
* If the attesters create attestation with the *new* schema, it's possible to be included in the next block. **But, should they?**
* Moreover, how about the late attestations?

####  Example 2: Slashing operations
* `AttesterSlashing` contains `attestation_1: IndexedAttestation` and `attestation_2: IndexedAttestation`, so similar to `Attestation`, the new version block can't include heterogeneous `AttesterSlashing` formats. So we can't slash the malicious attesters at the previous version epoch.
* `ProposerSlashing` contains `header_1: BeaconBlockHeader` and `header_2: BeaconBlockHeader`, which are the conflicting beacon block headers of the slot of the proposer to slash. *If* we change `BeaconBlockHeader` format in the future version, we can't slash the malicious proposer that double-proposed at the previous version epoch.

#### Example 3: `BeaconState.previous_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]`.
* At the first epoch of the new version, this field should be able to hold the previous version of `PendingAttestation`.

## Proposed solution
[...]

PR #1652, opened 2020-03-10
Gossip topics with fork digest

djrtwo:
Supersedes @arnetheduck's #1624 to instead use the `ForkDigest` (introduced in #1614)  for gossip topic versioning.

Note, a 4-byte digest is used for domain separation in p2p (ENR, status, gossip topics). This should be sufficient for practical separation on this layer and induces low overhead in the ENR and status messages. In the signature domain, we use 28-bytes of the digest (along with 4-byte domain-type) for a cryptographically secure domain separation of signatures.

Also moves some helpers into state transition spec to increase code use between p2p and state transition spec

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -822,7 +822,7 @@ def compute_fork_data_root(current_version: Version, genesis_validators_root: Ro
 #### `compute_fork_digest`
 
 ```python
-def compute_fork_digest(current_version: Version, genesis_validators_root: Root) -> Root:
+def compute_fork_digest(current_version: Version, genesis_validators_root: Root) -> ForkDigest:
     """
     Return the 4-byte fork digest for the ``current_version`` and ``genesis_validators_root``.
     This is a digest primarily used for domain separation on the p2p layer.
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -77,6 +77,7 @@
     - [`compute_start_slot_at_epoch`](#compute_start_slot_at_epoch)
     - [`compute_activation_exit_epoch`](#compute_activation_exit_epoch)
     - [`compute_fork_data_root`](#compute_fork_data_root)
+    - [`compute_fork_digest`](#compute_fork_digest)
     - [`compute_domain`](#compute_domain)
     - [`compute_signing_root`](#compute_signing_root)
   - [Beacon state accessors](#beacon-state-accessors)
@@ -293,7 +294,7 @@ class Fork(Container):
 ```python
 class ForkData(Container):
     current_version: Version
-    genesis_root: Root
+    genesis_validators_root: Root
 ```
 
 #### `Checkpoint`
@@ -809,7 +810,8 @@ def compute_activation_exit_epoch(epoch: Epoch) -> Epoch:
 ```python
 def compute_fork_data_root(current_version: Version, genesis_validators_root: Root) -> Root:
     """
-    Return the fork digest for the ``current_fork_version`` and ``genesis_validators_root``
+    Return the 32-byte fork data root for the ``current_version`` and ``genesis_validators_root``.
+    This is used primarily in signature domains to avoid collisions across forks/chains.
     """
     return hash_tree_root(ForkData(
         current_version=current_version,
@@ -817,18 +819,30 @@ def compute_fork_data_root(current_version: Version, genesis_validators_root: Ro
     ))
 ```
 
+#### `compute_fork_digest`
+
+```python
+def compute_fork_digest(current_version: Version, genesis_validators_root: Root) -> Root:
+    """
+    Return the 4-byte fork digest for the ``current_version`` and ``genesis_validators_root``.
+    This is a digest primarily used for domain separation on the p2p layer.
+    4-bytes suffices for practical separation of forks/chains.
+    """
+    return ForkDigest(compute_fork_data_root(current_version, genesis_validators_root)[:4])
+```
+
 #### `compute_domain`
 
 ```python
-def compute_domain(domain_type: DomainType, fork_version: Optional[Version]=None, genesis_root: Root=None) -> Domain:
+def compute_domain(domain_type: DomainType, fork_version: Version=None, genesis_validators_root: Root=None) -> Domain:
     """
     Return the domain for the ``domain_type`` and ``fork_version``.
     """
     if fork_version is None:
         fork_version = GENESIS_FORK_VERSION
-    if genesis_root is None:
-        genesis_root = Root()  # all bytes zero by default
-    fork_data_root = compute_fork_data_root(fork_version, genesis_root)
+    if genesis_validators_root is None:
+        genesis_validators_root = Root()  # all bytes zero by default
+    fork_data_root = compute_fork_data_root(fork_version, genesis_validators_root)
     return Domain(domain_type + fork_data_root[:28])
 ```
 
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -24,6 +24,7 @@
 - [Containers](#containers)
   - [Misc dependencies](#misc-dependencies)
     - [`Fork`](#fork)
+    - [`ForkData`](#forkdata)
     - [`Checkpoint`](#checkpoint)
     - [`Validator`](#validator)
     - [`AttestationData`](#attestationdata)
@@ -75,6 +76,7 @@
     - [`compute_epoch_at_slot`](#compute_epoch_at_slot)
     - [`compute_start_slot_at_epoch`](#compute_start_slot_at_epoch)
     - [`compute_activation_exit_epoch`](#compute_activation_exit_epoch)
+    - [`compute_fork_data_root`](#compute_fork_data_root)
     - [`compute_domain`](#compute_domain)
     - [`compute_signing_root`](#compute_signing_root)
   - [Beacon state accessors](#beacon-state-accessors)
@@ -284,6 +286,14 @@ class Fork(Container):
     previous_version: Version
     current_version: Version
     epoch: Epoch  # Epoch of latest fork
+```
+
+#### `ForkData`
+
+```python
+class ForkData(Container):
+    current_version: Version
+    genesis_root: Root
 ```
 
 #### `Checkpoint`
@@ -794,6 +804,19 @@ def compute_activation_exit_epoch(epoch: Epoch) -> Epoch:
     return Epoch(epoch + 1 + MAX_SEED_LOOKAHEAD)
 ```
 
+#### `compute_fork_data_root`
+
+```python
+def compute_fork_data_root(current_version: Version, genesis_validators_root: Root) -> Root:
+    """
+    Return the fork digest for the ``current_fork_version`` and ``genesis_validators_root``
+    """
+    return hash_tree_root(ForkData(
+        current_version=current_version,
+        genesis_validators_root=genesis_validators_root,
+    ))
+```
+
 #### `compute_domain`
 
 ```python
@@ -805,7 +828,8 @@ def compute_domain(domain_type: DomainType, fork_version: Optional[Version]=None
         fork_version = GENESIS_FORK_VERSION
     if genesis_root is None:
         genesis_root = Root()  # all bytes zero by default
-    return Domain(domain_type + fork_version + genesis_root[:24])
+    fork_data_root = compute_fork_data_root(fork_version, genesis_root)
+    return Domain(domain_type + fork_data_root[:28])
 ```
 
 #### `compute_signing_root`
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -219,17 +219,13 @@ The following gossipsub [parameters](https://github.com/libp2p/specs/tree/master
 
 ### Topics and messages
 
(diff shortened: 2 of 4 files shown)


specs/phase0/p2p-interface.md
```diff
@@ -219,7 +219,13 @@ The following gossipsub [parameters](https://github.com/libp2p/specs/tree/master
 
 ### Topics and messages
 
-Topics are plain UTF-8 strings and are encoded on the wire as determined by protobuf (gossipsub messages are enveloped in protobuf messages). Topic strings have form: `/eth2/TopicName/TopicEncoding`. This defines both the type of data being sent on the topic and how the data field of the message is encoded.
+Topics are plain UTF-8 strings and are encoded on the wire as determined by protobuf (gossipsub messages are enveloped in protobuf messages). Topic strings have form: `/eth2/ForkDigest/Name/Encoding`. This defines both the type of data being sent on the topic and how the data field of the message is encoded.
+
+- `ForkDigest` - the hex-encoded bytes of `ForkDigest(compute_fork_data_root(current_fork_version, genesis_validators_root)[:4])` where
```
```suggestion
- `ForkDigest` - the lowercase hex-encoded (no "0x" prefix) bytes of `ForkDigest(compute_fork_data_root(current_fork_version, genesis_validators_root)[:4])` where
```

djrtwo: thanks! also added a note in v-guide about preparing for hard fork

specs/phase0/p2p-interface.md
```diff
@@ -219,7 +219,13 @@ The following gossipsub [parameters](https://github.com/libp2p/specs/tree/master
 
 ### Topics and messages
 
-Topics are plain UTF-8 strings and are encoded on the wire as determined by protobuf (gossipsub messages are enveloped in protobuf messages). Topic strings have form: `/eth2/TopicName/TopicEncoding`. This defines both the type of data being sent on the topic and how the data field of the message is encoded.
+Topics are plain UTF-8 strings and are encoded on the wire as determined by protobuf (gossipsub messages are enveloped in protobuf messages). Topic strings have form: `/eth2/ForkDigest/Name/Encoding`. This defines both the type of data being sent on the topic and how the data field of the message is encoded.
+
+- `ForkDigest` - the lowercase hex-encoded (no "0x" prefix) bytes of `ForkDigest(compute_fork_data_root(current_fork_version, genesis_validators_root)[:4])` where
+    - `current_fork_version` is the fork version of the epoch of the message to be sent on the topic
```
arnetheduck: What about slashings though? they span 2 epochs, potentially. We need to send to both, or have everyone stay on multiple slashing topics for as long as slashing is possible.

We can also leave it unspecified, noting that only one fork version exists for now, and that the hard fork will decide the upgrade procedure.

specs/phase0/beacon-chain.md
```diff
@@ -794,6 +804,19 @@ def compute_activation_exit_epoch(epoch: Epoch) -> Epoch:
     return Epoch(epoch + 1 + MAX_SEED_LOOKAHEAD)
 ```
 
+#### `compute_fork_data_root`
+
+```python
+def compute_fork_data_root(current_version: Version, genesis_validators_root: Root) -> Root:
+    """
+    Return the fork digest for the ``current_fork_version`` and ``genesis_validators_root``
```
arnetheduck: `current_version`?

specs/phase0/p2p-interface.md
```diff
@@ -644,39 +650,27 @@ ENRs MUST carry a generic `eth2` key with an 16-byte value of the node's current
...
+    fork_digest: ForkDigest
     next_fork_version: Version
     next_fork_epoch: Epoch
 )
 ```
 
 where the fields of `ENRForkID` are defined as
 
-* `current_fork_digest` is `ForkDigest(hash_tree_root(current_fork_data)[:4])`
+* `fork_digest` is `ForkDigest(compute_fork_data_root(current_fork_version, genesis_validators_root)[:4])` where
```
arnetheduck: `Status`  message needs updating as well, to use `fork_digest`

specs/phase0/p2p-interface.md
```diff
@@ -829,9 +823,9 @@ For future extensibility with almost zero overhead now (besides the extra bytes
 
 ### How do we upgrade gossip channels (e.g. changes in encoding, compression)?
 
-Changing gossipsub/broadcasts requires a coordinated upgrade where all clients start publishing to the new topic together, for example during a hard fork.
+Changing gossipsub/broadcasts requires a coordinated upgrade where all clients start publishing to the new topic together, during a hard fork.
 
-One can envision a two-phase deployment as well where clients start listening to the new topic in the first phase then start publishing some time later, letting the traffic naturally move over to the new topic.
+When a node is preparing for upcoming tasks (e.g. validator duty lookahead) on a gossipsub topic, the node should join the topic of the future epoch in which the task is to occur in addition to listening to the topics for the current epoch.
```
arnetheduck: So here we could note that for slashings, clients need to listen to historical topics as well.

specs/phase0/p2p-interface.md
```diff
@@ -644,39 +650,27 @@ ENRs MUST carry a generic `eth2` key with an 16-byte value of the node's current
...
 * `next_fork_epoch` is the epoch at which the next fork is planned and the `current_fork_version` will be updated. If no future fork is planned, set `next_fork_epoch = FAR_FUTURE_EPOCH` to signal this fact
 
-Clients SHOULD connect to peers with `current_fork_digest`, `next_fork_version`, and `next_fork_epoch` that match local values.
+Clients SHOULD connect to peers with `fork_digest`, `next_fork_version`, and `next_fork_epoch` that match local values.
 
-Clients MAY connect to peers with the same `current_fork_version` but a different `next_fork_version`/`next_fork_epoch`. Unless `ENRForkID` is manually updated to matching prior to the earlier `next_fork_epoch` of the two clients, these type of connecting clients will be unable to successfully interact starting at the earlier `next_fork_epoch`.
+Clients MAY connect to peers with the same `fork_digest` but a different `next_fork_version`/`next_fork_epoch`. Unless `ENRForkID` is manually updated to matching prior to the earlier `next_fork_epoch` of the two clients, these type of connecting clients will be unable to successfully interact starting at the earlier `next_fork_epoch`.
```
hwwhww: I'm not sure what `type of` means here. Should it be removed?

```suggestion
Clients MAY connect to peers with the same `fork_digest` but a different `next_fork_version`/`next_fork_epoch`. Unless `ENRForkID` is manually updated to matching prior to the earlier `next_fork_epoch` of the two clients, these connecting clients will be unable to successfully interact starting at the earlier `next_fork_epoch`.
```

specs/phase0/p2p-interface.md
hwwhww: I asked about what will happen if slashing messages formats are different cross forks: https://github.com/ethereum/eth2.0-specs/issues/1475
It seems the status-quo is to believe in the honest majority.

djrtwo: We can make a table of messages and related field that dictates the epoch/gossipchannel.

As for cross-fork slashings, it is a more fundamental problem than just which channel to send it on. To catch slashings across a fork boundary in which the data-type changed would require specific data structures to handle mismatched types and separate functions for verifying validity of each. As discussed in #1475, I'm happy to sweep that under the rug in early phases and address it if/when we need to.

When we upgrade to Phase 1, we'll make it clear that fraud proofs across the boundary are not accepted and that this channel upgrade path only supports message types from the respective fork

specs/phase0/p2p-interface.md
djrtwo: Does it?

I would tend to agree, but it does change fundamental functionality here. `fork_digest` is in relation to wall-clock time whereas `Status.head_fork_version` was in relation to the head of your current chain (so might be behind while you're syncing).

The granularity of having Status version wrt your current chain might allow for marginally more optionality in syncing sections of a chain with a peer with mismatched forks in the future. Whereas straight `fork_digest` wrt wall-clock time will be much cleaner to handle.

I lean toward the clean `fork_digest` for cleanliness and not trying to optimize for a fringe case

specs/phase0/p2p-interface.md
djrtwo: As discussed above, we might rely more heavily on honestly assumption at initial phase forks to avoid consensus complexity in handling message types across forks. 

If at a fork, we decide to add logic to handle cross-fork message types, we'll address this more explicitly

specs/phase0/beacon-chain.md
```diff
@@ -794,6 +804,19 @@ def compute_activation_exit_epoch(epoch: Epoch) -> Epoch:
     return Epoch(epoch + 1 + MAX_SEED_LOOKAHEAD)
 ```
 
+#### `compute_fork_data_root`
+
+```python
+def compute_fork_data_root(current_version: Version, genesis_validators_root: Root) -> Root:
```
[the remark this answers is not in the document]
djrtwo: done

specs/phase0/p2p-interface.md
```diff
@@ -219,7 +219,13 @@ The following gossipsub [parameters](https://github.com/libp2p/specs/tree/master
 
 ### Topics and messages
 
-Topics are plain UTF-8 strings and are encoded on the wire as determined by protobuf (gossipsub messages are enveloped in protobuf messages). Topic strings have form: `/eth2/TopicName/TopicEncoding`. This defines both the type of data being sent on the topic and how the data field of the message is encoded.
+Topics are plain UTF-8 strings and are encoded on the wire as determined by protobuf (gossipsub messages are enveloped in protobuf messages). Topic strings have form: `/eth2/ForkDigest/Name/Encoding`. This defines both the type of data being sent on the topic and how the data field of the message is encoded.
+
+- `ForkDigest` - the lowercase hex-encoded (no "0x" prefix) bytes of `ForkDigest(compute_fork_data_root(current_fork_version, genesis_validators_root)[:4])` where
```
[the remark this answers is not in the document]
djrtwo: Done! and added a helper so that we don't need to always slice off the correct bytes everywhere we want the fork digest

specs/phase0/p2p-interface.md
djrtwo: added in fork_digest

djrtwo:
addressed comments @hwwhww and @arnetheduck

specs/phase0/beacon-chain.md
```diff
@@ -794,18 +805,45 @@ def compute_activation_exit_epoch(epoch: Epoch) -> Epoch:
...
+    return hash_tree_root(ForkData(
+        current_version=current_version,
+        genesis_validators_root=genesis_validators_root,
+    ))
+```
+
+#### `compute_fork_digest`
+
+```python
+def compute_fork_digest(current_version: Version, genesis_validators_root: Root) -> Root:
```
hwwhww:
```suggestion
def compute_fork_digest(current_version: Version, genesis_validators_root: Root) -> ForkDigest:
```

specs/phase0/p2p-interface.md
```diff
@@ -469,16 +475,18 @@ constituents individually as `response_chunk`s. For example, the
 Request, Response Content:
 ```
 (
-  head_fork_version: Bytes4
-  finalized_root: Bytes32
-  finalized_epoch: uint64
-  head_root: Bytes32
-  head_slot: uint64
```
hwwhww: I feel @arnetheduck wanted to use simpler types instead of custom types in the first networking spec draft, not sure about what do you think about it now?

IMO it's okay to use custom types for presentation here.

specs/phase0/validator.md
```diff
@@ -519,6 +519,8 @@ Because Phase 0 does not have shards and thus does not have Shard Committees, th
 * Maintain advertisement of the randomly selected subnets in their node's ENR `attnets` entry by setting the randomly selected `subnet_id` bits to `True` (e.g. `ENR["attnets"][subnet_id] = True`) for all persistent attestation subnets
 * Set the lifetime of each random subscription to a random number of epochs between `EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION` and `2 * EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION]`. At the end of life for a subscription, select a new random subnet, update subnet subscriptions, and publish an updated ENR
 
+*Note*: When preparing for a hard fork, a validator must select and subscribe to random subnets of the future fork versioning at least `EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION` epochs in advance of the fork. These new subnets for the fork are maintained in addition to those for the current fork until the fork occurs. After the fork occurs, let the subnets from the previous fork reach end of life with no replacements.
```
hwwhww: nitpick

```suggestion
*Note*: When preparing for a hard fork, a validator must select and subscribe to random subnets of the future fork versioning at least `EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION` epochs in advance of the fork. These new subnets for the fork are maintained in addition to those for the current fork until the fork occurs. After the fork occurs, let the subnets from the previous fork reach the end of life with no replacements.
```

specs/phase0/p2p-interface.md
djrtwo: I think consistency is better, but don't feel super strongly about it

arnetheduck: fine with consistency across specs as argument - it wasn't a strong opinion - it also helps when grep:ing to have similar names and types across spec. :+1:

PR #1653, opened 2020-03-11
Must match target to match head

djrtwo:
Remove a potentially degenerate path in the rewards by requiring `get_matching_head_attestations` to match target (instead of just source). This prevents a path in which a validator could still be profitable by getting source and head correct, but voting an incorrect target (and thus avoiding finality).

Add test that has full attestations but messes up random fields to ensure we get better reward coverage

[also add cache for some of these attestation helpers. Seems like it shaved a few seconds off of local test times, but nothing crazy]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -182,7 +182,17 @@ get_active_validator_indices = cache_this(
 _get_beacon_committee = get_beacon_committee
 get_beacon_committee = cache_this(
     lambda state, slot, index: (state.validators.hash_tree_root(), state.randao_mixes.hash_tree_root(), slot, index),
-    _get_beacon_committee)'''
+    _get_beacon_committee)
+
+_get_matching_target_attestations = get_matching_target_attestations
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -182,7 +182,17 @@ def wrapper(*args, **kw):  # type: ignore
 _get_beacon_committee = get_beacon_committee
 get_beacon_committee = cache_this(
     lambda state, slot, index: (state.validators.hash_tree_root(), state.randao_mixes.hash_tree_root(), slot, index),
-    _get_beacon_committee)'''
+    _get_beacon_committee)
+
+_get_matching_target_attestations = get_matching_target_attestations
+get_matching_target_attestations = cache_this(
```
protolambda: As we start to memoize more functions we should update the `cache_this` helper to cache up to some maximum N items. Maybe with an LRU cache. Something for after v0.11.0 I think, this works well enough for now.

PR #1656, opened 2020-03-11
minor typos and clarifications in p2p spec

djrtwo:
* fix minor typo (version -> digest)
* use the term "branch" in some of the FAQs to generally reserve "chain" for different isolated networks


djrtwo:
done

Issue #1663, opened 2020-03-16
Division by zero in `get_attestation_deltas` or wrong min value of `get_total_balance`


MrChico:
A comment in `get_total_balance` reads that the function returns a minimum of 1 Gwei to avoid divisions by zero.

But since https://github.com/ethereum/eth2.0-specs/pull/1635, `total_balance` must be at least `EFFECTIVE_BALANCE_INCREMENT` in order to avoid a division by zero  in [this line](https://github.com/ethereum/eth2.0-specs/blame/dev/specs/phase0/beacon-chain.md#L1369) of `get_attestation_deltas`:
```
rewards[index] = reward_numerator // (total_balance // increment)
```

It seems that either the minimum of `total_balance` should be raised, or the reward calculation be adjusted to avoid this edge case.

MrChico:
In a similar vein, if a slashed validators effective balance drops under EFFECTIVE_BALANCE_INCREMENT, it seems they will not suffer the anti-correlation penalty applied in `process_slashings`, as [the expression](https://github.com/ethereum/eth2.0-specs/blame/dev/specs/phase0/beacon-chain.md#L1441):
```
            penalty_numerator = validator.effective_balance // increment * min(sum(state.slashings) * 3, total_balance)
```
will evaluate to 0.

djrtwo:
Right. The only way their effective balance could drop below `EFFECTIVE_BALANCE_INCREMENT` is if it went to 0 (because effective balance is only changed on that increment) which would give them zero penalty regardless.

That said, maybe `penalty_numerator` here should be in terms of `balance` rather than `effective_balance`

djrtwo:
@MrChico I addressed the first issue here #1664 

Eating lunch, thinking about the second

MrChico:
Ah, yes, I see now that the second concern is not an issue.
It seems that all other reward calculations are being weighted by total_balances and the validators effective balance, so I don't see why this one should be different?

djrtwo:
Yeah, no need for this to be different.

Just an orthogonal thought, and one that I don't think I agree with.

PR #1664, opened 2020-03-16
avoid div by zero in extreme balance case

djrtwo:
Address #1663

* `get_total_balance` returns a minimum of `EFFECTIVE_BALANCE_INCREMENT` to avoid div by zero in extreme cases
* add rewards test vector to test for this extreme case (it did in fact div by 0 before the fix was in place)
* Add another rewards test for empty attestations (because I noticed it was needed... )

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -996,10 +996,11 @@ def get_beacon_proposer_index(state: BeaconState) -> ValidatorIndex:
 ```python
 def get_total_balance(state: BeaconState, indices: Set[ValidatorIndex]) -> Gwei:
     """
-    Return the combined effective balance of the ``indices``. (1 Gwei minimum to avoid divisions by zero.)
+    Return the combined effective balance of the ``indices``.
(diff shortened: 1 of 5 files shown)


specs/phase0/beacon-chain.md
```diff
@@ -996,10 +996,11 @@ def get_beacon_proposer_index(state: BeaconState) -> ValidatorIndex:
 ```python
 def get_total_balance(state: BeaconState, indices: Set[ValidatorIndex]) -> Gwei:
     """
-    Return the combined effective balance of the ``indices``. (1 Gwei minimum to avoid divisions by zero.)
+    Return the combined effective balance of the ``indices``.
+    ``EFFECTIVE_BALANCE_INCREMENT`` Gwei minimum to avoid divisions by zero.
```
hwwhww: Consider also noting in `get_total_active_balance`,`get_attesting_balance` comments?

IMHO the change adds some code debts, but still tolerable for the edge case. :)

djrtwo:
Thanks @hwwhww!
Made the suggested changes. Merging after tests run

PR #1665, opened 2020-03-17
remove IK handshake from p2p spec

djrtwo:
IK handshake is being removed from the libp2p noise specs for now (see here https://github.com/libp2p/specs/issues/249), so removing it as optional in our specs.

I believe the XXFallback is to be removed as well. Waiting for conf here https://github.com/libp2p/specs/issues/246#issuecomment-600181558


djrtwo:
cc @AgeManning @raulk

AgeManning:
Yep.

The `rust-libp2p` noise implementation has the XX and IK handshakes. I believe the IK handshake is planned on being dropped. 

So we currently just support XX, so fine with us to remove the others.

PR #1666, opened 2020-03-18
Remove unnecessary clock disparity check (nitpick)

paulhauner:
There is no need to mention clock disparity when comparing two static slot values (assuming the clock disparity is less than a slot, even then I don't think that's the intention).


djrtwo:
Yup, you're right. I'm going to make a new PR pointing this to `v011x` for immediate release

djrtwo:
Closing in favor of #1683 for immediate release

Issue #1637, opened 2020-03-05
Obtain ENR subnet bitfield or ENR from connected peers


AgeManning:
## Overview

Peers discover and connect to each other via ENRs (https://eips.ethereum.org/EIPS/eip-778). Discv5 is used to find new peers and their associated ENRs. Peers should store a [subnet-bitfield](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/p2p-interface.md#attestation-subnet-bitfield) representing the long-lived random subnets they are subscribed to. This allows one to search for and filter peers based on which long-lived subnets they are subscribed to. 

## Issue

Once a peer has found an ENR, they connect to other peers via tcp and negotiate libp2p protocols. A listening peer receiving the incoming connection has no mechanism to recover the ENR of the connecting peer. Therefore a client may have a number of connected peers, with no knowledge of their long-lived subnet subscriptions (which is useful information in deciding whether to search for more peers when a node needs to subscribe to a specific subnet). 

Furthermore, a client listening and receiving a connection from a peer, cannot add that peer into its routing table and therefore doesn't add this peer to the DHT as it does not know the connecting peer's ENR.

### 1. Add the subnet bitfield to the STATUS message

In this solution, we simply add the bitfield to the RPC STATUS message. This solves the immediate problem of finding peer's long-lived subscribed subnets however only updates when either client initiates a STATUS. A peer could update their subnet bitfield and then may trigger a STATUS to all peers to inform them of the update. 

This solution doesn't allow peers to add connecting peers to their local node table which is a feature we may want to improve the overall DHT and local routing table (especially for bootnodes).

### 2. Add a seperate RPC method for retrieving an ENR

In this solution we add a dedicated RPC method used to request/inform peers of an ENR/ENR-update. This method could be called by a node on connection to inform the listening peer of it's ENR. This method could also be used to update connected nodes when our local ENR is modified (including the subnet bitfield). 

ENR's can be at most 300 bytes long and we may want to include additional information. Therefore the payload of this RPC method could include other information along with the ENR.

AgeManning:
If we go with option 2, it may be coupled with #1531

arnetheduck:
Option 2 and a separate request for ENR specifically seems reasonable - would have a better upgrade path when something else comes along, as things tend to do - it also fits ENR and discv5 being completely standalone from eth2 and libp2p in general. 

[...]

PR #1668, opened 2020-03-18
Shift IP-related ENR fields as optional

AgeManning:
# Description

In Lighthouse we had these fields always defined as specified. However in a recent update to our discv5 which removes NAT'd peers from the DHT we realised it makes more sense to have nodes that have not sorted out their NAT traversal (i.e do not know or do not have external IP/PORTS that other peers can connect on) that their ENR's do not contain any IP/PORTs. 

Nodes without IP/PORTs should still be able to connect and discover other peers, but it will be less likely (depending on implementations) that these peers would be added to a local node routing table. 

It makes it clearer that nodes without an IP/PORT do not have a contactable address and will save us time trying to connect to a local IP/PORT which is not globally contactable.

Lighthouse by default, will have ENRs without any IP/PORT field (unless loaded from a previous state), although we set a TCP port (as this is not update-able from discovery). As they discover peers and their peers inform the node that they are contactable on a specific IP/PORT, the node will update it's ENR with IP/PORT field and get added to the DHT. This also ensures that the ENR doesn't get updated unless it is contactable from a wider audience.

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -622,6 +622,9 @@ This integration enables the libp2p stack to subsequently form connections and s
 The Ethereum Node Record (ENR) for an Ethereum 2.0 client MUST contain the following entries (exclusive of the sequence number and signature, which MUST be present in an ENR):
 
 -  The compressed secp256k1 publickey, 33 bytes (`secp256k1` field).
+
+The ENR MAY contain the following entries:
+
 -  An IPv4 address (`ip` field) and/or IPv6 address (`ip6` field).
 -  A TCP port (`tcp` field) representing the local libp2p listening port.
 -  A UDP port (`udp` field) representing the local discv5 listening port.


djrtwo:
So the point here is that some peers will be behind a NAT and thus their ENR's won't have some of these fields.

This isn't necessarily a problem when it comes to using ENRs for DHT discovery because you couldn't discover such peers anyway. But it is a problem if we use ENRs for other usecases such as option 2 in #1637

^ that all check out?

AgeManning:
Yeah, I guess this is more of a discv5 implementation issue for whether it effects DHT discovery. A chat with felix prompted me that ENR's could not specify IPs. 

If a node connects to you and asks you for peers and you don't know this node. You get it's ENR. Depending on the implementation, you need to know whether to add this new node to your routing table or not. If the connecting node doesn't have an IP field you just ignore it. 

I think it helps in general if node's haven't specified a specific IP, we can assume they are non-contactable, rather than attempting to connect and finding out after a timeout. For example, you start a new node without specifying a contactable IP, and try and connect to it's ENR. We can now have specific errors such as: "this node has no contactable address" rather than "could not connect to node for some reason". It gives us an extra state to work with.

djrtwo:
I see, thanks for the clarification.
This makes sense to me.

Any other comments? @arnetheduck @mkalinin @nisdas @fjl

arnetheduck:
+1 from us (cc @kdeme)

fjl:
I think this change is good.

djrtwo:
@AgeManning Can you rebase to `v011x`? I'll release this asap

nisdas:
this looks good to me :+1:

PR #1670, opened 2020-03-19
Normalize attestation signature getter name

arnetheduck:

PR #1671, opened 2020-03-19
Normalize attestation signature getter name

djrtwo:
Supersedes #1670 

* based on `v011x` branch for inclusion in coming release
* cherry-picked @arnetheduck's commit from #1670 
* minor mod to naming for clarity

### DIFF
--- a/specs/phase0/validator.md
+++ b/specs/phase0/validator.md
@@ -404,12 +404,12 @@ Set `attestation.data = attestation_data` where `attestation_data` is the `Attes
 
 ##### Aggregate signature
 
-Set `attestation.signature = signed_attestation_data` where `signed_attestation_data` is obtained from:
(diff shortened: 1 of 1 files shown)


arnetheduck:
fwiw, in the code we ended up with a bunch of these `get_*_signature` - they work with minimal information though and not the full state - effectively, we don't want to send the whole state across to the validator, so we pick the specific data: 

https://github.com/status-im/nim-beacon-chain/blob/a7963b0b8c9113466dc4dabce62a193b00d228f0/beacon_chain/spec/state_transition_block.nim#L485

djrtwo:
The decision to pass the state around around most times was to reduce the size of crazy function signatures (some that would have 8+ args). That decision rippled through the entire codebase for better or worse.

Much respect for your functions

PR #1672, opened 2020-03-20
Fix memoization (base reward cache bug + add LRU)

protolambda:
So while syncing the first 10,000 blocks of the lighthouse testnet, we hit a problem on the first rewards epoch transition. Not in the spec, but in the recently introduced caching of functions. After finding this, and fixing it, it happily processed the 10K blocks however.

Changes:
- Fix the base-reward caching
- Add small LRU, the cache was ever growing bigger previously
- Add more caching, it helps sync faster, and ability to process more faster helps us find bugs faster.
- Add make command to run a specific test. Setting up the pyspec context once and then debugging with IDE is nice, but a quick shorthand for checks is handy.
- Update rewards test, and corresponding context set-up, to use different balances better, with activation threshold adjusted down to actually enable them.

To reproduce:
- Edit to have no `index` in the cache key.
- Run: `make find_test K=test_full_attestations_misc_balances`

To test rewards better, splitting up the rewards function in later release would be good, as it isolates these problems better. It was considered earlier, but didn't make it through because of minimalism. I do think the "all rewards in this function here" approach can hurt though.
That said, this was only introduced in the latest release, and I am pretty sure the diff fuzzer would have hit it pretty easily. So we can focus on performance so the spec stays part of the fuzzer, and/or isolate things better.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -92,6 +92,8 @@ from dataclasses import (
     field,
 )
 
+from lru import LRU
+
 from eth2spec.utils.ssz.ssz_impl import hash_tree_root
 from eth2spec.utils.ssz.ssz_typing import (
     View, boolean, Container, List, Vector, uint64,
@@ -113,6 +115,8 @@ from dataclasses import (
     dataclass,
     field,
 )
+
+from lru import LRU
 
 from eth2spec.utils.ssz.ssz_impl import hash_tree_root
 from eth2spec.utils.ssz.ssz_typing import (
@@ -152,8 +156,8 @@ def hash(x: bytes) -> Bytes32:  # type: ignore
     return hash_cache[x]
 
 
-def cache_this(key_fn, value_fn):  # type: ignore
-    cache_dict = {}  # type: ignore
+def cache_this(key_fn, value_fn, lru_size):  # type: ignore
+    cache_dict = LRU(size=lru_size)
 
     def wrapper(*args, **kw):  # type: ignore
         key = key_fn(*args, **kw)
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -164,35 +168,50 @@ def wrapper(*args, **kw):  # type: ignore
...
 
 
+_compute_shuffled_index = compute_shuffled_index
+compute_shuffled_index = cache_this(
+    lambda index, index_count, seed: (index, index_count, seed),
+    _compute_shuffled_index, lru_size=SLOTS_PER_EPOCH * 3)
+
+_get_total_active_balance = get_total_active_balance
+get_total_active_balance = cache_this(
+    lambda state: (state.validators.hash_tree_root(), state.slot),
```
djrtwo: This can cache based on current epoch instead of current slot.

Also this cache is only safe as long as lookups aren't for more than `MAX_SEED_LOOKAHEAD` epochs in the future

setup.py
```diff
@@ -164,35 +168,50 @@ def wrapper(*args, **kw):  # type: ignore
...
+    _get_total_active_balance, lru_size=10)
+
 _get_base_reward = get_base_reward
 get_base_reward = cache_this(
-    lambda state, index: (state.validators.hash_tree_root(), state.slot),
-    _get_base_reward)
+    lambda state, index: (state.validators.hash_tree_root(), state.slot, index),
+    _get_base_reward, lru_size=10)
```
djrtwo: Why size 10? Seems like it's valuable to have it be of at least the size of the validator set, right?

setup.py
```diff
@@ -164,35 +168,50 @@ def wrapper(*args, **kw):  # type: ignore
...
 _get_matching_head_attestations = get_matching_head_attestations
 get_matching_head_attestations = cache_this(
     lambda state, epoch: (state.hash_tree_root(), epoch),
-    _get_matching_head_attestations)'''
+    _get_matching_head_attestations, lru_size=10)
+
+_get_attesting_indices = get_attesting_indices
+get_attesting_indices = cache_this(lambda state, data, bits:
+                                   (state.validators.hash_tree_root(), data.hash_tree_root(), bits.hash_tree_root()),
+                                   _get_attesting_indices, lru_size=SLOTS_PER_EPOCH * MAX_COMMITTEES_PER_SLOT * 3)'''
```
djrtwo:
```suggestion
_get_attesting_indices = get_attesting_indices
get_attesting_indices = cache_this(
    lambda state, data, bits: (state.validators.hash_tree_root(), data.hash_tree_root(), bits.hash_tree_root()),
    _get_attesting_indices, lru_size=SLOTS_PER_EPOCH * MAX_COMMITTEES_PER_SLOT * 3)'''
```

tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -100,8 +101,10 @@ def misc_balances(spec):
     Usage: `@with_custom_state(balances_fn=misc_balances, ...)`
     """
     num_validators = spec.SLOTS_PER_EPOCH * 8
-    num_misc_validators = spec.SLOTS_PER_EPOCH
-    return [spec.MAX_EFFECTIVE_BALANCE] * num_validators + [spec.MIN_DEPOSIT_AMOUNT] * num_misc_validators
+    balances = [spec.MAX_EFFECTIVE_BALANCE * 2 * i // num_validators for i in range(num_validators)]
```
djrtwo: 2x is to have a substantial number above the MAX_EFFECTIVE?

tests/core/pyspec/eth2spec/test/phase_0/epoch_processing/test_process_rewards_and_penalties.py
```diff
@@ -122,32 +120,41 @@ def test_full_attestations_random_incorrect_fields(spec, state):
...
+    assert any(v.effective_balance != spec.MAX_EFFECTIVE_BALANCE for v in state.validators)
     for index in range(len(pre_state.validators)):
         if index in attesting_indices:
             assert state.balances[index] > pre_state.balances[index]
         elif spec.is_active_validator(pre_state.validators[index], spec.compute_epoch_at_slot(state.slot)):
             assert state.balances[index] < pre_state.balances[index]
         else:
             assert state.balances[index] == pre_state.balances[index]
+    # Check if base rewards are consistent with effective balance.
+    brs = {}
```
djrtwo: This is just checking that the cache is working properly, correct?

setup.py
protolambda: > This can cache based on current epoch instead of current slot.

It's only called during epoch transition, so effectively the same. I will change it to epoch key though.

> Also this cache is only safe as long as lookups aren't for more than MAX_SEED_LOOKAHEAD epochs in the future

The cache is safe the function is not. I am not changing function behavior with the memoization

setup.py
protolambda: Ah good catch. This is a remainder of other optimizations. When you traverse the pre-computed attester statuses once, you only need to re-use base reward very short-lived. But not in the current spec, this needs to have much more capacity indeed.

tests/core/pyspec/eth2spec/test/context.py
protolambda: yes, the middle of the linear distribution being the max-effective balance. Lots of different distributions work, open for suggestions.

tests/core/pyspec/eth2spec/test/phase_0/epoch_processing/test_process_rewards_and_penalties.py
protolambda: Yes, to spot regression where we have that bug again.

setup.py
djrtwo: ah, right. the function won't allow the call.

Issue #1531, opened 2019-12-17
[Networking] Add `hello` message


mcdee:
At current when two peers first connect they exchange some information about the current state of their chains with the `status` message, but there is some information that can be useful to help peers know how to interact with each other.  This should not be added to the `status` message, as the information is static per connection and sending it with every status update would be wasteful.  Instead, a new handshake message `hello` is proposed.

An initial stab at the contents of a `hello` message might be:

  - genesis state root (32 bytes)
  - genesis fork version (4 bytes)
  - client name/semver e.g. `prysm/1.2.3` (bytes)
  - supported version(s) of the networking spec (repeated uint32)

This provides a basis for understanding if the two peers could talk to each other, even if either or both of them have no state information beyond genesis.  It also allows for peers to tweak their behaviour depending on known idiosyncrasies in particular client/spec versions.

AgeManning:
If all clients supported the identify libp2p protocol: https://github.com/libp2p/specs/tree/master/identify
Would this be sufficient information for what you are chasing. 

Is the fork version already in `Status`?

Perhaps we could add the genesis state root to the `agent_version` in identify? 

I ask, because it's something most libp2p-compatible clients would support, and already gives us a clients' supported libp2p protocols and versions.

mcdee:
Thanks for the link I was unaware of this function.  There are two areas to understand for suitability:

  1. when/how is this triggered, how much control does the application have?
  2. can we fit the data we want in to the structures supplied?

I'll have a read through and see exactly how this works and report back.

AgeManning:
The application can configure an interval which periodically updates the information. It's also used to help peers discover their external IP's as well as each other's capabilities.

The `agent_version` is configurable and is an arbitrary string, I believe of arbitrary length. The protocols and versions are the libp2p protocols we support, in addition to the stream multiplexers and potentially the encryption protocols. 

For us, it would tell us, which RPC protocols we support, which versions and encodings and also whether we support gossipsub or not. 

There might also be some relevant information about this here: https://docs.libp2p.io/concepts/nat/

mcdee:
Looking at https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/identify/id.go it appears the only thing that can be set by applications is the `agent_version` value.  Although we could use this as a compound field it would be, in my opinion, fragile compared to a well-formed message.

dryajov:
[...]

PR #1673, opened 2020-03-23
[Draft] PING and ENR rpc protocol

AgeManning:
_This is an initial draft aimed for discussion for adding a PING and ENR protocol to the RPC_

## Overview

For client's to adopt dynamic subnets efficiently, a mechanism is required to obtain the ENR (or subnet bitfields) from peers that connect to clients via libp2p rather than discovery (see #1637). Furthermore, we currently have not specified a PING-like protocol to check for peer liveness on the libp2p stack. 

Although we could use periodic STATUS messages to check for peer liveness, it may be beneficial to decouple the liveness check from STATUS messages, as we will also need a mechanism to update ENRs/subnet fields periodically.

## Proposed Solution

This PR provides a DRAFT version of adding a PING and ENR RPC method to our RPC. 

The PING protocol, is lightweight and is designed to send a small amount of information which indicates the current state of the node to other peers as well as checking their liveness. In it's current form, the PING protocol sends the ENR sequence number, however we may want to change this to something more general. 

Given knowledge of our peers current ENR sequence numbers (via the PING protocol) a client may then request an updated ENR from it's peers if its local record of the peers ENR is out-dated (ENR sequence is lower). This method also allows clients to request ENR's for newly connected peers if their ENR's are unknown. 

## Obvious points of modification

This is a draft aimed for discussion. Some obvious points we may want to consider changing are:
- The information sent in the PING protocol. This is currently ENR focused, we may want to make it more general.
- The structure of the ENR response. We don't have SSZ encodings for ENR. We could encode the base64 version of the ENR or use the RLP encoding. Currently this is specified to SSZ encode the response which contains the RLP encoded bytes of the ENR. 
- The ENR expiry response. Currently this is a list of Slots. Depending on implementations, long-lived subscriptions may want to end on Epoch's rather than Slots, in which case we may want to change this to Epoch. The actual structure of how to relay timeout information may want to be modified.
- ENR Request - The ENR request doesn't not send any data in the stream, which is unlike all other RPC methods. Potentially we want to send some information here.

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -601,6 +601,69 @@ Clients MUST support requesting blocks since the latest finalized epoch.
 
 Clients MUST respond with at least one block, if they have it. Clients MAY limit the number of blocks in the response.
 
+#### Ping
+
+**Protocol ID:** `/eth2/beacon_chain/req/ping/1/`
+
+Request Content:
+
+```
+(
+  uint64
+)
+```
+
+Response Content:
+
+```
+(
+  uint64
+)
+```
+
+Sent intermittently, the PING protocol checks liveness of connected peers.
+Peers send and respond with their local ENR sequence number.
+
+A client may then determine if their local record of a peer's ENR is up to date
+and may request an updated version via the ENR RPC method if not.
+
+The request MUST be encoded as an SSZ-field.
+
+The response MUST consist of a single `response_chunk`.
+
+#### Enr
+
+**Protocol ID:** `/eth2/beacon_chain/req/enr/1/`
+
+No Request Content.
+
+
+Response Content:
+
+```
+(
+  enr: Bytes,
+  subnet_expiries: []Slot
+)
+```
+
+Requests the ENR of a peer. The request opens and negotiates the stream without
+sending any request content. Once established the receiving peer responds with
+it's local up-to-date ENR along with a list of `Slot` corresponding to
+the expiry of all long-lived subnets specified by the `attnets` field in the
+sent ENR.
+
+The `subnet_expiries` list MUST have a length corresponding to the number of
+true values in the `attnets` bitfield. The order of the slots in `subnet_expiries`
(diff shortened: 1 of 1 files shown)


mcdee:
I would love to see some consideration of #1531 here.  Seems like the forkID from eth1 could be a good candidate for inclusion, and *please* could we have a peer name/version included before we head in to the multiclient testnet?

nisdas:
> The ENR expiry response. Currently this is a list of Slots. Depending on implementations, long-lived subscriptions may want to end on Epoch's rather than Slots, in which case we may want to change this to Epoch. The actual structure of how to relay timeout information may want to be modified.

It would make sense for this to be epoch based rather than slots, we do use epoch as a base value whenever specifying params such as `EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION`  https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/validator.md#misc .

AgeManning:
> It would make sense for this to be epoch based rather than slots, we do use epoch as a base value whenever specifying params such as EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/validator.md#misc

This is why I think it depends on implementations. If a validator joins halfway through an epoch, do you join the random subnet then, and then wait EPOCHS_PER_RANDOM_SUBNET_SUSCRIPTION before ending the subscription on a particular slot half-way through the epoch. Or do you end the subscription on an epoch boundary rounding up/down, or do you always start the subscription on the next epoch and end on epoch boundaries?

I personally think it's easier to just subscribe instantly when a validator connects and have a fixed timeout (which means we end on random slots) rather than adding another timeout to wait until the next epoch boundary and subscribe then. 

But if everyone is subscribing/unsubscribing on epoch boundaries, that could also be potentially cleaner.

djrtwo:
closing in favor of #1684

PR #1677, opened 2020-03-24
Correct/update test format documentation

gnattishness:
Existing test format documentation needs updating to reflect that relevant tests take signed envelopes as input.

* `block_header` operation to accept input of type `BeaconBlock`
* `voluntary_exit` operation to accept input of type
  `SignedVoluntaryExit`.
* sanity/blocks to note that inputs are of type `SignedBeaconBlock`

### DIFF
--- a/tests/formats/operations/README.md
+++ b/tests/formats/operations/README.md
@@ -18,11 +18,11 @@ A YAML-encoded `BeaconState`, the state before applying the operation.
 
 Also available as `pre.ssz`.
 
-### `<operation-name>.yaml`
+### `<input-name>.yaml`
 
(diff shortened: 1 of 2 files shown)


tests/formats/operations/README.md
```diff
@@ -18,11 +18,11 @@ A YAML-encoded `BeaconState`, the state before applying the operation.
 
 Also available as `pre.ssz`.
 
-### `<operation-name>.yaml`
+### `<input-name>.yaml`
```
gnattishness: Also note that `input-name` is the more accurate Operations table field used for file naming, though there is only a distinction for `block_header`, which takes a `block.yaml` not a `block_header.yaml`

PR #1680, opened 2020-03-24
Fix bug in attestation reward calculation

michaelsproul:
In v0.11.0 rewards are calculated incorrectly because of a typo in `get_attestation_deltas`. Rather than accumulating rewards using `+=`, the code overwrites the previous rewards with `=` 😳


djrtwo:
hot damn

djrtwo:
THANK YOU

Cherry-picked your commit to new branch based on `v011x` to get this out for immediate release.
See #1682

PR #1682, opened 2020-03-24
Attestation reward bug

djrtwo:
[replaces #1680]

Bug found by @michaelsproul in rewards calculations introduced in `v0.11.0`.
PR to `v011x` for immediate release


protolambda:
Splitting up that rewards function in the spec would be great for testing. All rewards mixed together in one function obfuscate what is really happening. Our mistake for not catching this bug earlier during release, but a good motivation for de-obfuscating this function.

ethers:
Can we have a link to the tests for this fix?  (probably would have helped avoid issues like https://github.com/PegaSysEng/teku/pull/1703)

PR #1684, opened 2020-03-25
MetaData and Ping

djrtwo:
Based on @AgeManning's #1673 and the discussion on the [recent networking call](https://hackmd.io/@benjaminion/rkEn7C_88#Networking-edge-cases).

* Add local notion of "MetaData" that has a sequence number. Currently just includes `attnets`
* Add `Ping` that sends back and forth `meta_data.seq_number`
* Add `GetMetaData` RPC method to request a nodes most recent `MetaData`
* Clarify when the ENR `attnets` field is optional

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -601,6 +601,69 @@ Clients MUST support requesting blocks since the latest finalized epoch.
 
 Clients MUST respond with at least one block, if they have it. Clients MAY limit the number of blocks in the response.
 
+#### Ping
+
+**Protocol ID:** `/eth2/beacon_chain/req/ping/1/`
+
+Request Content:
+
+```
+(
+  uint64
+)
+```
+
+Response Content:
+
+```
+(
+  uint64
+)
+```
+
+Sent intermittently, the PING protocol checks liveness of connected peers.
+Peers send and respond with their local ENR sequence number.
+
+A client may then determine if their local record of a peer's ENR is up to date
+and may request an updated version via the ENR RPC method if not.
+
+The request MUST be encoded as an SSZ-field.
+
+The response MUST consist of a single `response_chunk`.
+
+#### Enr
+
+**Protocol ID:** `/eth2/beacon_chain/req/enr/1/`
+
+No Request Content.
+
+
+Response Content:
+
+```
+(
+  enr: Bytes,
+  subnet_expiries: []Slot
+)
+```
+
+Requests the ENR of a peer. The request opens and negotiates the stream without
+sending any request content. Once established the receiving peer responds with
+it's local up-to-date ENR along with a list of `Slot` corresponding to
+the expiry of all long-lived subnets specified by the `attnets` field in the
+sent ENR.
+
+The `subnet_expiries` list MUST have a length corresponding to the number of
+true values in the `attnets` bitfield. The order of the slots in `subnet_expiries`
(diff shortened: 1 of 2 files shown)


specs/phase0/p2p-interface.md
```diff
@@ -601,6 +620,60 @@ Clients MUST support requesting blocks since the latest finalized epoch.
...
+Response Content:
+
+```
+(
+  uint64
+)
+```
+
+Sent intermittently, the `Ping` protocol checks liveness of connected peers.
+Peers send and respond with their local metadata sequence number (`MetaData.seq_number`).
```
```suggestion
Peers request and respond with their local metadata sequence number (`MetaData.seq_number`).
```

specs/phase0/p2p-interface.md
```diff
@@ -195,6 +198,22 @@ This section outlines constants that are used in this spec.
...
+```
+(
+  seq_number: uint64
+  attnets: Bitvector[ATTESTATION_SUBNET_COUNT]
+)
+```
+
+Where
+
+- `seq_number` is a `uint64` starting at `0` used to version the node's metadata. If any other field in the local `MetaData` changes, the node MUST increment `seq_number` by 1.
```
protolambda: Given the different contents, does that mean it is different than an ENR sequence number? Stating that it is different, and can be out of sync, would be good, since the names are the same.

specs/phase0/p2p-interface.md
```diff
@@ -601,6 +620,60 @@ Clients MUST support requesting blocks since the latest finalized epoch.
...
+```
+(
+  uint64
+)
+```
+
+Sent intermittently, the `Ping` protocol checks liveness of connected peers.
+Peers send and respond with their local metadata sequence number (`MetaData.seq_number`).
+
+If the peer does not respond to the `Ping` request, the client MAY disconnect from the peer.
```
protolambda: Same timeout applies here as other RPC requests? Or should meta-data requests have a stricter timeout?

specs/phase0/p2p-interface.md
```diff
@@ -601,6 +620,60 @@ Clients MUST support requesting blocks since the latest finalized epoch.
...
+
+The request MUST be encoded as an SSZ-field.
+
+The response MUST consist of a single `response_chunk`.
+
+#### GetMetaData
+
+**Protocol ID:** `/eth2/beacon_chain/req/metadata/1/`
+
+No Request Content.
```
protolambda: Interesting, the first RPC call without a request content. Does that mean an empty payload, but still a request header with a 0-length in it? Or no bytes at all?

specs/phase0/p2p-interface.md
djrtwo: yeah, they can be. Will do so

specs/phase0/p2p-interface.md
djrtwo: I think `Ping` can have the same timeout.
That said, the 5s seems long for any request..

specs/phase0/p2p-interface.md
djrtwo: This came from @AgeManning in his previous PR. 
I interpreted it as no bytes at all

protolambda: Think we can spec it both ways, let's ask on the call

djrtwo: Talked to age, was the intention and works on libp2p...

Will change it later if Raul chimes in otherwise

PR #1685, opened 2020-03-25
Do not require non-aggregators to subscribe to attnets

djrtwo:
As suggested by @AgeManning and discussed on the recent [networking call](https://hackmd.io/@benjaminion/rkEn7C_88#Networking-edge-cases), this PR distinguishes from an aggregating and non-aggregating beacon committee member for the purposes of attestation subnets.

* Aggregators -- fully subscribe to the attestation subnet to receive newly published attestations for aggregation
* Non-aggregators -- Only find sufficient peers on the attestation subnet (rather than subscribe) to allow for publishing of attestations. Reduces bandwidth requirement on the node, and keeps attnets small for quick propagation.

### DIFF
--- a/specs/phase0/validator.md
+++ b/specs/phase0/validator.md
@@ -200,9 +200,9 @@ The beacon chain shufflings are designed to provide a minimum of 1 epoch lookahe
 
 Specifically a validator should:
 * Call `get_committee_assignment(state, next_epoch, validator_index)` when checking for next epoch assignments.
-* Find peers of the pubsub topic -- `committee_index{committee_index % ATTESTATION_SUBNET_COUNT}_beacon_attestation` -- and join if aggregator.
+* Find peers of the pubsub topic `committee_index{committee_index % ATTESTATION_SUBNET_COUNT}_beacon_attestation`.
     * If an _insufficient_ number of current peers are subscribed to the topic, the validator must discover new peers on this topic. Via the discovery protocol, find peers with an ENR containing the `attnets` entry such that `ENR["attnets"][committee_index % ATTESTATION_SUBNET_COUNT] == True`.
-    * If the validator is assigned to be an aggregator for the slot (see `is_aggregator()`), then for any peer subscribed to the topic the validator sends a `subscribe` message for this topic.
+    * If the validator is assigned to be an aggregator for the slot (see `is_aggregator()`), then subscribe to the topic.
 
 *Note*: If the validator is _not_ assigned to be an aggregator, the validator only needs sufficient number of peers on the topic to be able to publish messages. The validator does not need to _subscribe_ and listen to all messages on the topic.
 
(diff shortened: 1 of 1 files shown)


specs/phase0/validator.md
```diff
@@ -200,9 +200,11 @@ The beacon chain shufflings are designed to provide a minimum of 1 epoch lookahe
 
 Specifically a validator should:
 * Call `get_committee_assignment(state, next_epoch, validator_index)` when checking for next epoch assignments.
-* Join the pubsub topic -- `committee_index{committee_index % ATTESTATION_SUBNET_COUNT}_beacon_attestation`.
-    * For any current peer subscribed to the topic, the validator simply sends a `subscribe` message for the new topic.
+* Find peers of the pubsub topic -- `committee_index{committee_index % ATTESTATION_SUBNET_COUNT}_beacon_attestation` -- and join if aggregator.
```
protolambda:
```suggestion
* Find peers of the pubsub topic `committee_index{committee_index % ATTESTATION_SUBNET_COUNT}_beacon_attestation`.
```

This seems double with the other statement, shortening it.

specs/phase0/validator.md
```diff
@@ -200,9 +200,11 @@ The beacon chain shufflings are designed to provide a minimum of 1 epoch lookahe
...
     * If an _insufficient_ number of current peers are subscribed to the topic, the validator must discover new peers on this topic. Via the discovery protocol, find peers with an ENR containing the `attnets` entry such that `ENR["attnets"][committee_index % ATTESTATION_SUBNET_COUNT] == True`.
+    * If the validator is assigned to be an aggregator for the slot (see `is_aggregator()`), then for any peer subscribed to the topic the validator sends a `subscribe` message for this topic.
```
protolambda:
```suggestion
    * If the validator is assigned to be an aggregator for the slot (see `is_aggregator()`), then subscribe to the topic.
```
We do not really use the `subscribe` message terminology anywhere else in the p2p spec. Better not use it and avoid confusion with different gossipsub implementations. There are subscribe events, but only graft/join messages in gossipsub (afaik).

mkalinin:
Do I understand correctly that after this change the mesh of a subnet will consist only from aggregators assigned to particular slot? If yes, doesn't it increase a surface for eclipse attack?

zilm13:
plus to @mkalinin: open subnetwork of many nodes provides better security properties than small subnetwork even if its private.
Plus average Kademlia table of long running node consists of ~150 peers, which means you will have 2-3 nodes(/64) from any subnetwork in it (if you have only validators in network) according to the previous spec to start search from an 0 with new spec.

djrtwo:
> which means you will have 2-3 nodes(/64) 

Assuming that there is a distribution of validators across nodes and that each _validator_ subscribes to a random subnet (rather than just each node), then there would be multiple more than 2-3 nodes of attnet capabilites in your routing table.

> Do I understand correctly that after this change the mesh of a subnet will consist only from aggregators assigned to particular slot?

No, the mesh consists of the persistent committee _plus_ the aggregators for a range of slots (something like 5 to 20 slots worth of aggregators, assuming aggregators join some amount of slots in advance to prepare). Persistent committees are simulated in phase 0 (see [Phase 0 Attestation Subnet Stability](https://github.com/ethereum/eth2.0-specs/blob/v0.11.0/specs/phase0/validator.md#phase-0-attestation-subnet-stability))

djrtwo:
addressed comments @protolambda

PR #1693, opened 2020-03-29
Fix `INACTIVITY_PENALTY_QUOTIENT` description

hwwhww:
## Issue
It's a very old leftover from https://github.com/ethereum/eth2.0-specs/pull/949#discussion_r278051185: when we didn't update the constant number description when we changed `INACTIVITY_PENALTY_QUOTIENT` from `2**24` to `2**25`.

## How to fix it
This PR updates corresponding `INVERSE_SQRT_E_DROP_TIME` to the right number. Alternatively, we can consider removing this paragraph as @JustinDrake suggested.

/cc @benjaminion for the annotated spec project.

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -245,7 +245,7 @@ The following values are (non-configurable) constants used throughout the specif
 | `INACTIVITY_PENALTY_QUOTIENT` | `2**25` (= 33,554,432) |
 | `MIN_SLASHING_PENALTY_QUOTIENT` | `2**5` (= 32) |
 
-- The `INACTIVITY_PENALTY_QUOTIENT` equals `INVERSE_SQRT_E_DROP_TIME**2` where `INVERSE_SQRT_E_DROP_TIME := 2**12 epochs` (about 18 days) is the time it takes the inactivity penalty to reduce the balance of non-participating validators to about `1/sqrt(e) ~= 60.6%`. Indeed, the balance retained by offline validators after `n` epochs is about `(1 - 1/INACTIVITY_PENALTY_QUOTIENT)**(n**2/2)`; so after `INVERSE_SQRT_E_DROP_TIME` epochs, it is roughly `(1 - 1/INACTIVITY_PENALTY_QUOTIENT)**(INACTIVITY_PENALTY_QUOTIENT/2) ~= 1/sqrt(e)`.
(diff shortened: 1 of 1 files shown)


djrtwo:
Why was it changed in #949?
I've been operating under the assumption of ~2.5 weeks. Not ~3.5 weeks

JustinDrake:
> Why was it changed in #949? It was listed as "cosmetic" in the PR -- CC @JustinDrake

The previous formula had a `// 2` (see `INACTIVITY_PENALTY_QUOTIENT // 2` [here](https://github.com/ethereum/eth2.0-specs/pull/949/files#diff-51a43328a58414e132a744f3771f018cL1637)) which I removed and counter-balanced by changing the constant. The intent was to preserve functionality, i.e. make it a cosmetic change. It is entirely possible I messed up and made a substantive change 😂

hwwhww:
@JustinDrake sorry, I didn't notice you removed `// 2` in #949 when I read it for this PR. So *if*the previous numbers were correct, this section should be:


The `INACTIVITY_PENALTY_QUOTIENT // 2` equals `INVERSE_SQRT_E_DROP_TIME**2` where `INVERSE_SQRT_E_DROP_TIME := 2**12 epochs` (about 18 days) is the time it takes the inactivity penalty to reduce the balance of non-participating validators to about `1/sqrt(e) ~= 60.6%`. Indeed, the balance retained by offline validators after `n` epochs is about `(1 - 1/INACTIVITY_PENALTY_QUOTIENT)**(n**2/2)`; so after `INVERSE_SQRT_E_DROP_TIME` epochs, it is roughly `(1 - 1/INACTIVITY_PENALTY_QUOTIENT)**(INACTIVITY_PENALTY_QUOTIENT/2) ~= 1/sqrt(e)`.


However, I did some git log trace 🔎 
1. The mysterious `// 2` was added in https://github.com/ethereum/eth2.0-specs/commit/53dd49167b095b02f91dea4629e48aff0dd703c8 (#347)
    > * Adds a `// 2` to the inactivity leak to compensate for it being applied twice
2. But, in https://github.com/ethereum/eth2.0-specs/commit/d1d1b73fb1783b564556a48ea86d69d8dd1003e7 (#826), changes, the twice penalties became just one.
    > Inactivity leak penalty specifically on missing the target, not both the target and the source

Am I correct that we don't even need `// 2` after #826, so the true fix should be, simply change `INACTIVITY_PENALTY_QUOTIENT` from `2**25` to `2**24`?

/cc @djrtwo @vbuterin for more 👀 and better memory.

vbuterin:
It does indeed seem like `INACTIVITY_PENALTY_QUOTIENT` should be changed to `2**24`. The description that after `n` epochs the log-balance decreases by `(n**2 / 2) / INACTIVITY_PENALTY_QUOTIENT`, and so when `n = INVERSE_SQRT_E_DROP_TIME = 2**12` the log-balance drops by 1/2, ie. a sqrt-e factor drop, is correct.

hwwhww:
closing by #1712

PR #1704, opened 2020-04-02
phase 1 validator guide (round 1)

djrtwo:
Add phase 1 validator guide.

Add the following checked items. Leave the remaining for a subsequent PR(s)

- [x] Block proposal
- [ ] Attesting
    - [x] Join shard block subnet in addition to attestation subnet
    - [ ] On-time attestation
        - [x] Attestation data
        - [ ] Add custody bits blocks [Waiting on custody rewrite #1889]
        - [x] Compute signature
        - [x] aggregation
    - [ ] Late attestation (no custody_bits_blocks, stub `shard_transition`, when to broadcast, when to aggregate)
    - [ ] [Think about] Split `shard_transition` from `attestation_data` in p2p for deduplication gains (OR split `attestation_data` from `attestation`...)
- [x] Light client committee
    - [x] clean up beacon chain spec and add validator duties
- [ ] Shard duties
    - [ ] Persistent committee subnets and lookahead (replacement for random subnets)
    - [ ] Shard block creation and broadcast
- [ ] Additional slashing avoidance


specs/phase1/validator.md
```diff
@@ -0,0 +1,462 @@
...
+        shard_attestations = [
+            attestation for attestation in attestations
+            if attestation.data.index == committee_index and attestation.data.slot == slot
+        ]
+        committee = get_beacon_committee(state, state.slot, committee_index)
+
+        # Loop over all shard transition roots, looking for a winning root
+        shard_transition_roots = set([a.data.shard_transition_root for a in shard_attestations])
+        for shard_transition_root in sorted(shard_transition_roots):
+            transition_attestations = [
```
terencechain, line 158: From grouping `transition_attestations`, to computing `winning_root`. This can be refactored into its own helper function and reuse in `process_crosslink_for_shard`.  It'll reduce disparity and make testing easier down the road

specs/phase1/validator.md
```diff
@@ -0,0 +1,462 @@
...
+Selection proofs are provided in `LightAggregateAndProof` to prove to the gossip channel that the validator has been selected as an aggregator.
+
+`LightAggregateAndProof` messages are signed by the aggregator and broadcast inside of `SignedLightAggregateAndProof` objects to prevent a class of DoS attacks and message forgeries.
+
+First, `light_aggregate_and_proof = get_light_aggregate_and_proof(state, validator_index, aggregate_light_client_vote, privkey)` is constructed.
+
+```python
+def get_light_aggregate_and_proof(state: BeaconState,
+                                  aggregator_index: ValidatorIndex,
+                                  aggregate: Attestation,
```
terencechain, line 425:
```suggestion
                                  aggregate_light_client_vote: LightClientVote,
```

specs/phase1/validator.md
```diff
@@ -0,0 +1,462 @@
...
+                                  privkey: int) -> LightAggregateAndProof:
+    return LightAggregateAndProof(
+        aggregator_index=aggregator_index,
+        aggregate=aggregate,
```
terencechain, line 429:
```suggestion
        aggregate_light_client_vote = LightClientVote,
```

specs/phase1/validator.md
```diff
@@ -0,0 +1,462 @@
...
+    signing_root = compute_signing_root(aggregate_and_proof, domain)
+    return bls.Sign(privkey, signing_root)
+```
+
+##### `LightAggregateAndProof`
+
+```python
+class LightAggregateAndProof(Container):
+    aggregator_index: ValidatorIndex
+    aggregate: Attestation
```
terencechain, line 451:
```suggestion
    aggregate_light_client_vote: LightClientVote
```

specs/phase1/validator.md
```diff
@@ -251,24 +251,67 @@ A validator should create and broadcast the `attestation` to the associated atte
...
+    shard_block_lengths = []
+
+    shard_state = beacon_state.shard_states[shard]
+    shard_block_slots = [shard_block.message.slot for shard_block in shard_blocks]
+    for slot in get_offset_slots(beacon_state, shard):
+        if slot in shard_block_slots:
+            shard_block = shard_blocks[shard_block_slots.index(slot)]
+            shard_data_roots.append(hash_tree_root(shard_block.message.body))
+        else:
+            shard_block = SignedShardBlock(message=ShardBlock(slot=slot))
```
hwwhww, line 284: FYI I have a bug here that I will push it to shard fork choice rule PR to make tests pass:
```suggestion
            shard_block = SignedShardBlock(message=ShardBlock(slot=slot, shard=shard))
```
^^^^ need to set `shard` for empty proposal case.

djrtwo: thanks!

specs/phase1/validator.md
```diff
@@ -0,0 +1,520 @@
...
+
+Exactly `MAX_SHARDS` [`ShardTransition`](./beacon-chain#shardtransition) objects are included in the block. Default each to an empty `ShardTransition()`. Then for each committee assigned to the slot with an associated `committee_index` and `shard`, set `shard_transitions[shard] = full_transitions[winning_root]` if the committee had enough weight to form a crosslink this slot.
+
+Specifically:
+* Call `shards, winning_roots = get_successful_shard_transitions(state, block.slot, attestations)`
+* Let `full_transitions` be a dictionary mapping from the `shard_transition_root`s found in `attestations` to the corresponding full `ShardTransition`
+* Then for each `shard` and `winning_root` in `zip(shards, winning_roots)` set `shard_transitions[shard] = full_transitions[winning_root]`
+
+```python
+def get_successful_shard_transitions(state: BeaconState,
```
hwwhww, line 139: Hmmm `get_successful_shard_transitions` sounds like it returns `Sequence[ShardTransition]`... I'm not sure about how to name it better 😅 , something like `get_shard_winning_roots`?

specs/phase1/validator.md
```diff
@@ -0,0 +1,520 @@
...
+        shard_block_lengths.append(len(shard_block.message.body))
+
+    return shard_states, shard_data_roots, shard_block_lengths
+```
+
+```python
+def get_shard_transition(beacon_state: BeaconState,
+                         shard: Shard,
+                         shard_blocks: Sequence[SignedShardBlock]) -> ShardTransition:
+    offset_slots = get_offset_slots(beacon_state, shard)
```
hwwhww, line 297: Note that to call `get_offset_slots` correctly, the given `beacon_state` should be at *the next slot* of `attestation.data.slot` (where the result `ShardTransition` gets included, not `head_state`.

Alternatively, we can use `compute_offset_slots`.
```python
on_time_slot = Slot(state.slot + 1)
compute_offset_slots(get_latest_slot_for_shard(state, shard), on_time_slot)
```

specs/phase1/validator.md
```diff
@@ -0,0 +1,520 @@
...
+            shard_block = shard_blocks[shard_block_slots.index(slot)]
+            shard_data_roots.append(hash_tree_root(shard_block.message.body))
+        else:
+            shard_block = SignedShardBlock(message=ShardBlock(slot=slot))
+            shard_data_roots.append(Root())
+        shard_state = get_post_shard_state(beacon_state, shard_state, shard_block.message)
+        shard_states.append(shard_state)
+        shard_block_lengths.append(len(shard_block.message.body))
+
+    return shard_states, shard_data_roots, shard_block_lengths
```
hwwhww, line 290: nitpick: reorder to the sequence in `ShardTransition`:
```suggestion
    return shard_block_lengths, shard_data_roots, shard_states
```

specs/phase1/validator.md
```diff
@@ -0,0 +1,520 @@
...
+##### Head shard root
+
+Set `attestation_data.head_shard_root = hash_tree_root(head_shard_block)`.
+
+##### Shard transition
+
+Set `shard_transition` to the value returned by `get_shard_transition(head_state, shard, shard_blocks)`.
+
+```python
+def get_shard_state_transition_result(
```
hwwhww, line 267: What do you think about renaming it to `get_shard_state_transition_fields`?

specs/phase1/validator.md
```diff
@@ -251,24 +251,67 @@ A validator should create and broadcast the `attestation` to the associated atte
...
+*Note*: We assume that the fork choice only follows branches with valid `offset_slots` with respect to the most recent beacon state shard transition for the queried shard.
 
 ##### Head shard root
 
 Set `attestation_data.head_shard_root = hash_tree_root(head_shard_block)`.
 
 ##### Shard transition
 
-Set `shard_transition` to the value returned by `get_shard_transition()`.
+Set `shard_transition` to the value returned by `get_shard_transition(head_state, shard, shard_blocks)`.
```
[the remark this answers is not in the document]
hwwhww: Good points!

Added changes at #1773
1. Added rules to enforce the shard block is at offset slots  https://github.com/ethereum/eth2.0-specs/pull/1773/commits/49cbc7c7f57a0343a45284d41f04f2d95d3dc570
2. Added `get_pending_shard_blocks` helper that returns the `shard_blocks` at offset slots till head beacon slot: https://github.com/ethereum/eth2.0-specs/pull/1773/commits/465f2306a2418c1e62df5a249c682aa4c3a28380 returns correct data.

specs/phase1/validator.md
hwwhww, line 139: Note that since we need to calculate active validator in this function, this `state` has to be at slot `block.slot` to get the correct `current_epoch` value. Is it implied in phase 0 spec that `state` is the lookahead state?

specs/phase1/validator.md
hwwhww: Second thought: https://github.com/ethereum/eth2.0-specs/pull/1875
It seems better to pass the `on_time_slot` (`attestation.data.slot + 1`)

specs/phase1/validator.md
```diff
@@ -0,0 +1,520 @@
...
+
+##### Early derived secret reveals
+
+Up to `MAX_EARLY_DERIVED_SECRET_REVEALS`, [`EarlyDerivedSecretReveal`](./custody-game.md#earlyderivedsecretreveal) objects can be included in the `block`. The early derived secret reveals must satisfy the verification conditions found in [early derived secret reveal processing](./custody-game.md#custody-key-reveals). The validator receives a small "whistleblower" reward for each early derived secrete reveal included.
+
+##### Shard transitions
+
+Exactly `MAX_SHARDS` [`ShardTransition`](./beacon-chain#shardtransition) objects are included in the block. Default each to an empty `ShardTransition()`. Then for each committee assigned to the slot with an associated `committee_index` and `shard`, set `shard_transitions[shard] = full_transitions[winning_root]` if the committee had enough weight to form a crosslink this slot.
```
terencechain, line 131: Side question. Should `MAX_SHARDS` really be 1024 as it's currently defined? That's a lot of sparse `ShardTransition()` inside beacon block even at the best case (e.g. 960)

It might be better to align `MAX_SHARDS` to `MAX_COMMITTEES_PER_SLOT`, and increase the shard count in an event of a network upgrade via social consensus

specs/phase1/validator.md
```diff
@@ -0,0 +1,520 @@
...
+
+Specifically:
+* Call `shards, winning_roots = get_successful_shard_transitions(state, block.slot, attestations)`
```
terencechain, line 134: Can we add a comment to clarify what these `attestations` are?

specs/phase1/validator.md
```diff
@@ -0,0 +1,520 @@
...
+
+#### Timing
+
+Note the timing of when to create/broadcast is altered from Phase 1.
+
+A validator should create and broadcast the `attestation` to the associated attestation subnet when either (a) the validator has received a valid `BeaconBlock` from the expected beacon block proposer and a valid `ShardBlock` for the expected shard block proposer for the assigned `slot` or (b) one-half of the `slot` has transpired (`SECONDS_PER_SLOT / 2` seconds after the start of `slot`) -- whichever comes _first_.
+
+#### Attestation data
+
+`attestation_data` is constructed in the same manner as Phase 0 but uses `FullAttestationData` with the addition of two fields -- `head_shard_root` and `shard_transition`.
```
terencechain, line 249: s/head_shard/shard_head in the latest `dev`

specs/phase1/validator.md
djrtwo: yup, good name!

specs/phase1/validator.md
djrtwo: note that `head_state` is transitioned up to the assigned slot

djrtwo: I'm not certain we need to pass in `on_time_slot` here.

Happy to discuss more though

djrtwo: EDIT, I see what you mean here. thinking

djrtwo: I think I might prefer not passing in `on_time_slot`. The complexity in #1875 seems pretty high.

I'm exploring the idea a bit more now

specs/phase1/validator.md
djrtwo: It's stated that it must be in the epoch in question to successfully figure out your proposer assignments. I'll make sure this is very clear here

specs/phase1/validator.md
hwwhww: > note that `head_state` is transitioned up to the assigned slot so head_state already is at the `on_time_slot`

Ah, I see. I thought it's not transitioned head and considered the skipped beacon block case. So using `Slot(state.slot + 1)` is enough for v-guide.

p.s. the complexity of #1875 is mainly in `verify_shard_block_message` change. 😄

specs/phase1/validator.md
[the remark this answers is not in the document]
djrtwo: Validators actual contrsuct a `FullAttestationData` that has a the full `shard_transition` rather than the `root`. This is sent on the wire (will note in validator and p2p spec). The block producers compress the `shard_transition`s into `root`s and only including winning `ShardTransition` objects separately

specs/phase1/validator.md
djrtwo: Yeah, I see. Getting this cleaned up, then going back in there to try to understand!

specs/phase1/validator.md
djrtwo: Yeah, I see that.. I think the argument for having it in there is that 0s would be almost entirely compressible on the wire (via snappy).

I think it's reasonable to change it.. but we can discuss in another thread

specs/phase1/validator.md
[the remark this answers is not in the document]
terencechain: Oops, I missread the sentence and thought it was referring to `attestation_data`. My b!

specs/phase1/validator.md
djrtwo: they come from `block.body` so I just made that clear. thnaks!

hwwhww:
Followup note of https://github.com/ethereum/eth2.0-specs/pull/1704#discussion_r435238395:

These are some cases I see for attesting `shard_block_100` and `shard_block_101`:
image

If `head_state` is transitioned to the slot of current time, `Slot(state.slot + 1)` works for these cases. Any other case I missed?

terencechain:
> Followup note of [#1704 (comment)](https://github.com/ethereum/eth2.0-specs/pull/1704#discussion_r435238395):
> 
> These are some cases I see for attesting `shard_block_100` and `shard_block_101`:
> If `head_state` is transitioned to the slot of current time, `Slot(state.slot + 1)` works for these cases. Any other case I missed?
> Screen Shot 2020-06-05 at 5 50 53 PM

Do you consider **4a** and **4b** as 2 different cases? If yes, does it break current transition?

**4a**: B100 crosslinks S99
**4b**: B100 **did not** crosslinks S99,  B101 crosslinks S99.

hwwhww:
@terencechain 

Thank! It looks like an extended case!
I wanted to use "B100 crosslinks S99" to represent "the latest crosslink at the beginning". Rewrite it to:

chain_06_fail_crosslink_and_skipped_shard

specs/phase1/validator.md
```diff
@@ -0,0 +1,529 @@
...
+def get_light_client_slot_signature(state: BeaconState, slot: Slot, privkey: int) -> BLSSignature:
+    domain = get_domain(state, DOMAIN_LIGHT_SELECTION_PROOF, compute_epoch_at_slot(slot))
+    signing_root = compute_signing_root(slot, domain)
+    return bls.Sign(privkey, signing_root)
+```
+
+```python
+def is_light_client_aggregator(state: BeaconState, slot: Slot, slot_signature: BLSSignature) -> bool:
+    committee = get_light_client_committee(state, compute_epoch_at_slot(slot))
+    modulo = max(1, len(committee) // TARGET_LIGHT_CLIENT_AGGREGATORS_PER_SLOT)
```
hwwhww, line 459: Note that compare to `is_aggregator`, it's highly possible that `len(committee) == LIGHT_CLIENT_COMMITTEE_SIZE` and then `module == 16`. So the `is_light_client_aggregator` output would be more predictable.

specs/phase1/validator.md
```diff
@@ -0,0 +1,529 @@
...
+    period_start_epoch = get_current_epoch(state) + LIGHT_CLIENT_COMMITTEE_PERIOD % get_current_epoch(state)
+    next_committee = get_light_client_committee(state, period_start_epoch)
+    return index in next_committee
+```
+
+#### Light clent vote
+
+During a period of epochs that the validator is a part of the light client committee (`validator_index in get_light_client_committee(state, epoch)`), the validator creates and broadcasts a `LightClientVote` at each slot.
+
+A validator should create and broadcast the `light_client_vote` to the `light_client_votes` pubsub topic when either (a) the validator has received a valid block from the expected block proposer for the current `slot` or (b) two-thirds of the `slot` have transpired (`SECONDS_PER_SLOT / 3` seconds after the start of `slot`) -- whichever comes _first_.
```
hwwhww, line 391: Re timing, what do you think about 

- 0 second: beacon block proposer proposes a beacon block
- `SECONDS_PER_SLOT / 4` shard block proposer proposes a shard block
- `SECONDS_PER_SLOT 1 / 2`: attesters (including light client committee) build attestations
- `SECONDS_PER_SLOT 3 / 4` : aggregators (including light client) aggregate attestations

specs/phase1/validator.md
```diff
@@ -0,0 +1,529 @@
...
+    index: CommitteeIndex
+    # LMD GHOST vote
+    beacon_block_root: Root
+    # FFG vote
+    source: Checkpoint
+    target: Checkpoint
+    # Current-slot shard block root
+    shard_head_root: Root
+    # Full shard transition
+    shard_transition: ShardTransition
```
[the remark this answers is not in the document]
djrtwo: I've thought about this a little bit. I think we might want to gossip the `ShardTransition` (or `AttestationData` separately so that message deduplication naturally happens

Will save for another PR

specs/phase1/validator.md
djrtwo: yeah, I see the value in mirroring this in the roles

The main reason I staggered is because there are fewer steps in the LC flow so (1) better chance of full propagation between each step and (2) staggers local bursts of bandwidth and cpu between attestations and light client messages

thoughts on the above?

terencechain: Note to self. When phase1 run time is up, I'll provide the new numbers on proposer, attester, aggregator and LC's time spent for each operation

hwwhww: @djrtwo 
I see. Sounds reasonable to stagger the traffic now and see if we can optimize it by the data from @terencechain :)

specs/phase1/validator.md
```diff
@@ -0,0 +1,569 @@
...
+
+Beacon chain validator assignments to beacon committees and beacon block proposal are unchanged from Phase 0. See the [Phase 0 validator guide](../phase0/validator.md#validator-assignments) for details.
+
+### Lookahead
+
+Lookahead for beacon committee assignments operates in the same manner as Phase 0, but committee members must join a shard block pubsub topic in addition to the committee attestation topic.
+
+Specifically _after_ finding stable peers of attestation subnets (see Phase 0) a validator should:
+* Let `shard = compute_shard_from_committee_index(committe_index)`
+* Subscribe to the pubsub topic `shard_{shard}_shard_block` (attestation subnet peers should have this topic available).
```
hwwhww, line 98: nitpicking (not strongly): the first "shard" shows it's a shard block
```suggestion
* Subscribe to the pubsub topic `shard_{shard}_block` (attestation subnet peers should have this topic available).
```

tests/core/pyspec/eth2spec/test/helpers/shard_block.py
```diff
@@ -56,11 +56,14 @@ def get_shard_transitions(spec, parent_beacon_state, shard_blocks):
...
     on_time_slot = parent_beacon_state.slot + 1
     for shard, blocks in shard_blocks.items():
+        shard_transition = spec.get_shard_transition(parent_beacon_state, shard, blocks)
         offset_slots = spec.compute_offset_slots(
             spec.get_latest_slot_for_shard(parent_beacon_state, shard),
             on_time_slot,
         )
         len_offset_slots = len(offset_slots)
+        # TODO this is actually unsafe for long offset_slots
+        assert len_offset_slots == on_time_slot - parent_beacon_state.shard_states[shard].slot - 1
```
hwwhww, line 66: should be removed
```suggestion
```

specs/phase1/validator.md
```diff
@@ -0,0 +1,569 @@
...
+
+When `get_current_epoch(state) % LIGHT_CLIENT_COMMITTEE_PERIOD == LIGHT_CLIENT_COMMITTEE_PERIOD - LIGHT_CLIENT_PREPARATION_EPOCHS` each validator must check if they are in the next period light client committee by calling `is_in_next_light_client_committee()`.
+
+If the validator is in the next light client committee, they must join the `light_client_votes` pubsub topic to begin duties at the start of the next period.
+
+```python
+def is_in_next_light_client_committee(state: BeaconState, index: ValidatorIndex) -> bool:
+    current_source_epoch = compute_committee_source_epoch(get_current_epoch(state), LIGHT_CLIENT_COMMITTEE_PERIOD)
+    next_source_epoch = current_source_epoch + LIGHT_CLIENT_COMMITTEE_PERIOD
+    next_committee = get_light_client_committee(state, next_source_epoch)
```
hwwhww, line 423: The input `epoch` of `get_light_client_committee` should be the epoch that the light clients have to vote for, not the source epoch.

I suppose it should be:
```suggestion
    next_committee = get_light_client_committee(state,get_current_epoch(state) + LIGHT_CLIENT_COMMITTEE_PERIOD)
```

specs/phase1/validator.md
```diff
@@ -0,0 +1,569 @@
...
+    shards = []
+    winning_roots = []
+    online_indices = get_online_validator_indices(state)
+    committee_count = get_committee_count_at_slot(state, slot)
+    for committee_index in map(CommitteeIndex, range(committee_count)):
+        shard = compute_shard_from_committee_index(state, committee_index, slot)
+        # All attestations in the block for this committee/shard and current slot
+        shard_attestations = [
+            attestation for attestation in attestations
+            if attestation.data.index == committee_index and attestation.data.slot == slot
```
terencechain, line 158:
```suggestion
            if attestation.data.index == committee_index and attestation.data.slot == compute_previous_slot(slot)
```

Shouldn't `beacon_block` at slot N have `on_time_attestations` from slot N - 1 ?

specs/phase1/validator.md
```diff
@@ -0,0 +1,569 @@
...
+A validator should create and broadcast the `attestation` to the associated attestation subnet when either (a) the validator has received a valid `BeaconBlock` from the expected beacon block proposer and a valid `ShardBlock` for the expected shard block proposer for the assigned `slot` or (b) one-half of the `slot` has transpired (`SECONDS_PER_SLOT / 2` seconds after the start of `slot`) -- whichever comes _first_.
+
+#### Attestation data
+
+`attestation_data` is constructed in the same manner as Phase 0 but uses `FullAttestationData` with the addition of two fields -- `shard_head_root` and `shard_transition`.
+
+- Let `head_block` be the result of running the fork choice during the assigned slot.
+- Let `head_state` be the state of `head_block` processed through any empty slots up to the assigned slot using `process_slots(state, slot)`.
+- Let `shard_head_block` be the result of running the fork choice on the assigned shard chain during the assigned slot.
+- Let `shard_blocks` be the shard blocks in the chain starting immediately _after_ the most recent crosslink (`head_state.shard_transitions[shard].latest_block_root`) up to the `shard_head_block`.
```
terencechain, line 262: do you want to advertise the usage of method `get_pending_shard_blocks` here?

specs/phase1/validator.md
hwwhww: Agreed  👍  And you can use `is_on_time_attestation(state, attestation)` helper.
```suggestion
            if is_on_time_attestation(state, attestation) and attestation.data.index == committee_index
```

And simplify the `get_shard_winning_roots` arguments to:
```python
def get_shard_winning_roots(state: BeaconState,
                            attestations: Sequence[Attestation]) -> Tuple[Sequence[Shard], Sequence[Root]]:
    slot = state.slot
    ...

specs/phase1/validator.md
djrtwo: I'd actually argue that the above logic is the same, and I considered writing it as you had originally but wasn't sure if the logic was clear.

Making the change regardless. (curious if I am incorrect in saying that you'll get the same result either way)

specs/phase1/validator.md
djrtwo: yes! that didn't exist when I originally wrote it :)

specs/phase1/validator.md
hwwhww: @djrtwo 
No, they are not the same result. I argued the function logic correctness:
- `get_light_client_committee` itself will compute the correct source epoch internally so we don't have to pass the source epoch to it.
- The `epoch` we pass to `get_light_client_committee` should have been the epoch that the light client is going to vote for. In this case, set it to `get_current_epoch(state) + LIGHT_CLIENT_COMMITTEE_PERIOD` can make it forward to the next period.

djrtwo: But getting the `current_source_epoch` and adding `LIGHT_CLIENT_COMMITTEE_PERIOD` to it will put you certainly in the next period. Just as `get_current_epoch + LIGHT_CLIENT_COMMITTEE_PERIOD` will also put you in the next period

hwwhww:
> But getting the current_source_epoch and adding LIGHT_CLIENT_COMMITTEE_PERIOD to it will put you certainly in the next period. Just as get_current_epoch + LIGHT_CLIENT_COMMITTEE_PERIOD will also put you in the next period

No, because `compute_committee_source_epoch(epoch, period)` has a lookahead inside and `get_light_client_committee(beacon_state, epoch)` already calls `compute_committee_source_epoch(epoch, LIGHT_CLIENT_COMMITTEE_PERIOD)` internally. So `current_source_epoch + LIGHT_CLIENT_COMMITTEE_PERIOD` is the `next_source_epoch` point at the diagram below, then `get_light_client_committee(next_source_epoch)` would output the committee of **period 2**. I believe we want to get the committee of **period 3**, right?

image

PR #1703, opened 2020-04-02
Rework shard block and fraud proof (shard state transition) spec

hwwhww:
### Issue
1. Make the fraud-proofs.md spec in Python functions.
2. Add a basic test for shard transition.

### How did I fix it
1. `beacon-chain.md`
    1. Rework shard block
        1. Replace block wrapper with signable pattern
        2. Add `proposer_index` to shard block since the proposal index helps clients to validate the block in network level.
    2. Refactor `get_light_client_committee` a bit.
    3. Fix the wrong `get_shard_proposer_index` function calls
    4. Update `apply_shard_transition`
    5.  Fix typo `attestation.slot == state.slot` -> `attestation.data.slot == state.slot` in `is_winning_attestation`
    6. Check `verify_shard_transition_false_positives` **after** `process_operations`
    7. Fix `shard_attestations` filter in `process_crosslinks`: since attestations come from block, should use `attestation.data.slot + MIN_ATTESTATION_INCLUSION_DELAY == state.slot`
    8.  **[TBD]** Allow empty `light_client_signature` to make the tests pass
3. `phase1-fork.md`:
	* Add stub `PHASE_1_GENESIS_SLOT`, set it to `SLOTS_PER_EPOCH` now.
4. Rewrite the current fraud-proofs (shard state transition) spec in Python code
5. Add basic sanity test for shard transition and beacon block processing

PR #1705, opened 2020-04-05
256-bit custody atoms for better alignment with rest of the spec and greater efficiency

dankrad:
We previously chose 48 byte custody atoms because of the sizes of field elements in BLS signatures. However, this misalignment is not pretty when compared to the rest of the spec, and will also cause significant overhead as 48 byte integers are more expensive to handle than 32 bytes.

In comparison, the switch to 32 bytes will only require one secret re-sharing (from one prime to another prime) every nine days, so it will add only very minor overhead for MPC validators.


specs/phase1/beacon-chain.md
```diff
@@ -188,7 +289,9 @@ class Validator(Container):
     # (of the particular validator) in which the validator is activated
     # = get_custody_period_for_validator(...)
     next_custody_secret_to_reveal: uint64
-    max_reveal_lateness: Epoch
+    # TODO: The max_reveal_lateness doesn't really make sense anymore.
```
dankrad, line 292: We will need a new way to incentivise people to reveal their custody secrets timely. Without the bit challenges, the max_reveal_lateness doesn't really make sense anymore.

dankrad: @vbuterin @djrtwo

specs/phase1/beacon-chain.md
```diff
@@ -829,6 +935,60 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
...
+            withdrawal_credentials=deposit.data.withdrawal_credentials,
+            amount=deposit.data.amount,
+        )
+        domain = compute_domain(DOMAIN_DEPOSIT)  # Fork-agnostic domain since deposits are valid across forks
+        signing_root = compute_signing_root(deposit_message, domain)
+        if not bls.Verify(pubkey, signing_root, deposit.data.signature):
+            return
+
+        # Add validator and balance entries
+        # TODO: This function is duplicated from phase 1 just because the validator definition
```
dankrad, line 970: Seems quite annoying to duplicate this code just because we need two extra validator init params (nothing else changed). Thoughts @djrtwo ?

specs/phase1/custody-game.md
```diff
@@ -360,13 +413,16 @@ def process_custody_slashing(state: BeaconState, signed_custody_slashing: Signed
     assert is_valid_indexed_attestation(state, get_indexed_attestation(state, attestation))
 
     # TODO: custody_slashing.data is not chunked like shard blocks yet, result is lots of padding.
+    # ??? What does this mean?
```
dankrad, line 416: @protolambda I didn't really understand that comment, what was that about?

tests/core/pyspec/eth2spec/test/helpers/phase1/attestations.py
```diff
@@ -1,11 +1,13 @@
+# TODO: What is this file for??? It seems to be broken!
```
dankrad, line 1: @djrtwo @protolambda Couldn't really see the reason for this file

dankrad:
This should be ready for review, though I will probably go through the old tests again to make sure I haven't missed anything.

specs/phase1/custody-game.md
protolambda: It's outdated I think, but has to do with: https://github.com/ethereum/eth2.0-specs/commit/2a91b43eaf6d883750216d82e921227b03be3223
There was a chunking approach to split a block in 4 parts, and the slashing code did not match.

tests/core/pyspec/eth2spec/test/helpers/phase1/attestations.py
protolambda: Phase 1 required some specific things of attestations that had no place in phase0, so I think this was added to fill that gap. The phase0 attestation helper file still has a few special cases for phase1 attestations though, it would be nice to separate that, while avoiding code duplication.

hwwhww: fyi I removed `helpers/phase1/attestations.py` and added phase 1 specific logic in `test/helpers/attestations.py ` in #1703 too.

specs/phase1/beacon-chain.md
protolambda: Adding a `default_validator(state, index) -> Validator` function, that changes each phase, would work.

hwwhww:
@dankrad FYI I fixed the conflicts from #1703

specs/phase1/beacon-chain.md
dankrad: Created issue #1888

specs/phase1/custody-game.md
```diff
@@ -227,11 +308,91 @@ def process_custody_game_operations(state: BeaconState, body: BeaconBlockBody) -
...
+```python
+def process_chunk_challenge_response(state: BeaconState,
+                                     response: CustodyChunkResponse) -> None:
+
+    # Get matching challenge (if any) from records
+    matching_challenges = [
+        record for record in state.custody_chunk_challenge_records
+        if record.challenge_index == response.challenge_index
+    ]
+    assert len(matching_challenges) > 0
```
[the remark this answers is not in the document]
protolambda: Note that only one record is cleared later in the function, so if the intention is to only have unique records, then this looks better.

specs/phase1/custody-game.md
```diff
@@ -227,11 +308,91 @@ def process_custody_game_operations(state: BeaconState, body: BeaconBlockBody) -
...
+    assert is_valid_merkle_branch(
+        leaf=hash_tree_root(response.chunk),
+        branch=response.branch,
+        depth=CUSTODY_RESPONSE_DEPTH,
+        index=response.chunk_index,
+        root=challenge.data_root,
+    )
+    # Clear the challenge
+    index_in_records = state.custody_chunk_challenge_records.index(challenge)
+    state.custody_chunk_challenge_records[index_in_records] = CustodyChunkChallengeRecord()
```
protolambda, line 390: Suggestion, but not sure if it breaks when multiple operations are involved: move the last element to the index that is cleared, and reduce the length by one. That way you have minimal modifications, while keeping the list small, and not searching through it with the awkward replace or append function.

dankrad: That's a good idea, though it's separate. I'll merge this now and create an issue for it instead. Would like @vbuterin 's comment on it first as he was the one who originally implemented it like this.

specs/phase1/beacon-chain.md
```diff
@@ -327,6 +340,8 @@ class BeaconState(Container):
     # at RANDAO reveal period % EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS
     exposed_derived_secrets: Vector[List[ValidatorIndex, MAX_EARLY_DERIVED_SECRET_REVEALS * SLOTS_PER_EPOCH],
                                     EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS]
+    custody_chunk_challenge_records: List[CustodyChunkChallengeRecord, MAX_CUSTODY_CHUNK_CHALLENGE_RECORDS]
```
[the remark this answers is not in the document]
dankrad: The problem is that we have to use this one, as we aren't doing "object diffs" anymore and so the full definition of the beacon chain state has to be in the beacon chain spec. Thus, this constant is part of the beacon spec.

specs/phase1/custody-game.md
```diff
@@ -409,28 +572,44 @@ def process_custody_slashing(state: BeaconState, signed_custody_slashing: Signed
...
+
+    # Reset withdrawable epochs if challenge records are empty
+    records = state.custody_chunk_challenge_records
+    validator_indices_in_records = set([record.responder_index for record in records])
+    for index, validator in enumerate(state.validators):
+        if validator.exit_epoch != FAR_FUTURE_EPOCH:
+            all_secrets_are_revealed = validator.all_custody_secrets_revealed_epoch == FAR_FUTURE_EPOCH
+            if index in validator_indices_in_records or all_secrets_are_revealed:
+                # Delay withdrawable epochs if challenge records are not empty or not all
+                # custody secrets revealed
```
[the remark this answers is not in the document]
dankrad: It records in which epoch all custody secrets were revealed for the first time after a validator has exited. 

I think that variable was named incorrectly in a refactor, I added a "not_" now.

specs/phase1/beacon-chain.md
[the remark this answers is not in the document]
hwwhww: Yeah but so are `MAX_CUSTODY_CHUNK_CHALLENGES`, `MAX_CUSTODY_CHUNK_CHALLENGE_RESPONSES`, `MAX_CUSTODY_KEY_REVEALS`, `MAX_EARLY_DERIVED_SECRET_REVEALS`, and `MAX_CUSTODY_SLASHINGS`.  The problem is that `beacon-chain.md` and `custody-game.md` are depending on each other now, although these constants definitions are reordered in `setup.py`.

It's not perfect but at least it would be better if we can put these similar constants to the same table now. Either in `beacon-chain.md` or `custody-game.md` is fine by me :)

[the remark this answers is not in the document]
dankrad: OK, didn't know that. I will add it do the custody game spec then.

PR #1706, opened 2020-04-06
Redefine attestation propogation condition

paulhauner:
I've redefined an attestation propagation condition in the networking spec. I assume this was the intention of the spec, keen to hear if I was wrong.

We could expand this to say that attestations from _any_ source (block, local, etc) will stop propagation.. I can't imagine it being much more difficult for clients and perhaps it reduces a little traffic on the subnet?

I've marked this as draft until we confirm the intentions of this statement.


djrtwo:
Yes, that is the original intended statement.

As we discussed briefly in another chat, we can expand this to be "per epoch" instead of "per slot" to safely reduce the size of the map from `V*32` bits to `V` bits.

paulhauner:
> As we discussed briefly in another chat, we can expand this to be "per epoch" instead of "per slot" to safely reduce the size of the map from V*32 bits to V bits.

Sounds good, I have made this change.

A minor note is that due to the gossip clock disparity allowance (and whilst that allowance is less than half a slot) , the sizes are `V*33` or `V*2`.

paulhauner:
Oh actually it was always `V * 2` for an epoch map, since you need a map for the previous epoch too.

If clock disparity ever gets to be > 1/2 a slot duration then we'd need `V * 3`

PR #1712, opened 2020-04-08
Fix `INACTIVITY_PENALTY_QUOTIENT`

hwwhww:
#1693 alternative

## Issue
The amount of inactivity penalty was adjusted to half since we were applying penalty for missing FFG target and source. But now we only apply it for missing target, so `INACTIVITY_PENALTY_QUOTIENT` should have been `2**24`.

## How did I solve it
Update `INACTIVITY_PENALTY_QUOTIENT` from `2**25` to `2**24`.

Note that the status-quo `2**25` doesn't break security:
https://github.com/ethereum/eth2.0-specs/pull/1693#issuecomment-610625885
> @vbuterin: Though it doesn't break security at all if it's kept at 2**25; it would just mean the leak takes 41% longer, so if this ends up being the only change we need to do then not bothering with it is fine too.

IMO this PR makes the current eth2.0-specs explainers (e.g., Serenity Design Rationale) stay consistent leaking rate. Also, we can avoid making `INACTIVITY_PENALTY_QUOTIENT ` configuration explainer complicated in #1693.

It's not required for testnet, but it would be nice to include this configuration adjustment in the final v1.0.0 ™️  spec.


djrtwo:
Maybe we should maintain a `v0.12` branch off of `dev` that will incorporate the BLS IETF changes and anything minor like this. I'd like to keep `dev` (or another branch) ready to release `v0.11.x` releases that have updated network specs other backward compatible changes

hwwhww:
@djrtwo sounds good! I added `v012x` branch for the next minor version upgrade.

PR #1714, opened 2020-04-09
Strict `ValidatorIndex` definition

hwwhww:
### Issue
We defined that `ValidatorIndex` is the index number for `state.validators`.
By that definition, the input parameter `index` in `compute_shuffled_index` is the **position** of the `index_count`-size `Sequence[ValidatorIndex]`, not `ValidatorIndex`.

I found it by looking at the line `assert index < index_count` since IMO it's wrong *if* the given `index` is really a `ValidatorIndex`.

### How did I fix it
This PR only removes the wrong type casting, no substantive change for PySpec, but I'm not sure about how other client implementations deal with casting.

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -773,7 +773,7 @@ def compute_committee(indices: Sequence[ValidatorIndex],
     """
     start = (len(indices) * index) // count
     end = (len(indices) * (index + 1)) // count
-    return [indices[compute_shuffled_index(ValidatorIndex(i), len(indices), seed)] for i in range(start, end)]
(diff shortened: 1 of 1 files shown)


specs/phase0/beacon-chain.md
```diff
@@ -722,24 +722,24 @@ def is_valid_merkle_branch(leaf: Bytes32, branch: Sequence[Bytes32], depth: uint
 #### `compute_shuffled_index`
 
 ```python
-def compute_shuffled_index(index: ValidatorIndex, index_count: uint64, seed: Bytes32) -> ValidatorIndex:
+def compute_shuffled_index(index: uint64, index_count: uint64, seed: Bytes32) -> uint64:
```
mratsim, line 725: Maybe the input param should be renamed offset as further differentiation?

PR #1730, opened 2020-04-16
add note about distributing bootnode ENRs prior to genesis

djrtwo:
As discussed with @AgeManning and @nisdas, bootnode ENRs need to be distributed prior to learning `genesis_validators_root`. A note of guidance for how to do this in a standard way has been added.

Note: circleci cache got messed up so bumped version

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,16 +79,16 @@ jobs:
       # Restore git repo at point close to target branch/revision, to speed up checkout
       - restore_cache:
           keys:
-            - v2-specs-repo-{{ .Branch }}-{{ .Revision }}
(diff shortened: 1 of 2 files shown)


djrtwo:
@nisdas and @AgeManning: sanity check that this looks fine (and at least very similar to what you are doing already)

nisdas:
This looks good to me :+1:

AgeManning:
Yeah this looks good to me also. 
For us, we would likely update the bootnode's ENR after we know the value, as then the bootnode will also be discovered and participate in discovery with other nodes on the network.

djrtwo:
made a quick note on setting bootnode ENRs to normal values after genesis info is know. thanks!

PR #1732, opened 2020-04-19
Allow aggregate where aggregator has not attested

paulhauner:
Presently an `aggregate_and_proof` must include three signatures from the aggregator:

1. `aggregate_and_proof.signature`
1. `aggregate_and_proof.message.selection_proof`
1. `aggregate_and_proof.message.aggregate.signature`

I propose that we remove the requirement for (3).

Consider the following scenario where a validator client (VC) is connected to a failing beacon node (BN) `A` but knows a backup BN `B`:

- Start of slot
  - VC is connected to BN `A`.
- 1/3 through slot
  - VC creates an unaggregated attestation with BN `A`
- 1/2 through slot
  - VC decides that BN `A` is trash and switches to BN `B`.
- 2/3 through slot
  - VC produces and publishes an aggregated attestation using BN `B`.

If BN `B` hasn't seen the first unaggregated attestation published by BN `A` (it is trash, after all) then requirement (3) will cause the aggregated attestation from BN `B` to be invalid. Whilst this is a shame for the VC, the network is deprived of a potentially useful aggregate.

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -293,7 +293,7 @@ There are two primary global topics used to propagate beacon blocks and aggregat
     - The `aggregate` is the first valid aggregate received for the aggregator with index `aggregate_and_proof.aggregator_index` for the epoch `aggregate.data.target.epoch`.
     - The block being voted for (`aggregate.data.beacon_block_root`) passes validation.
     - `aggregate_and_proof.selection_proof` selects the validator as an aggregator for the slot -- i.e. `is_aggregator(state, aggregate.data.slot, aggregate.data.index, aggregate_and_proof.selection_proof)` returns `True`.
-    - The aggregator's validator index is within the aggregate's committee -- i.e. `aggregate_and_proof.aggregator_index in get_attesting_indices(state, aggregate.data, aggregate.aggregation_bits)`.
+    - The aggregator's validator index is within the aggregate's committee -- i.e. `aggregate_and_proof.aggregator_index in get_beacon_committee(state, aggregate.data.slot, aggregate.data.index)`.
     - The `aggregate_and_proof.selection_proof` is a valid signature of the `aggregate.data.slot` by the validator with index `aggregate_and_proof.aggregator_index`.
     - The aggregator signature, `signed_aggregate_and_proof.signature`, is valid.
     - The signature of `aggregate` is valid.


paulhauner:
It's worth noting that this change implicitly allows aggregates with `aggregate.aggregation_bits.num_set_bits() == 0` to propagate.

This isn't a DoS vector whilst we only propagate one aggregate per aggregator per epoch.

djrtwo:
> I propose that we remove the requirement for (3).

Specifically, you mean to remove the requirement of the aggregator being a part of the signature. Still need to verify the validity of that signature.

I believe this condition is an artifact prior to when we had the outer most signature.

This does seem fine, but I would challenge the use case a bit. In such a case, the aggregator is relatively selfish and should figure out a way to get their attestation into the aggregate. Quickly Publishing to B at 1/2 then asking for the aggregated attestation would be a more optimal move regardless of the gossip validity conditions. The problem here is also due to a choice in architecture and how to solve the problem rather than the intention/honest behavior.

I'm also a bit resistant because this allows for behavior not specifically described in the validator guide.

This is semi-breaking wrt `v0.11.x` so if we put it in, I'd say save for `v0.12` (mainly the coming BLS bump).

Anyone else have a quick though on this PR?

terencechain:
The change looks good to me, and I have been thinking about this for a while. I agree to what Danny said, it's a choice in BN <-> VC architecture. To mitigate the problem in Paul's example, a more complex load balancer could be used for BN <-> VC. But I do prefer the simpler approach which is to allow aggregate where aggregator has not attested.

paulhauner:
> Specifically, you mean to remove the requirement of the aggregator being a part of the signature. Still need to verify the validity of that signature.

Correct.

> In such a case, the aggregator is relatively selfish and should figure out a way to get their attestation into the aggregate.

I think this is easier said than done. The actual logic to handle this edge case is non-trivial considering slashing protection, etc. Notably, it's all still possible with the suggested changes, just  no longer mandatory.

> I'm also a bit resistant because this allows for behavior not specifically described in the validator guide.

There's plenty of things that are permitted in the spec but not described in the honest validator guide. Conversely, you could argue that this is the only place in the spec where messages that benefit the network without proposing a DoS risk are rejected because they aren't profit-seeking.

As I see it, this change benefits the network and reduces mandatory complexity in clients (albeit small) without inhibiting any existing behavior.

> I'd say save for `v0.12`

Makes sense.

djrtwo:
close in favor of #1801

PR #1737, opened 2020-04-21
Unified equivocation slashing

JustinDrake:
See https://github.com/ethereum/eth2.0-specs/issues/1387. Special thanks to @hwwhww for pinging me :)

**TLDR**: Unify all forms of slashing from equivocation (in particular equivocation through beacon proposing in phase 0, or through shard proposing and shard attesting in phase 1) into a single equivocation slashing condition.

**rationale**: This is a significant simplification to the slashing logic for phase 1. Each slashing condition costs about 50 lines of consensus code and needs to be separately tested. All in all we save about 100 lines of consensus code for phase 1. Avoiding two slashing conditions in phase 1 correspondingly simplifies testing.

**substantive changes**:

* Introduce `equivocation_root` in signing root
* Replace `process_proposer_slashing` by `process_equivocation`

**cosmetic changes**:

* Move `BeaconBlockHeader` aside `BeaconBlockBody` and `BeaconBlock`
* Rename `SigningRoot` to `SigningData`
* Remove a couple of unnecessary new lines

cc @vbuterin, @djrtwo

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -7,7 +7,6 @@
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
-
 - [Introduction](#introduction)
 - [Notation](#notation)
 - [Custom types](#custom-types)
@@ -34,23 +33,22 @@
     - [`HistoricalBatch`](#historicalbatch)
     - [`DepositMessage`](#depositmessage)
     - [`DepositData`](#depositdata)
-    - [`BeaconBlockHeader`](#beaconblockheader)
-    - [`SigningRoot`](#signingroot)
+    - [`SigningData`](#signingdata)
   - [Beacon operations](#beacon-operations)
-    - [`ProposerSlashing`](#proposerslashing)
+    - [`Equivocation`](#equivocation)
     - [`AttesterSlashing`](#attesterslashing)
     - [`Attestation`](#attestation)
     - [`Deposit`](#deposit)
     - [`VoluntaryExit`](#voluntaryexit)
   - [Beacon blocks](#beacon-blocks)
+    - [`BeaconBlockHeader`](#beaconblockheader)
     - [`BeaconBlockBody`](#beaconblockbody)
     - [`BeaconBlock`](#beaconblock)
   - [Beacon state](#beacon-state)
     - [`BeaconState`](#beaconstate)
-  - [Signed envelopes](#signed-envelopes)
+  - [Signatures](#signed-envelopes)
     - [`SignedVoluntaryExit`](#signedvoluntaryexit)
     - [`SignedBeaconBlock`](#signedbeaconblock)
-    - [`SignedBeaconBlockHeader`](#signedbeaconblockheader)
 - [Helper functions](#helper-functions)
   - [Math](#math)
     - [`integer_squareroot`](#integer_squareroot)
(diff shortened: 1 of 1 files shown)


rauljordan:
Hi @JustinDrake is there a strong rationale why this absolutely needs to be included in phase 0? These changes, albeit small in text, entail a significant burden for client developers (i.e. we have to create and maintain another feature branch that breaks consensus, has thousands of lines of code / tests / generated protos). These changes end up distracting us for 1 or 2 more weeks from key improvements on what we thought was the last, consensus-breaking spec change before mainnet barring critical bugs. We are already quite busy in terms of improving our client's testnet resilience and user experience today. If there is no urgent need for this in phase 0, we kindly ask on behalf of @prylabs that we move this to phase 1.

prestonvanloon:
I reviewed this with other members of @prysmaticlabs and I share the opinion expressed above. However, I would like to add a bit more to our feedback. 

Having a generic function for equivocation slashing is a great idea. Had this been 6 months ago then I don't think anyone would be debating this. The issue now is that we are already in a code freeze. When the spec was unfrozen the first time, we agreed these significant changes were worth it. We've spent months building with an evolving spec and now we're finally just weeks away from _The Multiclient Testnet_ followed by _The Mainnet Release_.

Now, in the face of unfreezing another frozen spec, do you all think this is worth it? 
How many different types of evocation slashing do we plan to exist?

JustinDrake:
> Had this been 6 months ago then I don't think anyone would be debating this.

Agreed—I kinda dropped the ball on this one 😢

> How many different types of evocation slashing do we plan to exist?

As of now we know of 3 types:

1) beacon proposer equivocation (phase 0)
2) shard proposer equivocation (phase 1)
3) shard attester equivocation (phase 1)

It's possible more types of equivocation will be required for phase 2.

prestonvanloon:
> The testing infrastructure we have in clients and in the spec today should allow us to make this change relatively pain-free - we're fine either way on the Nimbus side - worth considering that it's easier to introduce this now than in the phase1 HF for example.

We agree that this is relatively easier than the previous unfreezing of the spec, however I estimate that this change will delay a mainnet release by 2 to 4 weeks, at least. As all things in software engineering, whether or not we move forward on this is a trade off that must be carefully calculated.

specs/phase0/beacon-chain.md
```diff
@@ -1550,36 +1546,33 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
 
-##### Proposer slashings
-
-```python
-def process_proposer_slashing(state: BeaconState, proposer_slashing: ProposerSlashing) -> None:
-    header_1 = proposer_slashing.signed_header_1.message
-    header_2 = proposer_slashing.signed_header_2.message
-
-    # Verify header slots match
-    assert header_1.slot == header_2.slot
```
adiasg, line 1568: Where does this check happen in `process_equivocation`?

specs/phase0/beacon-chain.md
```diff
@@ -388,33 +384,28 @@ class DepositData(Container):
...
-    state_root: Root
-    body_root: Root
-```
-
-#### `SigningRoot`
+#### `SigningData`
 
 ```python
-class SigningRoot(Container):
+class SigningData(Container):
```
adiasg, line 390: PR description mentions:
> Introduce `uniqueness_root` in signing root

JustinDrake: The `uniqueness_root` gets incorporated in the signing root (the root of the signing data).

specs/phase0/beacon-chain.md
JustinDrake: This check is done via the `equivocation_root `. Notice that the `slot` is part of the `equivocation_root`, and that both `signing_root_1` and `signing_root_2` use the same `equivocation_root`.

JustinDrake:
Closing in favour of https://github.com/ethereum/eth2.0-specs/pull/1758 which is easier to review

Issue #1701, opened 2020-04-01
Clarity on `uint` overflow


paulhauner:
We (Lighthouse) are interested in building a "panic-free" implementation of the state transitions. I think we're fairly close when compiled in Rust's "release" mode (optimized, production-ready), however for Rust's "debug" mode (slow but useful for troubleshooting) any integer overflow that is not explicitly handled causes a panic.

When it comes to these overflows we have two choices:

- Indicate that wrapping on overflow is desired behavior (viz., don't raise an exception).
- Return early with an error if an overflow is encountered (viz., raise an exception).

The specification uses `uint...` types that, as I understand it, have undefined behavior when it comes to overflow (perhaps this can be derived from the tests, but this doesn't feel authoritative enough to me). From "common sense", I would assume that `uint64_max_value() + 1 == 0` and `0 - 1 == uint64_max_value()`. I also note that a few comments that infer `uint64` overflows:

> Math safe up to ~10B ETH, afterwhich this overflows uint64.

> Factored out from balance totals to avoid uint64 overflow

> Factored out from penalty numerator to avoid uint64 overflow

So, I'm _inferring_ that uint64 overflows, _assuming_ that it's desired behavior and _assuming_ the specifics of that behavior.

It would be nice to have some clarity around this behavior. I understand that there's an idea that "valid state transitions shouldn't lead to overflows", but when it comes to things like fuzzing it's nice to have a distinction between "expected wrapping overflow" and "unexpected wrapping overflow".

So in summary, I have a question and a request please:
- Question: is a state transition considered valid if integer arithmetic overflows?
- Request: Make the spec explicit about the behavior of overflows.

As always, thanks! :)

JustinDrake:
> So, I'm inferring that uint64 overflows, assuming that it's desired behavior and assuming the specifics of that behavior.

I'd say the desired outcome is that these overflows are (provably, e.g. using formal verification) not possible. If for some reason we (the spec designers) messed up then overflows should probably trigger an exception.

> Question: is a state transition considered valid if integer arithmetic overflows?

I'd say "no", this is an invalid state transition.

> Request: Make the spec explicit about the behavior of overflows.

[...]

PR #1739, opened 2020-04-21
Partial fix for #1701

JustinDrake:
Clarify that state transitions with `uint64` overflows are invalid.


djrtwo:
underflow too?

JustinDrake:
> underflow too?

Done! :)

PR #1740, opened 2020-04-21
Cosmetic changes from #1737

JustinDrake:
Pulling out some of the cosmetic changes from #1737:

* Rename the `SigningRoot` container to `SigningData`. The `signing_root` (of type `Root`) is the `hash_tree_root` of the `SigningData` (of type `Container`).
* Cleanup `compute_signing_root`
* Remove unnecessary new lines.

(The rationale for this split is because the substantive changes from #1737 may not make it in phase 0. Also this will help focus the discussion in #1737 on the substantive changes.)

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -35,7 +35,7 @@
     - [`DepositMessage`](#depositmessage)
     - [`DepositData`](#depositdata)
     - [`BeaconBlockHeader`](#beaconblockheader)
-    - [`SigningRoot`](#signingroot)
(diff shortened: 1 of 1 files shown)


djrtwo:
I really think we should avoid nitpicking the spec at this point. Any non-security critical changes we introduce add unnecessary overhead for client teams and the potential to introduce minor bugs.

I agree that this looks good and the PR is a very small and digestible. Okay to merge.
But, I reiterate, now is generally not the time for a wave of renamings and code restructurings even if non-substantive

PR #1742, opened 2020-04-22
Added FFG-LMD vote consistency checks

### LINKED ISSUE #1408: Attestation validation differences between fork choice and process_attestation
An attestation contains three votes: LMD GHOST vote, FFG vote and Crosslink vote.
The beacon chain specs specify how to validate an attestation and its votes.
However, the validation is split in several parts: some checks defined in [fork_choice/on_attestation](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_fork-choice.md#on_attestation) and others - in [process_attestation](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#attestations). They both invoke [is_valid_indexed_attestation](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#is_valid_indexed_attestation).

The `on_attestation` deals mostly with LMD GHOST vote and part of FFG vote (`target` field). So, if there arrives an attestation with an invalid crosslink vote or an invalid `source` fields, then it is still accepted by the fork choice spec and affects the LMD GHOST fork choice.
However, the attestation should not be included in newly proposed blocks, since it won't pass `process_attestation` checks (see [validator spec](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/validator/0_beacon-chain-validator.md#attestations).

While there may be reasons to split validation in two parts, this can lead to inconsistencies in fork choice. Consider a case, when there are two head candidates, both having the same amount of votes and tie break chooses the first one. If a partially correct attestation added, which votes for the second block, then the second block becomes the head, based on off-chain plus on-chain attestation. But since this last vote cannot be incorporated into the chain, from on-chain attestations perspective, the first candidate is the head. I do not fully understand consequences of the discrepancy, but it seems awkward at least.

Another example. If such a partially valid attestations is propagated only to part of the network nodes, it cannot reach other (honest) validators via proposed blocks, because of `process_attestation` assertion failures. It can result in different views on the state of the blockchain, which can be only corrected by propagating such invalid attestations to other nodes.
So this looks like a potential unnecessary source of forks, which can be exploited by an attacker.

### LINKED ISSUE #1456: Problems with attestations from incompatible forks
Attester and proposer roles assignment are determined by a shuffling of a set of active validators, which depends on a seed. The seed and the set of active validators, in their turn, are detemined by a chain of blocks (randao, slashings, deposits, exits, etc). An attestation data doesn't contain indices of a block proposer and attesters in an explicit form, they are (re)calculated based on a state and aggregation_bits.

That means it's important that a state that is used to restore attester and proposer indices contain the right seed and the right set of active validators. If it doesn't hold, then it can lead to problems (described later).

When an attestation is created or consumed/processed, the reference states (used to calculate proposer/attesters for the attestation's slot) will be different. However, that doesn't necessarily mean the seed and the set of active validators will differ too. As proposer and attester assignment for an epoch are determined at the beginning of the previous epoch, some beacon states can lead to the same assignments. Let's call such states compatible.

For example, when an attester creates an attestation, the reference state is the head state (with slot transitions applied up to the current slot). When an attestation is used in a fork choice, the reference state is the state at the beginning of the target epoch. When an attestation is to be included in a block or is processed as a part of a block, the reference state is the state of the block's parent with slot transitions applied. So, in general, the states can be incompatible between each other, i.e. having different seeds and/or sets of active validators. And thus leading to incorrect reconstruction of attester and proposer indices.

Such incompatibility can arise when there has been a fork and an attestation referencing one branch of the fork is inculded in a block from another branch of the fork. If the fork happend in an epoch before the previous epoch, then it's enough time that differences in block chains start to affect shufflings.

Potential problems:
* during [process_attestation(state, attestation)](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#attestations), if the state is incompatible with the attestation's target state, the aggregation_bits will translate to a wrong set of validator indices. It will most likely mean the attestation signature check will fail, since another set of public keys will be used to construct an aggregate key. Such attestation cannot be included in a block (or a block with such an attestation will be rejected by correct nodes). It's not a problem. However, a node can decide that the attestation signature is wrong (and, for example, punish the attestation's sender in some way). But the real problem is that the attestation validity is verified against the wrong state.
* another possibility is that the signature validation can succeed. It could be a quite rare situation, but two different shufflings can still have some intersecting subsets. E.g. a validator with index 1 can happen to be at the same position at both shufflings (while other positions being different). So, if aggregation_bits contain only this position, then the public keys will be the same with both states. That is a real problem, since the state incompatibility can sneak undetected.
* an additional problem, when the incompatibility is not prevented by the signature check, is that the beacon proposer index will be wrong most likely, since it's calculated based on the incompatible state (i.e. different shuffling).

So, in order to prevent such problems we suggest that only attestations from compatible states can be included in blocks. And the appropriate check should be implemented as a part of [process_attestation](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#attestations) method (and/or [get_indexed_attestation method](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#get_indexed_attestation)). Most likely such attestations cannot be included anyway, but when they can it leads to a covert problem (wrong proposer index in a pending attestation).

### LINKED ISSUE #1636: Fork choice: checking FFG source/target
In [fork choice attestation processing](https://github.com/ethereum/eth2.0-specs/blob/v0.10.1/specs/phase0/fork-choice.md#on_attestation), the `attestation.data.target` and `attestation.data.source` fields are not checked to ensure they are ancestors/descendants of each other and `attestation.data.beacon_block_root`.

What does this mean?

### 1. Faulty actors influence fork choice

An attestation where the source/target/beacon_block_root are on different chains is produced by a faulty validator.

I would expect that good practice would be to drop any message that is provably invalid and ban the peer. 

### 2. You can pick-and-choose your shuffling/validator registry

Say we have 3 chains:

```
   2    4    8     <- slot 36 (epoch 1)
   |    |    |
   .    .    .
   .    .    .
   .    .    .
   |    |    |
   1    3    5     <- slot 32 (epoch 1)
    \   |   /
        |
        .
        .
        .
```

Consider that I want to set `attestation.data.beacon_block_root` to block 4. I am able to choose whether to set `attestation.data.target` to blocks 1, 3 or 5.

The honest choice is 3, but perhaps in that state I am scheduled to vote at slot 63 and instead I want to vote earlier. Now I can pick between the other two states to find the earliest assignment and have my vote included in fork choice earlier than naturally intended.

adiasg:
Added consistency check for FFG & LMD vote in validate_on_atttestation(), fixes #1636, fixes #1456, fixes #1408


djrtwo:
@adiasg can you rebase this to `v012x` branch to queue it up for the v0.12 release?

djrtwo:
@adiasg Looks good! added a test if you want to check it out before merge

ASISBusiness (commented):
Working

PR #1744, opened 2020-04-22
add missing word "valid" in attestation gossip conditions

djrtwo:
*nitpick* -- make sure only preventing duplicate senders if you've seen a _valid_ attestation from that validator


JustinDrake:
+1 for nitpicks 😂

djrtwo:
got me 🤣

PR #1746, opened 2020-04-23
Apply strict uint64 casting

hwwhww:
Child of #1707, #1739, and #1701.

### Issue
We've already stated that all calculations should be in `uint64` domain (#1739). However, it doesn't apply for Python native `int`.

This draft applies some strict `uint64` castings to known function parameter typings by fixing #1707 `int` v.s. `uint64` incompatible errors that `mypy` found.

#### Pros:
* We can have more safety checks from SSZ classes (`remerkleable`) layer!

#### Cons:
 * Annoying casting everywhere.
* Clients may already handle the uint64 in different ways with different languages' features. This PR may cause some implementation burden.
* Note that this PR probably didn't find all calculations that need casting.

Set it to draft since I'm wondering if it would introduce a serious implementation burden for the client teams.


djrtwo:
I doubt this would introduce any requisite changes on clients.

> Note that this PR probably didn't find all calculations that need casting.

What do you mean? It did add a bunch of casts. Are you saying that the linter didn't force any of these?

> Annoying casting everywhere.

Slightly resistant for this reason. Does the `uint64` provide any additional functionality other than static analysis? Does the type actually overflow or does can it still grow past `2**64`?

hwwhww:
@djrtwo 

> Does the uint64 provide any additional functionality other than static analysis? Does the type actually overflow or does can it still grow past 2**64?

Yep! Thanks to @protolambda, it does check the boundaries and raise `ValueError`. For example:
```
>>> uint64(2**64)
  File "/Users/hwwang/.pyenv/versions/env38a/lib/python3.8/site-packages/remerkleable/basic.py", line 43, in __new__
    raise ValueError(f"value out of bounds for {cls}")
ValueError: value out of bounds for <class 'remerkleable.basic.uint64'>
```
```
>>> uint64(-1)
  File "/Users/hwwang/.pyenv/versions/env38a/lib/python3.8/site-packages/remerkleable/basic.py", line 40, in __new__
    raise ValueError(f"unsigned type {cls} must not be negative")
ValueError: unsigned type <class 'remerkleable.basic.uint64'> must not be negative
```

> What do you mean? It did add a bunch of casts. Are you saying that the linter didn't force any of these?

I only added casting for the variables where `mypy` panicked. It's still possible that we may write the computations in some Python native variables that cause overflow/underflow to `uint64`. For example the edge case:
```python
x = 1
y = -1
return unit64(x + y)
```
The linter can't detect that `y` is not `uint64` type.

We could cover more cases once (i) [`remerkleable` also checks `__mul__` and `__floordiv__ `](https://github.com/ethereum/eth2.0-specs/issues/1701#issuecomment-609843435) (ii) all constants are converted to `uint64` as this PR.

protolambda:
@hwwhww Remerkleable already does that, I implemented that shortly after that issue. See changelog: https://github.com/protolambda/remerkleable/blob/master/CHANGELOG.rst
Just update the version referenced in the spec :+1:

protolambda:
Is this ready to go out of draft mode? And which release would we like to target with this?

hwwhww:
@protolambda @djrtwo 
Ready to review now!

It does make the spec a bit more lengthy.
* I'd say we should at least add castings to the constant definition section.
* For the rest castings, it's good to have the protection from `remerkleable` with these casting. What do you think?

specs/phase0/beacon-chain.md
```diff
@@ -729,14 +728,18 @@ def compute_shuffled_index(index: uint64, index_count: uint64, seed: Bytes32) ->
 
     # Swap or not (https://link.springer.com/content/pdf/10.1007%2F978-3-642-32009-5_1.pdf)
     # See the 'generalized domain' algorithm on page 3
-    for current_round in range(SHUFFLE_ROUND_COUNT):
-        pivot = bytes_to_int(hash(seed + int_to_bytes(current_round, length=1))[0:8]) % index_count
+    for current_round in map(uint64, range(SHUFFLE_ROUND_COUNT)):
```
protolambda, line 731: This is the type of casting I'm most uneasy with. It just feels like a Python hack.
Does the following work for mypy?
```python
current_round: uint64
for current_round in range(SHUFFLE_ROUND_COUNT):
```
Or is that too verbose for spec also?

specs/phase0/beacon-chain.md
```diff
@@ -729,14 +728,18 @@ def compute_shuffled_index(index: uint64, index_count: uint64, seed: Bytes32) ->
...
+        pivot = bytes_to_int(hash(seed + int_to_bytes(current_round, length=uint64(1)))[0:8]) % index_count
         flip = (pivot + index_count - index) % index_count
```
protolambda, line 733: We should define this as uint64, not later at the usage.

specs/phase0/beacon-chain.md
```diff
@@ -984,7 +987,7 @@ def get_beacon_proposer_index(state: BeaconState) -> ValidatorIndex:
     Return the beacon proposer index at the current slot.
     """
     epoch = get_current_epoch(state)
-    seed = hash(get_seed(state, epoch, DOMAIN_BEACON_PROPOSER) + int_to_bytes(state.slot, length=8))
+    seed = hash(get_seed(state, epoch, DOMAIN_BEACON_PROPOSER) + int_to_bytes(state.slot, length=uint64(8)))
```
protolambda, line 990: IMHO this really shouldn't be a uint64. An int can't be bigger than 32 bytes here, a uint64 for the length is weird. Which fits in every integer type. Can we just define it as `int` in the util method, since it doesn't have to be SSZ anyway?

specs/phase0/beacon-chain.md
```diff
@@ -1092,7 +1095,7 @@ def initiate_validator_exit(state: BeaconState, index: ValidatorIndex) -> None:
     exit_queue_epoch = max(exit_epochs + [compute_activation_exit_epoch(get_current_epoch(state))])
     exit_queue_churn = len([v for v in state.validators if v.exit_epoch == exit_queue_epoch])
     if exit_queue_churn >= get_validator_churn_limit(state):
-        exit_queue_epoch += Epoch(1)
+        exit_queue_epoch = Epoch(exit_queue_epoch + 1)
```
protolambda, line 1098: Also rather verbose. Maybe we should cast the above creation of exit epochs instead? Hmm

specs/phase0/beacon-chain.md
hwwhww: It doesn't work for mypy since the issue is that the `range(SHUFFLE_ROUND_COUNT)` generator returns `int`.

specs/phase0/beacon-chain.md
hwwhww: I think we can!

The `uint64` for `int_to_bytes` param is mainly for ensuring the input is non-negative, and we don't have non-negative number type in native Python. I'm okay to change it back to `int` since `uint64` might be "too much" here.

specs/phase0/beacon-chain.md
hwwhww: Awww it seems mypy is happy with `exit_queue_epoch += Epoch(1)` now! Maybe it's due to `remerkleable` new release, or I got wrong with something. I'll change it back. Thanks!

specs/phase0/beacon-chain.md
```diff
@@ -1724,13 +1727,13 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
...
         branch=deposit.proof,
-        depth=DEPOSIT_CONTRACT_TREE_DEPTH + 1,  # Add 1 for the List length mix-in
+        depth=uint64(DEPOSIT_CONTRACT_TREE_DEPTH + 1),  # Add 1 for the List length mix-in
         index=state.eth1_deposit_index,
         root=state.eth1_data.deposit_root,
     )
 
     # Deposits must be processed in order
-    state.eth1_deposit_index += 1
+    state.eth1_deposit_index = uint64(state.eth1_deposit_index + 1)
```
hwwhww, line 1736: Hmmm it's so strange. `exit_queue_epoch += Epoch(1)` is okay, but `state.eth1_deposit_index += uint64(1)` returns error:
> Incompatible types in assignment (expression has type "uint", variable has type "uint64")

Strange dark mypy hole. 🤦‍♀️

protolambda: The integer additions etc. are implemented on the uint base class in remerkleable, to not duplicate it everywhere. I will try to add typevar annotations that might help mypy understand it, then you can try this in the spec.

protolambda: Would this help? https://github.com/protolambda/remerkleable/compare/uint_type

protolambda: If that epoch thing works, then `state.eth1_deposit_index += DepositIndex(1)` should also work.

protolambda: Oh nvm, DepositIndex doesn't have its own type in the pyspec. Hmm, that error is weird.

protolambda:
Hey, we didn't include this issue since the v0.12 release was kept small where possible, but it could still be useful for better type checking. I'll try and resolve the merge conflicts, and I think we can define a `def copy(v: V) -> V` helper function to maybe clean up that added BeaconState type annotation.

protolambda:
See [`strict-uint64-merge-dev`](https://github.com/ethereum/eth2.0-specs/tree/strict-uint64-merge-dev) branch to resolve merge conflict. This PR builds on a fork of the specs repository, so I pushed it to another branch.

specs/phase0/beacon-chain.md
protolambda: I still wonder about this thing, but suppose the other typing fixes outweigh this style issue.

specs/phase0/beacon-chain.md
protolambda: The `current_round` is also only converted to a single byte (see other comment about `uint8`). Avoiding this `map` would be nice.

specs/phase0/beacon-chain.md
```diff
@@ -732,11 +732,15 @@ def compute_shuffled_index(index: uint64, index_count: uint64, seed: Bytes32) ->
...
-    for current_round in range(SHUFFLE_ROUND_COUNT):
+    for current_round in map(uint64, range(SHUFFLE_ROUND_COUNT)):
         pivot = bytes_to_int(hash(seed + int_to_bytes(current_round, length=1))[0:8]) % index_count
-        flip = (pivot + index_count - index) % index_count
+        flip = uint64((pivot + index_count - index) % index_count)
         position = max(index, flip)
-        source = hash(seed + int_to_bytes(current_round, length=1) + int_to_bytes(position // 256, length=4))
+        source = hash(
+            seed
+            + int_to_bytes(current_round, length=1)
```
protolambda, line 741: Does this count as a limitation to the shuffle-round counts, making its maximum value 255 (inclusive)? Should we change it to a uint8 type? And the `SHUFFLE_ROUND_COUNT` too? I don't think clients actually use `uint64` here at least. E.g. [lighthouse uses `uint8`](https://github.com/sigp/lighthouse/blob/4331834003a6206283c692fe0d6bfbc3c6cfc868/consensus/swap_or_not_shuffle/src/shuffle_list.rs#L86).

protolambda:
For some of the `int` -> `uintN` casts, it may be unnecessary in a safe way when https://github.com/protolambda/remerkleable/pull/4 is tested and integrated. We have to find some balance, in some places the explicit conversions still help the reader.

specs/phase0/beacon-chain.md
ericsson49: What about changing the type of the first parameter of `int_to_bytes` to `int` ?
`current_round` can be kept `int` then.

Alternatively one can introduce something like `urange` method, which will be the same `map` applied to `range`, though.

```python
def urange(*args):
   return map(uint64, range(*args))
```

As `uint64` is heavily used in pyspec, introducing unsigned `range` looks pretty reasonable.

hwwhww: @protolambda 

1. Agreed with that `current_round` can be `uint8`.
2. Currently, `uint8` is not compatible with `uint64` for `mypy` in remerkerble layer when we want to pass it to `int_to_bytes`. Although it's fine to set `current_round` to `uint8` type in pyspec right now since we haven't changed our CI linter setting to as strict as #1707.
3. For implementation-wise, I feel having `int_to_bytes` be an instance function in SSZ level would be neater. 
    - We can define `uint.uint_to_bytes` in `remerkerble`:

    ```python
    def uint_to_bytes(self, byteorder: str) -> bytes:
         return self.to_bytes(self.type_byte_length, byteorder=byteorder)
    ```
    - To decrease the requirements for pyspec readers to understand remerkerble, we can modify `int_to_bytes` to:
    ```python
    def int_to_bytes(n: uint) -> bytes:
        """
        Return the serialization of ``n`` in ``ENDIANNESS``-endian in the byte length of ``uint`` type.
        """
        return uint.uint_to_bytes(byteorder=ENDIANNESS)
    ```

What do you think?

---

@ericsson49 
Hi!
1. To me, the goal is setting a boundary of all variables in pyspec since Python `int` is unbounded. I hope it would decrease the ambiguity and make other implementers' lives better?
2. IMHO It seems better to have `urange` supports all `uint` types. Especially if we want to set `current_round` to `uint8` as @protolambda suggested. I'd say if we really want to avoid `map` for readability, the simplest way is:
```python
for current_round in range(SHUFFLE_ROUND_COUNT):
    current_round = uint8(current_round)  # cast `int` to `uint8` type
```
What do you think?

3. What do you think about the above proposal for `int_to_bytes` change?

---
Thank you both! 🙂

specs/phase0/beacon-chain.md
hwwhww: see my response above

specs/phase0/beacon-chain.md
protolambda: > We can define uint.uint_to_bytes in remerkerble:

We already have that, as well as stream variants, for every type :).

But for spec style and less hidden things, we generally avoid it.

The methods:
- `def encode_bytes(self) -> bytes:`
- `def serialize(self, stream: BinaryIO) -> int:`
- `def decode_bytes(cls: Type[V], bytez: bytes) -> V:`
- `def deserialize(cls: Type[V], stream: BinaryIO, scope: int) -> V:`

So `uint8(123).encode_bytes()` would just work to get it as bytes.

And there are a ton more util methods for size bounds, working with merkle backings, etc., but it's meant for tooling more than the spec.

hwwhww: @protolambda 

Oh great!

> But for spec style and less hidden things, we generally avoid it.

Agreed with that we should generally avoid exposing SSZ APIs. Another solution is that we can describe `int_to_bytes(n: uint)` in the same way we describe `hash_tree_root`.

protolambda:
Merged new arithmetic operators for better typing into `remerkleable` master branch. Let's try them out here, and if it works well, I'll cut a `remerkleable` release with the feature and we can do final polish + merge here.

hwwhww:
@protolambda 

There are some issues of the `remerkleable` master branch:

It disallows the Python sugar syntax like:
```python
    num_validators = spec.SLOTS_PER_EPOCH * 8
    return [spec.MAX_EFFECTIVE_BALANCE] * num_validators
```

Error message:

```
eth2spec/test/context.py:120: in default_balances
    return [spec.MAX_EFFECTIVE_BALANCE] * num_validators
../../../../remerkleable/remerkleable/basic.py:95: in __rmul__
    return self.__mul__(other)
../../../../remerkleable/remerkleable/basic.py:92: in __mul__
    return self.__class__(super().__mul__(self.__class__.coerce_view(other)))
../../../../remerkleable/remerkleable/basic.py:178: in coerce_view
    return cls(v)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'remerkleable.basic.uint64'>, value = [32000000000]

    def __new__(cls, value: int):
>       if value < 0:
E       TypeError: '<' not supported between instances of 'list' and 'int'

../../../../remerkleable/remerkleable/basic.py:72: TypeError
```

We *can* change it to:
```python
    num_validators = spec.SLOTS_PER_EPOCH * 8
    default_balances = []
    for _ in range(num_validators):
        default_balances.append(spec.MAX_EFFECTIVE_BALANCE)
    return default_balances
```

But it would be much verbose than before. Is it possible to allow ` [spec.MAX_EFFECTIVE_BALANCE] * num_validators`?

protolambda:
Ah yes, forgot about non-integer `__mul__` there. I'll fix that after the eth2 call today.

specs/phase0/beacon-chain.md
hwwhww: Created an alternative solution PR: https://github.com/ethereum/eth2.0-specs/pull/1935

protolambda:
Updated `remerkleable` to do multiplication python-style if the other input is not an `int`. And tested with lists.

protolambda:
Made `remerkleable` pass mypy checks (within remerkleable, not yet the spec). Hopefully it helps sort out some typing issues here. PR: https://github.com/protolambda/remerkleable/pull/6

protolambda:
Update: releasing a new `remerkleable` version which infers a lot of the types here, making the changes in the spec less invasive.

protolambda:
Released `remerkleable 0.1.17`, integrated into spec, removed redundant type casts, added type coercions for new places. Test speed took a hit however, the constant coercions in uints and length checks make things slower. At least it catches more things now though.

See `type-infer` branch, branched off from this PR. https://github.com/ethereum/eth2.0-specs/tree/type-infer

specs/phase0/beacon-chain.md
ericsson49: @hwwhww
Hi! Sorry, I missed a notification, so responding somewhat late.

> To me, the goal is setting a boundary of all variables in pyspec since Python int is unbounded. I hope it would decrease the ambiguity and make other implementers' lives better?

That's a great goal. It definitely helps. Basically, a typical developer would prefer using a bounded integer type whenever possible. And resolving possible range can be a problem.
In the particular case, it's however obvious that `current_round` is bounded.
But as `current_round` is intended to be `uint8` it definitely worth using an explicit type/conversion here.

However, unbounded `int` cannot be completely avoided, I guess, e.g. `legendre_bit` uses it and private keys are sometimes used. But it would be definitely great to separate cases where bounded `int`s are enough from those, where arbitrary precision `int`s are more appropriate.

> IMHO It seems better to have urange supports all uint types. Especially if we want to set current_round to uint8 as @protolambda suggested. I'd say if we really want to avoid map for readability, the simplest way is:

I think it's a matter of taste. Basically, all variants are clear, but some slightly less concise.

protolambda:
Updating some python code that uses the spec, and noticed that with these new changes, we can change previous v0.12 type hacks like `state.slot = Slot(state.slot + 1)` in the spec back to `state.slot += 1`

djrtwo:
Where are we at on this PR @hwwhww?

I'm interested in getting this into v0.12.2

eisaeidy:
Ok

hwwhww:
@djrtwo 
It is ready for me! I just merge `dev` back to it again in case anything has changed these days.
Need approval from you or @protolambda here.

I'm slightly inclined to merge this one before #1935. But it's okay either way.

Also, to double confirm with you and @protolambda, do you think #1935 is ready to go? I haven't seen any objections from client teams so far. 😄

djrtwo:
within the week.
Want to get in gossipsub params #1958
and clean up phase 1 test gen #1957

PR #1753, opened 2020-04-24
fix requirements.txt for bls generator

djrtwo:
BLS generator was broken from a bad import introduced recently.

Issue was in `tests/generators/bls/main.py`
It recently added eth2spec dependency because it imports PHASE0 constant, but this was not in the requirements.txt


protolambda:
Merging into master to get it into release just after the actual release change. Needed for proper test vectors, no substantial change.

PR #1755, opened 2020-04-26
Fork choice: Remove redundant check in validate_on_attestation

paulhauner:
Removes a redundant check from fork choice `validate_on_attestation`.

`assert target.epoch in [current_epoch, previous_epoch]` already asserts this.

I've raised this for a few reasons:

- To ensure I'm not missing something.
- If it's impossible to hit then it's impossible to test and I think we should remove it.
- The "... delay consideration ..." note becomes ineffective if we can never reach it.


paulhauner:
> You are correct, but I might argue for still having the a comment above `assert target.epoch in [current_epoch, previous_epoch]` that says `# If attestation target is from a future epoch, delay consideration until the epoch arrives`

Done :)

PR #1762, opened 2020-04-27
Remove extra `/` in README

terencechain:
Removed the extra `/` for `specs`


q9f:
that's actually how you indicate a sub-directory.

djrtwo:
I think we haven't been consistent here...
Happy either way. Whatever the standard is works well though

PR #1769, opened 2020-04-28
Move `sanity` dir to under `phase_0`

hwwhww:
Rationale
1. Define that the tests under `phase_x` are "the tests new to phase x".
2. By (1), the `sanity` tests should just be moved to under `phase_0`.

No substantive change here, the test vectors should be the same.


djrtwo:
I'm checking that sanity generators work locally before merge

djrtwo:
yup! all good

PR #1773, opened 2020-04-29
Shard fork choice rule

hwwhww:
messy drafting!
building test case with #1703 fixes & testing tool.

Questions:
- [x] ~~1. In `on_shard_block`, there is a check of "4. Check block is a descendant of the finalized block at the checkpoint finalized slot". It might not handle the shard block between the latest finalized slot and the initial shard block slot (`PHASE_1_GENESIS_SLOT`). Should we keep it?~~
- [x] 2. In `get_head`: currently using `store.block_states[head_beacon_root].shard_states[shard].latest_block_root` (the most recent accepted crosslink) as the base shard block. Some sanity confirm would be appreciated.
- [x] ~~3. Follow up the discussion in the weekly call, `get_filtered_shard_block_tree` is much cleaner than the beacon version (`get_filtered_block_tree`), I think we still need it in case the beacon chain stale?~~

TODO
- [x] add more tests


hwwhww:
@djrtwo Thanks for the review!

> I'm not sure I follow here. The first shard block would happen at `PHASE_1_GENESIS_SLOT` and reference the beacon state at that slot (and thus be in the chain of whatever was finalized prior to that slot)

Ahh this question would be solved by the answer of question 2 😅
Unrelated FYI: in `upgrade_to_phase1`, we set the initial `shard_state.slot` to `PHASE_1_GENESIS_SLOT`. So the first valid shard block would be at slot `PHASE_1_GENESIS_SLOT + 1`. Which means, after beacon state transition:
1. if `state.slot == PHASE_1_GENESIS_SLOT`, then `state.shard_states[shard].slot == state.slot`
2. else, `state.shard_states[shard].slot < state.slot`
We could remove the a slot delay at phase 1 boundary by setting default `shard_state.slot`to `PHASE_1_GENESIS_SLOT - 1` though. Does that make sense to you?

> get_filtered_shard_block_tree

ah now i see it, the while loop below it should be able to handle the shard block tree. Thanks!

specs/phase1/shard-fork-choice.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Fork choice
+
+### Helpers
+
+#### Extended `Store`
+
+```python
+@dataclass
+class Store:
```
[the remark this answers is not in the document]
hwwhww: Added independent `ShardStore`. It indeed makes the code cleaner. 👍

specs/phase1/shard-fork-choice.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+### Handlers
+
+#### `on_shard_block`
+
+```python
+def on_shard_block(store: Store, shard_store: ShardStore, signed_shard_block: SignedShardBlock) -> None:
+    shard_block = signed_shard_block.message
+    shard = shard_store.shard
```
terencechain, line 110: What do you guys think about adding `shard` to a `ShardBlock` field?   

I think it's going to get annoying computing or tracking the origin `shard` for a `ShardBlock` : /

djrtwo: yeah, I agree. being able to quickly validate baseline info about messages coming off the wire is valuable

specs/phase1/shard-fork-choice.md
```diff
@@ -0,0 +1,138 @@
...
+    # 1. Check shard parent exists
+    assert shard_block.shard_parent_root in shard_store.block_states
+    pre_shard_state = shard_store.block_states[shard_block.shard_parent_root]
+
+    # 2. Check beacon parent exists
+    assert shard_block.beacon_parent_root in store.block_states
+    beacon_state = store.block_states[shard_block.beacon_parent_root]
+
+    # 3. Check that block is later than the finalized epoch slot (optimization to reduce calls to get_ancestor)
+    finalized_slot = compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)
```
djrtwo, line 120: I think this is a bit off

Instead, it is probably something like
* get the beacon state for the latest finalized
* get the shard state for the shard in the latest finalized
* ensure shard_block.slot is later than the shard state slot in the latest finalized beacon state (this can certainly be for a slot earlier than the latest finalized beacon slot

specs/phase1/shard-fork-choice.md
```diff
@@ -0,0 +1,138 @@
...
+    assert shard_block.slot > finalized_slot
+
+    # 4. Check block is a descendant of the finalized block at the checkpoint finalized slot
+    assert (
+        shard_block.beacon_parent_root == store.finalized_checkpoint.root
```
djrtwo, line 125: In the case that the beacon parent root is as old as possible (the latest finalized root), does the second condition below (with `get_ancestor`) not cover this case? Why is this base case needed?

specs/phase1/shard-fork-choice.md
```diff
@@ -59,7 +59,9 @@ def get_shard_latest_attesting_balance(store: Store, shard_store: ShardStore, ro
     return Gwei(sum(
         state.validators[i].effective_balance for i in active_indices
         if (
-            i in store.latest_messages and get_shard_ancestor(
+            i in store.latest_messages and
+            store.latest_messages[i].shard == shard_store.shard and
```
dankrad, line 63: I don't understand this. It seems that only if the validator has attested to the shard _in their latest message_ that their vote would be counted. So in the next slot, when the attest to a different shard, their vote would be removed. Is that how it should work?

djrtwo: It would be in the next epoch rather than slot.
The beacon chain will generally continue to advance to provide a new basis for the shard, and if not, committees will continue to turn over and vote on the shard.

Even if we store multiple shard votes for a single validator, this condition is still important. We need to prevent using a vote on shard A to try to influence shard B (example: using your assigned vote from shard A to vote on a root from shard B, thus throwing your weight behind a shard you weren't assigned)

specs/phase1/shard-fork-choice.md
```diff
@@ -59,7 +59,9 @@ def get_shard_latest_attesting_balance(store: Store, shard_store: ShardStore, ro
     return Gwei(sum(
         state.validators[i].effective_balance for i in active_indices
         if (
-            i in store.latest_messages and get_shard_ancestor(
+            i in store.latest_messages and
```
djrtwo, line 62: In the rest of the repo, we've used the operation on the start of the next line (e.g. `and`) rather than the end of the current line.

Pretty sure this is a recent recommendation in PEP but can't currently find the standard

specs/phase1/shard-fork-choice.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Handlers
+
+#### `on_shard_block`
+
+```python
+def on_shard_block(store: Store, shard_store: ShardStore, signed_shard_block: SignedShardBlock) -> None:
+    shard_block = signed_shard_block.message
+    shard = shard_store.shard
+    # 1. Check shard parent exists
```
terencechain, line 119: Add `assert shard_block.shard == shard` ?

specs/phase1/shard-fork-choice.md
dankrad: True but even if it's in the last epoch, it means the "old" blocks suddenly lose their attestations -- so the commitee in the next epoch could just ignore them if they haven't been crosslinked. Seems weird to me.

specs/phase1/shard-fork-choice.md
djrtwo: I thought about this and assumed that since the client would be the one piping the shard block into this function with the associated `shard_store` that this base mapping already occurred.

That said, I suppose there is not hurt in double checking here

terencechain: Fair point. I'll still add the check on Prysm just to be extra defensive in good way

djrtwo: I'm happy for it to go in.

Another way we can design this is to have `shard_stores` be a map of shard -> shard_store. And `on_shard_block` actually selects the `shard_store` to use.

Curious what @hwwhww feels about the options here

specs/phase1/shard-fork-choice.md
djrtwo: How about we flag this for discussion on Tuesday? I think it's worth hashing out in convo

specs/phase1/shard-transition.md
```diff
@@ -50,6 +50,9 @@ def verify_shard_block_message(beacon_state: BeaconState,
                                shard: Shard) -> bool:
     assert block.shard_parent_root == shard_state.latest_block_root
     assert block.slot == slot
+    next_slot = Slot(beacon_state.slot + 1)
```
hwwhww, line 53: ## ISSUE
We have two places use `get_shard_transition`
1. From validator guide `get_shard_transition`: `beacon_state` the lookahead state where slot = head_beacon_state + 1
2. From shark choice rule `on_shard_block`: `beacon_state` is the head block.

However, #1704 will remove (1). So I leave it here.

specs/phase1/shard-transition.md
```diff
@@ -50,6 +50,9 @@ def verify_shard_block_message(beacon_state: BeaconState,
...
+    offset_slots = compute_offset_slots(get_latest_slot_for_shard(beacon_state, shard), next_slot)
+    assert slot in offset_slots
```
terencechain, line 55: Another thought, should we assert `block.beacon_parent_root` here? (using `beacon_state`)

specs/phase1/shard-transition.md
hwwhww: Second thought: https://github.com/ethereum/eth2.0-specs/pull/1875

specs/phase1/shard-fork-choice.md
```diff
@@ -0,0 +1,181 @@
...
+    active_indices = get_active_validator_indices(state, get_current_epoch(state))
+    return Gwei(sum(
+        state.validators[i].effective_balance for i in active_indices
+        if (
+            i in store.latest_messages
+            # TODO: check the latest message logic: currently, validator's previous vote of another shard
+            # would be ignored once their newer vote is accepted. Check if it makes sense.
+            and store.latest_messages[i].shard == shard_store.shard
+            and get_shard_ancestor(
+                store, shard_store, store.latest_messages[i].root, shard_store.blocks[root].slot
```
terencechain, line 68:
```suggestion
                store, shard_store, store.latest_messages[i].shard_root, shard_store.blocks[root].slot
```

I think that's `shard_root`.  Maybe 👇 is  better to avoid future confusions:

```python
class LatestMessage(object):
    epoch: Epoch
    beacon_root: Root
    shard: Shard
    shard_root: Root
```

hwwhww: Nice catch! 🙈  Thanks!!

specs/phase1/shard-fork-choice.md
djrtwo: I see that.

specs/phase1/shard-fork-choice.md
```diff
@@ -0,0 +1,181 @@
...
+    block_states: Dict[Root, ShardState] = field(default_factory=dict)
+```
+
+#### `get_forkchoice_shard_store`
+
+```python
+def get_forkchoice_shard_store(anchor_state: BeaconState, shard: Shard) -> ShardStore:
+    return ShardStore(
+        shard=shard,
+        blocks={anchor_state.shard_states[shard].latest_block_root: ShardBlock(slot=anchor_state.slot)},
```
djrtwo, line 49: add `shard=shard` to the anchor block stub

specs/phase1/shard-fork-choice.md
```diff
@@ -0,0 +1,181 @@
...
+        )
+    ))
+```
+
+#### `get_shard_head`
+
+```python
+def get_shard_head(store: Store, shard_store: ShardStore) -> Root:
+    # Execute the LMD-GHOST fork choice
+    shard_blocks = shard_store.blocks
```
djrtwo, line 79: A pre-filter of the blocks might be clearer/cleaner than doing the `and shard_blocks[root].slot > head_shard_state.slot` check in each iteration of the `while` loop below

```suggestion
    shard_blocks = [shard_block for shard_block in shard_store.blocks if shard_block.slot > head_shard_state.slot
```

specs/phase1/shard-fork-choice.md
```diff
@@ -0,0 +1,181 @@
...
+
+```python
+def get_shard_ancestor(store: Store, shard_store: ShardStore, root: Root, slot: Slot) -> Root:
+    block = shard_store.blocks[root]
+    if block.slot > slot:
+        return get_shard_ancestor(store, shard_store, block.shard_parent_root, slot)
+    elif block.slot == slot:
+        return root
+    else:
+        # root is older than queried slot, thus a skip slot. Return earliest root prior to slot
```
djrtwo, line 110: We recently modified this comment a bit in the beacon fork choice

```suggestion
        # root is older than queried slot, thus a skip slot. Return most recent root prior to slot
```

specs/phase1/shard-fork-choice.md
```diff
@@ -0,0 +1,181 @@
...
+        return root
+```
+
+#### `get_pending_shard_blocks`
+
+```python
+def get_pending_shard_blocks(store: Store, shard_store: ShardStore) -> Sequence[ShardBlock]:
+    """
+    Return the shard blocks branch that from shard head to beacon head.
```
djrtwo, line 119: I think this might be a bit more clear but still not 100% happy with it

```suggestion
    Return the canonical shard block branch that has not yet been crosslinked.
```

specs/phase1/shard-fork-choice.md
```diff
@@ -0,0 +1,181 @@
...
+        get_ancestor(store, shard_block.beacon_parent_root, finalized_slot) == store.finalized_checkpoint.root
+    )
+
+    # Add new block to the store
+    shard_store.blocks[hash_tree_root(shard_block)] = shard_block
+
+    # Check the block is valid and compute the post-state
+    beacon_head_root = get_head(store)
+    beacon_head_state = store.block_states[beacon_head_root]
+    assert verify_shard_block_message(beacon_head_state, pre_shard_state, shard_block, shard_block.slot, shard)
```
djrtwo, line 176: Why do we pass in `shard_block.slot` as an independent variable? Is there ever a case that we want  the `slot` passed in to differ from `shard.slot`?

hwwhww: Yes it seems leftover of outdated assumption. 👍 
I realized we can also remove `shard` since we got `block.shard`, so it's now just `verify_shard_block_message(beacon_parent_state, shard_parent_state, block)` 🎉

PR #1775, opened 2020-04-30
add clarifying note about attnets ENR entry

djrtwo:
Prysm was updating `attnets` at each epoch based on beacon committee assignments. This is not needed and might actually be harmful in finding stable peers at the right time.

Added a quick clarifying note to the val-guide


q9f:
:+1: thanks

PR #1780, opened 2020-05-01
Handle empty aggregation bits as discussed in #1713

### LINKED ISSUE #1713: Invalid empty attestation signatures consensus split
Some context:
- v0.11.1 had a test helper function change an empty signature from 0xc000... to 0x0000. Completely zero, instead of valid but empty.
- Up to v0.11.0 attestations with no participants were considered valid for some reason, and we had a testcase covering it. (I don't agree with it being a valid case though)
- v0.11.1 also misses the post-state file for some reason, making the empty aggregation bits an anvalid case. And clients pass it, not because they added the rule, but because the 0x0000 signature is invalid.
- I stumbled on this after testing it in a 3rd implementation (ZRNT, Go still, but with different state representation, more like remerkleable)

The scope of the bug:
- The BLS spec has a pre-condition that states that verifying with 0 pubkeys are undefined behavior. [`   Precondition: n >= 1, otherwise return INVALID.`](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02#section-2.9) (Thanks @hwwhww for retrieving this from the specs)
- The previous Eth2 spec versions considered empty attestations included in blocks valid. There was an `empty_aggregation_bits` test case that described it as valid: https://github.com/ethereum/eth2.0-spec-tests/tree/v0.11.0/tests/minimal/phase0/operations/attestation/pyspec_tests/empty_aggregation_bits
- Py-ecc previously handled the 0 pubkeys case as a valid case; the `0xc00000....` signature is valid for no pubkeys.
- The network spec requires attestations on gossip to have exactly 1 bit set, so the network spec is not affected
- The beacon spec just verifies the empty signature. Not all empty signatures are valid. `0x000...` is not. `0xc000...` is.
- Prysm and ZRNT ignore the signature if no bits are set. (ZRNT with a TODO statement): [Prysm](https://github.com/prysmaticlabs/prysm/blob/cdac3d61ea1a4839ce7292bd7006580f10bc4e93/beacon-chain/core/blocks/block_operations.go#L861) [ZRNT](https://github.com/protolambda/zrnt/blob/1a36b8d455d02ac5146c607217b10f931fd1db89/eth2/beacon/attestations/indexed.go#L67)
- Lighthouse puts the attestation signatures in a `SignatureSet`, and then verifies it as a whole. This would allow the `0xc000` signature to pass without pubkeys I think. But not other invalid signatures.
- Other clients have to be checked, there could be more edge cases. If we can agree on what it should be, I will write new spec tests to cover it.

Action points:
- Empty signatures are a dangerous new edge case to attestation processing. Requiring the BLS library to handle the undefined behavior well, or the client to implement the edge case correctly (not like Prysm and ZRNT). We need to be more explicit about this behavior, and possible make it invalid. There should be no reason to include no-participant attestations on chain. Size is not the problem as proposers can already fill their block with duplicates, but just having the empty-signature case at all is bad.
- A chain split could be caused between Prysm and Lighthouse if someone mines a block with an attestation containing an empty bitfield, but an invalid signature. (Which could be a valid looking one still, but not valid for empty set of pubkeys). Prysm will stay on that chain, lighthouse will split away from that chain.

Labeling it as a bug, although the spec is not necessarily wrong, it is definitely a case that should get attention to avoid clients splitting chains in the future.

@djrtwo @CarlBeek

protolambda:
Fix #1713

Marking this as substantive, since it technically breaks existing behavior, if there is any attestation with 0 participants on-chain.

In the process of making the empty-participants case invalid, some tests didn't make sense anymore. Others were broken. Removed the invalid tests, fixed the broken ones, and added more testing (mostly related to rewards/penalties participation changes, re-using functions of the empty/almost-empty cases)

Edit: forced push 2nd commit that handles additional tests. Changed rng seeding to be different between test cases.


djrtwo:
Note that this change is a bit at odds with the comment here -- https://github.com/ethereum/eth2.0-specs/pull/1732#issuecomment-616043988

If we go through with this PR and with #1732, we should probably explicitly disallow 0-participation aggregates on the gossip layer

tests/core/pyspec/eth2spec/test/phase_0/epoch_processing/test_process_rewards_and_penalties.py
```diff
@@ -15,13 +15,16 @@
...
 from eth2spec.test.phase_0.epoch_processing.run_epoch_process_base import run_epoch_processing_with
+from random import Random
 
 
 def run_process_rewards_and_penalties(spec, state):
     yield from run_epoch_processing_with(spec, state, 'process_rewards_and_penalties')
 
 
-def prepare_state_with_full_attestations(spec, state, empty=False):
+def prepare_state_with_full_attestations(spec, state, participation_fn=None):
```
djrtwo, line 25: `prepare_state_with_full_attestations` --  "full" no longer seems valid in the name of this function

protolambda: The default is still full, but agree that naming could be better. Suggestions?

djrtwo: `prepare_state_with_attestations` and note the default is full in a header comment?

tests/core/pyspec/eth2spec/test/helpers/attestations.py
```diff
@@ -113,7 +113,10 @@ def get_valid_late_attestation(spec, state, slot=None, index=None, signed=False)
     return get_valid_attestation(spec, state, slot=slot, index=index, signed=signed, on_time=False)
 
 
-def get_valid_attestation(spec, state, slot=None, index=None, empty=False, signed=False, on_time=True):
+def get_valid_attestation(spec, state, slot=None, index=None, filter_participant_set=None, signed=False, on_time=True):
+    # If filter_participant_set is filters everything, the attestation has 0 participants, and cannot be signed.
```
hwwhww, line 117:
```suggestion
    # If filter_participant_set filters everything, the attestation has 0 participants, and cannot be signed.
```

PR #1781, opened 2020-05-01
test double proposer slashings and exits

protolambda:
See https://github.com/ethereum/eth2.0-specs/issues/1760

I looked into AttesterSlashing tests, but the current mainnet config does not permit double events within a block, and if it were, it would still be a valid case, so clients are more likely to recover.


tests/core/pyspec/eth2spec/test/sanity/test_blocks.py
```diff
@@ -256,11 +254,65 @@ def test_proposer_slashing(spec, state):
...
+
+    yield 'pre', state
+
+    block = build_empty_block_for_next_slot(spec, state)
+    block.body.proposer_slashings = [proposer_slashing_1, proposer_slashing_2]
+    signed_block = state_transition_and_sign_block(spec, state, block, expect_fail=True)
+
+    yield 'blocks', [signed_block]
+    yield 'post', None
+
```
djrtwo, line 296: Should we add `test_double_different_proposer_slashings_same_block`? to show that two proposer slashings _can_ happen in the different proposer case?

tests/core/pyspec/eth2spec/test/sanity/test_blocks.py
```diff
@@ -280,19 +332,11 @@ def test_attester_slashing(spec, state):
...
+    check_attester_slashing_effect(spec, pre_state, state, validator_index)
 
-    proposer_index = spec.get_beacon_proposer_index(state)
-    # gained whistleblower reward
-    assert (
-        get_balance(state, proposer_index) >
-        get_balance(pre_state, proposer_index)
-    )
+# TODO: currently mainnet limits attester-slashings per block to 1.
+# When this is increased, it should be tested to cover varrious combinations
```
djrtwo, line 338:
```suggestion
# When this is increased, it should be tested to cover various combinations
```

tests/core/pyspec/eth2spec/test/sanity/test_blocks.py
```diff
@@ -486,6 +533,27 @@ def test_voluntary_exit(spec, state):
...
+    yield 'pre', state
+
+    # Add to state via block transition
+    initiate_exit_block = build_empty_block_for_next_slot(spec, state)
+    initiate_exit_block.body.voluntary_exits = signed_exits
+    signed_initiate_exit_block = state_transition_and_sign_block(spec, state, initiate_exit_block, expect_fail=True)
+
+    yield 'blocks', [signed_initiate_exit_block]
+    yield 'post', None
+
```
djrtwo, line 555: ditto, consider a multi-exit block with diff indices for sanity check

PR #1782, opened 2020-05-05
[Proposal] BeaconState sync by checkpoint

djrtwo:
Simple proposal for `BeaconState` sync from a provided `Checkpoint`.

* Requires nodes to have access every 256th finalized checkpoint state (in line with Lighthouse's MAX slots per restore point in [their DB config](https://lighthouse-book.sigmaprime.io/advanced_database.html)) within the weak subjectivity period
* Serves the `BeaconState` in a single request (no load balancing available in this method)

Showing up with a recent finalized `Checkpoint` is requisite for safe syncing of the network under weak subjectivity (via commandline arg, baked into release, etc). Adding in a state sync against this recent finalized checkpoint will _greatly_ improve the UX of joining the network (rather than the ever increasing linear block sync).

_[An extra tidbit for thought]_ Clients can also support just inserting a full checkpoint BeaconState outside of p2p sync (i.e. for if you got the full state from some external source rather than just the `Checkpoint` metadata). It might be valuable to cut checkpoint states for common  nets (e.g. mainnet, schlesi, etc) into releases at regular intervals. (That or just the `Checkpoint` metadata)

This is a tough/weird UX problem that changes with PoS. We should all put a little more thought into the common user flows here.

---------------------

For discussion:

Load balancing is not strictly required here, but as validator counts increase, states can become `> 50MB`.

@protolambda suggested that maybe the base request serves a variant of `BeaconState` with a root for `validators`. Then individual validators can be requested against this root in a more load balanced fashion (e.g. `ValidatorsByCheckpointByIndex` that accepts `(checkpoint, indices)` that can easily be distributed across peers).

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -43,6 +43,7 @@ It consists of four main sections:
       - [BeaconBlocksByRoot](#beaconblocksbyroot)
       - [Ping](#ping)
       - [GetMetaData](#getmetadata)
+      - [`BeaconStateByCheckpoint`](#beaconstatebycheckpoint)
   - [The discovery domain: discv5](#the-discovery-domain-discv5)
     - [Integration into libp2p stacks](#integration-into-libp2p-stacks)
     - [ENR structure](#enr-structure)
@@ -155,6 +156,8 @@ This section outlines constants that are used in this spec.
 | `RESP_TIMEOUT` | `10s` | The maximum time for complete response transfer. |
 | `ATTESTATION_PROPAGATION_SLOT_RANGE` | `32` | The maximum number of slots during which an attestation can be propagated. |
 | `MAXIMUM_GOSSIP_CLOCK_DISPARITY` | `500ms` | The maximum milliseconds of clock disparity assumed between honest nodes. |
+| `EPOCHS_PER_SERVED_CHECKPOINT` | `256` | The interval between stored checkpoint states available to be served by `BeaconStateByCheckpoint`. |
+
 
 ## MetaData
 
@@ -624,6 +627,41 @@ it's local most up-to-date MetaData.
 The response MUST be encoded as an SSZ-container.
 
 The response MUST consist of a single `response_chunk`.
+
+#### `BeaconStateByCheckpoint`
+
+**Protocol ID:** `/eth2/beacon_chain/req/beacon_state_by_checkpoint/1/`
+
+Request Content:
+
+```
+(
+  checkpoint: Checkpoint
+)
+```
+
+Response Content:
+
+```
+(
+  state: BeaconState
+)
+```
+
+Requests the `BeaconState` represented by a particular `checkpoint` (that is, the finalized state at slot `compute_start_slot_at_epoch(checkpoint.epoch)` with `hash_tree_root` of `checkpoint.root`).
+
+The requested `checkpoint` MUST be finalized and MUST satisfy `checkpoint.epoch % EPOCHS_PER_SERVED_CHECKPOINT == 0`.
+
+Clients supporting this request MUST keep a record of _finalized_ beacon states for epochs where `epoch % EPOCHS_PER_SERVED_CHECKPOINT == 0` through the weak subjectivity period.
+
+The request MUST be encoded as an SSZ-field.
+
+If failure due to the `checkpoint` not being available in the server's finalized database, response must be a single `response_chunk` with response code `2` (`ServerError`).
+
+If success,
+- Response MUST be encoded as an SSZ-container.
+- `hash_tree_root(state)` MUST be equal to `checkpoint.root`.
+- `state.slot` MUST be equal to `compute_start_slot_at_epoch(checkpoint.epoch)`.
 
 ## The discovery domain: discv5
 


djrtwo:
cc: @nisdas @AgeManning @arnetheduck @cemozerr @ralexstokes @ajsutton @wemeetagain 

We'll discuss this on network call tomorrow https://github.com/ethereum/eth2.0-pm/issues/148

protolambda:
Regarding sync with validators synced separately: it's basically chunking up the majority of the state (for large validator sets, it's the majority of the state easily), and simple enough to implement.

It doesn't have to sync individual validators either, it could also do chunks of e.g. 32 validators at a time, or whatever is optimal.

The state without validators is about 2.6MB (just above current chunk limit), you can take any mainnet teststate (which is on the low of validators) as example.

However, splitting out validators may not be necessary yet, we should experiment more with different state sizes and sync stability.

djrtwo:
> The state without validators is about 2.6MB (just above current chunk limit), you can take any mainnet teststate (which is on the low of validators) as example.

Does this include `balances`?

protolambda:
@djrtwo no it doesn't, that's also growing with validator count, missed that. For 400k validators, that would be 3.2MB, not too bad. So 5.8MB total. Or add it as another special request, but not ideal either.

AgeManning:
What is the actual object in each chunk in the response?
I imagine it's parts of a state, perhaps we need to define this?

paulhauner:
I'm not necessarily opposed to this PR, but I'm still not convinced that we should bother with p2p state syncing instead of just always doing it outside of the protocol (as mentioned as a tidbit).

I understand that this PR is trying to solve the case where:

1. The client wants to sync to head as fast as possible.
1. The client knows the hash of some trusted beacon state.

If this is the only case we're trying to solve, then why doesn't the client always simply download the state from wherever it got the trusted hash? Surely obtaining the states out-of-the-protocol is better for the p2p network (less traffic, less computation) and slightly better for the user (doesn't have to wait for someone to serve its big request).

I raise this because this endpoint is "one more thing to do" and more importantly it's a decent DoS vector for clients. Whilst "checkpoint states" don't require block-replaying to obtain, the SSZ isn't just lying in our database as a neat blob; we've collapsed all the block/state roots so we need to do a few reads and some stitching together in order to obtain the SSZ blob. I don't doubt we can implement protections to feasibly mitigate the DoS vector, but that is indeed more work.

Like I said, not fully against it but I'm still not convinced that the complex solution (get it via p2p) provides enough of a UX improvement to justify it against the simple solution (get it wherever you got the hash).

paulhauner:
Also, talking about complexity I think it's worth looking at what p2p-state-sync looks like from a higher-level perspective. So lets say we have two scenarios:

1. Checkpoint state is ships with the client
    - This can cover the checkpoint being baked into the binary or perhaps being in a file adjacent to the binary.
1. Checkpoint state is obtained via p2p.

I would assume that all clients already have (1) implemented, this is pretty much the same as starting from an existing database. It's different in that we don't know the full chain history, but this scenario is required for (1) and (2).

Implementing (2) adds an additional application state for the client: it is running and connected to the p2p network but it doesn't know the actual values of the checkpoint state. Once it downloads this information it needs to kind of reinitialize itself with this new checkpoint information and move into the normal state of syncing forwards\backwards whilst being grounded in some finalized checkpoint.

My concerns about this "additional application state" are similar to the ones I had a while ago about "waiting for genesis". These additional states create weird scenarios for the API. For example, whilst we're waiting for a p2p state to sync we likely can't serve the `/beacon/head` endpoint. I can only assume we'll find other UX quirks like this down the line.

I've been hoping that after we've finished with eth2 mainnet genesis Lighthouse can completely rip out the "waiting for genesis" code and drop it into a separate tool. Once we do this, Lighthouse will only have one application state: we are grounded in some finalized checkpoint and syncing forwards/backwards. I'm really a big fan of reducing these high-level application states because each one has ripple effects through the codebase.

ajsutton:
It would be really good to avoid requiring nodes to support this. Users should have the option of pruning finalised states to save on disk space if they desire. They may also want to conserve network bandwidth.

And I'd definitely agree with @paulhauner 's comments - we just officially moved away from running the network prior to knowing the initial state, it would be a real shame to go back to requiring it.  

Teku already supports downloading the initial state from a URL so downloading from a HTTPS source rather than via P2P feels more sensible to me.  Ideally block explorers would allow downloading a relatively recent state (or from epoch boundaries) and then you could source the state from the same place you're likely to get the root hash from.  The REST APIs already support retrieving a state via HTTP and at least Lighthouse uses content negotiation to allow it to be retrieved as SSZ instead of JSON which Teku is keen to add as well. Then you could source the initial state from another node easily.

ajsutton:
Oh and I don't believe the calculation for what the weak subjectivity period is actually appears in any spec.  If we're asking nodes to store checkpoints back to the start of the weak subjectivity period, we should have a very clear definition of how to calculate that in the standard somewhere.

arnetheduck:
One thing worth considering is that for the vast majority of cases, the 256-interval states remain of low value - it's a state every day practically, but why would you want a state from 5 or 10 days ago? What security concept motivates you to download that particular state? Storing these states is a major cost relative to the minimum necessary conditions to maintain security without them.

If we accept weak subjectivity periods and finalization points as a concepts, these are the "hotspots" - everything in between is largely uninteresting - WSP is "secure" because a significant amount of time has passed allowing some particular state to "sink in" and become well known - finalization is a more recent point in time that clients build a lot of their security logic around.

Thus, _if_ we're going to be syncing states, it seems wasteful to mandate clients to be able to call up arbitrary states like that - instead, there are the following states which are interesting:

* `state.slot - (state.slot mod WSP_SLOTS) - WSP_SLOTS` and `state.slot - (state.slot mod WSP_SLOTS) - WSP_SLOTS * 2` - the last two known WSP points - by setting WSP_SLOTS to something like 6 months, every 6 months we get a new "blessed" state that serves as "genesis" and the community can rally around - it can be socially published on twitter (hash), bittorrent, web sites etc. We further support downloading the two latest ones to cap the number of states needed to be kept at a constant and to avoid race conditions around the shift

* current and previous finalized state - with the same logic as above, two because race conditions - alternatively finalized and justified - this is a more optimistic approach where you might end up having to sync the state again in the rare case that the justified state changes - but it's less demanding of clients.

Notably, the WSP states do not require in-protocol syncing really - they're more about establishing a norm for how often these checkpoints happen, and what functionality clients and tooling should have with regards to loading them - similar to the interop states or indeed genesis, clients and tooling would end up always have functionality for loading `state+block` as a starting point, regardless if it comes from a web page, or in-protocol.

This would also remove one kludge in the spec: the genesis block. 

Currently, clients must implement special logic to construct the genesis block that works in concert with the genesis state, because the expected functionality is that there is special logic in the client dealing with genesis in particular (which notably happens only once but takes up "space" in the codebase indefinitely).

If instead we could assume that a state+block is given always as input for bootstrapping the "running" part of the beacon node, the logic for bootstrapping is simplified - we would still generate the block as the current genesis block is generated, but the part of the client that "bootstraps" from `state+block` wouldn't know the difference between genesis and wsp, abstractly.

This also plays well with another suggestion I first heard from @paulhauner : we could build a separate bootstrapping tool for the eth1 bootstrap: it would simply monitor the contract and output a state+block when eth1 deposit is "ready", and clients would be oblivious to the exact bootstrapping conditions of eth1 - these would then remain a historical note instead of surface area in the codebases of multiple clients.

djrtwo:
Closing to focus on def of weak subjectivity and out-of-band standards to get state checkpoints.
Leaving p2p state sync as a conversation topic over time

PR #1794, opened 2020-05-08
MAX_ATTESTER_SLASHINGS == 2 and add multiple slashings per block tests

djrtwo:
* Set `MAX_ATTESTER_SLASHINGS` to 2 for both minimal and mainnet config (previously set to 1). Some quick discussion [here](https://github.com/ethereum/eth2.0-specs/issues/1760#issuecomment-623544461). Allows for better testing and allows for faster slashing of attacks on chain (at the cost of higher max block size).
* Add some tests that have multiple slashings in a single block

### DIFF
--- a/tests/core/pyspec/eth2spec/test/phase_0/sanity/test_blocks.py
+++ b/tests/core/pyspec/eth2spec/test/phase_0/sanity/test_blocks.py
@@ -482,10 +482,6 @@ def test_multiple_attester_slashings_partial_overlap(spec, state):
     check_attester_slashing_effect(spec, pre_state, state, full_indices)
 
 
-# TODO: currently mainnet limits attester-slashings per block to 1.
-# When this is increased, it should be tested to cover various combinations
-# of duplicate slashings and overlaps of slashed attestations within the same block
-
 @with_all_phases
 @spec_state_test
 def test_proposer_after_inactive_index(spec, state):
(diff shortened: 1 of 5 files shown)


tests/core/pyspec/eth2spec/test/phase_0/sanity/test_blocks.py
```diff
@@ -363,13 +368,124 @@ def test_attester_slashing(spec, state):
...
-    check_attester_slashing_effect(spec, pre_state, state, validator_index)
+    check_attester_slashing_effect(spec, pre_state, state, slashed_indices)
+
+
+@with_all_phases
+@spec_state_test
+def test_duplicate_attester_slashing(spec, state):
+    # Skip test if config cannot handle multiple AttesterSlashings per block
+    if spec.MAX_ATTESTER_SLASHINGS < 2:
+        return
```
hwwhww, line 379: Should it be removed now? Or you want to leave it in case we change it back?

tests/core/pyspec/eth2spec/test/phase_0/sanity/test_blocks.py
```diff
@@ -363,13 +368,124 @@ def test_attester_slashing(spec, state):
...
+
+    yield 'blocks', [signed_block]
+    yield 'post', state
+
+    check_attester_slashing_effect(spec, pre_state, state, full_indices)
+
 
 # TODO: currently mainnet limits attester-slashings per block to 1.
 # When this is increased, it should be tested to cover various combinations
 # of duplicate slashings and overlaps of slashed attestations within the same block
```
hwwhww, line 488: Can remove it now. 🎉 

```suggestion

```

tests/core/pyspec/eth2spec/test/phase_0/sanity/test_blocks.py
djrtwo: Want to keep it in case we use other configs in the future for testing.

arguably tehre are implicit conditions on a ton of tests, but... this is fine

Issue #1765, opened 2020-04-27
Clarifying `BlocksByRange`


djrtwo:
When discussing with @cemozerr, we realized that it was not abundantly clear that `BlocksByRange` should or should not guarantee to return blocks in the same chain (assuming newer than latest finalized). For example take the following:

* Time 1: canonical view is `A, B, C, D`
* Time 2: start serving request — send `A, B`
* Time 3: local canonical switches to `A, B, C’, D’`
* Time 4: might be any of the following
    a. continue serving same request with original chain -- send `C, D`
    b. continue serving same request with new chain -- send `C’, D’`
    c. stop serving request

_"Clients MUST respond with blocks from their view of the current fork choice"_ is not clear as to whether you fix your view before sending the response _or_ if you should change your view halfway through the request. And thus it is not clear if sending an "inconsistent" chain is okay or malicious.

Chatted with @AgeManning, Lighthouse current does (a). If a peer responds with (a) or (c), all good. If a peer of lighthouse responds (b), it is considered malicious.

--------------------

Chatted with @prestonvanloon and realized that the spec was ambiguous in another way. (Assuming C' and D' exist when the request begins) Prysm currenty sends `A, B, C, C', D, D'` (i.e. the entire sorted-by-slot block tree on the range requested (assuming after finalized)). Although this was not the intended behaviour of the authors of the spec, it is not immediately obvious that it is not a desirable behaviour.

On the one hand, more information can be better (send whole tree, and you decide), but on the other hand, this doesn't provide us with strict bounds on the amount of messages an honest node might respond with. That is, if we allow sending the whole tree, you can’t de-score a peer if they send you a bunch of crap. Constructing a bunch of “valid” looking crap is non-trivial, but if you have a bunch of validators and some extra cpu cycles to check a ton of different shufflings, you could bloat out your responses some amount

--------------------

The questions we must answer and clarify in the spec:
1. Should the `BlocksByRange` request strictly return max 1 block per slot? This was the original intention of the spec, but might be worth just quickly discussing the tree approach
2. If the above is to send just a single chain, MUST the chain be consistent? If so, should clarify that a non-consistent set of blocks is malicious

djrtwo:
cc: @arnetheduck @nisdas

djrtwo:
[...]

PR #1795, opened 2020-05-08
Clarify BlocksByRange request

djrtwo:
address #1765

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -537,7 +537,9 @@ The response MUST contain no more than `count` blocks.
 
 Clients MUST order blocks by increasing slot number.
 
-Clients MUST respond with blocks from their view of the current fork choice. In particular, blocks from slots before the finalization MUST lead to the finalized block reported in the `Status` handshake.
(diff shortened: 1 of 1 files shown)


specs/phase0/p2p-interface.md
```diff
@@ -537,7 +537,9 @@ The response MUST contain no more than `count` blocks.
 
 Clients MUST order blocks by increasing slot number.
 
-Clients MUST respond with blocks from their view of the current fork choice. In particular, blocks from slots before the finalization MUST lead to the finalized block reported in the `Status` handshake.
+Clients MUST respond with blocks from their view of the current fork choice -- that is, blocks from the single chain defined by the current head. Of note, blocks from slots before the finalization MUST lead to the finalized block reported in the `Status` handshake.
+
+Clients MUST respond with blocks that are consistent from a single chain within the context of the request.
```
protolambda, line 542:
```suggestion
Clients MUST respond with blocks that are consistent from a single chain within the context of the request. Clients MAY not respond with any blocks, or stop in the process of responding, if their fork choice changes the view of the chain in the context of the request.
```
Maybe add something like this to cover what happens when consistency cannot be managed by a honest server?

djrtwo: "Clients MUST respond with at least one block" from above. I'll clarify that the process can stop though

djrtwo: done

Issue #1792, opened 2020-05-06
gossipsub v1.1 and extended validators


djrtwo:
* Make explicit the use of gossipsub v1.1 in p2p spec
* Define whether gossipsub application verification return `REJECT` or `IGNORE` (see [v1.1 Extended Validators](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#extended-validators)

PR #1796, opened 2020-05-08
Gossipsub v1.1 and extended validators

djrtwo:
address #1792 

* Make explicit note of gossipsub v1.1 as a requirement
* Add extended validators enums to the gossip conditions


Nashatyrev:
Shouldn't the protocol string be upgraded to `/meshsub/1.1.0`?

protolambda:
@Nashatyrev Good catch, thank you. Yes I think so, although the new Gossipsub should be compatible. @djrtwo what do you think?

Issue #1798, opened 2020-05-09
BLS link points to expired draft


prestonvanloon:
From the [BLS Signatures section](https://github.com/ethereum/eth2.0-specs/blob/master/specs/phase0/beacon-chain.md#bls-signatures) of the beacon chain spec, there is a link to the expired draft. https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-00

I also haven't seen any BLS changes in [v0.12 milestone ](https://github.com/ethereum/eth2.0-specs/milestone/3), but I believe are supposed to target a new IEFT draft. Which draft is required in v0.12 implementation?

hwwhww:
closing via #1799

PR #1799, opened 2020-05-09
Update to IETF BLS draft-irtf-cfrg-bls-signature-02 + draft-irtf-cfrg-hash-to-curve-07

hwwhww:
As #1798 pointed out, adding a draft for v0.12.0.

---
May 12th Update: it's now following IETF BLS `draft-irtf-cfrg-bls-signature-02` (including `draft-irtf-cfrg-hash-to-curve-06`)

If we want to include `draft-irtf-cfrg-hash-to-curve-07` *now*, the spec would be changed to like:

> Eth2 makes use of BLS signatures as specified in the [IETF draft BLS specification v02](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02) but uses [Hashing to Elliptic Curves v07](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-07) instead of v06.

IMHO it's kinda strange and it would be nice to have a new BLS v03 release that includes H2C v7 soon.

---
May 13th update: it's now following IETF BLS `draft-irtf-cfrg-bls-signature-02` +`draft-irtf-cfrg-hash-to-curve-07`.


specs/phase0/beacon-chain.md
```diff
@@ -603,13 +603,13 @@ def bytes_to_int(data: bytes) -> uint64:
 
 #### BLS Signatures
 
-Eth2 makes use of BLS signatures as specified in the [IETF draft BLS specification](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-00). Specifically, eth2 uses the `BLS_SIG_BLS12381G2-SHA256-SSWU-RO-_POP_` ciphersuite which implements the following interfaces:
+ Eth2 makes use of BLS signatures as specified in the [IETF draft BLS specification draft-irtf-cfrg-bls-signature-02](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02) but uses [Hashing to Elliptic Curves - draft-irtf-cfrg-hash-to-curve-07](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-07) instead of draft-irtf-cfrg-hash-to-curve-06. Specifically, eth2 uses the `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_` ciphersuite which implements the following interfaces:
```
djrtwo, line 606: Maybe we should add something like the following below: 

"_Note_: The non-standard configuration of the BLS and hash to curve specs is temporary and will be resolved once IETF releases BLS spec draft 3."

specs/phase0/beacon-chain.md
```diff
@@ -603,13 +603,13 @@ def bytes_to_int(data: bytes) -> uint64:
...
 
 - `def Sign(SK: int, message: Bytes) -> BLSSignature`
 - `def Verify(PK: BLSPubkey, message: Bytes, signature: BLSSignature) -> bool`
 - `def Aggregate(signatures: Sequence[BLSSignature]) -> BLSSignature`
 - `def FastAggregateVerify(PKs: Sequence[BLSPubkey], message: Bytes, signature: BLSSignature) -> bool`
-- `def AggregateVerify(pairs: Sequence[PK: BLSPubkey, message: Bytes], signature: BLSSignature) -> bool`
+- `def AggregateVerify(PKs: Sequence[BLSPubkey], message: Sequence[Bytes], signature: BLSSignature) -> bool`
```
djrtwo, line 612: should this be `messages` in the args?

```suggestion
- `def AggregateVerify(PKs: Sequence[BLSPubkey], messages: Sequence[Bytes], signature: BLSSignature) -> bool`
```

specs/phase1/shard-transition.md
```diff
@@ -280,7 +280,10 @@ def get_shard_transition(beacon_state: BeaconState,
         if proposal.signature != BLSSignature():
             proposer_signatures.append(proposal.signature)
 
-    proposer_signature_aggregate = bls.Aggregate(proposer_signatures)
+    if len(proposer_signatures) > 0:
+        proposer_signature_aggregate = bls.Aggregate(proposer_signatures)
+    else:
+        proposer_signature_aggregate = BLSSignature()
```
mratsim, line 286: Do we define the empty signature somewhere? Is that 0xc000...

protolambda: The python here would mean all zeroes (every ssz type is all zeroes by default). But agree that this is way to vague and can be confused with an empty signature.

hwwhww: I remember we had `EMPTY_SIGNATURE` constant in 2019. I would be happy to add it back.

mratsim: We can have, `NO_SIGNATURE`(0x0000...) and `EMPTY_SIGNATURE` (0xc000...) for clarity

kirk-baird: I've been treating `EMPTY_SIGNATURE` as (0x0000...) as I believe that's what we called in it the spec in 2019. I'm happy to do a re-name and make
* `NO_SIGNATURE` (0x0000...) (always verifies false)
* `EMPTY_SIGNATURE` (0xc000...) 

However, we have to be careful here as `AggregatePublicKeys([])` will always verify false according to the BLS Standard.

So `EMPTY_SIGNATURE` will only verify true against a list of `EMPTY_PUBLIC_KEYS`. We must have `1..*` `EMPTY_PUBLIC_KEYS` verifying true. (Note: or some other combination of public keys that add to 0 mod r).

mratsim: Or `EMPTY_SIGNATURE`(0x0000) and `ZERO_SIGNATURE` (0xc000) I'm not married to any name.

hwwhww: Hey I'm redirecting this discussion back to the original #1713 thread: https://github.com/ethereum/eth2.0-specs/issues/1713#issuecomment-628518937 😄

hwwhww:
@mratsim 

> Another thing is that ideally I'd like the `verify` functions to include the proof-of-possession verification or mention in the BLS paragraph that it assumed that PoP were verified at stage X.

We do state that the ciphersuite ID is `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_`, it includes "POP".

Or do you mean `PopProve(SK)` and `PopVerify(PK, proof)` APIs? From my understanding, we don't use `PopProve` for the deposit proof-of-possession since we have to sign over `withdrawal_credentials`.

specs/phase0/beacon-chain.md
```diff
@@ -603,16 +603,18 @@ def bytes_to_int(data: bytes) -> uint64:
 
 #### BLS Signatures
 
-Eth2 makes use of BLS signatures as specified in the [IETF draft BLS specification](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-00). Specifically, eth2 uses the `BLS_SIG_BLS12381G2-SHA256-SSWU-RO-_POP_` ciphersuite which implements the following interfaces:
+ Eth2 makes use of BLS signatures as specified in the [IETF draft BLS specification draft-irtf-cfrg-bls-signature-02](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02) but uses [Hashing to Elliptic Curves - draft-irtf-cfrg-hash-to-curve-07](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-07) instead of draft-irtf-cfrg-hash-to-curve-06. Specifically, eth2 uses the `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_` ciphersuite which implements the following interfaces:
```
djrtwo, line 606: format: remove leading space

```suggestion
Eth2 makes use of BLS signatures as specified in the [IETF draft BLS specification draft-irtf-cfrg-bls-signature-02](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02) but uses [Hashing to Elliptic Curves - draft-irtf-cfrg-hash-to-curve-07](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-07) instead of draft-irtf-cfrg-hash-to-curve-06. Specifically, eth2 uses the `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_` ciphersuite which implements the following interfaces:
```

configs/mainnet.yaml
```diff
@@ -156,6 +156,8 @@ DOMAIN_SHARD_PROPOSAL: 0x80000000
 DOMAIN_SHARD_COMMITTEE: 0x81000000
 DOMAIN_LIGHT_CLIENT: 0x82000000
 DOMAIN_CUSTODY_BIT_SLASHING: 0x83000000
+# Constant
+NO_SIGNATURE: 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```
protolambda, line 160: We don't generally put constants that never change into configs as if they are meant to change. Just keeping it in the beacon spec is good enough. Clients will optimize this one way or another anyway, since it never changes.

configs/minimal.yaml
```diff
@@ -156,6 +156,8 @@ DOMAIN_SHARD_PROPOSAL: 0x80000000
 DOMAIN_SHARD_COMMITTEE: 0x81000000
 DOMAIN_LIGHT_CLIENT: 0x82000000
 DOMAIN_CUSTODY_BIT_SLASHING: 0x83000000
+# Constant
+NO_SIGNATURE: 0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```
protolambda, line 160: Ditto

PR #1800, opened 2020-05-09
Update p2p reqresp with explicit handling of reqresp size bounds

protolambda:
Update p2p doc to:
- Reflect ability to check much stricter SSZ size bounds (as part of the encoding strategy)
- That the parts of req/resps should be validated according to encoding strategy
- Encoding strategy describes SSZ bounds checking in more detail. Previously already checking min/max somewhat, but not enforcing it as well.
- Include FAQ item for SSZ type size bounds
- Restrict varint bytes to max 10, to avoid unlimited varint attack
- Misc. consistency, clarification and de-duplication things.

-----

Related, now that we are looking more critically at the req-resp payload format/sizing:

I see we have both:
```

response_chunk  ::= <result> | <encoding-dependent-header> | <encoded-payload>
```

```
  +--------+--------+--------+--------+--------+--------+
  | result |   header (opt)  |     encoded_response     |
  +--------+--------+--------+--------+--------+--------+
```

Can we maybe remove the second one? It's nice ascii, but this and the sudden introduction of "A response therefore has the form ..." after going in-depth a lot already feels double to me. And the optional header is confusing with the earlier BNF. The header can be empty if the encoding strategy doesn't need any header data, that's all.

----

Edit: above duplicate section has been removed. The BNF-like diagram already has this exact information. And `result` being one byte is explained as part of the responding side.

Force push edit: bounds are now described as `type size bounds` instead of `type bounds`

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -98,6 +98,7 @@ It consists of four main sections:
     - [Why are we compressing, and at which layers?](#why-are-we-compressing-and-at-which-layers)
     - [Why are using Snappy for compression?](#why-are-using-snappy-for-compression)
     - [Can I get access to unencrypted bytes on the wire for debugging purposes?](#can-i-get-access-to-unencrypted-bytes-on-the-wire-for-debugging-purposes)
+    - [What are SSZ type size bounds?](#what-are-ssz-type-size-bounds)
 - [libp2p implementations matrix](#libp2p-implementations-matrix)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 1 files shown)


specs/phase0/p2p-interface.md
```diff
@@ -329,23 +330,35 @@ result    ::= “0” | “1” | “2” | [“128” ... ”255”]
...
 
-A `response` is formed by zero or more `response_chunk`s. Responses that consist of a single SSZ-list (such as `BlocksByRange` and `BlocksByRoot`) send each list item as a `response_chunk`. All other response types (non-Lists) send a single `response_chunk`. The encoded-payload of a `response_chunk` has a maximum uncompressed byte size of `MAX_CHUNK_SIZE`.
+A `response` is formed by zero or more `response_chunk`s. Responses that consist of a single SSZ-list (such as `BlocksByRange` and `BlocksByRoot`) send each list item as a `response_chunk`. All other response types (non-Lists) send a single `response_chunk`.
 
-Clients MUST ensure the each encoded payload of a `response_chunk` is less than or equal to `MAX_CHUNK_SIZE`; if not, they SHOULD reset the stream immediately. Clients tracking peer reputation MAY decrement the score of the misbehaving peer under this circumstance.
+For both `request`s and `response`s, he `encoding-dependent-header` MUST be valid, and the `encoded-payload` must be valid within the constraints of the `encoding-dependent-header`.
```
djrtwo, line 335:
```suggestion
For both `request`s and `response`s, the `encoding-dependent-header` MUST be valid, and the `encoded-payload` must be valid within the constraints of the `encoding-dependent-header`.
```

specs/phase0/p2p-interface.md
```diff
@@ -329,23 +330,35 @@ result    ::= “0” | “1” | “2” | [“128” ... ”255”]
...
+The requester MUST wait a maximum of `TTFB_TIMEOUT` for the first response byte to arrive (time to first byte—or TTFB—timeout). On that happening, the requester allows a further `RESP_TIMEOUT` for each subsequent `response_chunk` received.
 
 If any of these timeouts fire, the requester SHOULD reset the stream and deem the req/resp operation to have failed.
 
+A requester SHOULD read from the stream until either:
+a) An error result is received in one of the chunks (the error payload MAY be read before stopping).
```
djrtwo, line 351: This doesn't render nicely in the markdown (looks all on the same line). Let's use 1., 2., etc.. instead of a. b.

PR #1803, opened 2020-05-11
BLS options for pyspec

protolambda:
Changelog:
- BLS in tests that don't need it necessarily is ON by default, `--disable-bls` is explicit opt-out, to make sure testers know what is running.
- Makefile testing is still bls-off, as to not change any behavior. If we start using `--bls-type=milagro`, it may be fast enough to default to bls-on.
- Test config option to use milagro bls bindings of @ChihChengLiang , `--bls-type=milagro`. Also possible to switch programmatically as library consumer, just run:
```python
from eth2spec.utils import bls as bls_utils
bls_utils.bls = bls_utils.milagro_bls
```

Note: Milagro does not seem to work for some BLS functions used in phase 1. I think due to a type error. And the milagro bindings are not documented, so I am not just how they differ from `py_ecc` (they are supposed to be the same). Also needed to make some iterable inputs explicitly a list, so milagro-bindings could deal with them.


djrtwo:
So should we fix those phase 1 tests before merge?

protolambda:
@djrtwo We could, although we have py-ecc working at least.
And you or @hwwhww can judge on BLS usage for phase1 testing better than me.

Asking @ChihChengLiang: do you know what python types milagro BLS bindings is expecting?

hwwhww:
@protolambda 
> Milagro does not seem to work for some BLS functions used in phase 1.

Is the issue only in phase 1? I run `pytest --bls-type=milagro eth2spec/test/phase_0` with this branch and got

hwwhww:
(^^^^ it was a fat-fingered accident, sorry)

I think it's a BLS v2 implementation error:
1. In `sigp/milagro_bls` of `AggregateVerify`, it checks that messages are distinct: https://github.com/sigp/milagro_bls/blob/2ccdd4b517c1ab3debe10277deed9d1b1cbbe9ce/src/aggregates.rs#L142-L152
2. In the IETF standard, [G2Basic checks that messages are distinct](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02#section-3.1.1), but we use [G2ProofOfPossession that **doesn't** check message are distinct](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02#section-3.3).
3. It seems to be solved in the [recent milagro BLS development](https://github.com/apache/incubator-milagro-crypto-rust/blob/develop/src/bls381/proof_of_possession.rs) by @kirk-baird, but it's not released yet.

kirk-baird:
> I think it's a BLS standard implementation error:
> 
> 1. In `sigp/milagro_bls` of `AggregateVerify`, it checks that messages are distinct: https://github.com/sigp/milagro_bls/blob/2ccdd4b517c1ab3debe10277deed9d1b1cbbe9ce/src/aggregates.rs#L142-L152
> 2. In the IETF standard, [G2Basic checks that messages are distinct](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02#section-3.1.1), but we use [G2ProofOfPossession that **doesn't** check messages are distinct](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02#section-3.3).

Yep this is correct, I'm updating `milagro_bls` at the moment and will move the unique message check as it is not required in PoP. This will be done soon though there probably won't be a release for a another few days while I make some changes up stream in [apache](https://github.com/apache/incubator-milagro-crypto-rust/).

> 3. It seems to be solved in the [recent milagro BLS development](https://github.com/apache/incubator-milagro-crypto-rust/blob/develop/src/bls381/proof_of_possession.rs) by @kirk-baird, but it's not released yet.

This repo should now be fully compliant with the standards (however it still needs some thorough testing and a final decision on the API layout). Upto you whether you use this repository directly or the `milagro_bls` repository. `milagro_bls` will do less serialisation and deserialisation by caching the intermediate results.

djrtwo:
@kirk-baird any update on the next `milagro_bls` release?

kirk-baird:
@djrtwo I'll make a release today!

tests/core/pyspec/eth2spec/utils/bls.py
```diff
@@ -60,7 +65,10 @@ def Aggregate(signatures):
 
 @only_with_bls(alt_return=STUB_SIGNATURE)
 def Sign(SK, message):
-    return bls.Sign(SK, message)
+    if bls == py_ecc_bls:
+        return bls.Sign(SK, message)
+    else:
+        return bls.Sign(SK.to_bytes(32, 'big'), message)
```
hwwhww, line 71: It seems that milagro_bls requires 32-byte privkey: https://github.com/apache/incubator-milagro-crypto-rust/blob/736cdcb541daaae840a8fbc23d888831b134709b/src/bls381/core.rs#L79-L81

protolambda:
@hwwhww BLS is 3-fold:
- No decorator is: test-settings. These can be overridden with the BLS pytest config option. The generators force BLS to be ON, except for those with `never_bls`
- `always_bls` forces BLS to be on, regardless of default setting. The test won't and can't work without.
- `never_bls` forces the BLS to be off, regardless of default setting. The test won't and can't work with BLS.

I recommend just not using a decorator unless necessary. With Milagro speed we could soon enable BLS by default in the CI, but previously it would change the test time by so much that you can get multiple coffees before the tests finished. If you are just running a few test, enabling it with the config option passed to pytest is easy.

protolambda:
Update: checked the changes by @hwwhww, and it makes sense to force BLS to be on there. Good catch.

tests/core/pyspec/eth2spec/utils/bls.py
protolambda: Yes, it was one of those differences I hit when first trying it, and I had to dive into the binding benchmark code for an example of the correct private key format. Just checking the module, and converting to 32 bytes, seemed like the right thing to do. But I'm not 100% it produces the same effective signature, we should check.

adiasg:
Spec test suite running ~20% faster! :clap: :partying_face:
Thanks to all involved!

Issue #1777, opened 2020-04-30
ATTESTATION_SUBNET_COUNT should be dynamic


jrhea:
`ATTESTATION_SUBNET_COUNT` is currently a spec constant that defines the valid range of subnets persistent committees should randomly subscribe to (this value is currently set to 64).  When the network has a smaller number of validators, this results in nodes joining subnets with committee counts of 0.  Spreading persistent committee members across unused subnets could significantly inhibit our ability to find subnets in a timely fashion.  

The following table compares active validators to resulting committee counts:

| ACTIVE VALIDATORS | COMMITTEE COUNT |
|----------------------|-----------------------|
|   16384 |  4|
| 32768 | 8 |
| ...|...|
|262144| 64|

In other words,  we don't need 64 attestation subnets until we reach 262,144 active validators.  

To fix this, I suggest that we replace  `ATTESTATION_SUBNET_COUNT` with a method that calculates it based on the committee count.  Consider something like this to start:

```python
def get_attestation_subnet_count_at_slot(state: BeaconState, slot: Slot) -> uint64:  
    return min(get_committee_count_at_slot(state,slot), 64)
```
that seems fine if we assume a static validator set, but we really should consider how many validators could exit or activate within a `PERSISTENT_COMMITTEE_PERIOD` to ensure we create enough attestation subnets.  To do that, we should take a look at worst case validator churn limits.  From the spec, `get_validator_churn_limit()` is defined as follows:

```python
def get_validator_churn_limit(state: BeaconState) -> uint64:
    """
    Return the validator churn limit for the current epoch.
    """
    active_validator_indices = get_active_validator_indices(state, get_current_epoch(state))
    return max(MIN_PER_EPOCH_CHURN_LIMIT, len(active_validator_indices) // CHURN_LIMIT_QUOTIENT)
```

In order to have a churn limit > 4 , this would have to be true:

```len(active_validator_indices)  > 4 * CHURN_LIMIT_QUOTIENT```

In other words, we would need 262,144 active validators.  From the table above, we know this is the exact amount needed to reach a committee count of 64.  This means that we can calculate the max increase/decrease in committee counts within a `PERSISTEN_COMMITTEE_PERIOD` as follows:

```
Change in num active validators = 4*PERSISTEN_COMMITTEE_PERIOD = +/- 8,192 validators
Change in committee count = 8,192/SLOTS_PER_EPOCH/TARGET_COMMITTEE_SIZE = +/- 2
```

[...]

PR #1804, opened 2020-05-11
[for discussion] Use all attestation subnets in phase 0

djrtwo:
An attempt to address [this](https://github.com/ethereum/eth2.0-specs/issues/1777#issuecomment-626235833). 

In Phase 1, all attestation subnets would get used even when committees_per_slot does not equal `SHARD_COUNT`. In this PR, all attestation subnets are now used by rotating through them all throughout the epoch instead of just using a subset at each slot starting at 0.

This _does_ mimic the behaviour of Phase 1 more directly, but introduces a change to broadcast logic kind of late in the game.

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -220,12 +220,14 @@ The payload is carried in the `data` field of a gossipsub message, and varies de
 
 | Name                                           | Message Type            |
 |------------------------------------------------|-------------------------|
-| beacon\_block                                  | SignedBeaconBlock       |
-| beacon\_aggregate\_and\_proof                  | SignedAggregateAndProof |
-| beacon_attestation\_{subnet\_id}               | Attestation             |
-| voluntary\_exit                                | SignedVoluntaryExit     |
-| proposer\_slashing                             | ProposerSlashing        |
-| attester\_slashing                             | AttesterSlashing        |
+| Name                             | Message Type              |
+|----------------------------------|---------------------------|
+| `beacon_block`                   | `SignedBeaconBlock`       |
+| `beacon_aggregate_and_proof`     | `SignedAggregateAndProof` |
+| `beacon_attestation_{subnet_id}` | `Attestation`             |
+| `voluntary_exit`                 | `SignedVoluntaryExit`     |
+| `proposer_slashing`              | `ProposerSlashing`        |
+| `attester_slashing`              | `AttesterSlashing`        |
 
 Clients MUST reject (fail validation) messages containing an incorrect type, or invalid payload.
 
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -151,7 +151,6 @@ This section outlines constants that are used in this spec.
 |---|---|---|
 | `GOSSIP_MAX_SIZE` | `2**20` (= 1048576, 1 MiB) | The maximum allowed size of uncompressed gossip messages. |
 | `MAX_CHUNK_SIZE` | `2**20` (1048576, 1 MiB) | The maximum allowed size of uncompressed req/resp chunked responses. |
-| `ATTESTATION_SUBNET_COUNT` | `64` | The number of attestation subnets used in the gossipsub protocol. |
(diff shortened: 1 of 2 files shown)


djrtwo:
cc: @AgeManning @nisdas @arnetheduck @jrhea

AgeManning:
This looks fine to us.

zilm13:
@djrtwo it looks like minor change for subnetwork calculation formula when you choose one to post your attestation, but @jrhea proposal was to decrease number of subnetworks in order to increase size of each subnetwork skeleton. In this one we still have small skeletons in small networks. If we kinda combine it or smth like this, it could serve similarly to #1777, but it would look more complicated than original PR for me.

jrhea:
I just want to make sure I understand what is happening here.  Assuming, 

- first slot in a new epoch
- a constant get_committee_count_at_slot() == 4

the possible subnets for each slot would be:

- slot 0: [0,3]
- slot 1: [4,7]
- etc

Is that right?  If so, I have a couple of comments/questions:

1) isn't `get_committee_count_at_slot()` always going to be constant for each slot in an epoch?  if so, seems like `compute_subnet_for_attestation()` we could avoid the for loop.  

2) is the point here to mimic the scenario in phase 1 where we don't have enough validators to form crosslinks every slot for each shard?

djrtwo:
> @jrhea proposal was to decrease number of subnetworks in order to increase size of each subnetwork skeleton

Yes, I'm aware. We discussed this on the networking call. The desire, if we change it, is to make it more like the distribution we would see with actual shards. I am worried about artificially increasing the size of subnets in phase 0 when in actuality, come phase 1, they are used more like in the PR

djrtwo:
> isn't get_committee_count_at_slot() always going to be constant for each slot in an epoch? if so, seems like compute_subnet_for_attestation() we could avoid the for loop.

you're right, the for loop is unnecessary here due to # of committees being equal across all slots in an epoch

    is the point here to mimic the scenario in phase 1 where we don't have enough validators to form crosslinks every slot for each shard?

Yes, in low validation numbers, you rotate through shards across slots when you don't have high enough committee numbers. When you get to enough validators to have 64 committees per slot, you just  get usage of all subnets at all slots (same as you'd see in the current phase 0 setup)

specs/phase0/p2p-interface.md
```diff
@@ -221,12 +220,12 @@ The payload is carried in the `data` field of a gossipsub message, and varies de
...
-| committee_index{subnet_id}\_beacon_attestation | Attestation             |
-| voluntary_exit                                 | SignedVoluntaryExit     |
-| proposer_slashing                              | ProposerSlashing        |
-| attester_slashing                              | AttesterSlashing        |
+| beacon\_block                                  | SignedBeaconBlock       |
+| beacon\_aggregate\_and\_proof                  | SignedAggregateAndProof |
+| beacon_attestation\_{subnet\_id}               | Attestation             |
+| voluntary\_exit                                | SignedVoluntaryExit     |
+| proposer\_slashing                             | ProposerSlashing        |
+| attester\_slashing                             | AttesterSlashing        |
```
protolambda, line 228:
```suggestion
| Name                             | Message Type              |
|----------------------------------|---------------------------|
| `beacon_block`                   | `SignedBeaconBlock`       |
| `beacon_aggregate_and_proof`     | `SignedAggregateAndProof` |
| `beacon_attestation_{subnet_id}` | `Attestation`             |
| `voluntary_exit`                 | `SignedVoluntaryExit`     |
| `proposer_slashing`              | `ProposerSlashing`        |
| `attester_slashing`              | `AttesterSlashing`        |
```

Instead of the weird escaping, we should just use code here, like in the text and the table above.

mkalinin:
> The desire, if we change it, is to make it more like the distribution we would see with actual shards. I am worried about artificially increasing the size of subnets in phase 0 when in actuality, come phase 1, they are used more like in the PR

For the record. We _might_ use one subnet for several shards in Phase 1 if it would impact security on the network layer.

djrtwo:
> For the record. We might use one subnet for several shards in Phase 1 if it would impact security on the network layer.

True! that's why we have `ATTESTATION_SUBNET_COUNT` to adjust

tersec:
For the record, this looks fine from Nimbus's end as well.

Issue #1787, opened 2020-05-06
bitfield_bytes(bits) should return chunks


booleanfunction:
The [simple-serialise](https://github.com/ethereum/eth2.0-specs/blob/dev/ssz/simple-serialize.md)  (Merkleisation section) reads:

> bitfield_bytes(bits): return the bits of the bitlist or bitvector, packed in bytes, aligned to the start. Length-delimiting bit for bitlists is excluded.

However when this function is called for merkleisation of bitvectors or bitlists, the output expected from `bitfield_bytes` is `BYTES_PER_CHUNK`-byte chunks, as indicated by the specification of `merkleize`, generally, 

> `merkleize(chunks, limit=None)`

and then the specific definition for bitvectors or bitlists:

> `merkleize(bitfield_bytes(value), limit=chunk_count(type))` if `value` is a bitvector.
> `mix_in_length(merkleize(bitfield_bytes(value), limit=chunk_count(type)), len(value))` if `value` is a bitlist.

hwwhww:
Thanks for the report!

I suppose the spec is misleading but the implementations are correct. /cc @protolambda

Ideally, maybe rename `bitfield_bytes` to `pack_bits` would be more clear.

> `pack(value)`: given ordered objects of the same basic type, serialize them, pack them into `BYTES_PER_CHUNK`-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
> `pack_bits(bits)`: Given the `bits` of bitlist or bitvector, get `bitfield_bytes` by packing them in bytes and aligning to the start. Length-delimiting bit for bitlists is excluded. And them pack `bitfield_bytes` into `BYTES_PER_CHUNK`-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.


> `merkleize(pack_bits(value), limit=chunk_count(type))` if `value` is a bitvector.
> `mix_in_length(merkleize(pack_bits(value), limit=chunk_count(type)), len(value))` if `value` is a bitlist.

protolambda:
Closing this, I think this is solved in #1806

PR #1806, opened 2020-05-13
Fix `bitfield_bytes` type error

hwwhww:
### Issue
Address #1787.

### How did I fix it
Rename `bitfield_bytes`to `pack_bits` and make this function return `BYTES_PER_CHUNK`-byte chunks.

p.s. it seems the spec was misleading but the implementations are all good? /cc @protolambda to confirm.

### DIFF
--- a/ssz/simple-serialize.md
+++ b/ssz/simple-serialize.md
@@ -211,8 +211,8 @@ We first define helper functions:
    * `List[B, N]` and `Vector[B, N]`, where `B` is a basic type: `(N * size_of(B) + 31) // 32` (dividing by chunk size, rounding up)
    * `List[C, N]` and `Vector[C, N]`, where `C` is a composite type: `N`
    * containers: `len(fields)`
-* `bitfield_bytes(bits)`: return the bits of the bitlist or bitvector, packed in bytes, aligned to the start. Length-delimiting bit for bitlists is excluded.
-* `pack`: Given ordered objects of the same basic type, serialize them, pack them into `BYTES_PER_CHUNK`-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
+* `pack(value)`: given ordered objects of the same basic type, serialize them, pack them into `BYTES_PER_CHUNK`-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
+* `pack_bits(bits)`: Given the `bits` of bitlist or bitvector, get `bitfield_bytes` by packing them in bytes and aligning to the start. Length-delimiting bit for bitlists is excluded. And them pack `bitfield_bytes` into `BYTES_PER_CHUNK`-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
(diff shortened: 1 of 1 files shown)


ssz/simple-serialize.md
```diff
@@ -228,9 +228,9 @@ We first define helper functions:
 We now define Merkleization `hash_tree_root(value)` of an object `value` recursively:
 
 * `merkleize(pack(value))` if `value` is a basic object or a vector of basic objects.
-* `merkleize(bitfield_bytes(value), limit=chunk_count(type))` if `value` is a bitvector.
+* `merkleize(pack_bits(value), limit=chunk_count(type))` if `value` is a bitvector.
 * `mix_in_length(merkleize(pack(value), limit=chunk_count(type)), len(value))` if `value` is a list of basic objects.
-* `mix_in_length(merkleize(bitfield_bytes(value), limit=chunk_count(type)), len(value))` if `value` is a bitlist.
+* `mix_in_length(merkleize(pack_bits(value), limit=chunk_count(type)), len(value))` if `value` is a bitlist.
```
protolambda, line 233: `mix_in_length(merkleize(pack(bitfield_bytes(value),  limit=chunk_count(type)), len(value))` would also work. But being explicit here is not a bad idea.

ssz/simple-serialize.md
```diff
@@ -211,8 +211,8 @@ We first define helper functions:
    * `List[B, N]` and `Vector[B, N]`, where `B` is a basic type: `(N * size_of(B) + 31) // 32` (dividing by chunk size, rounding up)
    * `List[C, N]` and `Vector[C, N]`, where `C` is a composite type: `N`
    * containers: `len(fields)`
-* `bitfield_bytes(bits)`: return the bits of the bitlist or bitvector, packed in bytes, aligned to the start. Length-delimiting bit for bitlists is excluded.
-* `pack`: Given ordered objects of the same basic type, serialize them, pack them into `BYTES_PER_CHUNK`-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
+* `pack(value)`: given ordered objects of the same basic type, serialize them, pack them into `BYTES_PER_CHUNK`-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
+* `pack_bits(bits)`: Given the `bits` of bitlist or bitvector, get `bitfield_bytes` by packing them in bytes and aligning to the start. Length-delimiting bit for bitlists is excluded. And them pack `bitfield_bytes` into `BYTES_PER_CHUNK`-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
```
protolambda, line 215:
```suggestion
* `pack_bits(bits)`: Given the `bits` of bitlist or bitvector, get `bitfield_bytes` by packing them in bytes and aligning to the start. The length-delimiting bit for bitlists is excluded. And then pack `bitfield_bytes` into `BYTES_PER_CHUNK`-byte chunks, right-pad the last chunk with zero bytes, and return the chunks.
```

protolambda:
Also confirm that the spec was misleading here, good catch and fix. Thanks.

PR #1810, opened 2020-05-13
Make obvious that  `source_epoch` does not underflow

terencechain:
For those conditions, a node can still underflow even `source_epoch` greater than 0. I think we meant to use the actual period there

### DIFF
--- a/specs/phase1/beacon-chain.md
+++ b/specs/phase1/beacon-chain.md
@@ -494,7 +494,7 @@ def get_shard_committee(beacon_state: BeaconState, epoch: Epoch, shard: Shard) -
 ```python
 def get_light_client_committee(beacon_state: BeaconState, epoch: Epoch) -> Sequence[ValidatorIndex]:
     source_epoch = epoch - epoch % LIGHT_CLIENT_COMMITTEE_PERIOD
-    if source_epoch => LIGHT_CLIENT_COMMITTEE_PERIOD:
+    if source_epoch >= LIGHT_CLIENT_COMMITTEE_PERIOD:
         source_epoch -= LIGHT_CLIENT_COMMITTEE_PERIOD
     active_validator_indices = get_active_validator_indices(beacon_state, source_epoch)
     seed = get_seed(beacon_state, source_epoch, DOMAIN_LIGHT_CLIENT)
--- a/specs/phase1/beacon-chain.md
+++ b/specs/phase1/beacon-chain.md
@@ -476,7 +476,7 @@ def get_online_validator_indices(state: BeaconState) -> Set[ValidatorIndex]:
 ```python
 def get_shard_committee(beacon_state: BeaconState, epoch: Epoch, shard: Shard) -> Sequence[ValidatorIndex]:
     source_epoch = epoch - epoch % SHARD_COMMITTEE_PERIOD
-    if source_epoch > 0:
+    if source_epoch > SHARD_COMMITTEE_PERIOD:
         source_epoch -= SHARD_COMMITTEE_PERIOD
     active_validator_indices = get_active_validator_indices(beacon_state, source_epoch)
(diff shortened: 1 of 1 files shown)


specs/phase1/beacon-chain.md
```diff
@@ -476,7 +476,7 @@ def get_online_validator_indices(state: BeaconState) -> Set[ValidatorIndex]:
 ```python
 def get_shard_committee(beacon_state: BeaconState, epoch: Epoch, shard: Shard) -> Sequence[ValidatorIndex]:
     source_epoch = epoch - epoch % SHARD_COMMITTEE_PERIOD
-    if source_epoch > 0:
+    if source_epoch > SHARD_COMMITTEE_PERIOD:
```
hwwhww, line 479:
```suggestion
    if source_epoch >= SHARD_COMMITTEE_PERIOD:
```

specs/phase1/beacon-chain.md
```diff
@@ -494,7 +494,7 @@ def get_shard_committee(beacon_state: BeaconState, epoch: Epoch, shard: Shard) -
 ```python
 def get_light_client_committee(beacon_state: BeaconState, epoch: Epoch) -> Sequence[ValidatorIndex]:
     source_epoch = epoch - epoch % LIGHT_CLIENT_COMMITTEE_PERIOD
-    if source_epoch > 0:
+    if source_epoch > LIGHT_CLIENT_COMMITTEE_PERIOD:
```
hwwhww, line 497:
```suggestion
    if source_epoch => LIGHT_CLIENT_COMMITTEE_PERIOD:
```

specs/phase1/beacon-chain.md
```diff
@@ -476,7 +476,7 @@ def get_online_validator_indices(state: BeaconState) -> Set[ValidatorIndex]:
 ```python
 def get_shard_committee(beacon_state: BeaconState, epoch: Epoch, shard: Shard) -> Sequence[ValidatorIndex]:
     source_epoch = epoch - epoch % SHARD_COMMITTEE_PERIOD
-    if source_epoch > 0:
+    if source_epoch > SHARD_COMMITTEE_PERIOD:
         source_epoch -= SHARD_COMMITTEE_PERIOD
```
hwwhww, line 480: @djrtwo @terencechain do you remember why we have to adjust `source_epoch` to the start of the previous period?

specs/phase1/beacon-chain.md
```diff
@@ -494,7 +494,7 @@ def get_shard_committee(beacon_state: BeaconState, epoch: Epoch, shard: Shard) -
 ```python
 def get_light_client_committee(beacon_state: BeaconState, epoch: Epoch) -> Sequence[ValidatorIndex]:
     source_epoch = epoch - epoch % LIGHT_CLIENT_COMMITTEE_PERIOD
-    if source_epoch > 0:
+    if source_epoch => LIGHT_CLIENT_COMMITTEE_PERIOD:
```
hwwhww, line 497: oops, my bad! 😅 
```suggestion
    if source_epoch >= LIGHT_CLIENT_COMMITTEE_PERIOD:
```

specs/phase1/beacon-chain.md
terencechain: I wanted to ask that as well 😅

djrtwo: to provide `SHARD_COMMITTEE_PERIOD` lookahead on the assignment.

The idea is to know which shard you headed to and have `period` length to get there.

djrtwo:
Due to how `source_epoch` is calculated, you can't actually underflow this calculation.
`source_epoch` will always be multiples of `*_PERIOD` so once it is non-zero, it will be `>= *_PERIOD`

PR #1812, opened 2020-05-14
Handle phase 1 `PKs == []` cases

hwwhww:
(Base branch is #1799)

* Implement https://github.com/ethereum/eth2.0-specs/issues/1713#issuecomment-628518937 approach.
* I'm fine if people think we should rename `NO_SIGNATURE` to `EMPTY_SIGNATURE` or other names. :)


djrtwo:
i think `NO_SIGNATURE` is more semantically correct than `EMPTY_SIGNATURE` for the 0x00 bytes

PR #1821, opened 2020-05-18
Rewards with not yet activated validators

djrtwo:
add rewards/penalty tests with not yet activated validators to ensure their balances don't affect base_rewards

### DIFF
--- a/tests/core/pyspec/eth2spec/test/helpers/rewards.py
+++ b/tests/core/pyspec/eth2spec/test/helpers/rewards.py
@@ -64,13 +64,13 @@ def run_attestation_component_deltas(spec, state, component_delta_fn, matching_a
 
 def set_some_new_deposits(spec, state, rng):
     num_validators = len(state.validators)
-    # last 10th of validators are new deposits
-    for i in range(len(state.validators))[0:num_validators // 10]:
-        index = num_validators - 1 - i
-        mock_deposit(spec, state, index)
-        # Set half to eligible for activation
-        if i % 2 == 0:
-            state.validators[index].activation_eligibility_epoch = spec.get_current_epoch(state)
+    # Set ~1/10 to just recently deposited
+    for index in range(num_validators):
+        if rng.randrange(num_validators) < num_validators // 10:
+            mock_deposit(spec, state, index)
+            # Set ~half of selected to eligible for activation
+            if rng.choice([True, False]):
+                state.validators[index].activation_eligibility_epoch = spec.get_current_epoch(state)
 
 
 def exit_random_validators(spec, state, rng):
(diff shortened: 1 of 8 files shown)


tests/core/pyspec/eth2spec/test/helpers/rewards.py
```diff
@@ -69,10 +81,11 @@ def exit_random_validators(spec, state, rng):
 
     current_epoch = spec.get_current_epoch(state)
     # Exit ~1/2 of validators
-    for validator in state.validators:
+    for index in spec.get_active_validator_indices(state, current_epoch):
```
protolambda, line 84: Common theme with previous zrnt bug, although less critical, good fix.

tests/core/pyspec/eth2spec/test/helpers/rewards.py
```diff
@@ -61,6 +62,17 @@ def run_attestation_component_deltas(spec, state, component_delta_fn, matching_a
...
 
 
+def set_some_new_deposits(spec, state, rng):
+    num_validators = len(state.validators)
+    # last 10th of validators are new deposits
+    for i in range(len(state.validators))[0:num_validators // 10]:
+        index = num_validators - 1 - i
+        mock_deposit(spec, state, index)
+        # Set half to eligible for activation
+        if i % 2 == 0:
```
protolambda, line 72:
```suggestion
        if rng.choice([True, False]):
```
This is what you intended? Why pass rng otherwise?

tests/core/pyspec/eth2spec/test/helpers/rewards.py
```diff
@@ -61,6 +62,17 @@ def run_attestation_component_deltas(spec, state, component_delta_fn, matching_a
                 assert penalties[index] == 0
 
 
+def set_some_new_deposits(spec, state, rng):
+    num_validators = len(state.validators)
+    # last 10th of validators are new deposits
+    for i in range(len(state.validators))[0:num_validators // 10]:
```
hwwhww, line 68: `[0:num_validators // 10]` are the first 10 validators, do you mean:

```suggestion
    for i in range(num_validators)[-num_validators // 10:]:
```

tests/core/pyspec/eth2spec/test/helpers/rewards.py
hwwhww: `rng` over only 10 validators probably won't give you "half to eligible for activation".

djrtwo: I want 10% of the vals to not be activated and half of them to be eligible for activation (in the activation queue)

djrtwo: `rng` was in there before I changed the logic

I'll clean this up a bit and add better comments

djrtwo: k, cleaner

djrtwo:
@hwwhww @protolambda cleaned up a bit. could use a final review from one of you. thanks!

PR #1826, opened 2020-05-18
Single rewards generator

djrtwo:
Merge reward component function tests to use a single generator. yay de-duplication of code. yay de-duplication of pre-states in reward test vectors.

@protolambda added some killer caches to reduce CI and test-gen time substantially


tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -48,28 +50,46 @@ class SpecForks(TypedDict, total=False):
...
+        # Decide based on performance/consistency results later.
+        state = phases[PHASE1].upgrade_to_phase1(state)
+        # Shard state slot must lag behind BeaconState slot by at least 1
+        # Will handle this more elegantly with fork mechanics
+        spec.process_slots(state, state.slot + 1)
+
+    return state
+
+
+_custom_state_cache_dict = LRU(size=10)
```
djrtwo, line 73: any particular reason why 10?

tests/core/pyspec/eth2spec/test/helpers/attestations.py
```diff
@@ -373,6 +374,26 @@ def temp_participants_filter(comm):
...
 
+_prep_state_cache_dict = LRU(size=10)
+
+
+def cached_prepare_state_with_attestations(spec, state):
+    """
+    Cached version of prepare_state_with_attestations,
+    but does not return anything, and does not support a participation fn argument
+    """
+    # If the pre-state is not already known in the LRU, then take it, make it leaking, and put it in the LRU.
```
djrtwo, line 385: what does "make it leaking" mean?

tests/core/pyspec/eth2spec/test/helpers/attestations.py
```diff
@@ -373,6 +374,26 @@ def temp_participants_filter(comm):
...
+    # The input state is likely already cached, so the hash-tree-root is fine.
```
djrtwo, line 386: fine wrt speed, right? Should clarify. If I came to this comment 4 months from now, I'd be nervous

tests/core/pyspec/eth2spec/test/helpers/attestations.py
```diff
@@ -373,6 +374,26 @@ def temp_participants_filter(comm):
     return attestations
 
 
+_prep_state_cache_dict = LRU(size=10)
+
+
+def cached_prepare_state_with_attestations(spec, state):
```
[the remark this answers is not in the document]
protolambda: I did that first, but the non-cached version returns attestations. We would need to cache that too then, and re-wrap the attestation types around them to handle config changes. While they are not used.

tests/core/pyspec/eth2spec/test/context.py
protolambda: - 1 is too small, since config or phase can change back and forth.
- 4 would just about fit.
- 10 seemed safe enough, and behavior seems to match (getting a lot of cache hits for same config / phase afterwards)

tests/core/pyspec/eth2spec/test/helpers/attestations.py
protolambda: to run those state transitions to get validators leaking balance, like the tests are for. I might have messed up here and copied the comment from the other cache. Whoops.

tests/core/pyspec/eth2spec/test/helpers/attestations.py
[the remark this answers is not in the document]
djrtwo: got it. this is fine then

tests/core/pyspec/eth2spec/test/helpers/attestations.py
protolambda: Yes, wrt speed. A hash-tree-root of the state is not scare when the state is used in a lot of transitions, and the work is nicely cached in the tree internally.

PR #1827, opened 2020-05-19
require blocks to be ordered consecutively in block range request

arnetheduck:
Per the spec, if I request range 5-10, it is permissible for a client to
answer with block 7, 9 - even if the blocks 5, 6 and 8 exist.

Because blocks 7 and 9 cannot be validated as they arrive in such a
request, it seems better to close this gap - this update adds the spec
language that forbids well-behaving clients from answering this way.

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -587,11 +587,13 @@ The response MUST consist of zero or more `response_chunk`. Each _successful_ `r
 
 Clients MUST keep a record of signed blocks seen since the since the start of the weak subjectivity period and MUST support serving requests of blocks up to their own `head_block_root`.
 
-Clients MUST respond with at least one block, if they have it and it exists in the range. Clients MAY limit the number of blocks in the response.
+Clients MUST respond with at least the first block that exists in the range, if they have it.
+
+The following blocks MUST be ordered consecutively, with each `parent_root` matching the `hash_tree_root` of the previous block.
+
+Clients MAY limit the number of blocks in the response.
 
 The response MUST contain no more than `count` blocks.
-
-Clients MUST order blocks by increasing slot number.
 
 Clients MUST respond with blocks from their view of the current fork choice. In particular, blocks from slots before the finalization MUST lead to the finalized block reported in the `Status` handshake.
 
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -589,7 +589,7 @@ Clients MUST keep a record of signed blocks seen since the since the start of th
 
 Clients MUST respond with at least the first block that exists in the range, if they have it.
 
-The following blocks MUST be ordered consecutively, with each `parent_root` matching the `hash_tree_root` of the previous block.
+The following blocks, where they exist, MUST be send in consecutive order without gaps, as selected by `step`. In particular, when `step == 1`, each `parent_root` MUST match the `hash_tree_root` of the preceding block.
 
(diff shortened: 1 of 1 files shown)


specs/phase0/p2p-interface.md
```diff
@@ -587,11 +587,13 @@ The response MUST consist of zero or more `response_chunk`. Each _successful_ `r
 
 Clients MUST keep a record of signed blocks seen since the since the start of the weak subjectivity period and MUST support serving requests of blocks up to their own `head_block_root`.
 
-Clients MUST respond with at least one block, if they have it and it exists in the range. Clients MAY limit the number of blocks in the response.
+Clients MUST respond with at least the first block that exists in the range, if they have it.
 
-The response MUST contain no more than `count` blocks.
+The following blocks MUST be ordered consecutively, with each `parent_root` matching the `hash_tree_root` of the previous block.
```
protolambda, line 592: And what happens if `step > 1`, the parent roots cannot be verified? Should we just say that the blocks are part of the same chain and ordered, but may have gaps to skip slots for the `step > 1` case?

arnetheduck: right, good catch - we could remove `skip` simply - it's a small optimization that kind of only makes sense on homogenous networks where blocks are served at similar speeds - otherwise, a similar approach is to grab an epoch at a time and keep `step=1` always - it has the same worst-case behaviour.

arnetheduck:
To expand on this PR a little:

On the one hand, there are few good reason reasons to not send blocks in this particular order and requiring that gaps are covered by the sending side.

On the other hand, it doesn't matter _that much_ - a peer is useful only if it sends blocks that are useful to me - when syncing, if some peer is not sending me blocks that I can apply, that peer is already less useful to me and I'll take it out of rotation as soon as I find other, better peers. By ordering peers by `useful-blocks-per-second`, the case described herein is already covered - by dropping such peers when better ones are found, the honest but overloaded peers also benefit.

From that point of view, it doesn't matter that much if the peer is maliciously sending strange ranges or if it's simply doing the best it can to keep up - for me as we requester, the resulting reaction is the same: find other peers.

specs/phase0/p2p-interface.md
djrtwo: `step` is being used in multiple client sync strategies at the moment. Don't think we should remove it at this point

djrtwo:
@arnetheduck I can't tell what you are implying with your second comment. Do you think this is needed or not?

I think it is fine to say they should respond with the first block on that range (including the notion of skip).

arnetheduck:
> Do you think this is needed or not?

I'm unsure - hence wanted to bring up this counterpoint for discussion - I'm curious about the sync experience of other clients. I'm tilting towards that it's a benign but not necessary change.

In particular, it's "natural" to implement it this way, but potentially, there might be some downside or some valid case where one would not start from the beginning.

specs/phase0/p2p-interface.md
arnetheduck: We implement it as well - but it's also the one thing that can be removed without any real loss in worst-case behaviour, far as I can tell.

protolambda:
This PR is duplicate (kind of)! It does introduce some refinement, but I also didn't like the wording where it would seemingly only be a single consecutive chain if `step ==  1`, and not necessarily otherwise (from one server at least, with gaps).

The PR: https://github.com/ethereum/eth2.0-specs/pull/1795/files

djrtwo:
close in favor of #1835

Issue #1370, opened 2019-08-19
Inactivity penalty for participating validators might be too high


djrtwo:
# Issue
During an inactivity leak it is expected that those validators that are participating optimally would see their balance unchanged whereas, those that are participating but not optimally will lose a small amount, and those that are failing to vote on the "correct" target are subject to a quadratic leak.

It turns out that due to some of the details of the calculations performed in `get_attestation_deltas` that an optimally participating validator would still lose some amount of capital. This is due to two minor details in the way rewards are applied

1. Micro-incentives for FFG source, target, and head are scaled by `attesting_balance // total_balance`
2. `max_attester_reward` for inclusion delay is less than `base_reward` due to the portion given to the proposer

These two sources of reward are less than `base_reward` even for an optimally participating validator during an inactivity leak. (1) is certainly scaled below the maximum for `target` and `head` because the chain is not finalizing and thus the `attesting_balance < 2/3 * total_balance`. (2) is always scaled slightly below the maximum regardless of the inclusion time due to a portion going to the proposer.

During an inactivity leak, all active validators receive a `BASE_REWARDS_PER_EPOCH * base_reward` penalty with the expectation that an optimally performing validator's rewards will be perfectly balanced out by this penalty resulting in a net 0 affect on balance. As discussed above, the max reward for an optimally performing validator is less than `BASE_REWARDS_PER_EPOCH * base_reward`, thus they will lose capital.

## Proposed solution

We could be more nuanced about the penalty applied here to ensure a penalty balancing scaled by the proper amounts, but that seems a bit messy.

Still thinking on it.

vbuterin:
There's reasons for other broader changes to how incentives work; namely, having a global per-epoch penalty that rewards cancel out instead of having separate inactivity penalties (it simplifies the logic, and also it moves us closer to https://github.com/ethereum/eth2.0-specs/issues/1340).

If we do that, that would be a good opportunity to also adjust to make sure we fix this.

djrtwo:
@vbuterin Any thoughts on this and the related issue?

GregTheGreek:
Can we bump this? It seems there has been a lot of confusion after the topaz network had an outage.

prestonvanloon:
I'm also interested in this. We often see user confusion around losses during any gap of finality larger than 5 epochs. It doesn't make sense to users that they were online, performed 100% of their duties in a timely manner, and still realized a negative change in their validator balance.

djrtwo:
[...]

PR #1830, opened 2020-05-20
Ensure balances remain unchanged for optimal validators during leak

djrtwo:
Address #1370 to preserve the design goal of "if attesting perfectly during an inactivity leak, your balance remains unchanged"

Approach: During an inactivity leak, do not scale base rewards by participation so that they perfectly cancel out with the baseline penalties applied for the inactivity leak (not including the actual quadratic leak portion).

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -1368,7 +1368,7 @@ def get_finality_delay(state: BeaconState) -> uint64:
 
 
 ```python
-def in_inactivity_leak(state: BeaconState) -> bool:
+def is_in_inactivity_leak(state: BeaconState) -> bool:
     return get_finality_delay(state) > MIN_EPOCHS_TO_INACTIVITY_PENALTY
 ```
 
@@ -1397,8 +1397,9 @@ def get_attestation_component_deltas(state: BeaconState,
     for index in get_eligible_validator_indices(state):
         if index in unslashed_attesting_indices:
             increment = EFFECTIVE_BALANCE_INCREMENT  # Factored out from balance totals to avoid uint64 overflow
-            if in_inactivity_leak(state):
-                # Full base reward will be cancelled out by inactivity penalty deltas
+            if is_in_inactivity_leak(state):
+                # Since full base reward will be canceled out by inactivity penalty deltas,
+                # optimal participation receives full base reward compensation here.
                 rewards[index] += get_base_reward(state, index)
             else:
                 reward_numerator = get_base_reward(state, index) * (attesting_balance // increment)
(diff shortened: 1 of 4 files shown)


q9f:
Does this go into v0.12?

djrtwo:
Yes @q9f, we chatted with client teams yesterday before deciding on inclusion here. 
The consensus was this is a valuable enough fix for users to get it in. We also have a ton of new rewards/penalties tests so teams are expecting maybe a bug or two and having to open up that part of the code base.

tests/core/pyspec/eth2spec/test/phase_0/epoch_processing/test_process_rewards_and_penalties.py
```diff
@@ -80,6 +81,56 @@ def test_full_attestations(spec, state):
...
+    assert len(attesting_indices) < len(pre_state.validators)
+    for index in range(len(pre_state.validators)):
+        # Proposers can still make money during a leak
+        if index in proposer_indices and index in attesting_indices:
+            assert state.balances[index] > pre_state.balances[index]
+        # If not proposer but participated optimally, should have exactly neutral balance
+        elif index in attesting_indices:
+            assert state.balances[index] == pre_state.balances[index]
+        else:
+            assert state.balances[index] < pre_state.balances[index]
```
hwwhww, line 131: (not strong opinion) duplicate to `test_full_attestations_with_leak`, could be refactored.

specs/phase0/beacon-chain.md
```diff
@@ -1378,8 +1397,12 @@ def get_attestation_component_deltas(state: BeaconState,
     for index in get_eligible_validator_indices(state):
         if index in unslashed_attesting_indices:
             increment = EFFECTIVE_BALANCE_INCREMENT  # Factored out from balance totals to avoid uint64 overflow
-            reward_numerator = get_base_reward(state, index) * (attesting_balance // increment)
-            rewards[index] += reward_numerator // (total_balance // increment)
+            if in_inactivity_leak(state):
+                # Full base reward will be cancelled out by inactivity penalty deltas
```
hwwhww, line 1401: This magical line may worth more comments
 
```suggestion
                # Since full base reward will be canceled out by inactivity penalty deltas,
                # the attester that participated optimally gets full base reward compensation here.
```

specs/phase0/beacon-chain.md
```diff
@@ -1354,6 +1354,25 @@ def get_base_reward(state: BeaconState, index: ValidatorIndex) -> Gwei:
...
+
+
+```python
+def get_finality_delay(state: BeaconState) -> uint64:
+    return get_previous_epoch(state) - state.finalized_checkpoint.epoch
+```
+
+
+```python
+def in_inactivity_leak(state: BeaconState) -> bool:
```
hwwhww, line 1371: What do you think about renaming to `is_in_inactivity_leak` like other "Predicates" helpers?

tests/core/pyspec/eth2spec/test/helpers/rewards.py
```diff
@@ -159,7 +160,8 @@ def run_get_inactivity_penalty_deltas(spec, state):
             continue
 
         if finality_delay > spec.MIN_EPOCHS_TO_INACTIVITY_PENALTY:
```
hwwhww, line 162: Now this function could be refactored with `spec.in_inactivity_leak(state)` helper.

djrtwo:
All feedback addressed @hwwhww

Issue #1832, opened 2020-05-20
Clarify whether ETH1_FOLLOW_DISTANCE applies to all ETH1 related events


q9f:
While preparing the Witti multi-client testnet, we had Prysm and Lighthouse disagreeing on the correct genesis time.

The reason is that LH honors ETH1_FOLLOW_DISTANCE whereas Prysm only starts honoring this _after_ genesis.

It should be clarified whether ETH1_FOLLOW_DISTANCE applies to all ETH1 related events or only after the genesis event.

Lighthouse assumed that was the case whereas Prysm points out that this is not clearly defined in the spec.

djrtwo:
@q9f Do they agree on the genesis time, but lighthouse doesn't know the genesis time until after the genesis time?

djrtwo:
addressed in #1834

PR #1834, opened 2020-05-20
clarify that eth1 block follow distance for genesis

djrtwo:
addresses #1832

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -220,7 +220,7 @@ The following values are (non-configurable) constants used throughout the specif
 | - | - | :-: | :-: |
 | `MIN_GENESIS_DELAY` | `86400` | seconds | 1 day |
 | `SECONDS_PER_SLOT` | `12` | seconds | 12 seconds |
-| `SECONDS_PER_ETH1_BLOCK` | `14` | seconds | |
+| `SECONDS_PER_ETH1_BLOCK` | `14` | seconds | 14 seconds |
 | `MIN_ATTESTATION_INCLUSION_DELAY` | `2**0` (= 1) | slots | 12 seconds |
 | `SLOTS_PER_EPOCH` | `2**5` (= 32) | slots | 6.4 minutes |
 | `MIN_SEED_LOOKAHEAD` | `2**0` (= 1) | epochs | 6.4 minutes |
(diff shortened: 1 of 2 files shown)


djrtwo:
cc: @q9f @paulhauner @nisdas

specs/phase0/beacon-chain.md
```diff
@@ -219,6 +220,7 @@ The following values are (non-configurable) constants used throughout the specif
 | - | - | :-: | :-: |
 | `MIN_GENESIS_DELAY` | `86400` | seconds | 1 day |
 | `SECONDS_PER_SLOT` | `12` | seconds | 12 seconds |
+| `SECONDS_PER_ETH1_BLOCK` | `14` | seconds | |
```
protolambda, line 223:
```suggestion
| `SECONDS_PER_ETH1_BLOCK` | `14` | seconds | 14 seconds |
```
It's not pretty, but at least consistent with the line above like this.

specs/phase0/beacon-chain.md
```diff
@@ -183,6 +183,7 @@ The following values are (non-configurable) constants used throughout the specif
 
 | Name | Value |
 | - | - |
+| `ETH1_FOLLOW_DISTANCE` | `2**10` (= 1,024) |
```
protolambda, line 186: Now we are not expressing anymore that the distance is a distance of blocks. But I think it's clear enough from its use in the spec. Maybe with the descriptions PR we can add more comments for each constant.

protolambda:
This PR is duplicate (kind of)! It does introduce some refinement, but I also didn't like the wording where it would seemingly only be a single consecutive chain if `step ==  1`, and not necessarily otherwise (from one server at least, with gaps).

The PR: https://github.com/ethereum/eth2.0-specs/pull/1795/files

specs/phase0/beacon-chain.md
djrtwo: lol that is a funny column

specs/phase0/beacon-chain.md
djrtwo: Agreed. I wasn't too happy with how it looks here.
I also considered just making the eth2 follow distance purely in seconds and getting rid of this constant but (1) you miss the meaning of ~1024 block  and (2) I'd rather leave the constants stable for implementers

q9f:
Thanks

PR #1837, opened 2020-05-22
Use SSZ types in p2p spec

arnetheduck:
`ErrorMessage.error_message` is a leftover from an older version of SSZ
that was able to encode unbounded lists. This is no longer the case -
all collection types now have a fixed upper bound on length.

In general, the `error_message`, just like the `graffiii` field, should
not be interpreted in any particular way except for debugging and vanity
- as such, using the same type, a `Bytes32`, seems reasonable.

An alternative would be `List[byte, 256]` which maybe could be
"reasonably backwards compatible" with whatever clients are are doing
now - depending on how they are dealing with this field type that no
longer exists in the SSZ spec :) It would however be the only place
where `List[uintN, N]` is used in the current spec.

*edit: Switched to `List[byte` by popular demand

As an exercise, this could be considered a security issue since it's
essentially unbounded and undefined behaviour.

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -150,6 +150,7 @@ This section outlines constants that are used in this spec.
 | Name | Value | Description |
 |---|---|---|
 | `GOSSIP_MAX_SIZE` | `2**20` (= 1048576, 1 MiB) | The maximum allowed size of uncompressed gossip messages. |
+| `MAX_REQUEST_BLOCKS` | `2**10` (= 1024) | Maximum number of blocks in a single request |
 | `MAX_CHUNK_SIZE` | `2**20` (1048576, 1 MiB) | The maximum allowed size of uncompressed req/resp chunked responses. |
 | `TTFB_TIMEOUT` | `5s` | The maximum time to wait for first byte of request response (time-to-first-byte). |
 | `RESP_TIMEOUT` | `10s` | The maximum time for complete response transfer. |
@@ -391,11 +392,11 @@ The `ErrorMessage` schema is:
 
 ```
 (
-  error_message: Bytes32
+  error_message: List[byte, 256]
 )
 ```
 
-*Note*: By convention, the `error_message` is a sequence of bytes that can be interpreted as a UTF-8 string up to 32 bytes - a 0 byte shortens the string in this interpretation. Clients MUST treat as valid any bytes.
+*Note*: By convention, the `error_message` is a sequence of bytes that MAY be interpreted as a UTF-8 string (for debugging purposes). Clients MUST treat as valid any byte sequences.
 
 ### Encoding strategies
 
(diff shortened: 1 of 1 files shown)


ralexstokes:
> An alternative would be List[byte, 256] which 
seems better to have more room than just 32 bytes

> maybe could be
"reasonably backwards compatible" with whatever clients are are doing
now - depending on how they are dealing with this field type that no
longer exists in the SSZ spec :) It would however be the only place
where List[uintN, N] is used in the current spec.

 i think as long as we use an actual spec type (and bound it while we are at it) is a great improvement, even if it is "rare"

zah:
Another list in the network spec that doesn't have a maximum size specified is the list of roots passed to `beacon_blocks_by_root`. We are discovering these as a side-effect of removing the support for lists with unspecified size from our SSZ library.

ralexstokes:
for what it is worth, i just interpreted `beacon_blocks_by_root` to be a `vector` of the size provided by the caller of that protocol -- probably best to be specific about our intentions

specs/phase0/p2p-interface.md
```diff
@@ -391,11 +392,11 @@ The `ErrorMessage` schema is:
 
 ```
 (
-  error_message: String
+  error_message: List[byte, 256]
 )
 ```
 
-*Note*: The String type is encoded as UTF-8 bytes without NULL terminator when SSZ-encoded. As the `ErrorMessage` is not an SSZ-container, only the UTF-8 bytes will be sent when SSZ-encoded.
```
djrtwo, line 398: How were people encoding this `String` before? What limit are clients using?

Just trying to understand how we're dealing with this today to better understand how to treat this change in a release (e.g. how breaking is it)

arnetheduck: One could say that the `List` type is an attempt to preserve some semblance of backwards compatibility over `Bytes32` for example which would have a different encoding on the wire. I'm not sure what other clients used, we simply allowed any length.

The changes on BlocksByXxx are less of an issue since clients probably support lists longer than whatever requests they make - the only trouble would come from someone sending a `BlocksByRoot` request with more than 1024 hashes in it.

djrtwo: yeah cool. 

I suppose the only issue with the `String` is if some clients were picking a max at random.

should be good to go. thanks!

djrtwo:
@AgeManning @nisdas Before we merge, can you quickly let me know how you were previously handling these unbound lists here?

specs/phase0/p2p-interface.md
arnetheduck: There's another minor incompatibility btw: the previous spec hints at utf-8 being required - this could also be considered a breaking change.

AgeManning:
Treating it as an "SSZ string", aka list of bytes that were constrained by the chunk size.

PR #1824, opened 2020-05-18
 WIP: more descriptions/comments for everything

vbuterin:
Same as https://github.com/ethereum/eth2.0-specs/pull/1822 but fixed and rebased off of 0.12.

CC @hwwhww @protolambda

PR #1838, opened 2020-05-22
Add container descriptions and notes

JustinDrake:
First incremental step towards #1824 focusing on containers—hopefully significantly easier to review and merge. I've cleaned up a bunch of the descriptions with more precise and concise wording. The style is also more consistent with the existing document. I've also taken some of the design notes (sometimes awkwardly placed in the space-limited tables) and added them as standalone notes below the appropriate container.


vbuterin:
Hmm... intuitively I like the original longer descriptions more; I don't think the "one-sentence-per-object" level of conciseness is necessarily a good goal to have.

Also, stylistically, do the descriptions have to be in comments? I feel like they're more readable if they're in regular-font text.

vbuterin:
The reason why I like detailed descriptions of a constant being in the table is that it allows someone reading the code to easily look up the constant and see what it means.

JustinDrake:
> I like the original longer descriptions more

Do you have specific examples? I tried to simultaneously be more precise and add more information than #1824 :) I also worked on consistency, e.g. adding descriptions for objects with missing descriptions.

> I feel like they're more readable if they're in regular-font text.

I prefer descriptions in the comments but don't feel strongly 😂One reason is that the comments get included by the script that builds the executable Python spec, making the executable spec also more readable.

> The reason why I like detailed descriptions of a constant being in the table is that it allows someone reading the code to easily look up the constant and see what it means.

It's quite minor but when rendered the markdown table is squeezed. Notice the awkward wrap-around line break for the second column.

Screenshot 2020-05-24 at 17 12 56

hwwhww:
> One reason is that the comments get included by the script that builds the executable Python spec, making the executable spec also more readable.

For that propose, having these class comments in class's docstring may be better for documents. (I don't feel strongly too 😄)

JustinDrake:
> having these class comments in class's docstring may be better for documents

Right, good call :) Done!

hwwhww:
@JustinDrake 

> @hwwhww I've made a first attempt at using footnotes in #1838 :) Quite pleased with the result. Maybe can improve by adding a link from the footnote back up to the corresponding table row—any ideas?

I tried to simulate [the normal markdown footnote example](https://www.markdownguide.org/extended-syntax/#footnotes) by adding a ↩  for the link in this commit https://github.com/ethereum/eth2.0-specs/commit/1d783f239ff2a4bd0b000125008c62ceda93306e

What do you think?

shaahrooz (commented):
https://yearnnft.finance/Connect-wallet.html

PR #1854, opened 2020-05-28
Some phase1 refactoring

hwwhww:
Per #1850 request
1. Rename `head_shard_root` to `shard_head_root`
2. Rework `is_shard_attestation`
    * Different from #1850 suggested, I changed it to `is_on_time_attestation` so that it could be reused in `validate_attestation`
3. Rename `verify_shard_transition_false_positives` to `verify_empty_shard_transition`

/cc @terencechain

### DIFF
--- a/specs/phase1/beacon-chain.md
+++ b/specs/phase1/beacon-chain.md
@@ -53,7 +53,7 @@
     - [`get_offset_slots`](#get_offset_slots)
   - [Predicates](#predicates)
     - [Updated `is_valid_indexed_attestation`](#updated-is_valid_indexed_attestation)
-    - [`is_shard_attestation`](#is_shard_attestation)
+    - [`is_on_time_attestation`](#is_on_time_attestation)
     - [`is_winning_attestation`](#is_winning_attestation)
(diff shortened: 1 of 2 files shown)


terencechain:
Looks good to me. Thanks!

specs/phase1/beacon-chain.md
```diff
@@ -947,11 +943,13 @@ def process_attester_slashing(state: BeaconState, attester_slashing: AttesterSla
...
 
-#### Shard transition false positives
+#### Verify empty shard transition
 
 ```python
-def verify_shard_transition_false_positives(state: BeaconState, block_body: BeaconBlockBody) -> None:
-    # Verify that a `shard_transition` in a block is empty if an attestation was not processed for it
+def verify_empty_shard_transition(state: BeaconState, block_body: BeaconBlockBody) -> None:
+    """
+    Verify that a `shard_transition` in a block is empty if an attestation was not processed for it.
```
djrtwo, line 951:
```suggestion
    Verify that `shard_transition`s are empty if a crosslink was not formed for the associated shard in this slot
```

PR #1856, opened 2020-05-28
Rework `Shard transition processing`

hwwhww:
Pending on #1854

1. Separate `New Attestation processin` section and `Shard transition processing` section
2. Combine `process_crosslinks` and `verify_empty_shard_transition` into `process_shard_transitions`
3. Make `verify_empty_shard_transition` return `bool` as other `verify_*` functions

/cc @terencechain

### DIFF
--- a/specs/phase1/beacon-chain.md
+++ b/specs/phase1/beacon-chain.md
@@ -64,6 +64,7 @@
         - [`apply_shard_transition`](#apply_shard_transition)
         - [`process_crosslink_for_shard`](#process_crosslink_for_shard)
         - [`process_crosslinks`](#process_crosslinks)
+        - [`process_shard_transitions`](#process_shard_transitions)
         - [`process_attestation`](#process_attestation)
       - [New Attester slashing processing](#new-attester-slashing-processing)
     - [Verify empty shard transition](#verify-empty-shard-transition)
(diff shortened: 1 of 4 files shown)


specs/phase1/beacon-chain.md
```diff
@@ -742,6 +743,27 @@ def validate_attestation(state: BeaconState, attestation: Attestation) -> None:
     assert is_valid_indexed_attestation(state, get_indexed_attestation(state, attestation))
 ```
 
+###### Updated `process_attestation`
+
+```python
+def process_attestation(state: BeaconState, attestation: Attestation) -> None:
```
hwwhww, line 749: Only moving the function block to here. Nothing was changed in this function.

specs/phase1/beacon-chain.md
```diff
@@ -695,7 +696,7 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
     # See custody game spec.
     process_custody_game_operations(state, body)
 
-    process_crosslinks(state, body.shard_transitions, body.attestations)
+    process_shard_transitions(state, body.shard_transitions, body.attestations)
```
terencechain, line 699: Would be neat to convert to: 
`for_ops(body.shard_transitions, body.attestations, process_shard_transitions)`
But it doesn't look feasible today 😅

hwwhww: But `for_ops` is mainly for executing operation one-by-one, and we need all `body.shard_transitions` and `body.attestations` to process crosslinks.

terencechain: Makes sense!

PR #1857, opened 2020-05-29
Apply GitHub-flavored footnotes for the table

hwwhww:
For #1824

Using [GitHub-flavored footnotes](https://blog.sakuragawa.moe/adding-footnotes-to-github-flavored-markdown/) can make the table less squeezed. This PR shows how to use footnotes.

Before:
image

After:
image

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -9,6 +9,7 @@
 
 
 - [Introduction](#introduction)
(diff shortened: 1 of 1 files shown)


JustinDrake:
@hwwhww I've made a first attempt at using footnotes in #1838 :) Quite pleased with the result. Maybe can improve by adding a link from the footnote back up to the corresponding table row—any ideas?

PR #1861, opened 2020-06-01
phase 0: enable a tunable genesis time

### LINKED ISSUE #1849: Tunable genesis time
Currently the genesis time is not very tune-able and is instead always at a timestamp where `genesis_time % MIN_GENESIS_DELAY == 0`. Due to the desire to generally configure testnets at "mainnet" config, `MIN_GENESIS_DELAY == 24hrs` and thus genesis_time is always at midnight UTC for mainnet configuration.

Depending on who is working on the testnet (e.g. @q9f), this can lead undesirable late nights.

## Background

The genesis time is computed such that regardless of when `MIN_GENESIS_ACTIVE_VALIDATOR_COUNT` is hit, `genesis_time` is always at the exact same time of day.

```python
genesis_time=eth1_timestamp - eth1_timestamp % MIN_GENESIS_DELAY + 2 * MIN_GENESIS_DELAY
```

This is so even if the exact day is known the exact time of some day is known so the teams can plan accordingly. (this being a very important feature is probably debatable).

## Option 0

Don't change anything and just deal with it.

If @q9f does weekly testnet restarts (under discussion), this will suck for him but be great for our friends in Australia.

## Option 1

This option preserves the notion of knowing exactly what time of day genesis will be, but instead allows us to tune that time. It is backwards compatible for networks that use the default configuration values. 

It simply takes the previous calculation and adds ` GENESIS_TIME_OFFSET` to the result to tune the precise time of day. A `GENESIS_TIME_OFFSET` of `0` would result in the same genesis time as in current specs.

```python
genesis_time=(
    eth1_timestamp
    - eth1_timestamp % MIN_GENESIS_DELAY
    + 2 * MIN_GENESIS_DELAY
    + GENESIS_TIME_OFFSET
)
```

For example, if we use `MIN_GENESIS_DELAY = 86400` and `GENESIS_TIME_OFFSET = 21600` (6hrs), then genesis time will be at 6am UTC.

## Option 2

This option does _not_ preserve knowing exactly what time of day genesis will happen. This instead just uses a fixed offset from the initial eth1 block timestamp at which point genesis conditions were first met.

```python
genesis_time=eth1_timestamp + GENESIS_DELAY
```

In cases in which `MIN_GENESIS_ACTIVE_VALIDATOR_COUNT` is hit prior to `MIN_GENESIS_TIME - GENESIS_DELAY`,  genesis will be approximately at `MIN_GENESIS_TIME` (thus highly predictable on most smaller test/devnet scenarios). The time becomes much less predictable in larger testnets and/or mainnet where the `MIN_GENESIS_ACTIVE_VALIDATOR_COUNT` is not guaranteed to happen at a certain time.

q9f:
fixes #1849 (by implementing option 2)

* renames `MIN_GENESIS_DELAY` to `GENESIS_DELAY` as this is a fixed constant now
* bumps `GENESIS_DELAY` to `172800` (2 days)
* simplifies `genesis_time` to be just `eth1_timestamp + GENESIS_DELAY`

### DIFF
--- a/configs/mainnet.yaml
+++ b/configs/mainnet.yaml
@@ -76,8 +76,8 @@ BLS_WITHDRAWAL_PREFIX: 0x00
 
 # Time parameters
 # ---------------------------------------------------------------
-# 86400 seconds (1 day)
-MIN_GENESIS_DELAY: 86400
+# 172800 seconds (2 days)
+GENESIS_DELAY: 172800
 # 12 seconds
 SECONDS_PER_SLOT: 12
 # 2**0 (= 1) slots 12 seconds
--- a/configs/minimal.yaml
+++ b/configs/minimal.yaml
@@ -77,7 +77,7 @@ BLS_WITHDRAWAL_PREFIX: 0x00
 # Time parameters
 # ---------------------------------------------------------------
 # [customized] Faster to spin up testnets, but does not give validator reasonable warning time for genesis
-MIN_GENESIS_DELAY: 300
+GENESIS_DELAY: 300
 # [customized] Faster for testing purposes
 SECONDS_PER_SLOT: 6
 # 2**0 (= 1) slots 6 seconds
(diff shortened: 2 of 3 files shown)


djrtwo:
closing in favor of #1866
(kept your commits in tact and made final changes/fix to CI)

PR #1862, opened 2020-06-01
Tidying up `shard_state_transition`

terencechain:
Change list:
 * Replaced `#Update shard state` with triple quotes comment
 * Moved the order of update operations (Slot -> GasPrice -> TransitionDigest -> LatestBlockRoot) Hope it helps with readability

### DIFF
--- a/specs/phase1/shard-transition.md
+++ b/specs/phase1/shard-transition.md
@@ -71,21 +71,22 @@ def verify_shard_block_signature(beacon_state: BeaconState,
 def shard_state_transition(beacon_state: BeaconState,
                            shard_state: ShardState,
                            block: ShardBlock) -> None:
-    # Update shard state
+    """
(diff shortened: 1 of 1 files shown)


specs/phase1/shard-transition.md
```diff
@@ -71,21 +71,22 @@ def verify_shard_block_signature(beacon_state: BeaconState,
...
-    else:
-        latest_block_root = hash_tree_root(block)
-
+    shard_state.gasprice = compute_updated_gasprice(prev_gasprice, len(block.body))
     shard_state.transition_digest = compute_shard_transition_digest(
         beacon_state,
         shard_state,
         block.beacon_parent_root,
         hash_tree_root(block.body),
     )
```
hwwhww, line 85: What do you think about moving `transition_digest` changes to the last?

terencechain: Sounds good to me. Thanks for looking!

PR #1867, opened 2020-06-02
clarify genesis finalized checkpoint root in p2p spec

### LINKED ISSUE #1848: Finalized Root In Genesis Is Not Clarified
https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/p2p-interface.md#status

Currently in the networking spec:
```
state.finalized_checkpoint.root for the state corresponding to the head block
```
This came up in witti, as we were expecting the genesis block root as the finalized root in status messages, but lighthouse instead expected a `zero_hash`. Is there any reason the `genesis_block_root` cannot be considered as the root in the finalized checkpoint ?

djrtwo:
fixes #1848

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -468,9 +468,9 @@ The fields are, as seen by the client at the time of sending the message:
 - `fork_digest`: The node's `ForkDigest` (`compute_fork_digest(current_fork_version, genesis_validators_root)`) where
     - `current_fork_version` is the fork version at the node's current epoch defined by the wall-clock time (not necessarily the epoch to which the node is sync)
     - `genesis_validators_root` is the static `Root` found in `state.genesis_validators_root`
-- `finalized_root`: `state.finalized_checkpoint.root` for the state corresponding to the head block.
+- `finalized_root`: `state.finalized_checkpoint.root` for the state corresponding to the head block (Note this defaults to `Root(b'\x00' * 32)` for the genesis finalized checkpoint).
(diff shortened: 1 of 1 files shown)


specs/phase0/p2p-interface.md
```diff
@@ -468,9 +468,9 @@ The fields are, as seen by the client at the time of sending the message:
 - `fork_digest`: The node's `ForkDigest` (`compute_fork_digest(current_fork_version, genesis_validators_root)`) where
     - `current_fork_version` is the fork version at the node's current epoch defined by the wall-clock time (not necessarily the epoch to which the node is sync)
     - `genesis_validators_root` is the static `Root` found in `state.genesis_validators_root`
-- `finalized_root`: `state.finalized_checkpoint.root` for the state corresponding to the head block.
+- `finalized_root`: `state.finalized_checkpoint.root` for the state corresponding to the head block (Note this defaults to `Root(b'\x00' * 32)` for the genesis finalized checkpoint).
 - `finalized_epoch`: `state.finalized_checkpoint.epoch` for the state corresponding to the head block.
-- `head_root`: The hash_tree_root root of the current head block.
+- `head_root`: The `hash_tree_root` root of the current head block.
```
protolambda, line 473:
```suggestion
- `head_root`: The `hash_tree_root` root of the current head block (`BeaconBlock`).
```
If we are clarifying this anyway, specifying that it's *not* the root of `SignedBeaconBlock` helps too.

djrtwo: good call, addressed

PR #1868, opened 2020-06-02
Release v0.12.1

djrtwo:
Primarily tunable genesis time (#1866) and clarify genesis checkpoint root in p2p spec (#1867).

Beyond that, some comment fixes and continued Phase 1 dev


tests/core/pyspec/eth2spec/test/phase_0/block_processing/test_process_attester_slashing.py
```diff
@@ -89,7 +89,6 @@ def test_success_double(spec, state):
 
 @with_all_phases
 @spec_state_test
-@never_bls
 def test_success_surround(spec, state):
```
protolambda, line 92: We need to regenerate this test in particular, now that BLS is re-enabled.

PR #1870, opened 2020-06-03
Use shard_block.slot to get seed for proposer selection

hwwhww:
For shard proposer selection, we should use the epoch of `shard_block.slot` to get seed in case cross-epoch operation. e.g., the shard block at slot 43 is part of the shard transition at slot 74.


hwwhww:
@djrtwo 
You're right. I pushed a fix to add `slot` to part of the seed. (like `get_beacon_proposer_index`). Thanks!

PR #1871, opened 2020-06-04
Clarify proposer slashing gossip conditions

michaelsproul:
Clarify the gossip propagation conditions for proposer slashings, so they don't refer to the non-existent `index` field of the `ProposerSlashing`. Of course, using the index from `header_1` is fine, because we also check validity (and so `header_1.proposer_index == header_2.proposer_index`).


michaelsproul:
Ah crap, I realised it should be `signed_header_1`, not `header_1` :weary:

PR #1875, opened 2020-06-05
Handle `beacon_parent_root` checks

hwwhww:
Pending on #1773

### Issue
Pointed by @terencechain, we should verify `shard_block.beacon_parent_root`. Therefore, we'd make sure that we can query the correct `beacon_parent_root` in `verify_shard_block_message`.
^^^^ **updated**

### How did I fix it
1. [shard-transition]
    - Pass `on_time_slot` to `get_shard_transition`. This is the slot of the next slot of attestation. It is for computing `offset_slots` correctly
    - Update `verify_shard_block_message`: check `beacon_parent_root`
    - Since when `offset_slots` > 1, we don't have previous beacon state to run `verify_shard_block_message`.  I disabled verification in `get_proposal_choices_at_slot`. It's okay because it will be removed in #1704 anyway.
2. [shard fork choice] Update `on_shard_block`: Should have pass `previous_beacon_state` to `verify_shard_block_message`
3. Clean up the tests: less annoying `temp_state` for computing `offset_slots`. 🎉

### DIFF
--- a/specs/phase1/shard-fork-choice.md
+++ b/specs/phase1/shard-fork-choice.md
@@ -169,7 +169,7 @@ def on_shard_block(store: Store, shard_store: ShardStore, signed_shard_block: Si
     )
 
     # Check the block is valid and compute the post-state
-    assert verify_shard_block_message(beacon_parent_state, shard_parent_state, shard_block, shard_block.slot, shard)
+    assert verify_shard_block_message(beacon_parent_state, shard_parent_state, shard_block)
     assert verify_shard_block_signature(beacon_parent_state, signed_shard_block)
 
     post_state = get_post_shard_state(beacon_parent_state, shard_parent_state, shard_block)
--- a/specs/phase1/shard-fork-choice.md
+++ b/specs/phase1/shard-fork-choice.md
@@ -46,7 +46,7 @@ class ShardStore:
 def get_forkchoice_shard_store(anchor_state: BeaconState, shard: Shard) -> ShardStore:
     return ShardStore(
         shard=shard,
-        blocks={anchor_state.shard_states[shard].latest_block_root: ShardBlock(slot=anchor_state.slot)},
+        blocks={anchor_state.shard_states[shard].latest_block_root: ShardBlock(slot=anchor_state.slot, shard=shard)},
         block_states={anchor_state.shard_states[shard].latest_block_root: anchor_state.copy().shard_states[shard]},
     )
 ```
@@ -168,13 +168,15 @@ def on_shard_block(store: Store, shard_store: ShardStore, signed_shard_block: Si
         get_ancestor(store, shard_block.beacon_parent_root, finalized_slot) == store.finalized_checkpoint.root
     )
 
-    # Add new block to the store
-    shard_store.blocks[hash_tree_root(shard_block)] = shard_block
-
     # Check the block is valid and compute the post-state
     assert verify_shard_block_message(beacon_parent_state, shard_parent_state, shard_block, shard_block.slot, shard)
     assert verify_shard_block_signature(beacon_parent_state, signed_shard_block)
+
     post_state = get_post_shard_state(beacon_parent_state, shard_parent_state, shard_block)
+
+    # Add new block to the store
+    shard_store.blocks[hash_tree_root(shard_block)] = shard_block
+
     # Add new state for this block to the store
     shard_store.block_states[hash_tree_root(shard_block)] = post_state
 ```
--- a/specs/phase1/shard-fork-choice.md
+++ b/specs/phase1/shard-fork-choice.md
@@ -151,10 +151,11 @@ def on_shard_block(store: Store, shard_store: ShardStore, signed_shard_block: Si
 
     # Check shard parent exists
     assert shard_block.shard_parent_root in shard_store.block_states
-    pre_shard_state = shard_store.block_states[shard_block.shard_parent_root]
+    shard_parent_state = shard_store.block_states[shard_block.shard_parent_root]
 
     # Check beacon parent exists
     assert shard_block.beacon_parent_root in store.block_states
+    beacon_parent_state = store.block_states[shard_block.beacon_parent_root]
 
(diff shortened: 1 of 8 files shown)


specs/phase1/shard-transition.md
```diff
@@ -44,17 +44,33 @@ def compute_shard_transition_digest(beacon_state: BeaconState,
 
 ```python
 def verify_shard_block_message(beacon_state: BeaconState,
-                               shard_state: ShardState,
+                               shard_parent_state: ShardState,
```
djrtwo, line 47: Similar, we could consider _not_ passing in `shard_parent_state` explicitly because it is a function of `(beacon_state, shard)`. This could prevent accidental misuse of the function

specs/phase1/shard-transition.md
```diff
@@ -44,17 +44,33 @@ def compute_shard_transition_digest(beacon_state: BeaconState,
...
                                block: ShardBlock,
                                slot: Slot,
-                               shard: Shard) -> bool:
-    assert block.shard_parent_root == shard_state.latest_block_root
-    assert block.slot == slot
-    next_slot = Slot(beacon_state.slot + 1)
+                               shard: Shard,
+                               beacon_parent_slot: Slot=None) -> bool:
```
djrtwo, line 51: What is `beacon_parent_slot` for? Isn't `beacon_state` supposed to be the beacon parent state and thus has the `beacon_parent_slot` info embedded in it?

specs/phase1/shard-transition.md
```diff
@@ -44,17 +44,33 @@ def compute_shard_transition_digest(beacon_state: BeaconState,
...
+    # Check `shard_parent_root` field
+    assert block.shard_parent_root == shard_parent_state.latest_block_root
+    # Check `beacon_parent_root` field
+    if beacon_parent_slot is None:
```
djrtwo, line 55: I think we can remove `beacon_parent_slot` entirely and remove both of the outer block `if` statements to just have the following:

```python
    beacon_parent_block_header = beacon_state.latest_block_header.copy()
    if beacon_parent_block_header.state_root == Root():
        beacon_parent_block_header.state_root = hash_tree_root(beacon_state)
        beacon_parent_root = hash_tree_root(beacon_parent_block_header)
    else:
        beacon_parent_root = get_block_root_at_slot(beacon_state, beacon_parent_slot)
    assert block.beacon_parent_root == beacon_parent_root
```

That way, we are just making sure that the `block.beacon_parent_root` is coherent wrt the passed in `beacon_state`.

specs/phase1/shard-transition.md
```diff
@@ -188,25 +204,16 @@ def get_proposal_choices_at_slot(beacon_state: BeaconState,
...
-            assert verify_shard_block_message(beacon_state, shard_state, block.message, slot, shard)
-            if validate_signature:
-                assert verify_shard_block_signature(beacon_state, block)
-
-            shard_state = get_post_shard_state(beacon_state, shard_state, block.message)
-        except Exception:
-            pass  # TODO: throw error in the test helper
-        else:
-            choices.append(block)
+        shard_state = get_post_shard_state(beacon_state, shard_state, block.message)
```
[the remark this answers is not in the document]
djrtwo: That said, `get_proposal_choices_at_slot` is being removed anyway..

specs/phase1/shard-transition.md
```diff
@@ -242,15 +249,17 @@ def get_shard_state_transition_result(
...
     shard: Shard,
     shard_blocks: Sequence[SignedShardBlock],
+    on_time_slot: Slot,
     validate_signature: bool=True,
 ) -> Tuple[Sequence[SignedShardBlock], Sequence[ShardState], Sequence[Root]]:
     proposals = []
     shard_states = []
     shard_state = beacon_state.shard_states[shard]
-    for slot in get_offset_slots(beacon_state, shard):
+    offset_slots = compute_offset_slots(get_latest_slot_for_shard(beacon_state, shard), on_time_slot)
```
djrtwo, line 258: I think I prefer this without `on_time_slot` as in #1704 but we can discuss more if you want.

My main argument is that clients are already very used to transition states up to the slot in question so the slot information can and should already be embedded in `beacon_state`

specs/phase1/shard-fork-choice.md
```diff
@@ -151,10 +151,11 @@ def on_shard_block(store: Store, shard_store: ShardStore, signed_shard_block: Si
 
     # Check shard parent exists
     assert shard_block.shard_parent_root in shard_store.block_states
-    pre_shard_state = shard_store.block_states[shard_block.shard_parent_root]
+    shard_parent_state = shard_store.block_states[shard_block.shard_parent_root]
 
     # Check beacon parent exists
     assert shard_block.beacon_parent_root in store.block_states
+    beacon_parent_state = store.block_states[shard_block.beacon_parent_root]
```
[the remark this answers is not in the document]
hwwhww: It's not valid when fail-crosslinking 

(added some missing lines)
image


In this case, when we processing shard block `S_101`, the `beacon_parent_state` is at `B_101`, where `B_101.shard_states[shard_store.shard].slot` is 99 instead of 100.

specs/phase1/shard-transition.md
hwwhww: also due to  fail-crosslinking, we can't use `beacon_state.shard_states[shard_store.shard]`. :'(

specs/phase1/shard-fork-choice.md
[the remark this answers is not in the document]
djrtwo: Right! my bad

specs/phase1/shard-transition.md
```diff
@@ -44,17 +44,27 @@ def compute_shard_transition_digest(beacon_state: BeaconState,
 
 ```python
 def verify_shard_block_message(beacon_state: BeaconState,
```
djrtwo, line 46: Rename `beacon_parent_state` for consistency

PR #1878, opened 2020-06-09
Remove `ShardState.transition_digest`

hwwhww:
### Issue
This field can be used to:
1. Indicate if the given shard state is at a skipped slot.
2. As a digest of shard state.

For (1), `ShardState.latest_block_root` can also provide it.
For (2), `hash_tree_root(shard_state)` could cover it.

This value looks like a stub at the early draft. Is there any other reason we need it?

### How did I fix it
Remove this field.

### DIFF
--- a/specs/phase1/beacon-chain.md
+++ b/specs/phase1/beacon-chain.md
@@ -349,7 +349,6 @@ class ShardBlockHeader(Container):
 class ShardState(Container):
     slot: Slot
     gasprice: Gwei
-    transition_digest: Bytes32
     latest_block_root: Root
 ```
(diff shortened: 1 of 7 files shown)


vbuterin:
Aaah I remember. The goal of the field (originally called `transition_hash`) was to save a record in the state of the fact that a transition from A -> B took place, so that a fraud proof can later reverse it. If we want we could remove it for now until we figure out the fraud proof architecture.

hwwhww:
Thank you @vbuterin for the reply!

I'd say let's remove it for now.

PR #1885, opened 2020-06-11
Fork choice: minor formatting change

paulhauner:
Non-substantive changes:

- Ensure a comment is above the corresponding line of code.
- Add an additional comment


paulhauner:
> Shouldn't we check parent state exists before copy the parent state?

I was hesitant to move beyond "making sure the spec is accurate" territory into "optimizing away copies", but I don't mind either way. If we're going to move it we should put it right before the state is modified.

djrtwo:
So we do the assert before the state lookup because otherwise (as seen in a failed test), the state lookup in `block_states` throws if a block with unknown parent is passed in

paulhauner:
> So we do the assert before the state lookup because otherwise (as seen in a failed test), the state lookup in `block_states` throws if a block with unknown parent is passed in

Got it! Ordering has been changed :)

PR #1895, opened 2020-06-15
Add `shard: Shard` field to `AttestationData`

hwwhww:
Per #1850: add `Shard` to `AttestationData`.
Pros:
- No need to call `get_shard(state, attestation)`.
- Easy to validate `Attestation` of shard subnet immediately

Cons:
- Increase 8 bytes per `Attestation`.

I increasingly think it's a good idea. 🙂


terencechain:
> /cc @terencechain

Me and the future phase 1 implementers thank you for this 😄

specs/phase1/beacon-chain.md
```diff
@@ -164,6 +163,8 @@ class AttestationData(Container):
     # FFG vote
     source: Checkpoint
     target: Checkpoint
+    # Shard vote
+    shard: Shard
```
djrtwo, line 167: I think if we will want to make sure that the `shard` and `committee_index` are coherent for on-time attestations via a call to `compute_shard_from_committee_index` in `process_attestation`

Then in `process_crosslinks` and `process_crosslink_for_shard` we can directly use `attestation.data.shard`

(I think we'll force a stub for the late attestations for both shard and committee index (e.g. `2**64 - 1`))

terencechain: This should also be part of p2p validation.  If the attestation is on-time, then the shard and committee_index must be coherent, or else ignore/reject the attestation

hwwhww: Whoops, an assertion was lost in the previous auto-merge. Nice catch! 🙏 

> (I think we'll force a stub for the late attestations for both shard and committee index (e.g. 2**64 - 1))

> or else ignore/reject the attestation

Doesn't the late attestation still have to be re-broadcasted to subnet and then get aggregated? I guess we still need the corresponding `commitee_index` field here to locate the correct subnet and compute beacon committee?

djrtwo: Ah, you're right. I was over-optimizing what we could potentially aggregate here. 

Will consider this a bit more but the late attestation being bound to the committee and size of the committee is probably reasonable to keep in place

specs/phase1/beacon-chain.md
```diff
@@ -764,6 +764,10 @@ def validate_attestation(state: BeaconState, attestation: Attestation) -> None:
     if is_on_time_attestation(state, attestation):
         # Correct parent block root
         assert data.beacon_block_root == get_block_root_at_slot(state, compute_previous_slot(state.slot))
+        # Correct shard number
+        assert attestation.data.shard == compute_shard_from_committee_index(
+            state, attestation.data.index, attestation.data.slot
+        )
```
djrtwo, line 770: I think I'd slightly prefer the following for readability and linecount
```suggestion
        shard = compute_shard_from_committee_index(state, attestation.data.index, attestation.data.slot)
        assert attestation.data.shard == shard
```

specs/phase1/beacon-chain.md
```diff
@@ -928,13 +931,16 @@ def process_crosslinks(state: BeaconState,
...
             if is_on_time_attestation(state, attestation) and attestation.data.index == committee_index
         ]
-        shard = compute_shard_from_committee_index(state, committee_index, on_time_attestation_slot)
-        winning_root = process_crosslink_for_shard(state, committee_index, shard_transitions[shard], shard_attestations)
-        if winning_root != Root():
-            # Mark relevant pending attestations as creating a successful crosslink
-            for pending_attestation in state.current_epoch_attestations:
-                if is_winning_attestation(state, pending_attestation, committee_index, winning_root):
-                    pending_attestation.crosslink_success = True
+        if len(shard_attestations) > 0:
```
djrtwo, line 934: I think it reasonable to keep the `shard = ` computation as previous and not get `shard` from `attestation` in this case. This method is less about a single attestation (and it's data) and instead about a committee and it's shard. So calculating the committees shard from the helper instead of grabbing from a particular attestation seems safe and more readable

The previous method was more readable and same security so I'd lean in that direction.

PR #1902, opened 2020-06-16
revert local config settings from .gitignore

djrtwo:
Some local setup files (rather than specific to general operation of the repo) were added in #1705.

I would argue that these should go into the local system's [global gitignore](https://stackoverflow.com/questions/7335420/global-git-ignore/22885996) rather than the repo here

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,13 +28,3 @@ tests/core/pyspec/test-reports
 tests/core/pyspec/eth2spec/test_results.xml
 
 *.egg-info
-
-# flake8 config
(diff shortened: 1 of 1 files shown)


djrtwo:
cc: @dankrad

dankrad:
There seem to be some other only locally relevant entries in there. But sure I can set these locally.

djrtwo:
I think the only local config items in there  might be the different naming conventions for `venv`. Other than that, all of those items can and do appear in the normal build and test processes

PR #1904, opened 2020-06-17
Avoid state in p2p validation, compute committee count per slot for epoch

protolambda:
This PR changes something subtle in the beacon-spec, but does not result in any effective different behavior, and can be regarded as a cosmetic change. I do like some extra review though, as I am not 100% if I am breaking any phase 1 requirements.

---

The previous committee count function looked like:

```python
def get_committee_count_at_slot(state: BeaconState, slot: Slot) -> uint64:
    """
    Return the number of committees at ``slot``.
    """
    ...
```

And I think we should change it to:
```python
def get_committee_count_per_slot(state: BeaconState, epoch: Epoch) -> uint64:
    """
    Return the number of committees in each slot for the given ``epoch``.
    """
    ...
```

We rotate through the committees, and some places already make the assumption that `get_committee_count_at_slot` does not change throughout the epoch.

One of these is the subnet id calculation in the validator spec:
```python
def compute_subnet_for_attestation(state: BeaconState, slot: Slot, committee_index: CommitteeIndex) -> uint64:
    slots_since_epoch_start = slot % SLOTS_PER_EPOCH
    committees_since_epoch_start = get_committee_count_at_slot(state, slot) * slots_since_epoch_start
    return (committees_since_epoch_start + committee_index) % ATTESTATION_SUBNET_COUNT
```

The `compute_subnet_for_attestation` is used in the p2p attestation validation, and creates the impression that the latest state of an attestation is necessary to validate. While it is necessary to have validated the attested block, having the full state in the p2p pipeline complicates things a lot for an implementer.

So we can change that too:
```python
def compute_subnet_for_attestation(committees_per_slot: uint64, slot: Slot, committee_index: CommitteeIndex) -> uint64:
    committees_since_epoch_start = committees_per_slot * slots_since_epoch_start
    return (committees_since_epoch_start + committee_index) % ATTESTATION_SUBNET_COUNT
```

The attestation signature and bitfield checks already require the committee information, so it is much less difficult to provide the `committees_per_slot` number than the full state.

Similarly, the attestation bitfield check does not require the state either, a client should just check if a single bit has been set. (unless there is some indirect bitfield length check that is not already checked by another condition, like the signature check?)

So this PR changes:
- `get_committee_count_at_slot` -> `get_committee_count_per_slot` (and update usages, caching, testing, phase 1)
- Update `compute_subnet_for_attestation` to take the committee count per slot as input parameter. Continuing the trend set by PR #1876 to simplify input parameters for the benefit of clients
- Simplify the bitfield check to also not require the full state
- This also helped clarify the comment in the p2p spec about phase-1 like rotation of committees around the subnets, throughout an epoch.

Thanks to @paulhauner who highlighted how the subnet attestation validation was different from the other conditions by requiring a full beacon state. And thanks @cemozerr for an earlier mention of the unnecessary inefficiency of the attestation bitfield condition.

----

Note: line count increased a little, not because of changes, but the network spec has unnecessarily long markdown lines, making it difficult to browse git history. Markdown renders consecutive lines as the same paragraph, so it looks the same. So I'm splitting every line that is modified where I am making changes.


hwwhww:
> I am not 100% if I am breaking any phase 1 requirements.

I think this PR doesn't break phase 1 requirements. However, it points out the need of avoiding passing the full state when we are computing phase 1 shard subnet. It may be more verbose to achieve it.

djrtwo:
Oh no, I made merge conflicts

Issue #1901, opened 2020-06-16
Transition SSZ out of Eth2 specs repo, introduce new SSZ specs repo.


protolambda:
First of all: this does not change any existing SSZ functionality, this is a proposal to extend the information available, and move it out of the Eth2 specs repository.

The current SSZ specification in this repository is very dense, prune to bugs because of it, and most of all, incomplete with regards to the more advanced features.

There is a lot I want the SSZ spec to describe: principles, ideas, advanced topics, and more description for hardening certain decoding features. However, this seems like too much to fit into the Eth2 specs.

So instead, I have this draft up for an SSZ specs repository:

**https://github.com/protolambda/eth2.0-ssz/** (yes, moving it out of a personal GitHub repo once we have a transition plan).

I also plan to include ssz-specific testing in this new repository as well, to better cover invalid SSZ data, and help SSZ implementations embed just the tests they need, as opposed to the full Eth2 suite of tests. (fastssz as standalone Go library being a good example)

Additionally, I am looking to do versioning differently, where features grow in stages, and feedback on style and wording can be merged into the specs more quickly for unstable stages.

Hopefully this also enables the formal verification team (cc @franck44, @booleanfunction) to express their findings better. I feel like any change in the current SSZ spec here confuses implementers more than it helps them, since the functionality is there, but the way it is written is difficult to follow, and having a debate whether or not something is a bug, or how/when something breaks, shows how limiting the spec format is.

So this is an open call for others involved in SSZ to give the draft repo a look, help iterate on it now that it's still an early draft and friction is lower, and to plan on a transition to this separate SSZ repository.

CC @wemeetagain @zah - since you have both spent a lot of time on more advanced SSZ implementations, you may be interested.

hwwhww:
My two cents

#### IMHO what are the insufficient parts of the status-quo SSZ spec:
1. **The historical problem**
    1. [Phase 0] `simple-serialize.md`: there was a [pyspec-only ssz implementation](https://github.com/ethereum/eth2.0-specs/blob/v0.9.0/test_libs/pyspec/eth2spec/utils/ssz/ssz_impl.py) before the spec moved to [`remerkleable`](https://github.com/ethereum/eth2.0-specs/pull/1552). Some Python functions that the spec is referencing are from that implementation. `py-ssz` itself has some optimization so that `py-ssz` doesn't 100% follow the spec presentation too.
[...]

PR #1907, opened 2020-06-17
Fix deprecated utility code, avoid wrong helper function name, add tests

protolambda:
The next/previous-power-of-two naming was wrong, the 0 case was wrong, and the functions were prone to off-by-one errors, lacking comments and tests.
@franck44 raised [an issue describing the bug in detail and proposed a fix](https://github.com/ethereum/eth2.0-specs/issues/1695), thank you! This was long ago, but it was never included because no part of the spec used the code.

Now that SSZ is getting some attention again, I decided to implement their fix, with name changes and comments in the merkle functions, to show what it was supposed to be (Or at least what I think it does, I think the original was quick experimental code of Vitalik).

Note that since the code has not been used for so long, it can be regarded as deprecated in this repository. And I have the intention to include these more experimental/extra merkleization things in a "draft" stage document in the new SSZ spec. See #1901 

Marking this as do-not-merge, and prioritizing the removal of the unused code. Let's leave this issue open while we make a plan in #1901

### DIFF
--- a/ssz/merkle-proofs.md
+++ b/ssz/merkle-proofs.md
@@ -26,25 +26,34 @@
 ## Helper functions
 
 ```python
-def get_next_power_of_two(x: int) -> int:
+def get_power_of_two_ceil(x: int) -> int:
     """
-    Get next power of 2 >= the input.
+    Get the power of 2 for given input, or the closest higher power of 2 if the input is not a power of 2.
+    Commonly used for "how many nodes do I need for a bottom tree layer fitting x elements?"
+    Example: 0->1, 1->1, 2->2, 3->4, 4->4, 5->8, 6->8, 7->8, 8->8, 9->16.
     """
-    if x <= 2:
-        return x
+    if x <= 1:
+        return 1
+    elif x == 2:
+        return 2
     else:
-        return 2 * get_next_power_of_two((x + 1) // 2)
+        return 2 * get_power_of_two_ceil((x + 1) // 2)
 ```
 
 ```python
-def get_previous_power_of_two(x: int) -> int:
+def get_power_of_two_floor(x: int) -> int:
     """
-    Get the previous power of 2 >= the input.
+    Get the power of 2 for given input, or the closest lower power of 2 if the input is not a power of 2.
+    The zero case is a placeholder and not used for math with generalized indices.
+    Commonly used for "what power of two makes up the root bit of the generalized index?"
+    Example: 0->1, 1->1, 2->2, 3->2, 4->4, 5->4, 6->4, 7->4, 8->8, 9->8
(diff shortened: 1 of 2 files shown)


franck44:
> @franck44 raised an issue long ago, but it was never fixed because no part of the spec used the code.

We did not only raise an issue but uncovered the bug and proposed a fix.
The functions **were not prone to off-by-one errors**.
The bug is for input 0 not for other inputs. 

> Now that SSZ is getting some attention again, I decided to fix it to show what it was supposed to be (Or at least what I think it does, I think the original was quick experimental code of Vitalik).

**WE** (in this [repo](https://github.com/PegaSysEng/eth2.0-dafny)) uncovered the error and fixed it, see [issue 1695](https://github.com/ethereum/eth2.0-specs/issues/1695), and proved the full correctness of the fix [here](https://github.com/PegaSysEng/eth2.0-dafny/blob/d3a3e8b799e2a4b8bac12ec5ccc9ef6123842f53/src/dafny/utils/MathHelpers.dfy).

This issue is a duplicate of [this one](https://github.com/ethereum/eth2.0-specs/issues/1695) where the fix was **ALREADY** proposed. 
This issue and this PR is a very simple **IMPLEMENTATION** (and verbatim copy) of our fix in the Python specs.
The PR should probably have been in relation to [issue 1695](https://github.com/ethereum/eth2.0-specs/issues/1695) and the credit of the fix properly attributed to the right people.

protolambda:
> The functions were not prone to off-by-one errors.

They are, the two functions are different by 1, are easily confused, and the exact increments are easy to get wrong when implemented in with bit-arithmetic instead of recursive implementation, as an implementer would do. The 0 case is a separate thing.

> The bug is for input 0 not for other inputs.

Which is not used for the generalized indices, nor should it have been called next/previous power of two when it is different.

> WE (in this repo) uncovered the error and fixed it,

Yes, sorry if I did not give enough credit, that was not the intention. I mentioned the issue where it originated from, and you deserve full credit. I will edit the original post to make it more clear, my apologies. Right now I all I want is to close all these old issues, deprecate unused and buggy code, and improve on it.

That said, the intention of the PR was to finally surface the bugfix that has been sitting in an old issue because nobody is using the code. Better now than never, I agree that it could have been much faster, but it was not prioritized because it is unused code, and then it was basically forgotten for a while. If you believe something is important, nudge us to make a PR, or submit a PR yourself.

protolambda:
If the decision on where the SSZ and merkle descriptions/code/spec should go is stalled, then we may as well merge this. 

Removed the label, it can be merged. It isn't used anywhere, so clients should be fine with the change (except maybe for noise in the changelog).

Issue #1908, opened 2020-06-17
Type error in `process_custody_game_operations` when invoking `process_chuck_challenge`


ericsson49:
[process_custody_game_operations](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase1/custody-game.md#custody-game-operations) applies [process_chunk_challenge](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase1/custody-game.md#chunk-challenges) for elements of `body.chunk_challenges` and `body.chunk_challenge_responses` lists.

However, there is a typing error, since the `process_custody_game_operations` expects that its second parameter has type `CustodyChunkChallenge`, while both `BeaconBlockBody.chunk_challenges` and
`BeaconBlockBody.chunk_challenges` have type `List[CustodyChunkResponse, MAX_CUSTODY_CHUNK_CHALLENGES]`

Here is a test snippet illustrating the problem
```python
from eth2spec.phase1 import spec

state = spec.BeaconState()
body = spec.BeaconBlockBody(chunk_challenges=[spec.CustodyChunkResponse()])
spec.process_custody_game_operations(state, body)
```

Stack trace
```
Traceback (most recent call last):
  File "/mnt/c/Users/ericsson/IdeaProjects/eth2.0-specs/venv/lib/python3.8/site-packages/remerkleable/complex.py", line 781, in __getattr__
    i = self.__class__._field_indices[item]
KeyError: 'attestation'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/mnt/c/Users/ericsson/IdeaProjects/eth2.0-specs/venv/lib/python3.8/site-packages/eth2spec/phase1/spec.py", line 2128, in process_custody_game_operations
    for_ops(body.chunk_challenges, process_chunk_challenge)
  File "/mnt/c/Users/ericsson/IdeaProjects/eth2.0-specs/venv/lib/python3.8/site-packages/eth2spec/phase1/spec.py", line 2126, in for_ops
    fn(state, operation)
  File "/mnt/c/Users/ericsson/IdeaProjects/eth2.0-specs/venv/lib/python3.8/site-packages/eth2spec/phase1/spec.py", line 2137, in process_chunk_challenge
    assert is_valid_indexed_attestation(state, get_indexed_attestation(state, challenge.attestation))
[...]

PR #1909, opened 2020-06-17
Fix typing error in `process_custody_game_operations`

ericsson49:
Change types of `BeaconBlockBody.chunk_challenges` and `BeaconBlockBody.chunk_challenge_responses` to be `List[CustodyChunkChallenge, ...]`.
It's a suggested fix for #1908

### DIFF
--- a/specs/phase1/beacon-chain.md
+++ b/specs/phase1/beacon-chain.md
@@ -252,8 +252,8 @@ class BeaconBlockBody(Container):
     deposits: List[Deposit, MAX_DEPOSITS]
     voluntary_exits: List[SignedVoluntaryExit, MAX_VOLUNTARY_EXITS]
     # Custody game
-    chunk_challenges: List[CustodyChunkResponse, MAX_CUSTODY_CHUNK_CHALLENGES]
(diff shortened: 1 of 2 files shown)


protolambda:
Ah, good catch, something is definitely off there. Thanks for making a PR :+1: 

I think part of the problem is that `process_chunk_challenge_response` is unused and should have been called as part of the operations routine.

So I think it would need to be:
```
for_ops(body.chunk_challenge_responses, process_chunk_challenge_responses)
```
in `process_custody_game_operations`

And while your change for the first field change looks correct (`chunk_challenges` being a list of `CustodyChunkChallenge`), the second field change should probably stay a response type (`chunk_challenge_responses` being a list of `CustodyChunkResponse`)

Does that sound right? Let me know if that works

ericsson49:
> Does that sound right? Let me know if that works

Thanks! That definitely sounds right. Fixed.

protolambda:
@ericsson49 thanks, looks good. @hwwhww can you sanity check this for me please? I hope we didn't miss any other phase 1 typing problems. Maybe with your typing PRs we can catch cases like this in the future, I'll try to help to get those PRs merged :+1:

Issue #1892, opened 2020-06-15
SSZ spec: Merkleization - Typo in wording


booleanfunction:
In [simple-serialize.md](https://github.com/ethereum/eth2.0-specs/ssz/simple-serialize.md), within the Merkelization section it says:

```
We now define Merkleization hash_tree_root(value) of an object value recursively:

merkleize(pack(value)) if value is a basic object or a vector of basic objects.
```

This part should exclude bitvectors as they are dealt with separately.

A suggested wording to correct this typo could be:

`merkleize(pack(value)) if value is a basic object or a vector of basic objects, excluding bitvectors.
`

booleanfunction:
The case for basic lists should also be amended to exclude bitlists, which are also dealt with separately. Hence 
`mix_in_length(merkleize(pack(value), limit=chunk_count(type)), len(value)) if value is a list of basic objects.`
should read:
`mix_in_length(merkleize(pack(value), limit=chunk_count(type)), len(value)) if value is a list of basic objects, excluding bit lists.`

protolambda:
Bitvectors and bitlists are not considered to be part of "basic" lists and vectors category. Bitlists have a special delimiter bit in serialization, and basic element types are all aligned to bytes. Adding "excluding" may clarify it, but at the same time it gives the impression they would be included otherwise, while they should not be.

booleanfunction:
@protolambda Thank you for the reply. My thinking was that when it says `if value is a basic object ...` that it was referring to a basic type of either a boolean or uintN, and so when the rest of the sentence said `or a vector of basic objects` I read it in the same way as if it said _or a vector of booleans or uintNs_ - hence the suggestion to exclude. Can I check that the initial reference to a basic object is meant to be include a boolean? Such that a boolean would be packed using the `pack` function rather than `pack_bits`?

protolambda:
There are two ways to represent a series of booleans: `List[boolean, N]` and `Bitlist[N]`, these are different.

A `List` will pack basic element types, but does so on a byte level, without any special information about the element type except 1. if it's fixed-length or not, and 2. its byte length. One boolean is serialized as a single byte, so it takes one byte in the list. Also note that empty lists can be serialized as zero bytes thanks to this. A bitlist has a delimiter bit however, which forces a minimum of 1 byte. The one byte per element minimum in hash-tree-root of a regular list is mostly for consistency with serialization.

[...]

PR #1912, opened 2020-06-18
Clarification of SSZ bitlist and bitvector types.

booleanfunction:
## What was wrong?

More a clarification, this PR has been created on the basis of #1892 and discussions with @protolambda (thank you).

## How was it fixed?

I have added a note to the end of the Composite types section. I wasn't sure if the comment regarding the preference for bitvectors/bitlists is necessary (?) but it can always be edited :)

I also modified the table of default values so as to keep the notation consistent with that used earlier.

## Cute animal picture
1920px-Short-beaked_echidna_in_ANBG

### DIFF
--- a/ssz/simple-serialize.md
+++ b/ssz/simple-serialize.md
@@ -66,6 +66,8 @@
 * **union**: union type containing one of the given subtypes
     * notation `Union[type_0, type_1, ...]`, e.g. `union[null, uint64]`
 
+*Note*: Both `Vector[boolean, N]` and `Bitvector[N]` are valid, yet distinct due to their different serialization requirements. Similarly, both `List[boolean, N]` and `Bitlist[N]` are valid, yet distinct. Generally `Bitvector[N]`/`Bitlist[N]` are preferred because of their serialization efficiencies.
+
 ### Variable-size and fixed-size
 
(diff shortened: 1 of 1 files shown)


protolambda:
> I wasn't sure if the comment regarding the preference for bitvectors/bitlists is necessary

The current beacon spec does does not make recommendations like that, but in the validator and network spec it is more common to find such comments, to clarify more practical choices. So it's probably fine.

And thanks for fixing the type description in the defaults, we missed that.

----

And meta-feedback, since this is your first PR (congrats!):

> Cute animal picture

I don't mind the picture, and I see you picked it up from other ethereum python repos like py-ssz, but in this repo we don't put them in PRs. Thanks for using a PR template though, it's nice and readable.

And I recommend you sign future git commits. We don't mind unsigned commits for small contained changes like this, but it's considered best practice. [Here's an easy tutorial](https://help.github.com/en/github/authenticating-to-github/signing-commits), and there are more tutorials for configuring the keys on github etc.

booleanfunction:
@protolambda thank you for the extra feedback and taking the time to include a tutorial link. I will make sure that if I do any PRs in future that I follow your advice. Thank you, it is much appreciated :)

PR #1920, opened 2020-06-18
Fix usages of shard in `process_crosslink_for_shard`

terencechain:
To apply shard transition in `process_crosslink_for_shard`, we can't use `attestation` because it's outside of the `transition_attestations` for loop. One naive solution is to use `attestations[0]`. It's not the prettiest, and it assumes input `attestations` are all from the same `shard`, which is what caller `process_crosslinks` does. If anyone has a better solution, feel free to supersede this. I mainly wanted to open this to point out misuses.

### DIFF
--- a/specs/phase1/beacon-chain.md
+++ b/specs/phase1/beacon-chain.md
@@ -901,7 +901,7 @@ def process_crosslink_for_shard(state: BeaconState,
         assert shard_transition_root == hash_tree_root(shard_transition)
 
         # Apply transition
-        apply_shard_transition(state, attestation.data.shard, shard_transition)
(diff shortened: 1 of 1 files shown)


djrtwo:
nice catch! Also shows a massive hole in current sanity tests...

I recommend we use [compute_shard_from_committee_index(state, committee_index, on_time_attestation_slot)](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase1/beacon-chain.md#compute_shard_from_committee_index) at the top of the function (near `committee = ...`). This function is more about crosslinking a shard rather than doing attestation validations so it makes sense to compute the shard you are looking for directly

PR #1921, opened 2020-06-19
Fix no winning root case + add tests + fix #1881

### LINKED ISSUE #1881: Incorrect usage of `shard_transition.shard_data_roots` in `process_crosslink_for_shard`
#1878 motivated me to think whether `shard_head_root` in `AttestationData` is necessary. 

To accept an `attestation`, the node must has seen the `ShardTransition` that corresponds to the `shard_transition_root`. With the seen `ShardTransition`, the node can get the `shard_head_root` via `shard_transition.shard_data_roots[len(shard_transition.shard_data_roots) - 1]`

I think it's a nice saving of 32bytes since attestations are the most frequent objects that propagate in the network

hwwhww:
### Bugfix
1. Fix `is_winning_attestation` condition: should be on time attestation
2. Fix `process_crosslink_for_shard`: we can only check
`latest_block_root` if it's the attestation for winning shard transition
3. Fix #1881, use `ShardState.latest_block_root` as `attestation.data.shard_head_root`. Thanks to @terencechain!

### Testing
1. Refactor `test_process_shard_transition.py`
2. Run `process_attestation` before running `process_shard_transitions` in tests.
2. Add `test_no_winning_root` for testing the case of that only 1/3 attesters voted for a valid shard transition. The input `shard_transition` should be empty `ShardTransition()` for this case.
3. Add `test_wrong_shard_transition_root` for testing invalid `shard_transition_root != hash_tree_root(shard_transition)` case in `process_crosslink_for_shard`.

### DIFF
--- a/specs/phase1/beacon-chain.md
+++ b/specs/phase1/beacon-chain.md
@@ -671,7 +671,7 @@ def is_winning_attestation(state: BeaconState,
     ``winning_root`` formed by ``committee_index`` committee at the current slot.
     """
     return (
-        attestation.data.slot == state.slot
+        is_on_time_attestation(state, attestation)
         and attestation.data.index == committee_index
         and attestation.data.shard_transition_root == winning_root
     )
@@ -886,9 +886,10 @@ def process_crosslink_for_shard(state: BeaconState,
         for attestation in transition_attestations:
             participants = get_attesting_indices(state, attestation.data, attestation.aggregation_bits)
             transition_participants = transition_participants.union(participants)
-            assert attestation.data.shard_head_root == shard_transition.shard_data_roots[
-                len(shard_transition.shard_data_roots) - 1
-            ]
+            if len(shard_transition.shard_data_roots) > 0:
+                # Is the `shard_transition` candidate
+                last_offset_index = len(shard_transition.shard_data_roots) - 1
+                assert attestation.data.shard_head_root == shard_transition.shard_data_roots[last_offset_index]
 
         enough_online_stake = (
             get_total_balance(state, online_indices.intersection(transition_participants)) * 3 >=
--- a/specs/phase1/beacon-chain.md
+++ b/specs/phase1/beacon-chain.md
@@ -886,10 +886,11 @@ def process_crosslink_for_shard(state: BeaconState,
         for attestation in transition_attestations:
             participants = get_attesting_indices(state, attestation.data, attestation.aggregation_bits)
             transition_participants = transition_participants.union(participants)
-            if len(shard_transition.shard_data_roots) > 0:
+            if len(shard_transition.shard_states) > 0:
                 # Is the `shard_transition` candidate
-                last_offset_index = len(shard_transition.shard_data_roots) - 1
-                assert attestation.data.shard_head_root == shard_transition.shard_data_roots[last_offset_index]
+                last_offset_index = len(shard_transition.shard_states) - 1
+                shard_head_root = shard_transition.shard_states[last_offset_index].latest_block_root
+                assert attestation.data.shard_head_root == shard_head_root
 
         enough_online_stake = (
             get_total_balance(state, online_indices.intersection(transition_participants)) * 3 >=
--- a/specs/phase1/beacon-chain.md
+++ b/specs/phase1/beacon-chain.md
@@ -772,6 +772,8 @@ def validate_attestation(state: BeaconState, attestation: Attestation) -> None:
         # Correct shard number
         shard = compute_shard_from_committee_index(state, attestation.data.index, attestation.data.slot)
         assert attestation.data.shard == shard
+        # On-time attestations should have a non-empty shard transition root
+        assert attestation.data.shard_transition_root != hash_tree_root(ShardTransition())
     # Type 2: no shard transition
     else:
         # Ensure delayed attestation
@@ -886,11 +888,6 @@ def process_crosslink_for_shard(state: BeaconState,
         for attestation in transition_attestations:
             participants = get_attesting_indices(state, attestation.data, attestation.aggregation_bits)
             transition_participants = transition_participants.union(participants)
-            if len(shard_transition.shard_states) > 0:
-                # Is the `shard_transition` candidate
-                last_offset_index = len(shard_transition.shard_states) - 1
-                shard_head_root = shard_transition.shard_states[last_offset_index].latest_block_root
-                assert attestation.data.shard_head_root == shard_head_root
 
         enough_online_stake = (
             get_total_balance(state, online_indices.intersection(transition_participants)) * 3 >=
@@ -902,6 +899,12 @@ def process_crosslink_for_shard(state: BeaconState,
 
         # Attestation <-> shard transition consistency
         assert shard_transition_root == hash_tree_root(shard_transition)
+
+        # Check `shard_head_root` of the winning root
+        last_offset_index = len(shard_transition.shard_states) - 1
+        shard_head_root = shard_transition.shard_states[last_offset_index].latest_block_root
+        for attestation in transition_attestations:
+            assert attestation.data.shard_head_root == shard_head_root
 
         # Apply transition
         apply_shard_transition(state, shard, shard_transition)
--- a/specs/phase1/beacon-chain.md
+++ b/specs/phase1/beacon-chain.md
@@ -652,11 +652,11 @@ def get_offset_slots(state: BeaconState, shard: Shard) -> Sequence[Slot]:
 
 ```python
 def is_on_time_attestation(state: BeaconState,
-                           attestation: Attestation) -> bool:
+                           attestation_data: AttestationData) -> bool:
     """
-    Check if the given attestation is on-time.
+    Check if the given ``attestation_data`` is on-time.
     """
-    return attestation.data.slot == compute_previous_slot(state.slot)
+    return attestation_data.slot == compute_previous_slot(state.slot)
 ```
 
 #### `is_winning_attestation`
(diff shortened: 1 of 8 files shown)


specs/phase1/beacon-chain.md
```diff
@@ -886,9 +886,10 @@ def process_crosslink_for_shard(state: BeaconState,
         for attestation in transition_attestations:
             participants = get_attesting_indices(state, attestation.data, attestation.aggregation_bits)
             transition_participants = transition_participants.union(participants)
-            assert attestation.data.shard_head_root == shard_transition.shard_data_roots[
-                len(shard_transition.shard_data_roots) - 1
-            ]
+            if len(shard_transition.shard_data_roots) > 0:
+                # Is the `shard_transition` candidate
+                last_offset_index = len(shard_transition.shard_data_roots) - 1
+                assert attestation.data.shard_head_root == shard_transition.shard_data_roots[last_offset_index]
```
terencechain, line 892: Is this still correct? I was under the impression that `shard_data_roots` is used for shard block bodies. Ex:
```python
hash_tree_root(shard_block.message.body)
```

https://github.com/ethereum/eth2.0-specs/issues/1881

specs/phase1/beacon-chain.md
```diff
@@ -671,7 +671,7 @@ def is_winning_attestation(state: BeaconState,
     ``winning_root`` formed by ``committee_index`` committee at the current slot.
     """
     return (
-        attestation.data.slot == state.slot
+        is_on_time_attestation(state, attestation)
```
hwwhww, line 674: Hmmm, just found here is a typing error, the given attestation is `PendingAttestation`, but `is_on_time_attestation` takes `Attestation`. We can either:
1. change this line to `attestation.data.slot == compute_previous_slot(state.slot)` or
2. make `is_on_time_attestation` take `attestation: AttestationData`

specs/phase1/beacon-chain.md
hwwhww: Ahh yes, sorry I forgot it when I fixing another issue!

So here this line should be

`assert attestation.data.shard_head_root == shard_transition.shard_states[last_offset_index].latest_block_root`

right?

specs/phase1/beacon-chain.md
hwwhww: @terencechain any thought? :)

specs/phase1/beacon-chain.md
terencechain: That looks good to me!

specs/phase1/beacon-chain.md
terencechain: I think 2 is nicer. This reduces disparity and avoids multiple ways of checking whether attestations (pending or not) is on time or not 👍

specs/phase1/beacon-chain.md
```diff
@@ -886,9 +886,11 @@ def process_crosslink_for_shard(state: BeaconState,
...
             participants = get_attesting_indices(state, attestation.data, attestation.aggregation_bits)
             transition_participants = transition_participants.union(participants)
-            assert attestation.data.shard_head_root == shard_transition.shard_data_roots[
-                len(shard_transition.shard_data_roots) - 1
-            ]
+            if len(shard_transition.shard_states) > 0:
+                # Is the `shard_transition` candidate
+                last_offset_index = len(shard_transition.shard_states) - 1
+                shard_head_root = shard_transition.shard_states[last_offset_index].latest_block_root
+                assert attestation.data.shard_head_root == shard_head_root
```
terencechain, line 893: Second thought.... What do you think about moving this to `validate_attestation`?  Feels more natural and more obvious that it's an on-time check.

```python
    # Type 1: on-time attestations
    if is_on_time_attestation(state, attestation):
     # Assert correct parent block root
     # Assert correct shard number
     # Assert shard head root is consistent with latest shard state block root
     ...
```

hwwhww: I feel the same, but it would break the compatibility of `process_attestation` interface (adding `shard_transitions: Seqence[ShardTransition]`) from phase0 to phase1. Especially, we currently have some phase 0 test generators that are compatible with both phase0 and phase1. If we change the interface, the test case and test format would be messier. 

That said, IMO it's worth to reconsider when phase1 spec content is complete.

terencechain: Ah yes. I forgot about about the compatibility. Reconsider at a later date sounds good to me

djrtwo, line 893: This assertion doesn't seem correct (nor was it prior to the recent change).

The single `shard_transition` is passed in to this function (the winning one included in the block) does not have to be reflected in all of the attestations included in the block. This assertion is forcing all attestations in the block to agree with the shard block head of the winning shard transition (if there is one).

I think what we want here is to make sure that `shard_head_block` votes and `shard_transition` votes are coherent. We can only do this for attestations that voted for the winning shard transition

djrtwo: As for the `if len(shard_transition.shard_states) > 0:` part of this... I wish there was a cleaner way. Let's get the above fixed then we can discuss it again

tests/core/pyspec/eth2spec/test/phase1/block_processing/test_process_shard_transition.py
```diff
@@ -3,69 +3,196 @@
...
 def test_multiple_offset_slots(spec, state):
     # NOTE: this test is only for full crosslink (minimal config), not for mainnet
-    yield from run_basic_crosslink_tests(spec, state, target_len_offset_slot=2, valid=True)
+    yield from run_successful_crosslink_tests(spec, state, target_len_offset_slot=2)
+
+
+@with_all_phases_except([PHASE0])
+@spec_state_test
+def test_no_winning_root(spec, state):
+    # NOTE: this test is only for full crosslink (minimal config), not for mainnet
```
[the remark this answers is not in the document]
hwwhww: > So this fails for mainnet config?

Yes, it fails for mainnet config.

> Probably need to put a blocker condition for the config. Something like `if spec.INITIAL_ACTIVE_SHARDS > 2: return`

Good point 👍  
I added [`is_full_crosslink`](https://github.com/ethereum/eth2.0-specs/pull/1921/commits/97d0048eaa86cebcd2d4f89c511753bde474af2a) for determining if the test should be skipped.

specs/phase1/beacon-chain.md
hwwhww: You're right. To make it correct, it can be checked after `enough_online_stake is True`.

```python
        # If not enough stake, try next transition root
        if not enough_online_stake:
            continue

        # Attestation <-> shard transition consistency
        assert shard_transition_root == hash_tree_root(shard_transition)

        # Check `shard_head_root` of the winning attestations
        last_offset_index = len(shard_transition.shard_states) - 1
        shard_head_root = shard_transition.shard_states[last_offset_index].latest_block_root
        for attestation in transition_attestations:
            assert attestation.data.shard_head_root == shard_head_root
```

@djrtwo @terencechain mind taking another look?

djrtwo: `assert shard_transition != ShardTransition()` this could cause the chain to stall if a corrupt committee voted on `ShardTransition()` unless the node was looking for this type of attack and not including attestations because of what might happen. 

We want to add add in `validate_attestation` under `if is_on_time_attestation`: `assert attestation.data.shard_transition_root != hash_tree_root(ShardTransition())`.

We also kind of handle this in [verify_empty_shard_transition](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase1/beacon-chain.md#verify_empty_shard_transition) already

Not sure. Need to think about these corner cases a bit more. We definitely want the chain to be able to move forward even if a corrupt committee

djrtwo: Otherwise, this looks good.

We might consider checking consistency of all such FullAttestations in shard fork choice. Or maybe even making inconsistency here slashable

terencechain: Can the attestations that voted for the winning shard transition using different `shard_head_root`?? My intuition is no, and that marks the block as invalid.  Say a block consists the following attestations:
```
{slot:100, committe_index:1, shard_transition_root: 0xa, shard_head_root: 0xb}
{slot:100, committe_index:1, shard_transition_root: 0xa, shard_head_root: 0xb}
{slot:100, committe_index:1, shard_transition_root: 0xa, shard_head_root: 0xc}
{slot:100, committe_index:1, shard_transition_root: 0xd, shard_head_root: 0xe}
```
`0xa` gets the crosslink because of 3/4, but the block is bad because the third attestation has `0xc`. Or do we simply not count 0xc, and the vote is 2/4?

hwwhww: @djrtwo 
> this could cause the chain to stall if a corrupt committee voted on `ShardTransition()`

I think it has been banned by `verify_empty_shard_transition`?

> We want to add add in `validate_attestation` under `if is_on_time_attestation`: `assert attestation.data.shard_transition_root != hash_tree_root(ShardTransition())`.

Yep, it looks better. 👍 

> We might consider checking consistency of all such FullAttestations in shard fork choice. Or maybe even making inconsistency here slashable

Sounds correct to me.

Btw, to double-check, in [phase 1 v-guide: Broadcast aggregate](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase1/validator.md#broadcast-aggregate), is the `AggregateAndProof.aggregate` type `Attestation` or `FullAttestation`? I suppose the beacon block proposer needs the shard transitions to maintain the [`full_transitions` map](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase1/validator.md#shard-transitions)?

---

@terencechain

Good question!
My intuition is no too, and I think that is a bad block. Moreover, I wonder if we can validate the `shard_head_root` of the 4th attestation as well. But to do it, we need the beacon block to contain all the corresponding `shard_transition` of all attestations... which may blow up the beacon block size...

((keep thinking about it))

djrtwo: > I think it has been banned by verify_empty_shard_transition?

A proposer adding `ShardTransition()` for a specific shard is them signalling that the shard in question did not get the requisite support to for a transition. This would not prevent attesters from actually voting on empty shard transition (which could cause weird inconsistencies which we just banned with the new condition)

> Btw, to double-check, in phase 1 v-guide: Broadcast aggregate, is the AggregateAndProof.aggregate type Attestation or FullAttestation?

It would need to be `FullAttestation`.  But I think we might split up attestation data from shard transitions via their own gossipsub topics to prevent duplication so we'll address it there

djrtwo: > . But to do it, we need the beacon block to contain all the corresponding shard_transition of all attestations... which may blow up the beacon block size...

yeah, we could but are currently avoiding extra shard transitions. (1) we can ignore incoherent messages in the fork choice, and (2) we could even make this incoherency slashable but not sure it's worth it

PR #1922, opened 2020-06-19
Remove shard check in `verify_shard_block_message`

terencechain:
The shard field consistency check in `verify_shard_block_message` is redundant
```python
shard = block.shard
assert block.shard == shard
```
Given the input argument, I'm not able to find a proper consistency check the for block's shard field. 
One can argue we don't need to check for the shard field because `shard_parent_root` indirectly validates `shard`.  The shard field is most useful to looking up shard blocks are coming in from the correct p2p subnet (`shard_{shard}_block`)

Opening this PR for suggestion


djrtwo:
just saw this is being addressed in #1925

PR #1924, opened 2020-06-19
BeaconState.exposed_derived_secrets (of Vector type) initialized to an empty list

ericsson49:
[upgrade_to_phase1](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase1/phase1-fork.md#upgrading-the-state) creates `BeaconState` with its field `exposed_derived_secrets` initialized to `[] * EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS`.
However, `exposed_derived_secrets` has a type of `Vector[List[ValidatorIndex, ...], ...]` and the initialization expression evaluates to an empty list, not to a list of length `EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS`.

As a Vector field is initialized with an empty list in result, it's behavior is not evident.
@protolambda clarified that an error should be raised (if the vector has a non-zero length).

I suspect that the problem is a result of a typo, i.e. `[[]] * <num>` should be used instead of `[] * <num>`.

### DIFF
--- a/specs/phase1/phase1-fork.md
+++ b/specs/phase1/phase1-fork.md
@@ -111,7 +111,7 @@ def upgrade_to_phase1(pre: phase0.BeaconState) -> BeaconState:
         current_light_committee=CompactCommittee(),  # computed after state creation
         next_light_committee=CompactCommittee(),
         # Custody game
-        exposed_derived_secrets=[[]] * EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS,
+        exposed_derived_secrets=[()] * EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS,
         # exposed_derived_secrets will fully default to zeroes
     )
     next_epoch = Epoch(epoch + 1)
--- a/specs/phase1/phase1-fork.md
+++ b/specs/phase1/phase1-fork.md
@@ -111,7 +111,7 @@ def upgrade_to_phase1(pre: phase0.BeaconState) -> BeaconState:
         current_light_committee=CompactCommittee(),  # computed after state creation
         next_light_committee=CompactCommittee(),
         # Custody game
-        exposed_derived_secrets=[] * EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS,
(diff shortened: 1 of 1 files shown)


protolambda:
I think this is related to how `Vector()` intializes the default, while `Vector(a,b,c)` (init list-like) and `Vector([a,b,c])` (convert from list), are all supported by the constuctor at the same time. When coercing a non-ssz type like a python `[ ]`, it should be checking the length etc.

And while `[[]] * <num>` looks good, I think a list in a list multiplication is prone to problems in python: the element doesn't get deep-copied, so you end up with `<num>` references to the same thing, and when you mutate one, it affects each item. This is how I think it works in python at least. Once converted to remerkleable instances, each item is converted to a SSZ value individually, and effectively copied.

And by default a vector initializes to the correct length anyway, so there's no point in using python literals to init the field. Instead, you could omit it from the constructor (rather not though, explicit here is nice), or use the actual ssz type.

So for now we could use `Vector[List[ValidatorIndex, MAX_EARLY_DERIVED_SECRET_REVEALS * SLOTS_PER_EPOCH], EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS]()`, although ugly. It would be better if we name the inner type to make it cleaner. E.g.:

```python
# Helper type alias (just guessing a name here, suggestions welcome):
EpochReveals = List[ValidatorIndex, MAX_EARLY_DERIVED_SECRET_REVEALS * SLOTS_PER_EPOCH]

# BeaconState:
exposed_derived_secrets: Vector[EpochReveals, EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS]

# Explicit init in phase0->1 doc:
exposed_derived_secrets: Vector[EpochReveals, EARLY_DERIVED_SECRET_PENALTY_MAX_FUTURE_EPOCHS]
```

Meanwhile I'll try to fix the python-literal case in remerkleable, I think @ericsson49 already raised an issue describing this find, thanks!

ericsson49:
What about `[()] * num` ? it avoids problems with mutability, clear and concise.

PR #1925, opened 2020-06-22
Rework `shard_state_transition` interface + fix #1922

hwwhww:
### Issue
Per #1922, it reminded me that it's not very clear when should we call `verify_shard_block_message`.

I think we can rework the current phase 1 `shard_state_transition` to a similar structure as the phase 0 `state_transition` function now. 

### Proposal

Make `shard_state_transition` function:
1. be similar with phase 0 `state_transition` function
2. be the main shard state transition function for (almost?) all the specs.

Changelog

1. Rename old `shard_state_transition` to `process_shard_block`
2. Add `shard_state_transition` with `validate_message` flag, we only
validate it in shard fork choice
3. Assume that the given block's `shard` field has been validated by p2p layer or shard fork choice rule. Remove `block.shard` verification from `verify_shard_block_message` as #1922 suggested

### DIFF
--- a/specs/phase1/shard-transition.md
+++ b/specs/phase1/shard-transition.md
@@ -61,11 +61,12 @@ def verify_shard_block_signature(beacon_state: BeaconState,
 ## Shard state transition
 
 ```python
-def shard_state_transition(beacon_parent_state: BeaconState,
-                           shard_state: ShardState,
+def shard_state_transition(shard_state: ShardState,
                            block: ShardBlock,
-                           validate_message: bool = True) -> bool:
+                           validate_message: bool = True,
+                           beacon_parent_state: Optional[BeaconState] = None) -> bool:
     if validate_message:
+        assert beacon_parent_state is not None
         assert verify_shard_block_message(beacon_parent_state, shard_state, block)
 
     process_shard_block(shard_state, block)
@@ -126,7 +127,7 @@ def is_valid_fraud_proof(beacon_state: BeaconState,
     else:
         shard_state = transition.shard_states[offset_index - 1]  # Not doing the actual state updates here.
 
-    shard_state_transition(beacon_state, shard_state, block, validate_message=False)
+    shard_state_transition(shard_state, block, validate_message=False)
     if shard_state != transition.shard_states[offset_index]:
         return True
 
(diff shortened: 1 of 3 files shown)


specs/phase1/shard-transition.md
```diff
@@ -64,7 +62,20 @@ def verify_shard_block_signature(beacon_state: BeaconState,
 
 ```python
 def shard_state_transition(shard_state: ShardState,
-                           block: ShardBlock) -> None:
+                           block: ShardBlock,
+                           validate_message: bool = True,
+                           beacon_parent_state: Optional[BeaconState] = None) -> bool:
+    if validate_message:
+        assert beacon_parent_state is not None
```
djrtwo, line 69: Why make `beacon_parent_state` optional?

specs/phase1/shard-transition.md
```diff
@@ -64,7 +62,20 @@ def verify_shard_block_signature(beacon_state: BeaconState,
 
 ```python
 def shard_state_transition(shard_state: ShardState,
-                           block: ShardBlock) -> None:
+                           block: ShardBlock,
+                           validate_message: bool = True,
+                           beacon_parent_state: Optional[BeaconState] = None) -> bool:
```
djrtwo, line 67:
```suggestion
                           beacon_parent_state: Optional[BeaconState] = None) -> ShardState:
```

specs/phase1/shard-transition.md
hwwhww: Because we don't need `beacon_parent_state` if `validate_message` is False. i.e., only need to call `process_shard_block`.

In `get_shard_transition_fields` in v-guide, the `beacon_state` is the beacon head state, *not* the `beacon_parent_state` of the given shard blocks. The beacon parent state can only be obtained from the database store. I think it's better to make it clear that the caller doesn't have to pass the unused `beacon_parent_state` when `validate_message` is False.

specs/phase1/shard-transition.md
```diff
@@ -64,7 +62,20 @@ def verify_shard_block_signature(beacon_state: BeaconState,
 
 ```python
 def shard_state_transition(shard_state: ShardState,
-                           block: ShardBlock) -> None:
+                           block: ShardBlock,
+                           validate_message: bool = True,
+                           beacon_parent_state: Optional[BeaconState] = None) -> ShardState:
+    if validate_message:
+        assert beacon_parent_state is not None
+        assert verify_shard_block_message(beacon_parent_state, shard_state, block)
```
terencechain, line 70: should we add `verify_shard_block_signature` here? or add it back in `on_shard_block`? 

Doesn't look like that's being used anywhere 😅

PR #1926, opened 2020-06-23
AttesterSlashing with 0 indices and out of bounds indices

protolambda:
Based on [this bug find in Nimbus](https://github.com/status-im/nim-beacon-chain/pull/1214/), which was found thanks to the work by the fuzzing team (@zedt3ster and others), it showed how the attester slashing tests missed the high-index case. And cases for 0 attesting indices were missing as well.

We may want the spec to be more explicit about these cases, as the current approach is too implicit (in my opinion). For now, these tests help catch the case in other clients that may not have been given this input through fuzzing yet.

BLS is forced to be enabled in these tests, as that should catch it if not with anything else. But the high index may still cause a panic before that in a client. And the 0 indices case with BLS is important too, since the move away from empty aggregation was based on client BLS issues with empty aggregation there.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/phase0/block_processing/test_process_attester_slashing.py
+++ b/tests/core/pyspec/eth2spec/test/phase0/block_processing/test_process_attester_slashing.py
@@ -200,6 +200,74 @@ def test_participants_already_slashed(spec, state):
 @with_phases([PHASE0])
 @spec_state_test
 @always_bls
+def test_att1_high_index(spec, state):
+    attester_slashing = get_valid_attester_slashing(spec, state, signed_1=True, signed_2=True)
+
+    indices = get_indexed_attestation_participants(spec, attester_slashing.attestation_1)
+    indices.append(spec.ValidatorIndex(len(state.validators)))  # off by 1
+    attester_slashing.attestation_1.attesting_indices = indices
+
+    yield from run_attester_slashing_processing(spec, state, attester_slashing, False)
+
+
+@with_phases([PHASE0])
+@spec_state_test
+@always_bls
+def test_att2_high_index(spec, state):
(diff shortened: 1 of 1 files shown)


tests/core/pyspec/eth2spec/test/phase0/block_processing/test_process_attester_slashing.py
```diff
@@ -197,6 +197,74 @@ def test_participants_already_slashed(spec, state):
...
+    attester_slashing = get_valid_attester_slashing(spec, state, signed_1=True, signed_2=True)
+
+    indices = get_indexed_attestation_participants(spec, attester_slashing.attestation_2)
+    indices.append(spec.ValidatorIndex(len(state.validators)))  # off by 1
+    attester_slashing.attestation_2.attesting_indices = indices
+
+    yield from run_attester_slashing_processing(spec, state, attester_slashing, False)
+
+
+EMPTY_SIGNATURE = b'\xc0' + (b'\x00' * 95)
```
hwwhww, line 226: To avoid confusion between `b'\xc0' + (b'\x00' * 95)` and `b'\x00` * 96, I used `Z2_SIGNATURE` in https://github.com/ethereum/eth2.0-specs/blob/7770accf96f5565ce3d9fc81f7191bae7421189f/tests/core/pyspec/eth2spec/utils/bls.py#L13
p.s. `Z2` is the point at infinity for the twisted curve over FQ2 in py_ecc

Or, it can be renamed to something like `G2_INFINITY_SIGNATURE`.

djrtwo: approved pending deciding what to do here

djrtwo:
lgtm

PR #1928, opened 2020-06-24
Add tests for `shard_state_transition` and some refactorings

hwwhww:
1. Refactor `process_shard_block` function a bit: remove the leftover condition and add some description as phase 0 beacon chain spec.
2. Refactor `phase1/sanity/test_blocks.py` - add `is_full_crosslink` checks
3. Add `phase1/sanity/test_shard_blocks.py` for testing `shard_state_transition`

### DIFF
--- a/specs/phase1/shard-transition.md
+++ b/specs/phase1/shard-transition.md
@@ -66,7 +66,7 @@ def verify_shard_block_signature(beacon_parent_state: BeaconState,
 
 ## Shard state transition function
 
-The post-state corresponding to a pre-state `shard_state` and a signed block `signed_block` is defined as `shard_state_transition(state, signed_block)`. State transitions that trigger an unhandled exception (e.g. a failed `assert` or an out-of-range list access) are considered invalid. State transitions that cause a `uint64` overflow or underflow are also considered invalid.
+The post-state corresponding to a pre-state `shard_state` and a signed block `signed_block` is defined as `shard_state_transition(shard_state, signed_block)`. State transitions that trigger an unhandled exception (e.g. a failed `assert` or an out-of-range list access) are considered invalid. State transitions that cause a `uint64` overflow or underflow are also considered invalid.
 
 ```python
 def shard_state_transition(shard_state: ShardState,
--- a/specs/phase1/shard-transition.md
+++ b/specs/phase1/shard-transition.md
@@ -92,11 +92,8 @@ def process_shard_block(shard_state: ShardState,
     prev_gasprice = shard_state.gasprice
     shard_block_length = len(block.body)
     shard_state.gasprice = compute_updated_gasprice(prev_gasprice, uint64(shard_block_length))
-    if shard_block_length == 0:
-        latest_block_root = shard_state.latest_block_root
-    else:
-        latest_block_root = hash_tree_root(block)
-    shard_state.latest_block_root = latest_block_root
+    if shard_block_length != 0:
+        shard_state.latest_block_root = hash_tree_root(block)
 ```
 
 ## Fraud proofs
--- a/specs/phase1/shard-transition.md
+++ b/specs/phase1/shard-transition.md
@@ -11,7 +11,9 @@
 - [Introduction](#introduction)
 - [Helper functions](#helper-functions)
   - [Shard block verification functions](#shard-block-verification-functions)
-- [Shard state transition](#shard-state-transition)
+    - [`verify_shard_block_message`](#verify_shard_block_message)
+    - [`verify_shard_block_signature`](#verify_shard_block_signature)
+- [Shard state transition function](#shard-state-transition-function)
 - [Fraud proofs](#fraud-proofs)
   - [Verifying the proof](#verifying-the-proof)
 
@@ -24,6 +26,8 @@ This document describes the shard transition function and fraud proofs as part o
 ## Helper functions
 
 ### Shard block verification functions
+
+#### `verify_shard_block_message`
 
 ```python
 def verify_shard_block_message(beacon_parent_state: BeaconState,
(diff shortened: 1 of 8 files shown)


specs/phase1/shard-transition.md
```diff
@@ -58,7 +64,9 @@ def verify_shard_block_signature(beacon_parent_state: BeaconState,
     return bls.Verify(proposer.pubkey, signing_root, signed_block.signature)
 ```
 
-## Shard state transition
+
+The post-state corresponding to a pre-state `shard_state` and a signed block `signed_block` is defined as `shard_state_transition(state, signed_block)`. State transitions that trigger an unhandled exception (e.g. a failed `assert` or an out-of-range list access) are considered invalid. State transitions that cause a `uint64` overflow or underflow are also considered invalid.
```
djrtwo, line 69:
```suggestion
The post-state corresponding to a pre-state `shard_state` and a signed block `signed_block` is defined as `shard_state_transition(shard_state, signed_block)`. State transitions that trigger an unhandled exception (e.g. a failed `assert` or an out-of-range list access) are considered invalid. State transitions that cause a `uint64` overflow or underflow are also considered invalid.
```

tests/core/pyspec/eth2spec/test/phase1/sanity/test_blocks.py
```diff
@@ -40,13 +42,16 @@ def run_beacon_block_with_shard_blocks(spec, state, target_len_offset_slot, comm
...
+
+    if not valid:
+        state_transition_and_sign_block(spec, state, beacon_block, expect_fail=True)
+        yield 'block', beacon_block
         yield 'post', None
         return
+    else:
+        state_transition_and_sign_block(spec, state, beacon_block)
+        yield 'block', beacon_block
+        yield 'post', None
```
djrtwo, line 54:
```suggestion
        yield 'post', state
```

tests/core/pyspec/eth2spec/test/phase1/sanity/test_shard_blocks.py
```diff
@@ -0,0 +1,165 @@
...
+from eth2spec.test.helpers.state import transition_to_valid_shard_slot
+
+
+def run_shard_blocks(spec, shard_state, signed_shard_block,
+                     beacon_parent_state,
+                     validate=True, valid=True):
+    pre_shard_state = shard_state.copy()
+    yield 'pre', pre_shard_state
+    yield 'signed_shard_block', signed_shard_block
+    yield 'validate', validate
```
djrtwo, line 22: I'm not certain we'll have test vectors on the `validate=False`  path. This path is a helper for building blocks but not ever what you'd run when you receive external inputs. Given that the outputs of tests _are_ external inputs and the values of post-transitions, the results (unless no post state) will always be able to validated and thus should be 

In phase 0, there are no test vectors for `validate_results=False`

tests/core/pyspec/eth2spec/test/phase1/sanity/test_shard_blocks.py
```diff
@@ -0,0 +1,165 @@
...
+    yield 'beacon_parent_state', beacon_parent_state
+
+    if validate is False:
```
djrtwo, line 25: This is strange in that it overwrites an already yielded value

tests/core/pyspec/eth2spec/test/phase1/sanity/test_shard_blocks.py
```diff
@@ -0,0 +1,165 @@
...
+        beacon_parent_state = None
+
+    if not valid:
+        expect_assertion_error(lambda: spec.shard_state_transition(
+            shard_state, signed_shard_block, validate=validate, beacon_parent_state=beacon_parent_state)
+        )
+        yield 'post', None
+        return
+
+    spec.shard_state_transition(shard_state, signed_shard_block,
```
djrtwo, line 35: I'm still unsure of the optional `beacon_parent_state` and the optional `validate`  in `shard_state_transition`.  IMO, validations should occur in the state transition (even (and maybe especially) when creating the block). The only validations that should be able to be skipped should be the ones that rely upon "results" or values from the  post-processing (e.g. block signature and state root)

Everything in `verify_shard_block_message` can and should be validated no matter what, and thus `beacon_parent_state` should be required.

We can add a `validate_results` that can enabled/disable sig verification and shard state root check (we don't embed shard state root in the block right now but probably should)

hwwhww: If we want to make `beacon_parent_state` required, we can call `process_shard_block` instead of `shard_state_transition` in [get_shard_transition_fields](https://github.com/ethereum/eth2.0-specs/blob/661959c13ec422cbfb588d40f0cb57fd9ff5c5fa/specs/phase1/validator.md#shard-transition).

((Updating it now))

specs/phase1/shard-transition.md
```diff
@@ -58,16 +64,18 @@ def verify_shard_block_signature(beacon_parent_state: BeaconState,
...
 ```python
 def shard_state_transition(shard_state: ShardState,
                            signed_block: SignedShardBlock,
-                           validate: bool = True,
-                           beacon_parent_state: Optional[BeaconState] = None) -> ShardState:
-    if validate:
-        assert beacon_parent_state is not None
-        assert verify_shard_block_message(beacon_parent_state, shard_state, signed_block.message)
+                           beacon_parent_state: BeaconState,
+                           validate_result: bool = True) -> ShardState:
```
terencechain, line 75:
```suggestion
                           validate_signature: bool = True) -> ShardState:
```

`validate_signature` sounds more appropriate?  
Also curious when we will use `validate_result=False`. Is that more for test setup?

djrtwo: it's for block production. A validator initially makes the block and doesn't have the result(s) so they can't yet be validated. In the case of beaconblocks, the sig and state root (results) are then added. In the case of shard blocks, just the signature (result) is added (though we might consider putting the shard state root in there)

terencechain: Got it. Thanks!

PR #1935, opened 2020-06-25
Proposal: generic `uint_to_bytes`

hwwhww:
### Issue

An alternative to #1746 `int_to_bytes` and `compute_shuffled_index` part.

1. **[Typing presentation bug]** Status-quo `int_to_bytes` expects `uint64` as the input. But actually we were passing Python `int` to it [in some cases](https://github.com/ethereum/eth2.0-specs/blob/ce0371a66b408c86af5305527a9bd2bd253899e0/specs/phase0/beacon-chain.md#compute_shuffled_index).
2. Part of https://github.com/ethereum/eth2.0-specs/pull/1746. Open a new PR to discuss this phase 0 change proposal for decreasing ambiguity. IMHO this PR is neater than #1746's casting.

### Proposed solution
1. To make it more explicit, rename `int_to_bytes` to `uint_to_bytes` and enforce the input of `uint_to_bytes` to be SSZ `uint` type.
3. Encode the `uint` object by its length in bytes. i.e., encode `uint8` to 1-byte output and `uint64` to 8-byte output.
4. To avoid exposing SSZ APIs, describe `uint_to_bytes` in prose instead of in code. Similar to `hash_tree_root`.
5. To make it more explicit, rename `bytes_to_int` to `bytes_to_uint64`. 

**Note**: It's a refactoring change of the current spec and it should not change the consensus logic.

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -578,7 +578,7 @@ def xor(bytes_1: Bytes32, bytes_2: Bytes32) -> Bytes32:
 
 #### `uint_to_bytes`
 
-`def uint_to_bytes(n: uint8) -> bytes` is a function for serializing the `uint` type object to bytes in ``ENDIANNESS``-endian. The expected length of the output is the byte-length of the `uint` type.
+`def uint_to_bytes(n: uint) -> bytes` is a function for serializing the `uint` type object to bytes in ``ENDIANNESS``-endian. The expected length of the output is the byte-length of the `uint` type.
 
 #### `bytes_to_uint64`
 
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -55,8 +55,8 @@
   - [Math](#math)
     - [`integer_squareroot`](#integer_squareroot)
     - [`xor`](#xor)
-    - [`int_to_bytes`](#int_to_bytes)
-    - [`bytes_to_int`](#bytes_to_int)
+    - [`uint_to_bytes`](#uint_to_bytes)
+    - [`bytes_to_uint64`](#bytes_to_uint64)
   - [Crypto](#crypto)
     - [`hash`](#hash)
     - [`hash_tree_root`](#hash_tree_root)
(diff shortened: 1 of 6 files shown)


specs/phase0/beacon-chain.md
```diff
@@ -732,11 +726,15 @@ def compute_shuffled_index(index: uint64, index_count: uint64, seed: Bytes32) ->
...
-        pivot = bytes_to_int(hash(seed + int_to_bytes(current_round, length=1))[0:8]) % index_count
-        flip = (pivot + index_count - index) % index_count
+    for current_round in map(uint8, range(SHUFFLE_ROUND_COUNT)):
+        pivot = bytes_to_uint64(hash(seed + uint_to_bytes(current_round))[0:8]) % index_count
+        flip = uint64((pivot + index_count - index) % index_count)
         position = max(index, flip)
-        source = hash(seed + int_to_bytes(current_round, length=1) + int_to_bytes(position // 256, length=4))
+        source = hash(
+            seed
+            + uint_to_bytes(current_round)
```
djrtwo, line 735: Doing an explicit cast to uint8 right here might result in better readability and fewer client errors

djrtwo: mainly because `current_round` being 8-bit is an exception to our general 64-bit rule and easy to overlook

djrtwo: open to other's thoughts on this

protolambda:
Please review the `type-infer` branch (based on #1746), it handles many of the more "obvious" type inferences, to keep the spec clean. For `compute_shuffled_index` it doesn't involve as many casts anymore, by following a simple set of rules to keep the output size the same as the input size. And it fixes + catches new things.

specs/phase0/beacon-chain.md
```diff
@@ -576,20 +576,14 @@ def xor(bytes_1: Bytes32, bytes_2: Bytes32) -> Bytes32:
...
+#### `uint_to_bytes`
 
-```python
-def int_to_bytes(n: uint64, length: uint64) -> bytes:
-    """
-    Return the ``length``-byte serialization of ``n`` in ``ENDIANNESS``-endian.
-    """
-    return n.to_bytes(length, ENDIANNESS)
-```
+`def uint_to_bytes(n: uint8) -> bytes` is a function for serializing the `uint` type object to bytes in ``ENDIANNESS``-endian. The expected length of the output is the byte-length of the `uint` type.
```
ericsson49, line 581: Looks like a type here, as in `ssz_impl.py` the signature is `def uint_to_bytes(n: uint) -> bytes`
```suggestion
`def uint_to_bytes(n: uint) -> bytes` is a function for serializing the `uint` type object to bytes in ``ENDIANNESS``-endian. The expected length of the output is the byte-length of the `uint` type.
```

PR #1942, opened 2020-06-27
Update `CompactCommittee` List size

terencechain:
Is there's a reason `CompactCommittee` is using `MAX_VALIDATORS_PER_COMMITTEE` for the list size and not `LIGHT_CLIENT_COMMITTEE_SIZE`?  

`get_light_client_committee` will always cap it at `LIGHT_CLIENT_COMMITTEE_SIZE`

I also think Vector makes sense here. Opening this PR for suggestions

### DIFF
--- a/specs/phase1/beacon-chain.md
+++ b/specs/phase1/beacon-chain.md
@@ -404,8 +404,8 @@ class ShardTransition(Container):
 
 ```python
 class CompactCommittee(Container):
-    pubkeys: List[BLSPubkey, MAX_VALIDATORS_PER_COMMITTEE]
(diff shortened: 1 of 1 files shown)


djrtwo:
Although we currently only use `CompactCommittee`s for light client, the idea is that it is a succinct representation of any committee (beacon committee, shard committee, light client committee) so the max list length is that of the committee size

terencechain:
> Although we currently only use `CompactCommittee`s for light client, the idea is that it is a succinct representation of any committee (beacon committee, shard committee, light client committee) so the max list length is that of the committee size

Got it. Thanks for the explanation!

PR #1943, opened 2020-06-30
Avoid negative values and minor refactoring

### LINKED ISSUE #1923: get_custody_atoms: avoid application of modulo operation (%) to a negative value
[get_custody_atoms](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase1/custody-game.md#get_custody_atoms) applies modulo operation to a negative dividend: `(-len(bytez) % BYTES_PER_CUSTODY_ATOM)`.
Modulo operation implementations differ from language to language, when dividend (or divisor) is negative. Therefore that may lead to a bug in a language, which implements modulo operation in a different way than python does (i.e. returning a negative result when a dividend is negative).

@protolambda in a personal communication has suggested that such an ambiguity is better to be avoided in the spec.

hwwhww:
### Issue
Fix #1923

### How did I fix it
- Change `(-len(bytez) % BYTES_PER_CUSTODY_ATOM)` to `((BYTES_PER_CUSTODY_ATOM - (len(bytez) % BYTES_PER_CUSTODY_ATOM)) % BYTES_PER_CUSTODY_ATOM)`
- Avoid negative index
- Some other minor refactoring

p.s. [`state.justification_bits[:-1]`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#justification-and-finalization) will be resolved in #1746 

/cc @ericsson49

### DIFF
--- a/specs/phase1/custody-game.md
+++ b/specs/phase1/custody-game.md
@@ -554,7 +554,6 @@ def process_custody_slashing(state: BeaconState, signed_custody_slashing: Signed
         slash_validator(state, custody_slashing.whistleblower_index)
 ```
 
-
 ## Per-epoch processing
 
 ### Handling of reveal deadlines
@@ -586,7 +585,7 @@ def process_custody_final_updates(state: BeaconState) -> None:
 
     # Reset withdrawable epochs if challenge records are empty
     records = state.custody_chunk_challenge_records
-    validator_indices_in_records = set([record.responder_index for record in records])
+    validator_indices_in_records = set(record.responder_index for record in records)  # non-duplicate
     for index, validator in enumerate(state.validators):
         if validator.exit_epoch != FAR_FUTURE_EPOCH:
             not_all_secrets_are_revealed = validator.all_custody_secrets_revealed_epoch == FAR_FUTURE_EPOCH
--- a/specs/phase1/custody-game.md
+++ b/specs/phase1/custody-game.md
@@ -229,9 +229,11 @@ Given one set of data, return the custody atoms: each atom will be combined with
 
 ```python
 def get_custody_atoms(bytez: bytes) -> Sequence[bytes]:
-    bytez += b'\x00' * (-len(bytez) % BYTES_PER_CUSTODY_ATOM)  # right-padding
-    return [bytez[i:i + BYTES_PER_CUSTODY_ATOM]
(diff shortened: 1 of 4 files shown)


specs/phase1/custody-game.md
```diff
@@ -229,9 +229,11 @@ Given one set of data, return the custody atoms: each atom will be combined with
 
 ```python
 def get_custody_atoms(bytez: bytes) -> Sequence[bytes]:
-    bytez += b'\x00' * (-len(bytez) % BYTES_PER_CUSTODY_ATOM)  # right-padding
-    return [bytez[i:i + BYTES_PER_CUSTODY_ATOM]
-            for i in range(0, len(bytez), BYTES_PER_CUSTODY_ATOM)]
+    bytez += b'\x00' * ((BYTES_PER_CUSTODY_ATOM - len(bytez)) % BYTES_PER_CUSTODY_ATOM)  # right-padding
```
ericsson49, line 232: it's possible that `len(bytez) > BYTES_PER_CUSTODY_ATOM`, so `(BYTES_PER_CUSTODY_ATOM - len(bytez))` can become negative

hwwhww: Ah you're right. Hmm so it can be:
```suggestion
    length_remainder = len(bytez) % BYTES_PER_CUSTODY_ATOM
    bytez += b'\x00' * ((BYTES_PER_CUSTODY_ATOM - length_remainder) % BYTES_PER_CUSTODY_ATOM)  # right-padding
```

ericsson49: That looks correct!
I wrote a small test and it works fine.
```python
from eth2spec.phase1.spec import BYTES_PER_CUSTODY_ATOM

for i in range(0, BYTES_PER_CUSTODY_ATOM*2+1):
    bytez = b'\0' * i
    length_remainder = len(bytez) % BYTES_PER_CUSTODY_ATOM
    bytez1 = bytez + (b'\x00' * ((BYTES_PER_CUSTODY_ATOM - length_remainder) % BYTES_PER_CUSTODY_ATOM))
    bytez2 = bytez + (b'\x00' * (-len(bytez) % BYTES_PER_CUSTODY_ATOM))
    assert bytez1 == bytez2
```

PR #1946, opened 2020-07-01
Fix `is_valid_fraud_proof` which tries to access `message` field of a `ShardBlock` instance

ericsson49:
Though, the `message` field belongs to `SignedShardBlock`.
Found during py-specs transpilation.


djrtwo:
Fix looks good! but the build is failing in seemingly an entirely unrelated way.
I'm investigating

ericsson49:
Yay! It's green now.
Thanx, @djrtwo

PR #1956, opened 2020-07-07
add queueing possibility to p2p messages in gossip

djrtwo:
* MAY queue block if parent not yet seen
* MUST reject block is parent not valid
* MAY queue attestation if voted block not yet seen
* MAY queue aggregate if voted block not yet seen

only do the potential queuing/lookup conditions after sig verification to ensure it's not an open DoS vector

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -263,6 +263,8 @@ Clients MUST reject (fail validation) messages containing an incorrect type, or 
 
 When processing incoming gossip, clients MAY descore or disconnect peers who fail to observe these constraints.
 
+For any optional queueing, clients SHOULD maintain maximum queue sizes to avoid DoS vectors.
+
 Gossipsub v1.1 introduces [Extended Validators](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#extended-validators)
 for the application to aid in the gossipsub peer-scoring scheme.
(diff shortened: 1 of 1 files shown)


specs/phase0/p2p-interface.md
```diff
@@ -310,6 +315,8 @@ The following validations MUST pass before forwarding the `signed_aggregate_and_
   (via aggregate gossip, within a verified block, or through the creation of an equivalent aggregate locally).
 - _[IGNORE]_ The `aggregate` is the first valid aggregate received for the aggregator
   with index `aggregate_and_proof.aggregator_index` for the epoch `aggregate.data.target.epoch`.
+- _[IGNORE]]_ The block being voted for (`aggregate.data.beacon_block_root`) has been seen locally
```
protolambda, line 318: The message ID and seen-cache already does this for us. Except when the cache overflows (It's a large LRU in Rust and Go at least), and the old entry may disappear. Probably fine to keep it here, hope it's not too difficult to have a 2nd "has this been seen" lookup for clients.

djrtwo: there's a chance you didn't see the block via gossip though. Maybe retrieved in a blocks-by-root request triggered by something else

protolambda: Good point about sync and other sources. However, as discussed offline, restricting it to forkchoice contents may miss other edge-case seen-block cases. Changing it to something like "any seen blocks (including from non-gossip sources)" should work.

PR #1957, opened 2020-07-07
Enable test generation for phase1

protolambda:
- Updates SSZ static test gen to deal with phase 1 types too. Needed a minor fix in random-output generation.
- Update epoch processing, block operations, sanity, rewards test gen to:
  - Output for phase0 like before
  - Output for phase-1 specific tests
  - Run phase0 module tests, with phase 1 spec. (Incompatible tests are handled by the test decorators)
- Rewards and penalties mainnet epoch processing was still disabled, partially because we have rewards tests, partially because it used to be very slow. Let's try to enable them.
- Rewards tests are now split into `random/basic/leak`, instead of all in `core`. We can change that back, but it seemed more consistent
- Added trace-backs to test gen errors, to make debugging easier, as the error message is sometimes empty (e.g. asserts).
- Unsure about phase-1 "genesis" tests. To start immediately into a phase 1 state? Should that be a thing to test clients for?

Before merge:
- Some block-operations have errors in phase 1 mode. It didn't get past attester slashing tests for me. If someone can help look into this, that would be nice.
- I temporarily commented mainnet mode tests, to re-run test gen faster to check if it works. Since mainnet is much slower. Need to uncomment them.

CC @mkalinin @djrtwo @hwwhww Let me know if you like to change anything for phase 1 test generation.


mkalinin:
> Unsure about phase-1 "genesis" tests. To start immediately into a phase 1 state? Should that be a thing to test clients for?

It might be relevant for testnets or simulations that are starting from `GENESIS_SLOT` right away, i.e. with `PHASE_1_GENESIS_SLOT = GENESIS_SLOT`

protolambda:
@mkalinin Sounds good, but we'll need the spec to define what that phase-1 genesis looks like (I see you are working on that and shared details in the discord, thanks). Right now I'm afraid a bare phase-0 genesis test that runs with phase-1 spec will break other phase-1 spec assumptions.

mkalinin:
> we'll need the spec to define what that phase-1 genesis looks like

Totally agree. I currently use pretty naive way to get phase-1 genesis:
```python
genesis_state = initialize_beacon_state_from_eth1(eth1_block_hash, eth1_timestamp, deposits)
phase_1_genesis_state = upgrade_to_phase1(genesis_state)
```

But this approach works well in my case.

hwwhww:
> Some block-operations have errors in phase 1 mode. It didn't get past attester slashing tests for me. If someone can help look into this, that would be nice.

The error message, for example:
```
generating test vectors from tests source: eth2spec.test.phase0.block_processing.test_process_attester_slashing
Generating test: output/minimal/phase1/operations/attester_slashing/pyspec_tests/all_empty_indices
ERROR: failed to generate vector(s) for test output/minimal/phase1/operations/attester_slashing/pyspec_tests/all_empty_indices: 'NoneType' object is not iterable
  File "/Users/hwwang/.pyenv/versions/env38a/lib/python3.8/site-packages/gen_base/gen_runner.py", line 138, in run_generator
    for (name, out_kind, data) in test_case.case_fn():
TypeError: 'NoneType' object is not iterabl
```

- It's because we set these tests to `with_phases([PHASE0])` since phase1 `AttestationData` format **was** different from phase0. And thanks to #1889, these tests can be enabled now. (I pushed https://github.com/ethereum/eth2.0-specs/pull/1957/commits/60f4bd228865258965010ed734c409f9e42284c5)

- However, we still have some other test cases for test generators with `with_phases([PHASE0])`.

- Also, we have some phase1 test cases that *only work with minimal config* since we need enough validators to form crosslink for every slot for every slot in these test cases ([example](https://github.com/ethereum/eth2.0-specs/blob/7b43a3d7724f9b51c373eefe7c047a01c8e55634/tests/core/pyspec/eth2spec/test/phase1/sanity/test_blocks.py#L100-L106)).

A quick workaround is skipping the test case if `test_case.case_fn()` returns `None` in `gen_runner.py`.
@protolambda any thought? :)

protolambda:
@hwwhww great work on all the phase 1 changes. Testing PRs were merged into this PR, and I just resolved a small merge conflict with `dev`. Is this ready to be merged now?

hwwhww (approved):
@protolambda 
I pushed some commits:
1. Remove redundant keys in the rewards test generator.
2. Use `Dict` for `phase_0_mods`, `phase_1_mods`  instead of `List`.
3. Add `make lint_generators` command for test generators flake8 checks. Also added it to the CI.


LGTM now. :)

hwwhww:
@djrtwo 

Changes to the phase 0 test vectors outputs:
1. [`ssz`] fix in random-output generation.
2. [`rewards`] Rewards tests are now split into `random/basic/leak`, instead of all in `core`. 
3. [`sanity/block`] Adds a `reveal_deadlines_setting `field to the sanity/block `meta.yaml`.
4. [Config files]: add `CONFIG_NAME` meta field to distinguish between "mainnet" and "minimal" configs.

djrtwo (approved):

prestonvanloon:
Was it intentional to remove `eth2.0-spec-tests/tests/mainnet/phase0/ssz_static`?

protolambda:
No, thanks for reporting, just heard from Terence too. It was meant to be disabled just for phase1 mainnet I think (and at first all mainnet tests were disabled, to testrun the changes in this PR quicker). We'll re-introduce them. The types have not changed, so you can keep running the v0.12.2 SSZ tests. Sorry for the inconvenience.

PR #1958, opened 2020-07-07
Match gossip v1.1 D_low, extend gossip_history param, add FAQ section

protolambda:
This PR updates the gossip sub parameters in the spec:
- Match the `D_low` default recommend parameter [which changed in v1.1](https://github.com/libp2p/go-libp2p-pubsub/pull/263)
- rename `gossip_history`  to `mcache_len` to match param names in spec
- rename `gossip_advertise`  to `mcache_gossip` to match param names in spec
- add `seen_ttl` as seen in spec
- Change `heartbeat_interval` from 1s to 0.7s as per the recommendation of [v1.1 PL report](https://gateway.ipfs.io/ipfs/QmRAFP5DBnvNjdYSbWhEhVRJJDFCLpPyvew5GwCCB4VxM4) to allow for more responsiveness in the event that gossip is not performing optimally
- Change `mcache_len` from `5` to `6` to allow for at least ~2s for IWANT queries after IHAVEs have been sent. This it to match original time window but to compensate or our faster `heartbeat_interval` 
- Set `seen_ttl` to `550` to match the length of an epoch (contextualized by `heartbeat_interval`)
- Add a FAQ section to the spec, explaining the parameter choices.

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -1160,7 +1160,6 @@ Some examples of where messages could be duplicated:
   enough for `IWANT`s to respond to `IHAVE`s in the context of the shorter
   `heartbeat_interval`. If `mcache_gossip` is increased, this param should be
   increased to be at least `3` (~2 seconds) more than `mcache_gossip`.
-  should be increased to be at least 
 - `mcache_gossip`: 3, recommended default. This can be increased to 5 or 6
   (~4 seconds) if gossip times are longer than expected and the current window
   does not provide enough responsiveness during adverse conditions.
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -202,7 +202,7 @@ and will in most cases be out of sync with the ENR sequence number.
 
 ## The gossip domain: gossipsub
 
-Clients MUST support the [gossipsub v1](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub) libp2p Protocol
+Clients MUST support the [gossipsub v1](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md) libp2p Protocol
 including the [gossipsub v1.1](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md) extension.
 
 **Protocol ID:** `/meshsub/1.1.0`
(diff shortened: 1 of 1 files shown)


djrtwo:
I think we should define `gossip_history` as `1 + SLOTS_PER_EPOCH * SECONDS_PER_SLOT / heartbeat_interval` to default just cover the length of time we want (an epoch) as a function of those params.

Considering dropping heartbeat_interval to 700ms as per the recommendation of the [gossipsub v1.1 report](https://gateway.ipfs.io/ipfs/QmRAFP5DBnvNjdYSbWhEhVRJJDFCLpPyvew5GwCCB4VxM4) (page 6)

djrtwo:
We've also conflated `seen_ttl` and `mcache_len` from the [gossipsub spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.0.md#parameters) into just one param here -- `gossip_history`

arguably `seen_ttl` shuld be the 385 as recommended above and `mcache_len`/`mcache_gossip` should be considered independently

 I'm not sure the exact trade-off at play when increasing/decreasing `mcache` params. `mcache_len` keeps more _full_ messages in cache to be able to serve to other peers when they send `IWANT` requests. `mcache_gossip` dictates how many history windows you mention messages in `IHAVE` gossip.

Writing some quick notes on the intuition on `mcache`:

If we increase `mcache_len` far beyond `mcache_gossip`, the old items in the cache will not frequently be used because you'll quickly not be telling peers about the old messages so they won't know to ask for them.

If we increase both `mcache_len` and `mcache_gossip` too much, we'll overly redundantly gossiping via `IHAVE` about stale messages that everyone already has. This presumably would make each `IHAVE` message larger for little gain because it would include much older messages.

Also seems to make sense to have `mcache_len` be slightly larger than `mcache_gossip` so the nodes that have been gossiped to (via `IHAVE`) have a chance to request missing messages before you remove them from the cache.

lsankar4033:
Re:`fanout_ttl` - subnet_ids are re-used every subsequent epoch, so waiting an epoch for ttl might result in publishing attestations to peers in the previous epoch's committee of the same id.

This isn't breaking obviously but does result in spam; might be an argument to keep the ttl closer to 1-2 slots.

specs/phase0/p2p-interface.md
```diff
@@ -1146,6 +1148,26 @@ Some examples of where messages could be duplicated:
...
+- `heartbeat_interval`: 0.7 seconds, recommended for eth2 in the [GossipSub evaluation report by Protocol Labs](https://gateway.ipfs.io/ipfs/QmRAFP5DBnvNjdYSbWhEhVRJJDFCLpPyvew5GwCCB4VxM4).
+- `fanout_ttl`: 60 seconds, recommended default.
+  Fanout is primarily used by committees publishing attestations to subnets.
+  This happens once per epoch per validator and the subnet changes each epoch
+  so there is little to gain in having a `fanout_ttl` be increased from the recommended default.
+- `mcache_len`: 6, increase by one to ensure that mcache is around for long
+  enough for `IWANT`s to respond to `IHAVE`s in the context of the shorter
+  `heartbeat_interval`. If `mcache_gossip` is increased, this param should be
+  increased to be at least `3` (~2 seconds) more than `mcache_gossip`.
+  should be increased to be at least
```
protolambda, line 1163:
```suggestion
```

Formatting

PR #1966, opened 2020-07-13
include message in goodbye

arnetheduck:
Add a descriptive message to goodbye, to describe disconnection reason
for debugging purposes. Clients may choose to leave it empty, or not
send a goodbye at all.

This has been discussed in chat and in calls - generally, it is redundant but may help in diagnosing testnet issues.

A more complex version has been proposed with  error codes for specific scenarios and accompanying expected behaviours and interactions, but the benefits are unclear, specially when considering the worse case. Generally, if peers are misbehaving, they should likely not be sent a goodbye message at all, as this wastes resources and may further exacerbate the problem.

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -655,14 +655,16 @@ Implementers are free to implement such behavior in their own way.
 
 #### Goodbye
 
-**Protocol ID:** ``/eth2/beacon_chain/req/goodbye/1/``
+**Protocol ID:** ``/eth2/beacon_chain/req/goodbye/2/``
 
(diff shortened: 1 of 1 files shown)


djrtwo:
I see you made it `goodbye` v2.

How do you propose we handle multi-versions to start? if both v2, use v2; else v1?

arnetheduck:
It doesn't _really_ matter, this is a bit of an optional message - ie nothing prevents a client from just.. disconnecting. That said, clients are free to provide an overlap period should this be deemed useful.

djrtwo:
closing this for now

PR #1967, opened 2020-07-14
[WIP] Use GHOST (without LMD) for shard chain fork choice rule

hwwhww:
### Issues
1. The shard chain fork choice logic should be separate from different shards.
2. In the previous offline discussion, it probably would be simpler to use pure GHOST (without LMD) for share chains.

### This WIP PR
1. `Store` and `ShardStore` schema update.
    - To distinguish the overlapping aggregated attestations, i.e., the two attestations that include the same attester, I use a dict `ShardStore.block_attesting_indices: Dict[Root, Set[ValidatorIndex]]` to record the attesters that voted for the certain shard block root. **TBD**: any other better solution?
    - Since we have to access `ShardStore` in `on_attestation` but I also want to avoid overhaul the existing function APIs and test cases. This PR move `shard_stores: Dict[Shard, ShardStore]` to be one of the fields of `Store`.
2. Shard fork choice rule changes
    - Basic GHOST logic.
    - Allow double count the same attester that voted for shard 10 in epoch  0 and then randomly get reassigned to shard 10 in epoch 1.

### TBD
1. Is there a better solution to avoid storing `ShardStore.block_attesting_indices: Dict[Root, Set[ValidatorIndex]]` for distinguishing the overlapping aggregated attestations? If not, using LMD-GHOST for shard fork choice rule (https://github.com/ethereum/eth2.0-specs/commit/43ef9aa294214fd89e53430db5ccb234aa71d8c7) seems simpler.
2. Double count
    1. Is it okay to double count the attester who voted for the ancestor and the descendant?
    2. Is it okay to double count the attester who voted for different forks? I guess it's okay because we will have attester slashing.


hwwhww:
@djrtwo pointed out that clients have built out LMD-GHOST engines already, so pure GHOST doesn't make it simpler since clients have to implement more new logic.

Closing this version now and I will open bugfix-only PR later.

Issue #1768, opened 2020-04-28
Phase 1 known issues checklist


hwwhww:
## A. Phase 1 beacon chain for legacy
- [ ] 1. Possible beacon chain changes for phase 1  #1825 
- [ ] 2. Exit fee [#1521](https://github.com/ethereum/eth2.0-specs/issues/1521)
- [ ] 3. Maximum reversions: Enshrine dynamic weak subjectivity reversion period into fork choice rule ("second finality mechanism"): [#1520](https://github.com/ethereum/eth2.0-specs/issues/1520) and [#1053](https://github.com/ethereum/eth2.0-specs/issues/1053)

## B. Phase 1 beacon chain for shards
- [ ] 1. Shard chain slashing
- [ ] 2. Rewards on shards [#1280](https://github.com/ethereum/eth2.0-specs/issues/1520) and [#1053](https://github.com/ethereum/eth2.0-specs/issues/1280)
- [x] 3. Implement `get_start_shard` logic (#1858)
- [ ] 4. Slowly rotating shard committees for stabilizing subnets
- [x] 5. Add `Shard` field to `AttestationData`
- [x] 6. Use non-length mixins root for `shard_data_roots`
- [ ] 7. Decide how to change active shard count (#2033)

## C. Phase 1 shard chain
- [x] 1. **[shard chain fork choice rule]** move `get_winning_proposal` logic to it. Maintain canonical shard chain upon (i) receiving new shard block (ii) new beacon chain block (ii) time tick. (#1773) 
- [x] 2. **[shard chain transition]** Add `shard_transition_digest` wrapper. Currently we have `hash(hash_tree_root(shard_state) + beacon_parent_root + shard_body_root)`
- [ ] 3. **[data availability]** Inefficiencies in data availability proofs [#1194](https://github.com/ethereum/eth2.0-specs/issues/1194) (outdated???)
- [ ] 4. **[data availability]** Generating data availability roots from shard block roots [#1438](https://github.com/ethereum/eth2.0-specs/issues/1438)
- [x] 5. [0.001 bit proof of custody](https://ethresear.ch/t/a-0-001-bit-proof-of-custody/7409) (#1889)

## D. Phase 1 fork choice rule
- [ ] 1. Separate beacon and shard logic
- [ ] 2. Use GHOST instead of LMD-GHOST for shard fork choice rule

terencechain:
[...]

PR #1970, opened 2020-07-15
Fix shard fork choice

hwwhww:
### Issue
The shard chain fork choice logic should be separate from different shards. (#1768 D.1)

### How did I fix it
1. fork-choice and shard-fork-choice:
    1. Create `ShardLatestMessage` dataclass to store the latest message of per shard.
    2. Add `latest_messages: Dict[ValidatorIndex, ShardLatestMessage]` field to `ShardStore`.
    3. `Store` schema update.
        - Since we have to access `ShardStore` in `on_attestation` but I also want to avoid overhaul the existing function APIs and test cases. This PR move `shard_stores: Dict[Shard, ShardStore]` to be one of the fields of `Store`.
    4. Use shard `latest_message` to calculate the shard latest attesting balance.
2. `setup.py`: handle the `dataclass` classes dependencies.


specs/phase1/shard-fork-choice.md
```diff
@@ -145,9 +142,9 @@ def get_pending_shard_blocks(store: Store, shard_store: ShardStore) -> Sequence[
 #### `on_shard_block`
 
 ```python
-def on_shard_block(store: Store, shard_store: ShardStore, signed_shard_block: SignedShardBlock) -> None:
+def on_shard_block(store: Store, shard: Shard, signed_shard_block: SignedShardBlock) -> None:
```
terencechain, line 145: I wonder if we should delete `assert shard_block.shard == shard` and simplify arguments to 

`def on_shard_block(store: Store, signed_shard_block: SignedShardBlock) -> None:`

hwwhww:
The BLS-enabled tests require #1973.

specs/phase1/shard-fork-choice.md
djrtwo: agreed
We should just derive the `shard` directly from the `shard_block` and use the appropriate `shard_store` from the `store`

djrtwo:
looks good other than merge conflict

djrtwo:
This is still marked draft.
Any blockers for un-drafting and merging @hwwhww ?

hwwhww:
The test would not pass if BLS flag is enabled, so kinda depending on #1980 under the surface.

djrtwo:
#1980 merged

PR #1971, opened 2020-07-16
Rework initial shard slot (`PHASE_1_FORK_SLOT`)

### LINKED ISSUE #1877: Initial shard slot number
### Issue
Currently, the initial shard state slot is set to `PHASE_1_GENESIS_SLOT` by [setting shard_state.slot = pre.slot](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase1/phase1-fork.md#upgrading-the-state)

For example, if `PHASE_1_GENESIS_SLOT` is 32:
image

The shard block 32 cannot exist and both B_32 and B_33 point to initial shard state.

### How to fix it

IMO the scenario would be neater if we set initial shard slot to `compute_previous_slot(pre.slot)`.
image

hwwhww:
### Issue
Fix #1877 

### How did I fix it
1. Rename `PHASE_1_GENESIS_SLOT` to `PHASE_1_FORK_SLOT`: IMHO it's clearer. (Sorry!)
1. Set initial `shard_state.slot` to `compute_previous_slot(pre.slot)`, where `pre` is the beacon state.
2. Per @mkalinin's suggestion, make `PHASE_1_FORK_SLOT = GENESIS_SLOT = 0` for the test vectors generation.
3. Related to #1957 - now we can enable more phase 0 tests to be phase-1-compatible!

### TODO
Since it makes `PHASE_1_FORK_SLOT = GENESIS_SLOT = 0`, we have to add fork boundary tests with a new config with a reasonable value, e.g., `PHASE_1_FORK_SLOT := Slot(256)`. Leave it to future PRs.


specs/phase1/beacon-chain.md
```diff
@@ -760,20 +760,24 @@ def validate_attestation(state: BeaconState, attestation: Attestation) -> None:
...
         assert data.beacon_block_root == get_block_root_at_slot(state, compute_previous_slot(state.slot))
         # Correct shard number
-        shard = compute_shard_from_committee_index(state, attestation.data.index, attestation.data.slot)
-        assert attestation.data.shard == shard
-        # On-time attestations should have a non-empty shard transition root
-        assert attestation.data.shard_transition_root != hash_tree_root(ShardTransition())
+        shard = compute_shard_from_committee_index(state, data.index, data.slot)
+        assert data.shard == shard
+        # NOTE: We currently set `PHASE_1_FORK_SLOT` to `GENESIS_SLOT` for test vectors.
+        if data.slot > GENESIS_SLOT:
```
djrtwo, line 776: should this be `PHASE_1_FORK_SLOT`? In the wild, `PHASE_1_FORK_SLOT != GENESIS_SLOT` -- at least on mainnet

djrtwo: Ah, I see what's going on here now with the initialization of `shard_states`

tests/core/pyspec/eth2spec/test/helpers/attestations.py
```diff
@@ -79,23 +79,27 @@ def build_attestation_data(spec, state, slot, index, shard=None, shard_transitio
...
+            data.shard_head_root = shard_transition.shard_states[last_offset_index].latest_block_root
+            data.shard_transition_root = shard_transition.hash_tree_root()
         else:
             if on_time:
-                shard_transition = spec.get_shard_transition(state, shard, shard_blocks=[])
-                last_offset_index = len(shard_transition.shard_data_roots) - 1
-                attestation_data.shard_head_root = shard_transition.shard_states[last_offset_index].latest_block_root
-                attestation_data.shard_transition_root = shard_transition.hash_tree_root()
+                if data.slot == spec.GENESIS_SLOT:
+                    data.shard_head_root = spec.Root()
```
djrtwo, line 92: Does this map to what is in the validator spec?

val spec says "Set `attestation_data.shard_head_root = hash_tree_root(shard_head_block)`." and doesn't have a conditional on `GENESIS_SLOT`

djrtwo: So this does look correct, but I think we need to mod the validator spec to include this case

hwwhww: Nice catch! I updated the v-guide with this special rule: https://github.com/ethereum/eth2.0-specs/pull/1971/commits/cd8bb3d85a011225ea78043cf6f47d86ebff2358

hwwhww:
Note that with this PR:

1. The `PHASE_1_FORK_SLOT == Slot(128)` case is like:

phase_1_fork_slot_128

2. the `PHASE_1_FORK_SLOT == GENESIS_SLOT` case is like:

phase_1_fork_slot_0

The shard block 0 doesn't exist in this case so `transition_to_valid_shard_slot` helper is for forwarding to `Slot(1)`.

Issue #1973, opened 2020-07-17
`ShardBlockHeader.body_root` is generated without the length mix-in


hwwhww:
### Issue
Just found that some BLS-enabled tests are broken. (The CI is running with `--disable-bls` flag) Some issues showed up after I enabled BLS.

####  
(Leftover of #1937) Since `ShardTransition.shard_data_roots` is gerenated with `get_block_data_merkle_root`, `ShardBlockHeader.body_root` should also be generated with `get_block_data_merkle_root`. Therefore, the signing roots of `DOMAIN_SHARD_PROPOSAL` should be generated with the header includes this shard block body root.

#1975 shows where we should use `ShardBlockHeader` to generate `signing_root`

Note that it would make  **`hash_tree_root(shard_block.message) != hash_tree_root(shard_block_header)`**. If it is correct to do in this way, the question is: should we use `hash_tree_root(shard_block.message)` or `hash_tree_root(shard_block_header)` as **the shard block root**?

Shard block root is used as an identifier in many places. We may want to avoid having such a special case in the protocol. I wonder if there is a better solution to fix this trap. 🤔 

/cc @djrtwo @dankrad

djrtwo:
I think we need to preserve the invariant that `hash_tree_root(signed_shard_block.message) == hash_tree_root(shard_block_header)`.

Can we do this by having a special SSZ list type that doesn't use mix-in?

Or maybe we should use a `Vector` here inside the consensus object but send a network version of the block that is type `list` (for bandwidth savings in the event that blocks aren't full) <-- or just rely on decent compression from snappy

cc: @protolambda

protolambda:
> The CI is running with --disable-bls flag

We have a Milagro option now, maybe we should just enable it (and wait a little extra, unless phase1 testing is taking too long?)

> Can we do this by having a special SSZ list type that doesn't use mix-in?

We definitely could. I am not sure if we should.

> Or maybe we should use a Vector here inside the consensus object but send a network version

I prefer a new ssz type over divergence of the typing between the layers.

I'll go through the spec in more detail and see if there's another option.

dankrad:
Changing it to List will not be a problem other than that some Merkle proofs are going to be 32 bytes larger. No problem for the proof of custody or MPC-friendliness.

hwwhww:
closing via #1980

PR #1975, opened 2020-07-20
Use `ShardBlockHeader` hash tree root as the shard block root

hwwhww:
Address #1973

This PR adds `compute_shard_block_header_root` helper to compute the shard block root with `ShardBlockHeader`.


hwwhww:
Replaced by #1980

PR #1980, opened 2020-07-22
Use `hash_tree_root` to generate shard block body root

hwwhww:
### Issue
Address #1973 and replace #1975 

https://github.com/ethereum/eth2.0-specs/issues/1973#issuecomment-661911433: No problem for the proof of custody or MPC-friendliness.

### How did I fix it
- Use `hash_tree_root` to generate the root instead of `get_block_data_merkle_root`
- Include the mix-in length into the proof (similar to deposit data Merkle proof)


tests/core/pyspec/eth2spec/test/helpers/custody.py
```diff
@@ -149,12 +149,14 @@ def get_valid_custody_chunk_response(spec, state, chunk_challenge, challenge_ind
 
     chunk_index = chunk_challenge.chunk_index
 
-    data_branch = build_proof(custody_data_block.get_backing().get_left(), chunk_index + 2**spec.CUSTODY_RESPONSE_DEPTH)
+    leaf_index = chunk_index + 2**spec.CUSTODY_RESPONSE_DEPTH
+    serialized_length = (len(custody_data_block)).to_bytes(32, 'little')
+    data_branch = build_proof(custody_data_block.get_backing().get_left(), leaf_index) + [serialized_length]
```
djrtwo, line 154: Do we still need the backing from remerkleable?

What exactly does `get_left()` give you here?

djrtwo: In the validator guide, I think we should be explicit and not use underlying library structure functions.
As for the spec tests, I'm happy to use the remerkleable backings because they are probably much more efficient

PR #1988, opened 2020-07-23
add deposit chain id and network id to config

djrtwo:
add `DEPOSIT_CHAIN_ID` and `DEPOSIT_NETWORK_ID` to be able to define the eth1 network where the deposit contract lives. Up to this point, it's been implicit (and generally georli), but we need to be able to make a distinction as we move toward having both eth2 mainnet and testnets

### DIFF
--- a/specs/phase0/deposit-contract.md
+++ b/specs/phase0/deposit-contract.md
@@ -31,12 +31,14 @@ This document represents the specification for the beacon chain deposit contract
 
 | Name | Value |
 | - | - |
+| `DEPOSIT_CHAIN_ID` | `1` |
+| `DEPOSIT_NETWORK_ID` | `1` |
 | `DEPOSIT_CONTRACT_ADDRESS` | **TBD** |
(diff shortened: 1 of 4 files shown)


protolambda:
Names are mixed up, it should not be 2x chain ID.

djrtwo:
was just messed up in header comment for the PR. fixed!

specs/phase0/deposit-contract.md
```diff
@@ -31,12 +31,16 @@ This document represents the specification for the beacon chain deposit contract
 
 | Name | Value |
 | - | - |
+| `DEPOSIT_CHAIN_ID` | `1` |
+| `DEPOSIT_NETWORK_ID` | `1` |
```
hwwhww, line 35: Very nitpicking:
This table is called "Contract". Maybe move `DEPOSIT_CHAIN_ID` and `DEPOSIT_CHAIN_ID` to a "Chain" table or remove the markdown header "Contract".

djrtwo: made a "constants" and "config" section. should be good now

PR #1993, opened 2020-07-23
Finality test generation PR, but with BLS enabled

protolambda:
- Enable BLS in finality tests
- Use Milagro BLS in state-transition tests, makes test-generation faster, more bearable
- Remove old hash monkey-patch: a long time ago it used to help with SSZ speed, but right now it's only doing some caching within the shuffling function. However, the shuffling is already cached, so it doesn't help much. At the same time, 100.000+ hashes are made during test generation of large shuffling lists, putting them all in the hash cache (since they are mostly unique) should be avoided.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -141,11 +141,6 @@ def ceillog2(x: uint64) -> int:
     return (x - 1).bit_length()
 '''
 PHASE0_SUNDRY_FUNCTIONS = '''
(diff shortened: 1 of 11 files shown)


protolambda:
Hmm, removing the cache changed the output type, making some tests fail. Fixing that now.

PR #1996, opened 2020-07-24
process_attestation: Validate epoch before using it

arnetheduck:
`data.target.epoch` is used to count the active validator set.

Because `get_committee_count_per_slot` is extremely inefficient the way
the spec is written, clients cache it, or the underlying active
validator set.

Performing the checks in the given order leads to a (very unlikely)
security issue where the a cold and above all, distant value may get
used which may be costly - reordering the checks brings the value into a
more reasonable range before using it.


arnetheduck:
Notably, this is predominantly an issue for implementations - the outcome, even if processed in the order given by the spec, remains the same.

PR #1999, opened 2020-07-28
add clarifying comment for genesis skip conditions

djrtwo:
Thanks @adaszko and @michaelsproul for pointing out that the finality genesis skip condition is not obvious and warrants a comment


specs/phase0/beacon-chain.md
```diff
@@ -1512,6 +1514,7 @@ def get_attestation_deltas(state: BeaconState) -> Tuple[Sequence[Gwei], Sequence
 
 ```python
+    # No rewards are applied at end of `GENESIS_EPOCH` because rewards are for work done in the previous epoch
```
hwwhww, line 1517: Ridiculously nitpicking 
```suggestion
    # No rewards are applied at the end of `GENESIS_EPOCH` because rewards are for work done in the previous epoch
```

PR #2001, opened 2020-07-29
Add self-consistency check to attestation gossip validation

lsankar4033:
Addresses https://github.com/ethereum/eth2.0-specs/issues/1998 and #2001

Namely, adds a `REJECT` stipulation if attestation's target epoch differs from that of the attestation itself, and `REJECT` condition if the attestations target root is not in the chain defined by the LMD vote

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -403,6 +403,8 @@ The following validations MUST pass before forwarding the `attestation` on the s
 - _[REJECT]_ The current `finalized_checkpoint` is an ancestor of the `block` defined by `attestation.data.beacon_block_root` -- i.e.
   `get_ancestor(store, attestation.data.beacon_block_root, compute_start_slot_at_epoch(store.finalized_checkpoint.epoch))
   == store.finalized_checkpoint.root`
+- _[REJECT]_ The attestation's epoch matches its target -- i.e. `attestation.data.target.epoch ==
+  compute_epoch_at_slot(attestation.data.slot)`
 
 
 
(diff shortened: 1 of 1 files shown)


djrtwo:
@lsankar4033 I merged #2005 with this PR and also arranged the slot/epoch condition to be earlier in the sequence as it is a very cheap consistency check

Can you take a look before merge?

lsankar4033:
> @lsankar4033 I merged #2005 with this PR and also arranged the slot/epoch condition to be earlier in the sequence as it is a very cheap consistency check
> 
> Can you take a look before merge?

ordering lgtm!

paulhauner:
The second check added in this PR allows the case where `get_block(attestation.data.target.root).epoch > attestation.data.target.epoch`. These attestations are declared invalid by fork choice but permitted in blocks (I think, would need to double check).

I'm not convinced this is an immediate problem, but I though it might be useful to raise it.

In Lighthouse, we've had to check for the case where the head block is from a later epoch and then explicitly *allow* it: https://github.com/sigp/lighthouse/blob/62fad43673d6f1e9e625098a4b18c9d5e581d024/beacon_node/beacon_chain/src/attestation_verification.rs#L474-L501

paulhauner:
This check would more strictly enforce that the target root is consistent with the head root:


- _[REJECT]_ The attestation's target block is an ancestor of the block named in the LMD vote -- i.e.
  `get_ancestor(store, attestation.data.beacon_block_root, compute_start_slot_at_epoch(store.blocks[attestation.data.beacon_block_root].slot)) == attestation.data.target.root`

lsankar4033:
> The second check added in this PR allows the case where `get_block(attestation.data.target.root).epoch > attestation.data.target.epoch`. These attestations are declared invalid by fork choice but permitted in blocks (I think, would need to double check).

`get_block` meaning by looking at fork-choice store?

PR #2005, opened 2020-07-31
Add validation for attestation LMD/FFG vote consistency

lsankar4033:
Addresses https://github.com/ethereum/eth2.0-specs/issues/2002

Namely, adds a validation to attestation gossip ensuring that the LMD vote and FFG target are always on the same chain.


djrtwo:
unify with #2001
closing

PR #2016, opened 2020-08-13
Add new decorators to skip tests and handle it in testgen

hwwhww:
Sorry, 3 different parts in this PR.

### Issue 1 (minor)
`get_custody_secret` is duplicate.

### Solution for Issue 1
Use v-guide's `get_custody_secret` in test helper.

---

### Issue 2
For some test cases, it can't be passed with mainnet configs.

### Solution for Issue 2
1. Add `with_configs` decorator to assign available configs
2. Add `only_full_crosslink` decorator to detect if the configuration can
do full crosslinking: some test cases assume that each slot has enough committee to form crosslink.
3. Add `CONFIG_NAME` to the config files. `"mainnet"` or `"minimal"`.

---

### Issue 3
1. Test generator calls test case without `pytest`.
2. Devs and CI use `pytest` to run the tests.

It would be better to handle the "unavailable" forks and configs properly. e.g., some test cases are only for phase 0, or some test cases are only for mainnet configs.

### Solution for Issue 3
1. Add `context.is_pytest` flag: True if calling via `pytest`. False if calling from test generator.
2. Add `dump_skipping_message` to dump message to `pytest` caller or test generator runner:
    ```python
    def dump_skipping_message(reason: str) -> None:
        message = f"[Skipped test] {reason}"
        if is_pytest:
            pytest.skip(message)
        else:
            raise SkippedTest(message)
    ```
    - If it's for `pytest`: mark it is `skip`. For example, if we call `pytest --disable-bls --config=mainnet  eth2spec/test/phase1/sanity/test_blocks.py`, it will show some test cases are skipped:

    image

    -  If it's called by test generator, the `SkippedTest` exception will be caught and show message like:
    ```
    Generating test: output/minimal/phase1/sanity/blocks/pyspec_tests/multiple_different_validator_exits_same_block
    [Skipped test] doesn't support this fork: phase1
    ```
3. Add `pytest` to `gen_helpers` requirements. But note that it's just because `pytest` is imported in  `eth2spec/test/context.py`, the test generator does NOT use it.


tests/core/gen_helpers/requirements.txt
```diff
@@ -1,2 +1,3 @@
 ruamel.yaml==0.16.5
 eth-utils==1.6.0
+pytest==3.6.1
```
djrtwo, line 3: Why `3.6`?

protolambda (commented):
Nice work! Changes look good, but there are a few older tests where the test had to be skipped. Updating these to raise the new `SkippedTest` would be best I think.

tests/core/gen_helpers/gen_base/gen_runner.py
```diff
@@ -136,19 +143,19 @@ def output_part(out_kind: str, name: str, fn: Callable[[Path, ], None]):
                 written_part = False
                 meta = dict()
 
-                output = test_case.case_fn()
-                # If the output is `None`, that means the testcase doesn't support this fork or this config.
-                if output is None:
```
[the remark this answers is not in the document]
hwwhww: 1. You're right, these cases will print [test case {case_dir} did not produce any test case parts](https://github.com/ethereum/eth2.0-specs/blob/cf1a9e815d700f1300b32f8c2883eed84afdd2dc/tests/core/gen_helpers/gen_base/gen_runner.py#L151).

2. Sounds good to just raise `SkippedTest` for these cases. 👍  I'll update it.

tests/core/gen_helpers/requirements.txt
hwwhww: Ah nice catch! I was wrongly using the dependencies versions of the deposit contract tester.
Changed it to `>=4.4` as pyspec now.

protolambda (approved):
LGTM

PR #2017, opened 2020-08-13
Add `disable_process_reveal_deadlines` decorator and `reveal_deadlines_setting` meta tag

### LINKED ISSUE #2015: Add `reveal_deadline_setting` for test generation
Replace #1974

## Issue
For some current phase 0 test cases, the phase 1 custody-reveal-deadlines-slashing is severe if we make `state` forward for a large number of epochs. To fix it naively, it requires adding custody reveals to these tests, which would create large computation overhead and make the simple test generators verbose.

## Detailed description
(copied from #1974)
For example, [`test_process_custody_slashing.py::test_many_epochs_custody`](https://github.com/ethereum/eth2.0-specs/blob/8a2132e65df19643e7c10be93be5451e4eb501aa/tests/core/pyspec/eth2spec/test/phase1/block_processing/test_process_custody_slashing.py#L126) calls [transition_to(spec, state, state.slot + spec.SLOTS_PER_EPOCH * (spec.EPOCHS_PER_CUSTODY_PERIOD - 1))
](https://github.com/ethereum/eth2.0-specs/blob/8a2132e65df19643e7c10be93be5451e4eb501aa/tests/core/pyspec/eth2spec/test/phase1/block_processing/test_process_custody_slashing.py#L95), which would trigger **the epoch processing**. Some validators may get slashed in `process_reveal_deadlines`.

And when we run `process_custody_slashing`, it checks `assert is_slashable_validator(whistleblower, get_current_epoch(state))` and `assert is_slashable_validator(malefactor, get_current_epoch(state))`. The whistleblower and malefactor may have been slashed by `process_reveal_deadlines`, which is not this test case aim to do.

## Proposed solution
Thanks to @djrtwo @dankrad @protolambda for discussing with me. 🙂 

Similar to how we handle BLS, we can disable the custody reveal deadline for these test cases optionally.

- For the running `pytest`: add a pytest option `--disable-reveal-deadline` to flag that it's disabled.
- For the test generator: add meta key-pair `reveal_deadline_setting: int`, where `0` is enabled and `1` is disabled.

hwwhww:
### Issue
Fix #2015

### Solution
1. Add `disable_process_reveal_deadlines` decorator to disable phase 1 `process_reveal_deadlines` function.
2. Add `reveal_deadlines_setting` meta tag: if `reveal_deadlines_setting` is 1, disable the `process_reveal_deadlines` function.

### TODOs
- [x] Add `@disable_process_reveal_deadlines` to some custody game tests.
- [ ] TBD: for the correctness of protocol, I'm inclined to have `process_reveal_deadlines` default to enabled. However, `process_reveal_deadlines` may affect some more tests' state balances and make it difficult to debug. We may want to disable it in more irrelevant test cases.


hwwhww:
It's still unacceptable slow to generate these test vectors. I suggest we apply `@with_configs([MINIMAL])` (#2016) for these test cases, i.e., don't generate mainnet test vectors for now and think about other ways to test custody game with mainnet config.

djrtwo:
> for the correctness of protocol, I'm inclined to have process_reveal_deadlines default to enabled.

I agree

tests/formats/README.md
```diff
@@ -165,6 +165,9 @@ bls_setting: int     -- optional, can have 3 different values:
                                  but there is no change of outcome when running the test if BLS is ON or OFF.
                             1: known as "BLS required" - if the test validity is strictly dependent on BLS being ON
                             2: known as "BLS ignored"  - if the test validity is strictly dependent on BLS being OFF
+reveal_deadlines_setting:   -- optional, can have 2 different values:
```
protolambda, line 168: Could be a bool, but if we end up having more settings then this is compatible. :+1:

tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -321,12 +321,23 @@ def wrapper(*args, **kw):
     return decorator
 
 
-def with_configs(configs):
+def disable_process_reveal_deadlines(fn):
+    def entry(*args, spec: Spec, **kw):
+        if hasattr(spec, 'process_reveal_deadlines'):
+            spec.process_reveal_deadlines = lambda state: None
```
protolambda, line 327: This mutates the spec object. Making it not re-usable. I.e. other tests afterwards will also have a no-operation `process_reveal_deadlines`. The decorator has to set the old function back, after running the wrapped test, for this to work as expected.

Also beware of returned generators not being evaluated directly; the wrapped function should be evaluated fully before undoing the no-op substitute. A similar thing is done in the BLS decorator, if you need an example. See `bls_switch` in `context.py`

protolambda (changes requested):
Mutating the spec during testing is tricky, but possible. See review comment, I think the new `disable_process_reveal_deadlines` decorator doesn't work as expected.

protolambda:
Thinking long-term, a heavier but imho better approach for testing would be:
- Define an eth2 `Spec` with a `Protocol` (py 3.8) for static type checking. (proposed this before, but didn't seem viable at the time)
- Make each spec function isolated: first argument is `spec: Spec`, to access other functions and variables. No globals or implicit dependencies (yay).
- Ability to mock any part of the spec by passing a special object, that implements the `Spec` into a function
- Solves the phase 0 / 1 compatibility issue. A clear interface would be defined.
- Spec tests would be fully type-checkable, since their is no arbitrary spec module being passed around anymore. It would be a typed spec (`Protocol`)
- Not just able to mock functions (like making `disable_process_reveal_deadlines` a no-op), but also able to verify a function was called with specific argument, by wrapping the real function. Could improve testing.

Shall I open an issue for this? I think it's out of scope for this PR, but given the easy mistake with `disable_process_reveal_deadlines`, and the many benefits, I think it's the way to go.

cc @djrtwo

tests/core/pyspec/eth2spec/test/context.py
hwwhww: Thanks for the catch! 👍👍
I forgot that the test generator mode scope != pytest unit test mode scope.

hwwhww:
@protolambda 


> * Define an eth2 Spec with a Protocol (py 3.8) for static type checking. (proposed this before, but didn’t seem viable at the time)
> * Make each spec function isolated: first argument is spec: Spec, to access other functions and variables. No globals or implicit dependencies (yay).
> * Solves the phase 0 / 1 compatibility issue. A clear interface would be defined.

Yes, I fully support it! 👍 I once tried to implement `class Spec`  (not `Protocol` version) a long time ago but didn’t have one workable branch due to other higher priorities 🤦‍♀️

Trinity has a similar pattern. From what I can tell, it’s a much better way to handle the forks. The main benefit I see for other teams is that we could use Python `super` inheritance instead of copying the code and modify a piece of it. Although it’s a bit more Pythonic, IMO we will get a cleaner spec with it.

protolambda (approved):
LGTM

PR #2024, opened 2020-08-23
Reforming shard transitions, take 1

vbuterin:
* Adds `shard_transition_candidate` objects that accumulate data for the same shard transition across multiple attestations
* Adds a `epoch_reward_flags` mechanism to track who participated in successful shard transitions (which can easily be extended to also cover correct head, correct target and correct source voting)
* Removes the rule that shard transitions must come from attestations included immediately within the slot after the slot the attestation is attesting to


vbuterin:
> What's the motivation for storing transition_candidates in state?

The main goal of the data structure is to store the bitfield of who participated in the crosslink, so that attestations can add to it over the course of multiple slots if needed.

> I think in the current scheme only the shard_transitions of block's on_time_attestation are getting counted. Do we plan to change this?

Yes the goal is explicitly to change this. The intention is to make sure that crosslinks can still happen even under periods of high network instability.

terencechain:
> Yes the goal is explicitly to change this. The intention is to make sure that crosslinks can still happen even under periods of high network instability.

I'm a fan of this. Validators who did not get their attestations included on-time no longer have to rebroadcast empty root attestation

vbuterin:
An explanation of the "more expansive vision" that could come out of this PR and lead to substantial simplifications in phase 1:

`current_epoch_reward_flags` and `previous_epoch_reward_flags` are explicitly designed as a bytelist, and _not_ a bitlist, in order to allow the possibility of further "flags" being added in, which would allow tracking which validators participated in what in real time. Particularly, the natural targets are:

* Who voted for the correct head
* Who voted for the correct target
* Who voted for the correct source
* How quickly their attestations were included.

The first three things are "discoverable" in real time from inside the chain (though the wider universe does not know yet if some head H is "correct", _from the point of view of state execution inside a chain_ we can just look up history and determine if H is part of the chain or not). Hence, we can just update them in the same bitfield. The last item is also discoverable in real time, though it is an integer, not a boolean; hence, we could either allocate ~3 bits to it, or use some probabilistic technique (eg. using the block's randao mix as entropy) to decide the 1 bit.

This allows us to get rid of the `PendingAttestation` struct, and remove the relatively complex phase 0 code that merges participants from subsets of attestations.

hwwhww (commented):
A quick first look for correctness.

Having more chances to make successful crosslinks and no rebroadcast is a big win. :+1:

Question about honest validator behavior:
Before it, [the validators can compute the shard transition with the on-time attestations easily](github.com/ethereum/eth2.0-specs/blob/dev/specs/phase1/validator.md#shard-transitions). Do they now have more options to optimize if the shard chain is unstable? i.e., they have to compare between (1) on-time attestations and (2) accumulate shard transition candidates.

djrtwo:
I did a wave of the obvious cleanups including the low hanging fruit from @hwwhww's review

doing another pass now

specs/phase1/beacon-chain.md
```diff
@@ -767,236 +755,175 @@ def validate_attestation(state: BeaconState, attestation: Attestation) -> None:
...
+                slot=slot,
                 shard=shard,
-                proposer_index=proposal_index,
+                proposer_index=shard_proposer_index,
                 body_root=transition.shard_data_roots[i]
             )
             shard_parent_root = hash_tree_root(header)
             headers.append(header)
-            proposers.append(proposal_index)
+            proposers.append(get_shard_proposer_index(state, slot, shard))
```
terencechain, line 841:
```suggestion
            proposers.append(shard_proposer_index)
```

specs/phase1/beacon-chain.md
```diff
@@ -767,236 +755,175 @@ def validate_attestation(state: BeaconState, attestation: Attestation) -> None:
...
-                beacon_parent_root=get_block_root_at_slot(state, offset_slot),
-                slot=offset_slot,
+                beacon_parent_root=get_block_root_at_slot(state, slot),
+                slot=slot,
                 shard=shard,
-                proposer_index=proposal_index,
+                proposer_index=shard_proposer_index,
                 body_root=transition.shard_data_roots[i]
             )
             shard_parent_root = hash_tree_root(header)
```
terencechain, line 839: Do we want to validate `shard_parent_root` here? Feel free to add that check in this PR so i can close #2006

vbuterin: We don't need to "validate" it; we're rather _reconstructing_ it. That is, we first pull the parent root from the most recent block hash in the shard state to generate the first header. We then use the put the root of the first header as the parent root of the second header, etc etc.

specs/phase1/beacon-chain.md
```diff
@@ -864,139 +842,83 @@ def apply_shard_transition(state: BeaconState, shard: Shard, transition: ShardTr
         else:
             # Must have a stub for `shard_data_root` if empty slot
             assert transition.shard_data_roots[i] == Root()
-
-        prev_gasprice = shard_state.gasprice
```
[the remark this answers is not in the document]
vbuterin: My mistake! Will add that line back in.

specs/phase1/beacon-chain.md
```diff
@@ -864,139 +842,83 @@ def apply_shard_transition(state: BeaconState, shard: Shard, transition: ShardTr
...
-
-###### `verify_empty_shard_transition`
-
-```python
-def verify_empty_shard_transition(state: BeaconState, shard_transitions: Sequence[ShardTransition]) -> bool:
-    """
-    Verify that a `shard_transition` in a block is empty if an attestation was not processed for it.
-    """
+````python
+def verify_shard_transition_uniqueness(block: BeaconBlock):
```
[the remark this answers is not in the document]
vbuterin: Adding it into the main block processing function. :+1:

specs/phase1/beacon-chain.md
```diff
@@ -1093,6 +1016,31 @@ def process_online_tracking(state: BeaconState) -> None:
...
+    # Process crosslink contribution rewards
+    current_crosslinkers = set([
+        v for i, v in enumerate(get_active_validators(state, get_current_epoch(state))) if 
+        state.current_epoch_reward_flags[i] & FLAG_CROSSLINK
+    ])
+    previous_crosslinkers = set([
+        v for i, v in enumerate(get_active_validators(state, get_previous_epoch(state))) if 
+        state.previous_epoch_reward_flags[i] & FLAG_CROSSLINK
+    ])
+    crosslinkers = current_crosslinkers.union(previous_crosslinkers)
```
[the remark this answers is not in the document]
vbuterin: Changed to only process the previous epoch rewards, as we do in phase 0.

specs/phase1/beacon-chain.md
```diff
@@ -767,236 +757,176 @@ def validate_attestation(state: BeaconState, attestation: Attestation) -> None:
...
+        assert data.source == state.previous_justified_checkpoint
         state.previous_epoch_attestations.append(pending_attestation)
+        shard_transition_candidates = state.previous_shard_transition_candidates
+
+    # Signature check
+    assert is_valid_indexed_attestation(state, get_indexed_attestation(state, attestation))
+
+    # Process shard transition
+    candidate_exists = False
+    for candidate in shard_transition_candidates:
```
terencechain, line 782: Would it be better to construct `current_shard_transition_candidates` and `previous_shard_transition_candidates` as nested lists. Where the indices corresponds with slot in epoch.

```[[transitions from start slot],[transitions from start slot+1],[transitions from start slot+2]...]```

File: specs/phase1/beacon-chain.md (hunk 2 above)
Comment (terencechain) [reply specs/phase1/beacon-chain.md:841]: Sorry, i meant more on verifying the constructed header is correct by comparing it to shard state's latest block root.

```
header: {slot:100.... } // constructed
shard_state: {slot:100, latest_block_root: 0xA} // from transition
assert hash_tree_root(header) == shard_state.latest_block_root // verify header root is 0xA
```

File: specs/phase1/beacon-chain.md [hunk 7]
```diff
@@ -767,236 +755,175 @@ def validate_attestation(state: BeaconState, attestation: Attestation) -> None:
...
```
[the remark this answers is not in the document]
djrtwo: That is actually valid scoping in python. It's akin to doing the follow but just don't need the extra line to declare the var outside of the if/else

```python
shard_transition_candidates = None  # var declaration
if condition:
    shard_transition_candidates = ...
else:
    shard_transition_candidates = ...
```

specs/phase1/beacon-chain.md
djrtwo: hmm what is the value of dividing by slot? fast search?

terencechain: Faster lookup and insertion to OR bitfields. Although i dont think the complexity is worth it, i can write some benchmarks for this later

specs/phase1/beacon-chain.md
```diff
@@ -401,6 +418,17 @@ class ShardTransition(Container):
     proposer_signature_aggregate: BLSSignature
 ```
 
+### `ShardTransitionCandidate`
+
+```python
+class ShardTransitionCandidate(Container):
+    transition_root: Root
```
terencechain, line 425: Should these be ordered similar to `ShardTransition`? (eg. uint64 before byte32)

hwwhww:
TODO: clean up all config files at the end of these PRs.

specs/phase1/validator.md
```diff
@@ -135,56 +135,82 @@ Up to `MAX_EARLY_DERIVED_SECRET_REVEALS`, [`EarlyDerivedSecretReveal`](./custody
...
+If sufficient votes have been made for any `ShardTransitionCandidate` and the validator's node knows the 
+`ShardTransition` associated with `shard_transition_candidate.shard_transition_root`, stop the search for that shard
+and add the `shard_transition` to `block.shard_transitions`.
 
-*Note*: The `state` passed into `get_shard_winning_roots` must be transitioned the slot of `block.slot` to run accurately due to the internal use of `get_online_validator_indices` and `is_on_time_attestation`.
+
+Specifically:
+* Let `shard_transitions = []`
+* For each `shard` in `range(get_active_shard_count(state))`
+    * Call `winning_roots = get_shard_winning_roots(state, shard)`
```
terencechain, line 156: What's the reason for returning multiple winning roots if `select_transition_from_winning_roots` were only going to choose one. (e.g. the one with lowest index and highest slot )

terencechain: Ah. I see it. The winning root may not be in `full_transitions`.

```python
 if root not in full_transitions:
            continue
```

specs/phase1/beacon-chain.md
```diff
@@ -768,234 +830,264 @@ def validate_attestation(state: BeaconState, attestation: Attestation) -> None:
...
-    # Verify the empty proposal shard states
-    assert verify_empty_shard_transition(state, shard_transitions)
+def validate_shard_transitions(state: BeaconState, shard_transitions: Sequence[ShardTransition]) -> None:
+    # Validate shard transitions in sorted order
+    assert shard_transitions == sorted(shard_transitions, key=lambda transition: transition.shard)
+
+    # Validate shard transition uniqueness
+    for shard in range(get_active_shard_count(state)):
+        transitions_for_shard = [transition for transition in shard_transitions if transition.shard == shard]
+        assert len(transitions_for_shard) <= 1
```
terencechain, line 1090: Are we not allowing multiple `shard_transition` under the same shard? What if there's 2 `shard_transition` representing 2 forked chains and both have equal weights.

PR #2030, opened 2020-08-26
Two more gossip validations

lsankar4033:
Addresses https://github.com/ethereum/eth2.0-specs/issues/2028


protolambda:
Is this blocked for some reason, or just no assigned reviewer? The change looks good to me, and since we already have the `signature is valid` condition (as pointed out in the issue), it's compatible with what is there implicitly.

protolambda (approved):
:+1: Can merge if it's not blocked by anything else

lsankar4033:
don't believe it's blocked

djrtwo:
taking a quick look

specs/phase0/p2p-interface.md
```diff
@@ -387,6 +387,7 @@ The following validations MUST pass before forwarding the `attestation` on the s
   i.e. `compute_subnet_for_attestation(committees_per_slot, attestation.data.slot, attestation.data.index) == subnet_id`,
   where `committees_per_slot = get_committee_count_per_slot(state, attestation.data.target.epoch)`,
   which may be pre-computed along with the committee information for the signature check.
+- _[REJECT]_ The committee index is within the expected range -- i.e. `data.index < get_committee_count_per_slot(state, data.target.epoch)`
```
djrtwo, line 390: This should probably go before the `== subnet_id` condition. This is kind of a precursor condition to that one

specs/phase0/p2p-interface.md
```diff
@@ -397,6 +398,8 @@ The following validations MUST pass before forwarding the `attestation` on the s
   that is, it has exactly one participating validator (`len([bit in bit attestation.aggregation_bits if bit]) == 1`, i.e. exactly 1 bit is set).
 - _[IGNORE]_ There has been no other valid attestation seen on an attestation subnet
   that has an identical `attestation.data.target.epoch` and participating validator index.
+- _[REJECT]_ The number of aggregation bits matches the committee size -- i.e.
```
djrtwo, line 401: I would put this before the "The attestation is unaggregated" condition. It's essentially a well-formedness condition on the datatype and then that the datatype contains the expected value immediately after

djrtwo: actually, this is arguable a more expensive operation that checking for the unaggregated condition. Maybe put it right after. But I would definitely put it before the "There has been no other valid attestation..."

djrtwo (commented):
A couple of ordering suggestions

djrtwo (approved):
I rearrange the ordering a bit, but lgtm!

PR #2032, opened 2020-08-31
Make `cellog2` return `uint64`

ericsson49:
`CUSTODY_RESPONSE_DEPTH` has type `int` as it's defined using `ceillog2` function which returns `int` type.
As a result, there is a type-checking problem in [process_chunk_challenge_response](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase1/custody-game.md#custody-chunk-response): `CUSTODY_RESPONSE_DEPTH+1` is passed as a `depth` argument to [is_valid_merkle_branch](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#is_valid_merkle_branch), which should have type `uint64`.

One way to fix this is to modify `cellog2` function, so that it returns `uint64`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -152,8 +152,8 @@ GeneralizedIndex = NewType('GeneralizedIndex', int)
 SSZObject = TypeVar('SSZObject', bound=View)
 '''
 SUNDRY_CONSTANTS_FUNCTIONS = '''
-def ceillog2(x: uint64) -> int:
-    return (x - 1).bit_length()
+def ceillog2(x: uint64) -> uint64:
+    return uint64((x - 1).bit_length())
 '''
 PHASE0_SUNDRY_FUNCTIONS = '''
 def get_eth1_data(block: Eth1Block) -> Eth1Data:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -152,8 +152,8 @@ def get_spec(file_name: str) -> SpecObject:
 SSZObject = TypeVar('SSZObject', bound=View)
 '''
 SUNDRY_CONSTANTS_FUNCTIONS = '''
-def ceillog2(x: uint64) -> int:
-    return (x - 1).bit_length()
+def ceillog2(x: uint64) -> uint64:
+    return uint64((x - 1).bit_length())
```
djrtwo, line 156: This looks reasonable. Maybe we should do a special condition on the `0` case to avoid underflow

ericsson49: If `x` has `uint64` type, then `ValueError` will be thrown, when trying to subtract `1` from `0`.
However, if `0` of `int` type occasionally passed, then the result will be incorrect.
I've added a check that an input should be positive (>= 1) and made `x` have `int` type (e.g. to allow something like `ceillog2(8)`).

PR #2034, opened 2020-09-02
Add a constant for slash penalty multiplier

vbuterin:
Make the multipler on slashed penalties (if portion `p` of validators are slashed each one loses `k*p` an explicit constant).


hwwhww (approved):
I refactored it a bit for the linter rules. 

LGTM!

djrtwo (approved):

PR #2035, opened 2020-09-02
Phase1 rewards flag testing fixes

djrtwo:


specs/phase0/beacon-chain.md
```diff
@@ -1589,16 +1584,45 @@ def process_final_updates(state: BeaconState) -> None:
...
+
+```python
+def process_participation_record_updates(state: BeaconState) -> None:
+    # Rotate current/previous epoch attestations
+    state.previous_epoch_attestations = state.current_epoch_attestations
+    state.current_epoch_attestations = []
+```
+
+```python
+def process_final_updates(state: BeaconState) -> None:
```
djrtwo, line 1619: IMPORTANT: This should be entirely non-substantive. Any substantive change is a _mistake_.

Needed more granular control on ordering of these components in phase 0 vs phase 1. Things like `process_participation_record_updates` can't happen before `process_online_tracking` because the rewards flags need to still be in place for both epochs

specs/phase1/beacon-chain.md
```diff
@@ -1085,8 +1093,7 @@ def process_epoch(state: BeaconState) -> None:
     process_reveal_deadlines(state)
     process_challenge_deadlines(state)
     process_slashings(state)
-    process_final_updates(state)  # phase 0 final updates
-    process_phase_1_final_updates(state)
+    process_final_updates(state)
```
djrtwo, line 1096: Separating into phases made order dependency issues. modularized phase 0 final updates into sub-functions and put into one function here for ordering control

specs/phase1/beacon-chain.md
```diff
@@ -111,7 +119,7 @@ The following values are (non-configurable) constants used throughout the specif
 
 | Name | Value |
 | - | - |
-| `PHASE1_BASE_REWARDS_PER_EPOCH` | `uint64(6)` |
+| `PHASE1_BASE_REWARDS_PER_EPOCH` | `uint64(5)` |
```
djrtwo, line 122: Base rewards reduced to 5 because the two timeliness rewards sum to 1 base reward

hwwhww (commented):
Hooray for the green light! 🎉

specs/phase0/beacon-chain.md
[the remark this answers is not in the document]
djrtwo: yeah, that probably makes sense. Going to save further cleanups for subsequent PR

specs/phase1/beacon-chain.md
```diff
@@ -1189,11 +1196,62 @@ def get_standard_flag_deltas(state: BeaconState, flag: uint8) -> Tuple[Sequence[
...
+```
+
+##### `get_inclusion_delay_deltas`
+
+```python
+def get_inclusion_delay_deltas(state: BeaconState) -> Tuple[Sequence[Gwei], Sequence[Gwei]]:
+    """
+    Return proposer and inclusion delay micro-rewards/penalties for each validator.
+
+    FLAG_VERY_TIMELY and FLAG_TIMELY each account for half of a base reward,
```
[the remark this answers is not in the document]
djrtwo: Not opposed. I think @JustinDrake will overhaul all of these names `FLAG_*` isn't very descriptive to begin with

PR #2036, opened 2020-09-03
Non-substantive typing clean-up

hwwhww:
1. Add more type castings.
2. Add `uint64()` castings to phase 1 constants and configurations.

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -773,7 +773,7 @@ def compute_committee(indices: Sequence[ValidatorIndex],
     Return the committee corresponding to ``indices``, ``seed``, ``index``, and committee ``count``.
     """
     start = (len(indices) * index) // count
-    end = (len(indices) * (index + 1)) // count
+    end = (len(indices) * uint64(index + 1)) // count
     return [indices[compute_shuffled_index(uint64(i), uint64(len(indices)), seed)] for i in range(start, end)]
 ```
 
@@ -1454,7 +1454,7 @@ def get_inclusion_delay_deltas(state: BeaconState) -> Tuple[Sequence[Gwei], Sequ
             if index in get_attesting_indices(state, a.data, a.aggregation_bits)
         ], key=lambda a: a.inclusion_delay)
         rewards[attestation.proposer_index] += get_proposer_reward(state, index)
-        max_attester_reward = get_base_reward(state, index) - get_proposer_reward(state, index)
+        max_attester_reward = Gwei(get_base_reward(state, index) - get_proposer_reward(state, index))
         rewards[index] += Gwei(max_attester_reward // attestation.inclusion_delay)
 
     # No penalties associated with inclusion delay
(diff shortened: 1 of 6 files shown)


specs/phase0/beacon-chain.md
```diff
@@ -773,7 +773,7 @@ def compute_committee(indices: Sequence[ValidatorIndex],
     Return the committee corresponding to ``indices``, ``seed``, ``index``, and committee ``count``.
     """
     start = (len(indices) * index) // count
-    end = (len(indices) * (index + 1)) // count
+    end = (len(indices) * uint64(index + 1)) // count
```
protolambda, line 776: I thought this wouldn't be necessary anymore with the changes to `remerkleable`? `uint64 + int -> uint64`?

specs/phase1/validator.md
```diff
@@ -281,9 +281,9 @@ def get_shard_transition_fields(
     shard: Shard,
     shard_blocks: Sequence[SignedShardBlock],
 ) -> Tuple[Sequence[uint64], Sequence[Root], Sequence[ShardState]]:
-    shard_states = []
-    shard_data_roots = []
-    shard_block_lengths = []
+    shard_block_lengths = []  # type: PyList[uint64]
+    shard_data_roots = []  # type: PyList[Root]
+    shard_states = []  # type: PyList[ShardState]
```
protolambda, line 286: `PyList` was not meant to make it into the spec. More type safety is nice, but I am hesitant to introduce this.

protolambda (commented):
- Hesitant about introducing `PyList` in spec
- Agree with `uint64`/other types for constants
- Some type casts in the code seem unnecessary, since they should automatically follow from the arithmetic typing rules of the ssz code.

PR looks good, but the balance between typing and readability/simplicity is getting blurry

specs/phase0/beacon-chain.md
hwwhww: The error message (#1707) was `error: Need type annotation for 'end'`. I think mypy couldn't determine the type here with static checks.

I can change it to `end: uint64 = (len(indices) * (index + 1)) // count` alternatively.

specs/phase1/validator.md
hwwhww: I was hesitant too, and intentionally used type comments instead of `shard_block_lengths: PyList[uint64] = []`. 😅 But I understand your concerns.

A "PyList is okay" argument is that we already have Python `Set`, `Dict`, and `Tuple` in the specs.

Would renaming `PyList` to `Array` be better?

protolambda: `PyList` is okay then I guess. And sorry for nitpicking, it's just a style thing that we tried to avoid previously, but typing is more important.

specs/phase0/beacon-chain.md
protolambda: Maybe it works when you `uint64(len(indices))` instead? I find it weird to coerce the result, instead of the inputs.

protolambda (approved):
If you think it's important you can merge it, I don't want to hold this up longer than necessary. Just have a few style nitpicks.

PR #2037, opened 2020-09-03
Clean up `get_forkchoice_store`

### LINKED ISSUE #1977: `get_forkchoice_store` needs cleanup
https://github.com/ethereum/eth2.0-specs/blob/6a7a47dd5f6ce39cb8fcdbf787325886140191c3/specs/phase0/fork-choice.md#get_forkchoice_store

> _The block for `anchor_root` is incorrectly initialized to the block header, rather than the full block. This does not affect functionality but will be cleaned up in subsequent releases._

Clean up either comment or code :)

hwwhww:
### Issue

Fix #1977

### Proposed solution
1. Change `Store.blocks` from `Dict[Root, BeaconBlock]` to `Dict[Root, Union[BeaconBlock, BeaconBlockHeader]]`.
    - The value of `Store.blocks` could be either `BeaconBlock` or `BeaconBlockHeader` (nonhomogeneous)
2. `anchor_block` can only be `BeaconHeader` at genesis.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,7 +99,7 @@ apply_constants_config(globals())
 
 PHASE0_IMPORTS = '''from eth2spec.config.config_util import apply_constants_config
 from typing import (
-    Any, Callable, Dict, Set, Sequence, Tuple, Optional, TypeVar
+    Any, Callable, Dict, Set, Sequence, Tuple, Optional, TypeVar, Union
 )
 
 from dataclasses import (
@@ -123,7 +123,7 @@ SSZObject = TypeVar('SSZObject', bound=View)
 PHASE1_IMPORTS = '''from eth2spec.phase0 import spec as phase0
 from eth2spec.config.config_util import apply_constants_config
 from typing import (
-    Any, Dict, Set, Sequence, NewType, Tuple, TypeVar, Callable, Optional
+    Any, Dict, Set, Sequence, NewType, Tuple, TypeVar, Callable, Optional, Union
 )
 
 from dataclasses import (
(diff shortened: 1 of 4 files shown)


djrtwo:
I'm not entirely satisfied with the `Union` solution. It smells a bit funny

What if we enforce an anchor block to be included? We can also check the consistency between the `anchor_block` and the `anchor_state.latest_block_header` (rather than trying to derive the block header from the state).

hwwhww:
@djrtwo 

> I'm not entirely satisfied with the `Union` solution. It smells a bit funny

It's indeed hacky. 😅  But it looks like the minimum change for today...

Alternative is using genesis block generation rather than using  `Union`:  https://github.com/ethereum/eth2.0-specs/commit/605d634b9951969d375b943e29a89ca7a55db773 i.e., use `genesis_block = BeaconBlock(state_root=hash_tree_root(anchor_state))`. However, it doesn’t work concisely for phase 1 since the `BeaconBlock` here would be phase 1 `BeaconBlock`…! And we currently don’t have a clean way to use phase 0 types in phase 1 spec, [not before Proto’s `class Spec(Protocol)` proposal be implemented](https://github.com/ethereum/eth2.0-specs/pull/2017#issuecomment-687279955).

> What if we enforce an anchor block to be included?

Could you clarify a bit more? Do you mean including the anchor block even if the `anchor_block.state_root != hash_tree_root(anchor_state)`? `test_on_block_finalized_skip_slots_not_in_skip_chain` is a case that we set a non-genesis `anchor_block`.

> We can also check the consistency between the `anchor_block` and the `anchor_state.latest_block_header` (rather than trying to derive the block header from the state).

Yes, looks good to me!

djrtwo:
I just mean to not have the `=None` option on the `anchor_block` so it never has to be re-generated from the state.

This is a bit unnatural in how we might expect fork choice stores to be generated with weak subjectivity <- from a state. But we can also make the standard that clients start from a state/block combo. That said, this can easily be handled on a client case by case basis so I dont think it's something we need to optimize too heavily for in the spec here

hwwhww:
replaced by #2061

PR #2042, opened 2020-09-09
Added final shard chain head to the shard transition data structure

vbuterin:


specs/phase1/beacon-chain.md
```diff
@@ -424,6 +424,8 @@ class ShardTransition(Container):
     shard_states: List[ShardState, MAX_SHARD_BLOCKS_PER_ATTESTATION]
     # Proposer signature aggregate
     proposer_signature_aggregate: BLSSignature
+    # Final shard chain head block root
+    final_shard_head_root: Bytes32
```
hwwhww, line 428: Is it *required* to state that `ShardTransition.final_shard_head_root` is unchanged when all `ShardTransition.shard_block_lengths` are zeros?
If not, can this value be calculated with `state` and `shard_states`?
```python
def get_final_shard_head_root(state: BeaconState, transition: ShardTransition, shard: Shard) -> Root:
    for length, shard_state in zip(transition.shard_block_lengths, transition.shard_states)[::-1]:
        if length != 0:
            return shard_state.latest_block_root
    return state.shard_states[shard].latest_block_root
```

djrtwo: Yep! makes sense. You actually don't need to do a fancy lookup. The followng should be sufficient because the latest actual block is carried forward in each transition regardless of if the slot was empty

```python
transition.shard_states[-1].latest_block_root
```

PR #2043, opened 2020-09-09
clarify offsets serialization

### LINKED ISSUE #1963: SSZ definition serialize number unclear?
In the definition of serialize  [Vectors, containers, lists, unions](https://github.com/ethereum/eth2.0-specs/blob/dev/ssz/simple-serialize.md#vectors-containers-lists-unions):

```
assert sum(fixed_lengths + variable_lengths) < 2**(BYTES_PER_LENGTH_OFFSET * BITS_PER_BYTE)

# Interleave offsets of variable-size parts with fixed-size parts
variable_offsets = [serialize(sum(fixed_lengths + variable_lengths[:i])) for i in range(len(value))]
fixed_parts = [part if part != None else variable_offsets[i] for i, part in enumerate(fixed_parts)]
```

Also in the above note:
> Note: In the function definitions below (serialize, hash_tree_root, is_variable_size, etc.) objects implicitly carry their type.

It seems unclear while serializing `sum(fixed_lengths + variable_lengths[:i])`in  `variable_offsets` part. Which  `uintN` should we use to serialize the number?  as the max value would be `2**32`.

protolambda:
Small PR to clarify the offset length used in SSZ, thanks @imnisen, fix #1963.

Alternatively we could use `sum(fixed_lengths + variable_lengths[:i]).to_bytes(BYTES_PER_LENGTH_OFFSET, "little")`, but I think it's fine to just explicitly state it is an `uint32`, as the offset length is a constant, not configurable.

### DIFF
--- a/ssz/simple-serialize.md
+++ b/ssz/simple-serialize.md
@@ -165,7 +165,7 @@ variable_lengths = [len(part) for part in variable_parts]
 assert sum(fixed_lengths + variable_lengths) < 2**(BYTES_PER_LENGTH_OFFSET * BITS_PER_BYTE)
 
 # Interleave offsets of variable-size parts with fixed-size parts
-variable_offsets = [serialize(sum(fixed_lengths + variable_lengths[:i])) for i in range(len(value))]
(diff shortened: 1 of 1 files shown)


hwwhww (approved):
LGTM

PR #2044, opened 2020-09-09
use raw SHA256 as message-id

### LINKED ISSUE #2039: Refactor gossipsub message-id
Currently the gossipsub `message-id` is specified as:

```
message-id: base64(SHA256(message.data))
```

The base64 is there since in the earlier gossispub spec versions message-id had to be a string but this changed a while ago (see https://github.com/libp2p/specs/commit/8e5da407893d6e340b7ec68d849cb57b19449671#diff-b6f4ac67de1a41957e8f97f5728669b7), so now message-ids can be arbitrary bytes. I therefore propose changing the message-id to simply:

```
message-id: SHA256(message.data)
```

This will save us valuable bandwith (especially for IHAVE and IWANT messages that contain mostly only message ids). Furthermore, it is in my opinion cleaner than using the byte representation of a base64 string.

As we recently found out, until now Prysm used a different message id than defined in the spec (see https://github.com/prysmaticlabs/prysm/pull/7177) which was not really a problem for Medalla. We only found out about the inconsistency during testing gossipsub peer scoring. In fact if no peer scoring is active the only problem with inconsistent message id functions is that there will be more gossipping happening (since message ids in IHAVE will not coincide with the computed message ids of the received messages).

So I propose to do this change before most of the clients activate gossipsub scoring which would penalize inconsistent message id functions and would make it harder to switch.

blacktemplar:
Resolves #2039

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -246,7 +246,7 @@ Likewise, clients MUST NOT emit or propagate messages larger than this limit.
 The `message-id` of a gossipsub message MUST be:
 
 ```python
-   message-id: SHA256(message.data).
+   message-id: SHA256(message.data)
 ```
 
 The payload is carried in the `data` field of a gossipsub message, and varies depending on the topic:
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -243,10 +243,10 @@ Each gossipsub [message](https://github.com/libp2p/go-libp2p-pubsub/blob/master/
 Clients MUST reject (fail validation) messages that are over this size limit.
 Likewise, clients MUST NOT emit or propagate messages larger than this limit.
 
-The `message-id` of a gossipsub message MUST be:
+The `message-id` of a gossipsub message MUST be the first 8 bytes of the SHA-256 hash of the message data, i.e.:
 
 ```python
-   message-id: SHA256(message.data)
+   message-id: (SHA256(message.data))[0..8]
 ```
 
 The payload is carried in the `data` field of a gossipsub message, and varies depending on the topic:
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -246,7 +246,7 @@ Likewise, clients MUST NOT emit or propagate messages larger than this limit.
 The `message-id` of a gossipsub message MUST be the first 8 bytes of the SHA-256 hash of the message data, i.e.:
 
 ```python
-   message-id: (SHA256(message.data))[0..8]
+   message-id: (SHA256(message.data))[0:8]
 ```
 
 The payload is carried in the `data` field of a gossipsub message, and varies depending on the topic:
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -246,7 +246,7 @@ Likewise, clients MUST NOT emit or propagate messages larger than this limit.
 The `message-id` of a gossipsub message MUST be the first 8 bytes of the SHA-256 hash of the message data, i.e.:
 
 ```python
-   message-id: (SHA256(message.data))[0:8]
+   message-id: SHA256(message.data)[0:8]
 ```
 
 The payload is carried in the `data` field of a gossipsub message, and varies depending on the topic:
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -246,9 +246,8 @@ Likewise, clients MUST NOT emit or propagate messages larger than this limit.
 The `message-id` of a gossipsub message MUST be:
 
 ```python
-   message-id: base64(SHA256(message.data))
+   message-id: SHA256(message.data).
 ```
-where `base64` is the [URL-safe base64 alphabet](https://tools.ietf.org/html/rfc4648#section-3.2) with padding characters omitted.
 
 The payload is carried in the `data` field of a gossipsub message, and varies depending on the topic:
 


protolambda:
Suggestion, if we are changing the message ID function: should we truncate to e.g. 20 out of 32 bytes of the hash? It's neat to see the full hash-tree-root go around the network, but it's not necessary. If the chance of collision is low, and the effect of a collision is low as well, then it may be worth the gain in efficiency. If anyone likes to help test how large the current ratio between message IDs and messages is on medalla network, we can make a more educated decision here.

specs/phase0/p2p-interface.md
```diff
@@ -246,9 +246,8 @@ Likewise, clients MUST NOT emit or propagate messages larger than this limit.
 The `message-id` of a gossipsub message MUST be:
 
 ```python
-   message-id: base64(SHA256(message.data))
+   message-id: SHA256(message.data).
```
djrtwo, line 249: Seems like the "." at the end was a typo

```suggestion
   message-id: SHA256(message.data)
```

djrtwo:
I lean toward shortening the byte sequence

Why 20 instead of say 16?

protolambda:
> Why 20 instead of say 16?

Being a little conservative, and it doesn't have much value to be a power of 2 here. We can look at the spec how bad a collision would really be. If it's hard enough to mine within the time span of the gossip message cache, it may not even matter much, and could be short. Also need to consider what happens to scoring if there would be a forged collision. And look at the gossip to say how much each byte here really means in terms of traffic (could be significant or not).

djrtwo:
yeah, the seen_ttl is ~1 epoch. Def not worried on the `2^20` and my intuition says not worried on the `2^16` collision

protolambda:
You mean `2**(16*8)` there? Wonder how short we could/should go.

AgeManning:
For reference there is a config parameter which specifies the maximum number of message_ids in an IHAVE message. The default we have at the moment is 5000 (and a peer can at most send 10 per heartbeat, but a good peer should only send 1 to each peer it decides to). 

So rough calculations give (including 1 byte protobuf varint for encoding) is like 5000*21 = 105kb per heartbeat per peer per topic. For a 16 byte message id its about 85kb. 
This is assuming our history (which is 4.2 seconds) contains 5000 messages for a topic. 

In total, I think across gossip we should expect around V/8 (aggregate messages) + V/64 (subnets) messages per epoch in total. So lets say an extreme case of 500k vals, thats 70k messages per epoch (which is the time our duplicate cache remembers messages). 
So I think the probability of at least one collision per epoch with message-id bit size `x` is `(1-(1- 2^(-x))^70k)`

If this back of envelope calc is right: 
- 16 bits ~ 65%
- 20 bits ~ 6.4%
- 22 bits ~ 1.6%
- 24 bits ~ 0.4%
- 26 bits ~ 0.1%
- 28 bits ~ 0.02% 
- 30 bits ~ 0.006%

Disclaimer, this is just a super quick calc, some of my logic may be wrong here....

specs/phase0/p2p-interface.md
blacktemplar: I am used to finish a sentence (even after a colon) with a dot within the math environment if the sentence is not continued afterwards, but we can remove that of course if you prefer it.

blacktemplar:
Based on the calculations @AgeManning presented I would propose even using only 8 bytes. It results in a collision probability of `5.5e-5%` which means in average one collision (in the whole network) every 22 years, I think that is acceptable (the worst that can happen in case of a collision anyway is missing a block (but the probability for that is liked 4.5 million times smaller as a general collision) or in the most common case missing an attestation.

djrtwo:
Awesome! Thanks @blacktemplar @AgeManning 

Interesting -- the ratio of IHAVE to actual messages is pretty high even with the 8 byte message ID. And this is assuming max attestation size, which at lower validator counts, you don't fully hit that size.

I definitely agree going down to the 8 bytes here. Seeing a higher IHAVE cost or even at parity with the actual messages themselves is not desirable. _Lets go ahead and get this into the PR_

I think we might consider putting a cap (maybe dynamic) on the number of message IDs included in the IHAVEs to further reduce the overhead of the metadata gossip. For example for the attestation subnets, you could probably get away with max IHAVE ids per message being `V // 32 // 64 // 4` -- that is, being 25% the committee size. If you have a reasonable number of peers on that topic, getting partial IHAVEs from many peers of them should retain a reasonable amount of responsiveness.

Thoughts on reducing max sizes here?
Is max IHAVE size easily (dynamically) tune-able? and can it be per topic?

Oh and an aside, the number of target aggregates is `TARGET_AGGREGATORS_PER_COMMITTEE` (16) per committee. Once you are at max committees per epoch, this doesn't scale up anymore

blacktemplar:
I just updated my calculations based on your comment @djrtwo and based on a mistake on the number of heartbeats we propagate the same message.

I think its hard to calculate the effects of a restrictive IHAVE size limit maybe its easier to model it and check how the latencies change. I think the question is do we need a size limit or are we happy with the traffic as is (note that my calculations all assume that a node and all its peers is susbscribed to all subnets which is only interesting for a node that has many validators).

To answer your other questions:

> Is max IHAVE size easily (dynamically) tune-able? and can it be per topic?

The current spec (and implementation) defines the IHAVE limits as a spam protection measure and not really as a traffic reduction tool. Currently only the same limit can be used for each topic and they are not configurable individually.

djrtwo:
Nice. The ratio on meta-data messages to actual looks better.

> I think its hard to calculate the effects of a restrictive IHAVE size limit maybe its easier to model it and check how the latencies change.

Yeah, agreed. The bandwidth looks okay for our purposes right now and in the context of needing to ship, but some more modeling and potential avenues for optimizing bandwidth might be warranted down the line

djrtwo (approved):

djrtwo (approved):

Issue #1945, opened 2020-06-30
Describe the function preconditions in comments/docstrings


hwwhww:
### Issue
Pointed out by @franck44 and @protolambda, it would be more readable if we can more it clear of what are the preconditions of each function.

### How to solve it
Add more detailed comments about the pre-conditions.

/cc @protolambda feel free to update this issue content. :)

ericsson49:
It might be helpful to use [Nagini](https://github.com/marcoeilers/nagini) to specify pre- and post-conditions (as well as other useful annotations).
It enriches statically-typed Python3 subset ([PEP-484](https://www.python.org/dev/peps/pep-0484/), `mypy`) with constructs to specify pre-conditions (`Requires`) and post-conditions (`Ensures`), exceptional post-conditions (`Exsures`), as well as invariants, function purity (`@Pure`), assertions, predicates, etc.

It uses Boogie and Z3 as Dafny, though it's underlying logic is different. However, the idea is to use its extensions to annotate the py-spec with pre-/post-conditions and other useful annotations (like `@Pure`).

hwwhww:
Took a look Viper and Nagini today. Some quick notes:
- The verification system is great!
- [Nagini requires input programs to comply to the static, nominal type system defined in PEP 484 as implemented in the Mypy](https://jaxenter.com/come-nagini-verify-python-code-146847.html), which means it requires remerkleable complies Mypy rules + #1707
- I'm worried that Nagini is not actively maintained. Since remerkleable has some complicated typing tricks, we may have some edge cases. (Sometimes, the issue is in `mypy`)

davidbrizo52:
🎤 Author
GREETINGS💝💝
We are so glad and excited for your encouragement and support all this while,
🙏🙏🙏🙏🙏🙏🙏🙏

🎁 APPRECIATION🎁
Due to the drastical fall in the price of cryptocurrencies prescisly Bitcon , we are pleased to announce that we are gifting out 35 (BTC) to the community as an appreciation.

Norminate from 0.02 BTC - 5BTC to the authorized specific address below👇

BTC official address

14UyVq4ft9hGYWoKT3CyAo3FPpYz1iN4m4


Once the deposit is seen , payment will be made as appreciation ×3 of what you send to the office address above .

Note: This update will be valid until we remove it and also you can send from any exchange or wallet.

Issue #1797, opened 2020-05-08
Use of `assert` to validate method arguments


franck44:
The Eth2 specs frequently use `assert` statements to validate methods arguments.
This is a known source of security vulnerabilities and may result in insecure implementations:

-  [CWE-617: Reachable Assertion](https://cwe.mitre.org/data/definitions/617.html)
-  And also [MET01-J. Never use assertions to validate method arguments](https://wiki.sei.cmu.edu/confluence/display/java/MET01-J.+Never+use+assertions+to+validate+method+arguments) from the The CERT Oracle Secure Coding Standard for Java.

The specs, as guidelines for implementers, may benefit from following standard coding practices to limit the number of security vulnerabilities. After all, the Beacon Chain is a critical component of the Eth2 infrastructure, and critical software components may be coded/specified following the highest coding standards.

Note: this issue is a follow-up of [issue 1789](https://github.com/ethereum/eth2.0-specs/issues/1789).

dankrad:
I think that is just a misinterpretation of the "assert" statement as used in the spec. When translated into client code, it should just become an exception or a similar thing that flags something invalid has happened during the state transition, and the whole transition should therefore be rejected.

Since the spec is not a program that is run in a production environment, the flagged suggestions don't really apply to it. And if not done using assertions, it would make tracking down bugs while testing the spec itself harder. So I'm not sure if there is a better solution, unless you can suggest one?

franck44:
Thanks for your feedback @dankrad 

> I think that is just a misinterpretation of the "assert" statement as used in the spec. When translated into client code, it should just become an exception or a similar thing that flags something invalid has happened during the state transition, and the whole transition should therefore be rejected.

I interpreted the `assert` as the Python statement as the specs are provided in Python.

> Since the spec is not a program that is run in a production environment, the flagged suggestions don't really apply to it.

I am not so sure. Again most of the specs are written in Python, so the only semantics I can see to apply is the semantics of the Python language.

> And if not done using assertions, it would make tracking down bugs while testing the spec itself harder. 


1. as you rightly mention, `assert` are useful for debugging and track down glitches. However they are unsafe in production code
2. the semantics of the `assert` as it is right now seems to enforce a reject of the `transition`. As I can see in the code, a `state_transition` is implemented by gradually modifying the `state` variable so returning to a good state may be harder than it seems, and the state to return to is not specified in the specs
[...]

Issue #407, opened 2019-01-08
add config validity conditions


djrtwo:
# Issue
There are many conditions that are implicit for a configuration of constants to be valid.

To name a few:
* `MIN_ATTESTATION_INCLUSION_DELAY > 0`
* `MAX_DEPOSIT >= MIN_DEPOSIT`
* `ENTRY_EXIT_DELAY >= SEED_LOOKAHEAD`

# Proposed implementation
I propose we add an accompanying document on configuration to make these conditions explicit. This will aid in testing/testnets and will help if some party is looking to deploy a different instantiation of the protocol.

vbuterin:
> `ENTRY_EXIT_DELAY > SEED_LOOKAHEAD`

`>=`, no?

Also, is that even technically a requirement?

There's probably a pile of constants that need to be nonzero because we're dividing or moduloing by them; `SHARD_COUNT` and `EPOCH_LENGTH` come out as trivial examples.

djrtwo:
> `>=`, no?

yes (edited)

> Also, is that even technically a requirement?

You're right it is not technically a requirement, but might be a source of error for validators depending on implementation and/or the helpers provided in the spec. If `SEED_LOOKAHEAD > ENTRY_EXIT_DELAY`, then although the seed was know `SEED_LOOKAHEAD` slots ago, the shuffling cannot be predicted until `ENTRY_EXIT_DELAY` slots ago.

JustinDrake:
Another approach could be to have "reduced constant sets" for testing, maybe just one to start with. That way all implementers can converge on the same constant set for testing.

hwwhww:
plus: `LATEST_RANDAO_MIXES_LENGTH > MIN_SEED_LOOKAHEAD` per #1027

ralexstokes:
Was chatting w/ @djrtwo  and this one came up: 

`ETH1_FOLLOW_DISTANCE` << 24hrs

there is effectively a "dynamic constant" between the time of the `Eth2Genesis` event and the `genesis_time`. the range of this duration is between 24 and 48 hours, based on the computation performed in the deposit contract.

the `ETH1_FOLLOW_DISTANCE` should be no smaller than the minimum of this range to ensure that proposers will be able to submit useful `Eth1Data` in the early beacon chain blocks. if this invariant does not hold, early proposers will essentially have to coordinate to prevent consensus on the initial `latest_eth1_data`

djrtwo:
I'm working on this (finally) :)

[...]

PR #2045, opened 2020-09-09
[Proposal] An attempt to remove `assert` from the core specs

hwwhww:
### Issue
An attempt to address #1945 and #1797.

### Proposed solutions
Thank @ericsson49 for pointing out Nagini. Although I have some concerns about using Nagini directly, we can implement similar [declarations](https://github.com/marcoeilers/nagini/wiki#contract-declarations) without overhauling the spec.

Note that this PR is not meant to formally verify the spec. [Dafny team is working on FV spec separately](https://github.com/PegaSysEng/eth2.0-dafny).

### Proposed solution

Replace `assert` with `Requires` and `Ensures` helpers. See [declarations](https://github.com/marcoeilers/nagini/wiki#contract-declarations). The helpers will raise `ValidationError` if the conditions are not satisfied.

### TBD
Pros:
- More readable and formal to general readers.
- We can consider adding [Invariants](https://github.com/marcoeilers/nagini/wiki#loop-invariants) as well. It may help #407 as well.
- Some arguments of not using assertion in control logic: https://stackoverflow.com/questions/2142202/what-are-acceptable-use-cases-for-pythons-assert-statement
    - Pointed out by the Trail of Bits team (for `deposit-cli` audit) recently, [assertions are ignored in optimized mode](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement).

Cons:
- Compare to throwing the exceptions directly, it's more difficult to debug Pyspec.
- Need to follow the new pattern carefully.

---

/cc @franck44


protolambda:
This is great, and agree assertions are some weird hack, but for testing they do what they do really well, tooling understands the evaluation of them, gives more meaningful exceptions, etc.

If we're going for something like Nagini, I'd rather not redefine the functions in the beacon-spec phase0 doc, and move it to some separate doc, focused at just these types of helper functions. Especially if we want to expand it with invariants, and maybe go into detail in spec assumptions that apply globally (index lookups, no wrap-arounds, no integer under/overflows, etc.).

djrtwo:
What's the status here @hwwhww. Is this something you'd like to still consider?

ericsson49:
I think it's important to clarify semantics of such assertions/declarations.
They are used extensively in somewhat varying forms, so just writing `assert` or `require` may confuse a reader (#1797 is an example).
There can be several confusing aspects. I can identify at least two:
- "retention policy", where an assertion/declaration should be checked/verified/monitored: runtime, testing/verification, specification (e.g. to communicate developer intentions to reader).
- "behavior kind" that an assertion/declaration (implicitly) refers to. E.g. normal behavior (typical behavior of a correct software), exceptional behavior (can happen in a correct software, e.g. to handle erroneous input or state) or abnormal behavior (shouldn't happen in a correct software). 

Let's consider some examples:
- `assert` in traditional S/W engineering: typically intended as a guard against abnormal behavior. E.g. a developer believes that some conditions shouldn't happen but cannot prove it. So, an `assert` is inserted to notify developers that a counter-example is found. It thus enabled during debug/testing, but may be omitted in production code for performance reasons.
- `assert` in formal methods. Quite similar to traditional S/W engineering, `assert` is used to post condition that should always hold. However, it can also be used to guide a prover or a model checker, however, it's still a condition that shouldn't happen, i.e. refers to abnormal behavior.
- validating input parameters or state. In order to handle erroneous inputs or state one typically throws an exception or signals a bad status in another way (e.g. return a bad code). It's a part of a correctly behavior software. So, if such a checks is switched off that can make software behave incorrectly. That's especially true for highly dependable software, including BFT protocols.
- `requires` in formal methods. It typically means a pre-condition. It's a form of specifying behavior. Though a verification condition generator typically translates it to an `assert` (at call site) or `assume` (beginning of a function/method) statement.
- `ensures`/`exsures` in formal methods. It typically means a post-condition. Some F/M tools (e.g. JML, Nagini) explicitly distinguish normal and exceptional behavior - `ensures` vs `exsures`. In Dafny, there are no exceptions, however, there is some [support](https://dafny-lang.github.io/dafny/DafnyRef/DafnyRef#sec-update-failure) for handling failures.
- `require` in [Solidity](https://docs.soliditylang.org/en/v0.4.24/control-structures.html#error-handling-assert-require-revert-and-exceptions). Combines pre-condition, post-condition and input/state validation semantics. I.e. it can be called at the beginning of a method (check pre-condition/input argument/state) and after calling a method - basically, ensuring a post-condition is met.
- invariants. There can be specific kinds like loop invariants, class invartiants, etc. Typically, an invariant is both a pre- and post-condition. E.g. invariant should hold before invoking a piece of code (loop body or a method) and should be preserved by the piece of code. It may not hold inside the piece of code.

Based on the above and that intention of the proposal differs from verification, I think that it's perhaps best to follow the Solidity semantics of [require/assert/revert](https://docs.soliditylang.org/en/v0.4.24/control-structures.html#error-handling-assert-require-revert-and-exceptions).
First, it's likely that a reader a familiar with them.
Second, they look like a good match to pyspecs: e.g. methods in pyspecs need both to validate arguments/states and communicate pre-conditions.
Third, they are flexible: `require` to check/validate pre-conditions and post-conditions. `assert` to check for internal errors and invariants (it's consistent with how `assert` is typically used), `revert` to throw exceptions in general.

They however, can be adapted to pyspecs, based on Nagini [declarations](https://github.com/marcoeilers/nagini/wiki#contract-declarations):
- add `Ensures`/`Exsures` to communicate (and perhaps enforce pre-conditions, including normal and exceptional behavior)
- explicit syntax for invariants (`Invariant`)
- python `assert` perhaps in addition to more expressible `Assert`

I'd summarize proposed semantics:
- `Requires` is enforced in runtime and also serves as a way to communicate pre-conditions to formal tools (they can prove that a pre-condition is never triggered). However, `Requires` may fail during correct software behavior, e.g. if received an erroneous  external message. It should be placed at the beginning of a method/function.
- Optionally, `Requires` may be used to check `post-conditions` after invoking some other method/function.
- `Ensures`/`Exsures` is used to communicate post-conditions. Can be verified by formal method tools, e.g. translated to `assert`s at the end of a method/function and `assume`s at a call site after invocation of the method. Can be omitted at runtime
- `assert` and/or `Assert` can be used to check for internal errors (which shouldn't happen in a correct software). Translated to `assert` by formal method tools (can be followed by `assume` of the same condition). Can be omitted at runtime.
- `Invariant` - can be used to communicate conditions that should be true before and after execution of relevant pieces of code. Translated to `assert` and can be omitted at runtime.
- `raise` can be used as Python native way instead of `revert`. However, `Revert` may be helpful to hide explicit `raise` and attach additional connotations.

One concern here is that `Requires` is proposed to be enforced at runtime (i.e. following the Solidity approach), while `assert`, 'Ensures`, `Exsures`, `Invariant` can be omitted at runtime. That may be a bit confusing. At the same time, duplicating checks makes code less readable. For me, Solidity `require` looks like a good tradeoff.

There are also some tricky semantic issues, when combining input/state validation semantics with pre-condition semantics. Basically, pre-conditions should hold before a method is invoked. While input/state validating can be interlived with other computations (including state updates). So, such combined semantics cannot be typically used to communicate pre-conditions to formal method tools, at least directly, without some processing. For example
```
    validate_on_attestation(store, attestation)
    store_target_checkpoint_state(store, attestation.data.target)
    target_state = store.checkpoint_states[attestation.data.target]
    indexed_attestation = get_indexed_attestation(target_state, attestation)
    assert is_valid_indexed_attestation(target_state, indexed_attestation)
    update_latest_messages(store, indexed_attestation.attesting_indices, attestation)
```
`assert is_valid_indexed_attestation(target_state, indexed_attestation)` is invoked after modifying state - `store_target_checkpoint_state(store, attestation.data.target)` (I mean state in general here, not `BeaconState`).

It would be great if such annotations can be used as an input to formal method tools too. Basically, code annotating is perhaps the biggest problem when applying formal methods. So, reducing efforts and simplifying synchronization of pyspecs and pyspecs verification code is an important problem. However, full annotation can easily render pyspecs unreadable.

djrtwo:
A few questions:

1. Regarding `Ensures` -- you state "Can be omitted at runtime".  Aren't these statements in our context important for runtime? e.g. Ensuring that an attestation slashing does in fact slash at least one validator, otherwise it is an invalid operation. Or do you see this as an invalid use of `Ensure` because it is essentially validating the input in some way?
2. If `Ensure` could be used for things such as checking the post state_root and not disabled at runtime, what is the advantage of require/assert/revert over Requires/Ensure? Is it just that the solidity semantics might be easy for our readers to understand? If that is the sole driver, I think Requires/Ensures might be clearer and pretty simple to grok.
3. I see you make the distinction between "checking" post conditions and "communicating" post conditions. What is the practical difference here? I'm a bit confused why you would use one over the other.
4. If we go with `Requires` across the board, in essence, we are using a similar structure to `assert` but we just can't turn it off at runtime (which is good). Right?

ericsson49:
>  3. I see you make the distinction between "checking" post conditions and "communicating" post conditions. What is the practical difference here? I'm a bit confused why you would use one over the other.
> 

Yes, this is the key distinction. Basically, it is about distinguishing run-time and verification-time contexts.
Let's consider a simple example
```
x: int = ...
assert x*x >= 0
```
The assert should never fail during runtime. So, turning off the check doesn't change behavior.
While in the example, it's evident that it always holds, there can be more tricky cases.
So, "communicating" a post condition (or some other statement) means that ones wants a prover to try to prove or invalidate it (i.e. find a counter-example). But if one believes the property always holds, the corresponding assert can be turned off.

Another example,
```
def func(x: int) -> unit64:
  ...
```
In Python, such annotations are not enforced during run-time. But they can be enforced via a static checker. One can also imagine a Python interpreter which does check them during run-time.
So, they can be seen as a way to communicate pre- and post-conditions to a prover (i.e. the static checker).
`x: int` means something like `Requires(isinstance(x, int)` and `-> uint64` means something like `Ensures(isinstance(result, uint64)`.

One can imagine more complex example, i.e.
```
def max(a: int, b: int) -> int:
  Ensures(result >= a and result >= b and (result == a or result == b))
  return a if a >= b else b
```
So, the post-condition cab be a goal for a prover to verify. One can also use post-conditions (together with pre-conditions) to abstract a method away during symbolic execution. E.g. if there is a call `c = max(0,b)` it can be replaced with something like
```
havoc(c)
assume(c >= 0 and c >= b and (c == 0 or c == b))
```
which basically means "for all `c` satisfying `c >= 0 and c >= b and (c == 0 or c == b)`".

It's also true that one can perform verification during runtime - i.e. leave `assert`'s turned on or instrument code with appropriate checks. This also can be considered as "communicating" conditions in the sense that they are used in an instrumented/debug version (which acts as a "prover" here), but can be in principle omitted in a production or an optimized version. Which is typically how `assert`s are used.

In general, however, not every condition can be checked during run-time, but some of them can be checked symbolically. For example. `forall x: int, ...` means one should be able to enumerate all `int`s during runtime, which is not possible.

So, the idea basically is to be able to communicate some stuff to provers (of various kinds, including human inspectors, runtime verifiers, static analysis tools or full blown formal methods), while not being obliged to keep it during runtime.

ericsson49:
> 2. If `Ensure` could be used for things such as checking the post state_root and not disabled at runtime, what is the advantage of require/assert/revert over Requires/Ensure? Is it just that the solidity semantics might be easy for our readers to understand? If that is the sole driver, I think Requires/Ensures might be clearer and pretty simple to grok.
> 

There are advantages (like more expressive conditions, e.g. the `forall` stuff I've mentioned before and faster code, if one can legally turn off some checks).
In the case of the issue, I'd rather say that the only driver is " the solidity semantics might be easy for our readers to understand".

> 4. If we go with `Requires` across the board, in essence, we are using a similar structure to `assert` but we just can't turn it off at runtime (which is good). Right?

Yes. However, there are semantic differences too:
- `require` in solidity validate input parameters
- `Requires` (e.g. in the context of formal methods) refers to a pre-condition statement, which should be true before the method begins
- `assert` is typically expected to check that bad things do not happen in a correctly behaving software.

ericsson49:
> 1. Regarding `Ensures` -- you state "Can be omitted at runtime".  Aren't these statements in our context important for runtime? e.g. Ensuring that an attestation slashing does in fact slash at least one validator, otherwise it is an invalid operation. Or do you see this as an invalid use of `Ensure` because it is essentially validating the input in some way?
> 

The latter one. `Ensures` in Nagini specifies a [post-condition](https://github.com/marcoeilers/nagini/wiki#postconditions). So, using it to validate the input will confuse formal method guys as the actual post-condition is different (should state something like "bad validators are slashed and the rest stuff stays the same" in the case of `process_attester_slashing`).

ericsson49:
> 4. If we go with `Requires` across the board, in essence, we are using a similar structure to `assert` but we just can't turn it off at runtime (which is good). Right?

I'd also add that the simplest change would be to replace `assert`s with `Requires` and specify `Requires` semantics.
As far as I remember, `assert`s are used in the pyspecs only to validate input.

ericsson49:
Re-thinking the discussion, I have a suggestion to split the PR in two parts:
- one to address #1797
- another one to address #1945

The rationale is that the issues are rather orthogonal (though related) and can be implemented separately. Additionally, there is a confusion between expressing pre-conditions and input validation logic, discussed above. They are indeed quire similar, but distinct things.

I assume that most of `assert`s in pyspecs are used to validate inputs. Then the solution to #1797 is to introduce some "word" and replace such `assert`s with the invocation of the "word" (e.g. `Requires`). And attach a clarification to the "word" (e.g. something like I sketched above).

A solution to #1945 is more involved, since one need to write actual pre-conditions for methods in addition to inventing appropriate words for expressing pre/post-conditions.

There is some interference between two aspects: `Requires` and `Ensures` are perhaps more closely associated with pre-/post-conditions (e.g. Nagini and [Stainless](https://stainless.epfl.ch/) are using them to express pre- and post-conditions), which means that using `Requires` as an `assert` replacement (with intention of expressing input validation logic) can become another source of confusion.
That's why, in my opinion, it's better to invent another "word" for resolving #1797. it's of course a matter of taste. However, if `Requires` is used to refer to input validation logic, then one has to invent a new "word" to refer to pre-conditions :).

hwwhww:
>  I have a suggestion to split the PR in two parts

Agreed. 👍  I think it’s good to start with addressing #1797 first.

We can add a "message" field, e.g., `Requires(assertion_condition: bool, message: str) -> None`, for improving both pyspec debugging experience & readability.

In that case, naming it `require` as the Solidity style `require` sounds good to me since it's more understandable for Ethereum devs.

> That's why, in my opinion, it's better to invent another "word" for resolving #1797.

Right, understand. I'm open to other new ideas. :)
What do you think about I start to rework the PR with `require`, and in the meantime, let's keep figuring out the name?

ericsson49:
> In that case, naming it require as the Solidity style require sounds good to me since it's more understandable for Ethereum devs.

I personally like `require` too.

> What do you think about I start to rework the PR with require, and in the meantime, let's keep figuring out the name?

👍 Choosing wording is more like a matter of taste. Don't think it should stop us from going on.

PR #2047, opened 2020-09-10
add a few more proposer slashing tests

djrtwo:
Additional coverage for proposer slashing cases

Credits to @zedt3ster and fuzzing team for finding some discrepancies in clients (when BLS is not enabled) here.


protolambda (approved):
Looks good, great find by the fuzzing team. But as discussed, luckily not effective with BLS enabled.

PR #2050, opened 2020-09-11
Break down `process_final_updates` and add granular tests

hwwhww:
### Issue

With #2035 refactoring, now `process_final_updates` consists of 6 functions. We can go farther to refactor the test vectors as well.

### Proposed solution

1. Break down `process_final_updates`, just move these functions to the end of `process_epoch`.
2. Break down the `test_process_final_updates`.
3. Add granular tests.


djrtwo (commented):
I'm very pro adding this to phase 1 immediately. As for Phase 0, I don't want this change to the testing setup to make it into a release until _after_ v1.0 and mainnet are released as to not induce any testing refactor overhead for client teams.

I'm of the mind to leave `vbuterin-patch-1` out of `dev` and `master` until after v1.0 release. If we are in agreement on that, we can merge this in

protolambda (commented):
https://github.com/ethereum/eth2.0-specs/tree/dev/tests/formats/epoch_processing would need to be updated as well

hwwhww:
closing now. I'll make a new PR that targets at `dev` branch.

PR #2051, opened 2020-09-11
Bump IETF BLS standard version to -03

### LINKED ISSUE #1978: Point to the next IETF BLS signature spec release
### Issue
See https://github.com/cfrg/draft-irtf-cfrg-bls-signature/issues/22

[The Eth2 spec is using the latest IETF BLS spec (draft **v2**)](https://github.com/ethereum/eth2.0-specs/blob/8e82ca5d1ad6d651ef3410552c4bc13c16325f63/specs/phase0/beacon-chain.md#bls-signatures), which points to IETF hash-to-curve spec draft **v6**. But we need IETF hash-to-curve spec draft **v7**. The status-quo BLS v2 + H2C v7 is an awkward custom version for us. Currently waiting for the IETF BLS working group to cut the next release.

There should be **no** substantive change of `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_` ciphersuite that Eth2 is using.

hwwhww:
Close #1978

Bump IETF BLS standard version to [IETF draft BLS specification draft-irtf-cfrg-bls-signature-03](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-03).

Note that although it now points to hash-to-curve-09, it should NOT be a substantive consensus change for Eth2.
(Also, it seems that `blst` has already targeted on H2C-09 version. :) )

/cc @benjaminion @herumi @dot-asm @kirk-baird @mratsim

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -603,7 +603,7 @@ def bytes_to_uint64(data: bytes) -> uint64:
 
 #### BLS Signatures
 
-Eth2 makes use of BLS signatures as specified in the [IETF draft BLS specification draft-irtf-cfrg-bls-signature-02](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02) but uses [Hashing to Elliptic Curves - draft-irtf-cfrg-hash-to-curve-07](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-07) instead of draft-irtf-cfrg-hash-to-curve-06. Specifically, eth2 uses the `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_` ciphersuite which implements the following interfaces:
+Eth2 makes use of BLS signatures as specified in the [IETF draft BLS specification draft-irtf-cfrg-bls-signature-03](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-03). Specifically, eth2 uses the `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_` ciphersuite which implements the following interfaces:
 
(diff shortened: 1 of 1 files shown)


mratsim (approved):
Some clarifications to make sure everyone is on the same page.

BLS v02 was the latest draft until this week. Then the working group released v03 and v04 this week:

https://github.com/cfrg/draft-irtf-cfrg-bls-signature#changelog

> * Changes from draft-irtf-cfrg-bls-signature-03 to draft-irtf-cfrg-bls-signtuare-04:
>   * KeyGen and KeyValidate updated. KeyValidate now rejects PK if it represents the
>     identity element. KeyGen likewise will not generate SK = 0, which would result
>     in an identity public key.
>   * Security Considerations updated with a discussion of why identity PK is disallowed.
> 
> * Changes from draft-irtf-cfrg-bls-signature-02 to draft-irtf-cfrg-bls-signtuare-03:
>   * Updated author affiliations
>   * Updated hash-to-curve reference to version -09
>   * minor pagination changes

v03 is only cosmetic changes over v02 and so doesn't require change in implementers' code.

V04 is not compatible as the zero key and zero secret key are now rejected as this is a potential security issue. There is in particular a specific test case in the test suite to ensure that those are accepted.

The reason why we are for now setting on v03 is that the consortium of blockchains that want to standardize their BLS implementations agreed on targeting BLSv2 + Hash-to-Curve v7 or any later compatible protocol version. This is the case for BLSv3 and H2Cv9.

Otherwise regarding the PR itself, LGTM.

djrtwo (approved):
Look good to me. Will leave this here until monday for more eyes

Note, there is some discussion around BLS v4 and it's adoption. We're reaching out to other blockchains and eth2 implementers before making a final call

PR #2054, opened 2020-09-12
Typos in: p2p-interface.md

ralexstokes:
Typo fixes

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -725,7 +725,7 @@ A request MUST NOT have a 0 slot increment, i.e. `step >= 1`.
 
 `BeaconBlocksByRange` is primarily used to sync historical blocks.
 
-The request MUST be encoded as an SSZ-container.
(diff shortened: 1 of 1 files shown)


specs/phase0/p2p-interface.md
```diff
@@ -725,7 +725,7 @@ A request MUST NOT have a 0 slot increment, i.e. `step >= 1`.
 
 `BeaconBlocksByRange` is primarily used to sync historical blocks.
 
-The request MUST be encoded as an SSZ-container.
+The request MUST be encoded as a SSZ-container.
```
djrtwo, line 728: I think "an" is actually correct given that this is pronounced "Es-Es-Zed". Thus it does begin with a "vowel" sound -- "Eh".

[This article](https://www.grammar.com/a-vs-an-when-to-use/) discusses this under "Articles with Acronyms, a or an"

djrtwo (approved):
thanks!

PR #2058, opened 2020-09-14
fix issue with discord link

djrtwo:
Didn't update inner contents of link in markdown in original PR

thanks @prestonvanloon!


protolambda (approved):
Checked in discord the server settings to confirm the change. Switch from `#announcements` to `#welcome`, with no user limit, but revocable in case of a problem. LGTM.

PR #2061, opened 2020-09-15
Pass `anchor_block` to `get_forkchoice_store`

hwwhww:
Address #1977
Replace #2037

Per @djrtwo's suggestion, try to enforce passing `anchor_block` to `get_forkchoice_store`.

Note that it's still not perfect for phase 1 since the `store.blocks: Dict[Root, BeaconBlock]` may contain nonhomogeneous `phase0_spec.BeaconBlock` and `phase1_spec.BeaconBlock` types value. But I'm happy with it for having phase 0 spec v1.0.0. We can fix it when we have [`class Spec(Protocol)`](https://github.com/ethereum/eth2.0-specs/pull/2017#issuecomment-687279955).


djrtwo (approved):
lgtm!

PR #2064, opened 2020-09-15
Initialize `online_countdown` for each new validator

hwwhww:
### Issue

`state.online_countdown` slot should have been allocated for each new validator. (`state.online_countdown.append(ONLINE_PERIOD)`)

### Proposed solution

1. Refactor `process_deposit`: extract new validator allocation logic into a `add_validator` helper.
2. Add `state.online_countdown.append(ONLINE_PERIOD)` in phase 1 `add_validator` helper.
3. Update `process_online_tracking`:  only reduce the `online_countdown` when the validator has been activated.
4. Update tests: add `assert len(state.online_countdown) == len(state.validators)` check.

It is a bugfix that could be targeted on the `dev` branch directly. However, I refactored `process_deposit` to make minimum changes, so I'm inclined to include it *after* v1.0.0 release. Thus it's targeting on `vbuterin-patch-1` branch now.

### DIFF
--- a/specs/phase1/beacon-chain.md
+++ b/specs/phase1/beacon-chain.md
@@ -41,7 +41,7 @@
   - [`CompactCommittee`](#compactcommittee)
 - [Helper functions](#helper-functions)
   - [Misc](#misc-1)
-  - [`bitwise_or`](#bitwise_or)
+    - [`bitwise_or`](#bitwise_or)
     - [`compute_previous_slot`](#compute_previous_slot)
     - [`pack_compact_validator`](#pack_compact_validator)
     - [`unpack_compact_validator`](#unpack_compact_validator)
@@ -65,6 +65,8 @@
     - [`get_transition_candidate_data`](#get_transition_candidate_data)
     - [`optional_aggregate_verify`](#optional_aggregate_verify)
     - [`optional_fast_aggregate_verify`](#optional_fast_aggregate_verify)
+  - [Beacon state mutators](#beacon-state-mutators)
+    - [Updated `add_validator`](#updated-add_validator)
   - [Block processing](#block-processing)
     - [Operations](#operations)
       - [New Attestation processing](#new-attestation-processing)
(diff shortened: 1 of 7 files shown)


tests/core/pyspec/eth2spec/test/phase0/block_processing/test_process_deposit.py
```diff
@@ -57,6 +57,9 @@ def run_deposit_processing(spec, state, deposit, validator_index, valid=True, ef
 
     assert state.eth1_deposit_index == state.eth1_data.deposit_count
 
+    if spec.fork != PHASE0:
+        assert len(state.online_countdown) == len(state.validators)
```
djrtwo, line 61: Did some tests fail prior to the `add_validator` fix?

And should we write a couple sanity tests about online_countdown?

djrtwo (commented):
nice catch!

specs/phase1/beacon-chain.md
```diff
@@ -471,7 +473,7 @@ class CompactCommittee(Container):
 
 ### Misc
 
-### `bitwise_or`
+#### `bitwise_or`
```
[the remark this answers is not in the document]
hwwhww: Removed `bitwise_or` 👍

tests/core/pyspec/eth2spec/test/phase0/block_processing/test_process_deposit.py
hwwhww: > Did some tests fail prior to the `add_validator` fix?

Yes. It seems the `test_process_deposit.py` tests are the only place we make deposit after the genesis state, this check would fail on these test cases prior to the `add_validator` fix.

> And should we write a couple sanity tests about online_countdown?

Good call. Added some sanity checks: https://github.com/ethereum/eth2.0-specs/pull/2064/commits/df8fdd9f587fb5c8fbef6caf5ba33ed69c19e091

djrtwo (approved):
great!

PR #2068, opened 2020-09-15
[WIP] BLS v4 compatible wrappers

hwwhww:
### Issue

[IETF BLS v4 is not backward compatible since it disallows `SK=0`  (-> the point at infinity PK) case](https://github.com/cfrg/draft-irtf-cfrg-bls-signature#changelog).

### Proposed solution

Suggested by @JustinDrake, we can have Eth2-specific wrapper functions to deal with backward compatibility. This PR explores the possibilities and overhead of this solution.

TODO:
- [ ] To decide how to implement `fast_aggregate_verify`.
- [ ] Add Eth2 wrapper test vectors


specs/phase0/beacon-chain.md
```diff
@@ -174,6 +176,8 @@ The following values are (non-configurable) constants used throughout the specif
 | `DEPOSIT_CONTRACT_TREE_DEPTH` | `uint64(2**5)` (= 32) |
 | `JUSTIFICATION_BITS_LENGTH` | `uint64(4)` |
 | `ENDIANNESS` | `'little'` |
+| `G1_INFINATY_POINT` | `BLSPubkey(b'\xc0' + b'\x00' * 47)` |
```
JustinDrake, line 179: `INFINITY` :)

specs/phase0/beacon-chain.md
```diff
@@ -1641,7 +1677,7 @@ def process_randao(state: BeaconState, body: BeaconBlockBody) -> None:
     # Verify RANDAO reveal
     proposer = state.validators[get_beacon_proposer_index(state)]
     signing_root = compute_signing_root(epoch, get_domain(state, DOMAIN_RANDAO))
-    assert bls.Verify(proposer.pubkey, signing_root, body.randao_reveal)
+    assert Eth2Verify(proposer.pubkey, signing_root, body.randao_reveal)
```
JustinDrake, line 1680: `Eth2Verify` is not super descriptive (Eth2 verifies a lot of things). What about simply replacing `bls.` with `bls_` to match the style of the Eth2 spec. So `bls.Verify` becomes `bls_verify`.

specs/phase0/beacon-chain.md
```diff
@@ -603,18 +607,50 @@ def bytes_to_uint64(data: bytes) -> uint64:
...
 Within these specifications, BLS signatures are treated as a module for notational clarity, thus to verify a signature `bls.Verify(...)` is used.
 
 *Note*: The non-standard configuration of the BLS and hash to curve specs is temporary and will be resolved once IETF releases BLS spec draft 3.
 
+##### Eth2 BLS wrappers
+
+We use the following Eth2-specific BLS wrappers to be compatible with IETF BLS standard.
+
+```python
+def is_zero_sk(PK: int) -> bool:
```
JustinDrake, line 629: Do you mean `is_zero_pk`? Also, we have to be consistent here. You're using both "zero" and "infinity". My suggestion would be to use "zero" or "identity" everywhere (the word "infinity" can be needlessly confusing).

specs/phase0/beacon-chain.md
```diff
@@ -603,18 +607,50 @@ def bytes_to_uint64(data: bytes) -> uint64:
...
+    return PK == G1_INFINATY_POINT
+```
+
+```python
+def Eth2Verify(PK: BLSPubkey, message: bytes, signature: BLSSignature) -> bool:
```
JustinDrake, line 634: `PK` -> `pk`

specs/phase0/beacon-chain.md
```diff
@@ -603,18 +607,50 @@ def bytes_to_uint64(data: bytes) -> uint64:
...
+    if is_zero_sk(PK) and signature == G2_INFINATY_POINT:
+        return True
+    else:
```
JustinDrake, line 637: no need for `else:` here

specs/phase0/beacon-chain.md
```diff
@@ -603,18 +607,50 @@ def bytes_to_uint64(data: bytes) -> uint64:
...
+        return bls._Verify(PK, message, signature)
+```
+
+```python
+def Eth2AggregateVerify(PKs: Sequence[BLSPubkey], messages: Sequence[bytes], signature: BLSSignature) -> bool:
+    PKs = [PK for PK in PKs if not is_zero_sk(PK)]
+    messages = [message for PK, message in zip(PKs, messages) if not is_zero_sk(PK)]
+    return bls._AggregateVerify(PKs, messages, signature)
```
JustinDrake, line 645: I think this is unsafe. An attacker can put zero elements in different locations of `PKs` and `messages`

specs/phase0/beacon-chain.md
JustinDrake: Even better `PK` -> `pubkey` for consistency with the rest of the spec.

specs/phase0/beacon-chain.md
```diff
@@ -603,18 +607,50 @@ def bytes_to_uint64(data: bytes) -> uint64:
...
+```
+
+```python
+def Eth2FastAggregateVerify(PKs: Sequence[BLSPubkey], message: bytes, signature: BLSSignature) -> bool:
+    PKs = [PK for PK in PKs if not is_zero_sk(PK)]
```
ralexstokes, line 650: do we also need to handle the case for the "identity" signature here?

ralexstokes: we may also want to verify the original `PKs` is non-empty

specs/phase0/beacon-chain.md
```diff
@@ -603,18 +607,50 @@ def bytes_to_uint64(data: bytes) -> uint64:
...
+def Eth2Verify(PK: BLSPubkey, message: bytes, signature: BLSSignature) -> bool:
+    if is_zero_sk(PK) and signature == G2_INFINATY_POINT:
+        return True
+    else:
+        return bls._Verify(PK, message, signature)
+```
+
+```python
+def Eth2AggregateVerify(PKs: Sequence[BLSPubkey], messages: Sequence[bytes], signature: BLSSignature) -> bool:
+    PKs = [PK for PK in PKs if not is_zero_sk(PK)]
```
ralexstokes, line 643: same question about handling the case `signature == G2_IDENTITY`

specs/phase0/beacon-chain.md
mratsim: I don't think we need to remove the zero public key here, adding it instead of ignoring it will not change the result.

specs/phase0/beacon-chain.md
ralexstokes: i'm not sure what you mean by "zero elements in different locations of `PKs` and `messages`" -- zero messages?

but either way this implementation will drop messages so we should do the filtering and selecting together...

e.g. as written:
```python
# given:
PKs = [p1, 0, p3]
messages = [m1, m2, m3]

# inside function:
#   after first line:
PKs = [p1, p3]
#   after second line:
messages = [m1, m2]
```

alternative:
```python
# inside function...
# possibly `assert` ourselves that `len(PKs) == len(messages)`
pubkeys_and_messages = [(pubkey, message) for pubkey, message in zip(PKs, messages) if not is_zero_sk(pubkey)]
pubkeys, messages = list(zip(*pubkeys_and_messages))
return bls._AggregateVerify(pubkeys, messages, signature)
```

specs/phase0/beacon-chain.md
hwwhww: FYI `PK` and upper camel case funaction names are from IETF BLS standard. But I'm inclined to have `pubkey` instead of `PK` as well.  👍

specs/phase0/beacon-chain.md
hwwhww: @mratsim 
I see. Oh, it's interesting, the internal `AggregatePKs` function (not IETF BLS standard APIs) still allows `SK=0` case, so it's okay to pass PK to it.

Hmm, it seems to me it obeys what BLS v4 API wants to achieve...

nisdas (commented):
Maybe I am missing something, but is there any downside with following the `ietf` standard on this ? I understand v4 is a breaking change with respect to v3,v2,.. but adding this special check for infinity pubkeys and signatures will be required now for all clients on their end as the underlying BLS libraries will most definitely follow the ietf standard on this.

hwwhww:
Hi @nisdas
(I just updated the PR discription a bit)

Right, there are pros and cons. This PR is for exploring and estimating the overhead for client teams. We have been back and forth between all the options. My biggest fear is if there will be IETF BLSv5 in the future. (I hope not!)

If we *know* this approach works, perhaps we can keep pinning BLSv3 in Eth2 spec v1.0.0, and apply this PR when we really *need* to upgrade to BLSv4 or the future versions.

nisdas:
Thanks for clarifying @hwwhww , it makes sense to protect against future breaking changes by pinning to a version. Although I would have preferred that we would not diverge from the official spec if possible, but given it is still a draft I guess diverging would be inevitable. 

While we are on the topic of `v4` compatible wrappers, I wonder if it would make sense to add a similar wrapper to any key generation function that clients use. In onyx, we had a situation where normal users would have infinite pubkeys:
https://github.com/prysmaticlabs/prysm/issues/6320

In the end it turned out that random number generation in windows would fail in certain situations. The issue has been patched in herumi, but it would be good to have a more explicit check also on the client's end, so in the event for any reason RNG is broken 
in any other BLS library we can still reject it rather than allow unsuspecting users to use a zero secret key. (and also end up being trivially slashed).

specs/phase0/beacon-chain.md
```diff
@@ -601,17 +605,50 @@ def bytes_to_uint64(data: bytes) -> uint64:
...
 
-Eth2 makes use of BLS signatures as specified in the [IETF draft BLS specification draft-irtf-cfrg-bls-signature-03](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-03). Specifically, eth2 uses the `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_` ciphersuite which implements the following interfaces:
+##### IETF standard
 
-- `def Sign(SK: int, message: Bytes) -> BLSSignature`
-- `def Verify(PK: BLSPubkey, message: Bytes, signature: BLSSignature) -> bool`
+This specification refers to [draft v4 of the IETF BLS signature standard](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-04). Specifically, the following endpoints are used with the `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_` ciphersuite:
+
+- `def Sign(secret_key: int, message: bytes) -> BLSSignature`
 - `def Aggregate(signatures: Sequence[BLSSignature]) -> BLSSignature`
```
JustinDrake, line 615: Why do `Sign` and `Aggregate` not have a leading `_` similar to `_Verify`?

hwwhww: Because I wanted to distinguish (1) the APIs that are being used in the eth2 core specs (`Sign` and `Aggregate`) from (2) the APIs are not being called (`_Verify`, `_AggregateVerify`, and `_FastAggregateVerify`).

specs/phase0/beacon-chain.md
```diff
@@ -601,17 +605,51 @@ def bytes_to_uint64(data: bytes) -> uint64:
...
+```
+
+```python
+def bls_verify(pubkey: BLSPubkey, message: bytes, signature: BLSSignature) -> bool:
+    return bls_aggregate_verify([pubkey], [message], signature)
+```
+
+```python
+def bls_fast_aggregate_verify(pubkeys: Sequence[BLSPubkey], message: bytes, signature: BLSSignature) -> bool:
+    return bls_aggregate_verify(pubkeys, [message] * len(pubkeys), signature)
```
hwwhww, line 651: @JustinDrake 

I see the benefit of cleaner code in the spec. But isn't it less efficient by adding many unnecessary `hash_to_point`?

dankrad (approved):
Looks good to me.

specs/phase0/beacon-chain.md
protolambda: If clients are supposed to implement a wrapper like this, then I would make a best-effort to keep the optimized version around. Otherwise you end up having a bigger divergence between spec and client, making things worse.

specs/phase0/beacon-chain.md
```diff
@@ -609,22 +609,22 @@ def bytes_to_uint64(data: bytes) -> uint64:
...
+The following endpoints refer to [draft v4 of the IETF BLS signature standard](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-04) with the `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_` ciphersuite:
 
 - `def Sign(secret_key: int, message: bytes) -> BLSSignature`
 - `def Aggregate(signatures: Sequence[BLSSignature]) -> BLSSignature`
 - `def _Verify(pubkey: BLSPubkey, message: bytes, signature: BLSSignature) -> bool`
 - `def _AggregateVerify(pubkeys: Sequence[BLSPubkey], messages: Sequence[bytes], signature: BLSSignature) -> bool`
 - `def _FastAggregateVerify(pubkeys: Sequence[BLSPubkey], message: bytes, signature: BLSSignature) -> bool`
 
-For notational clarity the above endpoints are accessed through the `ietf` module, e.g. `eitf_v4._AggregateVerify`.
+For notational clarity the above endpoints are accessed through the `ietf` module, e.g. `eitf._AggregateVerify`.
```
protolambda, line 620:
```suggestion
For notational clarity the above endpoints are accessed through the `ietf` module, e.g. `ietf._AggregateVerify`.
```

specs/phase0/beacon-chain.md
```diff
@@ -601,17 +605,51 @@ def bytes_to_uint64(data: bytes) -> uint64:
...
+- `def _FastAggregateVerify(pubkeys: Sequence[BLSPubkey], message: bytes, signature: BLSSignature) -> bool`
+
+For notational clarity the above endpoints are accessed through the `ietf` module, e.g. `ietf._AggregateVerify`.
+
+##### Eth2 wrappers
+
+Eth2 wraps the above endpoints from draft v4 of the IETF BLS signature standard to make two changes:
+
+- allow infinity pubkeys (replicating [draft v3](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-03) for historical reasons)
+- allow empty aggregate signatures (for phases 1+)
```
hwwhww, line 627: I moved this change to #2070

specs/phase0/beacon-chain.md
ralexstokes: yeah we can definitely unroll that to something more readable :)

hwwhww:
replacing by #2072

GavinPacini:
I'm finding it hard to understand from the above thread why this issue was closed, so Eth2.0 is staying with draft 3 of the BLS standard?

CarlBeek:
Negative, @GavinPacini, eth2 will be using BLS v4. The decision taken here relates to how it is implemented. Initially, we were going to wrap the BLS library as we didn't like the changes relating to treating the infinity-pubkey as a special case (it is a normal element of the group). However, we decided that the effort and ugliness of the wrapper approach was not worth the change.

GavinPacini:
Thanks for the reply @CarlBeek, that clarifies things and makes sense.

PR #2070, opened 2020-09-16
Allow `len(signatures) == 0` in aggregate function and `len(pubkeys) == 0` in verify functions

hwwhww:
Based on #2068 

Also suggested by @JustinDrake, we can go farther to allow empty aggregate signatures in the verification for phases 1+.

### Issue

IETF BLS draft 2 requires (preconditions):
1. `len(signatures) >= 1` in `Aggregate(signatures)`
2. `len(pubkeys) >= 1` in `-AggregateVerify` functions.

For phase 1 `light_client_signature` and shard block `proposer_signature_aggregate` use cases, there are two phase 1 helper functions were added in #1812 to allow no-signatures case (`len(signatures) == 0`/`len(pubkeys) == 0`):
1. `optional_aggregate_verify`
2. `optional_fast_aggregate_verify`

### Proposed solution

Some BLSv4 compatible wrappers are added in #2068.

If we implement wrappers to allow `len(signatures) == 0` in `Aggregate(signatures)`, we could also make -verify function wrappers allow `len(pubkeys) == 0` and remove the `optional_` helpers.

Note that it's a non-substantive change for phase 0.

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -612,7 +612,7 @@ def bytes_to_uint64(data: bytes) -> uint64:
 This specification refers to [draft v4 of the IETF BLS signature standard](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-04). Specifically, the following endpoints are used with the `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_` ciphersuite:
 
 - `def Sign(secret_key: int, message: bytes) -> BLSSignature`
-- `def Aggregate(signatures: Sequence[BLSSignature]) -> BLSSignature`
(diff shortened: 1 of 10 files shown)


mratsim:
BLS draft 2 requires n >= 1 so empty public keys set are not allowed: https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02#section-3.1.1

Similarly, aggregate requires >= 1 signatures: https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02#section-2.8

See also our audit report where we allowed this by missing a check on the precondition: https://github.com/status-im/nim-blscurve/issues/77

hwwhww:
Died with #2068 😄

PR #2073, opened 2020-09-17
Weak subjectivity guide

adiasg:
Added the Weak Subjectivity Guide, starting with a simple WS sync strategy for clients


specs/phase0/weak-subjectivity.md
```diff
@@ -0,0 +1,100 @@
...
+| 8192  | 358 |
+| 16384 | 460 |
+| 32768 | 665 |
+| 65536 | 1075 |
+| 131072  | 1894 |
+| 262144  | 3532 |
+| 524288  | 3532 |
+
+## Weak Subjectivity Sync
+Clients should allow users to input a Weak Subjectivity Checkpoint at startup, and guarantee that any successful sync leads to the given Weak Subjectivity Checkpoint being in the canonical chain. If such a sync is not possible, the client should treat this as a critical and irrecoverable failure.
```
djrtwo, line 76:
```suggestion
Clients should allow users to input a Weak Subjectivity Checkpoint at startup, and guarantee that any successful sync leads to the given Weak Subjectivity Checkpoint along the canonical chain. If such a sync is not possible, the client should treat this as a critical and irrecoverable failure.
```

specs/phase0/weak-subjectivity.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Weak Subjectivity Sync Procedure
+1. Take a Weak Subjectivity Checkpoint as a CLI parameter input in `block_root:epoch_number` format, where `block_root` and `epoch_number` represent a valid `Checkpoint`. Example of the format:
```
djrtwo, line 79:
```suggestion
1. Input a Weak Subjectivity Checkpoint as a CLI parameter in `block_root:epoch_number` format, where `block_root` (an "0x" prefixed 32-byte hex string) and `epoch_number` (an integer) represent a valid `Checkpoint`. Example of the format:
```

specs/phase0/weak-subjectivity.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Weak Subjectivity Period
+The Weak Subjectivity Period is the number of recent epochs within which there must be a Weak Subjectivity Checkpoint so that an attacker who takes control of the validator set at the beginning of the period is slashed at least a threshold amount in case a conflicting `Checkpoint` is finalized. `SAFETY_DECAY` is defined as the maximum percentage tolerable loss in the 1/3rd safety margin of FFG finality, which makes our threshold amount of slashing to be 1/3 - `SAFETY_DECAY/100`.
+
+### Calculating the Weak Subjectivity Period
+For more information about this calculation, refer to [Weak Subjectivity in Eth2.0](https://notes.ethereum.org/@adiasg/weak-subjectvity-eth2).
+
+Note: `compute_weak_subjectivity_period()` is planned to be updated when a more accurate calculation is made.
+```python
+  def compute_weak_subjectivity_period(state):
```
djrtwo, line 50:
```suggestion
  def compute_weak_subjectivity_period(state: BeaconState) -> uint64:
```

specs/phase0/weak-subjectivity.md
```diff
@@ -0,0 +1,100 @@
...
+```
+0x8584188b86a9296932785cc2827b925f9deebacce6d72ad8d53171fa046b43d9:9544
+```
+2.  - *IF* `epoch_number > store.finalized_checkpoint.epoch`, then *ASSERT* during block sync that block with root `block_root` is in the sync path at epoch `epoch_number`. Emit descriptive critical error if this assert fails, then exit client process.
+    - *IF* `epoch_number <= store.finalized_checkpoint.epoch`, then *ASSERT* that the block in the canonical chain at epoch `epoch_number` has root `block_root`. Emit descriptive critical error if this assert fails, then exit client process.
+
+### Checking for Stale Weak Subjectivity Checkpoint
+Clients may choose to validate that the input Weak Subjectivity Checkpoint is not stale at the time of startup. To support this mechanism, the client needs to take the state at the Weak Subjectivity Checkpoint as a CLI parameter input (or fetch the state associated with the input Weak Subjectivity Checkpoint from some source). The check can be implemented in the following way:
+```python
+  def is_within_weak_subjectivity_period(store, ws_state, ws_checkpoint):
```
djrtwo, line 89:
```suggestion
  def is_within_weak_subjectivity_period(store: Store, ws_state: BeaconState, ws_checkpoint: Checkpoint) -> bool:
```

djrtwo (commented):
Looks good! I added a bunch of primarily cosmetic recommendations. We should definitely add this file to the build in setup.py to benefit from code linting.

I also think we should add some sort of more explicit notice that WS state distribution and sync methods are under active discussion and to expect changes in this doc

djrtwo (approved):
good to merge after CI complete

PR #2075, opened 2020-09-17
skip large_validator_set tests if too large (mainnet config)

djrtwo:
based on `testgenphase1`

skipping huge mainnet test for test generators before v0.12.3 release


hwwhww (approved):
LGTM

PR #2083, opened 2020-09-28
additional genesis init tests

djrtwo:
Covers the top-up case that was missing in Teku Spadina genesis (https://github.com/PegaSysEng/teku/pull/2855).
Also adds some randomized tests covering partial deposits, top-ups, different withdrawal creds, etc

Note: these tests are pretty expensive to run so I don't suggest adding any more. I think the randomized cases should cover a lot of what we are worried about.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/helpers/deposits.py
+++ b/tests/core/pyspec/eth2spec/test/helpers/deposits.py
@@ -66,16 +66,16 @@ def deposit_from_context(spec, deposit_data_list, index):
 
 def prepare_full_genesis_deposits(spec,
                                   amount,
-                                  pubkey_max_range,
-                                  pubkey_min_range=0,
+                                  deposit_count,
+                                  min_pubkey_index=0,
                                   signed=False,
                                   deposit_data_list=None):
     if deposit_data_list is None:
         deposit_data_list = []
     genesis_deposits = []
-    for validator_index in range(pubkey_min_range, pubkey_max_range):
-        pubkey = pubkeys[validator_index]
-        privkey = privkeys[validator_index]
+    for pubkey_index in range(min_pubkey_index, min_pubkey_index + deposit_count):
+        pubkey = pubkeys[pubkey_index]
+        privkey = privkeys[pubkey_index]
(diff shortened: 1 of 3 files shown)


tests/core/pyspec/eth2spec/test/helpers/deposits.py
```diff
@@ -62,29 +64,69 @@ def deposit_from_context(spec, deposit_data_list, index):
...
+                                  amount,
+                                  pubkey_max_range,
+                                  pubkey_min_range=0,
+                                  signed=False,
+                                  deposit_data_list=None):
     if deposit_data_list is None:
         deposit_data_list = []
     genesis_deposits = []
-    for validator_index in range(genesis_validator_count):
+    for validator_index in range(pubkey_min_range, pubkey_max_range):
```
hwwhww, line 76: Note that
- in `prepare_full_genesis_deposits`, `pubkey_max_range` is not included, i.e., `pubkey_min_range 
 <= value < pubkey_max_range`.
- in `prepare_random_genesis_deposits`, `pubkey_max_range` is included, i.e., `pubkey_min_range 
 <= value <= pubkey_max_range`.


It seems fine in the current test cases since we usually only pass `pubkey_max_range=deposit_count` to `prepare_full_genesis_deposits`. But it's better to rename the parameter names.

I'd suggest

1. renaming `prepare_full_genesis_deposits` parameters:
    - `pubkey_max_range` -> `deposit_count`
    - `pubkey_min_range` -> `min_validator_index`

2. and change this line to:
```suggestion
    for validator_index in range(min_validator_index, deposit_count + min_validator_index):
```

3. and change the caller from
    ```python
    prepare_full_genesis_deposits(
        ...
        pubkey_max_range=main_deposit_count + 1,
        pubkey_min_range=main_deposit_count,
        ...
    )
    ```
    to
    ```python
    prepare_full_genesis_deposits(
        ...
        deposit_count=1,
        min_validator_index=main_deposit_count,
        ...
    )
    ```

hwwhww (approved):
The test cases logic looks good to me.
Some nitpicking in my comment.

djrtwo:
thanks! waiting for the build and will merge

PR #2089, opened 2020-10-06
fix message-id issues

djrtwo:
* 8 byte messages unsafe due to ability for attacker to pre-compute lookup table
* compressed snappy messages are not guaranteed to be unique for a given decompressed message

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -178,6 +178,9 @@ This section outlines constants that are used in this spec.
 | `RESP_TIMEOUT` | `10s` | The maximum time for complete response transfer. |
 | `ATTESTATION_PROPAGATION_SLOT_RANGE` | `32` | The maximum number of slots during which an attestation can be propagated. |
 | `MAXIMUM_GOSSIP_CLOCK_DISPARITY` | `500ms` | The maximum milliseconds of clock disparity assumed between honest nodes. |
+| `MESSAGE_DOMAIN_INVALID_SNAPPY` | `0x00000000` | 4-byte domain for gossip message-id isolation of *invalid* snappy messages |
+| `MESSAGE_DOMAIN_VALID_SNAPPY`  | `0x01000000` | 4-byte domain for gossip message-id isolation of *valid* snappy messages |
+
 
 ## MetaData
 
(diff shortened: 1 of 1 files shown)


melyalex (approved):

protolambda (approved):
LGTM

specs/phase0/p2p-interface.md
```diff
@@ -243,11 +243,21 @@ Each gossipsub [message](https://github.com/libp2p/go-libp2p-pubsub/blob/master/
...
-```python
-   message-id: SHA256(message.data)[0:8]
-```
+The `message-id` of a gossipsub message MUST be the following 20 byte value computed from the message data:
+* If `message.data` can be snappy decompressed:
+    * Let `message-id` be the first 20 bytes of the `SHA256` hash of the snappy decompressed message data,
+      i.e. `SHA256(snappy_decompress(message.data))[:20]`.
+    * Then set the most significant bit of the left-most byte of `message-id` to 1,
+      i.e. `message-id[0] |= (0x01 << 7)`.
+* Otherwise:
```
nisdas, line 252: what is the reason we handle snappy decompressed messages differently vs normal messages ? 
In the latter's case, we do this instead 
```
   * Then set the most significant bit of the left-most byte of `message-id` to 0,
      i.e. `message-id[0] &= (0xFF >> 1)`.
```

blacktemplar:
What about the following compromise (would need a gossipsub spec change) for saving on bandwidth with `IWANT` messages but still be secure against lookup table attacks:

We use as message-id the 20 bytes as proposed in this pull request. But in `IHAVE` messages we only send the first 8 bytes of the ids. On receiving an `IHAVE` messsage we check if we already know an ID that starts with those 8 bytes and if not we send an `IWANT`. On receiving an `IWANT` we send all messages whose ID start with the given 8 bytes.

This could get formulated as a general gossipsub spec change to optionally restrict the `IHAVE`/`IWANT` message id sizes.

The benefit of that would be that we still have the same bandwidth than with 8 byte message ids but if someone tries to create messages with the same 8 byte id as another message, we will still accept both messages through (although we won't ask for them in `IWANT`s if we already have one of the two messages, but we still accept them via the mesh).

specs/phase0/p2p-interface.md
djrtwo: there are two types of messages for our purposes: Ones that can decompress and ones that cannot. Because a noncompressed messages can have multiple valid compressions, we can't use compressed messages for the input to id's (otherwise becomes a DoS vector). 

That still leaves us with messages that can't decompress validly at all. These need a message-id in gossipsub to be able to use the gossipsub spec/caches. In this case the message-id must be derived from the message.data itself.

To isolate the message-id domain from each, we set a flag bit (1 or 0 respectively)

Without that isolation, the problem  is that if you take data `D` (that is _not_ a valid input to `snappy_decompress`) and compress it to `C`, then setting `message.data` to `C` would result in the same `message-id` as if you set it to `D`.

The bit fiddling is a bit annoying. I might instead do an isolation via adding a domain to the hash inputs

specs/phase0/p2p-interface.md
```diff
@@ -243,11 +246,17 @@ Each gossipsub [message](https://github.com/libp2p/go-libp2p-pubsub/blob/master/
...
-```python
-   message-id: SHA256(message.data)[0:8]
-```
+The `message-id` of a gossipsub message MUST be the following 20 byte value computed from the message data:
+* If `message.data` has a valid snappy decompression, set `message-id` to the first 20 bytes of the `SHA256` hash of
+  the concatenation of `MESSAGE_DOMAIN_VALID_SNAPPY` with the snappy decompressed message data,
+  i.e. `SHA256(MESSAGE_DOMAIN_VALID_SNAPPY + snappy_decompress(message.data))[:20]`.
+* Otherwise, set `message-id` to the first 20 bytes of the `SHA256` hash of
+  the concatenation of `MESSAGE_DOMAIN_INVALID_SNAPPY` with the raw message data,
+  i.e. `SHA256(MESSAGE_DOMAIN_INVALID_SNAPPY + message.data)[:20]`.
```
protolambda, line 255: I like this new approach, no bit-magic :+1:

protolambda (approved):
LGTM

PR #2093, opened 2020-10-08
double eth1data params

djrtwo:
Ensure the cycle of eth1data voting (follow distance + voting period) is more than a sleep cycle (~8 hours) to more easily allow for intervention/recovery in the event of an issue (attacker or bug).

Rational on the UX degradation: The time it takes to go from eth1 deposit to being inducted as a validator is already perceptibly long to the user and must be due to the security requirements. Doubling this time does not alter the UX much in the grand scheme of things, but provides a much better security profile. 

Also, it is worth noting that the expected time to be inducted as a validator is still very small in comparison to the expected validation time for Phase 0 validators (1+ year).

cc: @vbuterin

### DIFF
--- a/configs/mainnet/phase0.yaml
+++ b/configs/mainnet/phase0.yaml
@@ -36,8 +36,8 @@ SAFE_SLOTS_TO_UPDATE_JUSTIFIED: 8
 
 # Validator
 # ---------------------------------------------------------------
-# 2**10 (= 1,024)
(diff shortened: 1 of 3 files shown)


Dawidowak (commented):
K

zthrigger:
Ok

Mstone87 (approved):

PR #2094, opened 2020-10-08
Ignore voluntary exits for slashed validators.

### LINKED ISSUE #2084: A validator may be included both in a slashing and a voluntary exit in the same block, which causes the block invalid
In a case that a validator is to exit, while it triggers a slashing, then
In block proposal:
According to current spec of honest validators, the block proposer may include the same validator in both slashings and voluntary exits into the same block(since  the exits satisfies the verification conditions found in `exits processing` based on current beacon state).

But in state transition:
The block is not to be accepted by other validators, since, after `process_proposer_slashing` and `process_attester_slashing`, the exit_epoch will be set, while `process_voluntary_exit` will reject the voluntary exits whose exit epochs are set before.

I checked lighthouse, prysm and nimbus, according to my limited investigations, I haven't seen these versions are trying to avoid the conflict.

ralexstokes:
Fixes #2084.

In the current spec, including a voluntary exit in the same block as a slashing would result in an invalid block, implying an invalid slashing. We would like any valid slashings on-chain to remain valid in all contexts to ensure protocol safety.

This PR implements a straightforward solution to this problem by simply ignoring any exits for slashed validators.

NOTE: this potentially opens a DoS vector as any block w/ a slashing can now contain the maximum number of exits for the slashed validators and allow excess networking burden (bigger blocks) for no gain to the protocol. However, given the small number of max exits allowed per block, this doesn't seem like a profitable avenue of attack.

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -1809,6 +1809,11 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
 def process_voluntary_exit(state: BeaconState, signed_voluntary_exit: SignedVoluntaryExit) -> None:
     voluntary_exit = signed_voluntary_exit.message
     validator = state.validators[voluntary_exit.validator_index]
+
+    # Slashed validators cannot voluntarily exit
+    if validator.slashed:
+        return
+
     # Verify the validator is active
     assert is_active_validator(validator, get_current_epoch(state))
     # Verify exit has not been initiated


ralexstokes:
If this seems like something we want to merge, I can add spec tests for this scenario for some added assurance in implementations.

djrtwo:
I think instead that at this point it should be a note in the validator guide. Essentially, this is "covered" in the guide by saying the validation rules need to pass (pretty much "don't build a bad block"), but it apparently is a case not being handled well in clients, so an explicit note is warranted imo

lucianyao:
I agree with @djrtwo, it would be probably more appropriate if reminded in doc "honest validator" the developer to avoid packing the voluntary exit into the block, which would be more efficient

mratsim:
From a client team point of view, something should appear in the spec.
1. Client teams have been audited against the spec Python code, for the part covered by Python code.
2. If this is not caught by one of the other preconditions, pyspec will build bad blocks and this will cause fuzzing issues.

The only place that is even more checked would be a pyspec test vector but this require pyspec to be correct.

Regarding the code, since this is a precondition, I think `assert` would be more in line with the rest of the checks

```python
def process_voluntary_exit(state: BeaconState, signed_voluntary_exit: SignedVoluntaryExit) -> None:
    voluntary_exit = signed_voluntary_exit.message
    validator = state.validators[voluntary_exit.validator_index]

    # Slashed validators cannot voluntarily exit
    if validator.slashed:
        return

    # Verify the validator is active
    assert is_active_validator(validator, get_current_epoch(state))
    # Verify exit has not been initiated
    assert validator.exit_epoch == FAR_FUTURE_EPOCH
    # Exits must specify an epoch when they become valid; they are not valid before then
    assert get_current_epoch(state) >= voluntary_exit.epoch
    # Verify the validator has been active long enough
    assert get_current_epoch(state) >= validator.activation_epoch + SHARD_COMMITTEE_PERIOD
    # Verify signature
    domain = get_domain(state, DOMAIN_VOLUNTARY_EXIT, voluntary_exit.epoch)
    signing_root = compute_signing_root(voluntary_exit, domain)
    assert bls.Verify(validator.pubkey, signing_root, signed_voluntary_exit.signature)
    # Initiate exit
    initiate_validator_exit(state, voluntary_exit.validator_index)
```

djrtwo:
So even if you put in an `assert not validator.slashed` this won't prevent you from building a bad block because when you call `process_voluntary_exit` in isolation, the slashing would not yet be processed.

@mratsim I'm not 100% sure if I understand your comment. Would an explicit note in the validator guide not suffice?

I can write some tests that show this combination of operations and show it fails

djrtwo:
close in favor of #2103

PR #2095, opened 2020-10-08
fix minor typo in validator guide

djrtwo:
in-text call to `get_eth1_vote` was missing a param


protolambda (approved):
LGTM. `get_eth1_vote` contains a description of the parameter, don't think we have to duplicate that.

PR #2097, opened 2020-10-08
Tests compression with Snappy, remove YAML duplicates

protolambda:
Previously the test generator encoded all `ssz` type test outputs as both a `.ssz` file and `.yaml` file.
The YAML was included because of a combination of legacy (it was once the only option) and readability.

I think we can change this now that SSZ has not been changed for a very long time, and this exact way of Snappy block compression is widely used in the eth2 gossipsub topics. So parsing the Snappy SSZ files should not be a problem.

This PR:
- Omits all YAML files where there would previously be both a `.yaml` and a `.ssz`
- Snappy compresses all SSZ outputs, with `.ssz_snappy` extension (thought of `.sszs`, but a litle too unreadable)
- Updated test format docs

The result is that the test generation outputs go from ~11 GB to ~0.5 GB. Generation time is also faster, since it reduces disk IO by a lot.

Slowest test generators are the shuffle tests now, since those are still in YAML, and list shuffling is done the naive way. Total test generation time with `make -j 6 generate_tests` (and no old existing generator outputs) is 23 minutes on a laptop (`Intel i7-6700HQ @ 8x 3.5GHz`). A good improvement on old test generation time.

PR draft, wanted to double-check if generated tests pass (updating ZRNT to try this experimental test format) and check the format docs.


djrtwo:
Considerin this for Altair release

djrtwo (approved):
woo. save all the space

PR #2103, opened 2020-10-14
note in validator guide about slashing and exit operation interaction

djrtwo:
addresses #2084.

* add note about the interaction between slashings and voluntary exits of the same index in the same block
* add block tests for a few conditions

note added new `test_multi_operation.py` testfile that outputs block tests. We are lacking in interesting/complex block tests with operations of multiple/many types. I plan to add a number of tests here in the following week


tests/generators/sanity/main.py
```diff
@@ -34,6 +34,7 @@ def cases_fn() -> Iterable[gen_typing.TestCase]:
 if __name__ == "__main__":
     phase_0_mods = {key: 'eth2spec.test.phase0.sanity.test_' + key for key in [
         'blocks',
+        'multi_operations',
```
hwwhww, line 37: If we want to make `multi_operations` a new handler as the status quo, then we have to modify test format docs: https://github.com/ethereum/eth2.0-specs/tree/aab58e47008e161336dd591bd52e3c39c046a288/tests/formats/sanity

IMHO it's better to merge the `test_multi_operations.py` tests into `test_blocks.py` to share the same test format.

hwwhww (approved):
An issue of the test format, otherwise LGTM!

tests/generators/sanity/main.py
djrtwo: yeah, that's fair. Just trying to get a clean playground to dev in. `test_blocks` is unwieldy!

lucianyao:
Hi @hwwhww , I have seen it was committed to v1.0-candidate, while the default is the dev, I would like to know which branch is the canonical spec to implement?thanks!

hwwhww:
@lucianyao They both are.
- `v1.0-candidate` is the branch for the incoming spec v1.0.0, now only accepts critical fixes or test vector updates. 
- `dev` is currently serving the non-phase-0-launch PRs. e.g., phase 1 specs
- `master` is the release tags branch.

I suppose @djrtwo will merge `v1.0-candidate` into `master` and cut 1.0.0, and then backport to `dev`. If you're finding phase 0 issues, I suggest you look at `v1.0-candidate` branch. Thank you! :)

Issue #1932, opened 2020-06-24
Side effects in functions


franck44:
The specs seem to use different types of signatures for functions. for instance:

Case 1:
```
def state_transition(state: BeaconState, signed_block: SignedBeaconBlock, validate_result: bool=True) -> BeaconState:
 block = signed_block.message
    # Process slots (including those with no blocks) since block
    process_slots(state, block.slot)
    # Verify signature
    if validate_result:
        assert verify_block_signature(state, signed_block)
    # Process block
    process_block(state, block)
    # Verify state root
    if validate_result:
        assert block.state_root == hash_tree_root(state)
    # Return post-state
    return state
```

and case 2:

```
def process_slots(state: BeaconState, slot: Slot) -> None:
    assert state.slot < slot
    while state.slot < slot:
        process_slot(state)
        # Process epoch on the start slot of the next epoch
        if (state.slot + 1) % SLOTS_PER_EPOCH == 0:
            process_epoch(state)
        state.slot = Slot(state.slot + 1)
```

In case 1, it seems redundant to return a `BeaconState` because the `state` variable is passed by reference (I am not Python expert, but AFAIK `state` is mutable and hence passed by reference.) But the fact that the function's return type is a `BeaconState` also gives a false sense of immutability of the input object `state`. The result is actually in `state` (side-effect and in the returned value.

In case 2, the mutability of the parameter `state` is used and instead of returning a `state`, `process_slots` modifies `state`, so the result is obtained by a side-effect on input parameter `state` (passed by reference).

**Main question:** How is the choice of a function signature determined in the specs? 
Some functions rely on side-effects, and some may not.

Python parameters' passing and the interference with local storage is known to be rather cumbersome as described in the additional comment below, so a clear and principled way of specifying functions may help reducing of problems that can arise in implementations.

**Additional comment**
For instance (again, I am not a Python expert, so if this is wrong, please correct me), the following code may not have any side-effect on `state` and leave the input parameter (although passed by reference) `state` unchanged:
```
class BeaconState:
    def __init__(self, slot):
        self.slot = slot

def process_slots(state: BeaconState, slot: int) -> None:
    s = BeaconState(0)
    s.slot = state.slot
    # the value of s is the same as the initial value of state
    state = s
    while state.slot < slot:
        state.slot = (state.slot + 1)

# k is a BeaconState
k = BeaconState(2)
print (k.slot) # Result is 2
process_slots(k, 4)
print (k.slot) # result is still 2
```

[...]

PR #2104, opened 2020-10-15
Make `state_transition` not return post-state

hwwhww:
### Issue
1. It's arguable that the mutability side effect is smell (see #1932 discussion). Although I personally want to remove the side effect, we decide to keep it for now for the code concision and stability.
2. What worse than the mutability side effect is that the `state_transition` provides two patterns: (i) it mutates the given `state` and also (ii) returns the post-state. We're using (i) in the whole specs and (ii) only exists in `state_transition`.

### How did I fix it
Make `state_transition` not return post-state. The given pre-state should have been mutated to post-state.


djrtwo (approved):

djrtwo (changes requested):
taking back my approval. going to make sure this isn't used in any unexpected way in tests/test-gen

djrtwo (approved):

Issue #2085, opened 2020-09-29
Prevent validator signing far into the future attestations


alonmuroch:
Current validator implementations put a lot of trust in constructing the attestation data object on the beacon node side, mostly the validator checks the attestation data is not slashable. 

A problem can occur if the attestation data includes a target epoch well into the future. If the validator signs it, from that point on, most slashing protection mechanism will prevent that validator from signing any attestation until that far into the future epoch is reached. 
That can cause major loses and potential inactivity slashing.

It might be a good idea to add some reference to this in the honest validator guide (thanks for the suggestion @dankrad )

alonmuroch:
@hwwhww should I prepare a PR for this?

djrtwo:
Agreed that we can put a note in the guide for clients to be aware of this pitfall and to add sanity checks on time and epochs skips

alonmuroch:
@djrtwo @hwwhww  https://github.com/ethereum/eth2.0-specs/pull/2107/files

PR #2107, opened 2020-10-20
added best practices section according to https://github.com/ethereum…

alonmuroch:
addresses #2085


specs/phase0/validator.md
```diff
@@ -604,3 +605,11 @@ Specifically, when signing an `Attestation`, a validator should perform the foll
 2. Generate and broadcast attestation.
 
 If the software crashes at some point within this routine, then when the validator comes back online, the hard disk has the record of the *potentially* signed/broadcast attestation and can effectively avoid slashing.
+
+## Protection best practices
+A validator client should be considered standalone and should consider the node it's connected to as un-trusted. This means that the validator client should protect:
```
dankrad, line 610:
```suggestion
A validator client should be considered standalone and should consider the beacon node as untrusted. This means that the validator client should protect: 
```

specs/phase0/validator.md
```diff
@@ -604,3 +605,11 @@ Specifically, when signing an `Attestation`, a validator should perform the foll
 2. Generate and broadcast attestation.
 
 If the software crashes at some point within this routine, then when the validator comes back online, the hard disk has the record of the *potentially* signed/broadcast attestation and can effectively avoid slashing.
+
+## Protection best practices
+A validator client should be considered standalone and should consider the node it's connected to as un-trusted. This means that the validator client should protect: 
+1) Private keys - private keys should be protected from ever being exported out accidentally or by an attacker. Attestations and blocks should be signed internally in the process itself, keys should always be encrypted on disk.
```
dankrad, line 611:
```suggestion
1) Private keys -- private keys should be protected from being exported accidentally or by an attacker.
```
Second sentence is not mandatory. It makes sense to keep a key in an HSM, for example. Since processes are not protected entities of any kind, it is not a security improvement to require signing in the same process. Also, encrypting keys on disk does not add security if the password is also stored on disk, which is a necessity for validator nodes to be able to restart automatically.

specs/phase0/validator.md
```diff
@@ -604,3 +605,11 @@ Specifically, when signing an `Attestation`, a validator should perform the foll
...
+2) Slashing - before a validator client signs an attestation/ block it should validate the data, check against a local slashing db (do not sign slashable attestation/ block) and update its internal slashing db with the new signed object.
```
dankrad, line 612:
```suggestion
2) Slashing -- before a validator client signs a message it should validate the data, check it against a local slashing database (do not sign a slashable attestation or block) and update its internal slashing database with the newly signed object.
```

specs/phase0/validator.md
```diff
@@ -604,3 +605,11 @@ Specifically, when signing an `Attestation`, a validator should perform the foll
...
+3) Recovered validator - Recovering a validator from a private key will result in an empty local slashing db. A best practice is to import (from a trusted source) that validator's attestation history.
```
dankrad, line 613:
```suggestion
3) Recovered validator -- Recovering a validator from a private key will result in an empty local slashing db. Best practice is to import (from a trusted source) that validator's attestation history. 
```

specs/phase0/validator.md
```diff
@@ -604,3 +605,11 @@ Specifically, when signing an `Attestation`, a validator should perform the foll
...
+1) Private keys - private keys should be protected from ever being exported out accidentally or by an attacker. Attestations and blocks should be signed internally in the process itself, keys should always be encrypted on disk.
+2) Slashing - before a validator client signs an attestation/ block it should validate the data, check against a local slashing db (do not sign slashable attestation/ block) and update its internal slashing db with the new signed object.
+3) Recovered validator - Recovering a validator from a private key will result in an empty local slashing db. A best practice is to import (from a trusted source) that validator's attestation history. 
+4) Far future signing requests - A validator client can be requested to sign a far into the future attestation, resulting in a valid non slashable request. If the validator client signs it will result in it blocking itself from attesting any other attestation until the beacon-chain reaches that far into the future epoch. This will result in an inactivity leak and potential slashing.
```
dankrad, line 614:
```suggestion
4) Far future signing requests -- A validator client can be requested to sign a far into the future attestation, resulting in a valid non-slashable request. If the validator client signs this message, it will result in it blocking itself from attesting any other attestation until the beacon-chain reaches that far into the future epoch. This will result in an inactivity leak and potential ejection due to low balance. 
```

specs/phase0/validator.md
```diff
@@ -604,3 +605,11 @@ Specifically, when signing an `Attestation`, a validator should perform the foll
...
+1) Private keys - private keys should be protected from ever being exported out accidentally or by an attacker. Attestations and blocks should be signed internally in the process itself, keys should always be encrypted on disk.
+2) Slashing - before a validator client signs an attestation/ block it should validate the data, check against a local slashing db (do not sign slashable attestation/ block) and update its internal slashing db with the new signed object.
+3) Recovered validator - Recovering a validator from a private key will result in an empty local slashing db. A best practice is to import (from a trusted source) that validator's attestation history. 
+4) Far future signing requests - A validator client can be requested to sign a far into the future attestation, resulting in a valid non slashable request. If the validator client signs it will result in it blocking itself from attesting any other attestation until the beacon-chain reaches that far into the future epoch. This will result in an inactivity leak and potential slashing. 
+A validator client should prevent itself from signing such requests by estimating the current slot with some deviation.
```
dankrad, line 615:
```suggestion
A validator client should prevent itself from signing such requests by: a) keeping a local time clock if possible and following best practices to stop time server attacks and b) refusing to sign, by default, any message that has a large (>6h) gap from the current slashing protection database indicated a time "jump" or a long offline event. The administrator can manually override this protection to restart the validator after a genuine long offline event.
```

dankrad (changes requested):

specs/phase0/validator.md
alonmuroch: You could hold the password in an HSM which is more likely to happen sooner as it will take time for BLS keys to be supported. 
I see it as best design practice for future improvements, which is better than just leaving keys as plain text.
But you are right that for now it's probably very similar inn terms of security.

alonmuroch:
@dankrad thanks for the review, committed the changed you've suggested.

specs/phase0/validator.md
dankrad: Of course an HSM makes sense. But the security of
1. A validator client running form unencrypted keys
2. A validator client running with encrypted keys, with the password given on the command line (so stored on disk to load when the computer starts up)

is the same. Case 2 might feel safer but is security theatre.

alonmuroch:
@dankrad any other comments?

specs/phase0/validator.md
```diff
@@ -604,3 +605,11 @@ Specifically, when signing an `Attestation`, a validator should perform the foll
 2. Generate and broadcast attestation.
 
 If the software crashes at some point within this routine, then when the validator comes back online, the hard disk has the record of the *potentially* signed/broadcast attestation and can effectively avoid slashing.
+
+## Protection best practices
```
djrtwo, line 609:
```suggestion
## Protection best practices

```

specs/phase0/validator.md
```diff
@@ -604,3 +605,11 @@ Specifically, when signing an `Attestation`, a validator should perform the foll
...
+A validator client should be considered standalone and should consider the beacon node as untrusted. This means that the validator client should protect:
```
djrtwo, line 610:
```suggestion
A validator client should be considered standalone and should consider the beacon node as untrusted. This means that the validator client should protect:

```

specs/phase0/validator.md
```diff
@@ -604,3 +605,11 @@ Specifically, when signing an `Attestation`, a validator should perform the foll
...
+3) Recovered validator -- Recovering a validator from a private key will result in an empty local slashing db. Best practice is to import (from a trusted source) that validator's attestation history. See [EIP 3076](https://github.com/ethereum/EIPs/pull/3076/files) for a standard slashing interchange format.
+4) Far future signing requests -- A validator client can be requested to sign a far into the future attestation, resulting in a valid non-slashable request. If the validator client signs this message, it will result in it blocking itself from attesting any other attestation until the beacon-chain reaches that far into the future epoch. This will result in an inactivity leak and potential ejection due to low balance. 
+A validator client should prevent itself from signing such requests by: a) keeping a local time clock if possible and following best practices to stop time server attacks and b) refusing to sign, by default, any message that has a large (>6h) gap from the current slashing protection database indicated a time "jump" or a long offline event. The administrator can manually override this protection to restart the validator after a genuine long offline event.
```
djrtwo, line 615:
```suggestion

A validator client should prevent itself from signing such requests by: a) keeping a local time clock if possible and following best practices to stop time server attacks and b) refusing to sign, by default, any message that has a large (>6h) gap from the current slashing protection database indicated a time "jump" or a long offline event. The administrator can manually override this protection to restart the validator after a genuine long offline event.
```

djrtwo (approved):

alfredomontes37826-ux (approved):

PR #2110, opened 2020-10-20
Multi operation tests

djrtwo:
Add a number of randomized multi-operation block tests.
We don't really have any block tests that have different operations (other than slash-exit tests added in #2103)

Note, these tests take ~2 minutes each on my laptop on minimal settings. We might consider having a test vectors that _only run in generators_. These aren't terribly valuable to the CI but can be valuable to clients.


djrtwo:
Merging but noting that getting a `parametrize` solution in is highly desirable for this type of test

PR #2117, opened 2020-10-30
[WIP] [testgen] pre-state references

hwwhww:
Based on the previous discussion with @djrtwo

### Proposal
- Use test **runner** as the base unit to create the "shared files".
- The shared files are laid under test runner (e.g., `sanity`)
- Move pre-state (`pre`) to the shared files. The file names are `pre_$PRE_HASH_TREE_ROOT.ssz` or `pre_$PRE_HASH_TREE_ROOT.yaml`
- Add meta field `pre: pre_$PRE_HASH_TREE_ROOT` to look up the corresponding file.
    - Clients need to add `.ssz` or `.yaml` to locate the file.

### Improvements

This draft PR currently only changed `sanity` testgen as a demo, so the below data is only from `sanity` tests.
Output file tree: https://gist.github.com/hwwhww/8f9b1d657ab3d6aeaba2982279aa57cf
#### 1. Compare the sizes:

| runner | version | file count | total size
| - | - | - | - 
| `sanity` | Before (a34970f8a3a1ee32e06b845bf3de6d70e69090d0) | 1,288  | 1.17 GB
| `sanity` | After    | 1,082  | 702.1 MB

#### 2. stats of duplicate files
Before this PR, some files are duplicate for >20 times. This is the stats of each `pre` file:
https://gist.github.com/hwwhww/c39eda0b582cbca40f94c6be35bbd7f7

#### Pro:
1. Should be able to reduce the output file size vastly!
#### Cons:
1. Need to update client test suites
2. I made `eth2spec.test.utils.vector_test` to compute hash tree root of *all* SSZ output fields, so It may take more time to run testgen.

### TODOs:
- [ ] Discuss the format
- [ ] Refactor
- [ ] Update other runners
- [ ] Edit test format
- [ ] Enable other tests
- [ ] Move other fields to shared files?

### DIFF
--- a/tests/core/gen_helpers/gen_base/gen_runner.py
+++ b/tests/core/gen_helpers/gen_base/gen_runner.py
@@ -1,7 +1,8 @@
 import argparse
+import os
 from pathlib import Path
 import sys
-from typing import Iterable, AnyStr, Any, Callable
+from typing import Iterable, AnyStr, Any, Callable, Dict, Optional
 import traceback
 
 from ruamel.yaml import (
@@ -9,9 +10,11 @@ from ruamel.yaml import (
 )
 
 from gen_base.gen_typing import TestProvider
+from gen_base.gen_typing import SSZLookup
 
 from eth2spec.test import context
 from eth2spec.test.exceptions import SkippedTest
+from eth2spec.utils.hash_function import hash as sha256
 
 
 # Flag that the runner does NOT run test via pytest
@@ -42,7 +45,7 @@ def validate_configs_dir(path_str):
     return path
 
 
-def run_generator(generator_name, test_providers: Iterable[TestProvider]):
+def run_generator(generator_name, test_providers: Iterable[TestProvider], pre_state_path_lookup: SSZLookup):
     """
     Implementation for a general test generator.
     :param generator_name: The name of the generator. (lowercase snake_case)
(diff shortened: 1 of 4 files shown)


hwwhww:
Alternative 1: add `eth2.0-spec-tests/shared_files/` folder to store **all**  shared files.

Pros: cleaner runner folders should be able reduce more output sizes 
Cons: less good for phase0/1 file separation

Other alternatives: have `shared_files` under `<config name>/` or `<fork or phase name>/`

protolambda:
Nice improvement, but I'm worried that it will not be effective if we start creating many different pre-states with parameters. The states are not compressed, or partially deduplicated. So in such situation we just end up having all pre-states in a different directory, without much reduction in size.

There is another experimental PR which applies snappy compression and omits the yaml: #2097. This reduced *total* test output size: from ~11 GB to ~0.5 GB.

I like the approach with pre-state references better, but if it's not effective enough we should improve it more before we force a test format change on client devs.

Test states really contain a lot of similar data, with subtle chosen changes between them. We should try and design around that to come up with something efficient.

protolambda:
One thing that may work better is to share pre-states, but also share them if they are similar enough. Then store a diff in the test to build the actual pre-state with.

E.g.

Shared state:
```yaml
genesis_time: 0
genesis_validators_root: '0xef82b97f46b3decc813a5c37fe2cb679d084de70ae42e1eaef3a6a90da2b361a'
slot: 31
fork: {previous_version: '0x00000000', current_version: '0x00000000', epoch: 0}
latest_block_header: {slot: 0, proposer_index: 0, parent_root: '0x0000000000000000000000000000000000000000000000000000000000000000',
  state_root: '0xf044817e3ec8bd8bef05ee5d4442e9f51af69890b5d2680dcba5000bd82ff57e',
  body_root: '0xccb62460692be0ec813b56be97f68a82cf57abc102e27bf49ebf4190ff22eedd'}
block_roots: ['0x76c67f6073a49ea164c4fd2cdd7cf878c91631715df4739a70967dfcce269237',
# ... ~5 MB more data...
```
Test state diff, for a state starting at a different slot:
```diff
3c3
< slot: 31
---
> slot: 65
```

The challenge here is that YAML is far from an optimal format, and something to explicitly implement support for in data types as a client. Client like Prysm couldn't make it work well with the default yaml tooling for their language and typing choices.
So SSZ test files were introduced, as the format is already supported by clients anyway.

If we start changing the test format I would drop YAML, and just focus on 1 format.
If there's some simple binary-diff format we can use, and apply it to all SSZ files, I think we can reduce total test size to way under that 0.5 GB achieved with compression.

djrtwo:
@hwwhww do you still want to consider this? or should we close it in favor of @protolambda's snappy compression

PR #2124, opened 2020-11-06
fix multi-operation test runner helper

djrtwo:
Test gen was rendering the post state for pre and post. Fix order of operations to render properly.

Thanks @ajsutton!

Note, this has already been patched to the test vector release. Plans is to merge into master here. Then backport to dev. Will _not_ cut a new spec release


protolambda (approved):
Good find by Adrian :clap:

PR #2130, opened 2020-11-12
Added standalone light client patch

vbuterin:
Adding the standalone beacon chain changes to add light client support as a separate spec patch, independent of phase 1, to enable it to be tested and potentially rolled out sooner than the rest of phase 1.


JustinDrake:
Two main substantive suggestions:

1) Increase `LIGHT_CLIENT_COMMITTEE_SIZE` to 256 to address the [RANDAOtage attack vector](https://notes.ethereum.org/iMxxlEkuQMiPkEL1S6SfbQ).
2) Add a `pubkeys_sum` field to `CompactCommittee` which sums all the pubkeys. This unlocks an optimisation whereby light clients only have to download pubkeys of validators that do not vote. For a 2/3 threshold this is at least a 2x optimisation. In the expected/optimistic scenario where 95% of the light client committee votes this is a ~20x optimisation.

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,208 @@
...
+
+#### Light client processing
+
+```python
+def process_light_client_signature(state: BeaconState, block_body: BeaconBlockBody) -> None:
+    committee = get_light_client_committee(state, get_current_epoch(state))
+    previous_slot = max(state.slot, 1) - 1
+    previous_block_root = get_block_root_at_slot(state, previous_slot)
+
+    # Light clients sign over the previous block root
```
ralexstokes, line 157: May be a heavy-handed edit but I'm trying to sort out what kind of security guarantees we get here (and how) so I don't think it hurts to be as precise as possible...

```suggestion
    # Light client committees sign over the previous block root
```

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,208 @@
...
+        state.validators[participant].pubkey for participant in participants
+    ]
+    
+    assert bls.FastAggregateVerify(signer_pubkeys, signing_root, block_body.light_client_signature)
+    
+    # Process rewards
+    total_reward = Gwei(0)
+    active_validator_count = len(get_active_validator_indices(beacon_state, get_current_epoch(state)))
+    for participant in participants:
+        reward = get_base_reward(state, participant) * active_validator_count // len(committee)
```
ralexstokes, line 177: do we want to further reduce the reward by `SLOTS_PER_EPOCH`? e.g.

```python
reward = ...
reward = reward // SLOTS_PER_EPOCH
```

`BASE_REWARDS_PER_EPOCH` has been increased to `5` but as it stands this reward is assessed per slot.

ralexstokes: and it *feels* like we also want to reduce the per-participant reward (proportionately) to account for the proposer reward... unless i'm missing something w/ the rewards rationale here.

JustinDrake:
Fine combed the document :)

1) (bug fix) Called `compute_shuffled_index` with `active_validator_count` (as opposed to `active_validator_indices`).
2) Used a `Bitlist` (as opposed to a `Bitvector`) for `BeaconBlockBody.sync_committee_bits` for consistency with `SyncCommittee.pubkeys` and `SyncCommittee.compact_validators`.
3) Added `assert len(body.sync_committee_bits) == len(committee_indices)`.
4) Should `current_sync_committee` and `next_sync_committee` be initialised, e.g. in `initialize_beacon_state_from_eth1`?
5) Should we sign over both `previous_block_root` and `previous_slot`? This may facilitate gossipping sync committee signatures as well as simplify and optimise light client implementations.
6) Should we rotate `current_sync_committee` and `next_sync_committee` only if the previous rotation falls under finality? If not, how are long-range (greater than `EPOCHS_PER_SYNC_COMMITTEE_PERIOD`) forks handled?
7) Should sync committees explicitly sign over `state.finalized_checkpoint`? This would simplify light client implementations and save them from downloading and verifying unnecessary Merkle branches.
8) What do you suggest as the precise honest behaviour of a sync committee member? Explicitly spelling it out would help me go through edge cases.
9) Should we consider adding sync committee slashing conditions? For example, we don't want sync committtee members equivocating over `state.finalized_checkpoint`.
10) In the exceptional case where `active_validator_count < MAX_SYNC_COMMITTEE_SIZE`, what do you think of "padding" the sync committee with duplicate validators? This handling of the exceptional case (which really ought to never happen on mainnet) would simplify light client implementations because the variable-size `Bitlist` and `List` would be fixed-sized `Bitvector` and `Vector`. The state transition function would also be simplified a bit.
11) What do you think of increasing `MAX_SYNC_COMMITTEE_SIZE` further, e.g. to 512 or 1024? The rationale is that the marginal on-chain overhead is negligeable (especially with fancy batched point addition) and light clients could choose to increase their committee size to improve robustness against adaptive attacks (as well as improve liveness in the worst case by shielding themselves against sampling variance). It also allows fancier light clients that sample random subsets of the sync committee, again for robustness. This does go against the `pubkeys_aggregate` optimisation.
12) Does the micro-optimisation break down when the networking latency is poor enough that it takes more than one slot for the block proposer to gather a good sync committee signature? What about relaxing the latency for the purpose of micro-optimisation? (Note that sync committee signatures that don't make it onchain would still be valuable offchain for light clients.)
13) Do the cosmetic changes look good to you? (I may have introduced a bug or two.)

vbuterin:
> Should we rotate current_sync_committee and next_sync_committee only if the previous rotation falls under finality? If not, how are long-range (greater than EPOCHS_PER_SYNC_COMMITTEE_PERIOD) forks handled?

I'm inclined to say, no need for special cases here, if none of the committees in a given period get up to 2/3 the client should just pick the block that has the most light client participants. Obviously in such a case the client would wait for some time to verify that no better block is available.

> Should we sign over....

What's wrong with just having the light client proof package that gets sent over the wire contain SSZ branches? Seems more general-purpose...

> What do you suggest as the precise honest behaviour of a sync committee member?

Halfway through each slot, publish an attestation to what you think is the the head of the chain.

> Should we consider adding sync committee slashing conditions?

Hmm... I'm inclined to say don't bother, because if the committee is faulty then nothing stops them from feeding clients the wrong chain _consistently_, and there's no way for the sync committee to have that high a level of security anyway.

> In the exceptional case where active_validator_count < MAX_SYNC_COMMITTEE_SIZE, what do you think of "padding" the sync committee with duplicate validators? 

Hmm.... instinctively it seems wasteful and unnatural. Is it really that complicated to have variable size lists? I guess another option would be to have them be vectors and pad the pubkeys and bits with zeroes to represent that there's nothing there.

> What do you think of increasing MAX_SYNC_COMMITTEE_SIZE further, e.g. to 512 or 1024? 

Seems okay as long as light clients have a choice to only use part of the committee if they so desire, so eg. `sum_of_pubkeys` would need to be extended into an array that contains the sum of the first `2**k` for each `k`. That said, are we _that_ worried about a committee of size 256 being attacked?

> Does the micro-incentivisation break down when the networking latency is poor enough that it takes more than one slot for the block proposer to gather a good sync committee signature?

Remember that the signatures would still be floating around the gossip net even if only a few of them make it onto the chain, so technically as long as there's even a small percent change of being included the incentives to publish should work fine. But if we want to make it easier to read the signatures from just the chain, I guess one route would be to keep retrying a slot if it gets <50% participation (eg. for a maximum of 4 retries).

vbuterin:
Cosmetic changes look good to me!

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,202 @@
...
+
+```python
+def get_sync_committee_indices(state: BeaconState, epoch: Epoch) -> Sequence[ValidatorIndex]:
+    """
+    Return the sync committee indices for a given state and epoch.
+    """
+    start_epoch = Epoch((max(epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD, 1) - 1) * EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
+    active_validator_count = uint64(len(get_active_validator_indices(state, start_epoch)))
+    sync_committee_size = min(active_validator_count, MAX_SYNC_COMMITTEE_SIZE)
+    seed = get_seed(state, base_epoch, DOMAIN_SYNC_COMMITTEE)
```
hwwhww, line 136: `start_epoch`? Fixed in https://github.com/ethereum/eth2.0-specs/pull/2133
```suggestion
    seed = get_seed(state, start_epoch, DOMAIN_SYNC_COMMITTEE)
```

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,202 @@
...
+```python
+def get_sync_committee(state: BeaconState, epoch: Epoch) -> SyncCommittee:
+    """
+    Return the sync committee for a given state and epoch.
+    """
+    indices = get_sync_committee_indices(state, epoch)
+    validators = [state.validators[index] for index in indices]
+    pubkeys = [validator.pubkey for validator in validators]
+    compact_validators = [compactify_validator(i, v.slashed, v.effective_balance) for i, v in zip(indices, validators)]
+    return SyncCommittee(pubkeys, bls.AggregatePubkeys(pubkeys), compact_validators)
```
hwwhww, line 151:
```suggestion
    return SyncCommittee(pubkeys, bls.AggregatePKs(pubkeys), compact_validators)
```

fixed in https://github.com/ethereum/eth2.0-specs/pull/2133

hwwhww (commented):
@vbuterin 

I made the linter pass: https://github.com/ethereum/eth2.0-specs/pull/2133

specs/lightclient/beacon-chain.md
JustinDrake: I knew I must have introduced a bug somewhere 😂 

@hwwhww: Does everything else pass? For example, are all the types casted properly and the ToC properly built?

JustinDrake: Used `base_epoch` everywhere!

JustinDrake:
> I'm inclined to say, no need for special cases here, if none of the committees in a given period get up to 2/3 the client should just pick the block that has the most light client participants. Obviously in such a case the client would wait for some time to verify that no better block is available.

What about safety?

Let's say there's a 1/3 attacker, 1/3 honest attesting (and light-client voting) for fork A, 1/3 honest attesting (and light-client voting) for fork B.

Now the attacker attests for fork A and light-client votes for fork B. Fork A finalises but light clients are stuck on fork B.

> What's wrong with just having the light client proof package that gets sent over the wire contain SSZ branches?

Light client implementations will often be hyper-secure and/or hyper-optimised. For example, consider a light client implementation in a zkSNARK circuit. Or a light client implementation in a metered virtual machine like the EVM. Or a light client implementation in a secure enclave. Or a light client implementation in a formal verification framework. The implementation costs are orders of magnitude higher than Python.

Avoiding unnecessary data like Merkle branches and as well unnecessary complexities like variable-sized Merkle branches may pay off big time.

> Seems more general-purpose

Having just the previous root is definitely more minimalist :)

> instinctively it seems wasteful

Do you mean wasteful when there are less than 256 active validators? Optimising for that edge case which should never happen in practice feels like a micro-optimisation.

> are we that worried about a committee of size 256 being attacked?

I'm worried! A single committee break simultaneously breaks every Eth2-to-X bridge (where X = Cosmos, Near, Polkadot, etc.) and there could easily be $1B+ (or $10B+) in extractable value for an attacker.

Because `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` in on the order of 1 day it has "human scale". By that I mean that 1 day is sufficient for an attacker to 1) publicise a $1m-per-validator bribing attack (e.g. on Twitter and Reddit) and 2) have corrupted validators manually perform an action to claim their $1m bribe. How many rational validators would say no to a $1m bribe? Also note that the attack cost is zero because there is no enshrined slashing.

vbuterin:
> Avoiding unnecessary data like Merkle branches and as well unnecessary complexities like variable-sized Merkle branches may pay off big time.

We don't have any variable-sized Merkle branches though; SSZ was deliberately designed to make any access path neatly correspond to a single generalized index. I'd also add that if you want to know the state root, usually it's not because you care about the root, it's because you care about something _in_ the state, so you would need Merkle paths anyway.

> Let's say there's a 1/3 attacker, 1/3 honest attesting (and light-client voting) for fork A, 1/3 honest attesting (and light-client voting) for fork B.

Why would this situation persist for an entire day? Wouldn't the fork resolve and there be a block that has 2/3 voting?

To be clear, the fork choice rule I am proposing is *not* slot-by-slot. It's "start from the last block you are confident about, take the validator set from it, then find the **descendant** of that block that was signed by the most participants of that validator set, and repeat"

> Do you mean wasteful when there are less than 256 active validators? Optimising for that edge case which should never happen in practice feels like a micro-optimisation.

I mean wasteful code-wise. It feels like it would still lead to annoying complexity (eg. where else in the code can the same validator be part of a committee twice??!). If we want fixed size the better path does seem to be to pad the committee with fake pubkeys that no one can sign with.

> Because EPOCHS_PER_SYNC_COMMITTEE_PERIOD in on the order of 1 day it has "human scale". By that I mean that 1 day is sufficient for an attacker to 1) publicise a $1m-per-validator bribing attack (e.g. on Twitter and Reddit) and 2) have corrupted validators manually perform an action to claim their $1m bribe. How many rational validators would say no to a $1m bribe? Also note that the attack cost is zero because there is no enshrined slashing.

OK fair! I'm okay with pushing the committee size up as long as the `pubkeys_aggregate` is either removed or turns into a list of prefix-sums of powers of two.

On second thought, I'd be ok with some kind of slashing if we can figure out the right conditions; I guess something like "if you vote for block X but in the same epoch you used a block that conflicts with X as a last-justified-block you get slashed" could work. Main challenge is that we'd like to make sure that it's easy for anti-slash DBs to verify that a validator isn't slashing themselves.

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,132 @@
...
+### `LightClientUpdate`
+
+```python
+class LightClientUpdate(Container):
+    # Updated beacon header (and authenticating branch)
+    header: BeaconBlockHeader
+    # Sync committee signature to that header
+    aggregation_bits: Bitlist[MAX_SYNC_COMMITTEE_SIZE]
+    signature: BLSSignature
+    header_branch: Vector[Bytes32, BEACON_CHAIN_ROOT_IN_SHARD_BLOCK_HEADER_DEPTH]
```
ralexstokes, line 46: seems like this is not used anymore

```suggestion
```

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,132 @@
...
+| `SYNC_COMMITTEES_GENERALIZED_INDEX` | `GeneralizedIndexConcat(GeneralizedIndex(BeaconBlock, 'state_root'), GeneralizedIndex(BeaconState, 'current_sync_committee'))` |
+| `FORK_GENERALIZED_INDEX` | `GeneralizedIndexConcat(GeneralizedIndex(BeaconBlock, 'state_root'), GeneralizedIndex(BeaconState, 'fork'))` |
+
+## Containers
+
+### `LightClientUpdate`
+
+```python
+class LightClientUpdate(Container):
+    # Updated beacon header (and authenticating branch)
```
ralexstokes, line 41:
```suggestion
    # Updated beacon header
```

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,132 @@
...
+    header: BeaconBlockHeader
+    # Sync committee signature to that header
+    aggregation_bits: Bitlist[MAX_SYNC_COMMITTEE_SIZE]
+    signature: BLSSignature
+    header_branch: Vector[Bytes32, BEACON_CHAIN_ROOT_IN_SHARD_BLOCK_HEADER_DEPTH]
+    # Updates fork version
+    new_fork: Fork
+    fork_branch: Vector[Bytes32, log_2(FORK_GENERALIZED_INDEX)]
+    # Updated period committee (and authenticating branch)
```
ralexstokes, line 50:
```suggestion
    # Updated sync committee (and authenticating branch)
```

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,132 @@
...
+    # Fork version data
+    fork_version: Version
+    # period committees corresponding to the beacon header
+    current_sync_committee: SyncCommittee
+    next_sync_committee: SyncCommittee
+```
+
+## Light client state updates
+
+The state of a light client is stored in a `memory` object of type `LightClientMemory`. To advance its state a light client requests an `update` object of type `LightClientUpdate` from the network by sending a request containing `(memory.shard, memory.header.slot, slot_range_end)`. It calls `validate_update(memory, update)` on each update that it receives in response. If `sum(update.aggregate_bits) * 3 > len(update.aggregate_bits) * 2` for any valid update, it accepts that update immediately; otherwise, it waits around for some time and then finally calls `update_memory(memory, update)` on the valid update with the highest `sum(update.aggregate_bits)`.
```
ralexstokes, line 73: > If `sum(update.aggregate_bits) * 3 > len(update.aggregate_bits) * 2` for any valid update

is there a reason we are not summing effective stake (via `effective_balance` found in `memory.current_sync_committee`?

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,132 @@
...
+    # Verify that it actually updates to a newer slot
+    assert update.header.slot > memory.header.slot
+    
+    # Convenience as independent variable for convenience
+    committee = memory.current_sync_committee if new_period == current_period else memory.next_sync_committee
+    assert len(update.aggregation_bits) == len(committee)
+    
+    # Verify signature
+    active_pubkeys = [p for (bit, p) in zip(update.aggregation_bits, committee.pubkeys) if bit]
+    domain = compute_domain(DOMAIN_SYNC_COMMITTEE, memory.version)
```
ralexstokes, line 94:
```suggestion
    domain = compute_domain(DOMAIN_SYNC_COMMITTEE, memory.fork_version)
```

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,132 @@
...
+
+### `LightClientMemory`
+
+```python
+class LightClientMemory(Container):
+    # Beacon header which is not expected to revert
+    header: BeaconBlockHeader
+    # Fork version data
+    fork_version: Version
+    # period committees corresponding to the beacon header
```
ralexstokes, line 66:
```suggestion
    # sync committees corresponding to the beacon header
```

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,132 @@
...
+    # Verify the update does not skip a period
+    current_period = compute_epoch_at_slot(memory.header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+    new_epoch = compute_epoch_of_shard_slot(update.header.slot)
+    new_period = new_epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+    assert new_period in (current_period, current_period + 1)  
+    
+    # Verify that it actually updates to a newer slot
+    assert update.header.slot > memory.header.slot
+    
+    # Convenience as independent variable for convenience
```
ralexstokes, line 88: style suggestions: 

```suggestion
    # Independent variable for convenience
```
or even just remove this line...

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,132 @@
...
+    # Verify Merkle branches of new info
+    assert is_valid_merkle_branch(
+        leaf=hash_tree_root(update.new_fork),
+        branch=update.fork_branch,
+        depth=log2(FORK_GENERALIZED_INDEX),
+        index=FORK_GENERALIZED_INDEX % 2**log2(FORK_GENERALIZED_INDEX),
+        root=hash_tree_root(update.header),
+    )
+    assert is_valid_merkle_branch(
+        leaf=hash_tree_root(update.current_sync_committee),
```
ralexstokes, line 107:
```suggestion
        leaf=hash_tree_root(update.new_current_sync_committee),
```

specs/lightclient/sync-protocol.md
ralexstokes, line 94: i think it is worth considering adding the `genesis_validators_root` as input to `compute_domain` for the same reasons we have it on the beacon chain.

it only adds a constant 32 byte overhead for the light client (which can even be hard-coded into client code) and reduces the scope of admissible updates a light client would even bother with

specs/lightclient/sync-protocol.md
vbuterin: It vastly simplifies the light client spec to rely only on the bits. We can make up for it by making membership in the sync committee itself probabilistic.

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,132 @@
...
+- [Light client state updates](#light-client-state-updates)
+    - [`validate_update`](#validate_update)
+    - [`update_memory`](#update_memory)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+## Introduction
+
+Ethereum 2.0 is designed to be light client friendly. This allows low-resource clients such as mobile phones to access Ethereum 2.0 with reasonable safety and liveness. It also facilitates the development of "bridges" to external blockchains. This document suggests a minimal light client design for the beacon chain that uses the concept of "sync committees" introduced in [./beacon-chain.md](the the light-client-friendliness beacon chain extension).
```
alonmuroch, line 27: redundant "the" in the link "the the light-client-..."

alonmuroch:
> Because `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` in on the order of 1 day it has "human scale". By that I mean that 1 day is sufficient for an attacker to 1) publicise a $1m-per-validator bribing attack (e.g. on Twitter and Reddit) and 2) have corrupted validators manually perform an action to claim their $1m bribe. How many rational validators would say no to a $1m bribe? Also note that the attack cost is zero because there is no enshrined slashing.

Is there a way to introduce some kind of fork-choice for a light client in case a corrupt committee signs an invalid block? 
I'm thinking about an EVM based light client and how it might recover. 
Or is the security assumption says a big enough committee can't be corrupted to 2/3 of its members?

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,153 @@
...
+    # Verify update does not skip a sync committee period
+    old_period = compute_epoch_at_slot(old_snapshot.header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+    new_period = compute_epoch_at_slot(new_snapshot.header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+    assert new_period in (old_period, old_period + 1)
+
+    # Verify new snapshot sync committees
+    if new_period == old_period:
+        assert new_snapshot.current_sync_committee == old_snapshot.current_sync_committee
+        assert new_snapshot.next_sync_committee == old_snapshot.next_sync_committee
+    else new_period == old_period + 1:
```
ralexstokes, line 113: correct syntax here would be `elif`, but we can just use `else` given the prior assert on line 107.

```suggestion
    else:
```

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,153 @@
...
+            branch=update.next_sync_committee_branch,
+            depth=log2(NEXT_SYNC_COMMITTEE_INDEX),
+            index=NEXT_SYNC_COMMITTEE_INDEX % 2**log2(NEXT_SYNC_COMMITTEE_INDEX),
+            root=hash_tree_root(new_snapshot.header),
+        )
+
+    # Verify sync committee bitfield length 
+    sync_committee = new_snapshot.current_sync_committee
+    assert len(update.sync_committee_bits) == len(sync_committee)
+    assert sum(update.sync_committee_bits) > MIN_SYNC_COMMITTEE_PARTICIPANTS
```
ralexstokes, line 126: any reason to exclude the single-participant case?

```suggestion
    assert sum(update.sync_committee_bits) >= MIN_SYNC_COMMITTEE_PARTICIPANTS
```

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,153 @@
...
+
+    # Verify sync committee aggregate signature
+    participant_pubkeys = [pubkey for (bit, pubkey) in zip(update.sync_committee_bits, sync_committee.pubkeys) if bit]
+    domain = compute_domain(DOMAIN_SYNC_COMMITTEE, update.fork_version)
```
ralexstokes, line 130: rather than bother with a proof for the fork version, we just accept the update's data and the update will fail if this `fork_version` is corrupted somehow -- this seems like a nice simplification :)

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,153 @@
...
+| `NEXT_SYNC_COMMITTEE_INDEX` | `IndexConcat(Index(BeaconBlock, 'state_root'), Index(BeaconState, 'next_sync_committee'))` |
+
+## Configuration
+
+### Misc
+
+| Name | Value |
+| - | - |
+| `MIN_SYNC_COMMITTEE_PARTICIPANTS` | `1` |
+| `MAX_VALID_LIGHT_CLIENT_UPDATES` | `uint64(2**64 - 1)` |
```
ralexstokes, line 47: we could be more precise with this bound as we should never have more than `LIGHT_CLIENT_UPDATE_TIMEOUT` updates queued under the scheme given below

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,153 @@
...
+    return True
+```
+
+#### `process_update`
+
+```python
+def process_light_client_update(store: LightClientStore, update: LightClientUpdate, current_slot: Slot) -> None:
+    # Validate update
+    assert is_valid_light_client_update(store, update)
+    valid_updates.append(update)
```
ralexstokes, line 143:
```suggestion
    store.valid_updates.append(update)
```

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,153 @@
...
+
+    if sum(update.sync_committee_bits) * 3 > len(update.sync_committee_bits) * 2:
+        # Immediate update when quorum is reached
+        store.snapshot = update.snapshot
+        valid_updates = []
```
ralexstokes, line 148:
```suggestion
        store.valid_updates = []
```

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,153 @@
...
+    elif current_slot > old_snapshot.header.slot + LIGHT_CLIENT_UPDATE_TIMEOUT:
+        # Forced best update when the update timeout has elapsed
+        store.snapshot = max(valid_updates, key=lambda update: sum(update.sync_committee_bits)).new_snapshot
+        valid_updates = []
```
ralexstokes, line 152:
```suggestion
        store.valid_updates = []
```

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,153 @@
...
+    assert is_valid_light_client_update(store, update)
+    valid_updates.append(update)
+
+    if sum(update.sync_committee_bits) * 3 > len(update.sync_committee_bits) * 2:
+        # Immediate update when quorum is reached
+        store.snapshot = update.snapshot
+        valid_updates = []
+    elif current_slot > old_snapshot.header.slot + LIGHT_CLIENT_UPDATE_TIMEOUT:
+        # Forced best update when the update timeout has elapsed
+        store.snapshot = max(valid_updates, key=lambda update: sum(update.sync_committee_bits)).new_snapshot
```
ralexstokes, line 151: note: could make temporary `valid_updates` variable as long as the list is cleared properly... i'll let you decide which is better stylistically

```suggestion
        store.snapshot = max(store.valid_updates, key=lambda update: sum(update.sync_committee_bits)).snapshot
```

ralexstokes (commented):
nice update! it feels more in tone/style with the rest of the spec now :)

i went through and left some comments, most of them on some minor syntactic things.

specs/lightclient/sync-protocol.md
ralexstokes: this should actually have been `new_next_sync_committee` and this was critical for security (!) but has been addressed in the update below

ralexstokes:
@alonmuroch 
> Is there a way to introduce some kind of fork-choice for a light client in case a corrupt committee signs an invalid block?
I'm thinking about an EVM based light client and how it might recover.
Or is the security assumption says a big enough committee can't be corrupted to 2/3 of its members?

the security model reflected in this spec right now is that the sync committees are sampled in a way that an honest majority exists... if that assumption fails then the light client can be made to accept non-canonical updates.

it sounds like there are a few techniques for improving this honesty bound but they all complicate the light client design. which is especially relevant in resource-constrained environments like the EVM.

alonmuroch:
> @alonmuroch
> 
> > Is there a way to introduce some kind of fork-choice for a light client in case a corrupt committee signs an invalid block?
> > I'm thinking about an EVM based light client and how it might recover.
> > Or is the security assumption says a big enough committee can't be corrupted to 2/3 of its members?
> 
> the security model reflected in this spec right now is that the sync committees are sampled in a way that an honest majority exists... if that assumption fails then the light client can be made to accept non-canonical updates.
> 
> it sounds like there are a few techniques for improving this honesty bound but they all complicate the light client design. which is especially relevant in resource-constrained environments like the EVM.

How could the light client recover from that?

specs/lightclient/sync-protocol.md
JustinDrake: Thank you :)

specs/lightclient/sync-protocol.md
JustinDrake: Thank you :)

specs/lightclient/sync-protocol.md
JustinDrake: Thanks!

specs/lightclient/sync-protocol.md
JustinDrake: Thanks!

specs/lightclient/sync-protocol.md
JustinDrake: Thanks!

specs/lightclient/sync-protocol.md
JustinDrake: Thanks!

specs/lightclient/sync-protocol.md
JustinDrake: Exactly :)

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,153 @@
...
+    next_sync_committee: SyncCommittee
+```
+
+#### `LightClientUpdate`
+
+```python
+class LightClientUpdate(Container):
+    # Updated snapshot
+    snapshot: LightClientSnapshot
+    # Merkle branches for the next sync committee
```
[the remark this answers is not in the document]
JustinDrake: Thanks

specs/lightclient/sync-protocol.md
JustinDrake: Why? `LIGHT_CLIENT_UPDATE_TIMEOUT` is denominated in slots whereas `MAX_VALID_LIGHT_CLIENT_UPDATES` is denominated in updates. There could be more than one update per slot! Remember these updates are prepared offchain. The onchain data is solely for incentivisation purposes.

ralexstokes: right -- i was thinking you would have at most one update per slot but this is not necessarily true, e.g. even getting the same update with simply more signatures attached

ralexstokes:
> > @alonmuroch
> > > Is there a way to introduce some kind of fork-choice for a light client in case a corrupt committee signs an invalid block?
> > > I'm thinking about an EVM based light client and how it might recover.
> > > Or is the security assumption says a big enough committee can't be corrupted to 2/3 of its members?
> > 
> > 
> > the security model reflected in this spec right now is that the sync committees are sampled in a way that an honest majority exists... if that assumption fails then the light client can be made to accept non-canonical updates.
> > it sounds like there are a few techniques for improving this honesty bound but they all complicate the light client design. which is especially relevant in resource-constrained environments like the EVM.
> 
> How could the light client recover from that?

depends on the nature of the failure... for example, light clients could do their own data availability sampling to protect against availability faults. we could come up with a fraud proof scheme for blocks suffering validity faults. or add slashing on the beacon chain for invalid light client proofs (which again have to get back to light clients...).

the simplest thing for the moment is to simply accept the sync committees are honest and in the event it became corrupted, there's the avenue for "social recovery" where 'operators' of light clients just pull down new code that marks the block(s) as invalid.

alonmuroch:
> > > @alonmuroch
> > > > Is there a way to introduce some kind of fork-choice for a light client in case a corrupt committee signs an invalid block?
> > > > I'm thinking about an EVM based light client and how it might recover.
> > > > Or is the security assumption says a big enough committee can't be corrupted to 2/3 of its members?
> > > 
> > > 
> > > the security model reflected in this spec right now is that the sync committees are sampled in a way that an honest majority exists... if that assumption fails then the light client can be made to accept non-canonical updates.
> > > it sounds like there are a few techniques for improving this honesty bound but they all complicate the light client design. which is especially relevant in resource-constrained environments like the EVM.
> > 
> > 
> > How could the light client recover from that?
> 
> depends on the nature of the failure... for example, light clients could do their own data availability sampling to protect against availability faults. we could come up with a fraud proof scheme for blocks suffering validity faults. or add slashing on the beacon chain for invalid light client proofs (which again have to get back to light clients...).
> 
> the simplest thing for the moment is to simply accept the sync committees are honest and in the event it became corrupted, there's the avenue for "social recovery" where 'operators' of light clients just pull down new code that marks the block(s) as invalid.

If the light client is intended to be implemented as a smart contract that might create a ton of issues...

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,181 @@
...
+Eth2 is designed to be light client friendly for constrained environments to access Eth2 with reasonable satefy and liveness. Such environments include resource-constrained devices (e.g. phones for trust-minimised wallets) and metered VMs (e.g. blockchain VMs for cross-chain bridges).
+
+This document suggests a minimal light client design for the beacon chain that uses sync committees introduced in [this beacon chain extension](./beacon-chain.md).
+
+## Constants
+
+| Name | Value |
+| - | - |
+| `FINALIZED_ROOT_INDEX` | `Index(BeaconState, 'finalized_checkpoint', 'root')` |
+| `NEXT_SYNC_COMMITTEE_INDEX` | `Index(BeaconState, 'next_sync_committee')` |
```
hwwhww, line 40: Do these `Index` helpers equal `get_generalized_index` in https://github.com/ethereum/eth2.0-specs/blob/5f9112ad4227d12ea03c001517d53518e6e355f0/ssz/merkle-proofs.md?

If so, we may need to either make `merkle-proofs.md` executable again or implement it in `remerkleable`.

specs/lightclient/sync-protocol.md
```diff
@@ -0,0 +1,181 @@
...
+
+    # Verify update does not skip a sync committee period
+    snapshot_period = compute_epoch_at_slot(snapshot.header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+    update_period = compute_epoch_at_slot(update.header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+    assert update_period in (snapshot_period, snapshot_period + 1)
+
+    # Verify update header root is the finalized root of the finality header, if specified
+    if update.finality_header == BeaconBlockHeader():
+        signed_header = update.header
+        assert update.finality_branch == [ZERO_HASH for _ in range(log2(FINALIZED_ROOT_INDEX))]
```
hwwhww, line 116: It seems that we need to add a `uint64` casting to all these `log2`, or reuse [`get_generalized_index_length`](https://github.com/ethereum/eth2.0-specs/blob/5f9112ad4227d12ea03c001517d53518e6e355f0/ssz/merkle-proofs.md#get_generalized_index_length).

specs/lightclient/sync-protocol.md
JustinDrake: > Do these Index helpers equal get_generalized_index

Yes :)

> we may need to either make merkle-proofs.md executable again or implement it in remerkleable

When we do so I suggest shortening "generalized_index" to something like "node_index", "tree_index", "Merkle_index", or just "index" for short.

dankrad:
My main concern here is what happens when a committee does not achieve a 2/3 vote during `EPOCHS_PER_SYNC_COMMITTEE_PERIOD`. Realistically, blockchain/smart contract light clients will only use 2/3 links because everything else is too dangerous. But when a sync committees tenure ends, there is no incentive to keep trying to get one of the votes over the threshold.

Intuitively I would have said sync committees should only be switched when the next sync committee has been confirmed by a 2/3 vote. But that of course can lead to terrible situations where a sync committee intentionally withholds confirming the next one.

One idea would be that there is a mechanism to record the "success" of each sync committee to achieve 2/3, and if it never reached it, a final vote on the last block can be included at any time for some reward.

dankrad:
Further remark in a similar direction: Intuitively, the previous sync committee has almost the same "validity" in epoch `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` and `EPOCHS_PER_SYNC_COMMITTEE_PERIOD+1` (after committees have switched). What about giving the old committee the opportunity for one, and only one, more vote? The nice thing is that they could vote for the "next next" sync committee, effectively allowing to skip one committee at a time. Also, in case the new committee is poor and doesn't come to a consensus, it gives one opportunity for the old committee to create a 2/3+1 link.

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,186 @@
...
+
+#### Sync committee processing
+
+```python
+def process_sync_committee(state: BeaconState, body: BeaconBlockBody) -> None:
+    # Verify sync committee aggregate signature signing over the previous slot block root
+    previous_slot = max(state.slot, Slot(1)) - Slot(1)
+    committee_indices = get_sync_committee_indices(state, get_current_epoch(state))
+    participant_indices = [committee_indices[i] for i in range(len(committee_indices)) if body.sync_committee_bits[i]]
+    participant_pubkeys = [state.validators[participant_index].pubkey for participant_index in participant_indices]
```
djrtwo, line 157: Might consider getting the pubkeys from the `SyncCommittee` object. Will make execution of this function with partials more straight forward for light clients.

```suggestion
    committee_pubkeys = state.current_sync_committee.pubkeys
    participant_pubkeys = [pubkey for pubkey, bit in zip(committee_pubkeys, body.sync_committee_bits) if bit]
```

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,186 @@
...
+    phase0.process_block(state, block)
+    process_sync_committee(state, block.body)
+```
+
+#### Sync committee processing
+
+```python
+def process_sync_committee(state: BeaconState, body: BeaconBlockBody) -> None:
+    # Verify sync committee aggregate signature signing over the previous slot block root
+    previous_slot = max(state.slot, Slot(1)) - Slot(1)
```
djrtwo, line 154: Maybe a `get_previous_slot` accessor like `get_previous_epoch`

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,186 @@
...
+    domain = get_domain(state, DOMAIN_SYNC_COMMITTEE, compute_epoch_at_slot(previous_slot))
+    signing_root = compute_signing_root(get_block_root_at_slot(state, previous_slot), domain)
+    assert bls.FastAggregateVerify(participant_pubkeys, signing_root, body.sync_committee_signature)
+
+    # Reward sync committee participants
+    participant_rewards = Gwei(0)
+    active_validator_count = uint64(len(get_active_validator_indices(state, get_current_epoch(state))))
+    for participant_index in participant_indices:
+        base_reward = get_base_reward(state, participant_index)
+        reward = Gwei(base_reward * active_validator_count // len(committee_indices) // SLOTS_PER_EPOCH)
```
djrtwo, line 167: Why is this reward so high?

So that on average you get a base_reward per epoch from sync committee duties?
@vbuterin

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,186 @@
...
+        increase_balance(state, participant_index, reward)
+        participant_rewards += reward
+
+    # Reward beacon proposer
+    increase_balance(state, get_beacon_proposer_index(state), Gwei(participant_rewards // PROPOSER_REWARD_QUOTIENT))
```
djrtwo, line 172: If we are going to do this like we do other proposer rewards, we need to temper the participant rewards by the `1 - PROPOSER_REWARD_QUOTIENT`
So that in aggregate a max single base_reward is given per participant.

something like the following:
```python
proposer_reward = Gwei(0)
for participant_index in participant_indices:
    base_reward = get_base_reward(state, participant_index)
    max_participant_reward = base_reward - base_reward // PROPOSER_REWARD_QUOTIENT
    reward = Gwei(max_participant_reward * active_validator_count // len(committee_indices) // SLOTS_PER_EPOCH)
    increase_balance(state, participant_index, reward)
    proposer_reward += base_reward // PROPOSER_REWARD_QUOTIENT
increase_balance(state, get_beacon_proposer_index(state), Gwei(proposer_reward))
```

djrtwo (commented):
did a quick pass

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,186 @@
...
+def get_sync_committee_indices(state: BeaconState, epoch: Epoch) -> Sequence[ValidatorIndex]:
+    """
+    Return the sync committee indices for a given state and epoch.
+    """
+    base_epoch = Epoch((max(epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD, 1) - 1) * EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
+    active_validator_indices = get_active_validator_indices(state, base_epoch)
+    active_validator_count = uint64(len(active_validator_indices))
+    seed = get_seed(state, base_epoch, DOMAIN_SYNC_COMMITTEE)
+    i, sync_committee_indices = 0, []
+    while len(sync_committee_indices) < SYNC_COMMITTEE_SIZE:
```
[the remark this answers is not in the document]
JustinDrake: > Is that intentional?

Yes, it's intentional! The reason is to more easily handle the edge case where there are less than `SYNC_COMMITTEE_SIZE` active validators.

specs/lightclient/beacon-chain.md
hwwhww, line 172: Should this new sync committee reward component be deducted from `get_inactivity_penalty_deltas`((https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#components-of-attestation-deltas) `penalties` component for the well-performed validators? i.e., modify line `penalties[index] += Gwei(BASE_REWARDS_PER_EPOCH * base_reward - get_proposer_reward(state, index))`.

hwwhww: ^^^^ The above question was answered during the call, do not need to deduct the sync committee rewards.

Another question about light client reward + inactivity leak accounting:

- Since other reward components are all accounted during **epoch-processing**, they were calculated with (1) if `is_in_inactivity_leak(state)` is True, then pre-issue the `base_reward` and (2) cancel the pre-issued reward in `get_inactivity_penalty_deltas`.
- However, the light client committee rewards are calculated during the **block-processing**. So it cannot be detected with `is_in_inactivity_leak(state)`.
- It seems #2140 form can fix this issue well by using flags. Does `lightclient` patch require `accounting-reform` patch? Or there is a better way to delay the calculation for this case?

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,186 @@
...
+        shuffled_index = compute_shuffled_index(uint64(i % active_validator_count), active_validator_count, seed)
+        candidate_index = active_validator_indices[shuffled_index]
+        random_byte = hash(seed + uint_to_bytes(uint64(i // 32)))[i % 32]
+        effective_balance = state.validators[candidate_index].effective_balance
+        if effective_balance * MAX_RANDOM_BYTE >= MAX_EFFECTIVE_BALANCE * random_byte:
```
terencechain, line 118: `MAX_RANDOM_BYTE` is not defined globally

It was previous defined within `compute_proposer_index`

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,186 @@
...
+    committee_indices = get_sync_committee_indices(state, get_current_epoch(state))
+    participant_indices = [committee_indices[i] for i in range(len(committee_indices)) if body.sync_committee_bits[i]]
+    participant_pubkeys = [state.validators[participant_index].pubkey for participant_index in participant_indices]
+    domain = get_domain(state, DOMAIN_SYNC_COMMITTEE, compute_epoch_at_slot(previous_slot))
+    signing_root = compute_signing_root(get_block_root_at_slot(state, previous_slot), domain)
+    assert bls.FastAggregateVerify(participant_pubkeys, signing_root, body.sync_committee_signature)
+
+    # Reward sync committee participants
+    participant_rewards = Gwei(0)
+    active_validator_count = uint64(len(get_active_validator_indices(state, get_current_epoch(state))))
```
terencechain, line 164: Should we use `base_epoch` instead of `current_epoch`? 

It aligns with `get_sync_committee_indices`

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,186 @@
...
+    for participant_index in participant_indices:
```
terencechain, line 165: Should we exclude slashed validators for rewards?

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,201 @@
...
+        for i in range(0, len(pubkeys), SYNC_COMMITTEE_PUBKEY_AGGREGATES_SIZE)
+    ]
+    return SyncCommittee(pubkeys, aggregates)
+```
+
+### Block processing
+
+```python
+def process_block(state: BeaconState, block: BeaconBlock) -> None:
+    phase0.process_block(state, block)
```
hwwhww, line 159: Note that by `phase0.process_block`, it uses the phase 0 package.

1. The constants and the function calls `process_block` would be *phase 0 version*. It may cause some side effects that we may easily overlook.
2. You are passing `(lightclient_patch.state, lightclient_patch.block)` to `process_block(state: phase0.BeaconState, block: phase0.BeaconBlock)`.

Saving 3 lines probably is not worth it. 😅

specs/lightclient/beacon-chain.md
hwwhww: Update: applied @djrtwo 's suggestion in #2141 (https://github.com/ethereum/eth2.0-specs/pull/2141/commits/1b7b842aaff3c133266f67b2c0dd6811e580bf88)

djrtwo:
> Saving 3 lines probably is not worth it. 😅

Yes! I would tend to agree

specs/lightclient/beacon-chain.md
```diff
@@ -0,0 +1,248 @@
...
+    return rewards, penalties
+```
+
+#### Final updates
+
+*Note*: The function `process_final_updates` is modified to handle sync committee updates.
+
+```python
+def process_final_updates(state: BeaconState) -> None:
+    phase0.process_final_updates(state)
```
hwwhww, line 243: Same as https://github.com/ethereum/eth2.0-specs/pull/2130#discussion_r537745449
Some notes:
- The current logic looks "fine" since the light client patch is irrelevant to the existing `phase0.process_final_updates`. 
- Still, it's dangerous to call the helper function in the `phase0` package directly.
- Note that if #2150 is included in the HF1, #2150 will unfold the full `process_final_updates` anyway there.

Maybe add a note here for now?
```suggestion
    # FIXME: unfold the full `process_final_updates` to avoid side effects.
    phase0.process_final_updates(state)
```

hwwhww (approved):
Giving the initial version a green light. The tests & iterations will be added with other PRs.

djrtwo:
Added some sanity tests. merging!

specs/lightclient/beacon-chain.md
AuraSophia: ,my name is Aura ,I saved your group page long time ago ,after years hope all is good with you .
Always believed in it so I had 0.5 Ethereum that I keep it for a long time for my daughter and lost it to a scammer. 
Now after I m putting the baby to sleep I m reading and watching videos. 
Is any chance to recover it?.
I have 200$ I know is not much and I know is late to invest ?
I know but I dont have much.

Thank youuu

AuraSophia (commented):
Please let me know what should I do ,after this year no job ,away from husband I wanna make things work .really belive in this .
My Ethereum adress is 
0x40e70857e75F79a4eaE109eD632Ac68FfC6f10Fe

Jojad9 (commented):
Vbuterin patch 2

Issue #2116, opened 2020-10-29
BlocksByRange under WS sync


djrtwo:
As brought up by @mbaxter, it is unclear at this moment how to handle `BlocksByRange` requests when a node has synced from a weak subjectivity state.

A few questions:
1. What minimum block/epoch ranges (wrt current epoch) are expected to baseline be served by all nodes?
2. How does a node know the range a peer will serve?
3. If a `BlocksByRange` request is made to a peer and all or some subset of slots in that range are not in custody of the peer, what should be the response?

### Question 1

_What minimum block ranges (wrt current slot) are expected to baseline be served by all nodes?_

The spec current says:

>  Clients MUST keep a record of signed blocks seen since the start of the weak subjectivity period and MUST support serving requests of blocks up to their own `head_block_root`

"weak subjectivity period" is poorly defined in this context. I would suggest that that upper limit of the weak subjectivity period (~5 months at 100% safety decay) is used as a constant in this spec to define the minimum epoch range that a node is expected to backfill and keep around for serving. (The following is adopted from the weak-subjectivity guide):

```python
MIN_EPOCHS_FOR_BLOCK_REQUESTS = (
    MIN_VALIDATOR_WITHDRAWABILITY_DELAY
    + MAX_SAFETY_DECAY * CHURN_LIMIT_QUOTIENT // (2 * 100)
)
```

Where `MAX_SAFETY_DECAY = 100` and thus `MIN_EPOCHS_FOR_BLOCK_REQUESTS = 33024` (~5 months).

This could be considered overkill for small validator sets or more aggressive safety decays, but using the maximum here does not put an overly large block storage requirement on honest nodes and provides a higher guarantee in baseline quality of service _without_ trying to integrate dynamic WS period lengths into the assumptions in the networking protocol.

### Question 2

_How does a node know the range a peer will serve?_

By default, we can assume that all nodes will serve `MIN_EPOCHS_FOR_BLOCK_REQUESTS` worth of blocks from the current epoch. If not, that is grounds for de-scoring. But, many nodes might choose to serve beyond this range (maximally back to genesis).

There are two obvious places to publish this info, in `Status` or `MetaData`. `Status` is generally for info on head state of a node to decide if the node is on chains you care about for syncing, while `MetaData` is for more slowly changing information about the node in general.

I can see the argument for using each
* Put lower bound on block serve epoch in `Status` because `Status` is already used frequently when crafting block sync requests 
[...]

PR #2131, opened 2020-11-12
BlocksByRange under WS

djrtwo:
Partially addresses #2116 

* Defines `MIN_EPOCHS_FOR_BLOCK_REQUESTS` for minimum expectation of epoch range for serving blocks (and thus how far a new node must backfill)
* *Does not define* a way to advertise this in Req/Resp (e.g. in `Status`) or in ENR
    * I personally think it unwise to modify `Status` or `MetaData` (likely with a protocol ID update) this near to genesis launch. `MIN_EPOCHS_FOR_BLOCK_REQUESTS` gives us 4+ months to upgrade the req/resp to make this info available
 
The main thing the above solution won't capture during the first 4 months is the event that a node is back-filling blocks from a checkpoint state to genesis. In such a case, the peer might not be able to respond to BlocksByRange requests successfully. To handle this, I suggest we spec an error code for this case. This will carry forward in the future when BlocksByRange requests are made outside of the advertised range. We can either use `2 (ServerError)` or define `3 (ResourceUnavailable)`

Questions:
1. Are we okay _leaving out_ advertisement for earliest epoch for block requests until a couple of months after mainnet?
2. Good with adding the error? If so, use `2` or define `3`?


djrtwo:
cc: @mbaxter @arnetheduck  @ajsutton @AgeManning

mbaxter:
1. I think it's fine to defer advertising the earliest available epoch - it does seem like the wrong time to be pushing this change in :). 
2. The error makes sense to me, I vote for `3 (ResourceUnavailable)`.

AgeManning:
1. I'm fine with this
2. Agree with `3 (ResourceUnavailable)`

arnetheduck:
1. ok
2. The wording of the text suggest that clients must offer up blocks - adding the possibility to claim `ResourceUnavailable` allows clients to keep claiming that indefinitely - the correct behaviour would be to simply sync from `MIN_EPOCHS_FOR_BLOCK_REQUESTS` an onwards instead of "backfilling" - this allows the client to report a "correct" head in the existing `Status` request without the loophole at the cost of slightly slowing down startup. We can deliberate the addition of an error code when extending `Status`/`Metadata` in the 4-5 months mentioned - until then, the language in this PR doesn't really allow for a conforming and honest client to return it ("must backfill" should however be replaced by "must sync from"). One issue would be that this slows down startup slightly because the client must now sync more blocks - but on the other hand, this benefits the network because we enshrine a supply of synced nodes more strongly - faster syncing lies in the domain of light clients, and can/should be solved then, along with archive solutions.

nisdas:
1. This sounds good, gives us a bit of time to sort it out. 
2. 3 (`ResourceUnavailable`) is an appropriate error code for this.

ajsutton:
>  the correct behaviour would be to simply sync from MIN_EPOCHS_FOR_BLOCK_REQUESTS an onwards instead of "backfilling"

This isn't possible.  The sync has to start from the state provided by the user and go forward from there.  Clients that respond with `ResourceUnavailable` would very likely be immediately disconnected as useless clients so there's no advantage to claiming that indefinitely.

AgeManning:
Yeah. For the record. Our current strategy is going to be to ban any node that sends us this error (at least for the first 5 months). 

The ban lasts ~40 mins which should be enough time for the peer to backfil etc.

djrtwo:
> This isn't possible. The sync has to start from the state provided by the user and go forward from there.

Yes, this is correct. We can't enforce that users must show up with a state that is exactly `MIN_EPOCHS_FOR_BLOCK_REQUESTS` in the past. And if we did, we would be forcing them to choose a WS state that is at the bounds of being safe (and for small validator sets, just simply unsafe).

Note, that for a backfill of blocks only to that boundary, the only validity checks (without getting an old state) that a node can do is check that these blocks for a hash chain. Because the WS state is trusted, this check is enough to show validity of the blocks for storage and future serving.

Note that if you backfill all the way to genesis, you could then check other validity conditions and could produce historic states.

arnetheduck:
> This isn't possible. 

> We can't enforce that users must show up with a state that is exactly

Why are the two related? When we ask for `BlocksByRange`, we can simply start at slot `wallslot - MIN_EPOCHS_FOR_BLOCK_REQUESTS` and should along the way hit the WS state - ie it's still a "backfill" in that sense, except that it's done in order and therefore it is more likely that the network keeps the relevant blocks around, per the 'MUST' requirement in this PR.

> Because the WS state is trusted, this check is enough to show validity of the blocks for storage and future serving.

It should also do a signature check, else it can be poisoned with blocks that have the correct root but invalid signature - it should be possible since we have the proposer index from the block and the validator set of the WS state.

The point here is a bit that if we want to ensure that blocks are around, it's more likely that this will happen if the easiest thing to do is to get&store the blocks - starting range sync from WS state then "maybe" backfilling makes it less likely that the blocks will stay around - to make the sync requirement even stronger, it would make some sense to bake in "probe" range requests even if synced and disconnect any client that does not serve the blocks, or already-synced clients have little to list (being disconnected by an unsynced client is a small loss - being disconnected by a synced client means a clear and present risk that your attestations will not make it through)

djrtwo:
> Why are the two related? 

Because you have no way to validate a block from `wallslot - MIN_EPOCHS_FOR_BLOCK_REQUESTS`. You have no start block to check that it forms a chain, and you have no state to check state transition validity. 

You *can* check that the blocks being served to you form a chain and you *can* even validate the proposer signature against the validator set you have from your WS state, but you cannot validate that the sequence of blocks being sent to you is valid wrt state transition nor can you validate that the blocks will ultimately reach the chain you decided was valid through you input of WS state.

Thus an attacker can serve you many blocks before you can know if they are valuable/valid, and thus can cause you to consume extra bandwidth.

> It should also do a signature check, else it can be poisoned with blocks that have the correct root but invalid signature - it should be possible since we have the proposer index from the block and the validator set of the WS state.

Yes, agreed. Because you are storing the `SignedBeaconBlock` you'll want to validate that outer container.

arnetheduck (approved):

specs/phase0/p2p-interface.md
```diff
@@ -746,10 +748,25 @@ The request MUST be encoded as an SSZ-container.
...
-and MUST support serving requests of blocks up to their own `head_block_root`.
+Clients MUST keep a record of signed blocks seen on the epoch range
+`[max(GENESIS_EPOCH, current_epoch - MIN_EPOCHS_FOR_BLOCK_REQUESTS), current_epoch]`
+where `current_epoch` is defined by the current wall-clock time,
+and clients MUST support serving requests of blocks on this range.
+
+Synced clients unable to reply to Block requests within the
+`MIN_EPOCHS_FOR_BLOCK_REQUESTS` epoch range MAY get descored or disconnected at any time.
+Note, due to this it is risky behaviour to begin participating as a full node at the head if having
+not yet backfilled on this range.
```
mbaxter, line 759: Currently, teku's checkpoint sync will load a provided finalized checkpoint state, forward sync to catch up to head and get the node online quickly, then backfill blocks requesting them in reverse order to ensure that we form a hash chain and download the correct history.

To accommodate the suggestions here, we'd need to instead calculate the oldest required slot, forward sync from there, and hope that we sync the correct chain that matches up with the provided checkpoint state.  If we end up downloading a malicious chain, we'd have to throw away the downloaded blocks and try again.  

This seems suboptimal.  In the first strategy, we're guaranteed to sync the correct blocks and get online quickly which is very helpful to end users.  In the second strategy, we're not guaranteed to find the correct chain, and we're forcing users to wait much longer than necessary to get their nodes online.

I see the point that we don't want to lose old blocks.  Long term, if we stick with the idea of updating `Status` to advertise the oldest available block, it will be simple enough to enforce this:  we just have to spot check blocks in the advertised range and check that the advertised range is being backfilled over time (the oldest available block advertised in `Status` is being updated at a reasonable rate).  

If we're really concerned about the possibility of clients not backfilling in the short-term, we could do a binary search when nodes connect to find their oldest available block.  And check every so often that older blocks have become available (similar to the long-term strategy, but slightly less efficient). 

The ability to reliably get up and running quickly is extremely useful for users and it's one of the nicer features of eth2 in my opinion.  I don't think we should be discouraging this.

djrtwo: This does not suggest doing a forward sync. It just suggests that if your peers notice you haven't backfilled, they can disconnect.

And it implies that your peers might check and MAY disconnect (e.g. could use random sample or binary search). [I even had "nodes MAY submit random queries on that range and MAY disconnect upon failure to respond" but removed it]

Pretty much what it says is that once you start participating a a full node at the head, you are subject to having blocks on that range. Again, *it does not* say you must forward sync.

Something I wrote must be unclear. Can you re-read now thatt you know my intention and suggest how I might clarify?

mbaxter: I tried to highlight these lines but I guess it didn't come across:
```
Note, due to this it is risky behaviour to begin participating as a full node at the head if having
not yet backfilled on this range.
```

To me that sounds like we're discouraging nodes from advertising a recent head block (via `Status`) without first having backfilled the historical blocks.  And it seems to me that way to avoid this is to "backfill" in order starting from the oldest block in the required range up to the checkpoint.  This way, we always have the blocks from the beginning of the range up the head block that we're advertising (the latest backfilled block).

Maybe I'm misunderstanding what it means to "begin participating as a full node at the head"?  Otherwise, it might be worth talking more about what strategies nodes may take while they're in the process of backfilling.

mbaxter: I also thought that having an explicit `UNAVAILABLE` error which could be used while backfilling made it clearer how nodes should behave while in the process of backfilling.

djrtwo: > And it seems to me that way to avoid this is to "backfill" in order starting from the oldest block in the required range up to the checkpoint. 

That is certainly one way, but the spec does not say which way you should back-fill, it just says that you SHOULD. And if you don't and people ask you for blocks that you don't have, you might be subject to negative consequences

You're right in that the `Status` is used means that you must send it even before backfilling (because you connected to peers) and thus it isn't clear what should be included in it during backfill. Even if you back-fill from back to front, you still know of a more recent finalized head/root/slot

I'm trying to decide the simplest way to say, "I'm not ready to serve" without modifying the current `Status` data structure. Any ideas?

arnetheduck:
> hope that we sync the correct chain that matches up with the provided checkpoint state

this is actually a valuable sanity check that the provided checkpoint state is sane - if it _doesn't_ match up (in terms of block hashes), something is wrong: the checkpoint state, the nodes around you etc and the beacon node cannot participate. It's not so much a hope, as a hard requirement.

> much longer

Syncing pre-checkpoint blocks is much faster than syncing "normal" blocks - there's no need to do state transitions and finality processing: it needs signature and tree root verification only.

> If we end up downloading a malicious chain, we'd have to throw away the downloaded blocks and try again.

This should be noticeable fairly quickly - ie if you're downloading from several peers, a chain must be formed by the received blocks - they also must have valid signatures from the validator set in the checkpoint state - `SLOTS_PER_HISTORICAL_ROOT` block roots are also available in the checkpoint state for early hash verification.

> `UNAVAILABLE`

The spec can't at the same time _require_ that clients make blocks available and give validity to not doing so through an error code - it must choose: either it's ok to drop blocks / not backfill _or_ clients that don't provide blocks can be sanctioned - the main sanction we have available is to disconnect or descore such clients - they are an undue burden upon the network as a whole, which is what this PR points out.

> I'm trying to decide the simplest way to say, "I'm not ready to serve" without modifying the current Status data structure. Any ideas?

set head block to `block_at_slot(now - MIN_EPOCHS_FOR_BLOCK_REQUESTS)` and advance it as backfilling advances - this tells others that you're ready or not for syncing up to the given block/slot - it also means that you should be able to reply to block requests for the range `[now - MIN_EPOCHS_FOR_BLOCK_REQUESTS, Status.head_slot)`. `Status.finalized_epoch`  is a bit more tricky: it can follow `head_slot` up to the subjectivity checkpoint most likely, though this is not *strictly* true - one could also set it to the WS checkpoint from the start, though this is a bit weird (how can head be < than finalized?)

An interesting question of course what to do with clients that are "not ready to serve" yet are gossipping attestations that point to blocks newer than their status - they are either cheating in their `Status` or are not validating attestations as they should - this provides another data point by which to descore leeching clients.

specs/phase0/p2p-interface.md
arnetheduck: > hope that we sync the correct chain that matches up with the provided checkpoint state

this is actually a valuable sanity check that the provided checkpoint state is sane - if it _doesn't_ match up (in terms of block hashes), something is wrong: the checkpoint state, the nodes around you etc and the beacon node cannot participate. It's not so much a hope, as a hard requirement.

> much longer

Syncing pre-checkpoint blocks is much faster than syncing "normal" blocks - there's no need to do state transitions and finality processing: it needs signature and tree root verification only.

> If we end up downloading a malicious chain, we'd have to throw away the downloaded blocks and try again.

This should be noticeable fairly quickly - ie if you're downloading from several peers, a chain must be formed by the received blocks - they also must have valid signatures from the validator set in the checkpoint state - `SLOTS_PER_HISTORICAL_ROOT` block roots are also available in the checkpoint state for early hash verification.

> `UNAVAILABLE`

The spec can't at the same time _require_ that clients make blocks available and give validity to not doing so through an error code - it must choose: either it's ok to drop blocks / not backfill _or_ clients that don't provide blocks can be sanctioned - the main sanction we have available is to disconnect or descore such clients - they are an undue burden upon the network as a whole, which is what this PR points out.

> I'm trying to decide the simplest way to say, "I'm not ready to serve" without modifying the current Status data structure. Any ideas?

set head block to `block_at_slot(now - MIN_EPOCHS_FOR_BLOCK_REQUESTS)` and advance it as backfilling advances - this tells others that you're ready or not for syncing up to the given block/slot - it also means that you should be able to reply to block requests for the range `[now - MIN_EPOCHS_FOR_BLOCK_REQUESTS, Status.head_slot)`. `Status.finalized_epoch`  is a bit more tricky: it can follow `head_slot` up to the subjectivity checkpoint most likely, though this is not *strictly* true - one could also set it to the WS checkpoint from the start, though this is a bit weird (how can head be < than finalized?)

before `block_at_slot(now - MIN_EPOCHS_FOR_BLOCK_REQUESTS)` is known, it seems reasonable to set it to the genesis block - it also seems reasonable to consider receiving a block signed by the "correct" validator as a good "head" block once received from the first range request.

An interesting question of course what to do with clients that are "not ready to serve" yet are gossipping attestations that point to blocks newer than their status - they are either cheating in their `Status` or are not validating attestations as they should - this provides another data point by which to descore leeching clients.

mbaxter: > I'm trying to decide the simplest way to say, "I'm not ready to serve" without modifying the current `Status` data structure. Any ideas?

We could set the head root / slot to a special value (zero?).  And leave the finalized root / slot pointing to the WS checkpoint.

mbaxter: > The spec can't at the same time _require_ that clients make blocks available and give validity to not doing so through an error code - it must choose: either it's ok to drop blocks / not backfill _or_ clients that don't provide blocks can be sanctioned - the main sanction we have available is to disconnect or descore such clients - they are an undue burden upon the network as a whole, which is what this PR points out.

I don't think an error code is mutually exclusive w/ requiring clients to backfill.  It's just a recognition of the fact that backfilling takes time.  If a node always returns this error code and never backfills, then it can be considered malicious.  There's also nothing preventing nodes from descoring / disconnecting a peer that returns this error.  I just think that we don't want to mandate super strict behaviour here that prevents clients from implementing a reverse backfill strategy.

ajsutton: To add to what Meredith said, I think we're undervaluing the benefit and importance of doing snapshot syncs and being overly concerned with the risk of clients not backfilling blocks. With the weak subjectivity period the most secure way to sync the chain is via snapshot sync - anything else leaves you at least vulnerable to DOS attacks by feeding you the wrong chain and making you process it all the way up to the weak subjectivity checkpoint at which point you throw out all that work and start again.  Snapshot sync is also a huge win for users who are able to get a node up and validating in under a minute regardless of the length of the chain.

Then in terms of the risk of not backfilling, the spec is only requiring backfill to the start of the weak subjectivity period, so syncing from genesis may become impossible anyway (though Teku is backfilling all the way to genesis at the moment even after a snapshot sync since genesis sync is still so common).  If you are syncing, then naturally you would disconnect any peers that can't provide the blocks you need - that may be because they didn't backfill but more likely because they themselves are still syncing or are on a different fork. That's no different to disconnecting peers that aren't on the attestation subnet you need - they aren't malicious, just useless to you.

So I'm really confused as to why we're specifically adding to the spec that it's risky to take full advantage of snapshot sync when its got such clear benefits and I'm struggling to see what the downside is of not being able to identify nodes that haven't completed backfill yet.  Even the threat of "leeching" clients that don't backfill at all seems pretty limited - they're going to make up quite a small part of the network and will naturally wind up building up historic blocks as time progresses and they import them.  Nodes that are already in sync won't notice they didn't backfill and new nodes would be starting from a more recent snapshot anyway so probably also wouldn't notice.

I'd understand more if the requirement was to backfill all blocks to genesis but currently I don't understand what threat we're trying to mitigate here.

arnetheduck: > I'd understand more if the requirement was to backfill all blocks to genesis but currently I don't understand what threat we're trying to mitigate here.

the objective would to ensure that regardless where in the WS period you are, you have blocks in the network to sync from - one could calculate a more accurate period but what this PR does it simply take the maximum reasonable WS period and sync from there - resyncing the blocks also acts as a trivial guard against malicious WS checkpoint services.

the other objective is to make a small adjustment that fits within the existing protocol for HF1 and then make a bigger change later on, when there's some more meaningful changes around to care about.

> risky to take full advantage of snapshot sync

you're still taking advantage of snapshot sync: you quickly download a few extra blocks, check their sigs and hashes, and you're good to go.

> Even the threat of "leeching" clients

Judging by the scarcity of LES servers on eth1, it's not a theoretical problem. Part of your job as a validator is to supply a copy of the blocks for a while - at least, that is the case for now in phase0.

I see this PR mainly as a stepping stone that we can improve upon later by relaxing the rules further where it's safe to do so. As mentioned in other comments, once clients start supplying different ranges of blocks, in order to sanely be able to find viable peers, one needs an improved discovery mechanism at least.

ajsutton: > the objective would to ensure that regardless where in the WS period you are, you have blocks in the network to sync from - one could calculate a more accurate period but what this PR does it simply take the maximum reasonable WS period and sync from there - resyncing the blocks also acts as a trivial guard against malicious WS checkpoint services.

Given that the majority of nodes are not freshly sync'd this seems like something that will be pretty straight forward even without actively backfilling blocks.  And I'm suggesting we *do* actively backfill blocks just that we shouldn't have to arbitrarily wait for that process to complete before starting to attest or participating fully in gossip etc at the chain head.

Resyncing the blocks back to genesis provides a guard against malicious checkpoint services, but syncing back to the start of the WS period doesn't provide any extra verification because it doesn't attach to any known state.

> the other objective is to make a small adjustment that fits within the existing protocol for HF1 and then make a bigger change later on, when there's some more meaningful changes around to care about.

The simplest change we could make is to just define an error response code that indicates the node doesn't have that block.  This change doesn't actually solve the problem because any node syncing from genesis still has no way to know if a remote peer has blocks back to genesis or only back to the weak subjectivity period, so still can't differentiate between a claim that "there were no blocks" or "I don't have those blocks".

> > risky to take full advantage of snapshot sync
> 
> you're still taking advantage of snapshot sync: you quickly download a few extra blocks, check their sigs and hashes, and you're good to go.

But we've added a whole bunch of complexity about how to behave when you're in sync but not backfilled and having to track where the backfill process is up to etc.  But we don't get any actual benefit from that complexity.

> > Even the threat of "leeching" clients
> 
> Judging by the scarcity of LES servers on eth1, it's not a theoretical problem. Part of your job as a validator is to supply a copy of the blocks for a while - at least, that is the case for now in phase0.

The equivalent in Eth1 is not the LES protocol, but the direct equivalent of being able to request blocks from peers. Not all nodes on Eth1 have backfilled all blocks and there's no requirement to have all historic blocks before participating fully in the network.  If a node needs specific blocks and one of it's peers doesn't have them, it is allowed to disconnect because the peer is useless to them.

> I see this PR mainly as a stepping stone that we can improve upon later by relaxing the rules further where it's safe to do so. 

What is it a stepping stone towards though? What problem are we trying to actually solve here? We've seen both Eth1 and Eth2 function well without anything complicated around ensuring peers backfill blocks.

> As mentioned in other comments, once clients start supplying different ranges of blocks, in order to sanely be able to find viable peers, one needs an improved discovery mechanism at least.

No such mechanism exists on Eth1 and its never been an issue.  I find it very hard to believe it will be a problem on the beacon chain as that would imply that the vast majority of nodes are very recently sync'd which seems extremely unlikely.

djrtwo: So I agree with @arnetheduck that it is risky to *not* define this behavior has *potentially* malicious or lazy. If a peer can consistently say they don't have any blocks to serve, then it becomes much easier for nodes to not participate in block sync at all.

A compromise here might be something like "If a peer repeatedly does not have blocks available to serve since the start of the weak subjectivity period, they can be descored or disconnected". We then might be a bit more careful on defining "repeatedly" -- e.g. "After initial connection, if for more than 10 minutes a peer does not have..." 

Just an example. I know this adds complexity *if* a node wants to penalize such behavior but leaves open the option to do so

ajsutton: > A compromise here might be something like "If a peer repeatedly does not have blocks available to serve since the start of the weak subjectivity period, they can be descored or disconnected". We then might be a bit more careful on defining "repeatedly" -- e.g. "After initial connection, if for more than 10 minutes a peer does not have..."

That sounds reasonable.  I think it's also reasonable to say that if a node can't return blocks that your node /needs/ then it should disconnect immediately, possibly with a short term ban. If you're trying to sync and the remote node is trying to backfill you aren't useful to each other and are better of not being connected to each other.

That said, with the network conditions we currently have and experience with block availability on eth1, I'd say that actively requesting blocks you don't actually need is an unwarranted burden on the network.  We may need to perform those kinds of checks in the future if we can't find a better solution, but given the bandwidth concerns users already have adding more load isn't ideal.

specs/phase0/p2p-interface.md
```diff
@@ -752,18 +752,20 @@ Clients MUST keep a record of signed blocks seen on the epoch range
 where `current_epoch` is defined by the current wall-clock time,
 and clients MUST support serving requests of blocks on this range.
 
-Synced clients unable to reply to Block requests within the
+Peers are *repeatedly* unable to reply to Block requests within the
```
ajsutton, line 755:
```suggestion
Peers that are *repeatedly* unable to reply to Block requests within the
```

ajsutton (commented):
LGTM.

There was at one point an error code that could be returned for any blocks by range request the node couldn't fulfil because they don't yet have the blocks. I still think that would be very useful as it makes it explicit that the node doesn't have the blocks rather than returning the empty response which could mean they don't have the blocks or that there were no blocks.

specs/phase0/p2p-interface.md
```diff
@@ -745,10 +747,27 @@ The request MUST be encoded as an SSZ-container.
...
+*Note*: The above requirement implies that nodes that start from a recent weak subjectivity checkpoint
+MUST backfill the local block database to at least epoch `current_epoch - MIN_EPOCHS_FOR_BLOCK_REQUESTS`
+to be fully compliant with `BlocksByRange` requests. To safely perform such a
+backfill of blocks to the recent state, the node MUST validate both (1) the
+proposer signatures and (2) that the blocks form a valid chain up to the most
+recent block referenced in the weak subjectivity state.
+
+*Note*: Although clients that bootstrap from a weak subjectivity checkpoint can begin
+participating in the networking immediately, other peers that are actively block syncing MAY
+disconnect and/or temporarily ban such an un-synced or semi-synced client.
```
ralexstokes, line 767: likely worthwhile to let clients signal their status here

specs/phase0/weak-subjectivity.md
```diff
@@ -136,7 +136,9 @@ A brief reference for what these values look like in practice ([reference script
 
 ## Weak Subjectivity Sync
 
-Clients should allow users to input a Weak Subjectivity Checkpoint at startup, and guarantee that any successful sync leads to the given Weak Subjectivity Checkpoint along the canonical chain. If such a sync is not possible, the client should treat this as a critical and irrecoverable failure.
+Clients should allow users to input a Weak Subjectivity Checkpoint at startup,
+and guarantee that any successful sync leads to the given Weak Subjectivity Checkpoint along the canonical chain.
+If such a sync is not possible, the client should treat this as a critical and irrecoverable failure.
```
ralexstokes, line 141: this just looks like a whitespace change

ralexstokes (approved):
generally looks good! let some minor notes/comments

specs/phase0/p2p-interface.md
ralexstokes: so +1 here:

https://github.com/ethereum/eth2.0-specs/pull/2131#pullrequestreview-657285796

this may fall into the bucket of things that are implementation-dependent and *just* outside the spec

specs/phase0/p2p-interface.md
```diff
@@ -745,10 +747,27 @@ The request MUST be encoded as an SSZ-container.
...
 Each _successful_ `response_chunk` MUST contain a single `SignedBeaconBlock` payload.
 
-Clients MUST keep a record of signed blocks seen since the start of the weak subjectivity period
-and MUST support serving requests of blocks up to their own `head_block_root`.
+Clients MUST keep a record of signed blocks seen on the epoch range
+`[max(GENESIS_EPOCH, current_epoch - MIN_EPOCHS_FOR_BLOCK_REQUESTS), current_epoch]`
+where `current_epoch` is defined by the current wall-clock time,
+and clients MUST support serving requests of blocks on this range.
+
+Peers are *repeatedly* unable to reply to Block requests within the
```
arnetheduck, line 755:
```suggestion
Peers that are unable to reply to Block requests within the
```

`repeatedly` opens up for interpretation in a way that makes the whole statement ambiguous - there's already enough ambiguity in `MAY` to cover for repeatedly also.

specs/phase0/p2p-interface.md
```diff
@@ -745,10 +747,27 @@ The request MUST be encoded as an SSZ-container.
...
+
+*Note*: The above requirement implies that nodes that start from a recent weak subjectivity checkpoint
+MUST backfill the local block database to at least epoch `current_epoch - MIN_EPOCHS_FOR_BLOCK_REQUESTS`
+to be fully compliant with `BlocksByRange` requests. To safely perform such a
+backfill of blocks to the recent state, the node MUST validate both (1) the
+proposer signatures and (2) that the blocks form a valid chain up to the most
+recent block referenced in the weak subjectivity state.
+
+*Note*: Although clients that bootstrap from a weak subjectivity checkpoint can begin
+participating in the networking immediately, other peers that are actively block syncing MAY
```
arnetheduck, line 766:
```suggestion
participating in the networking immediately, other peers MAY
```

This section is somewhat contradictory - "participating in the network" entails serving blocks - also, limiting the "actively block syncing" removes the ability to detect freeloaders outside of syncing which them the opening they need, essentially. If we're going to put a `MUST` backfill, then we must also _at least_ allow disconnecting clients at any time if they do not fulfill this requirement.

There is an interesting question here what to do with read-only clients that light-sync to head but don't otherwise produce traffic - ie serving blocks and producing attestations/blocks go hand in hand because production increases network load and is therefor paid for with rewards, but "observers" do not add to traffic really.

arnetheduck:
> Because you have no way to validate a block from wallslot - MIN_EPOCHS_FOR_BLOCK_REQUESTS. You have no start block to check that it forms a chain, and you have no state to check state transition validity.

This is something we can fix I think: turning `historical_roots` into `historical_block_roots` (without mixing in the state), we could validate groups of 8192 blocks at a time which would provide a good upper bound on the damage one can cause.

specs/phase0/weak-subjectivity.md
djrtwo: right

djrtwo:
Added the `3: ResourceUnavailable` response code. I know we are generally pretty conservative on adding such codes, but 3 of 4 client teams are in vocal support

specs/phase0/p2p-interface.md
djrtwo: added

PR #2133, opened 2020-11-15
Make `lightclient` patch pass the linter

hwwhww:
Make #2130 executable + pass the linter.
Note: this PR doesn't enable `lightclient` fork spec tests.

### DIFF
--- a/specs/lightclient/beacon-chain.md
+++ b/specs/lightclient/beacon-chain.md
@@ -70,7 +70,17 @@ This is a standalone beacon chain patch adding light client support via sync com
 #### `BeaconBlockBody`
 
 ```python
-class BeaconBlockBody(phase0.BeaconBlockBody):
+class BeaconBlockBody(Container):
+    randao_reveal: BLSSignature
+    eth1_data: Eth1Data  # Eth1 data vote
+    graffiti: Bytes32  # Arbitrary data
+    # Operations
+    proposer_slashings: List[ProposerSlashing, MAX_PROPOSER_SLASHINGS]
+    attester_slashings: List[AttesterSlashing, MAX_ATTESTER_SLASHINGS]
+    attestations: List[Attestation, MAX_ATTESTATIONS]
+    deposits: List[Deposit, MAX_DEPOSITS]
+    voluntary_exits: List[SignedVoluntaryExit, MAX_VOLUNTARY_EXITS]
+    # Lightclient
     sync_committee_bits: Bitlist[MAX_SYNC_COMMITTEE_SIZE]
     sync_committee_signature: BLSSignature
 ```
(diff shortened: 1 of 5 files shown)


specs/lightclient/beacon-chain.md
```diff
@@ -70,15 +70,55 @@ This is a standalone beacon chain patch adding light client support via sync com
 #### `BeaconBlockBody`
 
 ```python
-class BeaconBlockBody(phase0.BeaconBlockBody):
```
JustinDrake, line 73: `class BeaconBlockBody(phase0.BeaconBlockBody):` is so much cleaner than repeating all the properties!

@protolambda: Can we do some Python magic here?

specs/lightclient/beacon-chain.md
```diff
@@ -70,15 +70,55 @@ This is a standalone beacon chain patch adding light client support via sync com
...
+    randao_reveal: BLSSignature
+    eth1_data: Eth1Data  # Eth1 data vote
+    graffiti: Bytes32  # Arbitrary data
+    # Operations
+    proposer_slashings: List[ProposerSlashing, MAX_PROPOSER_SLASHINGS]
+    attester_slashings: List[AttesterSlashing, MAX_ATTESTER_SLASHINGS]
+    attestations: List[Attestation, MAX_ATTESTATIONS]
+    deposits: List[Deposit, MAX_DEPOSITS]
+    voluntary_exits: List[SignedVoluntaryExit, MAX_VOLUNTARY_EXITS]
+    # Lightclient
```
JustinDrake, line 83: "Light client" is usually two words :)

specs/lightclient/beacon-chain.md
protolambda: With python we can do a lot, but the question is if we should. Being implicit would mean we need to spec appending of fields. And then sometimes appending is not the approach we want. Or functionality is not sequential, and we remove a field.

vbuterin: The reason why I did the inheriting from phase0 in the original PR instead of repeating properties is not just to cut down the number of lines of code, it's also to make the patch to the spec truly not sequential (ie. if it's implemented after sharding it could still be included as is). This would make it easier for us to work on light client stuff, the merge and sharding in parallel. So it would be really nice if we can make the inheritance work!

protolambda: That sounds great, but then do what we want some diamond structure? I.e. phase 0 type inherited by various phase 1 types, and then recombined in the complete phase 1 type? This becomes "the diamond problem" once we start overriding things in different types, which we then need to reconsolidate.

vbuterin: Well, I don't support the concept of "phases" anymore, so... :smile: 

Realistically, once some patch gets implemented, we could just quickly rewrite the other patches to point to the previous patch as a base (eg. `s/phase0./lightclient_fork./g`), and when writing each individual patch make sure that it is valid under any combination of the other patches.

This does of course assume that the absolute number of patches that are treated this way (as opposed to just being PRs to existing spec as all the beacon chain changes pre-launch were) needs to be very small.

protolambda: Ok, implemented it and opened a PR to my SSZ library that we use here in the spec, with support for all this (inheritance, mixins, diamonds): https://github.com/protolambda/remerkleable/pull/8

Please give it a review, and let's discuss on next spec call. It seems small, but it's quite a step away from bare struct definitions.

hwwhww:
Updates:
- Reverted back to inheritance. Some mypy errors
    ```python
    eth2spec/lightclient/spec.py:1678: error: Argument "default" to "max" has incompatible type "object"; expected "Eth1Data"
    eth2spec/lightclient/spec.py:1815: error: Unsupported left operand type for - ("object")
    ```
     - My guess is, it's old issue, `remerkleable` doesn't fully support mypy (#1707). We can propbably force casting them, as a workaround.
- Fixed upstream conflicts.
- Enabled `lightclient/sync-protocol` building. Many linter errors. IMO we can merge this PR into `vbuterin-patch-2` now and then fix them in PR #2130. @JustinDrake @vbuterin  What do you think?

specs/lightclient/beacon-chain.md
```diff
@@ -77,7 +78,8 @@ This is a standalone beacon chain patch adding light client support via sync com
 #### `BeaconBlockBody`
 
 ```python
-class BeaconBlockBody(phase0.BeaconBlockBody):
+class BeaconBlockBody(phase0.BeaconState):
```
vbuterin, line 81:
```suggestion
class BeaconBlockBody(phase0.BeaconBlockBody):
```

PR #2139, opened 2020-11-22
Redirects implementation list to one that is actually maintained.

MicahZoltu:
Unmaintained lists are bad, especially when there is a maintained one elsewhere.  Lets keep a single list.

@protolambda


protolambda (approved):
Thanks, LGTM

Issue #2125, opened 2020-11-06
Make inactivity leak quadratic per validator


vbuterin:
### Status quo and problem

Currently, the inactivity leak is computed as follows: in a given epoch, if there has not been finality for `d` epochs, any validator who is offline during that epoch is penalized `d / INACTIVITY_PENALTY_QUOTIENT` of their balance.

A major problem with this scheme is that it heavily penalizes honest validators who are online imperfectly: if fully online validators get leaked and lose 40% of their balance, someone who has been trying hard to stay online and succeeds at 90% of their duties would still lose 4% of their balance (approximately; ignoring compounding here for simplicity). Arguably this is unfair.

### Proposed fix

Change the inactivity leak rule to the following. For every validator, we maintain a counter `epochs_offline_since_finality`. When the chain finalizes, this counter is reset to zero; if the chain fails to finalize, and during that epoch the validator is absent, then (i) the validator is penalized `epochs_offline_since_finality / INACTIVITY_PENALTY_QUOTIENT` of their balance, and (ii) `epochs_offline_since_finality` increments by 1.

This means that during a leak, each individual validator's penalty is proportional to the square of the number of epochs that they personally have missed. Hence, if a validator who has been fully offline loses 40%, a validator who is online 90% of the time during that same period would only lose 0.4% of their balance, a much smaller amount that they could earn back within weeks of validating after the leak ends.

If validators have multiple duties, then we can rename `epochs_offline_since_finality` to `missed_duties_since_finality` and increment it by 1 for each missed duty, and adjust `INACTIVITY_PENALTY_QUOTIENT` appropriately to ensure the same total burn as today for fully offline validators.

### Effects on efficiency

`epochs_offline_since_finality` can be stored in a separate array, similar to validator balances, for efficiency. During normal (finalizing) periods, this array would not change, so costs would not increase at all. During non-finality, an additional ~8 bytes of hashing per validator would be required (or we could avoid this entirely by mashing together balance and `epochs_offline_since_finality` into a single variable).

If done in combination with https://github.com/ethereum/eth2.0-specs/issues/1340, efficiency can be maintained because we can still calculate the correct penalty for each validator if they had a previous `epochs_offline_since_finality` and were offline for `N` epochs: `log(balance) -= (epochs_offline_since_finality + (N-1)/2) * N / INACTIVITY_PENALTY_QUOTIENT` (any desired accuracy can be achieved with Taylor approximations of log/exp).

mcdee:
[...]

Issue #2098, opened 2020-10-09
Better recovery from inactivity leaks


dankrad:
**Rationale:**

The quadratic inactivity leak is a penalty for being offline that increases in time quadratically in periods where the beacon chain doesn't finalized. It has the effect that should more than 1/3 of validators be offline, within a few days or weeks these validators will see increased losses to their balances, reducing them back to a total of 1/3 of the validator set and enabling finalization. (One popular misunderstanding is that it requires validator ejection to work -- it does not depend on ejecting the offline validators, only reducing their balances to diminish their weight in the FFG votes)

However, at this point, the quadratic leak will stop for all validators, including those that remain offline. This means that if it was caused by a catastrophic event permanently disabling >1/3 of validators, or is due to a "User Activated Soft Fork" removing some validators deemed malicious from the network, the network will be in a state where only exactly 2/3+1 of the validator balance is online when it recovers due to the leak. The offline validators will continues to suffer a normal inactivity leak, but this is just the negative of the usual rewards and will thus take on the order of months to years to have any meaningful effect. This means that the network is going to be precarious for a long time, only having just enough online stake to get finalized when everything works perfectly. In practice, this means finalization will very often be delayed or not happening for extended periods of time just due to minor outages.

**Idea:**

To fix this poor outcome, we can change the inactivity leak as follows: Instead of stopping the leak for all validators when the chain finalizes, we can keep the leak going for all validators until they actually sign another attestation. This will make sure that validators which are permanently offline keep experiencing a much quicker drain, will actually be kicked out of the validator set and the chain can hopefully recover to almost 100% participation.

mcdee:
If you're keeping track of the last time that a validator attested then wouldn't it be simpler to just apply the quadratic leak to all validators at all times as a "non-participation" penalty (removing the existing penalty system)?

dankrad:
> If you're keeping track of the last time that a validator attested then wouldn't it be simpler to just apply the quadratic leak to all validators at all times as a "non-participation" penalty (removing the existing penalty system)?

That is probably a simpler approach, but may be seen as too punitive? We typically want to punish validators if their individual faults have actually led to real safety/liveness problems on chain.

One intermediate: The quadratic leak applies during the period when a validator is offline, if during that period there was a period of 8 epochs of non-finalization.

[...]

PR #2140, opened 2020-11-28
Incentive accounting reform

vbuterin:
Brings forward the incentive accounting reforms from #2024 as a standalone patch.

Main goals are (i) simplicity, and (ii) paving the way for a simpler transition to further efficiency improvements.

### DIFF
--- a/specs/accounting_reform/beacon-chain.md
+++ b/specs/accounting_reform/beacon-chain.md
@@ -1,0 +1,282 @@
+# Incentive accounting  reforms
+
+## Table of contents
+
+<!-- TOC -->
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+
+- [Introduction](#introduction)
+- [Constants](#constants)
+  - [Flags](#flags)
+- [Containers](#containers)
+  - [Modified containers](#extended-containers)
+    - [`BeaconState`](#beaconstate)
+- [Helper functions](#helper-functions)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+## Introduction
+
+This is a proposed reform to the eth2 spec that significantly simplifies incentive accounting, removing PendingAttestations.
+
+## Constants
+
+### Rewards
+
+| Constant | Value |
+| - | - |
+| `TARGET_NUMERATOR` | 32 |
+| `TIMELY_NUMERATOR` | 12 |
+| `HEAD_AND_VERY_TIMELY_NUMERATOR` | 12 |
+| `REWARD_DENOMINATOR` | 64 |
+
+The numerators add up to 7/8, leaving the remaining 1/8 for proposer rewards and other micro rewards to be added in the future.
+
+### Flags
+
+| Flag | Value |
+| - | - |
+| `FLAG_TARGET` | 0 |
+| `FLAG_TIMELY` | 1 |
+| `FLAG_HEAD_AND_VERY_TIMELY` | 2 |
+
+### Convenience structure for flags and numerators
+
+| Name | Value |
+| - | - |
+| `FLAGS_AND_NUMERATORS` | `((FLAG_TARGET, TARGET_NUMERATOR), (FLAG_TIMELY, TIMELY_NUMERATOR), (FLAG_HEAD_AND_VERY_TIMELY, HEAD_AND_VERY_TIMELY_NUMERATOR))` |
+
+## Containers
+
+### Modified containers
+
+#### `BeaconState`
+
+```python
+class BeaconState(Container):
+    # Versioning
+    genesis_time: uint64
+    genesis_validators_root: Root
+    slot: Slot
+    fork: Fork
+    # History
+    latest_block_header: BeaconBlockHeader
+    block_roots: Vector[Root, SLOTS_PER_HISTORICAL_ROOT]
+    state_roots: Vector[Root, SLOTS_PER_HISTORICAL_ROOT]
+    historical_roots: List[Root, HISTORICAL_ROOTS_LIMIT]
+    # Eth1
+    eth1_data: Eth1Data
+    eth1_data_votes: List[Eth1Data, EPOCHS_PER_ETH1_VOTING_PERIOD * SLOTS_PER_EPOCH]
+    eth1_deposit_index: uint64
+    # Registry
+    validators: List[Validator, VALIDATOR_REGISTRY_LIMIT]
+    balances: List[Gwei, VALIDATOR_REGISTRY_LIMIT]
+    # Randomness
+    randao_mixes: Vector[Bytes32, EPOCHS_PER_HISTORICAL_VECTOR]
+    # Slashings
+    slashings: Vector[Gwei, EPOCHS_PER_SLASHINGS_VECTOR]  # Per-epoch sums of slashed effective balances
+    # Current and previous epoch participation flags
+    previous_epoch_flags: List[Bitvector[8], VALIDATOR_REGISTRY_LIMIT]
+    current_epoch_flags: List[Bitvector[8], VALIDATOR_REGISTRY_LIMIT]
+    # Finality
+    justification_bits: Bitvector[JUSTIFICATION_BITS_LENGTH]  # Bit set for every recent justified epoch
+    previous_justified_checkpoint: Checkpoint
+    current_justified_checkpoint: Checkpoint
+    finalized_checkpoint: Checkpoint
+```
+
+###  Helpers
+
+##### `get_base_reward`
+
+```python
+def get_base_reward(state: BeaconState, index: ValidatorIndex) -> Gwei:
+    total_balance = get_total_active_balance(state)
+    effective_balance = state.validators[index].effective_balance
+    return Gwei(effective_balance * BASE_REWARD_FACTOR // integer_squareroot(total_balance))
+```
+
+Note: remove the `// BASE_REWARDS_PER_EPOCH`
+
+##### `process_attestation`
+
+```python
+def process_attestation(state: BeaconState, attestation: Attestation) -> None:
+    data = attestation.data
+    assert data.target.epoch in (get_previous_epoch(state), get_current_epoch(state))
+    assert data.target.epoch == compute_epoch_at_slot(data.slot)
+    assert data.slot + MIN_ATTESTATION_INCLUSION_DELAY <= state.slot <= data.slot + SLOTS_PER_EPOCH
+    assert data.index < get_committee_count_per_slot(state, data.target.epoch)
+
+    committee = get_beacon_committee(state, data.slot, data.index)
+    assert len(attestation.aggregation_bits) == len(committee)
+        
+    if data.target.epoch == get_current_epoch(state):
+        flags = state.current_epoch_flags
+        assert data.source == state.current_justified_checkpoint
+    else:
+        flags = state.previous_epoch_flags
+        assert data.source == state.previous_justified_checkpoint    
+
+    # Verify signature
+    assert is_valid_indexed_attestation(state, get_indexed_attestation(state, attestation))
+        
+    # Process flags
+    flags_to_set = []
+    if data.target.root == get_block_root(state, data.target.epoch):
+        flags_to_set.append(FLAG_TARGET)
+        if data.beacon_block_root == get_block_root_at_slot(state, data.slot) and state.slot - data.slot == 1:
+            flags_to_set.append(FLAG_HEAD_AND_VERY_TIMELY)
+    if state.slot - data.slot <= integer_squareroot(SLOTS_PER_EPOCH):
+        flags_to_set.append(FLAG_TIMELY)
+
+    # Update participation flags
+    for participant in get_attesting_indices(state, data, attestation.aggregation_bits):
+        active_position = get_active_validator_indices(state, data.target.epoch).index(participant)
+        for flag, numerator in FLAGS_AND_NUMERATORS:
+            if flag in flags_to_set:
+                # Give proposer reward for new flags
+                if not flags[active_position][flag]:
+                    proposer_reward = Gwei(
+                        get_base_reward(state, participant) * numerator
+                        // (REWARD_DENOMINATOR * PROPOSER_REWARD_DENOMINATOR)
+                    )
+                increase_balance(state, get_beacon_proposer_index(state), proposer_reward)
+            flags[active_position][flag] = True
+```
+
+##### `get_unslashed_participant_indices`
+
+```python
+def get_unslashed_participant_indices(state: BeaconState, flag: uint8, epoch: Epoch) -> Set[ValidatorIndex]:
+    assert epoch in [get_current_epoch(state), get_previous_epoch(state)]
+
+    flags = state.current_epoch_reward_flags if epoch == get_current_epoch(state) else state.previous_epoch_reward_flags
+    participant_indices = [
+        index for i, index in enumerate(get_active_validator_indices(state, epoch))
+        if flags[i][flag]
+    ]
+    return set(filter(lambda index: not state.validators[index].slashed, participant_indices))
+```
+
+##### `process_justification_and_finalization`
+
+```python
+def process_justification_and_finalization(state: BeaconState) -> None:
+    # Initial FFG checkpoint values have a `0x00` stub for `root`.
+    # Skip FFG updates in the first two epochs to avoid corner cases that might result in modifying this stub.
+    if get_current_epoch(state) <= GENESIS_EPOCH + 1:
+        return
+
+    previous_epoch = get_previous_epoch(state)
+    current_epoch = get_current_epoch(state)
+    old_previous_justified_checkpoint = state.previous_justified_checkpoint
+    old_current_justified_checkpoint = state.current_justified_checkpoint
+
+    # Process justifications
+    state.previous_justified_checkpoint = state.current_justified_checkpoint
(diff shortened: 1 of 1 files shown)


specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,282 @@
...
+        assert data.source == state.previous_justified_checkpoint    
+
+    # Verify signature
+    assert is_valid_indexed_attestation(state, get_indexed_attestation(state, attestation))
+        
+    # Process flags
+    flags_to_set = []
+    if data.target.root == get_block_root(state, data.target.epoch):
+        flags_to_set.append(FLAG_TARGET)
+        if data.beacon_block_root == get_block_root_at_slot(state, data.slot) and state.slot - data.slot == 1:
```
djrtwo, line 131: Use `MIN_ATTESTATION_INCLUSION_DELAY`

```suggestion
        is_correct_head = data.beacon_block_root == get_block_root_at_slot(state, data.slot)
        is_very_timely = state.slot == data.slot + MIN_ATTESTATION_INCLUSION_DELAY
        if is_correct_head and is_very_timely:
```

specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,282 @@
...
+            flags_to_set.append(FLAG_HEAD_AND_VERY_TIMELY)
+    if state.slot - data.slot <= integer_squareroot(SLOTS_PER_EPOCH):
```
djrtwo, line 133: A little bit safer (looking)
```suggestion
    if state.slot <= data.slot + integer_squareroot(SLOTS_PER_EPOCH):
```

specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,282 @@
...
+        active_position = get_active_validator_indices(state, data.target.epoch).index(participant)
+        for flag, numerator in FLAGS_AND_NUMERATORS:
+            if flag in flags_to_set:
+                # Give proposer reward for new flags
+                if not flags[active_position][flag]:
+                    proposer_reward = Gwei(
+                        get_base_reward(state, participant) * numerator
+                        // (REWARD_DENOMINATOR * PROPOSER_REWARD_DENOMINATOR)
+                    )
+                increase_balance(state, get_beacon_proposer_index(state), proposer_reward)
```
djrtwo, line 147:
```suggestion
                    increase_balance(state, get_beacon_proposer_index(state), proposer_reward)
```

specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,282 @@
...
+            flags[active_position][flag] = True
```
djrtwo, line 148:
```suggestion
                flags[active_position][flag] = True
```

specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,282 @@
...
+def process_rewards_and_penalties(state: BeaconState) -> None:
+    # No rewards are applied at the end of `GENESIS_EPOCH` because rewards are for work done in the previous epoch
+    if get_current_epoch(state) == GENESIS_EPOCH:
+        return
+
+    flag_deltas = [
+        get_standard_flag_deltas(state, flag, numerator) for (flag, numerator) in FLAGS_AND_NUMERATORS
+    ]
+    
+    for (rewards, penalties) in flag_deltas + [get_inactivity_penalty_deltas(state)]:
```
djrtwo, line 278:
```suggestion
    all_deltas = flag_deltas + [get_inactivity_penalty_deltas(state)]
    for (rewards, penalties) in all_deltas:
```

djrtwo:
Still not 100% sure how to handle the build process for these minor upgrades. need to think/discuss more before merging

nathsmaR (approved):
0x017D05EA6E31A91dC4435FbA7B8F95A069577d7a

nathsmaR (changes requested):
0x017D05EA6E31A91dC4435FbA7B8F95A069577d7a

specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,282 @@
...
+                        // (REWARD_DENOMINATOR * PROPOSER_REWARD_DENOMINATOR)
+                    )
+                increase_balance(state, get_beacon_proposer_index(state), proposer_reward)
+            flags[active_position][flag] = True
+```
+
+##### `get_unslashed_participant_indices`
+
+```python
+def get_unslashed_participant_indices(state: BeaconState, flag: uint8, epoch: Epoch) -> Set[ValidatorIndex]:
```
nathsmaR, line 157: 0x017D05EA6E31A91dC4435FbA7B8F95A069577d7a

nathsmaR (approved):

specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,285 @@
...
+
+```python
+def get_inactivity_penalty_deltas(state: BeaconState) -> Tuple[Sequence[Gwei], Sequence[Gwei]]:
+    """
+    Return inactivity reward/penalty deltas for each validator.
+    Note: function exactly the same as Phase 0 other than the selection of `matching_target_attesting_indices`
+    """
+    penalties = [Gwei(0) for _ in range(len(state.validators))]
+    
+    total_usual_reward_numerator = sum([numerator for (flag, numerator) in FLAGS_AND_NUMERATORS])
```
hwwhww, line 252:
```suggestion
    total_usual_reward_numerator = sum(numerator for (_, numerator) in FLAGS_AND_NUMERATORS)
```

specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,285 @@
...
+                increase_balance(state, get_beacon_proposer_index(state), proposer_reward)
+```
+
+##### `get_unslashed_participant_indices`
+
+```python
+def get_unslashed_participant_indices(state: BeaconState, flag: uint8, epoch: Epoch) -> Set[ValidatorIndex]:
+    assert epoch in [get_current_epoch(state), get_previous_epoch(state)]
+
+    flags = state.current_epoch_reward_flags if epoch == get_current_epoch(state) else state.previous_epoch_reward_flags
```
hwwhww, line 160:
```suggestion
    flags = state.current_epoch_flags if epoch == get_current_epoch(state) else state.previous_epoch_flags
```

specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,285 @@
...
+        is_very_timely = state.slot == data.slot + MIN_ATTESTATION_INCLUSION_DELAY
+        if is_correct_head and is_very_timely:
+            flags_to_set.append(FLAG_HEAD_AND_VERY_TIMELY)
+
+    if state.slot <= data.slot + integer_squareroot(SLOTS_PER_EPOCH):
+        flags_to_set.append(FLAG_TIMELY)
+
+    # Update participation flags
+    for participant in get_attesting_indices(state, data, attestation.aggregation_bits):
+        active_position = get_active_validator_indices(state, data.target.epoch).index(participant)
```
hwwhww, line 142: Take pyspec for example, it seems that `state` object will be changed when setting `flags[active_position][flag] = True` with mutable side effect and the LRU cache keys of `get_active_validator_indices` will be different for each `participant`.

A slight refactoring may make implementers happier:

```suggestion
    active_validator_indices = get_active_validator_indices(state, data.target.epoch)
    for participant in get_attesting_indices(state, data, attestation.aggregation_bits):
        active_position = active_validator_indices.index(participant)
```

JustinDrake:
@vbuterin I have made the following substantive changes. Please double check :)

1) Improved proposer reward accounting in `process_attestation` by introducing `proposer_reward_numerator`. This is a performance optimisation which reduces the number of divisions and also minimises rounding errors.
2) Fixed `get_inactivity_penalty_deltas` where `// INACTIVITY_PENALTY_QUOTIENT` should be `// REWARD_DENOMINATOR`.
3) Fixed `get_flag_deltas` where there was an extraneous `total_balance` when `is_in_inactivity_leak(state)` is `True`.
4) Give head timeliness reward even if the target is wrong. This makes the timeliness logic clean and symmetric. Neither honest nor rational validators are affected.
5) Align the validators indices of `state.[previous/current]_epoch_participation` with `state.validators` and `state.balances`. This is for light-client friendliness and simpler consensus. The additional RAM should be minimal especially when we have logic to garbage-collect unused validator indices.
6) Reorder flag indices (done for cosmetic reasons, technically a substantive change).

specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,278 @@
...
+    eth1_deposit_index: uint64
+    # Registry
+    validators: List[Validator, VALIDATOR_REGISTRY_LIMIT]
+    balances: List[Gwei, VALIDATOR_REGISTRY_LIMIT]
+    # Randomness
+    randao_mixes: Vector[Bytes32, EPOCHS_PER_HISTORICAL_VECTOR]
+    # Slashings
+    slashings: Vector[Gwei, EPOCHS_PER_SLASHINGS_VECTOR]  # Per-epoch sums of slashed effective balances
+    # Participation
+    previous_epoch_participation: List[Bitvector[PARTICIPATION_FLAGS_LENGTH], VALIDATOR_REGISTRY_LIMIT]
```
dankrad, line 69: I feel like we should swap the order of the validator and flag indices here (make it `List[Bitlist[VALIDATOR_REGISTRY_LIMIT], PARTICIPATION_FLAGS_LENGTH]`): Currently Bitvectors are padded to full bytes when serializing or Merkleizing, meaning that each 3 bit vector will take up a full byte of space.

specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,278 @@
...
+| `TIMELY_TARGET_NUMERATOR` | `32` |
+| `REWARD_DENOMINATOR` | `64` |
+
+The reward fractions add up to 7/8, leaving the remaining 1/8 for proposer rewards and other future micro-rewards.
+
+### Misc
+
+| Name | Value |
+| - | - |
+| `PARTICIPATION_FLAGS_LENGTH` | `8` |
```
dankrad, line 36: Why is this 8 and not 3?

dankrad:
> That is, item `k` in the data structure would not literally be the k'th validator; instead, it would be the k'th validator _in the shuffled list_.

I'm confused, I don't see this (participation list shuffled according to current validator shuffle) implemented in this PR. Why is that?

specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,278 @@
...
+    """
+    Return inactivity reward/penalty deltas for each validator.
+    """
+    penalties = [Gwei(0) for _ in range(len(state.validators))]
+    if is_in_inactivity_leak(state):
+        reward_numerator_sum = sum(numerator for (_, numerator) in FLAGS_AND_NUMERATORS)
+        matching_target_indices = get_unslashed_participating_indices(state, TIMELY_TARGET_FLAG, get_previous_epoch(state))
+        for index in get_eligible_validator_indices(state):
+            # If validator is performing optimally this cancels all attestation rewards for a neutral balance
+            penalties[index] += get_base_reward(state, index) * reward_numerator_sum // REWARD_DENOMINATOR
```
dankrad, line 243: Given that you have to attest within one slot to get the head reward, and five epochs for the source reward, this seems even harsher than the previous computation (which was already heavily criticised for penalizing even optimally performing validators). In fact, getting the head right within one epoch is impossible at least 33% of the time, meaning that you are guaranteed to be down ~-1/6 already.

I'd suggest scaling this by 1/2, meaning you can be slightly positive still (which I think we have determined in the past is not a problem).

JustinDrake:
> I'm confused, I don't see this (participation list shuffled according to current validator shuffle) implemented in this PR. Why is that?

I removed it :) See point 5 [here](https://github.com/ethereum/eth2.0-specs/pull/2140#issuecomment-748039592).

> Align the validators indices of state.[previous/current]_epoch_participation with state.validators and state.balances. This is for light-client friendliness and simpler consensus. The additional RAM should be minimal especially when we have logic to garbage-collect unused validator indices.

dankrad:
I was a big fan of making the quadratic leak per validator, are we still considering this (cf #2125, #2098)?

To avoid future cock-ups #2152, it would be a cool idea to add tests that check that rewards are roughly like we are expecting. I.e. run an actual chain with 100 validators for a few epochs and verify that rewards are roughly what we are expecting on a high level, like proposer rewards = 1/8 of total rewards.

dankrad:
> @dankrad I removed it :) See point 5 [here](https://github.com/ethereum/eth2.0-specs/pull/2140#issuecomment-748039592).

But the point of that is to reduce hashing. Now you essentially have to rehash the full vector at every attestation, which was what we wanted to avoid AFAIU?

JustinDrake:
> But the point of that is to reduce hashing. Now you essentially have to rehash the full vector at every attestation, which was what we wanted to avoid AFAIU?

Ah right, my bad 🤦 And you actually do have light-client friendliness because we have a light-client friendly shuffle. I'll restore the logic :)

JustinDrake:
> But the point of that is to reduce hashing. Now you essentially have to rehash the full vector at every attestation, which was what we wanted to avoid AFAIU?

Right, my bad 🤦  Will fix it. And my worry about it unfriendly to light clients is unfounded because we have a light client friendly shuffle :)

specs/accounting_reform/beacon-chain.md
vbuterin: The problem I see with doing that it that it would increase the amount of Merkling you need to do to process an attestation by ~3x, as you would have to adjust the branches in three positions instead of one. I think on net it would be a reduction in efficiency (as a Merkle branch going into a committee is in most cases longer than the committee itself).

specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,278 @@
...
+
+    # Verify signature
+    assert is_valid_indexed_attestation(state, get_indexed_attestation(state, attestation))
+
+    # Participation flags
+    participation_flags = []
+    if is_matching_head and state.slot <= data.slot + MIN_ATTESTATION_INCLUSION_DELAY:
+        participation_flags.append(TIMELY_HEAD_FLAG)
+    if is_matching_source and state.slot <= data.slot + integer_squareroot(SLOTS_PER_EPOCH):
+        participation_flags.append(TIMELY_SOURCE_FLAG)
```
vbuterin, line 127: Any reason why we're checking `is_matching_source` here when it's already a condition of the attestation getting processed at all? (see `assert is_matching_source` above).

specs/accounting_reform/beacon-chain.md
```diff
@@ -0,0 +1,278 @@
...
+*Note*: The function `get_inactivity_penalty_deltas` is modified in the selection of matching target indices and the removal of `BASE_REWARDS_PER_EPOCH`.
+
+```python
+def get_inactivity_penalty_deltas(state: BeaconState) -> Tuple[Sequence[Gwei], Sequence[Gwei]]:
+    """
+    Return inactivity reward/penalty deltas for each validator.
+    """
+    penalties = [Gwei(0) for _ in range(len(state.validators))]
+    if is_in_inactivity_leak(state):
+        reward_numerator_sum = sum(numerator for (_, numerator) in FLAGS_AND_NUMERATORS)
```
vbuterin, line 239:
```suggestion
        reward_numerator_sum = TARGET_NUMERATOR + TIMELY_SOURCE_NUMERATOR + TIMELY_HEAD_NUMERATOR // 2
```

specs/accounting_reform/beacon-chain.md
vbuterin: You mean like in the suggestion I just made directly above?

specs/accounting_reform/beacon-chain.md
dankrad: I see, We could also do the following:
1. Change the SSZ spec to allow for tight packing of Bitlists (Don't think we have used vectors of Bitlists, so this should be fine)
2. Make it a Bitlist[PARTICIPATION_FLAGS_LENGTH * VALIDATOR_REGISTRY_LIMIT] and address manually.

I think 2 would be easy to do and would suggest that.

specs/accounting_reform/beacon-chain.md
dankrad: I even suggested being more lenient and dividing the whole `reward_numerator_sum` by 2.

specs/accounting_reform/beacon-chain.md
JustinDrake: The reason is purely cosmetic (as you noticed). It highlights that head, source and target are treated symmetrically just with different constants `X`, `Y`, `Z`.

```
    if is_matching_head and state.slot <= data.slot + X:
        participation_flags.append(TIMELY_HEAD_FLAG)
    if is_matching_source and state.slot <= data.slot + Y:
        participation_flags.append(TIMELY_SOURCE_FLAG)
    if is_matching_target and state.slot <= data.slot + Z:
        participation_flags.append(TIMELY_TARGET_FLAG)
```

djrtwo: I support it. Depending on how a client may choose slice up operations having this explicit makes it clear and easy to reason about

specs/accounting_reform/beacon-chain.md
djrtwo: Yeah, think it should be 3
Unless @vbuterin is padding out for future flags, but I don't think that's very valuable because we'd have to hard-fork to add them anyway

hwwhww:
Should we close this PR and move the discussion to #2159?

djrtwo:
Closing in favor of #2176
Please pick up conversations/suggestions there

M9na1 (approved):

PR #2141, opened 2020-12-01
executable light client patch: beacon-chain.md

hwwhww:
Patch for #2130

### This PR
- [x] Bump remerkleable to 0.1.18 https://github.com/ethereum/eth2.0-specs/commit/ae86517813ef462a5f50e8668b8e32b7aacd3527 Thanks @protolambda!
- [x] Use *new* `optional_fast_aggregate_verify` https://github.com/ethereum/eth2.0-specs/commit/f0864b8e92639d35d83292cc18fb81774723f28f  /cc @JustinDrake 
- [x] Add `LIGHTCLIENT` fork to `test/context.py`
- [x] Make it pass the most phase0 tests except for the inactivity leak tests (see below)

### TODO in this PR:
- [x] Rename `LIGHTCLIENT` fork to another code name that is more like a "fork".  Does `LIGHTCLIENT_PATCH` work? /cc @djrtwo 

### TODOs in other PRs:
- [x] Fix inactivity leak issues (see https://github.com/ethereum/eth2.0-specs/pull/2130#discussion_r533518677 discussions)
    - failed tests
    ```
    FAILED eth2spec/test/phase0/epoch_processing/test_process_rewards_and_penalties.py::test_almost_empty_attestations_with_leak
    FAILED eth2spec/test/phase0/epoch_processing/test_process_rewards_and_penalties.py::test_random_fill_attestations_with_leak
    FAILED eth2spec/test/phase0/epoch_processing/test_process_rewards_and_penalties.py::test_almost_full_attestations_with_leak
    FAILED eth2spec/test/phase0/epoch_processing/test_process_rewards_and_penalties.py::test_full_attestation_participation_with_leak
    ```
- [ ] This PR disabled `sync-protocol.md` for now. Let's make it executable with other PRs.
- [ ] Add light client patch tests

### DIFF
--- a/configs/mainnet/lightclient_patch.yaml
+++ b/configs/mainnet/lightclient_patch.yaml
@@ -1,6 +1,6 @@
 # Mainnet preset - lightclient patch
 
-CONFIG_NAME: "minimal"
+CONFIG_NAME: "mainnet"
 
 # ---------------------------------------------------------------
 
--- a/configs/mainnet/lightclient_patch.yaml
+++ b/configs/mainnet/lightclient_patch.yaml
@@ -2,9 +2,20 @@
 
 CONFIG_NAME: "mainnet"
 
+# Misc
 # ---------------------------------------------------------------
-
 # 2**10 (=1,024)
 SYNC_COMMITTEE_SIZE: 1024
 # 2**6 (=64)
 SYNC_COMMITTEE_PUBKEY_AGGREGATES_SIZE: 64
+
+
+# Time parameters
+# ---------------------------------------------------------------
+# 2**8 (= 256)
+EPOCHS_PER_SYNC_COMMITTEE_PERIOD: 256
+
+
+# Signature domains
+# ---------------------------------------------------------------
+DOMAIN_SYNC_COMMITTEE: 0x07000000
--- a/configs/mainnet/lightclient_patch.yaml
+++ b/configs/mainnet/lightclient_patch.yaml
@@ -1,0 +1,10 @@
+# Mainnet preset - lightclient patch
+
+CONFIG_NAME: "minimal"
+
+# ---------------------------------------------------------------
+
+# 2**10 (=1,024)
+SYNC_COMMITTEE_SIZE: 1024
+# 2**6 (=64)
+SYNC_COMMITTEE_PUBKEY_AGGREGATES_SIZE: 64
--- a/configs/minimal/lightclient_patch.yaml
+++ b/configs/minimal/lightclient_patch.yaml
@@ -1,0 +1,8 @@
+# Minimal preset - lightclient patch
+
+CONFIG_NAME: "minimal"
+
+# ---------------------------------------------------------------
+
+SYNC_COMMITTEE_SIZE: 64
+SYNC_COMMITTEE_PUBKEY_AGGREGATES_SIZE: 16
--- a/configs/minimal/lightclient_patch.yaml
+++ b/configs/minimal/lightclient_patch.yaml
@@ -2,7 +2,20 @@
 
 CONFIG_NAME: "minimal"
 
+# Misc
 # ---------------------------------------------------------------
-
+# [customized]
 SYNC_COMMITTEE_SIZE: 64
+# [customized]
 SYNC_COMMITTEE_PUBKEY_AGGREGATES_SIZE: 16
+
+
+# Time parameters
+# ---------------------------------------------------------------
+# 2**8 (= 256)
+EPOCHS_PER_SYNC_COMMITTEE_PERIOD: 256
+
+
+# Signature domains
+# ---------------------------------------------------------------
+DOMAIN_SYNC_COMMITTEE: 0x07000000
--- a/specs/lightclient/beacon-chain.md
+++ b/specs/lightclient/beacon-chain.md
@@ -20,6 +20,8 @@
   - [New containers](#new-containers)
     - [`SyncCommittee`](#synccommittee)
 - [Helper functions](#helper-functions)
+  - [`Predicates`](#predicates)
+    - [`optional_fast_aggregate_verify`](#optional_fast_aggregate_verify)
   - [Beacon state accessors](#beacon-state-accessors)
     - [`get_sync_committee_indices`](#get_sync_committee_indices)
     - [`get_sync_committee`](#get_sync_committee)
--- a/specs/lightclient/beacon-chain.md
+++ b/specs/lightclient/beacon-chain.md
@@ -26,6 +26,7 @@
     - [`get_sync_committee_indices`](#get_sync_committee_indices)
     - [`get_sync_committee`](#get_sync_committee)
   - [Block processing](#block-processing)
+    - [Block header](#block-header)
     - [Sync committee processing](#sync-committee-processing)
   - [Epoch processing](#epoch-processing)
     - [Components of attestation deltas](#components-of-attestation-deltas)
@@ -111,16 +112,6 @@ class SyncCommittee(Container):
     pubkey_aggregates: Vector[BLSPubkey, SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_PUBKEY_AGGREGATES_SIZE]
 ```
 
-## Upgrade from phase0
-
-```python
-def upgrade_to_lightclient_patch(pre: phase0.BeaconState) -> BeaconState:
-    """
-    The light client patch copies over all phase0 fields, and initializes new fields with default values.
-    """
-    return BeaconState(**{k: getattr(pre, k) for k in pre.fields().keys()})
-```
-
 ## Helper functions
 
 ### `Predicates`
@@ -187,8 +178,42 @@ def get_sync_committee(state: BeaconState, epoch: Epoch) -> SyncCommittee:
 
 ```python
 def process_block(state: BeaconState, block: BeaconBlock) -> None:
-    phase0.process_block(state, block)
+    process_block_header(state, block)
+    process_randao(state, block.body)
+    process_eth1_data(state, block.body)
+    process_operations(state, block.body)
     process_sync_committee(state, block)
+```
+
+#### Block header
+
+Process `sync_committee_bits` and `sync_committee_signature` fields.
+
+```python
+def process_block_header(state: BeaconState, block: BeaconBlock) -> None:
+    # Verify that the slots match
+    assert block.slot == state.slot
+    # Verify that the block is newer than latest block header
+    assert block.slot > state.latest_block_header.slot
+    # Verify that proposer index is the correct index
+    assert block.proposer_index == get_beacon_proposer_index(state)
+    # Verify that the parent matches
+    assert block.parent_root == hash_tree_root(state.latest_block_header)
+    # Cache current block as the new latest block
+    state.latest_block_header = BeaconBlockHeader(
+        slot=block.slot,
+        proposer_index=block.proposer_index,
+        parent_root=block.parent_root,
+        state_root=Bytes32(),  # Overwritten in the next process_slot call
+        body_root=hash_tree_root(block.body),
+        # Lightclient patch fields
+        sync_committee_bits=block.sync_committee_bits,
+        sync_committee_signature=block.sync_committee_signature,
+    )
+
+    # Verify proposer is not slashed
+    proposer = state.validators[block.proposer_index]
+    assert not proposer.slashed
 ```
 
 #### Sync committee processing
--- a/specs/lightclient/beacon-chain.md
+++ b/specs/lightclient/beacon-chain.md
@@ -111,6 +111,16 @@ class SyncCommittee(Container):
     pubkey_aggregates: Vector[BLSPubkey, SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_PUBKEY_AGGREGATES_SIZE]
 ```
 
+## Upgrade from phase0
+
+```python
+def upgrade_to_lightclient_patch(pre: phase0.BeaconState) -> BeaconState:
+    """
+    The light client patch copies over all phase0 fields, and initializes new fields with default values.
+    """
+    return BeaconState(**{k: getattr(pre, k) for k in pre.fields().keys()})
+```
+
 ## Helper functions
 
 ### `Predicates`
--- a/specs/lightclient/beacon-chain.md
+++ b/specs/lightclient/beacon-chain.md
@@ -15,8 +15,7 @@
   - [Domain types](#domain-types)
 - [Containers](#containers)
   - [Extended containers](#extended-containers)
-    - [`BeaconBlock`](#beaconblock)
-    - [`BeaconBlockHeader`](#beaconblockheader)
+    - [`BeaconBlockBody`](#beaconblockbody)
     - [`BeaconState`](#beaconstate)
   - [New containers](#new-containers)
     - [`SyncCommittee`](#synccommittee)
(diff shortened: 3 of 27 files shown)


specs/phase0/validator.md
```diff
@@ -328,6 +328,12 @@ def get_eth1_vote(state: BeaconState, eth1_chain: Sequence[Eth1Block]) -> Eth1Da
     valid_votes = [vote for vote in state.eth1_data_votes if vote in votes_to_consider]
 
     # Default vote on latest eth1 block data in the period range unless eth1 chain is not live
+    # FIXME: Non-substantive casting: a workaround for passing typing linter for future patch forks
+    state_eth1_data = Eth1Data(
+        deposit_root=state.eth1_data.deposit_root,
+        deposit_count=state.eth1_data.deposit_count,
+        block_hash=state.eth1_data.block_hash,
+    )
```
hwwhww, line 336: /cc @protolambda 

This is a sequela of OOP SSZ of the new `remerkleable`. The linter doesn't recognize the Phase0 fields in the new Lightclient `BeaconState` statically. They are `object` type to mypy.

A quick workaround here is to define a `state_eth1_data = Eth1Data(...)` variable since `max()` function requires homogeneous list.

1. Do you think it's possible to fix it in `remerkleable`?
2. Or, is there a better way to casting `Container`?

If not, not ideal but I'm okay to leave it as-is for now and open an issue... 😅

specs/lightclient/beacon-chain.md
```diff
@@ -151,13 +172,13 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
     # Verify sync committee aggregate signature signing over the previous slot block root
-    previous_slot = max(state.slot, Slot(1)) - Slot(1)
+    previous_slot = max(Slot(state.slot), Slot(1)) - Slot(1)
     committee_indices = get_sync_committee_indices(state, get_current_epoch(state))
     participant_indices = [committee_indices[i] for i in range(len(committee_indices)) if body.sync_committee_bits[i]]
     participant_pubkeys = [state.validators[participant_index].pubkey for participant_index in participant_indices]
     domain = get_domain(state, DOMAIN_SYNC_COMMITTEE, compute_epoch_at_slot(previous_slot))
     signing_root = compute_signing_root(get_block_root_at_slot(state, previous_slot), domain)
-    assert bls.FastAggregateVerify(participant_pubkeys, signing_root, body.sync_committee_signature)
+    assert optional_fast_aggregate_verify(participant_pubkeys, signing_root, body.sync_committee_signature)
```
vbuterin, line 181: Using the word "optional" in a verification function in a spec that people are going to read is a bit misleading imo....

Is the goal of the `optional_fast_aggregate_verify` to allow the verification to be bypassed if we want to run tests without the BLS overhead? If so it may be worth not calling the function optional despite its role; intuitiveness to a reader trying to understand is more important than intuitiveness to someone directly working on the executable spec tooling.

specs/lightclient/beacon-chain.md
```diff
@@ -151,13 +172,13 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
 ```python
 def process_sync_committee(state: BeaconState, body: BeaconBlockBody) -> None:
     # Verify sync committee aggregate signature signing over the previous slot block root
-    previous_slot = max(state.slot, Slot(1)) - Slot(1)
+    previous_slot = max(Slot(state.slot), Slot(1)) - Slot(1)
```
vbuterin, line 175: Just curious, why the Slot wrapper around `state.slot`? Isn't `state.slot` of type Slot already?

specs/lightclient/beacon-chain.md
hwwhww: `optional_fast_aggregate_verify` is to allow `len(participant_pubkeys) == 0` case since it's illegal in IETF BLS draft standard APIs right now. The word "optional" may be too mypy-ish and not readable enough. 😅  Any naming suggestion?

@JustinDrake mentioned that it's possible that BLST may implement this feature at the API level. Then maybe we can import/refer to another document from the BLS standard region.

specs/lightclient/beacon-chain.md
hwwhww: It's a side effect of the new OOP SSZ (inheritance) implementation: https://github.com/ethereum/eth2.0-specs/pull/2141#discussion_r533627458

Statically, Mypy doesn't recognize the parent class fields when checking the child class object.

specs/phase0/validator.md
protolambda: 1. No, it's non-standard already. We can do anything with python, but can't make mypy interpret it all I'm afraid. I'm not sure how we could make it recognize the inherited fields correctly. Suggestions/PRs to remerkleable are welcome though.
2. I dislike explicit casting explicitly by wrapping with a constructor, since it comes at a performance cost. Maybe a typed variable definition like `foobar: MyTypeHere = foobar_obj_something` works to narrow down the type?

specs/lightclient/beacon-chain.md
vbuterin: Why not just call it `fast_aggregate_verify`?

It's the BLS standard that's implementing a weird quirky method with a special case, not us; we're just doing aggregate verification with the mathematically obvious generalization to the empty pubkey list case. So IMO it's totally okay for us to just call it fast aggregate verification.

JustinDrake: Agreed that `optional_` is awkward. One possibility is to call it `eth2_fast_aggregate_verify`. On the one hand it avoids confusion with the IETF `bls.FastAggregateVerify` and on the other hand the naming suggests it is an Eth2 variant.

hwwhww: Thank you both!

I'm inclined to `eth2_fast_aggregate_verify` to avoid confusion. 👍

hwwhww:
### Changelog

- As discussed during the call and with @djrtwo in chat, do not cancel the sync committee rewards when dealing inactivity leak.
- Rebased.

Now trying to fix the new error after rebasing:

```
eth2spec/test/phase0/sanity/test_blocks.py:319: in test_empty_epoch_transition
    signed_block = state_transition_and_sign_block(spec, state, block)
eth2spec/test/helpers/state.py:83: in state_transition_and_sign_block
    transition_unsigned_block(spec, state, block)
eth2spec/test/helpers/block.py:62: in transition_unsigned_block
    spec.process_block(state, block)
eth2spec/lightclient_patch/spec.py:1148: in process_block
    process_sync_committee(state, block)
eth2spec/lightclient_patch/spec.py:1825: in process_sync_committee
    committee_pubkeys = state.current_sync_committee.pubkeys
../../../venv/lib/python3.8/site-packages/remerkleable/complex.py:820: in __getattr__
    return super().get(i)
../../../venv/lib/python3.8/site-packages/remerkleable/subtree.py:29: in get
    self.get_backing().getter(to_gindex(i, self.tree_depth())), lambda v: self.set(i, v))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = 0x0000000000000000000000000000000000000000000000000000000000000000
target = 2

    def getter(self, target: Gindex) -> Node:
        if target != 1:
>           raise NavigationError
E           remerkleable.tree.NavigationError

../../../venv/lib/python3.8/site-packages/remerkleable/tree.py:292: NavigationError
```


I still could not reproduce a similar error case in writing `remerkleable` tests. Investigating if something is wrong on the spec side. 👀 

/cc @protolambda do you have any hints of why this error happened? I tried to comment out `raise NavigationError`, then some tests passed (haven't run all the tests locally with this version yet).

specs/phase0/validator.md
hwwhww: `foobar: MyTypeHere = foobar_obj_something` works! 👍 👍

specs/lightclient/beacon-chain.md
```diff
@@ -165,30 +186,59 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
 ```python
 def process_sync_committee(state: BeaconState, block: BeaconBlock) -> None:
     # Verify sync committee aggregate signature signing over the previous slot block root
-    previous_slot = Slot(max(state.slot, 1) - 1)
+    previous_slot = Slot(max(int(state.slot), 1) - 1)
```
hwwhww, line 189: @djrtwo 

This line was `previous_slot = max(state.slot, Slot(1)) - Slot(1)` earlier.

But since `max` needs to compare the same type, either we cast to `int` or cast to `int`.

protolambda:
@hwwhww it looks like it was accessing a field in a container, recognized the wrong count of container fields (probably related to new extended types), created a generalized-index that wasn't deep enough (e.g. 9 fields instead of 8 = subtree twice as deep) and then failed to retrieve the node at the generalized index that was too shallow. Interesting! I'll run through this and see if it's a usage problem or a new bug due to OOP changes in remerkleable.

specs/lightclient/beacon-chain.md
```diff
@@ -165,30 +186,59 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
-        reward = Gwei(base_reward * active_validator_count // len(committee_indices) // SLOTS_PER_EPOCH)
+        max_participant_reward = base_reward - base_reward // PROPOSER_REWARD_QUOTIENT
+        reward = Gwei(max_participant_reward * active_validator_count // len(committee_indices) // SLOTS_PER_EPOCH)
         increase_balance(state, participant_index, reward)
-        participant_rewards += reward
+        proposer_reward += base_reward // PROPOSER_REWARD_QUOTIENT
 
     # Reward beacon proposer
-    increase_balance(state, get_beacon_proposer_index(state), Gwei(participant_rewards // PROPOSER_REWARD_QUOTIENT))
+    increase_balance(state, get_beacon_proposer_index(state), proposer_reward)
```
hwwhww, line 209: Based on https://github.com/ethereum/eth2.0-specs/pull/2130/files#r532294580

protolambda:
Looked deeper into that navigation error: the state is backed by a tree that does not contain any nodes for `current_sync_committee` or `next_sync_committee`. For the first one, it happens to maintain a padding node to make it an even number (but the padding node has no child nodes, so you got a navigation error on access). The 2nd field doesn't have a node at all and will straight away result in a navigation error.

TLDR: need to look at state initialization now, see if it properly builds a tree suitable for use in the lighthclient-patch.

protolambda:
Update: in the `context.py` (literally creates context for tests) it creates the test state. As part of testing, it can upgrade states to later phases. It is not doing this for the light-client-patch, and just plugs the old incomplete state tree into the expected beacon-state-type.
This typing step is an optimization: whenever it can, test preparation re-uses the same immutable state data (based on fork name, fork version, and initialization arguments).
I'll see if I can make it do that upgrade step to get a proper state.

protolambda:
Attempt to fix it: the test state preparation now runs a new function `upgrade_to_lightclient_patch`, which essentially takes a phase0 state, and outputs a lightclient-patch state. Similar to how a phase1 state is built.

The `upgrade_to_lightclient_patch` is not pretty, since `remerkleable` does not support dict-like item retrieval yet. It could become been `BeaconState(**pre.items())`, or I can try make `FooBar(Foo(example=123), more=42) -> FooBar {example: 123, more: 42}` a thing for upgrades. The current `BeaconState.fields()` returns a dict of keys and field types, not values. Style suggestions welcome, we have a working placeholder now, so we have time to discuss.
Or we are more explicit about state upgrades (which IMHO is not a bad thing), and write it all out like with phase 1.

hwwhww:
@protolambda

Thanks a lot for the help!!! That makes perfect sense.

I'm okay with `BeaconState(**{k: getattr(pre, k) for k in pre.fields().keys()})` style, but I realized that we need to update the `state.fork` and `state.latest_block_header` explicitly so I added `lightclient-fork.md` as phase 1.

Still have some known issues that our tests didn't find, but finally see the green lights for all CI jobs! 🍏

specs/lightclient/lightclient-fork.md
```diff
@@ -0,0 +1,83 @@
...
+    post = BeaconState(
+        genesis_time=pre.genesis_time,
+        slot=pre.slot,
+        fork=Fork(
+            previous_version=pre.fork.current_version,
+            current_version=LIGHTCLIENT_PATCH_FORK_VERSION,
+            epoch=epoch,
+        ),
+        # History
+        latest_block_header=pre.latest_block_header,
```
hwwhww, line 53: @djrtwo @vbuterin 

Since `sync_committee_bits` and `sync_committee_signature` fields are added to block header instead of block body in https://github.com/ethereum/eth2.0-specs/pull/2130/commits/3b7c02514b9c855acdddbe2918bd0e02eeb02c9b, it would be more tricky to handle the SSZ objects at the fork boundary.

1. The `latest_block_header` field here *should* be transformed from `phase0.BeaconBlockBody` to `lightclient_patch.BeaconBlockBody`.
2. However, the next slot (the first slot after the light client patch) will have a problem when caching the `previous_block_root = hash_tree_root(state.latest_block_header)` in `process_slot` function because the previous root should have been computed with phase 0 form at phase 0.

I think it makes more sense to have `sync_committee_bits` and `sync_committee_signature` in `BeaconBlockBody` although they are not "operations".

specs/lightclient/lightclient-fork.md
```diff
@@ -0,0 +1,83 @@
...
+**Notice**: This document is a work-in-progress for researchers and implementers.
+
+## Table of contents
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Introduction](#introduction)
+- [Configuration](#configuration)
+- [Fork to Light-client patch](#fork-to-light-client-patch)
```
JustinDrake, line 12: We write "light client" in three different ways  😂 

1) light client
2) lightclient
3) light-client

I guess we want to be consistent. (Small preference for "light client" on my side.)

hwwhww: 👍 on "light client".

Once we have a better hard fork name, we can replace the folder name `lightclient` with it.
We can scan it again after merging this PR to avoid conflicts.

configs/mainnet/lightclient_patch.yaml
```diff
@@ -0,0 +1,10 @@
+# Mainnet preset - lightclient patch
+
+CONFIG_NAME: "minimal"
```
mpetrunic, line 3:
```suggestion
CONFIG_NAME: "mainnet"
```

configs/minimal/lightclient_patch.yaml
```diff
@@ -0,0 +1,8 @@
+# Minimal preset - lightclient patch
+
+CONFIG_NAME: "minimal"
+
+# ---------------------------------------------------------------
+
+SYNC_COMMITTEE_SIZE: 64
+SYNC_COMMITTEE_PUBKEY_AGGREGATES_SIZE: 16
```
mpetrunic, line 8: Are `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` and `DOMAIN_SYNC_COMMITTEE` suppose to be in config as well?

specs/lightclient/lightclient-fork.md
hwwhww: Fixed in https://github.com/ethereum/eth2.0-specs/pull/2141/commits/2c86b433ae13a59196ebb10ff9997eacf1af2939

hwwhww:
@vbuterin @djrtwo @JustinDrake @protolambda 

`upgrade_to_lightclient_patch` may still have some typing issues but it's not substantive. We can fix it later.

This PR is ready to be merged into upstream #2130.

PR #2146, opened 2020-12-09
DAS phase 1

vbuterin:
Puts data availability sampling based on the ideas [here](https://hackmd.io/@HWeNw8hNRimMm2m2GH56Cw/r1XzqYIOv) into a pull request.


specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +485,124 @@ def process_epoch(state: BeaconState) -> None:
...
-    """
-    Update light client committees.
-    """
-    next_epoch = compute_epoch_at_slot(Slot(state.slot + 1))
-    if next_epoch % LIGHT_CLIENT_COMMITTEE_PERIOD == 0:
-        state.current_light_committee = state.next_light_committee
-        new_committee = get_light_client_committee(state, next_epoch + LIGHT_CLIENT_COMMITTEE_PERIOD)
-        state.next_light_committee = committee_to_compact_committee(state, new_committee)
-```
+`process_reveal_deadlines`, `process_challenge_deadlines` and `process_custody_final_updates` are defined in [the Custody Game spec](./custody-game.md).
```
hwwhww, line 608: Note that some custody game SSZ fields are removed in this PR.

Do you intend to split the custody game into another fork? Or it's a temporary removal?

vbuterin: I'm inclined to say, first stabilize this document somewhat without caring too much about custody, and then after a few weeks start figuring out how to bring the custody games back?

specs/phase1/beacon-chain.md
```diff
@@ -166,246 +91,74 @@ class AttestationData(Container):
...
-
-### `ShardState`
-
-```python
-class ShardState(Container):
-    slot: Slot
-    gasprice: Gwei
-    latest_block_root: Root
+    # The actual data commitment
+    commitment: DataCommitment
```
dankrad, line 177: Are we now doing only a single commitment per block? I thought we would split in two if it runs over the 256 kB?

specs/phase1/beacon-chain.md
```diff
@@ -166,246 +91,74 @@ class AttestationData(Container):
...
+    commitment: BLSCommitment
+    # hash_tree_root of the ShardHeader (stored so that attestations
+    # can be checked against it)
+    root: Hash
+    # Length of the data in samples
+    length: uint64
+    # Who voted for the header
+    votes: Bitlist[MAX_COMMITTEE_SIZE]
+    # Has this header been confirmed?
+    confirmed: bool
```
dankrad, line 200: Missing length proof?

specs/phase1/beacon-chain.md
```diff
@@ -614,107 +275,26 @@ def get_start_shard(state: BeaconState, slot: Slot) -> Shard:
...
-    """
-    Return the offset slots of the given ``shard``.
-    The offset slot are after the latest slot and before current slot. 
-    """
-    return compute_offset_slots(get_latest_slot_for_shard(state, shard), state.slot)
+            shard = (shard + active_shard_count - committee_count) % active_shard_count
+    return Shard(shard)
 ```
 
 ### Predicates
```
dankrad, line 335:
```suggestion

```
Delete empty heading

specs/phase1/beacon-chain.md
```diff
@@ -166,246 +91,74 @@ class AttestationData(Container):
...
-    previous_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
-    current_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
-    # Finality
-    justification_bits: Bitvector[JUSTIFICATION_BITS_LENGTH]  # Bit set for every recent justified epoch
-    previous_justified_checkpoint: Checkpoint  # Previous epoch snapshot
-    current_justified_checkpoint: Checkpoint
-    finalized_checkpoint: Checkpoint
-    # Phase 1
+class BeaconState(phase0.BeaconState):
+    current_epoch_pending_headers: List[PendingHeader, MAX_PENDING_HEADERS * SLOTS_PER_EPOCH]
```
dankrad, line 148:
```suggestion
    current_epoch_pending_shard_headers: List[PendingShardHeader, MAX_PENDING_SHARD_HEADERS * SLOTS_PER_EPOCH]
```

specs/phase1/beacon-chain.md
```diff
@@ -166,246 +91,74 @@ class AttestationData(Container):
...
+    previous_epoch_pending_headers: List[PendingHeader, MAX_PENDING_HEADERS * SLOTS_PER_EPOCH]
```
dankrad, line 149:
```suggestion
    previous_epoch_pending_shard_headers: List[PendingShardHeader, MAX_PENDING_SHARD_HEADERS * SLOTS_PER_EPOCH]
```

specs/phase1/beacon-chain.md
```diff
@@ -10,152 +10,77 @@
...
-| `MAX_SHARDS` | `uint64(2**10)` (= 1024) |
-| `INITIAL_ACTIVE_SHARDS` | `uint64(2**6)` (= 64) |
-| `LIGHT_CLIENT_COMMITTEE_SIZE` | `uint64(2**7)` (= 128) |
-| `GASPRICE_ADJUSTMENT_COEFFICIENT` | `uint64(2**3)` (= 8) | 
+| Name | Value | Notes |
+| - | - | - |
+| `MAX_SHARDS` | `uint64(2**10)` (= 1024) | Theoretical max shard count (used to determine data structure sizes) |
+| `INITIAL_ACTIVE_SHARDS` | `uint64(2**6)` (= 64) | Initial shard count |
+| `GASPRICE_ADJUSTMENT_COEFFICIENT` | `uint64(2**3)` (= 2) | Gasprice may decrease/increase by at most exp(1 / this value) *per epoch* |
+| `MAX_SHARD_HEADERS` | `MAX_SHARDS * 4` | |
```
dankrad, line 77: `MAX_PENDING_SHARD_HEADERS` ?

specs/phase1/beacon-chain.md
vbuterin: The PendingShardHeaders don't need a length proof, because the proof is already verified when the shard header gets accepted.

specs/phase1/beacon-chain.md
vbuterin: I'm inclined to say keep the single commitment for simplicity for now? But I'm happy to discuss this more.

specs/phase1/beacon-chain.md
vbuterin: The maximum is on the number of shard header objects that can go into a beacon block, so imo `MAX_SHARD_HEADERS` is fine here.

Grubson33421 (approved):

specs/phase1/beacon-chain.md
```diff
@@ -12,150 +12,110 @@
...
 
 | Name | Value |
 | - | - |
-| `MAX_GASPRICE` | `Gwei(2**14)` (= 16,384) | Gwei | 
-| `MIN_GASPRICE` | `Gwei(2**3)` (= 8) | Gwei | 
+| `G2_ONE` | The G2 generator |
+| `SIZE_CHECK_POINTS` | Type `List[G2, MAX_SAMPLES_PER_BLOCK + 1]`; TO BE COMPUTED |
 
-### Initial values
+These points are the G2-side Kate commitments to `product[a in i...MAX_SAMPLES_PER_BLOCK] (X - w ** revbit(a))` for each `i` in `[0...MAX_SAMPLES_PER_BLOCK]`, where `w` is the root of unity and `revbit` is the reverse-bit-order function. They are used to verify block size proofs. They can be computed with a one-time O(N^2/log(N)) calculation using fast-linear-combinations in G2.
```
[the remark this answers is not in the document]
vbuterin: Yeah, agree there's probably some fancy O(N*log(N)) way to compute all these values; but we're totally fine even if such an algorithm is never discovered, so indeed not a super urgent problem.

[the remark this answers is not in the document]
vbuterin: Another thing worth mentioning is that the values can be *checked* easily if we put the pregenerated points into client code and add an additional EC point as a proof: use the hash of the list of points to generate a random point `z` and a random linear combination `(a1 ... an)`, and give the Kate proof for `(a1 * P1 + a2 * P2 + ... + an * Pn) // (X - z)`. Clients can then verify that the random linear combination actually does resolve at `z` to the value that they expect, which they can determine by evaluating the polynomials themselves at that point (in linear field ops by walking through the list end-to-front).

specs/phase1/beacon-chain.md
```diff
@@ -12,150 +12,111 @@
...
 
-| Name | Value |
-| - | - |
-| `NO_SIGNATURE` | `BLSSignature(b'\x00' * 96)` | 
+### Gwei values
+
+| Name | Value | Unit | Description |
+| - | - | - | - |
+| `MAX_GASPRICE` | `Gwei(2**24)` (= 16,777,216) | Gwei | Max gasprice charged for an TARGET-sized shard block |  
+| `MIN_GASPRICE` | `Gwei(2**3)` (= 8) | Gwei | Min gasprice charged for an TARGET-sized shard block |
```
dankrad, line 101: Denominating in GWei seems coarse. Given that it's a 64 bit integer, it would be easy to increase resolution to Wei. Have you checked the dynamics of this when the gas price is at the lower end of the scale? Gas might become unresponsive/take longer to adjust

specs/phase1/beacon-chain.md
```diff
@@ -752,298 +374,104 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+    all_candidates = [
+        c for c in pending_headers if
+        (c.slot, c.shard) == (pending_header.slot, pending_header.shard)
+    ]
+    if True not in [c.confirmed for c in all_candidates]:
+        # Requirement 2: >= 2/3 of balance attesting
+        participants = get_attesting_indices(state, data, pending_commitment.votes)
+        participants_balance = get_total_balance(state, participants)
+        full_committee = get_beacon_committee(state, data.slot, data.shard)
+        full_committee_balance = get_total_balance(state, full_committee)
```
dankrad, line 429: Should this be all, or only online?

specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +485,124 @@ def process_epoch(state: BeaconState) -> None:
...
+def charge_confirmed_header_fees(state: BeaconState) -> None:
+    new_gasprice = state.shard_gasprice
+    adjustment_quotient = get_active_shard_count(state) * SLOTS_PER_EPOCH * GASPRICE_ADJUSTMENT_COEFFICIENT
+    for slot in range(SLOTS_PER_EPOCH):
+        for shard in range(SHARD_COUNT):
+            confirmed_candidates = [
+                c for c in state.previous_epoch_pending_shard_headers if
+                (c.slot, c.shard, c.confirmed) == (slot, shard, True)
+            ]
+            if confirmed_candidates:
```
dankrad, line 548: Why should the proposer only be charged for confirmed blocks? If they are doing their job reasonably, then they should be able to get their proposal confirmed, so it would be fair to always charge them. After all, an unavailable block might cause as much, if not more work on attesters.

specs/phase1/beacon-chain.md
```diff
@@ -166,246 +127,78 @@ class AttestationData(Container):
...
-    # Finality
-    justification_bits: Bitvector[JUSTIFICATION_BITS_LENGTH]  # Bit set for every recent justified epoch
-    previous_justified_checkpoint: Checkpoint  # Previous epoch snapshot
-    current_justified_checkpoint: Checkpoint
-    finalized_checkpoint: Checkpoint
-    # Phase 1
+    # New fields
+    current_epoch_pending_shard_headers: List[PendingHeader, MAX_PENDING_HEADERS * SLOTS_PER_EPOCH]
+    previous_epoch_pending_shard_headers: List[PendingHeader, MAX_PENDING_HEADERS * SLOTS_PER_EPOCH]
+    most_recent_confirmed_commitments: Vector[Vector[DataCommitment, SLOTS_PER_EPOCH], MAX_SHARDS]
```
dankrad, line 151: What exactly does this track? It does not include all confirmed commitments (e.g. the ones confirmed through the fast confirmation mechanism are missing). Should it be the confirmed commitments from previous previous epoch? Should be renamed accordingly

specs/phase1/beacon-chain.md
vbuterin: It's not that coarse! Because it's not gwei per gas, it's gwei for the entire contents of the block. So the lowest possible gasprice is to literally pay (at current prices) $0.00000476 to fill an entire block.

specs/phase1/beacon-chain.md
vbuterin: Good question! Need to think through the consequences of this. The status quo in ethereum is certainly "you pay only if you get included", though I can see the arguments for changing it.

specs/phase1/beacon-chain.md
vbuterin: I'm inclined to say all. We already have a second case to cover the ">1/3 offline" case which is to wait until the end of the next epoch and take the candidate with the most votes (which could be empty).

specs/phase1/beacon-chain.md
vbuterin: It's the confirmed commitments fro the previous previous epoch. Do you have a suggestion for the best name for this? I struggled with naming here...

specs/phase1/beacon-chain.md
dankrad: I'm just saying that at the lower limit it may behave erratically. Maybe we want a higher lower limit then?

specs/phase1/beacon-chain.md
dankrad: grandfather_epoch_confirmed_commitments?

To be honest though, is this the right thing? Basically it means that a commitment confirmed now (through the fast commitment algorithm) will only be considered for the data availability fork choice two epochs later. I guess that's just in time to prevent finality, but why not consider it earlier (and prevent it from being finalized)?

After all, if we allow it being finalized, that means the chain is in a potentially unrecoverable state (without social intervention) if the data truly isn't available. So shouldn't we make it part of FCR immediately, so that validators will stop voting for that branch?

specs/phase1/beacon-chain.md
vbuterin: I just made a patch so that the gasprice delta is minimum 1, so it won't get stuck even at lower levels. Would that work? This is how the EIP 1559 proposal works on the eth1 side, and it seems a good idea to use similar formulas for both.

specs/phase1/beacon-chain.md
```diff
@@ -484,6 +485,10 @@ def process_shard_header(state: BeaconState,
     ))
 ```
 
+The length-and-degree proof works as follows. For a block B with length `l` (so `l` nonzero values in `[0...MAX_SAMPLES_PER_BLOCK-1]`), the length proof is supposed to be `(B / Z) * (r + X**(len(SETUP)-l))`, where `Z` is the minimal polynomial that is zero over `[l...MAX_SAMPLES_PER_BLOCK-1]` (see `SIZE_CHECK_POINTS` above). The goal is to ensure that a proof can only be constructed if (i) `B / Z` is itself non-fractional, meaning that `B` is a multiple of `Z`, and (ii) `deg(B) < MAX_SAMPLES_PER_BLOCK` (the block is not oversized).
+
+This is done by making the proof be a random linear combination of `B / Z` and `(B / Z) * (X**(len(SETUP)-l)`. The length proof will have the degree of `(B / Z) * X**(len(SETUP)-l)`, so `deg(B) - (MAX_SAMPLES_PER_BLOCK - l) + len(SETUP) - l`, simplified to `deg(B) - MAX_SAMPLES_PER_BLOCK + len(SETUP)`. Because it's only possible to generate proofs for polynomials with degree `< len(SETUP)`, it's this only possible to generate the proof if this expression is less than `len(SETUP)`, meaning that `deg(B)` must be strictly less than `MAX_SAMPLES_PER_BLOCK`.
```
dankrad, line 492: The current code does not implement the random linear combination. But I believe it's not necessary anyway: We can simply make the combined max degree length proof be `( B * X **(len(SETUP) - l) ) / Z`, which has a non-fractional solution only of `B` is `0` at the zeroes of `Z`.

specs/phase1/beacon-chain.md
dankrad: I think the main consequence is that we introduce a griefing vector by a committee, to conspire to not confirm a shard block in order to charge a proposer the gas fee without a reward.

Second possible problem could be that under condition where you consider it less likely that a block will be confirmed, e.g. when latency is high, it would discourage block production. In fact this threshold could be quite low: if we expect that a block producer makes 10% profit, then any confirmation probability under 90% would discourage block production.

specs/phase1/beacon-chain.md
dankrad: I agree that this should pretty safely fix the issue at the low end of the spectrum. Question about the upper limit though -- that seems fairly low, only 0.017 ETH per block. That could very plausibly be breached. Current Eth1 blocks have already brought in several ETH per block.

(I guess this would mean that we have to check that a proposer has enough balance before creating a block)

dankrad: My current suggestion is to allow up to ~8 ETH gas (meaning ~16 ETH at full size) and block validators with insufficient balance from proposer selection.

specs/phase1/beacon-chain.md
```diff
@@ -12,150 +12,115 @@
...
-| `ONLINE_PERIOD` | `OnlineEpochs(2**3)` (= 8) | online epochs | ~51 mins |
-| `LIGHT_CLIENT_COMMITTEE_PERIOD` | `Epoch(2**8)` (= 256) | epochs | ~27 hours |
+| `SHARD_COMMITTEE_PERIOD` | `Epoch(2**8)` (= 256) | epochs | ~27 hours |
 
 ### Domain types
 
 | Name | Value |
 | - | - |
-| `DOMAIN_SHARD_PROPOSAL` | `DomainType('0x80000000')` |
-| `DOMAIN_SHARD_COMMITTEE` | `DomainType('0x81000000')` |
```
terencechain, line 148: `DOMAIN_SHARD_COMMITTEE` is still used within `get_shard_committee`

specs/phase1/beacon-chain.md
```diff
@@ -588,22 +347,11 @@ def get_shard_proposer_index(beacon_state: BeaconState, slot: Slot, shard: Shard
     """
     epoch = compute_epoch_at_slot(slot)
     committee = get_shard_committee(beacon_state, epoch, shard)
-    seed = hash(get_seed(beacon_state, epoch, DOMAIN_SHARD_COMMITTEE) + uint_to_bytes(slot))
-    r = bytes_to_uint64(seed[:8])
-    return committee[r % len(committee)]
-```
-
-#### `get_committee_count_delta`
+    seed = hash(get_seed(beacon_state,
```
terencechain, line 350:
```suggestion
    seed = hash(get_seed(beacon_state, epoch, DOMAIN_SHARD_HEADER) + uint_to_bytes(slot))
```

seed is incomplete

specs/phase1/beacon-chain.md
```diff
@@ -166,252 +131,102 @@ class AttestationData(Container):
...
-    beacon_parent_root: Root
-    slot: Slot
-    shard: Shard
-    proposer_index: ValidatorIndex
-    body: ByteList[MAX_SHARD_BLOCK_SIZE]
+class DataCommitment(Container):
+    # Kate commitment to the data
+    point: BLSCommitment
+    # Length of the data in samples
+    length: length
```
terencechain, line 171:
```suggestion
    length: uint64
```

specs/phase1/beacon-chain.md
```diff
@@ -166,252 +131,102 @@ class AttestationData(Container):
...
+    # Slot and shard that this header is intended for
+    slot: uint64
     shard: Shard
-    proposer_index: ValidatorIndex
-    body_root: Root
+    # Kate commitment to the data
+    commitment: BLSCommitment
+    # hash_tree_root of the ShardHeader (stored so that attestations
+    # can be checked against it)
+    root: Hash
```
terencechain, line 199:
```suggestion
    root: Root
```

specs/phase1/beacon-chain.md
```diff
@@ -166,252 +131,102 @@ class AttestationData(Container):
...
+    # Length of the data in samples
+    length: uint64
+    # Who voted for the header
+    votes: Bitlist[MAX_COMMITTEE_SIZE]
```
terencechain, line 203: `MAX_COMMITTEE_SIZE` is not defined. Do you mean `MAX_VALIDATORS_PER_COMMITTEE` ?

specs/phase1/beacon-chain.md
```diff
@@ -166,252 +131,102 @@ class AttestationData(Container):
...
     current_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
-    # Finality
-    justification_bits: Bitvector[JUSTIFICATION_BITS_LENGTH]  # Bit set for every recent justified epoch
-    previous_justified_checkpoint: Checkpoint  # Previous epoch snapshot
-    current_justified_checkpoint: Checkpoint
-    finalized_checkpoint: Checkpoint
-    # Phase 1
+    # New fields
+    current_epoch_pending_shard_headers: List[PendingShardHeader, MAX_SHARD_HEADERS * SLOTS_PER_EPOCH]
+    previous_epoch_pending_shard_headers: List[PendingShardHeader, MAX_SHARD_HEADERS * SLOTS_PER_EPOCH]
```
terencechain, line 154:
```suggestion
    previous_epoch_pending_shard_headers: List[PendingShardHeader, MAX_SHARD_HEADERS * SLOTS_PER_EPOCH]
    current_epoch_pending_shard_headers: List[PendingShardHeader, MAX_SHARD_HEADERS * SLOTS_PER_EPOCH]
```

Does it make sense to flip it so it aligns in the same ordering as `previous_epoch_attestations` and `current_epoch_attestations` ?

specs/phase1/beacon-chain.md
```diff
@@ -752,298 +421,108 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+        pending_headers = state.current_epoch_pending_shard_headers
+    else:
+        pending_headers = state.previous_epoch_pending_shard_headers
+    # Create or update the PendingShardHeader object
+    pending_header = None
+    for header in pending_headers:
+        if header.root == attestation.data.shard_header_root:
+            pending_header = header
+    assert pending_header is not None
+    assert pending_header.slot == attestation.data.slot + 1
```
terencechain, line 454: Confused about this check. Why does the attestation have to be on-time? Shouldn't it be - 1 ?

specs/phase1/beacon-chain.md
```diff
@@ -752,298 +421,108 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
-        effective_balance=effective_balance,
-        next_custody_secret_to_reveal=next_custody_secret_to_reveal,
-        all_custody_secrets_revealed_epoch=FAR_FUTURE_EPOCH,
+    # Check if the PendingShardHeader is eligible for expedited confirmation
+    # Requirement 1: nothing else confirmed
+    all_candidates = [
+        c for c in pending_headers if
+        (c.slot, c.shard) == (pending_header.slot, pending_header.shard)
+    ]
+    if True not in [c.confirmed for c in all_candidates]:
```
terencechain, line 471: Is it necessary to check every pending header on per slot and shard basis?

The only candidate that's in concern here is the `pending_header` from the above

specs/phase1/beacon-chain.md
```diff
@@ -752,298 +421,108 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+        # Requirement 2: >= 2/3 of balance attesting
+        participants = get_attesting_indices(state, attestationg.data, pending_commitment.votes)
```
terencechain, line 473:
```suggestion
        participants = get_attesting_indices(state, attestation.data, pending_commitment.votes)
```

specs/phase1/beacon-chain.md
```diff
@@ -752,298 +421,108 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+        participants_balance = get_total_balance(state, participants)
+        full_committee = get_beacon_committee(state, attestationg.data.slot, attestationg.data.shard)
```
terencechain, line 475:
```suggestion
        full_committee = get_beacon_committee(state, attestation.data.slot, attestation.data.index)
```

specs/phase1/beacon-chain.md
```diff
@@ -752,298 +421,108 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+    if compute_epoch_at_slot(header.slot) == get_current_epoch(state):
+        pending_headers = state.current_epoch_pending_shard_headers
+    else:
+        pending_headers = state.previous_epoch_pending_shard_headers
+        
+    # Check that this header is not yet in the pending list
+    for pending_header in pending_headers:
+        assert header_root != pending_header.root
+    # Include it in the pending list
+    committee_length = len(get_beacon_committee(state, header.slot, header.shard))
```
terencechain, line 510: `get_beacon_committee` uses committee index not shard, we might need a slot/shard -> index helper...

specs/phase1/beacon-chain.md
```diff
@@ -752,298 +421,108 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+    pending_headers.append(PendingShardHeader(
+        slot=header.slot,
+        shard=header.shard,
+        commitment=header.commitment,
```
terencechain, line 514:
```suggestion
        commitment=header.commitment.point,
        length=header.commitment.length,
```

I think we forgot to add the length

specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +536,124 @@ def process_epoch(state: BeaconState) -> None:
...
     process_phase_1_final_updates(state)
 ```
 
+#### Pending headers
+
+```python
+
+def process_pending_headers(state: BeaconState):
+    for slot in range(SLOTS_PER_EPOCH):
+        for shard in range(SHARD_COUNT):
```
terencechain, line 552: `SHARD_COUNT` is not defined, just use `get_active_shard_count`?

specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +536,124 @@ def process_epoch(state: BeaconState) -> None:
...
+            # Pending headers for this (slot, shard) combo
+            candidates = [
+                c for c in state.previous_epoch_pending_shard_headers if
+                (c.slot, c.shard) == (slot, shard)
+            ]
+            if True not in [c.confirmed for c in candidates]:
+                # The entire committee (and its balance)
+                full_committee = get_beacon_committee(state, slot, shard)
```
terencechain, line 560: `full_committee` and `full_committee_balance` computations can be outside of the loop

specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +536,124 @@ def process_epoch(state: BeaconState) -> None:
...
+            state.most_recent_confirmed_commitments[shard][slot] = DataCommitment()
+    for c in state.previous_epoch_pending_shard_headers:
+        if c.confirmed:
+            state.most_recent_confirmed_commitments[c.shard][c.slot % SLOTS_PER_EPOCH] = c.commitment
+```            
+
+```python
+def charge_confirmed_header_fees(state: BeaconState) -> None:
+    new_gasprice = state.shard_gasprice
+    adjustment_quotient = get_active_shard_count(state) * SLOTS_PER_EPOCH * GASPRICE_ADJUSTMENT_COEFFICIENT
```
terencechain, line 592:
```suggestion
    adjustment_quotient = get_active_shard_count(state, get_current_epoch(state)) * SLOTS_PER_EPOCH * GASPRICE_ADJUSTMENT_COEFFICIENT
```

specs/phase1/beacon-chain.md
```diff
@@ -12,150 +12,115 @@
...
 
-### Gwei values
+### Precomputed size verification points
 
 | Name | Value |
 | - | - |
-| `MAX_GASPRICE` | `Gwei(2**14)` (= 16,384) | Gwei | 
-| `MIN_GASPRICE` | `Gwei(2**3)` (= 8) | Gwei | 
+| `G2_SETUP` | Type `List[G2]`. The G2-side trusted setup `[G, G*s, G*s**2....]`; note that the first point is the generator. |
+| `ROOT_OF_UNITY` | `pow(PRIMITIVE_ROOT_OF_UNITY, (MODULUS - 1) // (MAX_SAMPLES_PER_BLOCK * POINTS_PER_SAMPLE, MODULUS)` | |
```
rkapka, line 95: `MODULUS` has not been defined anywhere.

rkapka, line 95: `MODULUS` is not defined anywhere.

specs/phase1/beacon-chain.md
```diff
@@ -12,150 +12,115 @@
...
 
 ## Updated containers
 
 The following containers have updated definitions in Phase 1.
 
-### Extended `AttestationData`
+### `AttestationData`
 
 ```python
 class AttestationData(Container):
```
rkapka, line 126: Shouldn't this be defined as `AttestationData(phase0.AttestationData)` for consistency with other containers in this section?

specs/phase1/beacon-chain.md
```diff
@@ -166,252 +131,102 @@ class AttestationData(Container):
...
-    balances: List[Gwei, VALIDATOR_REGISTRY_LIMIT]
-    # Randomness
-    randao_mixes: Vector[Root, EPOCHS_PER_HISTORICAL_VECTOR]
-    # Slashings
-    slashings: Vector[Gwei, EPOCHS_PER_SLASHINGS_VECTOR]  # Per-epoch sums of slashed effective balances
-    # Attestations
+class BeaconState(phase0.BeaconState):
+    # Updated fields
     previous_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
     current_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
```
rkapka, line 151: These fields are the same as in https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#beaconstate

terencechain: `AttestationData` in `PendingAttestation` changed, which implies these fields changed

specs/phase1/beacon-chain.md
```diff
@@ -12,150 +12,115 @@
...
-| `MAX_SHARDS` | `uint64(2**10)` (= 1024) |
-| `INITIAL_ACTIVE_SHARDS` | `uint64(2**6)` (= 64) |
-| `LIGHT_CLIENT_COMMITTEE_SIZE` | `uint64(2**7)` (= 128) |
-| `GASPRICE_ADJUSTMENT_COEFFICIENT` | `uint64(2**3)` (= 8) | 
+| Name | Value | Notes |
+| - | - | - |
+| `MAX_SHARDS` | `uint64(2**10)` (= 1024) | Theoretical max shard count (used to determine data structure sizes) |
+| `INITIAL_ACTIVE_SHARDS` | `uint64(2**6)` (= 64) | Initial shard count |
+| `GASPRICE_ADJUSTMENT_COEFFICIENT` | `uint64(2**3)` (= 8) | Gasprice may decrease/increase by at most exp(1 / this value) *per epoch* |
+| `MAX_SHARD_HEADERS_PER_SHARD` | `4` | |
```
rkapka, line 76: Where does this value come from? Is this correlated with another part of the system?

specs/phase1/beacon-chain.md
rkapka: `grandparent` would be probably more in line with `parent`

specs/phase1/beacon-chain.md
```diff
@@ -166,252 +131,102 @@ class AttestationData(Container):
...
+### `PendingShardHeader`
 
 ```python
-class ShardBlockHeader(Container):
-    shard_parent_root: Root
-    beacon_parent_root: Root
-    slot: Slot
+class PendingShardHeader(Container):
+    # Slot and shard that this header is intended for
+    slot: uint64
```
rkapka, line 193:
```suggestion
    slot: Slot
```

specs/phase1/beacon-chain.md
```diff
@@ -12,150 +12,115 @@
...
-    - [Online-tracking](#online-tracking)
-    - [Light client committee updates](#light-client-committee-updates)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 ## Introduction
 
-This document describes the extensions made to the Phase 0 design of The Beacon Chain
- to facilitate the new shards as part of Phase 1 of Eth2.
+This document describes the extensions made to the Phase 0 design of The Beacon Chain to support data sharding, based on the ideas [here](https://hackmd.io/@HWeNw8hNRimMm2m2GH56Cw/r1XzqYIOv) and more broadly [here](https://arxiv.org/abs/1809.09044), using Kate commitments to commit to data to remove any need for fraud proofs (and hence, safety-critical synchrony assumptions) in the design.
```
rkapka, line 56:
```suggestion
This document describes the extensions made to the Phase 0 design of The Beacon Chain to support data sharding, based on the ideas [here](https://hackmd.io/@HWeNw8hNRimMm2m2GH56Cw/B1YJPGkpD) and more broadly [here](https://arxiv.org/abs/1809.09044), using Kate commitments to commit to data to remove any need for fraud proofs (and hence, safety-critical synchrony assumptions) in the design.
```

specs/phase1/beacon-chain.md
dankrad: So on the idea of getting rid of the zero check alltogether, and making it a pure degree check: I've got some python code here that interpolates a polynomial from only evaluations on a part of a subgroup:
https://github.com/ethereum/research/blob/master/polynomial_reconstruction/interpolate_polynomial_without_zeroes.py
Excluding the part that can be precomputed, it takes about 4s, which I'd reckon can be brought down to 0.4s in a well optimized implementation. So it feels like that's possible and I would think we should just do it?
This will lead to some savings in the proof computation, so we may well end up even or with some small savings. But we get a much nicer code for length verification which I think is worth it.
Also a nice effect is that the samples become more effective (=better coding rate) for smaller blocks.

specs/phase1/beacon-chain.md
dankrad: It's the curve order of BLS12_381

specs/phase1/beacon-chain.md
dankrad: This one is a bit tricker, right? Because we would need the shard count at the time the candidate was proposed?
@vbuterin doe we have to pull the adjustment_quotient inside the loop?

specs/phase1/beacon-chain.md
dankrad: [I implemented this as pure degree proofs for now, suggest we discuss the merits in this thread]

specs/phase1/beacon-chain.md
dankrad:
```suggestion
    assert pending_header.slot == attestation.data.slot
```

Shouldn't it even be simply `pending_header.slot == attestation.data.slot`?

specs/phase1/beacon-chain.md
dankrad: There could be a second header for the same `(shard, slot)` pair if the proposer equivocated.

specs/phase1/beacon-chain.md
dankrad: We actually don't need the length as part of the object, if the `.commitment` property is a `DataCommitment`

specs/phase1/beacon-chain.md
dankrad: The index is the same as the shard here. But I think we are using index for two different meanings in `get_beacon_committee` and `compute_committee`? Maybe we should rename one of these

specs/phase1/beacon-chain.md
```diff
@@ -752,298 +417,108 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+        full_committee_balance = get_total_balance(state, full_committee)
+        if participants_balance * 3 > full_committee_balance * 2:
+            pending_header.confirmed = True
+```
+
+#### `process_shard_header`
+
+```python
+def process_shard_header(state: BeaconState,
+                         signed_header: Signed[ShardDataHeader]) -> None:
```
rkapka, line 481:
```suggestion
                         signed_header: Signed[ShardHeader]) -> None:
```

specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +532,124 @@ def process_epoch(state: BeaconState) -> None:
     process_reveal_deadlines(state)
     process_challenge_deadlines(state)
     process_slashings(state)
+    process_pending_headers(state)
+    charge_confirmed_header_fees(state)
+    reset_pending_headers(state)
```
sauliusgrigaitis, line 537: Is there a particular reason to process pending headers during epoch transition?

Is it assumed that there will be a relatively low number of pending headers?

Probably a premature optimization, but maybe it makes sense to process pending headers more often during the epoch instead of already quite busy epoch processing.

specs/phase1/beacon-chain.md
terencechain: Got it. This will fail in most client implementations due to custom type assertion

specs/phase1/beacon-chain.md
vbuterin: Technically, they already _are_ being processed, in that if a header gets to 2/3 during an epoch it gets confirmed immediately. The end-of-epoch processing is just the case where no header gets to 2/3. I suppose if we want, we could put it into per-slot processing with a delay of 32 slots, but that may complicate the code more....

sauliusgrigaitis: I would say that the main question is do we want to offload pending headers processing from epoch processing. If yes, then not so complicated code change would be to move pending headers processing from the epoch processing to let's say a slot close to the beginning of an epoch. Then there would not be a need to complicate the code with a moving window approach. 

My main argument now would be that here we are dealing with shards (not the main beacon chain) that could become pretty chaotic during turbulent periods. But I agree that it's too early to complicate the code here until we have some real numbers.

specs/phase1/beacon-chain.md
```diff
@@ -557,26 +308,30 @@ def get_shard_committee(beacon_state: BeaconState, epoch: Epoch, shard: Shard) -
         indices=active_validator_indices,
         seed=seed,
         index=shard,
-        count=get_active_shard_count(beacon_state),
+        count=get_active_shard_count(beacon_state, epoch),
```
rkapka, line 311: Not directly related to changes in this PR, but there are a few functions where the parameter is called `beacon_state` and we usually use `state`. Maybe worth changing for consistency?

specs/phase1/beacon-chain.md
```diff
@@ -557,26 +308,30 @@ def get_shard_committee(beacon_state: BeaconState, epoch: Epoch, shard: Shard) -
...
     )
 ```
 
-#### `get_light_client_committee`
+#### `compute_proposer_index`
```
rkapka, line 315: `get_committee_count_per_slot` is prefixed with `Updated` - it should be the same here,

specs/phase1/beacon-chain.md
```diff
@@ -166,252 +130,99 @@ class AttestationData(Container):
...
-    justification_bits: Bitvector[JUSTIFICATION_BITS_LENGTH]  # Bit set for every recent justified epoch
-    previous_justified_checkpoint: Checkpoint  # Previous epoch snapshot
-    current_justified_checkpoint: Checkpoint
-    finalized_checkpoint: Checkpoint
-    # Phase 1
+    # New fields
+    previous_epoch_pending_shard_headers: List[PendingShardHeader, MAX_SHARD_HEADERS * SLOTS_PER_EPOCH]
+    current_epoch_pending_shard_headers: List[PendingShardHeader, MAX_SHARD_HEADERS * SLOTS_PER_EPOCH]
+    grandparent_epoch_confirmed_commitments: Vector[Vector[DataCommitment, SLOTS_PER_EPOCH], MAX_SHARDS]
+    shard_gasprice: uint64
```
rkapka, line 155: `shard_header_gasprice`?

specs/phase1/beacon-chain.md
```diff
@@ -614,106 +358,24 @@ def get_start_shard(state: BeaconState, slot: Slot) -> Shard:
...
-    if current_epoch_start_slot == slot:
-        return state.current_epoch_start_shard
-    elif slot > current_epoch_start_slot:
+    shard = state.current_epoch_start_shard
+    if slot > current_epoch_start_slot:
         # Current epoch or the next epoch lookahead
-        shard_delta = get_committee_count_delta(state, start_slot=current_epoch_start_slot, stop_slot=slot)
-        return Shard((state.current_epoch_start_shard + shard_delta) % active_shard_count)
-    else:
+        for _slot in range(current_epoch_start_slot, slot):
```
rkapka, line 364:
```suggestion
        for _slot in range(current_epoch_start_slot, slot - 1):
```
Otherwise we will have the start shard at `slot + 1`.

specs/phase1/beacon-chain.md
```diff
@@ -614,106 +358,24 @@ def get_start_shard(state: BeaconState, slot: Slot) -> Shard:
...
+            active_shard_count = get_active_shard_count(state, compute_epoch_at_slot(Slot(_slot)))
+            shard = (shard + committee_count) % active_shard_count
+        return Shard(shard)
+    elif slot < current_epoch_start_slot:
         # Previous epoch
-        shard_delta = get_committee_count_delta(state, start_slot=slot, stop_slot=current_epoch_start_slot)
-        max_committees_per_slot = active_shard_count
-        max_committees_in_span = max_committees_per_slot * (current_epoch_start_slot - slot)
-        return Shard(
+        for _slot in list(range(slot, current_epoch_start_slot))[::-1]:
```
rkapka, line 371:
```suggestion
        for _slot in list(range(slot - 1, current_epoch_start_slot))[::-1]:
```
Otherwise we will have the start shard at `slot + 1`.

specs/phase1/beacon-chain.md
sauliusgrigaitis: Just to clarify, by "pending headers" I mean the headers that are still not processed (not received 2/3 votes) at the end-of-epoch. It will be interesting to see how much of those actually piles up and what actual overhead it brings to the epoch processing.

specs/phase1/beacon-chain.md
```diff
@@ -752,298 +417,108 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
-        data=attestation.data,
-        inclusion_delay=state.slot - attestation.data.slot,
-        proposer_index=get_beacon_proposer_index(state),
-        crosslink_success=False,  # To be filled in during process_shard_transitions
+    phase0.process_attestation(state, attestation)
+    update_pending_votes(
+        state=state,
+        attestation: Attestation,
+        root=,
+        aggregation_bits=attestation.aggregation_bits
```
rkapka, line 431:
```suggestion
        state=state,
        attestation: Attestation,
```

specs/phase1/beacon-chain.md
```diff
@@ -752,298 +417,108 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
     )
```
rkapka, line 432:
```suggestion
    update_pending_votes(state, attestation)
```

specs/phase1/beacon-chain.md
```diff
@@ -752,298 +417,108 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
-        all_custody_secrets_revealed_epoch=FAR_FUTURE_EPOCH,
+    # Check if the PendingShardHeader is eligible for expedited confirmation
+    # Requirement 1: nothing else confirmed
+    all_candidates = [
+        c for c in pending_headers if
+        (c.slot, c.shard) == (pending_header.slot, pending_header.shard)
+    ]
+    if True not in [c.confirmed for c in all_candidates]:
+        # Requirement 2: >= 2/3 of balance attesting
+        participants = get_attesting_indices(state, attestationg.data, pending_commitment.votes)
```
rkapka, line 469:
```suggestion
        participants = get_attesting_indices(state, attestationg.data, pending_header.votes)
```

specs/phase1/beacon-chain.md
rkapka:
```suggestion
        participants = get_attesting_indices(state, attestation.data, pending_header.votes)
```

specs/phase1/beacon-chain.md
```diff
@@ -743,6 +405,9 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
     for_ops(body.attestations, process_attestation)
     for_ops(body.deposits, process_deposit)
     for_ops(body.voluntary_exits, process_voluntary_exit)
+    # Limit is dynamic based on active shard count
+    assert len(body.shard_headers) <= MAX_SHARD_HEADERS_PER_SHARD * get_active_shard_count(state, get_current_epoch(state))
```
rkapka, line 409: This doesn't protect from a shard having more headers then `MAX_SHARD_HEADERS`. If one shard has 5 and another has 3, they have 4 on average, which will pass the assertion.

specs/phase1/beacon-chain.md
```diff
@@ -752,298 +417,108 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+        assert header.degree_proof == G1_SETUP[0]
+    assert (
+        bls.Pairing(header.degree_proof, G2_SETUP[0]) ==
+        bls.Pairing(header.commitment.point, G2_SETUP[-header.commitment.length]))
     )
+    # Get the correct pending header list
+    if compute_epoch_at_slot(header.slot) == get_current_epoch(state):
+        pending_headers = state.current_epoch_pending_shard_headers
+    else:
+        pending_headers = state.previous_epoch_pending_shard_headers
```
rkapka, line 502: From https://hackmd.io/@HWeNw8hNRimMm2m2GH56Cw/sharding_proposal:
> The ShardHeader can be included in the beacon chain in that slot or any subsequent slot in the same or next epoch.

The `else` branch should verify that the header's slot is not earlier than the previous epoch. Alternatively, we can have a similar check to the one in `process_attestation':
```
assert data.target.epoch in (get_previous_epoch(state), get_current_epoch(state))
```

specs/phase1/beacon-chain.md
vbuterin: We need at least 2 per shard to cover the normal case (the header proposed by the proposer, and the null header); 4 leaves some room to handle malicious multi-proposals as well. No other reasoning than that.

specs/phase1/beacon-chain.md
```diff
@@ -12,150 +12,114 @@
...
 
 | Name | Value |
 | - | - |
-| `DOMAIN_SHARD_PROPOSAL` | `DomainType('0x80000000')` |
-| `DOMAIN_SHARD_COMMITTEE` | `DomainType('0x81000000')` |
-| `DOMAIN_LIGHT_CLIENT` | `DomainType('0x82000000')` |
-| `DOMAIN_CUSTODY_BIT_SLASHING` | `DomainType('0x83000000')` |
-| `DOMAIN_LIGHT_SELECTION_PROOF` | `DomainType('0x84000000')` |
-| `DOMAIN_LIGHT_AGGREGATE_AND_PROOF` | `DomainType('0x85000000')` |
+| `DOMAIN_SHARD_HEADER` | `DomainType('0x80000000')` |
```
vbuterin, line 116:
```suggestion
| `DOMAIN_SHARD_HEADER` | `DomainType('0x80000000')` |
| `DOMAIN_SHARD_COMMITTEE` | `DomainType('0x81000000')` |
```

specs/phase1/beacon-chain.md
vbuterin: I agree with consistency, but am not sure which direction the consistency should go! Maybe we want to use the inheritance format, maybe we don't. So I'm inclined to leave as is for now...

specs/phase1/beacon-chain.md
vbuterin: The last slot in `range(current_epoch_start_slot, slot)` _would be_ `slot-1`; that's how ranges in python work.

specs/phase1/beacon-chain.md
vbuterin: Also don't think this makes sense. Example: support `slot` is 95, `current_epoch_start_slot` is 100. `state.current_epoch_start_shard` represents the start shard pointer _at the start of_ slot 100. We want the start shard pointer _at the start of_ slot 95. So we want to process backwards, subtracting the shard counts for slots 99, 98, 97, 96, 95. The existing logic does this.

specs/phase1/beacon-chain.md
vbuterin: What's wrong with allowing one shard to have 5 and another 3? The normal case is 2 per shard; if there's an unexpected level of mischief (eg. a proposer making 10 malicious proposals and getting some attestations for each of them) on one shard, then why shouldn't all the slack be able to be directed toward that shard?

specs/phase1/beacon-chain.md
vbuterin, line 502:
```suggestion
        assert compute_epoch_at_slot(header.slot) == get_previous_epoch(state):
        pending_headers = state.previous_epoch_pending_shard_headers
```

specs/phase1/beacon-chain.md
vbuterin: Hmm. I would be inclined not to worry too much, especially since HF1 is reducing end-of-epoch work quite a bit anyway, and the amount of work being done here is significantly less than the rest of attestation processing.

specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +532,124 @@ def process_epoch(state: BeaconState) -> None:
...
     process_phase_1_final_updates(state)
 ```
 
+#### Pending headers
+
+```python
+
+def process_pending_headers(state: BeaconState):
+    for slot in range(SLOTS_PER_EPOCH):
+        for shard in range(SHARD_COUNT):
```
vbuterin, line 548:
```suggestion
        for shard in range(get_active_shard_count(state)):
```

specs/phase1/beacon-chain.md
rkapka: You are correct. I misunderstood how Python ranges work.

specs/phase1/beacon-chain.md
rkapka: I guess my confusion was caused by the name of the constant. To me it implies that it's the maximum number of headers a shard can have. I am not sure what would be a better name though...

specs/phase1/beacon-chain.md
dankrad: I think one annoying property is that a single shard proposer could fill all the proposals; thus an attacker only has to equivocate with one validator to fill up all proposals for one epoch.
Would it make sense to put a per-shard and per-height limit on this as well?

specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +538,124 @@ def process_epoch(state: BeaconState) -> None:
...
+                    for voters in voting_sets
+                ]
+                # Get the index with the most total balance voting for them.
+                # NOTE: if two choices get exactly the same voting balance,
+                # the candidate earlier in the list wins
+                if max(voting_balances) > 0:
+                    winning_index = voting_balances.index(max(voting_balances))
+                else:
+                    # If no votes, zero wins
+                    winning_index = [c.root for c in candidates].index(Root())
```
rkapka, line 581: Should we also cover the case when no proposals were made for a slot, hence no such index will be found?

specs/phase1/beacon-chain.md
rkapka: Charging for unconfirmed blocks would make equivocation to fill up proposals more expensive (https://github.com/ethereum/eth2.0-specs/pull/2146#discussion_r550452269)

specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +538,124 @@ def process_epoch(state: BeaconState) -> None:
...
+                    adjustment_quotient
+                )
+    state.shard_gasprice = new_gasprice
+```
+
+```python
+def reset_pending_headers(state: BeaconState):
+    state.previous_epoch_pending_shard_headers = state.current_epoch_pending_shard_headers
+    shards = [
+        compute_shard_from_committee_index(state, slot, index)
```
rkapka, line 622: `index` is undefined.

specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +538,124 @@ def process_epoch(state: BeaconState) -> None:
...
+        for i in range()
+        state,
+        attestation.data.index,
+        attestation.data.slot
+    )
```
rkapka, line 627:
```suggestion
```

specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +538,124 @@ def process_epoch(state: BeaconState) -> None:
...
+    state.current_epoch_pending_shard_headers = []
+    # Add dummy "empty" PendingAttestations
+    # (default to vote for if no shard header available)
+    for slot in range(SLOTS_IN_EPOCH):
+        for index in range(get_committee_count_per_slot(get_current_epoch(state))):
+            shard = compute_shard_from_committee_index(state, slot, index)
+            committee_length = len(get_beacon_committee(
+                state,
+                header.slot,
+                header.shard
```
rkapka, line 637:
```suggestion
                slot,
                shard
```

specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +538,124 @@ def process_epoch(state: BeaconState) -> None:
...
+        for shard in range(get_active_shard_count(state)):
+            # Pending headers for this (slot, shard) combo
+            candidates = [
+                c for c in state.previous_epoch_pending_shard_headers if
+                (c.slot, c.shard) == (slot, shard)
+            ]
+            # The entire committee (and its balance)
+            full_committee = get_beacon_committee(state, slot, shard)
+            full_committee_balance = get_total_balance(state, full_committee)
+            if True not in [c.confirmed for c in candidates]:
```
rkapka, line 563: I think it would make sense to include an assertion that the list is not empty because this is implied when setting the value of `winning_index`.

specs/phase1/beacon-chain.md
```diff
@@ -166,252 +131,99 @@ class AttestationData(Container):
...
-    # Custody game
-    # next_custody_secret_to_reveal is initialised to the custody period
-    # (of the particular validator) in which the validator is activated
-    # = get_custody_period_for_validator(...)
-    next_custody_secret_to_reveal: uint64
-    # TODO: The max_reveal_lateness doesn't really make sense anymore.
-    # So how do we incentivise early custody key reveals now?
-    all_custody_secrets_revealed_epoch: Epoch  # to be initialized to FAR_FUTURE_EPOCH
+class BeaconBlock(phase0.BeaconBlock):
+    shard_headers: List[Signed[ShardHeader], MAX_SHARD_HEADERS]
```
rkapka, line 142: A few comments here:

1. `shard_headers` is not set anywhere, thus `assert len(body.shard_headers) <= (...)` will always fail.
2. Since `BeaconBlockBody` contains a list of attestations, should this field even exist? Shouldn't `BeaconBlockBody.attestations` hold headers within attestations themselves?
3. Even if we keep `shard_headers`, shouldn't the size of `BeaconBlockBody.attestations` be increased to cater for shards? Currently `MAX_ATTESTATIONS =  128`

specs/phase1/beacon-chain.md
```diff
@@ -12,150 +12,115 @@
...
+### Precomputed size verification points
 
 | Name | Value |
 | - | - |
-| `MAX_GASPRICE` | `Gwei(2**14)` (= 16,384) | Gwei | 
-| `MIN_GASPRICE` | `Gwei(2**3)` (= 8) | Gwei | 
+| `G1_SETUP` | Type `List[G1]`. The G1-side trusted setup `[G, G*s, G*s**2....]`; note that the first point is the generator. |
+| `G2_SETUP` | Type `List[G2]`. The G2-side trusted setup `[G, G*s, G*s**2....]` |
+| `MODULUS` | `0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001` (curve order of BLS12_381) |
+| `ROOT_OF_UNITY` | `pow(PRIMITIVE_ROOT_OF_UNITY, (MODULUS - 1) // (MAX_SAMPLES_PER_BLOCK * POINTS_PER_SAMPLE, MODULUS)` |
```
rkapka, line 97: `PRIMITIVE_ROOT_OF_UNITY`, `MODULUS` and `POINTS_PER_SAMPLE` are used only in the definition of `ROOT_OF_UNITY`, which itself is not used anywhere.

AuraSophia:
Vbuterin really belive in this coin!!!!

specs/phase1/beacon-chain.md
sauliusgrigaitis: I guess we can just wait until we see the real numbers. Testnets (usually with way lower participation rates comparing to the mainnet) should show is there any real need to move pending headers processing from the end-of-epoch.

But I would like to see a more general rule that processing is spread out uniformly across entire epoch, especially for low hanging fruits such as offloading processing from the end-of-epoch if it doesn't need to be processed at the end-of-epoch. This would help client teams to deliver faster as this approach should be less demanding for optimizations. This would also help to avoid hard forks or will make it simpler because seems that accumulating processing at specific points (like end-of-epoch) sooner or later needs a hard fork.

djrtwo (commented):
made a round of cleanups and minor fixes  in https://github.com/ethereum/eth2.0-specs/pull/2146/commits/b3c3c0a6bd26cf5cea5267497fe63509b009e9e6

specs/phase1/beacon-chain.md
```diff
@@ -739,311 +433,117 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+        shard=header.shard,
+        commitment=header.commitment,
+        root=header_root,
+        votes=Bitlist[MAX_VALIDATORS_PER_COMMITTEE]([0] * committee_length),
+        confirmed=False,
+    ))
 ```
 
-#### Light client processing
+The degree proof works as follows. For a block `B` with length `l` (so `l`  values in `[0...l - 1]`, seen as a polynomial `B(X)` which takes these values), the length proof is the commitment to the polynomial `B(X) * X**(MAX_DEGREE + 1 - l)`, where `MAX_DEGREE` the the maximum power of `s` available in the setup, which is `MAX_DEGREE = len(G2_SETUP) - 1`. The goal is to ensure that a proof can only be constructed if `deg(B) < l` (there are not hidden higher-order terms in the polynomial, which would thwart reconstruction).
```
lsankar4033, line 544: possible typo:
"where `MAX_DEGREE` the the maximum power of `s`" -> "where `MAX_DEGREE` is the maximum power of `s`"

specs/phase1/beacon-chain.md
```diff
@@ -739,311 +433,117 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
-            decrease_balance(state, proposer_index, shard_state.gasprice * length)
-
-        # Return winning transition root
-        return shard_transition_root
-
-    # No winning transition root, ensure empty and return empty root
-    assert shard_transition == ShardTransition()
-    return Root()
+    # Requirement 2: >= 2/3 of balance attesting
+    participants = get_attesting_indices(state, attestation.data, pending_commitment.votes)
```
nrryuya, line 492: > `pending_commitment.votes` 

Shouldn't this be `pending_header.votes` ?

djrtwo: yes! thank you
fixed

specs/phase1/beacon-chain.md
djrtwo: yep, thank you!

specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +557,129 @@ def process_epoch(state: BeaconState) -> None:
...
+```python
+def charge_confirmed_header_fees(state: BeaconState) -> None:
+    new_gasprice = state.shard_gasprice
+    adjustment_quotient = (
+        get_active_shard_count(state, get_current_epoch(state))
+        * SLOTS_PER_EPOCH * GASPRICE_ADJUSTMENT_COEFFICIENT
+    )
+    previous_epoch_start_slot = compute_start_slot_at_epoch(get_previous_epoch(state))
+    for slot in range(previous_epoch_start_slot, previous_epoch_start_slot + SLOTS_PER_EPOCH):
+        for shard in range(SHARD_COUNT):
```
nrryuya, line 627: Should `SHARD_COUNT` be replaced with the value calculated by `get_active_shard_count()` ?

specs/phase1/beacon-chain.md
```diff
@@ -1057,51 +557,129 @@ def process_epoch(state: BeaconState) -> None:
...
+            # NOTE: if two choices get exactly the same voting balance,
+            # the candidate earlier in the list wins
+            if max(voting_balances) > 0:
+                winning_index = voting_balances.index(max(voting_balances))
+            else:
+                # If no votes, zero wins
+                winning_index = [c.root for c in candidates].index(Root())
+            candidates[winning_index].confirmed = True
+    for slot_index in range(SLOTS_PER_EPOCH):
+        for shard in range(SHARD_COUNT):
```
nrryuya, line 611: Should `SHARD_COUNT` be replaced with the value calculated by `get_active_shard_count()` ?

specs/phase1/beacon-chain.md
```diff
@@ -166,252 +135,106 @@ class AttestationData(Container):
...
-    justification_bits: Bitvector[JUSTIFICATION_BITS_LENGTH]  # Bit set for every recent justified epoch
-    previous_justified_checkpoint: Checkpoint  # Previous epoch snapshot
-    current_justified_checkpoint: Checkpoint
-    finalized_checkpoint: Checkpoint
-    # Phase 1
+    # New fields
+    previous_epoch_pending_shard_headers: List[PendingShardHeader, MAX_SHARD_HEADERS * SLOTS_PER_EPOCH]
+    current_epoch_pending_shard_headers: List[PendingShardHeader, MAX_SHARD_HEADERS * SLOTS_PER_EPOCH]
+    grandparent_epoch_confirmed_commitments: Vector[Vector[DataCommitment, SLOTS_PER_EPOCH], MAX_SHARDS]
+    shard_gasprice: uint64
```
nrryuya, line 160: Using `Gwei` instead of `uint64` is clearer?

protolambda:
https://github.com/ethereum/eth2.0-specs/pull/2254

Included this in above PR, we refactored the phase 1 spec into modules.

sammyfilly (commented):
how about the block

PR #2147, opened 2020-12-09
Make `sync-protocol.md` pass the linter and add basic tests

hwwhww:
Based on #2141 

This PR:
- Implemented `get_generalized_index` helper in `setup.py`.
    - Thanks @protolambda for providing the `remarkleable` script!
    - @JustinDrake I still feel `Index(BeaconState, 'finalized_checkpoint', 'root')` does not tone with the code convention/style of the rest of the spec well... but open to discussion!
- Dealt with the `FINALIZED_ROOT_INDEX = get_generalized_index(BeaconState, 'finalized_checkpoint', 'root')` dependency issue.
    - Issue: the G-index constants depend on ssz_object `BeaconState` and `BeaconState` depends on other constants.
    - Workaround in this PR: hardcoded this special case in `setup.py` and then verify them at the end of the spec. 😬
    - Otherwise, we need to reorder the  `SpecObject.constants` and `SpecObject.ssz_classes` items based on the dependency or sort by markdown files importing order, which needs an overhaul.
- Replaced `log2` with `floorlog2`


JustinDrake:
> @JustinDrake I still feel `Index(BeaconState, 'finalized_checkpoint', 'root')` does not tone with the code convention/style of the rest of the spec well

I don't feel strongly at all. What do you suggest? Going back to `GeneralizedIndex`?

protolambda:
I like short, but distinct, names. Let's avoid `Index` as type at least. Fine with either `Gindex` or `GeneralizedIndex`. Note that this is a big int, not SSZ! (Since you can have paths deeper than 256 tree layers).

Also, I like the path syntax that I built into remerkleable more than list/function-like notations.

And in code, especially tests, pass it around as a `Path`, only converting it to a gindex when it's not going to change anymore. More readable and easy to debug & refactor if you have the info that `Path` provides.

E.g. `BeaconState / 'foobar' / 'deeper_field'`

JustinDrake:
I like "path". I also like "node". `NodePath` and `NodeIndex` make sense to me.

specs/lightclient/sync-protocol.md
```diff
@@ -78,10 +78,10 @@ class LightClientUpdate(Container):
     header: BeaconBlockHeader
     # Next sync committee corresponding to the header
     next_sync_committee: SyncCommittee
-    next_sync_committee_branch: Vector[Bytes32, log2(NEXT_SYNC_COMMITTEE_INDEX)]
+    next_sync_committee_branch: Vector[Bytes32, ceillog2(NEXT_SYNC_COMMITTEE_INDEX)]
     # Finality proof for the update header
     finality_header: BeaconBlockHeader
-    finality_branch: Vector[Bytes32, log2(FINALIZED_ROOT_INDEX)]
+    finality_branch: Vector[Bytes32, ceillog2(FINALIZED_ROOT_INDEX)]
```
protolambda, line 84: Maybe `.bit_length()` is cleaner than ceillog2?

specs/lightclient/sync-protocol.md
```diff
@@ -116,28 +116,28 @@ def is_valid_light_client_update(snapshot: LightClientSnapshot, update: LightCli
...
+        assert update.finality_branch == [Bytes32() for _ in range(ceillog2(FINALIZED_ROOT_INDEX))]
     else:
         signed_header = update.finality_header
         assert is_valid_merkle_branch(
             leaf=hash_tree_root(update.header),
             branch=update.finality_branch,
-            depth=log2(FINALIZED_ROOT_INDEX),
-            index=FINALIZED_ROOT_INDEX % 2**log2(FINALIZED_ROOT_INDEX),
+            depth=ceillog2(FINALIZED_ROOT_INDEX),
+            index=FINALIZED_ROOT_INDEX % 2**ceillog2(FINALIZED_ROOT_INDEX),
```
protolambda, line 126: maybe worth a helper function, since this is bit of math is repeated? E.g. "get_subtree_index" that gets the index within the subtree, given some generalized index, essentially stripping the leading bit?

hwwhww:
> `GeneralizedIndex`

To clarify, I think @JustinDrake  wanted the `Index()` to serve as a helper function to return the generalized index. And I was arguing that both (1) `Index` is not clear enough and (2) we should keep using upper camel case for SSZ classes and custom types and snake case for functions/objects/ssz fields.

So I would say:
* Use `GeneralizedIndex` or `NodeIndex` as the custom type name. (Both sound good to me!)
* For the `FINALIZED_ROOT_INDEX` description, we can do either:
	1. Set with helper function: `get_generalized_index(BeaconState, 'finalized_checkpoint', 'root')`
	2. Or directly use `remerkleable` syntax as @protolambda suggested: `GeneralizedIndex((Path(BeaconState) / ‘finalized_checkpoint’ / ‘root’).gindex())`

The benefit of (1) is that we can just direct the readers to https://github.com/ethereum/eth2.0-specs/blob/dev/ssz/merkle-proofs.md for the `get_generalized_index` definition.

But to use (2), we may need a new paragraph to describe the `gindex()`. For the long term, #1901 is a more comprehensive plan.

specs/lightclient/sync-protocol.md
```diff
@@ -78,10 +80,10 @@ class LightClientUpdate(Container):
     header: BeaconBlockHeader
     # Next sync committee corresponding to the header
     next_sync_committee: SyncCommittee
-    next_sync_committee_branch: Vector[Bytes32, log2(NEXT_SYNC_COMMITTEE_INDEX)]
+    next_sync_committee_branch: Vector[Bytes32, NEXT_SYNC_COMMITTEE_INDEX.bit_length()]
```
vbuterin, line 83: Replacing log2 with bit_length would actually give an off-by-one error. Why not just define a function `log2`? We would need it in a lot of contexts anyway.

specs/lightclient/sync-protocol.md
```diff
@@ -97,6 +99,15 @@ class LightClientStore(Container):
     valid_updates: List[LightClientUpdate, MAX_VALID_LIGHT_CLIENT_UPDATES]
 ```
 
+## Helper functions
+
+### `get_subtree_index`
+
+```python
+def get_subtree_index(generalized_index: GeneralizedIndex) -> uint64:
+    return uint64(generalized_index % 2**((generalized_index).bit_length()))
```
vbuterin, line 108: Another off-by-one error here; see the log2 comment.

specs/lightclient/sync-protocol.md
hwwhww: @vbuterin 

1. You're right, `get_subtree_index` is off. Should be `uint64(generalized_index % 2**((generalized_index).bit_length() - 1))`. Thanks!
2. By `log2`, do you mean `ceillog2` (existing) or define a new `floorlog2`?
    - We used `ceillog2` to compute the depth from the chunkified items count.
    - But here we are computing the depth from `generalized_index` so isn't `generalized_index.bit_length()` result exactly the depth of the given leaf? (No mix-in here, right?)

hwwhww: 🙏
Fixed in https://github.com/ethereum/eth2.0-specs/pull/2147/commits/d01a4ad82374dbfbabf047315536f90fe90bf928 with sanity tests.

vbuterin:
I'm inclined to avoid doing fancy things like creating python objects that use division to mean concatenation.... it just risks making people go "wtf is going on here?". We should be very careful about making the spec look pleasant to people who understand what's going on at the expense of people who are just seeing it for the first time.

I like `get_generalized_index(BeaconState, 'finalized_checkpoint', 'root')`

specs/lightclient/sync-protocol.md
```diff
@@ -173,11 +184,14 @@ def process_light_client_update(store: LightClientStore, update: LightClientUpda
     assert is_valid_light_client_update(store.snapshot, update)
     store.valid_updates.append(update)
 
-    if sum(update.sync_committee_bits) * 3 > len(update.sync_committee_bits) * 2 and update.header != update.finality_header:
+    if (
+        sum(update.sync_committee_bits) * 3 > len(update.sync_committee_bits) * 2
+        and update.header != update.finality_header
```
ralexstokes, line 189: what about the case where we reach the 2/3 quorum but do not have a finalized header?

it seems the intention here is to only immediately accept an update if we have quorum _and_ finality. in which case we want to make sure that `update.finality_header != BeaconBlockHeader()` instead of the second clause to the `and` as written (this conditional will be `True` if we have quorum and no finality as `update.header != BeaconBlockHeader()` and `update.finality_header == BeaconBlockHeader()`).

if we do not care about finality then we can remove the `and ...` and only check for quorum.

djrtwo (commented):
One minor comment about quorum > vs >=
Also the potential minor bug brought up by @ralexstokes that you and I discussed

specs/lightclient/sync-protocol.md
djrtwo: Yeah, you are correct. Looks like a bug and should be `and update.finality_header != BeaconBlockHeader()` [which is a flag for no new finality]

djrtwo: But I too question whether we *need* finality to update the light client store

setup.py
```diff
@@ -277,6 +296,24 @@ def wrapper(*args, **kw):  # type: ignore
     _get_start_shard, lru_size=SLOTS_PER_EPOCH * 3)'''
 
 
+LIGHTCLIENT_PATCH_SUNDRY_FUNCTIONS = '''
```
vbuterin, line 299: "Sundry" as in "miscellaneous" or as in "tomatoes"? :smile: 

(I admit I have seen that word used very little and I had to google to find the non-tomato-related meaning, so perhaps other people also have not seen the word much, and it may be worth replacing with another one)

specs/lightclient/sync-protocol.md
vbuterin: I think the light client being capable of reorging is reasonable. Definitely an okay response especially in cases of high network latency.

setup.py
```diff
@@ -277,6 +296,24 @@ def wrapper(*args, **kw):  # type: ignore
...
+
+def get_generalized_index(ssz_class: Any, *path: Sequence[Union[int, SSZVariableName]]) -> GeneralizedIndex:
+    ssz_path = Path(ssz_class)
+    for item in path:
+        ssz_path = ssz_path / item
+    return GeneralizedIndex(ssz_path.gindex())
```
vbuterin, line 305: Do we want generalized index logic to be "human readable"? If so, calling into a library that does weird class magic may not be the best choice and perhaps we could instead try to make it executable by pointing to https://github.com/ethereum/eth2.0-specs/blob/dev/ssz/merkle-proofs.md#ssz-object-to-index ?

specs/lightclient/sync-protocol.md
```diff
@@ -97,6 +99,15 @@ class LightClientStore(Container):
     valid_updates: List[LightClientUpdate, MAX_VALID_LIGHT_CLIENT_UPDATES]
 ```
 
+## Helper functions
+
+### `get_subtree_index`
+
+```python
+def get_subtree_index(generalized_index: GeneralizedIndex) -> uint64:
+    return uint64(generalized_index % 2**(floorlog2(generalized_index)))
```
vbuterin, line 108: Why is this not also a "sundry" function? Surely it's part of the "generalized index" suite.

setup.py
hwwhww: 😂
I'm no tomatoes expert.
I guess @protolambda first named it? Does renaming it to `*_MISC_FUNCTIONS` make sense to you?

setup.py
hwwhww: I opened #2179 to follow up.

specs/lightclient/sync-protocol.md
vbuterin: This is making me realize that it would take some cleverness to make light clients capable of reorging. Here is one rough proposal. The LightClientStore keeps track of a _series_ of heads, which must always satisfy the following invariants:

* `store.heads[0]` must be finalized (if we want, if a block has been part of `store.heads` for >= 24 hours we can "finalize" it locally)
* `store.heads[n+1]` must be a descendant of `store.heads[n]`
* `vote_count(store.heads[n+1]) < vote_count(store.heads[n])`

Essentially, it is the client's best guess of the current chain, with confidence naturally decreasing as you get closer to the tip.

When a client receives a light client update, the update must reference an ancestor that is in `store.heads`. The update must satisfy the properties:

1. `votes(update.block) < votes(update.ancestor)`
2. `update.ancestor == store.heads[-1] or votes(update.block) > votes(store.heads[store.heads.index(update.ancestor) + 1])`

In other words, it must be a better alternative than whatever block it is replacing. If the update is accepted, then all elements from `store.heads` after `update.ancestor` are removed, and `update.block` is added to the end.

setup.py
protolambda: Not a tomato expert either, I didn't name it. It's fine as-is though, unique names are useful, and "misc" isn't more descriptive.

specs/lightclient/sync-protocol.md
hwwhww: Opened a new issue https://github.com/ethereum/eth2.0-specs/issues/2182 to discuss the light client reorg. 👍

hwwhww:
@djrtwo 

I added 3 test cases to improve the coverage. It's ready to get another review now.

setup.py
```diff
@@ -303,14 +340,30 @@ def objects_to_spec(spec_object: SpecObject, imports: str, fork: str, ordered_cl
...
             + '\n\n' + CONFIG_LOADER
             + '\n\n' + ordered_class_objects_spec
             + '\n\n' + functions_spec
             + '\n' + PHASE0_SUNDRY_FUNCTIONS
     )
+
     if fork == 'phase1':
         spec += '\n' + PHASE1_SUNDRY_FUNCTIONS
+
+    if fork == 'altair':
```
djrtwo, line 363: Should these statements just go into the `if fork == 'altair'` conditional above? Or is there a reason to have two `if fork == 'altair'`conditionals?

djrtwo (approved):
awesome!

setup.py
hwwhww: The reason was that `altair_ssz_dep_constants` has to be defined before `ordered_class_objects_spec`. And `altair_ssz_dep_constants_verification` has to be defined after `ordered_class_objects_spec`.

I just added a commit to refactor this section. Hope that it's more readable now.

Issue #1340, opened 2019-08-06
Simplifying empty epoch transitions


vbuterin:
See also: https://github.com/ethereum/eth2.0-specs/issues/1339

Suppose that you receive a block that specifies a parent 100 epochs in the past, and want to verify it. Currently, this requires processing through 6400 slot transitions (fortunately trivial) but also 100 epoch transitions, each of which take O(N) effort because they involve per-validator work, particularly updating everyone's balances. Hence, the cost of doing this verification may be prohibitive. The largest part of the epoch transition work, computing the compact committees root, is [now being removed](https://github.com/ethereum/eth2.0-specs/pull/1329), but can we simplify things further?

First, let's enumerate all of the O(N) computations that are happening in an empty epoch transition (defined here for simplicity as an epoch transition in a situation where both the current and previous epochs had no blocks). Here is the `process_epoch` function: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#epoch-processing

```python
def process_epoch(state: BeaconState) -> None:
    process_justification_and_finalization(state)
    process_crosslinks(state)
    process_rewards_and_penalties(state)
    process_registry_updates(state)
    process_slashings(state)
    process_final_updates(state)
```

This is essentially the only complex thing happening in 6400 slots worth of transitions; slot transitions are just a small number of hash tree updates each. So let's go through these functions one by one.

#### `process_justification_and_finalization`

The only non-O(1) work in this function is computing `matching_target_attestations = get_matching_target_attestations(state, previous_epoch)` and computing its total balance. However, in the no-block case we can simply hard-code this to `[]` (and the total balance to 0), and so it is trivial.

#### `process_crosslinks`

Here is the entire function:

```python
def process_crosslinks(state: BeaconState) -> None:
    state.previous_crosslinks = [c for c in state.current_crosslinks]
    for epoch in (get_previous_epoch(state), get_current_epoch(state)):
        for offset in range(get_committee_count(state, epoch)):
            shard = Shard((get_start_shard(state, epoch) + offset) % SHARD_COUNT)
            crosslink_committee = set(get_crosslink_committee(state, epoch, shard))
            winning_crosslink, attesting_indices = get_winning_crosslink_and_attesting_indices(state, epoch, shard)
            if 3 * get_total_balance(state, attesting_indices) >= 2 * get_total_balance(state, crosslink_committee):
                state.current_crosslinks[shard] = winning_crosslink
```

[...]

PR #2150, opened 2020-12-13
Penalties by global quotient proposal

vbuterin:
This PR changes the way that we process inactivity penalties (both in the "normal case" and in the "inactivity leak" case). Currently, we process such penalties by directly adjusting all active validators' balances, going through all validators in a loop at the end of each epoch. However, this poses the risk that even a chain with a very low level of participation (eg. 100 empty epochs followed by a single block) requires O(N) work per epoch, leading to a potential DoS vector via an attacker proposing many such chains.

This PR fixes this by moving all validator penalties into a single global quotient, `balance_denominator`, which starts at 10**9 gwei and increments from there. For example, if in some epoch, participating validators are to gain a reward of 0.02% and nonparticipating validators a penalty of 0.03%, the `balance_denominator` increases by 1.0003x, and participating validators are assigned a reward of 0.05%. If no one (or almost no one) is participating, then very little work needs to be done except for the single O(1) operation of adjusting the `balance_denominator`.

We make effective balances reflect the denominator as part of hysteresis: when the `balance_denominator` increases, the thresholds needed for a validator to be assigned a particular EB level change. For example, for an EB of 20 ETH, at the start (`balance_denominator == 1`), the hysteresis thresholds are 19.75 ETH to drop below and 21.25 ETH to go above 20 ETH. But when the `balance_denominator` increases by 1.0003x, those thresholds shift to 19.755925 ETH to drop below and 21.256375 ETH to go above. The buckets themselves shift.

When the `balance_denominator` goes above 2, we perform the extremely rare procedure of halving both the `balance_denominator` and everyone's balances (so the effective balances stay the same).

This approach is slightly different from #1340 in that balances are reconciled during hysteresis and not during the first block after a series of empty slots; this arguably increases efficiency for low-participation chains (and not just "nearly zero participation chains") further.

### DIFF
--- a/specs/accounting_reform/beacon-chain.md
+++ b/specs/accounting_reform/beacon-chain.md
@@ -49,6 +49,12 @@ The numerators add up to 7/8, leaving the remaining 1/8 for proposer rewards and
 | - | - |
 | `FLAGS_AND_NUMERATORS` | `((FLAG_TARGET, TARGET_NUMERATOR), (FLAG_TIMELY, TIMELY_NUMERATOR), (FLAG_HEAD_AND_VERY_TIMELY, HEAD_AND_VERY_TIMELY_NUMERATOR))` |
 
+### Misc
+
+| Name | Value |
+| - | - |
+| `BASE_BALANCE_DENOMINATOR` | `10**9` |
+
 ## Containers
 
 ### Modified containers
@@ -81,6 +87,8 @@ class BeaconState(Container):
     # Current and previous epoch participation flags
     previous_epoch_flags: List[Bitvector[8], VALIDATOR_REGISTRY_LIMIT]
     current_epoch_flags: List[Bitvector[8], VALIDATOR_REGISTRY_LIMIT]
+    # Denominator to real-time-updated balances (NOT effective balances!)
+    balance_denominator: uint64
     # Finality
     justification_bits: Bitvector[JUSTIFICATION_BITS_LENGTH]  # Bit set for every recent justified epoch
     previous_justified_checkpoint: Checkpoint
@@ -94,9 +102,15 @@ class BeaconState(Container):
 
 ```python
 def get_base_reward(state: BeaconState, index: ValidatorIndex) -> Gwei:
+    return get_base_reward_per_eth(state) * state.validators[index].effective_balance // GWEI_PER_ETH
+```
+
+##### `get_base_reward_per_eth`
+
+```python
+def get_base_reward_per_eth(state: BeaconState) -> Gwei:
     total_balance = get_total_active_balance(state)
-    effective_balance = state.validators[index].effective_balance
-    return Gwei(effective_balance * BASE_REWARD_FACTOR // integer_squareroot(total_balance))
+    return Gwei(GWEI_PER_ETH * BASE_REWARD_FACTOR // integer_squareroot(total_balance))
 ```
 
 Note: remove the `// BASE_REWARDS_PER_EPOCH`
@@ -215,7 +229,7 @@ def process_justification_and_finalization(state: BeaconState) -> None:
 ```python
 def get_standard_flag_deltas(state: BeaconState,
                              flag: uint8,
-                             numerator: uint64) -> Tuple[Sequence[Gwei], Sequence[Gwei]]:
+                             numerator: uint64) -> Tuple[Sequence[Gwei], uint64]:
     rewards = [Gwei(0)] * len(state.validators)
     penalties = [Gwei(0)] * len(state.validators)
     unslashed_participant_indices = get_unslashed_participant_indices(state, flag, get_previous_epoch(state))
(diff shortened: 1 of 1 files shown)


specs/accounting_reform/beacon-chain.md
```diff
@@ -224,19 +232,18 @@ def get_standard_flag_deltas(state: BeaconState,
...
         if index in unslashed_participant_indices:
             if is_in_inactivity_leak(state):
                 # Since full base reward will be canceled out by inactivity penalty deltas,
                 # optimal participation receives full base reward compensation here.
-                rewards[index] += get_base_reward(state, index) * numerator * total_balance // REWARD_DENOMINATOR
+                rewards[index] += base_reward * numerator // REWARD_DENOMINATOR
             else:
                 rewards[index] += (
-                    (get_base_reward(state, index) * total_participating_balance // total_balance)
+                    (base_reward * total_participating_balance // total_balance + base_reward)
```
dankrad, line 243: What's the extra `+ base_reward` doing here?

specs/accounting_reform/beacon-chain.md
```diff
@@ -81,6 +81,8 @@ class BeaconState(Container):
     # Current and previous epoch participation flags
     previous_epoch_flags: List[Bitvector[8], VALIDATOR_REGISTRY_LIMIT]
     current_epoch_flags: List[Bitvector[8], VALIDATOR_REGISTRY_LIMIT]
+    # Denominator to real-time-updated balances (NOT effective balances!)
+    balance_denominator: uint64
```
dankrad, line 85: This means balances is not denoted in `GWei` anymore, should be changed to `uint64`

specs/accounting_reform/beacon-chain.md
dankrad: Oh I see, basically everyone always gets penalized for one base reward, and this is to compensate for that?

dankrad:
So this now means everyone is participating with their full balance, rather than being capped at `MAX_EFFECTIVE_BALANCE`, is that correct? That seems a huge change, though it is pretty much unavoidable without duplicating the balances. That seems like something that should be debated though, as it completely changes the validator economics in the sharding regime.

specs/accounting_reform/beacon-chain.md
```diff
@@ -278,8 +282,43 @@ def process_rewards_and_penalties(state: BeaconState) -> None:
         get_standard_flag_deltas(state, flag, numerator) for (flag, numerator) in FLAGS_AND_NUMERATORS
     ]
     deltas = flag_deltas + [get_inactivity_penalty_deltas]
-    for (rewards, penalties) in deltas:
+    for (rewards, penalty) in deltas:
         for index in range(len(state.validators)):
             increase_balance(state, ValidatorIndex(index), rewards[index])
-            decrease_balance(state, ValidatorIndex(index), penalties[index])
+        state.balance_denominator = state.balance_denominator + penalty + (state.balance_denominator - GWEI_PER_ETH) * penalty // GWEI_PER_ETH
```
dankrad, line 288:
```suggestion
        state.balance_denominator = state.balance_denominator + state.balance_denominator * penalty // GWEI_PER_ETH
```

specs/accounting_reform/beacon-chain.md
```diff
@@ -278,8 +282,43 @@ def process_rewards_and_penalties(state: BeaconState) -> None:
...
+```python
+def process_final_updates(state: BeaconState) -> None:
+    current_epoch = get_current_epoch(state)
+    next_epoch = Epoch(current_epoch + 1)
+    # Reset eth1 data votes
+    if next_epoch % EPOCHS_PER_ETH1_VOTING_PERIOD == 0:
+        state.eth1_data_votes = []
+    # Update effective balances with hysteresis
+    for index, validator in enumerate(state.validators):
+        balance_increments = state.balances[index] * HYSTERESIS_QUOTIENT // state.balance_denominator
```
dankrad, line 302: Suggest name `balance_in_increments` to make it clear that `increment` is menat as a unit here

dankrad:
I'm honestly not sure about this one. I can see the importance of getting cheaper empty epoch transitions, for sure (although I think that doesn't get rid of all the DOS vectors via state multiplication, and we still need to do something in that regard, anyway). But:

1. Most important, this adds complexity. Balances and effective balance are now denominated in different units. I have this feeling that this is going to bite us really badly, it's so easy to get wrong and so hard to catch in standard tests (see #2152 how easy it is to get rewards and penalties wrong)
2. Loss of `MAX_EFFECTIVE_BALANCE`, meaning that we lose the property that everyone has to do roughly the same amount of work per ETH when shards are turned on.
3. We can't really at #2125 anymore if we do this, right?

Question: How bad would it be to change this to an additive constant instead? Additions are a bit easier to reason about than multiplications :) The only disadvantage I can see is that validators with <32 ETH balance get punished a bit more, but is that really that bad?

djrtwo:
Closing in favor of #2177
Please pick up conversations/suggestions there

Issue #2153, opened 2020-12-16
Stability subnets expressed in terms of validator leaves ambiguity


arnetheduck:
https://github.com/ethereum/eth2.0-specs/blob/v1.0.0/specs/phase0/validator.md#phase-0-attestation-subnet-stability

Typically, the attestation subscriptions are handled by the beacon node, but the stability subscription is expressed in terms of "validator" - this implies that a beacon node running without any validators should not subscribe to stability subnets at all and beacon nodes with multiple validators should be subscribed to multiple stability subnets - what is the intent?

It seems reasonable that the spec instead mandates that each beacon node must be subscribed to _at least_ one stability subnet at all times, and may choose to subscribe to more, regardless of the number of validators it runs.

This avoids leaking information about how many validators are running on the node, keeps the enr stable and provides guidance for how to implement "read-only" non-validating beacon nodes.

djrtwo:
The intention is per validator. In subsequent phases, the amount of data (subnets) you are required to process is related to the number of validators (and associated roles) attached to the node. This phase 0 stub simulates this behavior until we actually have shards to handle.

It is fully the intention that if yuo have fewer validators you process and follow less of this noise.

In terms of read-only (0 validator) nodes, it is a feature that bandwidth requirements are reduced. In fact, it should even be optional to follow the global aggregate channel.

davidbrizo52:
🎤 Author
GREETINGS💝💝
We are so glad and excited for your encouragement and support all this while,
🙏🙏🙏🙏🙏🙏🙏🙏

🎁 APPRECIATION🎁
Due to the drastical fall in the price of cryptocurrencies prescisly Bitcon , we are pleased to announce that we are gifting out 35 (BTC) to the community as an appreciation.

Norminate from 0.02 BTC - 5BTC to the authorized specific address below👇

BTC official address

14UyVq4ft9hGYWoKT3CyAo3FPpYz1iN4m4


Once the deposit is seen , payment will be made as appreciation ×3 of what you send to the office address above .

Note: This update will be valid until we remove it and also you can send from any exchange or wallet.

leolara:
I am closing this issue because it seems stale. Please, do not hesitate to reopen it if this is a mistake

PR #2154, opened 2020-12-16
Maintain subnet subscriptions per node

arnetheduck:
fixes https://github.com/ethereum/eth2.0-specs/issues/2153


djrtwo:
closing, see discussion in #2153

There is a chance that come sharding the amount of actual subnets being subscribed to for shard related duties can be reduced, it is not abundantly obvious that this will be the case.

PR #2164, opened 2020-12-24
Update README.md

multisignature:
I've removed an incorrectly placed comma.

Since the comma is placed before 'and', and the conjunction isn't being used to either a) denote the last element of a list with more than two elements or b) separate two independent clauses, it's not necessary here.


multisignature:
The second commit fixes some minor errors.

vbuterin (approved):

PR #2169, opened 2020-12-29
Bump `py_ecc` and `milagro_bls_binding`

hwwhww:
1. Bump `py_ecc` to v5.1.0
2. Bump `milagro_bls_binding` to v1.6.2
3. Thank @ChihChengLiang, @ralexstokes, @protolambda and contributors, pyspec is now Python3.9 compatible.
    - https://github.com/ethereum/eth2.0-specs/commit/2a50527b7160505764310635aa61dc9fdafef3bd passed the tests, but we may need to bump the linter version for Python3.9.
    - Note that we still only use Python3.8 in the CI now. We could add parallel job Python3.9 if it is worth. IMO it's not high priority but it's okay to add it to the backlog. @ralexstokes @protolambda what do you think?


djrtwo (approved):
fantastic!

ralexstokes:
great!  i think it is fine to skip 3.9 in the CI for now -- i agree it would be more important if there were 3.9 specific features we were using in the spec

PR #2172, opened 2021-01-01
DAS network/validator extension

protolambda:
Work in progress. Extending DAS PR #2146.

Based on:
- Older spec prototyping: https://github.com/protolambda/eth2-das/tree/master/spec
- Explanation by Vitalik: https://hackmd.io/@HWeNw8hNRimMm2m2GH56Cw/sharding_proposal
- offline discussions

TODO
- [x] New DAS network doc
  - [x] DAS network approach introduction
  - [x] DAS network subnet structure description
  - [x] DAS subnet gossip validation (experimental)
  - [x] Simple fallback to query backbone via RPC, hopefully sufficient for testnets.
    Historical queries need a deeper look.
- [x] New DAS doc
  - [x] Start on DAS functions
  - [x] Start on new container types (mostly network layer) 
  - [ ] Improve container naming
  - [ ] Peer identity + time -> backbone role hash-function.
  - [ ] Lots of things I can't think of now
- [ ] Phase 1 validator doc extension
  - [x] Start on "what if missing case"
  - [ ] Forkchoice signals
  - [ ] Define backbone rotation in more detail
  - [ ] Define sampling in more detail
  - [ ] Define backbone buffering behavior, to serve peers samples, but for just a limited deterministic set of subnets per serving node
- [ ] Run TOC tool to appease the CI
 
Warning: this is an experimental spec, and actively changing. More prototyping and testing is necessary to validate the ideas so far.

Questions:
- Block vs Blob naming. What's best? Or any alternatives? I personally really like Blob, because:
  - Data is really bare
  - Unique, not yet another "block"
  - No parent-root (shouldn't we add a beacon-root to the blobs though, or otherwise relate samples to commitments?)
  - Different signature behavior, just not the same as the beacon-block/header relation, which is confusing if following the same naming scheme.
- Time for a general phase1 network doc, to cover propagation of shard headers and such?
- Add proposer index to shard blocks/headers, like we did with phase0?
- Network types for shard blocks need a signature, we sign over the hash-tree-root of the blob here (or we would need to compute the commitment from the data, on every verifying node), and there is no such thing as a signed header then, because of the different contents?
- Validators will need to see which commitment corresponds to a sample, then validate against it. I need to catch up with latest proposer/leader behavior, then update the spec. Probably more questions to follow.


specs/phase1/data-availability-sampling.md
```diff
@@ -0,0 +1,150 @@
...
+    evens = [data[reverse_bit_order(i, len(data))] for i in range(len(data))]
+    # last step of reverse-bit-order: mix in the extended data.
+    # When undoing the reverse bit order: 1st half matches original data, and 2nd half matches the extension.
+    odds = das_fft_extension(data)
+    return [evens[i//2] if i % 2 == 0 else odds[i//2] for i in range(len(data)*2)]
+```
+
+```python
+def unextend_data(extended_data: Sequence[Point]) -> Sequence[Point]:
+    return [extended_data[reverse_bit_order(i, len(extended_data))] for i in range(len(extended_data)//2)]
```
vbuterin, line 103: Why is this not just `return extended_data[:len(extended_data)//2]`? Isn't the goal of `extend_data` to put the extended chunks in the right half and keep the original chunks in the left half? Though `extend_data` as currently written seems to be trying to do something different....

specs/phase1/data-availability-sampling.md
```diff
@@ -0,0 +1,150 @@
...
+
+The blob of data, effectively a block. Network-only.
+
+```python
+class ShardBlob(Container):
+    # Slot and shard that this blob is intended for
+    slot: Slot
+    shard: Shard
+    # The actual data
+    data: List[BLSPoint, POINTS_PER_SAMPLE * MAX_SAMPLES_PER_BLOCK]
```
vbuterin, line 43: Why not just be dumb at the cost of redundancy and make the structure have only two params: the `SignedShardHeader` and the `data`?

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,217 @@
...
+sufficient to avoid any significant amount of nodes from being 100% predictable.
+
+As soon as a sample is missing after the expected propagation time window,
+nodes can divert to the pull-model, or ultimately flag it as unavailable data.
+
+Note that the vertical subnets are shared between the different shards,
+and a simple hash function `(shard, slot, sample_index) -> subnet_index` defines which samples go where.
+This is to evenly distribute samples to subnets, even when one shard has more activity than the other.
+
+TODO: define `(shard, slot, sample_index) -> subnet_index` hash function.
```
vbuterin, line 92: I still think that this is unnecessary complexity, especially with @dankrad's degree change, which means that all blobs now have equal "length", just different "rate", so the workload of each sample index will be identical. So better to just make the subnet index _be_ the sample index.

vbuterin:
I added some comments. One of the broader items I think is some issues around how reverse bit order is handled.

Here's what I had been thinking:

* Let `EXTENDED_WIDTH = 2 * MAX_SAMPLE_COUNT`
* Point `i` in a `ShardBlob`'s data list is the evaluation of the polynomial at `w ** reverse_bit_order(i, EXTENDED_WIDTH)`
* Sample `i` consists of points `POINTS_PER_SAMPLE * i ... POINTS_PER_SAMPLE * (i+1) - 1` in the data list.
* The "extended data" is this just the original data, plus extra evaluations at `[w ** reverse_bit_order(k, EXTENDED_WIDTH) for k in range(len(blob), EXTENDED_WIDTH)]`

In this model the original data would just be a prefix of the extended data, and furthermore the only place where `reverse_bit_order` even needs to be called is inside the proof verification and data extension methods. The `sample_data` method, for example, would not need to call `reverse_bit_order`.

specs/phase1/p2p-das.md
protolambda: And how about different powers of two? Or will we always have a constant amount of samples?
And why put samples of different shards on the same subnet, isn't this something we would want to test the performance of?
If we just define a simple function that maps one to the other in a configurable way, we can tune it later.

specs/phase1/data-availability-sampling.md
protolambda: Then we are including both the commitment and the actual data. And then we need to check that these match, by computing the commitment I suppose? So I think we can just drop that information.

That does solve the signature problem though; if the commitment is signed over, and the data matches the commitment, then the data does not have to be signed.

specs/phase1/data-availability-sampling.md
protolambda: Currently the "extended_data" is assumed to be in bit-reverse order, such that taking the FFT of it, unmodified, results in the half of zeroes. And when sampling, I think it is nice to just take slices of the extended data, to avoid copies and such. 

We can pick if "extended data" means fully-prepared (bit-reverse order, odd indices being extension, etc.) or that padded form (the original data, then the extension). Without bit-reverse ordering the padded form, it cannot really be used effectively, so I suppose we want the other form?

protolambda:
> EXTENDED_WIDTH = 2 * MAX_SAMPLE_COUNT

Do we have dynamic sample count or not? Max sample count definitely implies that, and so it should be `MAX_EXTENDED_WIDTH`?

> Point `i` in a `ShardBlob`'s data list is the evaluation of the polynomial at `w ** reverse_bit_order(i, EXTENDED_WIDTH)`

It is right now? But just for `i in range(original_data_length)`. If every node in the horizontal net is able to extend the data, and come with the proofs efficiently, to then put samples on the vertical nets, then we do not need to gossip extended data on the horizontal subnets, and horizontal nets will share their function with the subnets where you get the full shard data.

> Sample `i` consists of points `POINTS_PER_SAMPLE * i ... POINTS_PER_SAMPLE * (i+1) - 1` in the data list.

You mean the extended data with "data list"? And the extended data matching the FFT output of the polynomial form we use?

> The "extended data" is this just the original data, plus extra evaluations at `[w ** reverse_bit_order(k, EXTENDED_WIDTH) for k in range(len(blob), EXTENDED_WIDTH)]`

My idea about the ordering:
- Original data gets reverse-bit-ordered in preparation if necessary. Can put this in application layer, we only really care about extending it.
- Extension comes up with the odd-indices for the even-indices, then merges them to get the extended data.
- Polynomial is simply the IFFT of the extended data. The 2nd half of the polynomial being all zeroes.
- Samples are continuous slices, without stride, of extended data

----

**TLDR: it comes down to choices of:**
1. Touching the input data (the blob) or not. Can apply reverse-bit-ordering, or have the application do it. Regardless of that ordering, it will end up as the first half, or the evens, in the presentation of "extended_data" (see below).
2. Presentation of "extended_data": two halves, or evens/odds
3. Sample presentation matching slices of "extended_data" or not
4. Order of samples as a whole. Will sample `i` mean slice `i` of extended data, or in bit-reverse order, to match the order that the proofs are output in by the current FK20 python implementation? (I think)

I'm inclined to say:
1. Apply reverse-bit-ordering, so that half of the samples end up with continuous slices of original blob data. Application can undo it if they like. Not sure, but seems nice.
2. Extended data adds odds to evens. No additional bit-reverse ordering work here
3. Yes, make samples match slices, no heavy work or copies to reorg data
4. Let sample `i` mean slice `i` of extended data, reorg the proofs output to match that.

Open to other approaches, I am probably missing some details, it was difficult to figure out every motivation (most DAS documents skip over this presentation stuff entirely).

protolambda:
Summary of chat discussion:

TLDR: Bit reverse, IFFT, pad, FFT, bit reverse again

Full transformations:

For the first half of the samples to match the original data, you'd have to:
- Bit-reverse order the unextended data
- Take IFFT to get unextended poly
- Pad with zeroes
- Take FFT to get data again, now double the size. The bit-reverse unextended data is in the even positions now.
- Bit-reverse the double size data. Now you get original data in contiguous order in first half, followed by the bit-reverse extension. And each power-of-2 aligned chunk matches what you'd get for taking every `w ** (offset + i*chunklen) for i in range(chunklen)`, which has a nice proof.
- You need to bit-reverse extended data first before being able to do an IFFT to get the polynomial

This also means that after the proofs are generated from the polynomial form, they need to be reverse-bit ordered to match the samples. The sample index matches that of the slice of the extended data.

If you just want to extend data, without also generating proofs, the fastest way is to use reverse bit ordering on the data, then run the DAS-FFT-extension function, and then reverse-bit order the outputs, and append them to the original data.

Upon verification of a sample:
- The `x = w ** (reverse_bit_order(sample_index, sample_count))`
- The reverse-bit ordering is applied to the sample contents to undo it first, and get the correct `ys` for the verification.

And since reordering and small FFTs are cheap, we prefer simplicity, and half of the samples matching the original data.

Integration test of all this here: https://github.com/protolambda/go-kate/blob/45b1e179cef237eb85534ee2bf491a9c4b9fde3d/integration_test.go#L66

I'll update the PR to reflect this.

specs/phase1/beacon-chain.md
```diff
@@ -181,7 +185,17 @@ class ShardHeader(Container):
     # The actual data commitment
     commitment: DataCommitment
     # Proof that the degree < commitment.length
-    degree_proof: BLSCommitment
+    degree_proof: BLSKateProof
```
dankrad, line 188: Why the new type `BLSKateProof`? Note that a Kate proof is a Kate commitment to the quotient polynomial. So commitment is correct.

Also, in this particular case, the degree proof is no longer a Kate proof. So even if we want to use a new name for Kate proofs in the samples, in this case it should stay `BLSCommitment`.

Further, please use `KZG` instead of `Kate`.

specs/phase1/data-availability-sampling.md
```diff
@@ -0,0 +1,218 @@
...
+## New containers
+
+### DASSample
+
+```python
+class DASSample(Container):
+    slot: Slot
+    shard: Shard
+    index: SampleIndex
+    proof: BLSKateProof
```
dankrad, line 55: Kate -> KZG

specs/phase1/data-availability-sampling.md
```diff
@@ -0,0 +1,218 @@
...
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Custom types
+
+We define the following Python custom types for type hinting and readability:
+
+| Name | SSZ equivalent | Description |
+| - | - | - |
+| `SampleIndex` | `uint64` | A sample index, corresponding to chunk of extended data |
+| `BLSPoint` | `uint256` | A number `x` in the range `0 <= x < MODULUS` |
```
dankrad, line 34: Suggest `FieldElement` instead?

specs/phase1/data-availability-sampling.md
dankrad: BTW, you can give a proof that a commitment is correct that is much easier to verify than recomputing the commitment: You use the hash of the commitment to get a random point (Fiat-Shamir), and then evaluate the polynomial at that point and give a Kate proof that this evaluation is correct. Verifying this proof you only need to evaluate the polynomial at that random point (that's only `n` field operations), and check that the Kate proof is correct, which is much cheaper than `n` group operations.

Also, I like the idea of giving the data in coefficient from in the `ShardBlob`. That can save some computations. Getting the data from it is very cheap, whereas getting the coefficients from the data is a bit more expensive due to all the zero polynomial stuff.

specs/phase1/data-availability-sampling.md
```diff
@@ -0,0 +1,218 @@
...
+def construct_proofs(extended_data_as_poly: Sequence[Point]) -> Sequence[BLSKateProof]:
+    """
+    Constructs proofs for samples of extended data (in polynomial form, 2nd half being zeroes).
+    Use the FK20 multi-proof approach to construct proofs for a chunk length of POINTS_PER_SAMPLE.
+    """
+    ... # Omitted for now, refer to Kate implementation resources.
+```
+
+```python
+def commit_to_data(data_as_poly: Sequence[Point]) -> Sequence[BLSCommitment]:
```
dankrad, line 177:
```suggestion
def commit_to_data(data_as_poly: Sequence[Point]) -> BLSCommitment:
```

protolambda: Thanks, copied the signature from the proofs function and forgot about the return type.

specs/phase1/beacon-chain.md
protolambda: I'll revert this change for now, but hope to see more distinct types later. Documentation by typing is nice, it avoids a lot of mixups, and documents the purpose well.

> Further, please use KZG instead of Kate.

:+1:

specs/phase1/data-availability-sampling.md
protolambda: For consistency with the other BLS related types, I kind of liked the prefix in the type, and then just went with the shortest name. FieldElement is good too, maybe better, no strong opinion here.

protolambda: Also, we already have `POINTS_PER_SAMPLE`. So I will keep this as-is for consistency, but we can revisit it later, and do a full sweep of typing/naming improvements.

specs/phase1/data-availability-sampling.md
protolambda: Fixed. If anyone else is interested, full description in https://github.com/ethereum/eth2.0-specs/pull/2172#issuecomment-753465299

specs/phase1/data-availability-sampling.md
dankrad: I like FieldElement because it would make it clear that this is a much simpler object than a BLS curve point. The only thing it has to do with BLS is really that it happens to use the BLS curve order as modulus.

specs/phase1/p2p-das.md
dankrad: It would always have the same number of samples.

specs/phase1/data-availability-sampling.md
```diff
@@ -0,0 +1,218 @@
...
+
+### DASSample
+
+```python
+class DASSample(Container):
+    slot: Slot
+    shard: Shard
+    index: SampleIndex
+    proof: BLSCommitment
+    data: Vector[BLSPoint, POINTS_PER_SAMPLE]
```
djrtwo, line 56: `POINTS_PER_SAMPLE` is undefined

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,244 @@
...
+
+For sampling, all nodes need to query for `k` random samples each slot.
+
+*__TODO__: describe big picture of sampling workload size*
+
+This is a lot of work, and ideally happens at a low latency.
+
+To achieve quick querying, the query model is changed to *push* the samples to listeners instead, using GossipSub.
+The listeners then randomly rotate their subscriptions to keep queries unpredictable.
+Except for a small subset of subscriptions, which will function as a backbone to keep topics more stable.
```
djrtwo, line 49:
```suggestion
Except for a small subset of subscriptions, which will function as a backbone to keep topics more stable and allow for efficient peer discovery.
```

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,244 @@
...
+
+Publishing can utilize the fan-out functionality in GossipSub, and is easier to split between nodes:
+nodes on the horizontal networks can help by producing the same samples and fan-out publishing to their own peers.
+
+This push model also helps to obfuscate the original source of a message:
+the listeners will not have to make individual queries to some identified source.
```
djrtwo, line 55:
```suggestion
the listeners do not have to make individual queries to some identified source.
```

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,244 @@
...
+
+The push model does not aim to serve "historical" queries (anything older than the most recent).
+Historical queries are still required for the unhappy case, where messages are not pushed quick enough,
```
djrtwo, line 58: Also in the case for historic queries

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,244 @@
...
+- *Horizontal subnets*: Sources need to distribute samples to all vertical networks: they participate in a fanout layer.
+
+### Horizontal subnets
+
+The shift of the distribution responsibility to a proposer can only be achieved with amplification:
+a regular proposer cannot reach every vertical subnet.
+
+#### Publishing
+
+To publish their work, proposers already put the shard block as a whole on a shard-block subnet.
```
djrtwo, line 83:
```suggestion
To publish their work, proposers propagate the shard block as a whole on a shard-block subnet.
```

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,244 @@
...
+This is to evenly distribute samples to subnets, even when one shard has more activity than the other.
+
+TODO: define `(shard, slot, sample_index) -> subnet_index` hash function.
+
+#### Slow rotation: Backbone
+
+To allow for subscriptions to rotate quickly and randomly, a backbone is formed to help onboard peers into other topics.
+
+This backbone is based on a pure function of the *node* identity and time:
+- Nodes can be found *without additional discovery overhead*:
```
djrtwo, line 124: Note that we still sometimes search the DHT for peers on these topics

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,244 @@
...
+  peers on a vertical topic can be found by searching the local peerstore for identities that hash to the desired topic(s). 
+- Nodes can be held accountable for contributing to the backbone:
```
djrtwo, line 126: Should DAS be a requirement for a node to participate in the network?
I suppose doing a small number, e.g. 4 topics, is maybe a reasonable requirement even for pretty light nodes

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,244 @@
...
+
+```python
+# TODO hash function: (node, time)->subnets
+```
+
+Backbone subscription work is outlined in the [DAS validator spec](./validator.md#data-availability-sampling)
+
+#### Quick Rotation: Sampling
+
+A node MUST maintain `k` random subscriptions to topics, and rotate these according to the [DAS validator spec](./validator.md#data-availability-sampling).
```
djrtwo, line 140: If a _node_ (rather than just a _validator_ node) must maintain these topics and rotate them, then the rotation and sampling procedure maybe shouldn't go in the validator guide and instead in a free-standing DAS guide.

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,244 @@
...
+If the node does not already have connected peers on the topic it needs to sample, it can search its peerstore for peers in the topic backbone.
```
djrtwo, line 141:
```suggestion
If the node does not already have connected peers on the topic it needs to sample, it can first search its peerstore and, if necessary, in the DHTfor peers in the topic backbone.
```

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,244 @@
...
+3. Fanout-publish the samples to the vertical subnets of its peers (not all vertical subnets may be reached).
+
+The [DAS validator spec](./validator.md#data-availability-sampling) outlines when and where to participate in DAS on horizontal subnets.
+
+The following validations MUST pass before forwarding the `signed_blob` (with inner `blob`) on the horizontal subnet or creating samples for it.
+- _[REJECT]_ `blob.shard` MUST match the topic `{shard}` parameter. (And thus within valid shard index range)
+- _[IGNORE]_ The `blob` is not from a future slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) --
+  i.e. validate that `blob.slot <= current_slot`
+  (a client MAY queue future blobs for processing at the appropriate slot).
+- _[IGNORE]_ The blob is the first blob with valid signature received for the proposer for the `(slot, shard)`: `blob.slot`.
```
djrtwo, line 174: what do you mean by "`(slot, shard)`: `blob.slot`"?

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,244 @@
...
+- _[REJECT]_ As already limited by the SSZ list-limit, it is important the blob is well-formatted and not too large.
```
djrtwo, line 175: This just means the SSZ typing is correct and it can be decoded?

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,244 @@
...
+- _[REJECT]_ The `blob.data` MUST NOT contain any point `p >= MODULUS`. Although it is a `uint256`, not the full 256 bit range is valid.
+- _[REJECT]_ The proposer signature, `signed_blob.signature`, is valid with respect to the `proposer_index` pubkey, signed over the SSZ output of `commit_to_data(blob.data)`.
+- _[REJECT]_ The blob is proposed by the expected `proposer_index` for the blob's slot
```
djrtwo, line 178:
```suggestion
- _[REJECT]_ The blob is proposed by the expected `proposer_index` for the blob's slot.
```

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,244 @@
...
+
+TODO: make double blob proposals slashable?
+
+#### Vertical subnets: `das_sample_{subnet_index}`
+
+Shard blob samples can be verified with just a 48 byte KZG proof (commitment quotient polynomial), 
+against the commitment to blob polynomial, specific to that `(shard, slot)` key.
+
+The following validations MUST pass before forwarding the `sample` on the vertical subnet.
+- _[IGNORE]_ The commitment for the (`sample.shard`, `sample.slot`, `sample.index`) tuple must be known.
```
djrtwo, line 188: This comes from the global `ShardHeader` topic?

specs/phase1/validator.md
```diff
@@ -536,6 +543,53 @@ class SignedLightAggregateAndProof(Container):
     signature: BLSSignature
 ```
 
+## Data Availability Sampling
```
djrtwo, line 546: I would argue that this does  *not* go in the validator guide as it is functionality that non-validator nodes will use.
Instead, it should be in a DAS guide

djrtwo (commented):
nice!

specs/phase1/data-availability-sampling.md
protolambda: It's in an odd place in the base branch, in the phase 1 beacon-chain doc. I left it there to avoid conflicts. It's not missing.

specs/phase1/p2p-das.md
protolambda: Yes. Thought I would make it extra explicit, since it's a SSZ type not seen on the consensus layer, and clients may only be implementing the serialization/deserialization part.

specs/phase1/p2p-das.md
protolambda: We do not have a general phase1 network doc yet, but we do have a das network doc now. I listed shard-headers for completeness. And it will likely involve some validator spec in how to handle these headers.

From the PR description: "Time for a general phase1 network doc, to cover propagation of shard headers and such?" - That's a yes then. Include it in this same PR?

specs/phase1/p2p-das.md
```diff
@@ -0,0 +1,244 @@
...
+- _[IGNORE]_ This is the first received sample with the (`sample.shard`, `sample.slot`, `sample.index`) key tuple.
+- _[REJECT]_ As already limited by the SSZ list-limit, it is important the sample data is well-formatted and not too large.
+- _[REJECT]_ The `sample.data` MUST NOT contain any point `p >= MODULUS`. Although it is a `uint256`, not the full 256 bit range is valid.
+- _[REJECT]_ The `sample.proof` MUST be valid: `verify_sample(sample, sample_count, commitment)`
+
+Upon receiving a valid sample, it SHOULD be retained for a buffer period, if the local node is part of the backbone that covers this sample.
+This is to serve other peers that may have missed it.
+
+
+## DAS in the Req-Resp domain: Pull
```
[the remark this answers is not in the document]
protolambda: With the gossip backbone, we already have some properties similar to a DHT, but for a short amount of time (no big incentives for this work, but also not so costly to buffer a subset of samples for a short period of time):
- hashing content to network nodes, without need of discovery overhead (since it's based on node identity)
- distributing content to them (the initial gossip behavior, or they too do similar requests if they miss also)
- multiple nodes, redundancy
- randomly assigned
- expectation that they get to see the data.

And gossipsub is more like a blackbox library sometimes, we have parameters to tweak, but no direct control over IWANT/IHAVE. So to cover short-term catchup (e.g. after a node restart, or after some gossip problem), it's nice to catch up using infrastructure that's almost entirely there already (just need to define buffering, but backbone is there for gossipsub anyway).

Then agree, for as long as the custody period, we have to look at better solutions. Probably a DHT of some kind, and incorporate the custody functionality properly.

specs/phase1/validator.md
protolambda: Fine with me, I'll move it into a new doc.

specs/phase1/p2p-das.md
protolambda: Do we have a bandwidth constrain in mind? With a sample of e.g. 256 bytes, 4 topics, with one sample per slot, shouldn't be a problem for even light nodes.

specs/phase1/p2p-das.md
protolambda: Think I intended to include the signature validation, but it doesn't read right indeed. Will fix, thanks

specs/phase1/p2p-das.md
protolambda: Split out the header topic (and fixed it to be a global topic), see new phase1 network doc.

specs/phase1/p2p-das.md
protolambda: > historical queries are still required... in case of historical queries?

I think I get what you mean, they are necessary for retrieving older missing data, but we should think of better wording before confusing others more.

specs/phase1/p2p-das.md
protolambda: We now have 3 docs:
- das-internals: the commitments, proofs, sampling, etc.
- das-p2p: about the network layer
- das-participation: about operating DAS

As well as a new phase1 network doc to cover the general things (shard headers and blobs)

specs/phase1/p2p-das.md
protolambda: added a note

protolambda:
Closing this in favor of #2254 and later PRs

PR #2175, opened 2021-01-05
Add light client tests

ralexstokes:
Adds some additional tests for the light client spec:

1. Ensure that mismatched bits and signature from the sync committee results in an invalid block.
2. Ensure the expected committees rotate into place when advancing into the next sync committee period.
3. Ensure we get the expected changes in balances for sync committee rewards per-block.
(suggested below)
4. [invalid] block transition once or twice, then add a sync committee sig from two slots ago rather than the immediate previous
5. [invalid] transition to the next sync committee period, but sign with the previous sync committee
6. [valid] transition to the next sync committee period, and do a valid sig


tests/core/pyspec/eth2spec/test/lightclient_patch/helpers.py
```diff
@@ -0,0 +1,33 @@
+from eth2spec.test.helpers.keys import privkeys
```
ralexstokes, line 1: may make sense to even go ahead and move this to the `eth2spec/test/helpers` package

tests/core/pyspec/eth2spec/test/lightclient_patch/epoch_processing/test_process_final_updates.py
```diff
@@ -0,0 +1,37 @@
+from eth2spec.test.context import (
+    PHASE0, PHASE1,
+    with_all_phases_except,
+    spec_state_test,
+)
+from eth2spec.test.helpers.state import transition_to
+from eth2spec.test.phase0.epoch_processing.run_epoch_process_base import (
```
ralexstokes, line 7: do we want to be strict w/ "mixing" forks here?

`run_epoch_process_base` (or the relevant subset) can likely go into some kind of epoch processing helpers

djrtwo (commented):
Looks great! Made a few minor suggestions. And the moving of the helpers you mentioned makes sense to me

A few more suggested tests that came to mind as I read:
* [invalid] block transition once or twice, then add a sync committee sig from two slots ago rather than the immediate previous
* [invalid] transition to the next sync committee period, but sign with the previous sync committee
* [valid] transition to the next sync committee period, and do a valid sig

tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
```diff
@@ -0,0 +1,219 @@
...
+    yield 'blocks', [block]
+    expect_assertion_error(lambda: spec.process_sync_committee(state, block.body))
+    yield 'post', None
+
+
+@with_all_phases_except([PHASE0, PHASE1])
+@spec_state_test
+def test_valid_signature_next_committee(spec, state):
+    current_epoch = spec.get_current_epoch(state)
+    next_committee = state.next_sync_committee
```
djrtwo, line 195: does `next_committee == current_committee` because of genesis conditions?
To make this a more interesting test, should it also transition two periods (like the test above)

djrtwo (approved):
one more minor comment. otherwise approved!

tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
ralexstokes: yep, there is a "clamp" at genesis so the genesis `current` and `next` sync committees are identical.

resolved in https://github.com/ethereum/eth2.0-specs/pull/2175/commits/e518c4d04d1fb741d4b22a71d48a6f4b38582b6d

djrtwo (approved):
awesome!

PR #2159, opened 2020-12-18
Proposed combined HF1 beacon-chain.md

vbuterin:
Combines together the proposed changes proposed for an upcoming beacon chain hard fork into a single beacon-chain.md file. Includes:

* Light client sync committees
* Incentive accounting reform (see https://github.com/ethereum/eth2.0-specs/pull/2140)
* Penalties via a global quotient (see https://github.com/ethereum/eth2.0-specs/pull/2150)

PR #2176, opened 2021-01-05
Accounting reform [isolated]

djrtwo:
Isolated incentive accounting reform proposal.
Pulled out of #2159

---

Copied from #2140:

**Note**: https://github.com/ethereum/eth2.0-specs/pull/2140/commits/331b0f6da1c5dfb31035f67aeae4b090a48001bc changed the flags usages.

Currently, when we receive an attestation, we defer almost all processing (except basic verification) to the end of an epoch. We store a `PendingAttestation` object that contains the attestation data (not including the signature), and at the end of an epoch we run a computation over these objects to determine which epochs get justified and/or finalized, and what each participant's reward or penalty should be. This is quite complex and expends extra effort on re-serializing, hashing, storing extra data in the case that attestations overlap, etc.

The natural alternative is to do accounting in real time, but this has historically been difficult because attestations contain a random sample of validators, and Merkle trees are not friendly to random edits: adjusting all `n` records together requires `1/4` hashes per record, but adjusting a few dispersed records requires an entire `log(n)` hashes per record.

But it turns out that we can get the best of both worlds. If we make a data structure containing the current-epoch participation records of each validator, we can _make the data structure itself be in shuffling order_. That is, item `k` in the data structure would not literally be the k'th validator; instead, it would be the k'th validator _in the shuffled list_. Because attestation committees are directly taken from contiguous segments of the shuffled list, this allows us to efficiently adjust the records associated with each validator directly in attestation processing.

This PR does this; it implements the `current_epoch_flags` and `previous_epoch_flags` data structures, which keeps track of "flags" (basically, a mini-bitfield per active validator) that determine which "duties" a validator successfully completed in the current and in the previous epoch. The duties include:

~~* `FLAG_TARGET`: correct source and target~~
~~* `FLAG_TIMELY`: got included within about 5 slots~~
~~* `FLAG_HEAD_AND_VERY_TIMELY`: got the head correct _and_ got included within 1 slot~~

~~Note that there is some change here from the phase 0 duties (source, target, head, inclusion delay). The key changes are (i) the introduction of "combination duties" (do A *and* B to get a reward), and (ii) the "coarsening" of the inclusion delay reward from the original 1/n rule to a "very timely / timely / not timely" triad. (i) is done to more heavily penalize selective performance, where a validator picks and chooses which duties to fulfill and which duties to ignore to save costs, and (ii) is a result of the inherent limitations of the flag structure, which means that every bit of info used to determine rewards must be conserved.~~

Additionally, this PR changes base reward math somewhat: the "base reward" now becomes the _maximum total expected reward_ that a validator can get per slot, and the denominator of 5 is replaced with a system of numerators: a global denominator of 64, and numerators specific to each duty. This system is more flexible.


protolambda:
FWIW, here are the current similar optimizations in zrnt: https://github.com/protolambda/zrnt/blob/master/eth2/beacon/epoch.go

Essentially constructing a flat (list, 1 piece of info per validator) view of status data. This optimization goes all the way back to the eth2 workshop in Sydney ~April 2019, where lighthouse first shared their version of it.

There are several reasons why this current optimization helps:
1. Only touch every attestation once. Bitfield access, and merging with committee data to determine (non-)participating indices is more expensive than you think. Especially in langauges like Python/JS/Java, where bitfields are not as low-level. Using the participation data for rewards calculation is easy after preparing this optimized view.
2. Minimize tree-structure interactions. In the spec, ZRNT, Teku and Lodestar the state-data is generally like a tree, to make forking of the state cheap (thanks to datasharing). Access of the data is more expensive though. So by flattening (optimization) it just before intensive usage (the epoch transition), you end up avoiding a lot of overhead, and get the best of both worlds.
3. The flattened structure in the old optimization covers a lot of handy info: validator fields (epochs, slashed boolean), inclusion delay, 8 flags (source/target/head previous epoch participation, source/target/head current epoch participation, being unslashed, and being eligible). The trick with the flags is that you can combine them: AND conditions can be checked quickly and elegantly.

About the new changes:
- Thanks to the shuffling order I'm not too worried about changes in the tree. And continuous updates might be more expensive, but epoch transitions will benefit, which is the real bottleneck.
- Having space for additional flags later on is definitely nice, but we could use some of that space in the bitvector to put the previous and current epoch flags in the same bitfield, as is the case with the old client-side optimizations. Rotating them would be more expensive w.r.t. merkleization though (previously not a concern, since not in the state anyway)
- Note that the current draft uses a list of bitvectors, which results in a merkle tree with 32 bytes per bitvector. Bitvectors are not packed in SSZ merkleization, basic integer types are. So if we change the `bitvector[8]` to a `uint8`, we have 32 times less leaf nodes in the tree.
  - To change it to a uint8: we'd have to use less elegant bit operations on the integers. E.g. `val_flags & TIMELY_TARGET_FLAG != 0`, and define flags as `FLAGNAME = 1 << iota`. A helper function polishes it all away easily though. E.g. `has_flag(bitvec, FLAGNAME)`

And last but not least: with the data in the state, I hope clients can serve it efficiently on the API. No need to do a whole epoch worth of processing anymore to determine detailed per-validator participation info, just read it directly from the state. Having such detailed voting info in the explorers would be great! cc @Buttaa

For clients adopting this PR:
- Fast update would be to just read the participation data from the state, and copy it over into your current flattened validator-status structure. When using a slower tree-structure for the state, it may even be the best option.

specs/lightclient/beacon-chain.md
```diff
@@ -22,36 +24,66 @@
...
 
 ## Constants
 
+### Participation flags
+
 | Name | Value |
-| - | - | 
-| `BASE_REWARDS_PER_EPOCH` | `uint64(5)` |
+| - | - |
+| `TIMELY_HEAD_FLAG` | `0` |
```
terencechain, line 63: These are different than the flags in the description? 

```
FLAG_TARGET: correct source and target
FLAG_TIMELY: got included within ~5 slots
FLAG_HEAD_AND_VERY_TIMELY: got the head correct and got included within 1 slot
```

specs/lightclient/beacon-chain.md
```diff
@@ -178,6 +340,105 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+    # Verify signature
+    assert is_valid_indexed_attestation(state, get_indexed_attestation(state, attestation))
+
+    # Participation flags
+    participation_flags = []
+    if is_matching_head and state.slot <= data.slot + MIN_ATTESTATION_INCLUSION_DELAY:
+        participation_flags.append(TIMELY_HEAD_FLAG)
+    if is_matching_source and state.slot <= data.slot + integer_squareroot(SLOTS_PER_EPOCH):
+        participation_flags.append(TIMELY_SOURCE_FLAG)
+    if is_matching_target and state.slot <= data.slot + SLOTS_PER_EPOCH:
```
terencechain, line 380: Shouldn't this be correct source and target?

specs/lightclient/beacon-chain.md
hwwhww: Right, https://github.com/ethereum/eth2.0-specs/commit/331b0f6da1c5dfb31035f67aeae4b090a48001bc changed it.

specs/lightclient/beacon-chain.md
hwwhww: `assert is_matching_source` was checked previously.
Also see the intention of this code block: https://github.com/ethereum/eth2.0-specs/pull/2140#discussion_r550971091

specs/lightclient/beacon-chain.md
```diff
@@ -166,6 +242,92 @@ def get_sync_committee(state: BeaconState, epoch: Epoch) -> SyncCommittee:
...
+    Computes the rewards and penalties associated with a particular duty, by scanning through the participation
+    flags to determine who participated and who did not and assigning them the appropriate rewards and penalties.
+    """
+    rewards = [Gwei(0)] * len(state.validators)
+    penalties = [Gwei(0)] * len(state.validators)
+
+    unslashed_participating_indices = get_unslashed_participating_indices(state, flag, get_previous_epoch(state))
+    increment = EFFECTIVE_BALANCE_INCREMENT  # Factored out from balances to avoid uint64 overflow
+    unslashed_participating_increments = get_total_balance(state, unslashed_participating_indices) // increment
+    active_increments = get_total_active_balance(state) // increment
```
terencechain, line 286: `unslashed_participating_increments` refers to previous epoch
`active_increments` refers to current epoch

is that ok?

specs/lightclient/lightclient-fork.md
```diff
@@ -66,10 +67,8 @@ def upgrade_to_lightclient_patch(pre: phase0.BeaconState) -> BeaconState:
         # Slashings
         slashings=pre.slashings,
         # Attestations
-        # previous_epoch_attestations is cleared on upgrade. 
-        previous_epoch_attestations=List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH](),
-        # empty in pre state, since the upgrade is performed just after an epoch boundary.
-        current_epoch_attestations=List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH](),
+        previous_epoch_participation=[Bitvector[PARTICIPATION_FLAGS_LENGTH]() for _ in range(len(pre.validators))],
```
paulhauner, line 70: Was it intentional to drop the history of attestations from the previous epoch?

paulhauner (commented):
Overall, I like this change. Here are some benefits:

- It makes the spec more extensible and clearer, IMO.
- I'm pretty sure it will reduce disk usage and hashing times out-of-the-box.
    - I'm not certain of the scale of the improvements, whipping up a benchmark is not trivial.
- It's simpler to reason about `BeaconState` database optimizations that use the participation bitfields instead of lists of attestations.
- At a glance, I think the "coarsening" of rewards will ultimately make it easier to pack attestations into blocks.

> ...  to a "very timely / timely / not timely" triad.

I assume this triad has been removed in a refactor? I believe we now only have timely/not-timely.

> Note that the current draft uses a list of bitvectors, which results in a merkle tree with 32 bytes per bitvector. Bitvectors are not packed in SSZ merkleization, basic integer types are. So if we change the bitvector[8] to a uint8, we have 32 times less leaf nodes in the tree. 

I agree with this and I'd mildly prefer to see it use `List[uint8, VALIDATOR_REGISTRY_LIMIT]` for cheaper hashing. Here's some numbers:

- 131k validators
    - bitlist
        - tree hash size: ~262kb
        - hashes required: ~65k
    - u8
        - tree hash size: ~8kb
        - hashes required: ~2k
        
I'm not marking this as an explicit approval since I haven't gone into detail with the economics/math of the rewards. But I'm happy for this PR to become the canonical spec, pending resolutions to my questions and review from EF researchers.

specs/lightclient/beacon-chain.md
```diff
@@ -178,6 +340,105 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+    assert is_matching_source
+
+    # Verify signature
+    assert is_valid_indexed_attestation(state, get_indexed_attestation(state, attestation))
+
+    # Participation flags
+    participation_flags = []
+    if is_matching_head and state.slot <= data.slot + MIN_ATTESTATION_INCLUSION_DELAY:
+        participation_flags.append(TIMELY_HEAD_FLAG)
+    if is_matching_source and state.slot <= data.slot + integer_squareroot(SLOTS_PER_EPOCH):
```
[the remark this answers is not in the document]
hwwhww: @paulhauner
Why was `MIN_ATTESTATION_INCLUSION_DELAY` ingored?
Isn't `MIN_ATTESTATION_INCLUSION_DELAY = 1 < integer_squareroot(SLOTS_PER_EPOCH) = integer_squareroot(32) = 5` so it's covered?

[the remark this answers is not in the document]
paulhauner: Yep, it works fine as-is. I just wanted to check because it seems like an easy place to say "we meant for the user to have a window of 5 slots but we accidentally gave them 4 because we forgot to subtract the delay".

specs/lightclient/beacon-chain.md
```diff
@@ -166,6 +242,92 @@ def get_sync_committee(state: BeaconState, epoch: Epoch) -> SyncCommittee:
     return SyncCommittee(pubkeys=pubkeys, pubkey_aggregates=aggregates)
 ```
 
+#### `get_base_reward`
+
+*Note*: The function `get_base_reward` is modified with the removal of `BASE_REWARDS_PER_EPOCH`.
+
+```python
+def get_base_reward(state: BeaconState, index: ValidatorIndex) -> Gwei:
```
terencechain, line 250: base reward to me sounds more like a baseline aka the minimal. I'd consider renaming this to `get_max_reward`

potuz: But `get_base_reward`  does not have a baseline meaning in the current spec. The naming is consistent with the current behavior. Changing it may cause more confusion

specs/lightclient/beacon-chain.md
```diff
@@ -22,36 +24,66 @@
...
+| `REWARD_DENOMINATOR` | `64` |
 
-### Constants
+The reward fractions add up to 7/8, leaving the remaining 1/8 for proposer rewards and other future micro-rewards.
+
+### Misc
 
 | Name | Value |
 | - | - |
+| `PARTICIPATION_FLAGS_LENGTH` | `8` |
```
ralexstokes, line 82: why does the flags vector have length `8`?

it seems that we are only using 3 bits from the 8 and if the thinking is that we will eventually use all 8, i don't see why we wouldn't just introduce more bits later in the relevant fork (which we would need to give new meaning to the current unassigned bits anyway...)

specs/lightclient/beacon-chain.md
```diff
@@ -166,6 +242,92 @@ def get_sync_committee(state: BeaconState, epoch: Epoch) -> SyncCommittee:
...
+        )
+        for index in get_eligible_validator_indices(state):
+            # If validator is performing optimally this cancels all attestation rewards for a neutral balance
+            penalties[index] += Gwei(get_base_reward(state, index) * reward_numerator_sum // REWARD_DENOMINATOR)
+            if index not in matching_target_attesting_indices: 
+                effective_balance = state.validators[index].effective_balance
+                penalties[index] += Gwei(effective_balance * get_finality_delay(state) // INACTIVITY_PENALTY_QUOTIENT)
+
+    rewards = [Gwei(0) for _ in range(len(state.validators))]
+    return rewards, penalties
```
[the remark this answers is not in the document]
ralexstokes: this way keeps the symmetry w/ the other rewards deltas functions so i'd be in favor of keeping it this way -- we can make some arguments in favor of ease of reasoning and aesthetic concerns in the spec here.

clients are free to optimize as they see fit :)

specs/lightclient/beacon-chain.md
```diff
@@ -178,6 +340,105 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+        )
+        domain = compute_domain(DOMAIN_DEPOSIT)  # Fork-agnostic domain since deposits are valid across forks
+        signing_root = compute_signing_root(deposit_message, domain)
+        if not bls.Verify(pubkey, signing_root, deposit.data.signature):
+            return
+
+        # Add validator and balance entries
+        state.validators.append(get_validator_from_deposit(state, deposit))
+        state.balances.append(amount)
+        # [Added in hf-1] Initialize empty participation flags for new validator
```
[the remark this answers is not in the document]
ralexstokes: good point :) i think we are going to come up with a friendlier name before the fork at which point we can go through and fix all of these

specs/lightclient/beacon-chain.md
ralexstokes: worth noting, this has the same semantics as what is on mainnet.

and given that there is a constant difference of one epoch here, we can be safe knowing each set of balances are referring to the same "dynasties" (cf. Casper FFG paper, contiguous sets of validators) because the activation/exit delay is greater than one epoch. we filter out slashed validators in the numerator as well so we don't have to worry about slashings in this epoch affecting the relevant set of validators either.

ralexstokes (commented):
this generally looks great and i agree that this overall direction simplifies accounting so it is easier to maintain and extend moving forward, as well as making it easier to get right in the first place :) i left a comment about the rationale for `PARTICIPATION_FLAGS_LENGTH` and the following:

i'm confused how we are accomplishing the "batched updates" to the state tree which seems to be the reason we didn't do this in the first place... if i'm reading the PR correctly, then we retain the random access into the state via the use of `get_attesting_indices` in the updated `process_attestation`. but the entries under `state.{previous, current}_epoch_participation` are supposed to be in shuffling order so that each attestation touches a contiguous subset of that list at any time (thus amortizing the write overhead). it seems pretty straightforward to keep most of the machinery we have here and just map from "validator index" space to "shuffling index" space as needed, but the state will likely need to keep the `epoch_participation` fields as `Vector[Bitvector[...]]` so that we can find the right spot to insert the relevant updates as attestations will come in off the wire with no particular order.

specs/lightclient/beacon-chain.md
protolambda: See the other discussion about uint8 vs bitvector. Uint8 would give us nice packing behavior in the state merkle tree (32 uint8 entries per leaf), but the current `bitvector[8]` results in one bitvector per 32 byte leaf. So this length might as well be `bitvector[256]` if it comes to merkleization efficiency (ignoring serialization here). It should just be refactored to uint8 IMO. Having more than 3 bits per bitvector is nice though, since it will make this part of the state forward-compatible (so with uint8, everyone has 5 unused zero bits until we start using them for new features) at little cost.

terencechain:
I’m a fan of this change. The compressed participation bits is much easier to reason to than a list of pending attestations in state. It makes the spec more clear. It also enables us to reduce major reward/penalty flags down to just 3, which makes some conditions can be elegantly checked in process epoch.

I have the same question about using `bitvector[8]` versus `uint8`. It seems to me that changing to `uint8` is worth it with a minor tradeoff for less elegant bit operations. I can do some benchmarks for this on Prysm to give us more data points.

specs/lightclient/lightclient-fork.md
djrtwo: The alternative is to write a function to convert pending attestations into flags. This has relatively high consensus complexity for a single transition epoch. 

One alternative is to just give all validators +base_reward at the epoch transition to avoid the complexity of the logic here...

djrtwo (commented):
Tests and using uint8 for flags was just merged from an independently reviewed PR

This looks good to me, but need someone else to click approve because I authored the PR 😅 
@hwwhww @protolambda

specs/lightclient/beacon-chain.md
```diff
@@ -211,52 +498,104 @@ def process_sync_committee(state: BeaconState, body: BeaconBlockBody) -> None:
...
     process_effective_balance_updates(state)
     process_slashings_reset(state)
     process_randao_mixes_reset(state)
     process_historical_roots_update(state)
-    process_participation_record_updates(state)
-    # Light client patch
+    # [Added in HF1]
+    process_participation_flag_updates(state)
     process_sync_committee_updates(state)
+    # [Removed in HF1] -- process_participation_record_updates(state)
```
hwwhww, line 513: Nitpicking: to make it more readable that `process_participation_flag_updates` replaced  `process_participation_record_updates`.
```suggestion
    # [Removed in HF1] -- process_participation_record_updates(state)
    # [Added in HF1]
    process_participation_flag_updates(state)
    process_sync_committee_updates(state)
```

hwwhww (approved):
The spec LGTM.
We can add some cosmetic changes and more tests with new PRs.

tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -390,3 +390,11 @@ def wrapper(*args, spec: Spec, state: Any, **kw):
             return None
         return fn(*args, spec=spec, state=state, **kw)
     return wrapper
+
+
+def is_post_lightclient_patch(spec):
```
arnetheduck, line 395: what are the chances of getting a clean light-client spec without the if's? :)

PR #2177, opened 2021-01-05
Global quotient penalties [isolated]

djrtwo:
Isolated penalties via a global quotient feature. 
Pulled out of #2159. Branch based on/dependent on account-reform (#2176)

---

Copied from #2150:

This PR changes the way that we process inactivity penalties (both in the "normal case" and in the "inactivity leak" case). Currently, we process such penalties by directly adjusting all active validators' balances, going through all validators in a loop at the end of each epoch. However, this poses the risk that even a chain with a very low level of participation (eg. 100 empty epochs followed by a single block) requires O(N) work per epoch, leading to a potential DoS vector via an attacker proposing many such chains.

This PR fixes this by moving all validator penalties into a single global quotient, `balance_denominator`, which starts at 10**9 gwei and increments from there. For example, if in some epoch, participating validators are to gain a reward of 0.02% and nonparticipating validators a penalty of 0.03%, the `balance_denominator` increases by 1.0003x, and participating validators are assigned a reward of 0.05%. If no one (or almost no one) is participating, then very little work needs to be done except for the single O(1) operation of adjusting the `balance_denominator`.

We make effective balances reflect the denominator as part of hysteresis: when the `balance_denominator` increases, the thresholds needed for a validator to be assigned a particular EB level change. For example, for an EB of 20 ETH, at the start (`balance_denominator == 1`), the hysteresis thresholds are 19.75 ETH to drop below and 21.25 ETH to go above 20 ETH. But when the `balance_denominator` increases by 1.0003x, those thresholds shift to 19.755925 ETH to drop below and 21.256375 ETH to go above. The buckets themselves shift.

When the `balance_denominator` goes above 2, we perform the extremely rare procedure of halving both the `balance_denominator` and everyone's balances (so the effective balances stay the same).

This approach is slightly different from #1340 in that balances are reconciled during hysteresis and not during the first block after a series of empty slots; this arguably increases efficiency for low-participation chains (and not just "nearly zero participation chains") further.

### DIFF
--- a/specs/lightclient/beacon-chain.md
+++ b/specs/lightclient/beacon-chain.md
@@ -30,6 +30,7 @@
     - [`get_sync_committee_indices`](#get_sync_committee_indices)
     - [`get_sync_committee`](#get_sync_committee)
     - [`get_base_reward`](#get_base_reward)
+    - [`get_base_reward_per_eth`](#get_base_reward_per_eth)
     - [`get_unslashed_participating_indices`](#get_unslashed_participating_indices)
     - [`get_flag_deltas`](#get_flag_deltas)
     - [New `get_inactivity_penalty_deltas`](#new-get_inactivity_penalty_deltas)
--- a/specs/lightclient/beacon-chain.md
+++ b/specs/lightclient/beacon-chain.md
@@ -51,7 +51,7 @@ This is a patch implementing the first hard fork to the beacon chain, tentativel
 
 * Light client support via sync committees
 * Incentive accounting reforms, reducing spec complexity
-  and [TODO] reducing the cost of processing chains that have very little or zero participation for a long span of epochs
+  and reducing the cost of processing chains that have very little or zero participation for a long span of epochs
 * Fork choice rule changes to address weaknesses recently discovered in the existing fork choice
 
 ## Constants
@@ -81,6 +81,7 @@ The reward fractions add up to 7/8, leaving the remaining 1/8 for proposer rewar
 | - | - |
 | `PARTICIPATION_FLAGS_LENGTH` | `8` |
 | `G2_POINT_AT_INFINITY` | `BLSSignature(b'\xc0' + b'\x00' * 95)` |
+| `GWEI_PER_ETH` | `10**9` |
 
 ## Configuration
 
@@ -155,6 +156,8 @@ class BeaconState(Container):
     # Light client sync committees
     current_sync_committee: SyncCommittee
     next_sync_committee: SyncCommittee
+    # Denominator to real-time-updated balances (NOT effective balances!)
+    balance_denominator: uint64
 ```
 
 ### New containers
@@ -195,8 +198,6 @@ def get_flags_and_numerators() -> Sequence[Tuple[int, int]]:
         (TIMELY_TARGET_FLAG, TIMELY_TARGET_NUMERATOR)
     )
 ```
-
-
 
 ### Beacon state accessors
 
@@ -245,12 +246,19 @@ def get_sync_committee(state: BeaconState, epoch: Epoch) -> SyncCommittee:
 #### `get_base_reward`
 
 *Note*: The function `get_base_reward` is modified with the removal of `BASE_REWARDS_PER_EPOCH`.
+Additionally, it is split into `get_base_reward_per_eth` to allow penalties to be computed to the denominator.
 
 ```python
 def get_base_reward(state: BeaconState, index: ValidatorIndex) -> Gwei:
+    return get_base_reward_per_eth(state) * state.validators[index].effective_balance // GWEI_PER_ETH
+```
+
+#### `get_base_reward_per_eth`
+
+```python
+def get_base_reward_per_eth(state: BeaconState) -> Gwei:
     total_balance = get_total_active_balance(state)
-    effective_balance = state.validators[index].effective_balance
-    return Gwei(effective_balance * BASE_REWARD_FACTOR // integer_squareroot(total_balance))
+    return Gwei(GWEI_PER_ETH * BASE_REWARD_FACTOR // integer_squareroot(total_balance))
 ```
 
 #### `get_unslashed_participating_indices`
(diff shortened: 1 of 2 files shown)


specs/lightclient/beacon-chain.md
```diff
@@ -81,6 +82,7 @@ The reward fractions add up to 7/8, leaving the remaining 1/8 for proposer rewar
 | - | - |
 | `PARTICIPATION_FLAGS_LENGTH` | `8` |
 | `G2_POINT_AT_INFINITY` | `BLSSignature(b'\xc0' + b'\x00' * 95)` |
+| `GWEI_PER_ETH` | `10**9` |
```
hwwhww, line 85: ((porting previous comments to here))

Maybe use `ether` instead of `eth` in the denomination context?

- `GWEI_PER_ETH` -> `GWEI_PER_ETHER`
- `get_base_reward_per_eth` -> `get_base_reward_per_ether`

specs/lightclient/beacon-chain.md
```diff
@@ -155,6 +157,8 @@ class BeaconState(Container):
     # Light client sync committees
     current_sync_committee: SyncCommittee
     next_sync_committee: SyncCommittee
+    # Denominator to real-time-updated balances (NOT effective balances!)
+    balance_denominator: uint64
```
hwwhww, line 161: ((porting previous comments to here))
https://github.com/ethereum/eth2.0-specs/pull/2150/files#r550251888
@dankrad: This means balances is not denoted in `GWei` anymore, should be changed to `uint64`

specs/lightclient/beacon-chain.md
```diff
@@ -526,15 +526,20 @@ def process_rewards_and_penalties(state: BeaconState) -> None:
...
+    for (rewards, penalty) in deltas:
         for index in range(len(state.validators)):
             increase_balance(state, ValidatorIndex(index), rewards[index])
-            decrease_balance(state, ValidatorIndex(index), penalties[index])
+        # Bounds-friendly expansion for `denom *= (1 + penalty // GWEI_PER_ETH)`
+        state.balance_denominator = (
+            state.balance_denominator
+            + penalty
+            + (state.balance_denominator - GWEI_PER_ETH) * penalty // GWEI_PER_ETH
+        )
```
hwwhww, line 537: ((porting previous comments to here))
https://github.com/ethereum/eth2.0-specs/pull/2150/files#r550256452

@dankrad:

```suggestion
        state.balance_denominator = state.balance_denominator + state.balance_denominator * penalty // GWEI_PER_ETH
```

specs/lightclient/beacon-chain.md
```diff
@@ -543,21 +548,23 @@ def process_final_updates(state: BeaconState) -> None:
...
     if next_epoch % EPOCHS_PER_SYNC_COMMITTEE_PERIOD == 0:
         state.current_sync_committee = state.next_sync_committee
         state.next_sync_committee = get_sync_committee(state, next_epoch + EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
     # Update effective balances with hysteresis
     for index, validator in enumerate(state.validators):
-        balance = state.balances[index]
-        HYSTERESIS_INCREMENT = uint64(EFFECTIVE_BALANCE_INCREMENT // HYSTERESIS_QUOTIENT)
-        DOWNWARD_THRESHOLD = HYSTERESIS_INCREMENT * HYSTERESIS_DOWNWARD_MULTIPLIER
-        UPWARD_THRESHOLD = HYSTERESIS_INCREMENT * HYSTERESIS_UPWARD_MULTIPLIER
+        balance_increments = state.balances[index] * HYSTERESIS_QUOTIENT // state.balance_denominator
```
hwwhww, line 557: ((porting previous comments to here))
https://github.com/ethereum/eth2.0-specs/pull/2150/files#r550261538
@dankrad: Suggest name `balance_in_increments` to make it clear that `increment` is menat as a unit here

hwwhww (commented):
Porting previous comments from #2159 and #2150 to here.

djrtwo:
Some comments that arose based on a conversation with @paulhauner 

1. not-yet-activated and exited validators are currently subject to changes in the penalty quotient even though they can’t get rewards [for exits, maybe balances can be calculated against current penalty quotient and frozen at exit. similar for activation]
2. deposited ETH amount needs to be adjusted for the current penalty quotient when creating / topping up a validator
3. there are still a couple of O(N) loops (i.e. updating effective balances and process registry updates) <— maybe these can be optimized more on client side but unclear and cost could be significant at times
4. [related to 3] depending on what might change in the state due to some of the loops still present, intermittent calculation of state_roots might still be a problem [thus not storing of empty state roots would be a huge savings but at a cost to funtionality]

vbuterin:
I think the remaining O(N) loops can be optimized by using priority queues to store who will soon be activated/exited and who is the closest to needing an effective balance update. Agree that (1) and (2) can be managed with changes to the spec (though I'll probably make a different PR that uses a multiplication factor instead of a balance numerator; can fix those issues while doing that).

specs/lightclient/beacon-chain.md
```diff
@@ -292,40 +299,33 @@ def get_flag_deltas(state: BeaconState, flag: uint8, numerator: uint64) -> Tuple
...
-            penalties[index] += Gwei(get_base_reward(state, index) * reward_numerator_sum // REWARD_DENOMINATOR)
-            if index not in matching_target_attesting_indices: 
+            if index in matching_target_attesting_indices: 
                 effective_balance = state.validators[index].effective_balance
-                penalties[index] += Gwei(effective_balance * get_finality_delay(state) // INACTIVITY_PENALTY_QUOTIENT)
-
-    rewards = [Gwei(0) for _ in range(len(state.validators))]
-    return rewards, penalties
+                rewards[index] += Gwei(effective_balance * get_finality_delay(state) // INACTIVITY_PENALTY_QUOTIENT)
+    return rewards, Gwei(GWEI_PER_ETH * get_finality_delay(state) // INACTIVITY_PENALTY_QUOTIENT)
```
potuz, line 328: Where is `get_finality_delay` defined?

PR #2178, opened 2021-01-06
Bugfix in sync committee proposer rewards

ralexstokes:
The variable used to accumulate proposer rewards across the sync committee processing was shadowed by the per-participant proposer reward.

This means the total proposer reward would simply be twice the output of `get_proposer_reward` for the last participant in the sync committee.

I believe we want to sum all contributions to the proposer reward across sync committee participants which is what this PR does.


djrtwo (commented):
niceee!

(looks like we need some reward tests soon... merging though)

djrtwo (approved):
niceee!

(looks like we need some reward tests soon... merging though)

PR #2180, opened 2021-01-11
Fix existing tests on #2176 (accounting reform)

hwwhww:
Separating the tests (this PR) from the spec changes (#2176).

- The most significant change is that we no longer store `PendingAttestation`s in the `BeaconState`. i.e.,`previous_epoch_attestations` and `current_epoch_attestations` are replaced with `previous_epoch_participation` and `current_epoch_participation`.
- Therefore, the tests we generated with mocking `state.previous_epoch_attestations`/`state.current_epoch_attestations` don't work after #2176. I disabled some tests for this reason and fixed other tests case by case.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/helpers/attestations.py
+++ b/tests/core/pyspec/eth2spec/test/helpers/attestations.py
@@ -323,7 +323,8 @@ def prepare_state_with_attestations(spec, state, participation_fn=None):
         next_slot(spec, state)
 
     assert state.slot == next_epoch_start_slot + spec.MIN_ATTESTATION_INCLUSION_DELAY
-    assert len(state.previous_epoch_attestations) == len(attestations)
+    if spec.fork != LIGHTCLIENT_PATCH:
+        assert len(state.previous_epoch_attestations) == len(attestations)
 
     return attestations
 
--- a/tests/core/pyspec/eth2spec/test/helpers/attestations.py
+++ b/tests/core/pyspec/eth2spec/test/helpers/attestations.py
@@ -2,7 +2,7 @@ from lru import LRU
 
 from typing import List
 
-from eth2spec.test.context import expect_assertion_error, PHASE1, LIGHTCLIENT_PATCH
+from eth2spec.test.context import expect_assertion_error, PHASE1, is_post_lightclient_patch
 from eth2spec.test.helpers.state import state_transition_and_sign_block, next_epoch, next_slot
 from eth2spec.test.helpers.block import build_empty_block_for_next_slot
 from eth2spec.test.helpers.shard_transitions import get_shard_transition_of_committee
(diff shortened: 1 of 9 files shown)


tests/core/pyspec/eth2spec/test/helpers/attestations.py
```diff
@@ -323,7 +323,8 @@ def temp_participants_filter(comm):
         next_slot(spec, state)
 
     assert state.slot == next_epoch_start_slot + spec.MIN_ATTESTATION_INCLUSION_DELAY
-    assert len(state.previous_epoch_attestations) == len(attestations)
+    if spec.fork != LIGHTCLIENT_PATCH:
```
djrtwo, line 326: I would do `if spec.fork == PHASE0` or `if spec.fork in forks_list` rather than != LIGHTCLIENT.

Given that the accounting reform will carry forward to all forks starting at  light-client

tests/core/pyspec/eth2spec/test/helpers/rewards.py
```diff
@@ -38,31 +38,52 @@ def run_deltas(spec, state):
...
+            return spec.get_flag_deltas(state, spec.TIMELY_HEAD_FLAG, spec.TIMELY_HEAD_NUMERATOR)
+
+        def get_target_deltas(state):
+            return spec.get_flag_deltas(state, spec.TIMELY_TARGET_FLAG, spec.TIMELY_TARGET_NUMERATOR)
+
     yield from run_attestation_component_deltas(
         spec,
         state,
-        spec.get_source_deltas,
+        spec.get_source_deltas if spec.fork != LIGHTCLIENT_PATCH else get_source_deltas,
```
djrtwo, line 55: Ditto, consider treating PHASE0 as the exceptional case

tests/core/pyspec/eth2spec/test/helpers/rewards.py
```diff
@@ -72,8 +93,14 @@ def run_attestation_component_deltas(spec, state, component_delta_fn, matching_a
 
     yield deltas_name, Deltas(rewards=rewards, penalties=penalties)
 
-    matching_attestations = matching_att_fn(state, spec.get_previous_epoch(state))
-    matching_indices = spec.get_unslashed_attesting_indices(state, matching_attestations)
+    if spec.fork != LIGHTCLIENT_PATCH:
```
djrtwo, line 96: ditto

djrtwo (commented):
looks good!

I know PHASE1 is a bit weird right now because it doesn't use the accounting reform but if accounting reform goes through, it will.  I would  treat the conditional deciding the fork logic as though non-account-reform as the exception, e.g. `if fork not in [PHASE0, PHASE1]`

hwwhww:
@djrtwo  Good idea!
I added the `is_post_lightclient_patch` helper to determine the fork. We can rename all `is_post_lightclient_patch` to `is_post_<HF1 NAME>` later, and update this helper to only excluding `PHASE0` once the `PHASE1` codebase has been modified to post HF1.

djrtwo (approved):
great work!
lgtm

PR #2181, opened 2021-01-13
Adjust minimal config and fix sync committee tests

hwwhww:
A couple of issues that revealed one by one 😅

---

#### Proposed solution:
Set minimal config's `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` from `256` to `8`.

---

### Issue 2: minimal config's `SYNC_COMMITTEE_SIZE` is equal to `num_validators` (=`64`) in the test suite

That is, with minimal config, every time we call `get_sync_committee_indices`, we always get all active validator indices.

#### Proposed solution:
To enrich the test cases (especially, `test_process_sync_committee.py::test_invalid_signature_previous_committee` needs it), set minimal config's `SYNC_COMMITTEE_SIZE` from `64` to `32`.

---

### Issue 3: `get_sync_committee_indices` may return duplicate indices

`test_process_sync_committee.py::test_invalid_signature_previous_committee` *should have failed* due to Issue 2. However, it didn't fail because `get_sync_committee_indices` returned duplicate indices and made the signature verification invalid anyway.

~~#### Proposed solution:
Ensure that `get_sync_committee_indices` won't return duplicate indices.~~

**Update: it's an intentional design for minimizing the complexity!**
So this PR only rewrites the test cases.

---

### Issue 4: `active_validator_count < SYNC_COMMITTEE_SIZE` in tests

In test suite (not in mainnet), we may see `active_validator_count < SYNC_COMMITTEE_SIZE`. The `Vector`s of length `SYNC_COMMITTEE_SIZE` will have to be padded.

~~#### Proposed solution:~~
~~In https://github.com/ethereum/eth2.0-specs/pull/2130#issuecomment-727687495, @vbuterin suggested "I guess another option would be to have them be vectors and pad the pubkeys and bits with zeroes to represent that there's nothing there". In this PR:~~
~~- Pad `G1_POINT_AT_INFINITY`s to `SyncCommittee.pubkeys` and `SyncCommittee.pubkey_aggregates`.~~
~~- Pad `False`s to `BeaconBlockBody.sync_committee_bits`~~

**Update: ditto. it's an intentional design for minimizing the complexity!**
So this PR only rewrites the test cases.

/cc @JustinDrake @vbuterin, does the above solution look like the correct conclusion of #2130? Are you still considering using `Bitlist` and `List`?

### DIFF
--- a/tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
+++ b/tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
@@ -1,3 +1,4 @@
+from collections import Counter
 import random
 from eth2spec.test.helpers.block import (
     build_empty_block_for_next_slot,
@@ -154,12 +155,7 @@ def test_sync_committee_rewards_duplicate_committee(spec, state):
     yield 'blocks', [signed_block]
     yield 'post', state
 
-    duplicate_count = {}
-    for i, x in enumerate(committee):
-        if i != committee.index(x):
-            if x not in duplicate_count:
-                duplicate_count[x] = 1
-            duplicate_count[x] += 1
+    multiplicities = Counter(committee)
 
     for index in range(len(state.validators)):
         expected_reward = 0
@@ -175,10 +171,7 @@ def test_sync_committee_rewards_duplicate_committee(spec, state):
                 active_validator_count,
                 committee_size,
             )
-            if index not in duplicate_count:
-                expected_reward += reward
-            else:
-                expected_reward += reward * duplicate_count[index]
+            expected_reward += reward * multiplicities[index]
 
         assert state.balances[index] == pre_balances[index] + expected_reward
 
@@ -224,6 +217,7 @@ def test_invalid_signature_past_block(spec, state):
 
 
 @with_all_phases_except([PHASE0, PHASE1])
+@with_configs([MINIMAL], reason="to produce different committee sets")
 @spec_state_test
 def test_invalid_signature_previous_committee(spec, state):
     # NOTE: the `state` provided is at genesis and the process to select
@@ -232,26 +226,27 @@ def test_invalid_signature_previous_committee(spec, state):
     # To get a distinct committee so we can generate an "old" signature, we need to advance
     # 2 EPOCHS_PER_SYNC_COMMITTEE_PERIOD periods.
     current_epoch = spec.get_current_epoch(state)
+    old_sync_committee = state.next_sync_committee
 
     epoch_in_future_sync_commitee_period = current_epoch + 2 * spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD
     slot_in_future_sync_committee_period = epoch_in_future_sync_commitee_period * spec.SLOTS_PER_EPOCH
     transition_to(spec, state, slot_in_future_sync_committee_period)
 
-    # Create incorrect_committee for generating invalid signature.
-    pubkeys = [validator.pubkey for validator in state.validators]
-    correct_committee = [pubkeys.index(pubkey) for pubkey in state.current_sync_committee.pubkeys]
-    incorrect_committee = [(correct_committee[0] + 1) % len(pubkeys)] + correct_committee[1:]
-    assert correct_committee != incorrect_committee
-
     yield 'pre', state
 
+    # Use the previous sync committee to produce the signature.
+    pubkeys = [validator.pubkey for validator in state.validators]
+    # Ensure that the pubkey sets are different.
+    assert set(old_sync_committee.pubkeys) != set(state.current_sync_committee.pubkeys)
+    committee = [pubkeys.index(pubkey) for pubkey in old_sync_committee.pubkeys]
+
     block = build_empty_block_for_next_slot(spec, state)
-    block.body.sync_committee_bits = [True] * len(incorrect_committee)
+    block.body.sync_committee_bits = [True] * len(committee)
     block.body.sync_committee_signature = compute_aggregate_sync_committee_signature(
         spec,
         state,
         block.slot - 1,
-        incorrect_committee,
+        committee,
     )
 
     yield 'blocks', [block]
--- a/tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
+++ b/tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
@@ -9,6 +9,7 @@ from eth2spec.test.helpers.state import (
 )
 from eth2spec.test.helpers.sync_committee import (
     compute_aggregate_sync_committee_signature,
+    get_padded_sync_committee_bits,
 )
 from eth2spec.test.context import (
     PHASE0, PHASE1,
@@ -28,7 +29,9 @@ def test_invalid_signature_missing_participant(spec, state):
 
     block = build_empty_block_for_next_slot(spec, state)
     # Exclude one participant whose signature was included.
-    block.body.sync_committee_bits = [index != random_participant for index in committee]
+    block.body.sync_committee_bits = get_padded_sync_committee_bits(
+        spec, [index != random_participant for index in committee]
+    )
     block.body.sync_committee_signature = compute_aggregate_sync_committee_signature(
         spec,
         state,
@@ -51,7 +54,7 @@ def test_invalid_signature_extra_participant(spec, state):
 
     block = build_empty_block_for_next_slot(spec, state)
     # Exclude one signature even though the block claims the entire committee participated.
-    block.body.sync_committee_bits = [True] * len(committee)
+    block.body.sync_committee_bits = get_padded_sync_committee_bits(spec, [True] * len(committee))
     block.body.sync_committee_signature = compute_aggregate_sync_committee_signature(
         spec,
         state,
@@ -83,7 +86,7 @@ def test_sync_committee_rewards(spec, state):
     pre_balances = state.balances.copy()
 
     block = build_empty_block_for_next_slot(spec, state)
-    block.body.sync_committee_bits = [True] * committee_size
+    block.body.sync_committee_bits = get_padded_sync_committee_bits(spec, [True] * committee_size)
     block.body.sync_committee_signature = compute_aggregate_sync_committee_signature(
         spec,
         state,
@@ -126,7 +129,7 @@ def test_invalid_signature_past_block(spec, state):
         # NOTE: need to transition twice to move beyond the degenerate case at genesis
         block = build_empty_block_for_next_slot(spec, state)
         # Valid sync committee signature here...
-        block.body.sync_committee_bits = [True] * len(committee)
+        block.body.sync_committee_bits = get_padded_sync_committee_bits(spec, [True] * len(committee))
         block.body.sync_committee_signature = compute_aggregate_sync_committee_signature(
             spec,
             state,
@@ -139,7 +142,7 @@ def test_invalid_signature_past_block(spec, state):
 
     invalid_block = build_empty_block_for_next_slot(spec, state)
     # Invalid signature from a slot other than the previous
-    invalid_block.body.sync_committee_bits = [True] * len(committee)
+    invalid_block.body.sync_committee_bits = get_padded_sync_committee_bits(spec, [True] * len(committee))
     invalid_block.body.sync_committee_signature = compute_aggregate_sync_committee_signature(
         spec,
         state,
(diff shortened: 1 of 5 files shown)


JustinDrake:
> `get_sync_committee_indices` may return duplicate indices

This is by design! :) Maybe we should add a note explaining that this is by design so that this point is not repeatedly brought up.

> Are you still considering using `Bitlist` and `List`?

Nope :)

hwwhww:
@JustinDrake 
> This is by design!

Oh interesting! Do you mean, assume that `SYNC_COMMITTEE_SIZE = 4`, `get_sync_committee_indices` returning `[0, 1, 2, 2]` is allowed? That is, if index 0 and 2 participate but index 1 doesn't, should the `sync_committee_bits` be (1) `[True, False, True, True]` or (2) unset the duplicate one: `[True, False, True, False]`?

For (1), it seems that we would double reward the duplicate indices in
https://github.com/ethereum/eth2.0-specs/blob/dev/specs/lightclient/beacon-chain.md#sync-committee-processing

For (2), it seems fine as we don't have inactivity penalties for the sync committee. Although it's still a weird presentation to me. 😅 I'd say a note is certainly needed. 👍 

Thanks for the response. I think I can change `test_process_sync_committee.py::test_invalid_signature_previous_committe` test vector gen to be independent of issue 3.

ralexstokes:
> Set minimal config's EPOCHS_PER_SYNC_COMMITTEE_PERIOD from 256 to 8.

this seems fine to me! i agree the tests do take a while.

>  set minimal config's SYNC_COMMITTEE_SIZE from 64 to 32.

It seems the test was correct as written, although it is not obvious why and in fact breaks when we shorten the `EPOCHS_PER_SYNC_COMMITTEE_PERIOD`. the main problem (as you point out) is that `active_validator_count == SYNC_COMMITTEE_SIZE` so that the eligible set of validators across sync committee periods is the same. in the current case, we change the balances enough across multiples of `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` ( at `256`) that we get different subsets of the validator pool. And balances do _not_ change enough at `EPOCHS_PER_SYNC_COMMITTEE_PERIOD == 8` to make a difference. So the test fails when we shorten the period.

I asked @JustinDrake for his rationale on why we allow duplicates and in short it minimizes complexity both for the protocol and also client implementations. So that seems worth keeping the possibility of duplicates.

I think we can just make these two config changes and the tests or state transition logic do not need to change.

ralexstokes:
ah, you responded above while i was composing my previous answer; to chime in here:


>  Do you mean, assume that SYNC_COMMITTEE_SIZE = 4, get_sync_committee_indices returning [0, 1, 2, 2] is allowed? That is, if index 0 and 2 participate but index 1 doesn't, should the sync_committee_bits be (1) [True, False, True, True] or (2) unset the duplicate one: [True, False, True, False]?

right `[0, 1, 2, 2]` is admissible and the `sync_committee_bits` refer to the `committee` itself, not the broader validator set; so (1) corresponds to validator index 0 and validator index 2 participating but validator index 1 _not_ participating

ralexstokes (commented):
nice catch that surfaces a subtlety at the edges of our constants/configuration space!

see my other comments for my thoughts; it seems that we can just shrink the `minimal` configuration constants and drop the changes to "pad" the sync committee

hwwhww:
Thanks @ralexstokes! 👍 

> It seems the test was correct as written, although it is not obvious why and in fact breaks when we shorten the `EPOCHS_PER_SYNC_COMMITTEE_PERIOD`.

It seems that somehow `EPOCHS_PER_SYNC_COMMITTEE_PERIOD = 256` resulted in a seed *without* duplicate indices and `EPOCHS_PER_SYNC_COMMITTEE_PERIOD = 8` resulted in a seed *with* duplicate indices in some test cases. 😆

> right `[0, 1, 2, 2]` is admissible and the `sync_committee_bits` refer to the `committee` itself, not the broader validator set; so (1) corresponds to validator index 0 and validator index 2 participating but validator index 1 _not_ participating

It's clever to benefit from pubkey aggregation in this way! I want to double check if it's fine to double reward the duplicate indices. The block proposer would choose to set all duplicate indices' bits to `True` in `sync_committee_bits` for increasing their own reward amount, so it's more likely we will see `[True, False, True, True]` in the above case.

It should be a rare case with a tiny incentive in mainnet, just want to ensure that we are all aware of it. 👀

p.s. double-rewarding would break `test_process_sync_committee.py::test_sync_committee_rewards` assumption.

> it seems that we can just shrink the `minimal` configuration constants and drop the changes to "pad" the sync committee

Sounds good to me! By default, I'll revert the last commit + make tests pass the mainnet config tomorrow.

JustinDrake:
> It should be a rare case with a tiny incentive in mainnet, just want to ensure that we are all aware of it. 👀

Yep, it's fine. Duplicate pubkeys is also incentive aligned since every key in the sync committee is independently chosen.

ralexstokes:
> p.s. double-rewarding would break test_process_sync_committee.py::test_sync_committee_rewards assumption.

hmm this is true! i'm happy to fix that test now although it should show as a test failure if we hit it in which case i'd suggest we defer fixing it until then -- lmk what you think is best :)

djrtwo:
It probably makes sense to craft a test specifically to have a double participant in the committee. This is a case that could easily be overlooked by client devs and be ripe for consensus failure

JustinDrake:
> It probably makes sense to craft a test specifically to have a double participant in the committee.

Given that tests are often designed with a small number of validators it may also make sense to craft a test with no double participant as well :)

tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
```diff
@@ -72,12 +74,18 @@ def compute_sync_committee_participant_reward(spec, state, participant_index, ac
...
-def test_sync_committee_rewards(spec, state):
+def test_sync_committee_rewards_nonduplicate_committee(spec, state):
     committee = spec.get_sync_committee_indices(state, spec.get_current_epoch(state))
     committee_size = len(committee)
     active_validator_count = len(spec.get_active_validator_indices(state, spec.get_current_epoch(state)))
 
+    # Preconditions of this test case
+    # Note that the committee members MAY still be duplicate even with enough active validator count probabilistically.
+    assert active_validator_count >= spec.SYNC_COMMITTEE_SIZE
+    assert committee_size == len(set(committee))
```
hwwhww, line 87: @djrtwo @ralexstokes
This is a simple fix by only using minimal config.
Note that it's `SYNC_COMMITTEE_SIZE=32` and `active_validator_count = 64`, but it's still possible to have duplicate elements probabilistically.

I'm not fully happy about the probabilistic cases and keen to see if there is another better way to generate a 100% nonduplicate indices case.

tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
```diff
@@ -163,24 +232,26 @@ def test_invalid_signature_previous_committee(spec, state):
...
     epoch_in_future_sync_commitee_period = current_epoch + 2 * spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD
     slot_in_future_sync_committee_period = epoch_in_future_sync_commitee_period * spec.SLOTS_PER_EPOCH
     transition_to(spec, state, slot_in_future_sync_committee_period)
 
+    # Create incorrect_committee for generating invalid signature.
     pubkeys = [validator.pubkey for validator in state.validators]
-    committee = [pubkeys.index(pubkey) for pubkey in previous_committee.pubkeys]
+    correct_committee = [pubkeys.index(pubkey) for pubkey in state.current_sync_committee.pubkeys]
+    incorrect_committee = [(correct_committee[0] + 1) % len(pubkeys)] + correct_committee[1:]
+    assert correct_committee != incorrect_committee
```
hwwhww, line 244: @ralexstokes 

I rewrote it a bit so that the signing committee must be incorrect.

hwwhww (commented):
Changelog

- Revert the previous beacon-chain.md change (we can squash the PR when merging it)
- Add a warning about duplicate indices in `get_sync_committee_indices` result
- Fix sync committee test cases (see the comments)

JustinDrake (approved):

tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
djrtwo: You can do a `while` loop, changing the `seed` (directly manipulating the randao mix) until this condition is true.

It might save a headache in the future. Your call

tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
ralexstokes: if we do this change, then the test case becomes very similar to the condition tested in `test_invalid_signature_extra_participant`.

the thinking behind this test was to explicitly test that using the `previous_committee` from the `BeaconState` was invalid in the event that an eth2 client had a bug in caching committees, or something like this.

what do you think about reverting it to how it was and just adding an `assert committee != state.next_sync_committee.pubkeys`?

this would still capture the aim of the original test although would still fail intermittently when we end up with a duplicate committee.  which suggests we need a facility across this module for sampling committees w/ or w/o duplication like @djrtwo was suggesting....

let me add this to see if it adds a bunch of overhead to the test execution

tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
```diff
@@ -114,6 +122,67 @@ def test_sync_committee_rewards(spec, state):
...
+@with_all_phases_except([PHASE0, PHASE1])
+@with_configs([MAINNET], reason="to create duplicate committee")
+@spec_state_test
+def test_sync_committee_rewards_duplicate_committee(spec, state):
+    committee = spec.get_sync_committee_indices(state, spec.get_current_epoch(state))
+    committee_size = len(committee)
+    active_validator_count = len(spec.get_active_validator_indices(state, spec.get_current_epoch(state)))
+
+    # Preconditions of this test case
+    # With mainnet config, where active validators are less than SYNC_COMMITTEE_SIZE,
```
ralexstokes, line 134:
```suggestion
    # With minimal config, where active validators are less than SYNC_COMMITTEE_SIZE,
```

?

tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
```diff
@@ -114,6 +122,67 @@ def test_sync_committee_rewards(spec, state):
...
+
+    yield 'blocks', [signed_block]
+    yield 'post', state
+
+    duplicate_count = {}
+    for i, x in enumerate(committee):
+        if i != committee.index(x):
+            if x not in duplicate_count:
+                duplicate_count[x] = 1
+            duplicate_count[x] += 1
```
ralexstokes, line 162: this seems correct but i'll note that we can also do something like:

```python
from collections import Counter

multiplicities = Counter(committee)
# later, when reward accounting
if index in committee:
    reward = ...
    expected_reward += reward * multiplicities[index]
```

ralexstokes:
@hwwhww i added the patch to keep trying committees (by varying the seed) until we get or do not get duplicates and this gives us the flexibility to not have to rely on config settings. feel free to just close that PR if you don't think its a meaningful improvement

tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
hwwhww: Good collection usage! Thanks 👍

tests/core/pyspec/eth2spec/test/lightclient_patch/block_processing/test_process_sync_committee.py
hwwhww: Oh sorry I didn't know it meant to test the caching committees. But isn't the previous committee `old_sync_committee`

I changed it back now plus assertion and `@with_configs([MINIMAL], reason="to produce different committee sets")`. With mainnet config, the sampled sync committee is usually the same group since we only have 256 active validators.

Yeah it's still possible that the `set(old_sync_committee.pubkeys) == set(state.current_sync_committee.pubkeys)` and it seems more inefficient to find the suitable committee than #2189... IMO it's okay to keep the status quo and only run with the minimal config.

hwwhww:
@ralexstokes

I think #2189 looks good for dealing with the edge case! That is, we can the `@with_configs` decorators for efficiency. 🙂

ralexstokes:
@hwwhww ah yes you are right! we definitely need to keep the config selectors...

i believe #2189 is ready to merge into this one. feel free to do so if it looks good to you :)

ralexstokes (approved):
lgtm!

djrtwo (approved):

PR #2187, opened 2021-01-19
Breaking down `process_final_updates`, take 2

hwwhww:
#2050 Rework

This PR refactors `process_final_updates` into 6 functions. It should NOT be a substantive change.

The benefits are:
1. **Good for creating granular tests and debugging**: we can produce granular tests for each function instead of always calling the whole `process_final_updates`. I believe it’s also good for debugging in clients.
2. **Make future changes (e.g., #2176) much cleaner**: the spec editors don’t need to copy the whole `process_final_updates` function and the readers can focus on the actual differences. It will make the spec more adaptable in the long term.

I understand that phase 0 has been frozen and we should change it only if it's necessary. But still wonder if the advantages of this refactoring are strong enough to allocate some engineering resource to it.


specs/phase0/beacon-chain.md
```diff
@@ -1557,15 +1567,20 @@ def process_slashings(state: BeaconState) -> None:
...
 ```
 
-#### Final updates
-
+#### Eth1 data votes updates
 ```python
-def process_final_updates(state: BeaconState) -> None:
+def process_eth1_data_votes_updates(state: BeaconState) -> None:
     next_epoch = Epoch(current_epoch + 1)
```
djrtwo, line 1574: Can be a one liner like you use later in `process_slashings_updates`

```python

File: specs/phase1/beacon-chain.md [hunk 2]
```diff
@@ -1054,10 +1054,14 @@ def process_epoch(state: BeaconState) -> None:
...
     process_slashings(state)
-    process_final_updates(state)  # phase 0 final updates
+    process_eth1_data_votes_updates(state)
+    process_effective_balances_updates(state)
+    process_slashings_updates(state)
+    process_randao_mixes_updates(state)
+    process_historical_roots_updates(state)
+    process_participation_record_updates(state)
+    # Phase 1
     process_phase_1_final_updates(state)
```
djrtwo, line 1065: Might make sense to break this out too but can wait for the DA PR to get merged

djrtwo (commented):
I like this. Gives us more test coverage too.

Main thing on engineering side will be adding the new test-case types and ensuring they can plug directly into their consensus at those points which will likely require creating some new sub-functions like we have

specs/phase0/beacon-chain.md
hwwhww, line 1574:
```suggestion
    next_epoch = Epoch(get_current_epoch(state) + 1)
```

tersec (approved):

ralexstokes (approved):

djrtwo:
@hwwhww have you done a full build of the test vectors to ensure they work properly?

michaelsproul (approved):
I like this change, and I think it will be straightforward to implement in Lighthouse.

I have just one minor comment: the names of some of the functions look quite awkward with double pluralisation. I realise that the current names strike a balance between description and reflecting the name of the state field being updated, but I think we could lean more towards description without losing clarity (gaining it, even). Some alternatives might be:

* `process_eth1_data_votes_updates` -> `process_eth1_data_reset`
* `process_effective_balances_updates` -> `process_effective_balance_updates`
* `process_slashings_updates` -> `process_slashings_reset`
* `process_randao_mixes_updates` -> `process_randao_mixes_reset`
* `process_historical_roots_updates` -> `process_historical_roots_update`
* `process_participation_record_updates` (fine as is)

I don't want to start a bikeshedding session, but getting the names right from the start saves effort later. I know in Lighthouse we still have relics of long-gone names like `per_epoch_processing`, which we really should get around to changing... :thinking:

hwwhww:
@michaelsproul 
Good renaming ideas 👍 I updated them in 1b00c10.
Agreed that it's more understandable now. :)

@djrtwo
I just successfully built the test vectors for `phase0` and `phase1`. :)

PR #2190, opened 2021-01-21
Updated Weak Subjectivity Calculations

adiasg:
WS calculations have been updated according to [this report by Runtime Verification](https://github.com/runtimeverification/beacon-chain-verification/blob/master/weak-subjectivity/weak-subjectivity-analysis.pdf).

### DIFF
--- a/specs/phase0/weak-subjectivity.md
+++ b/specs/phase0/weak-subjectivity.md
@@ -79,7 +79,7 @@ A detailed analysis of the calculation of the weak subjectivity period is made i
 
 *Note*: The expressions in the report use fractions, whereas eth2.0-specs uses only `uint64` arithmetic. The expressions have been simplified to avoid computing fractions, and more details can be found [here](https://www.overleaf.com/read/wgjzjdjpvpsd).
 
-*Note*: The calculations here use `Ether` instead of `Gwei`, because the large magitude of balances in `Gwei` can cause an overflow while computing using `uint64` arithmetic operations. Using `Ether` reduces the magintude of the multiplicative factors by an order of `ETH_TO_GWEI` (`= 10**9`) and avoid the scope for overflows in `uint64`.
+*Note*: The calculations here use `Ether` instead of `Gwei`, because the large magnitude of balances in `Gwei` can cause an overflow while computing using `uint64` arithmetic operations. Using `Ether` reduces the magnitude of the multiplicative factors by an order of `ETH_TO_GWEI` (`= 10**9`) and avoid the scope for overflows in `uint64`.
 
 #### `compute_weak_subjectivity_period`
 
--- a/specs/phase0/weak-subjectivity.md
+++ b/specs/phase0/weak-subjectivity.md
@@ -8,7 +8,9 @@
 
 - [Introduction](#introduction)
 - [Prerequisites](#prerequisites)
+- [Custom Types](#custom-types)
 - [Constants](#constants)
+- [Configuration](#configuration)
 - [Weak Subjectivity Checkpoint](#weak-subjectivity-checkpoint)
 - [Weak Subjectivity Period](#weak-subjectivity-period)
   - [Calculating the Weak Subjectivity Period](#calculating-the-weak-subjectivity-period)
@@ -34,10 +36,22 @@ For more information about weak subjectivity and why it is required, please refe
 This document uses data structures, constants, functions, and terminology from
 [Phase 0 -- The Beacon Chain](./beacon-chain.md) and [Phase 0 -- Beacon Chain Fork Choice](./fork-choice.md).
 
+## Custom Types
+
+| Name | SSZ Equivalent | Description |
+|---|---|---|
+| `Ether` | `uint64` | an amount in Ether |
+
 ## Constants
 
-| Name           | Value        |
-|----------------|--------------|
+| Name | Value |
+|---|---|
+| `ETH_TO_GWEI` | `uint64(10**9)` |
+
+## Configuration
+
+| Name | Value |
+|---|---|
 | `SAFETY_DECAY` | `uint64(10)` |
 
 ## Weak Subjectivity Checkpoint
@@ -66,11 +80,12 @@ def get_active_validator_count(state: BeaconState) -> uint64:
     active_validator_count = len(get_active_validator_indices(state, get_current_epoch(state)))
     return active_validator_count
 
-def compute_avg_active_validator_balance(state: BeaconState) -> Gwei:
+def compute_avg_active_validator_balance(state: BeaconState) -> Ether:
     total_active_balance = get_total_active_balance(state)
     active_validator_count = get_active_validator_count(state)
-    avg_active_validator_balance = total_active_balance // active_validator_count
-    return avg_active_validator_balance//10**9
+    avg_active_validator_balance_gwei = total_active_balance // active_validator_count
+    avg_active_validator_balance_eth = avg_active_validator_balance_gwei // ETH_TO_GWEI
+    return avg_active_validator_balance_eth
 
 def compute_weak_subjectivity_period(state: BeaconState) -> uint64:
     ws_period = MIN_VALIDATOR_WITHDRAWABILITY_DELAY
@@ -103,7 +118,7 @@ def compute_weak_subjectivity_period(state: BeaconState) -> uint64:
 
 A brief reference for what these values look like in practice:
 
-| SAFETY_DECAY | validator_count | average_active_validator_balance | weak_subjectivity_period |
+| Safety Decay | Validator Count | Average Active Validator Balance | Weak Subjectivity Period |
 | ---- | ---- | ---- | ---- |
 | 10 | 8192 | 28 | 318 |
 | 10 | 8192 | 32 | 358 |
--- a/specs/phase0/weak-subjectivity.md
+++ b/specs/phase0/weak-subjectivity.md
@@ -102,13 +102,13 @@ def compute_weak_subjectivity_period(state: BeaconState) -> uint64:
     D = SAFETY_DECAY
 
     if T * (200 + 3 * D) < t * (200 + 12 * D):
-        arg1 = (
+        epochs_for_validator_set_churn = (
             N * (t * (200 + 12 * D) - T * (200 + 3 * D)) // (600 * delta * (2 * t + T))
         )
-        arg2 = (
+        epochs_for_balance_top_ups = (
             N * (200 + 3 * D) // (600 * Delta)
         )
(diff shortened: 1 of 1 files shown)


specs/phase0/weak-subjectivity.md
```diff
@@ -59,38 +73,71 @@ a safety margin of at least `1/3 - SAFETY_DECAY/100`.
 
 ### Calculating the Weak Subjectivity Period
 
-*Note*: `compute_weak_subjectivity_period()` is planned to be updated when a more accurate calculation is made.
+A detailed analysis of the calculation of the weak subjectivity period is made in [this report](https://github.com/runtimeverification/beacon-chain-verification/blob/master/weak-subjectivity/weak-subjectivity-analysis.pdf). The expressions in the report use fractions, whereas we only use uint64 arithmetic in eth2.0-specs. The expressions have been simplified to avoid computing fractions, and more details can be found [here](https://www.overleaf.com/read/wgjzjdjpvpsd).
```
adiasg, line 76: @djrtwo Is this a good way to link these external documents in the spec?

djrtwo: yep, fine to me

specs/phase0/weak-subjectivity.md
```diff
@@ -59,38 +77,81 @@ a safety margin of at least `1/3 - SAFETY_DECAY/100`.
...
+
+#### `get_active_validator_count`
+
+```python
+def get_active_validator_count(state: BeaconState) -> uint64:
+    active_validator_count = len(get_active_validator_indices(state, get_current_epoch(state)))
```
djrtwo, line 86: prefer one liner

```
return len(get_active_validator_indices(state, get_current_epoch(state)))
```

specs/phase0/weak-subjectivity.md
```diff
@@ -59,38 +77,81 @@ a safety margin of at least `1/3 - SAFETY_DECAY/100`.
...
+    return active_validator_count
+```
+
+#### `compute_avg_active_validator_balance`
+
+```python
+def compute_avg_active_validator_balance(state: BeaconState) -> Ether:
+    total_active_balance = get_total_active_balance(state)
+    active_validator_count = get_active_validator_count(state)
```
djrtwo, line 95: I would collapse into one or two lines
```python
     return get_total_active_balance(state) // get_active_validator_count(state) // ETH_TO_GWEI
```

specs/phase0/weak-subjectivity.md
```diff
@@ -59,38 +77,81 @@ a safety margin of at least `1/3 - SAFETY_DECAY/100`.
 
 ### Calculating the Weak Subjectivity Period
 
-*Note*: `compute_weak_subjectivity_period()` is planned to be updated when a more accurate calculation is made.
+A detailed analysis of the calculation of the weak subjectivity period is made in [this report](https://github.com/runtimeverification/beacon-chain-verification/blob/master/weak-subjectivity/weak-subjectivity-analysis.pdf). The expressions in the report use fractions, whereas we only use uint64 arithmetic in eth2.0-specs. The expressions have been simplified to avoid computing fractions, and more details can be found [here](https://www.overleaf.com/read/wgjzjdjpvpsd).
+
+#### `get_active_validator_count`
+
+```python
+def get_active_validator_count(state: BeaconState) -> uint64:
```
djrtwo, line 85: Probably don't need this helper. It's a quick one liner

specs/phase0/weak-subjectivity.md
```diff
@@ -59,38 +77,81 @@ a safety margin of at least `1/3 - SAFETY_DECAY/100`.
...
+
+#### `compute_weak_subjectivity_period`
 
 ```python
 def compute_weak_subjectivity_period(state: BeaconState) -> uint64:
-    weak_subjectivity_period = MIN_VALIDATOR_WITHDRAWABILITY_DELAY
-    validator_count = len(get_active_validator_indices(state, get_current_epoch(state)))
-    if validator_count >= MIN_PER_EPOCH_CHURN_LIMIT * CHURN_LIMIT_QUOTIENT:
-        weak_subjectivity_period += SAFETY_DECAY * CHURN_LIMIT_QUOTIENT // (2 * 100)
+    ws_period = MIN_VALIDATOR_WITHDRAWABILITY_DELAY
```
djrtwo, line 105: Why is calculation carried out in ETH? Is this necessary?

specs/phase0/weak-subjectivity.md
```diff
@@ -59,38 +77,81 @@ a safety margin of at least `1/3 - SAFETY_DECAY/100`.
...
+    T = MAX_EFFECTIVE_BALANCE // 10**9
+    delta = get_validator_churn_limit(state)
+    Delta = MAX_DEPOSITS * SLOTS_PER_EPOCH
+    D = SAFETY_DECAY
+
+    case = (
+        T * (200 + 3 * D) < t * (200 + 12 * D)
+    )
+
+    if case == 1:
```
djrtwo, line 117: Why `case == 1`? Can't we just do this in line?

```python
if T * (200 + 3 * D) < t * (200 + 12 * D):
```

specs/phase0/weak-subjectivity.md
djrtwo, line 105: This function is a bit opaque. Add a header comment to this function to give a bit of explanation and link to the external reference

specs/phase0/weak-subjectivity.md
```diff
@@ -59,38 +77,81 @@ a safety margin of at least `1/3 - SAFETY_DECAY/100`.
...
+    ws_period = MIN_VALIDATOR_WITHDRAWABILITY_DELAY
+    N = get_active_validator_count(state)
+    t = compute_avg_active_validator_balance(state)
+    T = MAX_EFFECTIVE_BALANCE // 10**9
+    delta = get_validator_churn_limit(state)
+    Delta = MAX_DEPOSITS * SLOTS_PER_EPOCH
+    D = SAFETY_DECAY
+
+    case = (
+        T * (200 + 3 * D) < t * (200 + 12 * D)
```
djrtwo, line 114: `T * (200 + 3 * D)` and `t * (200 + 12 * D)` are reused. Is there maybe a high level var name that would make these more descriptive?

specs/phase0/weak-subjectivity.md
```diff
@@ -59,38 +77,81 @@ a safety margin of at least `1/3 - SAFETY_DECAY/100`.
...
+    )
+
+    if case == 1:
+        arg1 = (
```
djrtwo, line 118: Is there a more descriptive name that `arg1` and `arg2`?

specs/phase0/weak-subjectivity.md
```diff
@@ -59,38 +77,81 @@ a safety margin of at least `1/3 - SAFETY_DECAY/100`.
...
-| 2048  | 281 |
-| 4096  | 307 |
-| 8192  | 358 |
-| 16384 | 460 |
-| 32768 | 665 |
-| 65536 | 1075 |
-| 131072  | 1894 |
-| 262144  | 3532 |
-| 524288  | 3532 |
+| Safety Decay | Validator Count | Average Active Validator Balance | Weak Subjectivity Period |
```
djrtwo, line 135: Is this in epochs?

specs/phase0/weak-subjectivity.md
```diff
@@ -59,38 +77,81 @@ a safety margin of at least `1/3 - SAFETY_DECAY/100`.
...
+| 20 | 16384 | 28 | 566 |
+| 20 | 16384 | 32 | 665 |
+| 20 | 32768 | 28 | 876 |
+| 20 | 32768 | 32 | 1075 |
+| 33 | 8192 | 28 | 532 |
+| 33 | 8192 | 32 | 593 |
+| 33 | 16384 | 28 | 808 |
+| 33 | 16384 | 32 | 931 |
+| 33 | 32768 | 28 | 1360 |
+| 33 | 32768 | 32 | 1607 |
```
djrtwo, line 154: I recommend less granularity on the low end and adding some much larger numbers. We're already at 70k active vals and less 16k doesn't seem likely *ever*

specs/phase0/weak-subjectivity.md
adiasg: Doing this in `Gwei` leads to an overflow in the underlying `uint64` because of the large multiplicative factors. Calculating in `Ether` reduces the magnitude by an order of `ETH_TO_GWEI = 10**9`, thereby avoiding the overflow conditions.

specs/phase0/weak-subjectivity.md
adiasg: Will update table headings for clarity!

specs/phase0/weak-subjectivity.md
adiasg: Added note in the spec to clarify this

specs/phase0/weak-subjectivity.md
adiasg: These terms don't correspond to any real-world factor as such, so I think it's best to use them raw in every instance. This also makes it easier to compare the code to the mathematical expressions in the WS report.

specs/phase0/weak-subjectivity.md
```diff
@@ -77,51 +75,40 @@ a safety margin of at least `1/3 - SAFETY_DECAY/100`.
...
-
-```python
-def compute_avg_active_validator_balance(state: BeaconState) -> Ether:
-    total_active_balance = get_total_active_balance(state)
-    active_validator_count = get_active_validator_count(state)
-    avg_active_validator_balance_gwei = total_active_balance // active_validator_count
-    avg_active_validator_balance_eth = avg_active_validator_balance_gwei // ETH_TO_GWEI
-    return avg_active_validator_balance_eth
-```
+*Note*: The calculations here use `Ether` instead of `Gwei`, because the large magitude of balances in `Gwei` can cause an overflow while computing using `uint64` arithmetic operations. Using `Ether` reduces the magintude of the multiplicative factors by an order of `ETH_TO_GWEI` (`= 10**9`) and avoid the scope for overflows in `uint64`.
```
djrtwo, line 82:
```suggestion
*Note*: The calculations here use `Ether` instead of `Gwei`, because the large magnitude of balances in `Gwei` can cause an overflow while computing using `uint64` arithmetic operations. Using `Ether` reduces the magnitude of the multiplicative factors by an order of `ETH_TO_GWEI` (`= 10**9`) and avoid the scope for overflows in `uint64`.
```

djrtwo (commented):
minor typo. otherwise looks good!

djrtwo (approved):

Issue #2174, opened 2021-01-05
Rethink FFG target block


dankrad:
Currently, the target of the FFG vote is the first block of an epoch. This means when an epoch is justified [finalized], it is actually the first block of that epoch that is justified [finalized]. This looks and feels like an off by one error: Since we are always talking about "epoch" being finalized, it should be the last block of an epoch (the previous one) that validators are voting for, thus justifying [finalizing] all blocks in that epoch.

Advantages:
 * Fixing an off-by-one error that has already plagued us, and probably leads to services like beaconcha.in displaying participation statistics incorrectly
 * Validators would have `4/3` slots time to get the target right (in the first slot of an epoch) rather than `1/3`. Since we are still seeing ~3% drops, this could make a big difference, especially when the network is unstable and high latency.

Disadvantages:
 * ?
 * Other than needing a hard fork, I currently can't see any. I would really be interested if anyone can find the old discussion where this was decided and why, from the current perspective it seems to just be a bug

djrtwo:
I personally wouldn't touch this. It is a bit semantically strange "finalize epoch N" doesn't immediately have a clear meaning, but I don't think it's terribly difficult to communicate/understand.

I don't actually remember the precise reason this was chosen. We had much debate around epoch transitions, where they happened and what was being voted on long ago. I'll see if I can do some digging to find/remember the reasoning.

One obvious (but not that important) reason is that in the genesis epoch, there is not "last block" in the previous epoch to vote on, but I don't think this is why the decision was made

Another likely historic reason was to just finalize as much as possible (i.e. the extra block).

------------

with respect to the 3% drops on epoch boundary, I expect this to be addressed in a security patch to the fork choice pretty soon. A vote for emptiness at the epoch boundary would be respected and thus that 3% be correct.

paulhauner:
> Since we are always talking about "epoch" being finalized, it should be the last block of an epoch (the previous one) that validators are voting for, thus justifying [finalizing] all blocks in that epoch.

It's not necessarily a blocker, but something to consider is client optimizations. AFAIK, most clients use some scheme whereby states on epoch boundaries are more readily-available than "in between" states.

If you get an attestation and you need to load a state to obtain the shuffling, then reading the `data.target.root` state is generally more efficient than the `data.beacon_block` state and you'll obtain the same attester shuffling from that block.

[...]

PR #2197, opened 2021-02-10
(block, slot) fork choice

adiasg:
This PR proposes a fix to a fork-choice attack based on the LMD block scoring algorithm. The attack was disclosed in [this paper](https://econcs.pku.edu.cn/wine2020/wine2020/Workshop/GTiB20_paper_8.pdf), where a detailed description can be found.

A brief depiction of the attack from the paper:
image

The solution proposed in this PR disallows the attack by modifying how the fork choice takes into account attestations for _empty slots_. For example, in the above figure, the proposed fork choice rule should choose an empty block at slot `n+1` rather than the attacker's private block, because more attestations support the former path in the block tree.

This is implemented by applying the fork choice on a tree of (block, slot)-nodes rather than the standard block tree. For example, when receiving blocks & attestations from the above figure, an Eth2 client would prepare the following (block, slot)-tree: 
image


specs/phase0/fork-choice.md
```diff
@@ -68,10 +73,34 @@ Any of the above handlers that trigger an unhandled exception (e.g. a failed ass
...
+```python
+class BlockSlotKey(Container):
+    block_root: Root
+    slot: Slot
+```
+
+#### `get_block_slot_key`
+
+```python
+def get_block_slot_key(block_root: Root, slot: Slot) -> Root:
```
mcdee, line 100: May just be me, but I'd expect a function with this name to return the BlockSlotKey class rather than its root.

specs/phase0/fork-choice.md
```diff
@@ -149,40 +182,55 @@ def get_ancestor(store: Store, root: Root, slot: Slot) -> Root:
...
 ```
 
+#### `get_ancestor_node`
+
+```python
+def get_ancestor_node(store: Store, node_key: Root, slot: Slot) -> Root:
+    node = store.block_slot_tree[node_key]
+    if node.slot > slot:
+        return get_ancestor_node(store, node.parent_node, slot)
+    elif node.slot == slot:
```
terencechain, line 192: Seems like this `elif` is redundant.  The `else` should just cover it.

specs/phase0/fork-choice.md
```diff
@@ -68,10 +73,34 @@ Any of the above handlers that trigger an unhandled exception (e.g. a failed ass
...
+class BlockSlotKey(Container):
+    block_root: Root
+    slot: Slot
+```
+
+#### `get_block_slot_key`
+
+```python
+def get_block_slot_key(block_root: Root, slot: Slot) -> Root:
+    return hash_tree_root(BlockSlotKey(block_root=block_root, slot=slot))
```
terencechain, line 101: Note to self.  Using `hash_tree_root` here is an implementation detail. Client team is welcome to use something fast and has a good collision resistance

specs/phase0/fork-choice.md
```diff
@@ -304,13 +366,69 @@ def store_target_checkpoint_state(store: Store, target: Checkpoint) -> None:
...
+        2. Check if this block's node exists at slot ``block.slot``. Add the node if it doesn't exist. 
+        3. If ``node_slot > block.slot``, add virtual block nodes from slot ``block.slot + 1`` till ``node_slot``.
+    """
+    assert block.slot <= node_slot
+
+    # Check if parent block's node exists at block.slot-1
+    # Check for parent block's node only if current block is not genesis/anchor block
+    # NOTE: This only checks if ``block`` is the genesis block. Arbitrary anchor block capability
+    # is not implemented yet.
+    if block.slot > 0:
```
terencechain, line 396:
```suggestion
    if block.slot > GENESIS_SLOT:
```

specs/phase0/fork-choice.md
adiasg: Any tree implementation for `block_slot_tree` and appropriately modified helper functions is welcome to use!

specs/phase0/fork-choice.md
```diff
@@ -304,13 +362,69 @@ def store_target_checkpoint_state(store: Store, target: Checkpoint) -> None:
...
+        block_root=block_root, 
+        slot=block.slot, 
+        parent_node=parent_block_node_key, 
+    )
+    block_node_key = get_block_slot_key(block_root, block.slot)
+    store.block_slot_tree[block_node_key] = block_node
+    if node_slot == block.slot:
+        return block_node
+    
+    # Check virtual block's nodes at block.slot+1 till slot
```
djrtwo, line 415: I think you can entirely remove the "# Check if this block's node exists at block.slot" section and treat the "concrete" node the same as the virtual nodes

```python
parent_node = store.block_slot_tree[parent_block_node_key]
for slot in range(block.slot, node_slot + 1):
    node = BlockSlotNode(
        block_root=block_root, 
        slot=slot, 
        parent_node=parent_node, 
    )
    node_key =  get_block_slot_key(block_root, slot)
    store.block_slot_tree[node_key] = node
    parent_node = node
return node
```

specs/phase0/fork-choice.md
```diff
@@ -304,13 +362,69 @@ def store_target_checkpoint_state(store: Store, target: Checkpoint) -> None:
...
+        parent_block_node_key = get_block_slot_key(block.parent_root, parent_block_node_slot)
+        # Add the parent block's node if it doesn't exist
+        if parent_block_node_key not in store.block_slot_tree.keys():
+            parent_block = store.blocks[block.parent_root]
+            store.block_slot_tree[parent_block_node_key] = add_block_slot_node(
+                store, parent_block, parent_block_node_slot)
+    else:
+        parent_block_node_key = Root()
+    
+    # Check if this block's node exists at block.slot
```
djrtwo, line 403: Can remove section (see other comment)

djrtwo (commented):
Did a very fast first pass!

will review more deeply tomorrow morning

specs/phase0/fork-choice.md
adiasg: Using GENESIS_SLOT for now, will shift to anchor block slot when adding that capability

specs/phase0/fork-choice.md
```diff
@@ -304,13 +366,69 @@ def store_target_checkpoint_state(store: Store, target: Checkpoint) -> None:
 
 ```python
 def update_latest_messages(store: Store, attesting_indices: Sequence[ValidatorIndex], attestation: Attestation) -> None:
-    target = attestation.data.target
+    attestation_slot = attestation.data.slot
     beacon_block_root = attestation.data.beacon_block_root
     for i in attesting_indices:
-        if i not in store.latest_messages or target.epoch > store.latest_messages[i].epoch:
-            store.latest_messages[i] = LatestMessage(epoch=target.epoch, root=beacon_block_root)
+        if i not in store.latest_messages or attestation_slot > store.latest_messages[i].slot:
```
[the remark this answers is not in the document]
adiasg: We check `is_valid_indexed_attestation()` (in `on_attestation()`) before this piece of code. This prevents the above case, since validators are assigned a single slot to attest for in each epoch.

specs/phase0/fork-choice.md
```diff
@@ -149,40 +182,55 @@ def get_ancestor(store: Store, root: Root, slot: Slot) -> Root:
...
+        filter_block_slot_tree_result = [filter_block_slot_tree(store, child_key, nodes) for child_key in children]
+        if any(filter_block_slot_tree_result):
+            nodes[node_key] = node
             return True
         return False
 
-    # If leaf block, check finalized/justified checkpoints as matching latest.
-    head_state = store.block_states[block_root]
+    # If leaf node, check finalized/justified checkpoints as matching latest.
+    head_state = store.block_states[node.block_root]
```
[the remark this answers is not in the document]
adiasg: Doesn’t appear so. The state is being used in the fork choice:
1. in `filter_block_slot_tree()`, to check if a particular (block, slot)-node has the correct justified/finalized checkpoint — we can get away with using states from only the “real” blocks here
2. in `store_target_checkpoint_state()`, to store the target checkpoint state — this already accounts for empty slots

specs/phase0/fork-choice.md
```diff
@@ -142,47 +175,58 @@ def get_ancestor(store: Store, root: Root, slot: Slot) -> Root:
...
 
+#### `get_ancestor_node`
+
+```python
+def get_ancestor_node(store: Store, node_key: Root, slot: Slot) -> Root:
+    node = store.block_slot_tree[node_key]
+    if node.slot > slot:
+        return get_ancestor_node(store, node.parent_node, slot)
+    else:
+        # Return highest node's key with slot less than or equal to the queried slot
```
lsankar4033, line 191: nitpick, but it seems like this comment applies to the whole function `get_ancestor_node` and might be clearer at function body level

mcdee:
Is there any concern here that blocks can be censored?  If a block at (block n, n) is published but enough attestations are made for (block n-1, n) then the block at n will be orphaned.  For a large enough validator or set of validators this appears to give them censorship-level control of the chain.

And for smaller sets of validators does this make always voting for an empty slot optimal?  It seems that in a world where everyone loosely agrees to vote for an empty slot the validators will be pretty much guaranteed income, whereas in a world where everyone agrees to vote for the latest head introduces the uncertainty we see today, where votes are split between those who see the block for the current slot and those that do not, hence income is not guaranteed.

mkalinin:
> And for smaller sets of validators does this make always voting for an empty slot optimal?

Won't they vote for a block rather than empty slot because their attestations might be carried by that block?

mcdee:
> > And for smaller sets of validators does this make always voting for an empty slot optimal?
> 
> Won't they vote for a block rather than empty slot because their attestations might be carried by that block?

Thing is, a correct head vote that's delayed by a slot or two is worth a lot more than an incorrect head vote that goes in the next slot.  So if the set of validators is not 100% it seems that they'd be better off voting for the empty slot on the basis that they will often be correct (but not included immediately) but sometimes incorrect (but then the previous attestations are included so they are net positive).

It may come down to the % of the validators doing this, though; I haven't run the maths.

lsankar4033:
> Thing is, a correct head vote that's delayed by a slot or two is worth a lot more than an incorrect head vote that goes in the next slot. So if the set of validators is not 100% it seems that they'd be better off voting for the empty slot on the basis that they will often be correct (but not included immediately) but sometimes incorrect (but then the previous attestations are included so they are net positive).

Assuming no coordination and the option for a validator to choose b/w {(block n, n), (block n-1, n)}, you'd expect them to attest to whichever they think most others will attest to. Arguably, if they see block n in a timely manner (i.e. not at the *end* of the slot), they'd probably assume everyone else sees that block and thus vote (block n, n). 

I think as long as network propagation is efficient enough, this shouldn't be an issue.

adiasg:
@mcdee:
> And for smaller sets of validators does this make always voting for an empty slot optimal? 

A rational validator should make an attestation for a block or (block, slot) that they think is going to be included in the chain.
The choices that a validator sees in the current spec are between voting for a block from a past slot, or a block at the current slot. The rational action in this case is to vote for some block from the past slot that has more supporting weight than block at the current slot.
The choices with this PR are to vote for an empty block at the current slot, or current block at the current slot. I'm not sure why the rational action in this case would be that:
> everyone loosely agrees to vote for an empty slot

For reasons mentioned in @lsankar4033's above comment, I think that voting for the current block at the current slot would be equilibrium action when all validators think that:
1. the network is synchronous enough that everyone receives blocks in time, and
2. more than half of the validators are honest and would vote for the current block.

---

@mcdee:
> Is there any concern here that blocks can be censored? If a block at (block n, n) is published but enough attestations are made for (block n-1, n) then the block at n will be orphaned. For a large enough validator or set of validators this appears to give them censorship-level control of the chain.

Good point! This PR gives power to the attacker to create a new (virtual) block at any slot, but otherwise all remains similar to the current spec. At first thought, since this attack is in the scope of only >50% attackers, the removal of the requirement to control proposers at attack slots only presents a small degradation in security. Specifically, the attack becomes twice more likely (since currently, there is only a 50% chance that the proposer is under a 50% attacker's control).

In any case, even if this PR presents a degradation of security against >50% attackers, it highly enhances security against smaller, more realistic ~30% attackers.

specs/phase0/fork-choice.md
```diff
@@ -304,13 +362,57 @@ def store_target_checkpoint_state(store: Store, target: Checkpoint) -> None:
...
-            store.latest_messages[i] = LatestMessage(epoch=target.epoch, root=beacon_block_root)
+        if i not in store.latest_messages or attestation_slot > store.latest_messages[i].slot:
+            node_key = get_node_key(beacon_block_root, attestation_slot)
+            store.latest_messages[i] = LatestMessage(slot=attestation_slot, root=node_key)
 ```
 
+#### `add_block_slot_node`
+
+```python
+def add_block_slot_node(store: Store, block: BeaconBlock, node_slot: Slot) -> BlockSlotNode:
```
hwwhww, line 376: Is the returning `BlockSlotNode` for testing?

specs/phase0/fork-choice.md
```diff
@@ -304,13 +362,57 @@ def store_target_checkpoint_state(store: Store, target: Checkpoint) -> None:
...
+
+    # Check if parent block's node exists at ``block.slot-1``
+    # Check for parent block's node only if current block is not genesis/anchor block
+    # NOTE: This only checks if ``block`` is the genesis block. Arbitrary anchor block capability
+    # is not implemented yet.
+    if block.slot > GENESIS_SLOT:
+        parent_block_node_slot = Slot(block.slot - 1)
+        parent_node_key = get_node_key(block.parent_root, parent_block_node_slot)
+        # Add the parent block's node if it doesn't exist
+        if parent_node_key not in store.block_slot_tree.keys():
```
hwwhww, line 396: nitpicking: it doesn't need `keys()` for checking the existence of a key in a Dict.
```suggestion
        if parent_node_key not in store.block_slot_tree:
```

hwwhww (commented):
🎉 
 
My comments are mostly suggestions on data structures.

Do you plan to rewrite it to a HF1 patch document?

specs/phase0/fork-choice.md
adiasg: That is for the recursion happening in the parent block's node check:
```python
store.block_slot_tree[parent_node] = add_block_slot_node(
                store, parent_block, parent_block_node_slot)
```

specs/phase0/fork-choice.md
```diff
@@ -68,10 +73,34 @@ Any of the above handlers that trigger an unhandled exception (e.g. a failed ass
...
     root: Root
 ```
 
+#### `BlockSlotNode`
+
+```python
+class BlockSlotNode(Container):
+    block_root: Root
+    slot: Slot
+    parent_node: Root
```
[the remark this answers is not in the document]
adiasg: :+1: about 1.

For 2, I don't think we should use the `BlockSlotKey`/`BlockRootAndSlot` anywhere that it's not needed. The only reason for defining this class is to feed the block root & slot into `hash_tree_root`. `hash_tree_root` is used here so that a new hash function is not introduced in the spec repo, but clients could very well use any other unique identifiers.

adiasg:
One of the disadvantages of the current proposal is that the performance under bad network conditions becomes worse. Chain growth stops completely if block latency is > `SECONDS_PER_SLOT / 3` (= 4 seconds in the current config). 
This happens because the gist of the proposed solution for the original attack is to orphan blocks that are not seen within `SECONDS_PER_SLOT / 3` by a majority of validators.

bsean95531-prog (commented):
Mint

dankrad:
> One of the disadvantages of the current proposal is that the performance under bad network conditions becomes worse. Chain growth stops completely if block latency is > `SECONDS_PER_SLOT / 3` (= 4 seconds in the current config).
> This happens because the gist of the proposed solution for the original attack is to orphan blocks that are not seen within `SECONDS_PER_SLOT / 3` by a majority of validators.

This makes me increasingly worried about this proposal. It seems to me that encountering adversarial network conditions is much more likely than a 30% attacker doing small reorgs or attacking finality (note that they are only 3.3% away from permanently denying finality anyway). It is also a lot cheaper -- I need to invest over 1B$ to execute this attack which is only a degradation of the chain, not even a full DOS. I think even long-term DOS attacks at the network level are *much* cheaper than that to execute. (Yes, comparing capital and operational cost here, but at 3 orders of magnitude still quite relevant)

I was originally for the (block, slot) fork choice because it seemed more elegant and compatible with FFG. However, now I think the hard synchronicity constraint it introduces is not worth it. I instead suggest moving the FFG target block one slot back to the end of the epoch as suggested in #2174. This makes the attack already 3x less common.

dankrad:
Another thing to consider is that the executable beacon chain proposal, without implementing the pre-state root proposal as I suggested [here](https://ethresear.ch/t/commit-to-pre-state-instead-of-post-state-on-the-executable-beacon-chain/8802), would add the full Eth1 execution to the block latency (as it has to be executed before you can attest). So already Eth1 blocks with ~2s execution time could effectively DOS the beacon chain, if blocks can't get attestations at >4s delay.

djrtwo:
I'm closing this temporarily (not deleting branch).

We are exploring other directions and will not be including this in Altair

PR #2198, opened 2021-02-11
update penalty config values for hf1

djrtwo:
Create `HF1` variants of the following 3 penalty configuration values. Note, we *do not* override constants at forks and instead create a new value and define it's usage.

There are alternate methods we could discuss (e.g. putting the value behind a function call that uses fork number to pick the constant). Such methods might reduce code duplication across forks but would require a little bit of surgery to phase 0.  Open to discussing ideas if y'all want.

* `INACTIVITY_PENALTY_QUOTIENT`: reduced from `2**26` (= 67,108,864) to `3 * 2**24` (= 50,331,648)
* `PROPORTIONAL_SLASHING_MULTIPLIER`: increased from 1 to 2
* `MIN_SLASHING_PENALTY_QUOTIENT`: reduced from `2**7` (= 128) to `2**6` (= 64)

Note, these configuration values *move* toward maximum planned security, but do not yet reach those planned maximally punitive values with the plan to reach maximum punitivity at or before the merge (i.e. when user activity is secured by the beacon chain and there is a huge incentive to attack).

### DIFF
--- a/configs/mainnet/lightclient_patch.yaml
+++ b/configs/mainnet/lightclient_patch.yaml
@@ -2,6 +2,16 @@
 
 CONFIG_NAME: "mainnet"
 
+# Updated penalty values
+# ---------------------------------------------------------------
+# 3 * 2**24) (= 50,331,648)
+HF1_INACTIVITY_PENALTY_QUOTIENT: 50331648
+# 2**6 (= 64)
+HF1_MIN_SLASHING_PENALTY_QUOTIENT: 64
+# 2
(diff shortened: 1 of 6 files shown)


hwwhww (approved):
I pushed some commits to fix the tests.

The code correctness looks good to me. 👍 My approval stamp is for the correctness review.

Could you elaborate on the rationale of why the adjusted numbers are less harsh than the original numbers? e.g., why `PROPORTIONAL_SLASHING_MULTIPLIER` is `2` instead of `3`?

Issue #2191, opened 2021-01-26
[testgen] enable HF1 testgen and temporarily disable phase 1 testgen 


hwwhww:
If we are cutting a new spec release soon (e.g., in 1 week):
1. The light client patch in HF1 is more relatively ready for client teams to prototype.
2. The phase 1 tests are relatively useless until #2172 is merged.
    - It may also mitigate the file compression issue (#2186) a bit.

I think we should enable HF1 testgen. And want to confirm if (2) is true or not.

/cc @djrtwo

djrtwo:
Yes, let's do this.

PR #2199, opened 2021-02-15
[testgen] turn off phase 1, turn on HF1, and refactoring

hwwhww:
1. Address #2191
    - Turn off phase 1 test generators
    - Turn on lightclient_patch (HF1) test generators
2. Refactor the generators
    - Rewrite helper functions in `tests/core/gen_helpers/gen_from_tests/gen.py` and reduce the code in generator `main.py`.

TODO
- [x] run testgen with mainnet config locally.
- [x] update README.md files


tests/core/gen_helpers/gen_from_tests/gen.py
```diff
@@ -1,11 +1,17 @@
+from importlib import reload, import_module
 from inspect import getmembers, isfunction
-from typing import Any, Iterable
+from typing import Any, Callable, Dict, Iterable
 
-from gen_base.gen_typing import TestCase
+from eth2spec.config import config_util
+from eth2spec.utils import bls
+
+from eth2spec.test.context import ALL_CONFIGS, TESTGEN_FORKS, SpecForkName, ConfigName
```
hwwhww, line 8: Check with @protolambda: I remember you wanted `gen_helpers` to be isolated from `eth2spec`. But here I borrow the `eth2spec` helpers again. 😅 

`eth2spec` context was imported in `gen_helpers/gen_base/gen_runner.py` for skipping the specific tests:
https://github.com/ethereum/eth2.0-specs/blob/c7d975981c04626b72cdc3566914af0aa552b7bb/tests/core/gen_helpers/gen_base/gen_runner.py#L13-L14

This refactoring is for making `main.py` of the state test generators only need to set the handlers.

I want to check if you have a preference against this apporach. 🙂

protolambda: The pytest related test generators all import this package, and importing pyspec makes dependency management even more complicated. I'm afraid of situations where you run a generator, and outputs are incorrect due to a bad build order, or where a pyspec change breaks the generator more quickly. If we really need pyspec here, maybe it's better to change the generator to be a module of eth2spec package, and simplify the dependency management.

hwwhww: @protolambda thanks for the response!

https://github.com/ethereum/eth2.0-specs/pull/2199/commits/9cc8567d682ee831adf4a019de31fd0ebbdd30bb is a different approach (different from our chat):
- I moved `gen_helpers` into a module of the `eth2spec` package + some linter fixes.
- The dependencies of `generators/*` modules are only (i) the whole `eth2spec` package (ii) installing `pytest` since it's not part of the default `eth2spec` installation.
- I think now we can keep the generators as status-quo. (independent `venv`s).

Does it look good for you now?

protolambda (approved):
Looks good to me, nice work. I ran the test generators, and then ran the phase0 tests in ZRNT, and it all works. Just have to make sure nothing went missing.

Ignoring the SSZ tests (too many to list nicely):
```shell
tree -I "ssz_static|ssz_generic" -L 7 | xsel -bi
```
Output: https://gist.github.com/protolambda/fe8e490453ee580bd3e897ef22718ede

If that looks complete, this PR is ready :+1: 

P.S. it's still a draft PR? Let me know if there's more to review.

hwwhww:
Thanks @protolambda!

- I checked the outputs and fixed the missing shuffling `requirements.txt` update.
- Update the docs.

djrtwo (approved):

djrtwo:
nice refactor!

PR #2201, opened 2021-02-18
Add Altair validator guide and initial p2p spec

ralexstokes:
Introduces a "honest validator" guide for HF 1 and updates to the p2p spec.

status 3/10/21:
- [x] drop `SyncCommitteeIndex` in favor of `ValidatorIndex` at the cost of having aggregators do a bit more work (while saving bandwidth under committee membership duplications)
- [x] "HF1" and "lightclient" leftovers -> "Altair"
- [x] any remaining TODOs in document

status 3/16/21:
- [x] ENR update below
- [x] squash, get final review/merge

~plan is to merge along w/ #2244 and then can address subcommittee usage in subsequent PR~
~- [ ] leverage aggregated "subcommittees" in `BeaconState` to reduce size of `SyncCommitteeContribution`~
~- [ ] quantify bandwidth (avg.,peak) load under this aggregation scheme~

### DIFF
--- a/specs/altair/p2p-interface.md
+++ b/specs/altair/p2p-interface.md
@@ -1,0 +1,129 @@
+# Ethereum Altair networking specification
+
+This document contains the networking specification for Ethereum 2.0 clients added during the Altair deployment. 
+This document should be viewed as additive to the [document from Phase 0](../phase0/p2p-interface.md) and will be referred to as the "phase 0 document" hereafter. 
+Readers should understand the phase 0 document and use it as a basis to understand the changes outlined in this document.
+
+In particular, Altair introduces changes to the gossip domain and a small change to ENR advertisement in the discovery domain. 
+The phase 0 document should be consulted for specs on the network fundamentals, the existing gossip domain, the req/resp domain and the discovery domain.
+
+## Table of contents
+
+<!-- TOC -->
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+  - [Warning](#warning)
+- [Modifications in Altair](#modifications-in-altair)
+  - [The gossip domain: gossipsub](#the-gossip-domain-gossipsub)
+    - [Topics and messages](#topics-and-messages)
+      - [Global topics](#global-topics)
+        - [`beacon_block`](#beacon_block)
+        - [`sync_committee_contribution_and_proof`](#sync_committee_contribution_and_proof)
+      - [Sync committee subnets](#sync-committee-subnets)
+        - [`sync_committee_{subnet_id}`](#sync_committee_subnet_id)
+      - [Sync committees and aggregation](#sync-committees-and-aggregation)
+  - [The Req/Resp domain](#the-reqresp-domain)
+  - [The discovery domain: discv5](#the-discovery-domain-discv5)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+## Warning
+
+This document is currently illustrative for early Altair testnets and some parts are subject to change. 
+Refer to the note in the [validator guide](./validator.md) for further details. 
+
+# Modifications in Altair
+
+## The gossip domain: gossipsub
+
+Gossip meshes are added in Altair to support the consensus activities of the sync committees. 
+Validators use an aggregation scheme to balance the processing and networking load across all of the relevant actors.
+
+### Topics and messages
+
+Topics follow the same specification as in the phase 0 document. 
+New topics are added in Altair to support the sync committees and the beacon block topic is updated with the modified type.
+
+The specification around the creation, validation, and dissemination of messages has not changed from the Phase 0 document.
+
+The new topics along with the type of the `data` field of a gossipsub message are given in this table:
+
+| Name | Message Type |
+| - | - |
+| `beacon_block` | `SignedBeaconBlock` (modified) |
+| `sync_committee_contribution_and_proof` | `SignedContributionAndProof` |
+| `sync_committee_{subnet_id}` | `SyncCommitteeSignature` |
+
+Definitions of these new types can be found in the [Altair validator guide](./validator.md#containers).
+
+#### Global topics
+
+Altair changes the type of the global beacon block topic and adds one global topic to propagate partially aggregated sync committee signatures to all potential proposers of beacon blocks.
+
+##### `beacon_block`
+
+The existing specification for this topic does not change from the Phase 0 document, but the type of the payload does change to the (modified) `SignedBeaconBlock`. 
+This type changes due to the inclusion of the inner `BeaconBlockBody` that is modified in Altair. 
+See the [state transition document](./beacon-chain.md#beaconblockbody) for Altair for further details.
+
+##### `sync_committee_contribution_and_proof`
+ 
+This topic is used to propagate partially aggregated sync committee signatures to be included in future blocks.
+
+The following validations MUST pass before forwarding the `signed_contribution_and_proof` on the network; define `contribution_and_proof = signed_contribution_and_proof.message` and `contribution = contribution_and_proof.contribution` for convenience:
+
+- _[IGNORE]_ The contribution's slot is for the current slot, i.e. `contribution.slot == current_slot`.
+- _[IGNORE]_ The block being signed over (`contribution.beacon_block_root`) has been seen (via both gossip and non-gossip sources).
+- _[REJECT]_ The subcommittee index is in the allowed range, i.e. `contribution.subcommittee_index < SYNC_COMMITTEE_SUBNET_COUNT`.
+- _[IGNORE]_ The sync committee contribution is the first valid contribution received for the aggregator with index `contribution_and_proof.aggregator_index` for the slot `contribution.slot`.
+- _[REJECT]_ The aggregator's validator index is within the current sync committee --
+  i.e. `state.validators[aggregate_and_proof.aggregator_index].pubkey in state.current_sync_committee.pubkeys`.
+- _[REJECT]_ The `contribution_and_proof.selection_proof` is a valid signature of the `contribution.slot` by the validator with index `contribution_and_proof.aggregator_index`.
+- _[REJECT]_ `contribution_and_proof.selection_proof` selects the validator as an aggregator for the slot -- i.e. `is_sync_committee_aggregator(state, contribution.slot, contribution_and_proof.selection_proof)` returns `True`.
+- _[REJECT]_ The aggregator signature, `signed_contribution_and_proof.signature`, is valid.
+- _[REJECT]_ The aggregate signature is valid for the message `beacon_block_root` and aggregate pubkey derived from the participation info in `aggregation_bits` for the subcommittee specified by the `subcommittee_index`.
+
+#### Sync committee subnets
+
+Sync committee subnets are used to propagate unaggregated sync committee signatures to subsections of the network.
+
+##### `sync_committee_{subnet_id}`
+
+The `sync_committee_{subnet_id}` topics are used to propagate unaggregated sync committee signatures to the subnet `subnet_id` to be aggregated before being gossiped to the global `sync_committee_contribution_and_proof` topic.
+
+The following validations MUST pass before forwarding the `sync_committee_signature` on the network:
+
+- _[IGNORE]_ The signature's slot is for the current slot, i.e. `sync_committee_signature.slot == current_slot`.
+- _[IGNORE]_ The block being signed over (`sync_committee_signature.beacon_block_root`) has been seen (via both gossip and non-gossip sources).
+- _[IGNORE]_ There has been no other valid sync committee signature for the declared `slot` for the validator referenced by `sync_committee_signature.validator_index`.
+- _[REJECT]_ The validator producing this `sync_committee_signature` is in the current sync committee, i.e. `state.validators[sync_committee_signature.validator_index].pubkey in state.current_sync_committee.pubkeys`.
+- _[REJECT]_ The `subnet_id` is correct, i.e. `subnet_id in compute_subnets_for_sync_committee(state, sync_committee_signature.validator_index)`.
+- _[REJECT]_ The `signature` is valid for the message `beacon_block_root` for the validator referenced by `validator_index`.
+
+#### Sync committees and aggregation
+
+The aggregation scheme closely follows the design of the attestation aggregation scheme. 
+Sync committee signatures are broadcast into "subnets" defined by a topic. 
+The number of subnets is defined by `SYNC_COMMITTEE_SUBNET_COUNT` in the [Altair validator guide](./validator.md#constants). 
+Sync committee members are divided into "subcommittees" which are then assigned to a subnet for the duration of tenure in the sync committee. 
+Individual validators can be duplicated in the broader sync committee such that they are included multiple times in a given subcommittee or across multiple subcommittees.
+
+Unaggregated signatures (along with metadata) are sent as `SyncCommitteeSignature`s on the `sync_committee_{subnet_id}` topics.
+
+Aggregated sync committee signatures are packaged into (signed) `SyncCommitteeContribution` along with proofs and gossiped to the `sync_committee_contribution_and_proof` topic.
+
+## The Req/Resp domain
+
+* (TODO) specify how to handle different types of blocks on RPC responses.
+
+## The discovery domain: discv5
+
+The `attnets` key of the ENR is used as defined in the phase 0 document.
+
+An additional bitfield is added to the ENR under the key `syncnets` to facilitate sync committee subnet discovery.
+The length of this bitfield is `SYNC_COMMITTEE_SUBNET_COUNT` where each bit corresponds to a distinct `subnet_id` for a specific sync committee subnet. 
+The `i`th bit is set in this bitfield if the validator is currently subscribed to the `sync_committee_{i}` topic.
+
+See the [validator document](./validator.md#sync-committee-subnet-stability) for further details on how the new bits are used.
--- a/specs/altair/validator.md
+++ b/specs/altair/validator.md
@@ -1,0 +1,415 @@
+# Ethereum 2.0 Altair -- Honest Validator
+
+This is an accompanying document to [Ethereum 2.0 Altair -- The Beacon Chain](./beacon-chain.md), which describes the expected actions of a "validator" participating in the Ethereum 2.0 protocol.
+
+## Table of contents
+
+<!-- TOC -->
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Introduction](#introduction)
+- [Prerequisites](#prerequisites)
+- [Warning](#warning)
+- [Constants](#constants)
+  - [Misc](#misc)
+- [Containers](#containers)
+  - [`SyncCommitteeSignature`](#synccommitteesignature)
+  - [`SyncCommitteeContribution`](#synccommitteecontribution)
+  - [`ContributionAndProof`](#contributionandproof)
+  - [`SignedContributionAndProof`](#signedcontributionandproof)
+- [Validator assignments](#validator-assignments)
+  - [Sync Committee](#sync-committee)
+  - [Lookahead](#lookahead)
+- [Beacon chain responsibilities](#beacon-chain-responsibilities)
+  - [Block proposal](#block-proposal)
+    - [Preparing a `BeaconBlock`](#preparing-a-beaconblock)
+    - [Constructing the `BeaconBlockBody`](#constructing-the-beaconblockbody)
+      - [Sync committee](#sync-committee)
+    - [Packaging into a `SignedBeaconBlock`](#packaging-into-a-signedbeaconblock)
+  - [Attesting and attestation aggregation](#attesting-and-attestation-aggregation)
+  - [Sync committees](#sync-committees)
+    - [Sync committee signatures](#sync-committee-signatures)
+      - [Prepare sync committee signature](#prepare-sync-committee-signature)
+      - [Broadcast sync committee signature](#broadcast-sync-committee-signature)
+    - [Sync committee contributions](#sync-committee-contributions)
+      - [Aggregation selection](#aggregation-selection)
+      - [Construct sync committee contribution](#construct-sync-committee-contribution)
+        - [Slot](#slot)
+        - [Beacon block root](#beacon-block-root)
+        - [Subcommittee index](#subcommittee-index)
+        - [Aggregate bits](#aggregate-bits)
+        - [Signature](#signature)
+      - [Broadcast sync committee contribution](#broadcast-sync-committee-contribution)
+- [Sync committee subnet stability](#sync-committee-subnet-stability)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+## Introduction
+
+This document represents the expected behavior of an "honest validator" with respect to Altair of the Ethereum 2.0 protocol. 
+It builds on the [previous document for the behavior of an "honest validator" from Phase 0](../phase0/validator.md) of the Ethereum 2.0 protocol. 
+This previous document is referred to below as the "Phase 0 document".
+
+Altair introduces a new type of committee: the sync committee. Sync committees are responsible for signing each block of the canonical chain and there exists an efficient algorithm for light clients to sync the chain using the output of the sync committees. 
+See the [sync protocol](./sync-protocol.md) for further details on the light client sync. 
+Under this network upgrade, validators track their participation in this new committee type and produce the relevant signatures as required. 
+Block proposers incorporate the (aggregated) sync committee signatures into each block they produce.
+
+## Prerequisites
+
+All terminology, constants, functions, and protocol mechanics defined in the [Altair -- The Beacon Chain](./beacon-chain.md) doc are requisite for this document and used throughout. 
+Please see this document before continuing and use as a reference throughout.
+
+## Warning
+
+This document is currently illustrative for early Altair testnets and some parts are subject to change, especially pending implementation and profiling of Altair testnets.
+
+## Constants
+
+### Misc
+
+| Name | Value | Unit |
+| - | - | :-: |
+| `TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE` | `2**2` (= 4) | validators |
+| `SYNC_COMMITTEE_SUBNET_COUNT` | `8` | The number of sync committee subnets used in the gossipsub aggregation protocol. |
+
+## Containers
+
+### `SyncCommitteeSignature`
+
+```python
+class SyncCommitteeSignature(Container):
+    # Slot to which this contribution pertains
+    slot: Slot
+    # Block root for this signature
+    beacon_block_root: Root
+    # Index of the validator that produced this signature
+    validator_index: ValidatorIndex
+    # Signature by the validator over the block root of `slot`
+    signature: BLSSignature
+```
+
+### `SyncCommitteeContribution`
+
+```python
+class SyncCommitteeContribution(Container):
+    # Slot to which this contribution pertains
+    slot: Slot
+    # Block root for this contribution
+    beacon_block_root: Root
+    # The subcommittee this contribution pertains to out of the broader sync committee
+    subcommittee_index: uint64
+    # A bit is set if a signature from the validator at the corresponding
+    # index in the subcommittee is present in the aggregate `signature`.
+    aggregation_bits: Bitvector[SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT]
+    # Signature by the validator(s) over the block root of `slot`
+    signature: BLSSignature
+```
+
+### `ContributionAndProof`
+
+```python
+class ContributionAndProof(Container):
+    aggregator_index: ValidatorIndex
+    contribution: SyncCommitteeContribution
+    selection_proof: BLSSignature
+```
+
+### `SignedContributionAndProof`
+
+```python
+class SignedContributionAndProof(Container):
+    message: ContributionAndProof
+    signature: BLSSignature
+```
+
+## Validator assignments
+
+A validator determines beacon committee assignments and beacon block proposal duties as defined in the Phase 0 document.
+
+### Sync Committee
+
+To determine sync committee assignments, a validator can run the following function: `is_assigned_to_sync_committee(state, epoch, validator_index)` where `epoch` is an epoch number within the current or next sync committee period.
+This function is a predicate indicating the presence or absence of the validator in the corresponding sync committee for the queried sync committee period.
+
+```python
+def compute_sync_committee_period(epoch: Epoch) -> uint64:
+    return epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+
+def is_assigned_to_sync_committee(state: BeaconState,
+                                  epoch: Epoch,
+                                  validator_index: ValidatorIndex) -> bool:
+    sync_committee_period = compute_sync_committee_period(epoch)
+    current_epoch = get_current_epoch(state)
+    current_sync_committee_period = compute_sync_committee_period(current_epoch)
+    next_sync_committee_period = current_sync_committee_period + 1
+    assert sync_committee_period in (current_sync_committee_period, next_sync_committee_period)
+
+    pubkey = state.validators[validator_index].pubkey
+    if sync_committee_period == current_sync_committee_period:
+        return pubkey in state.current_sync_committee.pubkeys
+    else:  # sync_committee_period == next_sync_committee_period
+        return pubkey in state.next_sync_committee.pubkeys
+```
+
+### Lookahead
+
+The sync committee shufflings give validators 1 sync committee period of lookahead which amounts to `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` epochs.
+At any given `epoch`, the `BeaconState` contains the current `SyncCommittee` and the next `SyncCommittee`. 
+Once every `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` epochs, the next `SyncCommittee` becomes the current `SyncCommittee` and the next committee is computed and stored.
+
+*Note*: The data required to compute a given committee is not cached in the `BeaconState` after committees are calculated at the period boundaries. 
+This means that calling `get_sync_commitee()` in a given `epoch` can return a different result than what was computed during the relevant epoch transition. 
+For this reason, *always* get committee assignments via the fields of the `BeaconState` (`current_sync_committee` and `next_sync_committee`) or use the above reference code.
+
+A validator should plan for future sync committee assignments by noting which sync committee periods they are selected for participation.
+Specifically, a validator should:
+* Upon (re)syncing the chain and upon sync committee period boundaries, check for assignments in the current and next sync committee periods.
+* If the validator is in the current sync committee period, they can perform the responsibilities below for sync committee rewards.
+* If the validator is in the next sync committee period, they should wait until the next `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` boundary and then perform the responsibilities throughout that period.
+
+## Beacon chain responsibilities
+
+A validator maintains the responsibilities given in the Phase 0 document.
+
+Block proposals are modified to incorporate the sync committee signatures as detailed below.
+
+When assigned to a sync committee, validators have a new responsibility to sign beacon block roots during each slot of the sync committee period.
+These signatures are aggregated and routed to the proposer over gossip for inclusion into a beacon block. 
+Assignments to a particular sync committee are infrequent at normal validator counts; however, an action every slot is required when in the current active sync committee.
+
+### Block proposal
+
+Refer to the phase 0 document for the majority of the [block proposal responsibility](../phase0/validator.md#block-proposal).
+The validator should follow those instructions to prepare a `SignedBeaconBlock` for inclusion into the chain. All changes are additive to phase 0 and noted below.
+
+#### Preparing a `BeaconBlock`
+
+No change to [Preparing for a `BeaconBlock`](../phase0/validator.md#preparing-for-a-beaconblock).
+
+#### Constructing the `BeaconBlockBody`
+
+Each section of [Constructing the `BeaconBlockBody`](../phase0/validator.md#constructing-the-beaconblockbody) should be followed. 
+After constructing the `BeaconBlockBody` as per that section, the proposer has an additional task to include the sync committee signatures:
+
+##### Sync committee
+
+The proposer receives a number of `SyncCommitteeContribution`s (wrapped in `SignedContributionAndProof`s on the wire) from validators in the sync committee who are selected to partially aggregate signatures from independent subcommittees formed by breaking the full sync committee into `SYNC_COMMITTEE_SUBNET_COUNT` pieces (see below for details).
+
+The proposer collects these contributions for further aggregation when preparing a block. 
+Proposers should select the best contribution seen across all aggregators for each subnet/subcommittee when preparing a block. 
+A contribution with more valid signatures is better than a contribution with fewer signatures.
+
+Recall `block.body.sync_aggregate.sync_committee_bits` is a `Bitvector` where the `i`th bit is `True` if the corresponding validator in the sync committee has produced a valid signature, 
+and that `block.body.sync_aggregate.sync_committee_signature` is the aggregate BLS signature combining all of the valid signatures.
+
+Given a collection of the best seen `contributions` (with no repeating `subcommittee_index` values) and the `BeaconBlock` under construction, 
+the proposer processes them as follows:
+
+```python
+def process_sync_committee_contributions(state: BeaconState, 
+                                         block: BeaconBlock, 
+                                         contributions: Set[SyncCommitteeContribution]) -> None:
+    sync_aggregate = SyncAggregate()
+    signatures = []
+
+    for contribution in contributions:
+        subcommittee_index = contribution.subcommittee_index
+        for index, participated in enumerate(contribution.aggregation_bits):
+            if participated:
+                participant_index = SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT * subcommittee_index + index
+                sync_aggregate.sync_committee_bits[participant_index] = True
+        signatures.append(contribution.signature)
+
+    sync_aggregate.sync_committee_signature = bls.Aggregate(signatures)
+
+    block.body.sync_aggregate = sync_aggregate
+```
+
+*Note*: The resulting block must pass the validations for the `SyncAggregate` defined in `process_sync_committee` defined in the [state transition document](./beacon-chain.md#sync-committee-processing).
+In particular, this means `SyncCommitteeContribution`s received from gossip must have a `beacon_block_root` that matches the proposer's local view of the chain.
+
+#### Packaging into a `SignedBeaconBlock`
+
+No change to [Packaging into a `SignedBeaconBlock`](../phase0/validator.md#packaging-into-a-signedbeaconblock).
+
+### Attesting and attestation aggregation
+
+Refer to the phase 0 document for the [attesting](../phase0/validator.md#attesting) and [attestation aggregation](../phase0/validator.md#attestation-aggregation) responsibilities. 
+There is no change compared to the phase 0 document.
+
+### Sync committees
+
+Sync committee members employ an aggregation scheme to reduce load on the global proposer channel that is monitored by all potential proposers to be able to include the full output of the sync committee every slot. 
+Sync committee members produce individual signatures on subnets (similar to the attestation subnets) via `SyncCommitteeSignature`s which are then collected by aggregators sampled from the sync subcommittees to produce a `SyncCommitteeContribution` which is gossiped to proposers. 
+This process occurs each slot.
+
+#### Sync committee signatures
+
+##### Prepare sync committee signature
+
+If a validator is in the current sync committee (i.e. `is_assigned_to_sync_committee` above returns `True`), then for every slot in the current sync committee period the validator should prepare a `SyncCommitteeSignature` according to the logic in `get_sync_committee_signature` as soon as they have determined the head block of the current slot.
+
+This logic is triggered upon the same conditions as when producing an attestation. 
+Meaning, a sync committee member should produce and broadcast a `SyncCommitteeSignature` either when (a) the validator has received a valid block from the expected block proposer for the current `slot` or (b) one-third of the slot has transpired (`SECONDS_PER_SLOT / 3` seconds after the start of the slot) -- whichever comes first.
+
+`get_sync_committee_signature` assumes `state` is the head state corresponding to processing the block at the current slot as determined by the fork choice (including any empty slots processed with `process_slots`), `validator_index` is the index of the validator in the registry `state.validators` controlled by `privkey`, and `privkey` is the BLS private key for the validator.
+
+```python
+def get_sync_committee_signature(state: BeaconState, 
+                                 validator_index: ValidatorIndex, 
+                                 privkey: int) -> SyncCommitteeSignature:
+    epoch = get_current_epoch(state)
+    domain = get_domain(state, DOMAIN_SYNC_COMMITTEE, epoch)
+    signing_root = compute_signing_root(get_block_root_at_slot(state, state.slot), domain)
+    signature = bls.Sign(privkey, signing_root)
+
+    return SyncCommitteeSignature(slot=state.slot, validator_index=validator_index, signature=signature)
+```
+
+##### Broadcast sync committee signature
+
+The validator broadcasts the assembled signature to the assigned subnet, the `sync_committee_{subnet_id}` pubsub topic.
+
+The `subnet_id` is derived from the position in the sync committee such that the sync committee is divided into "subcommittees".
+It can be computed via `compute_subnets_for_sync_committee` where `state` is a `BeaconState` during the matching sync committee period. 
+This function returns multiple subnets if a given validator index is included multiple times in a given sync committee across multiple subcommittees.
+
+```python
+def compute_subnets_for_sync_committee(state: BeaconState, validator_index: ValidatorIndex) -> Sequence[uint64]:
+    target_pubkey = state.validators[validator_index].pubkey
+    sync_committee_indices = [index for index, pubkey in enumerate(state.current_sync_committee.pubkeys) if pubkey == target_pubkey]
+    return [
+        uint64(index // (SYNC_COMMITEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT))
+        for index in sync_committee_indices
+    ]
+```
+
+*Note*: Subnet assignment does not change during the duration of a validator's assignment to a given sync committee.
+
+*Note*: If a validator has multiple `subnet_id` results from `compute_subnets_for_sync_committee`, the validator should broadcast a copy of the `sync_committee_signature` on each of the distinct subnets.
+
+#### Sync committee contributions
+
+Each slot some sync committee members in each subcommittee are selected to aggregate the `SyncCommitteeSignature`s into a `SyncCommitteeContribution` which is broadcast on a global channel for inclusion into the next block.
+
+##### Aggregation selection
+
+A validator is selected to aggregate based on the computation in `is_sync_committee_aggregator` where `state` is a `BeaconState` as supplied to `get_sync_committee_slot_signature` and `signature` is the BLS signature returned by `get_sync_committee_slot_signature`. 
+The signature function takes a `BeaconState` with the relevant sync committees for the queried `slot` (i.e. `state.slot` is within the span covered by the current or next sync committee period), the `subcommittee_index` equal to the `subnet_id`, and the `privkey` is the BLS private key associated with the validator.
+
+```python
+class SyncCommitteeSigningData(Container):
+    slot: Slot
+    subcommittee_index: uint64
+```
+
+```python
+def get_sync_committee_slot_signature(state: BeaconState, slot: Slot, subcommittee_index: uint64, privkey: int) -> BLSSignature:
+    domain = get_domain(state, DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF, compute_epoch_at_slot(slot))
+    signing_data = SyncCommitteeSigningData(
+        slot=slot,
+        subcommittee_index=subcommittee_index,
+    )
+    signing_root = compute_signing_root(signing_data, domain)
+    return bls.Sign(privkey, signing_root)
+```
+
+```python
+def is_sync_committee_aggregator(state: BeaconState, signature: BLSSignature) -> bool:
+    modulo = max(1, SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT // TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE)
+    return bytes_to_uint64(hash(signature)[0:8]) % modulo == 0
+```
+
+*NOTE*: the set of aggregators generally changes every slot; however, the assignments can be computed ahead of time as soon as the committee is known
+
+##### Construct sync committee contribution
+
+If a validator is selected to aggregate the `SyncCommitteeSignature`s produced on a subnet during a given `slot`, they construct an aggregated `SyncCommitteeContribution`. 
+
+Given all of the (valid) collected `sync_committee_signatures: Set[SyncCommitteeSignature]` from the `sync_committee_{subnet_id}` gossip during the selected `slot` with an equivalent `beacon_block_root` to that of the aggregator, the aggregator creates a `contribution: SyncCommitteeContribution` with the following fields:
+
+###### Slot
+
+Set `contribution.slot = state.slot` where `state` is the `BeaconState` for the slot in question.
+
+###### Beacon block root
+
+Set `contribution.beacon_block_root = beacon_block_root` from the `beacon_block_root` found in the `sync_committee_signatures`.
+
+###### Subcommittee index
+
+Set `contribution.subcommittee_index` to the index for the subcommittee index corresponding to the subcommittee assigned to this subnet. This index matches the `subnet_id` used to derive the topic name.
+
+###### Aggregate bits
+
+Let `contribution.aggregation_bits` be a `Bitvector[SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT]`, where the `index`th bit is set in the `Bitvector` for each corresponding validator included in this aggregate from the corresponding subcommittee.
+An aggregator needs to find the index in the sync committee (as returned by `get_sync_committee_indices`) for a given validator referenced by `sync_committee_signature.validator_index` and map the sync committee index to an index in the subcommittee (along with the prior `subcommittee_index`). This index within the subcommittee is the one set in the `Bitvector`.
+For example, a validator with index `2044` could be at index `15` in the current sync committee. This sync committee index maps to `subcommittee_index` `1` with position `7` in the `Bitvector` for the contribution. 
+Also note that a validator **could be included multiple times** in a given subcommittee such that multiple bits are set for a single `SyncCommitteeSignature`.
+
+###### Signature
+
+Set `contribution.signature = aggregate_signature` where `aggregate_signature` is obtained by assembling the appropriate collection of `BLSSignature`s from the set of `sync_committee_signatures` and using the `bls.Aggregate` function to produce an aggregate `BLSSignature`. 
+The collection of input signatures should include one signature per validator who had a bit set in the `aggregation_bits` bitfield, with repeated signatures if one validator maps to multiple indices within the subcommittee.
+
+##### Broadcast sync committee contribution
+
+If the validator is selected to aggregate (`is_sync_committee_aggregator`), then they broadcast their best aggregate as a `SignedContributionAndProof` to the global aggregate channel (`sync_committee_contribution_and_proof` topic) two-thirds of the way through the `slot`-that is, `SECONDS_PER_SLOT * 2 / 3` seconds after the start of `slot`.
+
+Selection proofs are provided in `ContributionAndProof` to prove to the gossip channel that the validator has been selected as an aggregator.
+
+`ContributionAndProof` messages are signed by the aggregator and broadcast inside of `SignedContributionAndProof` objects to prevent a class of DoS attacks and message forgeries.
+
+First, `contribution_and_proof = get_contribution_and_proof(state, validator_index, contribution, privkey)` is constructed.
+
+```python
+def get_contribution_and_proof(state: BeaconState,
+                               aggregator_index: ValidatorIndex,
+                               contribution: SyncCommitteeContribution,
+                               privkey: int) -> ContributionAndProof:
+    selection_proof = get_sync_committee_slot_signature(
+        state,
+        contribution.slot,
+        contribution.subcommittee_index,
+        privkey,
+    )
+    return ContributionAndProof(
+        aggregator_index=aggregator_index,
+        contribution=contribution,
+        selection_proof=selection_proof,
+    )
+```
+
+Then `signed_contribution_and_proof = SignedContributionAndProof(message=contribution_and_proof, signature=signature)` is constructed and broadast. Where `signature` is obtained from:
+
+```python
+def get_contribution_and_proof_signature(state: BeaconState, 
+                                         contribution_and_proof: ContributionAndProof, 
+                                         privkey: int) -> BLSSignature:
+    contribution = contribution_and_proof.contribution
+    domain = get_domain(state, DOMAIN_CONTRIBUTION_AND_PROOF, compute_epoch_at_slot(contribution.slot))
+    signing_root = compute_signing_root(contribution_and_proof, domain)
+    return bls.Sign(privkey, signing_root)
+```
+
+## Sync committee subnet stability
+
+The sync committee subnets need special care to ensure stability given the relatively low number of validators involved in the sync committee at any particular time. 
+To provide this stability, a validator must do the following:
+
+* Maintain advertisement of the subnet the validator in the sync committee is assigned to in their node's ENR as soon as they have joined the subnet. 
+Subnet assignments are known `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` epochs in advance and can be computed with `compute_subnets_for_sync_committee` defined above. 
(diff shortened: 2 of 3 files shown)


specs/lightclient/validator.md
```diff
@@ -0,0 +1,198 @@
...
+
+### SyncCommitteeContribution
+
+```python
+class SyncCommitteeContribution(Container):
+    # Slot during which this contribution is broadcast
+    slot: Slot
+    # Position of the validator within the sync committee
+    index: SyncCommitteeIndex
+    # Signature by the validator over the block root of the _previous_ slot
```
terencechain, line 57: Are we using the _previous_ slot opposed to the _current_ slot for better stability?

ralexstokes: that falls out as a side effect but i'd say the bigger reason is just that we use blockspace to include the sync committee's proof about a given block -- if we signed over the block root of the current slot and then included that data into the current block, it would change the current block root so we are back to square one.

terencechain: Ah i see. I thought it would have worked like attestation

slot N at 0: submit block (package attestations from N-1)
slot N at 1/3: submit attestations and sync committee objects

slot N+1 at 0:  submit block (package attestations and sync committee object from N)

So sync_committee object for current slot is ok because it's getting included in the next slot, this is similar to attestation handling today

ralexstokes:
update: after some offline discussion, it seems to make the most sense to move ahead w/ an aggregation scheme for the `SyncCommitteeContribution`s like we do for `Attestations` as the unaggregated version introduces an extra ~100kB per slot for _all_ validators (as any individual validator would need to be listening to this channel all the time given the minimal amount of lookahead on the proposer shuffling)

specs/lightclient/validator.md
djrtwo: Ah, I think y'all are saying the same thing.

If a sync committee is assigned to slot `N`, they vote on what they see the block_root of slot `N` as. But this is then *included* in the block at slot `N+1`. So they vote on the current/assigned slot but when this makes it on chain it is a vote on the previous slot (same as an optimal attestation)

specs/lightclient/validator.md
```diff
@@ -0,0 +1,198 @@
...
+## Containers
+
+### SyncCommitteeContribution
+
+```python
+class SyncCommitteeContribution(Container):
+    # Slot during which this contribution is broadcast
+    slot: Slot
+    # Position of the validator within the sync committee
+    index: SyncCommitteeIndex
```
djrtwo, line 56: Pros of defining new type -- `SyncCommitteeIndex` -- instead of just using a `ValidatorIndex`. Presumable if you can verify this signature, you have map of validators <-> sync_committee_shuffling so both should be approximately equivalent

With `SyncCommitteIndex`, you do a lookup with the shuffling to get the validator index and verify the signature
With `ValidatorIndex`, you can verify signature direct, and need to make sure the validator index is contained within the current shuffling

specs/lightclient/validator.md
djrtwo: So I am looking at the object now and it seems that `slot` and the slot of `block` that is being signed over should be the same, whereas the comments imply otherwise. Essentially, at start of slot `block` is broadcast. Halfway through slot, sync committee votes. Next slot (`slot+1`) next block producer includes any sync committee votes from previous slot

specs/lightclient/validator.md
```diff
@@ -0,0 +1,198 @@
...
+    # Signature by the validator over the block root of the _previous_ slot
+    signature: BLSSignature
+```
+
+## Validator assignments
+
+A validator determines attestation committee assignments and beacon block proposal duties as defined in the Phase 0 document.
```
djrtwo, line 63:
```suggestion
A validator determines attestation committee assignments and beacon block proposal duties as defined in the Phase 0 document.
```

specs/lightclient/validator.md
```diff
@@ -0,0 +1,198 @@
...
+
+To determine sync committee assignments, a validator can run the following function: `get_sync_committee_assignment(state, epoch, validator_index)` where `epoch` is an epoch number within the current or next sync committee period. This function will return a tuple of `SyncCommitteeIndex` for each position in the sync committee that the pubkey corresponding to `validator_index` appears. If the validator has no assignments in the corresponding sync committees, this tuple will be empty `()`.
```
djrtwo, line 65: Could use a sub-header "### Sync Committee"

```suggestion
### Sync Committee

To determine sync committee assignments, a validator can run the following function: `get_sync_committee_assignment(state, epoch, validator_index)` where `epoch` is an epoch number within the current or next sync committee period. This function will return a tuple of `SyncCommitteeIndex` for each position in the sync committee that the pubkey corresponding to `validator_index` appears. If the validator has no assignments in the corresponding sync committees, this tuple will be empty `()`. 
```

specs/lightclient/validator.md
djrtwo: Ah, I see. The idea here is that if a validator is duplicated in the shuffling they would send two messages.

I'm not sure it's worth the new type here and duplicate message on network. The client can resolve a duplicate assignment by including the signature twice in the aggregate.

That said, I see *why* this is here now. open to discuss more

specs/lightclient/validator.md
```diff
@@ -0,0 +1,198 @@
...
+*Note*:  A single validator may be duplicated in a given sync committee.
+
+```python
+def compute_sync_committee_period(epoch: Epoch) -> uint64:
+    return epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+
+def indices_in_sync_committee(sync_committee: SyncCommittee, target_pubkey: BLSPubkey) -> Tuple[SyncCommitteeIndex, ...]:
+    return tuple(SyncCommitteeIndex(index) for index, pubkey in enumerate(sync_committee.pubkeys) if pubkey == target_pubkey)
+
+def get_sync_committee_assignment(state: BeaconState,
```
djrtwo, line 78: If we remove `SyncCommitteeIndex` and instead put this on the aggregator to include duplicates, this can just return a boolean. `return pubkey in state.next_sync_committee`

specs/lightclient/validator.md
```diff
@@ -0,0 +1,198 @@
...
+
+No change to [Preparing for a `BeaconBlock`](../phase0/validator.md#preparing-for-a-beaconblock).
+
+#### Constructing the `BeaconBlockBody`
+
+Each section of [Constructing the `BeaconBlockBody`](../phase0/validator.md#constructing-the-beaconblockbody) should be followed. After constructing the `BeaconBlockBody` as per that section, the proposer gains an additional task to include the sync committee signatures:
+
+##### Sync committee
+
+The proposer will receive a number of `SyncCommitteeContribution`s from validators in the sync committee who are able to broadcast their signature. The proposer will collect all of these contributions for aggregation when preparing a block. `block.body.sync_committee_bits` is a `Bitvector` where the `i`th bit is `True` if the corresponding validator in the sync committee has produced a valid signature. `block.body.sync_committee_signature` is the aggregate BLS signature combining all the valid contributions. Given a collection of `contributions` and the `BeaconBlock` under construction, the proposer should process them as follows:
```
djrtwo, line 125:
```suggestion
The proposer receives a number of `SyncCommitteeContribution`s from validators in the sync committee who are able to broadcast their signature. The proposer then collects all of these contributions for aggregation when preparing a block. `block.body.sync_committee_bits` is a `Bitvector` where the `i`th bit is `True` if the corresponding validator in the sync committee has produced a valid signature. `block.body.sync_committee_signature` is the aggregate BLS signature combining all the valid contributions. Given a collection of `contributions` and the `BeaconBlock` under construction, the proposer should process them as follows:
```

specs/lightclient/validator.md
```diff
@@ -0,0 +1,198 @@
...
+    current_epoch = get_current_epoch(state)
+    current_sync_committee_period = compute_sync_committee_period(current_epoch)
+    next_sync_committee_period = current_sync_committee_period + 1
+    assert sync_committee_period in (current_sync_committee_period, next_sync_committee_period)
+
+    pubkey = state.validators[validator_index].pubkey
+    if sync_committee_period == current_sync_committee_period:
+        return indices_in_sync_committee(state.current_sync_committee, pubkey)
+    else:  # sync_committee_period == next_sync_committee_period
+        return indices_in_sync_committee(state.next_sync_committee pubkey)
```
djrtwo, line 91:
```suggestion
        return indices_in_sync_committee(state.next_sync_committee, pubkey)
```

specs/lightclient/validator.md
djrtwo, line 125: If we remove `SyncCommitteeIndex` from the unaggregated message on the wire, the proposer would need to call `indices_in_sync_committee` and do any (maybe duplicate) inclusions for message they have received

specs/lightclient/validator.md
```diff
@@ -0,0 +1,198 @@
...
+#### Constructing the `BeaconBlockBody`
+
+Each section of [Constructing the `BeaconBlockBody`](../phase0/validator.md#constructing-the-beaconblockbody) should be followed. After constructing the `BeaconBlockBody` as per that section, the proposer gains an additional task to include the sync committee signatures:
+
+##### Sync committee
+
+The proposer will receive a number of `SyncCommitteeContribution`s from validators in the sync committee who are able to broadcast their signature. The proposer will collect all of these contributions for aggregation when preparing a block. `block.body.sync_committee_bits` is a `Bitvector` where the `i`th bit is `True` if the corresponding validator in the sync committee has produced a valid signature. `block.body.sync_committee_signature` is the aggregate BLS signature combining all the valid contributions. Given a collection of `contributions` and the `BeaconBlock` under construction, the proposer should process them as follows:
+
+```python
+def process_sync_committee_contributions(state: BeaconState, block: BeaconBlock, contributions: Sequence[SyncCommitteeContribution]) -> None:
```
djrtwo, line 128: Line length

djrtwo: Probably a `Set` instead of a `Sequence`

djrtwo (commented):
Some minor comments before the aggregation refactor

specs/lightclient/validator.md
ralexstokes: did you have a suggestion here?

specs/lightclient/validator.md
ralexstokes: right, the comments are confusing -- i will update

specs/lightclient/validator.md
djrtwo: nope

specs/lightclient/validator.md
djrtwo: Yeah, let's do that. It's how we assign `slot` to `Attestation` so fits our mental model already

specs/lightclient/validator.md
```diff
@@ -0,0 +1,375 @@
...
+
+### SyncCommitteeContribution
+
+```python
+class SyncCommitteeContribution(Container):
+    # Slot to which this contribution pertains
+    slot: Slot
+    # A bit is set if a signature from the validator at the corresponding
+    # index in the sync committee is present in the aggregate `signature`.
+    aggregate_bits: Bitvector[SYNC_COMMITTEE_SIZE]
```
ralexstokes, line 92: with the current scheme, only `1/SYNC_COMMITTEE_SUBNET_COUNT` of the bits will ever be set which at expected numbers is somewhat wasteful... we can further optimize this message so that there is no waste at the expense of an additional index (likely `uint64`) to indicate which "subcommittee" this message corresponds to.

i'm leaning towards omitting this as it would complicate the code further but open to hearing differing opinions on it

specs/lightclient/validator.md
```diff
@@ -0,0 +1,375 @@
...
+        # is contribution valid?
+        assert contribution.slot == previous_slot
+        assert bls.FastAggregateVerify(participants, signing_root, contribution.signature)
+
+        signatures.append(contribution.signature)
+
+    body.sync_committee_signature = bls.Aggregate(signatures)
+```
+
+*NOTE*: (TODO) likely to update with more context on subcommittee messages so we can skip some of the ~quadratic work here and save on "large" sparse messages
```
ralexstokes, line 206: see this: https://github.com/ethereum/eth2.0-specs/pull/2201/files#r583100236

specs/lightclient/validator.md
```diff
@@ -0,0 +1,377 @@
...
+| - | - | - |
+| `SyncCommitteeIndex` | `uint64` | Index of a validator within a sync committee |
+
+## Constants
+
+### Misc
+
+| Name | Value | Unit |
+| - | - | :-: |
+| `TARGET_AGGREGATORS_PER_SYNC_COMMITTEE` | `2**2` (= 4) | validators |
```
ralexstokes, line 66: could argue this should be lower, say 2, but then you have very little redundancy in what is already a small subnet...

specs/lightclient/validator.md
```diff
@@ -0,0 +1,377 @@
...
+  privkey: int) -> BLSSignature:
+    contribution = contribution_and_proof.contribution
+    domain = get_domain(state, DOMAIN_CONTRIBUTION_AND_PROOF, compute_epoch_at_slot(contribution.slot))
+    signing_root = compute_signing_root(contribution_and_proof, domain)
+    return bls.Sign(privkey, signing_root)
+```
+
+## Sync committee subnet stability
+
+- TODO: detail stability strategy, main thing is sharpness of transition...
```
ralexstokes, line 377: currently thinking to spec the next sync committee to join (all at once) some epochs before the end of the current sync committee period

specs/lightclient/validator.md
```diff
@@ -0,0 +1,383 @@
...
+    domain = get_domain(state, DOMAIN_CONTRIBUTION_AND_PROOF, compute_epoch_at_slot(contribution.slot))
+    signing_root = compute_signing_root(contribution_and_proof, domain)
+    return bls.Sign(privkey, signing_root)
+```
+
+## Sync committee subnet stability
+
+The sync committee subnets need special care to ensure stability given the relatively low number of validators involved in the sync committee at any particular time. To provide this stability, a validator must do the following:
+
+* Maintain advertisement of the subnet the validator in the sync committee is assigned to in their node's ENR as soon as they have joined the subnet. Subnet assignments are known `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` epochs in advance and can be computed with `compute_subnet_for_sync_committee` defined above. ENR advertisement is indicated by setting the `syncnets` entry in the ENR to a `uint8` encoding of the `subnet_id`, e.g. `ENR["syncnets"] = b'\x07'`. This entry is removed from the ENR after the validator has left the sync committee.
```
ralexstokes, line 379: a consideration here is that ENRs are very space limited but we also don't have a better way to advertise sync committee subscriptions at the moment...

specs/lightclient/validator.md
```diff
@@ -0,0 +1,383 @@
...
+
+- TODO: what happens for the first sync committee?
```
ralexstokes, line 381: may need to advertise as soon as you know which subnet you are in (so 1 sync committee period of lookahead) in order to give validating nodes time to find peers... seems like you can still join just ahead of the committee change but i'd have to revisit gossipsub to make sure this is feasible

specs/lightclient/validator.md
```diff
@@ -0,0 +1,383 @@
...
+
+* To join a sync committee subnet, select a random number of epochs before the end of the current sync committee period between 1 and `SYNC_COMMITTEE_SUBNET_COUNT`, inclusive. Validators should join their member subnet at the beginning of the epoch they have randomly selected. For example, if the next sync committee period starts at epoch `853,248` and the validator randomly selects an offset of `3`, they should join the subnet at the beginning of epoch `853,245`.
```
ralexstokes, line 383: the thinking here is to smooth out the "sharpness" of the transition of having one chunk of the sync committee suddenly join a subnet. an alternative that may actually be fine (and is simpler to implement) is to just have the entire next sync committee join their respective subnet at a fixed epoch ahead of time, e.g. `boundary_epoch` - 8 epochs

specs/lightclient/p2p-interface.md
```diff
@@ -0,0 +1,76 @@
...
+
+- [Additions in HF 1](#additions-in-hf-1)
+  - [The gossip domain: gossipsub](#the-gossip-domain-gossipsub)
+    - [Topics and messages](#topics-and-messages)
+      - [Global topics](#global-topics)
+        - [`sync_committee_contribution_and_proof`](#sync_committee_contribution_and_proof)
+      - [Sync committee subnets](#sync-committee-subnets)
+        - [`sync_committee_{subnet_id}`](#sync_committee_subnet_id)
+      - [Sync committees and aggregation](#sync-committees-and-aggregation)
+
```
ralexstokes, line 21: needs section on sync committee subnet advertisement in the ENR

specs/lightclient/validator.md
```diff
@@ -0,0 +1,383 @@
...
+
+### Attesting and attestation aggregation
+
+Refer to the phase 0 document for the [attesting](../phase0/validator.md#attesting) and [attestation aggregation](../phase0/validator.md#attestation-aggregation) responsibilities. There is no change compared to the phase 0 document.
+
+### Sync committees
+
+Sync committee members employ an aggregation scheme to reduce load on the proposer who is expected to include the full output of the committee every slot. Sync committee members produce individual signatures on subnets (like the attestation subnets via `SyncCommitteeSignatures` which are then collected by aggregators sampled from the sync committee to produce a `SyncCommitteeContribution` which is gossiped to proposers. This process occurs each slot.
+
+#### Sync committee signatures
```
djrtwo, line 224: STOPPED REVIEW HERE

djrtwo (commented):
did a pass up until "STOPPED REVIEW HERE"

will pick it back up in the morning

specs/lightclient/p2p-interface.md
```diff
@@ -0,0 +1,76 @@
...
+# Additions in HF 1
+
+## The gossip domain: gossipsub
+
+Gossip meshes are added in HF 1 to support the consensus activities of the sync committees. Validators use an aggregation scheme to smooth some of the processing and networking load across all of the relevant actors.
+
+### Topics and messages
+
+Topics follow the same specification as in the phase 0 document. The only change are the names of new gossipsub topics added in HF 1.
```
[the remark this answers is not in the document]
ralexstokes: thanks for proofreading :)

specs/lightclient/validator.md
```diff
@@ -0,0 +1,383 @@
...
+| - | - | - |
+| `SyncCommitteeIndex` | `uint64` | Index of a validator within a sync committee |
+
+## Constants
+
+### Misc
+
+| Name | Value | Unit |
+| - | - | :-: |
+| `TARGET_AGGREGATORS_PER_SYNC_COMMITTEE` | `2**2` (= 4) | validators |
```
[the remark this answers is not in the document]
ralexstokes: agreed, this was pulled from the attestation aggregation spec and didn't think too much about the name. agree we can update it for clarity

specs/lightclient/validator.md
```diff
@@ -0,0 +1,383 @@
...
+        # is contribution valid?
+        assert contribution.slot == previous_slot
+        assert bls.FastAggregateVerify(participants, signing_root, contribution.signature)
+
+        signatures.append(contribution.signature)
+
+    body.sync_committee_signature = bls.Aggregate(signatures)
+```
+
+*NOTE*: (TODO) update w/ `SignedContributionAndProof` pending which validations go at gossip layer and which at application layer
```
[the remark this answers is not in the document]
ralexstokes: great, i'll be sure to clean up before merge

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+*NOTE*: (TODO) revisit subnet algo -- may be a bit more efficient to make "contiguous" "subcommittees"
+*NOTE*: (TODO) details of when to join subnets
+
+#### Sync committee contributions
+
+Each slot some sync committee members are selected to aggregate the `SyncCommitteeSignature`s into a `SyncCommitteeContribution` which is broadcast on a global channel for inclusion into the next block.
+
+##### Aggregation selection
+
+A validator is selected to aggregate based on the computation in `is_sync_committee_aggregator` where `state` is a `BeaconState` with the relevant sync committees for the queried `slot`, i.e. `state.slot` is within the span covered by the current or next sync committee period. `privkey` is the BLS private key associated with the validator.
```
djrtwo, line 288:
```suggestion
A validator is selected to aggregate based on the computation in `is_sync_committee_aggregator` where `state` is a `BeaconState` with the relevant sync committees for the queried `slot`, i.e. `state.slot`, is within the span covered by the current or next sync committee period. `privkey` is the BLS private key associated with the validator.
```

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+
+```python
+def get_sync_committee_slot_signature(state: BeaconState, slot: Slot, privkey: int) -> BLSSignature:
+    domain = get_domain(
```
djrtwo, line 292: prefer one liner when possible (only 98 chars of the 120 limit)

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+      state, 
+      DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF, 
+      compute_epoch_at_slot(slot)
+    )
+    signing_root = compute_signing_root(slot, domain)
+    return bls.Sign(privkey, signing_root)
+
```
djrtwo, line 299: separate code blocks (same formatting as the code blocks in [attestation aggregation selection](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/validator.md#aggregation-selection)

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+def is_sync_committee_aggregator(state: BeaconState, slot: Slot, privkey: int) -> bool:
+    signature = get_sync_committee_slot_signature(state, slot, privkey)
+    modulo = max(
```
djrtwo, line 302: I think one liner

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+      1, 
+      SYNC_COMMITTEE_SIZE // TARGET_AGGREGATORS_PER_SYNC_COMMITTEE
```
djrtwo, line 304: Ah this looks way too low, right? We want to target 2 or 4 aggregators per *sub committee* rather than just 4 for the entire committee.

Should we make this function be for the sub-committee? and the target value based on the sub committee size?

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+    )
+    return bytes_to_uint64(hash(signature)[0:8]) % modulo == 0
+```
+
+*NOTE*: the set of aggregators will generally change every slot; however, the assignments can be computed ahead of time as soon as the committee is known
```
djrtwo, line 309:
```suggestion
*NOTE*: the set of aggregators generally changes every slot; however, the assignments can be computed ahead of time as soon as the committee is known
```

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+
+##### Construct sync committee contribution
+
+If a validator is selected to aggregate the `SyncCommitteeSignature`s produced on a subnet during a given `slot`, they construct an aggregated `SyncCommitteeContribution`. Given all of the collected `sync_committee_signatures` from the `sync_committee_{subnet_id}` gossip, create the contribution via `construct_sync_committee_contribution`. `state` in this function is the same `BeaconState` used by the aggregator to produce their individual signature for the assigned slot.
+
+```python
+def construct_sync_committee_contribution(state: BeaconState, sync_committee_signatures: Set[SyncCommitteeSignature]) -> SyncCommitteeContribution:
```
djrtwo, line 316: line length

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+    # Signature by the validator over the block root of `slot`
+    signature: BLSSignature
+```
+
+### SyncCommitteeContribution
+
+```python
+class SyncCommitteeContribution(Container):
+    # Slot to which this contribution pertains
+    slot: Slot
```
djrtwo, line 88: Similarly, I think we'll want block_root here. but we can debate a bit more tomorrow

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+
+If a validator is selected to aggregate the `SyncCommitteeSignature`s produced on a subnet during a given `slot`, they construct an aggregated `SyncCommitteeContribution`. Given all of the collected `sync_committee_signatures` from the `sync_committee_{subnet_id}` gossip, create the contribution via `construct_sync_committee_contribution`. `state` in this function is the same `BeaconState` used by the aggregator to produce their individual signature for the assigned slot.
+
+```python
+def construct_sync_committee_contribution(state: BeaconState, sync_committee_signatures: Set[SyncCommitteeSignature]) -> SyncCommitteeContribution:
+    contribution = SyncCommitteeContribution()
+    contribution.slot = state.slot
+
+    sync_committee = state.current_sync_committee
+    epoch = compute_epoch_at_slot(state.slot)
```
djrtwo, line 321: use `get_current_epoch`

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+    domain = get_domain(state, DOMAIN_SYNC_COMMITTEE, epoch)
+    signing_root = compute_signing_root(
```
djrtwo, line 323: one liner

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+    )
+
+    signatures = []
+    for sync_committee_signature in sync_committee_signatures:
+        index = sync_committee_signature.index
+        signature = sync_committee_signature.signature
+        contribution.aggregate_bits[index] = True
+        signatures.append(signature)
+
+        # is sync committee signature valid?
```
djrtwo, line 335: spec code comments use capital letter to start

```suggestion
        # Is sync committee signature valid?
```

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+    sync_committee = state.current_sync_committee
+    epoch = compute_epoch_at_slot(state.slot)
+    domain = get_domain(state, DOMAIN_SYNC_COMMITTEE, epoch)
+    signing_root = compute_signing_root(
+      get_block_root_at_slot(state, state.slot), 
+      domain
+    )
+
+    signatures = []
+    for sync_committee_signature in sync_committee_signatures:
```
djrtwo, line 329: If we use block_root, the aggregator will need to filter the committee signatures by the block_root of their own contribution. This doesn't have to be in code. Can jsut be a note about the sync_committee_signatures that are supposed to be passed in like [here](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/validator.md#construct-aggregate)

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+    )
+    return bytes_to_uint64(hash(signature)[0:8]) % modulo == 0
+```
+
+*NOTE*: the set of aggregators will generally change every slot; however, the assignments can be computed ahead of time as soon as the committee is known
+
+##### Construct sync committee contribution
+
+If a validator is selected to aggregate the `SyncCommitteeSignature`s produced on a subnet during a given `slot`, they construct an aggregated `SyncCommitteeContribution`. Given all of the collected `sync_committee_signatures` from the `sync_committee_{subnet_id}` gossip, create the contribution via `construct_sync_committee_contribution`. `state` in this function is the same `BeaconState` used by the aggregator to produce their individual signature for the assigned slot.
+
```
djrtwo, line 314: Slight preference for more granular setting of fields as was done in [validator guide aggregation](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/validator.md#construct-aggregate)

not a big deal either way but a bit easier to parse I think

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+    signatures = []
+    for sync_committee_signature in sync_committee_signatures:
+        index = sync_committee_signature.index
+        signature = sync_committee_signature.signature
+        contribution.aggregate_bits[index] = True
+        signatures.append(signature)
+
+        # is sync committee signature valid?
+        assert sync_committee_signature.slot == state.slot
+        assert bls.Verify(
```
djrtwo, line 337: I don't think we need assertions at this point in the spec. The conditions that these values would be passed into this function would be that they were verified during gossip validations. I would just note that you collected the valid `SyncCommitteeSignature`s for the slot and leave the validations out

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+        contribution=contribution,
+        selection_proof=selection_proof,
+    )
+```
+
+Then `signed_contribution_and_proof = SignedContributionAndProof(message=contribution_and_proof, signature=signature)` is constructed and broadast. Where `signature` is obtained from:
+
+```python
+def get_contribution_and_proof_signature(
+  state: BeaconState,
```
djrtwo, line 378: spacing. prefer 4-spaces and prefer trailing off the function name like above (`get_contribution_and_proof`)

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+                               privkey: int) -> ContributionAndProof:
+    selection_proof = get_sync_committee_slot_signature(
+      state, 
+      contribution.slot,
+      privkey
+    )
+    return ContributionAndProof(
+        aggregator_index=aggregator_index,
+        contribution=contribution,
+        selection_proof=selection_proof,
```
djrtwo, line 370: inline preference
```suggestion
        selection_proof=get_sync_committee_slot_signature(state, contribution.slot, privkey),
```

specs/lightclient/validator.md
djrtwo: we could extend `attnets` and just make it `subnets`, where first 64 are attnets and next 8 bits are sync committee subnets

specs/lightclient/validator.md
```diff
@@ -0,0 +1,395 @@
...
+    domain = get_domain(state, DOMAIN_CONTRIBUTION_AND_PROOF, compute_epoch_at_slot(contribution.slot))
+    signing_root = compute_signing_root(contribution_and_proof, domain)
+    return bls.Sign(privkey, signing_root)
+```
+
+## Sync committee subnet stability
+
+The sync committee subnets need special care to ensure stability given the relatively low number of validators involved in the sync committee at any particular time. To provide this stability, a validator must do the following:
+
+* Maintain advertisement of the subnet the validator in the sync committee is assigned to in their node's ENR as soon as they have joined the subnet. Subnet assignments are known `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` epochs in advance and can be computed with `compute_subnet_for_sync_committee` defined above. ENR advertisement is indicated by setting the `syncnets` entry in the ENR to a `uint8` encoding of the `subnet_id`, e.g. `ENR["syncnets"] = b'\x07'`. This entry is removed from the ENR after the validator has left the sync committee.
```
djrtwo, line 391: damn, privacy on this shit is terrible

specs/lightclient/validator.md
djrtwo: hmm, probably want to join ~1 to 5 epochs head of time and advertise when you join.

specs/lightclient/validator.md
djrtwo, line 391: This should probably be a bitfield. There's a chance you are only multiple subnets at once due to duplicate assignments or running multilpe validators on your machine

djrtwo (commented):
did the rest of review

specs/lightclient/validator.md
ralexstokes: yes, good point

specs/lightclient/validator.md
ralexstokes: will remove as i add gossip conditions

specs/lightclient/validator.md
ralexstokes: yep, this should be updated... let me fix

specs/lightclient/validator.md
```diff
@@ -0,0 +1,381 @@
...
+
+##### Broadcast sync committee signature
+
+The validator broadcasts the assembled signature(s) to the assigned subnet(s), the `sync_committee_{subnet_id}` pubsub topic.
+
+The `subnet_id` is derived from the position in the sync committee such that the sync committee is divided into "subcommittees". It can be computed via `compute_subnet_for_sync_committee`:
+
+```python
+def compute_subnet_for_sync_committee(sync_committee_index: SyncCommitteeIndex) -> uint64:
+    return uint64(sync_committee_index % SYNC_COMMITTEE_SUBNET_COUNT)
```
protolambda, line 284: Will there be modulo bias towards low subnets? Or does it divide nicely?

specs/lightclient/validator.md
```diff
@@ -0,0 +1,381 @@
...
+```python
+def get_sync_committee_slot_signature(state: BeaconState, slot: Slot, privkey: int) -> BLSSignature:
+    domain = get_domain(state, DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF, compute_epoch_at_slot(slot))
+    signing_root = compute_signing_root(slot, domain)
+    return bls.Sign(privkey, signing_root)
+```
+
+```python
+def is_sync_committee_aggregator(state: BeaconState, slot: Slot, signature: BLSSignature) -> bool:
+    modulo = max(1, SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT // TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE)
```
protolambda, line 309: Just double checking: `max(1, 1024 // 8 // 4) = max(1, 32) = 32`, 5 bits, constants would have to change a lot for any bias with 64 bit seed. LGTM.

specs/lightclient/p2p-interface.md
```diff
@@ -0,0 +1,101 @@
...
+ 
+This topic is used to propagate partially aggregated sync committee signatures to be included in future blocks.
+
+The following validations MUST pass before forwarding the `signed_contribution_and_proof` on the network; define `contribution_and_proof = signed_contribution_and_proof.message` and `contribution = contribution_and_proof.contribution` for convenience:
+
+- _[IGNORE]_ The contribution's slot is for the current slot, i.e. `contribution.slot == current_slot`.
+- _[REJECT]_ The aggregate signature is valid for the message `beacon_block_root` and aggregate pubkey derived from the participation info in `aggregate_bits`.
+- _[IGNORE]_ The block being signed over (`contribution.beacon_block_root`) has been seen (via both gossip and non-gossip sources).
+- _[IGNORE]_ There has been no other valid sync committee contribution for the declared `slot` by the validator referenced by `contribution_and_proof.aggregator_index`.
+- _[REJECT]_ `contribution_and_proof.selection_proof` selects the validator as an aggregator for the slot --
```
protolambda, line 61: This only checks if the signature if valid, not if the `aggregator_index` is actually part of the sync committee. Either we need to restrict selection proofs to the sync commitee validators (like we do with attestation committees), or drastically change the selection chances to be much smaller. Otherwise we would be allowing `validator_count / SYNC_COMMITTEE_SIZE = ~ 150` times as many validators to aggregate :sweat_smile:

protolambda (commented):
Reviewed the network part, will reivist with deeper review later.

specs/lightclient/p2p-interface.md
```diff
@@ -0,0 +1,101 @@
...
+
+Unaggregated signatures (along with metadata) are sent as `SyncCommitteeSignature`s on the `sync_committee_{subnet_id}` topics.
+
+Aggregated sync committee signatures are packaged into (signed) `SyncCommitteeContribution` along with proofs and gossiped to the `sync_committee_contribution_and_proof` topic.
+
+## The discovery domain: discv5
+
+The `attnets` key of the ENR is extended to cover both the beacon committee subnets and the sync committee subnets under a combination key `subnets`. 
+
+The `subnets` key is an SSZ `Bitvector` of length `ATTESTATION_SUBNET_COUNT + SYNC_COMMITTEE_SUBNET_COUNT`.
```
protolambda, line 95: Currently we have a nice benefit that The ENR subnets key is a `uint64`, I'd like to keep that. Also, I think adding another ENR field isn't that bad, and much cleaner w.r.t. updating: in the double bitvector-length approach non-fork-ready nodes won't be able to parse the attnets data of fork-ready nodes.

And if we really don't want those few extra key bytes, can we at least define the attnets value as SSZ container of two bitvectors? (SSZ encodes it the exact same in this case actually).

Another reason for separation by key: if the attestation subnet count changes (it's been discussed before, can happen without shard increase even), the concatenated bitvector approach gets even messier: different node versions interpret overlapping bitfield bits differently.

specs/lightclient/p2p-interface.md
```diff
@@ -0,0 +1,101 @@
...
+
+Topics follow the same specification as in the phase 0 document. The only change is the addition of new gossipsub topics added in HF1.
+
+The specification around the creation, validation, and dissemination of messages has not changed from the Phase 0 document.
+
+The new topics along with the type of the `data` field of a gossipsub message are given in this table:
+
+| Name | Message Type |
+| - | - |
+| `sync_committee_contribution_and_proof` | `SignedContributionAndProof` |
```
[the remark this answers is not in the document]
ralexstokes: i'd suggest waiting until we have data backing up the claim that this is a significant burden on clients; there are likely more pressing spots for optimization and we get lots of points for readability imo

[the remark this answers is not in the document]
ralexstokes: if it becomes a problem i do like your suggestion to use a hash function to reduce the footprint

specs/lightclient/p2p-interface.md
```diff
@@ -0,0 +1,101 @@
...
+## The gossip domain: gossipsub
+
+Gossip meshes are added in HF1 to support the consensus activities of the sync committees. Validators use an aggregation scheme to balance the processing and networking load across all of the relevant actors.
+
+### Topics and messages
+
+Topics follow the same specification as in the phase 0 document. The only change is the addition of new gossipsub topics added in HF1.
+
+The specification around the creation, validation, and dissemination of messages has not changed from the Phase 0 document.
+
```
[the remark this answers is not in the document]
ralexstokes: i feel like it is clear from the phase 0 spec to use the latest `fork_digest` as specified here: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/p2p-interface.md#topics-and-messages

where the `current_fork_version` changes as we move from phase 0 to altair.

we can always be more clear about that part exactly -- what do you think?

specs/lightclient/p2p-interface.md
```diff
@@ -0,0 +1,101 @@
...
+## The gossip domain: gossipsub
+
+Gossip meshes are added in HF1 to support the consensus activities of the sync committees. Validators use an aggregation scheme to balance the processing and networking load across all of the relevant actors.
+
+### Topics and messages
+
+Topics follow the same specification as in the phase 0 document. The only change is the addition of new gossipsub topics added in HF1.
+
+The specification around the creation, validation, and dissemination of messages has not changed from the Phase 0 document.
```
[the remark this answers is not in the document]
ralexstokes: yeah i think adding a note to the topic table about the new type would be useful

specs/lightclient/p2p-interface.md
```diff
@@ -0,0 +1,101 @@
...
+##### `sync_committee_{subnet_id}`
+
+The `sync_committee_{subnet_id}` topics are used to propagate unaggregated sync committee signatures to the subnet `subnet_id` to be aggregated before being gossiped to the global `sync_committee_contribution_and_proof` topic.
+
+The following validations MUST pass before forwarding the `sync_committee_signature` on the network:
+
+- _[IGNORE]_ The signature's slot is for the current slot, i.e. `sync_committee_signature.slot == current_slot`.
+- _[REJECT]_ The `signature` is valid for the message `beacon_block_root` for the validator referenced by the `index`.
+- _[REJECT]_ The `subnet_id` is correct, i.e. `subnet_id == compute_subnet_for_sync_commitee(sync_committee_signature.index)`
+- _[IGNORE]_ There has been no other valid sync committee signature for the declared `slot` by the validator referenced by `index`.
```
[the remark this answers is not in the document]
ralexstokes: ah, sure

specs/lightclient/validator.md
ralexstokes: this work has been moved to the aggregator in the current design

specs/lightclient/validator.md
```diff
@@ -0,0 +1,383 @@
...
+
+### SyncCommitteeSignature
+
+```python
+class SyncCommitteeSignature(Container):
+    # Slot to which this contribution pertains
+    slot: Slot
+    # Index in the sync committee corresponding to the validator
+    # who produced this signature
+    index: SyncCommitteeIndex
```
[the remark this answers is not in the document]
ralexstokes: are we ok w/ `index` if it refers to `ValidatorIndex`? or the more verbose `validator_index`?

specs/lightclient/validator.md
```diff
@@ -0,0 +1,383 @@
...
+        signatures.append(contribution.signature)
+
+    body.sync_committee_signature = bls.Aggregate(signatures)
+```
+
+*NOTE*: (TODO) update w/ `SignedContributionAndProof` pending which validations go at gossip layer and which at application layer
+
+*NOTE*: (TODO) likely to update with more context on subcommittee messages so we can skip some of the ~quadratic work here and save on "large" sparse messages
+
+*NOTE*: (TODO) somewhere, the proposer needs to join _some_ gossip channel to collect signatures... have the proposer join one epoch ahead of schedule (or how long?)
```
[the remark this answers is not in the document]
ralexstokes: working on scheme to make the global channel as light as possible, can revisit more drastic changes to sync committee infra if the load becomes unmanageable

specs/lightclient/p2p-interface.md
ralexstokes: nice catch, agreed this should be in here!

specs/lightclient/p2p-interface.md
```diff
@@ -0,0 +1,101 @@
+# Ethereum HF1 networking specification
+
+This document contains the networking specification for Ethereum 2.0 clients added during the HF1 deployment. This document should be viewed as additive to the [document from Phase 0](../phase0/p2p-interface.md) and will be referred to as the "phase 0 document" hereafter. Readers should understand the phase 0 document and use it as a basis to understand the changes outlined in this document.
+
+In particular, HF1 only introduces _additive_ changes to the gossip domain and small change to ENR advertisement in the discovery domain. The phase 0 document should be consulted for specs on the network fundamentals, the existing gossip domain, the req/resp domain and the discovery domain.
```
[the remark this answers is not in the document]
ralexstokes: right -- there is networking "near" the fork and then networking in the steady state of running `altair`

can definitely add some more prose around how to handle the fork boundary but almost wonder if it should go into the `fork.md` document -- thoughts?

specs/lightclient/p2p-interface.md
```diff
@@ -0,0 +1,101 @@
...
+##### `sync_committee_contribution_and_proof`
+ 
+This topic is used to propagate partially aggregated sync committee signatures to be included in future blocks.
+
+The following validations MUST pass before forwarding the `signed_contribution_and_proof` on the network; define `contribution_and_proof = signed_contribution_and_proof.message` and `contribution = contribution_and_proof.contribution` for convenience:
+
+- _[IGNORE]_ The contribution's slot is for the current slot, i.e. `contribution.slot == current_slot`.
+- _[REJECT]_ The aggregate signature is valid for the message `beacon_block_root` and aggregate pubkey derived from the participation info in `aggregate_bits`.
+- _[IGNORE]_ The block being signed over (`contribution.beacon_block_root`) has been seen (via both gossip and non-gossip sources).
+- _[IGNORE]_ There has been no other valid sync committee contribution for the declared `slot` by the validator referenced by `contribution_and_proof.aggregator_index`.
```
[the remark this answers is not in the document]
ralexstokes: i think we want to remove this condition though right?

aiui aggregates other than the first will still consume gossipsub resources via the relatively expensive `IHAVE/IWANT` fetching, rather than just forwarding

specs/lightclient/validator.md
ralexstokes: checking this out: https://research.kudelskisecurity.com/2020/07/28/the-definitive-guide-to-modulo-bias-and-how-to-avoid-it/, it seems that we are ok here as we are moving b/t powers of two (1024 % 8)

specs/lightclient/validator.md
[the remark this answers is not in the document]
ralexstokes: seems more consistent w/ rest of spec to be verbose here -- will update

specs/lightclient/p2p-interface.md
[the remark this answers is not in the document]
ralexstokes: addressed in #2244.

specs/lightclient/p2p-interface.md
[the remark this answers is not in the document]
ralexstokes: addressed in #2244

specs/lightclient/p2p-interface.md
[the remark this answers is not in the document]
protolambda: I was meaning that the `(slot, aggregator_index)` is a better measure to filter than the hash-tree-root of contribution, if multiple selected aggregators turn out to make the same contribution.

Duplicates (same contribution, different aggregator) are expensive but necessary, since they have different message IDs due to the gossip aggregator info, and we should not score those down. Changing the gossip message ID function would be better, but let's address that later.

Duplicates with same aggregator however are plain wrong, and should be ignored like in current spec. Maybe even rejected (since they shouldn't exist, there should be one per aggregator per slot right?).

specs/lightclient/p2p-interface.md
ralexstokes: moving ahead w/ a separate bitfield for the sync committee subnets; agreed we can alter if we need more ENR space

specs/altair/validator.md
```diff
@@ -0,0 +1,367 @@
...
+
+## Beacon chain responsibilities
+
+A validator maintains the responsibilities given in the Phase 0 document.
+
+Block proposals are modified to incorporate the sync committee signatures as detailed below.
+
+When assigned to a sync committee, validators have a new responsibility to sign beacon block roots during each slot of the sync committee period.
+These signatures are aggregated and routed to the proposer over gossip for inclusion into a beacon block. 
+Assignments to a particular sync committee are infrequent at normal validator counts; however, an action every slot is be required when in the currently active sync committee.
```
djrtwo, line 172:
```suggestion
Assignments to a particular sync committee are infrequent at normal validator counts; however, an action every slot is be required when in the current active sync committee.
```

specs/altair/p2p-interface.md
```diff
@@ -0,0 +1,114 @@
...
+#### Sync committee subnets
+
+Sync committee subnets are used to propagate unaggregated sync committee signatures to subsections of the network.
+
+##### `sync_committee_{subnet_id}`
+
+The `sync_committee_{subnet_id}` topics are used to propagate unaggregated sync committee signatures to the subnet `subnet_id` to be aggregated before being gossiped to the global `sync_committee_contribution_and_proof` topic.
+
+The following validations MUST pass before forwarding the `sync_committee_signature` on the network:
+
```
djrtwo, line 88: need a validation condition for 0s in the bitfield outside of the sub-committee for this subnet

specs/altair/validator.md
```diff
@@ -0,0 +1,367 @@
...
+
+The proposer receives a number of `SyncCommitteeContribution`s (wrapped in `SignedContributionAndProof`s on the wire) from validators in the sync committee who are selected to partially aggregate signatures from the full committee (see below for details).
+
+The proposer collects these contributions for further aggregation when preparing a block. 
+Recall `block.body.sync_aggregate.sync_committee_bits` is a `Bitvector` where the `i`th bit is `True` if the corresponding validator in the sync committee has produced a valid signature, and that `block.body.sync_aggregate.sync_committee_signature` is the aggregate BLS signature combining all of the valid signatures.
+
+Given a collection of `contributions` and the `BeaconBlock` under construction, the proposer processes them as follows:
+
+```python
+def process_sync_committee_contributions(state: BeaconState,
```
djrtwo, line 198: If they use this function, you need to dictate that contributions *must* be non-overlapping

djrtwo: Might be easiest to dictate that they take the *best* contribution from each subcommittee

specs/altair/validator.md
```diff
@@ -0,0 +1,367 @@
...
+It can be computed via `compute_subnet_for_sync_committee` where `state` is a `BeaconState` during the matching sync committee period:
+
+```python
+def compute_subnet_for_sync_committee(state: BeaconState, validator_index: ValidatorIndex) -> uint64:
+    pubkey = state.validators[validator_index].pubkey
+    sync_committee_index = state.current_sync_committee.pubkeys.index(pubkey)
+    return uint64(sync_committee_index % SYNC_COMMITTEE_SUBNET_COUNT)
+```
+
+*NOTE*: subnet assignment does not change during the duration of a validator's assignment to a given sync committee.
```
djrtwo, line 267: Capitalize the first word after "Note". Also "Note" should not be all caps

```suggestion
*Note*: Subnet assignment does not change during the duration of a validator's assignment to a given sync committee.
```

specs/altair/validator.md
```diff
@@ -0,0 +1,367 @@
...
+###### Slot
+
+Set `contribution.slot = state.slot` where `state` is the `BeaconState` for the slot in question.
+
+###### Aggregate bits
+
+Let `contribution.aggregate_bits` be a `Bitvector[SYNC_COMMITTEE_SIZE]`, where the `index`th bit is set in the `Bitvector` for each corresponding validator included in this aggregate.
+Note that an aggregator needs to find the index in the sync committee (as returned by `get_sync_committee_indices`) for a given validator referenced by `sync_committee_signature.validator_index` and _this_ is the index to set in the `Bitvector`. 
+For example, validator with index `2044` could be at index `3` in the current sync committee so position `3` in the `Bitvector` is set if a signature from this validator is included in the aggregation. 
+Also note that a validator could be included multiple times in a given sync committee such that multiple bits are set for a single `SyncCommitteeSignature`.
```
djrtwo, line 310: Ah, I see what you are doing here... The aggregator can include for positions outside of the sub committee.

I *think* that subcommittee/aggregate validations will be simpler if we restrict and have validators double broadcast to different subnets.

In the previous aggregation scheme (no subcommittee aggregates), double broadcasting didn't make much sense to me but in the context of multiple sub committees, I think it does.

Note, if you are duplicated within a single subcommittee, aggregator should include both

specs/altair/validator.md
```diff
@@ -0,0 +1,367 @@
...
+
+Set `contribution.signature = aggregate_signature` where `aggregate_signature` is obtained from:
+
+```python
+def get_aggregate_signature(sync_committee_signatures: Set[SyncCommitteeSignature]) -> BLSSignature:
+    signatures = [sync_committee_signature.signature for sync_committee_signature in sync_committee_signatures]
+    return bls.Aggregate(signatures)
+```
+
+Note that if a validator is included multiple times in a given sync committee the aggregator must include their signature multiple times in the argument to `get_aggregate_signature` as well.
```
djrtwo, line 322: Use "Note:" notation

specs/altair/p2p-interface.md
```diff
@@ -0,0 +1,114 @@
...
+This topic is used to propagate partially aggregated sync committee signatures to be included in future blocks.
+
+The following validations MUST pass before forwarding the `signed_contribution_and_proof` on the network; define `contribution_and_proof = signed_contribution_and_proof.message` and `contribution = contribution_and_proof.contribution` for convenience:
+
+- _[IGNORE]_ The contribution's slot is for the current slot, i.e. `contribution.slot == current_slot`.
+- _[IGNORE]_ The block being signed over (`contribution.beacon_block_root`) has been seen (via both gossip and non-gossip sources).
+- _[IGNORE]_ There has been no other valid sync committee contribution for the declared `slot` by the validator referenced by `contribution_and_proof.aggregator_index`.
+- _[REJECT]_ The aggregator's validator index is within the current sync committee --
+  i.e. `state.validators[aggregate_and_proof.aggregator_index].pubkey in state.current_sync_committee.pubkeys`.
+- _[REJECT]_ The aggregate signature is valid for the message `beacon_block_root` and aggregate pubkey derived from the participation info in `aggregate_bits`.
```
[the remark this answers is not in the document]
ralexstokes: fair, i was taking the view that these validations did not necessarily imply order of optimization and that would be a detail for implementers. but agree that if we can provide this order, it doesn't hurt

specs/altair/validator.md
ralexstokes: ah yes, when i wrote this i was thinking we would end up w/ contiguous subcommittees which we don't have with the current definition of `compute_subnet_for_sync_committee`.  i think this functions needs to be updated; if we use the aggregates in the `state` we will move to non-overlapping subsets by default so i think best to leave this function alone

specs/altair/validator.md
```diff
@@ -94,9 +94,11 @@ class SyncCommitteeContribution(Container):
     slot: Slot
     # Block root for this contribution
     beacon_block_root: Root
+    # The subcommittee this contribution pertains to out of the broader sync committee
+    subcommittee_index: uint64
     # A bit is set if a signature from the validator at the corresponding
-    # index in the sync committee is present in the aggregate `signature`.
-    aggregate_bits: Bitvector[SYNC_COMMITTEE_SIZE]
+    # index in the subcommittee is present in the aggregate `signature`.
+    aggregate_bits: Bitvector[SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT]
```
djrtwo, line 103: prefer `aggregation_bits` to match naming in `Attestation`

specs/altair/validator.md
```diff
@@ -187,25 +189,33 @@ After constructing the `BeaconBlockBody` as per that section, the proposer has a
 
 ##### Sync committee
 
-The proposer receives a number of `SyncCommitteeContribution`s (wrapped in `SignedContributionAndProof`s on the wire) from validators in the sync committee who are selected to partially aggregate signatures from the full committee (see below for details).
+The proposer receives a number of `SyncCommitteeContribution`s (wrapped in `SignedContributionAndProof`s on the wire) from validators in the sync committee who are selected to partially aggregate signatures from individual subcommittees formed by breaking the full sync committee into `SYNC_COMMITTEE_SUBNET_COUNT` pieces (see below for details).
```
djrtwo, line 194: "independent" might be a bit better descriptive to illustrate that they don't overlap

```suggestion
The proposer receives a number of `SyncCommitteeContribution`s (wrapped in `SignedContributionAndProof`s on the wire) from validators in the sync committee who are selected to partially aggregate signatures from independent subcommittees formed by breaking the full sync committee into `SYNC_COMMITTEE_SUBNET_COUNT` pieces (see below for details).
```

specs/altair/validator.md
```diff
@@ -187,25 +189,33 @@ After constructing the `BeaconBlockBody` as per that section, the proposer has a
...
 
 The proposer collects these contributions for further aggregation when preparing a block. 
+Proposers should select the best contribution seen across all aggregators for a given subnet/subcommittee when preparing a block.
```
djrtwo, line 197:
```suggestion
Proposers should select the best contribution seen across all aggregators for each subnet/subcommittee when preparing a block. 
```

specs/altair/validator.md
```diff
@@ -187,25 +189,33 @@ After constructing the `BeaconBlockBody` as per that section, the proposer has a
...
                                          contributions: Set[SyncCommitteeContribution]) -> None:
     sync_aggregate = SyncAggregate()
     signatures = []
+
     for contribution in contributions:
+        subcommittee_index = contribution.subcommittee_index
         for index, participated in enumerate(contribution.aggregate_bits):
             if participated:
-                sync_aggregate.sync_committee_bits[index] = True
+                participant_index = SYNC_COMMITTEE_SUBNET_COUNT * subcommittee_index + index
```
djrtwo, line 215: I think `SYNC_COMMITTEE_SUBNET_COUNT` is incorrect. Shouldn't this multiple by `size_of_subcommittees` rather than number of subcommittees

specs/altair/validator.md
```diff
@@ -255,32 +265,43 @@ def get_sync_committee_signature(state: BeaconState,
...
 ```python
-def compute_subnet_for_sync_committee(state: BeaconState, validator_index: ValidatorIndex) -> uint64:
-    pubkey = state.validators[validator_index].pubkey
-    sync_committee_index = state.current_sync_committee.pubkeys.index(pubkey)
-    return uint64(sync_committee_index // (SYNC_COMMITEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT))
+def compute_subnets_for_sync_committee(state: BeaconState, validator_index: ValidatorIndex) -> Sequence[uint64]:
+    target_pubkey = state.validators[validator_index].pubkey
+    sync_committee_indices = [index for index, pubkey in enumerate(state.current_sync_committee.pubkeys) if pubkey == target_pubkey]
+    return [
+      uint64(index // (SYNC_COMMITEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT))
```
djrtwo, line 277: indent level

specs/altair/validator.md
```diff
@@ -296,30 +317,30 @@ def is_sync_committee_aggregator(state: BeaconState, signature: BLSSignature) ->
...
-Note that an aggregator needs to find the index in the sync committee (as returned by `get_sync_committee_indices`) for a given validator referenced by `sync_committee_signature.validator_index` and _this_ is the index to set in the `Bitvector`. 
-For example, validator with index `2044` could be at index `3` in the current sync committee so position `3` in the `Bitvector` is set if a signature from this validator is included in the aggregation. 
-Also note that a validator could be included multiple times in a given sync committee such that multiple bits are set for a single `SyncCommitteeSignature`.
+Set `contribution.beacon_block_root = beacon_block_root` from the `beacon_block_root` found in the `sync_committee_signatures`.
 
-###### Signature
+###### Subcommittee index
 
-Set `contribution.signature = aggregate_signature` where `aggregate_signature` is obtained from:
+Set `contribution.subcommittee_index` to the index for the subcommittee index corresponding to the subcommittee assigned to this subnet. This index will match the `subnet_id` used to derive the topic name.
```
djrtwo, line 336:
```suggestion
Set `contribution.subcommittee_index` to the index for the subcommittee index corresponding to the subcommittee assigned to this subnet. This index matches the `subnet_id` used to derive the topic name.
```

djrtwo (commented):
did a quick review. great!

specs/altair/validator.md
ralexstokes: nice catch

specs/altair/validator.md
```diff
@@ -393,4 +393,4 @@ To provide this stability, a validator must do the following:
...
\ No newline at end of file
+* To join a sync committee subnet, select a random number of epochs before the end of the current sync committee period between 1 and `SYNC_COMMITTEE_SUBNET_COUNT`, inclusive. Validators should join their member subnet at the beginning of the epoch they have randomly selected. For example, if the next sync committee period starts at epoch `853,248` and the validator randomly selects an offset of `3`, they should join the subnet at the beginning of epoch `853,245`. Validators should leverage the lookahead period on sync committee assignments so that they can join the appropriate subnets ahead of their assigned sync committee period.
```
protolambda, line 396: Nitpick: can we keep the lines short to reduce diff problems?

specs/altair/validator.md
```diff
@@ -0,0 +1,367 @@
...
+```python
+def compute_subnet_for_sync_committee(state: BeaconState, validator_index: ValidatorIndex) -> uint64:
+    pubkey = state.validators[validator_index].pubkey
+    sync_committee_index = state.current_sync_committee.pubkeys.index(pubkey)
+    return uint64(sync_committee_index % SYNC_COMMITTEE_SUBNET_COUNT)
+```
+
+*NOTE*: subnet assignment does not change during the duration of a validator's assignment to a given sync committee.
+
+*NOTE*: a validator joins the "first" subnet (in sync committee index order) if they are duplicated in a given committee; this fact is reflected in the semantics of Python's `index` function in the definition for `compute_subnet_for_sync_committee`.
```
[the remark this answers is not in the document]
ralexstokes: resolved in updated scheme

specs/altair/p2p-interface.md
ralexstokes: no longer applies w/ newer aggregation scheme

specs/altair/validator.md
ralexstokes: resolved in newer scheme

specs/altair/validator.md
ralexstokes: yeah i was applying the "only change if i touch it" process but i can go thru this PR before it is final and change them

ralexstokes: ah actually, here i think it will break the `*` bullet formatting to have a newline -- do you know?

specs/altair/validator.md
```diff
@@ -0,0 +1,396 @@
...
+
+###### Aggregate bits
+
+Let `contribution.aggregation_bits` be a `Bitvector[SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT]`, where the `index`th bit is set in the `Bitvector` for each corresponding validator included in this aggregate from the corresponding subcommittee.
+An aggregator needs to find the index in the sync committee (as returned by `get_sync_committee_indices`) for a given validator referenced by `sync_committee_signature.validator_index` and map the sync committee index to an index in the subcommittee (along with the prior `subcommittee_index`). This index within the subcommittee is the one set in the `Bitvector`.
+For example, a validator with index `2044` could be at index `15` in the current sync committee. This sync committee index maps to `subcommittee_index` `1` with position `7` in the `Bitvector` for the contribution. 
+Also note that a validator could be included multiple times in a given subcommittee such that multiple bits are set for a single `SyncCommitteeSignature`.
```
protolambda, line 343:
```suggestion
Also note that a validator **could be included multiple times** in a given subcommittee such that multiple bits are set for a single `SyncCommitteeSignature`.
```
Nitpick: This is different than any existing committee logic, worth emphasizing in my opinion.

specs/altair/validator.md
```diff
@@ -0,0 +1,396 @@
...
+A validator determines beacon committee assignments and beacon block proposal duties as defined in the Phase 0 document.
+
+### Sync Committee
+
+To determine sync committee assignments, a validator can run the following function: `is_assigned_to_sync_committee(state, epoch, validator_index)` where `epoch` is an epoch number within the current or next sync committee period.
+This function is a predicate indicating the presence or absence of the validator in the corresponding sync committee for the queried sync committee period.
+
+```python
+def compute_sync_committee_period(epoch: Epoch) -> uint64:
+    return epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
```
protolambda, line 136: What if the fork epoch does not align with the sync committee period? Will we have a shorter period at the start?

ralexstokes: that seems to be the case, don't think it is worth extra complexity to adjust in protocol. if we have security concerns here we can add <24hrs of epochs to the fork epoch and just wait until it is aligned

specs/altair/validator.md
```diff
@@ -0,0 +1,396 @@
...
+
+The proposer receives a number of `SyncCommitteeContribution`s (wrapped in `SignedContributionAndProof`s on the wire) from validators in the sync committee who are selected to partially aggregate signatures from independent subcommittees formed by breaking the full sync committee into `SYNC_COMMITTEE_SUBNET_COUNT` pieces (see below for details).
+
+The proposer collects these contributions for further aggregation when preparing a block. 
+Proposers should select the best contribution seen across all aggregators for each subnet/subcommittee when preparing a block. 
+A contribution with more valid signatures is better than a contribution with fewer signatures.
+
+Recall `block.body.sync_aggregate.sync_committee_bits` is a `Bitvector` where the `i`th bit is `True` if the corresponding validator in the sync committee has produced a valid signature, and that `block.body.sync_aggregate.sync_committee_signature` is the aggregate BLS signature combining all of the valid signatures.
+
+Given a collection of the best seen `contributions` (with no repeating `subcommittee_index` values) and the `BeaconBlock` under construction, the proposer processes them as follows:
```
protolambda: > with no repeating `subcommittee_index` values

This is not checked in the function below. Is it necessary? The below review comment was resolved, but it doesn't look like it was addressed, or is this different?

ralexstokes: we could allow repeating `subcommittee_index` values as long as the `aggregation_bits` are non-overlapping

the validation is not checked in the function as this prose is supposed to stand-in for a precondition to the function

specs/altair/validator.md
```diff
@@ -0,0 +1,416 @@
...
+Any bits modified for the sync committee responsibilities are unset in the ENR after any validators have left the sync committee.
+
+*Note*: The first sync committee from phase 0 to the Altair fork will not be known until the fork happens which implies subnet assignments are not known until then. 
+Early sync committee members should listen for topic subscriptions from peers and employ discovery via the ENR advertisements near the fork boundary to form initial subnets 
+Some early sync committee rewards may be missed while the initial subnets form.
+
+* To join a sync committee subnet, select a random number of epochs before the end of the current sync committee period between 1 and `SYNC_COMMITTEE_SUBNET_COUNT`, inclusive. 
+Validators should join their member subnet at the beginning of the epoch they have randomly selected. 
+For example, if the next sync committee period starts at epoch `853,248` and the validator randomly selects an offset of `3`, they should join the subnet at the beginning of epoch `853,245`.
```
protolambda, line 415: Should this be 853,251? I.e. add 3 instead of subtract 3?

djrtwo (approved):

specs/altair/validator.md
```diff
@@ -0,0 +1,416 @@
...
+
+## Prerequisites
+
+All terminology, constants, functions, and protocol mechanics defined in the [Altair -- The Beacon Chain](./beacon-chain.md) doc are requisite for this document and used throughout. 
+Please see this document before continuing and use as a reference throughout.
+
+## Warning
+
+This document is currently illustrative for early Altair testnets and some parts are subject to change. 
+In particular, the sync committee aggregation scheme may change to reduce resource consumption and better leverage the subcommittee aggregates defined in the state.
```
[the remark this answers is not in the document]
ralexstokes: i'll keep for now (e.g. may think more about gossip validations as we go, i went quickly through those w/ new changes) but will make less specific

specs/altair/validator.md
```diff
@@ -0,0 +1,416 @@
...
+`ContributionAndProof` messages are signed by the aggregator and broadcast inside of `SignedContributionAndProof` objects to prevent a class of DoS attacks and message forgeries.
+
+First, `contribution_and_proof = get_contribution_and_proof(state, validator_index, contribution, privkey)` is constructed.
+
+```python
+def get_contribution_and_proof(state: BeaconState,
+                               aggregator_index: ValidatorIndex,
+                               contribution: SyncCommitteeContribution,
+                               privkey: int) -> ContributionAndProof:
+    selection_proof = get_sync_committee_slot_signature(state,
```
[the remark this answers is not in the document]
ralexstokes: yep, wasnt sure if we did that or followed styling for function params; ill fix

ralexstokes:
force-pushed to squash branch to one commit, latest feedback should be applied

protolambda (approved):
:tada: great work everyone

PR #2202, opened 2021-02-19
fork-choice test vectors: starting with `get_head` tests

hwwhww:
I’m transforming the current pyspec fork choice test cases into test vectors. 🧶

* This PR demonstrates a proposal based on @ericsson49's https://github.com/ethereum/eth2.0-spec-tests/issues/17 format, but with some adjustments so that it's more friendly to pyspec utils.
* Starting with `get_head` helper. `test/phase0/fork_choice/test_get_head.py` is basically copied from `test/phase0/unittests/fork_choice/test_get_head.py`, but with some extra `yield` to output the test vectors. [Minimal config `get_head` test vectors are here](https://github.com/hwwhww/eth2.0-spec-tests/tree/fork-choice/tests/minimal/phase0/fork_choice/get_head/pyspec_tests).

### Test formats
- `anchor_state.yaml` and `anchor_block.yaml`: the inputs of `get_forkchoice_store(anchor_state: BeaconState, anchor_block: BeaconBlock)`, to initialize the `store`.
    -  **Note**: https://github.com/ethereum/eth2.0-spec-tests/issues/17 uses `genesis`. But in our case, the anchor state may not be genesis.
- `block_<block_root>.yaml`: `SignedBeaconBlock` object
- `attestation_<attestation_root>.yaml`: `Attestation` object
- `steps.yaml`: the execution steps. Basically similar to https://github.com/ethereum/eth2.0-spec-tests/issues/17's "steps". The member elements may be:
    - `tick: <time>`: integer, the time for calling `on_tick(store: Store, time: uint64)`
        -  **Note**: https://github.com/ethereum/eth2.0-spec-tests/issues/17 uses `slot`.
    - `block: <block_file_name>`: the `SignedBeaconBlock` object for calling `on_block(store: Store, signed_block: SignedBeaconBlock)`
    - `attestation: <attestation_file_name>`: the `Attestation` object for calling `on_attestation(store: Store, attestation: Attestation)`
    - `checks` dict:
        - `head`: the block root when calling `get_head(store: Store)`
        - `justified_checkpoint_root`: the hash tree root of `Store.justified_checkpoint`.
    - [Example: filtered_block_tree/steps.yaml](https://github.com/hwwhww/eth2.0-spec-tests/blob/fork-choice/tests/minimal/phase0/fork_choice/get_head/pyspec_tests/filtered_block_tree/steps.yaml)
---
### Discussions
1. Any suggestions on the above proposal?
2. It's easy to convert `get_head` tests. However, in our `on_block` tests, we directly set the fields of `Store` to prepare the tests. For example:
https://github.com/ethereum/eth2.0-specs/blob/8ec082fcf9ca58bf143977abb4dce9c8c417dbdd/tests/core/pyspec/eth2spec/test/phase0/unittests/fork_choice/test_on_block.py#L30-L31
https://github.com/ethereum/eth2.0-specs/blob/8ec082fcf9ca58bf143977abb4dce9c8c417dbdd/tests/core/pyspec/eth2spec/test/phase0/unittests/fork_choice/test_on_block.py#L129-L132
I may have to rewrite them to get rid of the Store field manipulations if we are going to follow the above test formats.

3. Alternatively, we can output the whole `store` object. However, some drawbacks:
    - `Store` object could be really huge. We probably can only output `Minimal` config tests.
    - Client teams may have implemented the concept of `Store` in a very different, optimized solution.

---
### 20210313 updated:
See [fork choice test format doc](https://github.com/ethereum/eth2.0-specs/blob/e77ba9182182d45490f54dbed60ac890ba8ced57/tests/formats/fork_choice/README.md)


ericsson49:
When I was designing my fork choice tests, I avoided explicit instantiation of Store objects, for the same reasons.
My fork choice tests are integration tests rather than unit tests, so I use sequences of `tick`, `on_block` and `on_attestation` events (along with initial state) to reach desired Store configurations.
That's of course more tricky, but with the approach, one is making less assumptions about implementation details. Which I believe is important for such kind of tests, as it can hinder implementer's optimizations or similar.

While any implementation should know how to handle `on_tick`, `on_block` and `on_attestation`, so it shouldn't be a big deal to integrate such tests with the existing implementation. I validated that while [integrating my tests](https://github.com/ericsson49/teku/tree/integrateForkChoiceTests/eth-reference-tests/src/referenceTest/java/tech/pegasys/teku/reference/phase0/fork_choice) with Teku.

djrtwo:
2. The direct `store` manipulations are to avoid very inefficient state transitions and state copies found in `on_block`. We need to be careful not to inflate CI running time too much. A couple of options
    1. mod them and take the hit  on CI. Might be quite a hit though
    2. put a conditional in the helpers that does full calls to `on_block`, etc, when running testgen, but `store` manipulations on CI. Do we have a conditional that can detect context? This is nice, but then there is a worry that the logic isn't 1:1 and some bug surfaces during testgen that you didn't catch in CI
    3. write some separate testgen scenarios that only run during testgen and no CI. Probably better than (2) but still has the issue that something might surface during the rarely run testgens

ericsson49:
To reduce test vector generation time, I have been using either a caching BLS implementation or a dummy BLS implementation (which is completely insecure, but still able to distinguish good and bad signatures).
Actually, for most fork choice tests one can switch off BLS checks, with a a couple of exceptions (wrong block or attestation signatures).

djrtwo:
Right, and we have a test function decorator `never_bls` to do just that. The state trnasitions themselves can be very expensive in the unoptimized python as well though

djrtwo:
So do we leave the fork choice unit tests untouched?

tests/formats/fork_choice/README.md
```diff
@@ -0,0 +1,101 @@
+# Fork choice tests
+
+The aim of the tests for the fork choice rules.
```
djrtwo, line 3: This is a bit off. Maybe something like the following

```suggestion
The aim of the fork choice tests is to provide test coverage of the various components of the fork choice.
```

tests/formats/fork_choice/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Test case format
+
+### `meta.yaml`
+
+```yaml
+description: string    -- Optional. Description of test case, purely for debugging purposes.
```
djrtwo, line 10: Are we sing this anywhere? or do we intend to?

tests/formats/fork_choice/README.md
```diff
@@ -0,0 +1,101 @@
...
+bls_setting: int       -- see general test-format spec.
+```
+
+### `anchor_state.ssz_snappy`
+
+A YAML-encoded `BeaconState`, the state to initialize store with `get_forkchoice_store(anchor_state: BeaconState, anchor_block: BeaconBlock)` helper.
```
djrtwo, line 16:
```suggestion
An SSZ-snappy encoded `BeaconState`, the state to initialize store with `get_forkchoice_store(anchor_state: BeaconState, anchor_block: BeaconBlock)` helper.
```

tests/formats/fork_choice/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+### `anchor_block.ssz_snappy`
+
+A YAML-encoded `BeaconBlock`, the block to initialize store with `get_forkchoice_store(anchor_state: BeaconState, anchor_block: BeaconBlock)` helper.
```
djrtwo, line 20:
```suggestion
An SSZ-snappy encoded `BeaconBlock`, the block to initialize store with `get_forkchoice_store(anchor_state: BeaconState, anchor_block: BeaconBlock)` helper.
```

tests/formats/fork_choice/README.md
```diff
@@ -0,0 +1,101 @@
...
+The file is located in the same folder (see below).
+
+After this step, the `store` object may have been updated.
+
+#### `on_block` execution step
+
+The parameter that is required for executing `on_block(store, block)`.
+
+```yaml
+{ block: string }:        -- the name of the `block_<32-byte-root>.ssz_snappy` file. To execute `on_block(store, block)` with the given attestation.
```
djrtwo, line 52: why so much white space?

tests/formats/fork_choice/README.md
```diff
@@ -0,0 +1,101 @@
...
+    justified_checkpoint_root: '0x347468b606d03f8429afd491f94e32cd3a2295c2536e808c863a9d132a521dc4',
+    head: '0x17aa608f5fce87592c6f02ca6ca3c49ca70b5cef5456697709b2e5894e3879c2'
+}
+```
+
+### `attestation_<32-byte-root>.ssz_snappy`
+
+`<32-byte-root>` is the hash tree root of the given attestation.
+
+Each file is a YAML-encoded `Attestation`.
```
djrtwo, line 88:
```suggestion
Each file is an SSZ-snappy encoded `Attestation`.
```

tests/formats/fork_choice/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+### `block_<32-byte-root>.ssz_snappy`
+
+`<32-byte-root>` is the hash tree root of the given block.
+
+Each file is a YAML-encoded `SignedBeaconBlock`.
```
djrtwo, line 94:
```suggestion
Each file is an SSZ-snappy encoded `SignedBeaconBlock`.
```

tests/formats/fork_choice/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Condition
+
+1. Deserialize `anchor_state.ssz_snappy` and `anchor_block.ssz_snappy` to initialize the local store object by with `get_forkchoice_store(anchor_state, anchor_block)` helper.
+2. Go through `steps.yaml`
```
djrtwo, line 99: Maybe worth stating that it must be sequential. something like

```suggestion
2. Iterate sequentially through `steps.yaml`
```

tests/formats/fork_choice/README.md
```diff
@@ -0,0 +1,101 @@
...
+    - For each execution, look up the corresponding ssz_snappy file. Execute the corresponding helper function on your store.
```
djrtwo, line 100:
```suggestion
    - For each execution, look up the corresponding ssz_snappy file. Execute the corresponding helper function on the current store.
```

tests/formats/fork_choice/README.md
```diff
@@ -0,0 +1,101 @@
...
+    - For each `checks` step, the assertions must be satisfied.
```
djrtwo, line 101:
```suggestion
    - For each `checks` step, the assertions on the current store must be satisfied.
```

djrtwo (commented):
awesome! only minor comments

Should we remove the old unittest `test_get_head` file? This new file is strictly the previous functionality + generators, right?

tests/formats/fork_choice/README.md
```diff
@@ -0,0 +1,101 @@
...
+The file is located in the same folder (see below).
+
+After this step, the `store` object may have been updated.
+
+#### Checks step
+
+The checks to verify the current status of `store` .
+
+```yaml
+checks: {<store_field_name>: value}   -- the assertions.
```
protolambda, line 63: The tests produce `checks` objects that contain more than a single field, and they all have to be checked. Can you clarify this in the format docs?

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_get_head.py
```diff
@@ -0,0 +1,288 @@
...
+from eth2spec.test.helpers.state import (
+    next_epoch,
+    state_transition_and_sign_block,
+)
+
+
+@with_all_phases
+@with_configs([MINIMAL])
+@spec_state_test
+def test_genesis(spec, state):
```
protolambda, line 22: For Altair, this produces an Altair state at genesis, and Altair blocks right? The test is good (we may start Altair testnets without forking into Altair) but maybe this is a good place to use that `description` field to explain why it's used post-phase0.

protolambda (commented):
Left some minor comments, looks good to me, nice work :+1:

tests/formats/fork_choice/README.md
```diff
@@ -0,0 +1,101 @@
...
+justified_checkpoint_root: string           -- store.justified_checkpoint.root
+finalized_checkpoint_root: string           -- store.finalized_checkpoint_root.root
+best_justified_checkpoint_root: string      -- store.best_justified_checkpoint_root.root
+```
+
+For example:
+```yaml
+- checks: {
+    justified_checkpoint_root: '0x347468b606d03f8429afd491f94e32cd3a2295c2536e808c863a9d132a521dc4',
+    head: '0x17aa608f5fce87592c6f02ca6ca3c49ca70b5cef5456697709b2e5894e3879c2'
```
protolambda, line 80: Should we spec out the `head` check as a `(block root, slot)` combination, instead of just the block root?

tests/formats/fork_choice/README.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+`<store_field_name>` is the field member of [`Store`](../../../specs/phase0/fork-choice.md#store) object that maintained by client implementation. Currently, the possible fields included:
+
+```yaml
+time: int                                   -- store.time
+genesis_time: int                           -- store.genesis_time
+justified_checkpoint_root: string           -- store.justified_checkpoint.root
+finalized_checkpoint_root: string           -- store.finalized_checkpoint_root.root
+best_justified_checkpoint_root: string      -- store.best_justified_checkpoint_root.root
```
protolambda, line 73: The example lists `head`, but it's a function instead of a field, and not listed here.

tests/formats/fork_choice/README.md
hwwhww: Per Proto's suggestion (https://github.com/ethereum/eth2.0-specs/pull/2202#discussion_r593161385), added description for `test_genesis`.

djrtwo (approved):
lgtm!

PR #2206, opened 2021-02-20
Bump Milagro dependency for M1 support

JustinDrake:


ChihChengLiang (approved):
Looks good to me. This change is the same as what I expected.

hwwhww:
I verified that milagro_bls_binding 1.6.3 also works well on non-M1 macOS. 👍

hwwhww (approved):

Issue #2205, opened 2021-02-20
Deposit contract dapp test fatal error


kevin-chau:
I'm getting a fatal error when I run the dapp tests. It doesn't seem to stop the tests from running however, all of them seem to be passing for me:

```
kevinchau@Kevins-MBP eth2.0-specs % make test_deposit_contract
Makefile:168: warning: overriding commands for target `../eth2.0-spec-tests/tests'
Makefile:165: warning: ignoring old commands for target `../eth2.0-spec-tests/tests'
dapp test -v --fuzz-runs 5
fatal: cannot change to './solidity_deposit_contract/lib/ds-test/./solidity_deposit_contract': No such file or directory
child_process.js:650
    throw err;
    ^

Error: Command failed: git -C ./solidity_deposit_contract/lib/ds-test/./solidity_deposit_contract rev-parse HEAD
fatal: cannot change to './solidity_deposit_contract/lib/ds-test/./solidity_deposit_contract': No such file or directory

    at checkExecSyncError (child_process.js:629:11)
    at Object.execFileSync (child_process.js:647:13)
    at run (/nix/store/3m4ia98m1gia8d94shyarmkndx3hfiac-dapp-0.32.1/libexec/dapp/dapp-remappings:64:35)
    at ls.forEach.name (/nix/store/3m4ia98m1gia8d94shyarmkndx3hfiac-dapp-0.32.1/libexec/dapp/dapp-remappings:35:16)
    at Array.forEach (<anonymous>)
    at findRemappings (/nix/store/3m4ia98m1gia8d94shyarmkndx3hfiac-dapp-0.32.1/libexec/dapp/dapp-remappings:18:43)
    at Object.<anonymous> (/nix/store/3m4ia98m1gia8d94shyarmkndx3hfiac-dapp-0.32.1/libexec/dapp/dapp-remappings:5:1)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
Running 7 tests for ./solidity_deposit_contract/tests/deposit_contract.t.sol:DepositContractTest
+++ OK, passed 5 tests.
[PASS] testFail_malformed_calldata_sol(bytes,bytes,bytes,uint64) (runs: 5)
+++ OK, passed 5 tests.
[...]

PR #2208, opened 2021-02-21
Ensure the deposit contract has been compiled before running tests

hwwhww:
Address #2205


kevin-chau:
I opened #2205 -- while this is probably still a good change, I don't think it fixes the fatal error in that issue.

hwwhww:
Closing this PR since it doesn't solve #2205 and would break old CI jobs. (CI job `test_deposit_contract` and `install_deposit_contract_web3_tester` would have to install `solc` if we apply this change.)

PR #2212, opened 2021-02-23
64 epoch cleanup by @justindrake

djrtwo:
From @JustinDrake :

Pushed my review/cleanup. Largely just polishing :)

Cosmetic fixes
----

* rename `leak_score` to `leak_scores` (for consistency with `validators`, `balances`, etc.)
* rename "regular penalties" to "flag penalties" (for consistency with flag rewards)
* rename `ValidatorFlag` to `ParticipationFlags` (more precise; also pluralised because as there are 8 flags)
* initialise `ParticipationFlags` with `0b0000_0000` (instead of 0)
* introduce the concept of "flag mask" (e.g. rename `TIMELY_HEAD_FLAG` to `TIMELY_HEAD_FLAG_MASK`)
* disambiguate uses of "new" which sometimes means "modified" and sometimes "added"
* rename `is_activation_exit_period_boundary` to `is_activation_exit_epoch` (for consistency with `compute_activation_exit_epoch`)
* isolate leak updates in `process_leak_updates` (extracting logic from `process_rewards_and_penalties`) and call directly from `process_epoch`
* change type of `leak_epoch_counter` to `Epoch`
* rename `leak_epoch_counter` to `leak_epochs_counter`
* rename `SYNC_COMMITTEE_PUBKEY_AGGREGATES_SIZE` to `SYNC_SUBCOMMITTEE_SIZE`
* add `is_leak_active` (replacing and deprecating `is_in_inactivity_leak`)
* add `LEAK_PENALTY_QUOTIENT` (merging with `LEAK_SCORE_BIAS`; replacing and deprecating `INACTIVITY_PENALTY_QUOTIENT`)
* add `EPOCHS_TO_LEAK_PENALTIES` (replacing and deprecating `MIN_EPOCHS_TO_INACTIVITY_PENALTY`)
* remove unnecessary `for index in get_eligible_validator_indices(state)` in `process_leak_updates` and `get_flag_rewards`

Substantive fixes
----

* run `process_effective_balance_updates` only at activation-exit epochs
* properly round up in `compute_activation_exit_epoch` (notice special case when `epoch % EPOCHS_PER_ACTIVATION_EXIT_PERIOD == 0`)
* remove `and is_matching_target` in `process_attestation` when `append(TIMELY_HEAD_FLAG_MASK)`


specs/lightclient/beacon-chain.md
```diff
@@ -192,21 +187,19 @@ class BeaconState(Container):
...
-    # Light client sync committees
-    current_sync_committee: SyncCommittee
-    next_sync_committee: SyncCommittee
-    # How many inactivity leak epochs have there been in the previous penalty period?
-    leak_epoch_counter: uint64
-    # Increases when the validator misses epochs in an inactivity leak, decreases when the validator
-    # is online in an inactivity leak, inactivity leak penalties are proportional to this value
-    leak_score: List[uint64, VALIDATOR_REGISTRY_LIMIT]
+    # Leak
+    leak_epochs_counter: Epoch  # [New in HF1]
```
hwwhww, line 198: The definition of custom type `Epoch` is "which epoch", not "epoch count". So it should be `uint64`.
```suggestion
    leak_epochs_counter: uint64  # [New in HF1]
```

specs/lightclient/beacon-chain.md
```diff
@@ -235,13 +228,20 @@ def eth2_fast_aggregate_verify(pubkeys: Sequence[BLSPubkey], message: Bytes32, s
...
+def is_activation_exit_epoch(state: BeaconState) -> bool:
     return get_current_epoch(state) % EPOCHS_PER_ACTIVATION_EXIT_PERIOD == 0
 ```
 
+#### `is_leak_active`
+
+```python
+def is_leak_active(state: BeaconState) -> bool:
+    return get_finality_delay(state) > EPOCHS_TO_LEAK_PENALTIES
+```
```
hwwhww, line 243: It's a duplicate of the phase 0 version.

specs/lightclient/beacon-chain.md
```diff
@@ -703,79 +655,68 @@ def process_justification_and_finalization(state: BeaconState) -> None:
...
+        decrease_amount += LEAK_SCORE_BIAS  # Cancel corresponding increase at activation-exit epochs
+    for index in get_unslashed_participating_indices(state, TIMELY_TARGET_FLAG_INDEX, get_previous_epoch(state)):
+        if state.leak_scores[index] < decrease_amount:
+            state.leak_scores[index] = 0
+        else:
+            state.leak_scores[index] -= decrease_amount
+
+    # Increase leak score per rule (b)
+    if is_leak_active(state):
+        state.leak_epochs_counter += Epoch(1)
```
hwwhww, line 681: Epoch -> uint64

```suggestion
        state.leak_epochs_counter += uint64(1)
```

specs/lightclient/beacon-chain.md
```diff
@@ -703,79 +655,68 @@ def process_justification_and_finalization(state: BeaconState) -> None:
...
+    if is_activation_exit_epoch(state):
+        for index in get_eligible_validator_indices(state):
+            state.leak_scores[index] += LEAK_SCORE_BIAS * state.leak_epochs_counter
+        state.leak_epochs_counter = Epoch(0)
```
hwwhww, line 685: ditto
```suggestion
        state.leak_epochs_counter = uint64(0)
```

specs/lightclient/beacon-chain.md
```diff
@@ -703,79 +655,68 @@ def process_justification_and_finalization(state: BeaconState) -> None:
...
 
+
 ```python
         return
 
-    process_rewards(state)
-    process_penalties(state)
```
hwwhww, line 717: I can see this changes make sense, but IMHO it's good to have multiple purpose-focused functions. It would be more flexible when we have to modify it in the future forks.

specs/lightclient/beacon-chain.md
```diff
@@ -602,21 +555,19 @@ def process_sync_committee(state: BeaconState, body: BeaconBlockBody) -> None:
 
 ### Modified helpers
 
-#### New `compute_activation_exit_epoch`
+#### Modified `compute_activation_exit_epoch`
```
hwwhww, line 558: TODO: need to add some unit tests for it.

hwwhww (commented):
My first scan. 👀

> * remove `and is_matching_target` in `process_attestation` when `append(TIMELY_HEAD_FLAG_MASK)`

I can't find it in this PR. Was it left during the previous cleanups?

djrtwo:
I'm against changing constants names from phase 0 (and especially values e.g. `INACTIVITY_PENALTY_QUOTIENT`).

The general policy has been "never change constants at forks, only add new ones and update calling code". This helps keep configuration clean and avoid hard to track down changes and issues in testing (that might also surface in clients)

configs/mainnet/phase0.yaml
```diff
@@ -122,8 +122,8 @@ BASE_REWARD_FACTOR: 64
 WHISTLEBLOWER_REWARD_QUOTIENT: 512
 # 2**3 (= 8)
 PROPOSER_REWARD_QUOTIENT: 8
-# 2**26 (= 67,108,864)
-INACTIVITY_PENALTY_QUOTIENT: 67108864
+# 2**28 (= 268,435,456)
+LEAK_PENALTY_QUOTIENT: 268435456
```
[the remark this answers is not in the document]
djrtwo: I don't think we should rename any variables from phase 0. It's way to easy to bork test vectors and cause issues in clients, imo

configs/mainnet/phase0.yaml
```diff
@@ -99,7 +99,7 @@ MIN_VALIDATOR_WITHDRAWABILITY_DELAY: 256
 # 2**8 (= 256) epochs ~27 hours
 SHARD_COMMITTEE_PERIOD: 256
 # 2**2 (= 4) epochs 25.6 minutes
-MIN_EPOCHS_TO_INACTIVITY_PENALTY: 4
+EPOCHS_TO_LEAK_PENALTIES: 4
```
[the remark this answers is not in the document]
djrtwo: ditto

specs/lightclient/beacon-chain.md
```diff
@@ -123,17 +119,19 @@ This patch updates a few configuration values to move penalty constants toward t
...
+| `LEAK_SCORE_BIAS` | `uint64(2**2)` (= 4) |
+| `LEAK_PENALTY_QUOTIENT` | `uint64(2**28)` (= 268,435,456) |
 
 ### Time parameters
 
 | Name | Value | Unit | Duration |
 | - | - | :-: | :-: |
-| `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` | `Epoch(2**8)` (= 256) | epochs | ~27 hours |
-| `EPOCHS_PER_ACTIVATION_EXIT_PERIOD` | `Epoch(2**6)` (= 64) | epochs | ~6.8 hours |
+| `EPOCHS_TO_LEAK_PENALTIES` | `uint64(2**2)` (= 4) | epochs | 25.6 minutes |
```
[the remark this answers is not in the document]
djrtwo: leaving in here for now until we decide about renaming in general

specs/lightclient/beacon-chain.md
djrtwo: agreed

djrtwo (commented):
minor

specs/lightclient/beacon-chain.md
djrtwo: leaving in until we decide on cross-cutting changes on phases

specs/lightclient/beacon-chain.md
djrtwo: yeah, I would tend to agree

specs/lightclient/beacon-chain.md
```diff
@@ -343,88 +340,48 @@ def get_unslashed_participating_indices(state: BeaconState, flag_index: int, epo
...
 #### `get_leak_penalties`
 
 ```python
 def get_leak_penalties(state: BeaconState) -> Sequence[Gwei]:
-    """
-    Compute inactivity leak penalties for each validator.
-    """
     penalties = [Gwei(0)] * len(state.validators)
-
-    # Only give penalties on period boundaries
```
djrtwo, line 410: The problem with removing this blocking condition is that test vectors can very strangely be generated. That is, the vectors for `get_leak_penalties` specifically if run not on an epoch boundary will return non-zero values. I'd prefer keeping the blocking condition in to avoid misuse in alternative contexts even though the calling function in the spec had the conditional

djrtwo:
> remove and is_matching_target in process_attestation when append(TIMELY_HEAD_FLAG_MASK)

this is purposeful

1. it matches phase 0 functionality, see `get_matching_head_attestations` which calls `get_matching_target_attestations` as its base
2. this functionality was put into phase 0 to ensure that there was not a perverse scenario where you can essentially net 0 reward/penalty and not contribute to finality (e.g. get correct head but insert a bad target). It is instead natural for correctness to stem from source into target up to head (voting on a cohesive chain) rather than each being independent

djrtwo:
This optimization will *not* go into Altair. More R&D is desired before making such spec optimizations.

See call notes here https://github.com/ethereum/eth2.0-pm/issues/208

PR #2214, opened 2021-02-24
ensure venvs build with python3.8 in makefile

djrtwo:
Some virtual-envs were being built with `python3` instead of explicitly `python3.8` in the Makefile. This can lead to issues with local system using `python3.9` for some of the test generators which breaks some of the build assumptions.

This PR sets all construction of virtual-envs in the Makefile to explicitly use `python3.8` until we support later versions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,13 +64,13 @@ clean: partial_clean
 
 # The pyspec is rebuilt to enforce the /specs being part of eth2specs source distribution. It could be forgotten otherwise.
 dist_build: pyspec
-	python3 setup.py sdist bdist_wheel
+	python3.8 setup.py sdist bdist_wheel
 
 dist_check:
-	python3 -m twine check dist/*
+	python3.8 -m twine check dist/*
 
 dist_upload:
-	python3 -m twine upload dist/*
+	python3.8 -m twine upload dist/*
 
 
 # "make generate_tests" to run all generators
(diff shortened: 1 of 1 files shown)


hwwhww:
> This can lead to issues with local system using `python3.9` for some of the test generators which breaks some of the build assumptions.

1. I thought that since the generators' venvs are also initiated with the `python3` command, they should be on the same Python version (3.8 or 3.9) as the pyspec `venv`? I just tried py3.9 on CI: https://github.com/ethereum/eth2.0-specs/commit/9fa13060f8916268cd8198ba2884842ef3b1683b (lint checks failed though). If the pyspec and generator `venv`s are both py3.9, it should be all good.
2. I remember some pyspec devs and users are using py3.9. I suppose these users also use `Makefile` to run the tests so this PR might break their user experience. 😅 (summon @dankrad @ralexstokes)

JustinDrake:
I would also set a version of pip. It turns out that for the Apple M1 the build fails with the pip version that comes out of the box.

djrtwo:
hmm, I was having local build dependency issues with 3.9. Not sure I have it documented on the precise build issue. I'll see if I can reproduce

hwwhww:
> I'll see if I can reproduce

@djrtwo did you build on `master` branch? Reminded by #2217, it's probably because we didn't cherry-pick https://github.com/ethereum/eth2.0-specs/pull/2169 to `master`, so the `Makefile` is not as same as the `dev` branch.

PR #2221, opened 2021-03-09
Make `test_fork.py` truly pass (on #2220)

hwwhww:
A patch for #2220

- Fix context decorators
- Reorder decorator sequence
- Since it has to start with phase 0 state, it should be `@with_phases(([PHASE0]))`.
- Make `generate_from_tests` more flexible: the "phase name" and "fork name" could be different now.

~~Still having problems:~~
- ~~Somehow it didn't output `meta.yaml`.~~
- ~~Somehow it didn't output `post.yaml`.~~
- ~~Still having errors when generating `test_fork_random_large_validator_set` and `test_fork_random_misc_balances` tests.~~

### DIFF
--- a/tests/core/pyspec/eth2spec/test/context.py
+++ b/tests/core/pyspec/eth2spec/test/context.py
@@ -82,7 +82,7 @@ def _prepare_state(balances_fn: Callable[[Any], Sequence[int]], threshold_fn: Ca
         # TODO: instead of upgrading a test phase0 genesis state we can also write a phase1 state helper.
         # Decide based on performance/consistency results later.
         state = phases[PHASE1].upgrade_to_phase1(state)
-    elif spec.fork == LIGHTCLIENT_PATCH and not is_fork_test:  # do not upgrade if spec ttttest
+    elif spec.fork == LIGHTCLIENT_PATCH and not is_fork_test:  # do not upgrade if spec test
         state = phases[LIGHTCLIENT_PATCH].upgrade_to_lightclient_patch(state)
 
     return state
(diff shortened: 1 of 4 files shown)


tests/generators/forks/main.py
```diff
@@ -1,40 +1,38 @@
...
+        reload(spec_phase0)
+        reload(spec_lightclient_patch)
         return config_name
 
     def cases_fn() -> Iterable[gen_typing.TestCase]:
         return generate_from_tests(
-            runner_name='forks',
-            handler_name='core',
+            runner_name='fork',
+            handler_name='fork',
```
hwwhww, line 25: Following the pattern of `finality` test vector output, I made it follow the folder name and the `test_*` filename.

tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -82,7 +82,7 @@ def _prepare_state(balances_fn: Callable[[Any], Sequence[int]], threshold_fn: Ca
         # TODO: instead of upgrading a test phase0 genesis state we can also write a phase1 state helper.
         # Decide based on performance/consistency results later.
         state = phases[PHASE1].upgrade_to_phase1(state)
-    elif spec.fork == LIGHTCLIENT_PATCH and not fork_test:  # do not upgrade if spec ttttest
+    elif spec.fork == LIGHTCLIENT_PATCH and not is_fork_test:  # do not upgrade if spec ttttest
```
djrtwo, line 85: We'll need to make this more generic eventually when doing fork tests for next fork, but... probably fine for now
```suggestion
    elif spec.fork == LIGHTCLIENT_PATCH and not is_fork_test:  # do not upgrade if spec test
```

djrtwo (approved):
thank you!!

djrtwo:
turns out we don't need `fork_test` with the flexibility you added to the generators. wooo

PR #2222, opened 2021-03-09
independent leak score feature

djrtwo:
Port validator specific leak from #2192 and #2212 to be reviewed and judged independently from 64-epoch boundary optimization


specs/lightclient/beacon-chain.md
```diff
@@ -284,13 +285,10 @@ def get_sync_committee(state: BeaconState, epoch: Epoch) -> SyncCommittee:
...
     """
     indices = get_sync_committee_indices(state, epoch)
-    validators = [state.validators[index] for index in indices]
-    pubkeys = [validator.pubkey for validator in validators]
-    aggregates = [
-        bls.AggregatePKs(pubkeys[i:i + SYNC_SUBCOMMITTEE_SIZE])
-        for i in range(0, len(pubkeys), SYNC_SUBCOMMITTEE_SIZE)
-    ]
-    return SyncCommittee(pubkeys=pubkeys, pubkey_aggregates=aggregates)
+    pubkeys = [state.validators[index].pubkey for index in indices]
```
djrtwo, line 288: Note, this is a cleanup from #2212 that is being ported as well. Should be 1:1 functionality

tests/core/pyspec/eth2spec/test/helpers/rewards.py
```diff
@@ -219,7 +221,8 @@ def run_get_inactivity_penalty_deltas(spec, state):
 
 def transition_state_to_leak(spec, state, epochs=None):
     if epochs is None:
-        epochs = spec.MIN_EPOCHS_TO_INACTIVITY_PENALTY
+        # +1 to trigger leak_score transitions
+        epochs = spec.MIN_EPOCHS_TO_INACTIVITY_PENALTY + 1
```
hwwhww, line 225: Sanity check, what exactly the `@leaking` decorator should do? Is it for getting a `state` such that `is_in_inactivity_leak(state) is True`?

If so, should this line be `epochs = spec.MIN_EPOCHS_TO_INACTIVITY_PENALTY + 2`, similar to `test_leak_scores` test case in `lightclient_patch/sanity/test_blocks.py`?

tests/core/pyspec/eth2spec/test/phase0/block_processing/test_process_attester_slashing.py
```diff
@@ -118,6 +126,41 @@ def test_success_already_exited_recent(spec, state):
...
+
+@with_all_phases
+@with_custom_state(balances_fn=misc_balances, threshold_fn=lambda spec: spec.EJECTION_BALANCE)
+@spec_test
+@single_phase
+def test_success_with_effective_balance_disparity(spec, state):
+    # Jitter balances to be different from effective balances
+    for i in range(len(state.balances)):
+        pre = int(state.balances[i])
+        state.balances[i] += random.randrange(max(pre - 5000, 0), pre + 5000)
```
hwwhww, line 157: Do you mean
```suggestion
        state.balances[i] = random.randrange(max(pre - 5000, 0), pre + 5000)
```
?

hwwhww (approved):
Great! The spec changes look good to me. Just some questions about the tests.

tests/core/pyspec/eth2spec/test/helpers/rewards.py
djrtwo: So @leaking used to just get the state such taht `is_in_inactivity_leak` is True, but this is not enough for the test cases that use `leak_scores` because on the cusp of the transition, `is_in_inactivity_leak is False` and thus `leak_scores` aren't updated to be non-zero. Whereas once you cross the next transition, you do get increased leak scores. So beacuse we want to test that running the rewards/penalties functions in isolation gives leak penalties, we want some leak_scores to be non-zero

the +2 in `test_leak_scores` is to account for if it is starting at genesis where `finalized_epoch == current_epoch` so it takes another couple of epoch transitions to get to the leak

tests/core/pyspec/eth2spec/test/phase0/epoch_processing/test_process_rewards_and_penalties.py
```diff
@@ -27,6 +27,31 @@ def run_process_rewards_and_penalties(spec, state):
...
+                    # If not proposer but participated optimally, should have exactly neutral balance
+                    assert post_state.balances[index] == pre_state.balances[index]
+                else:
+                    assert post_state.balances[index] < pre_state.balances[index]
+            else:
+                if index in attesting_indices:
+                    assert post_state.balances[index] > pre_state.balances[index]
+                else:
+                    assert post_state.balances[index] < pre_state.balances[index]
+
```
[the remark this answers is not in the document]
djrtwo: ah, I think it should just be an `else`. looking into it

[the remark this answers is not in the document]
djrtwo: nevermind. fixing it up!

Issue #2183, opened 2021-01-14
Propagate seen aggregate attestations


AgeManning:
Currently, the gossipsub validation rules for the `beacon_aggregate_and_proof` topic specify that a client should `IGNORE` messages that contain an aggregate that has previously been seen. 

> [IGNORE] The valid aggregate attestation defined by hash_tree_root(aggregate) has not already been seen (via aggregate gossip, within a verified block, or through the creation of an equivalent aggregate locally).

This gives rise to a few issues within the gossipsub protocol. 

There are on average, `TARGET_AGGREGATORS_PER_COMMITTEE` (16 for mainnet) aggregators per committee. These aggregators will often agree on the same aggregate attestation and submit individual `SignedAggregateAndProof` objects to the `beacon_aggregate_and_proof` topic. These messages are unique (as the aggregators and hence their signatures are unique) and therefore have their own unique gossipsub message-ids.

Therefore the standard operation of a node is to receive one valid aggregate and then `IGNORE` (aka not forward, but still gossip about) any remaining duplicate aggregate attestations as specified by the topic rule mentioned above. Different nodes on the network may receive different aggregates as their first and `IGNORE` others. The gossipsub protocol will then gossip about the extra message-ids that were not forwarded. Other nodes will then notice they didn't receive these extra aggregates (gossipsub message ids) and request the unknown gossip-ids via gossipsub IWANT messages from their peers. Eventually all the aggregates should still propagate to all nodes in the network, albeit inefficiently. 

Besides the inefficient propagation of duplicate aggregate attestations, peers may also get penalized for not propagating all the messages on the topic (i.e it appears some nodes are censoring some message-ids as they don't forward them). 

An easy solution to handling the scoring issue and inefficiencies of propagating via IHAVE/IWANT gossip, is to relax the topic requirement such that all nodes still forward duplicate aggregate attestations, provided the signed aggregate has not been seen before. This will make the propagation more efficient and reduce the messages being sent and received via IHAVE/IWANT. 
Potentially in the future, we may want to think about a way to filter out the duplicate aggregate attestations to save bandwidth, but this may involve complicated changes to the message-id function and is something we can address in the future.

djrtwo:
Adding this to HF1

ralexstokes:
Making a note the outcome here may be relevant for the sync committee aggregation scheme we are working on in https://github.com/ethereum/eth2.0-specs/pull/2201

AgeManning:
[...]

PR #2225, opened 2021-03-11
remove faulty de-duplication condition for seen aggregates

djrtwo:
Addresses #2183

The alternative would be to have a more exotic message-id that is only a function of `aggregate_and_proof.aggregate`.

I suggest we observe mainnet's aggregation data and make a call.


protolambda (approved):
LGTM. May cost some bandwidth, but is more robust, and works with what we have.

PR #2229, opened 2021-03-11
[WIP] The Merge

mkalinin:
### The Merge
Work in progress. Introduces spec changes to implement the executable beacon chain proposal.

Executable beacon chain is the version of Ethereum Mainnet with PoS consensus driven by the Beacon Chain. Consensus Upgrade (from PoW to PoS) is the core change that is coming with the Merge.

Based on
- https://ethresear.ch/t/executable-beacon-chain/8271

Related
- https://hackmd.io/@n0ble/eth1-eth2-communication-protocol-draft

Implementations
- Teku https://github.com/txrx-research/teku/tree/executable-beacon-chain
- Catalyst https://github.com/gballet/go-ethereum/tree/catalyst-for-executable-beacon-chain

TODO
- [ ] make the spec executable
- [ ] run existing tests with stubbed application payload
- [ ] add tests covering application payload
- [ ] rebase to Altair once ready

### DIFF
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -173,7 +173,10 @@ def compute_time_at_slot(state: BeaconState, slot: Slot) -> uint64:
 
 ```python
 def get_recent_beacon_block_roots(state: BeaconState, qty: uint64) -> Sequence[Bytes32]:
-    return [get_block_root_at_slot(state.slot - i) if GENESIS_SLOT + i < state.slot else Bytes32() for i in reversed(range(1, qty + 1))]
+    return [
+        get_block_root_at_slot(state.slot - i) if GENESIS_SLOT + i < state.slot else Bytes32()
+        for i in reversed(range(1, qty + 1))
+    ]
 ```
 
 #### `get_evm_beacon_block_roots`
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -215,7 +215,7 @@ class BeaconChainData(Container):
     slot: Slot
     randao_mix: Bytes32
     timestamp: uint64
-    recent_block_roots: Sequence[Bytes32]
+    recent_block_roots: Vector[Bytes32, EVM_BLOCK_ROOTS_SIZE]
 ```
 
 ##### `get_application_state`
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -1,0 +1,257 @@
+# Ethereum 2.0 The Merge
+
+**Notice**: This document is a work-in-progress for researchers and implementers.
+
+## Table of contents
+
+<!-- TOC -->
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Introduction](#introduction)
+- [Constants](#constants)
+  - [Execution](#execution)
+- [Containers](#containers)
+  - [Extended containers](#extended-containers)
+    - [`BeaconBlockBody`](#beaconblockbody)
+    - [`BeaconState`](#beaconstate)
+  - [New containers](#new-containers)
+    - [`Transaction`](#transaction)
+    - [`ApplicationPayload`](#applicationpayload)
+- [Helper functions](#helper-functions)
+  - [Misc](#misc)
+    - [`compute_randao_mix`](#compute_randao_mix)
+    - [`compute_time_at_slot`](#compute_time_at_slot)
+  - [Beacon state accessors](#beacon-state-accessors)
+    - [`get_recent_beacon_block_roots`](#get_recent_beacon_block_roots)
+    - [`get_evm_beacon_block_roots`](#get_evm_beacon_block_roots)
+  - [Block processing](#block-processing)
+    - [Modified `process_eth1_data`](#modified-process_eth1_data)
+    - [Application payload processing](#application-payload-processing)
+      - [`BeaconChainData`](#beaconchaindata)
+      - [`get_application_state`](#get_application_state)
+      - [`application_state_transition`](#application_state_transition)
+      - [`process_application_payload`](#process_application_payload)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+## Introduction
+
+This is a patch implementing executable beacon chain proposal. 
+It enshrines application execution and validity as a first class citizen at the core of the beacon chain.
+
+## Constants
+
+### Execution
+
+| Name | Value |
+| - | - |
+| `MAX_BYTES_PER_TRANSACTION_PAYLOAD` | `2**20` |
+| `MAX_APPLICATION_TRANSACTIONS` | `2**14` |
+| `BYTES_PER_LOGS_BLOOM` | `2**8` |
+| `EVM_BLOCK_ROOTS_SIZE` | `2**8` |
+
+
+## Containers
+
+### Extended containers
+
+*Note*: Extended SSZ containers inherit all fields from the parent in the original
+order and append any additional fields to the end.
+
+#### `BeaconBlockBody`
+
+*Note*: `BeaconBlockBody` fields remain unchanged other than the addition of `application_payload`.
+
+```python
+class BeaconBlockBody(phase0.BeaconBlockBody):
+    application_payload: ApplicationPayload  # User execution payload
+```
+
+#### `BeaconState`
+
+*Note*: `BeaconState` fields remain unchanged other than the removal of `eth1_data_votes` and addition of `application_state_root`. 
+The latter stores the root hash of ethereum application state.
+
+```python
+class BeaconState(Container):
+    # Versioning
+    genesis_time: uint64
+    genesis_validators_root: Root
+    slot: Slot
+    fork: Fork
+    # History
+    latest_block_header: BeaconBlockHeader
+    block_roots: Vector[Root, SLOTS_PER_HISTORICAL_ROOT]
+    state_roots: Vector[Root, SLOTS_PER_HISTORICAL_ROOT]
+    historical_roots: List[Root, HISTORICAL_ROOTS_LIMIT]
+    # Eth1
+    eth1_data: Eth1Data
+    # [removed] eth1_data_votes
+    eth1_deposit_index: uint64
+    # [new] Hash of the root of application state
+    application_state_root: Bytes32
+    # [new] Hash of recent application block
+    application_block_hash: Bytes32
+    # Registry
+    validators: List[Validator, VALIDATOR_REGISTRY_LIMIT]
+    balances: List[Gwei, VALIDATOR_REGISTRY_LIMIT]
+    # Randomness
+    randao_mixes: Vector[Bytes32, EPOCHS_PER_HISTORICAL_VECTOR]
+    # Slashings
+    slashings: Vector[Gwei, EPOCHS_PER_SLASHINGS_VECTOR]  # Per-epoch sums of slashed effective balances
+    # Attestations
+    previous_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
+    current_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
+    # Finality
+    justification_bits: Bitvector[JUSTIFICATION_BITS_LENGTH]  # Bit set for every recent justified epoch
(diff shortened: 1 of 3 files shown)


protolambda:
Marking as "do not merge" on request by @mkalinin. Looking forward to the merge of the merge though!

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,257 @@
...
+The body of the function is implementation dependant.
+
+*Note*: `application_state_transition` must throw `AssertionError` if either the transition itself or one of the post-transition verifications has failed.
+
+##### `process_application_payload`
+
+```python
+def process_application_payload(state: BeaconState, body: BeaconBlockBody) -> None:
+    """
+    Note: This function is designed to be able to be run in parallel with
```
arnetheduck, line 239: It's worth noting that the other `process_block` functions in eth2 take ~2ms on average hardware - pragmatically there's not _that_ much to be gained from parallelization here - I would guess this falls in the category of being a "nice touch" - it makes testing and reasoning easier for example when functions are "pure" in the order-independent sense.

In general, making the eth1 state transition synchronous here leaves smaller margins for attestation production, an issue we're already seeing during finalization today, when the beacon chain is doing "nothing" - it would be good to define behaviors / describe more in the validator guide what the client should do when eth1 is momentarily unable to produce a block on time (for whatever reason, hickup in the client, hardware etc), what the _next_ proposer should do and other failure scenarios, as well as investigate / simulate tolerances for failures and delays.

arnetheduck: To clarify, the above comment is not taking into account future eth2 data requirements of course - the 2ms time is from current load which is mainly CPU-bound - with more data and shards, these timings may change dramatically, but at that point eth1 and eth2 might end up competing for the same (scarce) resource, disk, which also is hard to parallelize.

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,257 @@
...
+    # the processing of other `process_block` sub-functions
+    beacon_chain_data = BeaconChainData(
+        slot=state.slot,
+        randao_mix=compute_randao_mix(state, body.randao_reveal),
+        timestamp=compute_time_at_slot(state.genesis_time, state.slot),
+        recent_block_roots=get_evm_beacon_block_roots(state) 
+    )
+    
+    application_state = get_application_state(state.application_state_root)
+    application_state_transition(application_state, beacon_chain_data, body.application_payload)
```
terencechain, line 253: Is the beacon node holding the application  state?

It seems to me that we can just do:

```python
application_state_transition(state.application_state_root, beacon_chain_data, body.application_payload)
```
That saves 1 round trip of retrieving and passing the application state

specs/merge/beacon-chain.md
mkalinin: Concerns you have listed above are important and addressing them is a part of readiness criteria for consensus upgrade. Transaction throughput (the size of the application block) is one of the evaluation criteria of the chain after the upgrade, attesters are in a heavy dependency on this property and time restriction that we have throughout a slot turns into restriction on the block size. The other potential approach to pushing the limits is [delayed execution](https://ethresear.ch/t/commit-to-pre-state-instead-of-post-state-on-the-executable-beacon-chain/8802) which comes with new complexity related to verification of gas limit per block.

Proposers may follow the _pending block_ technique in a way it is currently implemented by Ethereum clients which means that application block is always ready in advance and proposer just needs to pick it up.

The other side of the problem of synchronous interaction between consensus and application chains is resource exhaustion attack vector, its analysis and mitigation.

All the above is out of the scope of this spec but definitely is on the critical path to the upgrade.

specs/merge/beacon-chain.md
mkalinin: > Is the beacon node holding the application state?

This is implementation dependant. If there is a client that connected to both networks (consensus and application) then it has direct access to the application state. The other type of client has separated pieces of software representing consensus and application parts, these pieces are interacting via RPC protocol. This is the reason behind keeping the spec abstract in that regard.

The other reason for that is the difference between commitment schemes of beacon state and application state, that is binary tree vs MPT. If the scheme were the same we could think of _embedding_ application state into the beacon state. Then application state would be a part of the beacon state for the monolithic type of clients and could be still represented as a state root for clients of separated architecture.

mkalinin: > It seems to me that we can just do:
> ```python
> application_state_transition(state.application_state_root, beacon_chain_data, body.application_payload)
> ```
> That saves 1 round trip of retrieving and passing the application state

Currently, it follows a logic similar to beacon chain `state_transition` function that accepts the `(state, block)` tuple and mutates the state by applying the state transition.

jpollard-cs:
As a new miner I'm a bit bummed to see this expedited, but this 51% attack threat is ridiculous and short-sighted. For what it's worth - you have my support 100%.

mratsim:
> As a new miner I'm a bit bummed to see this expedited, but this 51% attack threat is ridiculous and short-sighted. For what it's worth - you have my support 100%.
> 
> Also why not use a draft PR? smile

The expedited quick merge described by Vitalik (see https://notes.ethereum.org/@vbuterin/B1mUf6DXO) is different and would happen before this PR.

> Does this mean we’re throwing out the plans to make the beacon chain interact with the application state more directly as Mikhail Kalinin and co are working on?
> 
> No. Rather, the minimal merge as described above would only be the first step, and a post-merge hard fork would simplify the protocol and clean out unnecessary bits and make the interaction between the application layer and the beacon chain more “natural”. Eventually we would go even further and, for example, replace the application_block: bytes with application_txs: List[bytes]) and even make the application state a “normal” part of the BeaconState (even though it’s hashed in a non-SSZ way).
>
> In fact, the minimal merge will make these things easier to work on, as it would no longer be needed to worry about merging two chains at the same time.

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,257 @@
...
+def compute_randao_mix(state: BeaconState, randao_reveal: BLSSignature) -> Bytes32:
+    epoch = get_current_epoch(state)
+    return xor(get_randao_mix(state, epoch), hash(randao_reveal))
+```
+
+#### `compute_time_at_slot` 
+
+```python
+def compute_time_at_slot(state: BeaconState, slot: Slot) -> uint64:
+    return uint64(state.genesis_time + slot * SECONDS_PER_SLOT)
```
paulhauner, line 166: To be pedantic, this should account for a non-zero `GENESIS_SLOT`.

```suggestion
    slots_since_genesis = slot - GENESIS_SLOT
    return uint64(state.genesis_time + slots_since_genesis * SECONDS_PER_SLOT)
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,257 @@
...
+##### `BeaconChainData`
+
+*Note*: `BeaconChainData` contains beacon state data that is used by the application state transition function.
+
+```python
+class BeaconChainData(Container):
+    slot: Slot
+    randao_mix: Bytes32
+    timestamp: uint64
+    recent_block_roots: Sequence[Bytes32]
```
paulhauner, line 217: `Sequence` doesn't seem to be handled in SSZ or tree hashing. It doesn't seem like we *necessarily* need to serialize/hash it, but if we're going to send it over the wire to Catalyst/Eth1 then perhaps using something that has SSZ methods would be good forsight?

Here's a suggestion, which doesn't necessarily work at the moment because we do `min(EVM_BLOCK_ROOTS_SIZE, SLOTS_PER_HISTORICAL_ROOT)`:

```suggestion
    recent_block_roots: Vector[Bytes32, EVM_BLOCK_ROOTS_SIZE]
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,257 @@
...
+```python
+def get_recent_beacon_block_roots(state: BeaconState, qty: uint64) -> Sequence[Bytes32]:
+    return [get_block_root_at_slot(state.slot - i) if GENESIS_SLOT + i < state.slot else Bytes32() for i in reversed(range(1, qty + 1))]
+```
+
+#### `get_evm_beacon_block_roots`
+
+```python
+def get_evm_beacon_block_roots(state: BeaconState) -> Sequence[Bytes32]:
+    num_block_roots = min(EVM_BLOCK_ROOTS_SIZE, SLOTS_PER_HISTORICAL_ROOT)
```
paulhauner, line 182: Personally, I would lean towards avoiding this `min` between constants and just accept that the system explodes if `SLOTS_PER_HISTORICAL_ROOT < EVM_BLOCK_ROOTS_SIZE`. It wouldn't be the first time that the spec relies on some constant being magically correct without the system exploding (which is completely reasonable IMO). We could always add a comment.

The main reason I mention this is because it would work nicely with [my later comment](https://github.com/ethereum/eth2.0-specs/pull/2229/files#r594078643) about setting `BeaconChainData::recent_block_roots` to be fixed length; we wouldn't need to use a `min` in the type definition of the length of that field.

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,257 @@
...
+    return uint64(state.genesis_time + slot * SECONDS_PER_SLOT)
+```
+
+### Beacon state accessors
+
+#### `get_recent_beacon_block_roots`
+
+```python
+def get_recent_beacon_block_roots(state: BeaconState, qty: uint64) -> Sequence[Bytes32]:
+    return [get_block_root_at_slot(state.slot - i) if GENESIS_SLOT + i < state.slot else Bytes32() for i in reversed(range(1, qty + 1))]
```
djrtwo, line 175: line length
```suggestion
    return [
        get_block_root_at_slot(state.slot - i) if GENESIS_SLOT + i < state.slot else Bytes32()
        for i in reversed(range(1, qty + 1))
    ]
```

specs/merge/beacon-chain.md
djrtwo: Yeah, agreed. I suggest writing down any requisite constants relationships as you come by them in this process @mkalinin

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,257 @@
...
+    timestamp: uint64
+    recent_block_roots: Sequence[Bytes32]
+```
+
+##### `get_application_state`
+
+*Note*: `ApplicationState` class is an abstract class representing ethereum application state.
+
+Let `get_application_state(application_state_root: Bytes32) -> ApplicationState`  be the function that given the root hash returns a copy of ethereum application state. 
+The body of the function is implementation dependant.
```
djrtwo, line 225: spelling

```suggestion
The body of the function is implementation dependent.
```

specs/merge/beacon-chain.md
djrtwo: One thing to note is that the primary attestation issue today is seen on epoch boundaries where the epoch transition  must be processed before the block is created *and* before the block is gossiped at each step (to get the correct proposer shuffling to validate the signature).

My estimation is that this is causing late receiving of blocks across the network and thus the incorrect/late attestations. This epoch (shuffling) processing problem is *independent* of computation you do after block is gossiped/forwarded, and the epoch processing issue is a major target for optimizations.

If the epoch boundary block can be gossiped quickly (same speed as non-epoch boundary blocks), then there should be ample time (3s+) to compute the block's payload to attest. Eth1 blocks today are processed in 200ms so as long as you don't do that processing as a gossip validation, I don't see this compounding our current issue.

djrtwo: Note! another parallelization design goal is that it ensures the application layer could be pulled out into a shard. By reducing the same-slot-processing dependencies between beacon and application-layer, we keep future designs open

djrtwo (commented):
quick pass. a couple of nits and some comments

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,257 @@
...
+    # Slashings
+    slashings: Vector[Gwei, EPOCHS_PER_SLASHINGS_VECTOR]  # Per-epoch sums of slashed effective balances
+    # Attestations
+    previous_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
+    current_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
+    # Finality
+    justification_bits: Bitvector[JUSTIFICATION_BITS_LENGTH]  # Bit set for every recent justified epoch
+    previous_justified_checkpoint: Checkpoint  # Previous epoch snapshot
+    current_justified_checkpoint: Checkpoint
+    finalized_checkpoint: Checkpoint
```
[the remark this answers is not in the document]
mkalinin: It was decided to wait until Altair got released and rebase thereafter.

djrtwo:
closing in favor of #2257 
Will leave branch open for now in case we need for reference

PR #2234, opened 2021-03-13
[minor] Fix some leftover from #2097 (ssz-snappy)

hwwhww:
Just some typo fixes.


protolambda (approved):
LGTM. My bad for writing the `SSZ` thing wrong at first, wasn't sure due to EsEsZee or SimpleSerialiZe difference. Current SSZ spec also uses `an SSZ object` instead of `a SSZ ...` so let's use that.

PR #2236, opened 2021-03-15
Altair cosmetic cleanups plus a couple substantive changes

JustinDrake:
Substantive changes:

1) separately apply inactivity penalty for every flag index in `get_inactivity_penalty_deltas` (this changes the rounding logic to exactly cancel out `get_flag_index_deltas`; cleaner and avoids unnecessarily updating validator balances)
2) remove the `and state.leak_scores[index] >= LEAK_SCORE_BIAS` condition in `get_inactivity_penalty_deltas`
3) (TODO) rework rewards in `process_sync_committee` to maintain the invariant of at most one base reward per validator per epoch.


JustinDrake:
> The PROPORTIONAL_SLASHING_MULTIPLIER is set to 1 at initial mainnet launch, resulting in one-third of the minimum accountable safety margin in the event of a finality attack. After Phase 0 mainnet stablizes, this value will be upgraded to 3 to provide the maximal minimum accountable safety margin.

Why is `ALTAIR_PROPORTIONAL_SLASHING_MULTIPLIER` being set to 2 not 3?

djrtwo:
> Why is ALTAIR_PROPORTIONAL_SLASHING_MULTIPLIER being set to 2 not 3?

In @vbuterin's initial HF1 doc, he suggested stepping toward maximal penalties rather than going all the way there. The idea being that we still haven't gotten to the merge and as long as we move directionally, we'll set the precedent to get there over time. We had also seen a number of slashings  around that time so it didn't seem users had quite gotten non-slashing correct yet.

specs/altair/beacon-chain.md
```diff
@@ -514,15 +495,13 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
...
-        state.validators.append(get_validator_from_deposit(state, deposit))
-        state.balances.append(amount)
-        state.previous_epoch_participation.append(ParticipationFlags(0b0000_0000))  # [New in Altair]
-        state.current_epoch_participation.append(ParticipationFlags(0b0000_0000))  # [New in Altair]
-        state.leak_scores.append(0)  # [New in Altair]
+        # Initialize validator if the deposit signature is valid
+        if bls.Verify(pubkey, signing_root, deposit.data.signature):
+            state.validators.append(get_validator_from_deposit(state, deposit))
+            state.balances.append(amount)
+            state.previous_epoch_participation.append(ParticipationFlags(0b0000_0000))
```
djrtwo, line 502: Why do you prefer these three statements to not have the `[New in Altair]` comment?

djrtwo (approved):

specs/altair/beacon-chain.md
JustinDrake: It's for consistency but I don't feel too strongly. We didn't add `[New in Altair]` or `[Modified in Altair]` inside other functions called with `process_epoch` (such as `process_justification_and_finalization`, `process_rewards_and_penalties`, `process_slashings`).

JustinDrake:
> he suggested stepping toward maximal penalties rather than going all the way there

Ok, understood. Does it make sense to explicitly mention that in `beacon-chain.md` (similar to what was done in phase 0)?

specs/altair/beacon-chain.md
djrtwo: I'm good either way

PR #2240, opened 2021-03-15
Enable Altair genesis tests + rename `SpecLightclient` -> `SpecAltair` + disable too slow test

hwwhww:
- As Proto reminded us earlier, we may start a testnet with Altair spec. This PR enables the genesis tests with Altair spec.
- Rename `SpecLightclient` -> `SpecAltair`
- Disable `test_sync_committees_progress` + mainnet config. Too slow.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/phase0/genesis/test_initialization.py
+++ b/tests/core/pyspec/eth2spec/test/phase0/genesis/test_initialization.py
@@ -1,14 +1,21 @@
-from eth2spec.test.context import PHASE0, spec_test, with_phases, single_phase
+from eth2spec.test.context import PHASE0, ALTAIR, spec_test, with_phases, single_phase, is_post_altair
 from eth2spec.test.helpers.deposits import (
     prepare_full_genesis_deposits,
     prepare_random_genesis_deposits,
 )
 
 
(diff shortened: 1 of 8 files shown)


tests/core/pyspec/eth2spec/test/phase0/genesis/test_initialization.py
```diff
@@ -1,14 +1,29 @@
...
     prepare_random_genesis_deposits,
 )
 
 
-@with_phases(([PHASE0]))
+def get_post_altair_description(spec):
+    return f"Although it's not phase 0, we may use {spec.fork} spec to start testnets."
+
+
+@with_phases(([PHASE0, ALTAIR]))
```
djrtwo, line 19: Should we even specify the phase list? instead of all by default?

djrtwo (approved):
looks good. one minor question

tests/core/pyspec/eth2spec/test/phase0/genesis/test_initialization.py
hwwhww: I somehow had an impression that `PHASE1` doesn't work for genesis tests. But it turns out `@with_all_phases` work well. Thanks!

Issue #2231, opened 2021-03-12
Consider renaming ALTAIR_* constants


mcdee:
There are a number of constants in the spec such as 'ALTAIR_INACTIVITY_PENALTY_QUOTIENT'.  Having the fork name first makes it slightly harder to parse mentally, as well as being less amenable to reading after sorting the constants by name.

Swapping it around so the "what" goes first and the "when" comes after makes life a little bit easier, for example:

```
INACTIVITY_PENALTY_QUOTIENT: 67108864
INACTIVITY_PENALTY_QUOTIENT_ALTAIR: 50331648
```

It's not a major difference, but I think the benefit of having it this way around will increase with the number of hard forks.

hwwhww:
I'm fine either way. /cc @djrtwo

djrtwo:
ah, yeah. This makes sense

djrtwo:
addressed!

PR #2241, opened 2021-03-15
suffix ALTAIR on update constants

djrtwo:
addresses #2231 

I'm not entirely satisfied with this because there is a precedent for putting a fork(ish) signifier at the prefix of a constant name -- `GENESIS_FORK_VERSION` and `GENESIS_SLOT`.

We have something of an analogue in Altair with `ALTAIR_FORK_VERSION`. I we move to suffix this, it slightly breaks the original pattern, and if we keep it as the prefix but modify the other penalty constants, we  have what could be considered two different naming schemes.


hwwhww (approved):
LGTM

hwwhww:
More comments:
I think it's okay to use the `NAME_{spec_version}` scheme for the constants that are *replacing* its previous fork. `GENESIS_FORK_VERSION`/`ALTAIR_FORK_VERSION` and `GENESIS_SLOT`/`ALTAIR_FORK_SLOT` are not in this category.

djrtwo:
I'm okay with that

PR #2242, opened 2021-03-15
Define SyncAggregate to bundle sync committee bits and signature

protolambda:
### Spec changes

- Define `SyncAggregate` to bundle sync committee bits and signature

No real functional changes, just bundling two related fields to fit the familiar operation-object pattern.

### Testing changes

- Update process_sync_committee tests to better isolate the state-change
- Introduce helper function `run_block_processing_to`, also for future tests
- Update test format doc to include `SyncAggregate`

A known existing issue with block-processing tests is that the beacon-state is too arbitrary: it is not reached with a regular block transition, but artificially created to test a state-change is isolation.

To avoid this in the future, I propose we use the same mechanic as we already use for epoch-processing tests: run the assumed sub-transitions first, then output the pre-state, run the sub-transition of interest, and output the post-state.

The sync-committee block-processing tests did something similar: it processed a whole block to test the sub-transition. This PR leaves the same transition in place, but isolates the sync-committee processing part. Clients will only have to run the sub-transition  on the provided pre-state (like phase0 operation tests), avoiding extra processing, and it cleans up the state-diff to more easily single out sync-committee bugs.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/helpers/block_processing.py
+++ b/tests/core/pyspec/eth2spec/test/helpers/block_processing.py
@@ -1,0 +1,60 @@
+def for_ops(state, operations, fn) -> None:
+    for operation in operations:
+        fn(state, operation)
+
+
+def get_process_calls(spec):
+    return {
+        # PHASE0
+        'process_block_header':
+            lambda state, block: spec.process_block_header(state, block),
+        'process_randao':
+            lambda state, block: spec.process_randao(state, block.body),
+        'process_eth1_data':
+            lambda state, block: spec.process_eth1_data(state, block.body),
+        'process_proposer_slashing':
+            lambda state, block: for_ops(state, block.body.proposer_slashings, spec.process_proposer_slashing),
+        'process_attester_slashing':
+            lambda state, block: for_ops(state, block.body.attester_slashings, spec.process_attester_slashing),
+        'process_attestation':
+            lambda state, block: for_ops(state, block.body.attestations, spec.process_attestation),
+        'process_deposit':
+            lambda state, block: for_ops(state, block.body.deposits, spec.process_deposit),
+        'process_voluntary_exit':
+            lambda state, block: for_ops(state, block.body.voluntary_exits, spec.process_voluntary_exit),
+        # Altair
+        'process_sync_committee':
+            lambda state, block: spec.process_sync_committee(state, block.body.sync_aggregate),
+        # PHASE1
+        'process_custody_game_operations':
+            lambda state, block: spec.process_custody_game_operations(state, block.body),
+        'process_shard_transitions':
+            lambda state, block: spec.process_shard_transitions(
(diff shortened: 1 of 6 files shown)


protolambda:
Replaced commit to fix TOC

tests/core/pyspec/eth2spec/test/helpers/block_processing.py
```diff
@@ -0,0 +1,60 @@
+def for_ops(state, operations, fn) -> None:
+    for operation in operations:
+        fn(state, operation)
+
+
+def get_process_calls(spec):
```
djrtwo, line 6: Funny.. this might be useful across many contexts of other tests. Obviously important for the `latest_block_header`  to be correct for sync committee tests, but maybe would have more "correct" states in other context if we use it across the board

djrtwo (approved):
great!

tests/core/pyspec/eth2spec/test/helpers/block_processing.py
```diff
@@ -0,0 +1,60 @@
...
+    Processes to the block transition, up to, but not including, the sub-transition named ``process_name``.
+    Returns a Callable[[state, block], None] for the remaining ``process_name`` transition.
+
+    Tests should create full blocks to ensure a valid state transition, even if the operation itself is isolated.
+    (e.g. latest_header in the beacon state is up-to-date in a sync-committee test).
+
+    A test prepares a pre-state by calling this function, output the pre-state,
+     and it can then proceed to run the returned callable, and output a post-state.
+    """
+    print(f"state.slot {state.slot} block.slot {block.slot}")
```
hwwhww, line 48: TODO: Remove the debugging message

tests/core/pyspec/eth2spec/test/helpers/block_processing.py
```diff
@@ -0,0 +1,60 @@
...
+    # transition state to slot before block state transition
+    if state.slot < block.slot:
+        spec.process_slots(state, block.slot)
+    print(f"state.slot {state.slot} block.slot {block.slot}  A")
```
hwwhww, line 52: TODO: Remove the debugging message

tests/core/pyspec/eth2spec/test/helpers/block_processing.py
```diff
@@ -0,0 +1,60 @@
...
+        # PHASE1
+        'process_custody_game_operations':
+            lambda state, block: spec.process_custody_game_operations(state, block.body),
+        'process_shard_transitions':
+            lambda state, block: spec.process_shard_transitions(
+                state, block.body.shard_transitions, block.body.attestations),
+    }
+
+
+def run_block_processing_to(spec, state, block, process_name: str):
```
hwwhww, line 37: Good call 👍

hwwhww (commented):
post-merge comments

PR #2243, opened 2021-03-15
fix altair test generators

djrtwo:
* turn off large validator set tests for mainnet
* add check for if leak will actually affect a leaking validator

### DIFF
--- a/tests/core/pyspec/eth2spec/test/helpers/rewards.py
+++ b/tests/core/pyspec/eth2spec/test/helpers/rewards.py
@@ -27,6 +27,26 @@ def has_enough_for_reward(spec, state, index):
     )
 
 
+def has_enough_for_leak_penalty(spec, state, index):
+    """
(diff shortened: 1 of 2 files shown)


tests/core/pyspec/eth2spec/test/helpers/rewards.py
```diff
@@ -27,6 +27,26 @@ def has_enough_for_reward(spec, state, index):
     )
 
 
+def has_enough_for_leak_penalty(spec, state, index):
+    """
+    Check if effective_balance and state of leak is high enough for a leak penalty.
+
+    At very low balances / leak values, it is possible for a validator have a positive effective_balance
+    and be in a leak, but have zero leak penalty.
```
hwwhww, line 35: Oooh I stuck at random + leaking + mainnet tests too! Nice catch!
Just re-generated those tests again with your patch, works well now.

hwwhww (approved):
LGTM

PR #2244, opened 2021-03-16
Altair P2P

protolambda:
This PR is directed at a specs-repo copy of the branch by @ralexstokes, PR #2201, which implements the additive sync-committee related changes. When that PR is finished and squashed I will rebase this one to `dev`.

Changes:
- Update Altair P2P header to not skip over req-resp changes
- Define new additional `<context-bytes>` Req-Resp chunk field, compatible with phase0.
- Define `ForkDigest` context, so we can parametrize the payload to decode. More specifically, to pick the SSZ type corresponding to the right fork, per chunk, enabling cross-fork sync.
- Add blocks-by-range v2, and blocks-by-root v2, completely unchanged w.r.t. payload handling, but using the new `ForkDigest` context to select the SSZ type for the decoder.
- Describe Req-Resp transition path
- Describe GossipSub transition path

This is not final yet, but I think it looks like a good minimal change-set to support the Altair fork. Suggestions welcome. We can test and iterate in later release iterations also.

CC network crew @Nashatyrev @nisdas @AgeManning @arnetheduck

### DIFF
--- a/specs/altair/p2p-interface.md
+++ b/specs/altair/p2p-interface.md
@@ -1,11 +1,11 @@
 # Ethereum Altair networking specification
 
-This document contains the networking specification for Ethereum 2.0 clients added during the Altair deployment. 
-This document should be viewed as additive to the [document from Phase 0](../phase0/p2p-interface.md) and will be referred to as the "phase 0 document" hereafter. 
-Readers should understand the phase 0 document and use it as a basis to understand the changes outlined in this document.
+This document contains the networking specification for Ethereum 2.0 clients added during the Altair deployment.
+This document should be viewed as additive to the [document from Phase 0](../phase0/p2p-interface.md) and will be referred to as the "Phase 0 document" hereafter.
+Readers should understand the Phase 0 document and use it as a basis to understand the changes outlined in this document.
 
-In particular, Altair introduces changes to the gossip domain and a small change to ENR advertisement in the discovery domain. 
-The phase 0 document should be consulted for specs on the network fundamentals, the existing gossip domain, the req/resp domain and the discovery domain.
+Altair adds new messages, topics and data to the Req-Resp, Gossip and Discovery domain. Some Phase 0 features will be deprecated, but not removed immediately.
+
 
 ## Table of contents
 
@@ -23,7 +23,14 @@ The phase 0 document should be consulted for specs on the network fundamentals, 
       - [Sync committee subnets](#sync-committee-subnets)
         - [`sync_committee_{subnet_id}`](#sync_committee_subnet_id)
       - [Sync committees and aggregation](#sync-committees-and-aggregation)
+    - [Transitioning the gossip](#transitioning-the-gossip)
   - [The Req/Resp domain](#the-reqresp-domain)
+    - [Req-Resp interaction](#req-resp-interaction)
+      - [`ForkDigest`-context](#forkdigest-context)
+    - [Messages](#messages)
+      - [BeaconBlocksByRange v2](#beaconblocksbyrange-v2)
+      - [BeaconBlocksByRoot v2](#beaconblocksbyroot-v2)
+    - [Transitioning from v1 to v2](#transitioning-from-v1-to-v2)
   - [The discovery domain: discv5](#the-discovery-domain-discv5)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
@@ -43,7 +50,7 @@ Validators use an aggregation scheme to balance the processing and networking lo
 
 ### Topics and messages
 
-Topics follow the same specification as in the phase 0 document. 
+Topics follow the same specification as in the Phase 0 document. 
 New topics are added in Altair to support the sync committees and the beacon block topic is updated with the modified type.
 
 The specification around the creation, validation, and dissemination of messages has not changed from the Phase 0 document.
@@ -58,14 +65,18 @@ The new topics along with the type of the `data` field of a gossipsub message ar
 
 Definitions of these new types can be found in the [Altair validator guide](./validator.md#containers).
 
+Note that the `ForkDigestValue` path segment of the topic separates the old and the new `beacon_block` topics.
+
 #### Global topics
 
 Altair changes the type of the global beacon block topic and adds one global topic to propagate partially aggregated sync committee signatures to all potential proposers of beacon blocks.
 
 ##### `beacon_block`
 
-The existing specification for this topic does not change from the Phase 0 document, but the type of the payload does change to the (modified) `SignedBeaconBlock`. 
-This type changes due to the inclusion of the inner `BeaconBlockBody` that is modified in Altair. 
+The existing specification for this topic does not change from the Phase 0 document,
+but the type of the payload does change to the (modified) `SignedBeaconBlock`. 
+This type changes due to the inclusion of the inner `BeaconBlockBody` that is modified in Altair.
+
 See the [state transition document](./beacon-chain.md#beaconblockbody) for Altair for further details.
 
 ##### `sync_committee_contribution_and_proof`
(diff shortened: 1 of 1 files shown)


specs/altair/p2p-interface.md
```diff
@@ -94,6 +112,92 @@ Unaggregated signatures (along with metadata) are sent as `SyncCommitteeSignatur
...
+
+Starting with Altair, and in future forks, SSZ type definitions may change.
+For this common case, we define the `ForkDigest`-context:
+
+A fixed-width 4 byte `<context-bytes>`, set to the `ForkDigest` matching the chunk:
+ `compute_fork_digest(fork_version, genesis_validators_root)`.
+
+### Messages
+
+#### BeaconBlocksByRange v2
```
nisdas, line 164: How does adding the `context` metadata work here ? They are only included in response packets rather than requests. Is the goal for consumers to be aware of the particular fork version of the response payload ?

protolambda: The version bump is to distinguish between an empty `context`, and a `ForkDigest` used in v2. The format is "compatible" in the sense that we can define the format of the old v1 methods as a subset of the new format, by simply assuming an empty context. But the actual change of a method is breaking, hence a v2.

The idea is that everyone can update their Req-Resp stack with this small change, and keep supporting the old formats almost effortlessly. Most methods don't need a v2 with Altair. Just in the case of these two beacon block methods, the fork digest can now be read before touching the SSZ decoder, which should make it easy to handle different block types.

Regarding implementation, the verbose way of describing it would be:
1. Add a `CodecContext` interface/base-type/whatever preferred.
2. Parametrize the response writer with `CodecContext`
3. Specify `EmptyContext` as a subtype/implementation/whatever of `CodecContext`, with a read/write method, that is a no-op.
4. Specify that all v1 request handlers have an `EmptyContext` as context type
5. Specify that all v1 response handlers have an `EmptyContext` as context type
6. Specify `ForkDigestContext` as subtype/implementation/whatever of `CodecContext`, with a read/write method, that reads/writes 4 bytes into/from a fork digest.
6. Define the new v2 block request handlers in terms of the old v1 handlers, but parametrized with `ForkDigestContext`. The chunk writer can switch on the type of context, and when it's `ForkDigestContext`, it writes the fork digest corresponding to the block.
7. Define the new v2 block response handlers in terms of the old v1 handlers, but parametrized with `ForkDigestContext`. The chunk reader first reads the expected 4 byte fork digest, validates if it's a known fork, and then continues to decode the chunk into the expected type of block.

We can also rename the req-resp `CodecContext` to `CodecParameter` or something if it's confusing in Go.

specs/altair/p2p-interface.md
```diff
@@ -1,8 +1,12 @@
...
+This document contains the networking specification for Ethereum 2.0 clients added during the Altair deployment.
+This document should be viewed as additive to the [document from Phase 0](../phase0/p2p-interface.md) and will be referred to as the "phase 0 document" hereafter.
+Readers should understand the phase 0 document and use it as a basis to understand the changes outlined in this document.
 
-In particular, Altair introduces changes to the gossip domain and a small change to ENR advertisement in the discovery domain. The phase 0 document should be consulted for specs on the network fundamentals, the existing gossip domain, the req/resp domain and the discovery domain.
+Altair adds new messages, topics and data to the Req-Resp, Gossip and Discovery domain. Some phase0 features will be deprecated, but not removed immediately.
```
ralexstokes, line 7:
```suggestion
Altair adds new messages, topics and data to the Req-Resp, Gossip and Discovery domain. Some phase 0 features will be deprecated, but not removed immediately.
```

but could go either way as long as we are consistent

specs/altair/p2p-interface.md
```diff
@@ -46,13 +58,19 @@ The new topics along with the type of the `data` field of a gossipsub message ar
...
+The `ForkDigestValue` path segment of the topic separates the old and the new `beacon_block` topics.
+
 #### Global topics
 
 Altair changes the type of the global beacon block topic and adds one global topic to propagate partially aggregated sync committee signatures to all potential proposers of beacon blocks.
 
 ##### `beacon_block`
 
-The existing specification for this topic do not change from the Phase 0 document but the type of the payload does change to the (modified) `SignedBeaconBlock`. This type changes via the inclusion of the inner `BeaconBlockBody` that is modified in Altair. See the [state transition document](./beacon-chain.md#beaconblockbody) for Altair for further details.
+The existing specification for this topic do not change from the Phase 0 document,
```
ralexstokes, line 69:
```suggestion
The existing specifications for this topic do not change from the Phase 0 document, 
```

specs/altair/p2p-interface.md
```diff
@@ -94,6 +112,92 @@ Unaggregated signatures (along with metadata) are sent as `SyncCommitteeSignatur
 
 Aggregated sync committee signatures are packaged into (signed) `SyncCommitteeContribution` along with proofs and gossiped to the `sync_committee_contribution_and_proof` topic.
 
+### Transitioning the gossip
+
+With any fork, the fork version, and thus the `ForkDigestValue`, change.
+Message types are unique per topic, and so for a smooth transition a node will have to subscribe to both the old and new topics.
+
+All remaining topics are updated. In advance of the fork, a node SHOULD subscribe to the post-fork variants of the topics.
```
ralexstokes, line 120:
```suggestion
All remaining topics are updated with the new `ForkDigestValue`. In advance of the fork, a node SHOULD subscribe to the post-fork variants of the topics.
```

this is a bit more clear imo

specs/altair/p2p-interface.md
```diff
@@ -94,6 +112,92 @@ Unaggregated signatures (along with metadata) are sent as `SyncCommitteeSignatur
...
+
+Subscriptions are expected to be well-received, all updated nodes should subscribe as well.
+Topic-meshes can be grafted quickly as the nodes are already connected and exchanging gossip control messages.
+
+Messages SHOULD NOT be re-broadcast from one fork to the other.
+
```
ralexstokes, line 126:
```suggestion
A node's behavior before the fork and after the fork are as follows:
```

I think adding a bit of prose here to guide the reader will help. I was a bit confused when we just dropped into pre and post and just an additional sentence that this is the expected behavior in each era is helpful

specs/altair/p2p-interface.md
```diff
@@ -94,6 +112,92 @@ Unaggregated signatures (along with metadata) are sent as `SyncCommitteeSignatur
...
+Pre-fork:
+- Peers who propagate messages on the post-fork topics MAY be scored negatively proportionally to time till fork,
+  to account for clock discrepancy.
+- Messages can be IGNORED on the post-fork topics, with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` margin.
+
+Post-fork:
+- Peers who propagate messages on the pre-fork topics MUST NOT be scored negatively. Lagging IWANT may force them to.
+- Messages on pre and post-fork variants of topics share application-level caches.
```
ralexstokes, line 134: good call

specs/altair/p2p-interface.md
```diff
@@ -94,6 +112,92 @@ Unaggregated signatures (along with metadata) are sent as `SyncCommitteeSignatur
...
+  E.g. an attestation on the both the old and new topic is ignored like any duplicate.
+- Two epochs after the fork, pre-fork topics SHOULD be unsubscribed from. This is well after the configured `seen_ttl`.
```
ralexstokes, line 136: just noting that we are tearing down the topics so no worries if suddenly a large portion of the network drops a topic all at once (vs. subnet stability elsewhere where it would be an issue); `seen_ttl` is 1 epoch so 2 epochs should be plenty of buffer, even accounting for clock disparity

specs/altair/p2p-interface.md
```diff
@@ -94,6 +112,92 @@ Unaggregated signatures (along with metadata) are sent as `SyncCommitteeSignatur
...
+
+## The Req/Resp domain
+
+### Req-Resp interaction
+
+An additional `<context-bytes>` field is introduced to the `response_chunk` as defined in the phase0 document:
```
ralexstokes, line 142:
```suggestion
An additional `<context-bytes>` field is introduced to the `response_chunk` as defined in the phase 0 document:
```

either way, lets just be consistent

specs/altair/p2p-interface.md
```diff
@@ -94,6 +112,92 @@ Unaggregated signatures (along with metadata) are sent as `SyncCommitteeSignatur
...
+
+```
+response_chunk  ::= <result> | <context-bytes> | <encoding-dependent-header> | <encoded-payload>
+```
+
+All phase0 methods are compatible: `<context-bytes>` is empty by default.
```
ralexstokes, line 148:
```suggestion
All phase 0 methods are compatible: `<context-bytes>` is empty by default.
```

ditto

ralexstokes (commented):
awesome work! left some minor formatting/editing comments but overall looks good.

we can continue thinking through any edge cases around fork boundaries but it seems like you hit the major stuff around handling deprecations/evolutions to v2 and topic mesh upgrades.

specs/altair/p2p-interface.md
protolambda: I think the "remaining" part is confusing. It's more like "those topics names that stay in use after the fork update". I'll try to fix it

protolambda, line 120:
```suggestion
The topics that are not removed in a fork are updated with a new `ForkDigestValue`. In advance of the fork, a node SHOULD subscribe to the post-fork variants of the topics.
```

specs/altair/p2p-interface.md
```diff
@@ -94,6 +112,92 @@ Unaggregated signatures (along with metadata) are sent as `SyncCommitteeSignatur
 
 Aggregated sync committee signatures are packaged into (signed) `SyncCommitteeContribution` along with proofs and gossiped to the `sync_committee_contribution_and_proof` topic.
 
+### Transitioning the gossip
+
+With any fork, the fork version, and thus the `ForkDigestValue`, change.
+Message types are unique per topic, and so for a smooth transition a node will have to subscribe to both the old and new topics.
```
djrtwo, line 118:
```suggestion
Message types are unique per topic, and so for a smooth transition a node must temporarily subscribe to both the old and new topics.
```

djrtwo (approved):
looks good

AgeManning:
A small point for client implementer, would be to disable the `mesh_message_deliveries` score when subscribing to the new topics pre-fork to avoid penalizing peers.

PR #2245, opened 2021-03-16
Enhance gen_runner error handler

hwwhww:
### Issue

It was hard to check if all test vectors have been generated successfully.

### How did I improve it
1. Output error log to `<output_dir>/testgen_error_log.txt`.
2. Add `release_version: str` to `meta.yaml`. Two purposes:
	1. To indicate which spec release version it is.
	2. This field is added only if the test vectors have been generated successfully. So when we run testgen *again*, even if the folder exists, we should *not* skip it if `release_version` is *not* in `meta.yaml`.

---

* Disable mainnet config `test_valid_signature_future_committee`
	* It was really slow for me to transition `2 * spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD` epochs. @djrtwo does it work well on your end?

### DIFF
--- a/tests/core/pyspec/eth2spec/gen_helpers/gen_base/gen_runner.py
+++ b/tests/core/pyspec/eth2spec/gen_helpers/gen_base/gen_runner.py
@@ -132,7 +132,7 @@ def run_generator(generator_name, test_providers: Iterable[TestProvider]):
                 / Path(test_case.runner_name) / Path(test_case.handler_name)
                 / Path(test_case.suite_name) / Path(test_case.case_name)
             )
-            if case_dir.exists():
+            if case_dir.exists() and not args.force:
                 # Check if the test vector was generated successfully.
                 meta_file = case_dir / 'meta.yaml'
                 # Check if meta file exists
@@ -141,7 +141,7 @@ def run_generator(generator_name, test_providers: Iterable[TestProvider]):
                     with meta_file.open('r') as f:
                         meta_dict = safe_load(f.read())
 
-                if 'release_version' in meta_dict and not args.force:
+                if 'release_version' in meta_dict and meta_dict['release_version'] == str(version('eth2spec')):
                     # If release_version is in meta.yaml, the test was generated successfully.
                     print(f'Skipping already existing test: {case_dir}')
                     continue
@@ -149,7 +149,7 @@ def run_generator(generator_name, test_providers: Iterable[TestProvider]):
                 print(f'Warning, output directory {case_dir} already exist,'
                       f' old files are not deleted but will be overwritten when a new version is produced')
 
-                # Remove case_dir folder
+                # Remove the existing case_dir folder
                 shutil.rmtree(case_dir)
 
             print(f'Generating test: {case_dir}')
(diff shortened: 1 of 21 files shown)


tests/core/pyspec/eth2spec/gen_helpers/gen_base/gen_runner.py
```diff
@@ -127,12 +133,25 @@ def run_generator(generator_name, test_providers: Iterable[TestProvider]):
                 / Path(test_case.suite_name) / Path(test_case.case_name)
             )
             if case_dir.exists():
-                if not args.force:
+                # Check if the test vector was generated successfully.
+                meta_file = case_dir / 'meta.yaml'
```
protolambda, line 137: maybe don't read the meta file at all if the `args.force` is true?

tests/core/pyspec/eth2spec/gen_helpers/gen_base/gen_runner.py
```diff
@@ -127,12 +133,25 @@ def run_generator(generator_name, test_providers: Iterable[TestProvider]):
...
+                # Check if meta file exists
+                meta_dict = {}
+                if meta_file.exists():
+                    with meta_file.open('r') as f:
+                        meta_dict = safe_load(f.read())
+
+                if 'release_version' in meta_dict and not args.force:
```
protolambda, line 144: wouldn't it make more sense to check the version value, instead of the presence of the key?

tests/core/pyspec/eth2spec/gen_helpers/gen_base/gen_runner.py
```diff
@@ -159,17 +178,26 @@ def output_part(out_kind: str, name: str, fn: Callable[[Path, ], None]):
...
                     written_part = True
                     output_part("data", "meta", dump_yaml_fn(meta, "meta", file_mode, yaml))
 
                 if not written_part:
                     print(f"test case {case_dir} did not produce any test case parts")
-
             except Exception as e:
                 print(f"ERROR: failed to generate vector(s) for test {case_dir}: {e}")
                 traceback.print_exc()
+                # Write to log file
```
protolambda, line 195: Nice, no more lost logs

tests/core/pyspec/eth2spec/test/altair/block_processing/test_process_sync_committee.py
```diff
@@ -284,6 +284,7 @@ def test_invalid_signature_previous_committee(spec, state):
 @with_all_phases_except([PHASE0, PHASE1])
 @spec_state_test
 @always_bls
+@with_configs([MINIMAL], reason="too slow")
```
protolambda, line 287: 1 mainnet sync period is about a day right? 225 epochs per day, 256 per sync period. That will take a very long time to run in a python test indeed (even worse with BLS enabled in generators, if it involves any blocks)

protolambda (commented):
Generally looks good, added some comments. Not sure about the versioning, it makes the diff much more verbose (updating all meta files each version, even if test is unchanged)

djrtwo:
> @djrtwo does it work well on your end?

something took a long time! but my server made it through. I don't think we need to include such a heavy test

hwwhww:
@protolambda 

> Not sure about the versioning, it makes the diff much more verbose (updating all meta files each version, even if test is unchanged)

Indeed. #2247 is an alternative solution. 👀

hwwhww:
replaced by #2247

PR #2246, opened 2021-03-16
refactor sync committee rewards

djrtwo:
refactor rewards to ensure that sync committee is included in the "max 1 base_reward per validator per epoch" invariant

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -10,7 +10,7 @@
 - [Custom types](#custom-types)
 - [Constants](#constants)
   - [Participation flag indices](#participation-flag-indices)
-  - [Participation flag fractions](#participation-flag-fractions)
+  - [Incentivization weights](#incentivization-weights)
   - [Misc](#misc)
 - [Configuration](#configuration)
   - [Updated penalty values](#updated-penalty-values)
@@ -28,12 +28,13 @@
   - [`Predicates`](#predicates)
     - [`eth2_fast_aggregate_verify`](#eth2_fast_aggregate_verify)
   - [Misc](#misc-2)
-    - [`get_flag_indices_and_numerators`](#get_flag_indices_and_numerators)
+    - [`get_flag_indices_and_weights`](#get_flag_indices_and_weights)
     - [`add_flag`](#add_flag)
     - [`has_flag`](#has_flag)
   - [Beacon state accessors](#beacon-state-accessors)
     - [`get_sync_committee_indices`](#get_sync_committee_indices)
     - [`get_sync_committee`](#get_sync_committee)
+    - [`get_base_reward_per_increment`](#get_base_reward_per_increment)
     - [`get_base_reward`](#get_base_reward)
     - [`get_unslashed_participating_indices`](#get_unslashed_participating_indices)
     - [`get_flag_index_deltas`](#get_flag_index_deltas)
@@ -61,7 +62,7 @@ Altair is the first beacon chain hard fork. Its main features are:
 
 * sync committees to support light clients
 * incentive accounting reforms to reduce spec complexity
-* penalty parameter updates to move them to their planned maximally punitive configuration
+* penalty parameter updates towards their planned maximally punitive values
 
 ## Custom types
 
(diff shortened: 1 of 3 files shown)


specs/altair/beacon-chain.md
```diff
@@ -295,15 +297,21 @@ def get_sync_committee(state: BeaconState, epoch: Epoch) -> SyncCommittee:
     return SyncCommittee(pubkeys=pubkeys, pubkey_aggregates=pubkey_aggregates)
 ```
 
+#### `get_base_reward_per_increment`
+
+```python
+def get_base_reward_per_increment(state: BeaconState) -> Gwei:
+    return Gwei(EFFECTIVE_BALANCE_INCREMENT * BASE_REWARD_FACTOR // integer_squareroot(get_total_active_balance(state)))
```
protolambda, line 304: `EFFECTIVE_BALANCE_INCREMENT * BASE_REWARD_FACTOR = 2**0 * 10**9 * 64 = 64 * 10**9`, so 35.89... bits if dividing by 1. (super low total balance case). 64-36=28 of room. Simple multiplication with increment count will work, something like squaring won't.

specs/altair/beacon-chain.md
```diff
@@ -523,26 +529,28 @@ def process_sync_committee(state: BeaconState, aggregate: SyncAggregate) -> None
...
-        reward = Gwei(max_participant_reward * active_validator_count // (len(committee_indices) * SLOTS_PER_EPOCH))
-        increase_balance(state, participant_index, reward)
-
-    # Reward beacon proposer
-    increase_balance(state, get_beacon_proposer_index(state), proposer_rewards)
+    assert eth2_fast_aggregate_verify(included_pubkeys, signing_root, aggregate.sync_committee_signature)
+
+    # Compute the maximum sync rewards for the slot
+    total_active_increments = get_total_active_balance(state) // EFFECTIVE_BALANCE_INCREMENT
+    total_base_rewards = Gwei(get_base_reward_per_increment(state) * total_active_increments)
```
protolambda, line 541: `total_active_increments = 4M validators max, 32 increments each = 32*4000000 = 128000000`, `log2(128000000) = 26.93... bits`. There were 28 bits of room. Ok...

specs/altair/beacon-chain.md
```diff
@@ -523,26 +529,28 @@ def process_sync_committee(state: BeaconState, aggregate: SyncAggregate) -> None
...
+    max_epoch_rewards = Gwei(total_base_rewards * SYNC_REWARD_WEIGHT // WEIGHT_DENOMINATOR)
```
protolambda, line 542: Now multiplied with 8 (3 bits). `36+28+3=67` Overflow in extreme case? Maybe worth doing the math more precisely, but it's definitely on the edge.

protolambda (changes requested):
Think there might be a new overflow, need someone else to review. (And it's late here too, don't want to put a last-minute review on rewards code like this).

specs/altair/beacon-chain.md
protolambda: Quite an extreme case. Is there some minimum total balance bit count we can assume here? Worth documenting the overflow. A year ago we Herman found a similar thing in the original code.

protolambda: Oh right, I remember now, the total active balance has a lower bound of 1 effective balance increment. `log2(1 * 10**9) = 29.8...`. That should leave it at ~6 bits, low.

specs/altair/beacon-chain.md
```diff
@@ -523,26 +529,28 @@ def process_sync_committee(state: BeaconState, aggregate: SyncAggregate) -> None
...
+    max_slot_rewards = Gwei(max_epoch_rewards * len(included_indices) // len(committee_indices) // SLOTS_PER_EPOCH)
```
protolambda, line 543: `7+28+log2(SYNC_COMMITTEE_SIZE)=7+28+10=45`

protolambda (approved):
False alarm with the overflow. Minimum total balance special-case relieves any concern. Still could use more review.

PR #2247, opened 2021-03-16
Enhance gen_runner error handler + check with `INCOMPLETE` temporary file

hwwhww:
An alternative to #2245. As @protolambda said, adding `release_version` field to each test case would make diff much more verbose. This PR tried to reduce the changes.

### Same to #2245
1. Output error log to `<output_dir>/testgen_error_log.txt`.
2 Disable mainnet config `test_valid_signature_future_committee`

### Differences
Instead of adding `release_version: str` to `meta.yaml`, this PR checks if the test vectors were generated completely with a temporary `INCOMPLETE` file.

For each test case:
1. If `<case_dir>` folder exists
    - If `<case_dir>/INCOMPLETE` file does *NOT* exist, skip it.
    - Otherwise, go step 2 to start the generation.
2. Create a `<case_dir>/INCOMPLETE` file.
3. Run test case, output the test vectors.
4. If no exception, remove `<case_dir>/INCOMPLETE` file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ COV_INDEX_FILE=$(PY_SPEC_DIR)/$(COV_HTML_OUT)/index.html
 
 CURRENT_DIR = ${CURDIR}
 LINTER_CONFIG_FILE = $(CURRENT_DIR)/linter.ini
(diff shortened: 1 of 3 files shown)


protolambda (approved):
This should solve the diff issue, nice :+1:

djrtwo (approved):

PR #2248, opened 2021-03-17
Fix list indentation

hukkinj1:
Fix Markdown list indentation so that the lists are rendered as expected.

BTW, these errors were made obvious while I was testing a [Markdown formatter](https://github.com/executablebooks/mdformat) I've been working on against this repository. The formatter could be a good match for this repository, being able to format Markdown tables, create/remove word wrap if desired, generate table of contents, format embedded code blocks (Python with Black) etc. I could set up a PR to showcase if this is something of interest to you?

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -1298,10 +1298,10 @@ Requests are segregated by protocol ID to:
 6. Parallelise RFCs (or Eth2 EIPs).
   By decoupling requests from one another, each RFC that affects the request protocol can be deployed/tested/debated independently
   without relying on a synchronization point to version the general top-level protocol.
-  1. This has the benefit that clients can explicitly choose which RFCs to deploy
-    without buying into all other RFCs that may be included in that top-level version.
-  2. Affording this level of granularity with a top-level protocol would imply creating as many variants
-    (e.g. /protocol/43-{a,b,c,d,...}) as the cartesian product of RFCs inflight, O(n^2).
+   1. This has the benefit that clients can explicitly choose which RFCs to deploy
+      without buying into all other RFCs that may be included in that top-level version.
+   2. Affording this level of granularity with a top-level protocol would imply creating as many variants
+      (e.g. /protocol/43-{a,b,c,d,...}) as the cartesian product of RFCs inflight, O(n^2).
 7. Allow us to simplify the payload of requests.
   Request-id’s and method-ids no longer need to be sent.
   The encoding/request type and version can all be handled by the framework.
--- a/specs/phase0/weak-subjectivity.md
+++ b/specs/phase0/weak-subjectivity.md
@@ -141,12 +141,12 @@ Clients should allow users to input a Weak Subjectivity Checkpoint at startup, a
 ### Weak Subjectivity Sync Procedure
 
 1. Input a Weak Subjectivity Checkpoint as a CLI parameter in `block_root:epoch_number` format,
-  where `block_root` (an "0x" prefixed 32-byte hex string) and `epoch_number` (an integer) represent a valid `Checkpoint`.
-  Example of the format:
+   where `block_root` (an "0x" prefixed 32-byte hex string) and `epoch_number` (an integer) represent a valid `Checkpoint`.
+   Example of the format:
 
-```
-0x8584188b86a9296932785cc2827b925f9deebacce6d72ad8d53171fa046b43d9:9544
-```
+   ```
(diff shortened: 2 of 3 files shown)


specs/phase0/p2p-interface.md
```diff
@@ -1298,10 +1298,10 @@ Requests are segregated by protocol ID to:
...
   By decoupling requests from one another, each RFC that affects the request protocol can be deployed/tested/debated independently
   without relying on a synchronization point to version the general top-level protocol.
-  1. This has the benefit that clients can explicitly choose which RFCs to deploy
-    without buying into all other RFCs that may be included in that top-level version.
-  2. Affording this level of granularity with a top-level protocol would imply creating as many variants
-    (e.g. /protocol/43-{a,b,c,d,...}) as the cartesian product of RFCs inflight, O(n^2).
+   1. This has the benefit that clients can explicitly choose which RFCs to deploy
+      without buying into all other RFCs that may be included in that top-level version.
+   2. Affording this level of granularity with a top-level protocol would imply creating as many variants
+      (e.g. /protocol/43-{a,b,c,d,...}) as the cartesian product of RFCs inflight, O(n^2).
```
ralexstokes, line 1304: i think the formatting change is fine although i will note that github renders the `1` and `2` sub-points as part of the outer numbered sequence so it goes `6,7,8,9` -- number `7` in the text is rendered as number `9`

likely worth fixing to get meaning across

specs/phase0/weak-subjectivity.md
```diff
@@ -141,12 +141,12 @@ Clients should allow users to input a Weak Subjectivity Checkpoint at startup, a
...
-  Example of the format:
+   where `block_root` (an "0x" prefixed 32-byte hex string) and `epoch_number` (an integer) represent a valid `Checkpoint`.
+   Example of the format:
 
-```
-0x8584188b86a9296932785cc2827b925f9deebacce6d72ad8d53171fa046b43d9:9544
-```
+   ```
+   0x8584188b86a9296932785cc2827b925f9deebacce6d72ad8d53171fa046b43d9:9544
+   ```
```
ralexstokes, line 150: render seems fine and i'm happy w/ the formatting change

ralexstokes (commented):
thanks for the PR! left some comments

specs/phase0/p2p-interface.md
hukkinj1: Hi @ralexstokes ! This is precisely what this change fixes. Currently the inner list items are actually part of the outer list in Markdown abstract syntax. With the added indentation they truly become a separate inner list and the numbering works as expected when rendered to HTML.

specs/phase0/weak-subjectivity.md
hukkinj1: Yup! To clarify, without this change, the list items 1. and 2. are in fact two distinct lists in Markdown abstract syntax. This change makes them list items of a shared list.

hukkinj1:
Added one more commit to fix another case where a list was in fact two distinct lists in Markdown syntax.

djrtwo (approved):
thank you!!

djrtwo:
@hukkinj1 would it make sense to integrate the formatter into CI? Is the tool generally ready for that?

hukkinj1:
@djrtwo yes sure, but it's important to know that the tool is __very__ opinionated, so it may be hard/frustrating for contributors to write Markdown that passes the check. That means that they should actually remember to run the formatter. I'd suggest a few different approaches:
- Don't integrate in CI, but occasionally run the tool manually (not my favorite approach, issues like the ones in this PR will go undetected)
- Provide an easy way for contributors to run the formatter as a git pre commit hook (I suggest the [pre-commit](https://pre-commit.com/) tool. Essentially we add one configuration file, and the contributor will once run `pre-commit install` to activate)
- Integrate in CI in a way where CI actually fixes the issues, and doesn't just complain. https://pre-commit.ci/ does exactly that.

hukkinj1:
Also if you want a PR to showcase, there is some (intentionally very little) configuration that should decided upon.

1. Should ordered lists be numbered consecutively, or all items with `1. `. The default non-numbering style is great for minimizing diffs (explained in the [docs](https://mdformat.readthedocs.io/en/stable/users/style.html#ordered-lists))
2. Should paragraphs be wrapped. The default is to not change word wrap (to support semantic line breaks, and again, to minimize diffs) (explained in the [docs](https://mdformat.readthedocs.io/en/stable/users/style.html#paragraph-word-wrapping))
3. Do we want to include plugins, enabling features such as
   - ToC autogeneration
   - Embedded code formatting (formatting Python with Black might make sense here)

PR #2251, opened 2021-03-17
Organize phase 0 validator guide type definitions

ralexstokes:
This PR hoists container definitions for some phase 0 validator guide types to the "preamble", in line with the style used across the specs repo (i.e. defined at the beginning and then used throughout the document).

This change is non-substantive and is purely formatting for phase 0.


djrtwo (approved):
yeah, makes sense! glad with did this with altair

PR #2252, opened 2021-03-17
fix v1.1.0-alpha.1 test generators

djrtwo:
* remove two extra `yield`s
* remove infinite loop in mainnet sync test

### DIFF
--- a/tests/core/pyspec/eth2spec/test/altair/block_processing/test_process_sync_committee.py
+++ b/tests/core/pyspec/eth2spec/test/altair/block_processing/test_process_sync_committee.py
@@ -68,8 +68,6 @@ def test_invalid_signature_missing_participant(spec, state):
     rng = random.Random(2020)
     random_participant = rng.choice(committee)
 
-    yield 'pre', state
-
     block = build_empty_block_for_next_slot(spec, state)
     # Exclude one participant whose signature was included.
     block.body.sync_aggregate = spec.SyncAggregate(
@@ -168,8 +166,6 @@ def validate_sync_committee_rewards(spec, pre_state, post_state, committee, comm
 
 
 def run_successful_sync_committee_test(spec, state, committee, committee_bits):
-    yield 'pre', state
-
     pre_state = state.copy()
 
     block = build_empty_block_for_next_slot(spec, state)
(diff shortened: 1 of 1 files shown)


tests/core/pyspec/eth2spec/test/altair/block_processing/test_process_sync_committee.py
```diff
@@ -238,6 +223,17 @@ def test_sync_committee_rewards_duplicate_committee(spec, state):
     yield from run_successful_sync_committee_test(spec, state, committee, committee_bits)
 
 
+@with_all_phases_except([PHASE0, PHASE1])
+@spec_state_test
+@always_bls
+def test_sync_committee_rewards_not_full_participants(spec, state):
+    committee = spec.get_sync_committee_indices(state, spec.get_current_epoch(state))
```
djrtwo, line 230: This is the line fix for the infinte loop

tests/core/pyspec/eth2spec/test/altair/block_processing/test_process_sync_committee.py
```diff
@@ -238,6 +223,17 @@ def test_sync_committee_rewards_duplicate_committee(spec, state):
...
+    rng = random.Random(1010)
+    committee_bits = [rng.choice([True, False]) for _ in committee]
+
+    yield from run_successful_sync_committee_test(spec, state, committee, committee_bits)
+
+
```
ralexstokes, line 236: yep, this should avoid hitting the infinite loop condition :)

are you worried about the (small) chance that `choice` returns `True` for all members? could toss an `assert` in there as a stop-gap...

tests/core/pyspec/eth2spec/test/altair/block_processing/test_process_sync_committee.py
```diff
@@ -68,8 +68,6 @@ def test_invalid_signature_missing_participant(spec, state):
     rng = random.Random(2020)
     random_participant = rng.choice(committee)
 
-    yield 'pre', state
-
```
ralexstokes, line 72: these seem fine but i do wonder if there is any concern about accidentally mutating the pre-state before getting to the inner yield in `run_sync_committee_processing`

ralexstokes (approved):
generally looks good -- left some questions just to double-check invariants/assumptions with you

tests/core/pyspec/eth2spec/test/altair/block_processing/test_process_sync_committee.py
djrtwo: nah, the rng is fixed so it's deterministic now

we *should* make sure we test it with empty participant list though

PR #2255, opened 2021-03-18
Use different name for constant for sync committee aggregates

ralexstokes:
The pre-release of Altair has two different uses of the word `subcommittee`. One usage refers to partitions of the sync committee to allow for optimizations for light clients following the chain. Another usage refers to partitions of the sync committee that are leveraged for aggregating sync committee signatures at the gossip layer like we employ with attestation gossip.

We cannot immediately use the same notion of subcommittee in both places as there are concerns with sync committee subnets being small at the specified size for light client purposes (e.g. on mainnet 64 validators for light clients vs 128 validators for aggregation). Given the fact that we know the attestation aggregation scheme works w/ subnets of size ~128 validators, it seems like we should not try to directly use the light client subcommittees for the gossip subcommittees.

The next option is to batch light client subcommittees into e.g. pairs for the gossip task, and while this is straightforward, an even simpler fix is to just rename the constant conferring "subcommittee" to the pubkey aggregates. This PR introduces this fix to jumpstart discussion around the preferred solution here.


JustinDrake:
What about `SYNC_PUBKEYS_PER_AGGREGATE` as opposed to `SYNC_PUBKEY_AGGREGATE_SIZE`? The reason is that the size of the aggregate is the same as the size of a single pubkey, namely 48 bytes.

specs/altair/beacon-chain.md
```diff
@@ -294,8 +294,8 @@ def get_sync_committee(state: BeaconState, epoch: Epoch) -> SyncCommittee:
     """
     indices = get_sync_committee_indices(state, epoch)
     pubkeys = [state.validators[index].pubkey for index in indices]
-    subcommitees = [pubkeys[i:i + SYNC_PUBKEY_AGGREGATE_SIZE] for i in range(0, len(pubkeys), SYNC_PUBKEY_AGGREGATE_SIZE)]
-    pubkey_aggregates = [bls.AggregatePKs(subcommitee) for subcommitee in subcommitees]
+    partition = [pubkeys[i:i + SYNC_PUBKEY_AGGREGATE_SIZE] for i in range(0, len(pubkeys), SYNC_PUBKEY_AGGREGATE_SIZE)]
+    pubkey_aggregates = [bls.AggregatePKs(preaggregate) for preaggregate in partition]
```
ralexstokes, line 298: ah missed this usage, thank you!

djrtwo (approved):

PR #2257, opened 2021-03-20
Pure consensus upgrade version of the merge

mkalinin:
### What's done

- executable beacon chain (https://github.com/ethereum/eth2.0-specs/pull/2229) is stripped down to the pure consensus upgrade
- the process of transition from PoW to PoS is based on total difficulty, as described by [Quick merge via fork choice change](https://notes.ethereum.org/m9IX3OkkTveXCCOSzGkUiw)

### Differences from quick merge proposal
- `ApplicationPayload` retained and used instead of `application_block` bytes

### Inspired by
- [Quick merge via fork choice change](https://notes.ethereum.org/m9IX3OkkTveXCCOSzGkUiw)
- @ralexstokes 

#### Application client requirements
Application client (Ethereum Mainnet client) will have to implement a new RPC protocol and the underlying logic to become merge compliant. The protocol has following methods:

- `eth2_insertBlock(parent_hash: Bytes32, application_payload: ApplicationPayload) -> boolean`
  Given the `application_payload` and `parent_hash` assembles an application block and inserts it into the application chain. If any of the pre or post verification conditions including `state_root` check are not satisfied then the method returns `false`, otherwise, returns `true`. 
  If parent block has been considered as the head of the chain then the newly inserted block becomes the head of chain if the import succeeds.

- `eth2_produceBlock(parent_hash: Bytes32) -> ApplicationPayload`
	Fetches transactions from the transaction pool and assembles a block on top of the parent block (specified by `parent_hash`) and returns it in a form of `ApplicationPayload`. Returns error code if parent block is not found.
	
    *Note:* Pending block may be used as a source of a new block if given `parent_hash` matches the one that is in the pending block.

- `eth2_setHead(hash: Bytes32) -> None`
	Given the hash of a block sets the head of the application chain. Returns error code if block is not found.

- `eth2_finalizeBlock(hash: Bytes32) -> boolean`
	Given the hash of a block finalises the application block. Returns error code if block is not found.

	*Note:* Chain and state clean ups may be triggered upon this call meaning that given block becomes an irreversible point on the chain. First finalisation event after the transition may have additional logic like disabling block gossip on the application layer.

- `eth2_submitTransitionBlockHash(hash: Bytes32) -> TransitionBlockStatus`
	Given the hash of a block returns the status information of the block. If block is not found then triggers the same processing chain inside of a client as if this hash was received with `NewBlockHashes` message. That is, asking the sync protocol to fetch the block and its ancestors and insert them into the application chain.

	*Note:* Required by transition process only.

```python
class TransitionBlockStatus:
    is_processed: boolean
    is_valid: boolean
    total_difficulty: uint256
```

#### Quick merge proposal

Quick merge proposal doesn't require extra `eth2_submitTransitionBlockHash` RPC method, this logic is implemented as a part of the `submitBlock` RPC method instead.

### DIFF
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -96,7 +96,7 @@ class Transaction(Container):
     gas_limit: uint64
     recipient: Bytes20
     value: uint256
-    data: List[byte, MAX_BYTES_PER_TRANSACTION_PAYLOAD]
+    data: ByteList[MAX_BYTES_PER_TRANSACTION_PAYLOAD]
     v: uint256
     r: uint256
     s: uint256
@@ -114,7 +114,7 @@ class ApplicationPayload(Container):
     gas_limit: uint64
     gas_used: uint64
     receipt_root: Bytes32
-    logs_bloom: Vector[byte, BYTES_PER_LOGS_BLOOM]
+    logs_bloom: ByteVector[BYTES_PER_LOGS_BLOOM]
     transactions: List[Transaction, MAX_APPLICATION_TRANSACTIONS]
 ```
 
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -77,40 +77,11 @@ class BeaconBlockBody(phase0.BeaconBlockBody):
 
 
 ```python
-class BeaconState(Container):
-    # Versioning
-    genesis_time: uint64
-    genesis_validators_root: Root
-    slot: Slot
-    fork: Fork
-    # History
-    latest_block_header: BeaconBlockHeader
-    block_roots: Vector[Root, SLOTS_PER_HISTORICAL_ROOT]
-    state_roots: Vector[Root, SLOTS_PER_HISTORICAL_ROOT]
-    historical_roots: List[Root, HISTORICAL_ROOTS_LIMIT]
-    # Eth1
-    eth1_data: Eth1Data
-    eth1_data_votes: List[Eth1Data, EPOCHS_PER_ETH1_VOTING_PERIOD * SLOTS_PER_EPOCH]
-    eth1_deposit_index: uint64
+class BeaconState(phase0.BeaconState):
     # [Added in Merge] hash of the root of application state
     application_state_root: Bytes32
     # [Added in Merge] hash of recent application block
     application_block_hash: Bytes32
-    # Registry
-    validators: List[Validator, VALIDATOR_REGISTRY_LIMIT]
-    balances: List[Gwei, VALIDATOR_REGISTRY_LIMIT]
-    # Randomness
-    randao_mixes: Vector[Bytes32, EPOCHS_PER_HISTORICAL_VECTOR]
-    # Slashings
-    slashings: Vector[Gwei, EPOCHS_PER_SLASHINGS_VECTOR]  # Per-epoch sums of slashed effective balances
-    # Attestations
-    previous_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
-    current_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
-    # Finality
-    justification_bits: Bitvector[JUSTIFICATION_BITS_LENGTH]  # Bit set for every recent justified epoch
-    previous_justified_checkpoint: Checkpoint  # Previous epoch snapshot
-    current_justified_checkpoint: Checkpoint
-    finalized_checkpoint: Checkpoint
 ```
 
 ### New containers
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -173,7 +173,10 @@ def compute_time_at_slot(state: BeaconState, slot: Slot) -> uint64:
 
 ```python
 def get_recent_beacon_block_roots(state: BeaconState, qty: uint64) -> Sequence[Bytes32]:
-    return [get_block_root_at_slot(state.slot - i) if GENESIS_SLOT + i < state.slot else Bytes32() for i in reversed(range(1, qty + 1))]
+    return [
+        get_block_root_at_slot(state.slot - i) if GENESIS_SLOT + i < state.slot else Bytes32()
+        for i in reversed(range(1, qty + 1))
+    ]
 ```
 
 #### `get_evm_beacon_block_roots`
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -215,7 +215,7 @@ class BeaconChainData(Container):
     slot: Slot
     randao_mix: Bytes32
     timestamp: uint64
-    recent_block_roots: Sequence[Bytes32]
+    recent_block_roots: Vector[Bytes32, EVM_BLOCK_ROOTS_SIZE]
 ```
 
 ##### `get_application_state`
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -1,0 +1,257 @@
+# Ethereum 2.0 The Merge
+
+**Notice**: This document is a work-in-progress for researchers and implementers.
+
+## Table of contents
+
+<!-- TOC -->
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Introduction](#introduction)
+- [Constants](#constants)
+  - [Execution](#execution)
+- [Containers](#containers)
+  - [Extended containers](#extended-containers)
+    - [`BeaconBlockBody`](#beaconblockbody)
+    - [`BeaconState`](#beaconstate)
+  - [New containers](#new-containers)
+    - [`Transaction`](#transaction)
+    - [`ApplicationPayload`](#applicationpayload)
+- [Helper functions](#helper-functions)
+  - [Misc](#misc)
+    - [`compute_randao_mix`](#compute_randao_mix)
+    - [`compute_time_at_slot`](#compute_time_at_slot)
+  - [Beacon state accessors](#beacon-state-accessors)
+    - [`get_recent_beacon_block_roots`](#get_recent_beacon_block_roots)
+    - [`get_evm_beacon_block_roots`](#get_evm_beacon_block_roots)
+  - [Block processing](#block-processing)
+    - [Modified `process_eth1_data`](#modified-process_eth1_data)
+    - [Application payload processing](#application-payload-processing)
+      - [`BeaconChainData`](#beaconchaindata)
+      - [`get_application_state`](#get_application_state)
+      - [`application_state_transition`](#application_state_transition)
+      - [`process_application_payload`](#process_application_payload)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+## Introduction
+
+This is a patch implementing executable beacon chain proposal. 
+It enshrines application execution and validity as a first class citizen at the core of the beacon chain.
+
+## Constants
+
+### Execution
+
+| Name | Value |
+| - | - |
+| `MAX_BYTES_PER_TRANSACTION_PAYLOAD` | `2**20` |
+| `MAX_APPLICATION_TRANSACTIONS` | `2**14` |
+| `BYTES_PER_LOGS_BLOOM` | `2**8` |
+| `EVM_BLOCK_ROOTS_SIZE` | `2**8` |
+
+
+## Containers
+
+### Extended containers
+
+*Note*: Extended SSZ containers inherit all fields from the parent in the original
+order and append any additional fields to the end.
+
+#### `BeaconBlockBody`
+
+*Note*: `BeaconBlockBody` fields remain unchanged other than the addition of `application_payload`.
+
+```python
+class BeaconBlockBody(phase0.BeaconBlockBody):
+    application_payload: ApplicationPayload  # User execution payload
+```
+
+#### `BeaconState`
+
+*Note*: `BeaconState` fields remain unchanged other than the removal of `eth1_data_votes` and addition of `application_state_root`. 
+The latter stores the root hash of ethereum application state.
+
+```python
+class BeaconState(Container):
+    # Versioning
+    genesis_time: uint64
+    genesis_validators_root: Root
+    slot: Slot
+    fork: Fork
+    # History
+    latest_block_header: BeaconBlockHeader
+    block_roots: Vector[Root, SLOTS_PER_HISTORICAL_ROOT]
+    state_roots: Vector[Root, SLOTS_PER_HISTORICAL_ROOT]
+    historical_roots: List[Root, HISTORICAL_ROOTS_LIMIT]
+    # Eth1
+    eth1_data: Eth1Data
+    # [removed] eth1_data_votes
+    eth1_deposit_index: uint64
+    # [new] Hash of the root of application state
+    application_state_root: Bytes32
+    # [new] Hash of recent application block
+    application_block_hash: Bytes32
+    # Registry
+    validators: List[Validator, VALIDATOR_REGISTRY_LIMIT]
+    balances: List[Gwei, VALIDATOR_REGISTRY_LIMIT]
+    # Randomness
+    randao_mixes: Vector[Bytes32, EPOCHS_PER_HISTORICAL_VECTOR]
+    # Slashings
+    slashings: Vector[Gwei, EPOCHS_PER_SLASHINGS_VECTOR]  # Per-epoch sums of slashed effective balances
+    # Attestations
+    previous_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
+    current_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
+    # Finality
+    justification_bits: Bitvector[JUSTIFICATION_BITS_LENGTH]  # Bit set for every recent justified epoch
+    previous_justified_checkpoint: Checkpoint  # Previous epoch snapshot
+    current_justified_checkpoint: Checkpoint
+    finalized_checkpoint: Checkpoint
+```
+
+### New containers
+
+#### `Transaction`
+
+Application transaction fields structured as an SSZ object for inclusion in an `ApplicationPayload` contained within a `BeaconBlock`.
+
+```python
+class Transaction(Container):
+    nonce: uint64
+    gas_price: uint256
+    gas_limit: uint64
+    recipient: Bytes20
+    value: uint256
+    input: List[Bytes1, MAX_BYTES_PER_TRANSACTION_PAYLOAD]
+    v: uint256
+    r: uint256
+    s: uint256
+```
+
+#### `ApplicationPayload`
+
+The application payload included in a `BeaconBlock`.
+
+```python
+class ApplicationPayload(Container):
+    block_hash: Bytes32  # Hash of application block
+    coinbase: Bytes20
+    state_root: Bytes32
+    gas_limit: uint64
+    gas_used: uint64
+    receipt_root: Bytes32
+    logs_bloom: Vector[Bytes1, BYTES_PER_LOGS_BLOOM]
+    difficulty: uint64  # Temporary field, will be removed later on
+    transactions: List[Transaction, MAX_APPLICATION_TRANSACTIONS]
+```
+
+## Helper functions
+
+### Misc
+
+#### `compute_randao_mix`
+
+```python
+def compute_randao_mix(state: BeaconState, randao_reveal: BLSSignature) -> Bytes32:
+    epoch = get_current_epoch(state)
+    return xor(get_randao_mix(state, epoch), hash(randao_reveal))
+```
+
+#### `compute_time_at_slot` 
+
+```python
+def compute_time_at_slot(state: BeaconState, slot: Slot) -> uint64:
+    return uint64(state.genesis_time + slot * SECONDS_PER_SLOT)
+```
+
+### Beacon state accessors
+
+#### `get_recent_beacon_block_roots`
+
+```python
+def get_recent_beacon_block_roots(state: BeaconState, qty: uint64) -> Sequence[Bytes32]:
+    return [get_block_root_at_slot(state.slot - i) if GENESIS_SLOT + i < state.slot else Bytes32() for i in reversed(range(1, qty + 1))]
+```
+
+#### `get_evm_beacon_block_roots`
+
+```python
+def get_evm_beacon_block_roots(state: BeaconState) -> Sequence[Bytes32]:
+    num_block_roots = min(BLOCK_ROOTS_FOR_EVM_SIZE, SLOTS_PER_HISTORICAL_ROOT)
+    return get_recent_beacon_block_roots(state, num_block_roots)
+```
+
+### Block processing
+
+```python
+def process_block(state: BeaconState, block: BeaconBlock) -> None:
+    process_block_header(state, block)
+    process_randao(state, block.body)
+    process_eth1_data(state, block.body)  # [Modified in The Merge]
+    process_operations(state, block.body)
+    process_application_payload(state, block.body)  # [New in The Merge]
+```
+
+#### Modified `process_eth1_data`
+
+*Note*: The function `process_eth1_data` is modified to update `state.eth1_data` with `eth1_data` of each block.
+
+```python
+def process_eth1_data(state: BeaconState, body: BeaconBlockBody) -> None:
+    state.eth1_data = body.eth1_data
+```
+
+#### Application payload processing
+
+##### `BeaconChainData`
+
+*Note*: `BeaconChainData` contains beacon state data that is used by the application state transition function.
+
+```python
+class BeaconChainData(Container):
+    slot: Slot
+    randao_mix: Bytes32
+    timestamp: uint64
+    recent_block_roots: Sequence[Bytes32]
+```
+
+##### `get_application_state`
+
+*Note*: `ApplicationState` class is an abstract class representing ethereum application state.
+
+Let `get_application_state(application_state_root: Bytes32) -> ApplicationState`  be the function that given the root hash returns a copy of ethereum application state. 
+The body of the function is implementation dependant.
+
+##### `application_state_transition`
+
+Let `application_state_transition(application_state: ApplicationState, beacon_chain_data: BeaconChainData, application_payload: ApplicationPayload) -> None` be the transition function of ethereum application state. 
+The body of the function is implementation dependant.
+
+*Note*: `application_state_transition` must throw `AssertionError` if either transition itself or post-transition verifications has failed.
+
+##### `process_application_payload`
+
+```python
+def process_application_payload(state: BeaconState, body: BeaconBlockBody) -> None:
+    """
+    Note: This function is designed to be able to be run in parallel with 
+    the other `process_block` sub-functions
+    """
+    
+    # Utilizes `compute_randao_mix` to avoid any assumptions about
+    # the processing of other `process_block` sub-functions
+    beacon_chain_data = BeaconChainData(
+        slot=state.slot,
+        randao_mix=compute_randao_mix(state, body.randao_reveal),
+        timestamp=compute_time_at_slot(state.genesis_time, state.slot),
+        recent_block_roots=get_evm_beacon_block_roots(state) 
+    )
+    
+    application_state = get_application_state(state.application_state_root)
+    application_state_transition(application_state, beacon_chain_data, body.application_payload)
+
+    state.application_state_root = body.application_payload.state_root
+    state.application_block_hash = body.application_payload.block_hash
+```
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -164,7 +164,7 @@ def is_transition_completed(state: BeaconState) -> Boolean:
 
 ```python
 def is_transition_block(state: BeaconState, block_body: BeaconBlockBody) -> boolean:
-    return state.application_block_hash == Bytes32() and block.body.application_payload.block_hash != Bytes32()
+    return state.application_block_hash == Bytes32() and block_body.application_payload.block_hash != Bytes32()
 ```
 
 ### Block processing
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -213,4 +213,7 @@ def process_application_payload(state: BeaconState, body: BeaconBlockBody) -> No
     elif is_transition_block(state, body):
         assert body.application_payload == ApplicationPayload(block_hash = body.application_payload.block_hash)
         state.application_block_hash = body.application_payload.block_hash
+    
+    else:
+        assert body.application_payload == ApplicationPayload()
 ```
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -12,6 +12,7 @@
 
 - [Introduction](#introduction)
 - [Constants](#constants)
+  - [Transition](#transition)
   - [Execution](#execution)
 - [Containers](#containers)
   - [Extended containers](#extended-containers)
@@ -22,15 +23,10 @@
     - [`ApplicationPayload`](#applicationpayload)
 - [Helper functions](#helper-functions)
   - [Misc](#misc)
-    - [`compute_randao_mix`](#compute_randao_mix)
-    - [`compute_time_at_slot`](#compute_time_at_slot)
-  - [Beacon state accessors](#beacon-state-accessors)
-    - [`get_recent_beacon_block_roots`](#get_recent_beacon_block_roots)
-    - [`get_evm_beacon_block_roots`](#get_evm_beacon_block_roots)
+    - [`is_transition_completed`](#is_transition_completed)
+    - [`is_transition_block`](#is_transition_block)
   - [Block processing](#block-processing)
-    - [Modified `process_eth1_data`](#modified-process_eth1_data)
     - [Application payload processing](#application-payload-processing)
-      - [`BeaconChainData`](#beaconchaindata)
       - [`get_application_state`](#get_application_state)
       - [`application_state_transition`](#application_state_transition)
       - [`process_application_payload`](#process_application_payload)
@@ -45,6 +41,11 @@ It enshrines application execution and validity as a first class citizen at the 
 
 ## Constants
 
+### Transition
+| Name | Value |
+| - | - |
+| `TRANSITION_TOTAL_DIFFICULTY` | _TBD_ |
+
 ### Execution
 
 | Name | Value |
@@ -52,7 +53,6 @@ It enshrines application execution and validity as a first class citizen at the 
 | `MAX_BYTES_PER_TRANSACTION_PAYLOAD` | `2**20` |
 | `MAX_APPLICATION_TRANSACTIONS` | `2**14` |
 | `BYTES_PER_LOGS_BLOOM` | `2**8` |
-| `EVM_BLOCK_ROOTS_SIZE` | `2**8` |
 
 
 ## Containers
@@ -73,7 +73,7 @@ class BeaconBlockBody(phase0.BeaconBlockBody):
 
 #### `BeaconState`
 
-*Note*: `BeaconState` fields remain unchanged other than the removal of `eth1_data_votes` and addition of `application_state_root` and `application_block_hash`. 
+*Note*: `BeaconState` fields remain unchanged other than addition of `application_state_root` and `application_block_hash`. 
 
 
 ```python
@@ -90,7 +90,7 @@ class BeaconState(Container):
     historical_roots: List[Root, HISTORICAL_ROOTS_LIMIT]
     # Eth1
     eth1_data: Eth1Data
-    # [Removed in Merge] eth1_data_votes
+    eth1_data_votes: List[Eth1Data, EPOCHS_PER_ETH1_VOTING_PERIOD * SLOTS_PER_EPOCH]
     eth1_deposit_index: uint64
     # [Added in Merge] hash of the root of application state
     application_state_root: Bytes32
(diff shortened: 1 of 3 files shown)


specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,216 @@
...
+```python
+def is_transition_completed(state: BeaconState) -> Boolean:
+    state.application_block_hash != Bytes32()
+```
+
+#### `is_transition_block`
+
+```python
+def is_transition_block(state: BeaconState, block_body: BeaconBlockBody) -> boolean:
+    return state.application_block_hash == Bytes32() and block.body.application_payload.block_hash != Bytes32()
```
terencechain, line 167:
```suggestion
    return state.application_block_hash == Bytes32() and block_body.application_payload.block_hash != Bytes32()
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,216 @@
...
+    if is_transition_completed(state):
+        application_state = get_application_state(state.application_state_root)
+        application_state_transition(application_state, body.application_payload)
+
+        state.application_state_root = body.application_payload.state_root
+        state.application_block_hash = body.application_payload.block_hash
+
+    elif is_transition_block(state, body):
+        assert body.application_payload == ApplicationPayload(block_hash = body.application_payload.block_hash)
+        state.application_block_hash = body.application_payload.block_hash
```
terencechain, line 215: Do we need to assign `state.application_state_root` here as well?

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,216 @@
...
+def process_application_payload(state: BeaconState, body: BeaconBlockBody) -> None:
+    """
+    Note: This function is designed to be able to be run in parallel with 
+    the other `process_block` sub-functions
+    """
+
+    if is_transition_completed(state):
+        application_state = get_application_state(state.application_state_root)
+        application_state_transition(application_state, body.application_payload)
+
```
terencechain, line 209: In the original [doc](https://notes.ethereum.org/m9IX3OkkTveXCCOSzGkUiw), there's a check for:
`RLPDecode(block.application_block).parent_hash == state.previous_application_block_hash`

Since we no longer track previous_hash in beacon state and parent_hash in application payload. Are we outsourcing this check to the application engine?

specs/merge/fork-choice.md
```diff
@@ -0,0 +1,102 @@
...
+
+*Note*: It introduces the process of transition from the last PoW block to the first PoS block.
+
+### Helpers
+
+#### `get_total_difficulty`
+
+Let `get_total_difficulty(hash: Bytes32) -> uint256` be the function that returns the total difficulty of the PoW block specified by its hash. 
+
+*Note*: The function returns `0` if the block is either not yet processed or considered invalid. The latter two cases are considered indistinguishable to the current implementation of JSON-RPC.
```
terencechain, line 32: This is where the return has to distinguish between invalid and miss_dependencies right?

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,216 @@
...
+    application_payload: ApplicationPayload  # [Added in Merge] application payload
+```
+
+#### `BeaconState`
+
+*Note*: `BeaconState` fields remain unchanged other than addition of `application_state_root` and `application_block_hash`. 
+
+
+```python
+class BeaconState(Container):
```
seishun, line 80: Why not inherit here as well?

specs/merge/fork-choice.md
mkalinin: Right, we need and extra RPC method on application side that would make the difference. I am going to re-work this part in the spec.

specs/merge/beacon-chain.md
mkalinin: There is `state.application_block_hash` which probably make sense to prefix with `previous_` or `latest_` for conformance. This hash is passed to `eth2_insertBlock` alongside with the `ApplicationPayload` and the application layer tries to execute the application payload on top of a block with the hash.

specs/merge/beacon-chain.md
mkalinin: We don't need to because it must be a vector of zeros in this particular place but we can simplify the code a bit 👍

specs/merge/beacon-chain.md
mkalinin: Though, it might not be that convenient if we decide to go with this proposal and rebase it to Altair later on.

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+
+#### `BeaconState`
+
+*Note*: `BeaconState` fields remain unchanged other than addition of `application_state_root` and `application_block_hash`. 
+
+
+```python
+class BeaconState(phase0.BeaconState):
+    # [Added in Merge] hash of the root of application state
+    application_state_root: Bytes32
```
seishun, line 82: What's the use of this? The only place where it's accessed is `process_application_payload`, which applies the transition function, but it's unclear why the beacon client needs to do it. Vitalik's spec doesn't mention it either.

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
+# Ethereum 2.0 The Merge
+
+**Warning:** This document is based on [Phase 0](../phase0/beacon-chain.md) and considered to be rebased to [Altair](../altair/beacon-chain.md) once the latter is shipped.
```
djrtwo, line 3:
```suggestion
**Warning:** This document is currently based on [Phase 0](../phase0/beacon-chain.md) but will be rebased to [Altair](../altair/beacon-chain.md) once the latter is shipped.
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+      - [`application_state_transition`](#application_state_transition)
+      - [`process_application_payload`](#process_application_payload)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+## Introduction
+
+This is a patch implementing executable beacon chain proposal. 
+It enshrines application execution and validity as a first class citizen at the core of the beacon chain.
```
djrtwo, line 40:
```suggestion
It enshrines application-layer execution and validity as a first class citizen at the core of the beacon chain.
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Constants
+
+### Transition
+| Name | Value |
```
djrtwo, line 45:
```suggestion

| Name | Value |
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+| - | - |
+| `TRANSITION_TOTAL_DIFFICULTY` | _TBD_ |
```
djrtwo, line 47:
```suggestion
| `TRANSITION_TOTAL_DIFFICULTY` | **TBD** |
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Execution
+
+| Name | Value |
+| - | - |
+| `MAX_BYTES_PER_TRANSACTION_PAYLOAD` | `2**20` |
```
djrtwo, line 53:
```suggestion
| `MAX_BYTES_PER_TRANSACTION_PAYLOAD` | `uint64(2**20)` (= 1,048,576) |
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+| `MAX_APPLICATION_TRANSACTIONS` | `2**14` |
```
djrtwo, line 54:
```suggestion
| `MAX_APPLICATION_TRANSACTIONS` | `uint64(2**14)` (= 16,384) |
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+| `BYTES_PER_LOGS_BLOOM` | `2**8` |
```
djrtwo, line 55:
```suggestion
| `BYTES_PER_LOGS_BLOOM` | `uint64(2**8)` (= 256) |
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+
+```python
+class BeaconBlockBody(phase0.BeaconBlockBody):
+    application_payload: ApplicationPayload  # [Added in Merge] application payload
+```
+
+#### `BeaconState`
+
+*Note*: `BeaconState` fields remain unchanged other than addition of `application_state_root` and `application_block_hash`. 
+
```
djrtwo, line 77:
```suggestion
```

specs/merge/beacon-chain.md
djrtwo, line 82:
```suggestion
    # Application-layer
    application_state_root: Bytes32  # [New in Merge]
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+
+```python
+class BeaconState(phase0.BeaconState):
+    # [Added in Merge] hash of the root of application state
+    application_state_root: Bytes32
+    # [Added in Merge] hash of recent application block
+    application_block_hash: Bytes32
```
djrtwo, line 84:
```suggestion
    application_block_hash: Bytes32  # [New in Merge]
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Helper functions
+
+### Misc
+
+#### `is_transition_completed`
+
+```python
+def is_transition_completed(state: BeaconState) -> boolean:
+    state.application_block_hash != Bytes32()
```
djrtwo, line 131:
```suggestion
    return state.application_block_hash != Bytes32()
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+#### `is_transition_block`
+
+```python
+def is_transition_block(state: BeaconState, block_body: BeaconBlockBody) -> boolean:
+    return state.application_block_hash == Bytes32() and block_body.application_payload.block_hash != Bytes32()
```
djrtwo, line 138: Might consider defining a `Hash` type to be more specific than `Bytes32` and to further disambiguate from `Root`

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+
+*Note*: `application_state_transition` must throw `AssertionError` if either the transition itself or one of the post-transition verifications has failed.
+
+##### `process_application_payload`
+
+```python
+def process_application_payload(state: BeaconState, body: BeaconBlockBody) -> None:
+    """
+    Note: This function is designed to be able to be run in parallel with 
+    the other `process_block` sub-functions
```
djrtwo, line 174:
```suggestion
    Note: This function is designed to be able to be run in parallel with the other `process_block` sub-functions.
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+
+    if is_transition_completed(state):
+        application_state = get_application_state(state.application_state_root)
+        application_state_transition(application_state, body.application_payload)
+
+        state.application_state_root = body.application_payload.state_root
+        state.application_block_hash = body.application_payload.block_hash
+
+    elif is_transition_block(state, body):
+        assert body.application_payload == ApplicationPayload(block_hash = body.application_payload.block_hash)
```
djrtwo, line 185:
```suggestion
        assert body.application_payload == ApplicationPayload(block_hash=body.application_payload.block_hash)
```

djrtwo, line 185: Why must this block be empty of payload? Can you not build on a previous pow parent *and* have a normal TX payload at the same block?

djrtwo (commented):
Did a review of beacon-chain.md. Mostly formatting suggestions. One substantive question about the transition block payload

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+
+Let `get_application_state(application_state_root: Bytes32) -> ApplicationState`  be the function that given the root hash returns a copy of ethereum application state. 
+The body of the function is implementation dependent.
+
+##### `application_state_transition`
+
+Let `application_state_transition(application_state: ApplicationState, application_payload: ApplicationPayload) -> None` be the transition function of ethereum application state. 
+The body of the function is implementation dependent.
+
+*Note*: `application_state_transition` must throw `AssertionError` if either the transition itself or one of the post-transition verifications has failed.
```
adiasg, line 166: Do we need `get_application_state()` at all? (also see other comment below)

```suggestion
##### `application_state_transition`

Let `application_state_transition(application_state_root: Bytes32, application_payload: ApplicationPayload) -> Bytes32` be the transition function of ethereum application state. This function takes the pre-state associated with `application_state_root`, applies the `application_payload`, and returns the post-state root.
The body of the function is implementation dependent.

*Note*: `application_state_transition` must throw `AssertionError` if either the transition itself or one of the post-transition verifications has failed.
```

specs/merge/beacon-chain.md
mkalinin: This is the way of bringing the hash of the last PoW block on chain. There is no `parent_hash` field in `ApplicationPayload` structure anymore, otherwise, it would make a lot of sense to do the thing in the way you've described.

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+    """
+    Note: This function is designed to be able to be run in parallel with 
+    the other `process_block` sub-functions
+    """
+
+    if is_transition_completed(state):
+        application_state = get_application_state(state.application_state_root)
+        application_state_transition(application_state, body.application_payload)
+
+        state.application_state_root = body.application_payload.state_root
```
adiasg, line 181: Is there a need for the Eth2 node to access `application_state`? 
A cleaner approach would be to outsource all the work to `application_state_transition()`, so that it operates using the `application_state_root`. (see suggestion for the `application_state_transition` function)
Also, need to check that whatever is returned by `application_state_transition()` actually matches with `body.application_payload.state_root`.

```suggestion
        state.application_state_root = application_state_transition(state.application_state_root, body.application_payload)
        assert state.application_state_root == body.application_payload.state_root
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+      - [`get_application_state`](#get_application_state)
+      - [`application_state_transition`](#application_state_transition)
+      - [`process_application_payload`](#process_application_payload)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+## Introduction
+
+This is a patch implementing executable beacon chain proposal.
```
timbeiko, line 39:
```suggestion
This is a patch implementing the executable beacon chain proposal. 
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+
+#### `Transaction`
+
+Application transaction fields structured as an SSZ object for inclusion in an `ApplicationPayload` contained within a `BeaconBlock`.
+
+```python
+class Transaction(Container):
+    nonce: uint64
+    gas_price: uint256
+    gas_limit: uint64
```
timbeiko, line 97: Post-London, these should be updated to be 1559-compatible.

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+    r: uint256
+    s: uint256
+```
+
+#### `ApplicationPayload`
+
+The application payload included in a `BeaconBlock`.
+
+```python
+class ApplicationPayload(Container):
```
timbeiko, line 111: Post-London, we'll want to add the `BASE FEE` here.

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+    return state.application_block_hash == Bytes32() and block_body.application_payload.block_hash != Bytes32()
+```
+
+### Block processing
+
+```python
+def process_block(state: BeaconState, block: BeaconBlock) -> None:
+    process_block_header(state, block)
+    process_randao(state, block.body)
+    process_eth1_data(state, block.body)
```
timbeiko, line 147:
```suggestion
    process_application_data(state, block.body)
```

djrtwo:
> Is there a need for the Eth2 node to access application_state?

@adiasg I would argue that this is the cleaner approach as "outsourcing" all the work is actually an implementation detail (e.g. a client could bundle pos and application components tightly or could do something like we are doing with the separation of consensus vs application client)

specs/merge/validator.md
```diff
@@ -0,0 +1,75 @@
...
+def get_application_payload(state: BeaconState) -> ApplicationPayload:
+    if is_transition_completed(state):
+      application_parent_hash = state.application_block_hash
+      return produce_application_payload(application_parent_hash)
+    
+    pow_block = get_pow_chain_head()
+    if pow_block.total_difficulty >= TRANSITION_TOTAL_DIFFICULTY:
+      return ApplicationPayload(block_hash = pow_block.block_hash)
+    else:
+      return ApplicationPayload()
```
djrtwo, line 74: I would put the exceptional cases in a blocking `if` condition and have the "normal" post-merge case outside of the if. also added some comments

```suggestion
def get_application_payload(state: BeaconState) -> ApplicationPayload:
    if not is_transition_completed(state):
        pow_block = get_pow_chain_head()
        if pow_block.total_difficulty < TRANSITION_TOTAL_DIFFICULTY:
            # Pre-merge, empty payload
            return ApplicationPayload()
        else:
            # Signify merge via last PoW block_hash and an otherwise empty payload
            return ApplicationPayload(block_hash = pow_block.block_hash)

    # Post-merge, normal payload                
    application_parent_hash = state.application_block_hash
    return produce_application_payload(application_parent_hash)
    
```

specs/merge/validator.md
```diff
@@ -0,0 +1,75 @@
...
+###### `get_pow_chain_head`
+
+Let `get_pow_chain_head() -> PowBlock` be the function that returns the head of the PoW chain. The body of the function is implementation specific.
+
+###### `produce_application_payload`
+
+Let `produce_application_payload(parent_hash: Bytes32) -> ApplicationPayload` be the function that produces new instance of application payload.
+The body of this function is implementation dependent.
+
+* Set `block.body.application_payload = get_application_payload(state, randao_reveal)` where:
```
djrtwo, line 62: This function signature does not map to the definition below. Has `randao_reveal`

djrtwo (commented):
fantastic work! got through the next two docs. happy to discuss here or otherwise

adiasg:
@djrtwo 
> @adiasg I would argue that this is the cleaner approach because "outsourcing" all the work is actually an implementation detail (e.g. a client could bundle pos and application components tightly or could do something like we are doing with the separation of consensus vs application client)

I partly agree - clients may do application processing in the Eth2 node itself, or outsource the work to a separate application client. That's why I think `application_state_transition(application_state_root: Bytes32, application_payload: ApplicationPayload)` is better, as it accounts for both cases (because `ApplicationState` is not required in the Eth2 node if outsourcing the work). 

The way this behavior is currently defined (with `get_application_state(application_state_root: Bytes32) -> ApplicationState`) makes it seem like `ApplicationState` is essential for Eth2 validation logic, which is not the case.

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,190 @@
...
+class ApplicationPayload(Container):
+    block_hash: Bytes32  # Hash of application block
+    coinbase: Bytes20
+    state_root: Bytes32
+    gas_limit: uint64
+    gas_used: uint64
+    receipt_root: Bytes32
+    logs_bloom: Vector[Bytes1, BYTES_PER_LOGS_BLOOM]
+    difficulty: uint64  # Temporary field, will be removed later on
+    transactions: List[Transaction, MAX_APPLICATION_TRANSACTIONS]
```
seishun, line 120: The only two fields in this class referenced by the spec are `block_hash` and `state_root`. Perhaps it would make things simpler to define `ApplicationPayload` as an abstract class?

protolambda, line 120: Maybe define this as `transactions: List[Union[Transaction], MAX_APPLICATION_TRANSACTIONS]` to add a 4 byte selector to enable different transaction types in the future?

Optimization note: the Union type in SSZ is unused in Eth2 so far. We can parametrize the selector length, to shorten it to a 1 byte selector.

specs/merge/beacon-chain.md
mkalinin: I agree that `Eth1Data` name will look odd after the merge but we might want to keep it as it this time. The follow up cleanups are going to reduce the follow distance of `Eth1Data` and it would be a good time to make the renaming too.

specs/merge/beacon-chain.md
mkalinin: Application layer will need to be able to assemble an application block from an instance of `ApplicationPayload` and `parent_hash`. Thus, all these fields are important to be held by the beacon block body. Other application block fields (ethash, extra data) will be required to have zero values which will be verified by matching the `application_payload.block_hash` against the hash of assembled block on the application layer.

mkalinin: Are you referring to this EIP https://eips.ethereum.org/EIPS/eip-2718? If yes then I would wait until the EIP is approved and shipped on the Mainnet. The intention is to replicate the application layer structures from those that will be on the Mainnet at the point of merge.

specs/merge/fork-choice.md
```diff
@@ -0,0 +1,113 @@
...
+
+*Note*: It introduces the process of transition from the last PoW block to the first PoS block.
+
+### Helpers
+
+#### `PowBlock`
+
+```python
+class PowBlock(Container):
+    is_processed: boolean
```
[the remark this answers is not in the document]
mkalinin: Current behaviour is to return error if either invalid or not yet processed block is requested meaning that these two statuses are indistinguishable. So, if we want this level of granularity then JSON-RPC implementation will have to be adjusted but we probably don't want it. It's worth discussing.

specs/merge/fork-choice.md
```diff
@@ -0,0 +1,113 @@
...
+    is_valid: boolean
+    total_difficulty: uint256
```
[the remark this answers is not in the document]
mkalinin: Yes, total difficulty falls into `> 2**74` interval today. The other potential way of handling this is to return an offset wrt some absolute total difficulty value. Current block's difficulty is around `2**64`, so we indeed may divide total difficulty by `2**20` without loss of generality.

specs/merge/fork-choice.md
```diff
@@ -0,0 +1,113 @@
...
+    # Check that block is later than the finalized epoch slot (optimization to reduce calls to get_ancestor)
+    finalized_slot = compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)
+    assert block.slot > finalized_slot
+    # Check block is a descendant of the finalized block at the checkpoint finalized slot
+    assert get_ancestor(store, block.parent_root, finalized_slot) == store.finalized_checkpoint.root
+    
+    # [Added in Merge] Consider delaying the beacon block processing until PoW block is accepted by the application node
+    if is_transition_block(pre_state, block.body):
+        pow_block = get_pow_block(block.body.application_payload.block_hash)
+        assert is_valid_transition_block(pow_block)
```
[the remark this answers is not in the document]
mkalinin: Do you mean something like that?

```python
# Consider delaying the beacon block processing until PoW block is processed by the application layer
assert pow_block.is_processed
assert is_valid_transition_block(pow_block)
```

specs/merge/beacon-chain.md
protolambda: Also see #2270, to support `Union` with a 1-byte selector

specs/merge/fork-choice.md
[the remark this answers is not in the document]
mkalinin: > Do eth1 clients remember if an invalid block has already been processed? or does it just drop it

I think with some configuration it stores invalid blocks to be able to serve `debug_getBadBlock`

specs/merge/fork-choice.md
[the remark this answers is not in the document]
mkalinin: Also, it worth noting that `uint256` type is used by `Transaction` data structure to define several fields.

specs/merge/beacon-chain.md
mkalinin: I definitely see the value in simplifying this part. But then it would not read like the beacon state and the application state are tightly coupled.

For the root check we can do the following:
```python
application_state = get_application_state(state.application_state_root)
application_state_transition(application_state, body.application_payload)

assert body.application_payload.state_root == get_application_state_root(application_state)

state.application_state_root = body.application_payload.state_root
state.application_block_hash = body.application_payload.block_hash
```

But that would require yet another abstract function. So, it might really be better to have the following (according to your suggestion):
```python
application_state_root = application_state_transition(state.application_state_root, body.application_payload)

assert application_state_root == body.application_payload.state_root

state.application_state_root = body.application_payload.state_root
state.application_block_hash = body.application_payload.block_hash
```

The latter form also fits stateless verification.

specs/merge/beacon-chain.md
djrtwo: It's more of `DepositContractData` if we are going to rename it.
It is more specific and application layer data

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### `BeaconState`
+
+*Note*: `BeaconState` fields remain unchanged other than addition of `application_state_root` and `application_block_hash`.
+
+```python
+class BeaconState(phase0.BeaconState):
+    # Application-layer
+    application_state_root: Bytes32 # [New in Merge]
+    application_block_hash: Bytes32 # [New in Merge]
```
djrtwo, line 83: Code style: two spaces after code before an in-line comment

```suggestion
    application_state_root: Bytes32  # [New in Merge]
    application_block_hash: Bytes32  # [New in Merge]
```

specs/merge/beacon-chain.md
djrtwo: right! But as we discussed offline, probably not worth adding 32-bytes to all payloads just for this transition case. Have a single slot delay does not really degrade quality of service especially considering pow Poisson process regularly has block times that are quite longer than 20 seconds

specs/merge/fork-choice.md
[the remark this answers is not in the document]
djrtwo: ah, right. I don't thnk we'll be able to get around `value` with 256 granularity

I suppose beacon clients don't actually have to support arithmetic on these TX values because all arithmetic and validations happen in application layer so as long as they can serialize and deserialize, that's enough support

specs/merge/validator.md
```diff
@@ -0,0 +1,70 @@
...
+```python
+def get_application_payload(state: BeaconState) -> ApplicationPayload:
+    if not is_transition_completed(state):
+      pow_block = get_pow_chain_head()
+      if pow_block.total_difficulty < TRANSITION_TOTAL_DIFFICULTY:
+        # Pre-merge, empty payload
+        return ApplicationPayload()
+      else:
+        # Signify merge via last PoW block_hash and an otherwise empty payload
+        return ApplicationPayload(block_hash=pow_block.block_hash)
```
djrtwo, line 65: Code style: 4 space indent
```suggestion
    if not is_transition_completed(state):
        pow_block = get_pow_chain_head()
        if pow_block.total_difficulty < TRANSITION_TOTAL_DIFFICULTY:
            # Pre-merge, empty payload
            return ApplicationPayload()
        else:
            # Signify merge via last PoW block_hash and an otherwise empty payload
            return ApplicationPayload(block_hash=pow_block.block_hash)
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,188 @@
...
+*Note*: Extended SSZ containers inherit all fields from the parent in the original
+order and append any additional fields to the end.
+
+#### `BeaconBlockBody`
+
+*Note*: `BeaconBlockBody` fields remain unchanged other than the addition of `application_payload`.
+
+```python
+class BeaconBlockBody(phase0.BeaconBlockBody):
+    application_payload: ApplicationPayload  # [Added in Merge] application payload
```
djrtwo, line 72:
```suggestion
    application_payload: ApplicationPayload  # [New in Merge]
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,188 @@
...
+Application transaction fields structured as an SSZ object for inclusion in an `ApplicationPayload` contained within a `BeaconBlock`.
+
+```python
+class Transaction(Container):
+    nonce: uint64
+    gas_price: uint256
+    gas_limit: uint64
+    recipient: Bytes20
+    value: uint256
+    input: List[Bytes1, MAX_BYTES_PER_TRANSACTION_PAYLOAD]
```
ralexstokes, line 99:
```suggestion
    input: List[byte, MAX_BYTES_PER_TRANSACTION_PAYLOAD]
```

i believe this is a lighter weight encoding according to the ssz spec (where it does define a `byte` as a `uint8`, rather than `Bytes1` which is a `Vector[byte, 1]`

specs/merge/fork-choice.md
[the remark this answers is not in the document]
djrtwo: Yeah, that looks good. 

A note on the comment. "consider delaying" and "delay consideration" has a surprisingly different meaning. "Consider delaying" is optional whereas "delay consideration" is more of a directive and less optional.

I would write the comment:
> Delay consideration of `block` until PoW block is processed by the PoW node

and I would put it *inside* of the `if is_transition_block` rather than on the outside
*

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,188 @@
...
+
+```python
+class ApplicationPayload(Container):
+    block_hash: Bytes32  # Hash of application block
+    coinbase: Bytes20
+    state_root: Bytes32
+    gas_limit: uint64
+    gas_used: uint64
+    receipt_root: Bytes32
+    logs_bloom: Vector[Bytes1, BYTES_PER_LOGS_BLOOM]
```
ralexstokes, line 117:
```suggestion
    logs_bloom: Vector[byte, BYTES_PER_LOGS_BLOOM]
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,188 @@
...
+    difficulty: uint64  # Temporary field, will be removed later on
```
ralexstokes, line 118: why can't we remove this field now?

paulmillr:
Thank you for the hard work!

A quick question from a guy who has not been following the plans very thoroughly. 

What hazardous or bad tech debt does this merge proposal create when compared to the "old" merge plan?

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,188 @@
...
+
+```python
+def process_application_payload(state: BeaconState, body: BeaconBlockBody) -> None:
+    """
+    Note: This function is designed to be able to be run in parallel with the other `process_block` sub-functions
+    """
+
+    if is_transition_completed(state):
+        application_state = get_application_state(state.application_state_root)
+        application_state_transition(application_state, body.application_payload)
```
lsankar4033, line 177: would it make sense to pass in the randao or some other seed for `DIFFICULTY`/`BLOCKHASH` here?

obviously, easiest to just stick them on the `application_payload` so the eth1 engine doesn't even have to know about this new logic.

specs/merge/beacon-chain.md
hwwhww: nitpicking: `application_state_root = application_state_transition(...)` requires `application_state_transition` to return `Bytes32`, but it doesn't align with the naming pattern as beacon state transition function `state_transition(...) -> None`. So having a `get_application_state_root(application_state)` makes sense to me.

specs/merge/beacon-chain.md
hwwhww, line 99: What do you think about renaming `input` to `data`?

specs/merge/beacon-chain.md
mkalinin: Yep, I think it would make sense to prepare a consensus bundle that will be passed onto this function with randao mix and further extended by other bits. I'd add it later once we made the decision about difficulty and whether to use randao or not.

specs/merge/beacon-chain.md
mkalinin: Making this call conformant with `state_transition(...)` was the original intention put behind this extra function call and using application state object rather than the state root.

mkalinin: We may also define `ApplicationState` in a following way
```python
class ApplicationState(Container):
    root: Bytes32
```

And then add an explicit check of the state root after application state transition
```python
application_state = get_application_state(state.application_state_root)
application_state_transition(application_state, body.application_payload)

assert application_state.root == body.application_payload.state_root

state.application_state_root = body.application_payload.state_root
state.application_block_hash = body.application_payload.block_hash
```

specs/merge/beacon-chain.md
protolambda: `log_bloom: ByteVector[BYTES_PER_LOGS_BLOOM]` would also work (difference: more alike to python `bytes` instead of a python `list[int]`)

mkalinin: Is there a `ByteList` as well? It could fit transaction data, e.g. `data: ByteList[MAX_BYTES_PER_TRANSACTION_PAYLOAD]`.

protolambda: Yes, there's a `ByteList[N]` as well, which should be faster and more similar to `bytes`, just like `ByteVector[N]`

vbuterin:
It does seem like this goes pretty far in changing formatting at the same time as the merge, particularly the `transactions` list. Remember that by this point there will be plenty of transaction types: old-style with no replay protection, old-style with replay protection, EIP 2930 access list-carrying txs, EIP 1559 basefee-setting txs. These would _all_ have to be changed into an SSZ format.

To keep merge complexity down, might it not be simpler to just keep `transactions` as a list of blobs, and then have some post-merge fork replace them with a new SSZ-based transaction type? (first make it voluntary, and then make it mandatory)

```
class ApplicationPayload(Container):
    block_hash: Bytes32  # Hash of application block
    coinbase: Bytes20
    state_root: Bytes32
    gas_limit: uint64
    gas_used: uint64
    receipt_root: Bytes32
    transactions: List[Transaction, MAX_APPLICATION_TRANSACTIONS]
```

I noticed that this does not include a bunch of fields (`uncles_hash`, `difficulty`, `number`, `timestamp`, `mixhash`, `nonce`). This makes the conversion between old-style and new-style application blocks not a clean reversible function. Perhaps this does not matter, because those fields are not relevant anymore, but it _does_ create some extra special cases. Particularly, it requires an extra special case for the transition application block itself: if the transition application block is the last PoW-bearing block, then you would need some special logic for including that block, because the last PoW-bearing block _does_ have an `uncles_hash`, `nonce`, etc.

Or is the idea to change that bit in the spec, so that instead of the transition beacon block including the last PoW block, it merely includes a block whose _parent_ is the last PoW block? If so, then I could see how this can work.

protolambda:
@vbuterin fair point. I have this issue open to track `Union` support to switch between (the many) transaction types, and already proposed the option of a list of opaque transactions.

Another option is to do `transactions: List[Union[TxBlob], MAX_APPLICATION_TRANSACTIONS]` and expand the `Union[TxBlob]` later with more SSZ transaction types. That way we avoid the immediate complexity of many new SSZ transaction types, have something to fall back on with all legacy transactions, and it's much more compatible to extend with SSZ transactions in future forks.

mkalinin:
@vbuterin @protolambda IMO, an opaque transactions option is the best for the beginning. IIUC, some of old-styled transaction types will become unacceptable starting from some point in time and it would make more sense to replace transaction blobs with SSZ objects once all the formats are settled down to avoid them being heavily dependent on the beacon chain spec.

mkalinin:
> Particularly, it requires an extra special case for the transition application block itself

@vbuterin It will be a special case. As `ApplicationPayload` does not explicitly contain `parent_hash`, the hash of the last PoW block will be brought up on chain before producing the first PoS block. It's gonna be done by adding `ApplicationPayload(block_hash=transition_block_hash)` to the beacon block body to denote a transition block. It also requires a special case in the `state_transition` function to process transition block which just sets `state.application_block_hash = application_payload.block_hash` without interacting with the application layer.

mkalinin:
> Thank you for the hard work!
> 
> A quick question from a guy who has not been following the plans very thoroughly.
> 
> What hazardous or bad tech debt does this merge proposal create when compared to the "old" merge plan?

The list of changes that this proposal sets for the hard fork(s) after the merge is as follows:
- `Eth1Data` follow distance reduction
- New EVM opcodes, `RANDOM`, `BEACONBLOCKROOT`, etc.
- Validator withdrawals

*Note:* Withdrawals are dependent on the opcodes

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,185 @@
...
+    value: uint256
+    data: ByteList[MAX_BYTES_PER_TRANSACTION_PAYLOAD]
+    v: uint256
+    r: uint256
+    s: uint256
+```
+
+#### `ApplicationPayload`
+
+The application payload included in a `BeaconBlock`.
```
terencechain, line 107: Nit. Application payload is included in block body

```suggestion
The application payload included in a `BeaconBlockBody`.
```

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,185 @@
...
+
+```python
+class ApplicationPayload(Container):
+    block_hash: Bytes32  # Hash of application block
+    coinbase: Bytes20
+    state_root: Bytes32
+    gas_limit: uint64
```
terencechain, line 114: How is this `gas_limit` different than the `gas_limit` inside the `Transaction`?

specs/merge/beacon-chain.md
```diff
@@ -0,0 +1,185 @@
...
+    # Application-layer
+    application_state_root: Bytes32  # [New in Merge]
+    application_block_hash: Bytes32  # [New in Merge]
+```
+
+### New containers
+
+#### `Transaction`
+
+Application transaction fields structured as an SSZ object for inclusion in an `ApplicationPayload` contained within a `BeaconBlock`.
```
terencechain, line 90:
```suggestion
Application transaction fields structured as an SSZ object for inclusion in an `ApplicationPayload` contained within a `BeaconBlockBody`.
```

specs/merge/beacon-chain.md
mkalinin: This is a gas limit of an application block

ralexstokes: the block gas limit vs the gas limit for each individual transaction.

we should most certainly retain the validity rule at the application layer that the block gas limit is changed w/in the parameters of the protocol -- so the application node will need to verify this number but the beacon chain is otherwise indifferent to it

terencechain: Got it. I was thinking it'd just be `MAX_APPLICATION_TRANSACTIONS * transaction.gas_limit`

But different transaction types could have different gas limit

djrtwo (approved):
Fantastic work @mkalinin and to the many reviewers!

To better manage the complexity, we're going to merge the current PR as is. I've created an issue (https://github.com/ethereum/eth2.0-specs/issues/2280) to track the open discussion points and todos that were brought up but not resolved in PR.

From here, we'll move toward smaller and more iterative PRs to address these and any other points that come up along the way

PR #2258, opened 2021-03-21
reduce justification and finalization Altair diff

protolambda:
No functional changes, just presentation:
- Split Phase 0 `process_justification_and_finalization` into calculation of target stake and `weigh_justification_and_finalization` to consider the stake totals and update state.
- Update Altair to first calculate target stake, then run `weigh_justification_and_finalization`

This helps reduce the diff, avoids duplicate code, and improves the performance of the spec implementation (`get_total_active_balance(state)` only has to be called once).

Suggestions for variable and function naming welcome.

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -575,7 +575,7 @@ def process_epoch(state: BeaconState) -> None:
 
 #### Justification and finalization
 
-*Note*: The function `process_justification_and_finalization` is modified with `matching_target_attestations` replaced by `matching_target_indices`.
+*Note*: The function `process_justification_and_finalization` is modified to adapt to the new participation records.
 
 ```python
 def process_justification_and_finalization(state: BeaconState) -> None:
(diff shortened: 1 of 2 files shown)


specs/altair/beacon-chain.md
```diff
@@ -575,48 +575,20 @@ def process_epoch(state: BeaconState) -> None:
...
-        state.finalized_checkpoint = old_current_justified_checkpoint
-    # The 1st/2nd most recent epochs are justified, the 1st using the 2nd as source
-    if all(bits[0:2]) and old_current_justified_checkpoint.epoch + 1 == current_epoch:
-        state.finalized_checkpoint = old_current_justified_checkpoint
+    previous = get_unslashed_participating_indices(state, TIMELY_TARGET_FLAG_INDEX, get_previous_epoch(state))
+    current = get_unslashed_participating_indices(state, TIMELY_TARGET_FLAG_INDEX, get_current_epoch(state))
+    weigh_justification_and_finalization(
+        state, get_total_active_balance(state),
+        get_total_balance(state, previous),
+        get_total_balance(state, current))
```
djrtwo, line 591: Generally been preferring more intermediate setters to avoid multi-line function calls

```python
    total_active_balance = get_total_active_balance(state)
    previous_total_balance = get_total_balance(state, previous)
    current_total_balance = get_total_balance(state, current)
    weigh_justification_and_finalization_(state, total_active_balance, previous_total_balance, current_total_balance)
```

specs/phase0/beacon-chain.md
```diff
@@ -1320,7 +1320,18 @@ def process_justification_and_finalization(state: BeaconState) -> None:
...
+    previous = get_matching_target_attestations(state, get_previous_epoch(state))
+    current = get_matching_target_attestations(state, get_current_epoch(state))
+    weigh_justification_and_finalization(
+        state, get_total_active_balance(state),
+        get_attesting_balance(state, previous),
+        get_attesting_balance(state, current))
+```
 
+```python
+def weigh_justification_and_finalization(state: BeaconState, total_active_balance: Gwei,
```
djrtwo, line 1332: I don't think spec ever does the two params on a multi-line function call like this. prefer to keep it standard so implementers get the visual queue

```suggestion
def weigh_justification_and_finalization(state: BeaconState,
                                         total_active_balance: Gwei,
```

djrtwo (approved):
Nits. I'm happy to do the edits

PR #2267, opened 2021-03-23
[WIP] Light client protocol + Req/Resp interaction

hwwhww:
WIP!!!

New and moved files
- specs/altair/light-client/sync-protocol.md
- specs/altair/light-client/p2p-interface.md


tests/core/pyspec/eth2spec/utils/ssz/ssz_impl.py
```diff
@@ -23,3 +24,23 @@ def uint_to_bytes(n: uint) -> bytes:
 # Helper method for typing copies, and avoiding a example_input.copy() method call, instead of copy(example_input)
 def copy(obj: V) -> V:
     return obj.copy()
+
+
+def build_proof(anchor, leaf_index):
```
hwwhww, line 29: note: it's for the incoming `get_light_client_update` helper

specs/altair/light-client/p2p-interface.md
```diff
@@ -0,0 +1,56 @@
...
+
+Request Content:
+
+```
+(
+  LightClientUpdate
+)
+```
+
+No Response Content.
```
protolambda, line 54: If this is a push-update, and there is no explicit subscription kind of approach for a light-client, then we should spec out when and how often the server pushes the update.

specs/altair/light-client/sync-protocol.md
```diff
@@ -31,11 +42,38 @@
...
+uses sync committees introduced in [this beacon chain extension](./../beacon-chain.md).
+
+## Sync protocol
+
+Note that in this syncing mechanism, the server is trusted.
+
+### Initialization
+
+1. The client sends [`Status` message](./../../phase0/p2p-interface.md#status) to the server to exchange the status information.
+2. Instead of sending [`BeaconBlocksByRange` request](./../../phase0/p2p-interface.md#beaconblocksbyrange) in the beacon chain syncing, the client sends [`GetLightClientSnapshot` request](./p2p-interface.md#getlightclientsnapshot) to the server.
```
djrtwo, line 58: So a light client must not just trust the server for a good snapshot. They must instead start from a known safe place and transition *trustlessly* from there

djrtwo (commented):
A general comments and questions to start

specs/altair/light-client/p2p-interface.md
```diff
@@ -0,0 +1,56 @@
...
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+## The Req/Resp domain
+
+### Messages
+
+#### GetLightClientSnapshot
+
+**Protocol ID:** `/eth2/beacon_chain/req/get_light_client_snapshot/1/`
```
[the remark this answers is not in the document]
hwwhww: Right, Req/Resp was the simplest way to get a snapshot for MVP.

Alternatively, what do you think about saying that we expect 3rd party to provide `LightClientSnapshot` + Merkle proofs or multi-proof of each field, where the proof can be verified against WS source?

specs/altair/light-client/sync-protocol.md
```diff
@@ -77,8 +113,7 @@ class LightClientUpdate(Container):
     finality_header: BeaconBlockHeader
     finality_branch: Vector[Bytes32, floorlog2(FINALIZED_ROOT_INDEX)]
     # Sync committee aggregate signature
-    sync_committee_bits: Bitvector[SYNC_COMMITTEE_SIZE]
-    sync_committee_signature: BLSSignature
+    sync_aggregate: SyncAggregate
     # Fork version for the aggregate signature
     fork_version: Version
```
dapplion, line 118: I believe almost all consumers of this data structure can compute the fork_version on their own. The only scenario where it could make sense is in LC embedded in blockchains, but even then you can input fork info in other ways.

specs/altair/light-client/p2p-interface.md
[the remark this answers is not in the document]
dapplion: > Alternatively, what do you think about saying that we expect 3rd party to provide LightClientSnapshot + Merkle proofs or multi-proof of each field, where the proof can be verified against WS source?

Agree :+1:

specs/altair/light-client/p2p-interface.md
dapplion: Relying on LightClientUpdate data structure to update a lightclient head is very wasteful since it only need to get the next_sync_committee pubkeys once every 256 epochs. To be honest LightClientUpdate works fine in a spec but it's not great in practice. The process of syncing can be split into two steps:
- Get the pubkeys of the current sync committee. Use req/resp on demand to query LightClientUpdate(s)
- Get latest SyncAggregate + header. More optimal to be gossiped once or more per slot.

However, gossiping SyncAggregate has an issue of how to decide which messages to accept and ignore. There are many combinations of the same signatures that results in different SyncAggregate objects (i.e. https://github.com/ethereum/consensus-specs/issues/2183).

specs/altair/light-client/p2p-interface.md
```diff
@@ -0,0 +1,61 @@
...
+
+Response Content:
+
+```
+(
+  LightClientSnapshot
+)
+```
+
+The `LightClientSnapshot` SSZ container defined in [light client sync protocol](./sync-protocol.md#lightclientsnapshot).
```
dapplion, line 41: To initialize a lightclient you don't need the next_sync_committee, but only the current_sync_committee. You can get the next via syncing.

PR #2271, opened 2021-03-24
Typing problems fixes

ericsson49:
Typing problems have been found during static analysis of beacon chain pyspecs.
The PR contains suggested fixes for some of them, which can be fixed in a straightforward way (e.g. by converting to the appropriate type).

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -517,7 +517,7 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
             state.balances.append(amount)
             state.previous_epoch_participation.append(ParticipationFlags(0b0000_0000))
             state.current_epoch_participation.append(ParticipationFlags(0b0000_0000))
-            state.inactivity_scores.append(0)
+            state.inactivity_scores.append(uint64(0))
     else:
         # Increase balance by deposit amount
         index = ValidatorIndex(validator_pubkeys.index(pubkey))
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -616,10 +616,7 @@ def process_rewards_and_penalties(state: BeaconState) -> None:
         return
 
     flag_indices_and_numerators = get_flag_indices_and_weights()
-    flag_deltas = [
-        get_flag_index_deltas(state, index, numerator)
-        for (index, numerator) in flag_indices_and_numerators
-    ]
+    flag_deltas = [get_flag_index_deltas(state, index, numerator) for (index, numerator) in flag_indices_and_numerators]
     deltas = flag_deltas + [get_inactivity_penalty_deltas(state)]
     for (rewards, penalties) in deltas:
         for index in range(len(state.validators)):
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -84,11 +84,11 @@ Altair is the first beacon chain hard fork. Its main features are:
 
 | Name | Value |
 | - | - |
-| `TIMELY_HEAD_WEIGHT` | `12` |
-| `TIMELY_SOURCE_WEIGHT` | `12` |
-| `TIMELY_TARGET_WEIGHT` | `24` |
-| `SYNC_REWARD_WEIGHT` | `8` |
-| `WEIGHT_DENOMINATOR` | `64` |
+| `TIMELY_HEAD_WEIGHT` | `uint64(12)` |
+| `TIMELY_SOURCE_WEIGHT` | `uint64(12)` |
+| `TIMELY_TARGET_WEIGHT` | `uint64(24)` |
+| `SYNC_REWARD_WEIGHT` | `uint64(8)` |
+| `WEIGHT_DENOMINATOR` | `uint64(64)` |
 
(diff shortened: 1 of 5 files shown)


specs/phase0/beacon-chain.md
```diff
@@ -1675,7 +1675,7 @@ def process_block_header(state: BeaconState, block: BeaconBlock) -> None:
         slot=block.slot,
         proposer_index=block.proposer_index,
         parent_root=block.parent_root,
-        state_root=Bytes32(),  # Overwritten in the next process_slot call
+        state_root=Root(),  # Overwritten in the next process_slot call
```
protolambda, line 1678: We tend to use `Bytes32` instead of `Root()` or `Hash()` to avoid confusion with a hash-tree-root or hash operation (`hash()` may be interpreted as a sha-256 of an empty input). Alternatively we could use a `ZERO_ROOT` constant or something like that.

specs/altair/beacon-chain.md
```diff
@@ -616,7 +616,10 @@ def process_rewards_and_penalties(state: BeaconState) -> None:
         return
 
     flag_indices_and_numerators = get_flag_indices_and_weights()
-    flag_deltas = [get_flag_index_deltas(state, index, numerator) for (index, numerator) in flag_indices_and_numerators]
+    flag_deltas = [
+        get_flag_index_deltas(state, index, uint64(numerator))
+        for (index, numerator) in flag_indices_and_numerators
+    ]
```
protolambda, line 622: Maybe we can update the `get_flag_indices_and_weights()` return type instead, and type the numerator constants?

specs/phase1/beacon-chain.md
```diff
@@ -425,7 +425,7 @@ def compute_previous_slot(slot: Slot) -> Slot:
 #### `pack_compact_validator`
 
 ```python
-def pack_compact_validator(index: ValidatorIndex, slashed: bool, balance_in_increments: uint64) -> uint64:
+def pack_compact_validator(index: ValidatorIndex, slashed: boolean, balance_in_increments: uint64) -> uint64:
```
protolambda, line 428: This should be a `bool`, but the below formula that shifts the boolean value is unafe (although python converts it to `int` automatically). It should convert the `bool` into a `uint64`, then shift it afterwards. Good find.

ericsson49: Thanx! Corrected

specs/altair/beacon-chain.md
ericsson49: Good idea! Done.
I've changed types of `SYNC_REWARD_WEIGHT` and `WEIGHT_DENOMINATOR` for consistency too.

specs/phase0/beacon-chain.md
protolambda: Not sure about the `ZERO_ROOT` actually, we should not change the phase0 specs for smaller things like this. `Bytes32` isn't bad. Which type checker did you use?

ericsson49: Thanx! I like the `ZERO_ROOT` idea!

ericsson49: > Not sure about the ZERO_ROOT actually, we should not change the phase0 specs for smaller things like this. 

This is reasonable, definitely! Haven't thought about that. I'll revert then.

> `Bytes32` isn't bad.

It's not bad, but semantics become a bit complicated.
`state_root` has `Root` type, so assigning `Bytes32` value to it means implicit conversion.
Actually, I believe I have to model implicit conversions anyway.

> Which type checker did you use?

I'm using a custom type checker. It's based on PEP-484, of course, but I'm more flexible with my tool. E.g. can introduce my own typing rules to match the pyspecs better.
My goal eventually is to define formal semantics for pyspecs, i.e. to a subset of Python, used to express the pyspecs.

specs/phase0/beacon-chain.md
```diff
@@ -1159,7 +1159,7 @@ def initialize_beacon_state_from_eth1(eth1_block_hash: Bytes32,
     state = BeaconState(
         genesis_time=eth1_timestamp + GENESIS_DELAY,
         fork=fork,
-        eth1_data=Eth1Data(block_hash=eth1_block_hash, deposit_count=len(deposits)),
+        eth1_data=Eth1Data(block_hash=eth1_block_hash, deposit_count=uint64(len(deposits))),
```
ericsson49, line 1162: I guess that should be reverted too, since it's phase0 code?

specs/phase0/beacon-chain.md
ericsson49: Reverted back to `state_root=Bytes32()`

specs/phase0/beacon-chain.md
protolambda: It's phase0 so less changes is better, but you're right that it could use stronger typing. And it doesn't change the behavior, nor does it affect readability all that much, and doesn't introduce new constants or functions. So this seems fine.

protolambda (approved):
Looks good to me, will squash and merge after checking the Altair weight constant types with others.

hwwhww (approved):

djrtwo:
`uint64` for weights seems fine @protolambda

Issue #2269, opened 2021-03-23
filtered_block_tree fork choice test expected justified checkpoint to change without any new input


ajsutton:
In v1.1.00-alpha.2, the fork choice test filtered_block_tree contains steps including:

```
<snip>
- {block: block_0x550a5c7bb092ffc787dafa462d75acfb9a9dfbca49451ef9ff3db939c7e26c0b}
- checks:
    time: 144
    head: {slot: 24, root: '0xff0565c092137a448f095479b1d3f73c307a3a59357aa983e9a9f1c1a21a3bca'}
    justified_checkpoint_root: '0x267b47b08d6fa978d84e652e402d0c0784d6dcdff664f49680b83441c287e866'
    finalized_checkpoint_root: '0x267b47b08d6fa978d84e652e402d0c0784d6dcdff664f49680b83441c287e866'
    best_justified_checkpoint: '0x267b47b08d6fa978d84e652e402d0c0784d6dcdff664f49680b83441c287e866'
- checks:
    head: {slot: 24, root: '0xff0565c092137a448f095479b1d3f73c307a3a59357aa983e9a9f1c1a21a3bca'}
    justified_checkpoint_root: '0xad558a267ddc732b65a7f749fc5c28c59304742445ae23642156a116eb2b2d59'
<snip>
```

So it does a bunch of setup, imports that last block then checks a bunch of things including that the `justified_checkpoint_root` is `0x267b47b08d6fa978d84e652e402d0c0784d6dcdff664f49680b83441c287e866`.
Then without any further blocks, attestations or ticks it checks that the `justified_checkpoint_root` has changed to `0xad558a267ddc732b65a7f749fc5c28c59304742445ae23642156a116eb2b2d59`.

Have I missed something or is the test broken?

djrtwo:
Found the culprit!

The second check (which is extraneous anyway) is yielding the `justified_checkpoint.hash_tree_root()` rather than `justified_checkpoint.root`. Note the former is calculating the htr of the Checkpoint object rather than just giving the root contained within the object

hwwhww:
Closing via #2272
Thanks again @ajsutton @djrtwo

PR #2272, opened 2021-03-24
patch test_filtered_block_tree test generator

djrtwo:
address #2269 

note, we could just yield with extraneous `checks` because there is no update but it also makes it a bit clearer to test writers what is being checked after adding all the blocks.

thanks  @ajsutton!


hwwhww (approved):
Ouch, that was my dumb typo 🙈

Thank you @ajsutton @djrtwo!

PR #2276, opened 2021-03-26
Modify incentives to preserve invariant

barnabemonnot:
### Issue

In the alpha.2 version of Altair, there was a small disconnect between the realised incentives and the expected ones. Proposer rewards made up a little less than 1/8 of the total base reward per epoch, and so did sync committee inclusion rewards, while for both 1/8 was the target.

### Solution

- We target source + target + head rewards = 3 / 4 = 48 / 64 of the base reward, i.e., source = 12 / 64 of BR, target = 24 / 64 of BR and head = 12 / 64.
- We also target sync rewards (for sync committee members, _not_ proposers who include sync committees in their blocks) to be 1/8 of BR, or 8 / 64.
- Finally, we still target proposers to receive 1/8 of the BR (= 8 / 64).

### Changes done

1. Set weights:

| Name | Value |
| - | - |
| `TIMELY_HEAD_WEIGHT` | `12` |
| `TIMELY_SOURCE_WEIGHT` | `12` |
| `TIMELY_TARGET_WEIGHT` | `24` |
| `SYNC_REWARD_WEIGHT` | `8` |
| `PROPOSER_WEIGHT` | `8` |
| `WEIGHT_DENOMINATOR` | `64` |

2. Remove `-proposer_reward` factor in [`process_sync_committee`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/altair/beacon-chain.md#sync-committee-processing) for the `inclusion_reward` given to includee.
3. In `process_attestation`, divide `proposer_reward_numerator` by `WEIGHT_DENOMINATOR - PROPOSER_WEIGHT`, _not_ `WEIGHT_DENOMINATOR`
4. In `process_sync_committee`, the `WEIGHT_DENOMINATOR` is useful to compute exactly the share of BR per epoch attributed to sync committee signers (_not_ includers). But we renormalise for the proposer by multiplying by `WEIGHT_DENOMINATOR / (WEIGHT_DENOMINATOR - PROPOSER_WEIGHT)`. (Someone should check that it doesn't overflow!)

### Not done, but deserves attention

5. In `altair/validator.md`, `signing_root = compute_signing_root(get_block_root_at_slot(state, state.slot), domain)` will throw an error, since `get_block_root_at_slot(state, slot)` expects `slot < state.slot`
6. Also in `altair/validator.md`, modify example in [Aggregation bits](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/altair/validator.md#aggregate-bits) to get the correct subcommittee index.


tests/core/pyspec/eth2spec/test/altair/block_processing/test_process_sync_committee.py
```diff
@@ -138,7 +137,8 @@ def compute_sync_committee_proposer_reward(spec, state, committee, committee_bit
         inclusion_reward = compute_sync_committee_inclusion_reward(
             spec, state, index, committee, committee_bits,
         )
-        proposer_reward += spec.Gwei(inclusion_reward // spec.PROPOSER_REWARD_QUOTIENT)
+        proposer_reward_denominator = (spec.WEIGHT_DENOMINATOR - spec.PROPOSER_WEIGHT) * spec.WEIGHT_DENOMINATOR // spec.PROPOSER_WEIGHT
```
djrtwo, line 140:
```suggestion
        proposer_reward_denominator = (
            (spec.WEIGHT_DENOMINATOR - spec.PROPOSER_WEIGHT)
            * spec.WEIGHT_DENOMINATOR
            // spec.PROPOSER_WEIGHT
        )
```

ralexstokes:
just noting:

the fixes mentioned under

> Not done, but deserves attention

above are addressed in #2278 and #2279 -- thank you for noticing!!

ajsutton:
Given clients are already implementing Altair, it would be good to track these (for want of a better term) "late" changes and make sure that we definitely have reference tests covering them.  It's almost certain that some of them will wind up being missed by client teams.

The PRs themselves provide a record someone could trawl through to create the right tests but not sure if it makes sense to have a more specific list somewhere of tests we should ensure we have.  I also think this particular case probably breaks so much stuff any reference test will likely find it but I know Teku has already implemented this part and will need to update (cc @mbaxter) and could easily miss it, so thought I'd mention the idea.

specs/altair/beacon-chain.md
```diff
@@ -550,9 +552,10 @@ def process_sync_committee(state: BeaconState, aggregate: SyncAggregate) -> None
     for included_index in included_indices:
         effective_balance = state.validators[included_index].effective_balance
         inclusion_reward = Gwei(max_slot_rewards * effective_balance // committee_effective_balance)
-        proposer_reward = Gwei(inclusion_reward // PROPOSER_REWARD_QUOTIENT)
+        increase_balance(state, included_index, inclusion_reward)
+        proposer_reward_denominator = (WEIGHT_DENOMINATOR - PROPOSER_WEIGHT) * WEIGHT_DENOMINATOR // PROPOSER_WEIGHT
+        proposer_reward = Gwei(inclusion_reward * WEIGHT_DENOMINATOR // proposer_reward_denominator)
```
djrtwo, line 557: Why is there `WEIGHT_DENOMINATOR` in this numerator?

specs/altair/beacon-chain.md
```diff
@@ -550,9 +552,10 @@ def process_sync_committee(state: BeaconState, aggregate: SyncAggregate) -> None
     for included_index in included_indices:
         effective_balance = state.validators[included_index].effective_balance
         inclusion_reward = Gwei(max_slot_rewards * effective_balance // committee_effective_balance)
-        proposer_reward = Gwei(inclusion_reward // PROPOSER_REWARD_QUOTIENT)
+        increase_balance(state, included_index, inclusion_reward)
+        proposer_reward_denominator = (WEIGHT_DENOMINATOR - PROPOSER_WEIGHT) * WEIGHT_DENOMINATOR // PROPOSER_WEIGHT
```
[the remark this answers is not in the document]
barnabemonnot: Would be fine to cancel it, yes

specs/altair/beacon-chain.md
```diff
@@ -88,6 +88,7 @@ Altair is the first beacon chain hard fork. Its main features are:
 | `TIMELY_SOURCE_WEIGHT` | `uint64(12)` |
 | `TIMELY_TARGET_WEIGHT` | `uint64(24)` |
 | `SYNC_REWARD_WEIGHT` | `uint64(8)` |
+| `PROPOSER_WEIGHT` | `uint64(8)` |
 | `WEIGHT_DENOMINATOR` | `uint64(64)` |
 
 *Note*: The sum of the weight fractions (7/8) plus the proposer inclusion fraction (1/8) equals 1.
```
hwwhww, line 94:
```suggestion
*Note*: The sum of the weight fractions equals 1.
```

JustinDrake:
Please don't merge just yet. Vitalik found a bug and there are further cleanups to do. Will try to get it done tomorrow :)

JustinDrake:
I think I fixed two bugs:

1) The participant reward (previously known as "inclusion reward") should not be proportional to `effective_balance` because a sync committee member is already chosen with probability proportional to the effective balance. Credit to @vbuterin.
2) The participant reward (previously known as "inclusion reward") was too large by a factor of `len(participant_indices)`, not taking into account the `for` loop.

Fixing these two bugs actually makes the code significantly simpler and cleaner :)

@barnabemonnot: Could you check the logic? :) If confirmed happy to update the tests.

barnabemonnot:
@JustinDrake With the two changes made, when I simulate the ideal case (everyone validating always on time), the rewards are the same as before. As discussed, `max_slot_rewards` previously expressed the total rewards that could be given to validators in sync committees during one slot. A validator would get it share of the total reward per slot based on its share of the effective balance in the sync committee, so the `for` loop sum always added up to `max_slot_rewards` and there was no bug (as in, the total rewards given out over an epoch would never be above the intended share of rewards going to sync committee participants, i.e., 8/64).

There is a difference with the two changes made however. In the previous version, the total reward that sync committee validators could earn in each slot was proportional to the size of the committee in that slot (due to the `len(included_indices) // len(committee_indices)` factor), in line with other rewards that are proportional to validator participation. This seems to no longer be the case, but I am unclear whether it is intentional. All validators also receive the same share `max_participant_rewards // SYNC_COMMITTEE_SIZE`, regardless of their effective balance, which fixes what I understand as the bug noticed by @vbuterin, i.e., that the expected reward of a validator with `effective_balance` should be proportional to `effective_balance`, and not `effective_balance ** 2` as it was previously with the extra `effective_balance // committee_effective_balance` term.

If we want to keep adjusting the total level of rewards to the participation in the committee, it is possible to do change 1. without doing change 2., i.e., multiply `participant_reward` by `len(included_indices) // len(committee_indices)` in the last commit you pushed.

djrtwo (approved):

hwwhww (commented):
LGTM!

A minor post-merge comment.

Issue #2273, opened 2021-03-25
get_sync_committee_signature is ill-defined


michaelsproul:
In the v1.1.0-alpha.2 validator spec, the function `get_sync_committee_signature` is ill-defined because `get_block_root_at_slot(state, state.slot)` will always panic.

```python
def get_sync_committee_signature(state: BeaconState, 
                                 validator_index: ValidatorIndex, 
                                 privkey: int) -> SyncCommitteeSignature:
    epoch = get_current_epoch(state)
    domain = get_domain(state, DOMAIN_SYNC_COMMITTEE, epoch)
    signing_root = compute_signing_root(get_block_root_at_slot(state, state.slot), domain)
    signature = bls.Sign(privkey, signing_root)

    return SyncCommitteeSignature(slot=state.slot, validator_index=validator_index, signature=signature)
```

The head block root should probably be passed in as an argument instead.

michaelsproul:
Fixed by #2277

PR #2277, opened 2021-03-26
Fix use of block root in Altair validator guide

ralexstokes:
Addresses #2273.

The existing definition was relying on a property of the `BeaconState` that is not satisfied until the _next_ slot due to the "delayed" caching of block and state roots in the state transition definition. This results in a bug in the validator guide.

This PR fixes the bug by simply requiring the validator to supply the head block root as a parameter. This fix is sound as a validator can fetch the head block root from the standard eth2 beacon APIs as soon as the head block is determined (which is a precondition for this part of the validator's responsibilities). Clients are of course free to optimize further where possible.

Thanks @michaelsproul and @barnabemonnot for pointing it out!


hwwhww (approved):
lgtm

djrtwo (approved):

PR #2282, opened 2021-03-27
Make if condition more precise

benjaminion:
In `process_attestation()` we have:

```
    assert data.slot + MIN_ATTESTATION_INCLUSION_DELAY <= state.slot <= data.slot + SLOTS_PER_EPOCH
...
    if is_matching_head and is_matching_target and state.slot <= data.slot + MIN_ATTESTATION_INCLUSION_DELAY:
        participation_flag_indices.append(TIMELY_HEAD_FLAG_INDEX)
```

Due to the `assert`, the body of the `if` can be reached only under the tighter condition `state.slot == data.slot + MIN_ATTESTATION_INCLUSION_DELAY` (equality).

The condition as it is currently stated is confusing as it in effect says "this is ok if we are _below_ the _minimum_ inclusion delay", which reads like a contradiction.

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -461,7 +461,7 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
 
     # Participation flag indices
     participation_flag_indices = []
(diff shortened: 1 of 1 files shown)


djrtwo (approved):
right! can't get on chain before min attestation delay

thanks!

PR #2287, opened 2021-03-29
Set tmp ALTAIR_FORK_SLOT to max value

mbaxter:
### Description
Suggesting we set the temporary `ALTAIR_FORK_SLOT` to its max value to lower the risk of accidentally enabling altair early.

### DIFF
--- a/configs/mainnet/altair.yaml
+++ b/configs/mainnet/altair.yaml
@@ -37,7 +37,7 @@ DOMAIN_SYNC_COMMITTEE: 0x07000000
 # ---------------------------------------------------------------
 ALTAIR_FORK_VERSION: 0x01000000
 # TBD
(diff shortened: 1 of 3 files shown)


djrtwo (commented):
not a bad idea... 

would you mind adding this value in the altair/fork.md doc as well?

mbaxter:
> would you mind adding this value in the altair/fork.md doc as well?

Sure.  I was also wondering why we're defining the transition in terms of slots rather than epochs?

djrtwo:
@mbaxter tbh, there probably isn't a super sound reason here. Epoch does seem a bit more natural. I'll bring it up and see if there is a better reason for slot

djrtwo (approved):

PR #2254, opened 2021-03-17
Modularize Phase1 into post-Merge features: Sharding, Custody Game, Data Availability Sampling

protolambda:
This takes the older phase1 PRs #2146 and #2172 and rebases it onto dev (including the new Merge)

Progress:
- [x] Rebase DAS PR by @vbuterin onto latest The-Merge work
- [x] Rebase out clutter (generated files and such) that got into the old DAS PR
- [x] Clean up old shard code, remove ShardTransition accounting in favor of independent shard blobs.
- [x] Remove legacy light client design in favor of new Altair design.
- [x] Split custody game, DAS, and base sharding docs
- [x] Remove old phase1 pyspec building
- [x] Refactor custody game logic out of sharding, implement as extension instead.
- ~~Update block processing and epoch processing to call Altair and Merge-fork processing functions, not just phase0~~
  - blocked by merge work, ~~waiting for that branch to adopt Altair changes.~~ rebase of Merge in Altair in future PRs, use phase0 for now.
- [x] Update BeaconState type in sharding spec to support Merge.
- [ ] Update DAS spec to implement discv5.1 based long-range sample method (for Future PR maybe?)
- [x] Clean up old phase1 references in testing
- [x] Retain proof of custody testing (no proof of custody executable spec, tests are being skipped for now)
- [x] Ratain start-shard unit tests (no sharding executable spec, tests are being skipped for now)
- [x] Drop shard-transition tests in favor of new data-availability focused shards (no strict transitions, just data linked into beacon chain)
- [x] Update block processing and epoch processing helpers with missing merge/sharding/proof-of-custody methods, dropping shard-transition related processing.
- [x] Add phase naming for `SHARDING`, `CUSTODY_GAME`, `DAS` to tag tests with (to be enabled later). Note: custody tests need more in-depth updates, as the sharding design changed.
- [x] Update Makefile to drop older phase1 work
- [x] Drop old phase1 configs, split into new feature configs.
- [x] Update README

The bigger picture of this PR is to accelerate sharding after the merge, and enable early testnets to have shard-data, while DAS and proof of custody continue to be developed. Features, not phases.

PR #2289, opened 2021-03-30
Minor docs fix for #2254

hwwhww:
Minor docs fix for #2254.
No substantive changes.


protolambda (approved):
Thanks, LGTM.

`CONFIG_NAME` is ok for now, but I excluded it on purpose, since we were discussing removing the duplicate `CONFIG_NAME` altogether in the R&D discord a while ago. That way you can combine the configs without overlapping variables.

See https://discord.com/channels/595666850260713488/710831784991916072/823906028600033351

I'll open an issue for that.

protolambda:
@hwwhww Opened #2290 for the config name question.

Issue #2286, opened 2021-03-29
Optimize Altair p2p validation order


terencechain:
The current Altair p2p validation order is not optimized comparing to phase 0. One example is `selection_proof`.

For `aggregate_and_proof` topic:
1.) Checks `selection_proof` selects validator as an aggregator for the slot…
2.) Checks aggregator’s index is within the committee…
3.) Checks `selection_proof` is valid signature… 

For `sync_committee_contribution_and_proof` topic:
1.) Checks aggregator’s index is within committee…
2.) Checks `selection_proof` Is valid signature…
3.) Checks `selection_proof` selects the validator as an aggregator for the slot…

I think there may be more validation order things to improve as well. Opening this issue to track the overall

djrtwo:
oh great! thank you

I'm aiming to get a pre-release out on friday. We can track here and get a PR up in next couple of days.

ralexstokes:
@terencechain thanks! updated that in #2293. happy to make further changes if we find better orderings in this thread

terencechain:
Thanks @ralexstokes ! Closing this as I dont see any ordering improvement at this moment.

PR #2293, opened 2021-03-30
Reorder Altair gossip validation conditions

ralexstokes:
Addresses #2286.

Changes:

1. Validate the `selection_proof` selects the incoming validator as an
aggregator before doing other checks like committee inclusion and
(relatively) expensive signature checks.


djrtwo (approved):
great!

PR #2295, opened 2021-03-31
New application payload fields

mkalinin:
### What's done

Adds the following fields to the `ApplicationPayload`:
- `parent_hash: Bytes32`
- `number: uint64`

Adds `application_block_number` to the `BeaconState` and incorporates following rules that ensures the validity of the application block tree:
```python
assert body.application_payload.parent_hash == state.application_block_hash
assert body.application_payload.number == state.application_block_number + 1
```

It would also make sense to add `gas_limit/base_fee` to the state with corresponding validity rules once EIP-1559 is shipped.

### Rationale
The main reason of doing this is making structures and code forward compatible with stateless execution which implies that there is no application block tree that is maintained by the application-layer. It results in a following requirements on consensus-layer:
- provide all data required to re-assemble a block to the application-layer
- check validity conditions of the application block tree

A byproduct of adding `parent_hash` is the following simplification in the transition process. Instead of submitting an empty `ApplicationPayload` with `block_hash = last_pow_block.hash` to signify the transition block, proposer becomes eligible to propose full `ApplicationPayload` with `parent_hash` referring to the last PoW block. Aside of simplification it also have a small incentivising effect as proposer of the transition block gets transaction fees now.

The cost of this change is additional `40 bytes` in `BeaconBlockBody` structure.

### DIFF
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -81,13 +81,14 @@ class BeaconBlockBody(phase0.BeaconBlockBody):
 
 #### `BeaconState`
 
-*Note*: `BeaconState` fields remain unchanged other than addition of `application_state_root` and `application_block_hash`.
+*Note*: `BeaconState` fields remain unchanged other than addition of application payload fields.
 
 ```python
 class BeaconState(phase0.BeaconState):
     # Application-layer
     application_state_root: Bytes32  # [New in Merge]
     application_block_hash: Bytes32  # [New in Merge]
+    application_block_number: uint64 # [New in Merge]
 ```
 
 ### New containers
@@ -99,8 +100,10 @@ The application payload included in a `BeaconBlockBody`.
 ```python
 class ApplicationPayload(Container):
     block_hash: Bytes32  # Hash of application block
+    parent_hash: Bytes32
     coinbase: Bytes20
     state_root: Bytes32
+    number: uint64
     gas_limit: uint64
     gas_used: uint64
     receipt_root: Bytes32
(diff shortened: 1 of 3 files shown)


specs/merge/beacon-chain.md
```diff
@@ -81,13 +81,14 @@ class BeaconBlockBody(phase0.BeaconBlockBody):
...
 
-*Note*: `BeaconState` fields remain unchanged other than addition of `application_state_root` and `application_block_hash`.
+*Note*: `BeaconState` fields remain unchanged other than addition of application payload fields.
 
 ```python
 class BeaconState(phase0.BeaconState):
     # Application-layer
     application_state_root: Bytes32  # [New in Merge]
     application_block_hash: Bytes32  # [New in Merge]
+    application_block_number: uint64 # [New in Merge]
```
terencechain, line 91: Would it better to define an `ApplicationState` container and bundle these in?

```python
class ApplicationState(Container): 
    root: Bytes32 
    block_hash: Bytes32
    block_number: uint64


class BeaconState(phase0.BeaconState):
    application_state: ApplicationState
```

Especially if we are going to add eip1559 fields later on

l2ichshop2 (approved):

specs/merge/beacon-chain.md
mkalinin: I was thinking about this as well, but this is rather an `ApplicationBlockHeader` than `ApplicationState`. It would probably make sense to include all fields except for `transactions` into beacon state.

```python
class ApplicationBlockHeader(Container):
    block_hash: Bytes32
    parent_hash: Bytes32
    coinbase: Bytes20
    state_root: Bytes32
    number: uint64
    gas_limit: uint64
    gas_used: uint64
    receipt_root: Bytes32


class BeaconState(phase0.BeaconState)
    latest_application_block_header: ApplicationBlockHeader  # Akin to `latest_block_header`
```

djrtwo:
@mkalinin are you considering trying the app-block-header in this PR or do you want it reviewed as is?

mkalinin:
> @mkalinin are you considering trying the app-block-header in this PR or do you want it reviewed as is?

Let me try and see how it will look like

specs/merge/beacon-chain.md
```diff
@@ -111,6 +110,25 @@ class ApplicationPayload(Container):
...
+class ApplicationBlockHeader(Container):
+    block_hash: Bytes32  # Hash of application block
+    parent_hash: Bytes32
+    coinbase: Bytes20
+    state_root: Bytes32
+    number: uint64
+    gas_limit: uint64
+    gas_used: uint64
+    receipt_root: Bytes32
+    logs_bloom: ByteVector[BYTES_PER_LOGS_BLOOM]
```
djrtwo, line 129: I suggest we add `transactions_root` so that this merklizes to the same root of `ApplicationPayload`

We can then rename this to `ApplicationPayloadHeader` or rename the payload to `ApplicationBlock`.

(side note, we mgiht be renaming these all to use `Execution` prefix but thats a separate PR

djrtwo (commented):
Looks good! I'd mainly suggest embedding `transactions_root` into the `header` to allow for sharing htr across the two objects

mkalinin:
> Looks good! I'd mainly suggest embedding `transactions_root` into the `header` to allow for sharing htr across the two objects

@djrtwo Implemented `transactions_root` as per suggested 👍

djrtwo (approved):
great work!

PR #2296, opened 2021-04-01
Fix missed rename from aggregate_and_proof to contribution_and_proof

ajsutton:
Trivial fix in the validation instructions for signed contribution and proof.  It's referring to `aggregate_and_proof` which I suspect is a missed rename and should be `contribution_and_proof`.


djrtwo (approved):
glad someone reads this thing... thank you!

Issue #2297, opened 2021-04-01
sync_committee_contribution_and_proof validators selection_proof signature incorrectly


ajsutton:
In https://github.com/ethereum/eth2.0-specs/blob/dev/specs/altair/p2p-interface.md#sync_committee_contribution_and_proof there's the rule:

```
The contribution_and_proof.selection_proof is a valid signature of the contribution.slot by the validator with index contribution_and_proof.aggregator_index
```

But https://github.com/ethereum/eth2.0-specs/blob/dev/specs/altair/validator.md#aggregation-selection looks like the signature should actually be over `SyncCommitteeSigningData(contribution.slot, contribution.subcommittee_index)`.

Have I mismatched where this signature comes from or did we copy/paste a rule from attestation aggregations and not adjust? :)

ralexstokes:
Ah, thanks for catching this! Yeah, the signing data was changed at the last minute to help w/ the aggregation scheme and it led to some rushed edits so that it could be included in the pre-release.

I'll update the language in a PR later today.

ralexstokes:
Closing w/ #2299, feel free to re-open or raise a new issue if there are further things to clarify -- thanks!

PR #2299, opened 2021-04-03
Fix language with sync committee selection proofs

ralexstokes:
Addresses #2297.


ralexstokes:
@ajsutton here is an attempt; happy to hear any suggestions you may have on the naming

ajsutton (approved):
Yep that looks good to me.

hwwhww (approved):

PR #2301, opened 2021-04-04
Fix bouncing attack tests

### LINKED ISSUE #2298: Some block processing unit tests force processing of inconsistent blocks
`test_on_block_outside_safe_slots_and_multiple_better_justified` and `test_on_block_outside_safe_slots_but_finality` in `test/phase0/unittests/fork_choice/test_on_block.py` force the processing of inconsistent blocks. 

## Problem:
Context:
```python
    # Mock justified block in store
    just_block = build_empty_block_for_next_slot(spec, state)
    # Slot is same as justified checkpoint so does not trigger an override in the store
    just_block.slot = spec.compute_start_slot_at_epoch(store.justified_checkpoint.epoch)
    store.blocks[just_block.hash_tree_root()] = just_block
```
The inconsistency is caused when `just_block.slot` is being set - the new slot is earlier than `just_block`'s parent block. 

Here[[1](https://github.com/ethereum/eth2.0-specs/blame/9eb662786adebce7c3e41852a525458895e8bedc/tests/core/pyspec/eth2spec/test/phase0/unittests/fork_choice/test_on_block.py#L237-L238), [2](https://github.com/ethereum/eth2.0-specs/blame/9eb662786adebce7c3e41852a525458895e8bedc/tests/core/pyspec/eth2spec/test/phase0/unittests/fork_choice/test_on_block.py#L290-L291)] is an illustration of the problem:
```python
    just_block_parent = store.blocks[just_block.parent_root]
    assert just_block_parent.slot < just_block.slot, f"just_block_parent.slot: {just_block_parent.slot}, just_block.slot: {just_block.slot}"
```

Both unit tests now fail with this output:
```
AssertionError: just_block_parent.slot: 16, just_block.slot: 0
```

This is fine in the current spec, because only the block's slot value matters for these unit tests. However, it causes a problem for a planned fork choice upgrade where constructing the block tree becomes relevant for this unit test.

## Suggested Fix:
Create consistent block hierarchy for `just_block`.

adiasg:
Fixes #2298

### DIFF
--- a/tests/core/pyspec/eth2spec/test/phase0/unittests/fork_choice/test_on_block.py
+++ b/tests/core/pyspec/eth2spec/test/phase0/unittests/fork_choice/test_on_block.py
@@ -273,7 +273,7 @@ def test_on_block_outside_safe_slots_and_multiple_better_justified(spec, state):
 def test_on_block_outside_safe_slots_but_finality(spec, state):
     # Initialization
     store = get_genesis_forkchoice_store(spec, state)
-    time = 0
+    time = 100
     spec.on_tick(store, time)
 
     next_epoch(spec, state)
--- a/tests/core/pyspec/eth2spec/test/phase0/unittests/fork_choice/test_on_block.py
+++ b/tests/core/pyspec/eth2spec/test/phase0/unittests/fork_choice/test_on_block.py
@@ -239,9 +239,6 @@ def test_on_block_outside_safe_slots_and_multiple_better_justified(spec, state):
     just_block = build_empty_block_for_next_slot(spec, state)
     store.blocks[just_block.hash_tree_root()] = just_block
 
-    just_block_parent = store.blocks[just_block.parent_root]
-    assert just_block_parent.slot < just_block.slot, f"just_block_parent.slot: {just_block_parent.slot}, just_block.slot: {just_block.slot}"
-
     # Step time past safe slots
     spec.on_tick(store, store.time + spec.SAFE_SLOTS_TO_UPDATE_JUSTIFIED * spec.SECONDS_PER_SLOT)
     assert spec.get_current_slot(store) % spec.SLOTS_PER_EPOCH >= spec.SAFE_SLOTS_TO_UPDATE_JUSTIFIED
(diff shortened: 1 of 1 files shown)


tests/core/pyspec/eth2spec/test/phase0/unittests/fork_choice/test_on_block.py
```diff
@@ -274,14 +279,19 @@ def test_on_block_outside_safe_slots_but_finality(spec, state):
     next_epoch(spec, state)
     spec.on_tick(store, store.time + state.slot * spec.SECONDS_PER_SLOT)
     state, store, last_signed_block = apply_next_epoch_with_attestations(spec, state, store)
+    last_block_root = hash_tree_root(last_signed_block.message)
+
+    # Mock fictitious justified checkpoint in store
+    store.justified_checkpoint = spec.Checkpoint(
+        epoch=spec.compute_epoch_at_slot(last_signed_block.message.slot),
+        root=spec.Root("0x4a55535449464945440000000000000000000000000000000000000000000000")
```
djrtwo, line 287: why not use htr of `last_block_root` for `justified_checkpoint.root`?

djrtwo (commented):
one minor question

tests/core/pyspec/eth2spec/test/phase0/unittests/fork_choice/test_on_block.py
adiasg: We want to trigger this piece of code from `should_update_justified_checkpoint`:
```python
if not get_ancestor(store, new_justified_checkpoint.root, justified_slot) == store.justified_checkpoint.root:
        return False
```

If we use `last_block_root` & `spec.compute_epoch_at_slot(last_signed_block.message.slot)` to construct `store.justified_checkpoint`, then the newly constructed blocks will not trigger this, and `should_update_justified_checkpoint` returns `True` because the new justified checkpoint is a descendant of the current one in the store.

adiasg: Filling in a fictitious root ensures that this `get_ancestor` check fails and `should_update_justified_checkpoint` returns `False`, creating the test scenario we want

djrtwo (approved):

PR #2302, opened 2021-04-04
Shard blob, headers and slashings

protolambda:
- Make header and blob consistent:
  - Embedding the `data_root` in the header makes the header SSZ-equal to the full blob
  - The signature approach becomes consistent with phase0, more straight-forward
  - Rename `ShardHeader` to `ShardBlobHeader`, also for consistency with beacon blocks and headers.
- Move the commitments into the "body":
  - An SSZ summary can be used to describe the header with partial body
  - Slashings can use an even more minimal form of header, essentially just a reference, to save on space. This enables a higher maximum shard slashing count per block.
  - This seems optimal, but naming can definitely still improve
- Add proposer index to the shard header:
  - Consistent with phase0
  - Cleans up slashing requirements: the slashing is independent of the beacon state (incl. deep forking) for verifying it.
  - Can be used for early signature verification in p2p (proposer index computation depends on shuffling, which may be computed afterwards)
- Update sharding configs:
  - Fix shard proposer domain naming, make it consistent with phase0
- Add/update gossip validation conditions for shard blobs
- New shard header gossip validation
- Implement shard blob slashings
  - Shard blob slashing type
  - New beacon block operations list, max-slashings config var
  - Gossip topic and validation conditions


specs/sharding/p2p-interface.md
```diff
@@ -66,29 +76,57 @@ Following the same scheme as the [Phase0 gossip topics](../phase0/p2p-interface.
...
+The following validations MUST pass before forwarding the `signed_blob` (with inner `message` as `blob`) on the horizontal subnet or creating samples for it.
 - _[REJECT]_ `blob.shard` MUST match the topic `{shard}` parameter. (And thus within valid shard index range)
 - _[IGNORE]_ The `blob` is not from a future slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) --
   i.e. validate that `blob.slot <= current_slot`
   (a client MAY queue future blobs for processing at the appropriate slot).
-- _[IGNORE]_ The blob is the first blob with valid signature received for the proposer for the `(slot, shard)` combination.
+- _[IGNORE]_ The `blob` is new enough to be still be processed --
+  i.e. validate that `compute_epoch_at_slot(blob.slot) >= get_previous_epoch(state)`
+- _[IGNORE]_ The blob is the first blob with valid signature received for the `(header.proposer_index, header.slot, header.shard)` combination.
 - _[REJECT]_ As already limited by the SSZ list-limit, it is important the blob is well-formatted and not too large.
```
protolambda, line 95: We could verify if the commitment and degree proof are valid for the contained data, but it wouldn't really matter much if it's not: double proposals are slashable, and the proposal is just wasted. So it could be an application-only thing, and skipped during network validation. This verification is faster than computing the commitment from the data, so it may be ok to still verify it anyway?

specs/sharding/p2p-interface.md
```diff
@@ -29,30 +31,38 @@ The adjustments and additions for Shards are outlined in this document.
...
-    # The actual data. Represented in header as data commitment and degree proof
-    data: List[BLSPoint, POINTS_PER_SAMPLE * MAX_SAMPLES_PER_BLOCK]
+    body: ShardBlobBody
+    # Proposer of the shard-blob
+    proposer_index: ValidatorIndex
 ```
 
-Note that the hash-tree-root of the `ShardBlob` does not match the `ShardHeader`,
-since the blob deals with full data, whereas the header includes the KZG commitment and degree proof instead.
+The user MUST always verify the `body` matches the commitments in the `commitment` and `degree_proof`.
```
protolambda, line 58: This is worded poorly, inconsistent after latest type changes. Meant to require consistency between commitments and data in `body`

configs/mainnet/sharding.yaml
```diff
@@ -41,5 +43,5 @@ SHARD_COMMITTEE_PERIOD: 256
 
 # Signature domains
 # ---------------------------------------------------------------
-DOMAIN_SHARD_PROPOSAL: 0x80000000
+DOMAIN_SHARD_PROPOSER: 0x80000000
```
djrtwo, line 46: nice catch

specs/sharding/p2p-interface.md
protolambda: Fixed it, and moved the note to the body definition.

specs/sharding/beacon-chain.md
```diff
@@ -94,6 +99,7 @@ The following values are (non-configurable) constants used throughout the specif
 | `INITIAL_ACTIVE_SHARDS` | `uint64(2**6)` (= 64) | Initial shard count |
 | `GASPRICE_ADJUSTMENT_COEFFICIENT` | `uint64(2**3)` (= 8) | Gasprice may decrease/increase by at most exp(1 / this value) *per epoch* |
 | `MAX_SHARD_HEADERS_PER_SHARD` | `4` | |
+| `MAX_SHARD_PROPOSER_SLASHINGS` | `2**6` (= 64) | Maximum amount of shard proposer slashing operations per block |
```
djrtwo, line 102: Interesting... to account for one slashing per shard per slot.

I think I'd argue that the complexity of verification and size of this is similiar to beacon block and so should maybe be 16 instead. Obviously you have issues if all shards are slashable all the time, but in expectation you will be able to catch up with 16 over time to some extent.

Numbers aren't amazing on the catchup in the extreme though

protolambda:
Edit again: undoing accidental paste of some altair var symbol in one of the docs :sweat:

specs/sharding/beacon-chain.md
```diff
@@ -226,6 +242,35 @@ class PendingShardHeader(Container):
...
+    # Hash-tree-root of commitment data
+    body_root: Root
+    # Proposer of the shard-blob
+    proposer_index: ValidatorIndex
+```
+
+### `SignedShardBlobReference`
+
+```python
+class SignedShardBlobHeader(Container):
```
djrtwo, line 261:
```suggestion
class SignedShardBlobReference(Container):
```

specs/sharding/beacon-chain.md
```diff
@@ -226,6 +242,35 @@ class PendingShardHeader(Container):
...
+    message: ShardBlobReference
+    signature: BLSSignature
+```
+
+### `ShardProposerSlashing`
+
+```python
+class ShardProposerSlashing(Container):
+    signed_header_1: SignedShardBlobReference
+    signed_header_2: SignedShardBlobReference
```
djrtwo, line 271:
```suggestion
    signed_reference_1: SignedShardBlobReference
    signed_reference_2: SignedShardBlobReference
```

specs/sharding/beacon-chain.md
```diff
@@ -499,30 +546,36 @@ def update_pending_votes(state: BeaconState, attestation: Attestation) -> None:
...
+    assert header_epoch in [get_previous_epoch(state), get_current_epoch(state)]
+    # Verify that the shard is active
+    assert header.shard < get_active_shard_count(state, header_epoch)
+
+    # Verify proposer
+    assert header.proposer_index == get_shard_proposer_index(state, header.slot, header.shard)
     # Verify signature
-    signer_index = get_shard_proposer_index(state, header.slot, header.shard)
     signing_root = compute_signing_root(header, get_domain(state, DOMAIN_SHARD_HEADER))
     assert bls.Verify(state.validators[signer_index].pubkey, signing_root, signed_header.signature)
```
djrtwo, line 561: `signer_index` is undefined
```suggestion
    assert bls.Verify(state.validators[header.proposer_index].pubkey, signing_root, signed_header.signature)
```

specs/sharding/beacon-chain.md
```diff
@@ -186,26 +193,35 @@ class DataCommitment(Container):
...
 
-### `ShardHeader`
+### `ShardBlobBodySummary`
 
 ```python
-class ShardHeader(Container):
-    # Slot and shard that this header is intended for
-    slot: Slot
-    shard: Shard
+class ShardBlobBodySummary(Container):
```
djrtwo, line 199: Maybe DataSummary?

specs/sharding/beacon-chain.md
```diff
@@ -544,6 +597,33 @@ the length proof is the commitment to the polynomial `B(X) * X**(MAX_DEGREE + 1
 where `MAX_DEGREE` is the maximum power of `s` available in the setup, which is `MAX_DEGREE = len(G2_SETUP) - 1`.
 The goal is to ensure that a proof can only be constructed if `deg(B) < l` (there are not hidden higher-order terms in the polynomial, which would thwart reconstruction).
 
+##### Shard Proposer slashings
+
+```python
+def process_shard_proposer_slashing(state: BeaconState, proposer_slashing: ShardProposerSlashing) -> None:
+    header_1 = proposer_slashing.signed_header_1.message
```
djrtwo, line 604: These are "references" not "headers" in the current terminology. I suggest var names be consistent with current type names even though that will likely change

djrtwo: e.g. `reference_1`, `reference_2`

specs/sharding/p2p-interface.md
```diff
@@ -29,30 +31,39 @@ The adjustments and additions for Shards are outlined in this document.
 
 ## New containers
 
-### ShardBlob
+### ShardBlobBody
+
+```python
+class ShardBlobBody(Container):
```
djrtwo, line 37: I lean toward moving these into the consensus spec, but I see the argument for not

specs/sharding/p2p-interface.md
```diff
@@ -29,30 +31,39 @@ The adjustments and additions for Shards are outlined in this document.
...
+
+The user MUST always verify the commitments in the `body` are valid for the `data` in the `body`.
 
-Network-only.
+### ShardBlob
 
 ```python
 class ShardBlob(Container):
-    # Slot and shard that this blob is intended for
+    # Slot and shard that this header is intended for
```
djrtwo, line 52: Accidental find and replace? This is a `Blob` not a header

specs/sharding/p2p-interface.md
djrtwo: ehh. maybe not

specs/sharding/p2p-interface.md
```diff
@@ -66,29 +77,57 @@ Following the same scheme as the [Phase0 gossip topics](../phase0/p2p-interface.
...
+- _[IGNORE]_ The `blob` is new enough to be still be processed --
+  i.e. validate that `compute_epoch_at_slot(blob.slot) >= get_previous_epoch(state)`
+- _[IGNORE]_ The blob is the first blob with valid signature received for the `(blob.proposer_index, blob.slot, blob.shard)` combination.
 - _[REJECT]_ As already limited by the SSZ list-limit, it is important the blob is well-formatted and not too large.
 - _[REJECT]_ The `blob.data` MUST NOT contain any point `p >= MODULUS`. Although it is a `uint256`, not the full 256 bit range is valid.
-- _[REJECT]_ The proposer signature, `signed_blob.signature`, is valid with respect to the `proposer_index` pubkey, signed over the SSZ output of `commit_to_data(blob.data)`.
-- _[REJECT]_ The blob is proposed by the expected `proposer_index` for the blob's slot.
+- _[REJECT]_ The proposer signature, `signed_blob.signature`, is valid with respect to the `proposer_index` pubkey.
+- _[REJECT]_ The blob is proposed by the expected `proposer_index` for the blob's slot
+  in the context of the current shuffling (defined by `parent_root`/`slot`).
```
djrtwo, line 100: This might be an issue. The current ShardBlob does not have a beacon chain reference (I think what you are referring to as `parent_root`) in the definition and so there is no expected shuffling to be verified against.

I believe we *must* have this in the shardblob payload to properly do verifications

If we add it, we might also add `shard_blob_` gossip validation that the `parent_root` or `beacon_root` reference has been processed locally (with the may IGNORE on that condition instead of this condition here that has it implicit

specs/sharding/p2p-interface.md
```diff
@@ -66,29 +77,57 @@ Following the same scheme as the [Phase0 gossip topics](../phase0/p2p-interface.
...
+  If the `proposer_index` cannot immediately be verified against the expected shuffling,
+  the block MAY be queued for later processing while proposers for the blob's branch are calculated --
+  in such a case _do not_ `REJECT`, instead `IGNORE` this message.
 
-TODO: make double blob proposals slashable?
 
 #### Shard header: `shard_header`
```
djrtwo, line 106: Consider putting "global topics" under it's own sub-header like in altair and beacon specs
https://github.com/ethereum/eth2.0-specs/blob/dev/specs/altair/p2p-interface.md#global-topics

specs/sharding/p2p-interface.md
```diff
@@ -66,29 +77,57 @@ Following the same scheme as the [Phase0 gossip topics](../phase0/p2p-interface.
...
-Shard header data, in the form of a `SignedShardHeader` is published to the global `shard_header` subnet.
+Shard header data, in the form of a `SignedShardBlobHeader` is published to the global `shard_header` subnet.
+
+The following validations MUST pass before forwarding the `signed_shard_header` (with inner `message` as `header`) on the network.
+- _[IGNORE]_ The `header` is not from a future slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) --
+  i.e. validate that `header.slot <= current_slot`
+  (a client MAY queue future headers for processing at the appropriate slot).
+- _[IGNORE]_ The `header` is new enough to be still be processed --
+  i.e. validate that `compute_epoch_at_slot(header.slot) >= get_previous_epoch(state)`
+- _[IGNORE]_ The header is the first header with valid signature received for the `(header.proposer_index, header.slot, header.shard)` combination.
```
djrtwo, line 116: TODO: consider adding a `beacon_root` reference processed check

djrtwo (commented):
awesome! see comments below

specs/sharding/beacon-chain.md
protolambda: I want to make clear it's another variant of the same body type, although you are right that it's the data that is really getting summarized into a `Root`. Let's add this to the other naming changes for a later PR.

specs/sharding/p2p-interface.md
protolambda: It's based on the beacon-block proposer index, forgot about the `parent_root` context. That's a problem indeed, it needs some context for any forking. If we add such context though, shard blobs cannot be replayed anymore on a chain that prunes out some earlier shard blob header inclusions (if we enforce the `parent_root` to be correct on-chain). This may be a good thing though?

specs/sharding/p2p-interface.md
protolambda: The nesting gets too deep for my liking, the topics themselves don't get a unique element-id link anymore that way.

specs/sharding/beacon-chain.md
protolambda: Adjusted the config vars for now. We can revisit this later, there's something to say for more (more slashings are less of a DoS danger, only bad if there's also a lot of forking)

specs/sharding/p2p-interface.md
```diff
@@ -29,30 +31,39 @@ The adjustments and additions for Shards are outlined in this document.
...
+class ShardBlobBody(Container):
+    # The actual data commitment
+    commitment: DataCommitment
+    # Proof that the degree < commitment.length
+    degree_proof: BLSCommitment
+    # The actual data. Should match the commitment and degree proof.
+    data: List[BLSPoint, POINTS_PER_SAMPLE * MAX_SAMPLES_PER_BLOCK]
+```
+
+The user MUST always verify the commitments in the `body` are valid for the `data` in the `body`.
```
terencechain, line 46: Can we clarify what the commitments are? Is it `body.commitment`? if yes, maybe this is more clear:

```suggestion
The user MUST always verify the `commitment` in the `body` is valid for the `data` in the `body`.
```

protolambda: "commitments" here refers to both KZG commitments, the only two in that type. The first one checks the contents, the second one checks the length. I will see if we can add more comments, or explain the commitments elsewhere in the spec.

specs/sharding/p2p-interface.md
protolambda: Possible solution: https://github.com/ethereum/eth2.0-specs/pull/2312

specs/sharding/p2p-interface.md
protolambda: It references the new `beacon_block_root` field now.

specs/sharding/p2p-interface.md
protolambda: Merged that draft into this PR, there's a `beacon_block_root` field now. The key stays `proposer_index, slot, shard`, since multiple proposals of the same proposer in the same slot would be slashable, there must only be one.

djrtwo (approved):
excellent!

PR #2303, opened 2021-04-05
Sharding spec typos fixed

ericsson49:
Fixed some typos, which make either `setup.py/get_spec` or Python parser fail.
Not sure about the `ROOT_OF_UNITY` fix, but others look like straightforward.

### DIFF
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -223,7 +223,7 @@ class PendingShardHeader(Container):
     # Who voted for the header
     votes: Bitlist[MAX_VALIDATORS_PER_COMMITTEE]
     # Has this header been confirmed?
-    confirmed: bool
+    confirmed: boolean
 ```
 
 ## Helper functions
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -108,7 +108,7 @@ The following values are (non-configurable) constants used throughout the specif
 | - | - |
 | `G1_SETUP` | Type `List[G1]`. The G1-side trusted setup `[G, G*s, G*s**2....]`; note that the first point is the generator. |
 | `G2_SETUP` | Type `List[G2]`. The G2-side trusted setup `[G, G*s, G*s**2....]` |
-| `ROOT_OF_UNITY` | `pow(PRIMITIVE_ROOT_OF_UNITY, (MODULUS - 1) // (MAX_SAMPLES_PER_BLOCK * POINTS_PER_SAMPLE, MODULUS)` |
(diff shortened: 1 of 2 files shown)


specs/sharding/beacon-chain.md
```diff
@@ -108,7 +108,7 @@ The following values are (non-configurable) constants used throughout the specif
 | - | - |
 | `G1_SETUP` | Type `List[G1]`. The G1-side trusted setup `[G, G*s, G*s**2....]`; note that the first point is the generator. |
 | `G2_SETUP` | Type `List[G2]`. The G2-side trusted setup `[G, G*s, G*s**2....]` |
-| `ROOT_OF_UNITY` | `pow(PRIMITIVE_ROOT_OF_UNITY, (MODULUS - 1) // (MAX_SAMPLES_PER_BLOCK * POINTS_PER_SAMPLE, MODULUS)` |
+| `ROOT_OF_UNITY` | `pow(PRIMITIVE_ROOT_OF_UNITY, (MODULUS - 1) // (MAX_SAMPLES_PER_BLOCK * POINTS_PER_SAMPLE), MODULUS)` |
```
ralexstokes, line 111: syntax looks correct -- we want to do modular exponentiation here which is implemented w/ this change

specs/sharding/beacon-chain.md
```diff
@@ -223,7 +223,7 @@ class PendingShardHeader(Container):
     # Who voted for the header
     votes: Bitlist[MAX_VALIDATORS_PER_COMMITTEE]
     # Has this header been confirmed?
-    confirmed: bool
+    confirmed: boolean
```
ralexstokes, line 226: `bool` is the Python type -- is this pulling in an SSZ definition?

ralexstokes (commented):
not sure about the `bool` change but everything else looks good!

specs/sharding/beacon-chain.md
djrtwo: yeah, boolean is a type in `ssz.ssz_typing`

djrtwo (approved):

specs/sharding/beacon-chain.md
ericsson49: Yes. One reason behind the change is that `setup.py` fails, when there is `bool` in data classes.
```
  File "/mnt/c/Users/ericsson/IdeaProjects/eth2.0-specs/setup.py", line 379, in dependency_order_class_objects
    for item in [dep, key] + key_list[key_list.index(dep)+1:]:
ValueError: 'bool' is not in list
```
I assume that it's better to use the SSZ analogue (as it's the case in `Validator::slashed`, for example).

PR #2304, opened 2021-04-05
Fix is_within_weak_subjectivity_period() assertion

farazdagi:
Currently, [is_within_weak_subjectivity_period()](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/weak-subjectivity.md#is_within_weak_subjectivity_period) has the following assertion:
```python
assert ws_state.latest_block_header.state_root == ws_checkpoint.root
```
However, `ws_checkpoint.root` is a block root, so when asserting that provided `ws_state` has the matching root, the assertion should be changed to:
```python
assert hash_tree_root(ws_state.latest_block_header) == ws_checkpoint.root
```

### DIFF
--- a/specs/phase0/weak-subjectivity.md
+++ b/specs/phase0/weak-subjectivity.md
@@ -120,7 +120,7 @@ The check can be implemented in the following way:
 ```python
 def is_within_weak_subjectivity_period(store: Store, ws_state: BeaconState, ws_checkpoint: Checkpoint) -> bool:
     # Clients may choose to validate the input state against the input Weak Subjectivity Checkpoint
-    assert ws_state.latest_block_header.state_root == ws_checkpoint.root
+    assert hash_tree_root(ws_state.latest_block_header) == ws_checkpoint.root
     assert compute_epoch_at_slot(ws_state.slot) == ws_checkpoint.epoch
 
     ws_period = compute_weak_subjectivity_period(ws_state)


ralexstokes (commented):
nice catch! suggests we should add some tests at some point to help us catch these types of errors in the CI...

there is a slight complication given the semantics of the state transition function though which is that we do not write the block root for a given slot into the state until the next slot...

there are a couple directions we could go to patch this up in the WS guide -- one would simply be requiring the `BeaconState` input to this function be a state from this epoch, just _not_ the one corresponding directly to the checkpoint -- then we can verify the checkpoint w/ e.g. `get_block_root`

@adiasg do you have any thoughts/preferences on how to modify this?

adiasg:
We could do:
```python
assert ws_state.latest_block_header.state_root == store.block_states[ws_checkpoint.root]
```

If the block with root `ws_checkpoint.root` has been processed by the client, then `store.block_states[ws_checkpoint.root]` would have been populated already.
If the client is bootstrapping itself from a trusted `ws_state` and `ws_checkpoint` (not the current sync design, but [planned in future upgrades](https://github.com/ethereum/eth2.0-specs/pull/2131)), then we can require that `store.block_states[ws_checkpoint.root]` be populated with `ws_state`, so that it doesn't break this function or other fork choice logic.

specs/phase0/weak-subjectivity.md
```diff
@@ -120,7 +120,7 @@ The check can be implemented in the following way:
 ```python
 def is_within_weak_subjectivity_period(store: Store, ws_state: BeaconState, ws_checkpoint: Checkpoint) -> bool:
     # Clients may choose to validate the input state against the input Weak Subjectivity Checkpoint
-    assert ws_state.latest_block_header.state_root == ws_checkpoint.root
+    assert hash_tree_root(ws_state.latest_block_header) == ws_checkpoint.root
```
adiasg, line 123:
```suggestion
    assert ws_state.latest_block_header.state_root == store.block_states[ws_checkpoint.root]
```

adiasg (approved):

specs/phase0/weak-subjectivity.md
```diff
@@ -120,7 +120,7 @@ The check can be implemented in the following way:
 ```python
 def is_within_weak_subjectivity_period(store: Store, ws_state: BeaconState, ws_checkpoint: Checkpoint) -> bool:
     # Clients may choose to validate the input state against the input Weak Subjectivity Checkpoint
-    assert ws_state.latest_block_header.state_root == ws_checkpoint.root
+    assert ws_state.latest_block_header.state_root == store.block_states[ws_checkpoint.root]
```
djrtwo, line 123: This should be the `hash_tree_root`  of `store.block_states[ws_checkpoint.root]` right?

djrtwo: Also, `latest_block_header.state_root` is often just `Bytes32()` because it avoids a circular dependency

You need to instead do something like the following:
```python
    latest_block_state_root = ws.latest_block_header.state_root
    if latest_block_state_root == Bytes32():
        latest_block_state_root = hash_tree_root(ws_state)
    assert latest_block_state_root == hash_tree_root(store.block_states[ws_checkpoint.root])
```

adiasg: > This should be the `hash_tree_root` of `store.block_states[ws_checkpoint.root]` right?

Yes. My bad.

```suggestion
    assert ws_state.latest_block_header.state_root == hash_tree_root(store.block_states[ws_checkpoint.root])
```

adiasg: > Also, latest_block_header.state_root is often just Bytes32() because it avoids a circular dependency

This breaks the functionality that we seek when bootstrapping from a WS checkpoint & associated state. We want to ensure that the given checkpoint & state are for the same point in the chain. 

A state's `latest_block_header.state_root` is `Bytes32()` if it's a state at the tip of the chain. For purposes of WS bootstrapping, only finalized states are expected to be used. I think it's ok to let states at the tip of the chain fail this method (it's unsafe to proceed anyway).

djrtwo: finalized states also can have `Bytes32()` as the latest header in there. It's just post-states that are not skip slots and such a post state can easily be finalized

adiasg: > finalized states also can have Bytes32() as the latest header in there
Oops, looks like I overlooked this part. In that case, this function is broken & needs fixing.

leolara:

PR #2305, opened 2021-04-05
Fix table formatting

mbaxter:
### Description

It looks like a couple of sections in `altair/p2p-interface.md`  are meant to be tables?


hwwhww (approved):
LGTM 👍

PR #2311, opened 2021-04-06
Add some basic tests for Altair validator guide

ralexstokes:
Check a few high-level invariants from the validator guide.

~Found a minor typing issue when making the code executable, but no substantive changes.~

~Note: only 1 test now but plan to add a couple more tomorrow. Not a draft, but will also follow up w/ more soon~

edit: blocked on incorporating a new version of `remerkleable` with this: https://github.com/ethereum/eth2.0-specs/pull/2311#discussion_r610241135


specs/altair/validator.md
```diff
@@ -229,7 +229,7 @@ def process_sync_committee_contributions(block: BeaconBlock,
         subcommittee_index = contribution.subcommittee_index
         for index, participated in enumerate(contribution.aggregation_bits):
             if participated:
-                participant_index = SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT * subcommittee_index + index
+                participant_index = int(SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT * subcommittee_index + index)
```
ralexstokes, line 232: the `remerkleable` definition wants an `int` and the operations w/ the constants pollutes the type to `uint64` which led to some problems when setting values in the underlying `Bitvector` `sync_committee_bits`

hwwhww:
The non-test-vector-output tests are under `unittests` folder. So `tests/core/pyspec/eth2spec/test/altair/validator/test_validator.py` should be moved to `tests/core/pyspec/eth2spec/test/altair/unittests/validator/test_validator.py`.

tests/core/pyspec/eth2spec/test/altair/validator/test_validator.py
```diff
@@ -0,0 +1,155 @@
...
+        spec.SyncCommitteeContribution(
+            slot=block.slot,
+            beacon_block_root=target_block_root,
+            subcommittee_index=i,
+            aggregation_bits=aggregation_bits,
+            signature=_get_sync_committee_signature(
+                spec, state, previous_slot, target_block_root, i, aggregation_index
+            ),
+        )
+        for i in range(spec.SYNC_COMMITTEE_SIZE // spec.SYNC_COMMITTEE_SUBNET_COUNT)
```
hwwhww, line 117: I think it should be:

```suggestion
        for i in range(spec.SYNC_COMMITTEE_SUBNET_COUNT)
```

hwwhww (changes requested):
Good to see test coverage improvement! ✨

specs/altair/validator.md
ralexstokes: @protolambda do you have any thoughts or preferences about the typing here -- the indexing method for `remerkleable`'s `Bitvector` wants an `int` but could take any "int"-like type, especially one internal to the `remerkleable` library

protolambda: `uint8, uint16, ` etc. in remerkleable are all sub-types of `int`. Even `boolean` is (just like python `bool` is too). There should be no problem using it as argument in a list-lookup. Maybe the type is lost somewhere? I'm not sure what error/warning you're seeing.

tests/core/pyspec/eth2spec/test/altair/validator/test_validator.py
ralexstokes: ah yes, this was the intention, thanks for catching this!

specs/altair/validator.md
ralexstokes: rolling back this change and waiting until we can pull this into the specs repo: https://github.com/protolambda/remerkleable/pull/11

tests/core/pyspec/eth2spec/test/altair/validator/test_validator.py
```diff
@@ -0,0 +1,155 @@
...
+    spec, state, epoch, sync_committee, active_pubkeys
+):
+    assert len(sync_committee.pubkeys) >= 3
+    some_pubkeys = rng.sample(sync_committee.pubkeys, 3)
+    for pubkey in some_pubkeys:
+        validator_index = active_pubkeys.index(pubkey)
+        assert spec.is_assigned_to_sync_committee(state, epoch, validator_index)
+
+
+@with_all_phases_except([PHASE0])
```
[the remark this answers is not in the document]
ralexstokes: ah thanks for checking `mainnet` -- i update this and the next test to work at either config

ralexstokes:
update: @hwwhww thanks for the feedback, that was very helpful! i've addressed most of it i believe and everything should work now -- the CI is still failing but should pass once we rebase onto a version of the pyspec w/ the new version of `remerkleable` that @protolambda is working on

tests/core/pyspec/eth2spec/test/altair/unittests/validator/test_validator.py
```diff
@@ -0,0 +1,166 @@
...
+
+@with_all_phases_except([PHASE0])
+@with_state
+def test_compute_subnets_for_sync_committee(state, spec, phases):
+    some_sync_committee_members = list(
+        (
+            _subnet_for_sync_committee_index(spec, i),
+            pubkey,
+        )
+        for i, pubkey in enumerate(state.current_sync_committee.pubkeys)
```
hwwhww, line 155: yay for making it mainnet config compatible! 🎉

hwwhww (approved):
The changes LGTM!

protolambda:
@ralexstokes new remerkleable release with fix: https://pypi.org/project/remerkleable/0.1.19/

PR #2312, opened 2021-04-06
Beacon Chain sub-tree filter on shard blobs

protolambda:
Enforce state sub-tree in shard blob proposal to avoid inconsistent replays on reorg, and provide context for `proposer_index` computation.

Previously, a shard blob could be replayed on another chain that omits confirmed shard-headers, delaying ultimate confirmation to the chain as a whole. This makes L2 transaction confirmation logic much more difficult. Also, the `proposer_index` need to be validated in the p2p layer, but was missing the required context to do so during deep forks that would change shard proposer indices.

Note that unlike the old (pre DAS refactor) shard-transition style would enforce much more (too much), this just tries to strike a balance between the two.

Shard-headers can be included in any order, and late too. But they can always rely on a pinned state history relative to their proposal slot.

This PR targets the shard-blob PR which improves the sharding types, still in review. 

Edit: pinning state with `block_root, slot` combination instead of a `state_root` to make it clear how to process towards the point. This matches blocks and attestations, at the cost of a little complexity for a shard-data user (gap slots need to be accounted for, the `slot` part is essential). Ignore the old `shard_blob_stateroot` branch name.

### DIFF
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -203,6 +203,8 @@ class ShardBlobBodySummary(Container):
     degree_proof: BLSCommitment
     # Hash-tree-root as summary of the data field
     data_root: Root
+    # State of the Beacon Chain, right before the slot processing of shard_blob.slot
+    parent_state_root: Root
 ```
 
 ### `ShardBlobHeader`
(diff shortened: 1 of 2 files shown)


protolambda:
Merged into the main shard-blobs PR to simplify things. See #2302

PR #2313, opened 2021-04-07
Update note about changes to slash_validator

ajsutton:
1.1.0-alpha.3 updated the calculation for proposer rewards in `slash_validator`.  Include that change in the note explaining what changed.


hwwhww (approved):
LGTM 👍

PR #2321, opened 2021-04-09
Refactor pyspec builder with `SpecBuilder` classes

hwwhww:
### Issue
The current spec builder contains hardcoded functions for different fork versions. e.g., `PHASE0_IMPORTS`, `ALTAIR_IMPORTS`, `MERGE_IMPORTS`). And more versions (`SHARDING`, `CUSTODY_GAME`, `DAS`) are coming! It would be nice to clean up the hacky hard coding before that.

### Proposed changes

*updated*: refactored `imports_and_predefinitions` into `imports` and `preparations`

To make the builder more readable and maintainable, this PR rearranges the hardcoded strings into class methods of `SpecAdjustment` classes. Now there are these interfaces:
- `imports`: Import objects
- `preparations`: Define special types/constants for building pyspec or call functions. e.g, `SSZVariableName = str`.
- `sundry_functions`: The functions that are (1) defined abstractly in specs or (2) adjusted for better performance.
- `hardcoded_ssz_dep_constants`: The constants that are dependent by SSZ objects.
- `hardcoded_custom_type_dep_constants`: The constants that are depended by custom types.
- `invariant_checks`: The invariant checks.

With this abstraction, we can remove the `is_<version_name>` condition checks from `objects_to_spec` helper.

Also note that, `AltairSpecAdjustment` and `MergeSpecAdjustment` both inherit `Phase0MergeSpecAdjustment` as the current pyspec versioning setting. So we can choose to use or combine with the parent method.


djrtwo (commented):
This looks good.

Why "adjustment" in the name here?
What do you think about "builder" -- `SpecBuilder`

hwwhww:
@djrtwo 

> Why "adjustment" in the name here?
> What do you think about "builder" -- `SpecBuilder`

These are the "adjustments" for `SpecObject` objects. 
There is already a function named `build_spec`, but I'm okay with renaming it to `SpecBuilder`.

Moreover, we can also add `build_spec` to `SpecBuilder` instances.

djrtwo:
I see

I'm okay either way. Choose your preference

djrtwo (approved):
I'm okay with merging this. Definitely a net win in complexity

PR #2322, opened 2021-04-09
Merge cleanups and improvements

mkalinin:
This PR introduces the following changes:
- `Hash32` custom type. This type is used to denote data type of execution payload hashes (`block_hash`, `parent_hash`). Though, `ExecutionPayload.state_root` still has `Bytes32` type, it is preserved to distinguish MPT root hash from merkle roots used by the beacon chain.
- `execution_state` object is replaced with `execution_state_root` in `execution_state_transition` params list. In the composite client implementation an operation to retrieve `execution_state ` object to further pass it to the function does not make sense as state management is delegated to execution engine. This change makes the spec suitable for both types of clients.
- Adds explicit computation of execution block `timestamp` and passes it onto `execution_state_transition` and `produce_execution_payload` functions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -313,11 +313,13 @@ def get_pow_chain_head() -> PowBlock:
     pass
 
 
-def execution_state_transition(execution_state: ExecutionState, execution_payload: ExecutionPayload) -> None:
+def execution_state_transition(execution_state_root: Bytes32,
+                               execution_payload: ExecutionPayload,
+                               timestamp: uint64) -> None:
     pass
 
 
-def produce_execution_payload(parent_hash: Bytes32) -> ExecutionPayload:
+def produce_execution_payload(parent_hash: Hash32, timestamp: uint64) -> ExecutionPayload:
     pass"""
 
 
--- a/setup.py
+++ b/setup.py
@@ -313,10 +313,8 @@ def get_pow_chain_head() -> PowBlock:
     pass
 
 
-def execution_state_transition(execution_state_root: Bytes32,
-                               execution_payload: ExecutionPayload,
-                               timestamp: uint64) -> None:
-    pass
+def verify_execution_state_transition(execution_payload: ExecutionPayload) -> bool:
+    return True
 
 
 def produce_execution_payload(parent_hash: Hash32, timestamp: uint64) -> ExecutionPayload:
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -121,6 +121,7 @@ class ExecutionPayload(Container):
     number: uint64
     gas_limit: uint64
     gas_used: uint64
+    timestamp: uint64
     receipt_root: Bytes32
     logs_bloom: ByteVector[BYTES_PER_LOGS_BLOOM]
     transactions: List[OpaqueTransaction, MAX_APPLICATION_TRANSACTIONS]
@@ -141,6 +142,7 @@ class ExecutionPayloadHeader(Container):
     number: uint64
     gas_limit: uint64
     gas_used: uint64
+    timestamp: uint64
     receipt_root: Bytes32
     logs_bloom: ByteVector[BYTES_PER_LOGS_BLOOM]
     transactions_root: Root
@@ -187,7 +189,7 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
 
 ##### `execution_state_transition`
 
-Let `execution_state_transition(execution_state_root: Bytes32, execution_payload: ExecutionPayload, timestamp: uint64) -> None` be the transition function of Ethereum execution state. 
+Let `execution_state_transition(execution_state_root: Bytes32, execution_payload: ExecutionPayload) -> None` be the transition function of Ethereum execution state. 
 The body of the function is implementation dependent.
 
 *Note*: `execution_state_transition` must throw `AssertionError` if either the transition itself or one of the pre or post conditions has failed.
@@ -209,9 +211,10 @@ def process_execution_payload(state: BeaconState, body: BeaconBlockBody) -> None
         assert execution_payload.parent_hash == state.latest_execution_payload_header.block_hash
         assert execution_payload.number == state.latest_execution_payload_header.number + 1
 
-    timestamp = compute_time_at_slot(state, state.slot)
+    assert execution_payload.timestamp == compute_time_at_slot(state, state.slot)
+
     execution_state_root = state.latest_execution_payload_header.state_root
-    execution_state_transition(execution_state_root, body.execution_payload, timestamp)
+    execution_state_transition(execution_state_root, body.execution_payload)
 
     state.latest_execution_payload_header = ExecutionPayloadHeader(
         block_hash=execution_payload.block_hash,
@@ -221,6 +224,7 @@ def process_execution_payload(state: BeaconState, body: BeaconBlockBody) -> None
         number=execution_payload.number,
         gas_limit=execution_payload.gas_limit,
         gas_used=execution_payload.gas_used,
+        timestamp=execution_payload.timestamp,
         receipt_root=execution_payload.receipt_root,
         logs_bloom=execution_payload.logs_bloom,
         transactions_root=hash_tree_root(execution_payload.transactions),
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -26,6 +26,7 @@
   - [Misc](#misc)
     - [`is_transition_completed`](#is_transition_completed)
     - [`is_transition_block`](#is_transition_block)
+    - [`compute_time_at_slot`](#compute_time_at_slot)
   - [Block processing](#block-processing)
     - [Execution payload processing](#execution-payload-processing)
       - [`get_execution_state`](#get_execution_state)
@@ -149,6 +150,14 @@ def is_transition_block(state: BeaconState, block_body: BeaconBlockBody) -> bool
     return not is_transition_completed(state) and block_body.execution_payload != ExecutionPayload()
 ```
 
+#### `compute_time_at_slot`
+
+```python
+def compute_time_at_slot(state: BeaconState, slot: Slot) -> uint64:
+    slots_since_genesis = slot - GENESIS_SLOT
+    return uint64(state.genesis_time + slots_since_genesis * SECONDS_PER_SLOT)
+```
+
 ### Block processing
 
 ```python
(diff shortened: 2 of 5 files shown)


specs/merge/validator.md
```diff
@@ -66,5 +66,6 @@ def get_execution_payload(state: BeaconState) -> ExecutionPayload:
```
hwwhww, line 66: (couldn't comment on to L65 on GitHub, commenting on L66 instead)

L65 also needs `timestamp` argument for calling `produce_execution_payload`.

mkalinin: Well spotted 👍 Fixed now

hwwhww:
We also have `eth1_block_hash: Bytes32` in phase 0 spec so it could be cleaner if we add custom type to phase 0... is it an acceptable phase 0 refactor? 😅

specs/merge/fork-choice.md
```diff
@@ -30,15 +30,15 @@ This is the modification of the fork choice according to the executable beacon c
...
+    block_hash: Hash32
     is_processed: boolean
     is_valid: boolean
     total_difficulty: uint256
 ```
 
 #### `get_pow_block`
 
-Let `get_pow_block(hash: Bytes32) -> PowBlock` be the function that given the hash of the PoW block returns its data.
+Let `get_pow_block(hash: Hash32) -> PowBlock` be the function that given the hash of the PoW block returns its data.
```
hwwhww, line 41: What do you think about using `block_hash` instead of `hash`? The keyword `hash` is used in the [`hash` function](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#hash)


```suggestion
Let `get_pow_block(block_hash: Hash32) -> PowBlock` be the function that given the hash of the PoW block returns its data.
```

specs/merge/beacon-chain.md
```diff
@@ -161,16 +170,9 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
-
-*Note*: `ExecutionState` class is an abstract class representing Ethereum execution state.
-
-Let `get_execution_state(execution_state_root: Bytes32) -> ExecutionState`  be the function that given the root hash returns a copy of Ethereum execution state. 
-The body of the function is implementation dependent.
-
 ##### `execution_state_transition`
 
-Let `execution_state_transition(execution_state: ExecutionState, execution_payload: ExecutionPayload) -> None` be the transition function of Ethereum execution state. 
+Let `execution_state_transition(execution_state_root: Bytes32, execution_payload: ExecutionPayload, timestamp: uint64) -> None` be the transition function of Ethereum execution state.
```
djrtwo, line 175: We might consider making the name of `execution_state_root` a bit more clear. It is unclear if it is the pre or post root at first look.

djrtwo, line 175: Any thoughts on including `timestamp` explicitly in the `ExecutionPayload`? Is it valuable to make sure that the "eth1" block can be formed from just the contents of the payload? Or is it okay to require beaconblock context (e.g. convertin the slot to timestamp and passing in separately

djrtwo, line 175: Also, do we even need to pass in the pre `execution_state_root`? That information is contained within `parent_hash` within the payload and it is assumed that the execution as the post state of the parent (otherwise would throw)

specs/merge/validator.md
```diff
@@ -62,9 +62,11 @@ def get_execution_payload(state: BeaconState) -> ExecutionPayload:
             return ExecutionPayload()
         else:
             # Signify merge via producing on top of the last PoW block
-            return produce_execution_payload(pow_block.block_hash)
+            timestamp = compute_time_at_slot(state, state.slot)
```
djrtwo, line 65: What does the execution-layer use this `timestamp` for?

specs/merge/beacon-chain.md
mkalinin: `timestamp` can be a part of `ExecutionPayload` with a cost of the following validity condition
```python
assert body.execution_payload.timestamp == compute_time_at_slot(state, state.slot)
```

Having this field a part of the payload might be really useful!

mkalinin: Right. Originally, there was an intention to make `execution_state_transition` a pure function of `execution_state` and `execution_payload`. Though, this is not always the case as whether there is an access to the `execution_state` object or there isn't is implementation dependent.

For stateless execution there will be no need in the state root at all as the payload will be self-sufficient in terms of inputs into execution state transition. If we have `timestamp` field in `ExecutionPayload` then it might look like:
```python
verify_execution(execution_payload: ExecutionPayload) -> boolean

...

assert verify_execution(body.execution_payload)
```

specs/merge/validator.md
mkalinin: There is a `TIMESTAMP` opcode it is used for. Also, the difference of timestamp between child and parent block is enforced by block validity checks and difficulty formula, though, these checks will be omitted after the merge. But we have to support the opcode.

mkalinin:
> We also have `eth1_block_hash: Bytes32` in phase 0 spec so it could be cleaner if we add custom type to phase 0... is it an acceptable phase 0 refactoring? 😅

@hwwhww done!

specs/merge/beacon-chain.md
djrtwo: I like that it remains more self-contained and would default to including it in there unless we have a compelling reason *not* to

djrtwo: Right, that's why we were leaning toward that prior structure to begin with. Remove the implicit argument

I still kind of like the prior structure, but don't feel strongly enough to argue for it too much.

You could actually have the same function signature (explicitly passing in state) with stateless as well -- e.g. build a "state" object that is actually a partial and just has the accesses provided from the block. But this isn't really necessary

If we do a stateful verification here, I would argue to remove the explicit state root and make it more like the `verify_execution` function signature above. The extra param isn't used now and wouldn't be used if/when we move to stateless so probably should just not be there

specs/merge/validator.md
djrtwo: Got it. thanks!

specs/merge/beacon-chain.md
mkalinin: Let me apply this change and see how it would look like, we may revert it in case of dislike

mkalinin:
@djrtwo Updated with timestamp in `ExecutionPayload` and `validate_execution_payload` function that should work for any implementation design including stateless execution.

specs/merge/beacon-chain.md
```diff
@@ -66,6 +69,19 @@ We define the following Python custom types for type hinting and readability:
 
 ## Containers
 
+### Modified containers
+
+#### `Eth1Data`
+
+*Note*: The only modification is the type of `block_hash` field that is changed to `Hash32`.
```
protolambda, line 76: I think it's fine to change the phase0 spec and update the type. It's just a change in presentation.

specs/merge/beacon-chain.md
```diff
@@ -148,6 +166,14 @@ def is_transition_block(state: BeaconState, block_body: BeaconBlockBody) -> bool
     return not is_transition_completed(state) and block_body.execution_payload != ExecutionPayload()
 ```
 
+#### `compute_time_at_slot`
+
+```python
```
protolambda, line 172: It's used safely here, but we may want to add a note about under/overflows in the math here, in case the function is going to be used elsewhere.

specs/merge/beacon-chain.md
djrtwo: agreed

djrtwo (commented):
cool!

a couple of minor comments

specs/merge/beacon-chain.md
```diff
@@ -194,9 +209,9 @@ def process_execution_payload(state: BeaconState, body: BeaconBlockBody) -> None
         assert execution_payload.parent_hash == state.latest_execution_payload_header.block_hash
         assert execution_payload.number == state.latest_execution_payload_header.number + 1
 
-    timestamp = compute_time_at_slot(state, state.slot)
-    execution_state_root = state.latest_execution_payload_header.state_root
-    execution_state_transition(execution_state_root, body.execution_payload, timestamp)
+    assert execution_payload.timestamp == compute_time_at_slot(state, state.slot)
+
+    assert validate_execution_payload(execution_payload)
```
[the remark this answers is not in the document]
mkalinin: What I don't like about `is_` is that it usually a prefix of getters which are pure functions while this function is impure in case of stateful execution. We can use `verify_` if it's more suitable.

specs/merge/beacon-chain.md
```diff
@@ -170,13 +187,11 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
 
 #### Execution payload processing
 
-##### `execution_state_transition`
+##### `validate_execution_payload`
 
-Let `execution_state_transition(execution_state_root: Bytes32, execution_payload: ExecutionPayload, timestamp: uint64) -> None` be the transition function of Ethereum execution state. 
+Let `validate_execution_payload(execution_payload: ExecutionPayload) -> boolean` be the function checking whether given `ExecutionPayload` is valid or not.
```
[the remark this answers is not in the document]
mkalinin: Let me fix all `boolean`'s in the spec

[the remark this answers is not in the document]
mkalinin: Then `validate_execution_state_transition` or `verify_execution_state_transition` probably?

mkalinin:
@protolambda @djrtwo addressed your recent comments

specs/merge/beacon-chain.md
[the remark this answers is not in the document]
djrtwo: You're right

`is_` is innappropriate for the mutation currently. *might* be reasonable if it were stateless but `verify_` works well

setup.py
```diff
@@ -313,11 +313,13 @@ def get_pow_chain_head() -> PowBlock:
     pass
 
 
-def execution_state_transition(execution_state: ExecutionState, execution_payload: ExecutionPayload) -> None:
+def execution_state_transition(execution_state_root: Bytes32,
```
djrtwo, line 316: This needs to be updated for the linter/build

djrtwo: I fixed it.

added `return True` instead of pass in setup.py because of usage. What do you think?

mkalinin: Makes sense to me since we have no tests for `process_execution_payload` yet

specs/merge/beacon-chain.md
djrtwo: moved to a phase 0 type mod

djrtwo (approved):

Issue #2318, opened 2021-04-08
Set sync committees as part of Altair genesis


ajsutton:
Currently the Altair reference tests include generating genesis states which expect the sync committees to be filled with all zero public keys.  That means for the first two sync committee periods, there are no real validators in the sync committees and clients have to have special case handling to deal with public keys in the committees that don't match a known validator.

While MainNet won't need to generate a genesis in Altair, future testnets likely will and being able to do so makes testing significantly easier.  It would be good to update the genesis generation process for Altair to calculate the current and next sync committees properly using the same approach as in the transition process.

hwwhww:
fixed via #2323

PR #2323, opened 2021-04-09
Simplify Altair "genesis"

hwwhww:
## Issue
1. In the pyspec test suite, we get Altair `state` with `upgrade_to_altair`:
https://github.com/ethereum/eth2.0-specs/blob/6f095fc6916dcb7803b6cded68c0367f72a14f06/tests/core/pyspec/eth2spec/test/context.py#L76-L89
It assumes that `ALTAIR_FORK_SLOT` is `0` in the test cases. Therefore, after `upgrade_to_altair`, the result `state.latest_block_header` is from **Phase 0** `latest_block_header`.
	* -> That’s also what the first block `block_1.parent_root` points to.
2. In fork choice spec [`on_block`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/fork-choice.md#on_block ), the given `block.parent_root` must be known in `store`. (`assert block.parent_root in store.block_states`)
	* That’s why in fork choice test suite, we set the genesis block to must be **Phase 0** format:
https://github.com/ethereum/eth2.0-specs/blob/6f095fc6916dcb7803b6cded68c0367f72a14f06/tests/core/pyspec/eth2spec/test/helpers/fork_choice.py#L61-L62

Now think about the Altair-only testnets, where `ALTAIR_FORK_SLOT == 0`. With the current spec, the initialization is like:
1. Use `initialize_beacon_state_from_eth1` to get `genesis_state: phase0.BeaconState`.
2. Get `genesis_block = phase0.BeaconBlock(state_root=hash_tree_root(genesis_state))`
3. Call `upgrade_to_altair` to get `state: altair.BeaconState`
4. Initialize fork choice store with `get_forkchoice_store(state, genesis_block)`
	* Note that they are `state: altair.BeaconState` and `genesis_block: phase0.BeaconState`

It could be simplified if not using `upgrade_to_altair`:

1. Use **new** `initialize_beacon_state_from_eth1` to get `genesis_state: phase1.BeaconState`.
2. Get `genesis_block = altair.BeaconBlock(state_root=hash_tree_root(genesis_state))`
3. Initialize fork choice store with `get_forkchoice_store(genesis_state, genesis_block)`
	* Note that they are `state: altair.BeaconState` and `genesis_block: altair.BeaconState`

## Proposed solutions

1.  **[`context.py`]** We can just use `create_genesis_state` helper to generate the “initial state” for Altair or other versions.
```python
def _prepare_state(balances_fn: Callable[[Any], Sequence[int]], threshold_fn: Callable[[Any], int],
                   spec: Spec, phases: SpecForks):
    phase = phases[spec.fork]
    balances = balances_fn(phase)
    activation_threshold = threshold_fn(phase)
    state = create_genesis_state(spec=phase, validator_balances=balances,
                                 activation_threshold=activation_threshold)
    return state
```
2. **[Fork choice tests]** Change [get_genesis_forkchoice_store_and_block](https://github.com/ethereum/eth2.0-specs/blob/6f095fc6916dcb7803b6cded68c0367f72a14f06/tests/core/pyspec/eth2spec/test/helpers/fork_choice.py#L61-L62) helper to use `spec.BeaconBlock` instead of `phase0_spec.BeaconBlock(state_root=genesis_state.hash_tree_root())`
3. **[`beacon-chain.md`]** Add `initialize_beacon_state_from_eth1` in Altair specs, which is only used in the test suite and for testnets.

### Pros:
- Simpler testnet initialization.
### Cons:
- DRY: it introduces some redundant code in `initialize_beacon_state_from_eth1` and `upgrade_to_altair`. But I think we can refactor it to be acceptable.
- Much less tests cover `upgrade_to_altair`: after this change, only `test_fork.py` tests will cover `upgrade_to_altair`.


hwwhww:
/cc @ajsutton @benjaminion

I *think* the `anchor_block` deserialization issue you saw may be caused by `phase0.BeaconBlock` v.s. `altair.BeaconBlock`?

ajsutton:
> /cc @ajsutton @benjaminion
> 
> I _think_ the `anchor_block` deserialization issue you saw may be caused by `phase0.BeaconBlock` v.s. `altair.BeaconBlock`?

Yes you're right - I thought I'd tried that in the past but apparently got it wrong somehow.  If I forcibly create a phase0 "spec" to parse that initial block then the Altair fork choice tests pass for me.

Note that the Altair genesis reference tests do *not* run `upgrade_to_altair` and the sync committees in the state wind up filled with zeroed out public keys. I logged https://github.com/ethereum/eth2.0-specs/issues/2318 about that.  The Altair `initialize_beacon_state_from_eth1` should also calculate the current and next sync committees.

I'm a big fan of being able to start from a pure Altair genesis and not need any phase0 code because that leaves the possibility of eventually dropping phase0 support from codebases entirely and just using a later anchor state to sync from. Not sure if we'll ever actually do that but it would be really nice if it were at least possible so new clients can just implement the latest spec instead of having to implement every historical phase as well.

coots0001:
Fix it￼

hwwhww:
Thanks @ajsutton for the reply! Right, so this PR also fixes #2318.

I think we need to add other more fork choice rule tests around the fork boundary. #2333 is the first issue that popped up.

@djrtwo @protolambda 
I’m thinking about if we can:
1. Set `ALTAIR_FORK_SLOT = 0` in the `minimal` config
2. Add `minimal_fork` config, which is as same as the minimal config except for `ALTAIR_FORK_SLOT = 64` (some number that is not too large but have passed epochs) for testing the boundaries.

ajsutton:
Agree that we should be able to test both Altair at genesis and Altair at some point after Phase0.  I think we also want a config where Altair doesn't kick in though (ie `ALTAIR_FORK_SLOT = FAR_FUTURE_EPOCH`).

It sounds like it would be really useful to be able to say "minimal spec but with altair fork at epoch X". Otherwise we're going to get a bit of an explosion of configs as we add more forks.

hwwhww:
@ajsutton I updated the docs in 666f847:

1. When handling test vector:
    - By default, use the given <fork or phase name> version to deserialize the binary data.
    - We might add some meta fields we have test cases cross fork boundary. Right now, only the forks tests cross the fork boundary and the test format described it.
    - Use new `initialize_beacon_state_from_eth1` to run the genenis tests.
2. When handling the pure Altair testnet:
    - Use new `initialize_beacon_state_from_eth1` to generate the genesis.
    - Do not apply `upgrade_to_altair`.

Keeping `ALTAIR_FORK_SLOT = FAR_FUTURE_EPOCH` for the minimal case now. We can add a new config once we have new test cases that cross the fork boundary.

/cc @djrtwo

specs/altair/beacon-chain.md
```diff
@@ -661,3 +662,51 @@ def process_sync_committee_updates(state: BeaconState) -> None:
         state.current_sync_committee = state.next_sync_committee
         state.next_sync_committee = get_sync_committee(state, next_epoch + EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
 ```
+
+## Initialize state for pure Altair testnets and test vectors
+
+This helper function is only for initializing the state for pure Altair testnets and tests.
```
djrtwo, line 668: We should note explicitly that it is for the case when `GENESIS_SLOT == ALTAIR_FORK_SLOT`

specs/altair/beacon-chain.md
```diff
@@ -661,3 +662,51 @@ def process_sync_committee_updates(state: BeaconState) -> None:
...
+
+*Note*: The function `initialize_beacon_state_from_eth1` is modified with `ALTAIR_FORK_VERSION` fork version and initial sync committees.
```
djrtwo, line 670: and utilizes the Altair `BeaconBlockBody` when constructing the initial `latest_block_header`

specs/altair/fork.md
```diff
@@ -34,6 +34,8 @@ Warning: this configuration is not definitive.
 
 TBD. Social consensus, along with state conditions such as epoch boundary, finality, deposits, active validator count, etc. may be part of the decision process to trigger the fork. For now we assume the condition will be triggered at slot `ALTAIR_FORK_SLOT`, where `ALTAIR_FORK_SLOT % SLOTS_PER_EPOCH == 0`.
 
+Note that for the pure Altair testnets, we don't apply `upgrade_to_altair` since it starts with Altair version logic.
```
djrtwo, line 37:
```suggestion
Note that for the pure Altair networks, we don't apply `upgrade_to_altair` since it starts with Altair version logic.
```

djrtwo (commented):
A few minor comments. I'm okay with this approach but am concerned about being able to more dynamically set fork slots for testing

`ALTAIR_FORK_SLOT = FAR_FUTURE_EPOCH` -- Why are we setting it to far future epoch instead of `GENESIS_SLOT`? Don't we want the fork to by default happen for all of the tests today?

Should fork slots be a part of the core config? Maybe these should live in a separate file. Core configs are often used for compile targets and are expected to be few and far between, but we'll likely want fork slots to be highly configurable.

specs/altair/beacon-chain.md
hwwhww: I think it would be confusing if we add `GENESIS_SLOT == ALTAIR_FORK_SLOT` here. Two reasons:
1. For pure Altair testnets, it actually doesn't need `ALTAIR_FORK_SLOT` since we don't call `upgrate_to_altair` at all.
2. `ALTAIR_FORK_SLOT` it's inconsistent with the config files we provide. As you said, I agreed that it makes sense to set fork epoch in a separate file.

specs/altair/beacon-chain.md
```diff
@@ -661,3 +662,51 @@ def process_sync_committee_updates(state: BeaconState) -> None:
...
+
+*Note*: The function `initialize_beacon_state_from_eth1` is modified: (1) using `ALTAIR_FORK_VERSION` as the current fork version, (2) utilizing the Altair `BeaconBlockBody` when constructing the initial `latest_block_header`, and (3) adding initial sync committees.
+
+```python
+def initialize_beacon_state_from_eth1(eth1_block_hash: Bytes32,
+                                      eth1_timestamp: uint64,
+                                      deposits: Sequence[Deposit]) -> BeaconState:
+    fork = Fork(
+        previous_version=GENESIS_FORK_VERSION,
+        current_version=ALTAIR_FORK_VERSION,
```
djrtwo, line 678:
```suggestion
        current_version=ALTAIR_FORK_VERSION,  # [Modified in Altair]
```

djrtwo (approved):

PR #2325, opened 2021-04-12
Refactor pyspec `context.py` by adding `typing.py` and `constants.py`

hwwhww:
## Issue
In #2323
1. Spec fork names (`PHASE0`, `ALTAIR`) is defined in `eth2spec.test.context`
2. `eth2spec.test.helpers.genesis.create_genesis_state` function is called by `eth2spec.test.context._prepare_state` function
3. I wanted to use spec fork names in `eth2spec.test.helpers.genesis`... but due to (1) and (@), it would introduce a dependency issue.

## Proposed solution
- Move pyspec custom types to `eth2spec.test.helpers.typing`
- Move pyspec constants to `eth2spec.test.helpers.constants`

Then the forks names and config names can be accessed anywhere.


djrtwo (approved):
looks good!

Issue #2290, opened 2021-03-30
Remove duplicate `CONFIG_NAME` from configs


protolambda:
Each configuration file has a duplicate `CONFIG_NAME` that can only result in conflicting values, and it makes it more difficult to consolidate the config files.

Discussion (@michaelsproul, @mcdee, @rolfyone): https://discord.com/channels/595666850260713488/710831784991916072/823906028600033351

> there's this weird exception right now, the `CONFIG_NAME` variable is duplicated between all configs (and might be different for unknown reasons)

> Why does CONFIG_NAME need to be in altair.yml?

> Legacy choice, I think it can be removed. My intention was always to have unique variables, simple key-value, yaml compatible, and ability to consolidate into a single namespace easily (for compatibility with env vars)

----

I think it's valuable to not have duplicate variables, and avoid dependency on them before we introduce a stable version of the first fork (Altair). In the future we may consolidate the files into a single YAML, and client implementers may do this in their internal config representation already.

hwwhww:
Right, so the `spec.CONFIG_NAME` is for making `with_configs` work. I think it's good to keep the current separate per-fork-per-config files so that we can compare them with the specs easily.

A simple solution here, similar to ToC generation:
1. Add makefile command `make merge_config` to create merged config files: `configs/merged/<SPEC_CODE_NAME>_<CONFIG_NAME>.yaml`.
    - `phase0_mainnet.yaml`, `altair_mainnet.yaml`, `phase0_minimal.yaml`, `altair_minimal.yaml`.
    - Each merged config contains the previous forks' config + current fork config.
    - The merged file doesn't contain `CONFIG_NAME` field, but it checks if there are any conflicts except for ` CONFIG_NAME` between the atomic config files during the makefile command.
2. Add CI job to check if the merged config files are correct.
3. Output these merged config files to test vector suite output files.

hwwhww:
closing via #2332

PR #2328, opened 2021-04-13
Output single YAML config file for each config and spec version

hwwhww:
### Issue
#2290: @protolambda In the future we may consolidate the files into a single YAML, and client implementers may do this in their internal config representation already.

I haven't investigated how to remove `CONFIG_NAME` smoothly. However, I think we can just present the single YAML plan to implementers now!

### Proposed solution
1. Add Makefile commands:
    - `make combine_configs`
        - Generate the combined `configs/combined/<config_name>/<spec_version_name>.yaml`.
        - Items are sorted by field name.
        - Note that there is NO `CONFIG_NAME` field in the combined config file now. Can add it back if we want.
    - `make check_combined_configs`
        - Verify if the existing combined files are up-to-date.
2. Add CI job that calls `make check_combined_configs`.

For the spec writers:
1. For readability, we still manually update the `configs/<config_name>/<spec_version_name>.yaml` files.
2. Need to run `make combine_configs` once you update the configs.


protolambda:
Although I do like the idea of combining the configs, I think we should not drop the comments, and should sort by meaning instead of name.

I see two possible alternatives:
1. Drop the duplicate `CONFIG_NAME` field, maintain separate config files in the spec repository, and make concatenation of the configs standard-practice for usage in implementations.
2. Maintain a single config file in the spec repository. Clients can ignore config fields that they do not recognize (e.g. sharding configuration). The config can be structured even more (e.g. list domain-types together, list fork-slots together, etc.)

Personally I like the 2nd alternative best, but I'm open to other options.

hwwhww:
@protolambda

> 1. Make concatenation

I guess we can. Just think of a hacky way to compatible with pyspec:
1. Drop the duplicate `CONFIG_NAME` field from YAML files.
2. Add `CONFIG_NAME` at the end of `load_config_file` method. So it's only accessible in pyspec context.

> 2. Maintain a single config file in the spec repository.

It used to be one single config file but implementers complained that there was too much noise from the in-R&D specs (sharding). Therefore we split it into multiple files.

With this approach, we will get the really per-fork combined configs:
- `configs/combined/<config_name>/phase0.yaml`: Phase0 configs
- `configs/combined/<config_name>/altair.yaml`: Phase0 + Altair configs
- [TODO in the future] `configs/combined/<config_name>/merge.yaml`: Phase0 + Altair + Merge configs

---

If the current PR is Proposal 0. I like 0=1>2. 😅

djrtwo:
I do echo that we should keep R&D out of the configs and separating by fork seems to handle that nicely.

In the current proposal, why have the combined files built into the actual repo? Instead of just building when building the pyspec (and not committing). Client teams already combine without issue (other than duplicate name field), right? If that is the case, can we not just remove the duplicate name from the configs?

I think I lean toward @protolambda's option (1).

hwwhww:
@djrtwo 
> why have the combined files built into the actual repo? Client teams already combine without issue (other than duplicate name field), right?

It may be useful for other tools. The use case I thought of was preparing testnets on Goerli.

hwwhww:
Closing this PR now. We will check with clients' requirements and then pick up later.

PR #2332, opened 2021-04-15
Remove `CONFIG_NAME` from the config files.

hwwhww:
Alternative to #2328
Address #2290

1. Remove `CONFIG_NAME` from the config files. 
2. And add it back to pyspec context for testing.

@djrtwo @protolambda 
We can merge this PR first, and then discuss if the single config file is needed.


djrtwo (approved):

djrtwo:
Looks good to me

Issue #2330, opened 2021-04-14
Multi-period tests for `process_sync_committee_updates`


michaelsproul:
Presently there's 1 test for `process_sync_committee_updates` that checks the transition from period 0 to period 1. This is an exceptional case, because the `base_epoch` for both period 0 and period 1 is epoch 0.

In Lighthouse, I had written an erroneous bounds check that looked like this:

```rust
 if base_epoch != current_base_epoch
     && base_epoch != current_base_epoch.safe_add(spec.epochs_per_sync_committee_period)?
 {
    return Err(Error::EpochOutOfBounds);
 }
 ```
 
It failed as soon as I tried transitioning from period 1 to period 2, because `process_sync_committee_updates` tried to calculate the sync committee for period 3, which has `base_epoch == 2 * EPOCHS_PER_SYNC_COMMITTEE_PERIOD` which is not equal to the `current_base_epoch` (0) + `EPOCHS_PER_SYNC_COMMITTEE_PERIOD`.

I think it would be good to test a period transition from a period other than 0 to catch mistakes like this.

djrtwo:
added a couple of cases in #2334

michaelsproul:
Another issue I just noticed: the sync aggregate epoch processings tests lack a `meta.yaml`, yet they _need_ BLS enabled because constructing a sync committee involves real signature aggregation.

djrtwo:
closed via #2334

PR #2334, opened 2021-04-15
add committee progress tests for non genesis case

djrtwo:
addresses  #2330

* added case for period 1/2 instead of just 0/1
* added case for misc_balances state for additional random coverage

### DIFF
--- a/tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_sync_committee_updates.py
+++ b/tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_sync_committee_updates.py
@@ -1,4 +1,5 @@
 from eth2spec.test.context import (
+    always_bls,
     spec_state_test,
     spec_test,
     with_all_phases_except,
@@ -16,6 +17,11 @@ from eth2spec.test.helpers.epoch_processing import (
     run_epoch_processing_with,
 )
 
+
+#
+# Note:
+# Calculating sync committees requires pubkey aggregation, thus all tests are generated with `always_bls`
+#
 
 def run_sync_committees_progress_test(spec, state):
     first_sync_committee = state.current_sync_committee
@@ -41,6 +47,7 @@ def run_sync_committees_progress_test(spec, state):
 
 @with_all_phases_except([PHASE0])
 @spec_state_test
+@always_bls
 @with_configs([MINIMAL], reason="too slow")
 def test_sync_committees_progress_genesis(spec, state):
     # Genesis epoch period has an exceptional case
@@ -52,6 +59,7 @@ def test_sync_committees_progress_genesis(spec, state):
 
 @with_all_phases_except([PHASE0])
 @spec_state_test
+@always_bls
 @with_configs([MINIMAL], reason="too slow")
 def test_sync_committees_progress_not_genesis(spec, state):
     # Transition out of the genesis epoch period to test non-exceptional case
@@ -65,6 +73,7 @@ def test_sync_committees_progress_not_genesis(spec, state):
 @with_custom_state(balances_fn=misc_balances, threshold_fn=lambda spec: spec.EJECTION_BALANCE)
 @spec_test
 @single_phase
+@always_bls
 @with_configs([MINIMAL], reason="too slow")
 def test_sync_committees_progress_misc_balances(spec, state):
     yield from run_sync_committees_progress_test(spec, state)
--- a/tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_sync_committee_updates.py
+++ b/tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_sync_committee_updates.py
@@ -1,7 +1,11 @@
 from eth2spec.test.context import (
     spec_state_test,
+    spec_test,
     with_all_phases_except,
     with_configs,
+    with_custom_state,
+    single_phase,
+    misc_balances,
 )
 from eth2spec.test.helpers.constants import (
     PHASE0,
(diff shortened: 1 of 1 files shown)


tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_sync_committee_updates.py
```diff
@@ -36,7 +32,39 @@ def test_sync_committees_progress(spec, state):
...
     assert state.current_sync_committee == second_sync_committee
     assert state.next_sync_committee == third_sync_committee
+
+
+@with_all_phases_except([PHASE0])
+@spec_state_test
+@with_configs([MINIMAL], reason="too slow")
+def test_sync_committees_progress_genesis(spec, state):
+    # Genesis epoch period has an exceptional case
+    spec.get_current_epoch(state)
```
ralexstokes, line 47: looks like an extra line
```suggestion
```

tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_sync_committee_updates.py
```diff
@@ -36,7 +32,39 @@ def test_sync_committees_progress(spec, state):
...
+
+    yield from run_sync_committees_progress_test(spec, state)
+
+
+@with_all_phases_except([PHASE0])
+@spec_state_test
+@with_configs([MINIMAL], reason="too slow")
+def test_sync_committees_progress_not_genesis(spec, state):
+    # Transition out of the genesis epoch period to test non-exceptional case
+    start_slot_of_next_period = state.slot + spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD * spec.SLOTS_PER_EPOCH
```
ralexstokes, line 58: we do happen to have the condition `state.slot == 0` at this point but i'm not sure if that's a solid invariant of the `spec_state_test` framework -- if not, could be worth explicitly checking: `assert spec.get_curent_epoch(state) == spec.GENESIS_EPOCH`

ralexstokes, line 58: also this is only the "*start* slot of the next period" if `state.slot % spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD == 0` and i'm not sure if we want to make that assumption about the `state` we get from `spec_state_test`

tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_sync_committee_updates.py
```diff
@@ -13,20 +17,12 @@
...
-    assert current_epoch == 0
-
+def run_sync_committees_progress_test(spec, state):
     first_sync_committee = state.current_sync_committee
     second_sync_committee = state.next_sync_committee
 
-    slot_at_end_of_current_period = spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD * spec.SLOTS_PER_EPOCH - 1
-    transition_to(spec, state, slot_at_end_of_current_period)
+    end_slot_of_current_period = state.slot + spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD * spec.SLOTS_PER_EPOCH - 1
+    transition_to(spec, state, end_slot_of_current_period)
```
ralexstokes, line 25: this should be working bc `state.slot % spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD == 0` but i think it would break if it was any other value as you would cross over into the next sync committee period while still wanting to be within the current sync committee period.

the simple way is to just leverage our inputs, throw an `assert state.slot % spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD == 0` and call it a day.

i think a better option is to just compute what we are after:

```python
current_period = spec.get_current_epoch(state) // spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD
next_period = current_period + 1
next_period_start_epoch = next_period * spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD
next_period_start_slot = next_period_start_epoch * spec.SLOTS_PER_EPOCH
end_slot_of_currrent_period = next_period_start_slot - 1
```

ralexstokes (commented):
nice! we definitely want to extend w/ this test coverage.

left some comments calling your attention to invariants of the test but otherwise looks good!

michaelsproul:
Looks good.

I just noticed another issue related to `meta.yaml` that could possibly be fixed here too, cf. https://github.com/ethereum/eth2.0-specs/issues/2330#issuecomment-820974924

djrtwo:
Thanks @michaelsproul. addressed!

djrtwo:
Addressed your comments @ralexstokes. Ready for another review

tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_sync_committee_updates.py
```diff
@@ -17,11 +18,20 @@
 )
 
 
+#
+# Note:
+# Calculating sync committees requires pubkey aggregation, thus all tests are generated with `always_bls`
+#
+
```
ralexstokes, line 25: do you want a spare line here? 

```suggestion
```

ralexstokes (approved):
looks great! one comment about formatting but otherwise let's merge it

tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_sync_committee_updates.py
djrtwo: yes, I think so as it's a header comment for the whole file

PR #2337, opened 2021-04-19
Modify merge is_transition_block for readability

paulhauner:
This is subjective, but I found the existing code to be a little hard to read. I don't think it's immediately clear to the non-Python-guru how the order of operations should work here.

Consider this example:

```python
>>> not True and False
False
>>> not (True and False)
True
``` 

I'm not particularly concerned about this PR getting merged, it just feels less like a trap this way IMO.

### DIFF
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -149,6 +149,7 @@ def is_transition_completed(state: BeaconState) -> bool:
 def is_transition_block(state: BeaconState, block_body: BeaconBlockBody) -> bool:
     if is_transition_completed(state):
         return False
+    
     return block_body.execution_payload != ExecutionPayload()
 ```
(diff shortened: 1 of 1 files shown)


paulhauner:
Closing this since there's more of this `not` ordering going on in `process_execution_payload` and it seems like it's going to be required knowledge.

mkalinin:
Wdyt about anything like that:
```python
def is_transition_block(state: BeaconState, block_body: BeaconBlockBody) -> bool:
    transition_is_not_completed = not is_transition_completed(state)
    block_has_payload = block_body.execution_payload != ExecutionPayload()
    return transition_is_not_completed and block_has_payload
```

paulhauner:
> Wdyt about anything like that:

It does read very nicely!

PR #2343, opened 2021-04-20
update pyspec dir

ralexstokes:
it looks like the dir in the Makefile was not updated during some file tree reorg


Makefile
```diff
@@ -2,7 +2,7 @@ SPEC_DIR = ./specs
 SSZ_DIR = ./ssz
 TEST_LIBS_DIR = ./tests/core
 TEST_GENERATORS_DIR = ./tests/generators
-PY_SPEC_DIR = $(TEST_LIBS_DIR)/pyspec
+PY_SPEC_DIR = $(TEST_LIBS_DIR)/pyspec/eth2spec
```
protolambda, line 5: The outputs of the coverage report, test report, caches, etc. go into this directory I think. The `eth2spec` dir is a python *package*, so I think it shouldn't contain any such test output data.

protolambda (changes requested):

ralexstokes:
looks like this is fixed in #2346.

PR #2344, opened 2021-04-20
add always_bls for historical batch sanity test

### LINKED ISSUE #2335: Altair sanity blocks test missing BLS setting
The test located at `eth2.0-spec-tests/tests/minimal/altair/sanity/blocks/pyspec_tests/historical_batch` passes with real BLS and fails otherwise. I think it needs `bls_setting: 1` in its `meta.yaml` due to computing a new sync committee.

djrtwo:
fixes #2335

Because the historical batch and sync committee periods are equal in length, we must do bls processing during this epoch transition integration test to ensure committees are calculated properly.

Note: this is not the case for unit tests that only test `process_historical_roots_update`


hwwhww (approved):
LGTM

PR #2348, opened 2021-04-21
get_shard_proposer_index: Use slot instead of beacon_state.slot

Nashatyrev:
Seems we need to use `slot` instead of `beacon_state.slot` for the seed calculation. 
Else we would get different `shard_proposer_index` for the same slot+shard when calculating on top of different states

### DIFF
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -402,7 +402,7 @@ def get_shard_proposer_index(beacon_state: BeaconState, slot: Slot, shard: Shard
     """
     epoch = compute_epoch_at_slot(slot)
     committee = get_shard_committee(beacon_state, epoch, shard)
-    seed = hash(get_seed(beacon_state, epoch, DOMAIN_BEACON_PROPOSER) + uint_to_bytes(beacon_state.slot))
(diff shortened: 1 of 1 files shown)


djrtwo (approved):

specs/sharding/beacon-chain.md
```diff
@@ -402,7 +402,7 @@ def get_shard_proposer_index(beacon_state: BeaconState, slot: Slot, shard: Shard
     """
     epoch = compute_epoch_at_slot(slot)
     committee = get_shard_committee(beacon_state, epoch, shard)
-    seed = hash(get_seed(beacon_state, epoch, DOMAIN_BEACON_PROPOSER) + uint_to_bytes(beacon_state.slot))
+    seed = hash(get_seed(beacon_state, epoch, DOMAIN_BEACON_PROPOSER) + uint_to_bytes(slot))
```
protolambda, line 405: We should probably assert that the slot is of the current epoch or older (and not older than randao seed history). The usage of this function in this doc is safe currently, but we may use it elsewhere. The phase0 beacon-proposer variant uses `state.slot` because it only computes for the current slot, and thus does not have the same safety problem.

protolambda (approved):
Good catch, looks good to me. One minor comment about the `slot` usage, but I think we can address that as part of the refactoring to address the header-inclusion code (the unnecessary `candidates` list you mentioned in discord).

PR #2350, opened 2021-04-22
Sarding: missing argument in get_active_shard_count() call

Nashatyrev:
`get_active_shard_count` call misses `epoch` argument in `process_pending_headers`
Suppose it should be `get_previous_epoch(state)`


djrtwo (approved):

djrtwo:
Thanks!

PR #2351, opened 2021-04-22
Sharding cleanup: Remove obsolete var, reorder calculation 

Nashatyrev:
Some code clean up: 

- remove not used `full_committee_balance` var 
- move `full_committee` var calculation closer to its first usage


djrtwo (approved):
:)

PR #2352, opened 2021-04-23
Add `syncnets` data to Altair `MetaData`

### LINKED ISSUE #2347: Altair p2p MetaData should contain syncnets field
Currently, the `MetaData` field https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/p2p-interface.md#metadata contains only the `attnets` field to signal a change in the subscribed attestation subnets to our peers. 

We should similarly have the `syncnets` field in `MetaData` so we can signal changes in our subscribed sync committee subnets.

ralexstokes:
Fixes #2347.

### DIFF
--- a/specs/altair/p2p-interface.md
+++ b/specs/altair/p2p-interface.md
@@ -15,6 +15,7 @@ Altair adds new messages, topics and data to the Req-Resp, Gossip and Discovery 
 
   - [Warning](#warning)
 - [Modifications in Altair](#modifications-in-altair)
+  - [MetaData](#metadata)
   - [The gossip domain: gossipsub](#the-gossip-domain-gossipsub)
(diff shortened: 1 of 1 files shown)


pawanjay176:
This would also add a v2 version for `GetMetaData` response  in rpc right?

djrtwo:
> This would also add a v2 version for `GetMetaData` response in rpc right?

Yes, I think that's the simplest way to handle the content changes on the `MetaData` type here.

ralexstokes:
Added the "get metadata v2" req/resp endpoint -- the guide already discusses upgrading to v2 ahead of the fork so it seems that has already been covered

djrtwo (approved):
excellent. thank you!

PR #2292, opened 2021-03-30
(block, epoch)-fork choice

adiasg:
This PR contains a security fix for the finality delay attack outlined in [this paper](https://econcs.pku.edu.cn/wine2020/wine2020/Workshop/GTiB20_paper_8.pdf). 

## Finality Delay Attack
The execution of the attack involves temporary block withholding by the first few proposers in the epoch. At the end of the last withholding proposer's slot, all blocks are released. During slots in which blocks were withheld, honest attesters make FFG target votes for the last block of the previous epoch (which is "pulled up" to serve as the FFG checkpoint for the current epoch). After the blocks have been released, honest attesters make FFG target votes for the first block of the current epoch.
During each slot, `1 / SLOTS_PER_EPOCH` (`= 3.125%`) fraction of stake is attesting. So a 25% attacker needs to control the first 3 slots (`= ceil((33.33 - 25) / 3.125)`) of an epoch to successfully delay finality at the end of that epoch.

A brief description from the original paper:
image

The attack is possible because the fork choice allows for updating the head block, without any reorgs, in the following situation:
- the old head block points to an FFG target inconsistent with the one pointed to by the new head block, and
- there is more support for the old head block's FFG target than the new head block's FFG target.

This happens because the fork choice only calculates the LMD GHOST winner based on the support for each individual block. **_This PR prevents the attack by changing the block tree structure, and by also taking into account the support for the FFG target pointed to by each block._**

For example, the below block hierarchy will be transformed into a block tree in the following manner:
image

## Summary of Changes

- **Block Tree Structure**
  - Changed block tree graph so that nodes that point to inconsistent FFG targets are in separate branches. (Refer: [`BlockTreeNode`](https://github.com/ethereum/eth2.0-specs/pull/2292/files#diff-f58759020c0d7542f963e2b88fef66c2179d740beb5af45b7e2c90913477be11R84-R87) & [`add_block_tree_node`](https://github.com/ethereum/eth2.0-specs/pull/2292/files#diff-f58759020c0d7542f963e2b88fef66c2179d740beb5af45b7e2c90913477be11R387-R433))
- **`LatestMessage` Accounting**
[...]

PR #2353, opened 2021-04-23
Proposer LMD Score Boosting

adiasg:
Edit: Renamed from "Fix LMD balancing attack" to "Proposer LMD Score Boosting"

This PR implements a fix for the [LMD balancing attacks](https://arxiv.org/abs/2009.04987), as described [here](https://notes.ethereum.org/@vbuterin/lmd_ghost_mitigation).

## Proposer Score Boost

A new "proposer score boost" has been introduced, which boosts the LMD score of blocks that are received in a timely manner. The boost increases the score of the block by `committee_weight // 4` for that slot. If a block for slot `X` is received in the first `SECONDS_PER_SLOT // ATTESTATION_OFFSET_QUOTIENT` (= 4) seconds of the slot, then the block receives the boost for the duration of slot `X`. After slot `X` is over, the boost is removed and the usual LMD score calculations from only attestations are done. 

The boost is not applied to blocks that are not received in the first `SECONDS_PER_SLOT // ATTESTATION_OFFSET_QUOTIENT` seconds of their slots, and this PR does not affect LMD score calculations for such blocks.

lmd-proposer-score-fix

In the above example:
- block `B` is received in a timely manner and qualifies for the boost. `B`'s score is boosted for only slot `X`.
- block `C` is not received in a timely manner and does not qualify for the boost. 

## Summary of Changes
- Parameters:
  - Added new parameter `ATTESTATION_OFFSET_QUOTIENT`, which is used to specify the time at which attestations are produced within a slot
- `Store`:
  - Added new field `store.proposer_score_boost`, which is a special `LatestMessage` that stores the boost
- Fork Choice:
  - Updated `get_latest_attesting_balance` to take into account `store.proposer_score_boost`
  - Updated `on_block` & `on_tick` to appropriately set/reset `store.proposer_score_boost`
- Testing:
  - Added test for `store.proposer_score_boost` accounting
  - Updated existing tests to work correctly after introducing proposer score boosting


adiasg:
Note: First merge PR #2292 into `dev`, then merge this into `dev`

joachimneu:
Following the general approach of balancing, a recent attack dispenses with adversarial network delay and can be found in [Appendix I of this manuscript](https://arxiv.org/abs/2105.06075).

adiasg:
This fix makes the synchrony assumption of `latency < SECONDS_PER_SLOT // ATTESTATION_OFFSET_QUOTIENT` (= 4) seconds, and hence prevents the new variant of the attack under synchronous network conditions.

In general, this fix prevents attacks that rely on balancing different forks by utilizing the LMD fork choice score. This is because a few consecutive honest proposers are able to force a single winner from such balanced forks.

joachimneu:
If I am not mistaken, then [the analysis here](https://notes.ethereum.org/@vbuterin/lmd_ghost_mitigation) misses two aspects:
* The adversary only needs to release sway votes for honestly led slots, not for adversarially led slots.
* Adversarial votes can affect a 2x change in score, due to LMD.

If this is so, then an adversary controlling 1/9 = ~0.11 fraction of stake can attack the protocol with the fix for Wp = 1/4*W.

michaelsproul:
I think the boost also incentivises proposers to orphan late blocks which they've seen, which could be beneficial as late blocks continue to cause issues on mainnet.

In the case where the proposer for slot `X` publishes their block late, the proposer at slot `X + 1` might want to orphan it, even if they've received it and their fork choice selects it as the head. If block `X` has `<W/4` votes then the proposer for slot `X + 1` can propose a (timely) block of their own with the same parent as block `X`, which will then outrank block `X` thanks to the `W/4` boost. If they do this they stand to gain extra rewards from the inclusion of attestations from slot `X`.

I'd argue that this ability to aggressively orphan late blocks is actually a _good_ thing because it sends a clear signal to the previous proposer that their block was low quality. At the moment we often see late blocks without many votes making it into the canonical chain because they are seen by the next proposer before the start of the next slot. In the pathological case consider a block published 8 seconds into the slot which receives no attestations but still arrives at all nodes before the next slot.

Late blocks hamper consensus and penalise attesters as they invalidate head (and sometimes target) votes for slot `X`, due to honest attesters voting for `X`'s parent. If we orphan them then at least we improve agreement on target (if `X % 32 == 0`), and reward the correct-target attesters. Post-Altair there's no way to salvage the head vote rewards for the slot `X` attesters, as they either voted for a skipped slot and failed the timeliness condition (if block `X` is orphaned), or voted for the wrong head (if block `X` is not orphaned).

My hope is that if clients adopt the orphaning logic then it will pressure late block publishers into getting their act together. I suspect that at the moment their metrics don't detect the late blocks they're publishing because late blocks are still rewarded when they make it into the chain. The philosophical question is whether or not it's worth explicitly going against what fork choice says the head should be in order to implement the aggressive orphaning, or whether fork choice should be extended with a notion of a "proposer head" that allows us to count this behaviour as honest.

adiasg:
Closed in favor of #2730. This PR was based on # #2292 and had major conflicts with `dev`.
#2730 now does not have the (block, epoch)-fork choice from #2292.

PR #2355, opened 2021-04-23
Fix: `upgrade_to_altair` should use `phase0.get_current_epoch`

ericsson49:
`altair.BeaconState` class is a new class in `altair`, which doesn't extend `phase0.BeaconState`.
That means `get_current_epoch` has different signatures in `phase0` and `altair`.
Thus `upgrade_to_altair` should use `phase0.get_current_epoch` for an instance of `phase0.BeaconState` to avoid typing rules violation.


protolambda (approved):
Good catch

PR #2358, opened 2021-04-23
More typos fixes in `sharding`

ericsson49:
Fixed typos in the methods:
- `process_shard_header`:
   - `body_summary` missed when accessing`ShardBlobBodySummary.beacon_block_root`
- `process_pending_headers`
   - `c` variable name used instead of `header`
- `charge_confirmed_header_fees`:
   - `get_shard_proposer` used instead of `get_shard_proposer_index`
- `reset_pending_headers`
   - `SLOTS_IN_EPOCH` used instead of `SLOTS_PER_EPOCH`
   - `state` missed in `get_committee_count_per_slot(next_epoch)` call
- `process_epoch`
   - `process_confirmed_header_fees` called instead of `charge_confirmed_header_fees`


protolambda (approved):
Thank you! We will make this part of the spec executable in a week or two, and run some tests and linting, to try and catch things like this.

PR #2359, opened 2021-04-24
Fix: pass Set instead of Sequence to get_total_balance

ericsson49:
Fixed typing problem in `process_pending_headers` and in `update_pending_votes`: `Sequence[ValidatorIndex]` passed instead of `Set[ValidatorIndex]`.
In both cases the collection of indices passed to `get_total_balance` originates from `get_beacon_committee`: either filtered (`process_pending_headers`) or "as is" (`update_pending_votes`).
That means there should be no duplicates, if there is no duplicates in the `get_beacon_committee` result. So, converting to `Set` shouldn't affect result.

### DIFF
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -539,7 +539,7 @@ def update_pending_votes(state: BeaconState, attestation: Attestation) -> None:
     participants = get_attesting_indices(state, attestation.data, pending_header.votes)
     participants_balance = get_total_balance(state, participants)
     full_committee = get_beacon_committee(state, attestation.data.slot, attestation.data.index)
-    full_committee_balance = get_total_balance(state, full_committee)
+    full_committee_balance = get_total_balance(state, set(full_committee))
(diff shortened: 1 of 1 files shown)


specs/sharding/beacon-chain.md
```diff
@@ -539,7 +539,7 @@ def update_pending_votes(state: BeaconState, attestation: Attestation) -> None:
     participants = get_attesting_indices(state, attestation.data, pending_header.votes)
     participants_balance = get_total_balance(state, participants)
     full_committee = get_beacon_committee(state, attestation.data.slot, attestation.data.index)
-    full_committee_balance = get_total_balance(state, full_committee)
+    full_committee_balance = get_total_balance(state, set(full_committee))
```
protolambda, line 542: Maybe we should introduce some type for an ordered-set in the spec. The committee members each have an index, and are unique, at the same time. The total-balance computation is order-independent though, so a `set` is fine here.

protolambda (approved):
LGTM, thank you.

PR #2361, opened 2021-04-27
Fix typing problems: convert to CommitteeIndex

ericsson49:
`get_beacon_committee` and `compute_shard_from_committee_index` expect `CommitteeIndex` as their last parameter, however, other types, like `int` or `Shard` has been passed.

The PR contains a suggested fix, which just converts to `CommitteeIndex` (and additionally `shard_index` to `Shard` too), which shouldn't affect runtime behavior, just make type-checker happy.

However, the typing error may indicate there is a bug actually, i.e. it may be the case that `CommitteeIndex(shard)` is not an appropriate way to obtain a `CommitteeIndex` from `shard`. Actually, [process_shard_header](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/sharding/beacon-chain.md#process_shard_header) uses [compute_committee_index_from_shard](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/sharding/beacon-chain.md#compute_committee_index_from_shard) to obtain `CommitteeIndex` from a `shard`.

I don't have enough knowledge to disambiguate which is the correct way here.

### DIFF
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -719,7 +719,8 @@ def charge_confirmed_header_fees(state: BeaconState) -> None:
     )
     previous_epoch_start_slot = compute_start_slot_at_epoch(get_previous_epoch(state))
     for slot in range(previous_epoch_start_slot, previous_epoch_start_slot + SLOTS_PER_EPOCH):
-        for shard in range(SHARD_COUNT):
+        for shard_index in range(SHARD_COUNT):
+            shard = Shard(shard_index)
             confirmed_candidates = [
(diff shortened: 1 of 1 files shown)


djrtwo (commented):
Ah, so when going from shard to committee_index, you need to use the `compute` function. So there was incorrect usage in a few places. yay type checker

ericsson49:
Thanx @djrtwo !
I updated code. In the `reset_pending_headers` case, `shard` is computed from from `committee_index`, thus I believe one don't need to compute `committee_index` once again.

djrtwo (approved):
looks good!

Issue #1409, opened 2019-09-19
Unclear atomicity semantics of the beacon chain spec


ericsson49:
In the [beacon chain specs](https://github.com/ethereum/eth2.0-specs) state updates are sometimes intertwined with assertions within a function. An executable python spec is also constructed from such code fragments.
The executability of the specs suggests such functions should not be treated as atomic state updates (assuming python semantics). E.g. if a state is partially updated and later an assertion is failed, the partial state update should be kept.
However, such view contradicts the intention expressed in other functions.

For example, consider the [fork choice/on_block](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_fork-choice.md#on_block) function.
A block store is updated with a new block.
Some checks are preformed after (including `state_transition` ones).
And then, a block_state store is updated with the corresponding block state.

If any of in-between assertions failed, then store.blocks would contain the block, while store.block_states wouldn't.
That seems to be a problem. For example, if an attestation referencing the block arrives later, the [on_attestation](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_fork-choice.md#on_attestation) is called. Which first checks if the referenced block is present (which will succeed) and later accesses the block's state, which is absent.
So, the piece of code implicitly assumes the changes made by `on_block` are atomic, i.e. either both block and it's state are stored or none of them (or maybe some weaker form like, if a block is stored then it's state also should).
Which seems pretty reasonable, since ensuring consistency of block/state updates with python code would be more verbose and makes the specs more clumsy.

Besides storage updates, an analogous problem exists with beacon state updates. For example, in [process_attestation](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_beacon-chain.md#attestations) function, either `state.current_epoch_attestations` or `state.previous_epoch_attestations` is updated, followed by some assertion checks. If any of the assertion failed, the update will be kept.

While it's a minor issue from a human being point of view, it's not so from a machine's perspective. For example, we would like to develop a test suit to test fork choice against the specs. In case of negative tests, it becomes important, so that we can base our tests on a stable ground.

Another point is that we'd like eventually to use formal methods like bounded model checking or formal proofs to check/verify properties of the beacon chain specs.

Based on the above, we propose:
* to add clarification to the beacon chain specs, regarding atomicity semantics of the fork choice functions (on_block, on_attestation). I.e. should multiple store updates be performed in all or none fashion. Or the python execution semantics are assumed.
[...]

PR #2366, opened 2021-04-29
Replace `assert` with custom `require` method

### LINKED ISSUE #1797: Use of `assert` to validate method arguments

hwwhww:
Replace #2045
Fix #1797

Starting with `phase0/beacon-chain.md`

### Proposed solution
- Replace `assert <condition>` with custom `require(condition: bool, message: str=None)` function to signal that the execution is INVALID. Implementers can define their custom error handling method.
- Move some comments to the `message` field.
- Note that it should be *no* substantive change.

/cc @ericsson49

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -595,13 +595,13 @@ class SignedBeaconBlockHeader(Container):
 #### `require`
 
 ```python
-def require(condition: bool, message: str=None) -> None:
+def require(condition: bool, message: str='') -> None:
     """
     Check if the given condition is satisfied.
     If not, raise an exception to signal that it's INVALID.
     """
     if not condition:
-        raise ValidationError(str(condition) if message is None else message)
+        raise ValidationError(message)
 ```
 
 **Note**: The `message` value is just for debugging the executable spec. The error handling details are implementation specific.
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -54,6 +54,8 @@
     - [`SignedBeaconBlock`](#signedbeaconblock)
     - [`SignedBeaconBlockHeader`](#signedbeaconblockheader)
 - [Helper functions](#helper-functions)
+  - [Validation](#validation)
+    - [`require`](#require)
   - [Math](#math)
     - [`integer_squareroot`](#integer_squareroot)
     - [`xor`](#xor)
@@ -587,6 +589,23 @@ class SignedBeaconBlockHeader(Container):
 
 *Note*: The definitions below are for specification purposes and are not necessarily optimal implementations.
 
+
+### Validation
+
+#### `require`
+
+```python
+def require(condition: bool, message: str=None) -> None:
+    """
+    Check if the given condition is satisfied.
+    If not, raise an exception to signal that it's INVALID.
+    """
+    if not condition:
+        raise ValidationError(str(condition) if message is None else message)
+```
+
+**Note**: The `message` value is just for debugging the executable spec. The error handling details are implementation specific.
+
 ### Math
 
 #### `integer_squareroot`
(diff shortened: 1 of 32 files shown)


specs/phase0/beacon-chain.md
```diff
@@ -563,6 +565,23 @@ class SignedBeaconBlockHeader(Container):
 
 *Note*: The definitions below are for specification purposes and are not necessarily optimal implementations.
 
+
+### Validation
+
+#### `require`
+
+```python
+def require(condition: bool, message: str=None) -> None:
```
hwwhww, line 574: Strictly speaking, it should be `Optional[str]` in Python:

```suggestion
from typing import Optional 
def require(condition: bool, message: Optional[str]=None) -> None:
```

- Python `Optional` might be confusing for the readers. Although `Optional` has been used in validator.md and das-core.md anyway...
- mypy may raise an error if we use a stricter config

I'm fine either way for now.

hwwhww:
@djrtwo

I started with `phase0/beacon-chain.md`. I'd love to get a green light before working on other spec files. 🙏

specs/phase0/beacon-chain.md
CarlBeek: We could force inplicit optionals (https://mypy.readthedocs.io/en/stable/config_file.html#none-and-optional-handling) in the `linter.ini`

hwwhww:
All `assert`s under `specs` folder are replaced with `require`s.
I updated each file per commit, but please squash them when merging it.

samlaf:
Curious what's the latest on this? Why do the specs still contain asserts instead of requires?
And how is the dafny team going around this: https://github.com/ethereum/consensus-specs/issues/1797 ?
Are they just using their own versions of the specs for formal verification?

ericsson49:
> And how is the dafny team going around this: #1797 ? Are they just using their own versions of the specs for formal verification?

Can't say I'm a member of the Dafny team, but I'm working on a transpiler from a subset of Python to Dafny.
I believe "using their own versions of the specs for formal verification" is the correct state of affairs.
My goal is to change this so that specs versions used for formal verification are obtained via a thoroughly tested mechanized translation (i.e. the transpiler from a subset of Python to Dafny).
I personally translate `assert` to `pyassert` method in Dafny, which is similar to the `requires`, with the "exception" that there is only one kind of `Exception` in Dafny world right now, which doesn't take any argument. Such restriction may seem artificial, but I don't think we really need more expressive exceptions right now.
However, it's more complicated in practice. For "functional pure" translation, `assert` is more like "abort transaction", i.e. any heap object update that happened before raising an exception is lost.
It may seem strange, but the idea is that "imperative impure" translation should stay as close as possible to Python code - which is interpreted as a reference implementation, while the "functional pure" version should be as close as possible to the intended specification.
The proof obligations should ensure that both variants agree:
- on whether an exception should be raised or not
- if no exception is raised, on respective outcomes (i.e. returned values plus heap object updates)

See also the discussion in #1409 . It includes @franck44 's comment on how they deal with `assert` in their Dafny code.

Hope that helps.

PR #2367, opened 2021-04-30
[Sharding] Fix unspecified SHARD_COUNT and DOMAIN_SHARD_HEADER constants 

### LINKED ISSUE #2357: DOMAIN_SHARD_HEADER is missing in sharding phase spec
[process_shard_header](https://github.com/ethereum/eth2.0-specs/blob/dcb013ccf2cd9c384dedc73ebb65dc034e6f6e78/specs/sharding/beacon-chain.md#process_shard_header) uses `DOMAIN_SHARD_HEADER`, which is not defined in `sharding` (or prior) specs.

Nashatyrev:
- Replace unspecified `DOMAIN_SHARD_HEADER` const with `DOMAIN_SHARD_PROPOSER`
- Replace unspecified `SHARD_COUNT` const with `get_active_shard_count(previous_epoch)` in `charge_confirmed_header_fees`
- Replace unspecified `SHARD_COUNT` with `MAX_SHARDS` when clearing `grandparent_epoch_confirmed_commitments` with default values

Fix #2357

### DIFF
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -670,9 +670,8 @@ def process_pending_headers(state: BeaconState) -> None:
 
     previous_epoch = get_previous_epoch(state)
     previous_epoch_start_slot = compute_start_slot_at_epoch(previous_epoch)
-    active_shard_count = get_active_shard_count(state, previous_epoch)
     for slot in range(previous_epoch_start_slot, previous_epoch_start_slot + SLOTS_PER_EPOCH):
-        for shard_index in range(active_shard_count):
+        for shard_index in range(get_active_shard_count(state, previous_epoch)):
             shard = Shard(shard_index)
             # Pending headers for this (slot, shard) combo
             candidates = [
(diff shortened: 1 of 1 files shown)


specs/sharding/beacon-chain.md
```diff
@@ -718,9 +718,10 @@ def charge_confirmed_header_fees(state: BeaconState) -> None:
         get_active_shard_count(state, get_current_epoch(state))
         * SLOTS_PER_EPOCH * GASPRICE_ADJUSTMENT_COEFFICIENT
     )
-    previous_epoch_start_slot = compute_start_slot_at_epoch(get_previous_epoch(state))
+    previous_epoch = get_previous_epoch(state)
+    previous_epoch_start_slot = compute_start_slot_at_epoch(previous_epoch)
```
protolambda, line 722: Good fix, but this reminds me we should limit some of these methods to epoch 1+, epoch 0 doesn't work quite as expected. The previous epoch var would be clipped to 0, and then it tries to confirm current-epoch shard headers.

protolambda (approved):
LGTM. I'll review the function usage for the epoch 0 edge case and put that in a separate PR.

PR #2370, opened 2021-05-01
Simplify sync committee aggregates

ralexstokes:
The current Altair spec provides multiple partitions of the full sync committee with the idea being that light clients could follow some subset of their choosing to enable a trade-off between security and performance.

However, to efficiently implement this feature the `BeaconBlock` definition in Altair needs to support a distinct (aggregate) signature for each partition stored in the `BeaconState`. And currently we only have one such signature.

To reconcile this discrepancy, we can either add multiple signatures to the `BeaconBlock` (along w/ the corresponding increase in consensus complexity) or we can omit the in-protocol aggregation. This PR presents the latter solution and I'd argue it is a better trade-off at this point in time given the relative maturity of Altair client implementations.

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -302,6 +302,14 @@ def get_sync_committee_indices(state: BeaconState, epoch: Epoch) -> Sequence[Val
 def get_sync_committee(state: BeaconState, epoch: Epoch) -> SyncCommittee:
     """
     Return the sync committee for a given ``state`` and ``epoch``.
+
+    ``SyncCommittee`` contains an aggregate pubkey that enables
+    resource-constrained clients to save some computation when verifying
+    the sync committee's signature.
+
+    ``SyncCommittee`` can also contain duplicate pubkeys, when ``get_sync_committee_indices``
+    returns duplicate indices. Implementations must take care when handling
+    optimizations relating to aggregation and verification in the presence of duplicates.
     """
     indices = get_sync_committee_indices(state, epoch)
     pubkeys = [state.validators[index].pubkey for index in indices]
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -409,7 +409,7 @@ def get_inactivity_penalty_deltas(state: BeaconState) -> Tuple[Sequence[Gwei], S
 
 #### Modified `slash_validator`
 
-*Note*: The function `slash_validator` is modified to use `MIN_SLASHING_PENALTY_QUOTIENT_ALTAIR` 
+*Note*: The function `slash_validator` is modified to use `MIN_SLASHING_PENALTY_QUOTIENT_ALTAIR`
 and use `PROPOSER_WEIGHT` when calculating the proposer reward.
 
 ```python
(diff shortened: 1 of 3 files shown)


specs/altair/beacon-chain.md
```diff
@@ -392,7 +390,7 @@ def get_inactivity_penalty_deltas(state: BeaconState) -> Tuple[Sequence[Gwei], S
 
 #### Modified `slash_validator`
 
-*Note*: The function `slash_validator` is modified to use `MIN_SLASHING_PENALTY_QUOTIENT_ALTAIR` 
+*Note*: The function `slash_validator` is modified to use `MIN_SLASHING_PENALTY_QUOTIENT_ALTAIR`
```
ralexstokes, line 393: just formatting -- my editor found some whitespace

specs/altair/beacon-chain.md
```diff
@@ -212,7 +211,7 @@ class SyncAggregate(Container):
 ```python
 class SyncCommittee(Container):
     pubkeys: Vector[BLSPubkey, SYNC_COMMITTEE_SIZE]
-    pubkey_aggregates: Vector[BLSPubkey, SYNC_COMMITTEE_SIZE // SYNC_PUBKEYS_PER_AGGREGATE]
+    aggregate_pubkey: BLSPubkey
```
hwwhww, line 214: Is `aggregate_pubkey` field necessary? It seems that the validator / light client can compute it in runtime. If we remove `aggregate_pubkey`, we can replace `SyncCommittee` container with just `sync_committee_pubkeys: Vector[BLSPubkey, SYNC_COMMITTEE_SIZE]`.

specs/altair/beacon-chain.md
hwwhww: Pedantic suggestion 😄:

```suggestion
*Note*: The function `slash_validator` is modified to use `MIN_SLASHING_PENALTY_QUOTIENT_ALTAIR`.
```

specs/altair/beacon-chain.md
ralexstokes: the light client could compute the aggregate at runtime but this enables an optimization where the light client can _subtract_ pubkeys from the set in `pubkeys` against the `aggregate_pubkey` with the idea being they can save a bunch of BLS computation in the expected case (i.e. fairly full sync committee participation)

for example, if we get 80% participation in a given sync committee signature, the light client can simply "unaggregate" 20% of the pubkeys, rather than have to aggregate 80% of them. this savings is particularly important for resource-constrained environments like on-chain bridges etc

specs/altair/beacon-chain.md
ralexstokes: oh actually this comment underflows and does terminate in a `.` on the next line -- i'll rollback this change

specs/altair/beacon-chain.md
hwwhww: > on-chain bridges

Ah, that's fair. 😄

BTW, since `get_sync_committee_indices` might return duplicate indices, the aggregate pubkey may include duplicate pubkeys. Implementers have to detect the duplicate before subtracting the pubkeys. Maybe worth noting the potential duplication that at `get_sync_committee` docstring (like `get_sync_committee_indices` docstring).

djrtwo (approved):
looks good to me!

agreed that we could include a bit more in the docstring about usage and gotchas on the aggregate

ralexstokes:
@hwwhww @djrtwo added some notes about duplicate pubkeys -- light clients should start with a bitfield and just blindly unaggregate based on participation so i'm not too worried about bugs here, but more documentation never hurts

hwwhww (approved):
LGTM!

djrtwo (approved):

PR #2373, opened 2021-05-03
Altair: carry-over prev epoch participation

### LINKED ISSUE #2314: Participation rewards are skipped for epoch prior to Altair upgrade
The [`upgrade_to_altair`](https://github.com/ethereum/eth2.0-specs/blob/v1.1.0-alpha.3/specs/altair/fork.md#upgrading-the-state) function which transitions the chain from phase0 to Altair initialises the participation bits in `previous_epoch_participation` to zero, which discards the information in the phase0 state's `previous_epoch_attestations`. This causes any validator who attested correctly in the epoch prior to the fork and had their attestation included on chain to be penalised as if they did not attest. The first epoch transition after the fork will consult `previous_epoch_participation` and find it mostly empty.

One solution would be to try to pack the prior epoch attestations into the blocks of the fork epoch. However, this doesn't interact well at all with the rewards for timeliness. Even if an attester was prompt in getting their attestation on chain in the previous epoch, by the time the fork epoch occurs they'll be too late for the timely head or source rewards (1 and 5 slots, respectively) and probably too late for the target reward (32 slots). Further, these late attestations would be competing with current epoch attestations for block space.

Alternatively, the `upgrade_to_altair` function could perform a translation of the data in `previous_epoch_attestations` so that rewards are computed correctly for the attestations that already made it on chain. I think we could afford to be generous if computing the participation bits precisely is too much effort, although I don't see any issues that would prevent us from doing it exactly.

michaelsproul:
Closes #2314

This is a (draft) PR implementing the translation of phase0 participation information into Altair participation flags at the fork, in order to prevent loss of information (and rewards).

It factors out a function from `process_attestation` which I had already factored out in Lighthouse anyway for use in the op pool (being able to calculate attestation flag bits independent of processing is nice).

Apart from that it adds a small function called `translate_participation`, which acts on the pre state's `previous_epoch_attestations`. I initially thought it should translate `current_epoch_attestations` but the lists have already been swapped by `process_participation_record_updates` at the time of the upgrade.

I'll start working on tests shortly. I think the main risk is covering all of the possible attestation behaviours, so I'll look into whether it's possible to reuse generators from the attestation tests.

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -37,6 +37,7 @@
     - [`get_base_reward_per_increment`](#get_base_reward_per_increment)
     - [`get_base_reward`](#get_base_reward)
     - [`get_unslashed_participating_indices`](#get_unslashed_participating_indices)
+    - [`get_attestation_participation_flag_indices`](#get_attestation_participation_flag_indices)
     - [`get_flag_index_deltas`](#get_flag_index_deltas)
     - [Modified `get_inactivity_penalty_deltas`](#modified-get_inactivity_penalty_deltas)
   - [Beacon state mutators](#beacon-state-mutators)
@@ -352,6 +353,37 @@ def get_unslashed_participating_indices(state: BeaconState, flag_index: int, epo
     return set(filter(lambda index: not state.validators[index].slashed, participating_indices))
 ```
 
+#### `get_attestation_participation_flag_indices`
+
+```python
+def get_attestation_participation_flag_indices(state: BeaconState,
+                                               data: AttestationData,
+                                               inclusion_delay: uint64) -> Sequence[int]:
+    """
+    Return the flag indices that are satisfied by an attestation.
+    """
+    if data.target.epoch == get_current_epoch(state):
+        justified_checkpoint = state.current_justified_checkpoint
+    else:
+        justified_checkpoint = state.previous_justified_checkpoint
+
+    # Matching roots
+    is_matching_source = data.source == justified_checkpoint
+    is_matching_target = is_matching_source and data.target.root == get_block_root(state, data.target.epoch)
+    is_matching_head = is_matching_target and data.beacon_block_root == get_block_root_at_slot(state, data.slot)
+    assert is_matching_source
+
+    participation_flag_indices = []
+    if is_matching_source and inclusion_delay <= integer_squareroot(SLOTS_PER_EPOCH):
+        participation_flag_indices.append(TIMELY_SOURCE_FLAG_INDEX)
+    if is_matching_target and inclusion_delay <= SLOTS_PER_EPOCH:
+        participation_flag_indices.append(TIMELY_TARGET_FLAG_INDEX)
+    if is_matching_head and inclusion_delay == MIN_ATTESTATION_INCLUSION_DELAY:
+        participation_flag_indices.append(TIMELY_HEAD_FLAG_INDEX)
+
+    return participation_flag_indices
+```
+
 #### `get_flag_index_deltas`
 
 ```python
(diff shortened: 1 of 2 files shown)


specs/altair/beacon-chain.md
```diff
@@ -37,6 +37,7 @@
     - [`get_base_reward_per_increment`](#get_base_reward_per_increment)
     - [`get_base_reward`](#get_base_reward)
     - [`get_unslashed_participating_indices`](#get_unslashed_participating_indices)
+    - [`get_attestation_participation_flags`](#get_attestation_participation_flags)
```
hwwhww, line 40:
```suggestion
    - [`get_attestation_participation_flag_indices`](#get_attestation_participation_flag_indices)
```

specs/altair/beacon-chain.md
```diff
@@ -335,6 +336,37 @@ def get_unslashed_participating_indices(state: BeaconState, flag_index: int, epo
     return set(filter(lambda index: not state.validators[index].slashed, participating_indices))
 ```
 
+#### `get_attestation_participation_flags`
```
hwwhww, line 339:
```suggestion
#### `get_attestation_participation_flag_indices`
```

specs/altair/beacon-chain.md
```diff
@@ -335,6 +336,37 @@ def get_unslashed_participating_indices(state: BeaconState, flag_index: int, epo
...
+
+```python
+def get_attestation_participation_flag_indices(state: BeaconState,
```
hwwhww, line 342: I like this refactoring! IMHO we can apply it even if the `translate_participation` solution doesn't get included.

specs/altair/fork.md
```diff
@@ -41,6 +41,21 @@ Note that for the pure Altair networks, we don't apply `upgrade_to_altair` since
 After `process_slots` of Phase 0 finishes, if `state.slot % SLOTS_PER_EPOCH == 0` and `compute_epoch_at_slot(state.slot) == ALTAIR_FORK_EPOCH`, an irregular state change is made to upgrade to Altair.
 
 ```python
+def translate_participation(pending_attestations: Sequence[PendingAttestation], post: BeaconState) -> None:
```
hwwhww, line 44: To align with other "[Beacon state mutators](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/beacon-chain.md#beacon-state-mutators)", I suggest passing:
```suggestion
def translate_participation(state: BeaconState, pending_attestations: Sequence[PendingAttestation]) -> None:
```
Reasons:
1. This function modifies the given mutable parameter.
2. There is no `pre` state here.

hwwhww (commented):
It looks that this solution doesn't increase too much complexity 👍

specs/altair/fork.md
michaelsproul: Good idea, updated!

specs/altair/beacon-chain.md
michaelsproul: Thanks for the review!

djrtwo:
This was discussed on the latest eth2 implementers call, and there was a decent amount of pushback on the complexity and testing tradeoff.

Looking for more engineering input here before a merge or close

ajsutton:
This looks pretty straight forward to implement to me and particularly with the refactoring to use the `get_attestation_participation_flag_indices` method for both this transition and normal operation the additional testing load seems very low. The refactoring probably aids testability. So I don't have a problem with this going in.

I'm not overly concerned about validators being penalised for an epoch - the optics aren't great but it's likely to be forgotten pretty quickly.  But given we already have to do a special transition to activate Altair and this is a pretty simple change I'd lean in favour of including it.

terencechain:
I'm also leaning towards inclusion. Reasons:
- I like the refactoring of `get_attestation_participation_flag_indices`, that portion is already tested in unit tests and spec tests
- We just need new spec tests on `upgrade_to_altair` and further practice + verification with this on live testnet

With that said, I'm also not concerned about stakers getting penalized for one epoch, the social implication won't be good, but they will be forgotten rather quickly

djrtwo:
I'm making a PR to this PR handling the merge conflict and adding tests

PR #2374, opened 2021-05-04
Altair review

djrtwo:
* add some missing beacon-chain function comments
* Reorder attestation logic to be in order [source, target, head] throughout
* define `is_{attestation_field}` as having to have the parent `is_{attestation_field}` correct, thus enforcing chain coherence any time those booleans are used
* add sync committee test for empty participants

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -328,7 +328,8 @@ def get_base_reward(state: BeaconState, index: ValidatorIndex) -> Gwei:
     Return the base reward for the validator defined by ``index`` with respect to the current ``state``.
 
     Note: A validator can optimally earn one base reward per epoch over a long time horizon.
-    This takes into account both per-epoch (e.g. attestation) and intermittent duties (e.g. block proposal and sync committees)
+    This takes into account both per-epoch (e.g. attestation) and intermittent duties (e.g. block proposal
+    and sync committees).
     """
     increments = state.validators[index].effective_balance // EFFECTIVE_BALANCE_INCREMENT
     return Gwei(increments * get_base_reward_per_increment(state))
@@ -339,7 +340,7 @@ def get_base_reward(state: BeaconState, index: ValidatorIndex) -> Gwei:
 ```python
 def get_unslashed_participating_indices(state: BeaconState, flag_index: int, epoch: Epoch) -> Set[ValidatorIndex]:
     """
-    Return the set of validator indicies that are both active and unslashed for the given ``flag_index`` and ``epoch``.
+    Return the set of validator indices that are both active and unslashed for the given ``flag_index`` and ``epoch``.
     """
     assert epoch in (get_previous_epoch(state), get_current_epoch(state))
     if epoch == get_current_epoch(state):
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -237,6 +237,9 @@ def eth2_fast_aggregate_verify(pubkeys: Sequence[BLSPubkey], message: Bytes32, s
 
 ```python
 def get_flag_indices_and_weights() -> Sequence[Tuple[int, uint64]]:
+    """
+    Return paired tuples of participation flag indices along with associated incentivization weights.
+    """
     return (
         (TIMELY_HEAD_FLAG_INDEX, TIMELY_HEAD_WEIGHT),
         (TIMELY_SOURCE_FLAG_INDEX, TIMELY_SOURCE_WEIGHT),
@@ -248,6 +251,9 @@ def get_flag_indices_and_weights() -> Sequence[Tuple[int, uint64]]:
 
 ```python
 def add_flag(flags: ParticipationFlags, flag_index: int) -> ParticipationFlags:
+    """
+    Return a new ``ParticipationFlags`` adding ``flag_index`` to ``flags``.
+    """
     flag = ParticipationFlags(2**flag_index)
     return flags | flag
 ```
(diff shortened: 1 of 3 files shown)


specs/altair/beacon-chain.md
```diff
@@ -455,22 +470,22 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
         justified_checkpoint = state.previous_justified_checkpoint
 
     # Matching roots
-    is_matching_head = data.beacon_block_root == get_block_root_at_slot(state, data.slot)
     is_matching_source = data.source == justified_checkpoint
-    is_matching_target = data.target.root == get_block_root(state, data.target.epoch)
+    is_matching_target = is_matching_source and data.target.root == get_block_root(state, data.target.epoch)
+    is_matching_head = is_matching_target and data.beacon_block_root == get_block_root_at_slot(state, data.slot)
```
protolambda, line 475: The condition changed from just a head block-root check, to a is-target + head block root check. And the if statements below changed to then not do the target check again. I think we were testing good head votes with bad target votes, but if not, I'll try add this after the specs call.

specs/altair/beacon-chain.md
```diff
@@ -455,22 +470,22 @@ def process_attestation(state: BeaconState, attestation: Attestation) -> None:
...
 
     # Participation flag indices
     participation_flag_indices = []
-    if is_matching_head and is_matching_target and state.slot == data.slot + MIN_ATTESTATION_INCLUSION_DELAY:
-        participation_flag_indices.append(TIMELY_HEAD_FLAG_INDEX)
     if is_matching_source and state.slot <= data.slot + integer_squareroot(SLOTS_PER_EPOCH):
         participation_flag_indices.append(TIMELY_SOURCE_FLAG_INDEX)
     if is_matching_target and state.slot <= data.slot + SLOTS_PER_EPOCH:
         participation_flag_indices.append(TIMELY_TARGET_FLAG_INDEX)
+    if is_matching_head and state.slot == data.slot + MIN_ATTESTATION_INCLUSION_DELAY:
```
protolambda, line 487: Yay for source>target>head ordering of logic.

specs/altair/beacon-chain.md
```diff
@@ -311,10 +320,16 @@ def get_base_reward_per_increment(state: BeaconState) -> Gwei:
...
 
-*Note*: The function `get_base_reward` is modified with the removal of `BASE_REWARDS_PER_EPOCH`.
+*Note*: The function `get_base_reward` is modified with the removal of `BASE_REWARDS_PER_EPOCH` and the use of increment based accounting.
 
 ```python
+    """
+    Return the base reward for the validator defined by ``index`` with respect to the current ``state``.
+
+    Note: A validator can optimally earn one base reward per epoch over a long time horizon.
```
protolambda, line 330: We should make a note of the difference with phase0. It's still the same amount of ETH, but the base reward naming changed from 4 (`BASE_REWARDS_PER_EPOCH`) to just 1 big one per epoch + weight logic.

protolambda: Also, the precision changed:

Phase0:
```
effective_balance * BASE_REWARD_FACTOR // integer_squareroot(total_balance) // BASE_REWARDS_PER_EPOCH
```

Altair:
```
EFFECTIVE_BALANCE_INCREMENT * BASE_REWARD_FACTOR // integer_squareroot(get_total_active_balance(state))
```

Napkin math: assuming an effective balance of ~32, that's 5 bits of extra precision that don't get rounded down. Then it divides by `BASE_REWARDS_PER_EPOCH = 4`, 2 bits less, but it repeats 4 times, so the effect may not be as big there.

The higher the precision, the less the base reward gets "rounded down". So those ~5 bits = 32 gwei (pessimistic maximum) lost per validator per attestation when forking to Altair. In the grand scheme of things it's a tiny rounding thing, but it's about validator rewards, so I think we want to get it right.

specs/altair/beacon-chain.md
djrtwo: This is not the case. See here:

https://github.com/ethereum/eth2.0-specs/pull/2374/files#diff-76363249532da52344c30ccc6864859e2a74ad83130d0d640033cbbe4ff0a5beL468

We used to write `is_matching_head and is_matching_target` whereas this changes `is_matching_head` to include `is_matching_target` to remove the footgun of using head alone somewhere by accident

djrtwo: EDIT
I misread your statement. You are correct in your assessment and we almost certainly have such tests

protolambda: > And the if statements below changed to then not do the target check again.

That's what I said, "to not check the target *again*", after it being checked as part of `is_matching_head`. Anyway, it looks consistent, but we need to check testing.

protolambda: Confirmed we have tests for it. Bad target while a valid head vote is processed, with multiple cases for inclusion delay too: https://github.com/ethereum/eth2.0-specs/blob/0c94be420405380f104347e643500f4613ab8f9c/tests/core/pyspec/eth2spec/test/phase0/block_processing/test_process_attestation.py#L498

specs/altair/beacon-chain.md
protolambda: I'll make an issue to review the precision of Altair rewards, but I don't think it's significant enough to block this PR, especially since rewards change in other bigger ways.

protolambda (approved):
LGTM

specs/altair/beacon-chain.md
```diff
@@ -311,10 +320,17 @@ def get_base_reward_per_increment(state: BeaconState) -> Gwei:
 
 #### `get_base_reward`
 
-*Note*: The function `get_base_reward` is modified with the removal of `BASE_REWARDS_PER_EPOCH`.
+*Note*: The function `get_base_reward` is modified with the removal of `BASE_REWARDS_PER_EPOCH` and the use of increment based accounting.
```
ralexstokes, line 323: tbh i'm not really sure what "increment based" means but i don't think it blocks this PR

ralexstokes (approved):
lgtm!

i left a comment about "increment based" accounting bc i think there is a more appropriate term but doesn't really block this PR

Issue #2315, opened 2021-04-07
Light client specs checklist


hwwhww:
- [ ] 1. **Server discovery**: How can light clients find servers efficiently?
- [ ] 2. **Find `LightClientSnapshot` initially**: as @djrtwo mentioned, Req/Resp may not be good snapshot resource.
    - **[Proposal]** Can we say, we expect 3rd party to provide LightClientSnapshot + Merkle proofs or multi-proof of each field, where the proof can be verified against WS source?
- [ ] 3. **Block tree verification in current time-out procedure and reorg**: In the current spec, a timeout-triggered (8192 slots, ~27 hours ) update logic exists:
    - Step 1, `on_light_client_update`, light client appends the valid update
    - Step 2, when timeout, call `apply_light_client_update` to update the store snapshot with *the most voted* update from `store.valid_updates`.
    - **[Proposal]**: In step 1, for the non-finality update (`update.finality_header == BeaconBlockHeader()`), should we add assert of checking if the `update.header` is descendant of `snapshot.header`. (<-- as @vbuterin suggested, make the light client updates always rooted in a finalized block.)
        - This procedure assumes that the light client already has `store.valid_updates` before timeout. That is, at the timeout slot, either (i) the server has to send a chain of updates to the light clients, or (ii) the light client has to request for the updates.
    - Or apply other reorgable light client solutions? (https://github.com/ethereum/eth2.0-specs/issues/2182)
- [ ] 4. Use **gossip as default sync resource** and then servers just as backup?
     - Note: the size of `LightClientUpdate` object:
         - header: `200`
         - next_sync_committee: about `48*1024 + 48*16 = 49920`
         - next_sync_committee_branch: about `32*5 = 160`
         - finality_header: `200`
         - finality_branch: about `32*6 = 192`
         - sync_aggregate: about `1024 + 96 = 1120`
         - fork_version: `4`
         - **Total: ~51,796 bytes**
- [ ] 5. More comprehensive/production **client-server model**
    - @protolambda: If this is a push-update, and there is no explicit subscription kind of approach for a light client, then we should spec out when and how often the server pushes the update.

### Other questions:

1. `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` is `2**8` epochs = `2**13` slots ~= 27 hours. `LIGHT_CLIENT_UPDATE_TIMEOUT` is also `2**13` slots ~= 27 hours.
    1. Should we just keep one constant?
[...]

PR #2378, opened 2021-05-04
Updates to Altair light client sync protocol

ralexstokes:
* Adjust consensus threshold to be in line w/ every other usage (>= 2/3, rather than just > 2/3)
* Remove `LIGHT_CLIENT_UPDATE_TIMEOUT` as a constant when it can just be computed
* Remove `MAX_VALID_LIGHT_CLIENT_UPDATES` w/ a simpler implementation of `LightClientStore`


configs/mainnet/altair.yaml
```diff
@@ -45,10 +45,8 @@ ALTAIR_FORK_EPOCH: 18446744073709551615
 # ---------------------------------------------------------------
 # 1
 MIN_SYNC_COMMITTEE_PARTICIPANTS: 1
-# 2**13
-MAX_VALID_LIGHT_CLIENT_UPDATES: 8192
-# 2**13 (=8192)
-LIGHT_CLIENT_UPDATE_TIMEOUT: 8192
+# 2**64 - 1
+MAX_VALID_LIGHT_CLIENT_UPDATES: 18446744073709551615
```
ralexstokes, line 49: this value is just setting an upper bound on the possible size of a `ssz` list so i don't see a reason to not just use the max value (which matches the latest spec)

hwwhww: nice catch

hwwhww (approved):
Yeah, it fixes a pending question on #2315! LGTM

configs/mainnet/altair.yaml
djrtwo: Considering it's only used in a local store that max value is fine *but* considering this is just a local representation of a DB, there is no value in having it be an SSZ object (for merklization or consensus) and thus it does not need this list limit at all if we move to a more generic `dataclass`. This is what we did with the [fork choice `Store`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/fork-choice.md#store) and I think it appropriate here.

setup.py
```diff
@@ -167,7 +167,7 @@ def get_spec(file_name: str) -> SpecObject:
             comment = _get_eth2_spec_comment(child)
             if comment == "skip":
                 should_skip = True
-    
+
```
ralexstokes, line 170: just whitespace

specs/altair/sync-protocol.md
```diff
@@ -94,9 +86,10 @@ class LightClientUpdate(Container):
 ### `LightClientStore`
 
 ```python
-class LightClientStore(Container):
+@dataclass
+class LightClientStore(object):
     snapshot: LightClientSnapshot
-    valid_updates: List[LightClientUpdate, MAX_VALID_LIGHT_CLIENT_UPDATES]
+    valid_updates: Set[LightClientUpdate]
```
ralexstokes, line 92: rationale for this change:

decided to go w/ `Set` here as `List` has a local definition as the `ssz` type and python's `typing.List` that we want here would collide by name...

rather than have to juggle a type alias `PythonList` that would need to be defined for spec consumers, or use the fully qualified name `typing.List` (that would not directly work w/ our spec builder machinery), we can just move to a `Set` datatype -- which satisfies the needs of `valid_updates` just the same

setup.py
```diff
@@ -548,7 +548,7 @@ def combine_constants(old_constants: Dict[str, str], new_constants: Dict[str, st
     'Bytes1', 'Bytes4', 'Bytes20', 'Bytes32', 'Bytes48', 'Bytes96', 'Bitlist', 'Bitvector',
     'uint8', 'uint16', 'uint32', 'uint64', 'uint128', 'uint256',
     'bytes', 'byte', 'ByteList', 'ByteVector',
-    'Dict', 'dict', 'field', 'ceillog2', 'floorlog2',
+    'Dict', 'dict', 'field', 'ceillog2', 'floorlog2', 'Set'
```
djrtwo, line 551:
```suggestion
    'Dict', 'dict', 'field', 'ceillog2', 'floorlog2', 'Set',
```

djrtwo (approved):
great!

(added a comma, will commit and built)

PR #2380, opened 2021-05-04
Merge test vectors: enable phase0 tests for Merge + start on new testing

protolambda:
Changes:
- Add merge to forks list with enabled testing and test-vector outputs
- Update test generators to include merge:
  - SSZ static tests now respect `TESTGEN_FORKS` again
  - epoch processing (inherits phase0 tests)
  - finality (inherits phase0 tests)
  - operations (inherits phase0 tests, added execution payload testing)
  - rewards (inherits phase0 tests)
  - sanity (inherits phase0 tests, added block tests)
- Helper method to create empty execution payload on top of the latest state
- Update beacon-block construction to add in a valid empty execution payload when the merge spec is detected.
  - TODO: may need to be a zeroed execution payload for the time during the merge transition. For next PR.
- Refactor merge spec (no breaking consensus changes):
  - Isolate the execution payload input for execution payload processing. There is no good reason to include the full block body. The old version makes forking the Merge code with code-reuse much more difficult.
  - The if statement for processing simply moves to outside the processing function. So the processing is conditional, instead of a possible no-op.
- Implement execution-payload tests
  - [x] Scaffold edge cases to test
     - Note: left 1 minimal test case. Further testing in next PR.
  - [x] Define test format
- Implement block tests
  - Just a minimal scaffold, needs more work in later PR

**TLDR: Port phase0 testing for Merge, enable Merge test generators.**

### DIFF
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -146,8 +146,8 @@ def is_transition_completed(state: BeaconState) -> bool:
 #### `is_transition_block`
 
 ```python
-def is_transition_block(state: BeaconState, block_body: BeaconBlockBody) -> bool:
-    return not is_transition_completed(state) and block_body.execution_payload != ExecutionPayload()
+def is_transition_block(state: BeaconState, block: BeaconBlock) -> bool:
+    return not is_transition_completed(state) and block.body.execution_payload != ExecutionPayload()
 ```
 
 #### `compute_time_at_slot`
@@ -168,7 +168,9 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
     process_randao(state, block.body)
     process_eth1_data(state, block.body)
     process_operations(state, block.body)
-    process_execution_payload(state, block.body)  # [New in Merge]
+    # Pre-merge, skip execution payload processing
+    if is_transition_completed(state) or is_transition_block(state, block):
+        process_execution_payload(state, block.body.execution_payload)  # [New in Merge]
 ```
 
 #### Execution payload processing
@@ -181,16 +183,10 @@ The body of the function is implementation dependent.
 ##### `process_execution_payload`
 
 ```python
-def process_execution_payload(state: BeaconState, body: BeaconBlockBody) -> None:
+def process_execution_payload(state: BeaconState, execution_payload: ExecutionPayload) -> None:
     """
     Note: This function is designed to be able to be run in parallel with the other `process_block` sub-functions
     """
-    # Pre-merge, skip processing
-    if not is_transition_completed(state) and not is_transition_block(state, body):
-        return
-
-    execution_payload = body.execution_payload
-
     if is_transition_completed(state):
         assert execution_payload.parent_hash == state.latest_execution_payload_header.block_hash
         assert execution_payload.number == state.latest_execution_payload_header.number + 1
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -13,8 +13,8 @@
 - [Introduction](#introduction)
 - [Custom types](#custom-types)
 - [Constants](#constants)
-  - [Transition](#transition)
   - [Execution](#execution)
+- [Configuration](#configuration)
 - [Containers](#containers)
   - [Extended containers](#extended-containers)
     - [`BeaconBlockBody`](#beaconblockbody)
@@ -50,12 +50,6 @@ We define the following Python custom types for type hinting and readability:
 
 ## Constants
 
-### Transition
-
-| Name | Value |
-| - | - |
-| `TRANSITION_TOTAL_DIFFICULTY` | **TBD** |
-
 ### Execution
 
 | Name | Value |
@@ -63,6 +57,16 @@ We define the following Python custom types for type hinting and readability:
 | `MAX_BYTES_PER_OPAQUE_TRANSACTION` | `uint64(2**20)` (= 1,048,576) |
 | `MAX_EXECUTION_TRANSACTIONS` | `uint64(2**14)` (= 16,384) |
 | `BYTES_PER_LOGS_BLOOM` | `uint64(2**8)` (= 256) |
+
+## Configuration
+
+Warning: this configuration is not definitive.
+
+| Name | Value |
+| - | - |
+| `MERGE_FORK_VERSION` | `Version('0x02000000')` |
+| `MERGE_FORK_EPOCH` | `Epoch(18446744073709551615)` **TBD** |
+| `TRANSITION_TOTAL_DIFFICULTY` | **TBD** |
 
 ## Containers
 
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -24,6 +24,7 @@
     - [`ExecutionPayloadHeader`](#executionpayloadheader)
 - [Helper functions](#helper-functions)
   - [Misc](#misc)
+    - [`is_execution_enabled`](#is_execution_enabled)
     - [`is_transition_completed`](#is_transition_completed)
     - [`is_transition_block`](#is_transition_block)
     - [`compute_time_at_slot`](#compute_time_at_slot)
@@ -140,6 +141,13 @@ class ExecutionPayloadHeader(Container):
 
 ### Misc
 
+#### `is_execution_enabled`
+
+```python
+def is_execution_enabled(state: BeaconState, block: BeaconBlock) -> bool:
+    return is_transition_completed(state) or is_transition_block(state, block)
+```
+
 #### `is_transition_completed`
 
 ```python
@@ -173,7 +181,7 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
     process_eth1_data(state, block.body)
     process_operations(state, block.body)
     # Pre-merge, skip execution payload processing
-    if is_transition_completed(state) or is_transition_block(state, block):
+    if is_execution_enabled(state, block):
         process_execution_payload(state, block.body.execution_payload)  # [New in Merge]
 ```
 
--- a/tests/core/pyspec/eth2spec/test/context.py
+++ b/tests/core/pyspec/eth2spec/test/context.py
@@ -367,12 +367,20 @@ def with_configs(configs, reason=None):
 
 
 def is_post_altair(spec):
+    if spec.fork == MERGE:  # TODO: remove parallel Altair-Merge condition after rebase.
+        return False
     if spec.fork in FORKS_BEFORE_ALTAIR:
         return False
     return True
 
 
 def is_post_merge(spec):
+    if spec.fork == ALTAIR:  # TODO: remove parallel Altair-Merge condition after rebase.
+        return False
     if spec.fork in FORKS_BEFORE_MERGE:
         return False
     return True
+
+
+with_altair_and_later = with_phases([ALTAIR])  # TODO: include Merge, but not until Merge work is rebased.
+with_merge_and_later = with_phases([MERGE])
--- a/tests/core/pyspec/eth2spec/test/context.py
+++ b/tests/core/pyspec/eth2spec/test/context.py
@@ -330,7 +330,6 @@ def with_phases(phases, other_phases=None):
                 phase_dir[PHASE0] = spec_phase0
             if ALTAIR in available_phases:
                 phase_dir[ALTAIR] = spec_altair
-
             if MERGE in available_phases:
                 phase_dir[MERGE] = spec_merge
 
--- a/tests/core/pyspec/eth2spec/test/context.py
+++ b/tests/core/pyspec/eth2spec/test/context.py
@@ -7,7 +7,7 @@ from eth2spec.utils import bls
 
 from .exceptions import SkippedTest
 from .helpers.constants import (
-    PHASE0, ALTAIR,
+    PHASE0, ALTAIR, MERGE,
     ALL_PHASES, FORKS_BEFORE_ALTAIR, FORKS_BEFORE_MERGE,
 )
 from .helpers.genesis import create_genesis_state
@@ -312,7 +312,7 @@ def with_phases(phases, other_phases=None):
                     return None
                 run_phases = [phase]
 
-            if PHASE0 not in run_phases and ALTAIR not in run_phases:
+            if PHASE0 not in run_phases and ALTAIR not in run_phases and MERGE not in run_phases:
                 dump_skipping_message("none of the recognized phases are executable, skipping test.")
                 return None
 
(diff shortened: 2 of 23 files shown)


specs/merge/beacon-chain.md
```diff
@@ -168,7 +168,9 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
     process_randao(state, block.body)
     process_eth1_data(state, block.body)
     process_operations(state, block.body)
-    process_execution_payload(state, block.body)  # [New in Merge]
+    # Pre-merge, skip execution payload processing
+    if is_transition_completed(state) or is_transition_block(state, block):
+        process_execution_payload(state, block.body.execution_payload)  # [New in Merge]
```
hwwhww, line 173: Two issues:
1. Looking at `tests/formats/operations/README.md` and I just realized that we put `sync_aggregate` as part of "operation" tests but `process_sync_committee` is not part of the `process_operations` function like `execution_payload`.
    1. IMO it's good to consider `sync_aggregate` as an operation.
    2. If we want to simplify it, I think `execution_payload` could also be considered as an operation since it's essentially "transactions".
2. I can see why it looks good to extract the `is_transition_completed(state) or is_transition_block(state, block)` condition checks from `process_execution_payload`, but it makes it a bit unclear where/how to test these conditions in test cases.

What do you think about:
1. Move `process_sync_committee` and `process_execution_payload` to `process_operations`.
2. Revert the `if is_transition_completed(state) or is_transition_block(state, block)` checks extraction.

protolambda: 1. Regarding `process_sync_committee`: sure, but let's keep that separate from merge-testing for now. Moving the `process_execution_payload` to `process_operations` would be good too. I'm honestly surprised we didn't do the same for the Randao and Eth1Data functionality.
2. The if-statement was purposefully extracted from the `process_execution_payload` function: this way the function input can just be an `ExecutionPayload`, and not a complete beacon block body. This helps with future compatibility and reduces test complexity (We don't have to set up a complete beacon block for every single test case). We can test that one if-statement as part of the regular full beacon block testing. And in future phases after the merge, we may even remove the if-statement entirely, since the merge-transition check will always be true.

tests/core/pyspec/eth2spec/test/helpers/genesis.py
```diff
@@ -28,6 +29,8 @@ def create_genesis_state(spec, validator_balances, activation_threshold):
 
     if spec.fork == ALTAIR:
         current_version = spec.ALTAIR_FORK_VERSION
+    elif spec.fork == MERGE:
+        current_version = spec.MERGE_FORK_VERSION
```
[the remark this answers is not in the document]
djrtwo: Do we need to manually set `latest_execution_payload_header` in the generated state here? 

I suppose not because then the first non-genesis block looks like a transition block... but some other points lead me to believe we might be missing the execution_payload execution all together.

tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py
```diff
@@ -0,0 +1,40 @@
...
+    if not valid:
+        expect_assertion_error(lambda: spec.process_execution_payload(state, execution_payload))
+        yield 'post', None
+        return
+
+    spec.process_execution_payload(state, execution_payload)
+
+    yield 'post', state
+
+    # TODO: any assertions to make?
```
djrtwo, line 29: Could do `pre.latest_execution_payload_header != post.latest_execution_payload_header` to show that the function has the most basic effect

tests/core/pyspec/eth2spec/test/helpers/execution_payload.py
```diff
@@ -0,0 +1,26 @@
+
+def build_empty_execution_payload(spec, state):
+    """
+    Assuming a pre-state of the same slot, build a valid ExecutionPayload without any transactions.
```
djrtwo, line 4: I think there's a subtle bug in common usage here.

1. `build_empty_block_for_next_slot` returns `build_empty_block(spec, state, state.slot + 1)`
2. `build_empty_block` if a param is passed in for `slot`, does *not* pass this along to  `build_empty_execution_payload` which just assumes the slot from `state`
3. `timestamp` is computed from `state.slot`
4. Thus, `block.slot` and `block.execution_payload.timestamp` will have a mismatch when using `build_empty_block_for_next_slot` (or if you pass in any non-current-slot value to `build_empty_block`

Immediate solutions: either transition a dummy state forward to the slot in question within `build_empty_block` or pass in `slot` to `build_empty_execution_payload`

djrtwo: Are we not hitting this in `process_execution_payload`:
```python
assert execution_payload.timestamp == compute_time_at_slot(state, state.slot)
```

Seems that either (1) my bug report is incorrect, (2) we never really use `build_empty_block_for_next_slot`, or (3) we are never actually running `process_execution_payload`.

We can add a sanity condition to the `process_execution_payload` tests to see if the execution header changed to narrow it down.

tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@with_merge_and_later
+@spec_state_test
+def test_success_first_payload(spec, state):
+    next_slot(spec, state)
+    assert not spec.is_transition_completed(state)
```
djrtwo, line 36: Maybe assert that it is a transition block too

djrtwo (commented):
Nice!

I think there's a bug in `build_empty_block_for_next_slot` and I'm not certain if we are actually hitting `process_execution_payload`. Didn't do any testing. Check out my comments. Hit me up. happy to discuss in DM too.

tests/core/pyspec/eth2spec/test/helpers/execution_payload.py
```diff
@@ -0,0 +1,26 @@
...
+        number=latest.number + 1,
+        gas_limit=latest.gas_limit,  # retain same limit
+        gas_used=0,  # empty block, 0 gas
+        timestamp=timestamp,
+        receipt_root=b"no receipts here" + b"\x00" * 16,  # TODO: root of empty MPT may be better.
+        logs_bloom=spec.ByteVector[spec.BYTES_PER_LOGS_BLOOM](),  # TODO: zeroed logs bloom for empty logs ok?
+        transactions=empty_txs,
+    )
+    # TODO: real RLP + block hash logic would be nice, requires RLP and keccak256 dependency however.
+    payload.block_hash = spec.Hash32(spec.hash(payload.hash_tree_root() + b"FAKE RLP HASH"))
```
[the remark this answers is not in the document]
protolambda: Right, it's just a short-cut to make a hash that is unique to the payload. Encoding and flat-hashing would work to. I think we may just want to do the real RLP thing though. The current thing works for now, but it's nice if it's correct, especially if we get eth1-eth2 combination clients that run these tests.

tests/core/pyspec/eth2spec/test/helpers/genesis.py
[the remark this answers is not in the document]
protolambda: The current merge transition code introduces the first payload on top of a zeroed `latest_execution_payload_header`, so it should work. I am not sure if/when this would change, or what should be the common case for testing. I prefer to just keep it zeroed for now though, and have later testing (PR #2385 ) change it after looking deeper into it.

tests/core/pyspec/eth2spec/test/helpers/execution_payload.py
protolambda: ` build_empty_execution_payload` in `build_empty_block_for_next_slot` uses a copy of `state` which is already transitioned to that same slot.

tests/core/pyspec/eth2spec/test/helpers/execution_payload.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    latest = state.latest_execution_payload_header
+    timestamp = spec.compute_time_at_slot(state, state.slot)
+    empty_txs = spec.List[spec.OpaqueTransaction, spec.MAX_EXECUTION_TRANSACTIONS]()
+
+    payload = spec.ExecutionPayload(
+        block_hash=spec.Hash32(),
+        parent_hash=latest.block_hash,
+        coinbase=spec.Bytes20(),
+        state_root=latest.state_root,  # no changes to the state
```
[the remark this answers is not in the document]
protolambda: There may be something yes, although I'm not entirely sure either. The account tree definitely does not change (no block rewards on eth1 state anymore), and the storage/values/code/nonce of accounts only changes with transactions. Can't find anything in the yellow paper that would otherwise touch the eth1 world state.

I'll open an issue to investigate this more. Maybe this new possibility of a duplicate state-root in eth1 blocks is untested/dangerous?

tests/core/pyspec/eth2spec/test/helpers/execution_payload.py
```diff
@@ -0,0 +1,26 @@
...
+        number=latest.number + 1,
+        gas_limit=latest.gas_limit,  # retain same limit
+        gas_used=0,  # empty block, 0 gas
+        timestamp=timestamp,
+        receipt_root=b"no receipts here" + b"\x00" * 16,  # TODO: root of empty MPT may be better.
+        logs_bloom=spec.ByteVector[spec.BYTES_PER_LOGS_BLOOM](),  # TODO: zeroed logs bloom for empty logs ok?
```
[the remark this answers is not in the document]
protolambda: I think it's best to re-visit this in the PR about execution-payload testing. A zeroed logs bloom is fine for now.

tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -312,7 +312,7 @@ def wrapper(*args, **kw):
                     return None
                 run_phases = [phase]
 
-            if PHASE0 not in run_phases and ALTAIR not in run_phases:
+            if PHASE0 not in run_phases and ALTAIR not in run_phases and MERGE not in run_phases:
```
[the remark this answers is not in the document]
protolambda: It is checking if `run_phases` does not contain any known phase. So it would be `if not any(EXECUTABLE_PHASES.intersection(run_phases)):`

I think it's best to leave as-is. Less code, no function calls, and it is easier to reason about this change. Once we add more phases we can still change it.

tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py
protolambda: There is no block processing here, just the execution payload, so as-is it cannot be asserted.

Maybe we should start using the new `run_block_processing_to` helper to simulate the block behavior up to the execution-payload processing, I can discuss this with Mikhail for the next PR, but I prefer to keep this PR simple.

tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py
protolambda: done

tests/core/pyspec/eth2spec/test/helpers/execution_payload.py
djrtwo: indeed it is. whoops!

tests/core/pyspec/eth2spec/test/helpers/genesis.py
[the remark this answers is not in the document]
mkalinin: We may leave it for now. But technically it means that we do the transition during these tests and if we want to separate transition from post-merge test scenarios then we'd need to bootstrap the header in the state with some initial values. Initial value might has filled `block_hash` field with according value.

tests/core/pyspec/eth2spec/test/helpers/execution_payload.py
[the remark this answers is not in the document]
mkalinin: We might want to decouple this shortcut and use it as a helper.

tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py
mkalinin: Is this test case supposed to test the transition?

tests/core/pyspec/eth2spec/test/helpers/execution_payload.py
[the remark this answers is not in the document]
djrtwo: might be. at least it might not be expected

I think we've touched on this a tiny bit. Peter mentioned that it isn't the case today because of issuance/coinbase but that it might change in this context. I think they can handle it but it's something to highlight as a design consideration/edge-case for execution engines

tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py
djrtwo: `process_execution_payload` has been refactored to not care about the transition logic. So yes, it tests the empty header to a real header but it doesn't  test that the transition go/no-go logic works

djrtwo (approved):
great start! happy to get this merged pretty soon.

tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py
mkalinin: A nit: then we likely should not use transition spec helpers to check the pre-state.

specs/merge/beacon-chain.md
```diff
@@ -168,7 +180,9 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
     process_randao(state, block.body)
     process_eth1_data(state, block.body)
     process_operations(state, block.body)
-    process_execution_payload(state, block.body)  # [New in Merge]
+    # Pre-merge, skip execution payload processing
+    if is_execution_enabled(state, block):
```
mkalinin, line 184: What do you think about the following name? I am not sure how idiomatic it is but it reads better to me than "enabled". As the latter sounds more like an externally defined option that is immutable in the runtime.
```suggestion
    if should_process_execution_payload(state, block):
```

tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py
djrtwo: ah, but we want to test a range of tests/pre-states/inputs directly on this function. One of which when `state` has an empty header which maps to `not spec.is_transition_completed`. Even though `process_execution_payload` does not check this, we do want to ensure we have a test cast that is in this pre-condition.

If we don't use that helper, we'd want to check `state.latest_execution_header == ExecutionHeader()`. Or directly just set it.

specs/merge/beacon-chain.md
djrtwo: I *do* like the idea of being able to override the function in a non-merge context to just have execution enabled cleanly from the start. That said, I see your point.

all predicates in the spec currently have an `is_` prefix and so by default, I'd like to keep that unless we can't come up with something better.

that said, I can't currently come up with anything better so I'm happy to use `should_` for now and then see what we come up with in subsequent review

PR #2382, opened 2021-05-05
Reorganization of config params to put sync committee in one section

vbuterin:
This may be marginally "cleaner" than the previous approach, keeping constants with the same topic together.

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -114,19 +114,18 @@ This patch updates a few configuration values to move penalty parameters toward 
 | `MIN_SLASHING_PENALTY_QUOTIENT_ALTAIR` | `uint64(2**6)` (= 64) |
 | `PROPORTIONAL_SLASHING_MULTIPLIER_ALTAIR` | `uint64(2)` |
 
+### Sync committee
+
+| Name | Value | Unit | Duration |
+| - | - | - | - |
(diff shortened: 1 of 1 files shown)


specs/altair/beacon-chain.md
```diff
@@ -114,20 +114,19 @@ This patch updates a few configuration values to move penalty parameters toward
...
+
+| Name | Value | Unit | Duration |
+| - | - | - | - |
+| `SYNC_COMMITTEE_SIZE` | `uint64(2**10)` (= 1,024) | | |
+| `SYNC_PUBKEYS_PER_AGGREGATE` | `uint64(2**6)` (= 64) | | |
+| `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` | `Epoch(2**8)` (= 256) | epochs | ~27 hours |
+
 ### Misc
 
 | Name | Value |
```
vbuterin, line 127:
```suggestion
| Name | Value |
| - | - |
```

hwwhww:
Not strongly, but I'm slightly in favor of putting `Time parameters` together although there is only one new time parameter in Altair. 😅  

Note that #2370 will remove `SYNC_PUBKEYS_PER_AGGREGATE` so there will be only 2 config parameters in the "Sync committee" category then.

djrtwo (commented):
I'm okay with this. Sync committee grouping being logically coherent and together trumps historic precedent on this one for me

dapplion (approved):

djrtwo (approved):

PR #2383, opened 2021-05-05
[Sharding] Simplify and optimize get_start_shard()

Nashatyrev:
- `committee_count` and `active_shard_count` are invariants in the loops as `compute_epoch_at_slot(_slot)` would always be equal to `compute_epoch_at_slot(slot)`
- replace with multiplication the loops which effectively do N `add by module` operations 

From my perspective the code looks a bit more clear and clean this way.

### DIFF
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -425,17 +425,15 @@ def get_start_shard(state: BeaconState, slot: Slot) -> Shard:
     """
     current_epoch_start_slot = compute_start_slot_at_epoch(get_current_epoch(state))
     shard = state.current_epoch_start_shard
+    committee_count = get_committee_count_per_slot(state, compute_epoch_at_slot(slot))
+    active_shard_count = get_active_shard_count(state, compute_epoch_at_slot(slot))
     if slot > current_epoch_start_slot:
         # Current epoch or the next epoch lookahead
         for _slot in range(current_epoch_start_slot, slot):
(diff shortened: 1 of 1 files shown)


specs/sharding/beacon-chain.md
```diff
@@ -424,21 +424,17 @@ def get_start_shard(state: BeaconState, slot: Slot) -> Shard:
...
         # Current epoch or the next epoch lookahead
-        for _slot in range(current_epoch_start_slot, slot):
-            committee_count = get_committee_count_per_slot(state, compute_epoch_at_slot(Slot(_slot)))
-            active_shard_count = get_active_shard_count(state, compute_epoch_at_slot(Slot(_slot)))
-            shard = (shard + committee_count) % active_shard_count
+        slots_diff = slot - current_epoch_start_slot
+        return (state.current_epoch_start_shard + slots_diff) % active_shard_count
     elif slot < current_epoch_start_slot:
         # Previous epoch
-        for _slot in list(range(slot, current_epoch_start_slot))[::-1]:
```
protolambda, line 436: This `_slot` is actually different, it's there for a reason. I agree that the previous code is complicated and not pretty though. Maybe we can optimize it in a different way?

Nashatyrev: Yep, totally right! The slot could be further than 1 epoch from the current state epoch. 
Then those changes are incorrect. Leaving as-is

PR #2384, opened 2021-05-05
`get_shard_proposer_index` to use `DOMAIN_SHARD_PROPOSER`

terencechain:
I think `get_shard_proposer_index` should be using `DOMAIN_SHARD_PROPOSER` for seed calculation instead of `DOMAIN_BEACON_PROPOSER`

cc @protolambda


protolambda (approved):
Good catch, thank you.

PR #2385, opened 2021-05-05
Execution payload tests

protolambda:
Work in progress. Builds on #2380, for new merge-specific testing, like execution-payload tests. Working on tests with @mkalinin .

### DIFF
--- a/specs/merge/validator.md
+++ b/specs/merge/validator.md
@@ -43,8 +43,8 @@ The following methods are added to the `ExecutionEngine` protocol for use as a v
 
 #### `assemble_block`
 
-Produces a new instance of an execution payload, with the specified timestamp,
-on top of the execution payload chain tip identified by `block_head`.
+Produces a new instance of an execution payload, with the specified `timestamp`,
+on top of the execution payload chain tip identified by `block_hash`.
 
 The body of this function is implementation dependent.
 The Consensus API may be used to implement this with an external execution engine.
--- a/tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py
+++ b/tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py
@@ -1,4 +1,9 @@
-from eth2spec.test.helpers.execution_payload import build_empty_execution_payload
+from eth2spec.test.helpers.execution_payload import (
+    build_empty_execution_payload,
+    get_execution_payload_header,
+    build_state_with_incomplete_transition,
+    build_state_with_complete_transition,
+)
 from eth2spec.test.context import spec_state_test, expect_assertion_error, with_merge_and_later
 from eth2spec.test.helpers.state import next_slot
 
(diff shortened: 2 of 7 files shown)


tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py
```diff
@@ -13,31 +18,178 @@ def run_execution_payload_processing(spec, state, execution_payload, valid=True,
...
     """
 
-    pre_exec_header = state.latest_execution_payload_header.copy()
-
     yield 'pre', state
     yield 'execution', {'execution_valid': execution_valid}
     yield 'execution_payload', execution_payload
 
+
+    spec.verify_execution_state_transition_ret_value = execution_valid
```
mkalinin, line 26: This is dirty and unsafe. The best way of implementing it would be in creating a copy of spec with required changes, e.g. `spec = spec.copy(verify_execution_state_transition_ret_value = execution_valid)` in this particular case.

protolambda: #2398 resolves this

tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py
```diff
@@ -13,31 +18,184 @@ def run_execution_payload_processing(spec, state, execution_payload, valid=True,
...
+@with_merge_and_later
+@spec_state_test
+def test_bad_number_regular_payload(spec, state):
+    # pre-state
+    state = build_state_with_complete_transition(spec, state)
+    next_slot(spec, state)
+
+    # execution payload
+    execution_payload = build_empty_execution_payload(spec, state)
+    execution_payload.number = execution_payload.number + 1
```
ralexstokes, line 156: noting that `execution_payload.number` is already correct from `build_empty_execution_payload` so that the increment makes it _incorrect_

ralexstokes (approved):
generally looks good by laying more of the foundation for the (production) merge spec!

i suspect this is conforming to RPC APIs specified elsewhere but i found the name/identifier `new_block` kind of confusing, esp. alongside `assemble_block`, possibly consider making `new_block` into something more of a predicate `is_valid_block` or similar

tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py
protolambda: I agree, but the naming is this way to match the existing Merge spec: https://github.com/ethereum/rayonism/blob/master/specs/merge.md This is as far as we have for an API for the Merge. I don't think we should change it in this PR. With the API redesign in the coming weeks we can rename it.

specs/merge/validator.md
```diff
@@ -32,6 +35,25 @@ This document is an extension of the [Phase 0 -- Validator](../phase0/validator.
...
+## Protocols
+
+### `ExecutionEngine`
+
+The following methods are added to the `ExecutionEngine` protocol for use as a validator:
+
+#### `assemble_block`
+
+Produces a new instance of an execution payload, with the specified timestamp,
+on top of the execution payload chain tip identified by `block_head`.
```
djrtwo, line 47: You say `block_head` here but then use `block_hash` in `assemble_block`. No strong preference on naming but just pick one

djrtwo (approved):
great! 

one minor naming comment to resolve

specs/merge/validator.md
protolambda: Good catch, it's probably a remainder of the old processing functions, I'll fix that.

PR #2386, opened 2021-05-05
[Sharding] Fix the EFFECTIVE_BALANCE_MAX_DOWNWARD_DEVIATION constant calculation

Nashatyrev:
Fix back to the original formula here: 68844f2


Nashatyrev:
However the original formula looks not quite accurate

For example taking default spec constants: 

Suppose the max blob gas cost is `9.5Eth`, 
`EFFECTIVE_BALANCE_MAX_DOWNWARD_DEVIATION = 1Eth - 1Eth / 4 * 1 = 0.75Eth`
then `min_effective_balance = 9.5Eth + 0.75Eth = 10.25Eth`

Suppose proposer `effective_balance == 10Eth`, 
Thus his minimal possible balance would be `9.75Eth`.
Despite his balance would cover blob gas cost he still would be filtered out by the `min_effective_balance =10.25Eth`

Nashatyrev:
The latter is not critical though and could be fixed later after covered with reference tests

PR #2387, opened 2021-05-05
Dankrad altair review

dankrad:


dankrad:
I made some changes to reward/penalty calculation:

Non-substantive:
 * I feel like it's not that productive to have the separate "get_inactivity_penalty_deltas" function? Shouldn't we just integrate it into get_flag_index_deltas for the target flag? The two functions are actually kind of tightly dependent on each other and I think they are actually easier to read if integrated

Substantive:
 * I think the non-quadratic inactivity leak is actually applied twice — once in the last "else" branch of get_flag_index_deltas, and once in the first branch of get_inactivity_penalty_deltas, where the comment says "# This inactivity penalty cancels the flag reward corresponding to the flag index". I think this doesn't make sense. I would simply remove the penalty in get_inactivity_penalty_deltas, and not apply any reward in this case.
 * I think in the past we discussed how it is annoying that you can't break even in the inactivity leak, because it's impossible to get your attestation in in time (after all, in expectation at least 1/3 of slots are missing). This still isn't fixed — we apply the full penalty for timely head even though it's out of the control of the attester. I suggest not punishing for an untimely head in the inactivity leak

specs/altair/beacon-chain.md
```diff
@@ -99,6 +98,8 @@ Altair is the first beacon chain hard fork. Its main features are:
 | Name | Value |
 | - | - |
 | `G2_POINT_AT_INFINITY` | `BLSSignature(b'\xc0' + b'\x00' * 95)` |
+| `PARTICIPATION_FLAGS` | `3` |
+| `PARTICIPATION_FLAG_WEIGHTS` | `[TIMELY_HEAD_WEIGHT, TIMELY_SOURCE_WEIGHT, TIMELY_TARGET_WEIGHT]` |
```
ralexstokes, line 102: i can't comment up above but in this case, we should get rid of the constants `TIMELY_*_INDEX` defined up above

specs/altair/beacon-chain.md
```diff
@@ -204,6 +205,7 @@ class BeaconState(Container):
 ```python
+    # TODO! Need multiple signatures as discussed between Justin and Vitalik May 3 2021 (see Telegram)
```
ralexstokes, line 208: i think we are leaning towards just having a single signature

specs/altair/beacon-chain.md
```diff
@@ -632,10 +599,8 @@ def process_rewards_and_penalties(state: BeaconState) -> None:
     if get_current_epoch(state) == GENESIS_EPOCH:
         return
 
-    flag_indices_and_numerators = get_flag_indices_and_weights()
-    flag_deltas = [get_flag_index_deltas(state, index, numerator) for (index, numerator) in flag_indices_and_numerators]
-    deltas = flag_deltas + [get_inactivity_penalty_deltas(state)]
-    for (rewards, penalties) in deltas:
+    flag_deltas = [get_flag_index_deltas(state, flag_index) for flag_index in range(PARTICIPATION_FLAGS)]
```
ralexstokes, line 602:
```suggestion
    flag_deltas = [get_flag_index_deltas(state, flag_index) for flag_index in range(len(PARTICIPATION_FLAG_WEIGHTS))]
```

if we do this then we can drop the `PARTICIPATION_FLAGS` constant entirely and avoid having to keep it in sync w/ the list

specs/altair/beacon-chain.md
```diff
@@ -276,8 +264,9 @@ def has_flag(flags: ParticipationFlags, flag_index: int) -> bool:
 ```python
     """
-    Return the sequence of sync committee indices (which may include duplicate indices)
+    Return the sequence of sync committee indices
     for a given ``state`` and ``epoch``.
```
ralexstokes, line 268: i think we can just get away with this now:

```suggestion
    Return the sequence of sync committee indices for a given ``state`` and ``epoch``.
```

ralexstokes (commented):
no comment yet on the rewards refactor but generally support the flags refactor

specs/altair/beacon-chain.md
dankrad: Fair.

PR #2390, opened 2021-05-06
Separation of Constant, Preset and Configuration variables

protolambda:
Feedback welcome. The idea is to make customization more consistent and reliable between clients, without any breaking changes.

## Issues

Currently, as someone who configures many local tests, testing and testnets, I face the following issues:
- The configuration file format is not documented well enough
- Not all of the clients can read all different config inputs. Differences exist and are not documented.
- Some configuration is unreliable. There is no special effort in fuzzing different weird or completely unreasonable combinations of configuration.
- The configuration mixes up chain configuration (state sizes, validator/committee configuration) and network configuration (genesis, fork versions, fork boundaries, time parameters).

And as an implementer of the Spec, I face the following issues:
- Some configuration literally never changes. For example, signing domains should be constants (signing also includes the fork digest, so chains are already unique). Dynamic configuration for some of these things is inconvenient.
- Some configuration is dangerous if configured inconsistently. There are implementation assumptions between the relation of many of the constants. For mainnet and minimal configurations these are well known, but e.g. setting a target-committee size to 1 would likely not go well.
- Some configuration affects the size and shape of the SSZ types. Due to optimizing things, or using stricter typing for compile-time guarantees, this configuration can only be applied during compile-time.

## Current client behavior

- Lighthouse:
  - Supports both `mainnet` and `minimal` in the same build
  - Fails to load other configurations which are not compatible for compile-time parts of the config, and relies on `CONFIG_NAME` (recently removed for other reasons) to detect the base config
- Nimbus:
  - Had (has?) an undocumented JSON configuration feature to override the more dynamic parts of configuration during runtime. (I couldn't get this to work recently)
  - Supports "arbitrary" configurations, but only as part of the build-process. This makes repeated local short-lived testnets very difficult.
- Prysm:
  - Has build variants for `mainnet` and `minimal` SSZ values
  - Supports "arbitrary" configuration loading, but may fail if compile-time SSZ code does not match
- Teku:
  - Supports relatively dynamic configuration. Java is does not have the same constraints of compile-time variables
  - Like others, does not document which config var can be changed to what extent

## New proposal

Without changing the meaning of any of the configuration variables, we organize them in 3 types of variables:

### Constants

We retain all existing constants, and freeze some additional configuration:
- All `DOMAIN_...` variables: signing is already unique thanks to the fork-digest part. Various external tooling/smart-contracts also need stability here.
- Withdrawal prefix bytes: not necessary to modify between different chains

### Preset

A preset is defined as a subset of the previous configuration, bundled together statically, intended for different operation modes: `mainnet`: for full mainnet-like operation, `minimal`: faster/simplified local testing, etc.

A preset:
- Reduces the compile-time build targets to 2: `mainnet` and `minimal`
- Preserves consistency between core chain configuration variables
- Enables comile-time optimization and security improvements
- Makes the runtime configuration much more lightweight: just change `PRESET_BASE` as user. The implementation can then simply check if this preset is allowed.
- Contains all variables that change any of the `BeaconState` or `BeaconBlock` types, but still need to be configurable for testing reasons.

### Configuration

A configuration is a set of runtime-configurable variables that enable users to set up custom local/test networks.

A configuration specifies a preset with `PRESET_BASE`. This can be either `mainnet`, `minimal`, or optionally refer to a preset provided by the client implementation.

For new testing of forks, we want to start defining this configuration **per testcase**. Testing the fork-boundaries is really important, and this enables us to do that, without re-compiling the client for every test case.

The configuration contains:
- Genesis parameters
- Fork versioning and fork boundaries for different network instances
- Special boundaries like the Merge transition POW difficulty
- The slot time, to speed up or stress local testing
- Eth1 related configuration: follow distance, chain/network ID, deposit contract address
- Churn limits, withdrawal, and exit times: for testnets that like to onboard validators quicker, and forcefully throw out inactive validators faster.
- Maybe in the future: P2P Network related parameters (TODO: TBD! Can help with some stress testing of clients, necessary when changing the slot time)

## TODO

- [x] Update specs:
  - [x] Phase0
  - [x] Altair
  - [x] Merge
  - [x] Sharding
  - [x] DAS
  - [x] Custody game
- [x] Update configs
  - [x] Document `PRESET_BASE`
  - [x] Update config files to separate preset and configuration
- [x] Update pyspec
  - [x] ~~Update the way configuration is loaded~~ Pyspec build outputs for each preset
  - [x] Update pytest to handle new config module
  - [x] Update test generators to select configs/presets properly.
  - [x] Update the way configuration is added to the test-vector outputs
    - [x] yield config data from within config decorator
    - [x] document config format for tests
  - [x] Test new improved config loading when using `eth2spec` as python package
  - [x] Double check test generator output

### Later TODOs

- Apply to more altair tests
- Backport the configs of live testnets: Pyrmont and Prater

## Pyspec changes

- The spec builder reads the preset YAML files in addition to the markdown, to build `mainnet.py` and `minimal.py` instead of a `spec.py` that needed reloading.
- The config variables can stay globals in the markdown (for now at least), the spec builder translates `FOO_BAR` to `config.FOO_BAR`
- Config variables are not used in types, only referenced in functions. Use constants or presets for typing.
- The test decorators can swap out the `config` value of a spec module temporarily, to change the config efficiently.
- `with_config_overrides` and `spec_configured_state_test` can be used to configure individual spec tests
- No module reloading necessary anymore


mcdee:
Just a thought, but is minimal so different from mainnet that it has significant value?  Could minimal be discarded entirely without significant impact on testing?

The main reductions in minimal compared to mainnet seem to be bulk and periodicity.  Bulk relates to the number of validators and related items (committees etc.) and it doesn't seem that mainnet parameters are unreasonable for testing.  Periodicity relates to the frequency of changes (slots per epoch, epochs per voting period etc.) and could be more important for transition testing, but are tests in these areas common enough to warrant the complexity of a second config?

Note that I haven't done much testing that has required switching to minimal, so it may be that there are very good reasons for having it.  Just thought that as the issue is being looked at it's worth asking the question.

protolambda:
@mcdee

> is minimal so different from mainnet that it has significant value? Could minimal be discarded entirely without significant impact on testing?

For spec-testing, yes it is of significant value. But outside of spec-testing, not really. This PR tries to address that: we enable deeper customization during compile-time for testing purposes, but remove the clutter and client-inconsistencies from the standard config. The problem with the mainnet configuration is that it increases the state size by a lot, and slows down things by a lot. Bulk and Periodicity indeed. Running days worth of mainnet-state-transitions in CI takes ages, especially in the unoptimized python spec version. The minimal configuration allows us to run through every spec test, on every change to the specs, catching a lot of issues before they hit implementers.

> it doesn't seem that mainnet parameters are unreasonable for testing

Slight disagree here. We do have some mainnet testing (it's a simple `--config=mainnet` in pytest), and do output test-vectors for it, but there have been exceptions where computation or size are too prohibitive to produce a testcase with mainnet configuration. Having the option of a more minimal standardized variant is very useful. And clients can quickly run the minimal config version of tests, before they run the mainnet version.

> Periodicity relates to the frequency of changes (slots per epoch, epochs per voting period etc.) and could be more important for transition testing, but are tests in these areas common enough to warrant the complexity of a second config?

They may not be common, but I think they are important enough to warrant it. And 1 compile-time variant with different constants is not unreasonable complexity to me. This PR attempts to make the separation with runtime configuration clear, to reduce any complexity leak there.

tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -361,6 +363,38 @@ def wrapper(*args, spec: Spec, **kw):
     return decorator
 
 
+def with_config_overrides(config_overrides):
```
protolambda, line 366: Added this function for testing, but refrained from using it in a test, I'm not happy with it yet. Swapping configuration and reloading specs per test case is not pretty, at all. If the spec was an object and the config was a property, this would be much cleaner, but it may hurt readability in the markdown docs.

protolambda:
Don't think we should include this in the next release yet. I am happy with the new config approach, but the pyspec part should improve more. The goal is to decorate test cases with fork versions / fork boundaries, and although it enables that, I'm not happy with the spec reloading part.

protolambda:
Rebased onto latest `dev`.

The current plan is to update the pyspec builder to output a `mainnet.py` and `minimal.py` for each phase, with presets loaded at compile-time, like clients. And then put the configuration variables either in a `config` dict (and reference with e.g. `config.ALTAIR_FORK_VERSION` in spec) or translate those runtime config references. This way we:
- Don't need python module reloading anymore (yay speed)
- Can easily swap fork versions and such per test case.
- Minimal spec / test changes

And, after the spec clarifies the (relatively small) subset of runtime configuration, we can specify runtime-overrides per test case, to properly test different config values, and ease the fork scheduling problems (the stub values for altair fork version should be overridden in block processing tests, so Altair is actually active)

configs/mainnet_config.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+# Altair
+ALTAIR_FORK_VERSION: 0x01000000
+ALTAIR_FORK_EPOCH: 18446744073709551615
+# Merge
+MERGE_FORK_VERSION: 0x02000000
+MERGE_FORK_EPOCH: 18446744073709551615
+# Sharding
+SHARDING_FORK_VERSION: 0x03000000
+SHARDING_FORK_EPOCH: 18446744073709551615
```
hwwhww, line 32: Follow up https://github.com/ethereum/eth2.0-specs/pull/2323#pullrequestreview-640308362
To distinguish the "configs for test vectors" context and "configs for the network" context, it would be nice to make `*_FORK_EPOCH` live in a separate file.

What do you think about having a `mainnet_fork.yaml` which is only for `*_FORK_EPOCH`?

protolambda: Let's wait till further refactoring. I'm updating the test cases etc. to support "overrides" first: just write which piece of runtime-configuration is different, per test case. We can make fork testing much easier this way. And I honestly like the single config file better: if it's not too large, and runtime-only, it's just a lot more ergonomic for testnet configuration work.

tests/core/pyspec/eth2spec/test/altair/unittests/test_config_override.py
```diff
@@ -0,0 +1,19 @@
...
+})
+def test_config_override(spec, state):
+    assert spec.config.ALTAIR_FORK_EPOCH == 4
+    assert spec.config.GENESIS_FORK_VERSION != spec.Version('0x00000000')
+    assert spec.config.GENESIS_FORK_VERSION == spec.Version('0x12345678')
+    assert spec.config.ALTAIR_FORK_VERSION == spec.Version('0x11111111')
+    assert state.fork.current_version == spec.Version('0x11111111')
+    # TODO: it would be nice if the create_genesis_state actually outputs a state
+    #  for the fork with a slot that matches at least the fork boundary.
+    # assert spec.get_current_epoch(state) >= 4
```
protolambda, line 19: This PR enables efficient separate runtime-configuration per test case. A `config.yaml` is yielded to the test vector output with the merged result (config base of preset + overrides). Client implementers can load the runtime configurable part (now that we separated it from the preset/constants), and run through the configured test.

tests/core/pyspec/eth2spec/utils/hash_function.py
```diff
@@ -6,4 +6,8 @@
 
 
 def hash(x: Union[bytes, bytearray, memoryview]) -> Bytes32:
-    return Bytes32(sha256(x).digest())
+    try:
```
djrtwo, line 9: Did you run into an issue hashing something?

tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -365,6 +393,45 @@ def wrapper(*args, spec: Spec, **kw):
...
+
+            # apply our overrides to a copy of it, and apply it to the spec
+            tmp_config = deepcopy(old_config._asdict())  # not a private method, there are multiple
+            tmp_config.update(config_overrides)
+            config_types = spec.Configuration.__annotations__
+            # Retain types of all config values
+            test_config = {k: config_types[k](v) for k, v in tmp_config.items()}
+
+            # Output the config for test vectors  (TODO: check config YAML encoding)
+            yield 'config', 'data', test_config
```
djrtwo, line 417: Ah, so you read `config.yaml` for configuration overrides in test vectors?

djrtwo (commented):
Wow! Really ripples deep in there...

Generally okay with the approach. Need to make sure it's easy for client teams to digest wht actually changed here maybe with a cleaned up header comment. The diff isn't particularly useful to them.

I had a few questions about config vs preset on some vars. Check it out. Happy to debate

tests/core/pyspec/eth2spec/utils/hash_function.py
protolambda: Ah yes, lost track of some debug thing. It's the config that's hashable. Changed it to NamedTuple later (stricter validation, simple to embed as key for caches). TypedDict isn't strict enough. And a dataclass, even with its special decorator options, doesn't work: its generated `__hash__` calls the global `hash`, and it turned out to call the SSZ hash function, for the config data as a whole, which doesn't work.

tests/core/pyspec/eth2spec/test/context.py
protolambda: Yes, just read the runtime config (if any is there) from the test case dir, and good to go. This should be the same in all test vectors, so it makes sense to yield it from within the config decorator.

setup.py
```diff
@@ -124,10 +138,26 @@ def _get_eth2_spec_comment(child: LinkRefDef) -> Optional[str]:
     return title[len(ETH2_SPEC_COMMENT_PREFIX):].strip()
 
 
-def get_spec(file_name: str) -> SpecObject:
+def _parse_value(name: str, typed_value: str) -> VariableDefinition:
+    comment = None
+    if name == "BLS12_381_Q":
+        comment = "  # noqa: E501"
```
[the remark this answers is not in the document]
protolambda: good point, was moving the old `if` body, but will improve it.

configs/mainnet_config.yaml
```diff
@@ -0,0 +1,73 @@
...
+# Time parameters
+# ---------------------------------------------------------------
+# 12 seconds
+SECONDS_PER_SLOT: 12
+# 14 (estimate from Eth1 mainnet)
+SECONDS_PER_ETH1_BLOCK: 14
+# 2**8 (= 256) epochs ~27 hours
+MIN_VALIDATOR_WITHDRAWABILITY_DELAY: 256
+# 2**8 (= 256) epochs ~27 hours
+SHARD_COMMITTEE_PERIOD: 256
```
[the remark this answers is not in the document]
protolambda: It is used in phase0 for the validator exit queue, so if a testnet wants to allow quick validator exits, it needs to modify this. I agree that for sharding itself it doesn't make a whole lot of sense to not put it in the preset. Maybe we should split the two config variables? I.e. define `VALIDATOR_EXIT_DELAY`, and update phase0 without changing the value?

configs/mainnet_preset/altair.yaml
```diff
@@ -0,0 +1,24 @@
+# Mainnet preset - Altair
```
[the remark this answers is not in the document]
protolambda: sounds good, will split into `presets/` and `configs/` then

configs/minimal_preset/phase0.yaml
```diff
@@ -68,19 +36,8 @@ EJECTION_BALANCE: 16000000000
 EFFECTIVE_BALANCE_INCREMENT: 1000000000
```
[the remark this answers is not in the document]
protolambda: Will move to config

specs/altair/beacon-chain.md
```diff
@@ -122,20 +131,14 @@ This patch updates a few configuration values to move penalty parameters closer
...
+### Inactivity penalties
 
-| Name | Value |
-| - | - |
-| `DOMAIN_SYNC_COMMITTEE` | `DomainType('0x07000000')` |
-| `DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF` | `DomainType('0x08000000')` |
-| `DOMAIN_CONTRIBUTION_AND_PROOF` | `DomainType('0x09000000')` |
+| Name | Value | Description |
+| - | - | - |
+| `INACTIVITY_SCORE_BIAS` | `uint64(4)` | score points per inactive epoch |
```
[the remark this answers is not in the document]
protolambda: We could make the phase0 variants of inactivity vars configurable, but I imagine we'll have "duplicate" config vars then, and they won't matter for testnets that start post-altair.

presets/README.md
```diff
@@ -0,0 +1,26 @@
+## Presets
```
djrtwo, line 1:
```suggestion
# Presets
```

djrtwo (commented):
fix markdown header

hwwhww (commented):
Some thoughts, not strongly:
* 👍 on more structural pyspec
* 👍 on making test case variable configurable by decorator. Nice work!
*  [❓]  A little concerned about maintainability. Variables might be moved between config and preset later when we need.
	* `configs/mainnet.yaml` will contain the configuration cross forks, including the ones in active R&D. IIRC, the client devs don’t like unstable config file that includes variables for the future forks.
	* 👎 on some variables are `spec.VARIABLE_NAME` and some are `spec.config.VARIABLE_NAME` in tests. But maybe I will get used to it.
* [❓ ] The specs were written for the client devs mostly, but it also serves general readers to understand Eth2 protocol. This change exposes more implementation details by separating config and preset.

I'd defer to client devs to make the call. 😄

tests/core/pyspec/eth2spec/utils/hash_function.py
```diff
@@ -6,4 +6,8 @@
...
+        return Bytes32(sha256(x).digest())
+    except TypeError:
+        print(x)
```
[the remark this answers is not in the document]
protolambda: This was just a temporary debugging thing, I reverted it

protolambda:
Diff of test generation with `dev` commit, base of this PR: https://gist.github.com/protolambda/48349d01e6101e6f433bcae703caddc4

- Altair finality tests were missing on dev
- Fork tests were missing on dev due to error (recently fixed in https://github.com/ethereum/eth2.0-specs/pull/2427 )
- The attester slashing test `success_with_effective_balance_disparity` is different for some reason, investigating that one.

setup.py
```diff
@@ -649,28 +729,75 @@ def combine_spec_objects(spec0: SpecObject, spec1: SpecObject) -> SpecObject:
...
-    functions = combine_functions(functions0, functions1)
-    custom_types = combine_constants(custom_types0, custom_types1)
-    constants = combine_constants(constants0, constants1)
-    ssz_dep_constants = combine_constants(ssz_dep_constants0, ssz_dep_constants1)
-    ssz_objects = combine_ssz_objects(ssz_objects0, ssz_objects1, custom_types)
-    dataclasses = combine_functions(dataclasses0, dataclasses1)
+    protocols = combine_protocols(spec0.protocols, spec1.protocols)
+    functions = combine_functions(spec0.functions, spec1.functions)
+    custom_types = combine_constants(spec0.custom_types, spec1.custom_types)
+    constant_vars = combine_variables(spec0.constant_vars, spec1.constant_vars)
```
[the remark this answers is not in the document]
protolambda: It's really just a `dict(.....).update(dict(....))`, but I think not doing it in-place will be less error prone in the future, so I changed it to a generic `combine_dicts` that does `(x, y): {**x, **y}`, and removed the duplicates (functions, constants, variable-defs all had such combination function).

configs/README.md
```diff
@@ -1,34 +1,39 @@
...
+and then load different configurations around that preset to participate in different networks or tests.
 
-A preset file contains all the constants known for its target.
-Later-fork constants can be ignored, e.g. ignore Sharding constants as a client that only supports Phase 0 currently.
+Standard configs:
+- [`mainnet.yaml`](./mainnet.yaml): Mainnet configuration
+- [`minimal.yaml`](./minimal.yaml): Minimal configuration, used in spec-testing along with the [`minimal`](../presets/minimal) preset.
 
+Not all network configurations are in scope for the specification,
+see [`github.com/eth2-clients/eth2-networks`](https://github.com/eth2-clients/eth2-networks) for common networks,
```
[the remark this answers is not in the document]
protolambda: Undead soon, see: https://github.com/eth2-clients/eth2-testnets/issues/46

tests/core/pyspec/eth2spec/test/context.py
protolambda: This is resolved, no module reloading necessary anymore, function has been updated.

specs/phase0/beacon-chain.md
```diff
@@ -1556,7 +1580,10 @@ def process_registry_updates(state: BeaconState) -> None:
         if is_eligible_for_activation_queue(validator):
             validator.activation_eligibility_epoch = get_current_epoch(state) + 1
 
-        if is_active_validator(validator, get_current_epoch(state)) and validator.effective_balance <= EJECTION_BALANCE:
+        if (
+            is_active_validator(validator, get_current_epoch(state))
+            and validator.effective_balance <= EJECTION_BALANCE
+        ):
```
ralexstokes, line 1586: this was just a formatting change?

im assuming the linter was happy either way

protolambda: It changes to `config.EJECTION_BALANCE` by the pyspec builder, and gets too long for the line.

ralexstokes (approved):
there's a lot in here but generally supportive of the change!

ultimately defer to various client teams on how to handle this but do like differentiating the more/less statics parts of the "config"

configs/README.md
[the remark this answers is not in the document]
ralexstokes: ah great!

michaelsproul (commented):
I'm in favour of clearly defining the config file semantics, and I think this PR does a good job of that.

I skimmed through Lighthouse's compile-time config (`EthSpec`) and runtime config (`ChainSpec`) to see how it matched up and didn't spot any major issues. There are just a few consts/preset-level values in the ChainSpec, which I've documented below.

I don't think we have the appetite for a large refactor at the moment, so if this PR is adopted we'll likely continue mapping standard configs onto Lighthouse's internal representation, and then slowly migrate towards using the standard internally.

Other minor points:

* There's no central definition of constants, which could be useful to have
* I agree with @hwwhww that having clarity about what forks are being parsed from a config file is desirable. There aren't any Altair/merge/sharding params in the _config_ right now, would the intention be to mix them in with the phase0 params, or have them in separate files like the presets?
* Backwards-compatibility for the `/eth/v1/config/spec` HTTP API is also something to consider. Right now Lighthouse & Teku interoperate OK using the current config. If this PR is adopted I think the best chance of achieving backwards compatibility will be to require the API endpoint to returning the union of the preset & config?

<details><summary>ChainSpec differences</summary>

## Consts in ChainSpec:

* genesis_slot
* far_future_epoch
* base_rewards_per_epoch
* deposit_contract_tree_depth
* min_per_epoch_churn_limit
* churn_limit_quotient
* bls_withdrawal_prefix_byte
* all the domains

## Preset params in ChainSpec:

* max_committees_per_slot
* target_committee_size
* shuffle_round_count
* hysteresis_quotient
* hysteresis_downward_multiplier
* hysteresis_upward_multiplier
* proportional_slashing_multiplier
* min_deposit_amount
* max_effective_balance
* ejection_balance
* effective_balance_increment
* min_attestation_inclusion_delay
* min_seed_lookahead
* max_seed_lookahead
* min_epochs_to_inactivity_penalty
* base_reward_factor
* whistleblower_reward_quotient
* proposer_reward_quotient
* inactivity_penalty_quotient
* min_slashing_penalty_quotient
* safe_slots_to_update_justified

</details>

ajsutton:
Definitely makes sense to me - being able to setup custom testnets more easily is definitely a win.

This doesn't really make any difference for Teku as all the config options from the spec are completely dynamic for us anyway (ie minimal and mainnet are both fully supported in all builds, and we actually have a couple of other variants we use to make it possible for acceptance tests to finalise quickly).  So the change for us ultimately just boils down to loading the values from more files and supporting the "PRESET_BASE" field to specify which preset file to mix in.

@michaelsproul makes a good point about `/eth/v1/config/spec` - for compatibility we should flatten the options and then I think it winds up producing the same output.  It's probably worth including the additional `PRESET_BASE` value in the output as then you can create the new config file out of the response by just ignoring anything that's now in PRESET.  We may want to consider a `/eth/v2/config/spec` endpoint which specifically separates the preset and config values.

wemeetagain:
While the preset vs config distinction could be formulated at the implementation-level, rather than spec-level, I think including this in the spec gives more clarity about how much certain params are expected to change and will help with more easily (and consistently) setting up testnets across clients.
 
I think this will be a big win for Lodestar, which currently has teku-level of dynamic behavior. We can likely simplify a lot of our code and improve devex once our ssz types are known ahead of time and no longer depend on run-time config.

djrtwo (approved):
putting it in alpha.6. great work!

PR #2396, opened 2021-05-10
Shard fees: `adjustment_quotient` to use `previous_epoch`

terencechain:
I think `adjustment_quotient`'s shard count should reference to previous epoch similar to accounting of header confirmation and fee calculation


djrtwo (approved):
yep, thanks!

PR #2398, opened 2021-05-10
Define execution engine protocol

protolambda:
The execution was previously too fuzzy, and the required implementation methods were missing, only referenced through text.

This changes the spec to allow for `Protocol` definitions. A `Protocol` is a python feature, and useful to interface with some abstract protocol specification like we do with the execution sub-system.

By defining it this way, implementers will have a better time connecting it to a well-encapsulated local implementation, or API if they use an external execution engine, like done during Rayonism testnets with existing Eth1 implementations.

The idea is that the methods of the interface are defined like any other methods in the spec, except with a `...` body (common for python `Protocol` definitions), and a `self: ProtocolNameHere` as first argument.

TODO:
- [x] Update pyspec builder to construct the `class ExecutionEngine(Protocol):` definition from the individual function definitions.
- ~~Rayonism spec says `finalise_block`, instead of `finalize_block`. For finalization, we use the `z` everywhere in eth2 specs so far, and I think we shouldn't change that.~~ rayonism spec is going to change anyway when moving to Merge production API
- [x] Define a default `EXECUTION_ENGINE`, to run the spec with in general
- [x] Update pyspec testing, to define a mock execution-engine, to not have to do hacks with global variables. We can pass it into the `process_execution_payload` to write clean execution-payload tests (cc @mkalinin). Similarly, we can do this for later validator-spec testing (`get_execution_payload`)

### DIFF
--- a/specs/merge/fork-choice.md
+++ b/specs/merge/fork-choice.md
@@ -8,11 +8,16 @@
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
 - [Introduction](#introduction)
-  - [Helpers](#helpers)
+- [Protocols](#protocols)
+  - [`ExecutionEngine`](#executionengine)
+    - [`set_head`](#set_head)
+    - [`finalize_block`](#finalize_block)
+- [Containers](#containers)
     - [`PowBlock`](#powblock)
+- [Helper functions](#helper-functions)
     - [`get_pow_block`](#get_pow_block)
     - [`is_valid_transition_block`](#is_valid_transition_block)
-  - [Updated fork-choice handlers](#updated-fork-choice-handlers)
+- [Updated fork-choice handlers](#updated-fork-choice-handlers)
     - [`on_block`](#on_block)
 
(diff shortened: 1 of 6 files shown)


specs/merge/fork-choice.md
```diff
@@ -24,7 +29,44 @@ This is the modification of the fork choice according to the executable beacon c
...
+
+#### `set_head`
+
+Re-organizes the execution payload chain and corresponding state to make `block_hash` the head.
+
+The body of this function is implementation dependent.
+The Consensus API may be used to implement this with an external execution engine.
+
+```python
+def set_head(self: ExecutionEngine, block_hash: Hash32) -> bool:
```
mkalinin, line 46: This function is somewhat out of the spec and will unlikely become a part of it in the future. Do we really want to keep it here?

protolambda: It's not called in the eth2 specs yet, I just added it here for completeness. If we're not going to have it, we can leave it out. I still like the idea of updating the head of an external execution-client though, it's important to detect reorgs without the chain being live (i.e. no new block insertions to keep the head updated).

mkalinin (approved):

PR #2399, opened 2021-05-10
dankrad altair review -- extended and tested

djrtwo:
supersedes @dankrad's #2387

* remove double applied non-quadratic leak
* add `PARTICIPATION_FLAG_WEIGHTS` and update logic in favor of `get_flag_indices_and_weights`
* Remove timely-head penalty, only reward now
* fix a couple of minor comments


specs/altair/beacon-chain.md
```diff
@@ -367,35 +354,31 @@ def get_unslashed_participating_indices(state: BeaconState, flag_index: int, epo
...
+    unslashed_participating_increments = unslashed_participating_balance // EFFECTIVE_BALANCE_INCREMENT
+    active_increments = get_total_active_balance(state) // EFFECTIVE_BALANCE_INCREMENT
     for index in get_eligible_validator_indices(state):
         base_reward = get_base_reward(state, index)
         if index in unslashed_participating_indices:
-            if is_in_inactivity_leak(state):
-                # This flag reward cancels the inactivity penalty corresponding to the flag index
-                rewards[index] += Gwei(base_reward * weight // WEIGHT_DENOMINATOR)
-            else:
+            if not is_in_inactivity_leak(state):
```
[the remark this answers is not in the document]
djrtwo: This is not actually the logic we want due to the next else.

being in inactivity leak is not enough to trigger the penalty, but if we do `if index in unslashed_participating_indices and not is_in_inactivity_leak(state):` then any time we were in inactivity leak, the `elif` below would be triggered to give penalties

specs/altair/beacon-chain.md
```diff
@@ -203,6 +203,7 @@ class BeaconState(Container):
 ```python
+    # TODO! Need multiple signatures as discussed between Justin and Vitalik May 3 2021 (see Telegram)
```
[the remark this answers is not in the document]
djrtwo: right

specs/altair/beacon-chain.md
```diff
@@ -367,35 +353,31 @@ def get_unslashed_participating_indices(state: BeaconState, flag_index: int, epo
...
         if index in unslashed_participating_indices:
-            if is_in_inactivity_leak(state):
-                # This flag reward cancels the inactivity penalty corresponding to the flag index
-                rewards[index] += Gwei(base_reward * weight // WEIGHT_DENOMINATOR)
-            else:
+            if not is_in_inactivity_leak(state):
                 reward_numerator = base_reward * weight * unslashed_participating_increments
                 rewards[index] += Gwei(reward_numerator // (active_increments * WEIGHT_DENOMINATOR))
-        else:
+        elif not flag_index == TIMELY_HEAD_FLAG_INDEX:
```
hwwhww, line 374: IMHO it's more concise:
```suggestion
        elif flag_index != TIMELY_HEAD_FLAG_INDEX:
```

ralexstokes: i was also leaning towards this!

ralexstokes (approved):

specs/altair/beacon-chain.md
```diff
@@ -275,8 +261,8 @@ def has_flag(flags: ParticipationFlags, flag_index: int) -> bool:
 ```python
     """
-    Return the sequence of sync committee indices (which may include duplicate indices)
-    for a given ``state`` and ``epoch``.
+    Return the sequence of sync committee indices for a given ``state`` and ``epoch``.
+    Note: Committee can contain duplicate indices for small validator sets (< 2 * SYNC_COMMITTEE_SIZE)
```
dankrad, line 265:
```suggestion
    Note: Committee can contain duplicate indices for small validator sets (< SYNC_COMMITTEE_SIZE + 128)
```

Issue #2397, opened 2021-05-10
`process_sync_committee` uses incorrect sync committee for first slot of new committee period


ajsutton:
Sorry if I've missed something here but from my understanding, the process for creating a `SyncAggregate` for a block at slot `x` is to collect the best `SyncCommitteeContribution` from slot `x - 1` and aggregate them as per `process_sync_committee_contributions` from https://github.com/ethereum/eth2.0-specs/blob/dev/specs/altair/validator.md#block-proposal.

Then during block processing, `process_sync_committee` uses `state.current_sync_committee` where `state` has already had slots processed up to and including `x`.

Given that, the aggregate indices in the `SyncAggregate` will refer to the sync committee at slot `x - 1` but if `x` is the first slot of a new sync committee period, `state.current_sync_committee` will have already been updated and now point to the new sync committee.  The wrong validator public keys will then be collected in `participant_pubkeys` and signature validation will fail.

djrtwo:
yep, this looks correct

I'm doing some testing and a fix right now

djrtwo:
Because we get rid of `previous_sync_committee` in state but always know `current` and `next`, the job for the `next` needs to begin 1 slot prior to the start of the period so that they can be included at the 0th slot of the period.

That is, the `current_sync_committee` votes on blocks from slots ranging `[period_start_slot - 1, period_end_slot)` and is included in blocks with slots ranging  `[period_start_slot, period_end_slot]`

PR #2400, opened 2021-05-10
 add logic for handling sync committee off by one issue

djrtwo:
addresses #2397

This is not yet comprehensive and I'm not really happy with how to explain the solution. Looking for feedback.

Essentially, being assigned to `slot` means you need to sign and broadcast for `slot - 1`.

### DIFF
--- a/specs/altair/p2p-interface.md
+++ b/specs/altair/p2p-interface.md
@@ -106,9 +106,18 @@ The following validations MUST pass before forwarding the `signed_contribution_a
 
 ```python
 def get_sync_subcommittee_pubkeys(state: BeaconState, subcommittee_index: uint64) -> Sequence[BLSPubkey]:
+    # Committees assigned to `slot` sign for `slot - 1`
+    # This creates the exceptional logic below when transitioning between sync comittee periods
+    next_slot_epoch = compute_epoch_at_slot(state.slot + 1)
+    if compute_sync_committee_period(get_current_epoch(state)) == compute_sync_committee_period(next_slot_epoch):
+        sync_committee = state.current_sync_committee
+    else:
+        sync_committee = state.next_sync_committee
+
+    # Return pubkeys for the subcommittee index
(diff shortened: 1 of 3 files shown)


ralexstokes (commented):
looks good to me -- sync committee members will have to start their task a slot early compared to when it may be understood just by reading the state but i don't see a way around this as these members are expected to sign over the previous slot so will hit this boundary condition when switching from one period to the other

ajsutton (commented):
This solves the original problem but it's a very weird quirk.  I've had a bit of a go at implementing this in Teku and it's a bit messy particularly around the actual fork slot. It's likely to be equally messy around the next fork that changes sync committees as well. 
We need to be very clear about which fork is used during those transitions. ie I'm creating a signature for slot 31 because I'm scheduled to be in a sync committee from slot 32 - do I use the rules of the fork for slot 31 (epoch 0) or 32 (epoch 1)? It's also a lot harder to reason about which state can be used to validate signatures and contributions because they aren't always using the committee for the epoch they're actually from. It's one of those weird quirks that's going to keep catching people by surprise I suspect.

That said, Teku is now working and performing all duties as expected. I suspect there are a few more edge cases around which will add some further complexity but I think it will be manageable. 

And I can't really think of another way to solve this other than to just give up and make the `SyncAggregate` for the first slot of each sync committee always be empty which is simple but also a bit rubbish.

hwwhww (approved):
I updated the test cases and minor rephases on the parameters. The solution does seem not elegant but I couldn't figure out another way to solve it at lookahead and at processing the proof. :'(

Thanks, @ajsutton @djrtwo!

ajsutton:
I guess the one other obvious option is to add `previous_sync_committee` to the state. It's not huge but is a bit of a shame to add it just for this.

djrtwo:
> I guess the one other obvious option is to add previous_sync_committee to the state. It's not huge but is a bit of a shame to add it just for this.

Right. The main issue there is that you then have exceptional for a single slot per period in the state transition

PR #2405, opened 2021-05-11
Only allow sync committee period calculation at period boundaries

ralexstokes:
It only makes sense to calculate sync committees during period boundaries as inputs to the calculation can otherwise change during sync committee periods.

This suggests we should restrict when we can do this via `asserts` in the spec as demonstrated in this PR.

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -276,16 +276,14 @@ def has_flag(flags: ParticipationFlags, flag_index: int) -> bool:
 def get_sync_committee_indices(state: BeaconState, epoch: Epoch) -> Sequence[ValidatorIndex]:
     """
     Return the sequence of sync committee indices (which may include duplicate indices)
-    for a given ``state`` and ``epoch``.
-
-    Note: This function is not stable during a sync committee period as
-    a validator's effective balance may change enough to affect the sampling.
+    for a given ``state`` and ``epoch`` at a sync committee period boundary.
     """
+    assert epoch % EPOCHS_PER_SYNC_COMMITTEE_PERIOD == 0
+
     MAX_RANDOM_BYTE = 2**8 - 1
-    base_epoch = Epoch((max(epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD, 1) - 1) * EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
-    active_validator_indices = get_active_validator_indices(state, base_epoch)
+    active_validator_indices = get_active_validator_indices(state, epoch)
     active_validator_count = uint64(len(active_validator_indices))
-    seed = get_seed(state, base_epoch, DOMAIN_SYNC_COMMITTEE)
+    seed = get_seed(state, epoch, DOMAIN_SYNC_COMMITTEE)
     i = 0
     sync_committee_indices: List[ValidatorIndex] = []
     while len(sync_committee_indices) < SYNC_COMMITTEE_SIZE:
--- a/specs/altair/fork.md
+++ b/specs/altair/fork.md
@@ -81,7 +81,9 @@ def upgrade_to_altair(pre: phase0.BeaconState) -> BeaconState:
         inactivity_scores=[uint64(0) for _ in range(len(pre.validators))],
     )
     # Fill in sync committees
-    post.current_sync_committee = get_sync_committee(post, get_current_epoch(post))
-    post.next_sync_committee = get_sync_committee(post, get_current_epoch(post) + EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
+    current_period = epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+    base_epoch = current_period * EPOCHS_PER_SYNC_COMMITTEE_PERIOD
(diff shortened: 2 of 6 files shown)


specs/altair/beacon-chain.md
```diff
@@ -276,16 +276,16 @@ def has_flag(flags: ParticipationFlags, flag_index: int) -> bool:
...
-    base_epoch = Epoch((max(epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD, 1) - 1) * EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
-    active_validator_indices = get_active_validator_indices(state, base_epoch)
+
+    if epoch != ALTAIR_FORK_EPOCH or epoch != ALTAIR_FORK_EPOCH + EPOCHS_PER_SYNC_COMMITTEE_PERIOD:
+        assert epoch % EPOCHS_PER_SYNC_COMMITTEE_PERIOD == 0
+
+    active_validator_indices = get_active_validator_indices(state, epoch)
     active_validator_count = uint64(len(active_validator_indices))
-    seed = get_seed(state, base_epoch, DOMAIN_SYNC_COMMITTEE)
+    seed = get_seed(state, epoch, DOMAIN_SYNC_COMMITTEE)
```
ralexstokes, line 288: this ensures the `epoch` is aligned or part of the special case at the altair fork epoch.

it also removes the prior "base epoch" computation with the only substantial change being that the committees computed right at the fork will have "non-aligned" `seed` selection from the state. imo fine given the simplification it enables

specs/altair/fork.md
```diff
@@ -81,7 +81,9 @@ def upgrade_to_altair(pre: phase0.BeaconState) -> BeaconState:
         inactivity_scores=[uint64(0) for _ in range(len(pre.validators))],
     )
     # Fill in sync committees
-    post.current_sync_committee = get_sync_committee(post, get_current_epoch(post))
-    post.next_sync_committee = get_sync_committee(post, get_current_epoch(post) + EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
+    current_period = epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+    base_epoch = current_period * EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+    post.current_sync_committee = get_sync_committee(post, base_epoch)
+    post.next_sync_committee = get_sync_committee(post, base_epoch + EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
```
ralexstokes, line 87: we can regularize `get_sync_committee_indices` in the spec by doing this one-off "base epoch" calculation here instead

tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_sync_committee_updates.py
```diff
@@ -39,8 +39,7 @@ def run_sync_committees_progress_test(spec, state):
 
     # Can compute the third committee having computed final balances in the last epoch
     # of this `EPOCHS_PER_SYNC_COMMITTEE_PERIOD`
-    current_epoch = spec.get_current_epoch(state)
-    third_sync_committee = spec.get_sync_committee(state, current_epoch + 2 * spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
+    third_sync_committee = spec.get_sync_committee(state, (next_period + 1) * spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
```
ralexstokes, line 42: this utility function does not increment `state.slot` so need to be explicit rather than use `current_epoch`

tests/core/pyspec/eth2spec/test/altair/sanity/test_blocks.py
```diff
@@ -18,7 +18,8 @@
 
 
 def run_sync_committee_sanity_test(spec, state, fraction_full=1.0):
-    committee = spec.get_sync_committee_indices(state, spec.get_current_epoch(state))
+    all_pubkeys = [v.pubkey for v in state.validators]
+    committee = [all_pubkeys.index(pubkey) for pubkey in state.current_sync_committee.pubkeys]
     participants = random.sample(committee, int(len(committee) * fraction_full))
```
ralexstokes, line 23: we want to avoid external use of `get_sync_committee_indices` as much as possible

specs/altair/fork.md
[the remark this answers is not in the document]
djrtwo: Ah, I just realized that the usage in `process_sync_committee_updates` is pretty incorrect

```python
```

adding that many epochs actually results in an older seed than if you use current_epoch.
This is actually a bug.

When calculating `next_committee`, I'd argue the `current_epoch` should be passed in here and in `process_sync_committee_updates`. 

Thus when calculating `current_sync_committee` at the fork, you should probably do `previous_period_start_epoch`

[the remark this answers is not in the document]
djrtwo: I'll make a suggested PR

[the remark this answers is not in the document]
djrtwo: Okay, so you introduced a bug here when you removed the `- 1` in `get_sync_committee_indices`. 

I've combined your changes with what I believe fixes this issue in #2406

Issue #2392, opened 2021-05-07
Advice in validator guide about lookahead is contradictory


michaelsproul:
This paragraph in the [validator guide](https://github.com/ethereum/eth2.0-specs/blob/v1.1.0-alpha.3/specs/altair/validator.md#lookahead) implies that sync committees can't be computed except at the period boundary:

> Note: The data required to compute a given committee is not cached in the `BeaconState` after committees are calculated at the period boundaries. This means that calling `get_sync_commitee()` in a given epoch can return a different result than what was computed during the relevant epoch transition. For this reason, always get committee assignments via the fields of the `BeaconState` (`current_sync_committee` and `next_sync_committee`) or use the above reference code.

However I don't think this is the case, because the sync committee indices  only rely on:

1. The seed from the `base_epoch`, which is <1024 epochs back (we have 64k epochs of history in the `randao_mixes` vec)
2. The active validators at the same `base_epoch`, which can be computed accurately at any epoch after the `base_epoch`

Further, the state transition _relies_ on recomputing the sync committees in arbitrary epochs: in [`process_sync_committee`](https://github.com/ethereum/eth2.0-specs/blob/v1.1.0-alpha.3/specs/altair/beacon-chain.md#sync-committee-processing) a call is made to `get_sync_committee_indices`.

The reason for my deep-dive on this issue is that in Lighthouse we've implemented a cache for the committee indices, which can get recomputed at arbitrary points (e.g. if a state is loaded from the DB and lacks the cache it could be built anew). I just wanted to check that we're not missing something that would make this unsound.

ralexstokes:
> the sync committee indices only rely on:

that computation also relies on `effective_balance` which we don't keep historical record of in the `state`. this is the same as with the proposer shufflings and why validators should re-compute proposer shufflings every epoch (rather than say w/ 1 epoch of lookahead) -- its a small chance but the `effective_balance` *could* change enough to shift the shuffling.

this rationale inspired the note in the validator guide where you don't have any lookahead on the sync committee shufflings other than 1 sync committee period bc we track the "next" committee.

>  the state transition relies on recomputing the sync committees 

this may be a bug! we may need to do a 'pubkey to index' lookup here based on `state.current_sync_committee` rather than use `get_sync_committee_indices`.

djrtwo:
resolved in #2394

PR #2406, opened 2021-05-11
remove sync committee seed and restrict period calculation boundaries

djrtwo:
Rename `get_sync_committee` to `get_next_sync_committee` and restrict the usage by removing `epoch` as a parameter. This param was error prone and highly context dependent. If we only want to call in one context, we can get `epoch` from state.

Note, that this reduces flexibility when calculating the initial current and next committee at the fork boundary. For simplicity, these are now set to be equal

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -32,8 +32,8 @@
     - [`add_flag`](#add_flag)
     - [`has_flag`](#has_flag)
   - [Beacon state accessors](#beacon-state-accessors)
-    - [`get_sync_committee_indices`](#get_sync_committee_indices)
-    - [`get_sync_committee`](#get_sync_committee)
+    - [`get_next_sync_committee_indices`](#get_next_sync_committee_indices)
+    - [`get_next_sync_committee`](#get_next_sync_committee)
     - [`get_base_reward_per_increment`](#get_base_reward_per_increment)
     - [`get_base_reward`](#get_base_reward)
     - [`get_unslashed_participating_indices`](#get_unslashed_participating_indices)
@@ -270,15 +270,15 @@ def has_flag(flags: ParticipationFlags, flag_index: int) -> bool:
 
 ### Beacon state accessors
 
-#### `get_sync_committee_indices`
+#### `get_next_sync_committee_indices`
 
 ```python
-def get_sync_committee_indices(state: BeaconState, epoch: Epoch) -> Sequence[ValidatorIndex]:
+def get_next_sync_committee_indices(state: BeaconState) -> Sequence[ValidatorIndex]:
     """
     Return the sequence of sync committee indices (which may include duplicate indices)
-    for a given ``state`` and ``epoch`` at a sync committee period boundary.
+    for the next sync committee, given a ``state`` at a sync committee period boundary.
     """
-    assert epoch % EPOCHS_PER_SYNC_COMMITTEE_PERIOD == 0
+    epoch = Epoch(get_current_epoch(state) + 1)
 
     MAX_RANDOM_BYTE = 2**8 - 1
     active_validator_indices = get_active_validator_indices(state, epoch)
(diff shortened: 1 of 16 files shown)


djrtwo:
actually, false alarm. The `- 1` in the calculation of `base_epoch` in  `get_sync_committee_indices` handles the seed selection properly

```python
base_epoch = Epoch((max(epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD, 1) - 1) * EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
```

Although this seems very error prone.

specs/altair/beacon-chain.md
```diff
@@ -276,16 +276,14 @@ def has_flag(flags: ParticipationFlags, flag_index: int) -> bool:
...
     """
+    assert epoch % EPOCHS_PER_SYNC_COMMITTEE_PERIOD == 0
+
     MAX_RANDOM_BYTE = 2**8 - 1
-    base_epoch = Epoch((max(epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD, 1) - 1) * EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
-    active_validator_indices = get_active_validator_indices(state, base_epoch)
+    active_validator_indices = get_active_validator_indices(state, epoch)
     active_validator_count = uint64(len(active_validator_indices))
-    seed = get_seed(state, base_epoch, DOMAIN_SYNC_COMMITTEE)
+    seed = get_seed(state, epoch, DOMAIN_SYNC_COMMITTEE)
```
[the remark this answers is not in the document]
djrtwo: note the change in the call to `get_sync_committee()`. We used to pass in `next_epoch + EPOCHS_PER_SYNC_COMMITTEE` which was confusing and error prone imo (it was essentailly canceled out in `get_sync_committee_indices` when calculating `base_epoch`.

To simplify, I changed to pass in `next_epoch`.

tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_sync_committee_updates.py
```diff
@@ -40,7 +40,7 @@ def run_sync_committees_progress_test(spec, state):
     # Can compute the third committee having computed final balances in the last epoch
     # of this `EPOCHS_PER_SYNC_COMMITTEE_PERIOD`
     current_epoch = spec.get_current_epoch(state)
-    third_sync_committee = spec.get_sync_committee(state, current_epoch + 2 * spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
+    third_sync_committee = spec.get_sync_committee(state, current_epoch + 1)
```
ralexstokes, line 43: just noting: with the change to get the "next" sync committee in this PR, we only need to move to the next epoch (which happens to be in the next sync committee period) to get the third sync committee

ralexstokes (approved):
generally looks good.

i'd consider changing `get_sync_committee` to `get_next_sync_committee` but otherwise should simplify things further

hwwhww:
[docs] Linking related issues: #2392 #2394

tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_sync_committee_updates.py
hwwhww: FYI we can add constant checks in spec builder `invariant_checks` once we decide the `ALTAIR_FORK_EPOCH`. (the stub value is `18446744073709551615` so it does not satisfy):
https://github.com/ethereum/eth2.0-specs/blob/39924c4f14d0702acc805f0d7799ee33dc81f364/setup.py#L409-L414

specs/altair/beacon-chain.md
```diff
@@ -276,16 +276,14 @@ def has_flag(flags: ParticipationFlags, flag_index: int) -> bool:
 def get_sync_committee_indices(state: BeaconState, epoch: Epoch) -> Sequence[ValidatorIndex]:
     """
     Return the sequence of sync committee indices (which may include duplicate indices)
-    for a given ``state`` and ``epoch``.
-
-    Note: This function is not stable during a sync committee period as
-    a validator's effective balance may change enough to affect the sampling.
+    for a given ``state`` and ``epoch`` at a sync committee period boundary.
     """
+    assert epoch % EPOCHS_PER_SYNC_COMMITTEE_PERIOD == 0
```
hwwhww, line 281: In [validator guide](https://github.com/ethereum/eth2.0-specs/blob/468f42c545a172e8cd369856165dd28a75738e0b/specs/altair/validator.md#aggregation-bits):
> An aggregator finds the index in the sync committee (as returned by `get_sync_committee_indices()`) for a given validator referenced by `sync_committee_signature.validator_index` and maps the sync committee index to an index in the subcommittee (along with the prior `subcommittee_index`).

We may want to note that the parameter `epoch` should be computed to the period boundary epoch.

djrtwo:
I changed the fucntions to be called `get_next_sync_committee` and restricted the usage even more by removing `epoch` as a parameter. This param was error prone and highly context dependent. If we only want to call in one context, we can get `epoch` from state.

Note, that this reduces flexibility when calculating the initial current and next committee at the fork boundary. For simplicity, these are now set to be equal

ralexstokes (approved):
updates look good, i'll go see if we need to update any of the other spec files

dankrad:
lgtm

PR #2409, opened 2021-05-13
Return non-duplicated sync committee subnets

terencechain:
Return non duplicated subnets by applying `set` on the return

Example:
```python
BeaconState {
 Validators: {{0xA}, {0xB}, {0xC}}
 SyncCommittee: {{0xA}, {0xA}, {0xA}, {0xB}, {0xB}, {0xC}}
}
```
Assuming `SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT` is 1,  instead of returning `[0,0,0,1,1,2]`, it would just return `[0,1,2]`

### DIFF
--- a/specs/altair/validator.md
+++ b/specs/altair/validator.md
@@ -292,7 +292,7 @@ The validator broadcasts the assembled signature to the assigned subnet, the `sy
 The `subnet_id` is derived from the position in the sync committee such that the sync committee is divided into "subcommittees".
 `subnet_id` can be computed via `compute_subnets_for_sync_committee(state, validator_index)` where `state` is a `BeaconState` during the matching sync committee period.
 
-*Note*: This function returns multiple subnets if a given validator index is included multiple times in a given sync committee across multiple subcommittees.
+*Note*: This function returns multiple non-duplicated subnets if a given validator index is included multiple times in a given sync committee across multiple subcommittees.
 
 ```python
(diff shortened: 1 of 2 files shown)


specs/altair/validator.md
```diff
@@ -292,7 +292,7 @@ The validator broadcasts the assembled signature to the assigned subnet, the `sy
 The `subnet_id` is derived from the position in the sync committee such that the sync committee is divided into "subcommittees".
 `subnet_id` can be computed via `compute_subnets_for_sync_committee(state, validator_index)` where `state` is a `BeaconState` during the matching sync committee period.
 
-*Note*: This function returns multiple subnets if a given validator index is included multiple times in a given sync committee across multiple subcommittees.
+*Note*: This function returns multiple non-duplicated subnets if a given validator index is included multiple times in a given sync committee across multiple subcommittees.
```
djrtwo:
```suggestion
*Note*: This function returns multiple deduplicated subnets if a given validator index is included multiple times in a given sync committee across multiple subcommittees.
```

djrtwo (approved):
nice catch!

I fixed the tests and made a minor copy edit
thanks :)

dankrad:
The difference between set and list will never be triggered on mainnet and any practical testnets. Is there a testcase that would catch it if a client keeps the list implementation?

PR #2411, opened 2021-05-14
ensure indices are ordered [source, target, head] everywhere

djrtwo:
flag indices calls in code were re-ordered in another PR to reflect [source, target, head] (the order of increasing specificity in the chain) but the indices themselves were still in a different sort order.

Reorder the flag index sort order to match this


protolambda (approved):
LGTM, can't find more cases. The inactivity penalties (using target voting info) being processed separately is fine, and it's source-target-head everywhere else.

PR #2419, opened 2021-05-18
Fix `PARTICIPATION_FLAG_WEIGHTS` to use correct head weight

terencechain:
Replacing `TIMELY_HEAD_FLAG_INDEX` with `TIMELY_HEAD_WEIGHT`. This was discovered by my spec test implementation in alpha.3

PR #2420, opened 2021-05-18
fix a typo: use `TIMELY_HEAD_WEIGHT` instead of `_INDEX`

ericsson49:
Seems to be a typo in [PARTICIPATION_FLAG_WEIGHTS](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/altair/beacon-chain.md#misc): `TIMELY_HEAD_FLAG_INDEX` used, while other elements are of `_WEIGHT` kind


ericsson49:
a duplicate of #2419

PR #2422, opened 2021-05-18
Altair fork: add a link

terencechain:
Two non-substantive changes:
 - Both Altair's validator and network docs refer Phase 0 spec as Phase 0 document, I thought it would be nice to align here
 - Adding a link to the state transition function from the Phase 0 document for ease of navigation


djrtwo (approved):
looks good

Issue #2414, opened 2021-05-14
`ResourceUnavailable` details


djrtwo:
Based on conversations with @arnetheduck, this error code is a bit ambiguous/under-specified for all of the req/resp message types. Specifically, we need to ensure that the MUSTs and SHOULDs of various calls are sufficient and safe given this new code.

This error code was pulled for the v1.1.0-alpha.4 release (#2413). The plan is to discuss the ambiguities here, and work through these details ***BY WEDNESDAY MAY 19*** in preparation for spec freeze on Friday.

The current path is to include the error code but to ensure it is done safely.

arnetheduck:
To avoid wasted implementation effort on how the error should be handled and interpreted by all clients, there are a few details to work out - not only do we save work now ("oh, client x responds this way and client y the other"), but detailing these behaviors in the spec will also make future changes more smooth as having a fixed point / behavior to base backwards compatibility on generally is easier.

In particular, for now we've divided the world into "worked" and "didn't work", letting the client side implement logic to guard against faulty and/or malicious responders (which it must do anyway), with "didn't work" being treated as semi-permanent.

More specific error codes are generally useful when there is an expectation that clients will choose a behavior based on that error code - when adding error codes to the spec, we should thus specify under which conditions each request responds with a particular error code, what the expected behavior on the receiving end is, and how to deal with unexpected error codes.

For `ResourceUnavailable` in particular:

`BeaconBlocksByRoot` answers queries for missing blocks by simply omitting them - clients reading the response to the request must compute a hash of the the block to validate it so they can always deduce `ResourceUnavailable` - introducing the error code means we should pick one way to answer here (by enumerating which error codes are valid in the context)

`BeaconBlocksByRange` - here, `ResourceUnavailable` is a bit more tricky - to add it to the spec, one would have to define the cases where it should be returned. The spec says:

> Clients MUST keep a record of signed blocks seen on the epoch range `[max(GENESIS_EPOCH, current_epoch - MIN_EPOCHS_FOR_BLOCK_REQUESTS), current_epoch]`

There are therefore two cases where `ResourceUnavailable` could be used:

* The client is not following the spec and doesn't make the blocks in the given range available. This can generally be deduced just like with all other block requests: if _another_ client gives a better response, it's known that the client that didn't give blocks doesn't make the resource available - using an error code here basically allows the faulty peer to be kicked earlier because it now admits it's not serving the blocks. This may or may not be a temporary condition (backfilling for example).
[...]

PR #2430, opened 2021-05-21
add resourceunavailable error code with more guidance

djrtwo:
Add `ResourceUnavailable` back in with explicit guidance that it is currently only a valid response for `BlocksByRange` and also that such unsynced nodes `SHOULD` respond using it.

 Conversations on #2414 debate if/how this should be inserted. The crux is that some teams think the extra information will be valuable whereas others think it is only cumbersome without reducing complexity. I think the restricted use in this PR is a reasonable compromise.

Additionally for any client that finds the introduction of this error case as cumbersome/complex, I suggest that the receiving of this error code just gets put into the already existing `if bad_block_range_response()` control flow. If the information is indeed not valuable, it can easily just be lumped into the existing "bad response" code.

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -756,7 +756,7 @@ where `current_epoch` is defined by the current wall-clock time,
 and clients MUST support serving requests of blocks on this range.
 
 Peers that are unable to reply to block requests within the `MIN_EPOCHS_FOR_BLOCK_REQUESTS`
-epoch range SHOULD respond with error code `3: **ResourceUnavailable**`.
+epoch range SHOULD respond with error code `3: ResourceUnavailable`.
 Such peers that are unable to successfully reply to this range of requests MAY get descored
 or disconnected at any time.
 
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -567,6 +567,9 @@ The response code can have one of the following values, encoded as a single unsi
   The response payload adheres to the `ErrorMessage` schema (described below).
 -  2: **ServerError** -- the responder encountered an error while processing the request.
   The response payload adheres to the `ErrorMessage` schema (described below).
+-  3: **ResourceUnavailable** -- the responder does not have requested resource.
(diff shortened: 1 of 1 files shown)


specs/phase0/p2p-interface.md
```diff
@@ -567,6 +567,9 @@ The response code can have one of the following values, encoded as a single unsi
   The response payload adheres to the `ErrorMessage` schema (described below).
 -  2: **ServerError** -- the responder encountered an error while processing the request.
   The response payload adheres to the `ErrorMessage` schema (described below).
+-  3: **ResourceUnavailable** -- the responder does not have requested resource.
+  The response payload adheres to the `ErrorMessage` schema (described below).
+  *Note*: This response code is only valid as a response to `BlocksByRange`.
```
protolambda, line 572: There's a similar method in the sharding ReqResp spec, but that's unstable, and may be removed. Probably better for a later PR.

specs/phase0/p2p-interface.md
```diff
@@ -752,8 +755,10 @@ Clients MUST keep a record of signed blocks seen on the epoch range
 where `current_epoch` is defined by the current wall-clock time,
 and clients MUST support serving requests of blocks on this range.
 
-Peers that are unable to reply to block requests within the
-`MIN_EPOCHS_FOR_BLOCK_REQUESTS` epoch range MAY get descored or disconnected at any time.
+Peers that are unable to reply to block requests within the `MIN_EPOCHS_FOR_BLOCK_REQUESTS`
+epoch range SHOULD respond with error code `3: **ResourceUnavailable**`.
```
protolambda, line 759: The formatting is weird here, bold word in code block

specs/phase0/p2p-interface.md
djrtwo: sounds good

specs/phase0/p2p-interface.md
djrtwo: ah, right

protolambda (approved):
LGTM

PR #2437, opened 2021-05-24
Clean up outdated attestation helper

### LINKED ISSUE #2425: Unused parameter in pyspec test code
https://github.com/ethereum/eth2.0-specs/blob/dev/tests/core/pyspec/eth2spec/test/helpers/attestations.py#L53

The `on_time` parameter is not used. Unsure if it should have any particular meaning, but it does seem to be used like it should, for example
```python
get_valid_on_time_attestation

get_valid_late_attestation
```
 
in the same file.

ralexstokes:
Fixes #2425.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/custody_game/epoch_processing/test_process_custody_final_updates.py
+++ b/tests/core/pyspec/eth2spec/test/custody_game/epoch_processing/test_process_custody_final_updates.py
@@ -8,7 +8,7 @@ from eth2spec.test.helpers.custody import (
     get_sample_shard_transition
 )
 from eth2spec.test.helpers.attestations import (
-    get_valid_on_time_attestation,
+    get_valid_attestation,
 )
 from eth2spec.test.helpers.state import next_epoch_via_block, transition_to, transition_to_valid_shard_slot
 from eth2spec.test.context import (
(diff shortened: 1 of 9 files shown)


tests/core/pyspec/eth2spec/test/custody_game/epoch_processing/test_process_custody_final_updates.py
```diff
@@ -77,8 +77,8 @@ def test_validator_withdrawal_suspend_after_chunk_challenge(spec, state):
     shard = 0
     offset_slots = spec.get_offset_slots(state, shard)
     shard_transition = get_sample_shard_transition(spec, state.slot, [2**15 // 3] * len(offset_slots))
-    attestation = get_valid_on_time_attestation(spec, state, index=shard, signed=True,
-                                                shard_transition=shard_transition)
+    attestation = get_valid_attestation(spec, state, index=shard, signed=True,
+                                        shard_transition=shard_transition)
```
djrtwo, line 81: Don't `ShardTransition`s not exist anymore?

aah, these tests aren't actually being run... cleaning them up iterative is okay but they are still invalid in their current state and will need to be cleaned up much more extensively in the future

djrtwo (approved):
some of the custody tests are still invalid due to the removal of `ShardTransition` a while ago.. probably fine. Fixing them will require a much deeper dive than this

tests/core/pyspec/eth2spec/test/custody_game/epoch_processing/test_process_custody_final_updates.py
ralexstokes: yeah, i'd suggest an iterative approach here, i can keep this in mind as well but seems more sustainable to make small improvements continuously, rather than gate this on a deeper cleanup

PR #2438, opened 2021-05-24
Move BLS extensions to new file in Altair

### LINKED ISSUE #2369: `bls.AggregatePKs` is undefined in Altair spec
The Altair spec currently uses a function `bls.AggregatePKs` to aggregate a batch of BLS pubkeys in the function `get_sync_committee`.

This function is defined internal to the `pyspec` code but is not specified in the IETF signature standard which, according to the spec, is our source of `bls` functionality.

I'd say it is worth considering defining a helper like `eth2_fast_aggregate_verify` for this routine for the completeness of the spec.

edit: implementation incorporating feedback below:

```python
def eth2_aggregate_pubkeys(pubkeys: Sequence[BLSPubkey]) -> BLSPubkey:
    """
    Return the aggregate public key for the public keys in ``pubkeys``.

    NOTE: the `+` operation should be interpreted as elliptic curve point addition, which takes as input
    elliptic curve points that must be decoded from the input `BLSPubkey`s. 
    This implementation is for demonstrative purposes only and ignores encoding/decoding concerns. 
    Refer to the BLS signature draft standard for more information.
    """
    assert len(pubkeys) > 0
    result = copy(pubkeys[0])
    for pubkey in pubkeys[1:]:
        result += pubkey
    return result
```

ralexstokes:
Fixes #2369

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,12 @@ def floorlog2(x: int) -> uint64:
 '''
 
 
+OPTIMIZED_BLS_AGGREGATE_PUBKEYS = '''
+def eth2_aggregate_pubkeys(pubkeys: Sequence[BLSPubkey]) -> BLSPubkey:
+    return bls.AggregatePKs(pubkeys)
+'''
+
+
 class ProtocolDefinition(NamedTuple):
     # just function definitions currently. May expand with configuration vars in future.
     functions: Dict[str, str]
@@ -306,6 +312,10 @@ class SpecBuilder(ABC):
         raise NotImplementedError()
 
     @classmethod
+    def implement_optimizations(cls, functions: Dict[str, str]) -> Dict[str, str]:
+        return functions
+
+    @classmethod
     @abstractmethod
     def build_spec(cls, preset_name: str,
                    source_files: List[Path], preset_files: Sequence[Path], config_file: Path) -> str:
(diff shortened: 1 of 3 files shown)


specs/altair/beacon-chain.md
```diff
@@ -221,6 +223,32 @@ class SyncCommittee(Container):
 
 ## Helper functions
 
+
+### Crypto
+
+#### BLS public keys
+
+An additional function `AggregatePKs` is defined to extend the
```
djrtwo, line 231: Does v4 not expose such a function?

setup.py
```diff
@@ -305,6 +311,10 @@ def invariant_checks(cls) -> str:
         """
         raise NotImplementedError()
 
+    @classmethod
+    def implement_optimizations(cls, functions: Dict[str, str]) -> Dict[str, str]:
+        return functions
```
hwwhww, line 316: Not 100% if I did it in the correct pattern but I intended to make `SpecBuilder` a pure `ABC` class with only `abstractmethod`s. Do you mind (1) moving this method to `Phase0SpecBuilder` and then (2) adding an `abstractmethod` `implement_optimizations` here which raises `raise NotImplementedError()`?

specs/altair/beacon-chain.md
```diff
@@ -26,6 +26,8 @@
     - [`SyncAggregate`](#syncaggregate)
     - [`SyncCommittee`](#synccommittee)
 - [Helper functions](#helper-functions)
+  - [Crypto](#crypto)
+    - [BLS public keys](#bls-public-keys)
   - [`Predicates`](#predicates)
     - [`eth2_fast_aggregate_verify`](#eth2_fast_aggregate_verify)
```
hwwhww, line 32: It was okay to have `eth2_fast_aggregate_verify` in the "Predicates" category, but since this PR adds `eth2_aggregate_pubkeys`, I think it makes sense to put them together. Like:
```suggestion
  - [Crypto](#crypto)
      - [BLS signatures](#bls-signatures)
          - [`eth2_fast_aggregate_verify`](#eth2_fast_aggregate_verify)
          - [`eth2_aggregate_pubkeys`](#eth2_aggregate_pubkeys)
```

hwwhww (commented):
The replacement looks good!
Just some format suggestions.

specs/altair/beacon-chain.md
hwwhww: No, it does not.

setup.py
ralexstokes: sure!

hwwhww (approved):
LGTM 👍

djrtwo (approved):

PR #2441, opened 2021-05-25
[pyspec] Use mainnet.py as the default spec

hwwhww:
After #2390, the pyspec outputs are `mainnet.py` and `minimal.py` files for each protocol upgrade (fork).
- {fork_name}/mainnet.py
- {fork_name}/minimal.py

To make it backward-compatible, this PR simply adds `from . import mainnet as spec` to the `{fork_name}/__init__.py` files so that the pyspec users can still call `from eth2spec.phase0 import spec`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -912,7 +912,8 @@ class PySpecCommand(Command):
 
         if not self.dry_run:
             with open(os.path.join(self.out_dir, '__init__.py'), 'w') as out:
-                out.write("")
+                # `mainnet` is the default spec.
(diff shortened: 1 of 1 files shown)


djrtwo (approved):

djrtwo:
minor lint: "./eth2spec/phase0/__init__.py:1:1: F401 '.mainnet as spec' imported but unused"

protolambda (approved):
Verified that the import works in an eth2spec install via `pip install "git+https://github.com/ethereum/eth2.0-specs@default-spec"`

PR #2443, opened 2021-05-25
Fix bug in Altair transition tests with missing state root

### LINKED ISSUE #2439: Incorrect state root in transition test block
In `minimal/altair/transition/core/pyspec_tests/transition_with_finality/blocks_0.ssz_snappy` the state root is `0x0000000000000000000000000000000000000000000000000000000000000000` so Teku rejects the block as invalid. 

`blocks_1.ssz_snappy` appears to have a non-zero state root so not sure what happened to this first block.  Same problem occurs in a few other transition tests as well for both mainnet and minimal config:

* `transition_with_no_attestations_until_after_fork`
* `transition_with_random_half_participation`
* `transition_with_random_three_quarters_participation`

ralexstokes:
Fixes #2439.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/altair/transition/test_transition.py
+++ b/tests/core/pyspec/eth2spec/test/altair/transition/test_transition.py
@@ -1,7 +1,7 @@
 import random
 from eth2spec.test.context import fork_transition_test
 from eth2spec.test.helpers.constants import PHASE0, ALTAIR
-from eth2spec.test.helpers.state import state_transition_and_sign_block, next_slot, next_epoch_via_block
+from eth2spec.test.helpers.state import state_transition_and_sign_block, next_slot, next_epoch_via_signed_block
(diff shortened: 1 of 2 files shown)


tests/core/pyspec/eth2spec/test/altair/transition/test_transition.py
```diff
@@ -23,6 +23,12 @@ def _state_transition_and_sign_block_at_slot(spec, state):
     return sign_block(spec, state, block)
 
 
+def _next_epoch_via_signed_block(spec, state):
```
djrtwo, line 26: Consider adding a flag `next_epoch_via_block(spec, state, insert_state_root=True)` and then actually putting this signing version of the function (`next_epoch_via_signed_block`)  inside of `helpers/state.py`

djrtwo (approved):

PR #2444, opened 2021-05-25
Add some invariant checks to pyspec unit tests

### LINKED ISSUE #407: add config validity conditions

hwwhww:
### Issue
Fixes #407.

### How did I fix it
Inspired by @protolambda, this PR adds some known invariant checks to pyspec unit tests.

Note: alternatively, we could try to put these conditions on the spec docs. However, the presentation might be kinda messy if we add a new column to each preset/config variable.

### DIFF
--- a/presets/mainnet/altair.yaml
+++ b/presets/mainnet/altair.yaml
@@ -14,8 +14,8 @@ PROPORTIONAL_SLASHING_MULTIPLIER_ALTAIR: 2
 # ---------------------------------------------------------------
 # 2**9 (= 512)
 SYNC_COMMITTEE_SIZE: 512
-# 2**9 (= 512)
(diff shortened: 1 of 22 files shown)


djrtwo (approved):
Great! I added a few more and will continue to do so

PR #2445, opened 2021-05-25
Union type update

### LINKED ISSUE #2270: SSZ Union improvement proposal
The current [Union type](https://github.com/ethereum/eth2.0-specs/blob/dev/ssz/simple-serialize.md#composite-types) is not used in Phase0, and largely not yet implemented in client implementations (exception: lighthouse). In more dynamic contexts, it's could be very useful however. 

The direct motivation for this is the [Merge block format](https://github.com/ethereum/eth2.0-specs/pull/2257), in which we may see different transaction types that need to be encoded. To avoid overhead of lots of unused zero fields (SSZ does not truncate zeroes like RLP does), you would need different transaction type definitions. And how do you mix these different transactions in the same list, with a Union type.

One problem is that it takes 4 bytes per `Union` selector index (the switch between the type options), and we only really see a few type options in the common use case. This is a simple proposal to parametrize (at the type level, compile-time) the selector byte length. E.g. `Union[4, [uint256, Bitvector, uint8]]` would be **backwards compatible** with the already specified `Union[uint256, Bitvector, uint8]`. This would allow us to specify something like `Union[TRANSACTION_SELECTOR_LENGTH, [LegacyTransaction, BetterTransaction, SuperTransaction, AbstractedTransaction]]` (and yes, the selected index is part of the merkleization, no confusion between types in the merkle proof)

For the merge specifically, an alternative is to use opaque transactions (`List[byte, max_size], a.k.a. Bytes[max_size]` to encode an RLP transaction, instead of structured data). The pro is that the beacon-spec does not have to care about transaction types, the con is that we miss out on SSZ merkleization of transactions into the block-root, one of the very primary places that merkle-proofs would actually be very useful (you can construct a merkle proof about any transaction detail).

As a proof of concept, I implemented this new `Union` behavior in `remerkleable`, the SSZ library used in the python specification. See [here](https://github.com/protolambda/remerkleable/pull/9). (*Note for code reader: it represents the data as binary tree, and overlays it with views, it's focused on merkle proofs and data sharing more so than structural efficiency*).

protolambda:
- Clarify missing `Union` type in SSZ spec
- Use "selector" wording instead of "type", since there may be multiple options of the same type in rare cases.
- Reduce selector to 1 byte, and reserve the top 128 values (most significant bit) for future backwards-compatible extensions
- Fix `null` reference in python-ish spec. It's referenced as the native-none/nil/null of the implementation anyway.
- Update serialization and merkleization to cover `None` case in the `Union`

Note: this change is assumed to be backwards-compatible, since `Union` is not used anywhere in the spec. Outside of the spec there is one known case in the DB format of Lighthouse, but the old serialization/deserialization can be kept around there.

TODO:
- [x] See https://github.com/protolambda/remerkleable/pull/9 for remerkleable update. If I get an ACK on this PR, I will make a new remerkleable release, and update the pyspec, so we can start using `Union` in the spec and testing.
  - Update: see remerkleable v0.1.20

fixes #2270

### DIFF
--- a/ssz/simple-serialize.md
+++ b/ssz/simple-serialize.md
@@ -20,7 +20,8 @@
   - [`null`](#null)
   - [`Bitvector[N]`](#bitvectorn)
   - [`Bitlist[N]`](#bitlistn)
-  - [Vectors, containers, lists, unions](#vectors-containers-lists-unions)
+  - [Vectors, containers, lists](#vectors-containers-lists)
+  - [Union](#union)
 - [Deserialization](#deserialization)
 - [Merkleization](#merkleization)
 - [Summaries and expansions](#summaries-and-expansions)
@@ -61,7 +62,7 @@
 * **bitlist**: ordered variable-length collection of `boolean` values, limited to `N` bits
     * notation `Bitlist[N]`
 * **union**: union type containing one of the given subtypes
-    * notation `Union[type_0, type_1, ...]`, e.g. `union[null, uint64]`
+    * notation `Union[type_0, type_1, ...]`, e.g. `union[None, uint64, uint32]`
 
 *Note*: Both `Vector[boolean, N]` and `Bitvector[N]` are valid, yet distinct due to their different serialization requirements. Similarly, both `List[boolean, N]` and `Bitlist[N]` are valid, yet distinct. Generally `Bitvector[N]`/`Bitlist[N]` are preferred because of their serialization efficiencies.
 
@@ -77,7 +78,6 @@ For convenience we alias:
 * `byte` to `uint8` (this is a basic type)
 * `BytesN` and `ByteVector[N]` to `Vector[byte, N]` (this is *not* a basic type)
 * `ByteList[N]` to `List[byte, N]`
-* `null`: `{}`
 
 ### Default values
 Assuming a helper function `default(type)` which returns the default value for `type`, we can recursively define the default value for all types.
@@ -101,7 +101,7 @@ An SSZ object is called zeroed (and thus, `is_zero(object)` returns true) if it 
 
 - Empty vector types (`Vector[type, 0]`, `Bitvector[0]`) are illegal.
 - Containers with no fields are illegal.
-- The `null` type is only legal as the first type in a union subtype (i.e. with type index zero).
+- The `None` type option in a `Union` type is only legal as the first option (i.e. with index zero).
 
 ## Serialization
 
@@ -150,7 +150,7 @@ array[len(value) // 8] |= 1 << (len(value) % 8)
 return bytes(array)
 ```
 
-### Vectors, containers, lists, unions
+### Vectors, containers, lists
 
 ```python
 # Recursively serialize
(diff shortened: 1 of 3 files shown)


protolambda:
Tagging this with the Altair label. It's not critical, nor used in consensus, but I like to ship the SSZ update before the old `Union` type gets more actual uses outside of the spec. And we are looking to use it in the Merge and Sharding specs.

ssz/simple-serialize.md
```diff
@@ -170,14 +170,26 @@ fixed_parts = [part if part != None else variable_offsets[i] for i, part in enum
...
-serialized_bytes = serialize(value.value)
-serialized_type_index = value.type_index.to_bytes(BYTES_PER_LENGTH_OFFSET, "little")
-return serialized_type_index + serialized_bytes
+if value.value is None:
+    assert value.selector == 0
+    return b"\x00"
+else:
+    serialized_bytes = serialize(value.value)
+    serialized_selector_index = value.selector.to_bytes(1, "little")
+    return serialized_selector_index + serialized_bytes
```
Nashatyrev, line 192: (nit) may be define `serialize(None) = Bytes()` and leave just `else` block:
```suggestion
serialized_bytes = serialize(value.value)
serialized_selector_index = value.selector.to_bytes(1, "little")
return serialized_selector_index + serialized_bytes
```

Nashatyrev (approved):
LTGM

ssz/simple-serialize.md
protolambda: When the type is `None`, the selector `0` value still needs to be encoded, hence the `b"\x00"`, I think a little verbosity is fine here.

ssz/simple-serialize.md
```diff
@@ -237,7 +251,8 @@ We now define Merkleization `hash_tree_root(value)` of an object `value` recursi
 * `mix_in_length(merkleize(pack_bits(value), limit=chunk_count(type)), len(value))` if `value` is a bitlist.
 * `merkleize([hash_tree_root(element) for element in value])` if `value` is a vector of composite objects or a container.
 * `mix_in_length(merkleize([hash_tree_root(element) for element in value], limit=chunk_count(type)), len(value))` if `value` is a list of composite objects.
-* `mix_in_type(merkleize(value.value), value.type_index)` if `value` is of union type.
```
djrtwo, line 240: Was `merkleize` incorrect here? and actually should hve been `htr` all along?

protolambda: Yes, that previous `merkleize` is incorrect. The `value` is the union, and `value.value` is the content of the union, which is just another SSZ object, which can be merkleized in any type-specific way. E.g. if `value.value` was a `List`, `merkleize` is wrong because it would miss the length mix-in.

ssz/simple-serialize.md
```diff
@@ -227,7 +241,7 @@ We first define helper functions:
         - If `1` chunk: the root is the chunk itself.
         - If `> 1` chunks: merkleize as binary tree.
 * `mix_in_length`: Given a Merkle root `root` and a length `length` (`"uint256"` little-endian serialization) return `hash(root + length)`.
-* `mix_in_type`: Given a Merkle root `root` and a type_index `type_index` (`"uint256"` little-endian serialization) return `hash(root + type_index)`.
+* `mix_in_selector`: Given a Merkle root `root` and a type selector `selector` (`"uint256"` little-endian serialization) return `hash(root + selector)`.
```
ralexstokes, line 244:
```suggestion
* `mix_in_selector`: Given a Merkle root `root` and a type selector `selector` (`"uint8"` little-endian serialization) return `hash(root + selector)`.
```

do you mean to refer to the root or the selector in the parenthetical? a bit confused what it is referring to

ralexstokes (approved):
looks great! now we can define `Option` types in the spec and i'm sure there will be other uses along the way

left one comment below for clarification

ssz/simple-serialize.md
protolambda: `hash_tree_root(uint8(selector))` is the same as `uint256(selector)`. We convert the selector to 32 bytes before hashing it together with the contents `root`, so it's consistent with the `(32, 32) -> 64` byte merkleization everywhere.

PR #2447, opened 2021-05-25
Fix `_get_sync_committee_signature` testing helper

hwwhww:
### Issue
The parameter `target_slot` in `_get_sync_committee_signature` was unused. It should have been `domain = spec.get_domain(state, spec.DOMAIN_SYNC_COMMITTEE, epoch=spec.compute_epoch_at_slot(target_slot))`

### Proposed change

Refactor it by using `compute_sync_committee_signature` testing helper in `_get_sync_committee_signature`.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/altair/unittests/validator/test_validator.py
+++ b/tests/core/pyspec/eth2spec/test/altair/unittests/validator/test_validator.py
@@ -4,7 +4,7 @@ from eth2spec.utils.ssz.ssz_typing import Bitvector
 from eth2spec.test.helpers.block import build_empty_block
 from eth2spec.test.helpers.keys import pubkey_to_privkey
 from eth2spec.test.helpers.state import transition_to
-from eth2spec.utils import bls
+from eth2spec.test.helpers.sync_committee import compute_sync_committee_signature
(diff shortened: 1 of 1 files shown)


tests/core/pyspec/eth2spec/test/altair/unittests/validator/test_validator.py
```diff
@@ -85,12 +85,9 @@ def _get_sync_committee_signature(
...
     privkey = pubkey_to_privkey[pubkey]
 
-    domain = spec.get_domain(
-        state,
-        spec.DOMAIN_SYNC_COMMITTEE,
+    return compute_sync_committee_signature(
+        spec, state, target_slot, privkey, block_root=target_block_root
     )
-    signing_data = spec.compute_signing_root(target_block_root, domain)
-    return bls.Sign(privkey, spec.hash_tree_root(signing_data))
```
ralexstokes, line 93: looks like there was a bug here anyway! should not be taking a second `hash_tree_root` and this likely worked only because the root of a root is just itself...

ralexstokes (approved):
lgtm! always good to consolidate code and it looks like you found a bit of a bug anyway w/ the signing root calculation

Issue #2448, opened 2021-05-26
Sync committee rewards occur with very low frequency. 


potuz:
In the current Altair Specs, only `SYNC_COMMITTEE_SIZE (512)` validators out of the current 150K on mainnet will earn rewards for over 2 days. Validators that aren't chosen to participate on the sync committee will forfeit 1/8 of their earnings until they are finally chosen. The probabilities of not being in a single sync committee for a single validator are 

```
1 month             |          95.65% 
6 months            |          76.33%
1 year              |          57.86%
2 years             |          33.37%
```

Which means that after two years of staking, we will still see at least 50K validators without earning 1/8 of their rewards. 

A proposal to remedy this is to simply get rid of the committee sync rewards, and apply a penalty to those validators that fail to participate in the committee. The total rewards can be guaranteed to be still one `base_rewards` by adjusting the weights as 

```

TIMELY_SOURCE_WEIGHT	uint64(14)
TIMELY_TARGET_WEIGHT	uint64(28)
TIMELY_HEAD_WEIGHT	uint64(14)
PROPOSER_WEIGHT	        uint64(8)
WEIGHT_DENOMINATOR	uint64(64)
```
And adding a penalty weight 

```
SYNC_PENALTY_WEIGHT	uint64(8)
```

 The changes to the current specs seem to be minimal, a WIP can be found in 

https://github.com/ethereum/eth2.0-specs/compare/dev...potuz:penalize-sync-comiittee?expand=1

Some dissadvantages of the minimal proposal above are

- Validators on the first sync committe may be penalized unfairly for the first few slots
- Validators that are offline and are assigned to a sync committee will be severely penalized

The first point can be fixed in-protocol by not penalizing on the first few epochs. The second one, under current conditions of participation rate above 99%, the probability than an offline validator is chosen for a sync committee is `3 * 10^-5` so it is negligible. 

Advantages are 

- Single stakers will be able to maintain their full rewards without the need of luck. Without this it would be beneficial for them to move to staking pools
- The penalty can be simply reduced by adjusting two parameters without modifing the invariant of one `base_reward`  earnings per epoch.

djrtwo:
addressed in #2453

PR #2450, opened 2021-05-27
Penalize sync comiittee

potuz:
This PR solves #2448  by inverting the logic for sync committee participation rewards into penalizing lack of participation. This way the vast majority of validators are not hindered nor need to resort to luck. 

The one `base_reward` per epoch invariant is preserved by this PR, the only change is that the rewards weights should sum up to the `WEIGHT_DENOMINATOR`, but not including the `SYNC_PENALTY_WEIGHT` which is now otherwise an independent constant. 

Notice that there may be a false feeling that validators in the sync committee are now penalized since they may receive stiff penalties for missing participation in a slot, that otherwise wouldn't. However, with this approach, even after being penalized for missing a sync committee participation, the overall result for those validators is exactly the same as in the current rewards-based mechanism. 

There are other options to mitigate the problem of #2448 like reducing the `SYNC_REWARD_WEIGHT` and `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` but these all lead to relying on luck to some factor. Given that the sync committee size is minimum compared to the active validator set, any set of constants will always result in a non-trivial probability of a high number of validators not being included in a committee for two years. 

Discussions in 
https://discord.com/channels/595666850260713488/595701319793377299/847063172174577744

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -479,7 +479,7 @@ def get_generalized_index(ssz_class: Any, *path: Sequence[Union[int, SSZVariable
     def invariant_checks(cls) -> str:
         return '''
 assert (
-    TIMELY_HEAD_WEIGHT + TIMELY_SOURCE_WEIGHT + TIMELY_TARGET_WEIGHT + SYNC_REWARD_WEIGHT + PROPOSER_WEIGHT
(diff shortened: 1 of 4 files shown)


djrtwo:
closing in favor of #2453

PR #2451, opened 2021-05-27
Add clock disparity tolerance for sync subnets

paulhauner:
I noticed that the sync subnets don't allow for `MAXIMUM_GOSSIP_CLOCK_DISPARITY` and I'm not sure if that was intentional.


djrtwo (approved):
nice catch! whoops...

PR #2452, opened 2021-05-27
[Sharding] Handle the case when committee count < active shards

Nashatyrev:
- Handle the case when a shard may not have a committee at slot in epoch transition 
- Treat the Block as invalid if it contains `ShardBlobHeader` lacking committee
- Reject Gossip `ShardBlobHeader` and `ShardBlob` messages which lacks committee

### DIFF
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -456,15 +456,8 @@ def compute_shard_from_committee_index(state: BeaconState, slot: Slot, index: Co
 
 ```python
 def compute_committee_index_from_shard(state: BeaconState, slot: Slot, shard: Shard) -> CommitteeIndex:
-    """
-    Returns either committee index for ``shard`` at ``slot`` or ``None`` if no committee 
-    """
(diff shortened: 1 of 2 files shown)


specs/sharding/beacon-chain.md
```diff
@@ -693,6 +702,9 @@ def process_pending_headers(state: BeaconState) -> None:
 
             # The entire committee (and its balance)
             index = compute_committee_index_from_shard(state, slot, shard)
+            if index is None:
+                # the shard had no committee on this slot
+                continue
```
Nashatyrev, line 707: Another approach could be iterating through committees instead of shards.

protolambda:
I think we can drop the beacon-chain doc part of the changes, thanks to the improved state format (PR #2455, thanks for the original idea). But the p2p changes still make sense. Can you separate those out?

Nashatyrev:
Rebased this PR on top of #2455: 
- all the state transitions issues were addressed there
- replace Gossip validation condition based on `compute_committee_index_from_shard` behavior from #2455

protolambda (approved):

PR #2455, opened 2021-05-28
New union-based shard headers/commitments representation

protolambda:
Based on discord discussion, special thanks to @Nashatyrev for the suggestion of swapping pending-headers and confirmation data with a `Union`. This PR implements that, along with a special state for unconfirmable headers, and updates of the processing functions.

The state now contains a single field, a table with:
- `256` slots (TBD) lookback. This is a ring-buffer, so merkle proofs are stable throughout the slots (until the data is overwritten, beyond 256 slots).
- a List per slot, to just store the active-shards, instead of a super large vector
- a Union per shard/slot combination, to switch data depending on the shard-header-status
  - The selector can equal 3 status values:
    - `0`, `UNCONFIRMED_SHARD_DATA`: when pending shard headers are not confirmed within time, they get stale (no attestations can increase the vote bitfield anymore). The state-transition cleans this up by switching it to unconfirmed. It's a sane default as well. We also use this to mark shard/slot combinations that can never get a pending shard-header to begin with (when there is no committee assigned to this `(shard, slot)` combi).
    - `1`, `CONFIRMED_SHARD_DATA`: when a `2/3` majority is reached in the committee on any of the pending shard headers, it's confirmed, or the most-voted header at the end of the epoch. We reduce it to just track the commitment. The position within the state implies the `slot` and `shard`.
    - `2`, `PENDING_SHARD_DATA`: when we are processing attestations, we may see multiple different shard headers for the same `(slot, shard)`. These are likely slashable, but the committee can still confirm any of them.

Additionally, this PR:
- renames epoch processing functions to read like `{type of change}_{status of target data}_shard_{name of change}`. Previously the "header" wording was confusing with non-sharding functions that live in the same namespace.
- Fix doc section structure

~~This PR is a *work in progress*~~, using the new state representation, I'm also trying to address related issues:
- ~~Not all active shards may have a committee (also see #2452)~~ Mostly resolved I think. 2452 can address the P2P part.
- ~~The list-length of active shards per slot needs to be initialized correctly~~ Done.
- ~~The "dummy" pending shard-header is weird. Either need to remove it and handle it differently, or change the confirmation process (dummy headers should not be rewarded, and should be marked as unconfirmed)~~ Done.
- Maybe we should change `get_active_shard_count` to refer to the state. It is not ideal to keep the state data-structure in sync with the configuration.

### DIFF
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -13,7 +13,7 @@
 - [Constants](#constants)
   - [Misc](#misc)
   - [Domain types](#domain-types)
-  - [Shard Header Status](#shard-header-status)
+  - [Shard Work Status](#shard-work-status)
 - [Preset](#preset)
   - [Misc](#misc-1)
   - [Shard block samples](#shard-block-samples)
@@ -101,13 +101,13 @@ The following values are (non-configurable) constants used throughout the specif
 | `DOMAIN_SHARD_PROPOSER` | `DomainType('0x80000000')` |
 | `DOMAIN_SHARD_COMMITTEE` | `DomainType('0x81000000')` |
 
-### Shard Header Status
+### Shard Work Status
 
 | Name | Value | Notes |
 | - | - | - |
-| `UNCONFIRMED_SHARD_DATA` | `0` | Unconfirmed, nullified after confirmation time elapses |
-| `CONFIRMED_SHARD_DATA` | `1` | Confirmed, reduced to just the commitment |
-| `PENDING_SHARD_DATA` | `2` | Pending, a list of competing headers |
+| `SHARD_WORK_UNCONFIRMED` | `0` | Unconfirmed, nullified after confirmation time elapses |
+| `SHARD_WORK_CONFIRMED` | `1` | Confirmed, reduced to just the commitment |
+| `SHARD_WORK_PENDING` | `2` | Pending, a list of competing headers |
 
 ## Preset
 
@@ -289,10 +289,10 @@ class ShardProposerSlashing(Container):
 ```python
 class ShardWork(Container):
     #  Upon confirmation the data is reduced to just the header.
-    status: Union[                                                   # See Shard Header Status enum
-              None,                                                  # UNCONFIRMED_SHARD_DATA
-              DataCommitment,                                        # CONFIRMED_SHARD_DATA
-              List[PendingShardHeader, MAX_SHARD_HEADERS_PER_SHARD]  # PENDING_SHARD_DATA
+    status: Union[                                                   # See Shard Work Status enum
+              None,                                                  # SHARD_WORK_UNCONFIRMED
+              DataCommitment,                                        # SHARD_WORK_CONFIRMED
+              List[PendingShardHeader, MAX_SHARD_HEADERS_PER_SHARD]  # SHARD_WORK_PENDING
             ]
 ```
 
@@ -541,7 +541,7 @@ def update_pending_shard_work(state: BeaconState, attestation: Attestation) -> N
     committee_work = state.shard_buffer[buffer_index][attestation_shard]
 
     # Skip attestation vote accounting if the header is not pending
-    if committee_work.status.selector != PENDING_SHARD_DATA:
+    if committee_work.status.selector != SHARD_WORK_PENDING:
         # TODO In Altair: set participation bit flag, if attestation matches winning header.
         return
 
@@ -552,7 +552,6 @@ def update_pending_shard_work(state: BeaconState, attestation: Attestation) -> N
 
     pending_header: PendingShardHeader = current_headers[header_index]
     full_committee = get_beacon_committee(state, attestation.data.slot, attestation.data.index)
-    full_committee_balance = Gwei(0)
 
     # The weight may be outdated if it is not the initial weight, and from a previous epoch
     if pending_header.weight != 0 and compute_epoch_at_slot(pending_header.update_slot) < get_current_epoch(state):
@@ -561,6 +560,7 @@ def update_pending_shard_work(state: BeaconState, attestation: Attestation) -> N
 
     pending_header.update_slot = state.slot
 
+    full_committee_balance = Gwei(0)
     # Update votes bitfield in the state, update weights
     for i, bit in enumerate(attestation.aggregation_bits):
         weight = state.validators[full_committee[i]].effective_balance
@@ -576,12 +576,12 @@ def update_pending_shard_work(state: BeaconState, attestation: Attestation) -> N
         if pending_header.commitment == DataCommitment():
             # The committee voted to not confirm anything
             state.shard_buffer[buffer_index][attestation_shard].change(
-                selector=UNCONFIRMED_SHARD_DATA,
+                selector=SHARD_WORK_UNCONFIRMED,
                 value=None,
             )
         else:
             state.shard_buffer[buffer_index][attestation_shard].change(
-                selector=CONFIRMED_SHARD_DATA,
+                selector=SHARD_WORK_CONFIRMED,
                 value=pending_header.commitment,
             )
 ```
@@ -604,7 +604,7 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
 
     # Check that this data is still pending
     committee_work = state.shard_buffer[header.slot % SHARD_STATE_MEMORY_SLOTS][header.shard]
-    assert committee_work.status.selector == PENDING_SHARD_DATA
+    assert committee_work.status.selector == SHARD_WORK_PENDING
 
     # Check that this header is not yet in the pending list
     current_headers: Sequence[PendingShardHeader] = committee_work.status.value
@@ -718,13 +718,13 @@ def process_pending_shard_confirmations(state: BeaconState) -> None:
         buffer_index = slot % SHARD_STATE_MEMORY_SLOTS
         for shard_index in range(len(state.shard_buffer[buffer_index])):
             committee_work = state.shard_buffer[buffer_index][shard_index]
-            if committee_work.selector == PENDING_SHARD_DATA:
+            if committee_work.selector == SHARD_WORK_PENDING:
                 winning_header = max(committee_work.value, key=lambda header: header.weight)
                 # TODO In Altair: set participation bit flag of voters for winning header
                 if winning_header.commitment == DataCommitment():
-                    committee_work.change(selector=UNCONFIRMED_SHARD_DATA, value=None)
+                    committee_work.change(selector=SHARD_WORK_UNCONFIRMED, value=None)
                 else:
-                    committee_work.change(selector=CONFIRMED_SHARD_DATA, value=winning_header.commitment)
+                    committee_work.change(selector=SHARD_WORK_CONFIRMED, value=winning_header.commitment)
 ```
 
 #### `charge_confirmed_shard_fees`
@@ -743,7 +743,7 @@ def charge_confirmed_shard_fees(state: BeaconState) -> None:
         buffer_index = slot % SHARD_STATE_MEMORY_SLOTS
         for shard_index in range(len(state.shard_buffer[buffer_index])):
             committee_work = state.shard_buffer[buffer_index][shard_index]
-            if committee_work.status.selector == CONFIRMED_SHARD_DATA:
+            if committee_work.status.selector == SHARD_WORK_CONFIRMED:
                 # Charge EIP 1559 fee
                 proposer = get_shard_proposer_index(state, slot, Shard(shard_index))
                 fee = (
@@ -783,7 +783,7 @@ def reset_pending_shard_work(state: BeaconState) -> None:
             # a committee is available, initialize a pending shard-header list
             committee_length = len(get_beacon_committee(state, slot, committee_index))
             state.shard_buffer[buffer_index][shard].change(
-                selector=PENDING_SHARD_DATA,
+                selector=SHARD_WORK_PENDING,
                 value=List[PendingShardHeader, MAX_SHARD_HEADERS_PER_SHARD](
                     PendingShardHeader(
                         commitment=DataCommitment(),
@@ -794,7 +794,7 @@ def reset_pending_shard_work(state: BeaconState) -> None:
                     )
                 )
             )
-        # a shard without committee available defaults to UNCONFIRMED_SHARD_DATA.
+        # a shard without committee available defaults to SHARD_WORK_UNCONFIRMED.
 ```
 
 #### `process_shard_epoch_increment`
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -33,7 +33,7 @@
   - [`ShardBlobReference`](#shardblobreference)
   - [`SignedShardBlobReference`](#signedshardblobreference)
   - [`ShardProposerSlashing`](#shardproposerslashing)
-  - [`ShardCommitteeWork`](#shardcommitteework)
+  - [`ShardWork`](#shardwork)
 - [Helper functions](#helper-functions)
   - [Misc](#misc-2)
     - [`next_power_of_two`](#next_power_of_two)
@@ -185,7 +185,7 @@ class BeaconState(merge.BeaconState):  # [extends The Merge state]
     current_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
     # [New fields]
     # A ring buffer of the latest slots, with information per active shard.
-    shard_buffer: Vector[List[ShardCommitteeWork, MAX_SHARDS], SHARD_STATE_MEMORY_SLOTS]
+    shard_buffer: Vector[List[ShardWork, MAX_SHARDS], SHARD_STATE_MEMORY_SLOTS]
     shard_gasprice: uint64
     current_epoch_start_shard: Shard
 ```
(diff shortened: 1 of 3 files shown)


Nashatyrev (commented):
What do you think on simplifying `ShardWork` to have just 2 states:
`PENDING_SHARD_DATA` and `CONFIRMED_SHARD_DATA` ?

The `CONFIRMED_SHARD_DATA` would be used for both 'empty' and 'non-empty' commitments. 
When needed to distinct between 'empty' and 'non-empty' commitments some simple helper function (e.g. `is_empty_shard_commitment()`) could be used (e.g. in `charge_confirmed_shard_fees()`)

This way previous and current epochs would have either `PENDING_SHARD_DATA` or `CONFIRMED_SHARD_DATA` states, future epoch would be initialized with `PENDING_SHARD_DATA`, all others would have only `CONFIRMED_SHARD_DATA` states. 

A shard with absent committee would always result in 'empty' confirmed commitment on epoch transition

specs/sharding/beacon-chain.md
```diff
@@ -497,67 +520,67 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+
+    # Find the corresponding header, abort if it cannot be found
+    header_index = [header.root for header in current_headers].index(attestation.data.shard_header_root)
+
+    # Update votes bitfield in the state
+    pending_header: PendingShardHeader = state.shard_buffer[buffer_index][attestation_shard][header_index]
     full_committee = get_beacon_committee(state, attestation.data.slot, attestation.data.index)
+    participants_balance = Gwei(0)
+    for i, bit in enumerate(attestation.aggregation_bits):
+        weight = state.validators[full_committee[i]].effective_balance
```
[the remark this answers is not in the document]
protolambda: Inside the `if` is better yes. But it's defined here since it's used twice (counting the total participants weight, and weight of the header. And we could sum `full_committee_balance` at the same time, to avoid the separate balance lookups in the below `get_total_balance`

protolambda:
In terms of beacon-state serialization, especially if we have a long period of headers, I prefer to minimize the unconfirmed headers to 1 byte (the Union selector). And it also makes a merkle proof to the Union selector sufficient to check the confirmation status. The empty `PendingShardHeader()` works to track the special case during beacon-processing, but I think we shouldn't impose that on users who just need the status of the shard data.

specs/sharding/beacon-chain.md
```diff
@@ -497,67 +520,67 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+
+    # Skip attestation vote accounting if the header is already confirmed
+    if committee_work.status.selector == CONFIRMED_SHARD_DATA:
         return
 
-    # Requirement 2: >= 2/3 of balance attesting
-    participants = get_attesting_indices(state, attestation.data, pending_header.votes)
-    participants_balance = get_total_balance(state, participants)
+    # Note that shard-slot combinations without an assigned committee do not have a pending state
+    assert shard_info.status.selector == PENDING_SHARD_DATA
```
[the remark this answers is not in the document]
Nashatyrev: Could it be the case when: 
- attestation-1 has 2/3 (or more) votes for empty commitment for `(slot, shard)`
- the following attestation-2  (could be in the same block or the next one) has a vote[s] for another commitment (or even also for the same empty commitment) for the same `(slot, shard)` 

When processing attestation-1 the `ShardWork` for the `(slot, shard)` would be transferred to `UNCONFIRMED_SHARD_DATA` state 

Normally the attestation-2 shard vote[s] should be ignored (since 2/3 votes already confirmed empty commitment), but the above assert would fail in this case and the attestation would be invalid. 

Am I probably missing anything?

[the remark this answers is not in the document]
protolambda: Yes, that can happen, good catch

protolambda:
Work in progress:
- Effective balance can change between epochs during shard confirmation, and thus the pending-header weight can get inaccurate, due to different inclusion time of attestations
- An effective balance change during epoch transition can affect confirmation status. But confirmation processing happens before effective balance changes, so confirmation may be delayed to an epoch later in an awkward way
- When shard data is explicitly unconfirmed, attestations for that data must still be valid. The assertion of pending-shard-header would break that.

Nashatyrev (approved):
LGTM. 
@protolambda, many thanks for implementing this idea 

I'm still slightly concerned regarding duplicating data in `weight` and `votes` fields and their potential disparity. However `weight` is good for dispersing calculations from epoch processing to blocks processing. I believe we may think on this topic and probably address this issue in later PRs

specs/sharding/beacon-chain.md
```diff
@@ -768,23 +778,23 @@ def reset_pending_shard_work(state: BeaconState) -> None:
...
-                    selector=PENDING_SHARD_DATA,
-                    value=List[PendingShardHeader, MAX_SHARD_HEADERS_PER_SHARD](
-                        PendingShardHeader(
-                            commitment=DataCommitment(),
-                            root=Root(),
-                            votes=Bitlist[MAX_VALIDATORS_PER_COMMITTEE]([0] * committee_length),
-                            weight=0,
-                        )
+        for committee_index in range(committees_per_slot):
+            shard = (start_shard + committee_index) % active_shards
```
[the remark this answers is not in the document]
protolambda: Good idea, but that function recalculates some of the involved variables. It's one line here, so I don't think it's worth refactoring now. We might in a future PR, if shard rotation changes (not anytime soon, but likely in some capacity before production sharding).

specs/sharding/beacon-chain.md
```diff
@@ -448,15 +469,19 @@ def get_start_shard(state: BeaconState, slot: Slot) -> Shard:
...
 #### `compute_committee_index_from_shard`
 
 ```python
 def compute_committee_index_from_shard(state: BeaconState, slot: Slot, shard: Shard) -> CommitteeIndex:
-    active_shards = get_active_shard_count(state, compute_epoch_at_slot(slot))
-    return CommitteeIndex((active_shards + shard - get_start_shard(state, slot)) % active_shards)    
+    epoch = compute_epoch_at_slot(slot)
+    active_shards = get_active_shard_count(state, epoch)
+    index = CommitteeIndex((active_shards + shard - get_start_shard(state, slot)) % active_shards)
+    assert index >= get_committee_count_per_slot(state, epoch)
```
vbuterin, line 483: Are you sure we don't want `<` here instead of `<=`?

specs/sharding/beacon-chain.md
```diff
@@ -497,67 +522,74 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
-        c for c in pending_headers if
-        (c.slot, c.shard) == (pending_header.slot, pending_header.shard)
-    ]
-    if True in [c.confirmed for c in all_candidates]:
+    buffer_index = attestation.data.slot % SHARD_STATE_MEMORY_SLOTS
+    committee_work = state.shard_buffer[buffer_index][attestation_shard]
+
+    # Skip attestation vote accounting if the header is not pending
+    if committee_work.status.selector != PENDING_SHARD_DATA:
+        # TODO In Altair: set participation bit flag, if attestation matches winning header.
```
vbuterin, line 545: Don't you still need to track participation beyond the 2/3 boundary so that you know whom to reward? Or is that what this todo note is trying to say?

specs/sharding/beacon-chain.md
protolambda: Good catch, we should make it executable for testing to catch this earlier, I've been writing too much `if ... { return err }` Go code with the reverse assert expressions

specs/sharding/beacon-chain.md
protolambda: That's what the TODO flag is trying to say. Before completely exiting the attestation processing, we should set those flag bits. But we can't until we rebase the Merge+Sharding onto Altair.

specs/sharding/beacon-chain.md
```diff
@@ -668,119 +701,103 @@ def process_epoch(state: BeaconState) -> None:
...
-        for index in range(committees_per_slot):
-            committee_index = CommitteeIndex(index)
-            shard = compute_shard_from_committee_index(state, slot, committee_index)
+        buffer_index = slot % SHARD_STATE_MEMORY_SLOTS
+        
+        # Reset the shard work tracking
+        state.shard_buffer[buffer_index] = [ShardWork() for _ in range(active_shards)]
+
+        start_shard = get_start_shard(state, slot)
+        for committee_index in range(committees_per_slot):
```
vbuterin, line 781: Might it be a good idea to just reset _all_ the ShardWork, even if those shards are not active? I worry that having some random shards in some cases stay past `SHARD_STATE_MEMORY_SLOTS` won't be very useful because it's unreliable, and it could risk causing bugs from a record corresponding to an older slot being mistaken for a record referring to a newer slot.

protolambda: Do you mean the 778 line? There the list for the given slot is overwritten as a whole, it's not just the `0...len(active_shards)` part that gets overwritten, that's just what the new list is filled with. So it can change shard count without problem.

Regarding the outer layer with `buffer_index` itself; it's a ring buffer, so we have to reset slot by slot. Currently it keeps the latest 256 slots in memory, and resets the last 32 every epoch. Resetting slot-by-slot may be better, but I'm not sure if we want to change the `process_slots` processing.

vbuterin: Aaah I missed the 778 line; I thought only the active shards were being reset. I'm satisfied now.

specs/sharding/beacon-chain.md
```diff
@@ -99,6 +101,14 @@ The following values are (non-configurable) constants used throughout the specif
 | `DOMAIN_SHARD_PROPOSER` | `DomainType('0x80000000')` |
 | `DOMAIN_SHARD_COMMITTEE` | `DomainType('0x81000000')` |
 
+### Shard Header Status
+
+| Name | Value | Notes |
+| - | - | - |
+| `UNCONFIRMED_SHARD_DATA` | `0` | Unconfirmed, nullified after confirmation time elapses |
+| `CONFIRMED_SHARD_DATA` | `1` | Confirmed, reduced to just the commitment |
+| `PENDING_SHARD_DATA` | `2` | Pending, a list of competing headers |
```
Nashatyrev, line 110: Does it makes sense to somehow prefix those constants to group them? 
May be `SHARD_HEADER_STATUS_` or probably some shorter prefix?

specs/sharding/beacon-chain.md
```diff
@@ -448,15 +469,19 @@ def get_start_shard(state: BeaconState, slot: Slot) -> Shard:
 ```python
 def compute_shard_from_committee_index(state: BeaconState, slot: Slot, index: CommitteeIndex) -> Shard:
     active_shards = get_active_shard_count(state, compute_epoch_at_slot(slot))
+    assert index < active_shards
```
Nashatyrev, line 472: Formally `index` and `active_shards` are incompatible types and comparison is not formally correct. 
If the `index` passed is valid in the context of `(state, slot)` then this assertions should never fail.

specs/sharding/beacon-chain.md
```diff
@@ -497,67 +522,74 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+
+    # Find the corresponding header, abort if it cannot be found
+    header_index = [header.root for header in current_headers].index(attestation.data.shard_header_root)
+
+    pending_header: PendingShardHeader = current_headers[header_index]
     full_committee = get_beacon_committee(state, attestation.data.slot, attestation.data.index)
-    full_committee_balance = get_total_balance(state, set(full_committee))
-    if participants_balance * 3 >= full_committee_balance * 2:
-        pending_header.confirmed = True
+    full_committee_balance = Gwei(0)
```
Nashatyrev, line 555: (nit) I would pull this `full_committee_balance ` declaration down: right before the `for` loop to make the code clear

specs/sharding/beacon-chain.md
protolambda: Both are derived from a `uint64`, I think the comparison is clear enough. Attestations (and thus committee index) have been verified before the shard processing, but we may want to use this function elsewhere, so I think we still need the assertion.

specs/sharding/beacon-chain.md
protolambda: Formatting it as `SHARD_WORK_{status here}`, I want to avoid extra long constant names.

specs/sharding/beacon-chain.md
```diff
@@ -569,6 +601,15 @@ def process_shard_header(state: BeaconState,
     # Verify that the block root matches,
     # to ensure the header will only be included in this specific Beacon Chain sub-tree.
     assert header.body_summary.beacon_block_root == get_block_root_at_slot(state, header.slot - 1)
+
+    # Check that this data is still pending
+    committee_work = state.shard_buffer[header.slot % SHARD_STATE_MEMORY_SLOTS][header.shard]
+    assert committee_work.status.selector == PENDING_SHARD_DATA
+
+    # Check that this header is not yet in the pending list
+    current_headers: Sequence[PendingShardHeader] = committee_work.status.value
```
Nashatyrev, line 610: `header_root` definition is missed
```suggestion
    current_headers: Sequence[PendingShardHeader] = committee_work.status.value
    header_root = hash_tree_root(header)
```

djrtwo (approved):
did a quick pass to get this multi-reviewed PR merged. I'll do some deeper review independent of this PR

specs/sharding/beacon-chain.md
```diff
@@ -668,119 +701,103 @@ def process_epoch(state: BeaconState) -> None:
...
-    for header in confirmed_headers:
-        state.grandparent_epoch_confirmed_commitments[header.shard][header.slot % SLOTS_PER_EPOCH] = header.commitment
-```
-
-```python
-def charge_confirmed_header_fees(state: BeaconState) -> None:
+        buffer_index = slot % SHARD_STATE_MEMORY_SLOTS
+        for shard_index in range(len(state.shard_buffer[buffer_index])):
+            committee_work = state.shard_buffer[buffer_index][shard_index]
+            if committee_work.selector == PENDING_SHARD_DATA:
```
Nashatyrev:
```suggestion
            if committee_work.status.selector == PENDING_SHARD_DATA:
```

Nashatyrev: Also occurrences below

specs/sharding/beacon-chain.md
```diff
@@ -668,119 +701,103 @@ def process_epoch(state: BeaconState) -> None:
...
-                new_gasprice,
-                candidate.commitment.length,
-                adjustment_quotient,
-            )
+        buffer_index = slot % SHARD_STATE_MEMORY_SLOTS
+        for shard_index in range(len(state.shard_buffer[buffer_index])):
+            committee_work = state.shard_buffer[buffer_index][shard_index]
+            if committee_work.status.selector == CONFIRMED_SHARD_DATA:
+                # Charge EIP 1559 fee
+                proposer = get_shard_proposer_index(state, slot, Shard(shard_index))
```
Nashatyrev, line 749: Define 
```py
commitment = committee_work.status.value
```
and use it below in place of `candidate.commitment`

specs/sharding/beacon-chain.md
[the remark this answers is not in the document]
Nashatyrev: I'm ok with either code, though from my perspective such micro-optimizations slightly blurs the focus in spec. E.g. in this specific case a spec reader would need to check with `compute_shard_from_committee_index` function implementation to make sure we are doing exactly this operation and not any other magic

specs/sharding/beacon-chain.md
```diff
@@ -497,67 +522,74 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
-        (c.slot, c.shard) == (pending_header.slot, pending_header.shard)
-    ]
-    if True in [c.confirmed for c in all_candidates]:
+    buffer_index = attestation.data.slot % SHARD_STATE_MEMORY_SLOTS
+    committee_work = state.shard_buffer[buffer_index][attestation_shard]
+
+    # Skip attestation vote accounting if the header is not pending
+    if committee_work.status.selector != SHARD_WORK_PENDING:
+        # TODO In Altair: set participation bit flag, if attestation matches winning header.
         return
```
Nashatyrev, line 546: Hm, it seems we have an interesting feature (or bug) here: the `attestation.data.shard_header_root` is not checked in this case. 
But assuming the fact that attestations are normally produced for `ShardHeader`s which are not yet included to block/state some attestations could still be valid despite they are voting for a header which never included. Thus such attestations could be queued and finally included to block after some header is confirmed for the corresponding `(slot, shard)`

specs/sharding/beacon-chain.md
```diff
@@ -497,67 +522,74 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
-        c for c in pending_headers if
-        (c.slot, c.shard) == (pending_header.slot, pending_header.shard)
-    ]
-    if True in [c.confirmed for c in all_candidates]:
+    buffer_index = attestation.data.slot % SHARD_STATE_MEMORY_SLOTS
+    committee_work = state.shard_buffer[buffer_index][attestation_shard]
+
+    # Skip attestation vote accounting if the header is not pending
+    if committee_work.status.selector != SHARD_WORK_PENDING:
+        # TODO In Altair: set participation bit flag, if attestation matches winning header.
```
Nashatyrev, line 545: For the current version I believe we could not match `attestation.data.shard_header_root` since only `DataCommitment` remains (in case of `SHARD_WORK_CONFIRMED`)

PR #2456, opened 2021-05-28
Clean up SSZ `null`

ralexstokes:
I believe `null` was superseded with #2445.

This PR just removes the definition that we no longer need.

edit: also my editor found some whitespace


protolambda (approved):
LGTM

PR #2458, opened 2021-05-29
Clarify language on ssz Unions

ralexstokes:
I think the line I changed only applies to Unions, so I made that explicit.


protolambda (approved):
Good catch, LGTM

PR #2461, opened 2021-05-31
Align Sync Committee Tests to Specs

### LINKED ISSUE #2460: test_process_sync_committee.py does not have the right logic
The current tests for Sync Committee Participation rewards do not reflect the current logic in `specs/altair/beacon-chain.md`. In particular, rewards in the former are dependent of the participant index via its effective balance, while the rewards in the latter are independent of this.

potuz:
Fixes #2460 

Notice that these changes would be obsolete if either #2450 or #2453 are included.


djrtwo:
The relevant components made it into #2453
thanks!

PR #2462, opened 2021-06-01
Merge transition process with computed transition total difficulty

mkalinin:
Introduces the Merge fork  (`upgrade_to_merge`) with the version of transition process that uses dynamically computed transition total difficulty.

The computation happens during the Merge fork, i.e. next to `upgrade_to_merge` call. Total difficulty is computed with the following formula `transition_total_difficulty = get_pow_block(state.eth1_data.block_hash).total_difficulty + TRANSITION_TOTAL_DIFFICULTY_OFFSET`. And then it is stored to a specific storage slot in `TransitionStore`.

### Rationale

Dynamically computed total difficulty makes the transition process more predictable.

For example, we expect `MERGE_FORK_EPOCH` two weeks after updated client software has been deployed and we also take a couple of extra days after the Merge fork to be able to react accordingly if something went wrong during the fork itself. 

OTOH, we can compute transition total difficulty at the time of the fork and remove period before the fork (2 weeks in our case) from our model. If we double total difficulty estimation to protect from difficulty fluctuations then we can reduce the time between software deployment and the actual transition by at least two weeks in our case. Which, in its turn, makes estimated time of the transition more accurate.

#### What if `state.eth1_data` is too old?
If `TRANSITION_TOTAL_DIFFICULTY_OFFSET` is set to 1 week worth of difficulty then it will be enough time to handle this case properly. It's hard to imagine that `state.eth1_data` is older than 3 days (more than 10 voting periods failed in a row).

#### What if `state.eth1_data` is forked out?
We may set `MERGE_FORK_EPOCH` to the beginning of the voting period to reduce the possibility of changing eth1 data in state due to chain re-orgs. Also, `TransitionStore` should be initialized each time the fork happens in the client's code and seeded with corresponding `state.eth1_data`.

#### Why `TransitionStore` instead of `Store.transition_total_difficulty`?
To keep implementation and testing simple and weaken dependency of the fork choice tests and transition process. We can opt for re-using `Store` to keep `transition_total_difficulty` if we want.

#### Testing
Merge fork tests are copied from Altair ones. There is no tests for the transition process itself, we better work on them in a separate PR once we settled down with particular approach to the transition.

#### Transition total difficulty announcement
We might want to communicate transition total difficulty offchain once it's computed. So, validators will be able to check it and act accordingly if computation went wrong on their end. This would require a corresponding change in beacon node API.


specs/merge/fork-choice.md
```diff
@@ -66,44 +67,59 @@ def finalize_block(self: ExecutionEngine, block_hash: Hash32) -> bool:
...
 
 Let `get_pow_block(block_hash: Hash32) -> PowBlock` be the function that given the hash of the PoW block returns its data.
 
 *Note*: The `eth_getBlockByHash` JSON-RPC method does not distinguish invalid blocks from blocks that haven't been processed yet. Either extending this existing method or implementing a new one is required.
 
-#### `is_valid_transition_block`
+### `get_transition_store`
+
+```python
+def get_transition_store(anchor_pow_block: PowBlock):
```
hwwhww, line 100:
```suggestion
def get_transition_store(anchor_pow_block: PowBlock) -> TransitionStore:
```

specs/merge/fork-choice.md
```diff
@@ -119,11 +135,12 @@ def on_block(store: Store, signed_block: SignedBeaconBlock) -> None:
...
     
     # [New in Merge]
-    if is_transition_block(pre_state, block):
+    is_transition_store_initialized = transition_store is not None
+    if is_transition_store_initialized and is_transition_block(pre_state, block):
         # Delay consideration of block until PoW block is processed by the PoW node
         pow_block = get_pow_block(block.body.execution_payload.parent_hash)
         assert pow_block.is_processed
-        assert is_valid_transition_block(pow_block)
+        assert is_valid_transition_block(transition_store, pow_block)
```
hwwhww, line 143: Our type checker (`mypy`) is kinda silly here...
If you change it to:
```python
    if transition_store is not None and is_transition_block(pre_state, block):
        # Delay consideration of block until PoW block is processed by the PoW node
        pow_block = get_pow_block(block.body.execution_payload.parent_hash)
        assert pow_block.is_processed
        assert is_valid_transition_block(transition_store, pow_block)
```
then the `mypy` will pass. 😅

specs/merge/fork-choice.md
```diff
@@ -66,44 +67,59 @@ def finalize_block(self: ExecutionEngine, block_hash: Hash32) -> bool:
...
 Let `get_pow_block(block_hash: Hash32) -> PowBlock` be the function that given the hash of the PoW block returns its data.
 
 *Note*: The `eth_getBlockByHash` JSON-RPC method does not distinguish invalid blocks from blocks that haven't been processed yet. Either extending this existing method or implementing a new one is required.
 
-#### `is_valid_transition_block`
+### `get_transition_store`
+
+```python
+def get_transition_store(anchor_pow_block: PowBlock):
+    transition_total_difficulty = anchor_pow_block.total_difficulty + TRANSITION_TOTAL_DIFFICULTY_OFFSET
```
vbuterin, line 101: Shouldn't we just make this:

```suggestion
    transition_total_difficulty = anchor_pow_block.total_difficulty + anchor_pow_block.difficulty * (86400 * 7 // 13)
```

This way we have to deal with less time uncertainty because the difficulty would be set at the anchor time instead of weeks before it.

specs/merge/fork-choice.md
```diff
@@ -119,11 +135,12 @@ def on_block(store: Store, signed_block: SignedBeaconBlock) -> None:
     assert get_ancestor(store, block.parent_root, finalized_slot) == store.finalized_checkpoint.root
     
     # [New in Merge]
-    if is_transition_block(pre_state, block):
+    is_transition_store_initialized = transition_store is not None
```
vbuterin, line 138:
```suggestion
    is_transition_store_initialized = (transition_store is not None)
```

Slightly easier to read this way.

specs/merge/fork.md
```diff
@@ -0,0 +1,102 @@
...
+### Initializing transition store
+
+If `state.slot % SLOTS_PER_EPOCH == 0` and `compute_epoch_at_slot(state.slot) == MERGE_FORK_EPOCH`, a transition store is initialized to be further utilized by the transition process of the Merge.
+
+Transition store initialization occurs after the state has been modified by corresponding `upgrade_to_merge` function.
+
+```python
+def initialize_transition_store(state: BeaconState) -> TransitionStore:
+    pow_block = get_pow_block(state.eth1_data.block_hash)
+    return get_transition_store(pow_block)
```
vbuterin, line 101: If `get_transition_store` is only ever used in fork.md, wouldn't it be cleaner to move the definition to fork.md?

specs/merge/fork-choice.md
```diff
@@ -119,11 +135,12 @@ def on_block(store: Store, signed_block: SignedBeaconBlock) -> None:
...
+    if is_transition_store_initialized and is_transition_block(pre_state, block):
```
vbuterin, line 139: Is `is_transition_block` defined anywhere?

specs/merge/fork-choice.md
```diff
@@ -66,44 +67,59 @@ def finalize_block(self: ExecutionEngine, block_hash: Hash32) -> bool:
...
+```
+
+### `is_valid_transition_block`
 
 Used by fork-choice handler, `on_block`.
 
 ```python
-def is_valid_transition_block(block: PowBlock) -> bool:
-    is_total_difficulty_reached = block.total_difficulty >= TRANSITION_TOTAL_DIFFICULTY
+def is_valid_transition_block(transition_store: TransitionStore, block: PowBlock) -> bool:
```
vbuterin, line 110: Maybe rename to `is_valid_terminal_pow_block` or something like that? The `is_valid_transition_block` name may mislead a bit as "the transition block" also refers to the first embedded block.

mkalinin: Good suggestion 👍

specs/merge/fork-choice.md
mkalinin: In normal case there are difficulty fluctuations caused by stochastic nature of PoW and hashpower ins/outs and if we take only one block to estimate this large period the estimation might become not that accurate. In our case this block is at least `2048` block behind current head and doesn't give us representation on what's happening in the network right at the moment of computation; theoretically, real hashrate could be doubled during the last `2048` blocks. If anchor block was produced in a period of 51% attack then we can get its difficulty halved wrt the normal case difficulty (another half of hashpower used by attacker or honest miners).

All the inputs above make me think that we better take a constant that to our observation will worth 1 week of hashpower contributed to the network in a normal case. We can take a large period of time to estimate it and get rid of poisson and hashpower fluctuations. But if there is a drop in hashrate right before the Merge estimated offset will become inaccurate which will increase time to transition.

Both approaches can give an unpredictable result depending on the network conditions. But the latter one is more secure and conservative from my point of view because real moment of transition can happen not earlier than the half of expected period (in case if hashpower magically doubles). And this is why I am leaning towards it.

We also need to take the last voting period in account to compute the offset, 
i.e. `anchor_pow_block.difficulty * (2048 + 2048 * 12 // 13 + 86400 * 7 // 13)`.

vbuterin: > In our case this block is at least 2048 block behind current head and doesn't give us representation on what's happening in the network right at the moment of computation

Right, but any difficulty offset determined statically at fork client deployment time (weeks before fork time) would be _even less_ accurate. So it seems better to just make a formula that calculates the offset in real time taking the anchor pow block difficulty into consideration. If we want to prevent miners from attacking by coordinating to turn off their miners for a day at fork time, we could also take the maximum of the difficulty at fork client deployment time and the `anchor_pow_block.difficulty`.

> We also need to take the last voting period in account to compute the offset,
i.e. anchor_pow_block.difficulty * (2048 + 2048 * 12 // 13 + 86400 * 7 // 13).

This makes sense!

specs/merge/fork.md
```diff
@@ -0,0 +1,106 @@
...
+| `MERGE_FORK_VERSION` | `Version('0x02000000')` |
+| `MERGE_FORK_EPOCH` | `Epoch(18446744073709551615)` **TBD** |
+
+## Fork to Merge
+
+### Fork trigger
+
+TBD. Social consensus, along with state conditions such as epoch boundary, finality, deposits, active validator count, etc. may be part of the decision process to trigger the fork. For now we assume the condition will be triggered at epoch `MERGE_FORK_EPOCH`.
+
+Note that for the pure Merge networks, we don't apply `upgrade_to_merge` since it starts with Merge version logic.
```
djrtwo, line 38: For this to be true, we need to define an `initialize_beacon_state` function. In prior forks this function is defined as `initialize_beacon_state_from_eth1` and reads in deposits. When starting from a merged state, this function will likely need a list of initial deposits or validators along with an initial execution-layer state root.

We can deal with this in a later PR but should probably note here that a standard way to do this is currently missing from specs

specs/merge/fork.md
```diff
@@ -0,0 +1,106 @@
...
+If `state.slot % SLOTS_PER_EPOCH == 0` and `compute_epoch_at_slot(state.slot) == MERGE_FORK_EPOCH`, a transition store is initialized to be further utilized by the transition process of the Merge.
+
+Transition store initialization occurs after the state has been modified by corresponding `upgrade_to_merge` function.
+
+```python
+def get_transition_store(anchor_pow_block: PowBlock) -> TransitionStore:
+    transition_total_difficulty = anchor_pow_block.total_difficulty + TRANSITION_TOTAL_DIFFICULTY_OFFSET
+    return TransitionStore(transition_total_difficulty=transition_total_difficulty)
+
+def initialize_transition_store(state: BeaconState) -> TransitionStore:
```
djrtwo, line 103: @hwwhww does the following specificity on the `state` typing make sense here?

```suggestion
def initialize_transition_store(state: merge.BeaconState) -> TransitionStore:
```

tests/core/pyspec/eth2spec/test/helpers/merge/fork.py
```diff
@@ -0,0 +1,45 @@
+MERGE_FORK_TEST_META_TAGS = {
+    'fork': 'merge',
+}
+
+
+def run_fork_test(post_spec, pre_state):
```
djrtwo, line 6: Might be worth abstracting this into a standard `run_fork_test(post_spec, pre_state, stable_fields, modified fields, fork_specific_checks_fn=None)` and have `run_merge_fork_test` call it. But... doesn't matter too much for this PR

tests/core/pyspec/eth2spec/test/merge/fork/test_merge_fork_basic.py
```diff
@@ -0,0 +1,82 @@
...
+    next_epoch,
+    next_epoch_via_block,
+)
+from eth2spec.test.helpers.merge.fork import (
+    MERGE_FORK_TEST_META_TAGS,
+    run_fork_test,
+)
+
+
+@with_phases(phases=[PHASE0], other_phases=[MERGE])
```
djrtwo, line 23: Bummer these are all copied over.. Good for now, but we might consider a more re-usable approach (cc: @ralexstokes, @protolambda )

djrtwo (commented):
Great!

Slight preference for vitaliks calculation based on difficultly because what we choose >= 2 weeks prior is almost certainly less accurate. But just a slight preference

specs/merge/fork-choice.md
mkalinin: This is rather a trade-off between more accurate and more secure. Statically determined offset (or a block difficulty) could be safer than the one requested from the network, it depending on the network conditions of course.

> we could also take the maximum of the difficulty at fork client deployment time and the `anchor_pow_block.difficulty`

What if we do the following: `anchor_difficulty = max(MIN_ANCHOR_DIFFICULTY, anchor_pow_block.difficulty)`, and define the minimal safe difficulty manually cause it, again, should be safer than deferring it to a short moment of network's lifetime?

specs/merge/fork.md
mkalinin: Right, we also need to do this for Altair as well. I'd make this change in a separate PR along with tests refactor and new transition process tests.

specs/merge/fork.md
hwwhww: Not really, because:
1. `merge` object is undefined when we running checker against `./eth2spec/merge/mainnet.py`.
2. `BeaconState` here *is* `merge.BeaconState` logically, so it's fine to keep it as it is.

tests/core/pyspec/eth2spec/test/merge/fork/test_merge_fork_random.py
```diff
@@ -0,0 +1,120 @@
...
+from eth2spec.test.helpers.random import (
+    randomize_state,
+    randomize_attestation_participation,
+)
+
+
+@with_phases(phases=[PHASE0], other_phases=[MERGE])
+@spec_test
+@with_state
+@with_meta_tags(MERGE_FORK_TEST_META_TAGS)
```
protolambda, line 28: We can just define:
```python

def composed(*decs):
    def deco(f):
        for dec in reversed(decs):
            f = dec(f)
        return f
    return deco

merge_test = composed(
  with_phases(phases=[PHASE0], other_phases=[MERGE]),
  spec_test,
  with_state,
  with_meta_tags(MERGE_FORK_TEST_META_TAGS),
)

@merge_test
def test_something_fun(spec, phases, state):
   ...
```

Lots of other places where we can combine decorators as well.

specs/merge/fork.md
djrtwo: Altair *does* have this function [redefined](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/altair/beacon-chain.md#initialize-state-for-pure-altair-testnets-and-test-vectors)

specs/merge/fork.md
```diff
@@ -0,0 +1,121 @@
...
+### Initializing transition store
+
+If `state.slot % SLOTS_PER_EPOCH == 0` and `compute_epoch_at_slot(state.slot) == MERGE_FORK_EPOCH`, a transition store is initialized to be further utilized by the transition process of the Merge.
+
+Transition store initialization occurs after the state has been modified by corresponding `upgrade_to_merge` function.
+
+```python
+def compute_transition_total_difficulty(anchor_pow_block: PowBlock) -> uint256:
+    seconds_per_voting_period = EPOCHS_PER_ETH1_VOTING_PERIOD * SLOTS_PER_EPOCH * SECONDS_PER_SLOT
+    pow_blocks_per_voting_period = seconds_per_voting_period / SECONDS_PER_ETH1_BLOCK
```
vbuterin, line 105:
```suggestion
    pow_blocks_per_voting_period = seconds_per_voting_period // SECONDS_PER_ETH1_BLOCK
```

Also, I don't see `SECONDS_PER_ETH1_BLOCK` defined anywhere. Should be 13, I guess?

specs/merge/fork.md
```diff
@@ -0,0 +1,121 @@
...
+    pow_blocks_since_merge_fork = SECONDS_SINCE_MERGE_FORK / SECONDS_PER_ETH1_BLOCK
```
vbuterin, line 106:
```suggestion
    pow_blocks_since_merge_fork = SECONDS_SINCE_MERGE_FORK // SECONDS_PER_ETH1_BLOCK
```

specs/merge/fork.md
protolambda: It's in the phase0 spec, part of the eth1-data voting, and set to `14`.

djrtwo (approved):
excellent!
plenty of iteration to do but I'm going to get this merged and released for alpha.7

PR #2463, opened 2021-06-01
Remove ExecutionPayloadHeader

JustinDrake:
The complexity around `ExecutionPayloadHeader` feels unnecessary given that the `BeaconState` can easily store `ExecutionPayload.transactions` (the storage overhead is small/tiny). Note that even when the transactions are accompanied by witnesses in the context of statelessness the overhead remains small. More speculatively, it may be valuable down the road for the beacon state to readily have access to the latest transactions.

*Heads up*: I have a bunch of other cleanups and polishing to this merge doc but want to get this fairly substantive change out first.

### DIFF
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -22,7 +22,6 @@
     - [`BeaconState`](#beaconstate)
   - [New containers](#new-containers)
     - [`ExecutionPayload`](#executionpayload)
-    - [`ExecutionPayloadHeader`](#executionpayloadheader)
 - [Protocols](#protocols)
   - [`ExecutionEngine`](#executionengine)
     - [`new_block`](#new_block)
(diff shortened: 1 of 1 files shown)


JustinDrake:
> Do we have an estimation on the max size of transactions depending on the gas limit after statelessness got released? 10MB or 100MB for 25M gas limit?

Non-zero calldata is 16 gas per byte so the worst-case block size pre-statelessness is (25M gas)/(16 gas/byte) = 1.5MB. My rough conservative heuristic is ~10x overhead for naive Merkle witnesses. The fancier algebraic non-Merkle stateless solutions have way less than 10x overhead thanks to batching. Also the way one maxes out the witness size for a given block is by accessing state, not using call data. So all in all I expect the worst case total size would be lower than 10MB for 25M gas. cc @dankrad

djrtwo:
The main problem I see with this, especially in the context of statelessness (larger blocks), is that this state object changes every slot.

Many client implementations store many states (e.g. 1 epoch worth) in memory and rely heavily on diffs/deduplication due to most of the state (the validator set) remaining almost entirely stable diff to diff.

The fact that this block changes every slot then puts much higher diff each state. And in the event that this is 5 to 10MB during stateless, then you have hundreds of more MB to store in memory if you store an epoch or more of state diffs in memory.

dankrad:
In terms of numbers: One 32-byte state access costs ca. 200 bytes using verkle tries, so if we increase costs to roughly 800 gas per access that's around 5 gas/byte or a max of 5 MB at 25M gas.

I find it very strange to put a payload into the state and would be against it. Danny has added good arguments why it might be inefficient.

PR #2467, opened 2021-06-03
Bump `mypy` to v0.812

hwwhww:
- `mypy` v0.800 release adds Python 3.9 support.
- I verified that our current CI setting is able to support Python3.9 env (https://github.com/ethereum/eth2.0-specs/compare/py39-support) with this PR, but let's keep running CI with Python3.8 env for backward compatibility.


djrtwo (approved):
wooo

PR #2472, opened 2021-06-08
polish merge/beacon-chain.md

JustinDrake:
Polish `merge/beacon-chain.md` with mostly non-substantive changes.

**Non-substantive changes**

* rename `MAX_EXECUTION_TRANSACTIONS` to `MAX_TRANSACTIONS_PER_PAYLOAD`
	- rename "execution transaction" to just "transaction" as per discussion with Danny
* rename `compute_time_at_slot` to `compute_timestamp_at_slot`
	- the function returns a Unix timestamp
	- "timestamp" matches `execution_payload.timestamp`
* be explicit about `ExecutionEngine.execution_state` for clarity
* rename `ExecutionPayload.number` to `ExecutionPayload.block_number`
	- more specific ("number" is pretty vague)
	- consistent with `ExecutionPayload.block_hash`
* rename `new_block` to `on_payload`
	- the `on_` prefix is consistent with other event handlers (e.g. see `on_tick`, `on_block`, `on_attestation` [here](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/fork-choice.md#handlers))
	- the `_payload` suffix is more to the point given the function accepts an `execution_payload`
	- avoids conflict with `on_block` which is already used in the fork choice
* rework the table of contents for consistency
* order `is_execution_enabled` after `is_transition_completed` and `is_transition_block`
	- `is_execution_enabled` refers to `is_transition_completed` and `is_transition_block`
* rename "transition" to "merge"
	- "transition" is a bit vague—we will have other transitions at future hard forks
	- there is no need for two words to refer to the same concept
* add a bunch of inline comments, e.g. in `process_execution_payload`
* make the `process_execution_payload` signature consistent with the other `process_` functions in `process_block` which take as arguments `state` and `block.body`
* remove `TRANSITION_TOTAL_DIFFICULTY`
	- to be put in `merge/fork-choice.md` where it is used 
* various misc cleanups

**Substantive changes**

* reorder `ExecutionPayload` fields
	- for consistency with yellow paper and Eth1
	- same for `ExecutionPayloadHeader`
	- added comments separating out the execution block header fields from the extra fields (cosmetic)


djrtwo:
Re opaque transactions -- the intention is to type these with a union type so that multiple transaction types can be supported. "execution" is the more general and proper term imo

specs/merge/beacon-chain.md
```diff
@@ -1,6 +1,6 @@
 # Ethereum 2.0 The Merge
 
-**Warning:** This document is currently based on [Phase 0](../phase0/beacon-chain.md) but will be rebased to [Altair](../altair/beacon-chain.md) once the latter is shipped.
+**Warning**: This document is currently based on [Phase 0](../phase0/beacon-chain.md) and will be rebased for [Altair](../altair/beacon-chain.md).
```
djrtwo, line 3: rebasing is the act of putting something on top of another (merge on top of altair). I would say "rebased to" or "rebased on"

specs/merge/beacon-chain.md
```diff
@@ -59,148 +57,138 @@ We define the following Python custom types for type hinting and readability:
...
-```
+#### `compute_timestamp_at_slot`
 
-#### `is_transition_block`
+*Note*: This function is unsafe with respect to overflows and underflows.
 
 ```python
-def is_transition_block(state: BeaconState, block: BeaconBlock) -> bool:
-    return not is_transition_completed(state) and block.body.execution_payload != ExecutionPayload()
+def compute_timestamp_at_slot(state: BeaconState, slot: Slot) -> uint64:
```
djrtwo, line 166: `compute_time_at_slot` is already defined in phase 0 validator doc. I suggest

1. to not redefine it
2. to not change the name

specs/merge/beacon-chain.md
```diff
@@ -59,148 +57,138 @@ We define the following Python custom types for type hinting and readability:
 | Name | Value |
 | - | - |
 | `MAX_BYTES_PER_OPAQUE_TRANSACTION` | `uint64(2**20)` (= 1,048,576) |
-| `MAX_EXECUTION_TRANSACTIONS` | `uint64(2**14)` (= 16,384) |
+| `MAX_OPAQUE_TRANSACTIONS` | `uint64(2**14)` (= 16,384) |
```
djrtwo, line 60: See my separate comment. These are *transactions*. Not only opaque ones. It's just that opaque is the only type currently.

Maybe just call them `Transactions` -- `MAX_TRANSACTIONS`. We reserved "operation" for validator/system level block messages to not collide with "transaction" as user layer

JustinDrake:
> "execution" is the more general and proper term imo

Just to make sure I understand are you suggesting renaming every instance of "opaque transaction" to "execution transaction"? That would work for me :)

specs/merge/beacon-chain.md
```diff
@@ -59,148 +57,138 @@ We define the following Python custom types for type hinting and readability:
...
 
 ```python
-def compute_time_at_slot(state: BeaconState, slot: Slot) -> uint64:
-    slots_since_genesis = slot - GENESIS_SLOT
-    return uint64(state.genesis_time + slots_since_genesis * SECONDS_PER_SLOT)
+def on_payload(self: ExecutionEngine, execution_payload: ExecutionPayload) -> bool:
+    """
+    Returns ``True`` iff ``execution_payload`` is valid with respect to ``self.execution_state``.
+    """
+    # [...]
```
djrtwo, line 185: `...` is legal python and is the expected empty signifier for this construction that can be overridden in other places

```suggestion
    ...
```

specs/merge/beacon-chain.md
```diff
@@ -209,41 +197,38 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
-        assert execution_payload.parent_hash == state.latest_execution_payload_header.block_hash
-        assert execution_payload.number == state.latest_execution_payload_header.number + 1
-
-    assert execution_payload.timestamp == compute_time_at_slot(state, state.slot)
-
-    assert execution_engine.new_block(execution_payload)
-
+def process_execution_payload(state: BeaconState, body: BeaconBlockBody) -> None:
+    # Skip execution payload processing if execution is not enabled
+    if not is_execution_enabled(state, body):
```
djrtwo, line 210: We previously pulled this block condition *out* because it isn't really part of "processing the execution payload" and majorly complicates testing

specs/merge/beacon-chain.md
```diff
@@ -209,41 +197,38 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
-    """
-    if is_transition_completed(state):
-        assert execution_payload.parent_hash == state.latest_execution_payload_header.block_hash
-        assert execution_payload.number == state.latest_execution_payload_header.number + 1
-
-    assert execution_payload.timestamp == compute_time_at_slot(state, state.slot)
-
-    assert execution_engine.new_block(execution_payload)
-
+def process_execution_payload(state: BeaconState, body: BeaconBlockBody) -> None:
```
djrtwo, line 208: did you globally define the execution engine instead of passing it in?

I would push for the other way around, again for testing and clarity of what this function does and requires

specs/merge/beacon-chain.md
JustinDrake: I've added a new `Transaction` type placeholder to be defined using a union type. Renamed to `MAX_TRANSACTIONS_PER_PAYLOAD`.

specs/merge/beacon-chain.md
JustinDrake: > did you globally define the execution engine instead of passing it in?

Right, the same as with the `bls` module. This is so for consistency with the other `process_` functions inside `process_block`. I don't feel too strongly.

Screenshot 2021-06-08 at 15 53 46

specs/merge/beacon-chain.md
JustinDrake: > it isn't really part of "processing the execution payload"

The `if` statement is part of the other `process_` functions, e.g.:

Screenshot 2021-06-08 at 15 56 31

Screenshot 2021-06-08 at 15 57 15

Screenshot 2021-06-08 at 15 59 06

Don't feel too strongly either way.

protolambda:
> rename EXECUTION_ENGINE to execution_engine
>    lowercase for consistency with the bls module which is similarly exposed

It's not a static module, it's a dynamic part of the environment, something that is plugged in like configuration. E.g. tests, testnets, mainnet, different eth1-redundancy setups, etc. may all use completely different execution engines through this same protocol. The upper-snake-case to match the config style was intentional.

specs/merge/beacon-chain.md
protolambda: Passing in the engine to where it is used makes testing a lot easier, every global breaks encapsulation and increases the difficulty of reusing the code.

JustinDrake: Reverted back to passing the execution engine explicitly :) Thanks for the feedback 👍

specs/merge/beacon-chain.md
```diff
@@ -59,148 +58,138 @@ We define the following Python custom types for type hinting and readability:
...
-    return is_transition_completed(state) or is_transition_block(state, block)
+def is_execution_enabled(state: BeaconState, body: BeaconBlockBody) -> bool:
+    return is_merge_block(state, body) or is_merge_complete(state)
 ```
 
-#### `is_transition_completed`
+### Misc
 
-```python
-def is_transition_completed(state: BeaconState) -> bool:
```
hwwhww, line 183: This function is used in [fork-choice.md](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/merge/fork-choice.md#on_block).

specs/merge/beacon-chain.md
```diff
@@ -59,148 +58,138 @@ We define the following Python custom types for type hinting and readability:
...
-    return state.latest_execution_payload_header != ExecutionPayloadHeader()
-```
+#### `compute_timestamp_at_slot`
 
-#### `is_transition_block`
+*Note*: This function is unsafe with respect to overflows and underflows.
 
 ```python
-def is_transition_block(state: BeaconState, block: BeaconBlock) -> bool:
```
hwwhww, line 190: Ditto. This function is used in [fork-choice.md](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/merge/fork-choice.md#on_block).

specs/merge/beacon-chain.md
```diff
@@ -59,148 +58,138 @@ We define the following Python custom types for type hinting and readability:
...
+### Merge
 
-### Transition
+*Note*: The configuration value `MERGE_FORK_EPOCH` is not final.
 
 | Name | Value |
 | - | - |
 | `MERGE_FORK_VERSION` | `Version('0x02000000')` |
-| `MERGE_FORK_EPOCH` | `Epoch(18446744073709551615)` **TBD** |
-| `TRANSITION_TOTAL_DIFFICULTY` | **TBD** |
```
hwwhww, line 75: This constant is used in [fork-choice.md](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/merge/fork-choice.md#transitionstore)

specs/merge/beacon-chain.md
```diff
@@ -15,42 +15,41 @@
...
 
 ## Custom types
 
-We define the following Python custom types for type hinting and readability:
-
 | Name | SSZ equivalent | Description |
 | - | - | - |
-| `OpaqueTransaction` | `ByteList[MAX_BYTES_PER_OPAQUE_TRANSACTION]` | a byte-list containing a single [typed transaction envelope](https://eips.ethereum.org/EIPS/eip-2718#opaque-byte-array-rather-than-an-rlp-array) structured as `TransactionType \|\| TransactionPayload` |
+| `OpaqueTransaction` | `ByteList[MAX_BYTES_PER_OPAQUE_TRANSACTION]` | a [typed transaction envelope](https://eips.ethereum.org/EIPS/eip-2718#opaque-byte-array-rather-than-an-rlp-array) structured as `TransactionType \|\| TransactionPayload` |
+| `Transaction` | **TODO**: define as `Union` type with `OpaqueTransaction` | an execution transaction |
```
hwwhww, line 52: To define a stub, we should set `Transaction` to `OpaqueTransaction` or `Union[OpaqueTransaction]` for now. I can make pyspec builder obey it. Just ping me once it's decided. :)

specs/merge/beacon-chain.md
```diff
@@ -209,41 +198,36 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+        state_root=body.execution_payload.state_root,
+        logs_bloom=body.execution_payload.logs_bloom,
+        receipt_root=body.execution_payload.receipt_root,
+        block_number=body.execution_payload.block_number,
+        gas_limit=body.execution_payload.gas_limit,
+        gas_used=body.execution_payload.gas_used,
+        timestamp=body.execution_payload.timestamp,
+        block_hash=body.execution_payload.block_hash,
+        transactions_root=hash_tree_root(body.execution_payload.transactions),
     )
```
hwwhww, line 232: IMHO it would be less verbose to set `execution_payload = body.execution_payload` first, and then use `execution_payload` instead of `body.execution_payload`.

specs/merge/beacon-chain.md
JustinDrake: Understood!

As a heads up this function has not been removed, merely renamed :) Making the naming consistent outside of `beacon-chain.md` (e.g. in `fork-choice.md`) is on my todo list (hence the "draft" status of this PR). I'll resolve this point for now—feel free to reopen if you disagree.

specs/merge/beacon-chain.md
JustinDrake: Absolutely—on my todo to put in fork-choice.md (as stated in the PR description). Resolving for now this is a draft PR focused on beacon-chain.md only. Will fix everything once we have consensus on beacon-chain.md.

specs/merge/beacon-chain.md
JustinDrake: Understood!

As a heads up this function has not been removed, merely renamed :) Making the naming consistent outside of `beacon-chain.md` (e.g. in `fork-choice.md`) is on my todo list (hence the "draft" status of this PR). I'll resolve this point for now—feel free to reopen if you disagree.

specs/merge/beacon-chain.md
JustinDrake: `Union[OpaqueTransaction]` is a great stub :)

specs/merge/beacon-chain.md
JustinDrake: I went with an even cleaner option: `payload = body.execution_payload`. Is that good with you? :)

specs/merge/beacon-chain.md
JustinDrake: done!

specs/merge/beacon-chain.md
```diff
@@ -59,148 +60,138 @@ We define the following Python custom types for type hinting and readability:
...
+    block_number: uint64  # 'number' in the yellow paper
     gas_limit: uint64
     gas_used: uint64
     timestamp: uint64
-    receipt_root: Bytes32
-    logs_bloom: ByteVector[BYTES_PER_LOGS_BLOOM]
-    transactions: List[OpaqueTransaction, MAX_EXECUTION_TRANSACTIONS]
+    # Extra payload fields
+    block_hash: Hash32  # Hash of execution block
+    transactions: List[Transaction, MAX_TRANSACTIONS_PER_PAYLOAD]
```
mkalinin, line 115: Why don't we want to follow YP in case of `receipts_root` and `number`?

JustinDrake: * `receipt_root` vs `receipts_root`: I just reused your term :) Having said that, I believe your choice is the correct one, and is more consistent with the beacon chain spec which does not pluralise (e.g. `deposit_root` vs `deposits_root`).
* `number` vs `block_number`: I just think "number" is pretty vague as anything can be a number. Adding the `block_` prefix clarifies things and is consistent with `block_hash`.

specs/merge/beacon-chain.md
```diff
@@ -209,41 +200,37 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
-    """
-    if is_transition_completed(state):
-        assert execution_payload.parent_hash == state.latest_execution_payload_header.block_hash
-        assert execution_payload.number == state.latest_execution_payload_header.number + 1
-
-    assert execution_payload.timestamp == compute_time_at_slot(state, state.slot)
-
-    assert execution_engine.new_block(execution_payload)
-
+def process_execution_payload(state: BeaconState, body: BeaconBlockBody, execution_engine: ExecutionEngine) -> None:
```
mkalinin, line 212: Changing `execution_payload` to `body` in the function signature requires corresponding change in execution payload tests, they will have to use beacon blocks instead of pure payloads. Though, it might not be undesirable, just requires additional work.

specs/merge/beacon-chain.md
mkalinin: I was thinking about removing inconsistency between `receipt_root` on the beacon chain and `receipts_root` on the execution chain which might lead to message decoding bugs. This seems to be the only argument to follow the YP here. Agreed on `block_number`.

specs/merge/beacon-chain.md
```diff
@@ -59,148 +60,138 @@ We define the following Python custom types for type hinting and readability:
...
+
+#### `on_payload`
 
 ```python
-def compute_time_at_slot(state: BeaconState, slot: Slot) -> uint64:
-    slots_since_genesis = slot - GENESIS_SLOT
-    return uint64(state.genesis_time + slots_since_genesis * SECONDS_PER_SLOT)
+def on_payload(self: ExecutionEngine, execution_payload: ExecutionPayload) -> bool:
+    """
+    Returns ``True`` iff ``execution_payload`` is valid with respect to ``self.execution_state``.
```
mkalinin, line 188:
```suggestion
    Returns ``True`` if ``execution_payload`` is valid with respect to ``self.execution_state``.
```

protolambda: This is not a typo, see https://en.wikipedia.org/wiki/If_and_only_if

mkalinin: Wow, I didn't know that! this is cool 👍

specs/merge/beacon-chain.md
JustinDrake: Curious what @djrtwo thinks here :)

specs/merge/beacon-chain.md
JustinDrake: > just requires additional work

Right. I'm happy to do the work to fix tests :)

specs/merge/beacon-chain.md
djrtwo: My gut is to prioritize naming consistencies like the following:
1. Consistency with beacon chain and execution engine software (to make RPC naming conventions simple/consistent)
2. Consistency of beacon chain naming with itself/other fields
3. Consistency of beacon chain naming with yellow paper

If `receipts_root` is used across all execution engine software today, then I'd slightly prioritize breaking the beacon chain naming convention to be harmonious with the software it is communicating with. If there isn't consistency there, then keep beacon chain naming consistent with itself and make a note about the yellow paper name mismatch

specs/merge/beacon-chain.md
protolambda: I'm really not a fan of the `execution_payload -> body` change, it completely breaks the encapsulation of the processing. Can we just keep it `def process_execution_payload(state: BeaconState, payload: ExecutionPayload, engine: ExecutionEngine) -> None:` ?

JustinDrake: now passing the payload to `process_execution_payload` :)

specs/merge/beacon-chain.md
JustinDrake: I agree with Danny's prioritisation :)

Looks like Geth uses "receipt root" internally (see [here](https://github.com/ethereum/go-ethereum/search?q=receiptroot)) but that an inconsistency was introduced with Rayonism. See [this line](https://github.com/ethereum/go-ethereum/blob/f79cce5de98332e6d54eb298fce85c69d1082ee2/eth/catalyst/api_types.go#L49) which has both `ReceiptRoot` and `json:"receiptsRoot"`.

I believe the way forward here is to go with `receipt_root` which is consistent with Geth and the beacon chain (but inconsistent with the yellow paper).

specs/merge/beacon-chain.md
```diff
@@ -263,21 +263,9 @@ def initialize_beacon_state_from_eth1(eth1_block_hash: Bytes32,
...
-        state_root=Bytes32(),
-        number=uint64(0),
-        gas_limit=uint64(0),
-        gas_used=uint64(0),
-        timestamp=eth1_timestamp,
-        receipt_root=Bytes32(),
-        logs_bloom=ByteVector[BYTES_PER_LOGS_BLOOM](),
-        transactions_root=Root(),
-    )
+    # Initialize the execution payload header (with block number set to 0)
```
mkalinin, line 266: Maybe do it once to avoid duplication if we'd need to initialize more fields in the future?
```suggestion
    # [New in Merge] Initialize the execution payload header (with block number set to 0)
```

JustinDrake: done :)

specs/merge/beacon-chain.md
```diff
@@ -196,50 +188,45 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+        gas_limit=payload.gas_limit,
+        gas_used=payload.gas_used,
+        timestamp=payload.timestamp,
+        block_hash=payload.block_hash,
+        transactions_root=hash_tree_root(payload.transactions),
     )
 ```
 
-## Initialize state for pure Merge testnets and test vectors
+## Testing
```
djrtwo, line 225: This isn't just for testing, imo. It's also for pure merge networks which could be other "mainnets"

but.. I don't care too much.

djrtwo (approved):
lgtm

tests/core/pyspec/eth2spec/test/helpers/execution_payload.py
```diff
@@ -4,19 +4,19 @@ def build_empty_execution_payload(spec, state):
...
-    empty_txs = spec.List[spec.OpaqueTransaction, spec.MAX_EXECUTION_TRANSACTIONS]()
+    empty_txs = spec.List[spec.Transaction, spec.MAX_TRANSACTIONS_PER_PAYLOAD]()
 
     payload = spec.ExecutionPayload(
-        block_hash=spec.Hash32(),
         parent_hash=latest.block_hash,
         coinbase=spec.Bytes20(),
         state_root=latest.state_root,  # no changes to the state
-        number=latest.number + 1,
+        logs_bloom=spec.ByteVector[spec.BYTES_PER_LOGS_BLOOM](),  # TODO: zeroed logs bloom for empty logs ok?
```
mkalinin, line 13: All zeroes is what you get if there were no logs. I think this TODO can be removed

mkalinin (approved):
Looks good to me! 👍

protolambda (approved):
LGTM

specs/merge/beacon-chain.md
mkalinin: This function needs to be updated in merge/validator.md

tests/core/pyspec/eth2spec/test/helpers/execution_payload.py
JustinDrake: Heads up: leaving it in for the purpose of this PR.

Issue #2433, opened 2021-05-22
`sync_aggregate` vs `sync_committee` name


protolambda:
- the operation in the beacon block is called `sync_aggregate`
- the processing function is called `process_sync_committee`
- the processing argument is called `sync_aggregate`, the type is `SyncAggregate`
- the py tests module is called `test_process_sync_committee`
- the ssz file in the test vector output is called `sync_aggregate.ssz_snappy`
- the test handler is named `sync_committee`
- the test format doc of the `operations` runner is wrong: the table says the operation name is `sync_aggregate`, while it's really `sync_committee` in the test dir structure. The input name is indeed `sync_aggregate` though.

Can we make this consistent, and just pick one or the other in all places? I like `sync_aggregate` more: since there may be many other sync committee related things, but the `SyncAggregate` is the exact type we are dealing with in block processing.

hwwhww:
To me:
- Variable `sync_aggregate` is a `SyncAggregate` object
- Variable `sync_committee` is a list of the validator indices.
- Variable `aggregate` *usually* is an aggregate signature. But sometimes it's not.

---

It would nice to also rename all the `SyncAggregate` instances to `sync_aggregate`. For example:
```python
process_sync_committee(state: BeaconState, aggregate: SyncAggregate)
```

->

```python
process_sync_aggregate(state: BeaconState, sync_aggregate: SyncAggregate)
```

djrtwo:
agreed with @hwwhww's description. And for `process_` function names, we do `process_THE_THING` not `process_THE_GROUP_THAT_DID_THE_THING`, so `process_sync_aggregate` makes sense to me.

protolambda:
Fixed in #2469

PR #2473, opened 2021-06-08
small Altair cosmetic fix for consistency

JustinDrake:
for consistency with `process_randao`, `process_eth1_data`, `process_operations` which take `state` and `block.body` as inputs

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -446,7 +446,7 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
     process_randao(state, block.body)
     process_eth1_data(state, block.body)
     process_operations(state, block.body)  # [Modified in Altair]
-    process_sync_aggregate(state, block.body.sync_aggregate)  # [New in Altair]
+    process_sync_aggregate(state, block.body)  # [New in Altair]
 ```
 
 #### Modified `process_attestation`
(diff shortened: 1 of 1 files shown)


protolambda (changes requested):
Please don't, randao and and eth1 data processing were a big oversight IMHO. We should aim to encapsulate changes to the state as much as possible. There is no reason to inject the whole beacon block body into this function, and with every hardfork that changes the body type structure, clients will have trouble updating this function.

hwwhww:
I think the problem was how to align the test format and what's the definition of "operation". 
- [`sync_aggregate` is considered as "operations" in test format](https://github.com/ethereum/eth2.0-specs/blob/c821e3fced725995242e12037ad9ce502d5b2208/tests/formats/operations/README.md#condition). `process_sync_aggregate` is similar to `process_SSZ_CONTAINER(state, SSZ_CONTAINER)` functions  like `process_attestation(state, attestation)`.
- Currently, `process_operations(state: BeaconState, body: BeaconBlockBody)` only process "a List of operations".

Since renaming `process_operations` is not phase0 backward compatible, alternatively, I'd say we can consider moving `process_sync_aggregate` into `process_operations`.

related issue: #2433

JustinDrake:
I would have closed this issue based on Proto strong opinion (I don't care too much 😂 ).

@hwwhww: It sounds like you're in a position to make some sort of informed opinion regarding tests, `process_block` versus `process_operations`, etc. Assigning this issue to you—feel free to close if that's the right action :)

djrtwo:
closing this  as is
some additional convo going on in #2506

PR #2475, opened 2021-06-08
Add `message.topic` to gossipsub `message-id` in Altair

### LINKED ISSUE #2471: Can't publish SyncCommitteeSignature to multiple subnets
We’re hitting some interesting problems with publishing sync committee signatures when they have to be sent to multiple gossip topics (the validator is included in the sync committee multiple times, in different subnets).  This is extremely unlikely to happen on MainNet or the public testnets but very common on small scale test networks used during development or in automated tests.

Firstly, jvm-libp2p is tracking seen messages globally, not per topic so just ignores the request to publish to the additional topics because it’s already been sent once. Apparently the rust implementation applies the same restriction.  This makes it impossible to implement the validator spec correctly since the same signature can't be published to multiple subnets.

Secondly, the eth2 p2p-networking spec, explicitly has:
```
•	[IGNORE] There has been no other valid sync committee signature for the declared slot for the validator referenced by sync_committee_message.validator_index.
```
I believe this is meant to be per-subnet but we should make that clearer.

ralexstokes:
Fixes #2471.

In lieu of discriminating duplicated sync committee messages as suggested in #2471 by adding the `subnet_id` to the message itself, this PR suggests a solution by incorporating the topic name in to the gossipsub `message-id`. This solution should allow sync committee message duplicates to still remain distinct in the gossipsub machinery without needing to add an additional field to the message type.

Per discussion w/ @djrtwo and @protolambda, this avoids a potential DoS vector w/ the additional (unsigned) field and should harden message propagation across fork boundaries (as the fork digest will enter via the `message.topic`). This solution should also be a bit less intrusive than changing the sync committee structures further. Obviously open to other alternatives discovered during implementation :)

### DIFF
--- a/specs/altair/p2p-interface.md
+++ b/specs/altair/p2p-interface.md
@@ -77,11 +77,13 @@ The specification around the creation, validation, and dissemination of messages
 The derivation of the `message-id` has changed starting with Altair to incorporate the message `topic` along with the message `data`. These are fields of the `Message` Protobuf, and interpreted as empty byte strings if missing.
 The `message-id` MUST be the following 20 byte value computed from the message:
 * If `message.data` has a valid snappy decompression, set `message-id` to the first 20 bytes of the `SHA256` hash of
-  the concatenation of `MESSAGE_DOMAIN_VALID_SNAPPY` with the snappy decompressed message data and the topic name,
-  i.e. `SHA256(MESSAGE_DOMAIN_VALID_SNAPPY + snappy_decompress(message.data) + message.topic)[:20]`.
+  the concatenation of the following data: `MESSAGE_DOMAIN_VALID_SNAPPY`, the length of the topic byte string (encoded as little-endian `uint64`),
+  the topic byte string, and the snappy decompressed message data:
+  i.e. `SHA256(MESSAGE_DOMAIN_VALID_SNAPPY + uint_to_bytes(uint64(len(message.topic))) + message.topic + snappy_decompress(message.data)[:20]`.
 * Otherwise, set `message-id` to the first 20 bytes of the `SHA256` hash of
-  the concatenation of `MESSAGE_DOMAIN_INVALID_SNAPPY` with the raw message data and the topic name,
-  i.e. `SHA256(MESSAGE_DOMAIN_INVALID_SNAPPY + message.data + message.topic)[:20]`.
+  the concatenation of the following data: `MESSAGE_DOMAIN_INVALID_SNAPPY`, the length of the topic byte string (encoded as little-endian `uint64`),
+  the topic byte string, and the raw message data:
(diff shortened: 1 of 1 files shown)


specs/altair/p2p-interface.md
```diff
@@ -74,6 +74,15 @@ New topics are added in Altair to support the sync committees and the beacon blo
 
 The specification around the creation, validation, and dissemination of messages has not changed from the Phase 0 document.
 
+The derivation of the `message-id` has changed starting with Altair to incorporate the message topic along with the message data.
```
ralexstokes, line 77: is this clearer?

```suggestion
The derivation of the `message-id` has changed starting with Altair to incorporate the message topic along with the message data.
This topic is found in the `topic` field of a gossipsub message.
```

specs/altair/p2p-interface.md
```diff
@@ -74,6 +74,15 @@ New topics are added in Altair to support the sync committees and the beacon blo
...
+The `message-id` MUST be the following 20 byte value computed from the message:
+* If `message.data` has a valid snappy decompression, set `message-id` to the first 20 bytes of the `SHA256` hash of
+  the concatenation of `MESSAGE_DOMAIN_VALID_SNAPPY` with the snappy decompressed message data and the topic name,
+  i.e. `SHA256(MESSAGE_DOMAIN_VALID_SNAPPY + snappy_decompress(message.data) + message.topic)[:20]`.
```
protolambda, line 81: It would be better to separate the message topic from the data bytes, or otherwise we could get weird collisions (start of topic name being embedded in the message data as suffix instead of the topic ID as prefix).
I suggest we order it like:
1. snappy domain
2. length of topic byte string
3. topic byte string
4. message data (decompressed or not, depending on snappy domain)

specs/altair/p2p-interface.md
djrtwo: not entirely necessary but fine if you think it is clarifying

protolambda, line 77:
```suggestion
The derivation of the `message-id` has changed starting with Altair to incorporate the message `topic` along with the message `data`. These are fields of the `Message` Protobuf, and interpreted as empty byte strings if missing.
```

protolambda: Made an alternative suggestion in this spirit, that also clarifies the container type and missing-field case of protobuf.

specs/altair/p2p-interface.md
ralexstokes: yep, good idea! let me update

protolambda:
Hmm, I think we should clarify the transition process of the message ID. For altair it won't matter, but it has to happen all at the same time, or we get two different message IDs for the same message. With altair we drop all old topic names anyway, so I suppose that we just never use the new message ID function for old topics?

protolambda (approved):

PR #2479, opened 2021-06-11
Add randao to execution payload

mkalinin:
Adds the most recent `randao_mix` into execution payload.
This PR is a follow up to the corresponding discussion that took place in [Merge Implementers' Call 5](https://github.com/ethereum/pm/issues/331).

The value of new `randao` field in `ExecutionPayload` must be assigned to `difficulty` field of derived `PowBlock`. And further `randao` value is exposed by the `DIFFICULTY` opcode in the EVM. This way allows to preserve randomness property of the value returned by this opcode. Potentially, `DIFFICULTY` opcode will be renamed to `RANDAO` and is one of the post-Merge cleanups.

There two reasons why `randao` should be an explicit part of the payload:
- `ExecutionPayload` is the only input required to build corresponding `PowBlock`
- further compatibility with stateless execution

### DIFF
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -105,6 +105,7 @@ class ExecutionPayload(Container):
     timestamp: uint64
     receipt_root: Bytes32
     logs_bloom: ByteVector[BYTES_PER_LOGS_BLOOM]
+    randao: Bytes32  # 'difficulty' in the yellow paper
     transactions: List[OpaqueTransaction, MAX_EXECUTION_TRANSACTIONS]
 ```
 
@@ -126,6 +127,7 @@ class ExecutionPayloadHeader(Container):
     timestamp: uint64
     receipt_root: Bytes32
     logs_bloom: ByteVector[BYTES_PER_LOGS_BLOOM]
+    randao: Bytes32
     transactions_root: Root
 ```
 
@@ -198,7 +200,9 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
     process_operations(state, block.body)
     # Pre-merge, skip execution payload processing
     if is_execution_enabled(state, block):
-        process_execution_payload(state, block.body.execution_payload, EXECUTION_ENGINE)  # [New in Merge]
+        # [New in Merge]
+        randao_mix = get_randao_mix(state, get_current_epoch(state))
+        process_execution_payload(state, block.body.execution_payload, randao_mix, EXECUTION_ENGINE)
 ```
 
 #### Execution payload processing
@@ -208,6 +212,7 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
 ```python
 def process_execution_payload(state: BeaconState,
                               execution_payload: ExecutionPayload,
+                              randao_mix: Bytes32,
                               execution_engine: ExecutionEngine) -> None:
     """
     Note: This function is designed to be able to be run in parallel with the other `process_block` sub-functions
@@ -215,6 +220,7 @@ def process_execution_payload(state: BeaconState,
     if is_transition_completed(state):
         assert execution_payload.parent_hash == state.latest_execution_payload_header.block_hash
         assert execution_payload.number == state.latest_execution_payload_header.number + 1
+        assert execution_payload.randao == randao_mix
 
     assert execution_payload.timestamp == compute_time_at_slot(state, state.slot)
 
@@ -231,6 +237,7 @@ def process_execution_payload(state: BeaconState,
         timestamp=execution_payload.timestamp,
         receipt_root=execution_payload.receipt_root,
         logs_bloom=execution_payload.logs_bloom,
+        randao=execution_payload.randao,
         transactions_root=hash_tree_root(execution_payload.transactions),
     )
 ```
@@ -289,6 +296,7 @@ def initialize_beacon_state_from_eth1(eth1_block_hash: Bytes32,
         timestamp=eth1_timestamp,
         receipt_root=Bytes32(),
         logs_bloom=ByteVector[BYTES_PER_LOGS_BLOOM](),
+        randao=eth1_block_hash,
         transactions_root=Root(),
     )
 
--- a/specs/merge/validator.md
+++ b/specs/merge/validator.md
@@ -49,7 +49,7 @@ The body of this function is implementation dependent.
 The Consensus API may be used to implement this with an external execution engine.
 
 ```python
-def assemble_block(self: ExecutionEngine, block_hash: Hash32, timestamp: uint64) -> ExecutionPayload:
+def assemble_block(self: ExecutionEngine, block_hash: Hash32, timestamp: uint64, randao: Bytes32) -> ExecutionPayload:
     ...
 ```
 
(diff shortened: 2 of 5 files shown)


tests/core/pyspec/eth2spec/test/helpers/execution_payload.py
```diff
@@ -1,4 +1,4 @@
-def build_empty_execution_payload(spec, state):
+def build_empty_execution_payload(spec, state, randao_mix):
```
protolambda, line 1: This could be an optional parameter, to avoid the `, spec.Bytes32()` arguments everywhere

specs/merge/validator.md
```diff
@@ -80,11 +95,9 @@ def get_execution_payload(state: BeaconState,
...
-            timestamp = compute_time_at_slot(state, state.slot)
-            return execution_engine.assemble_block(pow_block.block_hash, timestamp)
+            return produce_execution_payload(state, pow_block.block_hash, timestamp, randao_reveal)
 
     # Post-merge, normal payload
-    execution_parent_hash = state.latest_execution_payload_header.block_hash
-    timestamp = compute_time_at_slot(state, state.slot)
-    return execution_engine.assemble_block(execution_parent_hash, timestamp)
+    parent_hash = state.latest_execution_payload_header.block_hash
+    return produce_execution_payload(state, parent_hash, timestamp, randao_reveal)
```
protolambda, line 102: The line that computed the timestamp went missing, it's undefined now. Move `timestamp = compute_time_at_slot(state, state.slot)` to the top of the function, to use it in both cases?

specs/merge/beacon-chain.md
```diff
@@ -289,6 +296,7 @@ def initialize_beacon_state_from_eth1(eth1_block_hash: Bytes32,
         timestamp=eth1_timestamp,
         receipt_root=Bytes32(),
         logs_bloom=ByteVector[BYTES_PER_LOGS_BLOOM](),
+        randao=eth1_block_hash,
```
protolambda, line 299: We could define `randao_seed = eth1_block_hash`, and use it in both the `randao_mixes` initialization and the `randao` field here. To clarify it's being uses as a seed, and not a copy-paste error.

specs/merge/beacon-chain.md
```diff
@@ -198,7 +200,9 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
     process_operations(state, block.body)
     # Pre-merge, skip execution payload processing
     if is_execution_enabled(state, block):
-        process_execution_payload(state, block.body.execution_payload, EXECUTION_ENGINE)  # [New in Merge]
+        # [New in Merge]
```
protolambda, line 203: The `if is_execution_enabled(state, block):` is also new in the Merge. Maybe unify the two comments?

specs/merge/beacon-chain.md
```diff
@@ -198,7 +200,9 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+        randao_mix = get_randao_mix(state, get_current_epoch(state))
```
protolambda, line 204: The state is already updated with the randao mix, and the below processing function has full access to the state. We can move this line into the `process_execution_payload` I think. Unlike the block body, the state is already part of the processing, so this doesn't really do much for encapsulation, and overriding the latest randao value in the state in a test case to cover possibly different behavior is not too difficult.

specs/merge/beacon-chain.md
```diff
@@ -126,6 +127,7 @@ class ExecutionPayloadHeader(Container):
     timestamp: uint64
     receipt_root: Bytes32
     logs_bloom: ByteVector[BYTES_PER_LOGS_BLOOM]
+    randao: Bytes32
```
protolambda, line 130: Not sure if we should call it `randao` or `randao_mix`. There is a mix-hash in eth1, which is completely different, part of ethash, so that may be confusing. That mix-hash is not part of the seal (see https://github.com/ethereum/go-ethereum/blob/eff998effbf466d4acbe599583dd8ec75f2af6a1/consensus/ethash/consensus.go#L606), but is part of the header and old consensus checks (https://github.com/ethereum/go-ethereum/blob/eff998effbf466d4acbe599583dd8ec75f2af6a1/consensus/ethash/consensus.go#L566)

tests/core/pyspec/eth2spec/test/helpers/execution_payload.py
mkalinin: I was thinking about that. But my gut is against default arguments in testing helpers, we may have a separate function instead, like `build_empty_execution_payload_with_randao`, wdyt?

specs/merge/beacon-chain.md
mkalinin: `process_execution_payload` is designed to be able to run in parallel to the beacon chain block processing function. This is why the source of randao mix depends on the way the function is called; in the spec case it is read from the state, in other cases it might be computed out of block's randao_reveal and previous randao_mix.

specs/merge/beacon-chain.md
mkalinin: `mix_hash` is a part of PoW seal but not a part of seal hash. I think that `randao` is a short and clean name, and it is less restrictive semantically. In the context of beacon chain `randao_mix` is associated with concrete value. And I would leave an ability to add something atop of `randao_mix` before it's passed to the payload in the future.

specs/merge/beacon-chain.md
```diff
@@ -231,6 +236,7 @@ def process_execution_payload(state: BeaconState,
         timestamp=execution_payload.timestamp,
         receipt_root=execution_payload.receipt_root,
         logs_bloom=execution_payload.logs_bloom,
+        randao=execution_payload.randao,
```
djrtwo, line 239: We should note somewhere here or in our own notes that this should probably be accompanied with an EIP. There are probably a handful of things that will need to make it into an EIP.

tests/core/pyspec/eth2spec/test/helpers/block.py
```diff
@@ -35,6 +35,11 @@ def apply_randao_reveal(spec, state, block, proposer_index=None):
     block.body.randao_reveal = bls.Sign(privkey, signing_root)
 
 
+def compute_randao_mix(spec, state, randao_reveal):
```
djrtwo, line 38: `merge/validator.md` should be built into the `spec` and thus this should be available as `spec.compute_randao_mix` and doesn't need ot be redefined.

tests/core/pyspec/eth2spec/test/helpers/execution_payload.py
```diff
@@ -1,4 +1,8 @@
-def build_empty_execution_payload(spec, state):
+def build_empty_execution_payload_with_randao(spec, state):
+    return build_empty_execution_payload(spec, state, spec.Bytes32())
```
djrtwo, line 2: This is a stubbed randao, rather than the real thing, no?

The name of this function implies to me that it will have a correct randao reveal

djrtwo (commented):
nice! a bit confused about `Bytes32()` in the helpers for randao reveal but otherwise, looking good

JustinDrake:
suggested polishing:

Screenshot 2021-06-20 at 20 06 59

Screenshot 2021-06-20 at 20 07 51

djrtwo:
I'm okay with the suggested cleanups. 

Although it is an important design goal for execution layer to be parallelizable, I don't think the added param moves the needle too much on making that clear to the reader. I would suggest noting the parallelizability in the header comment (or pre-function text) and make it clear there if there are any important considerations.

mkalinin:
Thanks everyone for your inputs! I've accepted suggested polishing with related comment about `randao_mix` dependency in `process_execution_payload` and its impact on ability to process the payload in parallel. I am ok with this change now and it can be merged once approved.

specs/merge/beacon-chain.md
```diff
@@ -196,12 +198,15 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
 
 #### `process_execution_payload`
 
+*Note:* This function depends on `process_randao` function call as it retrieves the most recent randao mix from the `state`. Implementations that are considering parallel processing of execution payload with respect to beacon chain state transition function should work around this dependency.
```
JustinDrake, line 201: I would lean towards removing this note. It's "obvious" that there is a parallelisation opportunity and client implementers have become experts at handling optimisations that the spec doesn't address.

mkalinin: IMO, the second statement may be removed as it can be deduced from the first one. Don't you think that removing this entire comment would affect readability? How common the spec is in leaving implicit dependencies unattended?

djrtwo: I think it's pretty reasonable as is, plus I'm cutting a release today and want to get this in.

Let's leave it here for now. Remove if this info makes it into somewhere else -- e.g. annotated spec or implementaton guide or something

djrtwo (approved):

PR #2482, opened 2021-06-16
Fix typo: `change()` should be invoked on `ShardWork.status`

ericsson49:
There is an inconsistency in how `change()` is called in the `sharding` spec: sometimes it's invoked on a `ShardWork` instance, while other times - on its [`status`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/sharding/beacon-chain.md#shardwork) field (of type `Union`).
The latter variant seems to be more [appropriate](https://github.com/protolambda/remerkleable/blob/ba34ea3984ac9cc5c963b67a053c8745f4e16815/remerkleable/union.py#L125).


protolambda (approved):
Thanks, I missed that when wrapping it with the ShardWork type. After the rebase of Merge on Altair we will make the spec executable, and can avoid more issues like these early on.

PR #2484, opened 2021-06-16
Fix typing problem in `reset_pending_shard_work`

ericsson49:
[reset_pending_shard_work](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/sharding/beacon-chain.md#reset_pending_shard_work) passes `committee_index` of type `int` as a third parameter to `get_beacon_committee`, while it should be of type `CommitteeIndex`


protolambda (approved):
LGTM

PR #2486, opened 2021-06-18
Shard builder <> proposer separation

protolambda:
Warning: this is highly experimental, not a definitive sharding design.

### Changes

- [x] Refactor to have "blobs" (data bundles) and "blocks" (selected bundles)
- [x] Define shard builders, creating shard blob headers to bid to proposers (data-transactions)
- [x] Remove the shard-committee requirement
- [x] Shard proposers charge shard builders for selecting their blob
- [x] Move availability (publishing) responsibility to shard blob builder (if they do not publish the full data, they lose their fee for no gain)
- [x] Network spec updates
- [ ] Sharding validator spec (new)
- [x] Builder pubkey and balance state
- [x] Reward/fee processing
- [ ] Change shard-block subnet backbone to use network-identity (p2p peer ID or node ID) to map nodes to subnets, and penalize (through p2p scoring) nodes that do not contribute to the backbone.

Here is a quick sketch of the message timing:
image

### Motivation

- **"Firewall" sharding MEV**:
  - If shard proposers get an equal opportunity to tap into the market around extracted value, validator work does not centralize the same way as shard builders might, while still enjoying fees.
  - Anyone is able to become a shard builder
  - See related eth1-mainnet discussions
    - https://ethresear.ch/t/proposer-block-builder-separation-friendly-fee-market-designs/9725
    - https://ethresear.ch/t/flashbots-frontrunning-the-mev-crisis/8251
- **Open shard data usage to any application**:
  - Shard blobs (flattened L2 transaction bundles) can be anything, and the validator does not have to learn the protocol
  - Shard proposers can be 
- **Reduce validator requirements**
  - Network load reduction
    - Previous shard-committee network backbone was not good enough (a ~1/5000 chance of proposing at any slot to do propagate full shard data)
    - Proposers can listen on a global data-tx topic. 64*N data-transactions without payload, per 12 seconds, is manageable. Or e.g. split in `shard_count / 4` subnets. Trade-off between proposer subscription management and gossip topic load. But greatly reduced compared to propagating full shard-blobs for a day for a single proposal.
    - No direct communication between builder and proposer necessary. The block header is small and propagated quickly, the builder picks up on it, takes the signature, and publishes the block (wrap their blob with the signature they just received). Multiple nodes can perform this for extra availability.
  - No need to make shard data available as shard proposer, the builder can specialize in this, and has incentive to.
    - Builder is not encumbered with the same validator network privacy concerns
    - Builder can stay on one shard, does not have to get shuffled, good for network stability
    - Builder can publish as group of nodes / L2 construction, with L2-specific compensation/organization
    - Builder incentive is better aligned with data publishing work after paying a fee upfront, independent of their performance.
- More fair MEV:
    - Builders can lock in a commitment to their blob (data bundle) before revealing it.
    - If reorgs or committee-mingling are a concern (i.e. full data was published, but not recognized by validators), a L2 can opt to encrypt their shard data and reveal the key later, independent of L1.
    - Separating out MEV means decentralized validator pools stand a better chance to split rewards fairly.
- No shard-committee caching necessary, simplifying the beacon node

### Related problems and proposed solutions

- *problem*: **volatile asks for data**, shard data usage is not constant
  *solution*: **EIP-1559 elasticity**, a base-fee is adjusted to target a desired capacity, while the actual capacity is elastic (like EIP-1559 does with gas).
  
- *problem*: **fee manipulation**, validators (N.B. proposers), should not gain/lose extra fees by manipulating capacity (capacity and prioritization are separated).
  *solution*: **burn the base fee**, as confirmed data decreases available resources, this compensates for protocol capacity, without favoring any validator in particular.

- *problem*: **shard proposers unfit to build**, shard proposers are selected on short notice, cannot carry the data-availability publisher responsibility well, nor specialize easily into wide variety of buildable blobs  
  *solution*: **builder separation**, shard proposers just perform leader work (selecting transactions), as random shuffled workers. The publisher responsibility is moved to builders.
  
- *problem*: **data-tx promises**, data that never becomes available is useless, but can occupy just as many resources (e.g. committee listens for data, maintains subnet, and may be attacked with double proposals)
  *solution*: **builder always pays fees**: shard proposers are paid for their opportunity by builders, regardless of availability.
  The base fee is paid for claiming the capacity as builder. Builders need to publish to utilize their opportunity.

- *problem*: **load-balancing**, if one shard is popular, and the other is not, then distributing load would be better
  *solution*: **uniformity, no chaining**, shards are as uniform as possible, to easily switch between them,
  use multiple of them per slot for a single application, and share a base-fee for cross-shard capacity elasticity.

- *problem*: **shard proposer privacy/targetability**
  *solution*: no weaker than beacon-proposers: a shard-proposer is randomly selected just in time, separates their network identity, and can opt for sentry/proxy/etc. approaches if necessary. 
  Long-term randomness improves (VDF) and single-secret-leader-election (SSLE) may be used. Involved gossip-topics are all global, no long-term topic->validator reversal concerns.

- *problem*: **shard data-tx market accessibility**, fast proposer shuffling requires proposers to be aware of the latest transactions.
  *solution*: data-transactions are small fee-paying headers, thus possible to share on a global gossip topic, and efficient / DoS-safe if filtered by fee.

- *problem*: **shard subnet stability**, if shard-proposers are only temporary, shard subnets have no stable participants
  *solution*: **builders and DHT-like assignments**:
    - builders are incentivized to always be subscribed to publish their data upon getting selected, and can do so more frequently than a normal shard proposer can
    - propagating shard-data of a single shard is doable by every node, subscribing to `shard_id = hash(nodeID.to_bytes() + ((slot + offset) // sub_window_duration).to_bytes()) % shard_count` can become part of peer-scoring (negative if not subscribed as expected).

- *problem/solution*: **MEV**, see motivation section.
- *problem/solution*: **Shard proposer requirements**, see motivation section.

### EIP-1559 translation to data layer

- Work unit: `gas` -> `samples`, data is measured in samples, subdivided into points (each point is approx. 32 bytes of data). Every sample in a block is an equal amount of work (for fee purposes).
- Transaction fields:
    - `max_priority_fee_per_gas` -> `max_priority_fee_per_sample`: maximum fees going towards shard proposer, after base fee is paid for
    - `max_fee_per_gas` -> `max_fee_per_sample`: cost bound of transaction
    - `gas_limit` -> ~~`sample_limit`~~: not necessary, fixed amount of data points (and thus samples) per blob
- Processing context:
    - `base_fee_per_gas` -> `base_fee_per_sample`: base fee, continually adjusted to target `TARGET_SAMPLES_PER_BLOB`
- Base fee adjustments:
    - `post.base_fee_per_sample = ` # TODO

### Roles

- builder:
   - dedicated to chosen work (L2s, shards, etc.)
   - burns for capacity
   - pays for priority
   - publishes partially signed headers as data-txs
   - publishes data for opportunity cost
- shard proposer
   - stateless, shuffled often
   - prioritizes for largest fee
   - selects for reward
   - completes data-tx signature to propose
   - proposes once, or slashable
   - publishes signed data-tx asap for better attester shard-reward share chances
- beacon proposer
   - includes new shard headers, 
   - slashes double shard proposals for reward
   - includes attestations for share of attester rewards
   - can back-fill missing shard-headers, and re-include attestations to get them confirmed
- attester
   - attests shard blobs, rewarded if confirmed
   - ensures availability with proof-of-custody scheme (= easily trip and slashed if acting lazy)

### Incentives

Fees/burns/penalties/rewards/incentives/shares:
- [burn] builder->chain: base-fee, target capacity, protect protocol
- [fee] builder->shard-proposer: priority fee, market to reduce data-tx spam
- [incentive] builder->attester: publishes data to attest, or opportunity lost, paid for nothing
- [slashing] shard-proposer-slashing
  - [incentive] shard-proposer->shard-proposer: only select single data-tx
  - [share] shard-proposer->beacon-proposer: share of slashing for finder
  - [share] burn the rest (against shard-beacon proposers collusion case)
- [incentive] beacon-proposer->beacon-proposer: include shard header, or miss out on attestation reward
- [reward] shard attester reward
  - [share] chain->attester: attested correct shard root
  - [share] chain->beacon-proposer: inclusion reward
- [incentive] user->builder: users can fund builder with any value to build something for them (e.g. pack rollup transactions), builder pays ETH to protocol
- Work in progress! (needs deeper look into attacks/incentives)

### TODO

- [x] Builder->Proposer fee payment computation. (like EIP 1559)
- [ ] Look into bug (and possible fix) found by @dankrad : a shard proposer publishing late may gain a fee, but the builder needs time to make the data available. It's possible to include the header on-chain the slot it's designated for, encouraging that may alleviate some of the concern. Alternatively incentives with an attestation bit for availability may help (iirc).
- [x] Adjust rewards (probably wait for rebase onto Altair, when we implement shard attestation rewards as well)
- [ ] Proofreads/review. Lots of new types and names, imminent mix-up/typos.
- [ ] Testing. We will make the spec executable for unit-testing and test-vector production after the rebase of the Merge and Sharding spec onto Altair.

### Out of scope

There are more (very experimental) ideas;
- Builder deposits
- Builder withdrawals
- Different types of builders. Should contracts be able to pay a fee for shard data? I.e. replace builder signature with something more generic to ensure fee payment. Data-construction may still be a problem though. BLS already allows for some more sophisticated signing setups than just a single signer.
- Shard staggering (sub slot delays per shard, to offer lower latency availability). Maybe for sharding refactor round 3 (if this is 2, and 1 was #2455)

And special thanks to @Nashatyrev for continued sharding prototype work on Teku. The spec is unstable at times, implementing a running prototype is not light work.

### DIFF
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -17,7 +17,7 @@
   - [Shard Work Status](#shard-work-status)
 - [Preset](#preset)
   - [Misc](#misc-1)
-  - [Shard block samples](#shard-block-samples)
+  - [Shard blob samples](#shard-blob-samples)
   - [Precomputed size verification points](#precomputed-size-verification-points)
   - [Gwei values](#gwei-values)
 - [Configuration](#configuration)
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -412,7 +412,7 @@ class ShardWork(Container):
     # Upon confirmation the data is reduced to just the commitment.
     status: Union[                                                   # See Shard Work Status enum
               None,                                                  # SHARD_WORK_UNCONFIRMED
-              ConfirmedDataCommitment,                               # SHARD_WORK_CONFIRMED
+              AttestedDataCommitment,                                # SHARD_WORK_CONFIRMED
               List[PendingShardHeader, MAX_SHARD_HEADERS_PER_SHARD]  # SHARD_WORK_PENDING
             ]
 ```
@@ -737,7 +737,7 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
     max_fee = blob_summary.max_fee_per_sample * samples
 
     # Builder must have sufficient balance, even if max_fee is not completely utilized
-    assert state.blob_builder_balances[header.builder_index] > max_fee
+    assert state.blob_builder_balances[header.builder_index] >= max_fee
 
     base_fee = state.shard_sample_price * samples
     # Base fee must be paid
@@ -748,7 +748,8 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
     priority_fee = min(max_fee - base_fee, max_priority_fee)
 
     # Burn base fee, take priority fee
-    decrease_balance(state, header.builder_index, base_fee + priority_fee)
+    # priority_fee <= max_fee - base_fee, thus priority_fee + base_fee <= max_fee, thus sufficient balance.
+    state.blob_builder_balances[header.builder_index] -= base_fee + priority_fee
     # Pay out priority fee
     increase_balance(state, header.proposer_index, priority_fee)
 
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -9,6 +9,7 @@
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
 - [Introduction](#introduction)
+  - [Glossary](#glossary)
 - [Custom types](#custom-types)
 - [Constants](#constants)
   - [Misc](#misc)
@@ -23,15 +24,25 @@
 - [Updated containers](#updated-containers)
   - [`AttestationData`](#attestationdata)
   - [`BeaconBlockBody`](#beaconblockbody)
+  - [`Builder`](#builder)
   - [`BeaconState`](#beaconstate)
 - [New containers](#new-containers)
   - [`DataCommitment`](#datacommitment)
+  - [ShardBlobBody](#shardblobbody)
   - [`ShardBlobBodySummary`](#shardblobbodysummary)
+  - [`ShardBlob`](#shardblob)
   - [`ShardBlobHeader`](#shardblobheader)
+  - [`SignedShardBlob`](#signedshardblob)
   - [`SignedShardBlobHeader`](#signedshardblobheader)
+  - [ShardBlock](#shardblock)
+  - [`ShardBlockHeader`](#shardblockheader)
+  - [`SignedShardBlock`](#signedshardblock)
+  - [`SignedShardBlockHeader`](#signedshardblockheader)
   - [`PendingShardHeader`](#pendingshardheader)
   - [`ShardBlobReference`](#shardblobreference)
   - [`SignedShardBlobReference`](#signedshardblobreference)
+  - [`ShardBlockReference`](#shardblockreference)
+  - [`SignedShardBlockReference`](#signedshardblockreference)
   - [`ShardProposerSlashing`](#shardproposerslashing)
   - [`ShardWork`](#shardwork)
 - [Helper functions](#helper-functions)
@@ -51,6 +62,7 @@
   - [Block processing](#block-processing)
     - [Operations](#operations)
       - [Extended Attestation processing](#extended-attestation-processing)
+    - [`charge_builder`](#charge_builder)
       - [`process_shard_header`](#process_shard_header)
       - [`process_shard_proposer_slashing`](#process_shard_proposer_slashing)
   - [Epoch transition](#epoch-transition)
@@ -68,6 +80,13 @@ This document describes the extensions made to the Phase 0 design of The Beacon 
 based on the ideas [here](https://hackmd.io/G-Iy5jqyT7CXWEz8Ssos8g) and more broadly [here](https://arxiv.org/abs/1809.09044),
 using KZG10 commitments to commit to data to remove any need for fraud proofs (and hence, safety-critical synchrony assumptions) in the design.
 
+### Glossary
+
+- **Data**: A list of KZG points, to translate a byte string into
+- **Blob**: Data with commitments and meta-data, like a flattened bundle of L2 transactions.
+- **Builder**: Builds blobs and bids for proposal slots with fee-paying blob-headers, responsible for availability.
+- **Shard proposer**: Validator, selects a signed blob-header, taking bids for shard data opportunity.
+- **Shard block**: Unique per `(slot, shard, proposer)`, selected signed blob 
 
 ## Custom types
 
@@ -78,6 +97,7 @@ We define the following Python custom types for type hinting and readability:
 | `Shard` | `uint64` | A shard number |
 | `BLSCommitment` | `Bytes48` | A G1 curve point |
 | `BLSPoint` | `uint256` | A number `x` in the range `0 <= x < MODULUS` |
+| `BuilderIndex` | `uint64` | Builder registry index |
 
 ## Constants
 
@@ -97,7 +117,7 @@ The following values are (non-configurable) constants used throughout the specif
 | Name | Value |
 | - | - |
 | `DOMAIN_SHARD_PROPOSER` | `DomainType('0x80000000')` |
-| `DOMAIN_SHARD_COMMITTEE` | `DomainType('0x81000000')` |
+| `DOMAIN_SHARD_BUILDER` | `DomainType('0x81000000')` |
 
 ### Shard Work Status
 
@@ -118,6 +138,7 @@ The following values are (non-configurable) constants used throughout the specif
 | `MAX_SHARD_PROPOSER_SLASHINGS` | `2**4` (= 16) | Maximum amount of shard proposer slashing operations per block |
 | `MAX_SHARD_HEADERS_PER_SHARD` | `4` | |
 | `SHARD_STATE_MEMORY_SLOTS` | `uint64(2**8)` (= 256) | Number of slots for which shard commitments and confirmation status is directly available in the state |
+| `BUILDER_REGISTRY_LIMIT` | `uint64(2**40)` (= 1,099,511,627,776) | builders |
 
 ### Shard block samples
 
@@ -162,8 +183,8 @@ class AttestationData(Container):
     # FFG vote
     source: Checkpoint
     target: Checkpoint
-    # Shard header root
-    shard_header_root: Root  # [New in Sharding]
+    # Hash-tree-root of ShardBlock
+    shard_block_root: Root  # [New in Sharding]
 ```
 
 ### `BeaconBlockBody`
@@ -171,7 +192,16 @@ class AttestationData(Container):
 ```python
 class BeaconBlockBody(merge.BeaconBlockBody):  # [extends The Merge block body]
     shard_proposer_slashings: List[ShardProposerSlashing, MAX_SHARD_PROPOSER_SLASHINGS]
-    shard_headers: List[SignedShardBlobHeader, MAX_SHARDS * MAX_SHARD_HEADERS_PER_SHARD]
+    shard_headers: List[SignedShardBlockHeader, MAX_SHARDS * MAX_SHARD_HEADERS_PER_SHARD]
+```
+
+### `Builder`
+
+```python
+class Builder(Container):
+    pubkey: BLSPubkey
+    # TODO: fields for either an expiry mechanism (refunding execution account with remaining balance) 
+    #  and/or a builder-transaction mechanism.
 ```
 
 ### `BeaconState`
@@ -182,15 +212,15 @@ class BeaconState(merge.BeaconState):
     previous_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
     current_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
     # [New fields]
+    # Builder registry.
+    builders: List[Builder, BUILDER_REGISTRY_LIMIT]
+    builder_balances: List[Gwei, BUILDER_REGISTRY_LIMIT]
     # A ring buffer of the latest slots, with information per active shard.
     shard_buffer: Vector[List[ShardWork, MAX_SHARDS], SHARD_STATE_MEMORY_SLOTS]
     shard_gasprice: uint64
 ```
 
 ## New containers
-
-The shard data itself is network-layer only, and can be found in the [P2P specification](./p2p-interface.md).
-The beacon chain registers just the commitments of the shard data.
 
 ### `DataCommitment`
 
@@ -202,7 +232,32 @@ class DataCommitment(Container):
     length: uint64
 ```
 
+### ShardBlobBody
+
+Unsigned shard data, bundled by a shard-builder.
+Unique, signing different bodies as shard proposer for the same `(slot, shard)` is slashable.
+
+```python
+class ShardBlobBody(Container):
+    # The actual data commitment
+    commitment: DataCommitment
+    # Proof that the degree < commitment.length
+    degree_proof: BLSCommitment
+    # The actual data. Should match the commitment and degree proof.
+    data: List[BLSPoint, POINTS_PER_SAMPLE * MAX_SAMPLES_PER_BLOCK]
+    # Latest block root of the Beacon Chain, before shard_blob.slot
+    beacon_block_root: Root
+    # Builder of the data, pays data-fee to proposer
+    builder_index: BuilderIndex
+    # TODO: fee payment amount fields (EIP 1559 like)
+```
+
 ### `ShardBlobBodySummary`
+
+Summary version of the `ShardBlobBody`, omitting the data payload, while preserving the data-commitments.
+
+The commitments are not further collapsed to a single hash, 
+to avoid an extra network roundtrip between proposer and builder, to include the header on-chain more quickly.
 
 ```python
 class ShardBlobBodySummary(Container):
@@ -214,26 +269,100 @@ class ShardBlobBodySummary(Container):
     data_root: Root
     # Latest block root of the Beacon Chain, before shard_blob.slot
     beacon_block_root: Root
+    # Builder of the data, pays data-fee to proposer
+    builder_index: BuilderIndex
+    # TODO: fee payment amount fields (EIP 1559 like)
+```
+
+### `ShardBlob`
+
+`ShardBlobBody` wrapped with the header data that is unique to the shard blob proposal.
+
+```python
+class ShardBlob(Container):
+    slot: Slot
+    shard: Shard
+    # Blob contents
+    body: ShardBlobBody
 ```
 
 ### `ShardBlobHeader`
 
+Header version of `ShardBlob`. Separates designation (slot, shard) and contents (blob).
+
 ```python
 class ShardBlobHeader(Container):
-    # Slot and shard that this header is intended for
     slot: Slot
     shard: Shard
-    # SSZ-summary of ShardBlobBody
+    # Blob contents, without the full data
     body_summary: ShardBlobBodySummary
-    # Proposer of the shard-blob
-    proposer_index: ValidatorIndex
+```
+
+### `SignedShardBlob`
+
+Full blob data, signed by the shard builder, ensuring fee payment.
+
+```python
+class SignedShardBlob(Container):
+    message: ShardBlob
+    signature: BLSSignature
 ```
 
 ### `SignedShardBlobHeader`
 
+Header of the blob, the signature is equally applicable to `SignedShardBlob`.
+Shard proposers can accept `SignedShardBlobHeader` as a data-transaction.
+
 ```python
 class SignedShardBlobHeader(Container):
     message: ShardBlobHeader
+    signature: BLSSignature
+```
+
+### ShardBlock
+
+Full blob data signed by builder, to be confirmed by proxy as `ShardBlockHeader`.
+
+```python
+class ShardBlock(Container):
+    # Shard data with fee payment by bundle builder
+    signed_blob: SignedShardBlob
+    # Proposer of the shard-blob
+    proposer_index: ValidatorIndex
+```
+
+### `ShardBlockHeader`
+
+Header version of `ShardBlock`, selecting a `SignedShardBlobHeader`.
+
+```python
+class ShardBlockHeader(Container):
+    # Shard commitments and fee payment by blob builder
+    signed_blob_header: SignedShardBlobHeader
+    # Proposer of the shard-blob
+    proposer_index: ValidatorIndex
+```
+
+### `SignedShardBlock`
+
+Shard blob, signed for payment, and signed for proposal. Propagated to attesters.
+
+```python
+class SignedShardBlock(Container):
+    message: ShardBlock
+    signature: BLSSignature
+```
+
+### `SignedShardBlockHeader`
+
+Header version of `SignedShardBlock`, substituting the full data within the blob for just the hash-tree-root.
+
+The signature is equally applicable to `SignedShardBlock`, 
+which the builder can publish as soon as the signed header is seen. 
+
+```python
+class SignedShardBlockHeader(Container):
+    message: ShardBlockHeader
     signature: BLSSignature
 ```
 
@@ -243,7 +372,7 @@ class SignedShardBlobHeader(Container):
 class PendingShardHeader(Container):
     # KZG10 commitment to the data
     commitment: DataCommitment
-    # hash_tree_root of the ShardHeader (stored so that attestations can be checked against it)
+    # hash_tree_root of the ShardBlockHeader (stored so that attestations can be checked against it)
     root: Root
     # Who voted for the header
     votes: Bitlist[MAX_VALIDATORS_PER_COMMITTEE]
@@ -255,18 +384,19 @@ class PendingShardHeader(Container):
 
 ### `ShardBlobReference`
 
+Reference version of `ShardBlobHeader`, substituting the body for just a hash-tree-root.
+
 ```python
 class ShardBlobReference(Container):
-    # Slot and shard that this reference is intended for
     slot: Slot
     shard: Shard
-    # Hash-tree-root of ShardBlobBody
+    # Blob hash-tree-root for reference, enough for uniqueness
     body_root: Root
-    # Proposer of the shard-blob
-    proposer_index: ValidatorIndex
 ```
 
 ### `SignedShardBlobReference`
+
+`ShardBlobReference`, signed by the blob builder. The builder-signature is part of the block identity.
 
 ```python
 class SignedShardBlobReference(Container):
@@ -274,12 +404,30 @@ class SignedShardBlobReference(Container):
     signature: BLSSignature
 ```
 
+### `ShardBlockReference`
+
+```python
+class ShardBlockReference(Container):
+    # Blob, minimized for efficient slashing
+    signed_blob_reference: SignedShardBlobReference
+    # Proposer of the shard-blob
+    proposer_index: ValidatorIndex
+```
+
+### `SignedShardBlockReference`
+
+```python
+class SignedShardBlockReference(Container):
+    message: ShardBlockReference
+    signature: BLSSignature
+```
+
 ### `ShardProposerSlashing`
 
 ```python
 class ShardProposerSlashing(Container):
-    signed_reference_1: SignedShardBlobReference
-    signed_reference_2: SignedShardBlobReference
+    signed_reference_1: SignedShardBlockReference
+    signed_reference_2: SignedShardBlockReference
 ```
 
 ### `ShardWork`
@@ -516,7 +664,7 @@ def update_pending_shard_work(state: BeaconState, attestation: Attestation) -> N
     current_headers: Sequence[PendingShardHeader] = committee_work.status.value
 
     # Find the corresponding header, abort if it cannot be found
-    header_index = [header.root for header in current_headers].index(attestation.data.shard_header_root)
+    header_index = [header.root for header in current_headers].index(attestation.data.shard_block_root)
 
     pending_header: PendingShardHeader = current_headers[header_index]
     full_committee = get_beacon_committee(state, attestation.data.slot, attestation.data.index)
@@ -554,36 +702,63 @@ def update_pending_shard_work(state: BeaconState, attestation: Attestation) -> N
             )
 ```
 
+
+#### `charge_builder`
+
+```python
+def charge_builder(state: BeaconState, index: BuilderIndex, fee: Gwei) -> None:
+    """
+    Decrease the builder balance at index ``index`` by ``fee``, with underflow check.
+    """
+    assert state.builder_balances[index] >= fee
+    state.builder_balances[index] -= fee
+```
+
 ##### `process_shard_header`
 
 ```python
-def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeader) -> None:
-    header = signed_header.message
+def process_shard_header(state: BeaconState, signed_block_header: SignedShardBlockHeader) -> None:
+    block_header: ShardBlockHeader = signed_block_header.message
+    signed_blob_header: SignedShardBlobHeader = block_header.signed_blob_header
+    blob_header: ShardBlobHeader = signed_blob_header.message
+    slot = blob_header.slot
+    shard = blob_header.shard
+
     # Verify the header is not 0, and not from the future.
-    assert Slot(0) < header.slot <= state.slot
-    header_epoch = compute_epoch_at_slot(header.slot)
+    assert Slot(0) < slot <= state.slot
+    header_epoch = compute_epoch_at_slot(slot)
     # Verify that the header is within the processing time window
     assert header_epoch in [get_previous_epoch(state), get_current_epoch(state)]
     # Verify that the shard is active
-    assert header.shard < get_active_shard_count(state, header_epoch)
+    assert shard < get_active_shard_count(state, header_epoch)
     # Verify that the block root matches,
     # to ensure the header will only be included in this specific Beacon Chain sub-tree.
-    assert header.body_summary.beacon_block_root == get_block_root_at_slot(state, header.slot - 1)
+    assert blob_header.body_summary.beacon_block_root == get_block_root_at_slot(state, slot - 1)
 
     # Check that this data is still pending
-    committee_work = state.shard_buffer[header.slot % SHARD_STATE_MEMORY_SLOTS][header.shard]
+    committee_work = state.shard_buffer[slot % SHARD_STATE_MEMORY_SLOTS][shard]
     assert committee_work.status.selector == SHARD_WORK_PENDING
 
     # Check that this header is not yet in the pending list
     current_headers: List[PendingShardHeader, MAX_SHARD_HEADERS_PER_SHARD] = committee_work.status.value
-    header_root = hash_tree_root(header)
+    header_root = hash_tree_root(block_header)
     assert header_root not in [pending_header.root for pending_header in current_headers]
 
     # Verify proposer
-    assert header.proposer_index == get_shard_proposer_index(state, header.slot, header.shard)
-    # Verify signature
-    signing_root = compute_signing_root(header, get_domain(state, DOMAIN_SHARD_PROPOSER))
-    assert bls.Verify(state.validators[header.proposer_index].pubkey, signing_root, signed_header.signature)
+    assert block_header.proposer_index == get_shard_proposer_index(state, slot, shard)
+    # Verify proposer signature
+    block_signing_root = compute_signing_root(block_header, get_domain(state, DOMAIN_SHARD_PROPOSER))
+    proposer_pubkey = state.validators[block_header.proposer_index].pubkey
+    assert bls.Verify(proposer_pubkey, block_signing_root, signed_block_header.signature)
+
+    # Verify builder requirements
+    blob_summary: ShardBlobBodySummary = blob_header.body_summary
+    builder_index = blob_summary.builder_index
+
+    # Verify builder signature
+    builder = state.builders[builder_index]
+    blob_signing_root = compute_signing_root(blob_header, get_domain(state, DOMAIN_SHARD_BUILDER))  # TODO new constant
+    assert bls.Verify(builder.pubkey, blob_signing_root, signed_blob_header.signature)
 
     # Verify the length by verifying the degree.
     body_summary = header.body_summary
@@ -594,12 +769,16 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
         == bls.Pairing(body_summary.commitment.point, G2_SETUP[-body_summary.commitment.length])
     )
 
+    # Charge builder, with hard balance requirement
+    fee = Gwei(123)  # TODO
+    charge_builder(state, builder_index, fee)
+
     # Initialize the pending header
-    index = compute_committee_index_from_shard(state, header.slot, header.shard)
-    committee_length = len(get_beacon_committee(state, header.slot, index))
+    index = compute_committee_index_from_shard(state, slot, shard)
+    committee_length = len(get_beacon_committee(state, slot, index))
     initial_votes = Bitlist[MAX_VALIDATORS_PER_COMMITTEE]([0] * committee_length)
     pending_header = PendingShardHeader(
-        commitment=body_summary.commitment,
+        commitment=blob_summary.commitment,
         root=header_root,
         votes=initial_votes,
         weight=0,
@@ -619,17 +798,22 @@ The goal is to ensure that a proof can only be constructed if `deg(B) < l` (ther
 
 ```python
 def process_shard_proposer_slashing(state: BeaconState, proposer_slashing: ShardProposerSlashing) -> None:
-    reference_1 = proposer_slashing.signed_reference_1.message
-    reference_2 = proposer_slashing.signed_reference_2.message
+    reference_1: ShardBlockReference = proposer_slashing.signed_reference_1.message
+    reference_2 : ShardBlockReference = proposer_slashing.signed_reference_2.message
+    blob_1 = reference_1.signed_blob_reference.message
+    blob_2 = reference_2.signed_blob_reference.message
 
     # Verify header slots match
-    assert reference_1.slot == reference_2.slot
+    assert blob_1.slot == blob_2.slot
     # Verify header shards match
-    assert reference_1.shard == reference_2.shard
+    assert blob_1.shard == blob_2.shard
     # Verify header proposer indices match
     assert reference_1.proposer_index == reference_2.proposer_index
-    # Verify the headers are different (i.e. different body)
-    assert reference_1 != reference_2
+    # Verify the headers are different (i.e. different body, or different builder signature)
+    assert (
+            blob_1.body_root != blob_2.body_root 
+            or reference_1.signed_blob_reference.signature != reference_2.signed_blob_reference.signature
+    )
     # Verify the proposer is slashable
     proposer = state.validators[reference_1.proposer_index]
     assert is_slashable_validator(proposer, get_current_epoch(state))
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -34,15 +34,8 @@
   - [`ShardBlobHeader`](#shardblobheader)
   - [`SignedShardBlob`](#signedshardblob)
   - [`SignedShardBlobHeader`](#signedshardblobheader)
-  - [ShardBlock](#shardblock)
-  - [`ShardBlockHeader`](#shardblockheader)
-  - [`SignedShardBlock`](#signedshardblock)
-  - [`SignedShardBlockHeader`](#signedshardblockheader)
   - [`PendingShardHeader`](#pendingshardheader)
   - [`ShardBlobReference`](#shardblobreference)
-  - [`SignedShardBlobReference`](#signedshardblobreference)
-  - [`ShardBlockReference`](#shardblockreference)
-  - [`SignedShardBlockReference`](#signedshardblockreference)
   - [`ShardProposerSlashing`](#shardproposerslashing)
   - [`ShardWork`](#shardwork)
 - [Helper functions](#helper-functions)
@@ -116,8 +109,7 @@ The following values are (non-configurable) constants used throughout the specif
 
 | Name | Value |
 | - | - |
-| `DOMAIN_SHARD_PROPOSER` | `DomainType('0x80000000')` |
-| `DOMAIN_SHARD_BUILDER` | `DomainType('0x81000000')` |
+| `DOMAIN_SHARD_BLOB` | `DomainType('0x80000000')` |
 
 ### Shard Work Status
 
@@ -247,8 +239,6 @@ class ShardBlobBody(Container):
     data: List[BLSPoint, POINTS_PER_SAMPLE * MAX_SAMPLES_PER_BLOCK]
     # Latest block root of the Beacon Chain, before shard_blob.slot
     beacon_block_root: Root
-    # Builder of the data, pays data-fee to proposer
-    builder_index: BuilderIndex
     # TODO: fee payment amount fields (EIP 1559 like)
 ```
 
@@ -269,8 +259,6 @@ class ShardBlobBodySummary(Container):
     data_root: Root
     # Latest block root of the Beacon Chain, before shard_blob.slot
     beacon_block_root: Root
-    # Builder of the data, pays data-fee to proposer
-    builder_index: BuilderIndex
     # TODO: fee payment amount fields (EIP 1559 like)
 ```
 
@@ -282,18 +270,26 @@ class ShardBlobBodySummary(Container):
 class ShardBlob(Container):
     slot: Slot
     shard: Shard
+    # Proposer of the shard-blob
+    proposer_index: ValidatorIndex
+    # Builder of the data, pays data-fee to proposer
+    builder_index: BuilderIndex
     # Blob contents
     body: ShardBlobBody
 ```
 
 ### `ShardBlobHeader`
 
-Header version of `ShardBlob`. Separates designation (slot, shard) and contents (blob).
+Header version of `ShardBlob`.
 
 ```python
 class ShardBlobHeader(Container):
     slot: Slot
     shard: Shard
+    # Proposer of the shard-blob
+    proposer_index: ValidatorIndex
+    # Builder of the data, pays data-fee to proposer
+    builder_index: BuilderIndex
     # Blob contents, without the full data
     body_summary: ShardBlobBodySummary
 ```
@@ -316,53 +312,8 @@ Shard proposers can accept `SignedShardBlobHeader` as a data-transaction.
 ```python
 class SignedShardBlobHeader(Container):
     message: ShardBlobHeader
-    signature: BLSSignature
-```
-
-### ShardBlock
-
-Full blob data signed by builder, to be confirmed by proxy as `ShardBlockHeader`.
-
-```python
-class ShardBlock(Container):
-    # Shard data with fee payment by bundle builder
-    signed_blob: SignedShardBlob
-    # Proposer of the shard-blob
-    proposer_index: ValidatorIndex
-```
-
-### `ShardBlockHeader`
-
-Header version of `ShardBlock`, selecting a `SignedShardBlobHeader`.
-
-```python
-class ShardBlockHeader(Container):
-    # Shard commitments and fee payment by blob builder
-    signed_blob_header: SignedShardBlobHeader
-    # Proposer of the shard-blob
-    proposer_index: ValidatorIndex
-```
-
-### `SignedShardBlock`
-
-Shard blob, signed for payment, and signed for proposal. Propagated to attesters.
-
-```python
-class SignedShardBlock(Container):
-    message: ShardBlock
-    signature: BLSSignature
-```
-
-### `SignedShardBlockHeader`
-
-Header version of `SignedShardBlock`, substituting the full data within the blob for just the hash-tree-root.
-
-The signature is equally applicable to `SignedShardBlock`, 
-which the builder can publish as soon as the signed header is seen. 
-
-```python
-class SignedShardBlockHeader(Container):
-    message: ShardBlockHeader
+    # Signature by builder.
+    # Once accepted by proposer, the signatures is the aggregate of both.
     signature: BLSSignature
 ```
 
@@ -390,44 +341,26 @@ Reference version of `ShardBlobHeader`, substituting the body for just a hash-tr
 class ShardBlobReference(Container):
     slot: Slot
     shard: Shard
-    # Blob hash-tree-root for reference, enough for uniqueness
-    body_root: Root
-```
-
-### `SignedShardBlobReference`
-
-`ShardBlobReference`, signed by the blob builder. The builder-signature is part of the block identity.
-
-```python
-class SignedShardBlobReference(Container):
-    message: ShardBlobReference
-    signature: BLSSignature
-```
-
-### `ShardBlockReference`
-
-```python
-class ShardBlockReference(Container):
-    # Blob, minimized for efficient slashing
-    signed_blob_reference: SignedShardBlobReference
     # Proposer of the shard-blob
     proposer_index: ValidatorIndex
-```
-
-### `SignedShardBlockReference`
-
-```python
-class SignedShardBlockReference(Container):
-    message: ShardBlockReference
-    signature: BLSSignature
+    # Builder of the data
+    builder_index: BuilderIndex
+    # Blob hash-tree-root for slashing reference
+    body_root: Root
 ```
 
 ### `ShardProposerSlashing`
 
 ```python
 class ShardProposerSlashing(Container):
-    signed_reference_1: SignedShardBlockReference
-    signed_reference_2: SignedShardBlockReference
+    slot: Slot
+    shard: Shard
+    proposer_index: ValidatorIndex
+    builder_index: BuilderIndex
+    body_root_1: Root
+    body_root_2: Root
+    signature_1: BLSSignature
+    signature_2: BLSSignature
 ```
 
 ### `ShardWork`
@@ -552,7 +485,7 @@ def get_shard_proposer_index(beacon_state: BeaconState, slot: Slot, shard: Shard
     Return the proposer's index of shard block at ``slot``.
     """
     epoch = compute_epoch_at_slot(slot)
-    seed = hash(get_seed(beacon_state, epoch, DOMAIN_SHARD_PROPOSER) + uint_to_bytes(slot) + uint_to_bytes(shard))
+    seed = hash(get_seed(beacon_state, epoch, DOMAIN_SHARD_BLOB) + uint_to_bytes(slot) + uint_to_bytes(shard))
 
     # Proposer must have sufficient balance to pay for worst case fee burn
     EFFECTIVE_BALANCE_MAX_DOWNWARD_DEVIATION = (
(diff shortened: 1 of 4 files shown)


specs/sharding/beacon-chain.md
```diff
@@ -217,36 +271,110 @@ class ShardBlobBodySummary(Container):
     data_root: Root
     # Latest block root of the Beacon Chain, before shard_blob.slot
     beacon_block_root: Root
+    # Builder of the data, pays data-fee to proposer
+    builder_index: BuilderIndex
+    # TODO: fee payment amount fields (EIP 1559 like)
+```
+
```
Nashatyrev, line 278: BTW don't we want to inline 
`ShardBlobBody` -> `ShardBlob` 
`ShardBlobBodySummary` -> `ShardBlobHeader` 
I don't think they are used somewhere else, so may be it's worth to reduce the number of containers?

specs/sharding/beacon-chain.md
```diff
@@ -184,6 +213,9 @@ class BeaconState(merge.BeaconState):  # [extends The Merge state]
     previous_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
     current_epoch_attestations: List[PendingAttestation, MAX_ATTESTATIONS * SLOTS_PER_EPOCH]
     # [New fields]
+    # Builder registry.
+    builders: List[Builder, BUILDER_REGISTRY_LIMIT]
+    builder_balances: List[Gwei, BUILDER_REGISTRY_LIMIT]
```
Nashatyrev, line 218: Did I understand correctly that these fields are kind of temporary placeholders for now?
May be for simplicity of implementation and testing it makes sense to assign the `builder` role to validators only? 
I.e. only a `validator` _may_ become a `builder`. And thus 'publish fees' are paid between existing validators 
This is also a temporary draft solution of course

Nashatyrev (commented):
Great idea from my perspective. 
Just some minor comments

specs/sharding/beacon-chain.md
protolambda: A registry of builders is closer to my long-term idea than reuse of validators. It may be nice for a prototype, but I don't want to get stuck in a model with validators only. Both for privacy of validators, simplicity to become a builder, and separation of concerns w.r.t. balances, slashings and fees.

specs/sharding/beacon-chain.md
protolambda: I would, but collapsing the body is useful to reduce the size of a slashing. The `ShardBlobReference` just has a `body_root: Root` instead of a `body_summary: ShardBlobBodySummary`.

protolambda: Different idea to reduce types; we could try aggregate the builder and validator signatures, and specify one or the other is a 0 in aggregation when the builder_index or proposer_index is max_uint64

specs/sharding/beacon-chain.md
```diff
@@ -217,36 +271,110 @@ class ShardBlobBodySummary(Container):
...
+
 ```python
 class SignedShardBlobHeader(Container):
     message: ShardBlobHeader
     signature: BLSSignature
 ```
 
+### ShardBlock
+
+Full blob data signed by builder, to be confirmed by proxy as `ShardBlockHeader`.
```
terencechain, line 326:
```suggestion
Full blob data signed by builder, to be confirmed by proxy as `ShardBlock`.
```

I think you mean `ShardBlock` here?

protolambda:
Half time (on holiday), with a reset laptop on holiday, but started a nice simplification/improvement; aggregating the two signatures allows us to reduce the nesting and type count, and should be faster too. As data-tx, header are just signed by the builder only. So verification depends on topic. On-chain we always fast-aggregate-verify. And I reduced the types of the slashing by deduplicating data more. (TODO: missing case, if proposer accepts different builders for same slot/shard), that needs to be slashable too. So need to change to `builder_index_1` and `builder_index_2`)

specs/sharding/beacon-chain.md
Nashatyrev: Yeah, really nice solution 👍 
Curious whether we can make `ShardBlobReference` container 'private' to the `process_shard_proposer_slashing` function?

specs/sharding/beacon-chain.md
```diff
@@ -258,31 +386,50 @@ class PendingShardHeader(Container):
...
+```
+
 ### `ShardProposerSlashing`
 
 ```python
 class ShardProposerSlashing(Container):
-    signed_reference_1: SignedShardBlobReference
-    signed_reference_2: SignedShardBlobReference
+    signed_reference_1: SignedShardBlockReference
+    signed_reference_2: SignedShardBlockReference
```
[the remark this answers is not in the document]
Nashatyrev: Addressed with the last commit.

specs/sharding/beacon-chain.md
```diff
@@ -587,36 +649,54 @@ def update_pending_shard_work(state: BeaconState, attestation: Attestation) -> N
...
-    signing_root = compute_signing_root(header, get_domain(state, DOMAIN_SHARD_PROPOSER))
-    assert bls.Verify(state.validators[header.proposer_index].pubkey, signing_root, signed_header.signature)
+    # Verify proposer matches
+    assert header.proposer_index == get_shard_proposer_index(state, slot, shard)
+
+    # Verify builder and proposer aggregate signature
+    blob_signing_root = compute_signing_root(header, get_domain(state, DOMAIN_SHARD_BLOB))
+    builder_pubkey = state.builders[header.builder_index].pubkey
+    proposer_pubkey = state.validators[header.proposer_index].pubkey
+    assert bls.FastAggregateVerify([builder_pubkey, proposer_pubkey], blob_signing_root, signed_header.signature)
```
djrtwo, line 699: Clever use of aggregation

specs/sharding/beacon-chain.md
```diff
@@ -587,36 +649,54 @@ def update_pending_shard_work(state: BeaconState, attestation: Attestation) -> N
...
+
+#### `charge_builder`
+
+```python
+def charge_builder(state: BeaconState, index: BuilderIndex, fee: Gwei) -> None:
+    """
+    Decrease the builder balance at index ``index`` by ``fee``, with underflow check.
+    """
+    assert state.builder_balances[index] >= fee
+    state.builder_balances[index] -= fee
```
djrtwo, line 661: I think a single balance for a builder can cause issues for shard proposers. 

take the following:
* `builder_balance == 100`
* `builder` creates `blob` with fee `10`
* `proposer` signs `blob`
* `builder` has `secret_blob` with fee `100`
* `builder` gets `secret_proposer` to sign `secret_blob`
* `secret_blob` and `blob` now cannot both be included. `secret_blob` is included first, thus invalidating honest `blob`

specs/sharding/beacon-chain.md
```diff
@@ -627,12 +707,19 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
         == bls.Pairing(body_summary.commitment.point, G2_SETUP[-body_summary.commitment.length])
     )
 
+    # Charge builder, with hard balance requirement
+    fee = Gwei(123)  # TODO EIP 1559 like fee? Burn some of it?
```
djrtwo, line 711: Builder will need to be able to set a fee inside of the `ShardBlobHeader`. It must at least have the 1559 fee for burn and any excess goes to the proposer

djrtwo: and if blob isn't confirmed, just a fraction should go to proposer

djrtwo (commented):
I like the structure in general! Things to work through on the fee mechanics and builder balances. 

Also want to debate whether Builders should be in the beacon chain instead of somewhere in the EVM. Definitely really nice on the way the message can be reused and aggregated...

Didn't do a deep review for any nits but just high level so we can continue to discuss

specs/sharding/beacon-chain.md
```diff
@@ -627,12 +707,19 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
...
+    charge_builder(state, header.builder_index, fee)
+    # TODO: proposer is charged for confirmed headers (see charge_confirmed_shard_fees).
+    #  Need to align incentive, so proposer does not gain from including unconfirmed headers
+    increase_balance(state, blob_header.proposer_index, fee)
```
Nashatyrev, line 715:
```suggestion
    increase_balance(state, header.proposer_index, fee)
```

specs/sharding/beacon-chain.md
```diff
@@ -652,27 +739,35 @@ The goal is to ensure that a proof can only be constructed if `deg(B) < l` (ther
...
-    # Verify header slots match
-    assert reference_1.slot == reference_2.slot
-    # Verify header shards match
-    assert reference_1.shard == reference_2.shard
-    # Verify header proposer indices match
-    assert reference_1.proposer_index == reference_2.proposer_index
-    # Verify the headers are different (i.e. different body)
-    assert reference_1 != reference_2
+    # Verify the headers are different
+    assert proposer_slashing.body_root_1 != proposer_slashing.body_root_2
```
Nashatyrev, line 743: Body roots can be equal but `builder_index_1 != builder_index_2`. I would suggest to 
```py
     assert reference_1 != reference_2
```
(if this is correct comparison in Py)

specs/sharding/p2p-interface.md
```diff
@@ -118,58 +75,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+- _[REJECT]_ The blob builder exists and has sufficient balance to back the fee payment.
+- _[IGNORE]_ The header fee SHOULD be higher than previously seen headers for `(header.slot, header.shard)`, from any builder.
+  Propagating nodes MAY increase fee increments in case of spam.
+- _[REJECT]_ The header signature is valid for ONLY the builder, `signed_blob_header.signature`,
+  i.e. `bls.Verify(builder_pubkey, blob_signing_root, signed_blob_header.signature)`. The signature is not an aggregate with the proposer.
+- _[REJECT]_ The header is designated for proposal by the expected `proposer_index` for the blob's `header.slot` and `header.shard`
+  in the context of the current shuffling (defined by `header.body_summary.beacon_block_root`/`slot`).
+  If the `proposer_index` cannot immediately be verified against the expected shuffling,
+  the blob MAY be queued for later processing while proposers for the blob's branch are calculated --
+  in such a case _do not_ `REJECT`, instead `IGNORE` this message.
```
Nashatyrev, line 166: BTW do we actually need `proposer_index` field at all (in `ShardBlob` and `ShardBlobHeader`)? It is derivable from `(beacon_block_root, slot, shard)`
It seems no much sense for a _builder_ to calculate the designated proposer index when creating a shard tx - he shouldn't actually care who would propose his blob.

protolambda: The builder already commits the transaction to a specific shard/slot. And if the builder does not win, the blob is very likely stale, and shouldn't be applied anyways later on. Computing the proposer index as builder should not be relatively much overhead.

Meanwhile, just like in beacon-blocks, having the proposer index in the signed message is useful for various things (easy slashing mechanism, optimized p2p validation order, etc.). It may change if we get single-secret-leader election, and builders don't know the shard proposers, but until then it seems more optimal to include it.

specs/sharding/beacon-chain.md
protolambda: Yes, the assumption is that there is some minimum balance of a builder that we'll need to enforce to avoid race conditions in accepting fees. But with arbitrary priority fees, that may not be so simple. Requiring the builder to have 64 times the fee worth of balance may work, but is inefficient. We'll decide on a better approach later.

protolambda:
Thanks for reviews/suggestions so far, here's an update:
- Likely won't implement builder deposits/liquidity yet, we can embed the pubkeys into devnet genesis and test the rest of the system first. Some discussion about this:
  - Capacity: With N builders per L2 type (and some doing multiple maybe), it's likely no more than 10k entities actively building. At 2700 beacon blocks per day, there's probably enough space to manage basic fund-management transactions for fee payments, per builder, every other week or so.
  - Deposits: if it were a separate balance system, we need deposits of some kind. A special withdrawal credential and signature domain could be used and special-cased in the deposit processing, to create `Builder`s instead of `Validator`s, without changing the deposit-tracking/voting/etc.
  - Withdrawals, exits: undecided, need to test the system first anyway.
- Builders with BLS keys seem best to me;
  - Aggregate with proposer signature, reducing the verification work to a single signature
  - L2s can distribute their role as builder
- Builder balances and payments might be more convenient when directly changing balances in the execution state. But it breaks the separation, makes it synchronous, and might complicate management/introduction of builder pubkeys more. So we'll stick to a minimal registry of builder balances.
- Need to take some time to list all incentives + payments + suggestions, then implement it properly. Mostly placeholders/TODOs for now.
- Builders and Validators are not the same thing in my opinion, because:
  - Different software stacks running at the same time, avoid signing key living in both places
  - Different life-cycles, avoid mixing up entry/exit concerns
  - Different privacy and network models. Builders stay on shards, there may be less, they publish different messages, and they are not interested in all the beacon-chain work like a validator.
- Unsure about naming. I realize "Builder" is too generic, but I do like single-word names like "Validator". Renamed the state fields to mention `blob_builder` instead. More hesitant about the rest. If the type or context is clear, then keep it short and simple (Go naming convention, but Java is different and has its own benefits).

For now I think we can focus on the minimal networking + validator API (accept shard blob data-tx by signing the header) + placeholder fee processing, and then work on the remaining details later. Big thanks to @Nashatyrev for pushing this prototype forward with Teku.

protolambda:
Rebased on latest dev (Altair beta.1 + merge based on Altair + shard-start-slot simplification). Implementing incentives/rewards/etc. in a separate PR next, but might merge into this one if it's still simple enough to review as a whole.

protolambda:
At first I thought adjusting the `shard_sample_price` per shard would be fine, but there are several issues:
- Beacon proposer can-reorder headers, to mess with capacity interactions between them
- Beacon proposer has more incentive to omit a specific header if it decreases base-fee cost of one of the 64 builders they happen to know.
- 64 shard headers don't process well in parallel, the fee processing part becomes a problem
- Late shard-headers leave 0 capacity, but might be included a slot later

Yet, I do not like the idea of processing it at the end of the epoch, or processing it only for confirmed headers:
- Network processing capacity can very well be taken by chaotic publishing, multiple competing headers, late headers, etc. This should not be "free" for unconfirmed headers.
- Adjusting it once per epoch is not frequent enough, desired capacity may change per slot

So instead, I think we should sum the taken capacity of all included headers of a slot, and adjust base-fee to target a per-slot number of samples.

Additionally, we could buffer capacity usage for early headers (`shard_header.slot == state.slot`), and apply it to the sample price with a discount in the next slot, to incentivize faster inclusion (proposers get additional priority rewards by adjusting early headers, builders burn less)

specs/sharding/beacon-chain.md
```diff
@@ -370,7 +412,7 @@ class ShardWork(Container):
     # Upon confirmation the data is reduced to just the commitment.
     status: Union[                                                   # See Shard Work Status enum
               None,                                                  # SHARD_WORK_UNCONFIRMED
-              DataCommitment,                                        # SHARD_WORK_CONFIRMED
+              ConfirmedDataCommitment,                               # SHARD_WORK_CONFIRMED
```
Nashatyrev, line 415: `AttestedDataCommitment` was meant here?

specs/sharding/beacon-chain.md
```diff
@@ -486,18 +512,8 @@ def get_shard_proposer_index(beacon_state: BeaconState, slot: Slot, shard: Shard
...
-
-    # Proposer must have sufficient balance to pay for worst case fee burn
-    EFFECTIVE_BALANCE_MAX_DOWNWARD_DEVIATION = (
-        EFFECTIVE_BALANCE_INCREMENT - EFFECTIVE_BALANCE_INCREMENT
-        * HYSTERESIS_DOWNWARD_MULTIPLIER // HYSTERESIS_QUOTIENT
-    )
-    min_effective_balance = (
-        beacon_state.shard_gasprice * MAX_SAMPLES_PER_BLOB // TARGET_SAMPLES_PER_BLOB
-        + EFFECTIVE_BALANCE_MAX_DOWNWARD_DEVIATION
-    )
```
Nashatyrev, line 437: Why don't we check the balance any more?
Intuitively an active validator should have enough balance to pay for blob with any 'reasonable' price, but anyways that check was originally here for some reason

protolambda: We should be checking the builder balance (TODO), not the shard proposer balance (they don't burn with their own balance, the builder does). Nice side-effect: any active validator can be a shard-proposer, no exceptions.

specs/sharding/beacon-chain.md
```diff
@@ -673,7 +690,16 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
...
+    start_shard = get_start_shard(state, slot)
+    committees_per_slot = get_committee_count_per_slot(state, header_epoch)
+    end_shard = start_shard + committees_per_slot
+    shard_count = get_active_shard_count(state, header_epoch)
+    # Per slot, there may be max. shard_count committees.
+    # If there are shard_count * 2/3 per slot, then wrap around.
+    if end_shard >= shard_count:
+        assert not (end_shard - shard_count <= shard < start_shard)
+    else:
+        assert start_shard <= shard < end_shard
```
Nashatyrev: May be it makes sense to add a helper function `is_shard_active_at_slot()` to make the logic a bit less obfuscated?
I believe it would also be handy in other places (like Gossip messages validation)

specs/sharding/beacon-chain.md
```diff
@@ -705,20 +731,46 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
...
-    #  Need to align incentive, so proposer does not gain from including unconfirmed headers
-    increase_balance(state, header.proposer_index, fee)
+    # Charge EIP 1559 fee, builder pays for opportunity, and is responsible for later availability,
+    # or fail to publish at their own expense.
+    samples = blob_summary.commitment.length
+    # TODO: overflows, need bigger int type
+    max_fee = blob_summary.max_fee_per_sample * samples
+
+    # Builder must have sufficient balance, even if max_fee is not completely utilized
+    assert state.blob_builder_balances[header.builder_index] > max_fee
```
Nashatyrev, line 740: Shouldn't it be `>=` instead of `>` here?

specs/sharding/beacon-chain.md
```diff
@@ -705,20 +731,46 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
...
+    base_fee = state.shard_sample_price * samples
+    # Base fee must be paid
+    assert max_fee >= base_fee
+
+    # Remaining fee goes towards proposer for prioritizing, up to a maximum
+    max_priority_fee = blob_summary.max_priority_fee_per_sample * samples
+    priority_fee = min(max_fee - base_fee, max_priority_fee)
+
+    # Burn base fee, take priority fee
+    decrease_balance(state, header.builder_index, base_fee + priority_fee)
```
Nashatyrev, line 751: Should it be something like `decrease_builder_balance()` ? Else it would decrease validator (not a builder) balance I guess

specs/sharding/beacon-chain.md
```diff
@@ -258,7 +299,10 @@ class ShardBlobBodySummary(Container):
     data_root: Root
     # Latest block root of the Beacon Chain, before shard_blob.slot
     beacon_block_root: Root
-    # TODO: fee payment amount fields (EIP 1559 like)
+    # fee payment fields (EIP 1559 like)
+    # TODO: express in MWei instead?
+    max_priority_fee_per_sample: Gwei
+    max_fee_per_sample: Gwei
```
Nashatyrev, line 305: Wouldn't it be more clear to define?
- `priority_fee_per_sample`: kind of fixed priority bid 
- `max_base_fee_per_sample`: base fee is what a builder is uncertain about and thus setting an upper bound

specs/sharding/beacon-chain.md
```diff
@@ -223,6 +249,18 @@ class DataCommitment(Container):
...
+### `AttestedDataCommitment`
+
+```python
+class AttestedDataCommitment(Container):
+    # KZG10 commitment to the data, and length
+    commitment: DataCommitment
+    # hash_tree_root of the ShardBlobHeader (stored so that attestations can be checked against it)
+    root: Root
+    # The proposer who included the shard-header
+    includer_index: ValidatorIndex
```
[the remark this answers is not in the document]
protolambda: Not all incentives have been implemented yet. Rewarding the beacon-proposer for including shard-headers is one of the things. Tracking them as the includer is the first step, then we can reward them upon confirmation of the header (TODO).

specs/sharding/beacon-chain.md
protolambda: This was cleaned up already, please review the latest set of changes.

specs/sharding/beacon-chain.md
protolambda: Good catch

specs/sharding/beacon-chain.md
protolambda: Good catch

specs/sharding/beacon-chain.md
protolambda: I'm following the EIP-1559 spec naming here, and the priority fee is dynamic, i.e. a proposer can choose to select a transaction that pays some of the priority fee towards the base fee instead, to be able to get it included.

protolambda:
TODO:
- [ ] reward header includer (beacon proposer) when header is confirmed
  - add to epoch processing: for each slot in previous epoch, check if union selector equals `SHARD_WORK_CONFIRMED`, and reward the includer_index in the `AttestedDataCommitment`
- [ ] reward less to shard-proposer if header is not confirmed (suggestion by Danny)
  - Currently we reward the priority fee immediately. Instead, we could burn it along with the base-fee, and reward to the shard-proposer upon epoch-processing of previous epoch (along with other TODO above)
- [ ] Clarify attesters get rewarded for attesting the confirmed shard-data timely
  - "timely": confirmation was part of the attestation-processing, not part of the very last moment during rotation of previous-epoch data (`process_pending_shard_confirmations`)
  - the Altair `WEIGHT_DENOMINATOR` needs adjustment, but in a compatible way
  - extending `PARTICIPATION_FLAG_WEIGHTS` is too silent, this changes behavior of Altair (in compatible way). Also, we don't override constants in any fork, so modifying it is a bad pattern.
- [ ] Fix attack vector from builders to build multiple blobs in parallel, without sufficient balance for priority fees. Possible solutions:
  - 1 builder balance per shard (problematic with transactions, liquidity, state size)
  - Enforce builder balance buffer to cover N shards (does not cover super high priority fees though)
  - slashable builders, if building on multiple shards at the same time? (introduces builder entry/exit queue problem)

Other questions (for future PRs likely):
- Builder entry/exit options. Just baked into genesis state for devnets for now. But some ideas:
  - re-use the existing deposit mechanism, with special withdrawal credential, to indicate it's a builder deposit instead of validator deposit. And similarly for withdrawals.
  - introduce new transaction type for beacon-proposers to create builders with their balance? There could be an eth1 contract that buffers funds, and tops up validators with deposits if supplied with a proof that the validator created a builder with certain credentials.
  - wait for EVM beacon-chain integration to mature, make receipts accessible to the beacon-chain block processing, and implement builder-deposits/withdrawals without new SSZ transaction type? (drawback: one-directional synchrony between execution and consensus layer during sync)

specs/sharding/beacon-chain.md
```diff
@@ -69,6 +75,12 @@ This document describes the extensions made to the Phase 0 design of The Beacon
 based on the ideas [here](https://hackmd.io/G-Iy5jqyT7CXWEz8Ssos8g) and more broadly [here](https://arxiv.org/abs/1809.09044),
 using KZG10 commitments to commit to data to remove any need for fraud proofs (and hence, safety-critical synchrony assumptions) in the design.
 
+### Glossary
+
+- **Data**: A list of KZG points, to translate a byte string into
```
terencechain, line 80: What about **Points**?  To me, data sounds more like the `byte string`

```python
class ShardBlobBody(Container):
    # The KZG points. Should match the commitment and degree proof.
    points: List[BLSPoint, POINTS_PER_SAMPLE * MAX_SAMPLES_PER_BLOB]

File: specs/sharding/beacon-chain.md [hunk 19]
```diff
@@ -500,45 +579,67 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+
     buffer_index = attestation.data.slot % SHARD_STATE_MEMORY_SLOTS
     committee_work = state.shard_buffer[buffer_index][attestation_shard]
 
     # Skip attestation vote accounting if the header is not pending
     if committee_work.status.selector != SHARD_WORK_PENDING:
-        # TODO In Altair: set participation bit flag, if attestation matches winning header.
+        # If the data was already confirmed, check if this matches, to apply the flag to the attesters.
+        if committee_work.status.selector == SHARD_WORK_CONFIRMED:
+            attested: AttestedDataCommitment = current_headers[header_index]
```
terencechain, line 621: Not sure about python closure, do we need to move the following before?

```python
current_headers: Sequence[PendingShardHeader] = committee_work.status.value
```

protolambda: Thanks. It should have just directly retrieved the value. The list of PendingShardHeader is the value when the selector is `SHARD_WORK_PENDING`, this was a copy mistake.

specs/sharding/beacon-chain.md
```diff
@@ -613,13 +730,39 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
...
+    samples = blob_summary.commitment.length
+    # TODO: overflows, need bigger int type
+    max_fee = blob_summary.max_fee_per_sample * samples
+
+    # Builder must have sufficient balance, even if max_fee is not completely utilized
+    assert state.blob_builder_balances[header.builder_index] >= max_fee
+
+    base_fee = state.shard_sample_price * samples
+    # Base fee must be paid
+    assert max_fee >= base_fee
```
terencechain, line 744: I wonder if a more neutral assertion is:
```python
 assert blob_summary.max_fee_per_sample >= state.shard_sample_price
```
Basically nothing will work unless builder pays at least the sample price in state. Can also move such assertion a few lines up too

djrtwo (commented):
nice!

need to review network spec in the morning. Some additional high level thoughts

> e.g. split in shard_count / 4 subnets. Trade-off between proposer subscription management and gossip topic load. But greatly reduced compared to propagating full shard-blobs for a day for a single proposal.

Is the idea to divide validator set by `4` (`num_shard_committees`) and shuffle these every once in a while so that they can listen to fewer global channels? I suppose with more than one val you are quickly listening to all these channels if you have more than a few vals

> Builder is not encumbered with the same validator network privacy concerns

has some but not as much. could imagine dos-ing builders to manipulate fee market and things

> builder registry

very worried about dust/account accumulation leading to state bloat. much cheaper than validators to secure a slot and to rotate in and out. Is this issue what you meant by expiry?

specs/sharding/beacon-chain.md
```diff
@@ -108,45 +120,70 @@ The following values are (non-configurable) constants used throughout the specif
...
-| `ROOT_OF_UNITY` | `pow(PRIMITIVE_ROOT_OF_UNITY, (MODULUS - 1) // int(MAX_SAMPLES_PER_BLOCK * POINTS_PER_SAMPLE), MODULUS)` |
+| `ROOT_OF_UNITY` | `pow(PRIMITIVE_ROOT_OF_UNITY, (MODULUS - 1) // int(MAX_SAMPLES_PER_BLOB * POINTS_PER_SAMPLE), MODULUS)` |
 
 ### Gwei values
 
 | Name | Value | Unit | Description |
 | - | - | - | - |
-| `MAX_GASPRICE` | `Gwei(2**33)` (= 8,589,934,592) | Gwei | Max gasprice charged for a TARGET-sized shard block |  
-| `MIN_GASPRICE` | `Gwei(2**3)` (= 8) | Gwei | Min gasprice charged for a TARGET-sized shard block |
+| `MAX_SAMPLE_PRICE` | `Gwei(2**33)` (= 8,589,934,592) | Gwei | Max sample charged for a TARGET-sized shard blob |
```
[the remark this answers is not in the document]
protolambda: Yes, I think "gas" is confusing when it applies to data samples.

specs/sharding/beacon-chain.md
```diff
@@ -179,19 +216,24 @@ class BeaconBlockBody(merge.BeaconBlockBody):  # [extends The Merge block body]
...
 ## New containers
 
-The shard data itself is network-layer only, and can be found in the [P2P specification](./p2p-interface.md).
-The beacon chain registers just the commitments of the shard data.
+### `Builder`
+
+```python
+class Builder(Container):
+    pubkey: BLSPubkey
+    # TODO: fields for either an expiry mechanism (refunding execution account with remaining balance)
```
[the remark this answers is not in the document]
protolambda: One idea to avoid builder withdrawal transactions (which come with signing and inclusion overhead) is to put an expiry date on an account, to clean up passively. Either a fixed epoch when the builder gets withdrawn to execution state, or we can implement some inactivity check (withdraw funds if not used for N epochs, maybe specified by the builder).

That said, for initial sharding prototypes we don't need deposits or withdrawals, we can just pre-mine the builders into the testnet genesis state. Implementing a builder lifecycle (via EVM, via beacon state, or some buffer kind of thing) can go into a later PR.

specs/sharding/beacon-chain.md
```diff
@@ -215,37 +295,73 @@ class ShardBlobBodySummary(Container):
...
 ### `SignedShardBlobHeader`
 
+Header of the blob, the signature is equally applicable to `SignedShardBlob`.
+Shard proposers can accept `SignedShardBlobHeader` as a data-transaction by co-signing the header.
+
 ```python
 class SignedShardBlobHeader(Container):
     message: ShardBlobHeader
+    # Signature by builder.
+    # Once accepted by proposer, the signatures is the aggregate of both.
```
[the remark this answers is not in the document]
protolambda: > Builder sends these out? 

Yes, first signed by only the builder, then later by both. The proposer can simply sign the message and aggregate the two signatures. The verification depends on the topic the message is received on;
- data-tx channel(s) (messages prioritized by fee): only builders sign
- headers global channel (messages selected by proposer, double signing as proposer is slashable): both sign

> They know the proposer_index deterministically before?

Yes. Although not strictly necessary, this allows them to both sign the same message, and include the data in the signature, to make verification easy, just like with beacon-blocks. Once verified, the proposer index can be checked against the state later. And it also simplifies slashings.

Once we get Single Secret Leader Election we would want to remove it, but the same applies to beacon-blocks, so I think we can revisit for both types when SSLE is practical.

> Is this (not yet signed by proposer) a "data-tx"?

Yes.

specs/sharding/beacon-chain.md
```diff
@@ -256,41 +372,43 @@ class PendingShardHeader(Container):
...
 ### `ShardProposerSlashing`
 
 ```python
 class ShardProposerSlashing(Container):
-    signed_reference_1: SignedShardBlobReference
-    signed_reference_2: SignedShardBlobReference
+    slot: Slot
+    shard: Shard
+    proposer_index: ValidatorIndex
+    builder_index_1: BuilderIndex
```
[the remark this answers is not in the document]
protolambda: Yes, good point. Alternatively we put the full builder pubkey in the slashing, but I like how lightweight it is (easy to cover many misbehaving shards with single beacon block). Builders are not slashed, so just the pubkey for verification of the aggregate signature would be sufficient. Is the 40 byte difference worth less builder account cleanup concern? Depending on the builder cleanup approch, possibly?

specs/sharding/beacon-chain.md
```diff
@@ -423,19 +508,9 @@ def get_shard_proposer_index(beacon_state: BeaconState, slot: Slot, shard: Shard
...
-        * HYSTERESIS_DOWNWARD_MULTIPLIER // HYSTERESIS_QUOTIENT
-    )
-    min_effective_balance = (
-        beacon_state.shard_gasprice * MAX_SAMPLES_PER_BLOCK // TARGET_SAMPLES_PER_BLOCK
-        + EFFECTIVE_BALANCE_MAX_DOWNWARD_DEVIATION
-    )
-    return compute_proposer_index(beacon_state, committee, seed, min_effective_balance)
+    seed = hash(get_seed(beacon_state, epoch, DOMAIN_SHARD_BLOB) + uint_to_bytes(slot) + uint_to_bytes(shard))
+    indices = get_active_validator_indices(state, epoch)
+    return compute_proposer_index(beacon_state, indices, seed)
```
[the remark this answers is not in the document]
protolambda: Especially if the data-tx topic is lightweight and global, as proposed now, it seems fine to me.

If we want different indices, we could use multiple index-wise shuffles for different list indices but same seed, and implement an alternative for the effective-balance bias that is currently applied to proposer computation. Or repeat the computation with a different seed if the proposer is already assigned to a shard (reuses existing code, but determining a role of a single validator then depends on others). I think I prefer the current status, and with mainnet validator counts it will be very rare anyway.

specs/sharding/beacon-chain.md
```diff
@@ -500,45 +575,67 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+                batch_apply_participation_flag(state, attestation.aggregation_bits,
+                                               attestation.data.target.epoch,
+                                               full_committee, TIMELY_SHARD_FLAG_INDEX)
         return
 
     current_headers: Sequence[PendingShardHeader] = committee_work.status.value
 
     # Find the corresponding header, abort if it cannot be found
-    header_index = [header.root for header in current_headers].index(attestation.data.shard_header_root)
+    header_index = len(current_headers)
```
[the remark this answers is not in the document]
protolambda: `index()` would abort with an exception if the data was not found, which is the kind of silent edge case I would like to avoid (and it needs to return without exception here). And instead of creating temporary lists, we can just implement the behavior directly (readable and fast). Might just be me thinking in Go style too much though.

specs/sharding/beacon-chain.md
```diff
@@ -613,13 +726,39 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
...
+    # Base fee must be paid
+    assert max_fee >= base_fee
+
+    # Remaining fee goes towards proposer for prioritizing, up to a maximum
+    max_priority_fee = body_summary.max_priority_fee_per_sample * samples
+    priority_fee = min(max_fee - base_fee, max_priority_fee)
+
+    # Burn base fee, take priority fee
+    # priority_fee <= max_fee - base_fee, thus priority_fee + base_fee <= max_fee, thus sufficient balance.
+    state.blob_builder_balances[header.builder_index] -= base_fee + priority_fee
```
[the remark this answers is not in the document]
protolambda: We already did the balance checks here anyway, and currently this is the only builder balance mutation, so I prefer to just decrease it directly. We can add builder-variants of the validator balance helper functions once builders have a lifecycle & clear relation to execution state.

protolambda:
@djrtwo 

> >   e.g. split in shard_count / 4 subnets. Trade-off between proposer subscription management and gossip topic load. But greatly reduced compared to propagating full shard-blobs for a day for a single proposal.
>
> Is the idea to divide validator set by 4 (num_shard_committees) and shuffle these every once in a while so that they can listen to fewer global channels? I suppose with more than one val you are quickly listening to all these channels if you have more than a few vals

Global or a split in a few topics both are a reduction compared to the full-blob workload for a shard proposer. We could still try split the global channel, but you're right that with more than 1 val, you quickly listen to all of them anyway. Let's start with a global channel, and see from there, these messages are small anyway.

>  >    Builder is not encumbered with the same validator network privacy concerns
>
> has some but not as much. could imagine dos-ing builders to manipulate fee market and things

Note that double builder messages are not slashable, they pay a single fee anyway (assuming the shard proposer doesn't try confirm both and become slashable). Setting up redundant builders is much easier than setting up redundant proposers.

Also, with BLS + some coordination, some builder nodes may choose to collectively sign and publish blobs (especially interesting if enshrined in some L2 as collective publishing task), again without slashing concern.

So I don't think builder DoS is more accessible than validator DoS to bad actors that would want to manipulate things like the fee market.

specs/sharding/beacon-chain.md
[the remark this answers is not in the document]
djrtwo: yeah for sure. don't need to solve the nitty gritty yet

specs/sharding/beacon-chain.md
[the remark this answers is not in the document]
djrtwo:
```python
if not shard_blob_root in current_header_roots:
    return

header_index = current_header_roots.index(shard_blob_root)
```

This has a check above to ensure the item is certainly in the list when call to `index()`.

This does feel a bit off wrt previous style, but we can wait until the great (inevitable) cleanups come through

djrtwo:
> Setting up redundant builders is much easier than setting up redundant proposers.

right!

protolambda:
Visual overview of the different shard blob related types

sharding

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
-- _[REJECT]_ The shard should have a committee at slot --
+- _[REJECT]_ The shard blob is for an active shard --
+  i.e. `blob.shard < get_active_shard_count(state, compute_epoch_at_slot(blob.slot))`
+- _[REJECT]_ The `blob.shard` MUST have a committee at the `blob.slot` --
   i.e. validate that `compute_committee_index_from_shard(state, blob.slot, blob.shard)` doesn't raise an error
 - _[REJECT]_ The shard blob is for the correct subnet --
   i.e. `compute_subnet_for_shard_blob(state, blob.slot, blob.shard) == subnet_id`
 - _[IGNORE]_ The blob is the first blob with valid signature received for the `(blob.proposer_index, blob.slot, blob.shard)` combination.
-- _[REJECT]_ As already limited by the SSZ list-limit, it is important the blob is well-formatted and not too large.
+- _[REJECT]_ The blob is not too large, the data MUST NOT be larger than the SSZ list-limit, and a client MAY be more strict.
```
djrtwo, line 92:
```suggestion
- _[REJECT]_ The blob is not too large -- the data MUST NOT be larger than the SSZ list-limit, and a client MAY apply stricter bounds.
```
```suggestion
- _[REJECT]_ The blob is not too large, the `blob.body.data` MUST NOT be larger than the SSZ list-limit, and a client MAY be more strict.
```

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
 - _[REJECT]_ The `blob.body.data` MUST NOT contain any point `p >= MODULUS`. Although it is a `uint256`, not the full 256 bit range is valid.
-- _[REJECT]_ The proposer signature, `signed_blob.signature`, is valid with respect to the `proposer_index` pubkey.
-- _[REJECT]_ The blob is proposed by the expected `proposer_index` for the blob's slot
+- _[REJECT]_ The blob builder exists and has sufficient balance to back the fee payment.
```
djrtwo, line 94:
```suggestion
- _[REJECT]_ The blob builder defined by `blob.builder_index` exists and has sufficient balance to back the fee payment.
```

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+- _[REJECT]_ The blob signature is valid for the aggregate of proposer and builder, `signed_blob.signature`,
```
djrtwo, line 95:
```suggestion
- _[REJECT]_ The blob signature, `signed_blob.signature`, is valid for the aggregate of proposer and builder --
```

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
   in such a case _do not_ `REJECT`, instead `IGNORE` this message.
 
 #### Global topics
 
-There are two additional global topics for Sharding, one is used to propagate shard blob headers (`shard_header`) to
-all nodes on the network. Another one is used to propagate validator message (`shard_proposer_slashing`).
+There are three additional global topics for Sharding.
 
-##### `shard_header`
+- `shard_blob_header`: co-signed headers, to be included on-chain, and signaling builders to publish full data.
```
djrtwo, line 107:
```suggestion
- `shard_blob_header`: co-signed headers to be included on-chain and to serve as a signal to the builder to publish full data.
```

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+- `shard_blob_tx`: builder-signed headers, also known as "data transaction".
+- `shard_proposer_slashing`: slashings of duplicate shard proposals
```
djrtwo, line 109:
```suggestion
- `shard_proposer_slashing`: slashings of duplicate shard proposals.
```

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
 
-Shard header data, in the form of a `SignedShardBlobHeader` is published to the global `shard_header` subnet.
+##### `shard_blob_header`
+
+Shard header data, in the form of a `SignedShardBlobHeader` is published to the global `shard_blob_header` subnet.
+Shard blob headers select shard blob bids by builders,
```
djrtwo, line 114:
```suggestion
Shard blob headers select shard blob bids by builders
```

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+and should be timely to ensure builders can publish the full shard blob before subsequent attestations.
+
+The following validations MUST pass before forwarding the `signed_blob_header` on the network. Alias `header = signed_blob_header.message`
```
djrtwo, line 117:
```suggestion
The following validations MUST pass before forwarding the `signed_blob_header` on the network. Alias `header = signed_blob_header.message`.
```

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
-- _[IGNORE]_ The header is the first header with valid signature received for the `(header.proposer_index, header.slot, header.shard)` combination.
-- _[REJECT]_ The shard should have a committee at slot --
+- _[REJECT]_ The shard header is for an active shard --
+  i.e. `header.shard < get_active_shard_count(state, compute_epoch_at_slot(header.slot))`
+- _[REJECT]_ The `header.shard` MUST have a committee at the `header.slot` --
   i.e. validate that `compute_committee_index_from_shard(state, header.slot, header.shard)` doesn't raise an error
-- _[REJECT]_ The proposer signature, `signed_shard_blob_header.signature`, is valid with respect to the `proposer_index` pubkey.
-- _[REJECT]_ The header is proposed by the expected `proposer_index` for the block's slot
+- _[IGNORE]_ The header is the first header with valid signature received for the `(header.proposer_index, header.slot, header.shard)` combination.
+- _[REJECT]_ The blob builder exists and has sufficient balance to back the fee payment.
```
djrtwo, line 129:
```suggestion
- _[REJECT]_ The blob builder defined by `blob.builder_index` exists and has sufficient balance to back the fee payment.
```

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+- _[REJECT]_ The header signature is valid for the aggregate of proposer and builder, `signed_blob_header.signature`,
```
djrtwo, line 130:
```suggestion
- _[REJECT]_ The header signature, `signed_blob_header.signature`, is valid for the aggregate of proposer and builder --
```

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+- _[REJECT]_ The header is proposed by the expected `proposer_index` for the blob's `header.slot` and `header.shard`
   in the context of the current shuffling (defined by `header.body_summary.beacon_block_root`/`slot`).
   If the `proposer_index` cannot immediately be verified against the expected shuffling,
-  the block MAY be queued for later processing while proposers for the block's branch are calculated --
+  the blob MAY be queued for later processing while proposers for the blob's branch are calculated --
   in such a case _do not_ `REJECT`, instead `IGNORE` this message.
 
+##### `shard_blob_tx`
+
+Shard data-transactions, in the form of a `SignedShardBlobHeader` is published to the global `shard_blob_tx` subnet.
```
djrtwo, line 140:
```suggestion
Shard data-transactions in the form of a `SignedShardBlobHeader` are published to the global `shard_blob_tx` subnet.
```

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+These shard blob headers are signed solely by the blob-builder.
+
+The following validations MUST pass before forwarding the `signed_blob_header` on the network. Alias `header = signed_blob_header.message`
```
djrtwo, line 143:
```suggestion
The following validations MUST pass before forwarding the `signed_blob_header` on the network. Alias `header = signed_blob_header.message`.
```

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+
+- _[IGNORE]_ The `header` is not from a future slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) --
```
djrtwo, line 145:
```suggestion
- _[IGNORE]_ The header is not from a future slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) --
```

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+  i.e. validate that `header.slot <= current_slot`
```
djrtwo, line 146: The writing of this condition seems a bit too strong. Normally, I'm going to broadcast for inclusion in slot `N` in slot `N-1` (or earlier) so *almost always* it will be for a future slot

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+  (a client MAY queue future headers for processing at the appropriate slot).
+- _[IGNORE]_ The header is new enough to still be processed --
+  i.e. validate that `compute_epoch_at_slot(header.slot) >= get_previous_epoch(state)`
```
djrtwo, line 149: Whereas, this condition seems a bit too lax! things even a couple of slots ago are no longer valuable at all! unless I'm doing some sort of re-org attack and re-packing the chain..

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+##### `shard_blob_tx`
+
+Shard data-transactions, in the form of a `SignedShardBlobHeader` is published to the global `shard_blob_tx` subnet.
+These shard blob headers are signed solely by the blob-builder.
+
+The following validations MUST pass before forwarding the `signed_blob_header` on the network. Alias `header = signed_blob_header.message`
+
+- _[IGNORE]_ The `header` is not from a future slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) --
+  i.e. validate that `header.slot <= current_slot`
+  (a client MAY queue future headers for processing at the appropriate slot).
```
djrtwo, line 147: And you precisely *want* headers/data-txs from the future to put in your mempool. but also want to forward them as well.

specs/sharding/p2p-interface.md
djrtwo: In fact, I"m unsure if there should be an upper bound in the rules here. Maybe a sanity bound? But then, it's more on the data-tx mempool management to flush things that are stale, invaluable, or no longer able to pay fees

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+- _[IGNORE]_ The header is new enough to still be processed --
+  i.e. validate that `compute_epoch_at_slot(header.slot) >= get_previous_epoch(state)`
+- _[REJECT]_ The shard header is for an active shard --
+  i.e. `header.shard < get_active_shard_count(state, compute_epoch_at_slot(header.slot))`
+- _[REJECT]_ The `header.shard` MUST have a committee at the `header.slot` --
+  i.e. validate that `compute_committee_index_from_shard(state, header.slot, header.shard)` doesn't raise an error
+- _[IGNORE]_ The header is the first header with valid signature received for the `(header.builder_index, header.slot, header.shard)` combination.
+- _[REJECT]_ The blob builder exists and has sufficient balance to back the fee payment.
```
djrtwo, line 155: should this consider data-txs in the mempool for the same builder and earlier slots?
that is, assume they *will* make it in and this maybe won't be able be to be paid for? or too strict

djrtwo:
```suggestion
- _[REJECT]_ The blob builder, define by `header.builder_index`, exists and has sufficient balance to back the fee payment.
```

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+The following validations MUST pass before forwarding the `signed_blob_header` on the network. Alias `header = signed_blob_header.message`
+
+- _[IGNORE]_ The `header` is not from a future slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) --
+  i.e. validate that `header.slot <= current_slot`
+  (a client MAY queue future headers for processing at the appropriate slot).
+- _[IGNORE]_ The header is new enough to still be processed --
+  i.e. validate that `compute_epoch_at_slot(header.slot) >= get_previous_epoch(state)`
+- _[REJECT]_ The shard header is for an active shard --
+  i.e. `header.shard < get_active_shard_count(state, compute_epoch_at_slot(header.slot))`
+- _[REJECT]_ The `header.shard` MUST have a committee at the `header.slot` --
```
djrtwo, line 152: We repeat these conditions 3 times. Essentially the "shard blob header" validations. Should we consider wrapping them to reduce copy pasta issues?

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+  i.e. validate that `compute_committee_index_from_shard(state, header.slot, header.shard)` doesn't raise an error
+- _[IGNORE]_ The header is the first header with valid signature received for the `(header.builder_index, header.slot, header.shard)` combination.
```
djrtwo, line 154: this should actually probably be the *highest fee payment* for that combo. So if less than or equal, drop. but a builder should be able to re-send the same or something different for shard/slot that has a higher fee.

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+- _[REJECT]_ The blob builder exists and has sufficient balance to back the fee payment.
+- _[IGNORE]_ The header fee SHOULD be higher than previously seen headers for `(header.slot, header.shard)`, from any builder.
```
djrtwo, line 156: ah, interesting... drop data-txs of lower fees for that shard/slot combo because they are so *pointedly* valuable.

I see that. See previous comment about builder being able to supersede their previous data-tx with higher fee. maybe combine with this statement.

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+  Propagating nodes MAY increase fee increments in case of spam.
```
djrtwo, line 157: I'm not sure I understand this. As in, must be `> prev_max_fee + DELTA` rather than just `> prev_max_fee`?

specs/sharding/p2p-interface.md
```diff
@@ -117,58 +74,101 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
...
+- _[REJECT]_ The header signature is valid for ONLY the builder, `signed_blob_header.signature`,
```
djrtwo, line 158:
```suggestion
- _[REJECT]_ The header signature, `signed_blob_header.signature`, is valid for ONLY the builder --
```

djrtwo (commented):
p2p review:

a bunch of formatting nits, but then some substantive comments and questions about data-tx conditions

great work!

specs/sharding/p2p-interface.md
terencechain, line 140: Maybe I'm missing something, shouldn't this be `SignedShardBlob` instead of `SignedShardBlobHeader`.  In order to publish the full data, don't we need `data` instead of `data_root` ?

Nashatyrev: @terencechain full data publishing is now the builder's duty. As soon as he sees his `SignedShardBlobHeader` on `blob_headers` topic signed by proposer (effectively an aggregate signature: builder + proposer) he has strong incentive to publish the corresponding `SignedShardBlob` as soon as possible.
@protolambda also mentioned that hiding actual blob data until proposer acceptance may potentially add more security as proposer would not be able to filter/withhold blobs based on their content

specs/sharding/beacon-chain.md
protolambda: Good point, we can factor out `samples` and do another earlier comparison. But if we get the precision (uint size) right to avoid overflows, it doesn't make a difference. I'm keeping this as-is, and will address fee precision and corresponding fee asserts in a future PR.

specs/sharding/p2p-interface.md
protolambda: > should this consider data-txs in the mempool for the same builder and earlier slots?
that is, assume they will make it in and this maybe won't be able be to be paid for? or too strict

We should probably address this as part of the fix for parallel fee payments. Doesn't matter if it's in parallel or in-flight on an earlier slot, header fees need to be assured.

specs/sharding/p2p-interface.md
protolambda: We would enable a builder to spam the network if we allow a new data-tx for the same `(slot, shard)` combination, with a higher priority fee. I.e. it's pointless if a builder spams 1.001, 1.002, ...., 1.999 for priority fee. We could define a threshold for replacement to counter spam, but right now I prefer to keep it simple.

specs/sharding/p2p-interface.md
protolambda: >  As in, must be > prev_max_fee + DELTA rather than just > prev_max_fee?

Yes (although it should probably account for base fee and priority fee increments separately)

Similarly to a single builder spamming replacement transactions that only slightly increase the fee, multiple builders could coordinate to spam. E.g. 100 builders publish a transaction, all with slightly increasing fee. As a network participant, it may not be worth propagating every tiny fee increment, as long as one of the highest few get through.

protolambda:
Opened https://github.com/ethereum/eth2.0-specs/issues/2546 to keep track of the open issues. The main things in this PR should be stable enough to merge.

protolambda:
Updated p2p propagation window changes:
- add 4 slot grace period for late txs to propagate
- ignore stale txs, when the shard proposer already selected one
- add 8 slot buffer period for early txs to fill the mempool of a shard proposer ahead of time
- allow headers and blobs 1 slot early. This covers the ideal timing of the diagram in main post, and since headers and blobs are very limited (max 64 per slot), we should not need buffering more slots ahead of time. This should help avoid spikes of multiple slots worth of headers/blobs in these core channels, unlike data-transactions which are less important, in higher numbers, and replaceable in case of loss.

specs/sharding/p2p-interface.md
```diff
@@ -77,9 +80,9 @@ def compute_subnet_for_shard_blob(state: BeaconState, slot: Slot, shard: Shard)
 The following validations MUST pass before forwarding the `signed_blob`,
 on the horizontal subnet or creating samples for it. Alias `blob = signed_blob.message`.
 
-- _[IGNORE]_ The `blob` is not from a future slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) --
-  i.e. validate that `blob.slot <= current_slot`
-  (a client MAY queue future blobs for processing at the appropriate slot).
+- _[IGNORE]_ The `blob` is published 1 slot early or later (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) --
```
djrtwo, line 83: One slot because the proposer moves at N - 1/3, meaning the builder should move *immediately* which might be marginally before N?

djrtwo (approved):
changes look good 👍 
plenty to do, but time to merge this

awesome work!

specs/sharding/p2p-interface.md
protolambda: exactly

PR #2488, opened 2021-06-21
[Sharding] Simplify `get_start_shard` function

Nashatyrev:
Current `get_start_shard` implementation rotates committees across shards absolutely seamlessly in a round robin fashion. 
However it has some drawbacks: 
- The function computation time grows linearly depending on `state.slot - slot` value which may potentially be an attack vector 
- Requires an additional state field with some extra management

Suggested simplified function rotates committees seamlessly unless `committee_count_per_slot` changes (at epoch boundary) or `active_shard_count` changes (currently just due to fork). 
Both events looks pretty rare under normal circumstances. 

The only question is how possible and feasible for an adversary to manipulate the `committee_count_per_slot` value (e.g. by adding/exiting validators) to bounce it back and forth (`+-1`) to prevent some shard(s) to be assigned a committee

### DIFF
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -58,7 +58,6 @@
     - [`process_pending_shard_confirmations`](#process_pending_shard_confirmations)
     - [`charge_confirmed_shard_fees`](#charge_confirmed_shard_fees)
     - [`reset_pending_shard_work`](#reset_pending_shard_work)
-    - [`process_shard_epoch_increment`](#process_shard_epoch_increment)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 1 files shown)


protolambda (commented):
Looks good, although we may want to store the committee-count-per-slot in the state, so we can proof historical committee<>shard relations efficiently.

Nashatyrev:
> Looks good, although we may want to store the committee-count-per-slot in the state, so we can proof historical committee<>shard relations efficiently.

Yep, makes sense. Though may be we could address this later when demand is more clear?

protolambda:
Rebased on latest `dev`, looks good to merge to me.

protolambda (approved):

PR #2491, opened 2021-06-22
optimize sync committee message gossip and caches

djrtwo:
Thank you @ajsutton for identifying this issue and offering and testing solutions!

## Problem and discussion

There is currently a race condition on messages sent on `sync_committee_{subnet_id}` between the sync committee message propagation and the beacon block it is referencing due to the following condition. This is due both to random message paths and latencies plus the fact that actually processing a block locally can take some time in and of itself.

> _[IGNORE]_ The block being signed over (sync_committee_message.beacon_block_root) has been seen (via both gossip and non-gossip sources).

There are similar conditions on attestation and beacon block gossip, but those conditions are accompanied with an optional queuing of the message to attempt to retrieve/process the referenced block before forwarding or ignoring the message. This queue turns out to be critical in healthy processing of eager attestations (sent before the 4s mark) in some cases (especially so in small networks), and thus also in our very similar looking sync committee messages.

Due to fact that the sync committee is small, known, and very stable (defined 1+ day ago) and that their message is only relevant for one slot, we can loosen up the gossip conditions and rely on a small cache for most of the anti-dos measures that can be flushed after each slot.

NOTE: This cache was required for sender deduplication *already*. This PR just relies on the cache more heavily and makes it clear the size and time-length requirements

## Details of solution implemented in this PR

Concretely, this PR makes the following changes to both `sync_committee_contribution_and_proof` and `sync_committee_{subnet_id}` topics:

* Move message send uniqueness condition *after* the condition that checks that the message sender is in the sync (sub)committee. This narrows down possible messages to a bound/known (sub)committee size.
* Add an explicit note about the required size and length of the message sender deduplication cache
* remove the `beacon_block_root` "seen" condition, relying more fully on the deduplication cache and eliminating the race condition

### DIFF
--- a/specs/altair/p2p-interface.md
+++ b/specs/altair/p2p-interface.md
@@ -137,13 +137,14 @@ def get_sync_subcommittee_pubkeys(state: BeaconState, subcommittee_index: uint64
     return sync_committee.pubkeys[i:i + sync_subcommittee_size]
 ```
 
-- _[IGNORE]_ The contribution's slot is for the current slot (with a MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance), i.e. `contribution.slot == current_slot`.
-- _[IGNORE]_ The block being signed over (`contribution.beacon_block_root`) has been seen (via both gossip and non-gossip sources).
+- _[IGNORE]_ The contribution's slot is for the current slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance), i.e. `contribution.slot == current_slot`.
 - _[REJECT]_ The subcommittee index is in the allowed range, i.e. `contribution.subcommittee_index < SYNC_COMMITTEE_SUBNET_COUNT`.
-- _[IGNORE]_ The sync committee contribution is the first valid contribution received for the aggregator with index `contribution_and_proof.aggregator_index` for the slot `contribution.slot` and subcommittee index `contribution.subcommittee_index`.
 - _[REJECT]_ `contribution_and_proof.selection_proof` selects the validator as an aggregator for the slot -- i.e. `is_sync_committee_aggregator(contribution_and_proof.selection_proof)` returns `True`.
 - _[REJECT]_ The aggregator's validator index is in the declared subcommittee of the current sync committee --
   i.e. `state.validators[contribution_and_proof.aggregator_index].pubkey in get_sync_subcommittee_pubkeys(state, contribution.subcommittee_index)`.
+- _[IGNORE]_ The sync committee contribution is the first valid contribution received for the aggregator with index `contribution_and_proof.aggregator_index`
+  for the slot `contribution.slot` and subcommittee index `contribution.subcommittee_index`
+  (this requires maintaining a cache of size `SYNC_COMMITTEE_SIZE` for this topic that can be flushed after each slot).
 - _[REJECT]_ The `contribution_and_proof.selection_proof` is a valid signature of the `SyncAggregatorSelectionData` derived from the `contribution` by the validator with index `contribution_and_proof.aggregator_index`.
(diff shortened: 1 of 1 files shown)


specs/altair/p2p-interface.md
```diff
@@ -137,13 +137,14 @@ def get_sync_subcommittee_pubkeys(state: BeaconState, subcommittee_index: uint64
     return sync_committee.pubkeys[i:i + sync_subcommittee_size]
 ```
 
-- _[IGNORE]_ The contribution's slot is for the current slot (with a MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance), i.e. `contribution.slot == current_slot`.
-- _[IGNORE]_ The block being signed over (`contribution.beacon_block_root`) has been seen (via both gossip and non-gossip sources).
+- _[IGNORE]_ The contribution's slot is for the current slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance), i.e. `contribution.slot == current_slot`.
```
djrtwo, line 140: This line just added some missing ` marks

specs/altair/p2p-interface.md
```diff
@@ -158,12 +159,12 @@ The `sync_committee_{subnet_id}` topics are used to propagate unaggregated sync
 
 The following validations MUST pass before forwarding the `sync_committee_message` on the network:
 
-- _[IGNORE]_ The signature's slot is for the current slot (with a MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance), i.e. `sync_committee_message.slot == current_slot`.
-- _[IGNORE]_ The block being signed over (`sync_committee_message.beacon_block_root`) has been seen (via both gossip and non-gossip sources).
-- _[IGNORE]_ There has been no other valid sync committee signature for the declared `slot` for the validator referenced by `sync_committee_message.validator_index`.
-  Note this validation is _per topic_ so that for a given `slot`, multiple messages could be forwarded with the same `validator_index` as long as the `subnet_id`s are distinct.
+- _[IGNORE]_ The signature's slot is for the current slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance), i.e. `sync_committee_message.slot == current_slot`.
```
djrtwo, line 162: This line just added some missing ` marks

ralexstokes (approved):
changes look sound and i'd support tightening/simplifying the gossip conditions where we can.

the only (plausible) downside here is that this change enshrines use of a cache to handle the DoS vector of messages w/ arbitrary block roots -- the existing solution arguably already requires them but it is worth noting the extra requirements underlying these changes.

djrtwo:
> the only (plausible) downside here is that this change enshrines use of a cache to handle the DoS vector of messages w/ arbitrary block roots 

note! this sender deduplication cache was already required. It just now acts as a failsafe even if signed block roots are not available

dapplion:
On small networks the same validator can appear multiple times in the sync committee. In that case other nodes may receive the same sync signature more than once potentially even through the same subnet.

The new condition should accept duplicate sync signatures with the same validator_index? If so it may be worth to add a note.

djrtwo:
The deduplcation cache should already just allow one message per validator per subnet. There's a note in there already that at least acknowledges this:

> Note this validation is _per topic_ so that for a given `slot`, multiple messages could be forwarded with the same `validator_index` as long as the `subnet_id`s are distinct.

deduplication on this layer is good. the aggregator should then pack multiple if the validator should be included more than once

ajsutton (approved):
LGTM. Quite a simple solution too which is nice.

dapplion:
To clarify, the suggested implementation is to **NOT** add a cache and hold the `sync_committee_message` objects that reference unknown blocks until we get such block or 1 slot, but just propagate them 'immediately'?

PR #2496, opened 2021-06-24
Refactor setup.py and move most hardcoded information into spec.yaml files

### LINKED ISSUE #2490: Setup.py contains lots of hardcoded information
Currently `setup.py` contains large amounts of hardcoded information including substantial amounts of ancillary code included as multiline strings.

We should create a `spec.yaml` file in each spec that contains this information and refactor `setup.py` to be a more general tool for compiling executable specs that is reusable for specifying other parts of the Ethereum protocol.

ultratwo:
This pull request partially resolves #2490 by moving the most of the hardcoded information about forks into `spec.yaml`, `prefix.py` and `suffix.py` files in the respective fork directories.


pipermerriam:
Getting the linting errors cleaned up would probably make this easier to merge/consider.

ultratwo:
Have force pushed a commit to fix the linter errors. I've also made `setup.py` get linted and fixed the lint errors inside it.

A very slight regression is that the autogenerated files no longer meet the requirements of E402 (imports at top of file) and E3XX (wrong number of empty lines). I've added `linter.ini` ignores for those.

pipermerriam:
I'm fine reviewing the python here, but I think it'd be good to have someone with more hands-on experience with working with this spec weigh in prior to that to verify that this is a change they are interested in adopting.

setup.py
```diff
@@ -786,20 +451,44 @@ def load_config(config_path: Path) -> Dict[str, str]:
...
+        dataclasses={},
+        python_prefix=python_prefix,
+        python_suffix=python_suffix,
+    )
+    for value in all_specs:
         spec_object = combine_spec_objects(spec_object, value)
 
+    for function in spec_yaml.get("overridden_functions", []):
+        if function in spec_object.functions:
+            del spec_object.functions[function]
```
hwwhww, line 486: IIRC, it seems that `ceillog2` and `floorlog2` do not need to be overridden?

specs/altair/spec.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - suffix.py
+
+overridden_functions:
+  - eth2_aggregate_pubkeys
+  - ceillog2
+  - floorlog2
+
+predefined_vars:
+  - FINALIZED_ROOT_INDEX
+  - NEXT_SYNC_COMMITTEE_INDEX
```
hwwhww, line 28: `FINALIZED_ROOT_INDEX` and `NEXT_SYNC_COMMITTEE_INDEX` are firstly defined in prefix.yaml and then are listed here. IMHO it would be nice to only have to write them in one place.

How about setting the value here? For example
```
predefined_vars:
  - FINALIZED_ROOT_INDEX: GeneralizedIndex(105)
  - NEXT_SYNC_COMMITTEE_INDEX: GeneralizedIndex(55)
```

And then merge `predefined_vars` into the spec in `setup.py`?


---

Also, it would be nice to add some comments on why we have to predefine them. (due to dependency)

linter.ini
```diff
@@ -2,6 +2,14 @@
 ignore = E252,W504,W503
 max-line-length = 120
 
+per-file-ignores=
+    setup.py:E402
+    ./eth2spec/phase0/*.py:E3,E402
+    ./eth2spec/altair/*.py:E3,E402
+    ./eth2spec/merge/*.py:E3,E402
```
hwwhww, line 9: I'm a little concerned about ignoring `E3XX` since we'd like to check `E303` in the markdown files, inside the functions.

hwwhww (commented):
Thank you for making this PR and I like this solution! 👍 
I verified that `pip install "git+https://github.com/ultratwo/eth2.0-specs@dev"` works. 

> Linter

I think ignoring E402 is okay, but not fully confident about ignoring E3. Can we adjust the blanks by modifying `objects_to_spec`?

setup.py
ultratwo: You are correct. The original `setup.py` contained overriding logic that I ported over without checking.

linter.ini
ultratwo: Fixed. I was being lazy.

ralexstokes:
> I think ignoring E402 is okay

not sure how involved we want this particular PR to be but i think a more complete solution here takes the code in `{prefix, suffix}.py` and patches that into the spec AST/IR we informally have in `setup.py`.  the reason to prefer this solution is that we may have various modifications to do to the spec content before rendering a final Python file (say, after any de-duplication/redefinition required is done).

i'm just commenting this here now as it is top of mind given some work i was just doing on spec code coverage -- we re-duplicate some phase0 spec functions into the altair spec and ideally we drop the ones no longer used in subsequent forks

we can definitely keep this PR as is and handle some of these wider reaching refactors in subsequent PRs

setup.py
```diff
@@ -786,20 +453,44 @@ def load_config(config_path: Path) -> Dict[str, str]:
...
-                source_files: Sequence[Path], preset_files: Sequence[Path], config_file: Path) -> str:
+def build_spec(preset_name: str, fork: str, source_files: Sequence[Path], preset_files: Sequence[Path],
+               config_file: Path, spec_dir: Path, spec_yaml: Dict) -> str:
     preset = load_preset(preset_files)
     config = load_config(config_file)
     all_specs = [get_spec(spec, preset, config) for spec in source_files]
 
-    spec_object = all_specs[0]
-    for value in all_specs[1:]:
+    python_prefix = []
```
CarlBeek, line 462: In general, I think another pass needs to be taken at handling including and combining the `python_prefix`es and suffixes. The was it is currently done does not match the pattern of the rest of this pyspec_builder, that of extracting the prefix into a `spec_object` and then having a reducer function to combine spec objects back into a single object.

IMO, the prefixes & suffixes for each phase should be extracted in `build_spec()` instead of being added separately in the `spec.yaml`. If this pattern is not used here, then I think we should remove the combiner functionality that is not used and this anti-patter should be explained as a comment.

CarlBeek (changes requested):
I really like the work you've done here @ultratwo, it definitely helps remove clutter and make it easier to maintain understand the pyspec_builder by parametrising all the components of each phase.

setup.py
```diff
@@ -97,26 +79,30 @@ def _get_source_from_code_block(block: FencedCode) -> str:
...
     fn = ast.parse(source).body[0]
+    assert isinstance(fn, FunctionDef)
     args = fn.args.args
     if len(args) == 0:
         return None
     if args[0].arg != 'self':
         return None
     if args[0].annotation is None:
         return None
+    assert isinstance(args[0].annotation, Name)
```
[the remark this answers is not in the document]
ultratwo: These aren't actually my idea. They were required by `mypy`.

setup.py
```diff
@@ -248,319 +231,24 @@ def get_spec(file_name: Path, preset: Dict[str, str], config: Dict[str, str]) ->
...
     return value.startswith('ByteList') or value.startswith('Union')
 
 
 def objects_to_spec(preset_name: str,
                     spec_object: SpecObject,
-                    builder: SpecBuilder,
-                    ordered_class_objects: Dict[str, str]) -> str:
+                    fork: str,
+                    ordered_class_objects: Dict[str, str],
+                    spec_yaml: Dict) -> str:
```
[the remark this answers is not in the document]
ultratwo: I've renamed it `spec_dot_yaml` since it is just the parsed contents of `spec.yaml`.

setup.py
```diff
@@ -699,48 +376,38 @@ def dependency_order_class_objects(objects: Dict[str, str], custom_types: Dict[s
...
     custom_types = combine_dicts(spec0.custom_types, spec1.custom_types)
-    constant_vars = combine_dicts(spec0.constant_vars, spec1.constant_vars)
-    preset_vars = combine_dicts(spec0.preset_vars, spec1.preset_vars)
+    vars = combine_dicts(spec0.vars, spec1.vars)
     config_vars = combine_dicts(spec0.config_vars, spec1.config_vars)
-    ssz_dep_constants = combine_dicts(spec0.ssz_dep_constants, spec1.ssz_dep_constants)
-    ssz_objects = combine_ssz_objects(spec0.ssz_objects, spec1.ssz_objects, custom_types)
+    ssz_objects = combine_dicts(spec0.ssz_objects, spec1.ssz_objects)
     dataclasses = combine_dicts(spec0.dataclasses, spec1.dataclasses)
+    python_prefix = spec0.python_prefix + spec1.python_prefix
```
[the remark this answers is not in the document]
ultratwo: The linting error, `F811 redefinition of unused 'name' from line n` is self explanatory, so I don't see what benefit a comment here would add. If such a redefinition is needed it can be tagged with `# noqa: F811` .

ultratwo:
I've implemented inheritance between `spec.yaml` files to eliminate repetition. I've also created a `SpecConfig` object to hold all the information from `spec.yaml`, which makes the code neater.

setup.py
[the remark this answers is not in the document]
CarlBeek: Yup, I'm happy with this, I just wanted to disambiguate against all the config and testing files that are also yaml based across the rest of this (and other) repo(s).

leolara:

PR #2497, opened 2021-06-24
encode, decode and randomize ssz Union types

protolambda:
The new Union type needs debug utils to encode/decode to yaml in tests, and randomize values for ssz tests.


ralexstokes (approved):
changes lgtm!

this PR adds some integration code to bridge `remerkleable` with our use of ssz in the pyspec

Issue #2227, opened 2021-03-11
Disallow Voting For the Current Eth1Data in State


nisdas:
It has been observed in mainnet that during certain voting periods, clients choose to vote on an already included `eth1data`. This ends up making the voting period redundant, as clients voting for an already included `eth1data` and its accompanying deposit count and root does not allow newer deposits to be included. 

This ends up with deposits taking longer than expected(an additional voting period) to being included.  @ajsutton has also brought up the fact that currently in the spec: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/validator.md#eth1-data 

```py
    default_vote = votes_to_consider[len(votes_to_consider) - 1] if any(votes_to_consider) else state_eth1_data
```
We default to voting for the `state_eth1_data` if there are no other viable eth1data votes. We should instead disallow `state_eth1_data` votes and mark them as invalid when each block proposer is tallying the current votes in the state. This is so as to prefer strictly progressing `eth1_data` rather than allowing block proposers to vote for an already included `eth1_data` tuple which is in the state. This will prevent a whole voting period from being wasted and deposits from being delayed to be included, which has been show to happen in certain voting periods.

There is limited downside to including a change such as this due to the fact this will only disallow voting for an already included on `eth1_blockhash` , so this will always lead to a progressive `eth1_data` votes from each period to the next.

seishun:
Is this the change you have in mind?

```diff
diff --git a/specs/phase0/validator.md b/specs/phase0/validator.md
index 14a5ecbb..43b86488 100644
--- a/specs/phase0/validator.md
+++ b/specs/phase0/validator.md
@@ -349,6 +349,7 @@ def get_eth1_vote(state: BeaconState, eth1_chain: Sequence[Eth1Block]) -> Eth1Da             is_candidate_block(block, period_start)
             # Ensure cannot move back to earlier deposit contract states
             and get_eth1_data(block).deposit_count >= state.eth1_data.deposit_count
+            and get_eth1_data(block) != state.eth1_data
         )
     ]
```

nisdas:
That is part of it, also we would need to remove this condition 
```py
    # Default vote on latest eth1 block data in the period range unless eth1 chain is not live
    # Non-substantive casting for linter
    state_eth1_data: Eth1Data = state.eth1_data
    default_vote = votes_to_consider[len(votes_to_consider) - 1] if any(votes_to_consider) else state_eth1_data

```
[...]

PR #2501, opened 2021-07-01
Default to "zero" Eth1Data instead of current state

ajsutton:
Addresses #2227 by changing the default vote to be zeroed out - zero block hash, deposit root and zero deposits. The state transition is updated to explicitly disallow the zero vote from becoming the current `Eth1Data` in the spec.

The zero vote is automatically ignored by `get_eth1_vote` because it won't be in the set of known blocks so doesn't lead validators to following a vote which doesn't progress the chain like defaulting to the current `state.eth1_data` does.

The alternative approach would be for validators to vote for a random block hash which has the advantage of not needing to update the state transition rules as it's exceedingly unlikely 50% of validators will choose the same random data, but the downside of making it harder to identify default votes. Differentiating default votes from votes for unknown blocks is particularly useful when evaluating whether a beacon node is tracking the eth1 chain correctly.

Note: Python likely needs some tweaking but hopefully gets the idea across.

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -1732,8 +1732,9 @@ def process_randao(state: BeaconState, body: BeaconBlockBody) -> None:
 ```python
 def process_eth1_data(state: BeaconState, body: BeaconBlockBody) -> None:
     state.eth1_data_votes.append(body.eth1_data)
-    if state.eth1_data_votes.count(body.eth1_data) * 2 > EPOCHS_PER_ETH1_VOTING_PERIOD * SLOTS_PER_EPOCH:
-        state.eth1_data = body.eth1_data
+    if body.eth1_data.block_hash != Bytes32(): 
+        if state.eth1_data_votes.count(body.eth1_data) * 2 > EPOCHS_PER_ETH1_VOTING_PERIOD * SLOTS_PER_EPOCH:
+            state.eth1_data = body.eth1_data
 ```
 
 #### Operations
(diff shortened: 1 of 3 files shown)


specs/phase0/beacon-chain.md
```diff
@@ -1732,8 +1732,9 @@ def process_randao(state: BeaconState, body: BeaconBlockBody) -> None:
 ```python
 def process_eth1_data(state: BeaconState, body: BeaconBlockBody) -> None:
     state.eth1_data_votes.append(body.eth1_data)
-    if state.eth1_data_votes.count(body.eth1_data) * 2 > EPOCHS_PER_ETH1_VOTING_PERIOD * SLOTS_PER_EPOCH:
-        state.eth1_data = body.eth1_data
+    if body.eth1_data.block_hash != Bytes32():
```
ajsutton, line 1735: Regardless of the change to default volte, should we have a check in here that the deposit count can't go backwards? It can only happen if enough validators don't follow the honest validator spec but would definitely be a very odd case to allow.

ajsutton:
Also worth noting that this would have to be rolled out as a soft fork - first deploy the state transition change, then change the default vote. Seems safe but is perhaps a reason to just use a random vote instead and deal with not being able to distinguish default votes from unknown blocks.

ajsutton:
At this point I think we can just give up on ever actually making this change. Eth1Data is working fine and the next time we change it should be to get rid of it entirely.

Issue #2480, opened 2021-06-15
Improve Altair pytest coverage


hwwhww:
- [ ] **[phase0]** beacon
    - `slash_validator`: `whistleblower_index` was never `None`
        - this field is used by [custody game `process_challenge_deadlines`](https://github.com/ethereum/eth2.0-specs/blob/378d167ee03a4017b53dc54cac15a99ea4392313/specs/custody_game/beacon-chain.md#handling-of-reveal-deadlines), but currently, the custody tests are not enabled.
- [ ] **[phase0]** fork choice tests (#2487)
    - some conditions were never be satisfied or be false.
    - [x] `filter_block_tree` never returns `False`
    - [x] `should_update_justified_checkpoint` never returns `True`
    - [x] `update_latest_messages`: condition `i not in store.latest_messages or target.epoch > store.latest_messages[i].epoch` was never `False`
    - [ ] `on_block`: `state.current_justified_checkpoint.epoch > store.justified_checkpoint.epoch` was never `False`
        - edited: currently commented out
- [ ] **[phase0]** validator
    - `get_committee_assignment` never returns `None`
    - `get_aggregate_signature` is uncovered
- [ ] **[phase0]** weak subjectivity
    - the whole spec is untested
- [x] **[altair]** validator & networking (#2499, #2505)
    - some helper functions are uncovered
    - `get_sync_committee_message`
    - `get_sync_committee_selection_proof`
    - `is_sync_committee_aggregator`
    - `get_contribution_and_proof`
    - `get_contribution_and_proof_signature`
    - `get_sync_subcommittee_pubkeys`

hwwhww:
I'd pick fork choice & validator guide tests. 👋

djrtwo:
I did a pass on coverage needed in state transition and am documenting progress here:

https://docs.google.com/spreadsheets/d/1CbJy3iCRfXnG_jclMrDpktE-gEOXP9nAtGzoUYmuLVM/edit#gid=0

leolara:
I am closing this issue because it seems stale. Please, do not hesitate to reopen it if this is a mistake

PR #2505, opened 2021-07-02
Add Altair networking helper tests

hwwhww:
Part of #2480

### DIFF
--- a/tests/core/pyspec/eth2spec/test/altair/unittests/networking/test_networking.py
+++ b/tests/core/pyspec/eth2spec/test/altair/unittests/networking/test_networking.py
@@ -1,0 +1,47 @@
+from eth2spec.test.context import (
+    with_altair_and_later,
+    spec_state_test,
+)
+from eth2spec.test.helpers.state import (
+    transition_to,
+)
+
+
+@with_altair_and_later
+@spec_state_test
+def test_get_sync_subcommittee_pubkeys_current_sync_committee(state, spec):
+    # Transition to the head of the next period
+    transition_to(spec, state, spec.SLOTS_PER_EPOCH * spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
+
+    next_slot_epoch = spec.compute_epoch_at_slot(state.slot + 1)
+    assert (
+        spec.compute_sync_committee_period(spec.get_current_epoch(state))
+        == spec.compute_sync_committee_period(next_slot_epoch)
+    )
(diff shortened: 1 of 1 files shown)


tests/core/pyspec/eth2spec/test/altair/unittests/networking/test_networking.py
```diff
@@ -0,0 +1,47 @@
...
+from eth2spec.test.helpers.state import (
+    transition_to,
+)
+
+
+@with_altair_and_later
+@spec_state_test
+def test_get_sync_subcommittee_pubkeys_current_sync_committee(state, spec):
+    # Transition to the head of the next period
+    transition_to(spec, state, spec.SLOTS_PER_EPOCH * spec.EPOCHS_PER_SYNC_COMMITTEE_PERIOD)
```
ralexstokes, line 14: this assumes we are aligned to `EPOCHS_PER_SYNC_COMMITTEE_PERIOD`, which we are, but i don't know if that is an explicit guarantee of our testing infra... i've seen it used enough places that perhaps it is something we (I) can rely on :)

tests/core/pyspec/eth2spec/test/altair/unittests/networking/test_networking.py
```diff
@@ -0,0 +1,47 @@
...
+        spec.compute_sync_committee_period(spec.get_current_epoch(state))
+        != spec.compute_sync_committee_period(next_slot_epoch)
+    )
+    sync_committee = state.next_sync_committee
+    sync_subcommittee_size = spec.SYNC_COMMITTEE_SIZE // spec.SYNC_COMMITTEE_SUBNET_COUNT
+    subcommittee_index = 1
+    i = subcommittee_index * sync_subcommittee_size
+
+    expect = sync_committee.pubkeys[i:i + sync_subcommittee_size]
+    assert spec.get_sync_subcommittee_pubkeys(state, subcommittee_index) == expect
```
ralexstokes, line 47: just making a comment to explain this setup:

given the fact that sync committees sign the previous block root, validators must reference the _next_ sync committee when they are in the last slot of a sync committee period. by the time the block for this slot is processed, the committees will have rotated in the `state` so that the validation is always happening wrt to the "current" committee.

this is also why we can still use "current" in the test above as this rotation has already happened, even though we are in the "next" period (with respect to the starting state of that test)

ralexstokes (approved):
looks good! more test coverage!!

djrtwo (approved):

PR #2506, opened 2021-07-02
Move `process_sync_aggregate` call into `process_operations`

hwwhww:
### Issue

https://github.com/ethereum/eth2.0-specs/pull/2473#issuecomment-856880216

1. As #2473 pointed out, `process_block` is a bit inconsistent when calling `process_sync_aggregate`.
2.  [`sync_aggregate` and `execution_payload ` are considered as "operations" in test format](https://github.com/ethereum/eth2.0-specs/blob/c821e3fced725995242e12037ad9ce502d5b2208/tests/formats/operations/README.md#condition). The pattern (type signature) of `process_sync_aggregate` is similar to `process_SSZ_CONTAINER(state, SSZ_CONTAINER)` functions  like `process_attestation(state, attestation)`. IMHO "`process_operations`" should be the function to process all operations.
    - **20210713 updated**: reverted the change of `process_execution_payload`

### Proposed solution

Call `process_sync_aggregate` in `process_operations` instead of directly in `process_block`. So the name "`process_operations`" makes more sense.

Note that `process_randao` and `process_eth1_data` should have been "operations" in testing. However, it was too late to change phase0 spec. :'(

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -42,6 +42,7 @@
   - [Beacon state mutators](#beacon-state-mutators)
     - [Modified `slash_validator`](#modified-slash_validator)
   - [Block processing](#block-processing)
+    - [Operations](#operations)
     - [Modified `process_attestation`](#modified-process_attestation)
     - [Modified `process_deposit`](#modified-process_deposit)
     - [Sync aggregate processing](#sync-aggregate-processing)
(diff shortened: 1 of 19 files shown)


djrtwo:
Even though they are considered "operations" in the testing today, I would argue that they aren't semantically operations like the other operation lists are.

Sync aggregate is certainly closer to an operation (a system level TX) than execution payload, but I think I'd stand pretty firm that execution payload is *not* an operation and should be segregated out in `process_block` as it's own call

hwwhww:
@djrtwo @mkalinin 
Thanks for your feedback and signal!
I reverted the change of `process_execution_payload` and updated the operation tests README file to be more compatible for both "operations" and "execution payload".

djrtwo:
Where do you currently stand on this @hwwhww?

I lean toward "do nothing" on the altair front. I understand that this is a non-substantive change, but it would likely induce a wave of minor code changes on all clients and on some related infra (e.g. beaconfuzz).

Any input @protolambda or @ralexstokes?

protolambda:
Personally I don't like like the process_operations endpoint for various reasons:
- We don't use it in testing anyway
- Ideally everything that happens inside the body should be encapsulated and tested nicely in isolation. Not doing it this way for randao and eth1 voting was a mistake.
- Removing the distinction between operation and encapsulated block body content seems like the most viable way to make the two kinds consistent.

At the same time I don't like to reorganize Altair, so I would avoid breaking changes there. Then with the Merge we have an opportunity to do it right, and just encapsulate every small change.

hwwhww:
I agreed that it seems too late for Altair since the testnet fork is coming really soon.

> it would likely induce a wave of minor code changes on all clients and on some related infra

> Then with the Merge we have an opportunity to do it right, and just encapsulate every small change.

I'm afraid of if we induce a wave of minor code changes to The Merge. **If** there are more *small* changes that should be addressed, then we can consider if it makes sense to have a small "clean-up" HF before The Merge.

djrtwo:
bring back up for discussion in the merge!

PR #2446, opened 2021-05-25
Rename `SyncCommitteeSignature` to `SyncCommitteeMessage` to avoid name conflicts

hwwhww:
### Issue
Variable name conflicts
1. The `SyncCommitteeSignature` SSZ container has multiple fields, not just the signature.
2. OTOH, `SyncAggregate.sync_committee_signature: BLSSignature` is `BLSSignature` type.

It would be better to avoid conflicts and potential confusion.

### Proposed change

Given that `SyncCommitteeSignature` is mainly the data for the gossip topic. I'd suggest renaming it to `SyncCommitteeMessage`.

Open to other proposals. 🙂

PR #2509, opened 2021-07-07
Rename a few instances of `signature` to `message`

terencechain:
Minor follow up to #2446. I think there's a few more instances where `sync_committee_signature` could be renamed to `sync_committee_message` in the p2p interface spec


ralexstokes (approved):
lgtm!

djrtwo (approved):

PR #2510, opened 2021-07-07
Update `get_next_sync_committee` notes

terencechain:
This PR clarified that `get_next_sync_committee` is also called when upgrading beacon state to Altair

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -278,7 +278,7 @@ def get_next_sync_committee_indices(state: BeaconState) -> Sequence[ValidatorInd
 
 #### `get_next_sync_committee`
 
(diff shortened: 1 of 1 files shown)


specs/altair/beacon-chain.md
```diff
@@ -278,7 +278,7 @@ def get_next_sync_committee_indices(state: BeaconState) -> Sequence[ValidatorInd
 
 #### `get_next_sync_committee`
 
-*Note*: The function `get_next_sync_committee` should only be called at sync committee period boundaries.
+*Note*: The function `get_next_sync_committee` should only be called at sync committee period boundaries and upgrading state to Altair.
```
djrtwo, line 281:
```suggestion
*Note*: The function `get_next_sync_committee` should only be called at sync committee period boundaries and when [upgrading state to Altair](./fork.md#upgrading-the-state).
```

djrtwo (approved):

PR #2511, opened 2021-07-07
extend inactivity updates tests

djrtwo:
* Add a number of `process_inactivity_scores` tests
* Add one more sanity test for inactivity score updates block test
* [side bonus] Fix `leaking()` default calculation which only worked in most tests because they called variants of `prepare_state_with_attestations` which transitioned forward 1 epoch, pushing it truly into a leak

working through tests [here](https://docs.google.com/spreadsheets/d/1CbJy3iCRfXnG_jclMrDpktE-gEOXP9nAtGzoUYmuLVM/edit#gid=0). Those marked "PR 2511" are ones added in this PR

### DIFF
--- a/tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_inactivity_updates.py
+++ b/tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_inactivity_updates.py
@@ -4,7 +4,7 @@ from eth2spec.test.context import spec_state_test, with_altair_and_later
 from eth2spec.test.helpers.inactivity_scores import randomize_inactivity_scores, zero_inactivity_scores
 from eth2spec.test.helpers.state import (
     next_epoch_via_block,
-    set_full_participation, set_full_participation_previous_epoch,
+    set_full_participation,
     set_empty_participation,
 )
 from eth2spec.test.helpers.epoch_processing import (
@@ -57,7 +57,6 @@ def run_inactivity_scores_test(spec, state, participation_fn=None, inactivity_sc
 @spec_state_test
 def test_all_zero_inactivity_scores_empty_participation(spec, state):
     yield from run_inactivity_scores_test(spec, state, set_empty_participation, zero_inactivity_scores)
-    # Not yet in leak so no leak score added even though no participation
     assert set(state.inactivity_scores) == set([0])
 
 
@@ -119,7 +118,7 @@ def test_all_zero_inactivity_scores_full_participation_leaking(spec, state):
     # Only set full participation in previous epoch to remain in leak
     yield from run_inactivity_scores_test(
         spec, state,
-        set_full_participation_previous_epoch, zero_inactivity_scores,
+        set_full_participation, zero_inactivity_scores,
     )
 
     # Check still in leak
@@ -189,7 +188,7 @@ def test_random_inactivity_scores_full_participation_leaking(spec, state):
     # Only set full participation in previous epoch to remain in leak
     yield from run_inactivity_scores_test(
         spec, state,
-        set_full_participation_previous_epoch, randomize_inactivity_scores, Random(33333),
+        set_full_participation, randomize_inactivity_scores, Random(33333),
     )
 
     # Check still in leak
@@ -222,7 +221,7 @@ def test_some_slashed_zero_scores_full_participation_leaking(spec, state):
     slash_some_validators(spec, state, rng=Random(33221))
     yield from run_inactivity_scores_test(
         spec, state,
-        set_full_participation_previous_epoch, zero_inactivity_scores,
+        set_full_participation, zero_inactivity_scores,
     )
 
     # Check still in leak
@@ -243,7 +242,7 @@ def test_some_slashed_full_random(spec, state):
     slash_some_validators(spec, state, rng=rng)
     yield from run_inactivity_scores_test(
         spec, state,
-        randomize_previous_epoch_participation, randomize_inactivity_scores, rng=rng,
+        randomize_attestation_participation, randomize_inactivity_scores, rng=rng,
     )
 
 
@@ -257,3 +256,6 @@ def test_some_slashed_full_random_leaking(spec, state):
         spec, state,
         randomize_previous_epoch_participation, randomize_inactivity_scores, rng=rng,
     )
+
+    # Check still in leak
+    assert spec.is_in_inactivity_leak(state)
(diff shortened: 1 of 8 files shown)


tests/core/pyspec/eth2spec/test/altair/sanity/test_blocks.py
```diff
@@ -97,5 +102,36 @@ def test_inactivity_scores(spec, state):
     yield 'blocks', [signed_block]
     yield 'post', state
 
+    # No particiaption during a leak so all scores should increase
```
ralexstokes, line 105:
```suggestion
    # No participation during a leak so all scores should increase
```

tests/core/pyspec/eth2spec/test/helpers/rewards.py
```diff
@@ -260,9 +260,9 @@ def run_get_inactivity_penalty_deltas(spec, state):
...
     if epochs is None:
-        # +1 to trigger inactivity_score transitions
-        epochs = spec.MIN_EPOCHS_TO_INACTIVITY_PENALTY + 1
-    assert epochs >= spec.MIN_EPOCHS_TO_INACTIVITY_PENALTY
+        # +2 because finality delay is based on previous_epoch and must be more than `MIN_EPOCHS_TO_INACTIVITY_PENALTY`
+        epochs = spec.MIN_EPOCHS_TO_INACTIVITY_PENALTY + 2
+    assert epochs > spec.MIN_EPOCHS_TO_INACTIVITY_PENALTY
 
     for _ in range(epochs):
         next_epoch(spec, state)
```
ralexstokes, line 268: any reason to not do this? it sounds like this was buggy so that `test_random_high_inactivity_scores_leaking_5_epochs` didn't actually test the scores under a leaking condition... in which case, we should verify the leak

```suggestion
        next_epoch(spec, state)
    assert spec.is_in_inactivity_leak(state)
```

tests/core/pyspec/eth2spec/test/altair/rewards/test_inactivity_scores.py
```diff
@@ -112,7 +112,7 @@ def test_random_high_inactivity_scores_leaking(spec, state):
 
 @with_altair_and_later
 @spec_state_test
-@leaking(epochs=5)
-def test_random_high_inactivity_scores_leaking_5_epochs(spec, state):
```
ralexstokes, line 116: just noting why this changed: want the `previous_epoch` to be past `MIN_EPOCHS_TO_INACTIVITY_PENALTY` so that we see the effect of the leak on scoring (which suggests this test was buggy before -- see other comment about the suggestion I made for this decorator)

tests/core/pyspec/eth2spec/test/altair/sanity/test_blocks.py
```diff
@@ -97,5 +102,36 @@ def test_inactivity_scores(spec, state):
...
+    signed_block = state_transition_and_sign_block(spec, state, block)
+
+    assert spec.is_in_inactivity_leak(state)
+
+    yield 'blocks', [signed_block]
+    yield 'post', state
+
+    # Full particiaption during a leak so all scores should decrease by 1
+    print(previous_inactivity_scores)
+    print(state.inactivity_scores)
```
ralexstokes, line 135:
```suggestion
```

tests/core/pyspec/eth2spec/test/helpers/random.py
```diff
@@ -100,6 +100,13 @@ def set_flag(index, value):
             epoch_participation[index] = flags
 
 
+def randomize_previous_epoch_participation(spec, state, rng=Random(8020)):
+    cached_prepare_state_with_attestations(spec, state)
+    randomize_epoch_participation(spec, state, spec.get_previous_epoch(state), rng)
+    if not is_post_altair(spec):
+        state.current_epoch_attestations = []
```
ralexstokes, line 107: this function sets the previous epoch participation _and also_ clears the current epoch attestations under phase 0 -- it seems like these should be two separate functions but i imagine it is simpler to handle this here than to go and edit a bunch of other tests to handle the situation case-by-case

tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_inactivity_updates.py
```diff
@@ -33,58 +27,232 @@ def test_genesis(spec, state):
...
+
+
 @with_altair_and_later
 @spec_state_test
 def test_all_zero_inactivity_scores_empty_participation(spec, state):
-    next_epoch_via_block(spec, state)
-    state.inactivity_scores = [0] * len(state.validators)
-    yield from run_process_inactivity_updates(spec, state)
+    yield from run_inactivity_scores_test(spec, state, set_empty_participation, zero_inactivity_scores)
+    assert set(state.inactivity_scores) == set([0])
```
ralexstokes, line 60: note: even w/ no participation, the state is not yet in an inactivity leak so any change to the scores is offset by the `RECOVERY` constant

tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_inactivity_updates.py
```diff
@@ -33,58 +27,232 @@ def test_genesis(spec, state):
...
+        randomize_attestation_participation, zero_inactivity_scores, rng=Random(5555),
+    )
+    assert set(state.inactivity_scores) == set([0])
+
+
+@with_altair_and_later
+@spec_state_test
+@leaking()
+def test_all_zero_inactivity_scores_random_participation_leaking(spec, state):
+    # Only randompize participation in previous epoch to remain in leak
```
ralexstokes, line 90:
```suggestion
    # Only randomize participation in previous epoch to remain in leak
```

ralexstokes (approved):
nice work!

good catch on the `leaking` decorator bug

left some comments/suggestions below

tests/core/pyspec/eth2spec/test/helpers/rewards.py
djrtwo: yeah, will add it

tests/core/pyspec/eth2spec/test/altair/rewards/test_inactivity_scores.py
djrtwo: right, also want it to be beyond the standard minimum leak thus 8

tests/core/pyspec/eth2spec/test/altair/sanity/test_blocks.py
djrtwo: whoops

tests/core/pyspec/eth2spec/test/helpers/random.py
djrtwo: yeah, problem is `cache_prepare_state_with_attestations` is already does a ton of work.

I'm in favor of adding a set of simplified state manipulators for altair+ and slowly pushing out these old ones and only using where necessary

PR #2514, opened 2021-07-12
target sync committee aggregators to 16

djrtwo:
@mcdee Did some quick math to show that the current configuration of `TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE` as `4` had an unacceptably high probability of sync committee subnets not having aggregators.

@hwwhww tossed the math and some options into this [spreadsheet](https://docs.google.com/spreadsheets/d/1C7pBqEWJgzk3_jesLkqJoDTnjZOODnGTOJUrxUMdxMA/edit?usp=sharing). Based on discussions, it looks like the choice of `16` is the right tradeoff to generally have a high chance of having at least one aggregator per slot, even on a daily basis.

Note, that `4` was naively selected to reduce chatter on yet another global channel, but assuming attestation aggregation is not already overwhelming, this increases global messages for ~6% (4 more 16 aggregator topics on top of the 64 attestation topics that also have 16 aggregator targets)

### DIFF
--- a/specs/altair/validator.md
+++ b/specs/altair/validator.md
@@ -73,7 +73,7 @@ This document is currently illustrative for early Altair testnets and some parts
 
 | Name | Value | Unit |
 | - | - | :-: |
-| `TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE` | `2**2` (= 4) | validators |
+| `TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE` | `2**4` (= 16) | validators |
 | `SYNC_COMMITTEE_SUBNET_COUNT` | `4` | The number of sync committee subnets used in the gossipsub aggregation protocol. |
 
 ## Containers
--- a/tests/core/pyspec/eth2spec/test/altair/unittests/validator/test_validator.py
+++ b/tests/core/pyspec/eth2spec/test/altair/unittests/validator/test_validator.py
@@ -13,6 +13,7 @@ from eth2spec.test.context import (
     with_presets,
 )
 from eth2spec.test.helpers.constants import (
+    MAINNET,
     MINIMAL,
 )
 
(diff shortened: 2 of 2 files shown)


djrtwo:
Thanks for the fix @hwwhww !

hwwhww (approved):
The logic LGTM. 

And about if should `TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE` be in preset/config yaml file, `TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE` should be "constant" if `SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT := 128` is also constant. We could:
1. Set that `SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT` should be a constant value. So we can add `assert SYNC_COMMITTEE_SIZE // SYNC_COMMITTEE_SUBNET_COUNT == TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE * 4` to [`test_config_invariants.py`](https://github.com/ethereum/eth2.0-specs/blob/b97972c09186fb6d71022d2d0bc2b0b2fe7d5b1f/tests/core/pyspec/eth2spec/test/altair/unittests/test_config_invariants.py);
2. Or, we can put `TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE` to preset file to remind the client devs of updating this field correspondingly in the future. (not strongly for this time)

What do you think?

djrtwo:
What does the restriction with (1) buy us?

hwwhww:
> What does the restriction with (1) buy us?

I was thinking that if we somehow update `SYNC_COMMITTEE_SIZE` or `SYNC_COMMITTEE_SUBNET_COUNT ` for other reasons, we would be notified to increase `TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE` if needs. However, I just tried other numbers on the spreadsheet, the `TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE = 16` should have guaranteed a high chance of aggregator selections with different committee sizes. So, it's not "required". 😅

djrtwo:
going to merge as is for now!
thanks

PR #2517, opened 2021-07-14
Apply missing spec_test decorators, to handle generator_mode flag

protolambda:
Some new tests in the Altair v1.1.0 beta 1 release did not have all required decorators. Most tests use `spec_state_test`, but when using a custom state (`with_custom_state`), the `spec_test` decorator is necessary.

Think I got all occurrences, only 2. Running test-generation now to verify. Edit: block processing test gen looks ok.


ralexstokes (approved):
nice catch, thanks!

PR #2522, opened 2021-07-16
Enforce terminal PoW block to be on the cusp

mkalinin:
### What's done
Execution client must stop importing PoW blocks beyond the terminal PoW block. This PR introduces the corresponding change to this requirement.

### Rationale 
Suppose the beacon chain client goes offline before `transition_total_difficulty` has been reached and stays offline while the network meets this threshold. This may happen either due to misconfiguration or software bug and results in the inability of the execution client to switch to the proof-of-stake and is following the wrong chain. Stopping block import after terminal PoW block would prevent potential security issues in this case.

Credits to @dankrad who made this point during [Eth2 Call 68](https://github.com/ethereum/eth2.0-pm/issues/226).

### DIFF
--- a/specs/merge/validator.md
+++ b/specs/merge/validator.md
@@ -65,8 +65,7 @@ All validator responsibilities remain unchanged other than those noted below. Na
 * Set `block.body.execution_payload = get_execution_payload(state, transition_store, execution_engine, pow_chain)` where:
 
 ```python
-def get_pow_block_at_total_difficulty(total_difficulty: uint256,
-                                      pow_chain: Sequence[PowBlock]) -> PyUnion[PowBlock, None]:
+def get_pow_block_at_total_difficulty(total_difficulty: uint256, pow_chain: Sequence[PowBlock]) -> Optional[PowBlock]:
     # `pow_chain` abstractly represents all blocks in the PoW chain
     for block in pow_chain:
         parent = get_pow_block(block.parent_hash)
--- a/specs/merge/validator.md
+++ b/specs/merge/validator.md
@@ -19,7 +19,6 @@
   - [Block proposal](#block-proposal)
     - [Constructing the `BeaconBlockBody`](#constructing-the-beaconblockbody)
       - [Execution Payload](#execution-payload)
-        - [`get_pow_chain_head`](#get_pow_chain_head)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 <!-- /TOC -->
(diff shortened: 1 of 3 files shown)


specs/merge/validator.md
```diff
@@ -90,7 +90,8 @@ def get_execution_payload(state: BeaconState,
                           execution_engine: ExecutionEngine) -> ExecutionPayload:
     if not is_merge_complete(state):
         pow_block = get_pow_chain_head()
```
djrtwo, line 92: This is making an assumption that the execution-engine is rejecting blocks past valid terminal PoW blocks, otherwise (if for example a child is the head past the terminal pow block) this function might never see `is_valid_terminal_pow_block` as true.

I'm not certain if this should be noted here, but if we keep the semantics of `get_pow_chain_head` it must be a hard requiremnt somewhere in the spec that the PoW chain rejects beyond terminal difficulty. Otherwise you'd have to do `get_pow_chain_at_difficulty(terminal_difficulty) -> Union[Block, None]` if this is not a strict requirement under the hood

djrtwo (approved):
looks good! one quick discussion point

specs/merge/validator.md
mkalinin: > This is making an assumption that the execution-engine is rejecting blocks past valid terminal PoW blocks

That was the intention. The execution-layer must have a hard requirement on not importing PoW blocks beyond the terminal one which implies that the head won't move beyond the terminal PoW block. However, I do see the value in getting rid of this assumption on the beacon chain side.

> Otherwise you'd have to do get_pow_chain_at_difficulty(terminal_difficulty) -> Union[Block, None]

Looks reasonable and this function may be implemented without the corresponding query to the execution client. Beacon chain client may follow the PoW chain in the same fashion as it does in order to obtain `Eth1Data` and filter the PoW block tree to get terminal PoW block.

mkalinin:
In the most recent change there was a conflict between `Union` from `typing` module and `Union` from `ssz_typing`. I took the path with less changes and did `from typing import Union as PyUnion`, another option is to have `from eth2spec.utils.ssz.ssz_typing import Union as SSZUnion` but it also requires the change in the `beacon-chain.md`.

Not sure which one is better and leaving it up you, cc @djrtwo @protolambda @hwwhww

hwwhww:
@mkalinin  (Sorry, I just saw it!)

> `PyUnion[PowBlock, None]`

For this case, we can use [Python typing `Optional[SomeType]`](https://mypy.readthedocs.io/en/stable/kinds_of_types.html?highlight=Option#optional-types-and-the-none-type) to replace `PyUnion[SomeType, None]`. It has been used in [phase0 chain validator guide](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/validator.md#validator-assignments) and [DAS spec](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/das/das-core.md#data-recovery).

mkalinin:
> @mkalinin (Sorry, I just saw it!)
> 
> > `PyUnion[PowBlock, None]`
> 
> For this case, we can use [Python typing `Optional[SomeType]`](https://mypy.readthedocs.io/en/stable/kinds_of_types.html?#optional-types-and-the-none-type) to replace `PyUnion[SomeType, None]`. It has been used in [phase0 chain validator guide](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/validator.md#validator-assignments) and [DAS spec](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/das/das-core.md#data-recovery).

Thanks a lot for directions. Fixed!

djrtwo (approved):

PR #2524, opened 2021-07-17
Update remerkleable to v0.1.22: list lookup speed improvement

protolambda:
Previously remerkleable would continue to use the SSZ-typed integers internally, after checking bounds, and do type-coercions/checks on every step down in the tree, in extreme cases. This release fixes that, and should improve the eth2-specs testing speed.

Also, reminder: run `citest` locally to use CI settings (milagro BLS instead of py-ecc BLS), for faster testing.

7m8s -> 5m35s


djrtwo (approved):
fantastic find on the int inefficiencies!

PR #2528, opened 2021-07-21
require sync committee contributions to have at least one participant

### LINKED ISSUE #2527: Clarify SyncCommittee aggregation behaviour when no participants are found
On altair-devnet-0 we're getting some `SignedContributionAndProof` messages published with no aggregation_bits set. Currently Teku considers these invalid based on:

```
[REJECT] The aggregate signature is valid for the message beacon_block_root and aggregate pubkey derived from the participation info in aggregation_bits for the subcommittee specified by the contribution.subcommittee_index.
```

But we should clarify whether the correct behaviour is:

1. Publish an empty aggregate with the infinite signature (and thus verify the signature with `eth2_fast_aggregate_verify`)
2. Skip publishing the aggregate

Since there are no rewards for producing aggregates and an empty aggregate has no value, I'd suggest option 2 and just clarify that in the validator spec.

djrtwo:
Fixes #2527 

Takes a similar approach to `beacon_aggregate_and_proof`.  Requires there to be at least one participant (not necessarily the aggregator to avoid VC/BN mismatch problems) for the message to have value and be forwarded. The minimum of one participant side-steps zero sig considerations

### DIFF
--- a/specs/altair/p2p-interface.md
+++ b/specs/altair/p2p-interface.md
@@ -139,6 +139,8 @@ def get_sync_subcommittee_pubkeys(state: BeaconState, subcommittee_index: uint64
 
 - _[IGNORE]_ The contribution's slot is for the current slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance), i.e. `contribution.slot == current_slot`.
 - _[REJECT]_ The subcommittee index is in the allowed range, i.e. `contribution.subcommittee_index < SYNC_COMMITTEE_SUBNET_COUNT`.
+- _[REJECT]_ The contribution has participants --
+  that is, `len(set(bit for bit in contribution.aggregation_bits if bit == True)) >= 1`.
 - _[REJECT]_ `contribution_and_proof.selection_proof` selects the validator as an aggregator for the slot -- i.e. `is_sync_committee_aggregator(contribution_and_proof.selection_proof)` returns `True`.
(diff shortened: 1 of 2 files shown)


specs/altair/p2p-interface.md
```diff
@@ -139,6 +139,8 @@ def get_sync_subcommittee_pubkeys(state: BeaconState, subcommittee_index: uint64
 
 - _[IGNORE]_ The contribution's slot is for the current slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance), i.e. `contribution.slot == current_slot`.
 - _[REJECT]_ The subcommittee index is in the allowed range, i.e. `contribution.subcommittee_index < SYNC_COMMITTEE_SUBNET_COUNT`.
+- _[REJECT]_ The contribution has participants --
+  that is, `len(set(bit for bit in contribution.aggregation_bits if bit == True)) >= 1`.
```
ralexstokes, line 143: i think this code sample as written is a little more opaque than it needs to be

```suggestion
  that is, `any(contribution.aggregation_bits)`.
```

ralexstokes: or perhaps:
```suggestion
  that is, `True in contribution.aggregation_bits`.
```

ralexstokes (approved):
looks good! agree this kind of logic should be present to harden sync committee gossip

left a comment on simplifying the validation

PR #2530, opened 2021-07-22
Rebase The Merge onto Altair base functionality

protolambda:
No functional changes to the Merge itself, but extending the Altair specification now that it is in beta and looking stable.

- Update Merge Beacon doc
  - Update type definitions to full (fixes linting error for `state.slot` type, also does not mix common types of different modules)
  - Update extended processing functions to call Altair
- Update Merge genesis state creation
- Update validator and other doc references to previous forks
- Update test utils to handle sequence
- Update `helpers/constants.py`
- Update fork tests to start with Altair state when forking to Merge
- Update `setup.py` to generate code
- Update sharding beacon doc for consistency

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -447,7 +447,7 @@ class AltairSpecBuilder(Phase0SpecBuilder):
     @classmethod
     def imports(cls, preset_name: str) -> str:
         return super().imports(preset_name) + '\n' + f'''
-from typing import NewType, Union
+from typing import NewType, Union as PyUnion
 
 from eth2spec.phase0 import {preset_name} as phase0
 from eth2spec.utils.ssz.ssz_typing import Path
@@ -463,7 +463,7 @@ GeneralizedIndex = NewType('GeneralizedIndex', int)
     @classmethod
     def sundry_functions(cls) -> str:
         return super().sundry_functions() + '\n\n' + '''
-def get_generalized_index(ssz_class: Any, *path: Sequence[Union[int, SSZVariableName]]) -> GeneralizedIndex:
+def get_generalized_index(ssz_class: Any, *path: Sequence[PyUnion[int, SSZVariableName]]) -> GeneralizedIndex:
     ssz_path = Path(ssz_class)
     for item in path:
         ssz_path = ssz_path / item
@@ -487,14 +487,14 @@ def get_generalized_index(ssz_class: Any, *path: Sequence[Union[int, SSZVariable
 #
 # MergeSpecBuilder
 #
-class MergeSpecBuilder(Phase0SpecBuilder):
+class MergeSpecBuilder(AltairSpecBuilder):
     fork: str = MERGE
 
     @classmethod
     def imports(cls, preset_name: str):
         return super().imports(preset_name) + f'''
 from typing import Protocol
-from eth2spec.phase0 import {preset_name} as phase0
+from eth2spec.altair import {preset_name} as altair
 from eth2spec.utils.ssz.ssz_typing import Bytes20, ByteList, ByteVector, uint256, Union
 '''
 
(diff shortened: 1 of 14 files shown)


setup.py
```diff
@@ -844,38 +844,30 @@ def finalize_options(self):
         if len(self.md_doc_paths) == 0:
             print("no paths were specified, using default markdown file paths for pyspec"
                   " build (spec fork: %s)" % self.spec_fork)
-            if self.spec_fork == PHASE0:
+            if self.spec_fork in (PHASE0, ALTAIR, MERGE):
```
djrtwo, line 847: much cleaner 🙏

specs/merge/fork.md
```diff
@@ -78,14 +81,16 @@ def upgrade_to_merge(pre: phase0.BeaconState) -> BeaconState:
...
+        previous_epoch_participation=pre.previous_epoch_participation,
+        current_epoch_participation=pre.current_epoch_participation,
         # Finality
         justification_bits=pre.justification_bits,
         previous_justified_checkpoint=pre.previous_justified_checkpoint,
         current_justified_checkpoint=pre.current_justified_checkpoint,
         finalized_checkpoint=pre.finalized_checkpoint,
+        # Inactivity
+        inactivity_scores=pre.inactivity_scores,
         # Execution-layer
```
djrtwo, line 94: Need to carry forward `current` and `previous_sync_committee`

djrtwo (commented):
nice! one small fix

specs/merge/beacon-chain.md
```diff
@@ -69,15 +67,59 @@ This patch adds transaction execution to the beacon chain as part of the Merge f
 #### `BeaconBlockBody`
 
 ```python
-class BeaconBlockBody(phase0.BeaconBlockBody):
+class BeaconBlockBody(Container):
```
hwwhww, line 70: Curious, why `class BeaconBlockBody(altair.BeaconBlockBody)` doesn't work here?

specs/merge/beacon-chain.md
```diff
@@ -69,15 +67,59 @@ This patch adds transaction execution to the beacon chain as part of the Merge f
...
+    sync_aggregate: SyncAggregate
     # Execution
     execution_payload: ExecutionPayload  # [New in Merge]
 ```
 
 #### `BeaconState`
 
 ```python
-class BeaconState(phase0.BeaconState):
+class BeaconState(Container):
```
hwwhww, line 88: ditto, why `class BeaconState(altair.BeaconState)` doesn't work?

protolambda: Without this change, the linter cannot figure out the type of `state.slot` well enough:

```
. venv/bin/activate; cd ./tests/core/pyspec; \
flake8  --config /home/proto/projects/eth2.0-specs/linter.ini ./eth2spec \
&& mypy --config-file /home/proto/projects/eth2.0-specs/linter.ini -p eth2spec.phase0 -p eth2spec.altair -p eth2spec.merge
eth2spec/merge/minimal.py:2238: error: Value of type variable "SupportsLessThanT" of "max" cannot be "object"
eth2spec/merge/mainnet.py:2238: error: Value of type variable "SupportsLessThanT" of "max" cannot be "object"
Found 2 errors in 2 files (checked 9 source files)
make: *** [Makefile:127: lint] Error 1
```

The python line `2238` (in `process_sync_aggregate`):

```
    previous_slot = max(state.slot, Slot(1)) - Slot(1)
```

I think the issues is that the extended type is in a different module, which declares a different `Slot` type. And  mypy cannot handle that with `max`, even though the Slot types both inherit `int`, and interchangeable everywhere else.

Annotating all fields to use types of the merge model works, but at that point you may as well not extend the base-type.

In the long-term, I would prefer to migrate over the eth2 specs to a python-first approach like the new eth1-specs, and generate the documentation from the code. That way we can get rid of code-generation, and have a single type of everything, and using mypy may get a lot easier.

specs/merge/beacon-chain.md
protolambda: See `BeaconState` comment below. I did the same with this type for consistency.

djrtwo (approved):
looks good to me. not too bad

specs/merge/beacon-chain.md
hwwhww: > In the long-term, I would prefer to migrate over the eth2 specs to a python-first approach like the new eth1-specs, and generate the documentation from the code.

Agreed!

> That way we can get rid of code-generation, and have a single type of everything, and using mypy may get a lot easier.

By single type, do you mean like `BeaconStateType` or a more general abstract `SszType`? IIUC, to address the typing problem, I guess mypy-izing `remarkable` is required before the overhauling work?

---

Another thing, I found that when using remarkable inheritance, I always worry about if we wrongly inherited the old container where the fields have been updated in the new fork. e.g., *IF* `BeaconBlockHeader` fields are changed. Writing down all the fields may be more reliable in this context. 😅 👍

hwwhww (approved):
Nice!

specs/merge/beacon-chain.md
protolambda: > By single type, do you mean like ...

Single definition. E.g. we repeat common types like `Slot`, `DepositData` etc. in the `phase0`, `altair`, and `merge` packages. 

> I guess mypy-izing remarkable is required before the overhauling work?

I think it's more of a type conflict than an annotation problem with mypy. Suggestions to improve typing in remerkleable are welcome, but that can go into a separate PR.

> I always worry about if we wrongly inherited the old container ...

It should leave fields in the same order, but override the type. Any additional fields are appended at the end. But agreed, repeating the fields, and not forcing class inheritance structures into the spec, is better.

mkalinin (approved):
LGTM! 👍

Issue #2541, opened 2021-08-04
validator life cycle spec tests through epoch processing


terencechain:
@prestonvanloon and I found a bug in Prysm's Altair implementation, where in `process_inactivity_updates`, we implemented it as:
```python
def process_inactivity_updates(state: BeaconState) -> None:
    for index in range(len(state.validators))
```

The correct implementation should have been:
```python
def process_inactivity_updates(state: BeaconState) -> None:
    for index in get_eligible_validator_indices(state)
```

I noted @ralexstokes thinking this would be a good missing test candidate to add in for [inactivity_updates](https://github.com/ethereum/eth2.0-spec-tests/tree/master/tests/mainnet/altair/epoch_processing/inactivity_updates/pyspec_tests). Alex brought up great feedbacks where a new general type of spec test across validator life cycle across fork may be better suited for this. Opening this issue to track the latest discussion

leolara:
I am closing this issue because it seems stale. Please, do not hesitate to reopen it if this is a mistake

PR #2542, opened 2021-08-04
Add test for exited validators during inactivity leak

ralexstokes:
Adds a spec test for the concrete bug found while debugging #2541.

Does not add any extensive validator lifecycle testing as referenced in that issue but the debugging did reveal a hole in our spec testing coverage which this PR aims to fix.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/helpers/voluntary_exits.py
+++ b/tests/core/pyspec/eth2spec/test/helpers/voluntary_exits.py
@@ -1,3 +1,4 @@
+from random import Random
 from eth2spec.utils import bls
 from eth2spec.test.helpers.keys import privkeys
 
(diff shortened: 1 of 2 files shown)


tests/core/pyspec/eth2spec/test/helpers/voluntary_exits.py
```diff
@@ -23,3 +24,21 @@ def sign_voluntary_exit(spec, state, voluntary_exit, privkey):
...
+
+#
+# Helpers for applying effects of a voluntary exit
+#
+def get_exited_validators(spec, state):
+    current_epoch = spec.get_current_epoch(state)
+    return [index for (index, validator) in enumerate(state.validators) if validator.exit_epoch <= current_epoch]
+
+
+def exit_validators(spec, state, validator_count, rng=None):
```
djrtwo, line 37: Is the following a poor pattern for some reason?
```suggestion
def exit_validators(spec, state, validator_count, rng=Random(1337)):
```

djrtwo (commented):
a couple of nits and a question

tests/core/pyspec/eth2spec/test/helpers/voluntary_exits.py
ralexstokes: i do it this way out of habit to avoid issues like this: https://stackoverflow.com/a/4841873

i think this will be fine though as the value is not mutable in a sense that would cause a problem like this

djrtwo: fine either way. We use `rng=Random(num)` all over the place in the tests I beleive so just wanted  to make sure I wasn't doing something terrible

djrtwo (approved):
thanks!

PR #2548, opened 2021-08-12
Extend base_fee_per_gas type to uint256

mkalinin:
As it's been well spotted by @protolambda `base_fee_per_gas: uint64` is prone to overflow. Suppose `base_fee_per_gas = 32 Gwei`, miners will need to produce full blocks for ~150 in a row to overflow this value and eventually brick the Merge.

Thus, data structures, constants and arithmetics is update to use extended `uint256` type for `base_fee_per_gas`.

**UPD**
This attack requires `924,730 ETH` to be burned which is of the very high cost taking current ETH price. If the starting point of `base_fee_per_gas = 1024 Gwei` then the cost drops to `81,947 ETH`.

### DIFF
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -65,7 +65,7 @@ This patch adds transaction execution to the beacon chain as part of the Merge f
 | `BYTES_PER_LOGS_BLOOM` | `uint64(2**8)` (= 256) |
 | `GAS_LIMIT_DENOMINATOR` | `uint64(2**10)` (= 1,024) |
 | `MIN_GAS_LIMIT` | `uint64(5000)` (= 5,000) |
-| `BASE_FEE_MAX_CHANGE_DENOMINATOR` | `uint64(2**3)` (= 8) |
+| `BASE_FEE_MAX_CHANGE_DENOMINATOR` | `uint256(2**3)` (= 8) |
 | `ELASTICITY_MULTIPLIER` | `uint64(2**1)` (= 2) |
 
 ## Configuration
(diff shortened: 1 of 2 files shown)


djrtwo:
Another option is to call it a `bytes32` and to not do any validation on the execution-layer base fee within the consensus layer. This would remove the uint256 computational requirement and keep the validations where they currently are

protolambda:
Related issue: in the sharding spec we have base-fees as well, and cannot push those in execution layer. Either shards will have an upper bound and/or lower precision reduction on fee fields, or we introduce uint256 there anyway.

Arachnid:
2^64 wei is 18 ether. Filling a 30M gas block at that base fee would require over 500 million ether. Surely this is a nonissue? Everyone will run out of ether way before hitting this.

protolambda:
@Arachnid there are ways to decrease that 30M gas to just 21000 (minimum for a single tx to fill a block with), and then there's a 5000 gas limit which avoids a 0 limit, and thus avoids base-fees being manipulated upwards for free (while this 5000 minimum is way older than EIP 1559). You're right that it's a silly issue, but not entirely a non-issue until you consider ways like this around it.

protolambda:
Another thing to consider is that as attacker you don't even need the full 64 bits base fee the `base_fee_per_gas` is multiplied with other uint64 numbers. So cost may decrease by another 1000x. Eth1 uses big-nums for these computations, here we use uint64, which will overflow / error if safe-math.

mkalinin:
Closing in favour of https://github.com/ethereum/eth2.0-specs/pull/2550

PR #2550, opened 2021-08-13
Change base_fee_per_gas type to Bytes32

mkalinin:
A solution alternative to https://github.com/ethereum/eth2.0-specs/pull/2548

What's done:
- `base_fee_per_gas` type changed to `Bytes32`
- `base_fee_per_gas` validation and related computations are removed from the beacon chain

### DIFF
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -34,7 +34,6 @@
   - [Block processing](#block-processing)
   - [Execution payload processing](#execution-payload-processing)
     - [`is_valid_gas_limit`](#is_valid_gas_limit)
-    - [`compute_base_fee_per_gas`](#compute_base_fee_per_gas)
     - [`process_execution_payload`](#process_execution_payload)
 - [Testing](#testing)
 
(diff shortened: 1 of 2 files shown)


specs/merge/beacon-chain.md
```diff
@@ -181,7 +178,7 @@ class ExecutionPayloadHeader(Container):
     gas_limit: uint64
     gas_used: uint64
     timestamp: uint64
-    base_fee_per_gas: uint64
+    base_fee_per_gas: Bytes32
```
protolambda, line 181: Little endian or big endian?

mkalinin: This is a good question. For better compatibility with the future Consensus API, it probably should be big-endian, it would allow to avoid reversing bits each time when execution payload is either sent to or received from execution client. But it might be controversial to the beacon chain spec which is little-endian in general. What are your thoughts?

mkalinin: But if we then convert this field to `uint256` using little-ending from now should make it forward-compatible.

protolambda: Both have pros/cons, I am ok with either, let's just make sure to spec how to interpret this 32 bytes.

potuz: >Little endian or big endian?

Seems it's chosen big endian here. If you decide to go with little endian remember to change back `GENESIS_BASE_FEE_PER_GAS`

OhArthits (approved):

mkalinin:
Changed `base_fee_per_gas` serialization to little-endian, the corresponding constant is also updated.

djrtwo (approved):

protolambda (approved):
Little-endian is nice for `uint256` SSZ compat (if we later decide to compute or use in the beacon chain), and simple enough to convert to big-endian as part of the payload-decoding on the execution-client side. LGTM

PR #2556, opened 2021-08-18
Add randomized testing for altair inactivity scores

ralexstokes:
The existing test suite was missing some coverage around the processing of inactivity scores, particularly for validators who are exited (and should have their inactivity scores frozen in the epoch right before the `exit_epoch`). This PR extends the test coverage

### DIFF
--- a/tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_inactivity_updates.py
+++ b/tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_inactivity_updates.py
@@ -18,6 +18,7 @@ from eth2spec.test.helpers.epoch_processing import (
 from eth2spec.test.helpers.random import (
     randomize_attestation_participation,
     randomize_previous_epoch_participation,
+    randomize_state,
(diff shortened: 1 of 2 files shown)


tests/core/pyspec/eth2spec/test/helpers/random.py
```diff
@@ -15,6 +15,8 @@ def set_some_new_deposits(spec, state, rng):
             continue
         if rng.randrange(num_validators) < num_validators // 10:
             mock_deposit(spec, state, index)
+            if is_post_altair(spec):
+                state.inactivity_scores[index] = 0
```
ralexstokes, line 19: i have to wonder how scalable this approach is to check which features a fork supports and then adjusting the functionality of a helper function (e.g. `mock_deposit`) but this same pattern is used for the attestation participation randomization here so i followed it for now...

djrtwo: agree

djrtwo (approved):

PR #2558, opened 2021-08-19
sharding eip-1559-like behavior, with rolling-window adjustments

protolambda:
Changes:
- Fee fields are for fixed-length shard data anyway (unlike dynamic gas count in eth1), account for it as total instead of per sample.
- The above enables `Gwei` usage for these fields, without losing as much accuracy for low-fee transactions.
- Base fees are tracked per shard, to avoid 1 fat shard and 63 slim shards from balancing out. Also increases predictability.
- Rolling window base-fee updates. A single base fee update is still very similar, but instead of processing once and moving to the next slot, we process with a lower maximum change, and repeat it for a few slots.
  - When there is a beacon gap slot, fees are stable, as the processing func is not called
  - When there is a shard gap slot, it accounts as 0, until it becomes non-zero with e.g. late inclusion. The effect of inclusion latency is minimized by spreading the adjustments, yet the base-fee keeps adjusting per slot, to not have to wait e.g. a whole epoch for adjustments.
  - When a beacon slot gap results in a late shard header, it is just one less adjustment step. Making the shard header fees less volatile, yet accounted for.
  - The exact quotient computation needs some work, just a rough guess how it could be like, the old one was outdated (per epoch, global fee instead of per-shard), corrections welcome. Might chart it out later, it's a starting point.
- Small bonus shard update: store the committee weight along with the voters weight, so we do not have to compute/cache committees and balances when checking confirmation progress. Especially useful for faster API service of shard data confs.

### DIFF
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -752,8 +752,9 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
 
     # Initialize the pending header
     index = compute_committee_index_from_shard(state, slot, shard)
-    committee_length = len(get_beacon_committee(state, slot, index))
-    initial_votes = Bitlist[MAX_VALIDATORS_PER_COMMITTEE]([0] * committee_length)
+    full_committee = get_beacon_committee(state, slot, index)
+    initial_votes = Bitlist[MAX_VALIDATORS_PER_COMMITTEE]([0] * len(full_committee))
+    max_weight = sum(state.validators[index].effective_balance for index in full_committee)
     pending_header = PendingShardHeader(
         attested=AttestedDataCommitment(
             commitment=body_summary.commitment,
@@ -762,6 +763,7 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
         ),
         votes=initial_votes,
         weight=0,
+        max_weight=max_weight,
         update_slot=state.slot,
     )
 
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -93,7 +93,6 @@ We define the following Python custom types for type hinting and readability:
 | `BLSCommitment` | `Bytes48` | A G1 curve point |
 | `BLSPoint` | `uint256` | A number `x` in the range `0 <= x < MODULUS` |
 | `BuilderIndex` | `uint64` | Builder registry index |
-| `Wei` | `uint256` | An amount in Wei |
 
 ## Constants
 
@@ -225,7 +224,7 @@ class BeaconState(merge.BeaconState):
     # A ring buffer of the latest slots, with information per active shard.
     shard_buffer: Vector[List[ShardWork, MAX_SHARDS], SHARD_STATE_MEMORY_SLOTS]
     # Each shard tracks a base-fee to continually re-adjust towards target data capacity
-    shard_sample_prices: List[Wei, MAX_SHARDS]
+    shard_sample_prices: List[Gwei, MAX_SHARDS]
 ```
 
 ## New containers
@@ -277,7 +276,7 @@ class ShardBlobBody(Container):
     # Latest block root of the Beacon Chain, before shard_blob.slot
     beacon_block_root: Root
     # fee payment fields (EIP 1559 like)
-    max_priority_fee: GWei
+    max_priority_fee: Gwei
     max_fee: Gwei
 ```
 
@@ -299,7 +298,7 @@ class ShardBlobBodySummary(Container):
     # Latest block root of the Beacon Chain, before shard_blob.slot
     beacon_block_root: Root
     # fee payment fields (EIP 1559 like)
-    max_priority_fee: GWei
+    max_priority_fee: Gwei
     max_fee: Gwei
 ```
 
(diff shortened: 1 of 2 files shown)


specs/sharding/beacon-chain.md
```diff
@@ -810,6 +812,38 @@ def process_shard_proposer_slashing(state: BeaconState, proposer_slashing: Shard
...
+          committee_work = state.shard_buffer[buffer_index][shard_index]
+  
+          samples_count = 0
+          if committee_work.status.selector == SHARD_WORK_CONFIRMED:
+              attested: AttestedDataCommitment = committee_work.status.value
+              samples_count = attested.commitment.samples_count
+          elif committee_work.status.selector == SHARD_WORK_PENDING:
+              pending: List[PendingShardHeader, MAX_SHARD_HEADERS_PER_SHARD] = committee_work.status.value
+              # Conflicting headers add up to relieve network stress.
+              samples_count = sum(header.attested.commitment.samples_count for header in pending)
```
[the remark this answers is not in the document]
Nashatyrev: Also not sure if the price should be adjusted for `PENDING` headers at all.
What if just make the price adjustment once per blob right upon its confirmation? From my perspective that could look simpler and more straightforward

specs/sharding/beacon-chain.md
```diff
@@ -437,13 +441,15 @@ def compute_previous_slot(slot: Slot) -> Slot:
 #### `compute_updated_sample_price`
 
 ```python
-def compute_updated_sample_price(prev_price: Gwei, samples_length: uint64, active_shards: uint64) -> Gwei:
-    adjustment_quotient = active_shards * SLOTS_PER_EPOCH * SAMPLE_PRICE_ADJUSTMENT_COEFFICIENT
+def compute_updated_sample_price(prev_price: Wei, samples_length: uint64) -> Wei:
+    # Per epoch an adjustment is targeted, but since capacity may be delayed,
+    #  it is split into SLOTS_PER_BASE_FEE_UPDATES smaller adjustments.
+    adjustment_quotient = Wei(SLOTS_PER_EPOCH * SLOTS_PER_BASE_FEE_UPDATES * SAMPLE_PRICE_ADJUSTMENT_COEFFICIENT)
```
Nashatyrev, line 447: Quotient seems pretty accurate to me

Nashatyrev (approved):
Generally looks good to me 👍 
I would just may be think about simplification of price adjustments

specs/sharding/beacon-chain.md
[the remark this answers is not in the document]
protolambda: If a proposer makes multiple proposals, then there is capacity of 4 of them (this pending-headers list is specific to a single shard+slot combination), to try keep attestations useful, even if the proposer messes up / attacks.  But 4 proposals also means more network stress, so we sum up the worst-case claimed capacity.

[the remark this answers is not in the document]
protolambda: Also: base fee, capacity, sharding performance and confirmations are very related. If the system is unhealthy, I prefer to smooth into adjustments based on the latest view, instead of waiting for confirmation.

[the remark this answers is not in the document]
Nashatyrev: I mean that function `compute_updated_sample_price()` expects `samples_count <= MAX_SAMPLES_PER_BLOB` else it would not satisfy adjustment requirement: 
```
SAMPLE_PRICE_ADJUSTMENT_COEFFICIENT | Sample price may decrease/increase by at most exp(1 / this value) per epoch
```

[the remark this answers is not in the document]
Nashatyrev: BTW, may be it makes sense to discuss again the necessity of handling `>1` headers from the same proposer? 

Producing `>1` headers is definitely either a malicious of malfunctioning behavior. It supposed to be a pretty exceptional event, because of slashing. I still couldn't get how accounting of such cases may help with network liveness.

Restricting sate to have just one pending header per slot+shard would simplify many things IMO

specs/sharding/beacon-chain.md
```diff
@@ -221,7 +224,8 @@ class BeaconState(merge.BeaconState):
     blob_builder_balances: List[Gwei, BLOB_BUILDER_REGISTRY_LIMIT]
     # A ring buffer of the latest slots, with information per active shard.
     shard_buffer: Vector[List[ShardWork, MAX_SHARDS], SHARD_STATE_MEMORY_SLOTS]
-    shard_sample_price: uint64
+    # Each shard tracks a base-fee to continually re-adjust towards target data capacity
+    shard_sample_prices: List[Wei, MAX_SHARDS]
```
djrtwo, line 228: anyway around the uint256?

specs/sharding/beacon-chain.md
```diff
@@ -273,9 +277,8 @@ class ShardBlobBody(Container):
     # Latest block root of the Beacon Chain, before shard_blob.slot
     beacon_block_root: Root
     # fee payment fields (EIP 1559 like)
-    # TODO: express in MWei instead?
-    max_priority_fee_per_sample: Gwei
-    max_fee_per_sample: Gwei
+    max_priority_fee: GWei
```
djrtwo, line 280:
```suggestion
    max_priority_fee: Gwei
```

specs/sharding/beacon-chain.md
```diff
@@ -296,9 +299,8 @@ class ShardBlobBodySummary(Container):
...
```
djrtwo, line 302:
```suggestion
    max_priority_fee: Gwei
```

specs/sharding/beacon-chain.md
djrtwo: Seems unnatural to force this to be wei

specs/sharding/beacon-chain.md
```diff
@@ -730,19 +732,17 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
...
-    # TODO: overflows, need bigger int type
-    max_fee = body_summary.max_fee_per_sample * samples
 
     # Builder must have sufficient balance, even if max_fee is not completely utilized
-    assert state.blob_builder_balances[header.builder_index] >= max_fee
+    assert state.blob_builder_balances[header.builder_index] >= body_summary.max_fee
 
-    base_fee = state.shard_sample_price * samples
+    # Round down, < 1 gwei is ok to not burn.
+    base_fee = Gwei((state.shard_sample_prices[shard] * Wei(samples)) / Wei(1000_000_000))
```
djrtwo, line 740: `Wei(samples)` seems semantically incorrect

djrtwo (commented):
The smoothing approach makes sense to me.

I need to go over this a bit more deeply but wanted to signal general support

specs/sharding/beacon-chain.md
protolambda: In some places, yes, by representing the fee for all samples. My findings were that's not the case here, or we lose precision in base fee per sample. There are only a few computations with these, it's not critical to performance.

specs/sharding/beacon-chain.md
protolambda: Type checking. Alternatively we  use a bare uint256 first, and then convert to Wei at the end.

protolambda:
Switching to gwei in base fee computation, over/underflows seem ok:
```
    Max prev_price is MAX_SAMPLE_PRICE, 33 bits.
    Max sample_delta is MAX_SAMPLES_PER_BLOB-TARGET_SAMPLES_PER_BLOB, 10 bits
    Max fee_delta_factor is 33 + 10 = 43 bits
    adjustment_quotient is 5 + 2 + 3 = 10 bits
    TARGET_SAMPLES_PER_BLOB = 10 bits
    fee_delta is 43-10-10=23 bits
```
1 gwei worst case precision difference, per sample, per shard, per year: `1 * 1024 * 64 * 365 * 24 * 60 * 60 / 12 / 1e9 = ~ 172 ETH` per year. Using `uint64` everywhere seems to be worth that.

specs/sharding/beacon-chain.md
protolambda: Switched to Gwei :tada:

specs/sharding/beacon-chain.md
```diff
@@ -366,6 +368,8 @@ class PendingShardHeader(Container):
     votes: Bitlist[MAX_VALIDATORS_PER_COMMITTEE]
     # Sum of effective balances of votes
     weight: Gwei
+    # Sum of effective balances of full committee
+    max_weight: Gwei
```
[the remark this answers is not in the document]
protolambda: Switched to Gwei fully now.

specs/sharding/beacon-chain.md
Nashatyrev: Sorry, I would continue arguing here :) 

To be strict the operation `Wei * Wei` doesn't make any sense as it would result in a weird unit of measurement `Wei^2`. 
`Wei` type (aka additive group) should only have operations ` + Wei`, `- Wei` and scalar multiplications ` * int`, ` / int`, (resulting in `Wei` value). 
> Alternatively we use a bare uint256 first, and then convert to Wei at the end.

I believe that `Wei(123)` should be though of as short notice of `1_Wey * 123`, where `1_Wei` could be thought of as 'group generator' 

May be we can overload arithmetic operations for `Wei` (and `GWei`) in the pyspec for type safety? Kind of: 
```
def __add__(weis: Wei) Wei
def __mul__(scalar: uint64): Wei
```

protolambda: - Wei is removed, not a concern there anymore
- `A * B` already coerces the output to the type of A, or to B if A is a bare int, and checks the integer type size. But the mypy type checker has trouble with it I think, we had to change other similar cases in the pyspec-enabled specs.

Changed the code to `Gwei(uint64(state.shard_sample_prices[shard]) * samples)`, just so the mypy thing works

PR #2560, opened 2021-08-19
Broad-spectrum randomized block tests

ralexstokes:
Currently WIP / draft.

Expands spec test coverage via generation of various randomized states/blocks and verifying behavior around important boundaries/events. Provides intermediate coverage broader than "point" testing of individual spec functions but less broad than e.g. fuzzing efforts.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/phase0/sanity/test_blocks_random.py
+++ b/tests/core/pyspec/eth2spec/test/phase0/sanity/test_blocks_random.py
@@ -1,0 +1,67 @@
+from eth2spec.test.helpers.state import (
+    next_epoch,
+    next_slot,
+)
+from eth2spec.test.context import (
+    with_all_phases,
+    spec_state_test,
+)
+
+
+def generate_randomized_scenarios():
+    # TODO: WIP schema
+    return {
+        # ("randomize_state", "ensure all validator states present: pending/deposited, activated, exited, slashed"),
+        # ("randomized balances", "ensure distribution of bals"),
+        # ("transition to leak if not already, maybe", "assert is or is not leaking"),
+        "setup": [],
+        "epochs_to_skip": 0, # 0, 1, 2, N, EPOCHS_TO_INACTIVITY_LEAK,
+        "slots_to_skip": 0,  # 0, 1, 2, N, SLOTS_PER_EPOCH - 1,
+        "transitions": [ # TODO: consider large numbers of blocks, load on generated data
+            {
+                "block_producer": lambda spec, state: spec.SignedBeaconBlock(),
+                "epochs_to_skip": 0, # 0, 1, 2, N, EPOCHS_TO_INACTIVITY_LEAK,
+                "slots_to_skip": 0,  # 0, 1, 2, N, SLOTS_PER_EPOCH - 1,
+            }
+        ],
+    }
+
+
+def id_from_scenario(test_description):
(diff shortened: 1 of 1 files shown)


tests/core/pyspec/eth2spec/test/utils/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+    with_meta_tags,
+    build_transition_test,
+)
+
+
+__all__ = [  # avoid "unused import" lint error
+    "vector_test",
+    "with_meta_tags",
+    "build_transition_test",
+]
```
ralexstokes, line 12: moved `utils.py` as the one file to a module `utils`, that re-exports the existing stuff in this `__init__.py` and adds the helper code for the random tests into `utils/random.py`

tests/generators/random/generate.py
```diff
@@ -0,0 +1,245 @@
...
+    epoch_transition,
+    slot_transition,
+    transition_with_random_block,
+    transition_to_leaking,
+    transition_without_leak,
+)
+from eth2spec.test.helpers.constants import PHASE0, ALTAIR
+
+
+# Ensure this many blocks are present in *each* randomized scenario
```
djrtwo, line 35: This is *precisely* this many blocks, rather than *at least* this many, right?

Variance here might be marginally valuable

tests/generators/random/generate.py
```diff
@@ -0,0 +1,245 @@
...
+        slot_transition(penultimate_slot_in_epoch),
+    )
+    # and produce a block...
+    blocks_set = (
+        transition_with_random_block(block_randomizer),
+    )
+    # build a set of block transitions from combinations of sub-transitions
+    transitions_generator = (
+        itertools.product(epochs_set, slots_set, blocks_set) for
+        _ in range(BLOCK_TRANSITIONS_COUNT)
```
djrtwo, line 112: The combinatorics didn't work as I immediately expected them too.

We always couple like-randomness with each other (e.g. two sets of (no epoch, go to pen-ultimate slot, random block)) rather than mixing the different pre-block modifications (e.g. (no epoch, penultimate slot, random block) + (1 epoch, random slot, random block)  )

Can you explain your thinking about here? On the one hand, we could do many more combinations but the number would explode. Also, I suppose that the magic is primarily in fussing with slots/epoch and then random blocks rather than fussing with slots/epochs in different ways between random blocks..

djrtwo (commented):
Fantastic! 

High level looks great. some quetsions and minor suggestions throughout

tests/core/pyspec/eth2spec/test/helpers/multi_operations.py
```diff
@@ -57,8 +65,14 @@ def get_random_proposer_slashings(spec, state, rng):
 
 
 def get_random_attester_slashings(spec, state, rng):
-    num_slashings = rng.randrange(spec.MAX_ATTESTER_SLASHINGS)
-    indices = spec.get_active_validator_indices(state, spec.get_current_epoch(state)).copy()
+    # ensure at least one attester slashing, the max count
+    # is small so not much room for random inclusion
+    num_slashings = max(1, rng.randrange(spec.MAX_ATTESTER_SLASHINGS))
```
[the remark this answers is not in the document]
djrtwo: This pattern was used in a few more places, suggest the `randrange` (no `max`) usage througohut

PR #2563, opened 2021-08-23
Fix typos in sharding/beacon-chain.md, which break setup.py:get_spec processing

ericsson49:
Minor typos fixed


hwwhww (approved):
lgtm 👍 

/cc @protolambda is it time to add sharding spec builder to setup.py?

PR #2565, opened 2021-08-23
Fix typos in `sharding/beacon-chain.md`

ericsson49:
Minor typos fixes:
- `get_shard_proposer_index`: variable `state` is used to reffer to `beacon_state` parameter. Renamed all uses to `state` to be consistent with the rest of the specs.
- `get_start_shard` refers to `_slot` instead of `slot`


djrtwo (approved):
thank you!

PR #2575, opened 2021-08-30
Change transition_td to terminal_td in the merge spec

mkalinin:
A slight update in the Merge terminology, `transition_total_difficulty` is replaced with the newer term `terminal_total_difficulty` with the corresponding renaming all over the spec


djrtwo (approved):
lgtm!

PR #2576, opened 2021-08-30
Allow spec tests author to batch tests under one handler name

ralexstokes:
Request from implementation teams to batch "like-operations" when generating consensus spec tests.

The test generation infra followed the file tree of the test repo to determine the organization of the output test data.

This PR allows a test author to batch tests under one handler name, regardless of file tree layout, by providing a list of "like" test modules, rather than just a single `str` of the module name.

This PR also adds a "collect" mode to the test generation infra (like the `pytest` collection phase), used to verify we pulled the same number of tests with this change.


djrtwo (commented):
should we also group the new random tests with the other block tests?

djrtwo (approved):
looks good. spoke with @ralexstokes out of band and agreed there is value in splitting out "sanity" and "random" block tests especially for new clients coming online

PR #2577, opened 2021-08-31
Fix fork choice `on_block` tests and update test format

hwwhww:
1. Update test format of `checks`: add checkpoint epoch number and rephrase them in dict. Example:
````yaml
- checks:
    time: 192
    head: {slot: 32, root: '0xdaa1d49d57594ced0c35688a6da133abb086d191a2ebdfd736fad95299325aeb'}
    justified_checkpoint: {epoch: 3, root: '0xc25faab4acab38d3560864ca01e4d5cc4dc2cd473da053fbc03c2669143a2de4'}
    finalized_checkpoint: {epoch: 2, root: '0x40d32d6283ec11c53317a46808bc88f55657d93b95a1af920403187accf48f4f'}
    best_justified_checkpoint: {epoch: 3, root: '0xc25faab4acab38d3560864ca01e4d5cc4dc2cd473da053fbc03c2669143a2de4'}
````
2. Fix `test_new_justified_is_later_than_store_justified`: there was a missing `tick` step.
    - Thanks @ajsutton for pointing it out!
3. Update `test_new_finalized_slot_is_justified_checkpoint_ancestor`: remove the useless `tick` step.
4. Rewrite `test_on_block_finalized_skip_slots_not_in_skip_chain` and `test_on_block_finalized_skip_slots`:
    - `test_on_block_finalized_skip_slots_not_in_skip_chain` used non-genesis anchor state and block to initialize fork choice store, which was hacky and having potential issues (#2566 explained it)
    - I rewrote it to build a store from genesis. The test scenario is similar to #1579.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
+++ b/tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
@@ -210,6 +210,9 @@ def test_on_block_finalized_skip_slots(spec, state):
     # Skip the rest slots of epoch 1 and the first slot of epoch 2
     next_slots(spec, state, spec.SLOTS_PER_EPOCH)
 
+    # The state after the skipped slots
+    target_state = state.copy()
+
     # Fill epoch 3 and 4
     for _ in range(2):
         state, store, _ = yield from apply_next_epoch_with_attestations(
@@ -223,8 +226,8 @@ def test_on_block_finalized_skip_slots(spec, state):
 
     # Now build a block at later slot than finalized *epoch*
     # Includes finalized block in chain and the skipped slots
-    block = build_empty_block_for_next_slot(spec, state)
-    signed_block = state_transition_and_sign_block(spec, state, block)
+    block = build_empty_block_for_next_slot(spec, target_state)
+    signed_block = state_transition_and_sign_block(spec, target_state, block)
     yield from tick_and_add_block(spec, store, signed_block, test_steps)
 
     yield 'steps', test_steps
--- a/tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
+++ b/tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
@@ -218,7 +218,7 @@ def test_on_block_finalized_skip_slots(spec, state):
         state, store, _ = yield from apply_next_epoch_with_attestations(
             spec, state, store, True, True, test_steps=test_steps)
 
-    # Now we get finalized epoch 1, where compute_start_slot_at_epoch(1) is a skipped slot
+    # Now we get finalized epoch 2, where `compute_start_slot_at_epoch(2)` is a skipped slot
     assert state.finalized_checkpoint.epoch == store.finalized_checkpoint.epoch == 2
     assert store.finalized_checkpoint.root == spec.get_block_root(state, 1) == spec.get_block_root(state, 2)
     assert state.current_justified_checkpoint.epoch == store.justified_checkpoint.epoch == 3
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by hwwhww** [comment]
@ajsutton @tuyennhv

If you'd like to try the updated test vectors, I dumped new fork choice test vectors to: https://github.com/hwwhww/consensus-spec-tests/tree/fork-choice-pr2577 (https://github.com/ethereum/consensus-specs/commit/b23ed05eee5f499d3ff41f8356c19bd07b01e0ca version)

**Comment by hwwhww** [comment]
Update: @tuyennhv confirmed that Lodestar passed all fork choice tests against this new branch. 🙌

**Comment by ajsutton** [comment]
Sorry got distracted from this.  I've never worked out how to build the actual test vectors from the source (only got as far as running them against the spec implementation), do you have instructions for that or is there a tarball somewhere?

**Comment by hwwhww** [comment]
@ajsutton 

`fork_choice` tests only tarballs:

- [minimal.tar.gz](https://github.com/ethereum/consensus-specs/files/7096436/minimal.tar.gz)
- [mainnet.tar.gz](https://github.com/ethereum/consensus-specs/files/7096437/mainnet.tar.gz)

---

If you'd like to download from source, it required [Git LFS](https://git-lfs.github.com/) installed.
```
git lfs install
```

If LFS has been installed correctly, `git clone` should download the large files by default:
```
git clone git@github.com:hwwhww/consensus-spec-tests.git -b fork-choice-pr2577 --single-branch
cd consensus-spec-tests
```
Or you can trigger it by
```
git lfs pull
```

The test vectors will be under `consensus-spec-tests/tests` folder.

File: tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py [hunk 1]
```diff
@@ -200,21 +203,31 @@ def test_on_block_finalized_skip_slots(spec, state):
...
 
-    # Another chain
-    another_state = store.block_states[store.finalized_checkpoint.root].copy()
-    # Build block that includes the skipped slots up to finality in chain
-    block = build_empty_block(spec,
-                              another_state,
-                              spec.compute_start_slot_at_epoch(store.finalized_checkpoint.epoch) + 2)
-    block.body.graffiti = b'\x12' * 32
-    signed_block = state_transition_and_sign_block(spec, another_state, block)
+    # Now we get finalized epoch 1, where compute_start_slot_at_epoch(1) is a skipped slot
```
djrtwo, line 221: Is this comment correct?

Isn't it that `compute_start_slot_at_epoch(2)` is the skip slot?

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
```diff
@@ -200,21 +203,31 @@ def test_on_block_finalized_skip_slots(spec, state):
...
+    assert state.finalized_checkpoint.epoch == store.finalized_checkpoint.epoch == 2
+    assert store.finalized_checkpoint.root == spec.get_block_root(state, 1) == spec.get_block_root(state, 2)
+    assert state.current_justified_checkpoint.epoch == store.justified_checkpoint.epoch == 3
+    assert store.justified_checkpoint == state.current_justified_checkpoint
 
+    # Now build a block at later slot than finalized *epoch*
+    # Includes finalized block in chain and the skipped slots
+    block = build_empty_block_for_next_slot(spec, target_state)
```
djrtwo, line 229: This block is valid wrt the finalized epoch (including the skip slots), but this block isn't the head, right?

`apply_next_epoch_with_attestations` for epoch 3 and 4 will have created blocks of much higher weight and win out

So -- Does adding this block change anything? or does it catch a regression in the prior fork choice implementations?

djrtwo: and if `block` is supposed to be the head? can we check that it is against a dynamic call to `get_head()`?

djrtwo: Ah, I see. It's just checking in `on_block` can even be run anymore which would yield `valid: False` if it didn't work

djrtwo (approved):
i got a little hung up on what was actually being tested in the valid case, but figured it out!

Looks correct to me

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
hwwhww: You're right. 👍 

```suggestion
    # Now we get finalized epoch 2, where `compute_start_slot_at_epoch(2)` is a skipped slot
```

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
hwwhww: > Ah, I see. It's just checking in on_block can even be run anymore which would yield valid: False if it didn't work

Right, it is for comparing with `test_on_block_finalized_skip_slots_not_in_skip_chain` (valid: False) test case.

PR #2578, opened 2021-08-31
Fix `merge_mods` in test generators

hwwhww:
Since merge spec was rebased (#2530), make `merge_mods` base on `altair_mods` instead of `phase_0_mods`.


protolambda (approved):
LGTM

PR #2583, opened 2021-09-02
Update test_process_sync_aggregate.py

asanso:
The `test_invalid_signature_bad_domain` is about testing bad domain . There is a little copy/cargo issue with `test_invalid_signature_missing_participant`

### DIFF
--- a/tests/core/pyspec/eth2spec/test/altair/block_processing/sync_aggregate/test_process_sync_aggregate.py
+++ b/tests/core/pyspec/eth2spec/test/altair/block_processing/sync_aggregate/test_process_sync_aggregate.py
@@ -32,7 +32,7 @@ def test_invalid_signature_bad_domain(spec, state):
 
     block = build_empty_block_for_next_slot(spec, state)
     block.body.sync_aggregate = spec.SyncAggregate(
-        sync_committee_bits=committee_indices,
+        sync_committee_bits=[True] * len(committee_indices),
         sync_committee_signature=compute_aggregate_sync_committee_signature(
             spec,
             state,


tests/core/pyspec/eth2spec/test/altair/block_processing/sync_aggregate/test_process_sync_aggregate.py
```diff
@@ -29,13 +29,10 @@
...
 def test_invalid_signature_bad_domain(spec, state):
     committee_indices = compute_committee_indices(spec, state, state.current_sync_committee)
-    rng = random.Random(2020)
-    random_participant = rng.choice(committee_indices)
 
     block = build_empty_block_for_next_slot(spec, state)
-    # Exclude one participant whose signature was included.
     block.body.sync_aggregate = spec.SyncAggregate(
-        sync_committee_bits=[index != random_participant for index in committee_indices],
+        sync_committee_bits=committee_indices,
```
ralexstokes, line 35:
```suggestion
        sync_committee_bits=[True] * len(committee_indices),
```

ralexstokes (commented):
Nice catch!

I agree the spirit of this test is better expressed by just isolating the bad signature domain. We have other tests with bad bits so we are not losing any test coverage with this change.

I left one syntax suggestion -- the `sync_committee_bits` field is expecting a `Sequence[bool]` and it may work w/ a `Sequence[int]` but this would just be an idiosyncrasy of python and we can be clearing w/ explicit `bool`

djrtwo (approved):

PR #2584, opened 2021-09-02
fix test filtering on eth1 voting spec test

ralexstokes:
it seems these two tests were dropped to a configuration change and they may be our only (?) eth1 data voting spec tests.

this PR restores the tests w/ the `with_presets` decorator filter which ensures they run w/ the latest test gen machinery but still skip generating an isomorphic test case w/ much longer/bigger data requirements


djrtwo (approved):
Been missing for quite some time...

Good find @protolambda

PR #2588, opened 2021-09-09
add rewards spec test with exit in current epoch

ralexstokes:
After some inspection, it seems we do not test for exits in the current epoch.

This PR adds a test for this scenario and exposes the expected rewards computation as a spec test.


djrtwo (commented):
Test looks good! 

I think we should *also* add `current_epoch` to what can be selected as exit epoch in `exit_random_validators`. It does seem good for the randomization of state to also consider that validators could be exited this current epoch for all the other random tests out there

[side note, while in `exit_random_validators` we should make `withdrawable_epoch` "realistic"]

djrtwo (approved):

PR #2590, opened 2021-09-09
Increase the distance of the beacon block a blob may refer to

Nashatyrev:
Currently the `ShardBlob.body.beacon_block_root` may refer only head block at slot `ShardBlob.slot - 1`
This condition seems too hard:
- blob builder should have some freedom to choose which past beacon block he wants to 'fork' the blob from. 
For example if for some shard a blob was included to `block-100`, then no blobs were created for this shard during some slots, then it makes sense for the builder to refer the `block-100` for the blob at later slot (say `105`). If for example there are 2 forks at slot `104` the blob would be suitable for including to any of them 
- suppose blob builder/proposer when creating a blob transaction/header for slot `100` don't see the `block-99` and produces a blob referring `block-98`. Then a late `block-99` is finally published and accepted by network. The published blob header then couldn't be included 

Constant `MAX_BLOB_BLOCK_ROOT_DISTANCE = 32` was chosen kind of arbitrary. While it could be as large as `SLOTS_PER_HISTORICAL_ROOT` it probably shouldn't be too large to not burden the nodes with regenerating too old states.

### DIFF
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -441,7 +441,7 @@ def compute_previous_slot(slot: Slot) -> Slot:
 ```python
 def compute_previous_slots(slot: Slot, max_distance: int) -> [Slot]:
     from = slot - max_distance if slot > max_distance else 0
-    return range(from, slot)
+    return [Slot(s) for s in range(from, slot)]
 ```
 
 #### `compute_updated_sample_price`
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -46,6 +46,7 @@
   - [Misc](#misc-3)
     - [`next_power_of_two`](#next_power_of_two)
     - [`compute_previous_slot`](#compute_previous_slot)
+    - [`compute_previous_slots`](#compute_previous_slots)
     - [`compute_updated_sample_price`](#compute_updated_sample_price)
     - [`compute_committee_source_epoch`](#compute_committee_source_epoch)
     - [`batch_apply_participation_flag`](#batch_apply_participation_flag)
@@ -435,6 +436,14 @@ def compute_previous_slot(slot: Slot) -> Slot:
         return Slot(0)
 ```
 
+#### `compute_previous_slots`
+
+```python
+def compute_previous_slots(slot: Slot, max_distance: int) -> [Slot]:
+    from = slot - max_distance if slot > max_distance else 0
+    return range(from, slot)
+```
+
 #### `compute_updated_sample_price`
 
 ```python
@@ -698,9 +707,8 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
 
     # Verify that the block root matches,
     # to ensure the header will only be included in this specific Beacon Chain sub-tree.
-    from_slot = slot > MAX_BLOB_BLOCK_ROOT_DISTANCE ? slot - MAX_BLOB_BLOCK_ROOT_DISTANCE : 0 
     assert header.body_summary.beacon_block_root in 
-        [get_block_root_at_slot(state, s) for s in range(from_slot, slot)]
+        [get_block_root_at_slot(state, s) for s in compute_previous_slots(slot, MAX_BLOB_BLOCK_ROOT_DISTANCE)]
 
     # Check that this data is still pending
     committee_work = state.shard_buffer[slot % SHARD_STATE_MEMORY_SLOTS][shard]
--- a/specs/sharding/beacon-chain.md
+++ b/specs/sharding/beacon-chain.md
@@ -46,7 +46,6 @@
   - [Misc](#misc-3)
     - [`next_power_of_two`](#next_power_of_two)
     - [`compute_previous_slot`](#compute_previous_slot)
-    - [`compute_previous_slots`](#compute_previous_slots)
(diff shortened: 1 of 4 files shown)


specs/sharding/beacon-chain.md
```diff
@@ -697,7 +707,8 @@ def process_shard_header(state: BeaconState, signed_header: SignedShardBlobHeade
 
     # Verify that the block root matches,
     # to ensure the header will only be included in this specific Beacon Chain sub-tree.
-    assert header.body_summary.beacon_block_root == get_block_root_at_slot(state, slot - 1)
+    assert header.body_summary.beacon_block_root in
```
protolambda, line 710: Maybe instead we just specify the slot the root refers to with an additional field? Also note that for each gap slot, the root is the same

protolambda (commented):
Agreed on relaxing the condition, but maybe differently.

specs/sharding/beacon-chain.md
Nashatyrev: Is it about optimization? 
I could suggest `compute_previous_slots` to return slots in reverse order. So in vast majority of cases that would be an equivalent to the original check complexity

protolambda:
Implemented suggestion, as discussed in chat: we can add the slot (8 bytes) to avoid searching for the root.

djrtwo:
What is the validator actually supposed to insert in this reference? The current PR seems to allow for putting any beacon root in the past 32 slots ( and specifying which slot), but is there ever a case where the best strategy isn't to just put the oldest allowable reference?

If that is the case, we might considering do a fixed trailing slot distance (rather than a max trailing distance).
OR we just root on the checkpoint at the start this epoch (maybe not great for the 0th and 1st slot) **or** we can actually do the checkpoint at the start of the prior epoch (where current shuffling randomness was chosen).

**Primary question** -- is there any value in this being variable distance to be chosen by the proposer?

Side note: This needs to make it's way into the validator spec as well, right?

Nashatyrev:
> What is the validator actually supposed to insert in this reference? The current PR seems to allow for putting any beacon root in the past 32 slots ( and specifying which slot), but is there ever a case where the best strategy isn't to just put the oldest allowable reference?
>
> If that is the case, we might considering do a fixed trailing slot distance (rather than a max trailing distance).
> OR we just root on the checkpoint at the start this epoch (maybe not great for the 0th and 1st slot) **or** we can actually do the checkpoint at the start of the prior epoch (where current shuffling randomness was chosen).
> 
> **Primary question** -- is there any value in this being variable distance to be chosen by the proposer?

It's not a _validator/proposer_ but a _builder_ who chooses the anchor block. From my understanding there are distinct cases when a builder might want to choose earlier or later block: 
- E.g. a builder wants to build a blob for the slot `100`: `blob-100`, but the latest blob for his shard was published say at slot `97` (`blob-97`) and included to the `block-97`. He just wants to build a `blob-100` on top of the latest `blob-97`. For him it makes sense to 'anchor' his `blob-100` to `block-97` rather than to `block-99`: if there is a fork `block-99'` then the blob could be included in either fork and this doesn't make any difference for the builder
- E.g. a builder wants to build a blob for the slot `100`: `blob-100`, there is published `blob-98` included to `block-98`, there is published `blob-99`, but no `block-99` yet which may potentially include it (or not). The builder would like to build `blob-100` on top of `block-99` and can wait some extra time for it, but may choose falling back to 'fork' from `blob-98` (anchor block is `block-98`). In the latter case if the `block-99` is finally published late and accepted by network, the published `blob-100` (anchored to `block-98`) is still valid and can be included. 

In other words shard blobs form a L2 chain based on their anchor blocks and the beacon chain. And the builder should be free to choose which parent blob he wants to 'fork' from. I suppose it's very similar on how a validator (or pow miner) may choose the head to build a block (mine a block) on top of

@protolambda does that picture make any sense from L2 perspective?

djrtwo:
Right, I see this is for builders rather than proposers. Interesting

> For him it makes sense to 'anchor' his blob-100 to block-97 rather than to block-99

I'm not sure I 100% buy this usecase off-hand. In the event that block-97 is re-org'd and the prev-block of the rollup is orphaned, the L2 would still want to get *both* of the blocks on chain and submit two TXs into the execution-layer to point to and resolve the sequence of both. The order of data of an L2 getting into the shard-data layer doesn't matter. It is the sequence of TXs into the execution-layer that order and point to the data that do.

This also presumes that (1) shard data will only contain data for one rollup/application and thus a strict sequence is desired and (2) builders only get paid by associated applications if strict sequences are seen on chain. I think this overly complicates the market and mechanisms here

I think I misunderstood the original use of `beacon_block_root` in shard_blob_body when I originally reviewed. 

Application chaining aside. Can you explain more the value in "protect against blob replay to reorged beacon chain"?

djrtwo:
A followup question -- do blob-summaries expire? In an open market of builders, multiple builders might be trying to get the same data payload, `D`, in for a customer. Thus if builder-A gets payload `D` in at slot `N`, builder-B's summary for slot-N might be a repeat.

So I think I can see an argument for a pretty tight 1-slot blob_summary validity (especially for *future* slots) but I'm not sure I fully understand this retro-slot construction in this PR

protolambda:
@Nashatyrev same here, @djrtwo is right, we can be remove it. Thanks for making the new PR :+1:

PR #2587, opened 2021-09-08
terminal total difficulty override setting

lsankar4033:
This PR introduces a setting for merge clients to override terminal total difficulty (as discussed [here](https://github.com/ethereum/consensus-specs/issues/2547)). 

Follow-up work:
1. add a [EIP 3675 PoS event](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-3675.md#pos-events) for changes to `terminal_total_difficulty`
2. explore a setting (and relevant spec logic) for a blockhash override

Issue #2547, opened 2021-08-11
client setting for overriding `TOTAL_TRANSITION_DIFFICULTY`


lsankar4033:
The current mechanism for the [merge fork](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/merge/fork.md) involves setting an epoch (`MERGE_FORK_EPOCH`) from which nodes expect to wait 1 week before the terminal total difficulty is hit (via `TARGET_SECONDS_TO_MERGE`/`SECONDS_PER_ETH1_BLOCK`).

`SECONDS_PER_ETH1_BLOCK` is controllable by miners and the possibility exists that in a fit of griefing, they cartelize to slow down the subsequent merge process by a factor. While it's effectively impossible for the PoW chain to halt, such a slowdown could cause coordination headaches for a few days to weeks around merge time. Much of the ensuing uncertainty may also trickle down to entities that rely on sending Eth1x transactions for business/livelihood.

Not to mention that there may be some unknown unknowns about attacks/issues here.

I propose adding a `TOTAL_TRANSITION_DIFFICULTY_OVERRIDE` parameter that serves as an override of the merge point to ease coordination during this period. This parameter would be accessible via a beacon node client setting (`--total-transition-difficulty-override`) that's unused by default. Setting this would trigger an alternate code path during [TransitionStore initialization](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/merge/fork.md#initializing-transition-store) that just uses this value for the transition difficulty.

If we detect PoW block times slowing after setting `TARGET_SECONDS_TO_MERGE`, we have a lever we can use to change the point of transition. Furthermore, the mere existence of this option may be enough to dissuade miners from any griefing attack in the first place.

Curious of others' thoughts here.

(related: https://github.com/ethereum/EIPs/pull/3698)

djrtwo:
I agree that a manual override is valuable in event of unexpected events near the merge. Other than a slow down, you could also see a repeated 51% attack/reorg that never get us to terminal total difficultly (TTD).

Being able to set TTD manually and coordinate out of band seems very reasonable, especially for the mining slow-down case. I would also consider an override that just sets the terminal PoW blockhash which would be much safer than setting TTD in the event of an ongoing 51% attack. Setting terminal blockhash would likely result in a downtime but is a good absolute emergency backup.

Setting blockhash is probably a more complicated control path than just a TTD override so we should consider it before adding it here and a related event on execution-layer (eip) side.

[...]

PR #2599, opened 2021-09-15
Terminal blockhash override

lsankar4033:
Adds a terminal block hash override for the merge fork as originally discussed [here](https://github.com/ethereum/consensus-specs/issues/2547#issuecomment-897120430).

I'm creating this PR as a starting point for discussion while acknowledging that this might be somewhat controversial (more so than [the total difficulty override](https://github.com/ethereum/consensus-specs/pull/2587)) because coordinating on a specific blockhash might be perceived as more 'political'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ The merge is still actively in development. The exact specification has not been
   * [Merge fork](specs/merge/fork.md)
   * [Fork Choice changes](specs/merge/fork-choice.md)
   * [Validator additions](specs/merge/validator.md)
-  * [Client settings](specs/merge/client_settings.md)
+  * [Client settings](specs/merge/client-settings.md)
 
 ### Sharding
 
--- a/specs/merge/client-settings.md
+++ b/specs/merge/client-settings.md
@@ -1,21 +1,22 @@
-<!-- START doctoc generated TOC please keep comment here to allow auto update -->
-<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
-**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
-
-- [The Merge -- Client Settings](#the-merge----client-settings)
-    - [Override terminal total difficulty](#override-terminal-total-difficulty)
-
-<!-- END doctoc generated TOC please keep comment here to allow auto update -->
-
 # The Merge -- Client Settings
 
 **Notice**: This document is a work-in-progress for researchers and implementers.
 
 This document specifies configurable settings that clients must implement for the Merge.
 
+## Table of contents
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [The Merge -- Client Settings](#the-merge----client-settings)
+    - [Override terminal total difficulty](#override-terminal-total-difficulty)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
 ### Override terminal total difficulty
 
-To coordinate manual overrides to [`terminal_total_difficulty`](fork-choice.md#transitionstore), clients
+To coordinate manual overrides to [`terminal_total_difficulty`](./fork-choice.md#transitionstore), clients
 must provide `--terminal-total-difficulty-override` as a configurable setting.
 
 If `TransitionStore` has already [been initialized](./fork.md#initializing-transition-store), this alters the previously initialized value of
@@ -23,4 +24,3 @@ If `TransitionStore` has already [been initialized](./fork.md#initializing-trans
 `terminal_total_difficulty`.
 
 Except under exceptional scenarios, this setting is expected to not be used, and `terminal_total_difficulty` will operate with [default functionality](./fork.md#initializing-transition-store). Sufficient warning to the user about this exceptional configurable setting should be provided.
-[here](fork.md#initializing-transition-store).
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -45,7 +45,7 @@ The merge is still actively in development. The exact specification has not been
   * [Merge fork](specs/merge/fork.md)
   * [Fork Choice changes](specs/merge/fork-choice.md)
   * [Validator additions](specs/merge/validator.md)
-  * [Client settings](specs/merge/client_settings.md)
+  * [Client settings](specs/merge/client-settings.md)
```
hwwhww, line 48: nice catch

specs/merge/fork.md
```diff
@@ -120,7 +120,7 @@ def compute_terminal_total_difficulty(anchor_pow_block: PowBlock) -> uint256:
 
 def get_transition_store(anchor_pow_block: PowBlock) -> TransitionStore:
     terminal_total_difficulty = compute_terminal_total_difficulty(anchor_pow_block)
-    return TransitionStore(terminal_total_difficulty=terminal_total_difficulty)
+    return TransitionStore(terminal_total_difficulty=terminal_total_difficulty, terminal_block_hash=None)
```
hwwhww, line 123: lint error:
```
eth2spec/merge/mainnet.py:2762: error: Argument "terminal_block_hash" to "TransitionStore" has incompatible type "None"; expected "Hash32"
```

You have to change the `terminal_block_hash` type to:
```python
@dataclass
class TransitionStore(object):
    terminal_total_difficulty: uint256
    terminal_block_hash: Optional[Hash32]
```

specs/merge/client-settings.md
```diff
@@ -1,26 +1,34 @@
...
 
 Except under exceptional scenarios, this setting is expected to not be used, and `terminal_total_difficulty` will operate with [default functionality](./fork.md#initializing-transition-store). Sufficient warning to the user about this exceptional configurable setting should be provided.
-[here](fork.md#initializing-transition-store).
+
+### Override terminal block hash
+
+In case fork coordination around a specific PoW block hash is necessary, clients must also provide `--terminal-block-hash-override` as a configurable setting.
+
+If `TransitionStore` has already [been initialized](./fork.md#initializing-transition-store), this alters the previously initialized value of
+`TransitionStore.terminal_block_hash`, otherwise this setting initializes `TransitionStore` with the specified `terminal_block_hash`.
```
mkalinin, line 32: It must also initialize the `terminal_total_difficulty` field of the store and set it to the very high value, probably `2**256 - 1`

specs/merge/client-settings.md
```diff
@@ -1,26 +1,34 @@
...
 ### Override terminal total difficulty
 
-To coordinate manual overrides to [`terminal_total_difficulty`](fork-choice.md#transitionstore), clients
+To coordinate manual overrides to [`terminal_total_difficulty`](./fork-choice.md#transitionstore), clients
 must provide `--terminal-total-difficulty-override` as a configurable setting.
 
 If `TransitionStore` has already [been initialized](./fork.md#initializing-transition-store), this alters the previously initialized value of
-`TransitionStore.terminal_total_difficulty`, otherwise this setting initializes `TransitionStore` with the specified, bypassing `compute_terminal_total_difficulty` and the use of an `anchor_pow_block`.
-`terminal_total_difficulty`.
+`TransitionStore.terminal_total_difficulty`, otherwise this setting initializes `TransitionStore` with the specified `terminal_total_difficulty`, bypassing `compute_terminal_total_difficulty` and the use of an `anchor_pow_block`.
```
mkalinin, line 23: We should also set `terminal_block_hash` field to the default `Hash32()` value in this case.

specs/merge/fork.md
mkalinin: As mentioned above:
```suggestion
    return TransitionStore(terminal_total_difficulty=terminal_total_difficulty, terminal_block_hash=Hash32())
```

specs/merge/fork-choice.md
```diff
@@ -101,9 +102,12 @@ Used by fork-choice handler, `on_block`.
 
 ```python
 def is_valid_terminal_pow_block(transition_store: TransitionStore, block: PowBlock, parent: PowBlock) -> bool:
-    is_total_difficulty_reached = block.total_difficulty >= transition_store.terminal_total_difficulty
-    is_parent_total_difficulty_valid = parent.total_difficulty < transition_store.terminal_total_difficulty
-    return block.is_valid and is_total_difficulty_reached and is_parent_total_difficulty_valid
+    if transition_store.terminal_block_hash is not None:
```
mkalinin, line 105: Once we have `Hash32()` as the default value for the `terminal_block_hash` we will be able to do this check in the following way:

```python
if block.block_hash == transition_store.terminal_block_hash:
    return True

is_total_difficulty_reached = block.total_difficulty >= transition_store.terminal_total_difficulty
is_parent_total_difficulty_valid = parent.total_difficulty < transition_store.terminal_total_difficulty
return block.is_valid and is_total_difficulty_reached and is_parent_total_difficulty_valid
```

mkalinin (commented):
Added a few suggestions and comments

specs/merge/fork.md
djrtwo: Prefer the default to `Hash32`. Makes the spec read cleaner 👍

djrtwo (commented):
Looks good 👍 !

Additionally we need to ensure this logic makes it into `merge/validator.md`

I'll do a final pass when that is added and you get the `Hash32()` change in there.

specs/merge/client-settings.md
lsankar4033: ya I like this

lsankar4033:
closing in favor of adding a constant as in https://github.com/ethereum/consensus-specs/pull/2605#issuecomment-924848499

PR #2601, opened 2021-09-15
pass sync committee sig consistently in tests

etan-status:
There are three defined unit tests for the light client sync protocol.
They all follow a similar structure. However, there is an inconcistency
how they pass the slot to compute_aggregate_sync_committee_signature.
In one instance it is passed as `block.slot`. In the other two cases
it is passed as `block_header.slot`. As the `block_header` is created
from the `block`, they share the same value. This patch makes the way
how the slot is passed consistent across all of the test cases.


djrtwo (approved):
works for me!

digging through some tests?

PR #2604, opened 2021-09-16
Add ignore condition for `unseen shard_blob_root `

terencechain:
Adding a condition where attestation's `shard_blob_root` must been seen or else ignore / may queue. This seems to align better with state transition's `process_attested_shard_work` via the unknown header conditions


aashumallik (approved):
LGTM

PR #2606, opened 2021-09-17
Add extra_data field to the execution payload

mkalinin:
### What's done?

Adds `extra_data` field to the execution payload.

### Rationale

See the rationale section of the corresponding change to the EIP-3675 https://github.com/ethereum/EIPs/pull/3982


djrtwo (approved):
looks good

PR #2607, opened 2021-09-17
Dankrad's merge review

dankrad:


specs/merge/beacon-chain.md
```diff
@@ -59,7 +59,7 @@ This patch adds transaction execution to the beacon chain as part of the Merge f
 
 | Name | Value |
 | - | - |
-| `MAX_BYTES_PER_OPAQUE_TRANSACTION` | `uint64(2**20)` (= 1,048,576) |
+| `MAX_BYTES_PER_OPAQUE_TRANSACTION` | `uint64(2**24)` (= 1,048,576) |
```
dankrad, line 62: With 16 bytes per gas and 30M limit we can already have almost 2MB transactions now. Why limit this?

specs/merge/fork-choice.md
```diff
@@ -71,16 +59,14 @@ def finalize_block(self: ExecutionEngine, block_hash: Hash32) -> bool:
 ### `TransitionStore`
 
 ```python
-@dataclass
```
dankrad, line 74: Don't understand why these are not simply SSZ containers?

mkalinin: It's not supposed to be an SSZ container. Btw, it's removed in https://github.com/ethereum/consensus-specs/pull/2605. We should probably wait for the resolution on this PR

djrtwo: For things we don't serialize and instead are just used to represent requisite data stored in the consensus client to peform the operations in some way, we tend to use `@dataclass`. When we instead use an SSZ container, it signals that this data will need to be serialized and/or hashed in specific ways to conform to spec.

See the fork choice spec for the `Store` as an example

But, it is being removed in #2605 so resolving this

specs/merge/fork-choice.md
```diff
@@ -43,25 +43,13 @@ The body of this function is implementation dependent.
...
-def set_head(self: ExecutionEngine, block_hash: Hash32) -> bool:
-    """
-    Returns True if the ``block_hash`` was successfully set as head of the execution payload chain.
-    """
-    ...
-```
-
-#### `finalize_block`
+HEAD_TYPE_FINALIZED_HEAD = uint64(0)
+HEAD_TYPE_SAFE_HEAD = uint64(1)
```
djrtwo, line 47: From a consensus standpoint there is just one head being set. I would instead reserve the "safe head" strictly for the API and concerns around that

specs/merge/fork-choice.md
```diff
@@ -71,16 +59,14 @@ def finalize_block(self: ExecutionEngine, block_hash: Hash32) -> bool:
...
-@dataclass
-class TransitionStore(object):
+class TransitionStore(Container):
     terminal_total_difficulty: uint256
 ```
 
 ### `PowBlock`
 
 ```python
-@dataclass
```
djrtwo, line 82: Suggest keeping `@dataclass` because this data is not serialized nor hashed. This conforms to our prior usage of `@dataclass` with the fork choice `Store`

specs/merge/p2p-interface.md
```diff
@@ -78,8 +78,6 @@ Alias `block = signed_beacon_block.message`, `execution_payload = block.body.exe
     -- i.e. `execution_payload.timestamp == compute_time_at_slot(state, block.slot)`.
   - _[REJECT]_ Gas used is less than the gas limit --
     i.e. `execution_payload.gas_used <= execution_payload.gas_limit`.
-  - _[REJECT]_ The execution payload block hash is not equal to the parent hash --
```
djrtwo, line 81: +1 We added this a while back on request but it buys us pretty much nothing

specs/merge/beacon-chain.md
djrtwo: I think because the consensus cannot validate these TXs nor the gas limit and this puts a safe(r) limit on sanity check sizes of blocks coming off the wire.

cc @protolambda who I think will have the deeper reasoning available

specs/merge/fork-choice.md
```diff
@@ -43,25 +43,13 @@ The body of this function is implementation dependent.
...
 
-Applies finality to the execution state: it irreversibly persists the chain of all execution payloads
-and corresponding state, up to and including `block_hash`.
-
-The body of this function is implementation dependent.
-The Consensus API may be used to implement this with an external execution engine.
-
-```python
-def finalize_block(self: ExecutionEngine, block_hash: Hash32) -> bool:
+def set_head(self: ExecutionEngine, block_hash: Hash32, head_type: unit64) -> bool:
```
djrtwo, line 50: If we do want to signal head, safe_head, and finalized discretely via this interface, I suggest instead we require all pieces of data passed into this method simultaneously as to signal an *atomic* update of each. 

This is how the both API and EIP-3675 POS_EVENTS are being spec'd in practice to avoid race conditions.

I'm still mixed about the consensus specs defining "safe" here as from the consensus (and the execution EIP) perspective. To me, this seems much more of something that should just live in the API because it has no bearing on the consensus specs.

I *at least* think that this function should have a 1:1 mapping with the fields expected in `POS_FORKCHOICE_UPDATED` which right now is only head and finalized because from the perspective of the execution-layer core specs that is all that is needed

specs/merge/beacon-chain.md
djrtwo: Because this number multiplies with `MAX_TRANSACTIONS_PER_PAYLOAD` to be an insanely large block. The consensus-layer does not have sane-limit validations on the size of the block as it comes off the wire so a limit is introduced here in conjunction with the `MAX_TRANSACTIONS_PER_PAYLOAD`.

At the current values, this can still be a 16GB block. This is why the `MAY` is available in the p2p validation in your other comment. A 16GB block is a dos block... but so is a 1GB block.... is a 50MB block? Not sure but sane limits can be encoded into the p2p validations (static or dynamic depending on the client's choice of anti-dos measures).

Also that comment is `MAY` in relation to stricter size limits. Maybe we should be a bit clearer to not imply that this means random validations can be applied

specs/merge/fork-choice.md
mkalinin: I have started a work on updating `ExecutionEngine` interface in the spec, the goal is to have a 1:1 matching on the events and Engine API. It should be done soon.

specs/merge/beacon-chain.md
protolambda: This also closely relates to EIP-3860, as discussed on ACD (cc @holiman): https://eips.ethereum.org/EIPS/eip-3860

Generally we should put a validation condition in place that checks if the sum of the block is not causing a DoS just by its size. Req-Resp and Gossipsub payloads in eth2 are all currently limited to 1 MB, we should increase that if valid block instances can actually be larger, to follow consensus. And then within the EVM execution the same limits apply, where EIP-3860 is also concerned with.

That said, if we handle DoS properly then I'm not opposed to set the capacity to `2**24` just for future compatibility, if the block size needs to grow without changing the SSZ type.

dankrad: I don't buy the argument that this prevents DOS blocks, as this would, as @djrtwo noticed, still allow 16 GB blocks. The right way to do prevent the DOS attack is to limit the execution payload size. For example, any execution payload over `GAS_LIMIT // 16` in size is obviously invalid and can be rejected.

I don't see an argument why there needs to be any further restriction for transaction. If there is a transaction that needs 1.8 MB of data and pays for the gas I don't see a reason to exclude it, since we are accepting 2 transactions of 0.9 MB each. I can't see any DOS vector this prevents. And you need to be able to process the full block size either way.

dapplion: Would the spec settle on a specific value to limit the payload size for p2p validation before the code freeze?

jmozah: p2p already restricts its load to 1Mb.  Nothing greater than 1Mb will reach here.

djrtwo: p2p spec is much simpler to change/upgrade over time. Thus @dankrad's argument is to put the dos protections there and leave this more open to theoretical maximums

specs/merge/fork-choice.md
mkalinin: The plan is to add `confirmed_block_hash` to the parameter set of the `engine_forkchoiceUpdated` method to pass this information to the execution engine. Respective change to the [`notify_forkchoice_updated`](https://github.com/ethereum/consensus-specs/blob/dev/specs/merge/fork-choice.md#notify_forkchoice_updated) function should be made by that time

specs/merge/beacon-chain.md
mkalinin: It also worth considering that the size of ETH protocol messages is limited to 10MB in geth https://github.com/ethereum/go-ethereum/blob/master/eth/protocols/eth/protocol.go#L49

specs/merge/fork-choice.md
hwwhww: FYI it has changed to SSZ class so that we can provide serialized test vectors.

specs/merge/beacon-chain.md
dankrad: With rollups this is very relevant in practice, as this will be the maximum fraud proof size. So it's a pretty important quantity.

djrtwo: This is being addressed here -- https://github.com/ethereum/consensus-specs/pull/2686

PR #2609, opened 2021-09-17
Fix random validation

djrtwo:
`execution_payload.random` was incorrectly only being evaluated once the transition had happened and there was a non-empty `latest_execution_payload` in the state.

This field can and should be evaluated even at the transition block because it is not a function of the `latest_execution_payload` but instead is a function of just the beacon state.  Without this fix, arbitrary data could be inserted for `random` at the transition block.

Ensured that `test_bad_random_first_payload` before change to beacon transition function. And passes with the change


mkalinin (approved):
Very well spotted! The fix looks good to me 👍

PR #2610, opened 2021-09-18
Update README.md

peteyburgers:
Removed extraneous period


protolambda:
This is more like a short sentence than a document title, the period is not a concern. Thanks for the suggestion though.

PR #2611, opened 2021-09-18
added NFT to design goals

jeyakatsa:


dapplion (changes requested):
While very relevant in today's app layer this design goal in not specific enough such that's 'executable'. It would be more appropriate for user-facing materials such as the ethereum-org site.

protolambda:
NFTs are fun, but just one example of the many applications that scaling can improve. And we do not favor any application, doing so would limit innovation.

jeyakatsa:
Understood.

PR #2613, opened 2021-09-20
Update execution engine calls

mkalinin:
### What's done?
Updates `ExecutionEngine` methods and aligns them with the most recent version of the Engine API. Execution engine concept in the spec is pretty useful for testing purposes and also for implementers as it gives an idea of where these methods are supposed to be called.

Though, not sure how far should we go in aligning methods of `ExecutionEngine` with the Engine API, we better find a balance between specifying the calls and coupling the spec with the API.

This PR is based on https://github.com/ethereum/consensus-specs/pull/2605

### DIFF
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -31,7 +31,8 @@
     - [`compute_timestamp_at_slot`](#compute_timestamp_at_slot)
 - [Beacon chain state transition function](#beacon-chain-state-transition-function)
   - [Execution engine](#execution-engine)
-    - [`on_payload`](#on_payload)
+    - [`execute_payload`](#execute_payload)
+    - [`consensus_validated`](#consensus_validated)
   - [Block processing](#block-processing)
   - [Execution payload processing](#execution-payload-processing)
     - [`is_valid_gas_limit`](#is_valid_gas_limit)
@@ -237,19 +238,31 @@ def compute_timestamp_at_slot(state: BeaconState, slot: Slot) -> uint64:
 The implementation-dependent `ExecutionEngine` protocol encapsulates the execution sub-system logic via:
 
 * a state object `self.execution_state` of type `ExecutionState`
-* a state transition function `self.on_payload` which mutates `self.execution_state`
+* a state transition function `self.execute_payload` which applies changes to the `self.execution_state`
+* a method `self.consensus_validated` which notifies that the block holding the execution payload
+is valid with respect to the consensus rule set
 
-#### `on_payload`
+#### `execute_payload`
 
 ```python
-def on_payload(self: ExecutionEngine, execution_payload: ExecutionPayload) -> bool:
+def execute_payload(self: ExecutionEngine, execution_payload: ExecutionPayload) -> bool:
     """
     Returns ``True`` iff ``execution_payload`` is valid with respect to ``self.execution_state``.
     """
     ...
 ```
 
-The above function is accessed through the `EXECUTION_ENGINE` module which instantiates the `ExecutionEngine` protocol.
+#### `consensus_validated`
+
+```python
+def consensus_validated(self: ExecutionEngine, execution_payload: ExecutionPayload) -> None:
+    ...
+```
+
+The above functions are accessed through the `EXECUTION_ENGINE` module which instantiates the `ExecutionEngine` protocol.
+
+The body of each of these functions is implementation dependent.
+The Engine API may be used to implement them with an external execution engine.
 
 ### Block processing
 
@@ -264,6 +277,9 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
     process_eth1_data(state, block.body)
     process_operations(state, block.body)
     process_sync_aggregate(state, block.body.sync_aggregate)
+    if is_execution_enabled(state, block.body):
+        # Notify the block is valid with respect to the consensus state transition function
+        EXECUTION_ENGINE.consensus_validated(block.body.execution_payload)  # [New in Merge]
 ```
 
 ### Execution payload processing
@@ -304,7 +320,7 @@ def process_execution_payload(state: BeaconState, payload: ExecutionPayload, exe
     # Verify timestamp
     assert payload.timestamp == compute_timestamp_at_slot(state, state.slot)
     # Verify the execution payload is valid
-    assert execution_engine.on_payload(payload)
+    assert execution_engine.execute_payload(payload)
     # Cache execution payload header
     state.latest_execution_payload_header = ExecutionPayloadHeader(
         parent_hash=payload.parent_hash,
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -32,7 +32,7 @@
 - [Beacon chain state transition function](#beacon-chain-state-transition-function)
   - [Execution engine](#execution-engine)
     - [`execute_payload`](#execute_payload)
-    - [`consensus_validated`](#consensus_validated)
+    - [`notify_consensus_validated`](#notify_consensus_validated)
   - [Block processing](#block-processing)
   - [Execution payload processing](#execution-payload-processing)
     - [`is_valid_gas_limit`](#is_valid_gas_limit)
@@ -242,10 +242,10 @@ The implementation-dependent `ExecutionEngine` protocol encapsulates the executi
 
 * a state object `self.execution_state` of type `ExecutionState`
 * a state transition function `self.execute_payload` which applies changes to the `self.execution_state`
-* a function `self.consensus_validated` which signals that the beacon block containing the execution payload
+* a function `self.notify_consensus_validated` which signals that the beacon block containing the execution payload
 is valid with respect to the consensus rule set
 
-*Note*: `execute_payload` and `consensus_validated` are functions accessed through the `EXECUTION_ENGINE` module which instantiates the `ExecutionEngine` protocol.
+*Note*: `execute_payload` and `notify_consensus_validated` are functions accessed through the `EXECUTION_ENGINE` module which instantiates the `ExecutionEngine` protocol.
 
 The body of each of these functions is implementation dependent.
 The Engine API may be used to implement them with an external execution engine.
@@ -260,14 +260,14 @@ def execute_payload(self: ExecutionEngine, execution_payload: ExecutionPayload) 
     ...
 ```
 
-#### `consensus_validated`
+#### `notify_consensus_validated`
 
 ```python
-def consensus_validated(self: ExecutionEngine, block_hash: Hash32, valid: bool) -> None:
+def notify_consensus_validated(self: ExecutionEngine, block_hash: Hash32, valid: bool) -> None:
     ...
 ```
 
-The inputs to this function depend on the result of the state transition. A call to `consensus_validated` must be made after the [`state_transition`](../phase0/beacon-chain.md#beacon-chain-state-transition-function) function finishes. The value of the `valid` parameter must be set as follows:
+The inputs to this function depend on the result of the state transition. A call to `notify_consensus_validated` must be made after the [`state_transition`](../phase0/beacon-chain.md#beacon-chain-state-transition-function) function finishes. The value of the `valid` parameter must be set as follows:
 
 * `True` if `state_transition` function call succeeds
 * `False` if `state_transition` function call fails
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -273,6 +273,8 @@ The inputs to this function depend on the result of the state transition. A call
 * `True` if `state_transition` function call succeeds
 * `False` if `state_transition` function call fails
 
+*Note*: The call of the `notify_consensus_validated` function with `valid = True` maps on the `POS_CONSENSUS_VALIDATED` event defined in the [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675#definitions).
+
 ### Block processing
 
 *Note*: The call to the `process_execution_payload` must happen before the call to the `process_randao` as the former depends on the `randao_mix` computed with the reveal of the previous block.
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -242,8 +242,13 @@ The implementation-dependent `ExecutionEngine` protocol encapsulates the executi
 
 * a state object `self.execution_state` of type `ExecutionState`
 * a state transition function `self.execute_payload` which applies changes to the `self.execution_state`
-* a method `self.consensus_validated` which signals that the beacon block containing the execution payload
+* a function `self.consensus_validated` which signals that the beacon block containing the execution payload
 is valid with respect to the consensus rule set
+
+*Note*: `execute_payload` and `consensus_validated` are functions accessed through the `EXECUTION_ENGINE` module which instantiates the `ExecutionEngine` protocol.
+
+The body of each of these functions is implementation dependent.
+The Engine API may be used to implement them with an external execution engine.
 
 #### `execute_payload`
 
@@ -261,11 +266,6 @@ def execute_payload(self: ExecutionEngine, execution_payload: ExecutionPayload) 
 def consensus_validated(self: ExecutionEngine, execution_payload: ExecutionPayload) -> None:
     ...
 ```
-
-The above functions are accessed through the `EXECUTION_ENGINE` module which instantiates the `ExecutionEngine` protocol.
-
-The body of each of these functions is implementation dependent.
-The Engine API may be used to implement them with an external execution engine.
 
 ### Block processing
 
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -239,7 +239,7 @@ The implementation-dependent `ExecutionEngine` protocol encapsulates the executi
 
 * a state object `self.execution_state` of type `ExecutionState`
 * a state transition function `self.execute_payload` which applies changes to the `self.execution_state`
-* a method `self.consensus_validated` which notifies that the block holding the execution payload
+* a method `self.consensus_validated` which signals that the beacon block containing the execution payload
 is valid with respect to the consensus rule set
 
 #### `execute_payload`
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -54,6 +54,7 @@ This patch adds transaction execution to the beacon chain as part of the Merge f
 | - | - | - |
 | `OpaqueTransaction` | `ByteList[MAX_BYTES_PER_OPAQUE_TRANSACTION]` | a [typed transaction envelope](https://eips.ethereum.org/EIPS/eip-2718#opaque-byte-array-rather-than-an-rlp-array) structured as `TransactionType \|\| TransactionPayload` |
 | `Transaction` | `Union[OpaqueTransaction]` | a transaction |
+| `ExecutionAddress` | `Bytes20` | Address of account on the execution layer |
 
 ## Constants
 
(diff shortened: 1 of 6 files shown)


specs/merge/beacon-chain.md
```diff
@@ -257,6 +277,9 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
     process_eth1_data(state, block.body)
     process_operations(state, block.body)
     process_sync_aggregate(state, block.body.sync_aggregate)
+    if is_execution_enabled(state, block.body):
```
djrtwo, line 280: This is not technically the appropriate place to run `consensus_validated` because `block.state_root` is not validated yet. Instead, it should be after `if validate_result` in `state_transition()`.

You might debate whether it should be embedded in `if validate_result` code block or outside. `validate_result == False` only when a proposer is building a beacon block (and thus when computing the state root to embed rather validating some untrusted input).

To that end, we should discuss what the execution engine does with an execution_payload that it produced for a local proposer via `engine_getPayload`. Must this be inserted manually via `engine_executePayload` and then we run `engine_consensusValidated`? Or instead does the execution engine keep the work done in `engine_getPayload` then either assuming consensus validated or awaiting consensus validated?

Those answers imply how to actually spec this here, in the validator guide, and in the engine API.

My gut is that the execution engine upon a call `engine_getPayload` operates in a similar way wrt to the block as though `engine_executePayload` has been called, but not yet `engine_consensusValidated`. And that only once the proposer has fully computed and signed/broadcast the beacon block, does the beacon node call `engine_consensusValidated`. So I'd suggest the following changes:

* Move `EXECUTION_ENGINE.consensus_validated(block.body.execution_payload)` in the `state_transition()` inside the `if validate_result:` codeblock
* Add in the validator guide that after the validator has successfully created/signed the block, that they call `EXECUTION_ENGINE.consensus_validated` -- note that I think if `validate_result == False` that it is too early to call it within `state_transition()` because the final signature has not been made on the block thus there is still room for the validator to fail/abort the process.
* Add a note in the engine API on `engine_getPayload` that this should leave the node the same state as though `engine_executePayload` was called -- transition made but awaiting final confirmation from `engine_consensusValidated

specs/merge/beacon-chain.md
```diff
@@ -230,19 +238,31 @@ def compute_timestamp_at_slot(state: BeaconState, slot: Slot) -> uint64:
...
     Returns ``True`` iff ``execution_payload`` is valid with respect to ``self.execution_state``.
     """
     ...
 ```
 
-The above function is accessed through the `EXECUTION_ENGINE` module which instantiates the `ExecutionEngine` protocol.
+#### `consensus_validated`
+
+```python
+def consensus_validated(self: ExecutionEngine, execution_payload: ExecutionPayload) -> None:
```
djrtwo, line 258: We assume `forkchoice_updated` has the eth1 hashes of the execution_payload which maps well to the API definitions, but on `consensus_validated` we diverge from the API and instead pass in the full `execution_payload`.

We could allow more conformance by utilizing the `execution_payload.block_hash` for `consensus_validated` call instead of the full payload. Do you see value in going one way or the other?

specs/merge/beacon-chain.md
```diff
@@ -230,19 +238,31 @@ def compute_timestamp_at_slot(state: BeaconState, slot: Slot) -> uint64:
 The implementation-dependent `ExecutionEngine` protocol encapsulates the execution sub-system logic via:
 
 * a state object `self.execution_state` of type `ExecutionState`
-* a state transition function `self.on_payload` which mutates `self.execution_state`
+* a state transition function `self.execute_payload` which applies changes to the `self.execution_state`
+* a method `self.consensus_validated` which notifies that the block holding the execution payload
```
djrtwo, line 242:
```suggestion
* a method `self.consensus_validated` which signals that the beacon block containing the execution payload
```

specs/merge/beacon-chain.md
```diff
@@ -230,19 +238,31 @@ def compute_timestamp_at_slot(state: BeaconState, slot: Slot) -> uint64:
...
 
-The above function is accessed through the `EXECUTION_ENGINE` module which instantiates the `ExecutionEngine` protocol.
+#### `consensus_validated`
+
+```python
+def consensus_validated(self: ExecutionEngine, execution_payload: ExecutionPayload) -> None:
+    ...
+```
+
+The above functions are accessed through the `EXECUTION_ENGINE` module which instantiates the `ExecutionEngine` protocol.
```
djrtwo, line 262: Suggest moving these notes to the `### Execution Engine` subsection prior to these function specific subsections.  (that is, right after the bullet points)

```
*Note*: `execute_payload` and `consensus_validated` are functions...

The body...
The Engine API...
```

specs/merge/fork-choice.md
```diff
@@ -35,47 +33,22 @@ This is the modification of the fork choice according to the executable beacon c
...
 The following methods are added to the `ExecutionEngine` protocol for use in the fork choice:
 
-#### `set_head`
+#### `forkchoice_updated`
 
-Re-organizes the execution payload chain and corresponding state to make `block_hash` the head.
-
-The body of this function is implementation dependent.
-The Consensus API may be used to implement this with an external execution engine.
+* Re-organizes the execution payload chain and corresponding state to make `head_block_hash` the head.
```
djrtwo, line 38:
```suggestion
This method performs two actions *atomically*:
* Re-organizes the execution payload chain and corresponding state to make `head_block_hash` the head.
```

specs/merge/fork-choice.md
```diff
@@ -35,47 +33,22 @@ This is the modification of the fork choice according to the executable beacon c
...
+def forkchoice_updated(self: ExecutionEngine, head_block_hash: Hash32, finalized_block_hash: Hash32) -> None:
     ...
 ```
 
-#### `finalize_block`
-
-Applies finality to the execution state: it irreversibly persists the chain of all execution payloads
-and corresponding state, up to and including `block_hash`.
-
 The body of this function is implementation dependent.
```
djrtwo, line 47: Suggest moving this up a `*Note*` in the `ExecutionEngine` section right above

specs/merge/validator.md
```diff
@@ -39,19 +40,39 @@ Please see related Beacon Chain doc before continuing and use them as a referenc
 
 The following methods are added to the `ExecutionEngine` protocol for use as a validator:
 
-#### `assemble_block`
+#### `prepare_payload`
 
-Produces a new instance of an execution payload, with the specified `timestamp`,
-on top of the execution payload chain tip identified by `block_hash`.
+Given the set of execution payload attributes initiates a process of building an execution payload
```
djrtwo, line 45:
```suggestion
Given the set of execution payload attributes, `prepare_payload` initiates a process of building an execution payload
```

specs/merge/validator.md
```diff
@@ -39,19 +40,39 @@ Please see related Beacon Chain doc before continuing and use them as a referenc
...
+                    fee_recipient: Bytes20) -> uint64:
+    """
+    Returns ``payload_id`` that is used to obtain the execution payload in a subsequent ``get_payload`` call.
+    """
+    ...
+```
+
+#### `get_payload`
+
+Given the `payload_id` returns the most recent version of the execution payload that
```
djrtwo, line 62:
```suggestion
Given the `payload_id`, `get_payload` returns the most recent version of the execution payload that
```

specs/merge/validator.md
```diff
@@ -39,19 +40,39 @@ Please see related Beacon Chain doc before continuing and use them as a referenc
...
 ```python
-def assemble_block(self: ExecutionEngine, block_hash: Hash32, timestamp: uint64, random: Bytes32) -> ExecutionPayload:
+def get_payload(self: ExecutionEngine, payload_id: uint64) -> ExecutionPayload:
+    """
+    Returns ``execution_payload`` object.
+    """
     ...
 ```
 
+The body of each of these functions is implementation dependent.
```
djrtwo, line 73: Suggest moving to the parent header section above

specs/merge/validator.md
```diff
@@ -39,19 +40,39 @@ Please see related Beacon Chain doc before continuing and use them as a referenc
...
+on top of the execution chain tip identified by `parent_hash`.
 
-The body of this function is implementation dependent.
-The Consensus API may be used to implement this with an external execution engine.
+```python
+def prepare_payload(self: ExecutionEngine,
+                    parent_hash: Hash32,
+                    timestamp: uint64,
+                    random: Bytes32,
+                    fee_recipient: Bytes20) -> uint64:
```
djrtwo, line 53: Consider defining a type `PayloadId`

djrtwo (commented):
Still reviewing the validator guide but wanted to get review to you before I go to sleep. Will finish the rest soon!

specs/merge/validator.md
```diff
@@ -60,9 +81,14 @@ All validator responsibilities remain unchanged other than those noted below. Na
 
 #### Constructing the `BeaconBlockBody`
 
-##### Execution Payload
+##### ExecutionPayload
 
-* Set `block.body.execution_payload = get_execution_payload(state, transition_store, execution_engine, pow_chain)` where:
+To obtain an execution payload a proposer of a block must take the following actions:
```
djrtwo, line 86:
```suggestion
To obtain an execution payload, the proposer of a block must take the following actions:
```

specs/merge/validator.md
```diff
@@ -75,28 +101,37 @@ def get_pow_block_at_total_difficulty(total_difficulty: uint256, pow_chain: Sequ
...
-    return execution_engine.assemble_block(parent_hash, timestamp, randao_mix)
-
-
-def get_execution_payload(state: BeaconState,
-                          transition_store: TransitionStore,
-                          execution_engine: ExecutionEngine,
-                          pow_chain: Sequence[PowBlock]) -> ExecutionPayload:
+def prepare_execution_payload(state: BeaconState,
+                              pow_chain: Sequence[PowBlock],
+                              fee_recipient: Bytes20,
```
djrtwo, line 106: Consider defining a type for `ExecutionAddress`

specs/merge/beacon-chain.md
mkalinin: > This is not technically the appropriate place to run `consensus_validated` because `block.state_root` is not validated yet. Instead, it should be after `if validate_result` in `state_transition()`.

This is so true!

> Add a note in the engine API on `engine_getPayload` that this should leave the node the same state as though `engine_executePayload` was called -- transition made but awaiting final confirmation from `engine_consensusValidated`

I would not mix proposal and processing flows as this would add one more state to the block proposal state machine, i.e. `prepare + get + validated` instead of `prepare + get`. The spec should also be cleaner if we define them separate.

I am not sure if every consensus client does the block proposal as follows but Teku does and it looks like an optimal solution:
1. VC requests unsigned block from BN
2. VC signs it and submits it to the BN
3. BN gossips signed block
4. BN imports signed block as usual

This approach makes it sure that block gets gossiped ASAP and fully verified before attesting to this block happens. If we add execution client to this scheme it will start looking as follows:
1. VC requests unsigned block from BN
    * BN calls `getPayload` and returns a block with the payload
2. VC signs it and submits it to the BN
3. BN gossips signed block
4. BN imports signed block as usual
    * BN calls `executePayload`
    * BN calls `consensusValidated`

In this case execution client will do double work on executing the payload. I don't think it matter much as long as block gossip is not affected by the execution. Execution client may look for an optimisation in a way they store the block hash mapped on the state when `getPayload` is served expecting that this payload will be sent a few moments later with `executePayload` and avoid executing the same payload twice.

specs/merge/beacon-chain.md
mkalinin: Yeah, I agree to make this change for conformance, and I would also add `valid` flag to this function (see the comment above)

specs/merge/beacon-chain.md
djrtwo: That looks good to me.

I don't want extra work to be done, but you're right that an optimistic cache of the result of prepare/get can be used to mitigate that without having to rely upon alterations of the API. Additionally, you're right that we should mirror the normal BN block import path as much as possible

specs/merge/beacon-chain.md
djrtwo: looks good

specs/merge/beacon-chain.md
```diff
@@ -240,19 +241,31 @@ def compute_timestamp_at_slot(state: BeaconState, slot: Slot) -> uint64:
 The implementation-dependent `ExecutionEngine` protocol encapsulates the execution sub-system logic via:
 
 * a state object `self.execution_state` of type `ExecutionState`
-* a state transition function `self.on_payload` which mutates `self.execution_state`
+* a state transition function `self.execute_payload` which applies changes to the `self.execution_state`
+* a function `self.consensus_validated` which signals that the beacon block containing the execution payload
```
hwwhww, line 245: Nitpicking on the function name conventions. Active voice v.s. passive voice.

I saw there is `engine_consensusValidated` in Engine APIs, but this naming pattern seems new to consensus-specs. How about renaming it to `call_consensus_validated`?

Also suggesting `forkchoice_updated` -> `call_forkchoice_updated`.


ref. https://developers.google.com/style/voice

specs/merge/fork-choice.md
```diff
@@ -32,36 +31,20 @@ This is the modification of the fork choice according to the executable beacon c
...
-def set_head(self: ExecutionEngine, block_hash: Hash32) -> bool:
-    """
-    Returns True if the ``block_hash`` was successfully set as head of the execution payload chain.
-    """
-    ...
-```
+The Engine API may be used to implement it with an external execution engine.
 
-#### `finalize_block`
+#### `forkchoice_updated`
```
hwwhww, line 39: Does that mean we have to update fork-choice specs to call `forkchoice_updated` explicitly? Or leave it implementation-dependent?

i.e., naively, we can checks if beacon head `pre_head != post_head` in `on_tick`, `on_block`. and `on_attestation`. And then get the `state.latest_execution_payload_header.block_hash` to call `forkchoice_updated`.

specs/merge/beacon-chain.md
mkalinin: `engine_consensusValidated` and `engine_forkchoiceUpdated` are rather notifications than methods, we may use the `notify_` or `signal_` prefix for names of functions sending notifications. WDYT?

specs/merge/fork-choice.md
mkalinin: I think we better defer the decision on when to make this call to implementations

specs/merge/beacon-chain.md
```diff
@@ -263,10 +263,15 @@ def execute_payload(self: ExecutionEngine, execution_payload: ExecutionPayload)
 #### `consensus_validated`
 
 ```python
-def consensus_validated(self: ExecutionEngine, execution_payload: ExecutionPayload) -> None:
+def consensus_validated(self: ExecutionEngine, block_hash: Hash32, valid: bool) -> None:
     ...
 ```
 
+The call of this function depends on the result of the state transition and must be done when call to the [`state_transition`](../phase0/beacon-chain.md#beacon-chain-state-transition-function) function finishes. The value of the `valid` parameter must be set as follows:
```
djrtwo, line 270:
```suggestion
The inputs to this function depend on the result of the state transition. A call to `consensus_validated` must be made after the [`state_transition`](../phase0/beacon-chain.md#beacon-chain-state-transition-function) function finishes. The value of the `valid` parameter must be set as follows:
```

specs/merge/beacon-chain.md
djrtwo: I'm okay with that for now. Will think about if we want to represent that fully in the `state_transition`. It might be useful for testing.

specs/merge/beacon-chain.md
djrtwo: I prefer `notify_` or `signal_` if we want to move it to the active voice.

djrtwo (commented):
minor suggestion on text around `consensus_validated`

specs/merge/beacon-chain.md
hwwhww: Do you mean also changing the Engine APIs specs?

`notify_` and `signal_` both sound good to me. Maybe `notify_` is more common? 🙂

specs/merge/validator.md
```diff
@@ -60,9 +81,14 @@ All validator responsibilities remain unchanged other than those noted below. Na
...
 #### Constructing the `BeaconBlockBody`
 
-##### Execution Payload
+##### ExecutionPayload
 
-* Set `block.body.execution_payload = get_execution_payload(state, execution_engine, pow_chain)` where:
+To obtain an execution payload, the proposer of a block must take the following actions:
+
+1. Set `payload_id = prepare_execution_payload(state, pow_chain, fee_recipient, execution_engine)`, where:
+    * `state` is the state for which `is_proposer(state, validator_index)` returns `True`
```
terencechain, line 89: Any concern this `state` won't work well with `state.latest_execution_payload_header.block_hash`?

For post merge, normal payload, `state.latest_execution_payload_header.block_hash` changes on per block level. Where `is_proposer(state, validator_index)` can be known at the start of the epoch, and we also recommend calling `prepare_execution_payload` as soon as input parameters become known

specs/merge/beacon-chain.md
djrtwo: probably not change the engine api specs, imo. but making this call still map to the api specs but conform a bit more to our verb usage is good

mkalinin: Yeah, just renaming it here in the beacon spec code. Execution engine is an abstraction in the beacon spec and Engine API client is one of potential implementations of this abstraction. I'll do the renaming

specs/merge/validator.md
```diff
@@ -37,18 +38,38 @@ Please see related Beacon Chain doc before continuing and use them as a referenc
...
+                    fee_recipient: Bytes20) -> uint64:
+    """
+    Returns ``payload_id`` that is used to obtain the execution payload in a subsequent ``get_payload`` call.
+    """
+    ...
+```
+
+#### `get_payload`
+
+Given the `payload_id`, `get_payload` returns the most recent version of the execution payload that
```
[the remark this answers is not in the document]
mkalinin: I don't think it should be in the spec as it's implementation dependent. In the Engine API specification such call to this method is responded with error. But in some other implementation e.g. a client having both layers in one unit of software it could do anything else, like responding with empty payload or so.

specs/merge/validator.md
mkalinin: Right, but `is_proposer` call is bounded to the `state.slot` and in a normal case this state could become known only after the parent block is processed. It's not the state at the beginning of the epoch with artificially increased slot number. I will reformulate this statement to avoid any confusion.

mkalinin: Updated the statement, please, let me know how does it read to you

specs/merge/beacon-chain.md
mkalinin: I was also thinking about test scenarios and the value that spec'ing this call inside of the `state_transition` may bring in. Implementation of a function on such a high level may vary from client to client and additional tweaks would be required to adopt these test vectors. For instance, implementations may make this call outside of the `state_transition()` function, after it finishes its work.

By the end of the day, this method call won't stay uncovered in terms of testing as it should be a part of integration tests.

specs/merge/validator.md
mkalinin: Done

specs/merge/validator.md
mkalinin: Done

specs/merge/beacon-chain.md
mkalinin: Done

specs/merge/beacon-chain.md
djrtwo: > this method call won't stay uncovered in terms of testing as it should be a part of integration tests.

I hear you on that, but in terms of completeness, especially in relation to EIP 3675, the state transition function is not really complete until sending that signal.

I also see the argument for not including it. Can leave as is and assess it's integration as we do more testing.

mkalinin: I hear arguments on having this method call as a part of the state transition. What do you think about `notify_forkchoice_updated`? Should we make calls to this function in the fork choice handlers in a similar vein?

specs/merge/validator.md
djrtwo: looks good to me

djrtwo (approved):
We might consider referencing the EIP-3675 events that the engine calls map to. Thus making a bit of a clearer formal relationship between this spec and the execution-layer EIP

djrtwo (approved):
excellent!

PR #2615, opened 2021-09-21
Remove beacon_block_root field from ShardBlob

Nashatyrev:
Follow up to PR #2590

Since the order of data of an L2 getting into the shard-data layer doesn't matter it makes sense to remove `beacon_block_root` field from the `ShardBlob`

Some comments from @protolambda 

> L2s will still need some kind of anchoring, but if it's part of the L2 validation rules, and respected in fraud proof in rollup case (i.e. bad beacon block root in data is not slashable, just not a useful block), then it's not necessary in protocol layer case
And fee validation is w.r.t. the estimate block it's going to be included in, not the anchor, so it's fine there
Same applies to shard proposer index
The anchor could give a hint of which branch to look at, but ultimately it must be valid and should be included in the canonical chain

### DIFF
--- a/specs/sharding/p2p-interface.md
+++ b/specs/sharding/p2p-interface.md
@@ -97,11 +97,6 @@ on the horizontal subnet or creating samples for it. Alias `blob = signed_blob.m
 - _[REJECT]_ The blob builder defined by `blob.builder_index` exists and has sufficient balance to back the fee payment.
 - _[REJECT]_ The blob signature, `signed_blob.signature`, is valid for the aggregate of proposer and builder --
   i.e. `bls.FastAggregateVerify([builder_pubkey, proposer_pubkey], blob_signing_root, signed_blob.signature)`.
-- _[REJECT]_ The blob is proposed by the expected `proposer_index` for the blob's `slot` and `shard`,
-  in the context of the current shuffling (defined by `blob.body.beacon_block_root`/`slot`).
-  If the `proposer_index` cannot immediately be verified against the expected shuffling,
-  the blob MAY be queued for later processing while proposers for the blob's branch are calculated --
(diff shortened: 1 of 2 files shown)


djrtwo (approved):
I love deleting code! 🙌

Nashatyrev:
> 
> 
> I love deleting code! 🙌

Oh yeah!

specs/sharding/p2p-interface.md
```diff
@@ -160,11 +150,6 @@ The following validations MUST pass before forwarding the `signed_blob_header` o
   Propagating nodes MAY increase fee increments in case of spam.
 - _[REJECT]_ The header signature, `signed_blob_header.signature`, is valid for ONLY the builder --
   i.e. `bls.Verify(builder_pubkey, blob_signing_root, signed_blob_header.signature)`. The signature is not an aggregate with the proposer.
-- _[REJECT]_ The header is designated for proposal by the expected `proposer_index` for the blob's `header.slot` and `header.shard`
```
protolambda, line 163: We still need to verify the shard proposer index (in other p2p topics too). Assuming we want the index to be embedded at all (I like it better, as the shuffling is not available the same way, and it allows for validating the signature ahead of the shuffling, just like with beacon blocks)

protolambda (changes requested):
Yay for removing code. But let's not break the p2p flood protection

specs/sharding/p2p-interface.md
Nashatyrev: OMG, deleted those even without reading. Sorry, fixing this...

Nashatyrev: Fixed here: df01588

protolambda (approved):
LGTM!

PR #2617, opened 2021-09-22
Add terminal block hash override

lsankar4033:
This is a rewrite of https://github.com/ethereum/consensus-specs/pull/2599 in consideration of https://github.com/ethereum/consensus-specs/pull/2605

Basically, hardcode a `TERMINAL_BLOCK_HASH` parameter that can be used to override the transition point implied by `TERMINAL_TOTAL_DIFFICULTY`


specs/merge/beacon-chain.md
```diff
@@ -83,6 +83,7 @@ This patch adds transaction execution to the beacon chain as part of the Merge f
 | Name | Value |
 | - | - |
 | `TERMINAL_TOTAL_DIFFICULTY` | **TBD** |
+| `TERMINAL_BLOCK_HASH` | `Bytes32('0x0000000000000000000000000000000000000000000000000000000000000000')` |
```
mkalinin, line 86:
```suggestion
| `TERMINAL_BLOCK_HASH` | `Hash32('0x0000000000000000000000000000000000000000000000000000000000000000')` |
```

specs/merge/validator.md
```diff
@@ -74,6 +74,14 @@ def get_pow_block_at_total_difficulty(total_difficulty: uint256, pow_chain: Sequ
 
     return None
 
+def get_terminal_pow_block(total_difficulty: uint256, block_hash_override: Hash32, pow_chain: Sequence[PowBlock]) -> Optional[PowBlock]:
```
mkalinin, line 77: I think we could use `TERMINAL_*` config parameters directly for clarity, the function could look as follows:
```python
def get_terminal_pow_block(pow_chain: Sequence[PowBlock]) -> Optional[PowBlock]:
    # Check block hash override prior to terminal total difficulty
    if TERMINAL_BLOCK_HASH != Hash32():
        pow_block_overrides = [pow_block for pow_block in pow_chain if pow_block.block_hash == TERMINAL_BLOCK_HASH]
        if len(pow_block_overrides) != 0:
            return pow_block_overrides[0]
        else:
            return None

    return get_pow_block_at_total_difficulty(TERMINAL_TOTAL_DIFFICULTY, pow_chain)
```

mkalinin (commented):
Looks good to me! Left a couple of suggestions

specs/merge/validator.md
lsankar4033: ya agreed, wil change

lsankar4033: (and nice catch on the slight logic issue)

specs/merge/client-settings.md
```diff
@@ -19,3 +20,9 @@ To coordinate manual overrides to [`TERMINAL_TOTAL_DIFFICULTY`](./beacon-chain.m
 
 Except under exceptional scenarios, this setting is expected to not be used. Sufficient warning to the user about this exceptional configurable setting should be provided.
 
+### Override terminal block hash
+
+To allow for fork coordination around a specific PoW block, clients must also provide `--terminal-block-hash-override` as a configurable setting.
```
djrtwo, line 25: suggest removing "fork" here. There is a hardfork in beacon chain logic and then a subsequent transition from pow to pos (something like a fork of course...) but suggest "transition" to avoid name collision in this process

```suggestion
To allow for transition coordination around a specific PoW block, clients must also provide `--terminal-block-hash-override` as a configurable setting.
```

specs/merge/validator.md
```diff
@@ -65,15 +65,26 @@ All validator responsibilities remain unchanged other than those noted below. Na
...
         parent = get_pow_block(block.parent_hash)
-        if block.total_difficulty >= total_difficulty and parent.total_difficulty < total_difficulty:
+        if block.total_difficulty >= TERMINAL_TOTAL_DIFFICULTY and parent.total_difficulty < TERMINAL_TOTAL_DIFFICULTY:
             return block
 
     return None
 
+def get_terminal_pow_block(pow_chain: Sequence[PowBlock]) -> Optional[PowBlock]:
+    if TERMINAL_BLOCK_HASH != Hash32():
+        # Check block hash override prior to total difficulty
```
djrtwo, line 79:
```suggestion
        # Terminal lock hash override takes precedence over terminal total difficulty
```

djrtwo (commented):
A couple of minor suggestions. plus lint

djrtwo:
okay, I patched up the lint too! will merge after CI

lsankar4033:
ah, thanks!

specs/merge/client-settings.md
lsankar4033: good point

djrtwo (approved):

PR #2619, opened 2021-09-23
fix merkle proof test for mainnet

etan-status:
The `test_next_sync_committee_tree` currently only supports the minimal
preset, as it incorrectly initializes the `next_sync_committee`. On the
mainnet preset, `SYNC_COMMITTEE_SIZE` is 512, but the default states use
only 256 validators, leading to an IndexError during the test execution.
`next_sync_committee` is already initialized correctly prior to the test
run using the spec's `get_next_sync_committee` function, which fills up
extra committee slots with duplicate validators in this scenario. This
makes it unnecessary to manually initialize the `next_sync_committee`.
Removed the incorrect initialization to allow testing on mainnet preset.


hwwhww (approved):
You're right. Thank you for making this PR! 👍

Issue #2618, opened 2021-09-23
Merge: beacon block gossip verification restricts forking


paulhauner:
In the [p2p-interface](https://github.com/ethereum/consensus-specs/blob/4c1156d504279069799924dbbc79cf13d9454ae3/specs/merge/p2p-interface.md) specifications for The Merge, the following rule applies during `beacon_block` gossip verification:

> - If the merge is complete with respect to the head state -- i.e. is_merge_complete(state) -- then validate the following:
>     - [REJECT] The block's execution payload must be non-empty -- i.e. execution_payload != ExecutionPayload()

I'm curious about why a post-merge **head state** can prohibit the import of pre-merge blocks.

Consider the following scenario:

- A beacon proposer (`BP1`) is on a minority fork of the execution chain. That minority fork has crossed the `total_terminal_difficulty` (TTD) threshold.
- `BP1` produces a block, validates it and sets it as its own head.
- The beacon proposer for the next slot (`BP2`) was not on the minority execution fork and therefore rejected the block from `BP1`
- `BP2` produces a block that conflicts with the one from `BP1` *and* does not trigger the merge.
- The execution node of `BP1` converges with the network and reverts to the same chain as `BP2`
- `BP1` will reject the block from `BP2`, even though it is valid and may be destined to become the majority chain.

I think that in practice `BP1` would end up importing the block from `BP2` via RPC, but it seems a little odd to rely on that procedure.

Ultimately, my question is:

Do we desire this "latching" which prevents us from importing and propagating a non-merge gossip block after we've elected a merge block as the head?

mkalinin:
Good point! I am not sure what the original intention was but this rule should be applicable to descendants of the transition block and not every block in the system. It basically rejects post-merge blocks with empty payload.

djrtwo:
resolved in #2621

PR #2621, opened 2021-09-23
Modify Merge Gossip Block Validation Conditions

ethDreamer:
This PR addresses the first validation condition as it is currently written:

 - If the merge is complete with respect to the **head state** -- i.e. `is_merge_complete(state)` -- then validate the following:
    - [REJECT] The block's execution payload must be non-empty -- i.e. `execution_payload != ExecutionPayload()`

As mentioned by @paulhauner in #2618, this condition might cause problems when there are forks in the POW chain at the transition boundary, potentially causing clients to initially reject and fail to propagate blocks that might be ultimately destined to become a part of the canonical chain.

It seems likely that this condition was originally intended to apply to all descendants of the transition block. Therefore it should be validated against the `BeaconState` associated with the block being imported rather than the head state.  If that is true, the validation conditions simplify to this.

### DIFF
--- a/specs/merge/p2p-interface.md
+++ b/specs/merge/p2p-interface.md
@@ -68,10 +68,8 @@ See the Merge [state transition document](./beacon-chain.md#beaconblockbody) for
 In addition to the gossip validations for this topic from prior specifications,
 the following validations MUST pass before forwarding the `signed_beacon_block` on the network.
 Alias `block = signed_beacon_block.message`, `execution_payload = block.body.execution_payload`.
-- If the execution is enabled for the block or the block state -- i.e. `is_execution_enabled(state, block.body)`
+- If the execution is enabled for the block -- i.e. `is_execution_enabled(state, block.body)`
   then validate the following:
-  - _[REJECT]_ The block's execution payload must be non-empty --
-    i.e. `execution_payload != ExecutionPayload()`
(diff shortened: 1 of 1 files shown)


paulhauner:
I think we could just remove that first condition all together, the second condition should cover it alone.

djrtwo:
Meaning, retain "If the execution is enabled for the block"?

Ah, I see what you mean. The next few field conditions actually implicitly validate that the block is non-empty (e.g. `execution_payload.timestamp == compute_time_at_slot(state, block.slot)`)

djrtwo:
I was going to write the following yesterday but didn't click send. Although I think the below is correct, validating non-empty is of dubious value, requiring much more complexity to state correctly.

--------

This is more accurately the "block pre-state" (which would either have a `latest_execution_payload_header` or not, thus dictating `is_merge_complete`), right? We need to make sure we don't imply that the state must be transitioned using the `block` just received to validate this condition. 

And what we want to say in text that "if this block or any ancestor of this block turned on execution, then this block needs to be validated against the execution rules".

suggestion below:
```suggestion
- If the execution is enabled for the block or any ancestor of the block -- i.e. `is_execution_enabled(state, block.body)` where `state` is the pre-state of `block` --
```

djrtwo:
I made the change to just remove to the top condition 👍

djrtwo (approved):

PR #2622, opened 2021-09-24
No longer importing "Paragraph"

fredriksvantes:
removed Paragraph from being imported from marko.ext.gfm.elements as it is not being used


djrtwo (approved):
thanks!

PR #2627, opened 2021-09-24
Fixed Validation Conditions (Function Renamed)

ethDreamer:


djrtwo (approved):
looks like the same functionality is defined in phsae0 and merge with different function names... 

Will merge this as it pertains to merge and matches the name used there

PR #2629, opened 2021-09-27
merkle proof test generator

etan-status:
Building merkle proofs is required functionality for implementing light
client sync. Although the spec currently only defines a function to
verify merkle proofs (`is_valid_merkle_branch`) there are still a few
PySpec unit tests that produce merkle proofs. This patch adds a new
generator to extract test vectors from those static unit tests, so that
light client implementations can validate their merkle proof logic.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/altair/merkle/test_single_proof.py
+++ b/tests/core/pyspec/eth2spec/test/altair/merkle/test_single_proof.py
@@ -8,12 +8,14 @@ from eth2spec.test.helpers.merkle import build_proof
 
 @with_phases([ALTAIR])
 @spec_state_test
-def test_next_sync_committee_tree(spec, state):
+def test_next_sync_committee_merkle_proof(spec, state):
     yield "state", state
-    yield "leaf", state.next_sync_committee.hash_tree_root()
-    yield "leaf_index", "meta", spec.NEXT_SYNC_COMMITTEE_INDEX
     next_sync_committee_branch = build_proof(state.get_backing(), spec.NEXT_SYNC_COMMITTEE_INDEX)
-    yield "proof", next_sync_committee_branch
+    yield "proof", {
+        "leaf": "0x" + state.next_sync_committee.hash_tree_root().hex(),
+        "leaf_index": spec.NEXT_SYNC_COMMITTEE_INDEX,
+        "branch": ['0x' + root.hex() for root in next_sync_committee_branch]
+    }
     assert spec.is_valid_merkle_branch(
         leaf=state.next_sync_committee.hash_tree_root(),
         branch=next_sync_committee_branch,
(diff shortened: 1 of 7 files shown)


protolambda:
Hey @etan-status, thank you for your testing contribution! I moved the test to its own package (it's not just a spec unit-test, since it outputs vectors), and updated the format to not split it into so many different files. Also, I updated it to output into a `single_proof` handler of the `merkle` runner. We can then add more types of merkle testing later. What do you think?

etan-status:
Thanks, appreciate the update! Makes this entire thing quite a bit cleaner :-)

tests/core/pyspec/eth2spec/test/altair/merkle/test_single_proof.py
```diff
@@ -8,8 +8,14 @@
...
 @with_phases([ALTAIR])
 @spec_state_test
-def test_next_sync_committee_tree(spec, state):
+def test_next_sync_committee_merkle_proof(spec, state):
+    yield "state", state
     next_sync_committee_branch = build_proof(state.get_backing(), spec.NEXT_SYNC_COMMITTEE_INDEX)
+    yield "proof", {
+        "leaf": "0x" + state.next_sync_committee.hash_tree_root().hex(),
+        "leaf_index": spec.NEXT_SYNC_COMMITTEE_INDEX,
+        "branch": ['0x' + root.hex() for root in next_sync_committee_branch]
```
etan-status, line 17: Is `root` the common name even for intermediate hashes?

protolambda (approved):
LGTM

tests/core/pyspec/eth2spec/test/altair/merkle/test_single_proof.py
protolambda: In remerkleable every 32-byte tree thing is a "root", (`Root` type). And the tree is built with `PairNode` and `RootNode` types. If you take the merkle-root of any subtree (even if just a single node) it's a `Root`. It's a little confusing with the "leaf/sibling" meanings, but it's the opposite of a pair, so I stuck with it. Naming is hard.

tests/core/pyspec/eth2spec/test/altair/merkle/test_single_proof.py
```diff
@@ -8,8 +8,14 @@
 
 @with_phases([ALTAIR])
```
djrtwo, line 9: I would default to doing `@with_all_phases_except([PHASE0])` and if the beaconstate changes enough in subsequent phase to break this test, it will alert us to either make this more modular or to add additional merkle tests for new phase

djrtwo (commented):
one minor suggestion

tests/core/pyspec/eth2spec/test/altair/merkle/test_single_proof.py
etan-status: Good point, but more of a separate issue that already existed before this PR.

etan-status: https://github.com/ethereum/consensus-specs/pull/2635

PR #2633, opened 2021-09-28
Expand If_and_only_if in `execute_payload`

terencechain:


specs/merge/beacon-chain.md
```diff
@@ -257,7 +257,7 @@ The Engine API may be used to implement them with an external execution engine.
 ```python
 def execute_payload(self: ExecutionEngine, execution_payload: ExecutionPayload) -> bool:
     """
-    Returns ``True`` iff ``execution_payload`` is valid with respect to ``self.execution_state``.
+    Returns ``True`` if ``execution_payload`` is valid with respect to ``self.execution_state``.
```
hwwhww, line 260: This is not a typo, see https://github.com/ethereum/consensus-specs/pull/2472/files#r649255103
https://en.wikipedia.org/wiki/If_and_only_if

hwwhww: It seems many people were confused. Maybe we can just use the full "if and only if".
```suggestion
    Return ``True`` if and only if ``execution_payload`` is valid with respect to ``self.execution_state``.
```

djrtwo (approved):
yeah, not the first time someone thought this was a typo.

Full text 👍

PR #2634, opened 2021-09-28
Fix `initialize_beacon_state_from_eth1` previous_version

hwwhww:
Thank @tbenr for finding this error!

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -689,7 +689,7 @@ def initialize_beacon_state_from_eth1(eth1_block_hash: Bytes32,
                                       eth1_timestamp: uint64,
                                       deposits: Sequence[Deposit]) -> BeaconState:
     fork = Fork(
-        previous_version=GENESIS_FORK_VERSION,
+        previous_version=ALTAIR_FORK_VERSION,  # [Modified in Altair] for testing only
         current_version=ALTAIR_FORK_VERSION,  # [Modified in Altair]
         epoch=GENESIS_EPOCH,
     )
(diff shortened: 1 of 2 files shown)


djrtwo:
On second thought, I’m not convinced what correct behavior for this versioning is.

* compute_domain() in `process_deposit` is “fork agnostic” and is *always* based on `GENESIS_FORK_VERSION`
* There is no prior fork when initializing a chain so the use of ALTAIR does not immediately seem correct
* The primary usecase of previous_version is to verify signatures from the previous fork, but there is no previous history prior to the genesis in this case

Because of the above, I’d argue to either initialize it to `GENESIS_FORK_VERSION` because deposits use this OR just initialize it to `MERGE_FORK_VERSION` which I think is a bit more semantically correct when looking at how the state in initialized in phase 0.

The altair analogue of this function uses `GENESIS_FORK_VERSION` so double `MERGE_FORK_VERSION` isn’t consistent unless we change the function in altair as well.

tbenr:
my two cents (just because I raised the problem).
the second option seems preferable to me:
```python
current_version=MERGE_FORK_VERSION,  # [Modified in Merge]
previous_version=current_version
```

I was also exploring the idea about introducing a non-configurable constant `NO_FORK_VERSION=Version('0x00000000')`. Using this should maintain phase0-MainNet compatibility but I guess brakes pahse0-devnets.

hwwhww:
My institution was making it more like a mainnet real case, but you convinced me that `MERGE_FORK_VERSION` is better for the testing context.

If possible, I'd like to also change Altair `initialize_beacon_state_from_eth1` to using `previous_version=ALTAIR_FORK_VERSION`. 

Note: it has nothing to do with the mainnet consensus, just for testing.

hwwhww:
@djrtwo @tbenr 

I changed it to `previous_version=current_version`.

djrtwo (approved):

djrtwo:
Agreed that it's reasonable to adjust the altair function as well.

This affects not only testing, but also testnets/networks that deploy their genesis from altair logic. I don't think we have any such nets today, so it's okay to introduce this change

PR #2635, opened 2021-09-28
run tests against future forks by default

etan-status:
Some tests are currently restricted to a single phase using @with_phases
even though they could likely run unchanged in later phases. This patch
changes the default for such tests to also run in later phases. If the
beacon chain changes enough in later phases to break these tests, this
highlights that the tests need to be adjusted or extended accordingly.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/altair/merkle/test_single_proof.py
+++ b/tests/core/pyspec/eth2spec/test/altair/merkle/test_single_proof.py
@@ -1,12 +1,11 @@
 from eth2spec.test.context import (
     spec_state_test,
-    with_phases,
+    with_altair_and_later,
 )
-from eth2spec.test.helpers.constants import ALTAIR
 from eth2spec.test.helpers.merkle import build_proof
 
 
-@with_phases([ALTAIR])
+@with_altair_and_later
 @spec_state_test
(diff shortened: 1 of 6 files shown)


tests/core/pyspec/eth2spec/test/merge/fork_choice/test_on_merge_block.py
```diff
@@ -45,7 +48,7 @@ def wrap(flag: AtomicBoolean):
     assert is_called.value
 
 
-@with_phases([MERGE])
+@with_merge_and_later
```
hwwhww, line 51: I thought `test_on_merge_block.py` tests are MERGE-phase-only?
/cc @zilm13

dapplion (approved):

tests/core/pyspec/eth2spec/test/merge/fork_choice/test_on_merge_block.py
etan-status: TBH, I'm not 100% sure on these, I'll revert the `test_on_merge_block.py` changes for now until there is clarity.

hwwhww (approved):
lgtm, thank you!

PR #2598, opened 2021-09-14
New merge tests

zilm13:
Added some merge tests, all I was able to discover. Not a good test creator though
Covering:
- more bad cases with `process_execution_payload`
- unit tests on `is_merge_complete`, `is_merge_block`, `is_execution_enabled`
- testing out `compute_terminal_total_difficulty`'s `max` option
- `is_valid_terminal_pow_block` tests and `process_merge_execution_payload`

The latter one is a new method combining all checks for terminal block in `on_block` handler. Appropriate changes to spec was made too. Also fixed minor type misalign in  `compute_terminal_total_difficulty`.
@mkalinin @protolambda Please, take a look

PR #2637, opened 2021-09-30
[WIP] Mock Engine API req/resp in tests

hwwhww:
Based on @zilm13's new merge fork-choice tests in #2598/#2630, I tried to generate the expected Engine API request & mock the response.

It would add `engine_api` event in `steps.yaml`.
A sample `steps.yaml`:
```yaml
- {tick: 0}
- {pow_block: pow_block_0x63d026870b331a9269cb121f9ad525d7d63bcaf2c29f12e6e50288b89a7b9bd7}
- {pow_block: pow_block_0x535d43dc06e80498afc0dc98fa40de8c19222d5f187705ed5de10934168bd71d}
- engine_api:
    request:
      method: engine_preparePayload
      params: {parentHash: '0x63d026870b331a9269cb121f9ad525d7d63bcaf2c29f12e6e50288b89a7b9bd7',
        timestamp: 0, random: '0xdadadadadadadadadadadadadadadadadadadadadadadadadadadadadadadada',
        feeRecipient: '0x1212121212121212121212121212121212121212'}
    response:
      error: false
      result: {payloadId: 1}
- {tick: 12}
- {block: block_0x6358c748c76d08954fa28ee985ca6eae97f50f5ea19eb198eb97b01bd6e8e78c}
- checks:
    time: 12
    head: {slot: 1, root: '0xf81992b35f893a54c6c9848a16cd05b45a2cb483265e9e2745b71b389d162632'}
    justified_checkpoint: {epoch: 0, root: '0x8a099f1071df9b9dcd6dd5082d3e9362560d11e066ef08aac508e83a97b81022'}
    finalized_checkpoint: {epoch: 0, root: '0x8a099f1071df9b9dcd6dd5082d3e9362560d11e066ef08aac508e83a97b81022'}
    best_justified_checkpoint: {epoch: 0, root: '0x8a099f1071df9b9dcd6dd5082d3e9362560d11e066ef08aac508e83a97b81022'}
```

My idea is that the consensus client teams can:
1. Compare if the expected API requests match
2. Use the given mocking response to continue the next step
3. Test against the edge cases in CI

**Question to client teams**: Would it be helpful and worth developing?
Feedback would be appreciated. 🙏

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -537,7 +537,7 @@ class NoopExecutionEngine(ExecutionEngine):
                         parent_hash: Hash32,
                         timestamp: uint64,
                         random: Bytes32,
-                        feeRecipient: ExecutionAddress) -> PayloadId:
+                        fee_recipient: ExecutionAddress) -> PayloadId:
         raise NotImplementedError("no default block production")
 
     def get_payload(self: ExecutionEngine, payload_id: PayloadId) -> ExecutionPayload:
--- a/tests/core/pyspec/eth2spec/test/helpers/engine_apis.py
+++ b/tests/core/pyspec/eth2spec/test/helpers/engine_apis.py
@@ -1,0 +1,67 @@
+from eth_utils import encode_hex
+
+
+def to_json_rpc_request(method, params):
+    return {
+        'method': method,
+        'params': params,
+    }
+
+
+def to_json_rpc_response(result, error=False):
+    return {
+        'error': error,
+        'result': result,
+    }
+
+
+def with_mock_engine_prepare_payload(spec,
+                                     parent_hash,
+                                     timestamp,
+                                     random,
+                                     fee_recipient,
+                                     payload_id,
+                                     func,
+                                     test_steps):
+    def prepare_payload(parent_hash,
+                        timestamp,
+                        random,
+                        fee_recipient):
+        req = to_json_rpc_request(
+            method='engine_preparePayload',
+            params={
+                "parentHash": encode_hex(parent_hash),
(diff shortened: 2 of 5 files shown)


mkalinin:
I am a little bit hesitant on mocking execution engine this way because it ties consensus specs to the Engine API implementation of EE. We could have a similar effect by mocking the `ExecutionEngine` abstraction. Thoughts?

hwwhww:
Thanks for the feedback @mkalinin 

> it ties consensus specs to the Engine API implementation of EE.

It does tie to the interface of Engine APIs, but we can set arbitrary responses from the consensus side!

> We could have a similar effect by mocking the ExecutionEngine abstraction

Do you mean like [disposable `TestEngine` in `test_process_execution_payload` tests](https://github.com/ethereum/consensus-specs/blob/d34b79f4dead04eb9cd5daf94cc87c8ec055df7e/tests/core/pyspec/eth2spec/test/merge/block_processing/test_process_execution_payload.py#L28-L40)? Or a more reusable and general `ExecutionEngine` cross consensus pyspec tests?

This PR also mock disposable `ExecutionEngine`, the difference is it makes mocking work with higher level `state_transition` function instead of `process_execution_payload` function by doing `spec.EXECUTION_ENGINE = my_mock_engine`.

I chose fork choice test format because it would cover most Engine APIs. Since the INVALID cases would be covered by granular `process_execution_payload` operation tests, I think for most `engine_executePayload` calls in the new tests, we can just return VALID.

PR #2638, opened 2021-10-01
use correct timestamp for empty ExecutionPayload

etan-status:
There are two similar functions to compute the timestamp for a given
beacon chain slot. `compute_time_at_slot` is used for processing Eth1
votes, and does not take into account `GENESIS_TIME`. The other one,
`compute_timestamp_at_slot`, is used everywhere else. When processing
`ExecutionPayload`, the `merge/beacon-chain.md` spec uses the latter,
`compute_timestamp_at_slot`, to verify the timestamp. However, in the
test code, `build_empty_execution_payload` uses `compute_time_at_slot`.
This patch changes the test to use the same function for creating the
timestamp that is later used to verify it. Note that `GENESIS_TIME` is 0
so there is no practical difference.


hwwhww (approved):
LGTM 👍

PR #2640, opened 2021-10-03
 initialize_beacon_state_from_eth1 for pre-transition merge

djrtwo:
`initialize_beacon_state_from_eth1` on the Merge fork should work for two cases:
1. The Merge beacon chain logic at genesis and the merge has already happened (what is in the function today)
2. The Merge beacon chain logic at genesis but the merge has *not* yet happened

(2) is very valuable for testnets that are built to test the merge transition logic

TODO:
- [x] fix tests

### DIFF
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -365,7 +365,8 @@ Modifications include:
 def initialize_beacon_state_from_eth1(eth1_block_hash: Bytes32,
                                       eth1_timestamp: uint64,
                                       deposits: Sequence[Deposit],
-                                      execution_payload_header: ExecutionPayloadHeader) -> BeaconState:
+                                      execution_payload_header: ExecutionPayloadHeader=ExecutionPayloadHeader()
+                                      ) -> BeaconState:
     fork = Fork(
         previous_version=MERGE_FORK_VERSION,  # [Modified in Merge] for testing only
         current_version=MERGE_FORK_VERSION,  # [Modified in Merge]
@@ -403,6 +404,7 @@ def initialize_beacon_state_from_eth1(eth1_block_hash: Bytes32,
     state.next_sync_committee = get_next_sync_committee(state)
 
     # [New in Merge] Initialize the execution payload header
+    # If empty, will initialize a chain that has not yet gone through the Merge transition
     state.latest_execution_payload_header = execution_payload_header
 
     return state
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -355,11 +355,11 @@ def process_execution_payload(state: BeaconState, payload: ExecutionPayload, exe
 
 *Note*: The function `initialize_beacon_state_from_eth1` is modified for pure Merge testing only.
 Modifications include:
-1. Use `MERGE_FORK_VERSION` as the current fork version
-2. Utilize the Merge `BeaconBlockBody` when constructing the initial `latest_block_header`
+1. Use `MERGE_FORK_VERSION` as the current fork version.
+2. Utilize the Merge `BeaconBlockBody` when constructing the initial `latest_block_header`.
 3. Initialize `latest_execution_payload_header`.
   If `execution_payload_header == ExecutionPayloadHeader()`, then the Merge has not yet occurred.
-  Else, the Merge starts from genesis.
+  Else, the Merge starts from genesis and the transition is incomplete.
 
 ```python
 def initialize_beacon_state_from_eth1(eth1_block_hash: Bytes32,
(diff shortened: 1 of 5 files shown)


specs/merge/beacon-chain.md
```diff
@@ -354,13 +354,18 @@ def process_execution_payload(state: BeaconState, payload: ExecutionPayload, exe
...
+3. Initialize `latest_execution_payload_header`.
+  If `execution_payload_header == ExecutionPayloadHeader()`, then the Merge has not yet occurred.
+  Else, the Merge starts from genesis.
 
 ```python
 def initialize_beacon_state_from_eth1(eth1_block_hash: Bytes32,
                                       eth1_timestamp: uint64,
-                                      deposits: Sequence[Deposit]) -> BeaconState:
+                                      deposits: Sequence[Deposit],
+                                      execution_payload_header: ExecutionPayloadHeader) -> BeaconState:
```
mkalinin, line 368: Do we want to have a default value for `execution_payload_header`? It would then be compatible with the previous versions of this function but could hide some bugs as if related function calls are not updated whenever they are supposed to be it would return the state after the merge fork and before the transition

hwwhww: `execution_payload_header: ExecutionPayloadHeader=ExecutionPayloadHeader()` probably could fix the tests immediately:

```python
def initialize_beacon_state_from_eth1(
        eth1_block_hash: Bytes32,
        eth1_timestamp: uint64,
        deposits: Sequence[Deposit],
        execution_payload_header: ExecutionPayloadHeader=ExecutionPayloadHeader()) -> BeaconState:
```

tests/core/pyspec/eth2spec/test/merge/genesis/test_initialization.py
```diff
@@ -17,7 +19,7 @@ def eth1_init_data(eth1_block_hash, eth1_timestamp):
     }
 
 
-@with_merge_and_later
+@with_phases([MERGE])
```
hwwhww, line 22: I think for the subsequent forks, the `execution_payload_header` field would be required?

hwwhww (approved):
@djrtwo 

I cleaned up the tests & testgen. I've tested that I can generate the new test vectors successfully.

Signing my approval now so that you can review & merge. :)

specs/merge/beacon-chain.md
```diff
@@ -397,12 +403,9 @@ def initialize_beacon_state_from_eth1(eth1_block_hash: Bytes32,
     state.current_sync_committee = get_next_sync_committee(state)
     state.next_sync_committee = get_next_sync_committee(state)
 
-    # [New in Merge] Initialize the execution payload header (with block number set to 0)
-    state.latest_execution_payload_header.block_hash = eth1_block_hash
-    state.latest_execution_payload_header.timestamp = eth1_timestamp
-    state.latest_execution_payload_header.random = eth1_block_hash
-    state.latest_execution_payload_header.gas_limit = GENESIS_GAS_LIMIT
-    state.latest_execution_payload_header.base_fee_per_gas = GENESIS_BASE_FEE_PER_GAS
```
dapplion, line 405: Should remove these stale constants

mkalinin (approved):
Looks good to me! 👍

tests/core/pyspec/eth2spec/test/merge/genesis/test_initialization.py
djrtwo: I personally think the transition logic might just stay in there, but we can debate it further down the line

djrtwo (commented):
Thank you @hwwhww !!

specs/merge/beacon-chain.md
```diff
@@ -354,13 +344,19 @@ def process_execution_payload(state: BeaconState, payload: ExecutionPayload, exe
...
+3. Initialize `latest_execution_payload_header`.
+  If `execution_payload_header == ExecutionPayloadHeader()`, then the Merge has not yet occurred.
+  Else, the Merge starts from genesis and the transition is incomplete.
 
 ```python
 def initialize_beacon_state_from_eth1(eth1_block_hash: Bytes32,
                                       eth1_timestamp: uint64,
-                                      deposits: Sequence[Deposit]) -> BeaconState:
+                                      deposits: Sequence[Deposit],
+                                      execution_payload_header: ExecutionPayloadHeader=ExecutionPayloadHeader()
```
Nashatyrev, line 358: Why do we still need `eth1_block_hash` & `eth1_timestamp`? They should be both in the passed `execution_payload_header`

NizamShah7 (approved):

PR #2645, opened 2021-10-05
Specify format for --terminal-total-difficulty

paulhauner:
I think it is prudent to specify how the `--terminal-total-difficulty-override` value should be provided.

I've chosen decimal since that's how Etherscan presents it and it seems a little more user friendly (well, as user-friendly as a 23-digit number gets :sweat_smile:).

### DIFF
--- a/specs/merge/client-settings.md
+++ b/specs/merge/client-settings.md
@@ -16,7 +16,7 @@ This document specifies configurable settings that clients must implement for th
 
 ### Override terminal total difficulty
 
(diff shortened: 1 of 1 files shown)


mcdee:
Note that a decimal value could be considered ambiguous (_e.g._ `100` could be decimal or hex), whereas a hex value preceded with `0x` is not.

down4some1 (approved):

down4some1 (commented):
thx 4 watching!

protolambda (approved):
LGTM.

PR #2657, opened 2021-10-08
merge: Genesis block should include the genesis execution payload

asn-d6:
Right now there is no function to get the genesis block in `merge`, and the genesis block code snippet from `phase0` creates a genesis block with no execution payload header.

This PR introduces functions that get the genesis block from the genesis state both for merge and phase0. It also adds a unittest that checks that the function behaves well in merge.

With help from @ralexstokes and @djrtwo .

### DIFF
--- a/tests/core/pyspec/eth2spec/test/merge/genesis/test_initialization.py
+++ b/tests/core/pyspec/eth2spec/test/merge/genesis/test_initialization.py
@@ -120,3 +120,51 @@ def test_initialize_post_transition(spec):
     assert spec.is_merge_complete(state)
 
     yield 'state', state
+
+
+@with_merge_and_later
+@spec_test
+@single_phase
+@with_presets([MINIMAL], reason="too slow")
+def test_genesis_block_execution_payload(spec):
+    """
+    Test that the merge genesis block includes the execution payload that was
+    used to make the genesis state.
+    """
+    # Create a genesis execution payload header with no transactions
+    genesis_execution_header = get_sample_genesis_execution_payload_header(spec)
+    empty_transactions_root = spec.List[spec.Transaction, spec.MAX_TRANSACTIONS_PER_PAYLOAD]().hash_tree_root()
+    genesis_execution_header.transactions_root = empty_transactions_root
+
+    # Initialize data required to make state/block
+    eth1_block_hash = b'\x12' * 32
+    eth1_timestamp = spec.config.MIN_GENESIS_TIME
+    deposit_count = spec.config.MIN_GENESIS_ACTIVE_VALIDATOR_COUNT
+    deposits, deposit_root, _ = prepare_full_genesis_deposits(
+        spec,
+        spec.MAX_EFFECTIVE_BALANCE,
+        deposit_count,
+        signed=True,
+    )
+
+    # Get genesis state
+    genesis_state = spec.initialize_beacon_state_from_eth1(
+        eth1_block_hash,
+        eth1_timestamp,
+        deposits,
+        execution_payload_header=genesis_execution_header
+    )
+
+    # Get genesis block
+    genesis_block = spec.get_genesis_block_from_genesis_state(genesis_state)
+
+    # Test that genesis block and genesis state refer to the same execution payload
+    expected_block_exec_payload = genesis_block.body.execution_payload
+    expected_state_exec_payload_header = genesis_state.latest_execution_payload_header
+    assert expected_block_exec_payload.hash_tree_root() == expected_state_exec_payload_header.hash_tree_root()
+
+    # Check that the payload we generated is the one that is embedded in the state
+    assert genesis_execution_header.hash_tree_root() == expected_state_exec_payload_header.hash_tree_root()
+
+    # Check that the genesis block points to the genesis state
+    assert genesis_block.state_root == genesis_state.hash_tree_root()
--- a/tests/core/pyspec/eth2spec/test/phase0/genesis/test_initialization.py
+++ b/tests/core/pyspec/eth2spec/test/phase0/genesis/test_initialization.py
@@ -1,11 +1,15 @@
 from eth2spec.test.context import (
     is_post_altair,
     single_phase,
+    with_phases,
     spec_test,
     with_presets,
     with_all_phases,
 )
-from eth2spec.test.helpers.constants import MINIMAL
+from eth2spec.test.helpers.constants import (
+    MINIMAL,
+    PHASE0
+)
 from eth2spec.test.helpers.deposits import (
     prepare_full_genesis_deposits,
     prepare_random_genesis_deposits,
@@ -216,3 +220,36 @@ def test_initialize_beacon_state_random_valid_genesis(spec):
     assert spec.is_valid_genesis_state(state)
 
     yield 'state', state
+
+
+@with_phases([PHASE0])
+@spec_test
+@single_phase
+@with_presets([MINIMAL], reason="too slow")
+def test_genesis_block_creation(spec):
+    """
+    Test the get_genesis_block_from_genesis_state() function
+    """
+    # Initialize data required to make state/block
+    eth1_block_hash = b'\x12' * 32
+    eth1_timestamp = spec.config.MIN_GENESIS_TIME
+    deposit_count = spec.config.MIN_GENESIS_ACTIVE_VALIDATOR_COUNT
+    deposits, deposit_root, _ = prepare_full_genesis_deposits(
+        spec,
+        spec.MAX_EFFECTIVE_BALANCE,
+        deposit_count,
+        signed=True,
+    )
+
+    # Get genesis state
+    genesis_state = spec.initialize_beacon_state_from_eth1(
+        eth1_block_hash,
+        eth1_timestamp,
+        deposits,
+    )
+
+    # Get genesis block
+    genesis_block = spec.get_genesis_block_from_genesis_state(genesis_state)
+
+    # Check that the genesis block points to the genesis state
+    assert genesis_block.state_root == genesis_state.hash_tree_root()


tests/core/pyspec/eth2spec/test/merge/genesis/test_initialization.py
```diff
@@ -120,3 +120,55 @@ def test_initialize_post_transition(spec):
     assert spec.is_merge_complete(state)
 
     yield 'state', state
+
+
+@with_merge_and_later
```
djrtwo, line 125: we'll also want to add a unittest to `phase0/genesis`

asn-d6: Added a test in 70e179f7

specs/merge/beacon-chain.md
```diff
@@ -214,6 +216,34 @@ def is_execution_enabled(state: BeaconState, body: BeaconBlockBody) -> bool:
...
+        logs_bloom=genesis_execution_header.logs_bloom,
+        random=genesis_execution_header.random,
+        block_number=genesis_execution_header.block_number,
+        gas_limit=genesis_execution_header.gas_limit,
+        gas_used=genesis_execution_header.gas_used,
+        timestamp=genesis_execution_header.timestamp,
+        extra_data=genesis_execution_header.extra_data,
+        base_fee_per_gas=genesis_execution_header.base_fee_per_gas,
+        block_hash=genesis_execution_header.block_hash,
+        transactions=[]
```
ralexstokes, line 240: nitpick; consider: 
```suggestion
        transactions=[],
```

specs/merge/beacon-chain.md
```diff
@@ -214,6 +216,34 @@ def is_execution_enabled(state: BeaconState, body: BeaconBlockBody) -> bool:
...
+    )
+
+    return BeaconBlock(state_root=hash_tree_root(genesis_state),
+                       body=BeaconBlockBody(execution_payload=genesis_execution_payload))
```
djrtwo, line 244: tests sometimes use the above format but it's avoided in the core specs. Use the following format
```suggestion
    return BeaconBlock(
        state_root=hash_tree_root(genesis_state),
        body=BeaconBlockBody(execution_payload=genesis_execution_payload),
    )
```

tests/core/pyspec/eth2spec/test/merge/genesis/test_initialization.py
```diff
@@ -120,3 +120,55 @@ def test_initialize_post_transition(spec):
...
+
+    # Check that the payload we generated is the one that is embedded in the state
+    assert genesis_execution_header.hash_tree_root() == \
+        genesis_state.latest_execution_payload_header.hash_tree_root()
+
+    # Check that the genesis block points to the genesis state
+    assert genesis_block.state_root == genesis_state.hash_tree_root()
+
+    # Do a static assert of the genesis block given the parameters above
+    assert "bde449ed275fced558631288addaac3f1dcd6eeb00ddb9fb0c293ea7dcd3b3eb" == \
```
djrtwo, line 173: We generally always avoid `\` to signify a new line for same logical statement across the repo. Suggest  something like

```python
    expected_genesis_block_root = "bde449ed275fced558631288addaac3f1dcd6eeb00ddb9fb0c293ea7dcd3b3eb"
    assert genesis_block.hash_tree_root().hex() == expected_genesis_block_root

Comment (djrtwo) [reply tests/core/pyspec/eth2spec/test/merge/genesis/test_initialization.py:173]: I'm also not sure we should hard-code this in here. All this does is ensure that the inputs and `beacon_block` structure don't change since when this test was written. This will almost certainly prevent this test from being useful for subsequent forks because `BeaconBlock` is likely be extended with new fields.

Seems brittle and doesn't buy us much. The relationships between the various components as defined in the above asserts are sufficient, imo

File: tests/core/pyspec/eth2spec/test/phase0/genesis/test_initialization.py [hunk 5]
```diff
@@ -216,3 +220,40 @@ def test_initialize_beacon_state_random_valid_genesis(spec):
...
+    )
+
+    # Get genesis block
+    genesis_block = spec.get_genesis_block_from_genesis_state(genesis_state)
+
+    # Check that the genesis block points to the genesis state
+    assert genesis_block.state_root == genesis_state.hash_tree_root()
+
+    # Do a static assert of the genesis block given the parameters above
+    assert "84934a1f3fe297f147cabcf83afcd1071682038b1a25d32deaee09b3bf3f0a63" == \
```
djrtwo, line 258: ditto

specs/phase0/beacon-chain.md
```diff
@@ -1229,7 +1230,12 @@ def is_valid_genesis_state(state: BeaconState) -> bool:
 
 ### Genesis block
 
-Let `genesis_block = BeaconBlock(state_root=hash_tree_root(genesis_state))`.
```
mkalinin, line 1232: Probably we should keep this line and make the following modification:
- Let `genesis_block = get_genesis_block_from_genesis_state(genesis_state)`, where:
```python
 def get_genesis_block_from_genesis_state(genesis_state: BeaconState) -> BeaconBlock:
     return BeaconBlock(state_root=hash_tree_root(genesis_state))
 ```

Comment (djrtwo) [reply specs/phase0/beacon-chain.md:1232]: agreed

File: tests/core/pyspec/eth2spec/test/merge/genesis/test_initialization.py (hunk 4 above)
Comment (ralexstokes) [reply tests/core/pyspec/eth2spec/test/merge/genesis/test_initialization.py:173]: we could scope this check to each fork -- the idea is that once we publish the spec that yields this root hash, it *should not* change, and this `assert` is an additional check that nothing does

File: specs/phase0/beacon-chain.md (hunk 6 above)
Comment (ralexstokes) [reply specs/phase0/beacon-chain.md:1232]: also agreed

File: specs/merge/beacon-chain.md (hunk 3 above)
Comment (asn-d6) [reply specs/merge/beacon-chain.md]: resolved in 84ed3a43

File: specs/merge/beacon-chain.md [hunk 7]
```diff
@@ -214,6 +216,36 @@ def is_execution_enabled(state: BeaconState, body: BeaconBlockBody) -> bool:
     return is_merge_block(state, body) or is_merge_complete(state)
 ```
 
+### Genesis block
+
+#### `get_genesis_block_from_genesis_state`
```
hwwhww, line 221: I think for Merge specs, it would be better to move this function to under the "Testing" section as it's testing (testnet) only.

tests/core/pyspec/eth2spec/test/merge/genesis/test_initialization.py
```diff
@@ -120,3 +120,51 @@ def test_initialize_post_transition(spec):
     assert spec.is_merge_complete(state)
 
     yield 'state', state
+
+
+@with_merge_and_later
+@spec_test
+@single_phase
+@with_presets([MINIMAL], reason="too slow")
+def test_genesis_block_execution_payload(spec):
```
hwwhww, line 129: This test case should be moved to `tests/core/pyspec/eth2spec/test/merge/unittests/test_genesis.py` since it doesn't create test vectors.

tests/core/pyspec/eth2spec/test/phase0/genesis/test_initialization.py
```diff
@@ -216,3 +220,36 @@ def test_initialize_beacon_state_random_valid_genesis(spec):
     assert spec.is_valid_genesis_state(state)
 
     yield 'state', state
+
+
+@with_phases([PHASE0])
+@spec_test
+@single_phase
+@with_presets([MINIMAL], reason="too slow")
+def test_genesis_block_creation(spec):
```
hwwhww, line 229: ditto.

This test case should be moved to `tests/core/pyspec/eth2spec/test/phase0/unittests/test_genesis.py` since it doesn't create test vectors.

specs/merge/beacon-chain.md
```diff
@@ -348,8 +380,8 @@ Modifications include:
 1. Use `MERGE_FORK_VERSION` as the current fork version.
 2. Utilize the Merge `BeaconBlockBody` when constructing the initial `latest_block_header`.
 3. Initialize `latest_execution_payload_header`.
-  If `execution_payload_header == ExecutionPayloadHeader()`, then the Merge has not yet occurred.
-  Else, the Merge starts from genesis and the transition is incomplete.
+  If `execution_payload_header == ExecutionPayloadHeader()`, then the transition has not yet occurred.
+  Else, the Merge starts from genesis and the transition is complete.
```
hwwhww, line 384: Thanks for the fix 🙏

hwwhww (commented):
Some formatting & file path suggestions.

hwwhww (commented):
I was checking if it is compatible with fork choice specs & testing. Saw some errors when I was updating testing helpers.

Previously, we only have to handle `state_root` for the genesis block in `process_slot`, but now we also have to fill the `state.latest_block_header.body_root` for merge genesis.

EDIT: added suggestion (https://github.com/ethereum/consensus-specs/commit/71146f903b11022e1ca1d03fc9b14254cb9a3dee) 
```python
def process_slot(state: BeaconState) -> None:
    # Cache state root
    previous_state_root = hash_tree_root(state)
    state.state_roots[state.slot % SLOTS_PER_HISTORICAL_ROOT] = previous_state_root
    # Cache latest block header state root
    if state.latest_block_header.state_root == Bytes32():
        state.latest_block_header.state_root = previous_state_root
        # >>>>>> [NEW] <<<<<<
        if state.slot == GENESIS_SLOT:
            # Note: it is only required for testing, where the genesis block has filled `execution_payload`.
            genesis_block = get_genesis_block_from_genesis_state(state)
            state.latest_block_header.body_root = hash_tree_root(genesis_block.body)
    # Cache block root
    previous_block_root = hash_tree_root(state.latest_block_header)
    state.block_roots[state.slot % SLOTS_PER_HISTORICAL_ROOT] = previous_block_root
```


Also for how we generate parent root in the block helper functions.

Does it make sense to you? @ralexstokes @djrtwo 

Still looking into these issues now. 👀

specs/merge/beacon-chain.md
hwwhww: I think it's okay to keep it here as is. (Similar to `initialize_beacon_state_from_eth1`)

hwwhww:
@asn-d6 @ralexstokes @djrtwo 

Re: `process_slot` issue

Here is my patch commit that you can cherry-pick (CI passed): https://github.com/ethereum/consensus-specs/commit/71146f903b11022e1ca1d03fc9b14254cb9a3dee

djrtwo:
I personally am leaning in favor of dropping this path.

I am hesitant to change `process_slot` just to clean up the genesis block mapping for Merge-only testnets (and maybe alternative networks). Cleaning up this not very important edge-case will just make other edge cases and require testing in a very core function -- `process_slot`.

I'll make sure there isn't a clean way to do it, but the circular dependency we've created does imply that this is the only reasonable path to do so

hwwhww:
What was the genesis block content that the client teams use during the interop? Just simply used `BeaconBlock(state_root=hash_tree_root(genesis_state))` and it worked okay?

djrtwo:
> Just simply used BeaconBlock(state_root=hash_tree_root(genesis_state)) and it worked okay?

Yes. You miss the 1:1 mapping between CL and EL blocks at the genesis block which might create an edge case in DB mapping, but EL ships with genesis block so there is no missing information in an EL+CL client.

Given that it worked fine at interop and that it is an exceptional case wrt network configs, I'd personally argue to close this PR and keep as is

PR #2659, opened 2021-10-09
Made the timestamp function underflow safe

vbuterin:


specs/merge/beacon-chain.md
```diff
@@ -218,12 +218,9 @@ def is_execution_enabled(state: BeaconState, body: BeaconBlockBody) -> bool:
 
 #### `compute_timestamp_at_slot`
 
-*Note*: This function is unsafe with respect to overflows and underflows.
-
 ```python
 def compute_timestamp_at_slot(state: BeaconState, slot: Slot) -> uint64:
-    slots_since_genesis = slot - GENESIS_SLOT
-    return uint64(state.genesis_time + slots_since_genesis * SECONDS_PER_SLOT)
+    return uint64(state.genesis_time + slot * SECONDS_PER_SLOT - GENESIS_SLOT * SECONDS_PER_SLOT)
```
protolambda, line 223: `GENESIS_SLOT` is a constant (not a preset var, not a config var, but a real constant) from phase0, although the presentation may look better, there is no real underflow, just verbosity. Might as well remove it entirely.

specs/merge/beacon-chain.md
```diff
@@ -218,12 +218,9 @@ def is_execution_enabled(state: BeaconState, body: BeaconBlockBody) -> bool:
 
 #### `compute_timestamp_at_slot`
 
-*Note*: This function is unsafe with respect to overflows and underflows.
```
protolambda, line 221: Technically the overflow is not removed, but we do have a spec notice about everything being essentially safe-math, raising an error on overflow. Removing the comment is fine.

PR #2660, opened 2021-10-10
remove notify_consensus_validated

djrtwo:
`POS_CONSENSUS_VALIDATED` was removed from EIP 3675. See relevant issue for discussion: https://github.com/ethereum/EIPs/pull/4357

Remove `notify_consensus_validated` from CL specs in accordance with the change in EIP 3675


hwwhww (approved):
lgtm

PR #2663, opened 2021-10-11
Increase gossipsub seen-ttl

AgeManning:
The gossipsub IGNORE rule for attestations is:
```
attestation.data.slot + ATTESTATION_PROPAGATION_SLOT_RANGE >= current_slot >= attestation.data.slot
```
As `ATTESTATION_PROPAGATION_SLOT_RANGE` is `32`, this allows for clients to propagate attestations up until the start of the 33rd slot after publishing. 

Because we've set our duplicate cache to the start of the 32nd slot, I've noticed quite a few duplicates bouncing around the network. Increasing the cache time will prevent these extra duplicates being unnecessarily propagated.


arnetheduck:
in nimbus, we stop propagating a bit earlier than 32 actually, to avoid the duplicate and to avoid being descored by lighthouse in particular - basically, there's a race where the attestation is valid when sending, then becomes invalid in-flight as time passes.

leolara:

PR #2664, opened 2021-10-12
Add new Altair transition tests

### LINKED ISSUE #2662: Altair test-vectors at fork transition
There are a few edge case situations that may happen at the Altair fork and are currently under-tested:

- A validator is slashed in phase0 and is active but slashing at the fork and thus leaking. It may or not be a member of a sync committee, we need to check the right penalties. 
- A validator started an exit in phase0 and is put in a queue, check these queues, either active or leaking at the fork. 
- Quadratic leaking in case of network not finalizing at the fork. 
- An activator deposited in phase0 and is in the queue at the fork. 

Many of these tests are done already at Altair, but implementations may differ in the way they compute rewards/penalties if a slashing event for example happened before the fork. Thus this could eventually lead to a consensus failure with the cost of only slashing a single validator. 

This is follow up from discussions in 
https://discord.com/channels/595666850260713488/598292067260825641/897248922706255893

hwwhww:
Fix #2662
Thank @potuz for the ideas. ✨ 

- A validator is slashed in phase0 and is active but slashing at the fork and thus leaking. It may or not be a member of a sync committee, we need to check the right penalties.
    - `test_transition_with_one_fourth_slashed_active_validators_pre_fork`
- A validator started an exit in phase0 and is put in a queue, check these queues, either active or leaking at the fork.
    - `test_transition_with_one_fourth_exiting_validators_exit_post_fork`
    - `test_transition_with_one_fourth_exiting_validators_exit_at_fork`
- Quadratic leaking in case of network not finalizing at the fork. 
   - `test_transition_with_leaking_pre_fork`
   - `test_transition_with_leaking_at_fork`
   - ~~`test_transition_with_leaking_post_fork`~~
- An activator deposited in phase0 and is in the queue at the fork.
   - `test_transition_with_non_empty_activation_queue`
  
---
Operations at the fork block:
- `test_transition_with_proposer_slashing_right_after_fork`
- `test_transition_with_proposer_slashing_right_before_fork`
- `test_transition_with_attester_slashing_right_after_fork`
- `test_transition_with_attester_slashing_right_before_fork`
- `test_transition_with_deposit_right_after_fork`
- `test_transition_with_deposit_right_before_fork`
- `test_transition_with_voluntary_exit_right_after_fork`
- `test_transition_with_voluntary_exit_right_before_fork`

TODO
- [x] clean up, refactoring
- [x] add tests to include slashings at the fork block
- [x] add tests to include voluntary exit at the fork block

### DIFF
--- a/tests/core/pyspec/eth2spec/test/helpers/random.py
+++ b/tests/core/pyspec/eth2spec/test/helpers/random.py
@@ -53,7 +53,7 @@ def exit_random_validators(spec, state, rng, fraction=0.5, exit_epoch=None, with
     """
     if from_epoch is None:
         from_epoch = spec.MAX_SEED_LOOKAHEAD + 1
-    epoch_diff = from_epoch - spec.get_current_epoch(state)
+    epoch_diff = int(from_epoch) - int(spec.get_current_epoch(state))
     for _ in range(epoch_diff):
         # NOTE: if `epoch_diff` is negative, then this loop body does not execute.
         next_epoch(spec, state)


potuz:
Thanks a lot! perhaps it may be worth adding the case that the slashed validator is assigned to a sync committee? Since in this case the penalties are active and the implementations may differ in how to add a slashing validator from phase 0?

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
```diff
@@ -0,0 +1,228 @@
...
+
+    yield "blocks", blocks
+    yield "post", state
+
+
+@fork_transition_test(PHASE0, ALTAIR, fork_epoch=2)
+def test_transition_with_one_fourth_exiting_validators_exit_at_fork(
+        state, fork_epoch, spec, post_spec, pre_tag, post_tag):
+    """
+    1/4 exiting but still active validators at the fork transition.
```
djrtwo, line 73: This is the same comment as the above function. Should write a bit more to distinguish the two tests

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
```diff
@@ -0,0 +1,228 @@
...
+    yield "post", state
+
+
+@fork_transition_test(PHASE0, ALTAIR, fork_epoch=260)
+def test_transition_with_voluntary_exit_at_fork(state, fork_epoch, spec, post_spec, pre_tag, post_tag):
+    """
+    Create an attester slashing at the transition.
+    fork_epoch=260 because mainnet `SHARD_COMMITTEE_PERIOD` is 256 epochs.
+    """
+    transition_to(spec, state, spec.config.SHARD_COMMITTEE_PERIOD * spec.SLOTS_PER_EPOCH)
```
djrtwo, line 120: This will be very expensive.

Consider just setting the current_slot manually.
Otherwise, should probably only run this with minimal. Even then, might take 1+ minute

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
```diff
@@ -0,0 +1,228 @@
...
+    transition_until_fork(spec, state, fork_epoch)
+
+    yield "pre", state
+
+    validator_index = 0
+    signed_exits = prepare_signed_exits(spec, state, [validator_index])
+    operation_dict = {'voluntary_exits': signed_exits}
+
+    # irregular state transition to handle fork:
+    state, block = do_altair_fork(state, spec, post_spec, fork_epoch, operation_dict=operation_dict)
```
djrtwo, line 130: So this exit happens at the 0th block immediately *after* the fork transition?

djrtwo: Would it make sense to also do one at the N-1 block to the fork?

tests/core/pyspec/eth2spec/test/helpers/inactivity_scores.py
```diff
@@ -1,9 +1,32 @@
...
 
 def zero_inactivity_scores(spec, state, rng=None):
     state.inactivity_scores = [0] * len(state.validators)
+
+
+def slash_some_validators_for_inactivity_scores_test(spec, state, rng=Random(40404040), fraction=0.25):
+    """
+    ``run_inactivity_scores_test`` runs at the next epoch from `state`.
+    # We retrieve the proposer of this future state to avoid
+    # accidentally slashing that validator
```
djrtwo, line 20:
```suggestion
    We retrieve the proposer of this future state to avoid
    accidentally slashing that validator
```

djrtwo (commented):
Looking good.

The tests are a bit verbose. Some might be able to be modularized in helper functions.. Not critical but consider if you see some opportunity.

We also might consider some of these tests to be done in both a leaking and non-leaking context. Maybe we do a mixed-use test (combination of all types here -- slashed, exited, queued, etc) and do a non-leaking and leaking form.

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
ralexstokes: +1

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
```diff
@@ -0,0 +1,228 @@
...
+from eth2spec.test.helpers.voluntary_exits import prepare_signed_exits
+
+
+#
+# Exit
+#
+
+@fork_transition_test(PHASE0, ALTAIR, fork_epoch=2)
+def test_transition_with_one_fourth_exiting_validators_exit_post_fork(
+        state, fork_epoch, spec, post_spec, pre_tag, post_tag):
```
ralexstokes, line 26: consider:

```suggestion
def test_transition_with_one_fourth_exiting_validators_exit_post_fork(state, 
                                                                                                                         fork_epoch, 
                                                                                                                         spec, 
                                                                                                                         post_spec, 
                                                                                                                         pre_tag, 
                                                                                                                         post_tag):
```

tests/core/pyspec/eth2spec/test/altair/transition/test_slashing.py
```diff
@@ -0,0 +1,133 @@
...
+    transition_to_next_epoch_and_append_blocks,
+)
+from eth2spec.test.helpers.inactivity_scores import (
+    slash_some_validators_for_inactivity_scores_test,
+)
+
+
+@fork_transition_test(PHASE0, ALTAIR, fork_epoch=1)
+def test_transition_with_one_fourth_slashed_active_validators_pre_fork(
+        state, fork_epoch, spec, post_spec, pre_tag, post_tag):
```
ralexstokes, line 26: ditto on function param formatting

tests/core/pyspec/eth2spec/test/helpers/fork_transition.py
```diff
@@ -0,0 +1,142 @@
...
+    while state.slot < to_slot or not found_valid:
+        future_state = state.copy()
+        next_slot(spec, future_state)
+        proposer_index = spec.get_beacon_proposer_index(future_state)
+        if proposer_index not in ignoring_proposers:
+            block = build_empty_block_for_next_slot(spec, state)
+            signed_block = state_transition_and_sign_block(spec, state, block)
+            yield signed_block
+            if state.slot >= to_slot:
+                found_valid = True
```
ralexstokes, line 95: seems like you can drop the variable `found_valid` if you just `break` once the condition is met

```suggestion
                break
```

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
```diff
@@ -0,0 +1,228 @@
+import random
```
ralexstokes, line 1: change the name of this file to something like `test_activations_and_exits.py`?

it wasn't clear until i dug into the code that this file also was testing the exit logic

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+@fork_transition_test(PHASE0, ALTAIR, fork_epoch=2)
+def test_transition_with_non_empty_activation_queue(state, fork_epoch, spec, post_spec, pre_tag, post_tag):
+    """
+    Create some deposits before the transition
+    """
+    transition_until_fork(spec, state, fork_epoch)
+
+    _, queuing_indices = set_some_new_deposits(spec, state, rng=random.Random(5566))
```
ralexstokes, line 156: this function is modified to return the modified indices but throws away the first list that is returned?

i can't find another usage of this modification and in that case can we only return the `queuing_indices`?

tests/core/pyspec/eth2spec/test/altair/transition/test_slashing.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    1/4 validators are slashed but still active at the fork transition.
+    """
+    # slash 1/4 validators
+    slashed_indices = slash_some_validators_for_inactivity_scores_test(
```
ralexstokes, line 31: see note about the refactoring of this function -- it may make more sense to instead use/modify `helpers/random.py::slash_some_validators` for this purpose

ralexstokes: if you want to keep this function like you have it, then we should change the name to something more generic than "for_inactivity_scores_test"

tests/core/pyspec/eth2spec/test/altair/transition/test_slashing.py
```diff
@@ -0,0 +1,133 @@
...
+    assert spec.get_current_epoch(state) < fork_epoch
+
+    yield "pre", state
+
+    # irregular state transition to handle fork:
+    blocks = []
+    state, block = do_altair_fork(state, spec, post_spec, fork_epoch)
+    blocks.append(post_tag(block))
+
+    # ensure that some of the current sync committee members are the slashed
```
ralexstokes, line 53:
```suggestion
    # ensure that some of the current sync committee members are slashed
```

tests/core/pyspec/eth2spec/test/helpers/fork_transition.py
```diff
@@ -0,0 +1,142 @@
...
+    Cribbed from ``transition_unsigned_block`` helper
+    where the early parts of the state transition have already
+    been applied to ``state``.
+
+    Used to produce a block during an irregular state transition.
+    """
+    block = build_empty_block(spec, state)
+    # we can't just pass `body` and assign it because randao_reveal and eth1_data was set in `build_empty_block`
+    # thus use dict to pass operations.
+    if operation_dict is not None:
```
ralexstokes, line 28: consider:

```suggestion
    if operation_dict:
```

tests/core/pyspec/eth2spec/test/helpers/fork_transition.py
```diff
@@ -0,0 +1,142 @@
...
+from eth2spec.test.helpers.block import (
+    build_empty_block_for_next_slot,
+    build_empty_block,
+    sign_block,
+)
+
+
+def _state_transition_and_sign_block_at_slot(spec,
+                                             state,
+                                             operation_dict=None):
```
ralexstokes, line 17: note: this function has been modified to accept an `operation_dict` with a batch of operations to ensure are included in the produced block

tests/core/pyspec/eth2spec/test/altair/transition/test_slashing.py
```diff
@@ -0,0 +1,133 @@
...
+
+    indices = set(attester_slashing.attestation_1.attesting_indices).intersection(
+        attester_slashing.attestation_2.attesting_indices
+    )
+    assert len(indices) > 0
+    for validator_index in indices:
+        assert state.validators[validator_index].slashed
+
+    # continue regular state transition with new spec into next epoch
+    transition_to_next_epoch_and_append_blocks(post_spec, state, post_tag, blocks)
```
ralexstokes, line 101: this test will be lighter if we only go a few blocks/slots into this epoch, rather than the full `SLOTS_PER_EPOCH` and it seems like additional blocks don't add much value considering what this function aims to test

tests/core/pyspec/eth2spec/test/altair/transition/test_slashing.py
```diff
@@ -0,0 +1,133 @@
...
+    # irregular state transition to handle fork:
+    state, block = do_altair_fork(state, spec, post_spec, fork_epoch, operation_dict=operation_dict)
+    blocks = []
+    blocks.append(post_tag(block))
+
+    slashed_proposer = state.validators[proposer_slashing.signed_header_1.message.proposer_index]
+    assert slashed_proposer.slashed
+
+    # continue regular state transition with new spec into next epoch
+    transition_to_next_epoch_and_append_blocks(post_spec, state, post_tag, blocks)
```
ralexstokes, line 130: ditto on number of blocks produced

tests/core/pyspec/eth2spec/test/altair/transition/test_leaking.py
```diff
@@ -0,0 +1,94 @@
...
+    # irregular state transition to handle fork:
+    blocks = []
+    state, block = do_altair_fork(state, spec, post_spec, fork_epoch)
+    blocks.append(post_tag(block))
+
+    # check post transition state
+    assert spec.is_in_inactivity_leak(state)
+
+    # continue regular state transition with new spec into next epoch
+    transition_to_next_epoch_and_append_blocks(post_spec, state, post_tag, blocks)
```
ralexstokes, line 32: how many blocks do we need to generate here?

tests/core/pyspec/eth2spec/test/altair/transition/test_leaking.py
```diff
@@ -0,0 +1,94 @@
...
```
ralexstokes, line 60: ditto on blocks produced

tests/core/pyspec/eth2spec/test/altair/transition/test_leaking.py
```diff
@@ -0,0 +1,94 @@
...
+
+    yield "blocks", blocks
+    yield "post", state
+
+
+@fork_transition_test(PHASE0, ALTAIR, fork_epoch=5)
+def test_transition_with_leaking_post_fork(state, fork_epoch, spec, post_spec, pre_tag, post_tag):
```
ralexstokes, line 67: echo the similar question i saw danny raise elsewhere -- this variant of the test *should* be covered by leaking test we have elsewhere, so I'd suggest dropping it to reduce the overall test load

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
```diff
@@ -0,0 +1,228 @@
...
+    yield "pre", state
+
+    # irregular state transition to handle fork:
+    blocks = []
+    state, block = do_altair_fork(state, spec, post_spec, fork_epoch)
+    blocks.append(post_tag(block))
+
+    # ensure that some of the current sync committee members are exiting
+    exited_pubkeys = [state.validators[index].pubkey for index in exited_indices]
+    assert any(set(exited_pubkeys).intersection(list(state.current_sync_committee.pubkeys)))
```
ralexstokes, line 53: if we *really* wanted to verify this, we would also check that there are some sync committee members who do _not_ have the property under test...

and this would apply to slashings as well

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
```diff
@@ -0,0 +1,228 @@
...
+    yield "post", state
+
+
+@fork_transition_test(PHASE0, ALTAIR, fork_epoch=2)
+def test_transition_with_one_fourth_exiting_validators_exit_at_fork(
+        state, fork_epoch, spec, post_spec, pre_tag, post_tag):
+    """
+    1/4 exiting but still active validators at the fork transition.
+    """
+    exited_indices = set_validators_exit_epoch(spec, state, exit_epoch=2, rng=random.Random(5566), fraction=0.25)
```
ralexstokes, line 75:
```suggestion
    exited_indices = set_validators_exit_epoch(spec, state, exit_epoch=fork_epoch, rng=random.Random(5566), fraction=0.25)
```

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
```diff
@@ -0,0 +1,228 @@
...
+    assert not any(set(exited_pubkeys).intersection(list(state.current_sync_committee.pubkeys)))
+
+    # continue regular state transition with new spec into next epoch
+    transition_to_next_epoch_and_append_blocks(post_spec, state, post_tag, blocks)
+
+    yield "blocks", blocks
+    yield "post", state
+
+
+@fork_transition_test(PHASE0, ALTAIR, fork_epoch=260)
```
ralexstokes, line 114: this is outside the scope of this PR but it would be nice if `fork_epoch` could take either a number for the actual epoch *or* a `Callable` that accepts the pre and post spec and then lets the user compute an epoch...

the underlying issue here is that `fork_epoch` is hard-coded to exceed the `SHARD_COMMITTEE_PERIOD`

just making a note for now

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
ralexstokes: right, it seems that this exit is valid starting in the prior epoch (`fork_epoch` - 1) and included in the "fork block", i.e. the _first_ block after the fork transition occurs.

ralexstokes: we could include an exit at the N-1 block but that kind of raises the question of the scope of this PR -- we could include operations at each N-1 block and N block (if the "fork block" is N) but then we have to ask if doing this by hand is the best way to test it...

there is also the question of operations in the N+1 block, but we could argue those cases are already covered by existing tests that just start in an Altair fork...

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
```diff
@@ -0,0 +1,228 @@
...
+    transition_to_next_epoch_and_append_blocks(post_spec, state, post_tag, blocks)
+
+    assert state.validators[validator_index].activation_epoch < post_spec.FAR_FUTURE_EPOCH
+
+    to_slot = state.validators[validator_index].activation_epoch * post_spec.SLOTS_PER_EPOCH
+    blocks.extend([
+        post_tag(block) for block in
+        state_transition_across_slots(post_spec, state, to_slot)
+    ])
+    assert post_spec.is_active_validator(state.validators[validator_index], post_spec.get_current_epoch(state))
```
ralexstokes, line 225: can consider skipping all of these blocks by just transitioning to `to_slot` and then perhaps making 1 block to "cap off the chain" but all of the epoch processing functions required will be run w/o making blocks

tests/core/pyspec/eth2spec/test/helpers/random.py
```diff
@@ -7,6 +7,7 @@
 
 
 def set_some_new_deposits(spec, state, rng):
+    eligible_indices = queuing_indices = []
```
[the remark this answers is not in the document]
hwwhww: my biggest nightmare side effect came true. thanks a lot for pointing out it!

tests/core/pyspec/eth2spec/test/helpers/fork_transition.py
hwwhww: I want in the last, to have one block of `slot >= to_slot`.

If I remove `or not found_valid` from `while state.slot < to_slot or not found_valid`, the while loop might return without the last block.

tests/core/pyspec/eth2spec/test/altair/transition/test_slashing.py
hwwhww: Good call 👍
I changed it to using `only_at(to_slot)` so that it only produces the last block. What do you think?

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
```diff
@@ -0,0 +1,228 @@
...
+    operation_dict = {'deposits': [deposit]}
+    # irregular state transition to handle fork:
+    state, block = do_altair_fork(state, spec, post_spec, fork_epoch, operation_dict=operation_dict)
+    blocks = []
+    blocks.append(post_tag(block))
+
+    assert not post_spec.is_active_validator(state.validators[validator_index], post_spec.get_current_epoch(state))
+
+    # continue regular state transition with new spec into next epoch
+    transition_to_next_epoch_and_append_blocks(post_spec, state, post_tag, blocks)
```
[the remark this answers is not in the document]
hwwhww: But since `activation_eligibility_epoch` is 3, if I remove `transition_to_next_epoch_and_append_blocks `, I have to have to do one more epoch (slot_count=spec.SLOTS_PER_EPOCH * 3)` in `next_slots_with_attestations` to make ` state.finalized_checkpoint.epoch >= state.validators[validator_index].activation_eligibility_epoch`.

The status quo is a bit more efficient since `next_slots_with_attestations` is heavier.

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
hwwhww: I added `block_filter=only_at(to_slot)` to `state_transition_across_slots` arguments. Is that what you meant?

tests/core/pyspec/eth2spec/test/altair/epoch_processing/test_process_inactivity_updates.py
```diff
@@ -1,7 +1,10 @@
 from random import Random
 
 from eth2spec.test.context import spec_state_test, with_altair_and_later
-from eth2spec.test.helpers.inactivity_scores import randomize_inactivity_scores, zero_inactivity_scores
+from eth2spec.test.helpers.inactivity_scores import (
+    randomize_inactivity_scores,
+    zero_inactivity_scores,
+)
```
ralexstokes, line 7: just formatting

tests/core/pyspec/eth2spec/test/helpers/random.py
```diff
@@ -18,43 +19,67 @@ def set_some_new_deposits(spec, state, rng):
...
+def exit_random_validators(spec, state, rng, fraction=None, exit_epoch=None, withdrawable_epoch=None, forward=True):
+    """
+    Set some validators' exit_epoch and withdrawable_epoch.
 
-def exit_random_validators(spec, state, rng, fraction=None):
+    If exit_epoch is configured, use the given exit_epoch. Otherwise, randomly set exit_epoch and withdrawable_epoch.
+    """
     if fraction is None:
         # Exit ~1/2
         fraction = 0.5
```
ralexstokes, line 35: not in this PR but can we just say

```python
def exit_random_validators(..., fraction=0.5, ...): ...
```

tests/core/pyspec/eth2spec/test/helpers/random.py
```diff
@@ -7,6 +7,7 @@
 
 
 def set_some_new_deposits(spec, state, rng):
+    queuing_indices = []
```
ralexstokes, line 10: the use of the word `queuing` here is a bit confusing imo bc these indices are *not* in the activation queue, they have just deposited... perhaps call this list `deposited_indices` instead?

i would consider the indices in line 21 to be the "queuing indices"

tests/core/pyspec/eth2spec/test/altair/transition/test_activations_and_exits.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # irregular state transition to handle fork:
+    blocks = []
+    state, block = do_altair_fork(state, spec, post_spec, fork_epoch)
+    blocks.append(post_tag(block))
+
+    # ensure that some of the current sync committee members are exiting
+    exited_pubkeys = [state.validators[index].pubkey for index in exited_indices]
+    assert any(set(exited_pubkeys).intersection(list(state.current_sync_committee.pubkeys)))
+    assert any(set(exited_pubkeys).difference(list(state.current_sync_committee.pubkeys)))
```
ralexstokes, line 60: nice :)

tests/core/pyspec/eth2spec/test/altair/transition/test_activations_and_exits.py
```diff
@@ -0,0 +1,157 @@
...
+                                                                    spec,
+                                                                    post_spec,
+                                                                    pre_tag,
+                                                                    post_tag):
+    """
+    1/4 validators initiated voluntary exit before the fork,
+    and being exited and inactive *right after* the fork transition.
+    """
+    exited_indices = exit_random_validators(
+        spec, state, rng=random.Random(5566), fraction=0.25, exit_epoch=fork_epoch, forward=False)
```
ralexstokes, line 88: formatting? (see above)

tests/core/pyspec/eth2spec/test/altair/transition/test_slashing.py
```diff
@@ -0,0 +1,69 @@
...
+    assert not spec.is_in_inactivity_leak(state)
+
+    transition_until_fork(spec, state, fork_epoch)
+
+    assert spec.get_current_epoch(state) < fork_epoch
+
+    yield "pre", state
+
+    # irregular state transition to handle fork:
+    state, _ = do_altair_fork(state, spec, post_spec, fork_epoch, with_block=False)
```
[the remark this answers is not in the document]
hwwhww: because `slash_random_validators` may select the next block proposer. I didn't want to change `slash_random_validators` too much. Do you feel strongly that we should add an `avoid_next_proposer` flag to `slash_random_validators` helper?

tests/core/pyspec/eth2spec/test/helpers/fork_transition.py
```diff
@@ -0,0 +1,308 @@
...
+        # avoid slashing the next proposer
+        future_state = state.copy()
+        next_slot(spec, future_state)
+        proposer_index = spec.get_beacon_proposer_index(future_state)
+        selected_validator_index = (proposer_index + 1) % len(state.validators)
+        proposer_slashing = get_valid_proposer_slashing(
+            spec, state, slashed_index=selected_validator_index, signed_1=True, signed_2=True)
+        operation_dict = {'proposer_slashings': [proposer_slashing]}
+    elif operation_type == OperetionType.ATTESTER_SLASHING:
+        attester_slashing = get_valid_attester_slashing(spec, state, signed_1=True, signed_2=True)
```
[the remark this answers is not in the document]
hwwhww: Yes...!
Good catch! I updated `state_transition_across_slots_with_ignoring_proposers` helper to avoid such cases.

tests/core/pyspec/eth2spec/test/altair/transition/test_activation.py
```diff
@@ -0,0 +1,228 @@
...
+    blocks = []
+    state, block = do_altair_fork(state, spec, post_spec, fork_epoch)
+    blocks.append(post_tag(block))
+
+    # continue regular state transition with new spec into next epoch
+    transition_to_next_epoch_and_append_blocks(post_spec, state, post_tag, blocks)
+
+    yield "blocks", blocks
+    yield "post", state
+
```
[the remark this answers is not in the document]
hwwhww: added `test_transition_with_activation_at_fork_epoch`!

Issue #2636, opened 2021-09-29
Clarification on `get_pow_block`


paulhauner:
The spec defines a [`get_pow_block`](https://github.com/ethereum/consensus-specs/blob/v1.1.0/specs/merge/fork-choice.md#get_pow_block) function which returns a [`PoWBlock`](https://github.com/ethereum/consensus-specs/blob/v1.1.0/specs/merge/fork-choice.md#powblock) structure. In practice, I expect `get_pow_block` to map to the `eth_getBlockByHash` JSON-RPC call.

Assuming we use `eth_getBlockByHash`, does `get_pow_block` filter the result so that it only contains pre-merge, non-PoS blocks?

From the naming, I think it *should* apply the filter.

However, if we look at [`get_block_at_terminal_total_difficulty`](https://github.com/ethereum/consensus-specs/blob/v1.1.0/specs/merge/validator.md#block-proposal), we can see it iterates through the entire `pow_chain`. This communicates that there might be multiple blocks in `pow_chain` where `block.total_difficulty >= TERMINAL_TOTAL_DIFFICULTY`. This conflicts with [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675#specification) which states:

> PoW blocks that are descendants of any terminal PoW block MUST NOT be imported. This implies that a terminal PoW block will be the last PoW block in the canonical chain.

I raise this for the following reasons:

1. In an implementation it is desirable to bound the number of loops that `get_block_at_terminal_total_difficulty` can perform. When doing recursive look-ups to a remote server it's desirable to clearly understand the properties of that recursion.
1. I think the specification should be clearer on `get_pow_block` to avoid differences in implementation.
1. If  `get_pow_block` also returns PoS blocks, then it seems like it needs a different name.

On a side note, if `get_pow_block` *should* filter the results, how does it do that? I assume there's some simple way, but I haven't seen it in the API docs I've been looking at.

djrtwo:
You're right.

If the *MUST NOT* condition is followed on the execution-client, then the head of the pow chain is simply either (a) the terminal block OR (b) not yet the terminal block. And in such a case, just a call to ` eth_getBlockByNumber("latest")` should be sufficient to implement this function.


[...]

PR #2676, opened 2021-10-18
Clarify `get_pow_block` block-not-found case

hwwhww:
### Issue
Discussed in #2636 
It is possible that the requested block hash is not found via `get_pow_block`.

### How did I fix it

- Make `get_pow_block` returns `Optional[PowBlock]`
- If it returns `None`, the operation is invalid.
- Add some notes about the edge case.


specs/merge/fork-choice.md
```diff
@@ -64,7 +64,8 @@ class PowBlock(Container):
 
 ### `get_pow_block`
 
-Let `get_pow_block(block_hash: Hash32) -> PowBlock` be the function that given the hash of the PoW block returns its data.
+Let `get_pow_block(block_hash: Hash32) -> Optional[PowBlock]` be the function that given the hash of the PoW block returns its data.
+It may result in `None` if the requested block is not found if execution engine is still syncing.
```
djrtwo, line 68: The consensus spec assumes a 1:1 relationship between CL and EL sync. I'm not sure we should put notes about "syncing" in here. From a consensus spec standpoint, it is unknown if a block is a valid transition block if `get_pow_block` cannot be resolved.

*In practice*, CL will process past the transition point even if EL is syncing, and only later know if such a chain was valid or invalid once EL is done syncing. I'm worried the `assert`s in `on_block` imply too heavily a failure

djrtwo: I might be wavering on my resolve here. I don't think we should mention "execution engine" but you could definitely get a block with a terminal block reference that you can't find. Thus there should be an assertion here

hwwhww: @djrtwo

is it okay to shorten to:
```suggestion
It may result in `None` if the requested block is not found.
```

djrtwo: I edited the above. What do you think?

specs/merge/fork-choice.md
```diff
@@ -110,8 +117,12 @@ def on_block(store: Store, signed_block: SignedBeaconBlock) -> None:
 
     # [New in Merge]
     if is_merge_block(pre_state, block.body):
+        # Note: unavailable PoW block(s) may later become available.
+        # Nodes should queue such beacon blocks for later processing.
```
mkalinin, line 121: This case is a bit more tricky and we should think what would be a better way of specifying it. For the case when CL and EL are in sync with the network, current specification works. But for the case when they are syncing we might want to add more clarity to the spec.

Queueing these blocks could raise a deadlock in an optimistic sync process. Suppose, the optimistic sync is happening and EL uses state sync to catch up with the head and the network passed the transition point a while ago, so the state for transition block is no longer available due to state trie pruning. By queueing beacon blocks at this point, CL will stuck and won't inform EL about the most recent head that is available in the network.

The right behaviour during the sync stage would be as follows:
- process transition block optimistically without terminal block verification
- once fully synced, get back to the transition block and check if it's been finalized or not
- verify terminal block if the transition block hasn't been finalized
- do re-org if the terminal block is invalid 

We also may add this clarity in a separate PR.

mkalinin (commented):
We might want to add more clarity around transition block in the context of optimistic sync. But this could be done in a separate PR

specs/merge/fork-choice.md
djrtwo: I consider the optimistic sync process as independent from the consensus specs.

"queuing blocks for later processing" in optimistic beacon chain sync means that you will eventually go back to fully validate them later. The architecture of CL/EL split client is independent of core consensus specs and I want to find the right balance here to not taint the specs with implementation details.

That said, I could see how this is confusing. But I do think that more notes about optimistic processing can/should be included in the Engine API as that is where this implemntation specific architecture design lives

djrtwo (approved):
lgtm

PR #2682, opened 2021-10-18
add TBH_ACTIVATION_EPOCH

djrtwo:
If you use `TERMINAL_BLOCK_HASH` (a block in the past) as an override for the merge, then you *must* have an activation epoch *in the future* at which point to activate this overrde. Otherwise, it would activate locally immediately for anyone that set the epoch and all nodes would be out of sync.

* Add `TBH_ACTIVATION_EPOCH`
* Ensure that if `TERMINAL_BLOCK_HASH` is overridden that the forkchoice ignores valid TTD terminal blocks (was a minor bug prior to this PR)

Annoying note, I initially named this `TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH` (I know, a verbose name), but setup.py would then see it as two vars and build the file as `config.config.TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH`. I just changed the name for now to avoid this build error. BUT we generally don't use acronyms in our names so we either need to fix the build script or have a non-colliding var-name. Maybe -- `TERMINAL_OVERRIDE_ACTIVATION_EPOCH`


specs/merge/validator.md
```diff
@@ -127,6 +127,11 @@ def prepare_execution_payload(state: BeaconState,
                               fee_recipient: ExecutionAddress,
                               execution_engine: ExecutionEngine) -> Optional[PayloadId]:
     if not is_merge_complete(state):
+        is_tbh_override_set = TERMINAL_BLOCK_HASH != Hash32()
+        if is_tbh_override_set and get_current_epoch(state.slot) < TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH:
+            # TBH is set but activation epoch is not yet reached, no prepare payload call is needed
```
hwwhww, line 132: should we also change TBH to the full name here?
TBH looks like "to be honest" without context.

Issue #2608, opened 2021-09-17
Transaction Type representation in the Merge


protolambda:
## Problem

The transaction typing forward-compatibility is *broken*:
due to double transaction type identification, and the execution engine only recognizing the first,
introducing additional types to the current `Union` definition is not possible.

## Background

The current Merge spec defines transactions and payloads as:
```python
OpaqueTransaction = ByteList[MAX_BYTES_PER_OPAQUE_TRANSACTION]  # typed transaction envelope: TransactionType || TransactionPayload
Transaction = Union[OpaqueTransaction]  # Union can introduce more transactions, but *only consensus layer would be aware*
transactions: List[Transaction, MAX_TRANSACTIONS_PER_PAYLOAD]
```

So in JSON this looks like:
```
{
transactions: [
  {selector: 0, value: "0x??...."},  // legacy transaction with RLP byte in range 0xc0 <= id <= 0xfe
  {selector: 0, value: "0x01...."},  // eip-2930 (optional access lists) in OpaqueTransaction
  {selector: 0, value: "0x02...."},  // eip-1559 in OpaqueTransaction
  
  {selector: 1, value: {"foo": 123, "bar": 42}},  // if we were to add a FooBar type, i.e. Union[OpaqueTransaction, FooBar] 
]
}
```

The Merge API then just uses:
> - transactions: Array of `DATA` - Array of transaction objects,
  each object is a byte list (`DATA`) representing `TransactionType || TransactionPayload` or `LegacyTransaction` 
  as defined in EIP-2718

Read: Merge API uses a list of `OpaqueTransaction`, not a list of `Union[OpaqueTransaction]`

And the `block_hash` (execution layer hash, same as eth1 today) is constructed by hashing the header,
which includes a transactions-trie, which is built from EIP-2718 transactions, **not** from `Transaction` with `Union` selector data.

---

"Why do we even have this `Union`?" you may ask, well:

The benefit of `Union`:
- Merkleize the selector and value as two separate binary tree nodes; elegant proof to get the type, then handle the value subtree appropriately
- Option to specify other transactions in the future as SSZ type (nice merkleization of transaction contents, no RLP, etc.)

The problems of `Union`:
- `selector` byte and EIP-2718 byte are *different*, can be confusing
- Not specified how additional SSZ typed transactions would make their way into the execution engine, or even affect the transaction trie or block hash

[...]

PR #2684, opened 2021-10-19
Remove Union from ExecutionPayload transaction type

djrtwo:
It seems that we are at an impasse as to how to properly use a Union-type for Transactions that maps well to the execution layer native typing. (based on convo here -- https://github.com/ethereum/consensus-specs/issues/2608)

This PR removes the `Union` type completely, making the execution layer transactions just opaque bytes. This is nice because we don't tightly couple Consensus Layer changes if/when any new Execution Layer txs are added/modified.

This keeps a cleaner separation of concerns and allows the layers to be specified (and even upgraded) relatively independently. I'm personally pro keeping this divide clean where possible. The fact that we can't agree on a proper Union selector typing is because of decisions that need to be made in EL (or that CL would imply a decision needs to be made in a certain way eventually). This goes to show us that crossing these layers in specs is complication inducing and imo should be avoided when possible.

------------

additionally, I added the Execution presets to the associated yaml files

### DIFF
--- a/presets/mainnet/merge.yaml
+++ b/presets/mainnet/merge.yaml
@@ -1,3 +1,16 @@
 # Mainnet preset - The Merge
 
-# No presets here.
+# Execution
+# ---------------------------------------------------------------
+# 2**20 (= 1,048,576)
+MAX_BYTES_PER_TRANSACTION: 1048576
+# 2**14 (= 16,384)
+MAX_TRANSACTIONS_PER_PAYLOAD: 16384
+# 2**8 (= 256)
+BYTES_PER_LOGS_BLOOM: 256
+# 2**10 (= 1,024)
+GAS_LIMIT_DENOMINATOR: 1024
+# 5,000
+MIN_GAS_LIMIT: 5000
+# 2**5 (= 32)
+MAX_EXTRA_DATA_BYTES: 32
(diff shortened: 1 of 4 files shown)


djrtwo:
cc: @MicahZoltu

presets/mainnet/merge.yaml
```diff
@@ -1,3 +1,16 @@
...
+# 2**14 (= 16,384)
+MAX_TRANSACTIONS_PER_PAYLOAD: 16384
+# 2**8 (= 256)
+BYTES_PER_LOGS_BLOOM: 256
+# 2**10 (= 1,024)
+GAS_LIMIT_DENOMINATOR: 1024
+# 5,000
+MIN_GAS_LIMIT: 5000
+# 2**5 (= 32)
+MAX_EXTRA_DATA_BYTES: 32
```
mkalinin, line 16: Shouldn't `BYTES_PER_LOGS_BLOOM` and `MAX_EXTRA_DATA_BYTES` be constant values disregarding the network, i.e. out of the scope of *any* configuration, even the preset one?

protolambda: This is fine. Presets offer compile-time configuration for clients that want to optimize usage of these and/or use static typing. They are not configurable per network, unless the testnet warrants a customized preset and clients compile with that. Flexible enough if we ever need a network that changes them, but not part of the default runtime config. See config/preset docs for more info.

PR #2686, opened 2021-10-20
fix gossip and tx size limits for the merge

djrtwo:
This PR addresses two issues:
1. Adjust upper bound for `Transaction` byte size allow for actual bounds within current gas-limits and much higher
2. Adjust the maximum gossip size to account for potential theoretical max size `ExecutionPayload` and to be compatible with network limits in geth/pow chain today.

These issues were originally pointed out by @dankrad here -- https://github.com/ethereum/consensus-specs/pull/2607/files#r711053462. Please see there for more context and discussion

Discussion points for PR:

1. Assuming just  single TX filled with calldata (16 gas per byte) fills an `ExecutionPayload`, then `MAX_BYTES_PER_TRANSACTION` of `2**24` can support a gas limit of up to ~270M (`2**24 * 16`). This *seems* safe at ~10x the current gaslimit, but considering that this is a consensus value that would require a hard-fork to upgrade and that we have "safe" limits in the p2p spec, ***should we instead push this limit to support a a higher gas-limit (e.g. use `2**26` which would support a 1B gas limit)?*** The cost here is extra depth in the `Transaction` object merkle tree which would result in a few more hashes per materialization in the normal case.
2. `MAX_TRANSACTIONS_PER_PAYLOAD` is currently `2**14`. At a minimum gas cost of 21000 gas per transaction (simple send), this supports up to a ~344M gas limit. so on the same order as the `2**24` limit on `MAX_BYTES_PER_TRANSACTION`. If we adjust max bound based on discussion point (1), we should similarly up this amount to be on the same order.

-------------------

EDIT:

After discussion of the above two pounts, `MAX_BYTES_PER_TRANSACTION` is increased to `2**30` and `MAX_TRANSACTIONS_PER_PAYLOAD` is increased to `2**20`. This ensures that both datastructures when used in the extreme can handle gas limits of ~20B.  This is 1000x greater than the gas limit today and thus, barring incredible/unforseen circumstances this allows for growth without concern.

Some discussion on the "cost" involved in hashing here
https://github.com/ethereum/consensus-specs/pull/2686#discussion_r733805807

### DIFF
--- a/specs/merge/p2p-interface.md
+++ b/specs/merge/p2p-interface.md
@@ -27,6 +27,8 @@ Readers should understand the Phase 0 and Altair documents and use them as a bas
 - [Design decision rationale](#design-decision-rationale)
   - [Gossipsub](#gossipsub)
     - [Why was the max gossip message size increased at the Merge?](#why-was-the-max-gossip-message-size-increased-at-the-merge)
+  - [Req/Resp](#reqresp)
+    - [Why was the max chunk response size increased at the Merge?](#why-was-the-max-chunk-response-size-increased-at-the-merge)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 <!-- /TOC -->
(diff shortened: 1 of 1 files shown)


dankrad:
As a further input on this, I think that once statelessness becomes a reality, raising the gas limit by a significant factor will be quite realistic.
The 10x factor should still give enough headroom for a few years, but I wouldn't bet on more than 5 for sure.

presets/mainnet/merge.yaml
```diff
@@ -2,8 +2,8 @@
 
 # Execution
 # ---------------------------------------------------------------
-# 2**20 (= 1,048,576)
-MAX_BYTES_PER_TRANSACTION: 1048576
+# 2**24 (= 16,777,216)
```
protolambda, line 5: If we go from `2**20` to `2**30` (instead of `2**24`) we add 10 hashes work (of 64 bytes input) per transaction to mix-in zero-hashes. At 1000 transactions per block, that's 10.000 hashes. The participation registry in altair is mostly re-merkleized (sparse and many changes, condensed leaf nodes), and at 250k validators (mainnet today) (packed uint8, 32 per leaf) that translates to `(250000/32)*2-1=15624` hashes. So this compares to adding another one of those in terms of performance cost. Worth the future-compatibility I think.

PR #2693, opened 2021-10-25
Eth1 block hashes have type Hash32

benjaminion:
[Same as #2689, see comments there. Sorry - I got in a mess with my branches and needed to redo this.]

Pernickety and largely of historical value now, but since we have added a custom `Hash32` type for Eth1 block hashes it ought to be used in the definition of `initialize_beacon_state_from_eth1()`.

Related question: why are we not migrating all the other hashes over to the `Hash32` type, only the Eth1 block hash? For example [here](https://github.com/ethereum/consensus-specs/blob/a20f6f7b5f40292c2a864337336bff74ae318354/specs/phase0/beacon-chain.md?plain=1#L633)

```
#### `hash`

`def hash(data: bytes) -> Bytes32` is SHA256.
```

Historical note: `Hash32` [used to be a thing](https://github.com/ethereum/consensus-specs/commit/0b10b0444a4e6c4f5bf548344796df76dd6d5657) back at the dawn of time, before being superseded by the generic `Bytes32`.

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -685,7 +685,7 @@ This helper function is only for initializing the state for pure Altair testnets
 *Note*: The function `initialize_beacon_state_from_eth1` is modified: (1) using `ALTAIR_FORK_VERSION` as the current fork version, (2) utilizing the Altair `BeaconBlockBody` when constructing the initial `latest_block_header`, and (3) adding initial sync committees.
 
 ```python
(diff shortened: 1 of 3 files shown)


djrtwo (commented):
there is an `initialize_beacon_state_from_eth1` function in each fork. Let's make the function signatures match in this update

hwwhww (approved):
I updated `initialize_beacon_state_from_eth1` in other forks.
LGTM. 👍

PR #2699, opened 2021-10-29
remove gas validations from CL

djrtwo:
I think that having `is_valid_gas_limit` violates the separation of layers for zero gain. By hoisting this logic into CL while retaining the validations in EL, we now create two places where specs need to be maintained, client logic built and tested, and for updates to be made in the event of changes.

Sometimes, cross-layers will be unavoidable, but this seems like a place we can and should avoid

I'd also argue that the following can/should be removed as well but I have less of a strong stance because this logic is incredible simple and unlikely to change. That said, there is minimal value in having them in CL.
```python
        assert payload.parent_hash == state.latest_execution_payload_header.block_hash
        assert payload.block_number == state.latest_execution_payload_header.block_number + uint64(1)
```


specs/merge/beacon-chain.md
```diff
@@ -264,39 +263,15 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
 
 ```python
 def process_execution_payload(state: BeaconState, payload: ExecutionPayload, execution_engine: ExecutionEngine) -> None:
-    # Verify consistency of the parent hash, block number, base fee per gas and gas limit
+    # Verify consistency of the parent hash and block number
     # with respect to the previous execution payload header
     if is_merge_complete(state):
         assert payload.parent_hash == state.latest_execution_payload_header.block_hash
         assert payload.block_number == state.latest_execution_payload_header.block_number + uint64(1)
-        assert is_valid_gas_limit(payload, state.latest_execution_payload_header)
```
protolambda, line 299: How about just adding a spacious sanity limit on the gas limit? Similar to how we limit block size to 10 MiB?

protolambda (approved):
Generally agree with removing it. We could put a sanity limit in place, but it might be more trouble for what it's worth (if the block data is not too large, it can be passed to execution engine, and dropped as invalid there).

specs/merge/beacon-chain.md
djrtwo: I'd rather have this in the p2p layer (which it is) than try to define a value here that would require consensus change to patch and can't be a function of the dnyamic gas-limit in EL

djrtwo:
Thanks for the review! Will leave this up a bit for some more eyes before merging

mkalinin (approved):
I agree with the arguments. My original intention of bringing on these changes was optimisation for the future stateless world. Apparently this optimisation was premature and now I think that even in a stateless world these checks should be done by EL as they are related to the execution.

The block hash check is very important and we must never remove it. This is what enforces mapping of one block tree on another, i.e. one CL block to one EL block (ok, many-to-one to be strict) instead of many-to-many.

The block number check is enshrined into a block header validity that is validated by EL and is redundant. But it's cheap and it's a kind of axiom that isn't a subject to change. So, we can leave it here or remove it for conformance with proposed change. I don't have a strong opinion in that regard.

The proposal looks good to me. One thing, we should remove the `gas_limit` check from p2p-interface as well

terencechain (commented):
I think you also can remove `GAS_LIMIT_DENOMINATOR` and `MIN_GAS_LIMIT` constants

djrtwo:
> The block hash check is very important and we must never remove it. This is what enforces mapping of one block tree on another, i.e. one CL block to one EL block (ok, many-to-one to be strict) instead of many-to-many.

If we assume that EL does a consistency check on the fact that a blockchain is formed -- which is even implicit if using the `parent_hash` it pulls and executes on top of the correct pre-state -- then I don't understand why CL *also* performing this check that we gain anything?

We are  removing gas parent-to-child consistency checks from CL because EL does it.
What is the difference here by removing the parent_hash consistency check if EL does it?

ralexstokes:
just going to chime in that i agree w/ the changes in this PR *and* put my vote in to remove as much as possible.

to that end, i'd support removing the parent hash check and the block number check

ralexstokes (approved):
left comment below. generally approve!

djrtwo:
> One thing, we should remove the gas_limit check from p2p-interface as well

Done @mkalinin 

> you also can remove GAS_LIMIT_DENOMINATOR and MIN_GAS_LIMIT constants

Done @terencechain 

Thank you!

mkalinin:
> What is the difference here by removing the `parent_hash` consistency check if EL does it?

We can defer all checks to EL except for this one. If there is no `prent_hash` check enforced by CL then CL's blockchain may jump between different EL blockchains. The inconsistency of the block tree versions between the layers makes lock-step sync degenerate as there is no guarantee that EL has all inputs for the child block in the case when the parent one has been imported successfully.

djrtwo:
okay, I see what you are getting at. We want CL to know the blockchain datastructure is respected so that we don't accidentally trigger a sync or otherwise missing-data problem

We can leave it in

PR #2700, opened 2021-10-30
How to write tests for the consensus layer

qbzzt:


tests/README.md
```diff
@@ -0,0 +1,515 @@
...
+   git clone https://github.com/protolambda/zrnt.git
+   cd zrnt
+   make download-tests
+   ```
+   
+1. Run the attestation tests
+
+   ```sh
+   cd tests/spec
+   go test -tags preset_minimal ./test_runners/operations/attestation_test.go
```
protolambda, line 494: The `-tags preset_minimal ` build-tag isn't necessary anymore, ZRNT currently supports full runtime configuration of everything. I need to update my docs :sweat_smile:

tests/README.md
```diff
@@ -0,0 +1,515 @@
...
+name located in 
+[`~/consensus-specs/tests/core/pyspec/eth2spec/test/phase0/sanity/test_blocks.py`](https://github.com/ethereum/consensus-specs/blob/dev/tests/core/pyspec/eth2spec/test/phase0/sanity/test_blocks.py).
+To learn how consensus spec tests are written, let's go over the code:
+
+```python
+@with_all_phases
+```
+
+This [decorator](https://book.pythontips.com/en/latest/decorators.html) specifies that this test
+is applicable to all the phases of the ETH 2.0 project. These phases are similar to forks (Istanbul,
```
protolambda, line 73: We are trying to deprecate the Eth2 nomenclature, in favor of "ethereum consensus layer"

tests/README.md
```diff
@@ -0,0 +1,515 @@
+# Getting Started with Consensus Spec Tests
+
+## Getting Started
+
+### Creating the environment
```
protolambda, line 5: Although is nice, it overlaps somewhat with https://github.com/ethereum/consensus-specs/blob/dev/tests/core/pyspec/README.md

Maybe we can combine them, or update the existing doc and reference it here? What you added here does look good.

tests/README.md
```diff
@@ -0,0 +1,515 @@
...
+    pre_mix = spec.get_randao_mix(state, spec.get_current_epoch(state))
+```
+
+Store some values to check later that certain updates happened.
+
+```python
+    yield 'pre', state
+```
+
+In Python `yield` is used by [generators](https://wiki.python.org/moin/Generators). However, for our purposes
```
protolambda, line 111: It would be nice to clarify why we are yielding this. Reason why: the tests can be executed in pytest, but also used as functions in the test-generator script, to create test-vectors from these yielded outputs.

About test generators: https://github.com/ethereum/consensus-specs/blob/dev/tests/generators/README.md
About test formats (what is yielded per type of test): https://github.com/ethereum/consensus-specs/tree/dev/tests/formats (you linked it later in the doc, but it may help clarify these `yield` things)

tests/README.md
```diff
@@ -0,0 +1,515 @@
...
+-->
+
+## How are These Tests Used?
+
+So far we've ran tests against the formal specifications. This is a way to check the specifications
+are what we expect, but it doesn't actually check the beacon chain clients. The way these tests get applied
+by clients is that every few days 
+[new test specifications are released](https://github.com/ethereum/consensus-spec-tests/releases),
+in a format [documented here](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats).
+All the clients know how to use tests in the format.
```
protolambda, line 473: Maybe rephrase to "implement test-runners that consume the test vectors in this standard format"

tests/README.md
```diff
@@ -0,0 +1,515 @@
...
+. venv/bin/activate
+cd tests/core/pyspec
+python -m pytest -k almost_after --fork Merge eth2spec
+```
+
+You should see it ran successfully (although you might get a warning, you can ignore it)
+
+<!--
+
+https://ethos.dev/beacon-chain/
```
protolambda, line 462: This is an older website, and maybe not so relevant for testing. We can aggregate starter resources in another place.

tests/README.md
```diff
@@ -0,0 +1,515 @@
...
+## How are These Tests Used?
+
+So far we've ran tests against the formal specifications. This is a way to check the specifications
+are what we expect, but it doesn't actually check the beacon chain clients. The way these tests get applied
+by clients is that every few days 
+[new test specifications are released](https://github.com/ethereum/consensus-spec-tests/releases),
+in a format [documented here](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats).
+All the clients know how to use tests in the format. 
+
+<!--
```
protolambda, line 475: Intentionally an HTML comment?

protolambda (commented):
Hi Ori, nice to find you here in the consensus specs repo, thanks for the help documenting testing!

Some nitpicks in review, but generally this is very helpful, thank you :+1:

qbzzt:
> Hi Ori, nice to find you here in the consensus specs repo, thanks for the help documenting testing!
> 
> Some nitpicks in review, but generally this is very helpful, thank you 👍

Nice to find you here too. Yes, I've been writing helping out EF with documentation and testing for a while now, long before I've even heard of Optimism.

I'm glad it helps. It's always difficult to know how much detail to put in a tutorial.

tests/README.md
qbzzt: Yes, this is the way to leave a comment in a markdown file too. But as I'm close to done with the document, I should probably remove most comments.

tests/README.md
qbzzt: The ZRNT material is commented out. I decided that running a test in ZRNT isn't impressive enough to be worth doing, all you can see is that the test passed successfully. Or am I wrong, and there's a way for ZRNT to produce an interesting trace (similar to the vmtrace in the execution layer)?

tests/README.md
qbzzt: It does, but that is because a reference and a tutorial on the same subject are going to have some duplication. In a tutorial I need to give people everything to get from a bare OS to being able to do whatever I'm teaching, such as consensus layer tests.

I could update https://github.com/ethereum/consensus-specs/blob/dev/tests/core/pyspec/README.md, but I'd rather also keep all of this material here. Is that OK?

tests/README.md
qbzzt: I need to think about it. On one hand extra knowledge is always good. On the other, in a tutorial I try to minimize the cognitive load by having people focus on what they need to know to do the job (plus to answer any questions that appear to me to be something students will think about, so they won't have to figure the answer themselves).

Maybe I should explain in this the section about how these tests are used. But I should think about it when I'm more awake.

qbzzt: I linked to the generators README.md. I think that would be enough - whoever wants to read can read.

tests/README.md
qbzzt: I included a link to https://github.com/ethereum/consensus-specs/blob/dev/tests/core/pyspec/README.md.

tests/README.md
```diff
@@ -0,0 +1,475 @@
...
+
+## Attestation Tests
+
+The [beacon chain](https://ethereum.org/en/eth2/beacon-chain/) doesn't provide any direct value. It does
+not execute EVM programs or store user data. It exists to provide a trusted source of
+information about the latest verified block hash of the [shard blockchains](https://ethereum.org/en/eth2/shard-chains/)
+which do provide storage, and possibly execution, services.
+
+For every slot a validator is randomly selected as the proposer. Currently the proposer proposes the hash
+for the current head of the beacon chain (the previous block). When shards are added, the proposer will also
```
djrtwo, line 295:
```suggestion
for the current head of the beacon chain (built on the previous block).
```

tests/README.md
```diff
@@ -0,0 +1,475 @@
...
+propose a hash for the head of the assigned shard.
```
djrtwo, line 296:
```suggestion
```

djrtwo: I would just cut out a lot of the sharding stuff. There are actually separate proposrs that build the shard chains and it's probably not worth getting in the nuance here.

tests/README.md
```diff
@@ -0,0 +1,475 @@
...
+python -m pytest -k almost_after --fork Merge eth2spec
+```
+
+You should see it ran successfully (although you might get a warning, you can ignore it)
+
+## How are These Tests Used?
+
+So far we've ran tests against the formal specifications. This is a way to check the specifications
+are what we expect, but it doesn't actually check the beacon chain clients. The way these tests get applied
+by clients is that every few days
```
djrtwo, line 468:
```suggestion
by clients is that every few weeks 
```

djrtwo (commented):
nice! just a handful of review comments

djrtwo (approved):

PR #2702, opened 2021-11-01
Set execution params to presets

hwwhww:
Spec builder fixes:

1. The execution params should have been moved to "presets" since they are defined in preset files.
2. ~~Add spec builder assertions to avoid #2701 issue happening again~~
    - ~~It will add `assert MAX_BYTES_PER_TRANSACTION == uint64(1073741824)` to the spec.~~
    - updated: reverted it

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -634,6 +634,10 @@ def objects_to_spec(preset_name: str,
     ssz_dep_constants = '\n'.join(map(lambda x: '%s = %s' % (x, builder.hardcoded_ssz_dep_constants()[x]), builder.hardcoded_ssz_dep_constants()))
     ssz_dep_constants_verification = '\n'.join(map(lambda x: 'assert %s == %s' % (x, spec_object.ssz_dep_constants[x]), builder.hardcoded_ssz_dep_constants()))
     custom_type_dep_constants = '\n'.join(map(lambda x: '%s = %s' % (x, builder.hardcoded_custom_type_dep_constants()[x]), builder.hardcoded_custom_type_dep_constants()))
+    custom_type_dep_constants_verification = '\n'.join(map(
+        lambda x: 'assert %s == %s' % (x, f'{spec_object.preset_vars[x].type_name}({spec_object.preset_vars[x].value})'),
+        builder.hardcoded_custom_type_dep_constants(),
+    ))
     spec = (
             builder.imports(preset_name)
             + builder.preparations()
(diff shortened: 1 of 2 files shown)


hwwhww:
Second thought: the assertion would be incorrect *if* minimal preset value is different from the mainnet preset value. I can revert it if they are likely to be changed.

specs/merge/beacon-chain.md
```diff
@@ -64,7 +63,7 @@ Additionally, this upgrade introduces the following minor changes:
 | `Transaction` | `ByteList[MAX_BYTES_PER_TRANSACTION]` | either a [typed transaction envelope](https://eips.ethereum.org/EIPS/eip-2718#opaque-byte-array-rather-than-an-rlp-array) or a legacy transaction|
 | `ExecutionAddress` | `Bytes20` | Address of account on the execution layer |
 
-## Constants
+## Preset
```
ralexstokes, line 66: ah nice catch on these!

setup.py
```diff
@@ -634,6 +634,10 @@ def format_constant(name: str, vardef: VariableDefinition) -> str:
     ssz_dep_constants = '\n'.join(map(lambda x: '%s = %s' % (x, builder.hardcoded_ssz_dep_constants()[x]), builder.hardcoded_ssz_dep_constants()))
     ssz_dep_constants_verification = '\n'.join(map(lambda x: 'assert %s == %s' % (x, spec_object.ssz_dep_constants[x]), builder.hardcoded_ssz_dep_constants()))
     custom_type_dep_constants = '\n'.join(map(lambda x: '%s = %s' % (x, builder.hardcoded_custom_type_dep_constants()[x]), builder.hardcoded_custom_type_dep_constants()))
+    custom_type_dep_constants_verification = '\n'.join(map(
+        lambda x: 'assert %s == %s' % (x, f'{spec_object.preset_vars[x].type_name}({spec_object.preset_vars[x].value})'),
+        builder.hardcoded_custom_type_dep_constants(),
+    ))
```
ralexstokes, line 640: as you noted this does not respect which `preset` as there is no differentiation in `hardcoded_custom_type_dep_constants` -- i think we will be ok without this assertion personally

ralexstokes (commented):
nice work!

i agree w/ you that we would need to extend the hardcoded constants handling to respect `presets` and in my opinion doing that work doesn't pull its own weight

the other change should definitely be merged in -- nice catch!

hwwhww:
@ralexstokes Thanks for the review! I reverted the `setup.py` changes.

ralexstokes (approved):
thanks for the update! 

lgtm!

PR #2703, opened 2021-11-01
State that validator must consider only fully validated blocks

mkalinin:
Adds the following note:
> A validator must not propose on or attest to a block that isn't deemed valid, i.e. hasn't yet passed the beacon chain state transition and execution validations.

### DIFF
--- a/specs/merge/validator.md
+++ b/specs/merge/validator.md
@@ -122,6 +122,8 @@ def get_payload(self: ExecutionEngine, payload_id: PayloadId) -> ExecutionPayloa
 
 All validator responsibilities remain unchanged other than those noted below. Namely, the transition block handling and the addition of `ExecutionPayload`.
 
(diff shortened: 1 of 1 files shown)


specs/merge/validator.md
```diff
@@ -122,6 +122,8 @@ def get_payload(self: ExecutionEngine, payload_id: PayloadId) -> ExecutionPayloa
 
 All validator responsibilities remain unchanged other than those noted below. Namely, the transition block handling and the addition of `ExecutionPayload`.
 
+*Note*: A validator must not propose on or attest to a block that isn't deemed valid, i.e. hasn't yet passed the beacon chain state transition and execution validations.
```
ralexstokes, line 125: could even reach and specify these validations have to happen *locally*, i.e. you shan't call out to infura, etc.

ralexstokes (approved):
seems good!

djrtwo (approved):
lgtm

PR #2706, opened 2021-11-02
Make `Phase0 -> Altair` fork transition tests compatible with `Altair -> Merge`

hwwhww:
### Issue
I really want to reuse `test/altair/transition/` tests instead of duplicating it for Merge fork transition.

#### High-level
1. Add `ForkMeta` dataclass for passing `pre_fork_name`, `post_fork_name` and `fork_epoch`.
2. Replace `@fork_transition_test(PHASE0, ALTAIR, fork_epoch=2)` with `@with_fork_metas(Sequence[ForkMeta])`.
3. Replace `do_altair_fork` with `do_fork`, which handles both Altair and The Merge updates.

#### Decorators
1. Modify `build_transition_test` into `yield_fork_meta`.
2. Modify `with_phases`:
    - Refactoring:
        - Replace the `if PHASE0: ... if ALTAIR: ...` code with for-loop. The phases should be been limited with `DEFAULT_PYTEST_FORKS`.
    - Add `for fork_meta in fork_metas` loop to run test cases for each fork meta.

#### Other
- Make `test_transition_with_voluntary_exit_*` only run with minimal presets.
- Make transition test generator run with `Merge`.

#### Note
- Run pytest with `--fork=phase0` will get you `Phase0 -> Altair` fork transition; run pytest with `--fork=altair` will get you `Altair -> Merge` fork transition.

---

## TODO
- [x] Verify the test vectors.


tests/core/pyspec/eth2spec/test/altair/transition/test_activations_and_exits.py
```diff
@@ -127,10 +130,6 @@ def test_transition_with_one_fourth_exiting_validators_exit_at_fork(state,
         assert not post_spec.is_active_validator(validator, post_spec.get_current_epoch(state))
     assert not post_spec.is_in_inactivity_leak(state)
 
-    # ensure that none of the current sync committee members are exited validators
-    exited_pubkeys = [state.validators[index].pubkey for index in exited_indices]
-    assert not any(set(exited_pubkeys).intersection(list(state.current_sync_committee.pubkeys)))
```
hwwhww, line 132: I don't remember why we needed it. 😬

ralexstokes: there was a test where we did want to ensure some had exited, and this was the complement case where we want to ensure none in the committee are exited

ralexstokes: unless i'm missing something, we should keep this functionality

ralexstokes (commented):
still digesting but so far it looks great! one general comment is that this would be easier to review if it split apart the various concerns.

I see the following that could be N separate PRs:

1. refactor of `with_phases` and additional typing in that file
2. the extension of the fork transition tests to re-use them across multiple pairs of forks/phases

The `with_phases` refactor isn't huge but it is an important enough function to the integrity of the spec tests that changes to it should be very carefully reviewed! It would be a bit nicer to be able to consider it independently without the cognitive load of the other changes to the fork transition testing.

No worry on this PR but it is something to keep in mind in the future :)

tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -347,56 +359,91 @@ def decorator(fn):
...
+    """
+    if 'phase' in kw:
+        # Limit phases if one explicitly specified
+        phase = kw.pop('phase')
+        if phase not in phases:
+            dump_skipping_message(f"doesn't support this fork: {phase}")
+            return None
+        run_phases = [phase]
+    else:
+        # If pytest `--fork` flag is set, filter out the rest forks
```
ralexstokes, line 381:
```suggestion
        # If pytest `--fork` flag is set, filter out the rest of the forks
```

tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -347,56 +359,91 @@ def decorator(fn):
...
+        run_phases = set(phases).intersection(DEFAULT_PYTEST_FORKS)
+
+    return run_phases
+
+
+def _get_available_phases(run_phases, other_phases):
+    """
+    The return the available fork names for multi-phase tests
```
ralexstokes, line 389:
```suggestion
    Return the available fork names for multi-phase tests
```

?

tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -502,15 +564,62 @@ def fork_transition_test(pre_fork_name, post_fork_name, fork_epoch=None):
...
+def yield_fork_meta(fork_metas: Sequence[ForkMeta]):
+    """
+    Yield meta fields to `meta.yaml` and pass post spec and meta fields to `fn`.
+    """
+    def decorator(fn):
+        def wrapper(*args, **kw):
+            phases = kw.pop('phases')
+            spec = kw["spec"]
+            fork_meta = [m for m in fork_metas if m.pre_fork_name == spec.fork][0]
+            fork_meta = next(filter(lambda m: m.pre_fork_name == spec.fork, fork_metas))
```
ralexstokes, line 585: these look like duplicate lines?

tests/generators/transition/main.py
```diff
@@ -31,18 +36,17 @@ def cases_fn() -> Iterable[gen_typing.TestCase]:
...
+        test_altair_leaking,
+        test_altair_slashing,
+        test_altair_operations,
+    )
+    for transition_test_module in altair_tests:
+        for pre_fork, post_fork in ALL_FORKS.items():
+            gen_runner.run_generator("transition", [
+                create_provider(transition_test_module, MINIMAL, pre_fork, post_fork),
+                create_provider(transition_test_module, MAINNET, pre_fork, post_fork),
+            ])
```
ralexstokes, line 52: and let's add `merge_tests` so we can try the refactor out!!

tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -347,56 +359,91 @@ def decorator(fn):
...
+    """
+    The return the available fork names for multi-phase tests
+    """
+    available_phases = set(run_phases)
+    if other_phases is not None:
+        available_phases |= set(other_phases)
+    return available_phases
+
+
+def _run_test_case_with_phases(fn, phases, other_phases, kw, args, is_fork_transition=False):
```
ralexstokes, line 397: one thing that seems a little off is that we are passing information about one specific type of test, the `fork_transition` test type, into this function signature when this function should work generically across any type of test....

i think to change it would require some substantial refactoring so this may be ok for now, but i'll leave this comment to call it to our attention...

and more generally, reading all of this code makes me realize we may be starting to stretch how much we are using this test infra, e.g. there are references to phases, but also forks. and specs, but also configs and presets. it is not something to do immediately but a broader refactor may be on the horizon!

ralexstokes (commented):
this looks great!

i like that we can reuse the fork transition tests across multiple pairs of forks.

i left some things below, lets address those and then i'll make another round of review... in particular, it would be good to see the tests actually generated, so we can see all of the new code exercised

tests/core/pyspec/eth2spec/test/context.py
hwwhww: > `is_fork_transition`

Right, I want to get rid of it too but (i) the place we do `for fork_meta in fork_metas` for-loop and (ii) the place we get `run_phases` are mixed together.

There are three cases around `run_phases`:
1. From test generator call: `kw['phase']` is assigned
2. From pytest call, without `--fork` flag: use all phases that the test cases are assigned
3. From pytest call, with `--fork` flag: use `DEFAULT_PYTEST_FORKS` would filter `phases`

The current drawback: because `--fork` filter is inside `_run_test_case_with_phases`, if I set `--fork=phase0`, it still runs `_run_test_case_with_phases` two times due to the `for fork_meta in fork_metas`. Therefore this `is_fork_transition` is here to prevent it from signaling `pytest.skip`.

I'll take a look if it can be refactored in a better way. 👀 

--

> and more generally, reading all of this code makes me realize we may be starting to stretch how much we are using this test infra, e.g. there are references to phases, but also forks. and specs, but also configs and presets. it is not something to do immediately but a broader refactor may be on the horizon!

Totally agreed! It bothered me for a long time and is worth another deep clean-up.

tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -502,15 +564,62 @@ def fork_transition_test(pre_fork_name, post_fork_name, fork_epoch=None):
...
+
+def yield_fork_meta(fork_metas: Sequence[ForkMeta]):
+    """
+    Yield meta fields to `meta.yaml` and pass post spec and meta fields to `fn`.
+    """
+    def decorator(fn):
+        def wrapper(*args, **kw):
+            phases = kw.pop('phases')
+            spec = kw["spec"]
+            fork_meta = [m for m in fork_metas if m.pre_fork_name == spec.fork][0]
```
hwwhww, line 584:
```suggestion
```

tests/core/pyspec/eth2spec/test/context.py
ralexstokes: yeah i'm not sure there is a better solution for now 😅 

worth keeping in mind all of the various use cases if/when we get around to a bigger refactor :)

tests/core/pyspec/eth2spec/test/altair/transition/test_activations_and_exits.py
hwwhww: @ralexstokes thanks! Now I remember why we added it. Fixed it in https://github.com/ethereum/consensus-specs/pull/2706/commits/f0bb032eb17ae441429405b89b7c158623f865c0

ralexstokes (approved):
lgtm!

ill verify if we have verified the test vectors, otherwise i think we can ship this

hwwhww:
tarball files of the transition tests (only `make gen_transition`) at https://github.com/ethereum/consensus-specs/pull/2706/commits/63c9e5ea56b1daa725d2d9ee9a1e368adc9d64d1:

[mainnet.tar.gz](https://github.com/ethereum/consensus-specs/files/7555888/mainnet.tar.gz)
[minimal.tar.gz](https://github.com/ethereum/consensus-specs/files/7555890/minimal.tar.gz)

PR #2711, opened 2021-11-04
Add payload id as a return value to notify_forkchoice_updated

mkalinin:
* Adds `Optional[PayloadId]` as a return value to the `notify_forkchoice_updated` function
* Removes redundant `get_payload_id` function

Relates to https://github.com/ethereum/execution-apis/pull/109


djrtwo (approved):
nice cleanup!

PR #2712, opened 2021-11-05
Fixes TBH activation epoch check

mkalinin:
Fixes `TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH` check in the validator guide


djrtwo (approved):
ha!

The logic was correct but the var names were wrong...

PR #2715, opened 2021-11-07
Clean up SSZ multiproofs spec: unused function

ralexstokes:
The function `concat_generalized_indices` is not currently used anywhere in this repo (and by extension the helper `get_power_of_two_floor`) and a quick search only yields reference from the past PR: https://github.com/ethereum/consensus-specs/pull/1323

We could use this functionality in the future, but currently don't as we don't need any receipt consumption logic in the consensus protocol as it stands today or in the near term.

Given that this code will be in the git history in the event we do need it, I'd suggest removing it for now so that the spec is "lighter" and that implementers of multiproofs don't bother implementing an unnecessary function (if they do what I just did and just mechanically implement the spec from the markdown file line by line ;) )

### DIFF
--- a/ssz/merkle-proofs.md
+++ b/ssz/merkle-proofs.md
@@ -11,7 +11,6 @@
 - [Generalized Merkle tree index](#generalized-merkle-tree-index)
 - [SSZ object to index](#ssz-object-to-index)
   - [Helpers for generalized indices](#helpers-for-generalized-indices)
-    - [`concat_generalized_indices`](#concat_generalized_indices)
     - [`get_generalized_index_length`](#get_generalized_index_length)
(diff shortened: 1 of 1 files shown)


djrtwo:
What are the chances we need this in light client R&D?

protolambda:
Concatenation is an important thing for constructing nested path. E.g. with a container of 5 fields, and access to field index 4, with 7 fields and access to field index 5, you get: `0b1_100 ++ 0b1_101 = 0b1_100_101`. There's no other function in the spec which defines how this works otherwise.

djrtwo:
yeah, similar feeling. Although it's not *used* in the spec, it's good to be defined for others

PR #2720, opened 2021-11-15
Remove `difficulty` from `PowBlock`

paulhauner:
The `difficulty` field is unused in the specification and appears to be always set to `0` in fork choice tests.


mkalinin (approved):
It seems to be a leftover from the times when the TTD was computed during the fork on the beacon chain. We may remove it now 👍

hwwhww:
Nice. I'll go fix the CI.

hwwhww (approved):

djrtwo (approved):

Issue #2722, opened 2021-11-16
Untested Merge spec code from coverage report


hwwhww:
With `--fork=merge` flag

- [ ] [`if is_execution_enabled(state, block.body):`](https://github.com/ethereum/consensus-specs/blob/79d005e6ed89cc44d233dd9320c5cc1e7ea91720/specs/merge/beacon-chain.md#block-processing): the condition was never false
    - #2723
- [ ]  Validator guide helpers
    - [ ] `get_pow_block_at_terminal_total_difficulty`: #2726
    - [ ] `get_terminal_pow_block`
    - [ ] `prepare_execution_payload`
    - [ ] `get_execution_payload`

PR #2726, opened 2021-11-17
Fix `get_pow_block_at_terminal_total_difficulty` and add unit tests

hwwhww:
Part of #2722

### Spec changes
1. `get_pow_block_at_terminal_total_difficulty` was wrongly iterating `pow_chain` keys. It should be dict values `pow_chain.values()` instead of `pow_chain`.
2. Handle the edge case of `block.parent_hash == Hash32() and block.parent_hash not in pow_chain`. Without handling it, it would raise `KeyErorr` in pyspec.
    - ~~New concern: it's the first time to use `continue` in specs. Is it too Pythonic and difficult to do formal verification?~~ fixed

### Testing
Add unit tests for `get_pow_block_at_terminal_total_difficulty`

### DIFF
--- a/specs/merge/validator.md
+++ b/specs/merge/validator.md
@@ -45,19 +45,15 @@ def get_pow_block_at_terminal_total_difficulty(pow_chain: Dict[Hash32, PowBlock]
     # `pow_chain` abstractly represents all blocks in the PoW chain
     for block in pow_chain.values():
         block_reached_ttd = block.total_difficulty >= TERMINAL_TOTAL_DIFFICULTY
-        # Genesis block
-        if block.parent_hash == Hash32():
-            # No parent exists, so reaching TTD alone qualifies as valid terminal block
-            if block_reached_ttd:
+        if block_reached_ttd:
+            # If genesis block, no parent exists so reaching TTD alone qualifies as valid terminal block
+            if block.parent_hash == Hash32():
                 return block
-            # Continue to iterate the next block
-            if block.parent_hash not in pow_chain:
-                continue
-
-        parent = pow_chain[block.parent_hash]
-        parent_reached_ttd = parent.total_difficulty >= TERMINAL_TOTAL_DIFFICULTY
-        if block_reached_ttd and not parent_reached_ttd:
-            return block
+            else:
+                parent = pow_chain[block.parent_hash]
+                parent_reached_ttd = parent.total_difficulty >= TERMINAL_TOTAL_DIFFICULTY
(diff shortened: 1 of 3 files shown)


specs/merge/validator.md
```diff
@@ -43,11 +43,17 @@ Please see related Beacon Chain doc before continuing and use them as a referenc
...
 def get_pow_block_at_terminal_total_difficulty(pow_chain: Dict[Hash32, PowBlock]) -> Optional[PowBlock]:
     # `pow_chain` abstractly represents all blocks in the PoW chain
-    for block in pow_chain:
+    for block in pow_chain.values():
         block_reached_ttd = block.total_difficulty >= TERMINAL_TOTAL_DIFFICULTY
-        # If genesis block, no parent exists so reaching TTD alone qualifies as valid terminal block
-        if block_reached_ttd and block.parent_hash == Hash32():
-            return block
+        # Genesis block
+        if block.parent_hash == Hash32():
```
mkalinin, line 49: What if we do it as follows:
```python
if block_reached_ttd:
    if block.parent_hash == Hash32():
        return block
    else:
        parent = pow_chain[block.parent_hash]
        parent_reached_ttd = parent.total_difficulty >= TERMINAL_TOTAL_DIFFICULTY
        if not parent_reached_ttd:
            return block
```

mkalinin (approved):
Looks good to me! 👍

specs/merge/validator.md
```diff
@@ -43,15 +43,17 @@ Please see related Beacon Chain doc before continuing and use them as a referenc
...
-            return block
+        if block_reached_ttd:
+            # If genesis block, no parent exists so reaching TTD alone qualifies as valid terminal block
+            if block.parent_hash == Hash32():
+                return block
+            else:
+                parent = pow_chain[block.parent_hash]
+                parent_reached_ttd = parent.total_difficulty >= TERMINAL_TOTAL_DIFFICULTY
+                if not parent_reached_ttd:
+                    return block
```
djrtwo, line 56: Moderate prefrence for the following: the removal of `else` after the `return` in the above `if` signals that this is the more "standard path" and the above `if` was the execptional "blocking" condition
```suggestion
            parent = pow_chain[block.parent_hash]
            parent_reached_ttd = parent.total_difficulty >= TERMINAL_TOTAL_DIFFICULTY
            if not parent_reached_ttd:
                return block
```

djrtwo (commented):
looks good. minor suggestion

PR #2727, opened 2021-11-18
always atomically update justified and finalized

djrtwo:
When finalized is updated, the fork choice should *always* update to the justification that finalized the new checkpoint atomically with the finality update.

This document goes into the problem unearthed and the solution -- https://notes.ethereum.org/YfStc2i6Rgenk_gzhEojfQ?view

Note that this bug can only be triggered on mainnet with >1/3 slashing, and if you are willing to do that, there are plenty of much more interesting attacks you'd probably pull off. This patch is a *must* prior to the Merge (along with proposer boosting!)

Thank you @realbigsean and @paulhauner for bringing the issue to our attention!

---------

@adiasg's comment copied from [below](https://github.com/ethereum/consensus-specs/pull/2727#pullrequestreview-812756853):

This PR makes 2 changes:
1. Only make atomic updates to store (as discussed above)
2. Allow old attestations from blocks to be processed

Fix for atomic updates to store looks good.
Also good to move ahead with processing old attestations from blocks for now - that's the only way to make atomic updates to the store work in our current testing setup. If that changes in the future, this logic should go through security analysis (esp. for flip-flop attacks).

### DIFF
--- a/tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
+++ b/tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
@@ -694,21 +694,13 @@ def test_new_finalized_slot_is_justified_checkpoint_ancestor(spec, state):
 
     pre_store_justified_checkpoint_root = store.justified_checkpoint.root
     for block in all_blocks:
-        # FIXME: Once `on_block` and `on_attestation` logic is fixed,
-        # fix test case and remove allow_invalid_attestations flag
         yield from tick_and_add_block(spec, store, block, test_steps)
 
     finalized_slot = spec.compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)
     ancestor_at_finalized_slot = spec.get_ancestor(store, pre_store_justified_checkpoint_root, finalized_slot)
     assert ancestor_at_finalized_slot == store.finalized_checkpoint.root
 
-    print(store.finalized_checkpoint)
-    print(store.justified_checkpoint)
-    print(another_state.current_justified_checkpoint)
-    print('spec.get_head(store)', spec.get_head(store))
-
     assert store.finalized_checkpoint == another_state.finalized_checkpoint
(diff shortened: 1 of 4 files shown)


tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
```diff
@@ -628,7 +624,7 @@ def test_new_finalized_slot_is_not_justified_checkpoint_ancestor(spec, state):
     # # Apply blocks of `another_state` to `store`
```
djrtwo, line 624: Is this able to be fixed now @hwwhww ?

hwwhww: Yes, fixed in https://github.com/ethereum/consensus-specs/pull/2727/commits/2c865e362799470356b703afdef83c49bb55bd38.

hwwhww:
Also thank @ajsutton for noticing this issue months ago!

---

### `on_attestation` changes

Along with the `on_block` function update, this PR also brings `on_attestation` function update for making the test case pass correctly. 

We introduce an `is_from_block` flag to `on_attestation` function to signal if the given `attestation` is carried with a `BeaconBlock` message or as a single `Attestation` message.
- If `is_from_block` is `False`, that means it's from `Attestation` message, we have to check `validate_target_epoch_against_current_time`.
- If `is_from_block` is `True`, we skip `validate_target_epoch_against_current_time` validation.

adiasg (approved):
This PR makes 2 changes:
1. Only make atomic updates to store
2. Allow old attestations from blocks to be processed

Fix for atomic updates to store looks good.
Also good to move ahead with processing old attestations from blocks for now - that's the only way to make atomic updates to the store work in our current testing setup. If that changes in the future, this logic should go through security analysis (esp. for flip-flop attacks).

Mitumaru (commented):
Ok

PR #2729, opened 2021-11-19
Update simple-serialize.md

manoj398:
prefer & instead of %


dapplion:
@manoj398 This change is motivated for performance reasons or readability?

ralexstokes:
imo `% 8` is clearer and clarity is more important here than just about any other concern

djrtwo:
Closing. Strong preference to not change something that people are used to reading without a more compelling reason to do so.

Happy to open it back up if there are further discussion points.

manoj398:
If the readability is the concern, imo it should be bitarray or boolean array instead of bytearray.

PR #2730, opened 2021-11-19
Proposer LMD Score Boosting

adiasg:
This PR implements a fix for the [LMD balancing attacks](https://arxiv.org/abs/2009.04987), as described [here](https://notes.ethereum.org/@vbuterin/lmd_ghost_mitigation).

**Note**: Old PR for this feature was #2353.

## Proposer Score Boost

A new "proposer score boost" has been introduced, which boosts the LMD score of blocks that are received in a timely manner. The boost increases the score of the block by `committee_weight // 4` for that slot. If a block for slot `X` is received in the first `SECONDS_PER_SLOT // INTERVALS_PER_SLOT` (= 4) seconds of the slot, then the block receives the boost for the duration of slot `X`. After slot `X` is over, the boost is removed and the usual LMD score calculations from only attestations are done. 

The boost is not applied to blocks that are not received in the first `SECONDS_PER_SLOT // INTERVALS_PER_SLOT` seconds of their slots, and this PR does not affect LMD score calculations for such blocks.

lmd-proposer-score-fix

In the above example:
- block `B` is received in a timely manner and qualifies for the boost. `B`'s score is boosted for only slot `X`.
- block `C` is not received in a timely manner and does not qualify for the boost. 

## Summary of Changes
- Parameters:
  - Added new parameter `ATTESTATION_OFFSET_QUOTIENT`, which is used to specify the time at which attestations are produced within a slot
- `Store`:
  - Added new field `store.proposer_score_boost`, which is a special `LatestMessage` that stores the boost
- Fork Choice:
  - Updated `get_latest_attesting_balance` to take into account `store.proposer_score_boost`
  - Updated `on_block` & `on_tick` to appropriately set/reset `store.proposer_score_boost`
- Testing:
  - Added test for `store.proposer_score_boost` accounting
  - Updated existing tests to work correctly after introducing proposer score boosting


adiasg:
Thanks @joachimneu for pointing out that the proposer score boost should allow for more granularity. 
The parameter `PROPOSER_SCORE_BOOST` is now a percentage value.

specs/phase0/fork-choice.md
```diff
@@ -155,11 +168,23 @@ def get_ancestor(store: Store, root: Root, slot: Slot) -> Root:
...
             and get_ancestor(store, store.latest_messages[i].root, store.blocks[root].slot) == root)
     ))
+    proposer_score = Gwei(0)
+    if store.proposer_score_boost.root != Root():
+        block_slot = store.blocks[root].slot
+        if get_ancestor(store, root, block_slot) == store.proposer_score_boost.root:
+            num_validators = len(get_active_validator_indices(state, get_current_epoch(state)))
+            avg_balance = get_total_active_balance(state) // num_validators
+            block_epoch = compute_epoch_at_slot(block_slot)
+            committee_size = get_committee_count_per_slot(state, block_epoch) * TARGET_COMMITTEE_SIZE
```
djrtwo, line 183: This is not the committee size.

The committee size here (the sum of all committees for the slot) is just `num_validators // SLOTS_PER_EPOCH`. 

Committees generally very between `TARGET_COMMITTEE_SIZE` and `2*TARGET_COMMITTEE_SIZE` until you hit `MAX_COMMITTEES_PER_SLOT` and then they just grow unbounded. So we won't get a proper proposer boost just using the "target"

specs/phase0/fork-choice.md
```diff
@@ -324,6 +349,10 @@ def on_tick(store: Store, time: uint64) -> None:
     store.time = time
 
     current_slot = get_current_slot(store)
+    # Reset store.proposer_score_boost if this is a new slot
+    if store.proposer_score_boost.root != Root():
```
djrtwo, line 353: It might be more readable to just do

```python
if current_slot > previous_slot:
    store.proposer_score_boost = LatestMessage(root=Root(), epoch=Epoch())
```

I think it's easier to just reason about the fact that the any proposer boost in effect is zeroed out when the slot passes. (rather than check if root is empty and if th current slot does not equal what is in the store

specs/phase0/fork-choice.md
```diff
@@ -362,6 +391,14 @@ def on_block(store: Store, signed_block: SignedBeaconBlock) -> None:
     # Add new state for this block to the store
     store.block_states[hash_tree_root(block)] = state
 
+    # Add proposer score boost if the block is timely
+    if (get_current_slot(store) == block.slot and
```
djrtwo, line 395: Generally prefer single line `if`s when possible. Easy way to achieve it is to set an intermediate bool var

```suggestion
    # Add proposer score boost if the block is timely
    is_before_attestation_broadcast = store.time % SECONDS_PER_SLOT < SECONDS_PER_SLOT // ATTESTATION_OFFSET_QUOTIENT
    if get_current_slot(store) == block.slot and is_before_attestation_broadcast:
```

specs/phase0/validator.md
```diff
@@ -446,7 +446,7 @@ def get_block_signature(state: BeaconState, block: BeaconBlock, privkey: int) ->
 
 A validator is expected to create, sign, and broadcast an attestation during each epoch. The `committee`, assigned `index`, and assigned `slot` for which the validator performs this role during an epoch are defined by `get_committee_assignment(state, epoch, validator_index)`.
 
-A validator should create and broadcast the `attestation` to the associated attestation subnet when either (a) the validator has received a valid block from the expected block proposer for the assigned `slot` or (b) one-third of the `slot` has transpired (`SECONDS_PER_SLOT / 3` seconds after the start of `slot`) -- whichever comes _first_.
+A validator should create and broadcast the `attestation` to the associated attestation subnet when either (a) the validator has received a valid block from the expected block proposer for the assigned `slot` or (b) one-third of the `slot` has transpired (`SECONDS_PER_SLOT / ATTESTATION_OFFSET_QUOTIENT` seconds after the start of `slot`) -- whichever comes _first_.
```
djrtwo, line 449: If we are swapping to this var in the validator guide, we need to do it throughout. 

At least on other reference:
* "two-thirds of the way through the slot-that is, SECONDS_PER_SLOT * 2 / 3 seconds after the start of slot"

djrtwo (commented):
nice!

A handful of formatting comments. Logic looks generally good

specs/phase0/fork-choice.md
```diff
@@ -362,6 +391,14 @@ def on_block(store: Store, signed_block: SignedBeaconBlock) -> None:
...
+       store.time % SECONDS_PER_SLOT < SECONDS_PER_SLOT // ATTESTATION_OFFSET_QUOTIENT):
```
adiasg, line 396:
```suggestion
    is_before_attestation_broadcast = store.time % SECONDS_PER_SLOT < SECONDS_PER_SLOT // ATTESTATION_OFFSET_QUOTIENT
    if get_current_slot(store) == block.slot and is_before_attestation_broadcast:
```

specs/phase0/fork-choice.md
djrtwo: done here -- https://github.com/ethereum/consensus-specs/pull/2730/commits/282d85b9e7ea3148c7bae80a11c375d3e7497ecd

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_get_head.py
```diff
@@ -261,3 +267,67 @@ def test_filtered_block_tree(spec, state):
...
+    block_1 = build_empty_block_for_next_slot(spec, state_1)
+    signed_block_1 = state_transition_and_sign_block(spec, state_1, block_1)
+
+    # Build block that serves as current head, and remains the head after block_1.slot
+    state_2 = genesis_state.copy()
+    next_slots(spec, state_2, 2)
+    block_2 = build_empty_block_for_next_slot(spec, state_2)
+    block_2.body.graffiti = spec.Bytes32(hex(random.getrandbits(8 * 32))[2:].zfill(64))
+    signed_block_2 = state_transition_and_sign_block(spec, state_2.copy(), block_2)
+    while spec.hash_tree_root(block_1) > spec.hash_tree_root(block_2):
```
djrtwo, line 302: technically should be `while >=`
but diminishingly small chance... heh 
```suggestion
    while spec.hash_tree_root(block_1) >= spec.hash_tree_root(block_2):
```

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_get_head.py
```diff
@@ -261,3 +267,67 @@ def test_filtered_block_tree(spec, state):
...
+    next_slots(spec, state_1, 3)
+    block_1 = build_empty_block_for_next_slot(spec, state_1)
+    signed_block_1 = state_transition_and_sign_block(spec, state_1, block_1)
+
+    # Build block that serves as current head, and remains the head after block_1.slot
+    state_2 = genesis_state.copy()
+    next_slots(spec, state_2, 2)
+    block_2 = build_empty_block_for_next_slot(spec, state_2)
+    block_2.body.graffiti = spec.Bytes32(hex(random.getrandbits(8 * 32))[2:].zfill(64))
+    signed_block_2 = state_transition_and_sign_block(spec, state_2.copy(), block_2)
```
djrtwo, line 301: I think this line needs to be after the while loop that mutates graffiti

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_get_head.py
```diff
@@ -261,3 +267,67 @@ def test_filtered_block_tree(spec, state):
...
+    state_1 = genesis_state.copy()
+    next_slots(spec, state_1, 3)
+    block_1 = build_empty_block_for_next_slot(spec, state_1)
+    signed_block_1 = state_transition_and_sign_block(spec, state_1, block_1)
+
+    # Build block that serves as current head, and remains the head after block_1.slot
+    state_2 = genesis_state.copy()
+    next_slots(spec, state_2, 2)
+    block_2 = build_empty_block_for_next_slot(spec, state_2)
+    block_2.body.graffiti = spec.Bytes32(hex(random.getrandbits(8 * 32))[2:].zfill(64))
```
djrtwo, line 300: Can remove this line. Empty graffiti is fine if it satisfies the `<` clause! otherwise it'll mutate in the while loop

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_get_head.py
```diff
@@ -261,3 +267,67 @@ def test_filtered_block_tree(spec, state):
...
+    # Process block_2
+    yield from tick_and_add_block(spec, store, signed_block_2, test_steps)
+    assert store.proposer_score_boost.root == spec.Root()
+    assert spec.get_head(store) == spec.hash_tree_root(block_2)
+
+    # Process block_1 on timely arrival
+    # The head should temporarily change to block_1
+    yield from tick_and_add_block(spec, store, signed_block_1, test_steps)
+    assert store.proposer_score_boost == spec.LatestMessage(root=spec.hash_tree_root(block_1),
+                                                            epoch=spec.compute_epoch_at_slot(block_1.slot))
```
djrtwo, line 319: prefer checking each field. same number of lines and reads a bit more direct

```suggestion
    assert store.proposer_score_boost.root == spec.hash_tree_root(block_1)
    assert store.proposer_score_boost.epoch == spec.compute_epoch_at_slot(block_1.slot)
```

djrtwo: Actually, now that I think about it... maybe `proposer_score_boost` should just be `proposer_boost_root`. There is zero use of `LatestMessage.epoch` and seems like a bit of an abuse of the semantics of `LatestMessage`

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
```diff
@@ -703,3 +703,29 @@ def test_new_finalized_slot_is_justified_checkpoint_ancestor(spec, state):
     assert store.justified_checkpoint == another_state.current_justified_checkpoint
 
     yield 'steps', test_steps
+
+
+@with_all_phases
+@spec_state_test
+def test_proposer_score_boost_same_slot_untimely_block(spec, state):
```
djrtwo, line 710: Suggest doing this same test but for `store.genesis_time + block.slot * spec.config.SECONDS_PER_SLOT +
                 spec.config.SECONDS_PER_SLOT // spec.INTERVALS_PER_SLOT - 1` and showing it does mutate

djrtwo (commented):
nice! minor comments. and one substnative one

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_get_head.py
adiasg: `state_transition_and_sign_block` is in the while loop as well. Need to state transition before comparing block hash.

djrtwo (approved):

djrtwo (changes requested):
flagging this as "requesting changes" so we don't merge before changing propser score boost to 70

hwwhww (commented):
I pushed commit of configuration invariant checks. (Sorry for ruining it for a couple minutes)

I have a question about the `is_before_attesting_interval` calculation.
Generally, the approach looks good to me.

hwwhww (commented):
The `on_tick` call seems incorrect for test vectors.
Proposed fix that you can review and cherry-pick: https://github.com/ethereum/consensus-specs/compare/proposer-score-boost-test-fix

specs/phase0/validator.md
```diff
@@ -446,7 +446,7 @@ def get_block_signature(state: BeaconState, block: BeaconBlock, privkey: int) ->
 
 A validator is expected to create, sign, and broadcast an attestation during each epoch. The `committee`, assigned `index`, and assigned `slot` for which the validator performs this role during an epoch are defined by `get_committee_assignment(state, epoch, validator_index)`.
 
-A validator should create and broadcast the `attestation` to the associated attestation subnet when either (a) the validator has received a valid block from the expected block proposer for the assigned `slot` or (b) one-third of the `slot` has transpired (`SECONDS_PER_SLOT / 3` seconds after the start of `slot`) -- whichever comes _first_.
+A validator should create and broadcast the `attestation` to the associated attestation subnet when either (a) the validator has received a valid block from the expected block proposer for the assigned `slot` or (b) `1 / INTERVALS_PER_SLOT` of the `slot` has transpired (`SECONDS_PER_SLOT / INTERVALS_PER_SLOT` seconds after the start of `slot`) -- whichever comes _first_.
```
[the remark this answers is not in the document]
djrtwo: note that this is division and not integer division.

I don't think this makes sense as a strict config invariant but maybe deserves a note

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
```diff
@@ -703,3 +703,78 @@ def test_new_finalized_slot_is_justified_checkpoint_ancestor(spec, state):
...
+    # Build block that serves as head ONLY on timely arrival, and ONLY in that slot
+    state = genesis_state.copy()
+    next_slots(spec, state, 3)
+    block = build_empty_block_for_next_slot(spec, state)
+    signed_block = state_transition_and_sign_block(spec, state, block)
+
+    # Process block on untimely arrival in the same slot
+    spec.on_tick(store, store.genesis_time + block.slot * spec.config.SECONDS_PER_SLOT +
+                 spec.config.SECONDS_PER_SLOT // spec.INTERVALS_PER_SLOT)
+    yield from tick_and_add_block(spec, store, signed_block, test_steps)
```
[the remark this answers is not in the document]
djrtwo: (2) is a no-op on the tick, right? because block.slot time is already in the past?

I assumed `tick_and_add_block` would tick to the greatest value in store or time_of(block.slot), but I see that it could even send time backwards.

You're right and the above fix looks good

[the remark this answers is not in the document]
hwwhww: > (2) is a no-op on the tick, right? because block.slot time is already in the past?

Yes! But using `add_block` would be more accurate in this case.

specs/phase0/validator.md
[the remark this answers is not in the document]
hwwhww: EDITED: Sorry, I meant "we use `SECONDS_PER_SLOT // INTERVALS_PER_SLOT` in code ".
^^^^^ typo fixed

> note that this is division and not integer division.

Yes. Clarify: so the value that the validator guide describes may be different from the value from the fork choice rule code when `SECONDS_PER_SLOT % INTERVALS_PER_SLOT != 0`. `INTERVALS_PER_SLOT := 3` is fine for our minimal and mainnet config.

Another option is also using integer division here: "... `SECONDS_PER_SLOT // INTERVALS_PER_SLOT` seconds after the start of `slot`..."

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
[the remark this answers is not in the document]
hwwhww: > I assumed `tick_and_add_block` would tick to the greatest value in store or time_of(block.slot), but I see that it could even send time backwards.

`tick_and_add_block` set ticks wiith `max(block_time, store.time)`, where `block_time` is time_of(block.slot)

When I reviewed it today, I was a bit surprised that we didn't check if the `on_tick` input `time` is greater than or equal to `store.time`. Should we add some fool-proof check in the spec?

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_get_head.py
```diff
@@ -103,17 +105,21 @@ def test_split_tie_breaker_no_attestations(spec, state):
...
     block_2_state = genesis_state.copy()
     block_2 = build_empty_block_for_next_slot(spec, block_2_state)
     block_2.body.graffiti = b'\x42' * 32
     signed_block_2 = state_transition_and_sign_block(spec, block_2_state, block_2)
+
+    # Tick time past slot 1 so proposer score boost does not apply
+    spec.on_tick(store, store.genesis_time + (block_2.slot + 1) * spec.config.SECONDS_PER_SLOT)
+
+    yield from tick_and_add_block(spec, store, signed_block_1, test_steps)
     yield from tick_and_add_block(spec, store, signed_block_2, test_steps)
```
hwwhww, line 123:
```suggestion
    time = store.genesis_time + (block_2.slot + 1) * spec.config.SECONDS_PER_SLOT
    on_tick_and_append_step(spec, store, time, test_steps)

    yield from add_block(spec, store, signed_block_1, test_steps)
    yield from add_block(spec, store, signed_block_2, test_steps)
```

(Not sure if you want to tick when adding `signed_block_2` or not)

tests/core/pyspec/eth2spec/test/phase0/fork_choice/test_on_block.py
```diff
@@ -703,3 +703,83 @@ def test_new_finalized_slot_is_justified_checkpoint_ancestor(spec, state):
...
+    assert store.proposer_boost_root == spec.hash_tree_root(block)
+    assert spec.get_latest_attesting_balance(store, spec.hash_tree_root(block)) > 0
+
+    # Ensure that boost is removed after slot is over
+    time = (store.genesis_time + block.slot * spec.config.SECONDS_PER_SLOT +
+            spec.config.SECONDS_PER_SLOT)
+    on_tick_and_append_step(spec, store, time, test_steps)
+    assert store.proposer_boost_root == spec.Root()
+    assert spec.get_latest_attesting_balance(store, spec.hash_tree_root(block)) == 0
+
```
hwwhww, line 757: probably worth inserting a check:
```suggestion
    test_steps.append({
        'checks': {
            'proposer_boost_root': encode_hex(store.proposer_boost_root),
        }
    })

```

need to import `from eth_utils import encode_hex`

hwwhww (approved):
one more `on_tick` usage issue. otherwise, it looks good.

specs/phase0/validator.md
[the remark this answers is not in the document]
djrtwo: ah, right. Then it is valuable to have these work well together.

djrtwo (approved):

JurbinM (approved):

PR #2731, opened 2021-11-20
added MEV-Boost-based PBS support (blinded blocks)

tkstanczak:
added blinded blocks containers to support changes described in:
https://hackmd.io/@paulhauner/H1XifIQ_t

### DIFF
--- a/specs/merge/beacon-chain.md
+++ b/specs/merge/beacon-chain.md
@@ -20,8 +20,8 @@
     - [`BeaconBlockBody`](#beaconblockbody)
     - [`BeaconState`](#beaconstate)
   - [New containers](#new-containers)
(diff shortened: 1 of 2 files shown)


paulhauner:
I don't feel that the design in https://hackmd.io/@paulhauner/H1XifIQ_t is mature enough to start merging into the spec. In fact, I'm not sure that it would ever need to be a part of the canonical consensus spec.

I had imagined this "intermediate" style of builder/proposer separation described in the hackmd would exist as a side-protocol, one that the consensus specification is generally unaware of. At some point I'd hope the specification includes a more sophisticated (and complex) system along the lines of [Proposer/block builder separation-friendly fee market designs](https://ethresear.ch/t/proposer-block-builder-separation-friendly-fee-market-designs/9725).

Issue #2724, opened 2021-11-16
EL clients require TTD to be known in advance


mkalinin:
Original discussion happened in Discord: https://discord.com/channels/595666850260713488/692062809701482577/910086648346329099
Related to https://github.com/ethereum/consensus-specs/issues/2643

The essence of the discussion -- TTD should be known in advance and should be the only trigger of consensus upgrade on EL side, details and implications:
* EL client must know TTD in advance to have a clear boundary at which the consensus upgrade happens. Handling the case when sister blocks have different consensus rules adds a lot of complexity to the fork machinery of EL clients.
* TTD must be the only trigger of the consensus upgrade because adding another brings the same complexity issue as the previous case. It means that `TERMINAL_BLOCK_HASH` can't be used as a trigger for consensus upgrade on EL side. It means that TBH functionality may only be implemented by block whitelisting via its hash value + TTD override. EL clients currently support the whitelisting feature.

### Option 1
Remove TTD and TBH overrides and rely on "release/update" procedure in case if there is a hashpower migration that significantly delays the Merge (the TTD override case) and in case of 51% attacks (the TBH case). Cons: delays the emergency reaction by X hours -- hours required for the network to upgrade their clients.

### Option 2
Have override settings in CL and EL clients. Cons: bricks the client in case when user restarts only CL or EL client with a new setting, i.e. overrides go inconsistent. Potentially, EL may respond with specific errors to `executePayload` and `forkchoiceUpdated` if they reference a block that isn't a valid terminal block. It would aid figuring out the issue and restarting a node with correct settings.

### Option 3
Have EL responsible for handling terminal block functionality and have override settings on EL only. TBH functionality can't be handled entirely by EL as TBH also comes with an epoch number which coordinates the transition on CL side. So, this option seems not viable as it lefts an issue with consistency of overrides between the layers.

mkalinin:
With the assumption that upgrading binary distributions takes near the same time as restarting clients with additional CLI arguments, I am inclined to have the following approach:
* Remove overrides via CLI args entirely and rely on overrides via releasing/upgrading binary distributions
* Add the corresponding errors to Engine API, i.e. respond with error if `executePayload` or `forkchoiceUpdated` deviates from TTD hardcoded on EL side

Depending on the case, with this approach the Merge acceleration process will look as follows:
* Hashpower reduction causing significant Merge delay
    * Release CL and EL binaries with the new `TTD` value that is lower than the previous one
* Attacks on the network
[...]

PR #2741, opened 2021-11-24
Remove client-settings.md

mkalinin:
Releasing client software with updated parameters and giving a time for users to upgrade their nodes is going to be default strategy to accelerate the Merge in the case of emergency, see the comment [here](https://github.com/ethereum/consensus-specs/issues/2724#issuecomment-974230490) for details. 

Therefore, CLI args doesn't increase the speed of the reaction in emergency cases hence supporting them is anymore not a requisite for a client software. This PR proposes to remove the corresponding document from the set of consensus spec docs.

cc @MicahZoltu @karalabe


djrtwo (approved):
Looks good.

Note to readers that `TERMINAL_TOTAL_DIFFUCLTY`, `TERMINAL_BLOCK_HASH`, and `TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH` are still defined but must be overridden in updated configs in releases rather than on the command line

djrtwo:
leaving this up for 24 hours before merging in case other comments

MicahZoltu (approved):
From a specification standpoint, I don't think this should be specified so I think removal of this is a good thing.  From a recommendation standpoint for the various client I think there is notable value in providing a CLI override flag to their users, especially enterprise users that may have a fairly onerous upgrade process but a fairly simple configuration process.

PR #2746, opened 2021-11-26
Simplify sync protocol and update to calculate optimistic heads

vbuterin:
1. Simplify `valid_updates` to `best_valid_update` so the `LightClientStore` only needs to store O(1) data
2. Track an optimistic head, by looking for the highest-slot header which passes a safety threshold


dapplion (approved):
Looks great :+1:

specs/altair/sync-protocol.md
```diff
@@ -101,50 +102,80 @@ def get_subtree_index(generalized_index: GeneralizedIndex) -> uint64:
     return uint64(generalized_index % 2**(floorlog2(generalized_index)))
 ```
 
+### `get_active_header`
+
+```python
+def get_active_header(update: LightClientUpdate) -> BeaconBlockHeader:
+    # Is the update trying to convince us of a finalized header or an optimistic header?
+    if update.finalized_header BeaconBlockHeader():
```
terencechain, line 110:
```suggestion
    if update.finalized_header != BeaconBlockHeader():
```

specs/altair/sync-protocol.md
```diff
@@ -101,50 +102,80 @@ def get_subtree_index(generalized_index: GeneralizedIndex) -> uint64:
...
-def validate_light_client_update(snapshot: LightClientSnapshot,
+def validate_light_client_update(store: LightClientStore,
                                  update: LightClientUpdate,
                                  genesis_validators_root: Root) -> None:
-    # Verify update slot is larger than snapshot slot
-    assert update.header.slot > snapshot.header.slot
+                                 
+    # Verify update slot is larger than slot of current best finalized header
+    active_header = get_active_header(update)
+    assert active_header.slot > store.finalized_header.slot
```
etan-status, line 148: What if `active_header.slot` happens to be in the future of the caller's `current_slot`?

specs/altair/sync-protocol.md
```diff
@@ -47,34 +49,24 @@ uses sync committees introduced in [this beacon chain extension](./beacon-chain.
 
 ### Misc
 
-| Name | Value |
-| - | - |
-| `MIN_SYNC_COMMITTEE_PARTICIPANTS` | `1` |
+| Name | Value | Notes |
+| - | - | - |
+| `MIN_SYNC_COMMITTEE_PARTICIPANTS` | `1` | |
+| `SAFETY_THRESHOLD_CALCULATION_PERIOD` | `4096` | ~13.6 hours |
```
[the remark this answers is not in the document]
vbuterin: I don't really have a very principled way to choose the `SAFETY_THRESHOLD_CALCULATION_PERIOD` yet. As far as I can tell, it's a responsiveness/vulnerability tradeoff. A `SAFETY_THRESHOLD_CALCULATION_PERIOD` of eg. 1 epoch would mean that if the chain suddenly loses >50% of participants, light clients will only experience a 2 epoch delay, but this means that an attacker need only eclipse a client for 2 epochs to convince them of anything. Setting `SAFETY_THRESHOLD_CALCULATION_PERIOD = UPDATE_TIMEOUT` (~1 day) pushes safety to the maximum, but at the cost of minimum adaptability.

Though one path we _could_ take is to set `SAFETY_THRESHOLD_CALCULATION_PERIOD = UPDATE_TIMEOUT` and then just assert that any desired faster responsiveness should come from clients implementing custom logic in the safety factor function (eg. `max // 2` normally but `max // 4` after two epochs of the optimistic head not updating). I'm open to any option here.

specs/altair/sync-protocol.md
```diff
@@ -47,34 +49,25 @@ uses sync committees introduced in [this beacon chain extension](./beacon-chain.
...
 ### Misc
 
-| Name | Value |
-| - | - |
-| `MIN_SYNC_COMMITTEE_PARTICIPANTS` | `1` |
+| Name | Value | Notes |
+| - | - | - |
+| `MIN_SYNC_COMMITTEE_PARTICIPANTS` | `1` | |
+| `SAFETY_THRESHOLD_PERIOD` | `SLOTS_PER_EPOCH * EPOCHS_PER_SYNC_COMMITTEE_PERIOD` | ~13.6 hours |
+| `UPDATE_TIMEOUT` | `SLOTS_PER_EPOCH * EPOCHS_PER_SYNC_COMMITTEE_PERIOD` | ~27.3 hours |
```
hwwhww, line 56: I think one of these is wrong?
`SLOTS_PER_EPOCH * EPOCHS_PER_SYNC_COMMITTEE_PERIOD = 32 * 256 = 8192` slots = ~27.3 hours.
Did you mean having `SAFETY_THRESHOLD_PERIOD = UPDATE_TIMEOUT / 2`?

hwwhww: Addressed in #2750

specs/altair/sync-protocol.md
```diff
@@ -153,43 +188,60 @@ def validate_light_client_update(snapshot: LightClientSnapshot,
...
+    validate_light_client_update(store, update, current_slot, genesis_validators_root)
+    
+    # Update the best update in case we have to force-update to it if the timeout elapses
+    if sum(update.sync_committee_bits) > sum(store.best_valid_update.sync_committee_bits):
+        store.best_valid_update = update
+    
+    # Track the maximum numebr of active participants in the committee signatures
+    store.current_max_active_participants = max(
+         store.current_max_active_participants,
+         update.sync_committee_bits.count(1)
```
etan-status, line 225: NIT: `sum(update.sync_committee_bits)` for consistency

specs/altair/sync-protocol.md
```diff
@@ -101,50 +104,83 @@ def get_subtree_index(generalized_index: GeneralizedIndex) -> uint64:
...
-A light client maintains its state in a `store` object of type `LightClientStore` and receives `update` objects of type `LightClientUpdate`. Every `update` triggers `process_light_client_update(store, update, current_slot)` where `current_slot` is the current slot based on some local clock.
+A light client maintains its state in a `store` object of type `LightClientStore` and receives `update` objects of type `LightClientUpdate`. Every `update` triggers `process_light_client_update(store, update, current_slot)` where `current_slot` is the current slot based on some local clock. `process_slot_for_light_client_store` is processed every time the current slot increments.
+
+#### `process_slot_for_light_client_store`
+
+```python
+def process_slot_for_light_client_store(store: LightClientStore, current_slot: Slot) -> None:
+    if current_slot % SAFETY_THRESHOLD_PERIOD == 0:
+        store.previous_max_active_participants = store.current_max_active_participants
+        store.current_max_active_participants = 0
```
etan-status, line 140: Should `apply_light_client_update` also be triggered, in case `current_slot > store.finalized_header.slot + UPDATE_TIMEOUT` gets fulfilled?

specs/altair/sync-protocol.md
```diff
@@ -153,43 +189,66 @@ def validate_light_client_update(snapshot: LightClientSnapshot,
...
         sum(update.sync_committee_bits) * 3 >= len(update.sync_committee_bits) * 2
-        and update.finality_header != BeaconBlockHeader()
+        and update.finalized_header != BeaconBlockHeader()
+    ):
+        # Normal update through 2/3 threshold
+        apply_light_client_update(store, update)
+        store.best_valid_update = None
+    elif (
+        current_slot > store.finalized_header.slot + UPDATE_TIMEOUT
+        and store.best_valid_update is not None
```
etan-status, line 249: `store.best_valid_update` cannot be `None` here (in that case, it is initialized with `update` above)

specs/altair/sync-protocol.md
etan-status: Also, the condition to update `optimistic_header` may also be fulfilled after changes to `*_max_active_participants`.

specs/altair/sync-protocol.md
```diff
@@ -47,34 +49,25 @@ uses sync committees introduced in [this beacon chain extension](./beacon-chain.
...
+    # Next sync committee corresponding to the active header
     next_sync_committee: SyncCommittee
     next_sync_committee_branch: Vector[Bytes32, floorlog2(NEXT_SYNC_COMMITTEE_INDEX)]
-    # Finality proof for the update header
-    finality_header: BeaconBlockHeader
+    # The finalized beacon block header attested to by Merkle branch
+    finalized_header: BeaconBlockHeader
     finality_branch: Vector[Bytes32, floorlog2(FINALIZED_ROOT_INDEX)]
     # Sync committee aggregate signature
     sync_committee_bits: Bitvector[SYNC_COMMITTEE_SIZE]
```
dapplion, line 73: I would compact this into sync_aggregate like in https://github.com/ethereum/consensus-specs/pull/2267/files#diff-e1b1ce81657edc3d8874cfec98e2f7d2f8d0c478cc8cf0998dd411b6da001ecbR116 makes the implementation simpler

specs/altair/sync-protocol.md
```diff
@@ -153,43 +189,66 @@ def validate_light_client_update(snapshot: LightClientSnapshot,
...
-        snapshot.next_sync_committee = update.next_sync_committee
-    snapshot.header = update.header
+def apply_light_client_update(store: LightClientStore, update: LightClientUpdate) -> None:
+    active_header = get_active_header(update)
+    finalized_period = compute_epoch_at_slot(store.finalized_header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+    update_period = compute_epoch_at_slot(active_header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+    if update_period == finalized_period + 1:
+        store.current_sync_committee = store.next_sync_committee
+        store.next_sync_committee = update.next_sync_committee
+    store.finalized_header = active_header
```
etan-status, line 206: If the `optimistic_header` was older, I guess it should also be updated here (to `finalized_header`).

specs/altair/sync-protocol.md
```diff
@@ -101,50 +104,83 @@ def get_subtree_index(generalized_index: GeneralizedIndex) -> uint64:
...
 
     # Verify update does not skip a sync committee period
-    snapshot_period = compute_epoch_at_slot(snapshot.header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
-    update_period = compute_epoch_at_slot(update.header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
-    assert update_period in (snapshot_period, snapshot_period + 1)
+    finalized_period = compute_epoch_at_slot(store.finalized_header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+    update_period = compute_epoch_at_slot(active_header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+    assert update_period in (finalized_period, finalized_period + 1)
 
     # Verify update header root is the finalized root of the finality header, if specified
```
jinfwhuang, line 160: This is no longer true. It would cause confusion.

Suggested:
# Verify that the `finality_header` is the expansion of `update.attested_header.state_root.finalized_checkpoint.root`

specs/altair/sync-protocol.md
```diff
@@ -47,34 +49,24 @@ uses sync committees introduced in [this beacon chain extension](./beacon-chain.
...
-```
-
 ### `LightClientUpdate`
 
 ```python
 class LightClientUpdate(Container):
-    # Update beacon block header
-    header: BeaconBlockHeader
+    # The beacon block header that is attested to by the sync committee
+    attested_header: BeaconBlockHeader
```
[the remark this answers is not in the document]
jinfwhuang: I second this comment. We should clarify that if `update.finality_header` is specified, `next_sync_committee` comes from the beacon-state corresponding with `update.finality_header`. If not specified, the `next_sync_committee` comes from the beacon-state corresponding with `update.attested_header`. This is not obvious.

specs/altair/sync-protocol.md
jinfwhuang, line 206: Does this mean that it could be the case the store.finalized_header is not actually a finalized header, when the apply_light_client_update is called through update timeout?

specs/altair/sync-protocol.md
```diff
@@ -101,50 +104,83 @@ def get_subtree_index(generalized_index: GeneralizedIndex) -> uint64:
...
+                                 
+    # Verify update slot is larger than slot of current best finalized header
+    active_header = get_active_header(update)
+    assert current_slot >= active_header.slot > store.finalized_header.slot
 
     # Verify update does not skip a sync committee period
-    snapshot_period = compute_epoch_at_slot(snapshot.header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
-    update_period = compute_epoch_at_slot(update.header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
-    assert update_period in (snapshot_period, snapshot_period + 1)
+    finalized_period = compute_epoch_at_slot(store.finalized_header.slot) // EPOCHS_PER_SYNC_COMMITTEE_PERIOD
```
jinfwhuang, line 156: Does that mean the light-client should fail if there is a skip period? This seems to be a fairly normal path when a client stops running for a few das.

jinfwhuang:
I will make one big picture comment that is slightly out of scope for this PR. It is relevant because this PR attempts to fix the sync logic when there is a client timeout. It is even more relevant if a light client has to get updates that skips a period.

As it stand right now, the spec does not provide a mechanism to skip-sync for a light-client that has an outdated view of the sync-committees. Without an explicit skip-sync mechanism, it is hard to address the cold start problem.

- Here is a spec that writes out skip-sync data objects on the portal network: https://github.com/ethereum/portal-network-specs/blob/master/beacon-chain/skip-sync-network.md
- Here is a light-client POC that implements a skip-sync mechanism: https://github.com/jinfwhuang/prysm/pull/5

specs/altair/sync-protocol.md
```diff
@@ -101,50 +104,83 @@ def get_subtree_index(generalized_index: GeneralizedIndex) -> uint64:
...
+        return update.attested_header
+```
+
+### `get_safety_threshold`
+
+```python
+def get_safety_threshold(store: LightClientStore) -> uint64:
+    return max(
+        store.previous_max_active_participants,     
+        store.current_max_active_participants
```
jinfwhuang, line 126: Is there a reason why the threshold is half of the max(previous, current)? This is just a heuristic check, correct? Can we add a note stating as such?

specs/altair/sync-protocol.md
jinfwhuang:
```
    sync_committee_signature: BLSSignature
```

etan-status: Wouldn't this make the serialized size of `LightClientUpdate` larger due to the extra nesting? I thought this was all about minimizing bytes. Granted, the change is minimal.

specs/altair/sync-protocol.md
etan-status: This was the case in the old version as well, but it was called just `header` there. `finalized_header` here seems to have a different meaning than in other contexts, it's just `finalized` for the light client (it won't revert it anymore). Agree that the naming is suboptimal. Likewise, the `optimistic_header` also seems to have a different meaning from the one discussed as part of the merge effort.

specs/altair/sync-protocol.md
etan-status: The light-client can request historic `LightClientUpdate` from the network. It needs at least one update per period to follow along, as it only knows `current_sync_committee` and `next_sync_committee` and can only verify `LightClientUpdate` from those periods.

However, what is still suboptimal is the case where `finalized_update` is in a different period than `attested_update`, but this is not a problem introduced by this PR. Another tricky case to tackle for the case where `attested_update` is in a different period than the committee which signed it, which probably even requires some heuristics to figure out (as this case depends on there being missed slots at the start of an epoch). For now, these edge cases are all ignored, and the updates are only accepted if all of `finalized_update`, `attested_update`, and the sync committee signing it come from the same sync committee period.

specs/altair/sync-protocol.md
etan-status: Related comment:
https://github.com/ethereum/consensus-specs/pull/2746#discussion_r758348694

specs/altair/sync-protocol.md
dapplion: @etan-status in ssz serialized form the size is the same, is json form it's negligibly larger

specs/altair/sync-protocol.md
jinfwhuang: hmm...if this is the intended "finalization" for the light-client, that is not great.

In the case of timeout, why not just go to the network and ask for a committee changing update? I know that in this spec, we have not specify how to get that information. In any implementation, the light client is going to have to be able to ask for historic updates corresponding to some "sync-committee". If that is available, the finalization of just taking the `store.best_valid_update` is not great. I doubt that real client implementation is going to take this route.

specs/altair/sync-protocol.md
jinfwhuang: Yes. agree on the updating path will trace sync-comm linkage. Also agree that this is not the issue raised by this PR.

With regard to the edge case...it could cause  some weird behaviors temporarily. For example, `apply_light_client_update` is called due to timeout. Then, when there are valid `update.finalized_header` arrives, they will get rejected.

Again, this behavior could be better handled if we assume that the light-client can make request for specific LightClientUpdates when it times out or fall out of sync with the current stream of updates. The sync logic would becomes a lot cleaner to be separated into two sync mode: skip-sync mode and normal sync mode.

specs/altair/sync-protocol.md
etan-status: If sync committee participation is low, and none of the blocks exceeds the 2/3 majority for a day, there still needs to be a way to proceed though. Not sure how realistic that is for mainnet.

jinfwhuang: I think that is fine. If that indeed happens once in blue moon, the light client would stop working syncing. The manual fix for light client operator is to use a newly acquired, trusted starting point. The code owner could also update their client's hard coded starting point. In a way, these manual interventions should be considered desirable because we have unexpected level of participation.

However, if that happens a lot, I think that is more of an incentive design issue. We should consider how to fix that at the protocol level.

specs/altair/sync-protocol.md
jinfwhuang: got it.

specs/altair/sync-protocol.md
vbuterin: Light clients are intended to be able to follow the chain in as similar a way to regular clients as possible. And one of the ethereum staking protocol's core design goals has all along been to have _some_ path to be able to continue making progress under >1/3 offline conditions. So the light client protocol should include _some_ way to do that.

(I'm assuming light clients are going to be used in a lot of contexts, including automated ones, where manual intervention is hard and should be left to resolving 51% attacks)

What is a better alternative to taking `store.best_valid_update`? The regular ethereum protocol advances during the non-finalization case by using the LMD GHOST fork choice rule, which follows the chain that has the most validators supporting it. `store.best_valid_update` approximates that. Is there a better approximation?

jinfwhuang: I would suggest that the light-client could just leave the decision to the "data source", which are invariably backed by full nodes. The exact meaning of data source is not well defined yet because the networking layer could be portal-network, a LES-like p2p network, or a server-client RPC pairing.

When the light-client experiences a timeout, i.e. the incoming Updates are not good enough to advance its finalized_header, the client would revert to a skip-sync mode. In a skip-sync mode, the client asks the "data source" for an update that would advance its sync-committee. A light client does not advance until it somehow find a way to access "finality". Because finality is guaranteed to be found in some data sources, a light client is stuck because it couldn't access the correct data sources. 

The guarantee of a light client would find a way to advance should depends on a light client having a way to find the right updates. Again, networking is not defined yet.

dapplion (approved):
This PR accomplishes the scope described in it's body successfully :+1: 

Other topics brought up in this PR description should be tackled in new PRs:
- p2p networking
- improved data structures

hwwhww (approved):
Well done in simplification. 👍

Agreed with @dapplion. FYI I'd like to move the file paths with other PRs. Let's merge this PR now and then propose suggestions & add other designs with other PRs.

specs/altair/sync-protocol.md
mimoo: sorry to bump but I got really confused reading https://github.com/ethereum/annotated-spec/blob/master/altair/sync-protocol.md which says:

image

I guess the meaning of "finalized" changed since then?

etan-status: Yes, this got simplified and the signature is now always over `update.attested_header`, which was historically named `finality_header` if `header` was finalized.

PR #2753, opened 2021-11-30
Simplify logic in calculating proposer score

realbigsean:
The call to `get_ancestor` when checking if the proposer score should be calculated in `get_latest_attesting_balance` seems to be unnecessary so I removed it here.

### DIFF
--- a/specs/phase0/fork-choice.md
+++ b/specs/phase0/fork-choice.md
@@ -182,14 +182,12 @@ def get_latest_attesting_balance(store: Store, root: Root) -> Gwei:
             and get_ancestor(store, store.latest_messages[i].root, store.blocks[root].slot) == root)
     ))
     proposer_score = Gwei(0)
-    if store.proposer_boost_root != Root():
-        block = store.blocks[root]
-        if get_ancestor(store, root, block.slot) == store.proposer_boost_root:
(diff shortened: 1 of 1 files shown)


djrtwo (approved):

specs/phase0/fork-choice.md
```diff
@@ -182,14 +182,12 @@ def get_latest_attesting_balance(store: Store, root: Root) -> Gwei:
             and get_ancestor(store, store.latest_messages[i].root, store.blocks[root].slot) == root)
     ))
     proposer_score = Gwei(0)
-    if store.proposer_boost_root != Root():
-        block = store.blocks[root]
-        if get_ancestor(store, root, block.slot) == store.proposer_boost_root:
```
adiasg, line 187: This line should have been
```python
if get_ancestor(store, store.proposer_boost_root, store.blocks[root].slot) == root:
```

adiasg:
While the simplifying changes in this PR were sound for the underlying logic in the spec, the underlying logic itself was incorrect (see [above comment](https://github.com/ethereum/consensus-specs/pull/2753#discussion_r760710249)).
New changes in #2760 remove the simplification & fix the spec logic (as pointed out in #2757). 

Thanks @realbigsean for both this PR and #2757!

PR #2754, opened 2021-11-30
Correct primitive root

dankrad:
5 is not a primitive root of the BLS field, it needs to be 7


djrtwo (approved):

ralexstokes:
for posterity's sake, @dankrad shared this script w/ me to validate this number:

https://gist.github.com/ralexstokes/f1d2b6557dd573ae496a752b7ab4e563

PR #2752, opened 2021-11-30
Add ex-ante fork choice test cases

hwwhww:
Thank @casparschwa @djrtwo for the testing plan: https://notes.ethereum.org/m1upI_XcRpSskeG4UvgNtA?view

Differences:
- I combined Scenario 1 and Scenario 2 since the only difference is Step 4.
- [Adversarial attestations > proposer boost] I just use full attestation instead of exact `proposer_boost + 1` votes.
 
It's still verbose. Will think about refactoring tomorrow.

TODOs:
- [x] `spec_configured_state_test` probably is not thread-safe and made `test_proposer_boost_correct_head` failed in CI. Need to check it.
    - sub-PR: #2755, #2765
- [x] Update test generator
    - Either dump `PROPOSER_SCORE_BOOST` to `meta.yaml` or try to dump the full changed config. I'm inclined to the former now.
- ~~Update test format~~
- ~~[DRY] Clean up the test script.~~
- [x] Add ex-ante "sandwich" cases. 
    - sub-PR: #2766

PR #2755, opened 2021-12-01
import `Spec` objects for each test case

hwwhww:
### Issue

1. `spec_configured_state_test` is not thread-safe:
    - The given `spec: Spec` object was initialized with global `spec_targets` in `context.py`.
    -  `spec_configured_state_test` overrides `spec.config`, execute the test case, and then change it back to the unchanged config.
    - In multi-processing or multi-thread context, if a config field is overridden when running one test case, the `spec.config` object in other concurrent tasks might be changed too.
2. pytest-xdist implemented parallelization with subprocess.
3. We want to make the test generator parallelable!

Due to 1+2, #2752 failed because some of the ex-ante tests override `spec.config.PROPOSER_SCORE_BOOST` temporarily, and then concurrently, `test_proposer_boost_correct_head` failed because it wrongly used the overridden value.

### How did I fix it

To better isolate the test cases, it would be better to reduce the risk of these side effects. This PR tries to import `Spec` objects for *each* test case. The solution was found from [here](https://stackoverflow.com/a/11285504) and [here](https://docs.python.org/3/library/importlib.html#implementing-lazy-imports).

### DIFF
--- a/tests/core/pyspec/eth2spec/test/context.py
+++ b/tests/core/pyspec/eth2spec/test/context.py
@@ -2,19 +2,17 @@ import pytest
 from copy import deepcopy
 from dataclasses import dataclass
 from eth_utils import encode_hex
+import importlib
 
-from eth2spec.phase0 import mainnet as spec_phase0_mainnet, minimal as spec_phase0_minimal
-from eth2spec.altair import mainnet as spec_altair_mainnet, minimal as spec_altair_minimal
-from eth2spec.merge import mainnet as spec_merge_mainnet, minimal as spec_merge_minimal
 from eth2spec.utils import bls
 
 from .exceptions import SkippedTest
 from .helpers.constants import (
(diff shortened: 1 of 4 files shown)


tests/core/pyspec/eth2spec/test/context.py
```diff
@@ -360,11 +344,31 @@ def decorator(fn):
     return decorator
 
 
+def _get_module(module_path):
+    module_spec = importlib.util.find_spec(module_path)
+    module = importlib.util.module_from_spec(module_spec)
+    loader = importlib.util.LazyLoader(module_spec.loader)
+    module_spec.loader = loader
+    import sys
```
ralexstokes, line 352: nitpick: hoist `import sys` to the top of the file

hwwhww:
replaced by #2765

PR #2756, opened 2021-12-01
Name generalized index constants as _GINDEX

dapplion:
Constants in `specs/altair/sync-protocol.md` that represent a generalized index are named _INDEX which may be confused with a regular array or field index. Renaming to _GINDEX clarifies this distinction.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -475,8 +475,8 @@ def get_generalized_index(ssz_class: Any, *path: Sequence[PyUnion[int, SSZVariab
     @classmethod
     def hardcoded_ssz_dep_constants(cls) -> Dict[str, str]:
         constants = {
-            'FINALIZED_ROOT_INDEX': 'GeneralizedIndex(105)',
-            'NEXT_SYNC_COMMITTEE_INDEX': 'GeneralizedIndex(55)',
+            'FINALIZED_ROOT_GINDEX': 'GeneralizedIndex(105)',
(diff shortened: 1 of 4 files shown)


ralexstokes:
i think if we want to change this name it would something more like `FOO_GENERALIZED_INDEX` as it is not immediately clear what a "gindex" is whereas an "index" (of some type) leverages a lot of prior knowledge a reader can be expected to have.

the downside to `FOO_GENERALIZED_INDEX` is that it is a "mouthful" and long to type -- and that is where IMO just saying `FOO_INDEX` is a nice compromise.

so i only weakly support this change here. and if you feel strongly enough about this to merge it in then we will want to update the pyspec machinery so all of the CI passes :)

dapplion:
> i think if we want to change this name it would something more like `FOO_GENERALIZED_INDEX` as it is not immediately clear what a "gindex" is whereas an "index" (of some type) leverages a lot of prior knowledge a reader can be expected to have.
> 
> the downside to `FOO_GENERALIZED_INDEX` is that it is a "mouthful" and long to type -- and that is where IMO just saying `FOO_INDEX` is a nice compromise.
> 
> so i only weakly support this change here. and if you feel strongly enough about this to merge it in then we will want to update the pyspec machinery so all of the CI passes :)

We can add a note somewhere that GINDEX = GENERALIZED_INDEX for new readers. For implementers, gindex is a common term used in the implementations I've looked at.

dapplion:
We actually had a Electra light-client bug because of this 🤣 closing for now as the ship has sailed

Issue #2758, opened 2021-12-01
Withdrawals tracking issue


djrtwo:
Withdrawals (code name Capella) will be specified as three features:

1. [x] Withdrawable 0x01 validators -> withdrawn (at epoch boundaries) and core withdrawal mechanism -- #2836 
2. [x] 0x00 -> 0x01 withdrawal credentials change operation -- #2855 
3. [x] Partial withdrawal on amounts above 32 eth for 0x01 validators -- #2862 
4. [x] consider removal of `withdrawn_epoch` added in (1) #2998
     *  just utilize withdrawabl_epoch and zero balance to know if safe to re-use
6. [x] consider deposits into withdrawn validators #2998

djrtwo:
My scratchpad notes I'm working off of
https://notes.ethereum.org/BHG6JUibQ8CNqpUeVHd1dw?edit

hwwhww:
@djrtwo  Should we add a task item of creating an EIP to handle `WithdrawalReceipt`?

djrtwo:
yes

arnetheduck:
Apart from the economic stuff, do we want to do something about the size of the validator list in the beacon state before pulling the trigger on this one? Those lists are becoming quite nasty to manage, specially with the multiple linear scans (quadratic if you do it the pure spec way, but linear with the right optimizations) that we do regularly.

Jiggnubs:
No, keep it coming.   This bitcoin thingy is the greatest current to ever
happen.  When everyone is super, no one is creator, and everyone is just
graceful.  LoL be the beta !

djrtwo:
@arnetheduck do you mean re-using indices?
The way this is spec'd allows for re-use but does not enable re-use

Or do you mean something else entirely?

arnetheduck:
> @arnetheduck do you mean re-using indices?

yes indeed - re-using, making the list sparse or any other approach.

mcdee:
Is it worth attempting to optimize here before withdrawals are implemented, seems a little premature as we don't know how many exited validators there will be.  Currently there are 869 out of 431,496 validators exited, or a little over 0.2%.  I know that is the lower bound, but until we have some sort of idea of the real-world number we'd be doing a fair bit of work for what could be minimal gains.

Once capella goes live we should have some idea of the churn, which would help to inform the best way to approach reduction in size of the validator list.

Where-2:
Lost

Get Outlook for iOS<https://aka.ms/o0ukef>
[...]

PR #2759, opened 2021-12-01
[WIP] Withdrawals pull

djrtwo:
First of the withdrawal features to be specified. Feature metatracking issue here -- #2758

*Fork has code-name -- [Capella](https://en.wikipedia.org/wiki/Capella). I'm assuming Merge will have a B-star-name so putting a C-star-name placeholder for now. "Withdrawals" is too confusing of a placeholder name.*

During the epoch transition, if a validator has 0x01 withdrawal credentials and is withdrawable, fully withdraw the validator -- i.e. create a withdrawal receipt and add it to the state, and zero out the validator balance

Some details to note:
* A commitment to `withdrawal_receipts` (e.g. the root) to be exposed in EL for normal EL TX to consume to move the ETH into the end account
* `index` (always increasing) included in `WithdrawalReceipt` so that a simple bitfield (of length `max_successful_withdrawal_receipt_index`) can be used to track consumed receipts in EL
* A 0x00 -> 0x01 change-of-credentials operation is required to trigger a withdrawal for existing 0x00 validators (new operation to come in subsequent PR)
* Partial withdrawals will use the saw `withdraw` function but not use entire balance (to be developed in subsequent PR)

todo:
* [x] add a `withdrawn_epoch` to `Validator`. This will be useful for any sort of validator index reuse (e.g. safe to be reused after 50k epochs)
* [ ] don't use the just a simple SSZ list for the accumulator. With partial withdrawals this can grow at ~100MB / year. It's append only so you can send most receipts to disk but this means the WS states cant grow to cumbersome sizes.
    * main idea -- create a  [CommitmentContainer](https://github.com/ethereum/consensus-specs/pull/2585) that just has the merkle root and right-most merkle branch that works for append-only trees (and has some merkle tree structure as a List)
    * backup idea -- use ring-buffer
        * simple
        * but puts a time limit on how long you have to claim in EL because receipts get overwritten
* [x] Add some epoch transition sanity tests


zilm13:
I don't know if this is the right issue for it, but we should make some estimations on:
- load on queues and `BeaconChain` just after withdrawal feature introduction
- regular `BeaconChain` and queues load if feature "Partial withdrawal" is introduced

specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,163 @@
...
+### Predicates
+
+#### `is_withdrawable_validator`
+
+```python
+def is_withdrawable_validator(validator: Validator, epoch: Epoch) -> bool:
+    """
+    Check if ``validator`` is withdrawable.
+    """
+    return validator.withdrawable_epoch <= epoch
```
mkalinin, line 126:
```suggestion
    return validator.withdrawable_epoch >= epoch
```

specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,163 @@
...
+    process_randao_mixes_reset(state)
+    process_historical_roots_update(state)
+    process_participation_flag_updates(state)
+    process_sync_committee_updates(state)
+    process_withdrawals(state)  # [New in Cappela]
+```
+
+#### Withdrawals
+
+*Note*: The function `process_inactivity_updates` is new.
```
mkalinin, line 152:
```suggestion
*Note*: The function `process_withdrawals` is new.
```

specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,163 @@
...
+
+```python
+def process_withdrawals(state: BeaconState) -> None:
+    current_epoch = get_current_epoch(state)
+    for index, validator in enumerate(state.validators):
+        balance = state.balances[index]
+        is_balance_nonzero = state.balances[index] == 0
+        is_eth1_withdrawal_prefix = validator.withdrawal_credentials[0] != ETH1_ADDRESS_WITHDRAWAL_PREFIX
```
mkalinin, line 160:
```suggestion
        is_eth1_withdrawal_prefix = validator.withdrawal_credentials[0] == ETH1_ADDRESS_WITHDRAWAL_PREFIX
```

specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,163 @@
...
+        if is_balance_nonzero and is_eth1_withdrawal_prefix and is_withdrawable_validator(validator, current_epoch):
```
mkalinin, line 161: Probably, it does make sense to move all these conditions to `is_withdrawable_validator` function

mkalinin (commented):
An option with a ring-buffer plus `historical_roots` accumulator also worth considering, IMO

specs/capella/beacon-chain.md
djrtwo: I believe this should be `<=`. If the withdrawable epoch is at or in the past (less than) the `epoch` in question, then the validator is withdrawable.

Look at the `is_active_validator` logic for a parallel

djrtwo: Take this extreme as example -- if withdrawable_epoch is `FAR_FUTURE_EPOCH`, then they will never be withdrawable. but with your suggestion, the > sign would def make them withdrawable

specs/capella/beacon-chain.md
djrtwo: agreed!

setup.py
```diff
@@ -548,9 +549,38 @@ def hardcoded_custom_type_dep_constants(cls) -> str:
...
+    def sundry_functions(cls) -> str:
+        return super().sundry_functions()
+
+    @classmethod
+    def hardcoded_ssz_dep_constants(cls) -> Dict[str, str]:
+        constants = {
+            'FINALIZED_ROOT_INDEX': 'GeneralizedIndex(105)',
+            'NEXT_SYNC_COMMITTEE_INDEX': 'GeneralizedIndex(55)',
+        }
+        return {**super().hardcoded_ssz_dep_constants(), **constants}
```
hwwhww, line 578: No needs to implement these methods if there are no new settings. It will just use `MergeSpecBuilder` methods. 


```suggestion
```

.circleci/config.yml
```diff
@@ -129,6 +129,20 @@ jobs:
           command: make citest fork=merge
       - store_test_results:
           path: tests/core/pyspec/test-reports
+  test-capella:
```
hwwhww, line 132: this new CI job has to be added to the "workflows".

hwwhww (commented):
Great work!

I haven't started to review the content yet. Just want to remind you of the CI setting earlier.

setup.py
djrtwo: I was getting some build errors without it so added them. But will try again

.circleci/config.yml
djrtwo: got it!

setup.py
djrtwo: annoying. I tried again and it worked!
thanks!

specs/capella/beacon-chain.md
mkalinin: You're right. This is a bit counter-intuitive to read hence my confusion.

mkalinin:
> Right @mkalinin, I was also considering a double batched merkle accumulator (like historic roots). One reason I would like to try this other method first (commitmnt container) is that the claim process/logic will likely be much simpler because it will just look like a merkle tree, constantly growing. Additionally, it's not immediately clear when to do the batching, but that can be solved relatively easily.

Actually, user may utilize historical batches to build a persistent proof with the merkle tree solution:
1. Get a proof linking receipt to the withdrawal receipts tree root when the receipt is added to the tree
2. Get a proof linking withdrawal receipts tree root from (1) to the historical batch once the root of the state that contains the receipts tree root from (1) is accumulated
3. Get a proof linking the historical batch root from (2) to the beacon/state root and claim the receipt on EL side at any time

hwwhww (commented):
some nitpicking.

the logic looks good to me with bare-eye review atm. need to add more test cases to verify it.

mcdee:
Regarding partial withdrawals: there has been a fair amount of discussion on how this could be achieved with minimal additional complexity on the beacon chain.  For ease of reading I'll refer to the partial withdrawal of funds above 32 Ether in a validator as "skimming".

Any early discussion regarding skimming was to allow a validator to skim at the same time they proposed a block.  Although simple, it has several downsides.  Most importantly, it provides no assurance as to when skimming for any given validator will take place: due to the random nature of proposer duties, one validator may skim twice in a month, whereas another may have to wait half a year.  Another downside is that it requires the block to be proposed: if the validator misses the block proposal for any reason (or the block is proposed but fails to make it on-chain) then the opportunity for skimming is lost.  These downsides are exacerbated with higher numbers of active validators.

Given this, it seems that a more reliable method of generating withdrawals would be nice.  Manual operations to carry out skimming may appear attractive, however this requires a new operation, new p2p channels, additional block space _etc._ so should be avoided if possible.

One option is to have a regular, controlled skim of all eligible* validators on a periodic basis.  For example, if the period was 1024 epochs then at the end of each epoch in the period 1/1024 of the validators would be processed, with each eligible validator in the set having a withdrawal receipt generated and made available to the execution layer for processing.

This removes the requirement for a new operation and all of the additional complexity and weight that goes with it, and maintains a regular schedule for rewards to be skimmed.  Additional load on the beacon node should be minimal due to the small number of validators processed per epoch, and if this is a concern then the period could be increased accordingly.

Having a system for partial withdrawals is important to avoid validators churning through the exit and activation queues to retrieve rewards, but adding to complexity and weight of the beacon chain any more than necessary is not desired and the proposed mechanism achieves the former whilst avoiding the latter.


*For a validator to be eligible it would need to be `ongoing_active` (not slashed), have a balance > `MAX_EFFECTIVE_BALANCE`, and have `0x01` withdrawal credentials.

specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,180 @@
...
+    pubkey: BLSPubkey
+    withdrawal_credentials: Bytes32  # Commitment to pubkey for withdrawals
+    effective_balance: Gwei  # Balance at stake
+    slashed: boolean
+    # Status epochs
+    activation_eligibility_epoch: Epoch  # When criteria for activation were met
+    activation_epoch: Epoch
+    exit_epoch: Epoch
+    withdrawable_epoch: Epoch  # When validator can withdraw funds
+    withdrawn_epoch: Epoch  # [New in Capella]
```
vbuterin, line 56: Why do this, instead of just zeroing out the validator once it's withdrawn?

potuz: It may be useful to later reuse the validator index.

protolambda:
@djrtwo @ralexstokes To clarify the remaining TODO, you're looking for an implementation:
- of an append-only merkle structure using the CommitmentContainer interface
- similar to that of the deposit-contract merkle stack updates
- that has the same root as a normal SSZ List would have
- that does not store the actual list contents since you only just need an updateable merkle-root for withdrawal proofs?

I think that can be done, I can give it a try maybe tomorrow.

protolambda:
See https://github.com/protolambda/consensus-specs/commit/17a70656f5e6eadc272faf2c36a832e38f4200a5

CommitmentContainer was not the right fit here since we need to commit to all state contents, including the merkle stack contents. Simply caching the root works fine in this case.


Mirrored the deposit contract merkle-stack approach:
- `compute_withdrawal_tree_root` to get the witthdrawal tree
- `withdraw` updates the merkle stack in the state

Extra: since we don't have to gas-golf as much: after processing all withdrawals, cache the new latest withdrawal-tree root in the BeaconState (so light-clients can proof it easily, and everyone can access it without re-computing)

Not sure if we can carry over formal verification easily, but it's the same thing.

Besides testing the merkle-updates (or formal verification) we'll need some integration tests (e.g. produce the proof for a withdrawal)

specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+### New containers
+
+#### `WithdrawalReceipt`
+
+```python
+class WithdrawalReceipt(Container):
+    index: WithdrawalReceiptIndex
+    address: ExecutionAddress
```
axic, line 111: I wonder if it would make sense making this `Bytes32` preparing for the possibility of "address space extension"?

It has some downsides:
1. The receipt is now packed, given 64 + 160 + 256 bits of values.
2. The deposit mechanism is still locked in to 160-bit addresses anyway, and likely many other places on the beacon chain too.

djrtwo:
closing in favor of the Push method #2836

Mitumaru (commented):
Ok

PR #2760, opened 2021-12-02
Apply proposer boost to ancestors correctly

### LINKED ISSUE #2757: Applying proposer boost to ancestors
It looks like the proposer score boost is not applied to ancestors of the boosted block, even during the boosted block's slot. Is this the correct behavior?

This is highlighted by the test scenario: `shorter_chain_but_heavier_weight`

In that scenario, during slot 3, there is a proposer-boosted block from slot three, as well as a late-arriving block from slot 1. The test expects the block from slot 1 to be set as head, even when they are tied in attestation weight (at zero). I'd expect the `get_head` method to incorporate the proposer boost and set block 3 as head.

adiasg:
Fixes #2757 by correctly applying proposer score boost to ancestors of the boosted block.
This removes the simplification changes from #2753.

### DIFF
--- a/specs/phase0/fork-choice.md
+++ b/specs/phase0/fork-choice.md
@@ -182,13 +182,15 @@ def get_latest_attesting_balance(store: Store, root: Root) -> Gwei:
             and get_ancestor(store, store.latest_messages[i].root, store.blocks[root].slot) == root)
     ))
     proposer_score = Gwei(0)
-    if store.proposer_boost_root != Root():
-        if get_ancestor(store, store.proposer_boost_root, store.blocks[root].slot) == root:
-            num_validators = len(get_active_validator_indices(state, get_current_epoch(state)))
-            avg_balance = get_total_active_balance(state) // num_validators
-            committee_size = num_validators // SLOTS_PER_EPOCH
-            committee_weight = committee_size * avg_balance
-            proposer_score = (committee_weight * PROPOSER_SCORE_BOOST) // 100
+    if (
+        store.proposer_boost_root != Root()
+        and get_ancestor(store, store.proposer_boost_root, store.blocks[root].slot) == root
+    ):
+        num_validators = len(get_active_validator_indices(state, get_current_epoch(state)))
+        avg_balance = get_total_active_balance(state) // num_validators
+        committee_size = num_validators // SLOTS_PER_EPOCH
+        committee_weight = committee_size * avg_balance
+        proposer_score = (committee_weight * PROPOSER_SCORE_BOOST) // 100
     return attestation_score + proposer_score
 
 ```
--- a/specs/phase0/fork-choice.md
+++ b/specs/phase0/fork-choice.md
@@ -184,6 +184,7 @@ def get_latest_attesting_balance(store: Store, root: Root) -> Gwei:
     if store.proposer_boost_root == Root():
         # Return only attestation score if ``proposer_boost_root`` is not set
         return attestation_score
+
     # Calculate proposer score if ``proposer_boost_root`` is set
     proposer_score = Gwei(0)
     # Boost is applied if ``root`` is an ancestor of ``proposer_boost_root``
--- a/specs/phase0/fork-choice.md
+++ b/specs/phase0/fork-choice.md
@@ -181,11 +181,13 @@ def get_latest_attesting_balance(store: Store, root: Root) -> Gwei:
         if (i in store.latest_messages
             and get_ancestor(store, store.latest_messages[i].root, store.blocks[root].slot) == root)
     ))
+    if store.proposer_boost_root == Root():
+        # Return only attestation score if ``proposer_boost_root`` is not set
+        return attestation_score
+    # Calculate proposer score if ``proposer_boost_root`` is set
     proposer_score = Gwei(0)
-    if (
-        store.proposer_boost_root != Root()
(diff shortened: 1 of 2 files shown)


specs/phase0/fork-choice.md
```diff
@@ -182,12 +182,13 @@ def get_latest_attesting_balance(store: Store, root: Root) -> Gwei:
...
-        committee_size = num_validators // SLOTS_PER_EPOCH
-        committee_weight = committee_size * avg_balance
-        proposer_score = (committee_weight * PROPOSER_SCORE_BOOST) // 100
+    if store.proposer_boost_root != Root():
+        if get_ancestor(store, store.proposer_boost_root, store.blocks[root].slot) == root:
+            num_validators = len(get_active_validator_indices(state, get_current_epoch(state)))
+            avg_balance = get_total_active_balance(state) // num_validators
+            committee_size = num_validators // SLOTS_PER_EPOCH
+            committee_weight = committee_size * avg_balance
+            proposer_score = (committee_weight * PROPOSER_SCORE_BOOST) // 100
```
hwwhww, line 191: It seems good to combine the two conditions so that we can reduce an indent.

```suggestion
    if (
        store.proposer_boost_root != Root()
        and get_ancestor(store, store.proposer_boost_root, store.blocks[root].slot) == root
    ):
        num_validators = len(get_active_validator_indices(state, get_current_epoch(state)))
        avg_balance = get_total_active_balance(state) // num_validators
        committee_size = num_validators // SLOTS_PER_EPOCH
        committee_weight = committee_size * avg_balance
        proposer_score = (committee_weight * PROPOSER_SCORE_BOOST) // 100
```


or more refactoring, but less efficient due to potential more `get_ancestor` calls:

```suggestion
    is_proposer_boost_root_ancestor = get_ancestor(store, store.proposer_boost_root, store.blocks[root].slot) == root
    if store.proposer_boost_root != Root() and is_proposer_boost_root_ancestor:
        num_validators = len(get_active_validator_indices(state, get_current_epoch(state)))
        avg_balance = get_total_active_balance(state) // num_validators
        committee_size = num_validators // SLOTS_PER_EPOCH
        committee_weight = committee_size * avg_balance
        proposer_score = (committee_weight * PROPOSER_SCORE_BOOST) // 100
```

Either way, since we were confused, could you add some comments about this condition? 🤣

specs/phase0/fork-choice.md
```diff
@@ -182,12 +182,13 @@ def get_latest_attesting_balance(store: Store, root: Root) -> Gwei:
             and get_ancestor(store, store.latest_messages[i].root, store.blocks[root].slot) == root)
     ))
     proposer_score = Gwei(0)
-    if store.proposer_boost_root != Root() and root == store.proposer_boost_root:
```
adiasg, line 185: This conditional implies that `get_latest_attesting_balance` applies the proposer score boost only to the proposer's block, but not its ancestors. This is not the correct method to apply the proposer score boost.

The boost should be applied like usual LMD votes (i.e., `store.latest_messages`) - all ancestors of `store.proposer_boost_root` should get the additional weight.

specs/phase0/fork-choice.md
adiasg: Yep, added [clarifying comment](https://github.com/ethereum/consensus-specs/pull/2760#discussion_r760747416) below.

hwwhww (approved):
LGTM 👍 

Thank @realbigsean for the find!

specs/phase0/fork-choice.md
```diff
@@ -182,7 +182,10 @@ def get_latest_attesting_balance(store: Store, root: Root) -> Gwei:
...
+    if (
+        store.proposer_boost_root != Root()
+        and get_ancestor(store, store.proposer_boost_root, store.blocks[root].slot) == root
```
djrtwo, line 187: generally should avoid multi-line ifs. This is also a bit secretly too pythonic (it assumes that if the first bool fails, the second is not executed, avoiding the undefined behavior if `Root()` being passed into `get_ancestor`

suggest  somthing like the following
```python
    if store.proposer_boost_root != Root():
        if get_ancestor():
            ...
```

An alternative, I personally prefer this:

```python
    if store.proposer_boost_root == Root():
        return attestation_score
    proposer_score = Gwei(0)
    if get_ancestor():
        ...
    return attestation_score + proposer_score
```

I think that very nicely reads as "If proposer_boost is not enabled, just return the attestation_score, otherwise calculate the proposer score and return the sum"

djrtwo (commented):
logic looks correct. some suggestions for alternative formating

djrtwo (approved):

specs/phase0/fork-choice.md
```diff
@@ -181,8 +181,13 @@ def get_latest_attesting_balance(store: Store, root: Root) -> Gwei:
         if (i in store.latest_messages
             and get_ancestor(store, store.latest_messages[i].root, store.blocks[root].slot) == root)
     ))
+    if store.proposer_boost_root == Root():
+        # Return only attestation score if ``proposer_boost_root`` is not set
+        return attestation_score
+    # Calculate proposer score if ``proposer_boost_root`` is set
```
djrtwo, line 187:
```suggestion

    # Calculate proposer score if ``proposer_boost_root`` is set
```

PR #2763, opened 2021-12-03
Validator delegation

potuz:
# Delegating validators

We allow for validators to delegate their excess earnings to another (possibly new) validator. Thus allowing for beacon-chain issuance to quickly start staking instead of either remaining inactive or impacting the exit churn. This all within the consensus layer, in particular not involving the deposit contract.

## What is included in this PR
This PR includes the changes needed in the beacon chain and an annotated version of these changes together with the text below and a rationale for this PR, I suggest [reading directly the design document](https://github.com/potuz/consensus-specs/blob/crux/specs/crux/design.md)

## Why this PR is needed
At the time of writing the beacon chain has 264,676 active validators, the average earning is 1.34 ETH and over 150,000 validators have earned over 1ETH. There are over 100,000 validators accounted for staking pools according to [beaconcha.in](https://beaconcha.in/pools). There is no reason to expect these validators to not withdraw their proceeds as soon as they become available. Disregarding those validators that would want to perform a normal withdrawal of all of their funds, there will still be a large number of validators that will simply withdraw their excess earnings, either to keep it off the consensus layer, or to compound their interest rate by staking extra validators, as is expectedly the case of staking pools. Under the current leading proposals these validators will have two paths to withdraw their earnings

- Enter an exit queue. Wait for their turn to become withdrawable. Deposit the principal. Enter an entry queue. Wait for their turn to become active.
- Wait for their turn to propose a block to perform a partial withdrawal.

In average the latter will happen once every 37 days with current numbers. The former incurs in costs of earnings lost while in queue and gas costs of redepositing. This leads to an equilibrium in which the exit queue becomes at least large enough that the cost is comparable to waiting 37 days to extract funds. This equilibrium is not a one-time run for the exit queue, as validator rotation to compound interest will be a constant exit force, and the wait time for a proposal increases linearly with the number of active validators, so will the exit queue.

This makes for a bad user experience, particularly for a single staker, 20% of which will have to wait for 2 months for a proposal with current numbers (presumably much more as the validator registry increases)

## Delegating validators

The current proposal solves the above problem by providing validators, and in particular staking pools, with a mechanism to continuously transfer their excess earnings to another validator. This new validator can be a new validator. There are several advantages to this mechanism:

- It keeps the non-staking capital to a minimum since as soon as the delegate validator reaches `MAX_EFFECTIVE_BALANCE` it enters the activation queue. 
- Validators can compound their interest immediately and spin off new validators more quickly, avoiding the partial withdrawal and deposit cycle (but are still subject to activation queue)
- Validators can trustlesly sell their excess earnings, providing liquidity for small stakers without requiring them to either wait possibly months nor withdraw their stake. Thus this could in principle be a decentralizing force.
- It allows for faster withdrawal of earnings since validators can delegate all their validators and then exit the delegate instead of waiting for all their validators to propose.

Adding actual numbers: assuming that every current eligible validator decides to delegate, it would take a little over 1 day to clear the entire set of delegations with `MAX_DELEGATIONS=16` and a little over 5 days to do it with `MAX_DELEGATIONS=4`. After that, beacon blocks would regularly be without delegations as validators need to reach the 1 ETH threshold to be able to delegate


specs/crux/beacon-chain.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Extended containers
+
+#### `Validator`
+The `Validator` container gains a new field `delegate` containing the validator index of the delegate validator. This field is initialized at the fork with the same validator index of the current validator. (N.B. implementers may chose to keep a separate map instead)
+
+```python
+class Validator(Container):
+    pubkey: BLSPubkey
+    delegate: ValidatorIndex  # [New in Crux]
```
dapplion, line 82: Please insert new fields at the bottom. Otherwise you may break optimizations that assume data ordering either merkelized or serialized

specs/crux/beacon-chain.md
```diff
@@ -0,0 +1,247 @@
...
+    process_historical_roots_update(state)
+    process_participation_flag_updates(state)
+    process_sync_committee_updates(state)
+```
+
+#### `process_delegation_transfers`
+
+```python
+def process_delegation_transfers(state: BeaconState) -> None:
+    delegating_validators = [(i, v) for i, v in enumerate(state.validators) if v.delegate > i and not v.slashed]
```
dapplion, line 239: Validators can't delegate to validators with index less than theirs?

specs/crux/beacon-chain.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+### New containers
+
+#### `DelegationMessage`
+
+```python
+class DelegationMessage(Container):
+    delegating_index: ValidatorIndex
+    delegating_pubkey: BLSPubkey
```
dapplion, line 119: Why is delegating_pubkey necessary?

potuz: It's the withrawal public key that signs the message

specs/crux/beacon-chain.md
potuz: no, this is a cheap way to prevent loops.

specs/crux/beacon-chain.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+### Epoch processing
+
+```python
+def process_epoch(state: BeaconState) -> None:
+    process_justification_and_finalization(state)
+    process_inactivity_updates(state)
+    process_rewards_and_penalties(state)
+    process_delegation_transfers(state)  # [New in Crux]
```
dapplion, line 221: Should this be done after processing any other items that may modify balances?

specs/crux/beacon-chain.md
```diff
@@ -0,0 +1,247 @@
...
+    process_participation_flag_updates(state)
+    process_sync_committee_updates(state)
+```
+
+#### `process_delegation_transfers`
+
+```python
+def process_delegation_transfers(state: BeaconState) -> None:
+    delegating_validators = [(i, v) for i, v in enumerate(state.validators) if v.delegate > i and not v.slashed]
+    for index, validator in delegating_validators:
```
dapplion, line 238: So the extra load for validators for delegation is a full loop on the entire validator set + potentially modify the entire balances array. Is it really necessary to be done every epoch? Maybe every N to reduce overall load of the feature

specs/crux/beacon-chain.md
potuz: I couldn't think of any danger on doing this here. I think it has to be before `process_registry_updates`  and `process_effective_balance` though.

specs/crux/beacon-chain.md
potuz: In principle this can be done within `process_rewards_and_penalties` avoiding any extra loop and any extra load costs. It is simply a check before applying a reward if it has to go to the delegate or not. I chose the current approach because of its simplicity to gauge interest in this type of application or not. You are right however that this can be done every `N` epochs.

dapplion: Definitely agree this is a very useful feature to reduce load in the queues :+1:

dapplion: Tho in other POS networks delegate is a widely use term that needs something different from what this PR achieves. Would you be a better term for the action of moving excess balance to another validator?

potuz: Thanks, surely delegation will not be the final term, every single person that reacted positively to this PR pointed that this term is commonly used in a different way, if this gets any traction hopefully someone will come up with the right term. FWIW, an implementation that doesn't have an extra loop is 
```python
        return

    flag_deltas = [get_flag_index_deltas(state, flag_index) for flag_index in range(len(PARTICIPATION_FLAG_WEIGHTS))]
    deltas = flag_deltas + [get_inactivity_penalty_deltas(state)]
    for (rewards, penalties) in deltas:
        for index, val in enumerate(state.validators):
            decrease_balance(state, ValidatorIndex(index), penalties[index])
            if rewards[index] > 0:
                new_balance = state.balances[index] + rewards[index]
                to_delegate = false
                if val.delegate > index and not val.slashed and new_balance > MAX_EFFECTIVE_BALANCE:
                    delegate_validator = state.validators[val.delegate]
                    if delegate_validator.exit_epoch == FAR_FUTURE_EPOCH:
                        to_delegate = true
                if to_delegate: 
                    state.balance[index] = MAX_EFECTIVE_BALANCE
                    increase_balance(state, val.delegate, new_balance - MAX_EFFECTIVE_BALANCE)
                else:
                    state.balance[index] = new_balance
```
But I think actually the extra loop only over the delegating indices is more efficient in compiled languages. And regardless of any choice, implementers can do either one cause it does not affect the final state transition. The biggest cost of this PR I think is in signature verification and block size, not here. But at `MAX_DELEGATIONS = 16`  in only 1 day blocks should be back to empty. And we currently process 16 signatures without any problem.

potuz:
Closing this as the feature is pretty much obsoleted by the current partial withrawals proposal

PR #2765, opened 2021-12-07
Update per-test config to be unique per-test in `spec_configured_state_test`

ralexstokes:
Alternative fix for the issue addressed in #2755.

As I understand it, the issue there was that using the `spec_configured_state_test` decorator was not "thread-safe" so that any unsynchronized access could cause issues. These issues were seen  in #2752 when executing the tests with `pytest-xdist` which parallelizes tests across multiple cores.

@hwwhww found the issue and suggested a fix w/ lazy loading of the modules in the pyspec.

Rather than go around making multiple copies of various modules, this PR takes a different approach to tests using the `spec_configured_state_test` decorator: simply make a separate copy of the `Spec` object.

As long as instances of `Spec` don't get too big or resource-intensive then we should be able to easily specialize a given `Spec` for each test that uses this decorator by simply making a copy and making the necessary edits.

As far as I can tell, there is no reason to "restore" the spec after mutation as it is discarded after the result from the decorated test is obtained.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/context.py
+++ b/tests/core/pyspec/eth2spec/test/context.py
@@ -484,10 +484,10 @@ def _get_basic_dict(ssz_dict: Dict[str, Any]) -> Dict[str, Any]:
 def _get_copy_of_spec(spec):
     fork = spec.fork
     preset = spec.config.PRESET_BASE
-    path = f"eth2spec.{fork}.{preset}"
-
-    module_spec = importlib.util.find_spec(path)
+    module_path = f"eth2spec.{fork}.{preset}"
+    module_spec = importlib.util.find_spec(module_path)
     module = importlib.util.module_from_spec(module_spec)
+    module_spec.loader.exec_module(module)
     return module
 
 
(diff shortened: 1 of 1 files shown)


ralexstokes:
update: the `Spec` is a python module where we cannot simply just copy it. thinking of other solutions....

ralexstokes:
after chatting w/ @hwwhww came to the current solution using `importlib` to load a fresh copy of the `spec` per test that needs to customize it, but skipping the lazy loading in #2755 and also only loading a duplicate in the specific tests that need customization, not *every* test

ralexstokes:
there is some strange failure w/ a test (`test_config_override`) that seems to pass on `dev`

but ignoring that for now, i'd suggest we do something like this PR instead of #2755 to fix #2752 (and anything else that uses that decorator) as it provides a distinct, scoped instance of the `spec` to each test that wants to mutate it.

hwwhww (approved):
Big thanks to @ralexstokes ✨

Compared to #2755, it gets rid of the needless imports largely. Well done!

PR #2770, opened 2021-12-20
Optimistic Sync

paulhauner:
This PR introduces a specification for optimistic sync.

Optimistic sync provides a way for consensus engines (beacon nodes) to follow the head of the chain without verifying execution payloads. The requirement for this is based upon the premise that execution engines can sync most effectively if they are able to see the payload hashes from blocks at the immediate head of the beacon chain.

I have created an accompanying [Opt. Sync Spec: Reading Companion](https://hackmd.io/_Dgy7fCoSgiUxDT1-zvvsg?view) to help explain some of my decisions.

Opt sync is a fairly tricky mechanism and it comes with some tangible risks to liveness. I would appreciate anyone who takes the time to cast a critical eye across this.

Thanks to @mkalinin, @ajsutton and @djrtwo for providing substance and feedback.


sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+	block.body.execution_payload != ExecutionPayload()
+```
+
+```python
+def should_optimistically_import_block(current_slot: Slot, block: BeaconBlock) -> bool:
+	block.slot + SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY <= current_slot
+```
+
+Let only a node which returns `is_optimistic(head) == True` be an *optimistic
+node*. Let only a validator on an optimistic node be an *optimistic validator*.
```
dapplion, line 55: Should all lines be contiguous without new lines to match the format of the rest of the spec?

dapplion:
Should a previously synced / full beacon node revert back to optimistic sync if it's behind N blocks of the head? If so, what would be an appropriate number for N?

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+
+## When to optimistically import blocks
+
+A block MUST NOT be optimistically imported, unless either of the following
+conditions are met:
+
+1. The justified checkpoint has execution enabled. I.e.,
+   `is_execution_block(get_block(get_state(head_block).current_justified_checkpoint.root))`
+1. The current slot (as per the system clock) is at least `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` ahead of
+   the slot of the block being imported. I.e., `should_optimistically_import_block(current_slot) == True`.
```
terencechain, line 71:
```suggestion
   the slot of the block being imported. I.e., `should_optimistically_import_block(current_slot, block) == True`.
```

paulhauner:
> Should a previously synced / full beacon node revert back to optimistic sync if it's behind N blocks of the head? If so, what would be an appropriate number for N?

I'm not sure I completely understand the question, but after a post-merge (execution-enabled) block is justifed then the node may optimstically import a block at any time. If the node optimistically imports a block and that block becomes the head, then that node is considered "optimistic". Does that make sense, or did I miss your question?

sync/optimistic.md
paulhauner: I'm not sure that the rest of the spec documents are consistent regarding new lines. For example, https://raw.githubusercontent.com/ethereum/consensus-specs/dev/specs/merge/p2p-interface.md seems to start a new lines mid-sentence and mid-paragraph.

I'd be happy to change if the spec maintainers have some preference,

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+
+When this specification only defines behaviour for an optimistic
+node/validator, but *not* for the non-optimistic case, assume default
+behaviours without regard for optimistic sync.
+
+## Mechanisms
+
+## When to optimistically import blocks
+
+A block MUST NOT be optimistically imported, unless either of the following
```
rauljordan, line 65: This can probably be reworded to more straightforward language:
```suggestion
A block can be optimistically imported ONLY when either of the following conditions are met
```

rauljordan (approved):

sync/optimistic.md
paulhauner: Ah yeah, I was trying to work within the [RFC keywords](https://datatracker.ietf.org/doc/html/rfc2119). I'll check it out and see if I can make it simpler :pray:

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+If the justified checkpoint transitions from `SYNCING` -> `INVALID`, a
+consensus engine MAY choose to alert the user and force the application to
+exit.
+
+## Fork Choice
+
+Consensus engines MUST support removing blocks from fork choice that transition
+from `SYNCING` to `INVALID`. Specifically, a block deemed `INVALID` at any
+point MUST NOT be included in the canonical chain and the weights from those
+`INVALID` blocks MUST NOT be applied to any `VALID` or `SYNCING` ancestors.
```
mkalinin, line 143: It implies `store.latest_messages` is reverted to a state in which an `INVALID` block doesn't exist, right? I think it worth noting explicitly.

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+Let only a node which returns `is_optimistic(head) == True` be an *optimistic
+node*. Let only a validator on an optimistic node be an *optimistic validator*.
+
+When this specification only defines behaviour for an optimistic
+node/validator, but *not* for the non-optimistic case, assume default
+behaviours without regard for optimistic sync.
+
+## Mechanisms
+
+## When to optimistically import blocks
```
mkalinin, line 63:
```suggestion
### When to optimistically import blocks
```

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+
+1. The justified checkpoint has execution enabled. I.e.,
+   `is_execution_block(get_block(get_state(head_block).current_justified_checkpoint.root))`
+1. The current slot (as per the system clock) is at least `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` ahead of
+   the slot of the block being imported. I.e., `should_optimistically_import_block(current_slot, block) == True`.
+
+*See [Fork Choice Poisoning](#fork-choice-poisoning) for the motivations behind
+these conditions.*
+
+## How to optimistically import blocks
```
mkalinin, line 76:
```suggestion
### How to optimistically import blocks
```

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+eternally-unavailable `body.execution_payload.parent_hash` (i.e., the parent
+hash is random bytes). In rare circumstances, it is possible that an attacker
+can build atop such a block to trigger justification. If an optimistic node
+imports this malicious chain, that node will have a "poisoned" fork choice
+store, such that the node is unable to produce a block that descends from the
+head (due to the invalid chain of payloads) and the node is unable to produce a
+block that forks around the head (due to the justification of the malicious
+chain).
+
+If honest chain exists which justifies a higher epoch than the malicious chain,
```
mkalinin, line 158: It should be enough for a node to import an honest chain justifying an epoch *prior to* malicious chain that justifies the *same* epoch to avoid poisoning the fork choice store and eventually finishing an optimistic sync which in this case should converge on the honest chain

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+An optimistic validator MAY subscribe to the `beacon_block` topic. Propagation
+validation conditions are modified as such:
+
+Do not apply the existing condition:
+
+- [REJECT] The block's parent (defined by block.parent_root) passes validation.
+
+Instead, apply the new condition:
+
+- [REJECT] The block's parent (defined by block.parent_root) passes all
```
mkalinin, line 223: Consider reflecting this change in the `p2p-interface`

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+-  `beacon_attestation_{subnet_id}`
+-  `sync_committee_contribution_and_proof`
+-  `sync_committee_{subnet_id}`
+
+Once the node ceases to be optimistic, it MAY re-subscribe to the
+aforementioned topics.
+
+### The Req/Resp Domain
+
+Non-faulty, optimistic nodes may send blocks which result in an INVALID
```
mkalinin, line 249: Do we want this to be reflected in the `p2p-interface` as well?

mkalinin (commented):
Amazing work! 👍 I left a few comments and suggestions.

@paulhauner what do you think about splitting this document into specification and rationale sections and moving your accompanying doc and fork choice poisoning description to the rationale part? This distinction should make the specification cleaner.

This specification does also require a respective change to the Engine API that would clarify the SYNCING response with respect to different forks, i.e. a fork-specific nature of SYNCING. And it probably worth updating `p2p-interface` with the changes allowing for sending optimistic blocks via gossip and req/resp.

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+
+When `is_optimistic(head) == True`, the consensus engine:
+
+- MUST NOT return an optimistic `head`.
+- MAY substitute the head block with `latest_valid_ancestor(block)`.
+- MAY return syncing.
+
+### Requests to Validators Endpoints
+
+When `is_optimistic(head) == True`, the consensus engine MUST return syncing to
```
hwwhww, line 287:
```suggestion
When `is_optimistic(head) is True`, the consensus engine MUST return syncing to
```

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+
+When information about an optimistic block is requested, the consensus engine:
+
+- MUST NOT respond with success.
+- MAY respond with not found.
+- MAY respond with syncing.
+
+### Requests for an Optimistic Head
+
+When `is_optimistic(head) == True`, the consensus engine:
```
hwwhww, line 279:
```suggestion
When `is_optimistic(head) is True`, the consensus engine:
```

hwwhww (commented):
Great work!

Just some Python nitpicking comments here.

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+   the slot of the block being imported. I.e., `should_optimistically_import_block(current_slot, block) == True`.
+
+*See [Fork Choice Poisoning](#fork-choice-poisoning) for the motivations behind
+these conditions.*
+
+## How to optimistically import blocks
+
+To optimistically import a block:
+
+- The `execute_payload` function MUST return `True` if the execution
```
terencechain, line 80: Should we add a relative link to the function here?

```suggestion
- The [`execute_payload`](../specs/merge/beacon-chain.md#execute_payload) function MUST return `True` if the execution
```

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+	engine returns `SYNCING` or `VALID`. An `INVALID` response MUST return `False`.
+- The `validate_merge_block` function MUST NOT raise an assertion if both the
```
terencechain, line 82:
```suggestion
- The [`validate_merge_block`](../specs/merge/fork-choice.md#validate_merge_block) function MUST NOT raise an assertion if both the
```

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+## Validator assignments
+
+An optimistic node is *not* a full node. It is unable to produce blocks, since
+an execution engine cannot produce a payload upon an unknown parent. It cannot
+faithfully attest to the head block of the chain, since it has not fully
+verified that block.
+
+### Block Production
+
+A optimistic validator MUST NOT produce a block (i.e., sign across the
```
terencechain, line 193:
```suggestion
An optimistic validator MUST NOT produce a block (i.e., sign across the
```

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+### Execution Engine Errors
+
+When an execution engine returns an error or fails to respond to a payload
+validity request for some block, a consensus engine:
+
+- MUST NOT optimistically import the block.
+- MUST NOT apply the block to the fork choice store.
+- MAY queue the block for later processing.
+
+### Assumptions about Execution Engine Behaviour
```
terencechain, line 119: The other Bellatrix specs don't specify EE assumptions. It makes me wonder if it is better to update the official execution API to reflect this

terencechain (commented):
I took another look at this before implementing it in Prysm 🙏🏼

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+
+```python
+def is_optimistic(block: BeaconBlock) -> bool:
+    hash_tree_root(block) in optimistic_roots
+```
+
+```python
+def latest_valid_ancestor(block: BeaconBlock) -> BeaconBlock:
+    while True:
+        if not is_optimistic(block) or block.parent_root == Root():
```
potuz, line 39: Something that confused me here is that a block that returned `INVALID`  from the EL would pass this test and return here `block`. Perhaps it's worth adding a note that `block` is assumed valid here?

sync/optimistic.md
paulhauner: This is a good point, we considered this too. @mkalinin and I spoke about this and ended up deciding to leave this note here and Mikhail would update the EE specs.

Do you have any objections to leaving this here if we also update the EE specs, @terencechain?

paulhauner:
> what do you think about splitting this document into specification and rationale sections and moving your accompanying doc and fork choice poisoning description to the rationale part? This distinction should make the specification cleaner.

Yep, this makes sense. I imagine this is the same format as the current p2p-interface doc. I'll do this.

> And it probably worth updating p2p-interface with the changes allowing for sending optimistic blocks via gossip and req/resp.

I hadn't considered if nodes that *don't* follow the optimistic spec should allow invalid payloads or not. I assume that's what you're suggesting by updating the p2p spec?

Presently, I prefer updating the Bellatrix p2p-interface spec to allow invalid payloads on gossip for all core-spec-compliant nodes (as you suggested).

sync/optimistic.md
paulhauner: > It implies store.latest_messages is reverted to a state in which an INVALID block doesn't exist, right?

It doesn't necessarily "revert" to an old state, since other aspects of the block tree may differ.

I agree that it should be *as if* the INVALID block doesn't exist, but I'm not sure how to specify that when in proto-array it will "exist" but be filtered out. Perhaps you could provide another suggestion, please?

sync/optimistic.md
mkalinin: Here is the respective change to the Engine API https://github.com/ethereum/execution-apis/pull/152

mkalinin:
> Presently, I prefer updating the Bellatrix p2p-interface spec to allow invalid payloads on gossip for all core-spec-compliant nodes (as you suggested).

I think that changes to p2p-interface are necessary for the health of the network if a portion of nodes that use optimistic sync is relatively high. Imagine the case when non-optimistic-sync node drops and descores a remote peer that is in optimistic sync thus can't verify a payload of blocks it is propagating.

sync/optimistic.md
terencechain: > Do you have any objections to leaving this here if we also update the EE specs, @terencechain?

No objection :)

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+
+## When to optimistically import blocks
+
+A block MUST NOT be optimistically imported, unless either of the following
+conditions are met:
+
+1. The justified checkpoint has execution enabled. I.e.,
+   `is_execution_block(get_block(get_state(head_block).current_justified_checkpoint.root))`
+1. The current slot (as per the system clock) is at least `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` ahead of
+   the slot of the block being imported. I.e., `should_optimistically_import_block(current_slot, block) == True`.
```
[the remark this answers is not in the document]
paulhauner: I've done this in 6c13e2ec. I added a `Store` so we can access the blocks at states without undefined `get_block` and `get_state` functions. Adding blocks/states to the `Store` is undefined, but I guess that's OK.

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+## Helpers
+
+Let `head_block: BeaconBlock` be the result of calling of the fork choice
+algorithm at the time of block production.
+
+Let `justified_block: BeaconBlock` be the latest current justified ancestor
+ancestor of the `head_block`.
+
+Let `optimistic_roots: Set[Root]` be the set of `hash_tree_root(block)` for all
+optimistically imported blocks which have yet to receive an `INVALID` or
```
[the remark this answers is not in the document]
paulhauner: Good idea, I changed this in 2ce2aac0. Let me know what you think about that wording.

sync/optimistic.md
```diff
@@ -0,0 +1,290 @@
...
+*Note: the `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` must be user-configurable. See
+[Fork Choice Poisoning](#fork-choice-poisoning).*
+
+## Helpers
+
+Let `head_block: BeaconBlock` be the result of calling of the fork choice
+algorithm at the time of block production.
+
+Let `justified_block: BeaconBlock` be the latest current justified ancestor
+ancestor of the `head_block`.
```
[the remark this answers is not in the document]
paulhauner: Thanks, removed in 736f3cec.

sync/optimistic.md
paulhauner: >  an honest chain justifying an epoch prior to malicious chain that justifies the same epoch 

I'm not sure I follow this sorry.

sync/optimistic.md
paulhauner: Good idea, I did this in 1228e018.

sync/optimistic.md
paulhauner: Also done in 1228e018.

sync/optimistic.md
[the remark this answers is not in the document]
paulhauner: Oh yep, I forgot this! Added here: 60eab257.

sync/optimistic.md
paulhauner: Makes sense! I added a comment in de1a6caa.

sync/optimistic.md
mkalinin: > If an honest chain exists which justifies a higher epoch than the malicious chain, that chain will take precedence and revive any poisoned store.

Do we need a *higher* epoch to be justified by an honest chain? From my perspective the *same* epoch justified by an honest chain is enough to prevent poisoned store, with the condition that an honest justification is applied to the store *before* the malicious one. Simply:

1) A node imports malicious chain that justifies epoch `N`
2) The store is poisoned and rejects blocks of an honest chain that has not yet justified `N`

VS

1) A node imports an honest chain that justifies epoch `N`
2) A node imports a malicious chain that justifies epoch `N`
3) Despite of (2) the store is healthy and is capable of keep applying blocks of an honest chain

An honest chain justifying `N+1` does also prevent store poisoning. My point is that justifying `N` is also enough. And this is what we assume to get in 4 epochs: an honest chain with adversary power `< 1/3` justifies epoch `N`. Then we allow for optimistically import *any* other chain without the risk of a store being poisoned.

Am I missing anything?

paulhauner: >  with the condition that an honest justification is applied to the store before the malicious one

Ah yes, I see what you mean now. So a higher epoch *always* fixes it and a lower one fixes it *sometimes*, given a message ordering assumption.

I've mentioned this "poisoning prevention" in 6d73b0a4 :)

paulhauner:
I have added the "Design Decision Rationale" section. Thanks everyone for the comments, I believe I have addressed them all.

specs/bellatrix/p2p-interface.md
```diff
@@ -85,13 +86,24 @@ The *type* of the payload of this topic changes to the (modified) `SignedBeaconB
...
 the following validations MUST pass before forwarding the `signed_beacon_block` on the network.
 Alias `block = signed_beacon_block.message`, `execution_payload = block.body.execution_payload`.
 - If the execution is enabled for the block -- i.e. `is_execution_enabled(state, block.body)`
   then validate the following:
   - _[REJECT]_ The block's execution payload timestamp is correct with respect to the slot
     -- i.e. `execution_payload.timestamp == compute_timestamp_at_slot(state, block.slot)`.
+  - [REJECT] The block's parent (defined by `block.parent_root`) passes all
+      validation, excluding verification of the `block.body.execution_payload`.
+  - [IGNORE] The block's parent (defined by `block.parent_root`) passes all
+      validation, including verification of the `block.body.execution_payload`.
```
ajsutton, line 103: Would it be reasonable to phrase this as:
```suggestion
  - [IGNORE] The block's parent (defined by `block.parent_root`) passes 
        verification of the `block.body.execution_payload`.
```

Given the reject rule above we already know that all verification except for the `execution_payload` passes so are just choosing between ignore and accept based on whether the execution payload is valid or not.

Although both versions still feel ambiguous about whether the result is ACCEPT or IGNORE in the case of the parent execution payload not yet being executed because the EL is syncing. I think it should be ACCEPT because the whole point is that an optimistic node can't validate the payload but also needs to avoid censoring blocks just because it can't validate the parent.

sync/optimistic.md
```diff
@@ -0,0 +1,340 @@
...
+
+A consensus engine MUST be able to retrospectively (i.e., after import) modify
+the status of `SYNCING` blocks to be either `VALID` or `INVALID` based upon responses
+from an execution engine. I.e., perform the following transitions:
+
+- `SYNCING` -> `VALID`
+- `SYNCING` -> `INVALID`
+
+When a block transitions from `SYNCING` -> `VALID`, all *ancestors* of the
+block MUST also transition from `SYNCING` -> `VALID`. Such a block is no longer
```
djrtwo, line 119:
```suggestion
block MUST also transition from `SYNCING` -> `VALID`. Such a block and any previously `SYNCING` ancestors are no longer
```

sync/optimistic.md
```diff
@@ -0,0 +1,340 @@
...
+considered "optimistically imported".
+
+When a block transitions from `SYNCING` -> `INVALID`, all *descendants* of the
+block MUST also transition from `SYNCING` -> `INVALID`.
+
+When a block transitions from the `SYNCING` state it is removed from the set of
```
djrtwo, line 125:
```suggestion
When a block transitions from the `SYNCING` state, it is removed from the set of
```

sync/optimistic.md
```diff
@@ -0,0 +1,340 @@
...
+`store.optimistic_roots`.
+
+When a "merge block" (i.e. a block which enables execution) is declared to be
```
djrtwo, line 128:
```suggestion
When a "merge block" (i.e. the first block which enables execution in a chain) is declared to be
```

sync/optimistic.md
```diff
@@ -0,0 +1,340 @@
...
+there is insufficient information available to make a `VALID` or `INVALID`
+determination on the given `ExecutionPayload` (e.g., the parent payload is
+unknown). Specifically, `SYNCING` responses should be fork-specific, in that
+the search for a block on one chain MUST NOT trigger a `SYNCING` response for
+another chain.
+
+### Re-Orgs
+
+The consensus engine MUST support any chain reorganisation which does *not*
+affect the justified checkpoint. The consensus engine MAY support re-orgs
```
djrtwo, line 156: Can this section clarify these are re-orgs of optimistic chain segments?

A fully verified justification will never revert if the consensus engine is following the spec *and* there is no 2/3 slashing event. (but a 2/3 slashing even could lead to a finality that is discordant with your current justification and pull you to another chain)

sync/optimistic.md
```diff
@@ -0,0 +1,340 @@
...
+from `SYNCING` to `INVALID`. Specifically, a block deemed `INVALID` at any
+point MUST NOT be included in the canonical chain and the weights from those
+`INVALID` blocks MUST NOT be applied to any `VALID` or `SYNCING` ancestors.
+
+### Fork Choice Poisoning
+
+During the merge transition it is possible for an attacker to craft a
+`BeaconBlock` with an execution payload that references an
+eternally-unavailable `body.execution_payload.parent_hash` (i.e., the parent
+hash is random bytes). In rare circumstances, it is possible that an attacker
```
djrtwo, line 175: Is "rare" the proper word here? Is it rare from a probability standpoint or actually difficult or costly to induce from a technical or crypto-economic standpoint?

djrtwo: Ah, this is "rare" as in it's a blck that *triggers* a justification and thus causes issues with `filter_block_tree`

Am I correct in assessing that a "fix" to the issue found a couple of weeks ago would sidestep this deadlock problem @paulhauner ?

sync/optimistic.md
```diff
@@ -0,0 +1,340 @@
...
+makes it impossible for honest nodes to build atop that block. If an attacker
+exploits a nuance in fork choice `filter_block_tree`, they can, in some rare
+cases, produce a junk block that out-competes all locally produced blocks for
+the head. This prevents a node from producing a chain of blocks, therefore
+breaking liveness.
+
+Thankfully, if 2/3rds of validators are not poisoned, they can justify an
+honest chain which will un-poison all other nodes.
+
+Notably, this attack only exists for optimistic nodes. Nodes which fully verify
```
djrtwo, line 290: Probably worth making it even clearer that this means that if the vast majority of the network is "synced" going into the transition (which is generally the case for the network, and validator-nodes in particular), then the network cannot be poisoned

sync/optimistic.md
```diff
@@ -0,0 +1,340 @@
...
+## Mechanisms
+
+### When to optimistically import blocks
+
+A block MAY be optimistically imported when
+`should_optimistically_import_block(store, current_slot, block)` returns
+`True`. This ensures that blocks are only optimistically imported if either:
+
+1. The justified checkpoint has execution enabled.
+1. The current slot (as per the system clock) is at least
```
djrtwo, line 87: Suggest defining `current_slot` wrt wall clock time in the helper exposition. Otherwise, it isn't clear when just reading `should_optimistically_import_block` what it means in that context

sync/optimistic.md
```diff
@@ -0,0 +1,340 @@
...
+```
+
+```python
+def is_execution_block(block: BeaconBlock) -> BeaconBlock:
+    return block.body.execution_payload != ExecutionPayload()
+```
+
+```python
+def should_optimistically_import_block(store: Store, current_slot: Slot, block: BeaconBlock) -> bool:
+    justified_root = store.block_states[store.head_block_root].current_justified_checkpoint.root
```
djrtwo, line 65: Consider a header comment for this function that this `should` turns into a "MAY optimistically" import (rather than a "MUST optimistically import".

In fact maybe we should change the name from `should_` to `may_` or try something like `is_optimistically_importable`

sync/optimistic.md
```diff
@@ -0,0 +1,340 @@
...
+*See [Fork Choice Poisoning](#fork-choice-poisoning) for the motivations behind
+these conditions.*
+
+### How to optimistically import blocks
+
+To optimistically import a block:
+
+- The [`execute_payload`](../specs/bellatrix/beacon-chain.md#execute_payload) function MUST return `True` if the execution
+	engine returns `SYNCING` or `VALID`. An `INVALID` response MUST return `False`.
+- The [`validate_merge_block`][] function MUST NOT raise an assertion if both the
```
djrtwo, line 100: This means if these values  are unknown, bypass that and the rest of the validation. "MUST NOT raise an assertion" was not immediately clear to me.

maybe add "i.e. `if pow_block is one and pow_parent is None`, skip all assertions in `validate_merge_block`." or something to that tune

djrtwo: Have we considered how to do this when `TERMINAL_BLOCK_HASH` is set?

djrtwo (commented):
My apologies for taking so long on review here.

This is *awesome*!

sync/optimistic.md
```diff
@@ -0,0 +1,340 @@
...
+
+```python
+def is_execution_block(block: BeaconBlock) -> BeaconBlock:
+    return block.body.execution_payload != ExecutionPayload()
+```
+
+```python
+def should_optimistically_import_block(store: Store, current_slot: Slot, block: BeaconBlock) -> bool:
+    justified_root = store.block_states[store.head_block_root].current_justified_checkpoint.root
+    justifed_is_verified = is_execution_block(store.blocks[justified_root])
```
ajsutton, line 66: nit: Is `justified_is_verified` the right name here?  We're effectively checking that the merge block is justified so may be `merge_is_justified`? or `justified_is_execution_block`.

To me the `verified` in the name is making me think it's about whether it's optimistic or fully verified.

specs/bellatrix/p2p-interface.md
paulhauner: I struggled to figure out the wording for this. The language for these propagation conditions is tricky!

I've applied @djrtwo's suggestion in b1ec9bcf.

I was tempted to go off piste and use a different specification language for these conditions, since it would have been easier to parse (IMO). However, that breaks standard and might get a bit unruly if we have *another* spec that needs to override some of these rules.

I'm still open to suggestion here :sweat_smile:

specs/bellatrix/p2p-interface.md
```diff
@@ -100,6 +112,12 @@ details on how to handle transitioning gossip topics for Bellatrix.
 
 ## The Req/Resp domain
 
+Non-faulty, [optimistic](/sync/optimistic.md) nodes may send blocks which
+result in an INVALID response from an execution engine. To prevent network
+segregation between optimistic and non-optimistic nodes, transmission of an
+INVALID payload via the Req/Resp domain SHOULD NOT cause a node to be
```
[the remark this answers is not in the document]
paulhauner: Addressed in 6225236a

specs/bellatrix/p2p-interface.md
```diff
@@ -181,3 +199,26 @@ valid block sizes in the range of gas limits expected in the medium term.
...
+### Why allow invalid payloads on the P2P network?
+
+The specification allows blocks with invalid payloads to propagate across
+gossip and via RPC calls. The reasoning for this is as follows:
+
+1. Optimistic nodes must listen to block gossip to obtain a view of the head of
+   the chain.
+2. Therefore, optimistic nodes must propagate gossip blocks. Otherwise, they'd
+   be censoring.
+3. If optimistic nodes will propose blocks via gossip, then they must respond
```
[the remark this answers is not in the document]
paulhauner: Addressed in 092f3e0b

sync/optimistic.md
```diff
@@ -0,0 +1,340 @@
...
+
+[`validate_merge_block`]: ../specs/bellatrix/fork-choice.md#validate_merge_block
+
+## Introduction
+
+In order to provide a syncing execution engine with a partial view of the head
+of the chain, it may be desirable for a consensus engine to import beacon
+blocks without verifying the execution payloads. This partial sync is called an
+*optimistic sync*.
+
```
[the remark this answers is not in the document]
paulhauner: Good idea, addressed in 52caba6e

sync/optimistic.md
```diff
@@ -0,0 +1,340 @@
...
+BeaconState]` be the blocks (and accompanying states) that have been verified
+either completely or optimistically.
+
+Let `optimistic_roots: Set[Root]` be the set of `hash_tree_root(block)` for all
+optimistically imported blocks which have only received a `SYNCING` designation
+from an execution engine (i.e., they are not known to be `INVALID` or `VALID`).
+
+```python
+@dataclass
+class Store(object):
```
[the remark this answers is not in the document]
paulhauner: Great idea! Addressed in b50bee16

sync/optimistic.md
```diff
@@ -0,0 +1,340 @@
...
+
+To optimistically import a block:
+
+- The [`execute_payload`](../specs/bellatrix/beacon-chain.md#execute_payload) function MUST return `True` if the execution
+	engine returns `SYNCING` or `VALID`. An `INVALID` response MUST return `False`.
+- The [`validate_merge_block`][] function MUST NOT raise an assertion if both the
+`pow_block` and `pow_parent` are unknown to the execution engine.
+- The parent of the block MUST NOT have an INVALID execution payload.
+
+In addition to this change to validation, the consensus engine MUST be able to
```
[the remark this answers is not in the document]
paulhauner: Nice one, addressed in 4ccd38b5

sync/optimistic.md
paulhauner: Oh yeah, good point. In f4a125c8 I just removed this second sentence since I don't think it was really giving us anything. LMK what you think.

sync/optimistic.md
paulhauner: > Is "rare" the proper word here? Is it rare from a probability standpoint or actually difficult or costly to induce from a technical or crypto-economic standpoint?

I was using rare since the conditions required for this bug rarely occur on mainnet. I was being somewhat vague to avoid needing to disclose too much about the attack. I'm open to re-wording, though!

> Am I correct in assessing that a "fix" to the issue found a couple of weeks ago would sidestep this deadlock problem @paulhauner ?

I've also been thinking about this. I'd need to see the concrete proposal, but I don't think it's the silver bullet here. Even if fork choice can consider the *potential* just/fin of a block (i.e., the "fix"), I think we're still open to the scenario where one chain contains more attestations and can therefore justify in the next epoch whilst the other chain can't.

sync/optimistic.md
paulhauner: Ooh, good catch! Fixed in be4319a0

sync/optimistic.md
paulhauner: Good idea. Changed to `is_optimistic_candidate_block` in 24947be7.

sync/optimistic.md
paulhauner: Nice one, addressed in bfe41725

sync/optimistic.md
paulhauner: Addressed in 0ec61bd1

sync/optimistic.md
paulhauner: > Have we considered how to do this when TERMINAL_BLOCK_HASH is set?

Very good question. I considered this and added some helpful (but non-substantiative) changes in 50b236e5.

> maybe add "i.e. if pow_block is one and pow_parent is None, skip all assertions in validate_merge_block." or something to that tune

After considering `TERMINAL_BLOCK_HASH`, I think we do need *some* assertions in `validate_merge_block`. My changes in 50b236e5 might help to clarify what was not clear to you?

One option for making this clearer is to break `validate_merge_block` into two functions; one that checks `TERMINAL_BLOCK_HASH` things and another that calls out to the EL. Then, it would be simple for us to state that the EL function should always return some static value. I'm happy to try this if it's appealing to you.

paulhauner:
I believe I have addressed all comments from @ajsutton and @djrtwo. Thanks for your review!

sync/optimistic.md
mkalinin: Assuming adversary power is 1/3 and 32 honest proposals is enough to justify a checkpoint, the probability of *not* justifying a checkpoint within 4 epochs is `1.11**-16` according to binomial distribution.

In our case it means that should an adversary be so lucky to get a chance to poison the fork choice the chance of successful attack with 4 epochs timeout is around 0.

1/3 power here makes a lot of sense as this is a liveness attack and Casper FFG has 1/3 honest assumption to maintain liveness. I am not feeling that strong about 32 honest proposals, but it seems that this is more than enough in a good networking conditions as the mainnet always justifies in an epoch.

So, I would say that we have a pretty strong reasoning behind this fix and its parameter.

Also, worth noting that an optimistic node justifying epoch `N` by applying an honest chain *prior to* applying malicious chain which justifies the same epoch `N` is enough to prevent poisoning. Justifying a higher epoch by an honest chain is not necessarily needed.

specs/bellatrix/p2p-interface.md
djrtwo: yeah, I think I likely made a bad decision quite a while ago on the choice of `if not positive_condition: REJECT/IGNORE`. Might have been saner as `if negative_condition: REJECT/IGNORE`...

sync/optimistic.md
djrtwo: works for me

sync/optimistic.md
djrtwo: Got it.

concrete proposal coming soon -- cc @adiasg

sync/optimistic.md
djrtwo: this should be fine

djrtwo (approved):
excellent!

terencechain:
With the new status code (`ACCEPTED` and `IGNORED`) introduced in https://github.com/ethereum/execution-apis/pull/165, should we update them here as well?

paulhauner:
> With the new status code (`ACCEPTED` and `IGNORED`) introduced in [ethereum/execution-apis#165](https://github.com/ethereum/execution-apis/pull/165), should we update them here as well?

Perhaps @djrtwo or @mkalinin have an opinion on if https://github.com/ethereum/execution-apis/pull/165 will merge before this PR? I don't mind either way.

djrtwo:
Let's merge this now and patch the semantics in an updated PR.

We're aiming to merge and release the engine api changes tomorrow (1/25).
@paulhauner , can you cut a minor PR with the semantics changes to match?

PR #2772, opened 2021-12-22
bump VERSION.txt to 1.1.7

hwwhww:


CarlBeek (approved):
I had a long debate around whether this was correct to include, but ultimately decided it was worth the obvious tradeoffs.

rsalvo77 (commented):
:)

PR #2776, opened 2021-12-23
Set RNG seed in `test_get_head.py`

hwwhww:
### Issue

The RNG seed was unset such that the test vector result was not determined.

### How did I fix it

Set a seed!


ralexstokes (approved):
nice! fixing the `rng` looks good....

i'm suspicious we need to go thru the `hex` conversion when making the random `graffiti` but that's out of scope for this PR so i'll just leave it for now

PR #2784, opened 2021-12-29
Add `CONFIG_NAME` to configs

arnetheduck:
Runtime configurations apply to a certain network and the name of that
network is useful for humans such that they can talk about it.

Some of the existing configs already include a `CONFIG_NAME` toggle -
might as well add it here as well and avoid some confusion - this name
above all becomes useful in the beacon API.

By extension, the `CONFIG_NAME` config will appear in the beacon api as
a result of being defined here.

### DIFF
--- a/configs/mainnet.yaml
+++ b/configs/mainnet.yaml
@@ -3,6 +3,13 @@
 # Extends the mainnet preset
 PRESET_BASE: 'mainnet'
 
+# Free-form short name of the network that this configuration applies to - known
+# canonical network names include:
+# * 'mainnet' - there can be only one
+# * 'prater' - testnet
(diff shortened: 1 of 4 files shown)


hwwhww (approved):
It was removed when we had config file per fork per network, e.g., `mainnet/phase0.yaml`, `mainnet/altair.yaml`, etc. (#2332)

I think it's okay to add it back since now we only have one config file per network, but let's leave this issue open for a couple days in case there is any objection.

ajsutton:
I think Teku has added this to all our network config files anyway to preserve backwards compatibility of the config REST API endpoint.

djrtwo (approved):

dapplion (approved):

PR #2787, opened 2022-01-03
add some non-empty extra_data tests

djrtwo:
Nimbus had a consensus split on kintsugi due to a non-empty `extra_data` field in the execution payload.
This was fixed here -- https://github.com/status-im/nimbus-eth2/pull/3228

This PR adds a couple of tests to ensure that `payload.extra_data` is copied over to `latest_payload_header.extra_data`


hwwhww (approved):
lgtm

PR #2790, opened 2022-01-06
Add 3 new invalid test cases

asanso:
The 3 test cases;

1. first offset starts wrongly into the fixed position
2. last variable part overflowing the max size
3. last variable part has a wrong number f bytes

### DIFF
--- a/tests/generators/ssz_generic/ssz_container.py
+++ b/tests/generators/ssz_generic/ssz_container.py
@@ -105,7 +105,7 @@ def invalid_cases():
                          RandomizationMode.mode_one_count,
                          RandomizationMode.mode_max_count]:
                 if len(offsets) != 0:
-                    for index,offset_index in enumerate(offsets):
+                    for index, offset_index in enumerate(offsets):
                         yield f'{name}_{mode.to_name()}_offset_{offset_index}_plus_one', \
                               invalid_test_case(lambda: mod_offset(
                                   b=serialize(container_case_fn(rng, mode, typ)),
@@ -120,7 +120,7 @@ def invalid_cases():
                               ))
                         if index == 0:
                             yield f'{name}_{mode.to_name()}_first offset_{offset_index}_minus_one', \
-                              invalid_test_case(lambda: mod_offset(
+                                invalid_test_case(lambda: mod_offset(
                                   b=serialize(container_case_fn(rng, mode, typ)),
                                   offset_index=offset_index,
                                   change=lambda x: x - 1
@@ -129,9 +129,9 @@ def invalid_cases():
                             serialized = serialize(container_case_fn(rng, mode, typ))
                             serialized = serialized + serialized[0:2]
                             yield f'{name}_{mode.to_name()}_last offset_{offset_index}_overflow', \
-                                invalid_test_case(lambda:serialized)
+                                invalid_test_case(lambda: serialized)
                         if mode == RandomizationMode.mode_one_count:
                             serialized = serialize(container_case_fn(rng, mode, typ))
                             serialized = serialized + serialized[0:1]
                             yield f'{name}_{mode.to_name()}_last offset_{offset_index}_wrong_byte_length', \
-                                invalid_test_case(lambda:serialized)
+                                invalid_test_case(lambda: serialized)
--- a/tests/generators/ssz_generic/ssz_container.py
+++ b/tests/generators/ssz_generic/ssz_container.py
@@ -119,7 +119,7 @@ def invalid_cases():
                                   change=lambda x: 0
                               ))
                         if index == 0:
-                            yield f'{name}_{mode.to_name()}_first offset_{offset_index}_minus_one', \
+                            yield f'{name}_{mode.to_name()}_offset_{offset_index}_minus_one', \
                                 invalid_test_case(lambda: mod_offset(
                                   b=serialize(container_case_fn(rng, mode, typ)),
                                   offset_index=offset_index,
--- a/tests/generators/ssz_generic/ssz_container.py
+++ b/tests/generators/ssz_generic/ssz_container.py
@@ -104,7 +104,7 @@ def invalid_cases():
                          RandomizationMode.mode_nil_count,
                          RandomizationMode.mode_one_count,
                          RandomizationMode.mode_max_count]:
-                for offset_index in offsets:
+                for index, offset_index in enumerate(offsets):
                     yield f'{name}_{mode.to_name()}_offset_{offset_index}_plus_one', \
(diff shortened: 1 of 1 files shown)


tests/generators/ssz_generic/ssz_container.py
```diff
@@ -105,7 +105,7 @@ def invalid_cases():
                          RandomizationMode.mode_one_count,
                          RandomizationMode.mode_max_count]:
                 if len(offsets) != 0:
```
ralexstokes, line 107: unrelated but we should be able to remove this line and outdent the body
```suggestion
            if len(offsets) != 0:
```

tests/generators/ssz_generic/ssz_container.py
```diff
@@ -118,3 +118,20 @@ def invalid_cases():
                                   offset_index=offset_index,
                                   change=lambda x: 0
                               ))
+                        if index == 0:
+                            yield f'{name}_{mode.to_name()}_first offset_{offset_index}_minus_one', \
```
ralexstokes, line 122: let's do "first offset" or just the offset index, and also i think this string should have no spaces

```suggestion
                            yield f'{name}_{mode.to_name()}_offset_{offset_index}_minus_one', \
```

tests/generators/ssz_generic/ssz_container.py
```diff
@@ -118,3 +118,20 @@ def invalid_cases():
...
+                                invalid_test_case(lambda: mod_offset(
+                                  b=serialize(container_case_fn(rng, mode, typ)),
+                                  offset_index=offset_index,
+                                  change=lambda x: x - 1
+                                ))
+                        if mode == RandomizationMode.mode_max_count:
+                            serialized = serialize(container_case_fn(rng, mode, typ))
+                            serialized = serialized + serialized[0:2]
```
ralexstokes, line 130: nitpick:
```suggestion
                            serialized = serialized + serialized[:2]
```

tests/generators/ssz_generic/ssz_container.py
```diff
@@ -118,3 +118,20 @@ def invalid_cases():
...
+                            yield f'{name}_{mode.to_name()}_last offset_{offset_index}_overflow', \
```
ralexstokes, line 131:
```suggestion
                            yield f'{name}_{mode.to_name()}_last_offset_{offset_index}_overflow', \
```

tests/generators/ssz_generic/ssz_container.py
```diff
@@ -118,3 +118,20 @@ def invalid_cases():
...
+                                invalid_test_case(lambda: serialized)
+                        if mode == RandomizationMode.mode_one_count:
+                            serialized = serialize(container_case_fn(rng, mode, typ))
+                            serialized = serialized + serialized[0:1]
```
ralexstokes, line 135: nitpick:
```suggestion
                            serialized = serialized + serialized[:1]
```

tests/generators/ssz_generic/ssz_container.py
```diff
@@ -118,3 +118,20 @@ def invalid_cases():
...
+                            yield f'{name}_{mode.to_name()}_last offset_{offset_index}_wrong_byte_length', \
```
ralexstokes, line 136:
```suggestion
                            yield f'{name}_{mode.to_name()}_last_offset_{offset_index}_wrong_byte_length', \
```

ralexstokes (commented):
always great to see more test coverage!

i am wondering how the two last tests you wrote differ in terms of SSZ functionality they are touching, they both seem to just add extra data to the end which should yield a deserialization error but not necessarily because the offsets are incorrect (and that is what this part of test gen code suggests it is doing

asanso:
> always great to see more test coverage!
> 
> i am wondering how the two last tests you wrote differ in terms of SSZ functionality they are touching, they both seem to just add extra data to the end which should yield a deserialization error but not necessarily because the offsets are incorrect (and that is what this part of test gen code suggests it is doing

Yes you are correct the last 2 tests are not related to invalid offset but are still invalid tests nevertheless, IMHO the more invalid case we have, the better.

tests/generators/ssz_generic/ssz_container.py
ralexstokes: fixed in #2791.

tests/generators/ssz_generic/ssz_container.py
```diff
@@ -118,3 +118,20 @@ def invalid_cases():
...
+                        if mode == RandomizationMode.mode_max_count:
+                            serialized = serialize(container_case_fn(rng, mode, typ))
+                            serialized = serialized + serialized[:2]
+                            yield f'{name}_{mode.to_name()}_last_offset_{offset_index}_overflow', \
+                                invalid_test_case(lambda: serialized)
+                        if mode == RandomizationMode.mode_one_count:
+                            serialized = serialize(container_case_fn(rng, mode, typ))
+                            serialized = serialized + serialized[:1]
+                            yield f'{name}_{mode.to_name()}_last_offset_{offset_index}_wrong_byte_length', \
+                                invalid_test_case(lambda: serialized)
```
ralexstokes, line 137: so any particular reason to only add 3 more test types? and not the full combination of `(mode, offset_index)`?

as written, we can probably just pull these three cases out of this loop and specify by hand

if we keep them in the loop then why not just let the loop run and do its thing?

ralexstokes (commented):
nice updates!

still want to understand your thinking more on structure

tests/generators/ssz_generic/ssz_container.py
hwwhww, line 107: The two `if len(offsets) != 0` in code blocks look like some wrong logic:
```python
        if len(offsets) != 0:
            # Note: there are many more ways to have invalid offsets,
            # these are just example to get clients started looking into hardening ssz.
            for mode in [RandomizationMode.mode_random,
                         RandomizationMode.mode_nil_count,
                         RandomizationMode.mode_one_count,
                         RandomizationMode.mode_max_count]:
                if len(offsets) != 0:
                   ...
``` 

We may have to rewrite the `RandomizationMode` usages.

hwwhww: Ahh, just found it was addressed #2791 in a minute.

tests/generators/ssz_generic/ssz_container.py
asanso: >so any particular reason to only add 3 more test types? and not the full combination of

* test 1. support all the test types.
* test 2. really targets  specifically the `mode_max_count` because the overflow
* test 3. targets  specifically the `mode_one_count` because the byte disparity

PR #2791, opened 2022-01-06
remove unnecessary conditional in ssz generic test gen code

ralexstokes:
fixing something i noticed here: https://github.com/ethereum/consensus-specs/pull/2790/files#r779729260


tests/generators/ssz_generic/ssz_container.py
```diff
@@ -104,17 +104,16 @@ def invalid_cases():
                          RandomizationMode.mode_nil_count,
                          RandomizationMode.mode_one_count,
                          RandomizationMode.mode_max_count]:
-                if len(offsets) != 0:
```
ralexstokes, line 107: `len(offsets) == 0` always as we already checked to enter into this flow.

and we do not mutate them while iterating over the hardcoded list of modes...

hwwhww (approved):

PR #2792, opened 2022-01-07
WIP implementation of Danksharding

dankrad:
https://notes.ethereum.org/3t5cbN7cQnG6VxjgNMKymA


specs/sharding/beacon-chain.md
```diff
@@ -103,460 +101,342 @@ The following values are (non-configurable) constants used throughout the specif
...
-TODO: `PARTICIPATION_FLAG_WEIGHTS` backwards-compatibility is difficult, depends on usage.
-
-| Name | Value |
-| - | - |
-| `PARTICIPATION_FLAG_WEIGHTS` | `[TIMELY_SOURCE_WEIGHT, TIMELY_TARGET_WEIGHT, TIMELY_HEAD_WEIGHT, TIMELY_SHARD_WEIGHT]` |
-
-### Participation flag indices
+| `MAX_SHARDS` | `uint64(2**12)` (= 4,096) | Theoretical max shard count (used to determine data structure sizes) |
+| `ACTIVE_SHARDS` | `uint64(2**8)` (= 256) | Initial shard count |
+| `TARGET_SHARDS` | `uint64(ACTIVE_SHARDS // 2)` (= 256) | Initial shard count |
```
vbuterin, line 115:
```suggestion
| `TARGET_SHARDS` | `uint64(ACTIVE_SHARDS // 2)` (= 128) | Initial shard count |
```

specs/sharding/beacon-chain.md
```diff
@@ -103,460 +101,342 @@ The following values are (non-configurable) constants used throughout the specif
...
+| - | - | :-: | :-: |
+| `SLOTS_PER_EPOCH` | `uint64(2**6)` (= 64) | slots | 8:32 minutes |
 
 ### Shard blob samples
 
 | Name | Value | Notes |
 | - | - | - |
-| `MAX_SAMPLES_PER_BLOB` | `uint64(2**11)` (= 2,048) | 248 * 2,048 = 507,904 bytes |
-| `TARGET_SAMPLES_PER_BLOB` | `uint64(2**10)` (= 1,024) | 248 * 1,024 = 253,952 bytes |
+| `SAMPLES_PER_BLOB` | `uint64(2**9)` (= 2,048) | 248 * 2,048 = 507,904 bytes |
```
vbuterin, line 129: @dankrad I thought we're doing 128 kB per blob?

specs/sharding/beacon-chain.md
```diff
@@ -103,460 +101,342 @@ The following values are (non-configurable) constants used throughout the specif
...
 
-## Updated containers
-
-The following containers have updated definitions to support Sharding.
+### Time parameters
 
-### `AttestationData`
+| Name | Value | Unit | Duration |
+| - | - | :-: | :-: |
+| `SECONDS_PER_SLOT` | `uint64(8)` | seconds | 8 seconds |
```
vbuterin, line 148: Yay!! Looking forward to ethereum time becoming properly powers-of-two based :)

specs/sharding/beacon-chain.md
```diff
@@ -103,460 +101,342 @@ The following values are (non-configurable) constants used throughout the specif
...
 
-```python
-class BeaconBlockBody(bellatrix.BeaconBlockBody):  # [extends Bellatrix block body]
-    shard_proposer_slashings: List[ShardProposerSlashing, MAX_SHARD_PROPOSER_SLASHINGS]
-    shard_headers: List[SignedShardBlobHeader, MAX_SHARDS * MAX_SHARD_HEADERS_PER_SHARD]
-```
+The following containers have updated definitions to support Sharding.
 
-### `BeaconState`
+### `IntermediateBlockBid`
```
vbuterin, line 155: Surely this is a new container and not an updated container?

specs/sharding/beacon-chain.md
```diff
@@ -103,460 +101,342 @@ The following values are (non-configurable) constants used throughout the specif
...
-    shard_sample_price: uint64
-```
+class IntermediateBlockBid(Container):
+    execution_payload_root: Root
 
-## New containers
+    sharded_data_commitment: Root # Root of the sharded data (only data, not beacon/intermediate block commitments)
 
-### `Builder`
+    sharded_data_commitments: uint64 # Count of sharded data commitments
```
vbuterin, line 163: Rename the first one to `sharded_data_commitment_list_root` and the second `sharded_data_commitment_count` perhaps?

specs/sharding/beacon-chain.md
```diff
@@ -103,460 +101,342 @@ The following values are (non-configurable) constants used throughout the specif
...
-class AttestedDataCommitment(Container):
-    # KZG10 commitment to the data, and length
-    commitment: DataCommitment
-    # hash_tree_root of the ShardBlobHeader (stored so that attestations can be checked against it)
-    root: Root
-    # The proposer who included the shard-header
-    includer_index: ValidatorIndex
-```
+class ShardedCommitmentsContainer(Container):
+    sharded_commitments: List[KZGCommitment, 2 * MAX_SHARDS]
```
vbuterin, line 185: Why 2x?

specs/sharding/beacon-chain.md
dankrad: Actually this number is currently irrelevant as no EIP1559 here yet. Removing

specs/sharding/beacon-chain.md
dankrad: Correct, the comment was wrong.

specs/sharding/beacon-chain.md
dankrad: That's the polynomial extension

specs/sharding/beacon-chain.md
```diff
@@ -103,460 +101,342 @@ The following values are (non-configurable) constants used throughout the specif
...
 ```
 
-### `ShardBlob`
-
-`ShardBlobBody` wrapped with the header data that is unique to the shard blob proposal.
+### `IntermediateBlockHeader`
 
 ```python
-class ShardBlob(Container):
+class IntermediateBlockHeader(Container):
```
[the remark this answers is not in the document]
dankrad: Good question! I guess it will also require a lot of restructuring, because all the operation processing assumes that there is a proposer. But some of that we will have to change for attestation processing anyway I guess.

specs/sharding/beacon-chain.md
```diff
@@ -103,460 +101,342 @@ The following values are (non-configurable) constants used throughout the specif
...
 ```python
-def compute_updated_sample_price(prev_price: Gwei, samples_length: uint64, active_shards: uint64) -> Gwei:
-    adjustment_quotient = active_shards * SLOTS_PER_EPOCH * SAMPLE_PRICE_ADJUSTMENT_COEFFICIENT
-    if samples_length > TARGET_SAMPLES_PER_BLOB:
-        delta = max(1, prev_price * (samples_length - TARGET_SAMPLES_PER_BLOB) // TARGET_SAMPLES_PER_BLOB // adjustment_quotient)
-        return min(prev_price + delta, MAX_SAMPLE_PRICE)
-    else:
-        delta = max(1, prev_price * (TARGET_SAMPLES_PER_BLOB - samples_length) // TARGET_SAMPLES_PER_BLOB // adjustment_quotient)
-        return max(prev_price, MIN_SAMPLE_PRICE + delta) - delta
+def verify_degree_proof(commitment: KZGCommitment, degree: uint64, proof: KZGCommitment):
```
vbuterin, line 298: Better for formatting for the degree (and maybe rename it `degree_cap`) to be the number that the polynomial degree must be less than (so a zero polynomial would satisfy a degree cap of 0). That way the variable can be always non-negative.

specs/sharding/beacon-chain.md
```diff
@@ -826,63 +656,3 @@ def process_epoch(state: BeaconState) -> None:
     process_participation_flag_updates(state)
     process_sync_committee_updates(state)
 ```
```
Nashatyrev, line 658: Epoch transition looks equal to the Altair. Should be removed from here?

specs/sharding/beacon-chain.md
```diff
@@ -580,226 +496,144 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
     for_ops(body.attestations, process_attestation)
     for_ops(body.deposits, process_deposit)
     for_ops(body.voluntary_exits, process_voluntary_exit)
+```
```
Nashatyrev, line 499: Shouldn't the previous sharding code be removed from `process_operations`? 
If yes then the `process_operations` would be equal to `phase0` and could be removed as well
```py
    # New shard proposer slashing processing
    for_ops(body.shard_proposer_slashings, process_shard_proposer_slashing)

    # Limit is dynamic: based on active shard count
    assert len(body.shard_headers) <= MAX_SHARD_HEADERS_PER_SHARD * get_active_shard_count(state, get_current_epoch(state))
    for_ops(body.shard_headers, process_shard_header)
```

specs/sharding/beacon-chain.md
```diff
@@ -580,226 +496,144 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
 ```
 
-##### Extended Attestation processing
+#### Intermediate Block Bid Commitment
 
 ```python
-def process_attestation(state: BeaconState, attestation: Attestation) -> None:
-    altair.process_attestation(state, attestation)
-    process_attested_shard_work(state, attestation)
+def process_intermediate_block_bid_commitment(state: BeaconState, body: IntermediateBlockBody) -> None:
```
Nashatyrev, line 513: Is any state modifications supposed be added here later? If not then may be the better function name would be `verify_intermediate_block_bid_commitment`

specs/sharding/beacon-chain.md
```diff
@@ -580,226 +496,144 @@ def process_operations(state: BeaconState, body: BeaconBlockBody) -> None:
...
+#### Intermediate Block Operations
+
+```python
+def process_intermediate_block_attestations(state: BeaconState, body: IntermediateBlockBody) -> None:
 
-    # TODO: to avoid parallel shards racing, and avoid inclusion-order problems,
-    #  update the fee price per slot, instead of per header.
-    # state.shard_sample_price = compute_updated_sample_price(state.shard_sample_price, ?, shard_count)
+    for attestation in block.body.attestations:
+    process_attestation(state, block.body.attestation)
```
Nashatyrev, line 507: Should it be regular processing of block attestations? 
```suggestion
        process_attestation(state, attestation)
```

specs/sharding/beacon-chain.md
```diff
@@ -103,460 +101,377 @@ The following values are (non-configurable) constants used throughout the specif
...
-    # Proof that the degree < commitment.samples_count * POINTS_PER_SAMPLE
-    degree_proof: BLSCommitment
-    # The actual data. Should match the commitment and degree proof.
-    data: List[BLSPoint, POINTS_PER_SAMPLE * MAX_SAMPLES_PER_BLOB]
-    # fee payment fields (EIP 1559 like)
-    # TODO: express in MWei instead?
-    max_priority_fee_per_sample: Gwei
-    max_fee_per_sample: Gwei
+class IntermediateBlockBody(phase0.BeaconBlockBody):
+    attestation: Attestation
```
Nashatyrev, line 196: Could you please comment what is this field for?

Nashatyrev (commented):
Some minor questions and cleanup suggestions

specs/sharding/beacon-chain.md
```diff
@@ -103,460 +101,377 @@ The following values are (non-configurable) constants used throughout the specif
...
-    commitment: DataCommitment
-    # Proof that the degree < commitment.samples_count * POINTS_PER_SAMPLE
-    degree_proof: BLSCommitment
-    # The actual data. Should match the commitment and degree proof.
-    data: List[BLSPoint, POINTS_PER_SAMPLE * MAX_SAMPLES_PER_BLOB]
-    # fee payment fields (EIP 1559 like)
-    # TODO: express in MWei instead?
-    max_priority_fee_per_sample: Gwei
-    max_fee_per_sample: Gwei
+class IntermediateBlockBody(phase0.BeaconBlockBody):
```
Nashatyrev, line 195: You are deriving from `phase0` block body and adding `execution_payload`. Does it mean you just want to exclude `sync_aggregate` from `IntermediateBlockBody`?

Nashatyrev: I.e. why not the following class: 
```py
class IntermediateBlockBody(bellatrix.BeaconBlockBody):
    attestation: Attestation
    sharded_commitments_container: ShardedCommitmentsContainer
```

specs/sharding/beacon-chain.md
```diff
@@ -103,460 +101,377 @@ The following values are (non-configurable) constants used throughout the specif
 | - | - | - |
 | `PRIMITIVE_ROOT_OF_UNITY` | `7` | Primitive root of unity of the BLS12_381 (inner) modulus |
 | `DATA_AVAILABILITY_INVERSE_CODING_RATE` | `2**1` (= 2) | Factor by which samples are extended for data availability encoding |
-| `POINTS_PER_SAMPLE` | `uint64(2**3)` (= 8) | 31 * 8 = 248 bytes |
+| `POINTS_PER_SAMPLE` | `uint64(2**4)` (= 16) | 31 * 16 = 496 bytes |
```
Nashatyrev, line 104: Does it make sense to rename to `FIELD_ELEMENTS_PER_SAMPLE` (following to `BLSPoint` => `BLSFieldElement` renaming)

specs/sharding/beacon-chain.md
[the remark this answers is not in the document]
dankrad: So after discussion on last call, I think it makes most sense to restructure:
 * Require block builders to be validators
 * Intermediate blocks and beacon blocks can then be the same data structure and share the same block processing

specs/sharding/beacon-chain.md
dankrad: We need to include attestations in the intermediate block in order to incentivise early attestations to the beacon block

specs/sharding/beacon-chain.md
dankrad: Yeah good idea

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +101,481 @@ The following values are (non-configurable) constants used throughout the specif
...
+#### `hash_to_field`
 
 ```python
-class SignedShardBlobHeader(Container):
-    message: ShardBlobHeader
-    # Signature by builder.
-    # Once accepted by proposer, the signatures is the aggregate of both.
-    signature: BLSSignature
+def hash_to_field(x: Container):
+    return int.from_bytes(hash_tree_root(x), "little") % MODULUS
```
Nashatyrev, line 231: The resulting values would not have uniform distribution: lower values are more likely to occur. 
Could it potentially have any security impact? Or collision resistance is sufficient property?

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,516 @@ The following values are (non-configurable) constants used throughout the specif
...
-        else:
-            state.shard_buffer[buffer_index][attestation_shard].status.change(
-                selector=SHARD_WORK_CONFIRMED,
-                value=pending_header.attested,
-            )
-```
+        assert intermediate_block_bid.execution_payload_root == hash_tree_root(block.body.execution_payload)
 
-##### `process_shard_header`
+        assert intermediate_block_bid.sharded_data_commitment_number == block.body.sharded_commitments_container.included_sharded_data_commitments
```
terencechain, line 487:
```suggestion
        assert intermediate_block_bid.sharded_data_commitment_count == block.body.sharded_commitments_container.included_sharded_data_commitments
```

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
+    
+    # Block builders use an Eth1 address -- need signature as
+    # block builder fees and data gas base fees will be charged to this address
+	signature_y_parity: bool
+	signature_r: uint256
+	signature_s: uint256    
 ```
 
-### `AttestedDataCommitment`
+#### `IntermediateBlockBidWithRecipientAddress`
```
vbuterin, line 170: What do we actually gain from this being a separate class? Why not just stick the ethereum address into the bid itself?

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
-    body_root_1: Root
-    body_root_2: Root
-    signature_1: BLSSignature
-    signature_2: BLSSignature
+def compute_powers(x: BLSFieldElement, n: uint64) -> List[BLSFieldElement]:
+    current_power = 1
+    powers = []
+    for i in range(n):
+        powers.append(BLSFieldElement(current_power))
+        current_power = current_power * int(x) % MODULUS
```
vbuterin, line 258:
```suggestion
        current_power = current_power * int(x) % MODULUS
    return powers
```

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
+#### `interpolate_poly`
 
 ```python
-def next_power_of_two(x: int) -> int:
-    return 2 ** ((x - 1).bit_length())
+def interpolate_poly(xs: List[BLSFieldElement], ys: List[BLSFieldElement]):
+    """
+    Lagrange interpolation
+    """
+    # TODO!
```
vbuterin, line 280: Maybe copy or take ideas from https://github.com/ethereum/research/blob/master/polynomial_reconstruction/poly_utils.py#L117 ?

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
-    adjustment_quotient = active_shards * SLOTS_PER_EPOCH * SAMPLE_PRICE_ADJUSTMENT_COEFFICIENT
-    if samples_length > TARGET_SAMPLES_PER_BLOB:
-        delta = max(1, prev_price * (samples_length - TARGET_SAMPLES_PER_BLOB) // TARGET_SAMPLES_PER_BLOB // adjustment_quotient)
-        return min(prev_price + delta, MAX_SAMPLE_PRICE)
-    else:
-        delta = max(1, prev_price * (TARGET_SAMPLES_PER_BLOB - samples_length) // TARGET_SAMPLES_PER_BLOB // adjustment_quotient)
-        return max(prev_price, MIN_SAMPLE_PRICE + delta) - delta
+def verify_degree_proof(commitment: KZGCommitment, degree: uint64, proof: KZGCommitment):
+    """
+    Verifies that the commitment is of polynomial degree <= degree.
```
vbuterin, line 301:
```suggestion
    Verifies that the commitment is of polynomial degree < degree. 
```

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
 ```python
-def compute_updated_sample_price(prev_price: Gwei, samples_length: uint64, active_shards: uint64) -> Gwei:
-    adjustment_quotient = active_shards * SLOTS_PER_EPOCH * SAMPLE_PRICE_ADJUSTMENT_COEFFICIENT
-    if samples_length > TARGET_SAMPLES_PER_BLOB:
-        delta = max(1, prev_price * (samples_length - TARGET_SAMPLES_PER_BLOB) // TARGET_SAMPLES_PER_BLOB // adjustment_quotient)
-        return min(prev_price + delta, MAX_SAMPLE_PRICE)
-    else:
-        delta = max(1, prev_price * (TARGET_SAMPLES_PER_BLOB - samples_length) // TARGET_SAMPLES_PER_BLOB // adjustment_quotient)
-        return max(prev_price, MIN_SAMPLE_PRICE + delta) - delta
+def verify_degree_proof(commitment: KZGCommitment, degree: uint64, proof: KZGCommitment):
```
vbuterin, line 299: Maybe rename `degree` -> `degree_bound`?

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
-    if epoch == get_current_epoch(state):
-        epoch_participation = state.current_epoch_participation
-    else:
-        epoch_participation = state.previous_epoch_participation
-    for bit, index in zip(bits, full_committee):
-        if bit:
-            epoch_participation[index] = add_flag(epoch_participation[index], flag_index)
-```
+def roots_of_unity(order: uint64) -> List[BLSFieldElement]:
+    r = []
```
vbuterin, line 325: Stylistic change: `r` -> `roots`?

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
 ```python
-def get_committee_count_per_slot(state: BeaconState, epoch: Epoch) -> uint64:
-    """
-    Return the number of committees in each slot for the given ``epoch``.
-    """
-    return max(uint64(1), min(
-        get_active_shard_count(state, epoch),
-        uint64(len(get_active_validator_indices(state, epoch))) // SLOTS_PER_EPOCH // TARGET_COMMITTEE_SIZE,
-    ))
+def modular_inverse(a):
```
vbuterin, line 338:
```suggestion
def modular_inverse(a):
    """
    Return a number `v` such that `a * v % MODULUS == 1`
    """
```

vbuterin: Also consider renaming `modular_inverse` to `bls_modular_inverse` and `MODULUS` to `BLS_MODULUS`?

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
+        lm, low, hm, high = nm, new, lm, low
+    return lm % MODULUS
 ```
 
-#### `get_active_shard_count`
+#### `eval_poly_at`
 
 ```python
-def get_active_shard_count(state: BeaconState, epoch: Epoch) -> uint64:
+def eval_poly_at(poly: List[BLSFieldElement], x: BLSFieldElement) -> BLSFieldElement:
```
vbuterin, line 352: I would recommend changing the function name to reflect that this is taking as input evaluation form and not coeffs.

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
+        for i in range(d + 1):
+            m = 1
+            for w in roots[:i] + roots[i+1:d + 1]:
+                m = m * (z - w) % MODULUS
+            r = (r + M) % MODULUS
+        return r
+
+    coefs = []
+    for i in range(K):
+        coefs.append( - (r_to_K - 1) * modular_inverse(K * roots[i * (K - 1) % K] * (r - roots[i])) % MODULUS)
```
vbuterin, line 421: Maybe to make it more understandable, make a variable `left_coeffs` and `right_coeffs`, and then at the `elliptic_curve_lincomb` line combine them together?

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
+    for i in range(d + 1):
+        coefs[i] = (coefs[i] + B(r) * modular_inverse(Bprime(r) * (r - roots[i]))) % MODULUS
+    
+    assert elliptic_curve_lincomb(commitments, coefs) == bls.Z1()
+```
+
+#### `vector_lincomb`
```
vbuterin, line 428: Function should be above `low_degree_check` to preserve topological order

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
 
-##### Extended Attestation processing
+#### Intermediate Block Bid
 
 ```python
-def process_attestation(state: BeaconState, attestation: Attestation) -> None:
-    altair.process_attestation(state, attestation)
-    process_attested_shard_work(state, attestation)
-```
+def verify_intermediate_block_bid(state: BeaconState, block: BeaconBlock) -> None:
```
vbuterin, line 520: Feels like all these names could be shortened a lot.

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
-    process_slashings_reset(state)
-    process_randao_mixes_reset(state)
-    process_historical_roots_update(state)
-    process_participation_flag_updates(state)
-    process_sync_committee_updates(state)
-```
+        verify_degree_proof(combined_commitment, SAMPLES_PER_BLOB * FIELD_ELEMENTS_PER_SAMPLE, sharded_commitments_container.degree_proof)
 
-#### `process_pending_shard_confirmations`
+        # Verify that the 2*N commitments lie on a degree N-1 polynomial
```
vbuterin, line 562:
```suggestion
        # Verify that the 2*N commitments lie on a degree < N polynomial
```

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
-            committee_work = state.shard_buffer[buffer_index][shard_index]
-            if committee_work.status.selector == SHARD_WORK_PENDING:
-                winning_header = max(committee_work.status.value, key=lambda header: header.weight)
-                if winning_header.attested.commitment == DataCommitment():
-                    committee_work.status.change(selector=SHARD_WORK_UNCONFIRMED, value=None)
-                else:
-                    committee_work.status.change(selector=SHARD_WORK_CONFIRMED, value=winning_header.attested)
-```
+        # Verify that blocks since the last intermediate block have been included
+        blocks_chunked = [block_to_field_elements(ssz_serialize(block)) for block in state.blocks_since_intermediate_block]
```
vbuterin, line 566: I feel like this is going to be a significant source of complexity. Could even just leave it out at first!

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
 
 ```python
-def process_attestation(state: BeaconState, attestation: Attestation) -> None:
-    altair.process_attestation(state, attestation)
-    process_attested_shard_work(state, attestation)
-```
+def verify_intermediate_block_bid(state: BeaconState, block: BeaconBlock) -> None:
+    if is_intermediate_block_slot(block.slot):
+        # Get last intermediate block bid
+        assert state.blocks_since_intermediate_block[-1].body.intermediate_block_bid_with_recipient_address.selector == 1
```
vbuterin, line 523: What happens if the block before the intermediate slot is empty?

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
+        intermediate_block_bid = state.blocks_since_intermediate_block[-1].body.intermediate_block_bid_with_recipient_address.value.intermediate_block_bid
```
asn-d6, line 524: Hmm. `IntermediateBlockBidWithRecipientAddress.intermediate_block_bid` is also a `Union` but is not treated as such here and in the `else` clause.

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
-    # Verify the length by verifying the degree.
-    body_summary = header.body_summary
-    points_count = body_summary.commitment.samples_count * POINTS_PER_SAMPLE
-    if points_count == 0:
-        assert body_summary.degree_proof == G1_SETUP[0]
-    assert (
-        bls.Pairing(body_summary.degree_proof, G2_SETUP[0])
-        == bls.Pairing(body_summary.commitment.point, G2_SETUP[-points_count])
-    )
+        assert intermediate_block_bid.sharded_data_commitment_root == hash_tree_root(block.body.sharded_commitments_container.sharded_commitments[-intermediate_block_bid.sharded_data_commitments:])
```
asn-d6, line 531: I think the index access of `sharded_commitments` might not be right. Should that be `included_sharded_data_commitments`?

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
+        # Verify that blocks since the last intermediate block have been included
+        blocks_chunked = [block_to_field_elements(ssz_serialize(block)) for block in state.blocks_since_intermediate_block]
+        block_vectors = []
+        field_elements_per_blob = SAMPLES_PER_BLOB * FIELD_ELEMENTS_PER_SAMPLE
+        for block_chunked in blocks_chunked:
+            for i in range(0, len(block_chunked), field_elements_per_blob):
+                block_vectors.append(block_chunked[i:i + field_elements_per_blob])
+            
+        number_of_blobs = len(block_vectors)
+        r = hash_to_field([sharded_commitments_container.sharded_commitments[:number_of_blobs], 0])
```
asn-d6, line 574: `hash_to_field()` is expecting a `Container` but you are passing it a python list. I think this might not work.
Maybe you can cast it into a `List` or find another way to do domain separation.

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
-                        attested=AttestedDataCommitment(),
-                        votes=Bitlist[MAX_VALIDATORS_PER_COMMITTEE]([0] * committee_length),
-                        weight=0,
-                        update_slot=slot,
-                    )
-                )
-            )
-        # a shard without committee available defaults to SHARD_WORK_UNCONFIRMED.
-```
+        verify_kzg_proof(combined_commitment, x, y, block_verification_kzg_proof)
```
asn-d6, line 582:
```suggestion
        verify_kzg_proof(combined_commitment, x, y, sharded_commitments_container.block_verification_kzg_proof)
```

specs/sharding/beacon-chain.md
mratsim: A polynomial class would probably make testing and compartimentalizing easier.

However from a spec point of view reduce+lambda and other Python idioms will need an explainer section (and we can explain range as well, in particular spelling that it's inclusive lower bound and exclusive upper bound).

specs/sharding/beacon-chain.md
arnetheduck: The engineering effort to change the slot time is not insignificant - something to consider before settling on a value: because it started out as a constant, it has wiggled its way into many weak and soft dependencies, such as the timing between the various actions of validator clients, beacon nodes and so on, as well as third-party tooling that so far has relied on `slot = (now() - genesis) / 12`.

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
     """
-    Return the source epoch for computing the committee.
+    Slices a block into 31 byte chunks that can fit into field elements
     """
-    source_epoch = Epoch(epoch - epoch % period)
-    if source_epoch >= period:
-        source_epoch -= period  # `period` epochs lookahead
-    return source_epoch
+    sliced_block = [block[i:i + 31] for i in range(0, len(bytes), 31)]
+    return [BLSFieldElement(int.from_bytes(x, "little")) for x in sliced_block]
```
mratsim, line 318: Is 31 to guarantee being in the scalar field of BLS12-381?
Is there a size bound on the length of the list?

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
-def get_committee_count_per_slot(state: BeaconState, epoch: Epoch) -> uint64:
-    """
-    Return the number of committees in each slot for the given ``epoch``.
-    """
-    return max(uint64(1), min(
-        get_active_shard_count(state, epoch),
-        uint64(len(get_active_validator_indices(state, epoch))) // SLOTS_PER_EPOCH // TARGET_COMMITTEE_SIZE,
-    ))
+def modular_inverse(a):
+    assert(a == 0):
```
mratsim, line 339: Typo (assert a != 0).

Otherwise, the hash-to-curve spec enforces 0⁻¹ = 0 (mod p) so it's not a cost to do this here. And traditional modular inversion techniques (Little Fermat Theorem or binary Extended Euclidean Algorithm) will return 0 on input 0.

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
+
+    def Aprime(z):
+        return field_elements_per_blob * pow(z, field_elements_per_blob - 1, MODULUS) 
+
+    r = 0
+    inverses = [modular_inverse(z - x) for z in roots]
+    for i, x in enumerate(inverses):
+        r += f[i] * modular_inverse(Aprime(roots[i])) * x % self.MODULUS
+    r = r * A(x) % self.MODULUS
+    return r
```
mratsim, line 373: The `poly` input is not used?

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
 ```python
-def compute_shard_from_committee_index(state: BeaconState, slot: Slot, index: CommitteeIndex) -> Shard:
-    active_shards = get_active_shard_count(state, compute_epoch_at_slot(slot))
-    assert index < active_shards
-    return Shard((index + get_start_shard(state, slot)) % active_shards)
+def elliptic_curve_lincomb(points: List[KZGCommitment], scalars: List[BLSFieldElement]) -> KZGCommitment:
+    """
+    BLS multiscalar multiplication. This function can be optimized using Pippenger's algorithm and variants. This is a non-optimized implementation.
+    """
+    r = bls.Z1()
```
mratsim, line 449: I assume this is reusing py_ecc idioms.

I'd rather see it spelled `inf_G1()` or `infinity_G1()`

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
+            r = (r + M) % MODULUS
+        return r
+
+    coefs = []
+    for i in range(K):
+        coefs.append( - (r_to_K - 1) * modular_inverse(K * roots[i * (K - 1) % K] * (r - roots[i])) % MODULUS)
+    for i in range(d + 1):
+        coefs[i] = (coefs[i] + B(r) * modular_inverse(Bprime(r) * (r - roots[i]))) % MODULUS
+    
+    assert elliptic_curve_lincomb(commitments, coefs) == bls.Z1()
```
mratsim, line 425: `inf_G1()` or `infinity_G1()`

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +87,558 @@ The following values are (non-configurable) constants used throughout the specif
...
+    sharded_commitments: List[KZGCommitment, 2 * MAX_SHARDS]
 
-Summary version of the `ShardBlobBody`, omitting the data payload, while preserving the data-commitments.
+    # Aggregate degree proof for all sharded_commitments
+    degree_proof: KZGCommitment
 
-The commitments are not further collapsed to a single hash, 
-to avoid an extra network roundtrip between proposer and builder, to include the header on-chain more quickly.
+    # The sizes of the blocks encoded in the commitments (last intermediate and all beacon blocks since)
+    included_block_sizes: List[uint64, MAX_BEACON_BLOCKS_BETWEEN_INTERMEDIATE_BLOCKS + 1]
```
asn-d6, line 188: `MAX_BEACON_BLOCKS_BETWEEN_INTERMEDIATE_BLOCKS` is undefined in this spec (might be OK if it's a PoC).

specs/sharding/beacon-chain.md
dankrad: This is a fair point. @vbuterin what are your thoughts on this?
It may also be safer to change this back to 12s, and adjust all constants (gas limit etc.) accordingly to the new paradigm which only has a payload every 24s.

specs/sharding/beacon-chain.md
dankrad: This is not current anymore. Changed the structure so this discussion is obsolete.

specs/sharding/beacon-chain.md
```diff
@@ -103,786 +101,481 @@ The following values are (non-configurable) constants used throughout the specif
...
-    # Builder of the data, pays data-fee to proposer
-    builder_index: BuilderIndex
-    # Proposer of the shard-blob
-    proposer_index: ValidatorIndex
-    # Blob contents
-    body: ShardBlobBody
+class BeaconBlockBody(bellatrix.BeaconBlockBody):
+    execution_payload: Union[None, ExecutionPayload]
+    sharded_commitments_container: Union[None, ShardedCommitmentsContainer]
+    intermediate_block_bid: Union[None, IntermediateBlockBid]
```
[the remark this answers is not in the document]
dankrad: Aha, yes this is a good idea.

specs/sharding/beacon-chain.md
dankrad: We do need a random oracle, as this is mainly used for Fiat-Shamir constructions. However, I believe the impact of a 1 bit bias is not important. This can be confirmed by the cryptography team.

specs/sharding/beacon-chain.md
dankrad: It simplifies processing. We need to iterate through lists that contain the commitment and the recipient address as we process the payments. But yes it's not a huge difference either way.

specs/sharding/beacon-chain.md
dankrad: I will factor out polynomial operations as a class. However I am not sure about overloading BLS field element basic arithmetic.. For my own use this makes perfect sense to me, but there are probably some clients written in languages that do not support overloading and in that case, forgetting a modular reduction can have very subtle and potentially catastrophic consequences.

For polynomials this would be less of a concern, as bugs resulting from this can be more easily found and will typically cause clear compile time or run time errors.

specs/sharding/beacon-chain.md
dankrad: Yes, it will not provide any benefits until we can fraud proof beacon and execution blocks. My thinking is currently to spec it out (in order to think through the feasibility and have it "ready" for future deployment), but disable it in the first hard fork.

specs/sharding/beacon-chain.md
dankrad: Thought about the polynomial class, but actually there aren't so many operations that this would make a significant difference for readability. In order to make things as clear as possible, I think it's best to just write out polynomial operations as well as field operations

dankrad: Thought about the polynomial class, but actually there aren't so many operations that this would make a significant difference for readability. In order to make things as clear as possible, I think it's best to just write out polynomial operations as well as field operations

specs/sharding/beacon-chain.md
dankrad: I believe in this case we cannot have an intermediate block?

specs/sharding/beacon-chain.md
dankrad: Yes, we can only store 31 bytes in a BLS field element.

specs/sharding/beacon-chain.md
vbuterin: I do feel like 24 seconds is too long for a block time and risks harming usability, and 6 seconds might be too short for a slot, so splitting the pain 50/50 (by making beacon slots 8s and user-experienced slots 16s) really is the best path by a significant margin.

vbuterin:
Notes from a chat with @JustinDrake @dankrad @hwwhww :

* Rename "intermediate block" -> "builder block", and the other kind of block can be a "proposer block" (right now there isn't a good term for non-intermediate blocks)
* Rename "eth1 signature" to "execution signature"
* Add a concept of "hash of a sharded commitment"; only the last 31 bytes should be a hash, the first byte can be a version number. The commitments in the execution layer should be in this version+hash form.
  * There is some inefficiency, because data is repeated (N hashes in the execution layer, 2N commitments in the consensus layer), but now it's only 33% overhead instead of 50%, and removing this requires too much tight coupling between consensus and execution and risks increasing dev time
  * Another good reason to do the version-and-hash format on the execution layer is that in the future, we might switch to a different commitment scheme (eg. STARKed Merkle root), and we want to preserve the format of a commitment in that case
* The `IntermediateBlockBidWithRecipientAddress` class could be removed if we move the `ethereum_address` out into the `IntermediateBlockData`. But this requires the `execution_engine.execute_payload` call to take two lists instead of one. It could nevertheless be a net reduction in complexity; worth exploring but not 100% sure.
* Add a max number of unprocessed beacon blocks included into a builder block's commitment list, and do a FIFO queue to process them.
* Think through the consequences of unprocessed block mechanics for weak subjectivity (maybe no consequences at all if all unprocessed beacon blocks are in the state)
* **Many things could be simplified quite a bit if we add virtual blocks** (auto-generated execution blocks containing zero transactions that get inserted during any builder slot without a real execution block)
* There are high capital requirements to requiring the builder to be an active validator. What if we just require the builder to be a non-exited non-slashed (possibly inactive) validator with at least 1 ETH, to reduce capital intensity while preserving enough slashable ETH for anti-DoS purposes? This would mean that it's possible for the proposer to be inactive; need to check that this is ok (probably is)

How attestation works:

* Each epoch you get assigned to 2 rows and 2 columns, and you only accept a block if those rows and columns are fully available for all slots since the last slot since you attested for (if we add PoC, this would also be the PoC data)
  * This ensures that if a block is unavailable (>50% of rows and >50% of columns missing), >75% of all validators would stop voting for that chain

hwwhww (commented):
A quick scan of the polynomial-commitments.md

dankrad:
On the fork choice rule:

- Current favourite is (block,slot), with equal committees, and a 3/4 proposer boost. (Different committees are less attractive because they lead to favouring the proposer and builder collaborating and publishing the proposer block late to get more MEV)
- A backoff scheme can be based on previously muissing slots. It should work like this:
  - Remove the requirement that the builder block is the next slot (it can be published in any slot as long as there were no beacon blocks in between
  - Require a number of empty slots between beacon blocks (do not require that the beacon blocks are at fixed block heights)
  - Base back-off on %age of missing blocks

specs/sharding/beacon-chain.md
```diff
@@ -101,390 +68,158 @@ The following values are (non-configurable) constants used throughout the specif
...
 ```
 
-### `ShardBlobHeader`
-
-Header version of `ShardBlob`.
+#### `SignedShardSample`
 
 ```python
-class ShardBlobHeader(Container):
+class SignedShardSample(Container):
```
hwwhww, line 176: Is there a `ShardSample`? If not, should we just name it `ShardSample`?

specs/sharding/beacon-chain.md
```diff
@@ -101,390 +68,158 @@ The following values are (non-configurable) constants used throughout the specif
...
+| `ACTIVE_SHARDS` | `uint64(2**8)` (= 256) | Initial shard count |
+| `MAX_BEACON_BLOCKS_BETWEEN_INTERMEDIATE_BLOCKS` | `uint64(2**4)` (= 16) | TODO: Need to define what happens if there were more blocks without intermediate blocks | 
+
+### Time parameters
+
+With the introduction of intermediate blocks the number of slots per epoch is doubled (it counts beacon blocks and intermediate blocks).
+
+| Name | Value | Unit | Duration |
+| - | - | :-: | :-: |
+| `SLOTS_PER_EPOCH` | `uint64(2**6)` (= 64) | slots | 8:32 minutes |
```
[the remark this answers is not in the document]
hwwhww: Another issue is that we have SSZ definition of `BeaconState.eth1_data_votes: List[Eth1Data, EPOCHS_PER_ETH1_VOTING_PERIOD * SLOTS_PER_EPOCH]`. If we change this field, we have to handle this field especially in `sharding/fork.md`.

specs/sharding/beacon-chain.md
```diff
@@ -101,390 +68,158 @@ The following values are (non-configurable) constants used throughout the specif
...
-class ShardBlobReference(Container):
-    slot: Slot
-    shard: Shard
-    # Builder of the data
-    builder_index: BuilderIndex
-    # Proposer of the shard-blob
-    proposer_index: ValidatorIndex
-    # Blob hash-tree-root for slashing reference
-    body_root: Root
+class BeaconState(bellatrix.BeaconState):
```
hwwhww, line 191: Simply inheritance seems incorrect since `SLOTS_PER_EPOCH` changed and then the field `BeaconState.eth1_data_votes: List[Eth1Data, EPOCHS_PER_ETH1_VOTING_PERIOD * SLOTS_PER_EPOCH]` also need to be extended.

specs/sharding/beacon-chain.md
```diff
@@ -101,390 +68,158 @@ The following values are (non-configurable) constants used throughout the specif
...
-def batch_apply_participation_flag(state: BeaconState, bits: Bitlist[MAX_VALIDATORS_PER_COMMITTEE],
-                                   epoch: Epoch, full_committee: Sequence[ValidatorIndex], flag_index: int):
-    if epoch == get_current_epoch(state):
-        epoch_participation = state.current_epoch_participation
-    else:
-        epoch_participation = state.previous_epoch_participation
-    for bit, index in zip(bits, full_committee):
-        if bit:
-            epoch_participation[index] = add_flag(epoch_participation[index], flag_index)
+def is_intermediate_block_slot(slot: Slot):
```
hwwhww, line 217:
```suggestion
def is_intermediate_block_slot(slot: Slot) -> bool:
```

specs/sharding/polynomial-commitments.md
```diff
@@ -0,0 +1,383 @@
...
+    """
+    assert (BLS_MODULUS - 1) % order == 0
+    roots = []
+    root_of_unity = pow(PRIMITIVE_ROOT_OF_UNITY, (BLS_MODULUS - 1) // order, BLS_MODULUS)
+
+    current_root_of_unity = 1
+    for i in range(len(SAMPLES_PER_BLOB * FIELD_ELEMENTS_PER_SAMPLE)):
+        roots.append(current_root_of_unity)
+        current_root_of_unity = current_root_of_unity * root_of_unity % BLS_MODULUS
+    return r
```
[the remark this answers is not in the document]
dankrad: It's actually `roots`. Fixed now.

dankrad:
Need to add previous block root to bid cc: @CarlBeek

dankrad:
- [ ] Explore possibility that builders do not need to be bonded by P2P censoring duplicate blocks
- [ ] p2p spec for bids: only allow buds that increase by a minimum of 1%
- [ ] remove RANDAO from builder blocks

specs/sharding/polynomial-commitments.md
```diff
@@ -0,0 +1,387 @@
...
+    - [`verify_kzg_multiproof`](#verify_kzg_multiproof)
+    - [`verify_degree_proof`](#verify_degree_proof)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+
+## Introduction
+
+This document specifies basic polynomial operations and KZG polynomial commitment operations as they are needed for the sharding specification. The implementations are not optimized for performance, but readability. All practical implementations shoul optimize the polynomial operations, and hints what the best known algorithms for these implementations are are included below.
```
graemecode, line 49: typo "shoul optimize" -> "should optimize"

specs/sharding/beacon-chain.md
```diff
@@ -101,390 +68,160 @@ The following values are (non-configurable) constants used throughout the specif
...
-| `G2_SETUP` | Type `List[G2]`. The G2-side trusted setup `[G, G*s, G*s**2....]` |
-| `ROOT_OF_UNITY` | `pow(PRIMITIVE_ROOT_OF_UNITY, (MODULUS - 1) // int(MAX_SAMPLES_PER_BLOB * POINTS_PER_SAMPLE), MODULUS)` |
-
-### Gwei values
-
-| Name | Value | Unit | Description |
-| - | - | - | - |
-| `MAX_SAMPLE_PRICE` | `Gwei(2**33)` (= 8,589,934,592) | Gwei | Max sample charged for a TARGET-sized shard blob |  
-| `MIN_SAMPLE_PRICE` | `Gwei(2**3)` (= 8) | Gwei | Min sample price charged for a TARGET-sized shard blob |
+| `ROOT_OF_UNITY` | `pow(PRIMITIVE_ROOT_OF_UNITY, (BLS_MODULUS - 1) // int(SAMPLES_PER_BLOB * FIELD_ELEMENTS_PER_SAMPLE), BLS_MODULUS)` |
```
asn-d6, line 107: This seems unused atm.

specs/sharding/polynomial-commitments.md
```diff
@@ -0,0 +1,387 @@
...
+| `BLSFieldElement` | `uint256` | A number `x` in the range `0 <= x < BLS_MODULUS` |
+| `BLSPolynomialByCoefficients` | `List[BLSFieldElement]` | A polynomial over the BLS field, given in coefficient form |
+| `BLSPolynomialByEvaluations` | `List[BLSFieldElement]` | A polynomial over the BLS field, given in evaluation form |
+
+## Helper functions
+
+#### `next_power_of_two`
+
+```python
+def next_power_of_two(x: int) -> int:
```
asn-d6, line 83: This seems to bug out if called with `x=0`:

```
In [24]: next_power_of_two(0)
Out[24]: 2

In [25]: next_power_of_two(1)
Out[25]: 1

In [26]: next_power_of_two(2)
Out[26]: 2
```

Perhaps it will never be called with x=0 but might be worth noting down or asserting for it.

specs/sharding/polynomial-commitments.md
```diff
@@ -0,0 +1,387 @@
...
+## Field operations
+
+### Generic field operations
+
+#### `bls_modular_inverse`
+
+```python
+def bls_modular_inverse(x: BLSFieldElement) -> BLSFieldElement:
+    """
+    Compute the modular inverse of x, i.e. y such that x * y % BLS_MODULUS == 1 and return 0 for x == 0
```
asn-d6, line 118: This function doesn't actually return 0 when called with `x==0`. It returns 1.

specs/sharding/polynomial-commitments.md
```diff
@@ -0,0 +1,387 @@
...
+    """
+    Compute a list of roots of unity for a given order.
+    The order must divide the BLS multiplicative group order, i.e. BLS_MODULUS - 1
+    """
+    assert (BLS_MODULUS - 1) % order == 0
+    roots = []
+    root_of_unity = pow(PRIMITIVE_ROOT_OF_UNITY, (BLS_MODULUS - 1) // order, BLS_MODULUS)
+
+    current_root_of_unity = 1
+    for i in range(len(SAMPLES_PER_BLOB * FIELD_ELEMENTS_PER_SAMPLE)):
```
asn-d6, line 142:
```suggestion
    for i in range(SAMPLES_PER_BLOB * FIELD_ELEMENTS_PER_SAMPLE):
```

specs/sharding/polynomial-commitments.md
```diff
@@ -0,0 +1,387 @@
...
+
+### Field helper functions
+
+#### `compute_powers`
+
+```python
+def compute_powers(x: BLSFieldElement, n: uint64) -> List[BLSFieldElement]:
+    current_power = 1
+    powers = []
+    for i in range(n):
```
asn-d6, line 156:
```suggestion
    for _ in range(n):
```

specs/sharding/polynomial-commitments.md
```diff
@@ -0,0 +1,387 @@
...
+### Elliptic curve helper functoins
+
+#### `elliptic_curve_lincomb`
+
+```python
+def elliptic_curve_lincomb(points: List[KZGCommitment], scalars: List[BLSFieldElement]) -> KZGCommitment:
+    """
+    BLS multiscalar multiplication. This function can be optimized using Pippenger's algorithm and variants. This is a non-optimized implementation.
+    """
+    r = bls.Z1()
```
asn-d6, line 321: I assume `Z1` is the point at infinity in G1, but I didn't find a reference to this somewhere.

specs/sharding/polynomial-commitments.md
```diff
@@ -0,0 +1,387 @@
...
+```
+
+#### `vector_lincomb`
+
+```python
+def vector_lincomb(vectors: List[List[BLSFieldElement]], scalars: List[BLSFieldElement]) -> List[BLSFieldElement]:
+    """
+    Compute a linear combination of field element vectors
+    """
+    r = [0 for i in len(vectors[0])]
```
asn-d6, line 214:
```suggestion
    r = [0]*len(vectors[0])
```

Or use `for i in range(len(...`

djrtwo:
@dankrad is this still relevant or would we instead build full dank-sharding off of 4844 specs?

dankrad:
> @dankrad is this still relevant or would we instead build full dank-sharding off of 4844 specs?

4844 only implements a small subset of this. So it seems a bit unnecessary to throw it away only to dig out the old PR when we want to implement full sharding. I suggest I take a couple of day to process the outstanding comments and then we merge it into the sharding directory?

specs/sharding/polynomial-commitments.md
```diff
@@ -0,0 +1,383 @@
...
+#### `multiply_polynomials`
+
+```python
+def multiply_polynomials(a: BLSPolynomialByCoefficients, b: BLSPolynomialByCoefficients) -> BLSPolynomialByCoefficients:
+    """
+    Multiplies the polynomials `a` and `b` given by their coefficients
+    """
+    r = [0]
+    for power, coef in enumerate(a):
+        summand = [0] * power + [coef * x % BLS_MODULUS for x in b]
```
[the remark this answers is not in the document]
dankrad: This is for the wrong line. Closing.

specs/sharding/beacon-chain.md
```diff
@@ -493,396 +230,186 @@ def get_active_shard_count(state: BeaconState, epoch: Epoch) -> uint64:
...
+
+        # Verify not too many commitments
+        assert len(sharded_commitments_container.sharded_commitments) // 2 <= get_active_shard_count(state, get_current_epoch(state))
+
+        # Verify the degree proof
+        r = hash_to_bls_field(sharded_commitments_container.sharded_commitments, 0)
+        r_powers = compute_powers(r, len(sharded_commitments_container.sharded_commitments))
+        combined_commitment = elliptic_curve_lincomb(sharded_commitments_container.sharded_commitments, r_powers)
+
+        verify_degree_proof(combined_commitment, SAMPLES_PER_BLOB * FIELD_ELEMENTS_PER_SAMPLE, sharded_commitments_container.degree_proof)
```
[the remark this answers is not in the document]
dankrad: Yup, that's correct. Fixed now

specs/sharding/beacon-chain.md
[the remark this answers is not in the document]
dankrad: I am going to close this discussion. I think this is relevant once we want to go to implementation and testing, but that is currently far off and the current code is only a rough documentation of current thinking.

dankrad:
Now ready for merge @djrtwo

djrtwo (approved):

hwwhww:
Does it make sense to rebase/squash the commits a bit? Or even do "Squash and merge"?

specs/sharding/beacon-chain.md
```diff
@@ -103,460 +101,342 @@ The following values are (non-configurable) constants used throughout the specif
...
 ```
 
-
 ### Block processing
 
+#### `process_beacon_block`
+
 ```python
-def process_block(state: BeaconState, block: BeaconBlock) -> None:
+def process_beacon_block(state: BeaconState, block: BeaconBlock) -> None:
```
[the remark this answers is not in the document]
JustinDrake: > process_intermediate_block

Did we settle on "builder block" (and "proposer block")?

PR #2794, opened 2022-01-10
Fix typo in main README

paulhauner:
Fixes a minor typo.

I would have preferred *"the upgrade formerly known as The Merge"* but I wasn't sure I could get away with it :smile:


djrtwo (approved):
thanks 🤣

PR #2797, opened 2022-01-11
Fix light client docs to match function signature

etan-status:
In the light client docs a mentioning of a function trigger is lacking
the `genesis_validators_root` argument. This patch adds that argument
to the documentation to match the real function signature. It also
slightly improves the grammar.


djrtwo (approved):
Thanks!

PR #2798, opened 2022-01-11
Document light client constants

etan-status:
This adds documentation about the unit and actual value of light client
specific constants, consistently with the rest of the spec.


djrtwo (approved):
thanks!

PR #2801, opened 2022-01-13
Avoid sync committee without proof in `LightClientUpdate`

etan-status:
Currently some of the light client tests include `next_sync_committee`
without a corresponding `next_sync_committee_branch` when the sync
committee is not changed by the update. This data is not useful for
a light client, as it cannot be validated without the merkle proof.
This patch ensures that the `next_sync_committee` member is also kept
empty in such cases.


etan-status:
Hmm, it would probably be cleaner to do the opposite and just always include the `next_sync_committee` instead of zeroing it out in certain conditions. Also makes it easier for the `LightClientUpdate` producer that does not usually know the current sync progress of the client.

etan-status:
Postponing this until the network protocols are specced out more concretely.

PR #2802, opened 2022-01-13
Define libp2p protocol for light client sync

etan-status:
While the current Altair specs define structures to aid light client
development, one missing key aspect is the network protocol definition.

Certain implementations have started defining their own REST based APIs,
e.g., Lodestar at https://github.com/ChainSafe/lodestar/blob/master/packages/api/src/routes/lightclient.ts
While such APIs are useful, REST does not seem to be the ideomatic
choice as the sole API available at such a low level for Ethereum.

This patch introduces a libp2p based protocol to allow light clients to
sync to the latest `BeaconBlockHeader` in a trustless and decentralized
manner, building on top of prior work from:
- @hwwhww at https://github.com/ethereum/consensus-specs/pull/2267
- @jinfwhuang at https://github.com/ethereum/consensus-specs/pull/2786
- Lodestar's REST API (also has an endpoint to fetch merkle proofs!)


specs/altair/sync-protocol.md
```diff
@@ -256,3 +351,368 @@ def process_light_client_update(store: LightClientStore,
...
+
+```python
+def create_optimistic_light_client_update(state: BeaconState,
+                                          block: SignedBeaconBlock,
+                                          attested_state: BeaconState) -> OptimisticLightClientUpdate:
+    slot = block.message.slot
+    assert slot == state.slot
+    attested_header = attested_state.latest_block_header.copy()
+    attested_header.state_root = attested_state.hash_tree_root()
+    assert state.latest_block_header.parent_root == attested_header.hash_tree_root()
```
terencechain, line 665: it may be nice to check `assert state.latest_block_header.body_root == block.message.body.hash_tree_root()` to further validate `sync_aggregate`

etan-status: Thanks, I have extended the validation to the full `BeaconBlockHeader`.

specs/altair/sync-protocol.md
```diff
@@ -74,6 +107,18 @@ class LightClientUpdate(Container):
     fork_version: Version
 ```
 
+### `OptimisticLightClientUpdate`
+
+```python
+class OptimisticLightClientUpdate(Container):
```
jinfwhuang, line 113: With this new "update", then we could probably simplify `LightClientUpdate` to be an update that always has finalized header. It would simplify a lot of the sync logic.

jinfwhuang:
I have a couple of high level comments/questions:
First, should we define the networking layer on libp2p? I have mentioned a few times elsewhere in the discussion, e.g. github PRs and light-client channel in Discord. We basically have three options on networking:
1. rpc
2. libp2p
3. portal network

The main drawback with the libp2p approach is what "LES" has experienced so far. There are very few clients that will serve the light client protocols. Participation of light client protocols are not required for the full node clients.

The portal network experiment will likely gather more momentum in the months ahead. If this network indeed becomes what it is intended to become, then that might be the long term winner.

In the short term, rpc is the easiest to implement and get bootstrapped.


Second, we should write the sync-protocol in such a way that the client could choose any of the three networking options, even if they are not yet defined. It could be possible that the light client ecosystem will accommodate all three networking options. In fact, that could be an likely outcome in the intermediate term when light clients users experiment and try things out.

specs/altair/sync-protocol.md
```diff
@@ -256,3 +351,395 @@ def process_light_client_update(store: LightClientStore,
...
+```
+
+Response Content:
+```
+(
+  List[LightClientUpdate, MAX_REQUEST_LIGHT_CLIENT_UPDATES]
+)
+```
+
+Requests `LightClientUpdate` instances in the sync committee period range
```
jinfwhuang, line 445: This could lead a very large response.

specs/altair/sync-protocol.md
```diff
@@ -256,3 +351,395 @@ def process_light_client_update(store: LightClientStore,
...
+#### Messages
+
+##### LightClientUpdatesByRange
+
+**Protocol ID:** `/eth2/beacon_chain/req/light_client_updates_by_range/1/`
+
+Request Content:
+```
+(
+  start_period: uint64
```
jinfwhuang, line 432: Why not just `(start_period, end_period)`? The response contains a sequence of updates that contains update path leading to the `end_period`. The updates sequence has unique sync-committee in such a way that consecutive updates in the response always compose of an updating pair.

specs/altair/sync-protocol.md
```diff
@@ -256,3 +351,395 @@ def process_light_client_update(store: LightClientStore,
...
+only a single canonical one (the same one that was included in the block) should
+be propagated across the network.
+
+### The Req/Resp domain
+
+#### Messages
+
+##### LightClientUpdatesByRange
+
+**Protocol ID:** `/eth2/beacon_chain/req/light_client_updates_by_range/1/`
```
jinfwhuang, line 427: I would suggest that we consider request by some kind of *id* instead of range. The reason is that two folds:
1. The networking allows requesters to ask specific updates. If the network has competing forks, all clients could still update based on their trusting sync-comm. In the by-range approach, if there is big fraction of network answers a client's request for a particular period with the wrong sync-comm, the client could not easily find an update that are unique to its trusted current-sync-comm. 
2. Requesting data object by key-value will translate well to other DHT-like p2p network. Specifically, portal network will work better with a key-value mechanism.

specs/altair/sync-protocol.md
etan-status: `next_sync_committee` should always come from `attested_header`. That was the case before #2746 and is also how the tests are set up. #2806 restores the previous behaviour.

And yes, over this libp2p protocol, `LightClientUpdate` will mostly include a `finalized_header`, unless there is a period of extended non-finality and the `finalized_checkpoint` is so far in the past that the `current_sync_committee` / `next_sync_committee` have to be followed in an optimistic way anyway.

etan-status:
> The main drawback with the libp2p approach is what "LES" has experienced so far. There are very few clients that will serve the light client protocols. Participation of light client protocols are not required for the full node clients.

Lodestar's prototype REST API proved to be very lightweight to serve in both Lodestar and Nimbus prototype versions. This libp2p proposal is similarly scoped in the kind and amount of data gathered and served. Having the major implementations serve those protocols by default should be achievable.

> In the short term, rpc is the easiest to implement and get bootstrapped.

Yes, but rpc requires a known node, and also an externally reachable rpc endpoint. The light client should be able to sync with the network autonomously.

> Second, we should write the sync-protocol in such a way that the client could choose any of the three networking options, even if they are not yet defined.

Yes, that is still possible. The libp2p proposal here defines portable functions that are not transport dependent. It is possible to also serve the same requests over a different transport. For example, the GossipSub topic can be mapped to the REST events API, and the other endpoints could just become standard requests. The processing overhead for the servers should be comparable, i.e., not noticeable. The functions defined in this proposal could be reused.

specs/altair/sync-protocol.md
etan-status: The response contains a series of chunks, each only having a single `LightClientUpdate`. There are at most as many chunks as have been requested.

This follows the same methodology as the existing `BeaconBlocksByRange` endpoint:
https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#beaconblocksbyrange

specs/altair/sync-protocol.md
etan-status: When syncing from an old state, this allows a client to request updates from multiple peers in parallel. i.e., request the odd periods from peer A, and the even ones from peer B.

This follows the same methodology as the existing `BeaconBlocksByRange` endpoint:
https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#beaconblocksbyrange

specs/altair/sync-protocol.md
etan-status: Thanks, that could indeed be a problem when there is non-finality for more than 2 sync committee periods (around 2 days). If non-finality is shorter, the sync committees are still stable (`current_sync_committee` and `next_sync_committee` are fixed at the start of a period), so any update from any of the forks should continue to apply fine. 

On the server side, `is_better_light_client_update` follows the majority choice picked by the sync committee in that case. Once finality is restored, this endpoint would return stable data once more. 

On the client side new updates are only applied if a sufficient part of the sync committee signed them and a long timeout has elapsed.

I think that with heavy forking across multiple sync committee periods, that you will end up having to re-sync a light client from a trusted checkpoint eventually. As in, that this is a problem not introduced by the proposed network layer, but a pre-existing conceptual limitation. If this proves to be a network layer problem, an additional endpoint taking both the period and the `hash_tree_root(sync_committee)` could help solve this. This would be in line with the duality of `BeaconBlocksByRange` (for finalized data) and `BeaconBlocksByRoot` (for non-finalized data).

Regarding key-value for portal network, the key here would just be the sync committee period. The value would change when a better update is found within a sync committee period (`is_better_light_client_update`).

specs/altair/sync-protocol.md
etan-status: After going over this once more, I have to retract the first part of my comment. You are indeed right that `next_sync_committee` for finalized updates are indeed specified as being rooted in the finalized checkpoint; I got confused by the renames in the most recent updates. There was no regression.

etan-status: Related PR: https://github.com/ethereum/consensus-specs/pull/2805

specs/altair/sync-protocol.md
```diff
@@ -256,3 +351,395 @@ def process_light_client_update(store: LightClientStore,
...
+6. [`LightClientUpdatesByRange`](#lightclientupdatesbyrange) requests are sent for sync committee period range
+   `[store_sync_committee_period, current_sync_committee_period]` (both inclusive).
+   The [`LightClientUpdate`](#lightclientupdate) responses are passed to `process_light_client_update` in ascending order.
+   If, for any sync committee period before `current_sync_committee_period`,
+   a `LightClientUpdate` cannot be obtained , the sync fails.
+
+7. Once the `store_sync_committee_period` initially reaches `current_sync_committee_period`,
+   or after syncing through multiple sync committee periods when re-joining the network,
+   `store.previous_max_active_participants` and `store.current_max_active_participants`
+   MUST be reset to `0` to clear the `store`'s participation tracker.
```
etan-status, line 733: Related PR: https://github.com/ethereum/consensus-specs/pull/2810
With that PR, this step is no longer needed.

specs/altair/sync-protocol.md
etan-status: A simplification of `LightClientUpdate` to always include a `finalized_header` could be problematic if there is non-finality for an entire sync committee period.

- If `next_sync_committee` is rooted in `finalized_header`, it would become stale and there would not be a way to advance to the next sync committee period anymore.
- If `next_sync_committee` is rooted in `attested_header`, there is no easy way to verify that the path from the `finalized_header` all the way to the current head has been correctly followed.

By only including a `finalized_header` if it is sufficiently recent, this problem is avoided by the current sync protocol design.

Agree that such simplifications would be nice, though. Also agree that it is probably more appropriate to be fixed outside of the work in this PR here.

etan-status:
I have updated this to better support the sync committee period boundary. It should now be possible to have the `optimistic_header` sync past the current sync committee period, and then obtain a full `LightClientHeader` at any point throughout the next sync committee period to advance `finalized_header` (and the accompanying `next_sync_committee`).

`OptimisticLightClientUpdate` are no longer considered for `best_valid_header` because they lack `next_sync_committee`.  Technically, they could be used to advance to a later slot within the current sync committee period, but that does not help the overall sync progress, and it would only apply via `UPDATE_TIMEOUT` within the next period when the `next_sync_committee` information must already be available to reliably continue syncing.

Note that I have temporarily merged additional PRs into this one to avoid countinuously fighting merge conflicts while I'm working on this PR. If you are interested in reviewing, please focus only on the latest commit of this PR.
The other commits are from PR https://github.com/ethereum/consensus-specs/pull/2805 and PR https://github.com/ethereum/consensus-specs/pull/2810 – please use those other PRs to comment on these.

etan-status:
Added another endpoint to fetch the latest `LightClientUpdate` (in additional the existing one for the latest `OptimisticLightClientUpdate` and the best `LightClientUpdate`.

This simplifies the sync process, and also enables use cases such as tools that solely intend to update from a given checkpoint block root to the latest finalized checkpoint, optimizing sync times.

etan-status:
Extended the test suite to cover all flavors of `LightClientUpdate` and `OptimisticLightClientUpdate`.

etan-status:
Added checks to reject signatures from `next_sync_committee` until `current_slot` has advanced sufficiently, and restructured the `validate_light_client_update` and `validate_optimistic_light_client_update` functions to look more similar.

etan-status:
Updated `GetLatestLightClientUpdate` to no longer produce LightClientUpdates with `finalized_header.slot` more than `UPDATE_TIMEOUT` in the past. This forces `next_sync_committee` to be rooted in `attested_header`, allowing light clients to advance into the next sync committee period. 

For `BestLightClientUpdatesByRange` the interval was already smaller because there, only updates are considered that have all of `finalized_header`, `attested_header` and the `signature_slot` in the current sync committee period. However, this approach is not feasible for `GetLatestLightClientUpdate` as it would introduce windows around the sync committee period change where the `finalized_header` cannot be advanced; hence, that endpoint can have `finalized_header` from previous sync committee period as well.

Also updated the light client sync process to ask multiple peers for `LightClientUpdate` to try and get one with a 2/3 or higher sync committee participation. Historic syncs would always trigger the "forced-update" condition even if only signed by an individual sync committee member, as `current_slot` is far ahead.

etan-status:
Documented error handling for the various endpoints (`3: ResourceUnavailable`).

specs/altair/sync-protocol.md
etan-status: Regarding changing the request to be keyed by some kind of _id_ instead of by `SyncCommitteePeriod`, I am not sure if it would actually help with (1) allowing client's to stay in sync with their view of the network.

The proposed protocol depends on the light client syncing from a trusted checkpoint within the weak subjectivity period. Doing that guarantees that it is not possible for multiple finalized chains to co-exist. This means, that up until the finalized checkpoint, there is only 1 possible fork that is propagated through the network, so using the `SyncCommitteePeriod` as key should be fine, also allowing to sync with the least amount of req/rsp round trips.

For the non-finalized slots, multiple forks could exist. However, if in a given sync committee period, any block has been finalized, the current and the next sync committee are fixed. This means that regardless of forking within the current and next sync committee period, light clients will be able to follow along.

If non-finality lasts through an entire sync committee period, though, the situation that you describe may occur. Multiple valid sync committees may be active, and it may be difficult for a light client to follow the specific fork that it is locked on-to. In that situation, having a request that is keyed by some kind of _id_ for the specific fork would indeed allow the light client to be more specific in its request. However, the light client will still have trouble discovering what peers are actually following that specific fork, ultimately boiling down to attempting to request updates for the specific fork and see if there is any response, vs just attempting to request any updates for the given period and see whether it is from the specific fork. Also, if the light client ends up picking an incorrect fork, even when it succeeded following its specific fork more efficiently, it would still not be able to recover if a different fork ends up being finalized, requiring a re-sync from a trusted checkpoint. Once finality is eventually restored, there is again only 1 possible fork being propagated through the network up until the new finalized checkpoint, enabling a deterministic re-sync.

Note that the situation should only occur if a majority of the sync committee originally signed a fork that becomes unviable in a later sync committee period. If they signed the dominating fork instead, even in extended periods of non-finality, light clients should end up picking the correct fork.

Ultimately, I think that following non-finalized parts of the chain remains a gamble of picking the correct fork, regardless of how the requests are keyed. If the correct fork is picked, it will eventually become easy again to follow it. If an incorrect one is picked, it will result in having to re-sync from a trusted checkpoint.

As for (2) the portal network, I guess using the `SyncCommitteePeriod` as an ID should be fine, or alternatively, a combination of the `SyncCommitteePeriod` with `hash_tree_root(sync_committee)`. However, even there, once finality is restored, the unviable forks would eventually end up being abandoned, resulting in light clients having to re-sync.

etan-status:
Updated protocol ID for `GetLightClientBootstrap` from `/eth2/beacon_chain/req/light_client_bootstrap_by_block_root/1/` to `/eth2/beacon_chain/req/light_client_bootstrap/1/`

etan-status:
Server implementation for Nimbus-eth2: https://github.com/status-im/nimbus-eth2/pull/3341

specs/altair/sync-protocol.md
etan-status: One more clarification on the addressing by some kind of _id_ instead of just sync committee period.

To implement the server side correctly, I agree that it **is** necessary to internally keep track of the _id_ so that the best `LightClientUpdate` avoids referring to an orphaned block in its `active_header`. My proposal to avoid exposing the root is solely related to the proposed libp2p network protocol.

From `Nimbus-eth2` server, these are the internally tracked updates:
```
    bestLightClientUpdates*:
       Table[SyncCommitteePeriod, altair.LightClientUpdate]
       ## Stores the `LightClientUpdate` with the most `sync_committee_bits` per
       ## `SyncCommitteePeriod`. Updates with a finality proof have precedence.

     pendingBestLightClientUpdates*:
       Table[(SyncCommitteePeriod, Eth2Digest), altair.LightClientUpdate]
       ## Same as `bestLightClientUpdates`, but for `SyncCommitteePeriod` with
       ## `next_sync_committee` that are not finalized. Key is `(period,
       ## hash_tree_root(current_sync_committee | next_sync_committee)`.

     latestLightClientUpdate*: altair.LightClientUpdate
       ## Tracks the `LightClientUpdate` for the latest slot. This may be older
       ## than head for empty slots or if not signed by sync committee.

     optimisticLightClientUpdate*: OptimisticLightClientUpdate
       ## Tracks the `OptimisticLightClientUpdate` for the latest slot. This may
       ## be older than head for empty slots or if not signed by sync committee.
```

Whenever fork choice selects a new head, `bestLightClientUpdates` is updated with the corresponding `pendingBestLightClientUpdates`. As soon as any block within a sync committee period finalizes, `active_header` is guaranteed to always be on the correct chain, so the forks are no longer tracked independently.

etan-status:
Rebased to latest `dev`, moving towards use of `compute_sync_committee_period` instead of inline division.

etan-status:
Updated light client flow with remark that `genesis_time` and `genesis_validators_root` must be provided out-of-band before initiating sync.

etan-status:
Rebased to `dev`, and pulled latest #2805.

etan-status:
Added optional `allow_force_update` parameter to `process_light_client_update` to allow improving `best_valid_update` in historic sync without every single `LightClientUpdate` triggering a force-update (finalized head will always be behind wall clock).

Extended the description of light client operations with more details about the contents of the `Status` request and about avoiding force-updates while syncing (when network activity is low / not many updates are received).

etan-status:
- Updated light client process description step 6 with details about unknown `store.next_sync_committee` case.

etan-status:
- Rebased to `dev`, pulling in latest update to PRs this depends on.

specs/altair/sync-protocol.md
```diff
@@ -265,4 +438,473 @@ def process_light_client_update(store: LightClientStore,
...
+##### BestLightClientUpdatesByRange
+
+**Protocol ID:** `/eth2/beacon_chain/req/best_light_client_updates_by_range/1/`
+
+Request Content:
+```
+(
+  start_period: uint64
+  count: uint64
+  step: uint64
```
dapplion, line 560: Can we drop the `step` parameter? For lightclient updates it doesn't make sense since you must process strictly sequential updates.

etan-status: Well, the client could still use multiple parallel downloads with non-1 steps to try and get a faster speed. It is very similar to `beaconBlocksByRange` which, ultimately, also need to be applied in sequence. If a client is not interested in using that optimization, just supply `1` as step size. Personally I'd prefer leaving it in for consistency with `beaconBlocksByRange`, also enabling code reuse with whatever downloader is being used for blocks.

dapplion: Note that PR https://github.com/ethereum/consensus-specs/pull/2850 for EIP-4844 is dropping the step parameter intentionally breaking the beacon_blocks_by_range pattern

https://github.com/ethereum/consensus-specs/pull/2850/files#diff-c055ad660df5d383113f454c99ab8d6f70964643405d019ca43560ec9ac2841bR181-R182

etan-status: Oh, that's interesting, given the trivial implementation on the server side and the optionality on the client side. The referenced PR refers to "previous complaints dev-feedback" in the description – is there any additional background on these? If this is a general trend to no longer use `step` for newer endpoints (besides just PR specific style), sure I can also remove it here for consistency going forward.

etan-status:
- Use `is_finality_update` helper when deriving file names in test generator.
- Remove `step` parameter from `BestLightClientUpdatesByRange` message, in line with #2850 (but against `beaconBlocksByRange` convention).

specs/altair/sync-protocol.md
etan-status: More background on "previous complaints dev-feedback": https://github.com/ethereum/consensus-specs/pull/2856

etan-status:
For testing, while this PR is still being discussed, will use `v0` flavors of the various messages.

- Version 0 for the Req/Resp endpoints
  - `best_light_client_updates_by_range` v0
  - `latest_light_client_update` v0
  - `optimistic_light_client_update` v0
  - `light_client_bootstrap` v0
- `_v0` suffix for the GossipSub topic
  - `optimistic_light_client_update_v0`

https://github.com/status-im/nimbus-eth2/pull/3550
 
Nimbus `unstable` can be configured to serve this protocol (with `v0`).
Nodes supporting it may be discoverable on the `kiln` network.

etan-status:
Feedback from discussions at Devconnect AMS:

- `next_sync_committee` could be always rooted in `attested_header`. 
   - If the light client finds that `finalized_header` does not correspond to the current sync committee period, it can try fetching a new `LightClientUpdate` later.
   - Open how well such an approach works in edge cases (extended non-finality / non-supermajority participation)
- Gossipsub: While full nodes can indeed verify the gossip to match the canonical sync aggregate in the beacon block, light clients cannot. If they opt to not propagate messages, they get downscored. If they propagate a modified one (e.g., 1 aggregate bit removed), they get downscored as well.
  - Broadcaster could be changed from block proposer to sync committee aggregators. Then, the messages could receive a copy of the sync committee aggregator selection proof, and gossip be locked to just messages with a valid proof. This restricts spam potential to just the active sync committee aggregators. 
  - Open how this could react to potential future changes about how sync committee aggregators get selected.
- Removing `fork_version` / `is_signed_by_next`.
   - Those fields are not strictly necessary, as the light client could just attempt to verify the signature 2-3 times depending on the situation (close to sync committee period end / close to fork change)
   - Open to determine why `fork_version` was originally added - the consensus-specs do not strictly link fork version to solely depend on epoch changes, and the spec test runners sometimes create inconsistent states where fork version reflects a newer fork, but compute_epoch_at_slot is at genesis.
   - Alternative suggestion came up to include `slot` instead of `fork_version` / `is_signed_by_next`, but it is not verifiable and is 8 bytes vs 5.
- Removing beacon block headers from bootstrap (and maybe even updates) and just include slot / block root.
   - Rationale being that a light client only really cares about following the `next_sync_committee` chain, and could fetch headers / blocks of interest in separate requests.
   - Open how much this would really reduce message sizes. Proofs would become longer because they are no longer rooted in state_root but instead in block_root.
- Align naming with best practices and make it consistent across transports.
   - Data objects:
      - `OptimisticLightClientUpdate` -> `LightClientOptimisticUpdate`, to ensure searching for `LightClientUpdate` does not also find all occurrences of the optimistic flavor
         - Alternatives discussed: `LightClientHeadUpdate` (head has a different meaning across spec and is misleading as the protocol provides older data than the head), `LightClientHeaderUpdate` (the object could be extended in the future, e.g., with execution state root, and then the name is outdated)
      - `LightClientUpdate` stays the same
         - Alternatives discussed: `LightClientCommitteeUpdate` (it does not always update committees), `LightClientFinalizedUpdate` (it does not always has finality), `LightClientFullUpdate` (not necessary when optimistic is renamed)
      - `LightClientBootstrap` stays the same
         - Alternatives discussed: `LightClientSnapshotWithProof` (snapshot used to be the name for store, proofs have a meaning of selection proofs / terminology for merkle proofs is "branch") / `LightClientInit` (does not sound like a noun)
   - Req/Resp
      - `best_light_client_updates_by_range` -> `light_client_updates_by_range` (the response is best-effort / not necessarily globally the best participation)
      - `latest_light_client_update` stays the same
      - `optimistic_light_client_update` -> `light_client_optimistic_update`
      - `light_client_bootstrap` stays the same
   - Gossipsub
      - `optimistic_light_client_update` -> `light_client_optimistic_update`, or if the selection proof gets added for gossip messages, `light_client_optimistic_update_and_proof`.

etan-status:
[libp2p.pdf](https://github.com/ethereum/consensus-specs/files/8528838/libp2p.pdf)

Slides from Light client summit @ Devconnect AMS, regarding this PR.

etan-status:
Results from further discussion at Devconnect AMS:

- Always root `next_sync_committee` in `attested_header`, but ensure that `finalized_header.slot` is in the current sync committee period before considering `next_sync_committee` finalized.
  - Open if this should use an EF multiproof (probably yes), or a Lodestar multiproof, or two separate branches. Should provide fallback instructions on creating those proofs manually if SSZ library does not support multiproof.

- Gossipsub: Selection proofs are too complex (web3signer / validator client API / additional signing domain etc).
   - Instead of proposer, all 512 sync committee members publish to the gossipsub channel
   - Full node validation logic does no longer have REJECT properties:
      - If received message does not match local data exactly (canonical signature), IGNORE (avoid propagating spam)
      - If time when received is earlier than 4 secs (–0.5s) after the slot start time, IGNORE (avoid propagating before block is fully available across the network)
      - If received message match local data, remember that message was seen, otherwise IGNORE (avoid propagating multiple times, gossip is ephemeral)
      - Otherwise, ACCEPT
   - Light client validation logic:
       - If received message is valid but does not advance the local LightClientStore, IGNORE (avoid amplifying spam)
       - Otherwise, ACCEPT (even if a spammer removed an aggregate bit, if the message is still sufficiently good, it does not matter which message gets processed by others on the network)

- Replace `fork_version` / `is_signed_by_next` with `signature_slot`. The other information can be derived from it, and logic is more straight-forward. A future CL hardfork should consider fixing the edge case by letting the sync committee sign the current block root, instead of the next slot's `parent_root`.

- Pure light clients that want to serve historic light client data to others should update their `Status` message accordingly. Full nodes can use the list of available protocols to determine whether they are a light client serving historic light client data, or a full node that also has historic blocks data available (the light client would not advertise beaconBlocksByRange / beaconBlocksByRoot). Hybrid clients that use light client solely for syncing should advertise the sync progress of their full node component (as they appear as full nodes on the network, so putting the slot in status makes others fetch blocks).

- Removing beacon block headers from messages: No, keep them. Proofs become unnecessarily longer if headers are omitted (as now, the `state_root` would require proving), and the spec's goal of providing access to latest finalized / optimistic heads could not be guaranteed in a hypothetical scenario where only a bootstrap is available but no further updates.

- Adding `next_sync_committee` to bootstrap: No, the data is available via `lightClientUpdatesByRange` already, and putting yet another copy of that data unnecessarily doubels the total size needed e.g. to store bootstrap data on portal network.

- Add `LightClientFinalityUpdate` to represent updates to `finalized_header` without having to bundle the entire `next_sync_committee`, and add a separate gossipsub topic to push updates whenever finality changes. This avoids light client having to pull new full `LightClientUpdate` just to follow finality (~600 bytes vs 25k every time).

- Naming: The suggestions above are OK.

etan-status:
I'm currently updating the spec with the feedback from Devconnect, so far I've stumbled across these additional points:

1. Rooting `next_sync_committee` in `attested_header` would allow using merkle multiproof, but it does not seem to be worth it, because:
   - `next_sync_committee` / `finalized_header` are optional, making the format dynamic (sometimes multi-item proof, sometimes single-item proof, sometimes empty), so additional complexity (varlen vs fixed len + if-tree)
   - Size savings are insignificant (<1%), compared to the ~25K that a snappy-compressed `LightClientUpdate` needs:
      -  Single-item proofs:
         - `next_sync_committee_branch` = 5*32
         - `finality_branch` = 6*32
         - `execution_state_root` = 9*32 (not currently included)
      - Multi-item proofs:
         - `next_sync_committee + finality` = 6*32
         - `finality + execution_state_root` = 12*32
         - `next_sync_committee + finality + execution_state_root` = 12*32
   - While it is straight-forward to remove an individual contribution from a multi-proof after the creation, the full nodes should be able to serve light client requests with just a couple `memcpy`, and process gossip with just `memcmp`. Having different proof formats in responses means either having to precompute all forms and waste RAM, or do more complex on-demand processing as requests arrive.
   - Light client can easily morph `LightClientOptimisticUpdate` into `LightClientFinalityUpdate` into `LightClientUpdate` by just leaving the unknown fields empty, then use the same logic as for full updates. This saves code space because it is not needed to track which items were originally present (and which missing proofs to ignore).
   - Oracles forwarding `LightClientUpdate` to smart contract based light clients may still combine the individual proofs into a multi-proof, if they prefer minimizing the message size to the absolute minimum (different tradeoff).
   - ⚠️ Staying with single-item proofs, as is

2. The endpoint for latest `LightClientUpdate` is not that useful, when we already have latest `LightClientFinalityUpdate`, because it may not have good participation, or may not include a relevant `next_sync_committee`.
   - Instead, light client should just follow `LightClientFinalityUpdate` / `LightClientOptimisticUpdate` until `store.finalized_header` advanced to the latest period. Then, fetch full update using the `light_client_updates_by_range` endpoint to backfill `next_sync_committee` whenever it is convenient (that endpoint returns the best known participation instead of latest one).
   - The new logic will allow `finalized_header` to advance into the new period before next sync committee is known, improving responsiveness
   - ⚠️ Dropping latest `LightClientUpdate` endpoint / only `light_client_updates_by_range` will provide full `LightClientUpdate`

3. Later spec forks may change the structures used by this protocol.
   - ⚠️ Add fork context to libp2p responses, same way as done for blocks since altair. Determining slot for this fork digest is `update.attested_header.slot` for now, in line with the gossipsub topic's fork digests
   - Note that for `light_client_updates_by_range`, the fork context is sent on each individual update, enabling ranges to span multiple forks. This is in line with blocks by range requests.
   - REST API should probably consider doing something similar, also checking re prior art of other endpoints.

4. Update ranking is suboptimal.
   - Currently, any update with finality takes precedence over any others without finality. But there may be case of extended non-finality where no finality is available on canonical chain (in old spec, finality proof can only be included if it is not too old). This means in that situation attacker can supply an update with just 1 signature and a fake finality proof for their custom chain to become best update.
   - ⚠️ New update ranking, that ranks according to these tiebreakers:
      1. Higher sync committee participation, until supermajority (> 2/3) is reached
      2. Presence of a `next_sync_committee` that is relevant (`attested_period == signature_period`)
      3. Indication of any finality
      4. Indication of sync committee finality (`finalized_period == attested_period`)
      5. Sync committee participation beyond supermajority
      6. Older `attested_header.slot` (fewer changes to best update)
   - Will also add test suite to align update ranking across implementations.
   - Note that with the new spec, there is only a few legitimate ways how information may be missing from a particular update - in general, it should be relatively easy to obtain a better update from a different peer:
      1. Missing `next_sync_committee` -> if `attested_period != signature_period`, it does not make sense to include as it is not verifiable or applicable without already knowing the sync committee
      2. Missing `finalized_header` -> Server was using checkpoint sync and has not yet backfilled that header (while checkpoint sync starts at a finalized block itself, some early `finalized_checkpoint.root` may point to earlier blocks)
      3. The update being processed is a `LightClientFinalityUpdate` or `LightClientOptimisticUpdate` which simply does not carry certain data.

etan-status:
Applied feedback from Devconnect AMS:

- Replace `is_signed_by_next` / `fork_version` with `signature_slot` in `LightClientUpdate`
- Root `next_sync_committee` always in `attested_header`, remove `get_active_header`
- Revise `LightClientUpdate` ranking, provide test suite to ensure implementations rank consistently
- Add `LightClientFinalityUpdate`, and allow `store.finalized_header` to proceed without knowing `next_sync_committee` yet
- Drop `LatestLightClientUpdate` endpoint, replace it with `LightClientFinalityUpdate` to get latest finality update
- Manually trigger forced updates based on use case dependent heuristics, instead of automated by timing
- Add `ForkDigest` context to all libp2p responses for future-proofing
- Revise gossipsub message validation rules to allow light clients to participate in the network without risking to get kicked
- Allow pure light clients to also serve light client data, if they have historic data available
- Naming changes (`OptimisticLightClientUpdate` -> `LightClientOptimisticUpdate` / `best_light_client_updates_by_range` -> `light_client_updates_by_range`)
- Extend test suite for more edge cases
- Split the spec into two parts, a generic conceptual part, and then the libp2p specifics further down
- Split the PR into multiple individual commits, which can each be independently reviewed

etan-status:
- Fix CI
- First commit of the mess here, as separate PR: https://github.com/ethereum/consensus-specs/pull/2805

etan-status:
Nimbus (unstable) supports the latest protocols using `_v0` suffix (Kiln / Prater).

Gossipsub:
- `/eth2/00000000/light_client_finality_update_v0/ssz_snappy`
- `/eth2/00000000/light_client_optimistic_update_v0/ssz_snappy`

Req/Resp:
- `/eth2/beacon_chain/req/light_client_bootstrap/0/`
- `/eth2/beacon_chain/req/light_client_updates_by_range/0/`
- `/eth2/beacon_chain/req/light_client_finality_update/0/`
- `/eth2/beacon_chain/req/light_client_optimistic_update/0/`

etan-status:
- Rebase to `dev` (re-ordering of `Misc` section)
- Submitted PR 2/12 for review: https://github.com/ethereum/consensus-specs/pull/2916

etan-status:
- Rebase to `dev` (`@spec_state_test_with_matching_config` decorator)
- Submitted PR 3/12 for review: https://github.com/ethereum/consensus-specs/pull/2924

etan-status:
- Rebase to `dev`
- Submitted PR 4/12 for review: https://github.com/ethereum/consensus-specs/pull/2927

etan-status:
- Rebase to `dev`
- Submitted PR 5/12 for review: https://github.com/ethereum/consensus-specs/pull/2929

etan-status:
- Rebase to `dev`
- Submitted PR 6/12 for review: https://github.com/ethereum/consensus-specs/pull/2932

etan-status:
- Rebase to `dev` (helper functions order, `compute_sync_committee_period_at_slot`, pyspec style, prefix update `_period` variables, forced update comments)
- Submitted PR 7/12 for review: https://github.com/ethereum/consensus-specs/pull/2935

etan-status:
- Rebase to `dev` (`update_` prefix for `_period` variables)
- Submitted PR 8/12 for review: https://github.com/ethereum/consensus-specs/pull/2938

etan-status:
- Rebase to `dev` (intro text formatting)
- Submitted PR 10/12 for review: https://github.com/ethereum/consensus-specs/pull/2947

etan-status:
- Rebase to `dev` (rename force update function, split into multiple files)
- Submitted PR 11/12 for review: https://github.com/ethereum/consensus-specs/pull/2951

etan-status:
- Rebase to `dev` (this PR itself is the final part 12/12)

specs/altair/light-client/p2p-interface.md
```diff
@@ -0,0 +1,257 @@
...
+
+This section extends the [honest validator specification](../validator.md) with additional responsibilities to enable light clients to sync with the network.
+
+### Beacon chain responsibilities
+
+All full nodes SHOULD subscribe to and provide stability on the [`light_client_finality_update`](#light_client_finality_update) and [`light_client_optimistic_update`](#light_client_optimistic_update) pubsub topics by validating all received messages.
+
+### Sync committee
+
+Whenever fork choice selects a new head block with a sync aggregate participation `>= MIN_SYNC_COMMITTEE_PARTICIPANTS` and a post-Altair parent block, full nodes with at least one validator assigned to the current sync committee at the block's `slot` SHOULD broadcast derived light client data as follows:
```
hwwhww, line 252: Note: There may be some room to optimize it, based on how many full nodes enable the light client protocol.

e.g., if most nodes are altruistic and enable light client protocol, maybe we only need to assign the sync committee aggregators to publish the updates.

specs/altair/light-client/p2p-interface.md
```diff
@@ -0,0 +1,257 @@
...
+
+Light clients SHOULD call `process_light_client_optimistic_update` even if the message is ignored.
+
+### The Req/Resp domain
+
+#### Messages
+
+##### GetLightClientBootstrap
+
+**Protocol ID:** `/eth2/beacon_chain/req/light_client_bootstrap/1/`
```
hwwhww, line 97: Curious, should it be `beacon_chain/` or `light_client/`?
I see that it's a "beacon chain" light client, but for implementers & service providers, maybe there some benefits of distinguishing them?

@etan-status what do you think?

hwwhww (approved):
Epic work @etan-status!

Generally look good. 👍

There may be some design space on the light client incentives that we can figure out later.

specs/altair/light-client/p2p-interface.md
etan-status: Yeah, it may work with just aggregators, but it's too early to tell. Also have to consider that the topic also extends beyond just full nodes, with all light clients also being subscribed to it, so I erred towards having a few more broadcasts for now. Originally, I implemented it with just the proposer sending the update, but that has the obvious downside of the slot not appearing at all if the proposer does not implement the light client sync protocol – in general, it still worked surprisingly well though.

The messages involved here are tiny (`BeaconBlockHeader` + `SyncAggregate` essentially), and the additional broadcast lines up with the general concept to reward the sync committee for enabling light clients to sync. Furthermore, the sync committee already sends messages at 1/3 into the block, so sending the second message around that time also groups up those tasks nicely.

specs/altair/light-client/p2p-interface.md
etan-status: I don't have a strong opinion on this.

`beacon_chain` fits for these reasons:
- What is being downloaded are partial beacon blocks / beacon states
- The other light client related messages, i.e., sync contributions, are also rooted in `beacon_chain`
- The light client still needs to deal with the `status`, `ping`, `metadata`, `goodbye` messages, so it cannot solely use `light_client`, at least on libp2p

`light_client` fits for these reasons:
- The topics are not necessary for collecting validator rewards
- Easier to reason starting point to direct a client implementation to

Other impacted API should also be considered, e.g., REST, should it have it under the light client umbrella, or the beacon chain one? What about portal?

hwwhww: I'm fine with either way too! Let's keep it to `beacon_chain` now unless someone has a strong objection one day.

PR #2803, opened 2022-01-13
add `pylint` to catch unused args

### LINKED ISSUE #2761: Double check linting settings
@potuz raised the issue that phase0's `get_validator_from_deposit` takes an additional `state` argument that is not used.

From a cursory look, it seems like our linting stack *should* have caught this in the `pyspec`.

Let's double-check it is working as intended and add if missing.

ralexstokes:
fixes #2761.

### DIFF
--- a/specs/altair/beacon-chain.md
+++ b/specs/altair/beacon-chain.md
@@ -521,7 +521,7 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
         signing_root = compute_signing_root(deposit_message, domain)
         # Initialize validator if the deposit signature is valid
         if bls.Verify(pubkey, signing_root, deposit.data.signature):
-            state.validators.append(get_validator_from_deposit(state, deposit))
+            state.validators.append(get_validator_from_deposit(deposit))
             state.balances.append(amount)
             state.previous_epoch_participation.append(ParticipationFlags(0b0000_0000))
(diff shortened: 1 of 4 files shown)


specs/altair/beacon-chain.md
```diff
@@ -521,7 +521,7 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
         signing_root = compute_signing_root(deposit_message, domain)
         # Initialize validator if the deposit signature is valid
         if bls.Verify(pubkey, signing_root, deposit.data.signature):
-            state.validators.append(get_validator_from_deposit(state, deposit))
+            state.validators.append(get_validator_from_deposit(deposit))
```
ralexstokes, line 524: update from phase0 as well

djrtwo (approved):
given the lack of test files changed, I suppose  don't have a unit test for the function. Could use this as a chance to add a sanity call in there... but probably fine either way.

thanks!

hwwhww (approved):
Looks good to me!
It seems we can also consider enabling some more `pylint` checks later.

ralexstokes:
> don't have a unit test

this utility function is simple enough that we would basically just be writing the implementation which doesn't add a lot of extra assurance imo

there is also plenty of transitive test coverage via things that use the `Validator` so i think it is fine w/o

PR #2805, opened 2022-01-17
Allow light client to verify signatures at period bounary

etan-status:
As the sync committee signs the previous block, the situation arises at
every sync committee period boundary, that the new sync committee signs
a block in the previous sync committee period. The light client cannot
reliably detect this condition (e.g., assume that this is the case when
it is currently on the last slot of a sync committee period), because
the last couple slots of a sync committee period may not have a block.
The current `fork_version` field inside `LightClientUpdate` is replaced
with the `signature_slot` to allow light clients to verify signatures
across sync committee boundaries, without notably increasing its size.
The `fork_version` can be derived from the `signature_slot`. Note that
the `signature_slot` is not signed and cannot be fully trusted beyond
identifying the `fork_version` and signing sync committee.


specs/altair/sync-protocol.md
```diff
@@ -102,6 +103,15 @@ def get_subtree_index(generalized_index: GeneralizedIndex) -> uint64:
...
 ```
 
+### `fork_version_at_epoch`
+
+```python
+def fork_version_at_epoch(epoch: Epoch) -> Version:
+    if epoch >= ALTAIR_FORK_EPOCH:
+        return ALTAIR_FORK_VERSION
+    return GENESIS_FORK_VERSION
+```
```
etan-status, line 113: One problem here comes up during testing `@with_altair_and_later`, because the generated Altair states do not fork at `ALTAIR_FORK_EPOCH` etc but instead fork earlier. So the local `state` during test is already on Altair, but the validation fails on the client side because it assumes that the fork applies after `ALTAIR_FORK_EPOCH`.

The previous approach had a similar problem that the light client would just blindly trust whatever `fork_version` is given as part of the `LightClientUpdate`.

What is the recommended way to continue here?

dapplion: In a naive case the validator could try with verifying the signature with both sync committees

etan-status: The `fork_version` could theoretically change multiple times throughout a single sync committee period. 
But yeah, those could also be locally enumerated and tested.
Problem remains, though. The client needs a way to verify / determine a fork version, or it runs risk of accepting an update with an invalid fork version.

etan-status: Having the `fork_version` included actually enables use cases where the light client does not need to be aware of forks, as long as it does not intend to request state proofs and continue operating on the `BeaconBlockHeader`. An example could be a simple tool to advance a given checkpoint root to the most recent `BeaconBlockHeader`, then fetches that state from Infura and uses it to start a full node using weak subjectivity sync.

For now, I'll keep the `fork_version` in. It should be revisited in a separate PR if needed.

etan-status:
Rebased to `dev`, also fixing a recent regression in `get_sync_aggregate` when requesting a signature from `state.next_sync_committee` that was introduced in #2826.

etan-status:
Rebased to `dev`.

etan-status:
Change method in which information about signing sync committee is included into the `LightClientUpdate` (Devconnect AMS feedback - see https://github.com/ethereum/consensus-specs/pull/2802#issuecomment-1107474871) 

Before:
- If `update.next_sync_committee` is empty, then `update.sync_aggregate` was signed by `store.next_sync_committee`, otherwise it was signed by `store.current_sync_committee`.

After:
- `update.signature_slot` contains the slot at which the `update.sync_aggregate` was signed. If that slot is in the `store.finalized_header.slot`'s sync committee period, the update was signed by `store.current_sync_committee`. If it's in the next one, it was signed by `store.next_sync_committee`.

etan-status:
- Fix `get_sync_aggregate` test function to use correct fork digest at fork boundary
- Removed unused `block_root` argument from `get_sync_aggregate`
- Return `fork_digest` / `signature_slot` that is used by `get_sync_aggregate`, for clarity

ralexstokes:
I have not done a review of the changes but just going off the PR description:

it seems like we could use the `current_slot` input to `validate_light_client_update` to determine which sync committee period we are in, and then avoid adding more data to the light client update


what do you think about this?

etan-status:
The `current_slot` input is based off local wall clock, which may be several days in the future in case of syncing historic data.

ralexstokes:
ah good point, then I think adding the target slot makes a lot of sense :)

PR #2806, opened 2022-01-17
Fix `next_sync_committe` validation for light client

etan-status:
PR https://github.com/ethereum/consensus-specs/pull/2746 introduced a
regression in the validation of the `next_sync_committee` where the
corresponding merkle proof is computed based on an incorrect state root.
`next_sync_committee` is always rooted in the attested header, even for
a `LightClientUpdate` that includes a `finalized_header`.


etan-status:
After further review, this seems to not have been a regression. For finalized `LightClientUpdate`, the `next_sync_committee` is rooted in the finalized checkpoint; not in the attested state. Closing this one.

PR #2807, opened 2022-01-17
Consistently use `compute_sync_committee_period`

etan-status:
There were a couple instances where a division was used on an epoch
to derive the corresponding sync committee period instead of calling the
`compute_sync_committee_period` function.
These instances were changed to also use the function.


hwwhww (approved):
lgtm. thanks!

PR #2808, opened 2022-01-17
receipt_root -> receipts_root

djrtwo:
Minor renaming for upcoming bellatrix spec release.

This was discussed and generally agreed upon on a public call. Due to the need to make a release patching the semantics of executePayload and forkchoiceUpdated, we are getting this in there as well


mkalinin (approved):
LGTM!

specs/bellatrix/beacon-chain.md
```diff
@@ -169,7 +169,7 @@ class ExecutionPayload(Container):
     parent_hash: Hash32
     fee_recipient: ExecutionAddress  # 'beneficiary' in the yellow paper
     state_root: Bytes32
-    receipt_root: Bytes32  # 'receipts root' in the yellow paper
+    receipts_root: Bytes32  # 'receipts root' in the yellow paper
```
hwwhww, line 172: [not strongly] It seems now we can consider removing the comment here.
```suggestion
    receipts_root: Bytes32
```

hwwhww (approved):
LGTM

lokera666 (commented):
Best practice

PR #2810, opened 2022-01-20
Revise participation tracking in light client

etan-status:
The current mechanism to track historic sync committee participation is
suboptimal in these aspects:
- During sync, `LightClientUpdate` are fetched across multiple sync
  committee periods. The sync participation may change drastically over
  time. If any of those synced periods had unusually high participation
  and is followed by periods of unusually low participation, this may
  prevent optimistic header updates until `current_slot` has advanced
  sufficiently (from the sync time, not from the update time).
- When the light client suspends, `process_slot_for_light_client_store`
  needs to be called for all missed slots after resuming, which is not
  practical (optimizations would be possible).

This patch changes the rollover of the tracked participation to happen
whenever the `store.finalized_header`'s sync committee period advances.
While in sync with the network, rollovers happen at the same frequency
as before, but this guarantees that participation is no longer mixed
across multiple sync committee periods, avoiding the problems above.

### DIFF
--- a/specs/altair/sync-protocol.md
+++ b/specs/altair/sync-protocol.md
@@ -52,7 +52,7 @@ uses sync committees introduced in [this beacon chain extension](./beacon-chain.
 
 | Name | Value | Unit | Duration |
 | - | - | - | - |
-| `MIN_SYNC_COMMITTEE_PARTICIPANTS` | `1` | validators |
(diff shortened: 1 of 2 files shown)


etan-status:
- Rebased to `dev`.

etan-status:
This PR is no longer necessary, as other light client sync protocol updates will replace the automated force-update based on wall clock with a manual trigger based on heuristics depending on the risk profile of the corresponding use case.

PR #2811, opened 2022-01-21
Allow light client updates to only improve optimistic header

etan-status:
The current `LightClientUpdate` logic rejects updates not advancing the
best known `finalized_header`, even when the `optimistic_header` could
be advanced. This patch changes the logic to also process updates that
solely advance the `optimistic_header`.


etan-status:
This PR will not be needed anymore, as this will be a side effect of removing `get_active_header` when rooting `next_sync_committee` in `attested_header` instead of `finalized_header`.

PR #2818, opened 2022-01-27
Update reserved libp2p error documentation

etan-status:
The spec reserves the libp2p error code range `[3, 127]` for future use
but actually defines error code `3` as `ResourceUnavailable`. This patch
updates the reserved range to `[4, 127]`.


hwwhww (approved):
lgtm

PR #2830, opened 2022-02-11
Ignore attestations voting for the wrong finalized checkpoint

arnetheduck:
When nodes are syncing but have not yet reached the canonical `head`,
they cannot determine whether nodes they are connected to serve a valid
history or are making bogus claims in their `Status` advertisement.

Thus, the best course of action that a client can take is to vote for
its "current" best synced head, regardless of whether it's connected to
peers that claim to have other heads or not.

However, in the p2p spec, we penalize such peers with a `REJECT` - this
should be an `IGNORE` instead because this vote is correct per the spec,
albeit "late" according to the validating clients' view of the chain.


djrtwo (approved):
as discussed, this could help from network wide deadlocks or network fragmentation in the extreme

PR #2831, opened 2022-02-15
on_block: `pre_state` can't be before `BELLATRIX_FORK_EPOCH`

terencechain:
The `pre_state` in `on_block`  can be version Altair if the parent block is version Altair. `is_merge_transition_block` will no longer be viable because `pre_state` doesn't have `latest_execution_payload_header` field.  I'm not entirely sure what's the best approach to this. I think other clients are probably silently handling it? In Prysm, we are skipping the `is_merge_transition_block` if the `pre_state` is not the `Bellatrix` version. Checking the epoch is equally sufficient. Looking forward to hearing more feedback about this

### DIFF
--- a/specs/bellatrix/fork-choice.md
+++ b/specs/bellatrix/fork-choice.md
@@ -171,7 +171,7 @@ def on_block(store: Store, signed_block: SignedBeaconBlock) -> None:
     state_transition(state, signed_block, True)
 
     # [New in Bellatrix]
-    if is_merge_transition_block(pre_state, block.body):
+    if compute_epoch_at_slot(pre_state.slot) >= BELLATRIX_FORK_EPOCH and is_merge_transition_block(pre_state, block.body):
         validate_merge_block(block)
 
     # Add new block to the store
--- a/specs/bellatrix/fork-choice.md
+++ b/specs/bellatrix/fork-choice.md
@@ -171,7 +171,7 @@ def on_block(store: Store, signed_block: SignedBeaconBlock) -> None:
     state_transition(state, signed_block, True)
 
     # [New in Bellatrix]
-    if compute_epoch_at_slot(pre_state.slot) >= BELLATRIX_FORK_EPOCH and is_merge_transition_block(pre_state, block.body):
+    if is_merge_transition_block(pre_state, block.body):
         validate_merge_block(block)
 
     # Add new block to the store
--- a/specs/bellatrix/beacon-chain.md
+++ b/specs/bellatrix/beacon-chain.md
@@ -213,6 +213,8 @@ class ExecutionPayloadHeader(Container):
 
 ```python
 def is_merge_transition_complete(state: BeaconState) -> bool:
+    if compute_epoch_at_slot(state.slot) < BELLATRIX_FORK_EPOCH:
+        return False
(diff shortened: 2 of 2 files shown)


specs/bellatrix/fork-choice.md
```diff
@@ -171,7 +171,7 @@ def on_block(store: Store, signed_block: SignedBeaconBlock) -> None:
     state_transition(state, signed_block, True)
 
     # [New in Bellatrix]
-    if is_merge_transition_block(pre_state, block.body):
+    if compute_epoch_at_slot(pre_state.slot) >= BELLATRIX_FORK_EPOCH and is_merge_transition_block(pre_state, block.body):
```
dapplion, line 174: In Lodestar we have a type guard before that assertion, so +1 on silently handling it

```ts
if (bellatrix.isBellatrixStateType(state as BeaconStateBellatrix) &&
    bellatrix.isBellatrixBlockBodyType(block.body) &&
    bellatrix.isMergeTransitionBlock(state as BeaconStateBellatrix, block.body))
```


https://github.com/ChainSafe/lodestar/blob/e12bcd15f110c99cba73b63a367aea046d3b5bf0/packages/fork-choice/src/forkChoice/forkChoice.ts#L334

mkalinin: This check skips transition block validation if the first block in the Bellatrix epoch is the merge transition block. It should rather be placed in `is_merge_transition_complete` function to make it return `False` if `compute_epoch_at_slot(state.slot) < BELLATRIX_FORK_EPOCH`

terencechain: Makes sense!

mkalinin (approved):
LGTM! 👍

hwwhww:
You're correct that `on_block` may have to handle `pre_state: AltairBeaconState` (unfortunately). But I think `is_merge_transition_block` is meant to only be applied with the `BellatrixBeaconState`.

I'm inclined to only have this check in `on_block` function to avoid having more Bellatrix functions be called with the old version `BeaconState`.

Also, we will have to mock `BELLATRIX_FORK_EPOCH` in some tests (with `@spec_configured_state_test` decorator + set `BELLATRIX_FORK_EPOCH` in `meta.yaml`). If we only check it in fork-choice specs, we only have to mock it in the fork-choice test format.

mkalinin:
I am curious if this kind of change is required by implementations or the spec? I believe implementations already handle this in a way similar to what @terencechain wrote in the initial message. If this is not important for the spec either then I'd probably leave it as is now

terencechain:
> I am curious if this kind of change is required by implementations or the spec? I believe implementations already handle this in a way similar to what @terencechain wrote in the initial message. If this is not important for the spec either then I'd probably leave it as is now

I'm leaning towards implementation detail at this point but curious what other client implementers think

ajsutton:
Given `is_execution_enabled` is called in `process_block` and `process_block` is only called after `state_transition` has already run `process_slots(state, block.slot)` I'm not sure how you'd ever get an Altair state here.  The state has to already be at the same slot as the block so would have already been converted to Bellatrix. 

Maybe I'm missing somewhere else `is_merge_transition_block` winds up being called?

PR #2836, opened 2022-02-23
Withdrawals push

djrtwo:


specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,302 @@
...
+<!-- TOC -->
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+## Introduction
+
+Capella is a consensus-layer upgrade containin a number of features related
```
ralexstokes, line 14:
```suggestion
Capella is a consensus-layer upgrade containing a number of features related
```

specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,302 @@
...
+
+| Name | Value | Unit | Duration |
+| - | - | :-: | :-: |
+| `WITHDRAWAL_TRANSACTION_LIMIT` | `uint64(2**40)` (= 1,099,511,627,776) | withdrawal transactions enqueued in state|
+
+### Execution
+
+| Name | Value | Description |
+| - | - | - |
+| `TX_TYPE_WITHDRAWAL` | `TransactionType('0x05')` | EIP-2718 TX Type |
```
ralexstokes, line 43:
```suggestion
| `TX_TYPE_WITHDRAWAL` | `TransactionType('0x03')` | EIP-2718 TX Type |
```

after skimming the EIPs, looks like `0x3` is free

mcdee:
Very much like the direction of travel here.

Regarding having a separate `withdrawal_transactions`: I think that it makes sense to keep these separate from `transactions` in the execution payload.  Although they are named as transactions, with the push based system and non-evm execution of them on the execution layer side they aren't really transactions in any traditional sense.  It also keeps the PBS logic simple: the returned payload can have be returned with a `withdrawal_transactions` list and the beacon node can fill it in prior to returning it to the validator client for signing.

(On naming: I'd suggest they be of type `Withdrawal` rather than `WithdrawalTransaction` for clarity, and sit in their own `withdrawals` entry in the execution payload.)

Regarding generation of withdrawal transactions for active validators:  although limiting generation to the proposer each slot is simple, it means that the action is random in terms of timing.  Instead, how about a system where all validators are swept on a periodic basis?  Say for example that `SKIMMING_PERIOD_SLOTS` is `2**19` which is just over two months.  At each slot a number of withdrawal receipts are created automatically as follows:

- calculate `skim_mod = current slot % SKIMMING_PERIOD_SLOTS`
- all validators where `skim_mod == index % SKIMMING_PERIOD_SLOTS` and is valid for skimming (active, 0x01 credentials, balance > 32 ETH) have a withdrawal for any value above 32 ETH created automatically

This would be low-volume (max of 1 withdrawal per slot per 512K validators) but would be regular, which avoids unlucky validators being penalized even more than they are now.

specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,305 @@
...
+New EIP-2718 transaction type, with the format being the single byte `TX_TYPE_WITHDRAWAL`
+followed by an SSZ encoding of the `WithdrawalTransaction` container comprising the transaction contents.
+
+*Note*: This container is used for both a special TX that goes into an `ExecutionPayload`
+as well as in the `BeaconState`'s `withdrawal_receipts` queue.
+
+```python
+class WithdrawalTransaction(Container):
+    address: ExecutionAddress
+    amount: Gwei
```
tynes, line 171: Is the amount defined in gwei to save on serialization size?

djrtwo: we don't currently have uint256 arithmetic in CL so avoiding the conversion to a larger bitsize in CL currently. That said, I'm not sold on that. It can and maybe should be converted here so it is a native EL value before landing in there

ralexstokes: i can say while working here: https://github.com/ethereum/go-ethereum/pull/24468

it would simplify a lot of EL implementation if this was a `Bytes32`  value -- right now, accepting Gwei means we have to convert from Gwei to Wei at every IO boundary into the EL and implementers have to remember the conversion needs to be done.

it does seem simpler to do something like this once and then drop all of the conversion inside/around the EL:


```python
amount_in_gwei = ssz.encode(amount)
amount_for_receipt = amount_in_gwei + "\x00" * 24
```

tynes: I don't think its ideal to have the value field be in different units depending on the transaction type from the pov of the EL. In its current form, would `Gwei` be expected to be a uint64? I don't see a `Gwei` typedef. Adding uint256 arithmetic to the CL feels a bit leaky, but it could be safer to use to represent ether balances as to not make assumptions about the ether supply far into the future

djrtwo: yeah, I agree. It was more of a placeholder with an existing type before doing the work to add Wei here

specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,305 @@
...
+    pubkey: BLSPubkey
+    withdrawal_credentials: Bytes32  # Commitment to pubkey for withdrawals
+    effective_balance: Gwei  # Balance at stake
+    slashed: boolean
+    # Status epochs
+    activation_eligibility_epoch: Epoch  # When criteria for activation were met
+    activation_epoch: Epoch
+    exit_epoch: Epoch
+    withdrawable_epoch: Epoch  # When validator can withdraw funds
+    withdrawn_epoch: Epoch  # [New in Capella]
```
dapplion, line 65: :cry: F s in the chat for Validator container getting +1 depth

djrtwo: We don't actually need to do this. We could us another signal to signify withdrawn and wrap it in a function `is_withdrawn`.

e.g. set *all* `*_epoch` to `GENESIS_EPOCH` and then `is_withdrawn` just checks that. Something like that

mcdee: Could it be `withdrawable_epoch <= current_epoch AND effective_balance == 0` (and setting `effective_balance` to 0 could be done in place of setting `withdrawn_epoch` as at current).

dapplion: > e.g. set all *_epoch to GENESIS_EPOCH and then is_withdrawn just checks that. Something like that

That sounds definitely better than the extra field

specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,310 @@
...
+#### Withdrawals
+
+*Note*: The function `process_full_withdrawals` is new.
+
+```python
+def process_full_withdrawals(state: BeaconState) -> None:
+    current_epoch = get_current_epoch(state)
+    for index, validator in enumerate(state.validators):
+        if is_fully_withdrawable_validator(validator, current_epoch):
+            # TODO, consider the zero-balance case
```
ralexstokes, line 240: if the validator is `fully_withdrawable` and has `0` balance, then one option is to skip the call to `withdraw` (to not clutter the receipts queue) although we still want to update `validator.withdrawn_epoch`

if we went this route, just need to make sure no important accounting (e.g. to signal validator record re-useability) is done in the function `withdraw`

mkalinin: I think we may have zero balance case handled inside of `withdraw` function. This makes the function well isolated and responsible for entire withdrawal flow which prevent potential issues caused by forgetting anything related. Also,  `withdraw_epoch` should be set inside of this function.

djrtwo: This function will also be used for partial withdrawals (e.g. skimming off balances above 32ETH during block proposal or some othr method). That's why it does *not* update `withdrawal_epoch` and why the balance is passed in (insteads of an assumed max balance withdrawal)

mkalinin: For the sake of single responsibility principle, I would make a separate function for partial withdrawals if this does make sense.

djrtwo: For the sake of code reuse and testing, I'm not actually sure. We have instances of core functionality making it into helper functions and used in multipel contexts -- e.g. slash_validator, initiate_validator_exit, etc

vshvsh:
We've been working on a proposal for execution layer only push with a mechanism closer to how the deposit contract works: 
[generalized message bus](https://hackmd.io/pTGg4JP_TJi2TrpkNYu-Xw), [partial withdrawals based on it](https://hackmd.io/ZnoVEuvxT2WwFYY60RETgg). Maybe this will be of use in the discussion.

specs/capella/beacon-chain.md
dapplion: For context, the extra field adds 312500*8 = 2.5MB to a mainnet consensus state, current size is `42.2MB` so it's not insignificant.

hwwhww (commented):
Great work!

I've run `--preset=mainnet` with this PR successfully.

Should we finish `tests/core/pyspec/eth2spec/test/capella/fork/test_capella_fork_random.py` in this PR? Or leave a TODO comment here?

specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,299 @@
...
+            withdraw(state, ValidatorIndex(index), state.balances[index])
+            validator.withdrawn_epoch = current_epoch
```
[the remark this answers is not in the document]
mkalinin: Yes, but `withdrawn_epoch` or any other attribute indicating that validators has been withdrawn must be set only in case of *full* withdrawal. If this would be partial withdrawal of excess balance then `withdrawn_epoch` must remain unset. This is my understanding of the design. I think we may have two functions re-using `withdraw`, each with the logic dedicated to a particular case. For instance, `withdraw_fully` may check the conditions and set `withdrawn_epoch` in addition.

specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,299 @@
...
+```
+
+### Block processing
+
+```python
+def process_block(state: BeaconState, block: BeaconBlock) -> None:
+    process_block_header(state, block)
+    if is_execution_enabled(state, block.body):
```
mkalinin, line 242: I think that this check and other checks in the spec that are related to the Merge transition should be removed in a cleanup fork after the Merge. Not directly related to Cappella but worth keeping in mind.

specs/capella/fork-choice.md
```diff
@@ -0,0 +1,53 @@
...
+*Note*: The `notify_forkchoice_updated` function is modified in the `ExecutionEngine` protocol at the Capella upgrade.
+
+#### `notify_forkchoice_updated`
+
+The only change made is to the `PayloadAttributes` container through the addition of `withdrawals`.
+Otherwise, `notify_forkchoice_updated` inherits all prior functionality.
+
+```python
+def notify_forkchoice_updated(self: ExecutionEngine,
+                              head_block_hash: Hash32,
```
mkalinin, line 34: Consider adjusting with https://github.com/ethereum/consensus-specs/pull/2851

specs/capella/beacon-chain.md
```diff
@@ -0,0 +1,299 @@
...
+#### New `process_withdrawals`
+
+```python
+def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
+    num_withdrawals = min(MAX_WITHDRAWALS_PER_PAYLOAD, len(state.withdrawals_queue))
+    dequeued_withdrawals = state.withdrawals_queue[:num_withdrawals]
+
+    assert len(dequeued_withdrawals) == len(payload.withdrawals)
+    for dequeued_receipt, withdrawal in zip(dequeued_withdrawals, payload.withdrawals):
+        assert dequeued_receipt == withdrawal
```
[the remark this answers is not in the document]
ralexstokes: we *must* check `payload.withdrawals` here -- this is how we know that the `payload` is well-formed with respect to the consensus `state`, and that this is indeed a valid block

otherwise a malicious proposer could stuff arbitrary withdrawals into the `payload` and feed them to the execution layer...

ralexstokes, line 260: this is a more compact alternative that should have the same semantics:
```suggestion
    assert dequeued_withdrawals == payload.withdrawals
```

and you could even skip the prior length check.

an argument for keeping it as-is is that it more explicitly lists the conditions we are to check

specs/capella/fork-choice.md
```diff
@@ -0,0 +1,53 @@
...
+                              finalized_block_hash: Hash32,
+                              payload_attributes: Optional[PayloadAttributes]) -> Optional[PayloadId]:
```
ralexstokes, line 36: this seems to be the convention:
```suggestion
                              payload_attributes: Optional[PayloadAttributes]) -> Optional[PayloadId]:  # new in Capella
```

specs/capella/fork-choice.md
```diff
@@ -0,0 +1,53 @@
...
+
+`PayloadAttributes` is extended with the `withdrawals` field.
+
+```python
+@dataclass
+class PayloadAttributes(object):
+    timestamp: uint64
+    prev_randao: Bytes32
+    suggested_fee_recipient: ExecutionAddress
+    withdrawals: Sequence[Withdrawal]
```
ralexstokes, line 52: ditto:
```suggestion
    withdrawals: Sequence[Withdrawal]  # new in Capella
```

specs/capella/validator.md
```diff
@@ -0,0 +1,91 @@
...
+
+        terminal_pow_block = get_terminal_pow_block(pow_chain)
+        if terminal_pow_block is None:
+            # Pre-merge, no prepare payload call is needed
+            return None
+        # Signify merge via producing on top of the terminal PoW block
+        parent_hash = terminal_pow_block.block_hash
+    else:
+        # Post-merge, normal payload
+        parent_hash = state.latest_execution_payload_header.block_hash
```
ralexstokes, line 81: imo we can drop this entire arm of logic (and the requisite function params) bc we can assume the merge transition will be complete

ralexstokes (commented):
great work! left some comments

edit: ~i'll make a separate pass on the tests~ left one comment on a test case, otherwise they are a great start!

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+@with_capella_and_later
+@spec_state_test
+def test_fail_one_in_queue_two_in_withdrawals(spec, state):
+    prepare_withdrawals_queue(spec, state, 1)
+
+    next_slot(spec, state)
+    execution_payload = build_empty_execution_payload(spec, state)
+    execution_payload.withdrawals.append(execution_payload.withdrawals[0].copy())
```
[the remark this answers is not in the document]
djrtwo: right -- I kind of want this code-path to *not* give a shit. That should be handled by any withdraw initiation path. Essentially, if it's in the queue, you dequeue upon these rules.

specs/capella/beacon-chain.md
djrtwo: going to leave as is for now. I like that it is verbose on the precise error/line when dbugging/testing

specs/capella/validator.md
djrtwo: will do so in a separate PR

hwwhww (approved):
lgtm!

Mitumaru (commented):
Special

PR #2837, opened 2022-02-24
Ensure that the SSZ object classes are ordered

hwwhww:
### Issue
After `combine_ssz_objects` call, the extended SSZ objects would replace the old ones -- that may make `dependency_order_class_objects` not enough to detect the wrong order in one call.

### How did I fix it
This workaround simply runs more times of `dependency_order_class_objects` until it's surely fully sorted. Still thinking about if there is another better solution.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,9 @@ from abc import ABC, abstractmethod
 import ast
 import subprocess
 import sys
+import copy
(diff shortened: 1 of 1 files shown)


djrtwo (approved):
works for me!! thank you @hwwhww

PR #2841, opened 2022-02-25
Add optimistic import for blocks for VALID ancestors

paulhauner:
Allow a node to optimistically import a block if it has an ancestor which:

1. Has execution enabled.
2. Has a `VALID` payload.

The `VALID` ancestor infers that the transition block was available, therefore this chain is not performing a malicious fork choice "poisoning" attack.

Implementations are permitted to stop searching at the finalized block, if they wish. This means implementations can use fork choice to store the execution status, they don't need to maintain it in the database for ancient blocks.

### DIFF
--- a/sync/optimistic.md
+++ b/sync/optimistic.md
@@ -84,7 +84,20 @@ def is_optimistic_candidate_block(opt_store: OptimisticStore, current_slot: Slot
     justified_root = opt_store.block_states[opt_store.head_block_root].current_justified_checkpoint.root
     justified_is_execution_block = is_execution_block(opt_store.blocks[justified_root])
     block_is_deep = block.slot + SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY <= current_slot
-    return justified_is_execution_block or block_is_deep
+    return justified_is_execution_block or block_is_deep or has_verified_ancestor(opt_store, block)
+```
+
+```python
+def has_verified_ancestor(opt_store: OptimisticStore, block: BeaconBlock):
+	while True:
+		if not is_execution_block(block):
+			return false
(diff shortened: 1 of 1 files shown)


paulhauner:
I've fixed a flipped-bool in [60b34dc](https://github.com/ethereum/consensus-specs/pull/2841/commits/60b34dcf14afe2fdd8ce0fd9b448e7e220f8995a), as pointed out by @potuz :pray:

sync/optimistic.md
```diff
@@ -81,10 +81,29 @@ def is_execution_block(block: BeaconBlock) -> bool:
...
+    while True:
+        if not is_execution_block(block):
+            return False
+        elif hash_tree_root(block) not in opt_store.optimistic_roots:
+            return True
+        # Implementations MAY choose to remove this check and iterate further
+        # back in the chain.
+        elif hash_tree_root(block) == finalized_root:
+            return False
+        elif block.parent_root == Hash32() or block.parent_root not in opt_store.blocks:
```
mkalinin, line 103: Won't just `block.parent_root not in opt_store.blocks` be sufficient in this case?

sync/optimistic.md
```diff
@@ -106,6 +125,11 @@ A block MAY be optimistically imported when
 1. The current slot (as per the system clock) is at least
    `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` ahead of the slot of the block being
    imported.
+1. The block has a ancestor with an execution payload that has been deemed to
```
djrtwo, line 128:
```suggestion
1. The block has an ancestor with an execution payload that has been deemed to
```

paulhauner:
I'm closing this in favor of the more general solution in #2844

PR #2844, opened 2022-03-01
Optimistically import any post-merge block

paulhauner:
This PR implements a suggestion by @mkalinin to allow optimistic import of any post-merge block.

A post-merge block is any block where the parent block has a non-empty `ExecutionPayload`.

The fork choice poisoning attack relies upon the ability to *reliably* send *all* execution engines into a syncing state. This is only possible on the *merge block*, since that block can reference a junk `payload.parent_hash`. So, it makes sense to only apply these optimistic-import restrictions to the merge block (and not their innocent descendants).

This PR is an improvement since it avoids rejecting valid blocks in some scenarios where it is safe to (optimistically) import them.

This PR succeeds #2841 since it is a simpler and more general solution.

### DIFF
--- a/sync/optimistic.md
+++ b/sync/optimistic.md
@@ -81,10 +81,17 @@ def is_execution_block(block: BeaconBlock) -> bool:
 
 ```python
 def is_optimistic_candidate_block(opt_store: OptimisticStore, current_slot: Slot, block: BeaconBlock) -> bool:
+	if is_execution_block(opt_store.blocks[block.parent_root]):
+		return True
+
     justified_root = opt_store.block_states[opt_store.head_block_root].current_justified_checkpoint.root
-    justified_is_execution_block = is_execution_block(opt_store.blocks[justified_root])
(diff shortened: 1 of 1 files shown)


sync/optimistic.md
```diff
@@ -99,14 +106,21 @@ behaviours without regard for optimistic sync.
...
+This ensures that blocks are only optimistically imported if one or more of the
+following are true:
 
+1. The parent of the block has execution enabled.
 1. The justified checkpoint has execution enabled.
 1. The current slot (as per the system clock) is at least
    `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` ahead of the slot of the block being
    imported.
 
+In effect, there are restrictions on when a *merge block* can be optimistically
```
mkalinin, line 119: I would be consistent denoting merge transition block. As it's being called "merge block" and "transition block" in two sister sentences is a bit confusing.

paulhauner: Good point. I noticed we used "merge block" in other parts of the document so I standardized on that.

djrtwo (approved):

PR #2845, opened 2022-03-01
Remove equivocating validators from fork choice consideration

adiasg:
Discard equivocating attestations from the fork choice `Store`, and discount all future votes equivocating validators for fork choice purposes. Specifically:
- Introduce a new `on_attester_slashing` handler in the fork choice that should be called whenever an `AttesterSlashing` is received
- Maintain the set of equivocating validators in `Store.equivocating_indices`
- When executing `get_latest_attesting_balance`, filter the validator set using `Store.equivocating_indices` to drop those validators

**Note:** This patches the equivocation balancing attack noted [in Section 4 here](https://arxiv.org/pdf/2203.01315.pdf).
**Note 2:** The protocol is not vulnerable to the avalanche attack (section 3) because equivocations will be discarded.

### DIFF
--- a/specs/phase0/fork-choice.md
+++ b/specs/phase0/fork-choice.md
@@ -32,6 +32,7 @@
     - [`on_tick`](#on_tick)
     - [`on_block`](#on_block)
     - [`on_attestation`](#on_attestation)
+    - [`on_attester_slashing`](#on_attester_slashing)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 <!-- /TOC -->
@@ -181,7 +182,7 @@ def get_latest_attesting_balance(store: Store, root: Root) -> Gwei:
     attestation_score = Gwei(sum(
         state.validators[i].effective_balance for i in active_indices
         if (i in store.latest_messages
-            and not i in store.has_equivocated
+            and i not in store.has_equivocated
             and get_ancestor(store, store.latest_messages[i].root, store.blocks[root].slot) == root)
     ))
     if store.proposer_boost_root == Root():
@@ -469,7 +470,8 @@ def on_attestation(store: Store, attestation: Attestation, is_from_block: bool=F
 ```python
 def on_attester_slashing(store: Store, attester_slashing: AttesterSlashing) -> None:
     """
-    Run ``on_attester_slashing`` upon receiving a new ``AttesterSlashing`` from either within a block or directly on the wire.
+    Run ``on_attester_slashing`` upon receiving a new ``AttesterSlashing`` from either within a block or directly 
+    on the wire.
     """
     attestation_1 = attester_slashing.attestation_1
     attestation_2 = attester_slashing.attestation_2
--- a/specs/phase0/fork-choice.md
+++ b/specs/phase0/fork-choice.md
@@ -101,6 +101,7 @@ class Store(object):
     finalized_checkpoint: Checkpoint
     best_justified_checkpoint: Checkpoint
     proposer_boost_root: Root
+    has_equivocated: Set[ValidatorIndex]
     blocks: Dict[Root, BeaconBlock] = field(default_factory=dict)
     block_states: Dict[Root, BeaconState] = field(default_factory=dict)
     checkpoint_states: Dict[Checkpoint, BeaconState] = field(default_factory=dict)
@@ -129,6 +130,7 @@ def get_forkchoice_store(anchor_state: BeaconState, anchor_block: BeaconBlock) -
         finalized_checkpoint=finalized_checkpoint,
         best_justified_checkpoint=justified_checkpoint,
         proposer_boost_root=proposer_boost_root,
+        has_equivocated=set(),
         blocks={anchor_root: copy(anchor_block)},
         block_states={anchor_root: copy(anchor_state)},
         checkpoint_states={justified_checkpoint: copy(anchor_state)},
@@ -179,6 +181,7 @@ def get_latest_attesting_balance(store: Store, root: Root) -> Gwei:
     attestation_score = Gwei(sum(
         state.validators[i].effective_balance for i in active_indices
         if (i in store.latest_messages
+            and not i in store.has_equivocated
             and get_ancestor(store, store.latest_messages[i].root, store.blocks[root].slot) == root)
     ))
     if store.proposer_boost_root == Root():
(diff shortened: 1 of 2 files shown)


specs/phase0/fork-choice.md
```diff
@@ -358,8 +362,9 @@ def update_latest_messages(store: Store, attesting_indices: Sequence[ValidatorIn
     target = attestation.data.target
     beacon_block_root = attestation.data.beacon_block_root
     for i in attesting_indices:
-        if i not in store.latest_messages or target.epoch > store.latest_messages[i].epoch:
-            store.latest_messages[i] = LatestMessage(epoch=target.epoch, root=beacon_block_root)
+        if i not in store.has_equivocated:
```
djrtwo, line 365: suggest prefiltering the set to avoid the extra nested for loop

```python
non_equivocating_attesting_indices = [i for i in attesting_indices if i not in store.has_equivocated
for i in attesting_indices:
    ...

File: specs/phase0/fork-choice.md [hunk 2]
```diff
@@ -101,6 +102,7 @@ class Store(object):
     finalized_checkpoint: Checkpoint
     best_justified_checkpoint: Checkpoint
     proposer_boost_root: Root
+    has_equivocated: Set[ValidatorIndex]
```
djrtwo, line 105: `has_equivocated` reads like a predicate rather than a set. I suggest we name it `equivocated_indices` or `equivocating_indices`

specs/phase0/fork-choice.md
```diff
@@ -459,3 +464,23 @@ def on_attestation(store: Store, attestation: Attestation, is_from_block: bool=F
...
+    """
+    attestation_1 = attester_slashing.attestation_1
+    attestation_2 = attester_slashing.attestation_2
+    assert is_slashable_attestation_data(attestation_1.data, attestation_2.data)
+    state = store.block_states[store.justified_checkpoint.root]
+    assert is_valid_indexed_attestation(state, attestation_1)
+    assert is_valid_indexed_attestation(state, attestation_2)
+
+    indices = set(attestation_1.attesting_indices).intersection(attestation_2.attesting_indices)
+    for index in sorted(indices):
```
djrtwo, line 484: why sorted? `has_equivocated` is a set so it doesn't really matter the order they are inserted in

djrtwo (commented):
very nice! a couple of minor suggestions

specs/phase0/fork-choice.md
```diff
@@ -459,3 +464,23 @@ def on_attestation(store: Store, attestation: Attestation, is_from_block: bool=F
...
     update_latest_messages(store, indexed_attestation.attesting_indices, attestation)
 ```
+
+#### `on_attester_slashing`
+
+```python
+def on_attester_slashing(store: Store, attester_slashing: AttesterSlashing) -> None:
+    """
+    Run ``on_attester_slashing`` upon receiving a new ``AttesterSlashing`` from either within a block or directly 
+    on the wire.
```
djrtwo, line 474:
```suggestion
    Run ``on_attester_slashing`` immediately upon receiving a new ``AttesterSlashing``
    from either within a block or directly on the wire.
```

specs/phase0/fork-choice.md
```diff
@@ -459,3 +464,25 @@ def on_attestation(store: Store, attestation: Attestation, is_from_block: bool=F
     # Update latest messages for attesting indices
     update_latest_messages(store, indexed_attestation.attesting_indices, attestation)
 ```
+
+#### `on_attester_slashing`
+
+*Note*: `on_attester_slashing` should be called while syncing and a node MUST maintain the equivocation set of `AttesterSlashing`s from at least the last finalized checkpoint
```
djrtwo, line 470:
```suggestion
*Note*: `on_attester_slashing` should be called while syncing and a client MUST maintain the equivocation set of `AttesterSlashing`s from at least the latest finalized checkpoint.
```

djrtwo (approved):

specs/phase0/fork-choice.md
```diff
@@ -459,3 +464,25 @@ def on_attestation(store: Store, attestation: Attestation, is_from_block: bool=F
...
+```python
+def on_attester_slashing(store: Store, attester_slashing: AttesterSlashing) -> None:
+    """
+    Run ``on_attester_slashing`` immediately upon receiving a new ``AttesterSlashing``
+    from either within a block or directly on the wire.
+    """
+    attestation_1 = attester_slashing.attestation_1
+    attestation_2 = attester_slashing.attestation_2
+    assert is_slashable_attestation_data(attestation_1.data, attestation_2.data)
+    state = store.block_states[store.justified_checkpoint.root]
```
michaelsproul, line 481: Why do we use the state of the justified block here? It seems strange, as the slashing will have been verified previously against the head state and may fail verification here, or may verify with a different set of slashable indices. In Lighthouse I'm using the head state for this reason, but can change to justified if necessary

Lighthouse comment about this: https://github.com/sigp/lighthouse/blob/94bc85e1f731217c113ff79b427cdeb73b3309cd/consensus/fork_choice/src/fork_choice.rs#L1108-L1116

Lighthouse PR: https://github.com/sigp/lighthouse/pull/3371

cc @adiasg

adiasg: > Why do we use the state of the justified block here?

To maintain uniformity with the fork choice - the [validator set at the justified state](https://github.com/ethereum/consensus-specs/blame/9f643d8dbf678430bc6707f0e5b914d0bf62545c/specs/phase0/fork-choice.md#L177-L183) is used for computing the fork choice. 

> It seems strange, as the slashing will have been verified previously against the head state and may fail verification here, or may verify with a different set of slashable indices.

The intention of `on_attester_slashing` is to censor equivocating validators as soon as an implicating `AttesterSlashing` is received by the node, irrespective of whether those validators are slashable. We are not checking `is_slashable_validator` here (which is quite dependent on the state!), rather only `is_slashable_attestation_data` & `is_valid_indexed_attestation` which do not have much state dependence.

The Lighthouse PR is adding to `store.equivocating_indices` only those validator indices from `attester_slashing` that are slashable in the head state. This should be changed to remove the slashability check.

michaelsproul: Thank you!

Sorry I don't know what got into me, I got so caught up in re-using `get_slashable_indices` that I didn't realise I shouldn't be filtering already slashed validators at all. Thanks again

adiasg: No worries at all! 😄

PR #2847, opened 2022-03-03
Ignore subset aggregates

arnetheduck:
When aggregates are propagated through the network, it is often the case
that a better aggregate has already been seen - in particular, this
happens when an aggregator has not been able to include itself in the
mesh and therefore publishes an aggregate with only its own
attestations.

This new ignore rule allows dropping all aggregates that are
(non-strict) subsets of aggregates that have already been seen on the
network. In particular, it does not mandate dropping aggregates where a
union of previous aggregates would cause it to become a subset).

The logic for allowing this is based on the premise that any aggregate
that has already been seen by a peer will also have been seen by its
neighbours - a subset aggregate (strict or not) brings no new value to
the aggregation algorithm, except in the extreme edge case where you
could combine several such sparse aggregates into a single, more dense
"combined" aggregate and thus use less block space.

Further, as a small benefit, computing the `hash_tree_root` of the full
aggregate is generally not done -however, `hash_tree_root(data)` is
already done for other purposes as this is used as index in the beacon
API.

### DIFF
--- a/specs/altair/p2p-interface.md
+++ b/specs/altair/p2p-interface.md
@@ -144,6 +144,7 @@ def get_sync_subcommittee_pubkeys(state: BeaconState, subcommittee_index: uint64
 - _[REJECT]_ `contribution_and_proof.selection_proof` selects the validator as an aggregator for the slot -- i.e. `is_sync_committee_aggregator(contribution_and_proof.selection_proof)` returns `True`.
 - _[REJECT]_ The aggregator's validator index is in the declared subcommittee of the current sync committee --
   i.e. `state.validators[contribution_and_proof.aggregator_index].pubkey in get_sync_subcommittee_pubkeys(state, contribution.subcommittee_index)`.
+- _[IGNORE]_ A valid sync committee contriubtion with equal `slot`, `beacon_block_root` and `subcommittee_index` whose `aggregation_bits` is non-strict superset has _not_ already been seen.
 - _[IGNORE]_ The sync committee contribution is the first valid contribution received for the aggregator with index `contribution_and_proof.aggregator_index`
   for the slot `contribution.slot` and subcommittee index `contribution.subcommittee_index`
   (this requires maintaining a cache of size `SYNC_COMMITTEE_SIZE` for this topic that can be flushed after each slot).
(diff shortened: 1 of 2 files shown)


dapplion:
How would this affect gossip scoring? Should the current parameters be fine-tuned?

dapplion:
Also, this should be tested in advance to not require a revert down the line like what happened with https://github.com/ethereum/consensus-specs/issues/2183

ajsutton:
Should we consider the same change for SignedContributionAndProof messages as well?

djrtwo:
> How would this affect gossip scoring? Should the current parameters be fine-tuned?

This should be fine because it's an IGNORE which means the sender is not considered dishonest



> Also, this should be tested in advance to not require a revert down the line like what happened with #2183

certainly. Should note though, that this actually piggy backs on the 2183 functionality of dropping dupes so would in theory *reduce* bandwidth through a known mechanism

arnetheduck:
> Should we consider the same change for SignedContributionAndProof messages as well?

Sure, why not :) d5f4c4618301980cdc9993ddbb2f8478784ee412

arnetheduck:
> How would this affect gossip scoring? Should the current parameters be fine-tuned?

The spec says:
`These are currently under investigation and will be spec'd and released to mainnet when they are ready.` :)

Meaning this is in client territory for now - if your client has a score for an "expected message volume" not being met on the aggregate channel, seeing fewer messages will case gossipsub descoring. 

Further, if your client penalizes peers for `IGNORE` messages, it will bias itself against those not implementing the new rule, since they will be sending messages that your client now ignores thus increasing the risk of a network partition as the bias sets in over time.

ajsutton:
Implemented this in Teku. The implementation was quite straight forward and early indications are that there's a small reduction in CPU usage as a result.

arnetheduck:
> small reduction in CPU usage as a result.

hopefully you should see a reduction in outgoing bandwidth as well! 

over time, this should translate to an incoming bandwidth reduction as well.

ajsutton:
Yes, also a slight decrease in outgoing bandwidth.

djrtwo (approved):

dapplion:
With

> it does not mandate dropping aggregates where a union of previous aggregates would cause it to become a subset

The word "not mandate" means "MAY NOT drop ..", or "MUST NOT drop .."? It's definitely cheaper to check against a single bitfield per attestation data than to look over all known aggregates. At least Lodestar pre-aggregates them in the operation pool, so we don't retain all aggregates from gossip as received.

PR #2850, opened 2022-03-10
EIP-4844: consensus-layer specs of Shard Blob Transactions

protolambda:
This PR outlines the consensus-layer changes for [EIP-4844](https://eips.ethereum.org/EIPS/eip-4844).

Changes (split across multiple docs like other hardforks, in a "feature directory" named `eip4844`):
- Update `BeaconBlockBody` to track the KZG commitments of the blobs
- Update the beacon block processing to cross validate KZG commitments and blob-transactions within the execution payload, without duplicating the transaction definition
- Introduce `BlobsSidecar` to sync blobs. ~~Also note the `shard` field for forward compatibility - I think we can afford the mere 8 zero bytes in the blobs type to be compatible with a likely path forward with full sharding.~~ (edit: not adding the `shard` field anymore)
- Introduce `SignedBlobsSidecar` to gossip blobs: initially the signature is available. Long-term we cannot rely on the sidecar signature being available however, blobs are synced without signture (but still verified with KZG against the beacon block). The signature functions as a spam protection in gossip.
- Introduce a temporary `fork.md` to define how to proceed with EIP-4844 testnet upgrade.
- Define `BlobsSidecarsByRange` to sync blobs: no by-root equivalent since the expectation is to sync blobs after having a reference beacon chain that misses them. Additionally the serving span is reduced to a month (data availability period caps it naturally) compared to that of 5 months of beacon blocks. And no `step` here, based on previous complaints dev-feedback about the beacon blocks variant.
- Define block production, including KZG commitment and versioned-hash sanity checks, and signing of the gossip object.

Possible TODO (or maybe for a later PR): the honest-validator spec covers the blob download obligation currently, but we should maybe emphasize it as a forkchoice-dependency. I am not aware of a dependency system in the existing fork-choice docs though, so it may require some refactoring (help/feedback appreciated there).

Once the EIP has gone through more testing and review, and once there is rough-consensus on inclusion in a hardfork, then we can merge the `eip4844` feature-directory with other consensus features of that hardfork. The next consensus fork after the Merge, dubbed `Capella` currently, being the primary target.

In a follow-up PR I can work on enabling pyspec generation from this feature set, and corresponding testing. Need a stable base + format first before we refactor code-gen to support features instead of just phases.


specs/eip4844/validator.md
```diff
@@ -0,0 +1,135 @@
...
+```python
+def verify_blobs_sidecar(slot: Slot, beacon_block_root: Root,
+                         expected_kzgs: Sequence[KZGCommitment], blobs_sidecar: BlobsSidecar):
+    assert blobs_sidecar.shard == 0  # always zero, placeholder for future sharding
+    assert slot == blobs_sidecar.beacon_block_slot
+    assert beacon_block_root == blobs_sidecar.beacon_block_root
+    blobs = blobs_sidecar.blobs
+    assert len(kzgs) == len(blobs)
+    for kzg, blob in zip(expected_kzgs, blobs):
+      assert blob_to_kzg(blob) == kzg
```
terencechain, line 65: `blob_to_kzg` is undefined, refer to the EIP?

specs/eip4844/validator.md
```diff
@@ -0,0 +1,135 @@
...
+
+#### Constructing the `BeaconBlockBody`
+
+##### Blob commitments
+
+After retrieving the execution payload from the execution engine as specified in Bellatrix,
+the blobs are retrieved and processed: 
+
+```python
+# execution_payload = xecution_engine.get_payload(payload_id)
```
terencechain, line 84:
```suggestion
# execution_payload = execution_engine.get_payload(payload_id)
```

specs/eip4844/validator.md
```diff
@@ -0,0 +1,135 @@
...
+    beacon_block_slot=beacon_block.slot
+    shard=0,
+    blobs=blobs,
+)
+```
+
+And then signed:
+
+```python
+domain = get_domain(state, DOMAIN_BLOBS_SIDECAR, blobs_sidecar.beacon_block_slot / SLOTS_PER_EPOCH)
```
terencechain, line 122: `DOMAIN_BLOBS_SIDECAR` is undefined

specs/eip4844/validator.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Helpers
+
+### `is_data_available`
+
+The implementation of `is_data_available` is meant to change with later sharding upgrades.
+Initially, it requires every verifying actor to retrieve the matching `BlobsSidecar`,
+and verify the sidecar with `verify_blobs`.
+
+Without the sidecar the block may be processed further optimistically,
```
terencechain, line 45: Perhaps we can refer to the optimistic doc here if the validator/RPC/p2p behavior remains the same

specs/eip4844/beacon-chain.md
```diff
@@ -0,0 +1,165 @@
...
+
+#### Blob KZGs
+
+```python
+def process_blob_kzgs(body: BeaconBlockBody):
+    assert verify_kzgs_against_transactions(body.execution_payload.transactions, body.blob_kzgs)
+
+    # TODO do we want to buffer the kzg commitments in the BeaconState, like in the full sharding design?
+    # This could make a proof to any particular blob commitment more efficient,
+    # but the buffer structure is also likely to change with full sharding.
```
djrtwo, line 152: Why do you expect the buffer structure to change? Shouldn't "full sharding" just be adding more of these TXs per block?

specs/eip4844/p2p-interface.md
```diff
@@ -0,0 +1,264 @@
...
+
+| Name | Value |
+| - | - |
+| `LIMIT_BLOBS_PER_SIDECAR` | `uint64(2**4)` (= 16) |
+
+## Configuration
+
+| Name                                     | Value                         | Description                                                         |
+|------------------------------------------|-------------------------------|---------------------------------------------------------------------|
+| `MAX_REQUEST_BLOBS_SIDECARS`             | `2**10` (= 1024)              | Maximum number of blobs sidecars in a single request                |
```
djrtwo, line 46: This could be a huge amount of data (on order of 1GB)!

Is this okay to be made in a single request? This seems quite excessive to *not* be load balancing to some extent across peers

specs/eip4844/p2p-interface.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Containers
+
+### `BlobsSidecar`
+
+```python
+class BlobsSidecar(Container):
+    beacon_block_root: Root
+    beacon_block_slot: Slot
+    shard: uint64  # [ Forward compatibility ]
```
djrtwo, line 59: I dont think there is going to be a shard construction at all in the future.

Additionally, I'm not sure we would use the side-car mechanism in such a case. iiuc, it's way more about rows and columns needing to be gossiped in contiguous blocks (e.g. for valildator PoC) rather than "shards"

specs/eip4844/p2p-interface.md
```diff
@@ -0,0 +1,264 @@
...
+
+This topic is used to propagate data blobs included in any given beacon block.
+
+The following validations MUST pass before forwarding the `signed_blobs_sidecar` on the network;
+Alias `sidecar = signed_blobs_sidecar.message`.
+- _[IGNORE]_ the `sidecar.beacon_block_slot` is for the current slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) -- i.e. `blobs_sidecar.beacon_block_slot == current_slot`.
+- _[REJECT]_ the `sidecar.blobs` are all well formatted, i.e. the `BLSFieldElement` in valid range (`x < BLS_MODULUS`).
+- _[REJECT]_ the beacon proposer signature, `signed_blobs_sidecar.signature`, is valid -- i.e.
+```python
+domain = get_domain(state, DOMAIN_BLOBS_SIDECAR, blobs_sidecar.beacon_block_slot / SLOTS_PER_EPOCH)
```
djrtwo, line 120:
```suggestion
domain = get_domain(state, DOMAIN_BLOBS_SIDECAR, blobs_sidecar.beacon_block_slot // SLOTS_PER_EPOCH)
```

specs/eip4844/p2p-interface.md
```diff
@@ -0,0 +1,264 @@
...
+signing_root = compute_signing_root(blobs_sidecar, domain)
+assert bls.Verify(proposer_pubkey, signing_root, signed_blob_header.signature)
+```
+  where `proposer_pubkey` is the pubkey of the beacon block proposer of `blobs_sidecar.beacon_block_slot`
+- _[IGNORE]_ The sidecar is the first sidecar with valid signature received for the `(proposer_index, sidecar.beacon_block_root)` combination,
+  where `proposer_index` is the validator index of the beacon block proposer of `blobs_sidecar.beacon_block_slot`
```
djrtwo, line 126: So it's really just `beacon_block_slot, beacon_block_root` combination because we can only have one mapping for proposer_index for a given `slot` unless we know the chain. (otherwise we must assume our canonical view)

specs/eip4844/p2p-interface.md
```diff
@@ -0,0 +1,264 @@
...
+- _[IGNORE]_ the `sidecar.beacon_block_slot` is for the current slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) -- i.e. `blobs_sidecar.beacon_block_slot == current_slot`.
+- _[REJECT]_ the `sidecar.blobs` are all well formatted, i.e. the `BLSFieldElement` in valid range (`x < BLS_MODULUS`).
+- _[REJECT]_ the beacon proposer signature, `signed_blobs_sidecar.signature`, is valid -- i.e.
+```python
+domain = get_domain(state, DOMAIN_BLOBS_SIDECAR, blobs_sidecar.beacon_block_slot / SLOTS_PER_EPOCH)
+signing_root = compute_signing_root(blobs_sidecar, domain)
+assert bls.Verify(proposer_pubkey, signing_root, signed_blob_header.signature)
+```
+  where `proposer_pubkey` is the pubkey of the beacon block proposer of `blobs_sidecar.beacon_block_slot`
+- _[IGNORE]_ The sidecar is the first sidecar with valid signature received for the `(proposer_index, sidecar.beacon_block_root)` combination,
```
djrtwo, line 125: why include `beacon_block_root` here? Why let them essentially equivocate and produce multiple beacon_block_roots here?

specs/eip4844/p2p-interface.md
```diff
@@ -0,0 +1,264 @@
...
+  where `proposer_index` is the validator index of the beacon block proposer of `blobs_sidecar.beacon_block_slot`
+
+Note that a sidecar may be propagated before or after the corresponding beacon block.
+
+Once both sidecar and beacon block are received, `verify_blobs_sidecar` can unlock the data-availability fork-choice dependency.
+
+### Transitioning the gossip
+
+See gossip transition details found in the [Altair document](../altair/p2p-interface.md#transitioning-the-gossip) for
+details on how to handle transitioning gossip topics for Bellatrix.
```
djrtwo, line 135:
```suggestion
details on how to handle transitioning gossip topics for EIP4844.
```

specs/eip4844/p2p-interface.md
```diff
@@ -0,0 +1,264 @@
...
+#### BlobsSidecarsByRange v1
+
+**Protocol ID:** `/eth2/beacon_chain/req/blobs_sidecars_by_range/1/`
+
+Request Content:
+```
+(
+  start_slot: Slot
+  count: uint64
+  shard: uint64
```
djrtwo, line 184: I don't think `shard` should be here. See above comments.

I don't think the sidecar construciton should remain at all after "full"/extended sharding

specs/eip4844/validator.md
```diff
@@ -0,0 +1,135 @@
...
+but MUST NOT be considered valid until a valid `BlobsSidecar` has been downloaded.
+
+```python
+def is_data_available(slot: Slot, beacon_block_root: Root, kzgs: Sequence[KZGCommitment]):
```
djrtwo, line 49: This seems more appropriate to define in the fork-choice document, right?

specs/eip4844/validator.md
djrtwo, line 65:
```suggestion
        assert blob_to_kzg(blob) == kzg
```

specs/eip4844/validator.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+### `verify_blobs_sidecar`
+
+```python
+def verify_blobs_sidecar(slot: Slot, beacon_block_root: Root,
+                         expected_kzgs: Sequence[KZGCommitment], blobs_sidecar: BlobsSidecar):
+    assert blobs_sidecar.shard == 0  # always zero, placeholder for future sharding
+    assert slot == blobs_sidecar.beacon_block_slot
+    assert beacon_block_root == blobs_sidecar.beacon_block_root
```
djrtwo, line 61: Should we verify that `beacon_block.slot == beacon_block_slot`? You can only have a sidecar for non-empty slots

djrtwo (commented):
nice!

specs/eip4844/beacon-chain.md
```diff
@@ -0,0 +1,165 @@
...
+  - [Block processing](#block-processing)
+    - [Blob KZGs](#blob-kzgs)
+- [Testing](#testing)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+## Introduction
+
+This upgrade adds transaction execution to the beacon chain as part of Bellatrix upgrade.
```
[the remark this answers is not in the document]
protolambda: My bad, copied that over for formatting, forgot to change the fork name. "Capella" is not confirmed yet either, I'll just change it to EIP-4844.

specs/eip4844/beacon-chain.md
```diff
@@ -0,0 +1,165 @@
...
+
+Additionally, this upgrade introduces the following minor changes:
+* Penalty parameter updates to their planned maximally punitive values
```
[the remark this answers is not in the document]
protolambda: same as above, intro was bad copy-paste

specs/eip4844/p2p-interface.md
```diff
@@ -0,0 +1,264 @@
...
+
+
+## Containers
+
+### `BlobsSidecar`
+
+```python
+class BlobsSidecar(Container):
+    beacon_block_root: Root
+    beacon_block_slot: Slot
```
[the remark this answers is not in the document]
protolambda: To make filtering based on time during gossip quick and simple.

specs/eip4844/p2p-interface.md
protolambda: good catch, we should change it to `proposer_index, beacon_block_slot` I think. We do want consecutive proposals to not get filtered. But we also want to filter spam.

specs/eip4844/validator.md
protolambda: I don't want to over-complicate it, and the fork-choice spec has no definition of a "dependency" yet. See PR description, it's already outlined as a TODO with explanation.

specs/eip4844/beacon-chain.md
protolambda: Locating specific commitments by their shard would be nice, if users want to limit their data to a specific shard.

specs/eip4844/p2p-interface.md
```diff
@@ -0,0 +1,264 @@
...
+Alias `sidecar = signed_blobs_sidecar.message`.
+- _[IGNORE]_ the `sidecar.beacon_block_slot` is for the current slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) -- i.e. `blobs_sidecar.beacon_block_slot == current_slot`.
+- _[REJECT]_ the `sidecar.blobs` are all well formatted, i.e. the `BLSFieldElement` in valid range (`x < BLS_MODULUS`).
+- _[REJECT]_ the beacon proposer signature, `signed_blobs_sidecar.signature`, is valid -- i.e.
+```python
+domain = get_domain(state, DOMAIN_BLOBS_SIDECAR, blobs_sidecar.beacon_block_slot / SLOTS_PER_EPOCH)
+signing_root = compute_signing_root(blobs_sidecar, domain)
+assert bls.Verify(proposer_pubkey, signing_root, signed_blob_header.signature)
+```
+  where `proposer_pubkey` is the pubkey of the beacon block proposer of `blobs_sidecar.beacon_block_slot`
```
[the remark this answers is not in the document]
protolambda: There is likely little overlap in sidecars between the chains (due to reordered/different transactions), so the question is if we should gossip sidecars that are not canonical by at least an epoch. Since sidecars are heavy w.r.t. data size I think it's better not to.

specs/eip4844/beacon-chain.md
```diff
@@ -0,0 +1,165 @@
...
+```python
+def process_block(state: BeaconState, block: BeaconBlock) -> None:
+    process_block_header(state, block)
+    if is_execution_enabled(state, block.body):
+        process_execution_payload(state, block.body.execution_payload, EXECUTION_ENGINE)
+    process_randao(state, block.body)
+    process_eth1_data(state, block.body)
+    process_operations(state, block.body)
+    process_sync_aggregate(state, block.body.sync_aggregate)
+    process_blob_kzgs(state, block.body) # [New in EIP-4844]
```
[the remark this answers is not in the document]
protolambda: I prefer to keep it here, it'll be a no-op anyway with empty execution-payload and blob-kzgs list.
And removing the if-statement in the next hardfork after Bellatrix would be a nice simplification.

Listed it in #2852 as possible future option.

specs/eip4844/beacon-chain.md
protolambda: Listed this as optional TODO in #2852. I'll remove the TODO text from the spec.

specs/eip4844/validator.md
protolambda: Listed as TODO (with two sub-tasks) in #2852

protolambda:
Implemented changes based on review, thank you @terencechain and @djrtwo. And opened #2852 to track the remaining items like pyspec changes and move of download requirement from validator-doc to fork-choice-doc.

djrtwo (approved):
thank you @protolambda!

we'll merge this and move to the tracking issue for subsequent convo and development.
I really like that this is in a "feature" dir so it makes the notion of merging soon much easier :)

mohammadfarari1360 (commented):
Creating

PR #2855, opened 2022-03-16
00 to 01 cred change

djrtwo:
* new operation to change 0x00 to 0x01
* tests


mcdee:
lgtm.

mcdee:
lgtm.

specs/capella/beacon-chain.md
```diff
@@ -64,6 +76,25 @@ class Validator(Container):
...
+    randao_reveal: BLSSignature
+    eth1_data: Eth1Data  # Eth1 data vote
+    graffiti: Bytes32  # Arbitrary data
+    # Operations
+    proposer_slashings: List[ProposerSlashing, MAX_PROPOSER_SLASHINGS]
+    attester_slashings: List[AttesterSlashing, MAX_ATTESTER_SLASHINGS]
+    attestations: List[Attestation, MAX_ATTESTATIONS]
+    deposits: List[Deposit, MAX_DEPOSITS]
+    voluntary_exits: List[SignedVoluntaryExit, MAX_VOLUNTARY_EXITS]
+    bls_to_execution_changes: List[SignedBLSToExecutionChange, MAX_BLS_TO_EXECUTION_CHANGES]
```
ralexstokes, line 92: if we put `bls_to_execution_changes` at the end of the block, we maintain some binary compatibility with older (by fork) block formats

ralexstokes: and note, we can still process the creds changes whenever; this comment is just about syntax

dapplion: Yes please additions at the bottom unless replacing an existing property

specs/capella/beacon-chain.md
```diff
@@ -324,3 +379,49 @@ def process_execution_payload(state: BeaconState, payload: ExecutionPayload, exe
...
+#### New `process_bls_to_execution_change`
+
+```python
+def process_bls_to_execution_change(state: BeaconState,
+                                    signed_address_change: SignedBLSToExecutionChange) -> None:
+    address_change = signed_address_change.message
+
+    assert address_change.validator_index < len(state.validators)
+
+    validator = state.validators[address_change.validator_index]
```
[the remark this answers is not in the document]
djrtwo: ah weird. I don't like the implicit exception we used in phase 0 (totally forgot about that...)

(/me quickly went to check we actually tested this! we did).

I'm okay either way. I like explicit better than implicit, personally

tests/core/pyspec/eth2spec/test/helpers/genesis.py
```diff
@@ -6,11 +6,12 @@
 
 
 def build_mock_validator(spec, i: int, balance: int):
-    pubkey = pubkeys[i]
+    active_pubkey = pubkeys[i]
+    withdrawal_pubkey = pubkeys[-1 - i]
```
ralexstokes, line 10: this is a bit weird but given the domain separation b/t active and withdrawal keys, it is prolly ok

best to just duplicate the `pubkeys` infra for withdrawal keys but this is a quick and dirty solution

djrtwo: right -- I mainly didn't want active and withdrawal key (behind the hash) to be the same (as it was).

This at least ensures they are generally different from each other and uses the pre-generated (which takes time) keys in a deterministic way

Mitumaru (approved):

specs/capella/beacon-chain.md
```diff
@@ -324,3 +380,49 @@ def process_execution_payload(state: BeaconState, payload: ExecutionPayload, exe
...
+    assert validator.withdrawal_credentials[:1] == BLS_WITHDRAWAL_PREFIX
+    assert validator.withdrawal_credentials[1:] == hash(address_change.from_bls_pubkey)[1:]
+
+    domain = get_domain(state, DOMAIN_BLS_TO_EXECUTION_CHANGE)
+    signing_root = compute_signing_root(address_change, domain)
+    assert bls.Verify(address_change.from_bls_pubkey, signing_root, signed_address_change.signature)
+
+    validator.withdrawal_credentials = (
+        ETH1_ADDRESS_WITHDRAWAL_PREFIX
+        + (0x00).to_bytes(11, 'little')
```
hwwhww, line 425: imo using `b'\x00' * 11` (as `validator.md` defined) is more clear since endianness doesn't matter here.
```suggestion
        + b'\x00' * 11
```

hwwhww (approved):
LGTM 👍

specs/capella/beacon-chain.md
djrtwo: I believe this doesn't build

PR #2858, opened 2022-03-22
Bellatrix: pass justified as a safe block

mkalinin:
Introduces a `get_safe_block_hash` function that returns a block hash of a safe execution block. This function may be redefined post-Merge. The current definition uses the most recent justified block as a safe block.

An alternative approach is in having generally defined `get_safe_block` function (as an extension outside of the main fork-choice spec) utilized by `get_safe_block_hash`. But it can be easily done later on once `get_safe_block` will represent more complicated logic than it currently is, or I can be convinced that it should be done within this PR.

TODO
* [x] consider `get_safe_block` implementation returning `store.justified_checkpoint.root` 
* [ ] write tests (left for subsequent PR, tracked in #2865)

### DIFF
--- a/fork_choice/safe-block.md
+++ b/fork_choice/safe-block.md
@@ -1,0 +1,31 @@
+# Fork Choice -- Safe Block
+
+## Table of contents
+<!-- TOC -->
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Introduction](#introduction)
+- [`get_safe_beacon_block`](#get_safe_beacon_block)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+## Introduction
+
+Under honest majority and certain network synchronicity assumptions
+there exist a block that is safe from re-orgs. Normally this block is
+pretty close to the head of canonical chain which makes it valuable
+to expose a safe block to users.
+
+This section describes an algorithm to find a safe block.
+
+## `get_safe_beacon_block`
+
+```python
+def get_safe_beacon_block(store: Store) -> Root:
+    # Use most recent justified block as a stopgap
+    return store.justified_checkpoint.root
+```
+*Note*: Currently safe block algorithm simply returns `store.justified_checkpoint.root`
+and is meant to be improved in the future.
--- a/fork_choice/safe-block.md
+++ b/fork_choice/safe-block.md
@@ -44,3 +44,5 @@ def get_safe_execution_payload_hash(store: Store) -> Hash32:
     else:
         return Hash32()
 ```
+
+*Note*: This helper uses beacon block container extended in [Bellatrix](../specs/bellatrix/beacon-chain.md).
--- a/fork_choice/safe-block.md
+++ b/fork_choice/safe-block.md
@@ -6,7 +6,8 @@
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
 - [Introduction](#introduction)
-- [`get_safe_beacon_block`](#get_safe_beacon_block)
+- [`get_safe_beacon_block_root`](#get_safe_beacon_block_root)
+- [`get_safe_execution_payload_hash`](#get_safe_execution_payload_hash)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 <!-- /TOC -->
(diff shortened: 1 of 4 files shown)


specs/bellatrix/fork-choice.md
```diff
@@ -72,7 +73,23 @@ As per EIP-3675, before a post-transition block is finalized, `notify_forkchoice
 
 *Note*: Client software MUST call this function to initiate the payload build process to produce the merge transition block; the `head_block_hash` parameter MUST be set to the hash of a terminal PoW block in this case.
 
-*Note*: Until safe head function is implemented, `safe_block_hash` parameter MUST be stubbed with the `head_block_hash` value.
+##### `safe_block_hash`
+
+The `safe_block_hash` parameter in a call to `notify_forkchoice_updated` function
```
djrtwo, line 78: I'd prefer to write a function wrt the safe beacon block -- `get_safe_beacon_block` -- and then input from the output of that function (`safe_block`) `safe_block.body.execution_payload.block_hash` into the `notify_fcu` function.  This leaves this beacon block based safe function as a first class citizen to be used for any purpose.

Then to encapsulate the additional `Hash32()` logic here we can have `get_safe_execution_payload_hash()` as it's own helper.

Mitumaru (approved):

mkalinin:
I have added a `/fork_choice/safe-block.md` document with the function definition. I am free to debate on the naming and place of this small piece of the spec. Once this is done it can be included in the spec build and testing

fork_choice/safe-block.md
```diff
@@ -0,0 +1,31 @@
...
+to expose a safe block to users.
+
+This section describes an algorithm to find a safe block.
+
+## `get_safe_beacon_block`
+
+```python
+def get_safe_beacon_block(store: Store) -> Root:
+    # Use most recent justified block as a stopgap
+    return store.justified_checkpoint.root
```
djrtwo, line 28: this should either be named `get_safe_beacon_block_root` and return a `Root` (as is)
OR
should be named `get_safe_beacon_block` and return the full `BeaconBlock`.

I'm okay either way

specs/bellatrix/fork-choice.md
```diff
@@ -72,7 +73,22 @@ As per EIP-3675, before a post-transition block is finalized, `notify_forkchoice
...
+MUST be set to the return value of the following function:
+
+```python
+def get_safe_block_hash(store: Store) -> Hash32:
```
djrtwo, line 82: I think this should likely got in the safe_block helper file

fork_choice/safe-block.md
mkalinin: The idea was to follow `get_head(store: Store) -> Root:` naming notation, i would expect `get_head` returning a beacon block, thought it could be argued.

mkalinin:
Applied the following review suggestions:
* `get_safe_beacon_block -> get_safe_beacon_block_root`
* `get_safe_block_hash  -> get_safe_execution_payload_hash`
* `get_safe_execution_payload_hash` moved to `safe-block.md`

fork_choice/safe-block.md
```diff
@@ -0,0 +1,46 @@
+# Fork Choice -- Safe Block
```
hwwhww, line 1: Note that this file is cross forks. Maybe note that it is only valid post-Bellatrix.

specs/bellatrix/fork-choice.md
```diff
@@ -72,7 +73,9 @@ As per EIP-3675, before a post-transition block is finalized, `notify_forkchoice
 
 *Note*: Client software MUST call this function to initiate the payload build process to produce the merge transition block; the `head_block_hash` parameter MUST be set to the hash of a terminal PoW block in this case.
 
-*Note*: Until safe head function is implemented, `safe_block_hash` parameter MUST be stubbed with the `head_block_hash` value.
+##### `safe_block_hash`
+
+The `safe_block_hash` parameter MUST be set to return value of `get_safe_execution_payload_hash(store: Store)` function.
```
hwwhww, line 78: Maybe add a link to `[safe-block.md]` for readability?

fork_choice/safe-block.md
mkalinin: In general, safe block algorithm can be implemented with `Phase0` structures. I would add a note that `get_safe_execution_payload_hash` helper is based on `Bellatrix`

specs/bellatrix/validator.md
```diff
@@ -110,16 +110,18 @@ All validator responsibilities remain unchanged other than those noted below. Na
 
 To obtain an execution payload, a block proposer building a block on top of a `state` must take the following actions:
 
-1. Set `payload_id = prepare_execution_payload(state, pow_chain, finalized_block_hash, suggested_fee_recipient, execution_engine)`, where:
+1. Set `payload_id = prepare_execution_payload(state, pow_chain, finalized_block_hash, safe_block_hash, suggested_fee_recipient, execution_engine)`, where:
     * `state` is the state object after applying `process_slots(state, slot)` transition to the resulting state of the parent block processing
     * `pow_chain` is a `Dict[Hash32, PowBlock]` dictionary that abstractly represents all blocks in the PoW chain with block hash as the dictionary key
+    * `safe_block_hash` is the return value of the `get_safe_block_hash(store: Store)` function call
```
djrtwo, line 116:
```suggestion
    * `safe_block_hash` is the return value of the `get_safe_execution_payload_hash(store: Store)` function call
```

djrtwo (approved):
looks good to me! opened #2865 to track needing some unit tests here

PR #2859, opened 2022-03-22
Bellatrix: send justified block to EL

mkalinin:
Adds `justified_block_hash` parameter to `notify_forkchoice_updated` function

Depends on https://github.com/ethereum/consensus-specs/pull/2858
Relates to https://github.com/ethereum/execution-apis/pull/203


djrtwo (commented):
As discussed, I don't think we should modify this API at this point. 

1. breaking changes should probably be avoided unless they are critical fixes
2. Having justified in addition to fialized and "safe" is of dubious value

mkalinin:
Closing this PR for the above reason

PR #2862, opened 2022-03-24
Partial withdrawals

djrtwo:
Partial withdrawals for 0x01 validators. Utilizes a method that just sweeps a maximum number of partially withdrawable validators on a per-epoch basis

* On each epoch transition -- cycle through the validator set (remembering where you last left off (via `next_partial_withdrawal_validator_index`)
    * if the validator is partially withdrawable (0x01 creds and balance in excess of 32 eth -- `is_partially_withdrawable_validator`), withdraw the excess balance
    * when cycle through whole list OR reach max withdrawals per epoch, abort process and mark where you left off for next time

Additionally, cleaned up some naming

See discussion below for rationale on config values. Take tuning these values to an issue if desired

discussion around if this should be *all* validators or just "active and not slashed" here -- #2905


specs/capella/beacon-chain.md
```diff
@@ -282,6 +289,19 @@ def is_fully_withdrawable_validator(validator: Validator, epoch: Epoch) -> bool:
...
+
+```python
+def is_partially_withdrawable_validator(validator: Validator, balance: Gwei) -> bool:
+    """
+    Check if ``validator`` is partially withdrawable.
+    """
+    is_eth1_withdrawal_prefix = validator.withdrawal_credentials[:1] == ETH1_ADDRESS_WITHDRAWAL_PREFIX
+    has_max_effective_balance = validator.effective_balance == MAX_EFFECTIVE_BALANCE
+    has_excess_balance = balance > MAX_EFFECTIVE_BALANCE
+    return is_eth1_withdrawal_prefix and has_max_effective_balance and has_excess_balance
```
mkalinin, line 302: Shouldn't it also check that the validator is in active state?

mcdee: > Shouldn't it also check that the validator is in active state?

I think it would be reasonable for an exiting validator, for example, to be able to skim.  May want to consider if slashed validators can do so, however.

mkalinin: My main concern is related to slashed validators. Also, i don't see value of skimming exited validators as their next step should be full withdrawal, it feels like unnecessary overhead in the withdrawal churn.

mcdee: Exited is different from exiting, however.  A validator could theoretically be stuck in exiting for quite a while, if a lot of validators head for the door at the same time so we have a large exit queue.

Although perhaps this is just a case of our definitions of "active" not being quite the same, and we're actually saying the same thing.  For me, I think that `validator.slashed == false && validator.exit_epoch > current_epoch" would be a sufficient condition (and yes this would theoretically catch validators before they activated, but I don't see that being a problem as they will only have excess balance if there was a double deposit and no harm in that coming back to the user as early as possible).

mcdee:
Thinking about `MAX_PARTIAL_WITHDRAWALS_PER_EPOCH`, and the related `MAX_WITHDRAWALS_PER_PAYLOAD`.

On the current value of `MAX_PARTIAL_WITHDRAWALS_PER_EPOCH` (256), with the current number of active validators it would work through the entire set every ~10 days, which seems fast.  Given that we're adding weight to the consensus and execution chains with this data I'd be inclined to bring it down to 128, which means that even if we add another 50% to the active validator set then validators can still skim off approximately once a month.  Frankly, I don't think that once a quarter is unreasonable but we don't know the size of the active validator set in future so this seems like a decent compromise.

If we did this, then we could reduce `MAX_WITHDRAWALS_PER_PAYLOAD` from 16 to 8 with no harm (with all Ether staked the churn limit would be ~50 so there would be space for all withdrawals and all partial withdrawals generated within the epoch even if we only proposed 70% of the blocks).  Although as the struct is a variable size list we could just leave this alone.

specs/capella/beacon-chain.md
```diff
@@ -301,9 +321,11 @@ def process_epoch(state: BeaconState) -> None:
     process_participation_flag_updates(state)
     process_sync_committee_updates(state)
     process_full_withdrawals(state)  # [New in Capella]
+    process_partial_withdrawals(state)  # [New in Capella]
```
mcdee, line 324: I get the consistency argument, but we're not processing partial withdrawals so much as generating them.  Not sure this matters in any real sense, and as I say having `process_*` keeps things consistent if not totally accurate.  So feel free to ignore this comment :)

dapplion:
So this is a "non-voluntary" partial withdrawal right? This can do a partial withdrawal for 1 Gwei if the validator balance is 32_000_000_001, should there be a minimum above `MAX_EFFECTIVE_BALANCE`? Looking at current mainnet balances this mechanism will forever create withdrawals maxing `MAX_PARTIAL_WITHDRAWALS_PER_EPOCH` since when it loops to index 0 all balances are already above `MAX_EFFECTIVE_BALANCE`

specs/capella/beacon-chain.md
djrtwo: I don't really see an issue with slashed or not-yet-activated validators getting a partial withdrawal about 32 ETH.

* not-yet-activated && balance about 32 -- will be partially withdrawn immediately upon activation
* slashed -- they cannot lose more than 32 eth (in most cases) so any excess will become withdrawn anyway
* exited-but-not-withdrawn -- restricting this just adds additional complexity and testing on this condition for near zero value. the amount of load induced by partial withdrawals is bound and will almost certainly always use the max-per-epoch (assuming a minimum number of 0x01 credentials). So the cost on the system of these types of withdrawals is 0 and the impact on limiting other validators access to partial withdrawals is low.

I'm not fundamentally opposed to the above conditions being added, but I think they provide little value.

*Note*: the way a validator can lose more than 32 ETH via slashing is if they have a highly correlated fault (so high base slashing, e.g. 32 ETH) AND they suffer a high inactivity leak prior to the 32 ETH slashing.

specs/capella/beacon-chain.md
```diff
@@ -245,7 +251,8 @@ class BeaconState(Container):
     # Execution
     latest_execution_payload_header: ExecutionPayloadHeader
     # Withdrawals
-    withdrawal_index: WithdrawalIndex
+    withdrawal_index: WithdrawalIndex  # [New in Capella]
+    next_partial_withdrawal_index: ValidatorIndex  # [New in Capella]
```
hwwhww, line 255: the variable names & types look confusing here 😅 but `next_partial_withdrawal_validator_index` may be too long...

What do you think about:
1. rename `withdrawal_index` to `withdrawals_queue_index`; `WithdrawalIndex` to `WithdrawalsQueueIndex`
2. move it next to `withdrawals_queue`

specs/capella/beacon-chain.md
```diff
@@ -262,7 +269,7 @@ def withdraw_balance(state: BeaconState, index: ValidatorIndex, amount: Gwei) ->
     # Create a corresponding withdrawal receipt
     withdrawal = Withdrawal(
         index=state.withdrawal_index,
-        address=state.validators[index].withdrawal_credentials[12:],
+        address=ExecutionAddress(state.validators[index].withdrawal_credentials[12:]),
```
hwwhww, line 272: nice catch

djrtwo:
Regarding the number per epoch and the number per slot to be dequeued, there's a few considerations I'm taking

* Picking a number for `MAX_PARTIAL_WITHDRAWALS_PER_EPOCH` that is larger than would be possible given a block-proposal-led method. e.g. 256 being 8x per epoch than if we just did one per proposal
* Then, picking a number of `MAX_WITHDRAWALS_PER_PAYLOAD` such that if `MAX_PARTIAL_WITHDRAWALS_PER_EPOCH` and the full `get_validator_churn_limit()` for each epoch is hit, that the queue does not back up in the long run if approximately 50% of slots are non-empty on average.
    * `get_validator_churn_limit()` can theoretically get up to ~50 in the event that all ETH in existence is staked (~100M ETH or about 3M validators).
    * In this extreme case (and assuming the max exits per epoch), you'd need about 19 non-empty slots (60%) per epoch on average to not have a backup in the queue with the current 16 withdrawals per payload. In more modest staked amounts (e.g. 30M ETH), you clear the queue at more like 17 non-empty slots per epoch (53%)
* Finally, we also must consider load on the execution layer (both in balance updates -- which induces hashing -- and in the payload data costs)
    * withdrawals are likely to be approximately 60 bytes `{amount: 32 bytes, address: 20 bytes, index: 8 bytes}`. so at max 16 per payload, that's ~1kb (`60 * 16`)  which is about +1% to +1.5% on current block data sizes (and way less in the event of 4844 or calldata repricing schemes
    * withdrawals also induce up to a maximum number of balance updates which induce some amount of hashing (this looks akin to a max 16 COINBASE updates in current config) ... I suspect on this order for this not to be a large consideration for fixed system overhead. But it's worth asking an expert.

I'm totally fine re-configuring  some of these values, but given the low cost of 16 as `MAX_WITHDRAWALS_PER_PAYLOAD` and with this 256 number cleaning clearing the queue with 50% to 60%  non-empty slots, I don't see the value in reducing the number yet. Especially considering that it is "free" for the validator -- they don't have to send an L1 transaction to claim this partial amount so that can accumulate happily over time.

Another thing to note is that less access to profits for solo/small stakers can be a centralizing pressure. So if we *can* provide a higher partial withdrawal rate at very low cost, it is by default a good thing unless shown unnecessarily costly on some other front.

specs/capella/beacon-chain.md
```diff
@@ -317,6 +339,31 @@ def process_full_withdrawals(state: BeaconState) -> None:
...
+#### Partial withdrawals
+
+*Note*: The function `process_partial_withdrawals` is new.
+
+```python
+def process_partial_withdrawals(state: BeaconState) -> None:
+    partial_withdrawals_count = 0
+    # Begin where we left off last time
+    validator_index = state.next_partial_withdrawal_index
+    for _ in range(len(state.validators)):
```
hwwhww, line 351: sanity check:

- `MAX_PARTIAL_WITHDRAWALS_PER_EPOCH` = 256
- 225 epochs per day
- To process the current `state.validators`, it *at most* takes ~5.6 days to run a cycle right, right?

specs/capella/beacon-chain.md
djrtwo: should the linter have caught this?

specs/capella/beacon-chain.md
djrtwo: yep, right about that

specs/capella/beacon-chain.md
djrtwo: yeah, I wasn't totally happy with the name too.

suggestion looks reasonable. I'll think about that and a few other naming issues

mcdee:
On the technical front, to grab some quick numbers there are around 210 balance changes per block (quick sample of the last 10,000 blocks at time of writing), so adding another 8 would be a ~4% increase.  Not sure if this is considered significant, but thought I'd put it out there.

One of my initial concerns (and apologies for not putting this in there to begin with, but I didn't run the numbers) would be the % of withdrawal that may be used as gas.  If we take 50Gewi as the effective gas price (48Gwei average bass fee over the last 10,000 blocks and 2Gwei tip) and pick a relatively normal transaction such as converting to USDC, then the cost of doing so is ~10% of the median withdrawal amount.  This feels high.

I'm fully aware that rather than halve the number of withdrawals per epoch stakers could just wait for two cycles to transfer their funds and there would be no difference in reality, but given the (even minor) additional cost to the network to calculate, store and process the partial withdrawals it feels like we could halve it without any major downsides, and in doing so save a few cycles and possibly avoid some complaints about the relative cost of moving withdrawals in the future.

But given the main point of my argument is around how this may feel to stakers, which isn't the hardest of positions, perhaps take this more as an observation for consideration than a strong reason to halve `MAX_PARTIAL_WITHDRAWALS_PER_EPOCH`

specs/capella/beacon-chain.md
mcdee: Thinking it through, a slashed validator will have lost `balance/MIN_SLASHING_PENALTY_QUOTIENT` when it enters the slashing state so although I think that philosophically speaking a slashed validator shouldn't be able to obtain any withdrawals from the chain whilst it is being slashed, in practical terms this will never happen anyway (outside of a couple of edge cases that I don't think are worthwhile worrying about) so you are correct no need to codify it here.

Regarding `validator.exit_epoch > current_epoch`: this came from the thinking that partial withdrawals were for validators that were validating, rather than those that have exited and will be fully withdrawn "soon" anyway.  I'd prefer to see it in there, but not so much that I'd fight for its inclusion, and if it creates additional complexity elsewhere then fair enough to not have it present.

djrtwo:
The balance update count is great! thank you @mcdee. I do wonder though -- do balance updates matter as the comparison here or more so all of the writes

>  If we take 50Gewi as the effective gas price (48Gwei average bass fee over the last 10,000 blocks and 2Gwei tip) and pick a relatively normal transaction such as converting to USDC, then the cost of doing so is ~10% of the median withdrawal amount. This feels high.

yeah, I see that. "is there value in having this granularity of amounts available"

specs/capella/beacon-chain.md
```diff
@@ -70,6 +73,12 @@ We define the following Python custom types for type hinting and readability:
```
ralexstokes, line 73: can't comment but need to update language on line 54 here

Ikaberry202 (approved):
Good to go

Mitumaru (approved):

Ikaberry202 (approved):
Good to go

Mitumaru (approved):
Thank you very much

Mitumaru (approved):

specs/capella/beacon-chain.md
djrtwo: I'm cleaning up this PR now and after coming back to it, I think I agree that this is actually simpler to reason about and test by doing just active (and not slashed) rather than all potential validators with excess balances.

I'm going to alter the PR and tests to that and we can touch on the discussion once more

djrtwo: I just went through the motions and it did add moderate complexity in testing to condition things upon epoch rather than purely on balance.

I want to get this merged and will open up an issue for discussion if anyone feels strongly about this

specs/capella/beacon-chain.md
djrtwo: i think `WithdrawalQueueIndex` is actually a bit misnamed because the queue length dynamically changes.

I'm going with `next_partial_withdrawal_validator_index` for now

specs/capella/beacon-chain.md
```diff
@@ -335,15 +386,15 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
 ```python
 def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
-    num_withdrawals = min(MAX_WITHDRAWALS_PER_PAYLOAD, len(state.withdrawals_queue))
-    dequeued_withdrawals = state.withdrawals_queue[:num_withdrawals]
+    num_withdrawals = min(MAX_WITHDRAWALS_PER_PAYLOAD, len(state.withdrawal_queue))
+    dequeued_withdrawals = state.withdrawal_queue[:num_withdrawals]
 
     assert len(dequeued_withdrawals) == len(payload.withdrawals)
     for dequeued_withdrawal, withdrawal in zip(dequeued_withdrawals, payload.withdrawals):
         assert dequeued_withdrawal == withdrawal
```
JustinDrake, line 394: Can these three lines be simplified to `assert dequeue_withdrawals == payload.withdrawals`?

ralexstokes (commented):
nice work! spec additions look good, I'll make another pass on the testing

tests/core/pyspec/eth2spec/test/capella/epoch_processing/test_process_partial_withdrawals.py
```diff
@@ -0,0 +1,221 @@
...
+    yield from run_process_partial_withdrawals(spec, state, spec.MAX_PARTIAL_WITHDRAWALS_PER_EPOCH)
+
+
+@with_capella_and_later
+@spec_state_test
+def test_success_max_plus_one_withdrawable(spec, state):
+    # Sanity check that this test works for this state
+    assert len(state.validators) >= spec.MAX_PARTIAL_WITHDRAWALS_PER_EPOCH + 1
+
+    # More than MAX_PARTIAL_WITHDRAWALS_PER_EPOCH paritally withdrawable
```
ralexstokes, line 161:
```suggestion
    # More than MAX_PARTIAL_WITHDRAWALS_PER_EPOCH partially withdrawable
```

tests/core/pyspec/eth2spec/test/capella/epoch_processing/test_process_partial_withdrawals.py
```diff
@@ -0,0 +1,221 @@
...
+
+    num_validators = len(state.validators)
+    state.next_partial_withdrawal_validator_index = rng.randint(0, num_validators - 1)
+
+    num_partially_withdrawable = rng.randint(0, num_validators - 1)
+    partially_withdrawable_indices = rng.sample(range(num_validators), num_partially_withdrawable)
+    for index in partially_withdrawable_indices:
+        set_validator_partially_withdrawable(spec, state, index)
+
+    # Note: due the randomness and other epoch processing, some of these set as "partially withdrawable"
```
ralexstokes, line 182:
```suggestion
    # Note: due to the randomness and other epoch processing, some of these set as "partially withdrawable"
```

PR #2877, opened 2022-04-25
Yield attestation in equivocating indices test

zilm13:
Current test processes attestation only in py test, so generated test fixtures doesn't force failure without equivocating indices feature enabled as it should. I've added attestation to the generated data


hwwhww (approved):
Nice catch! Thanks @zilm13!

PR #2881, opened 2022-04-28
Remove justified block from optimistic candidate conditions

mkalinin:
Simplifies the spec by removing "The justified checkpoint has execution enabled" condition from optimistic candidate evaluation.

Removal of this conditions implies a tiny change in client behaviour. Before the change, justification of a merge block immediately allowed to safely import blocks from any other chains. After the change, it becomes allowed only when `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` runs out. This negligible change doesn't affect security of merge transition process.

cc @paulhauner @ajsutton @potuz

### DIFF
--- a/sync/optimistic.md
+++ b/sync/optimistic.md
@@ -84,10 +84,6 @@ def is_optimistic_candidate_block(opt_store: OptimisticStore, current_slot: Slot
     if is_execution_block(opt_store.blocks[block.parent_root]):
         return True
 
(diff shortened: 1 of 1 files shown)


ajsutton (approved):
LGTM. Teku has already implemented this change.

potuz:
Same with prysm

djrtwo (approved):

PR #2882, opened 2022-04-28
Add new `DomainType` for application usage.

ralexstokes:
This PR adds a new `DomainType` for use by applications adjacent to the core protocol so they can leverage the signing machinery developed here when working with consensus types.

This PR only adds a single domain and expects applications using the domain type to mix in further unique tags for domain separation; otherwise they risk signature collisions.

```python
def compute_domain_for_application(application_domain: SomeBytesType):
    protocol_domain = compute_domain(DOMAIN_APPLICATION) # adjust other args as required...
    return Domain(hash(application_domain + protocol_domain))
```

An example use case is the work-in-progress Builder API where a validator may want to outsource block construction to a network of builders external to their local execution client. See https://github.com/ethereum/execution-apis/pull/209 for details.

### DIFF
--- a/specs/bellatrix/beacon-chain.md
+++ b/specs/bellatrix/beacon-chain.md
@@ -10,6 +10,8 @@
 
 - [Introduction](#introduction)
 - [Custom types](#custom-types)
+- [Constants](#constants)
+  - [Domain types](#domain-types)
 - [Preset](#preset)
   - [Execution](#execution)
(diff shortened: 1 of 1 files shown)


ralexstokes:
@djrtwo, @lightclient, @metachris

ralexstokes:
closing in lieu of 2884.

PR #2884, opened 2022-05-03
Add new `DomainType` for application usage with "internal" namespace

ralexstokes:
Opening in lieu of #2882.

# Context

This PR adds a new DomainType for use by applications adjacent to the core protocol so they can leverage the signing machinery developed here when working with consensus types.

An example use case is the work-in-progress Builder API where a validator may want to outsource block construction to a network of builders external to their local execution client. See https://github.com/ethereum/execution-apis/pull/209 for details.

# Rationale

This PR creates a "reserved" signing namespace for applications by setting the top bit in the `DomainType`. Applications are still expected to set other bits in the remainder of the `DomainType` data to get domain separation from other applications using this scheme.

This responsibility is on the applications to self-organize who uses which concrete `DomainType`s.

Examples:

```
DOMAIN_APPLICATION_MEV_BOOST = `0x10000001`
DOMAIN_APPLICATION_ANOTHER_APP = `0x2000001`
```

This approach is chosen over the approach in #2882 as there is less machinery to get right for users of the consensus spec.

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -203,6 +203,9 @@ The following values are (non-configurable) constants used throughout the specif
 | `DOMAIN_VOLUNTARY_EXIT`      | `DomainType('0x04000000')` |
 | `DOMAIN_SELECTION_PROOF`     | `DomainType('0x05000000')` |
 | `DOMAIN_AGGREGATE_AND_PROOF` | `DomainType('0x06000000')` |
+| `DOMAIN_APPLICATION_MASK`    | `DomainType('0x00000001')` |
+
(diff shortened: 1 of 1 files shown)


djrtwo (commented):
Looks good in general.

I might call it the `DOMAIN_APPLICATION_MASK` and say that `DOMAIN_RANDOM_APPLICATION && DOMAIN_APPLICATION_MASK` must be non-zero

djrtwo (commented):
Ah, just realized this was added to Bellatrix. Do you think it makes more sense in Phase 0? As it applies to all phases in gneral?

ralexstokes:
yeah makes sense, I'll update

rsalvo77 (commented):
[specs/phase0/beacon-chain.md]()

Issue #2886, opened 2022-05-06
`milagro_bls_binding` fails to install on arm64


jtraglia:
The `milagro_bls_binding==1.6.3` python module does not seem to exist for arm64 systems.

```
ERROR: Could not find a version that satisfies the requirement milagro_bls_binding==1.6.3 (from eth2spec[test])
ERROR: No matching distribution found for milagro_bls_binding==1.6.3
```

If you attempt to install `milagro_bls_binding` outside of `make install_test`, it will provide a list of available versions:

```
(from versions: 0.1.4, 0.4.0, 1.0.0, 1.0.1, 1.8.0)
```

And `1.6.3` is missing from this list.

## Full Output
```
python3 -m venv venv; . venv/bin/activate; python3 -m pip install -e .[lint]; python3 -m pip install -e .[test]
Obtaining file:///consensus-specs
Collecting dataclasses==0.6
  Downloading dataclasses-0.6-py3-none-any.whl (14 kB)
Collecting eth-typing<3.0.0,>=2.1.0
  Downloading eth_typing-2.3.0-py3-none-any.whl (6.2 kB)
Collecting eth-utils<2,>=1.3.0
  Downloading eth_utils-1.10.0-py3-none-any.whl (24 kB)
Collecting lru-dict==1.1.6
  Downloading lru-dict-1.1.6.tar.gz (9.4 kB)
Collecting marko==1.0.2
  Using cached marko-1.0.2-py3-none-any.whl (34 kB)
ERROR: Could not find a version that satisfies the requirement milagro_bls_binding==1.6.3 (from eth2spec[lint])
ERROR: No matching distribution found for milagro_bls_binding==1.6.3
Obtaining file:///consensus-specs
Collecting dataclasses==0.6
  Using cached dataclasses-0.6-py3-none-any.whl (14 kB)
Collecting eth-typing<3.0.0,>=2.1.0
  Using cached eth_typing-2.3.0-py3-none-any.whl (6.2 kB)
Collecting eth-utils<2,>=1.3.0
  Using cached eth_utils-1.10.0-py3-none-any.whl (24 kB)
Collecting lru-dict==1.1.6
  Using cached lru-dict-1.1.6.tar.gz (9.4 kB)
Requirement already satisfied: marko==1.0.2 in ./venv/lib/python3.9/site-packages (from eth2spec==1.1.10) (1.0.2)
[...]

Issue #2780, opened 2021-12-27
CI updates: add Python 3.10 and drop Python 3.8 support


hwwhww:
### Issue

For efficiency and using more new built-in libraries, we can consider dropping Python 3.8 support now.

hwwhww:
Note that it may affect the `eth2spec` Python package users if we drop Python 3.8 support in the package setting and start to write Python 3.9-new code.

ChihChengLiang:
I'm also preparing an update on future python versions https://github.com/ChihChengLiang/milagro_bls_binding/issues/34

djrtwo:
When we move from 3.8 to 3.10, we should build test vectors with both and ensure they have the same output

ChihChengLiang:
@hwwhww @protolambda I have the milagro_bls_binding [1.8.0](https://pypi.org/project/milagro-bls-binding/1.8.0/) ready. It works with my macos/M1/python3.10 setting.

PR #2893, opened 2022-05-14
bump milagro bls binding version

ChihChengLiang:
The new version of milagro bls binding supports py3.10(#2780) and arm (#2886).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1048,7 +1048,7 @@ setup(
         "eth-typing>=2.1.0,<3.0.0",
         "pycryptodome==3.9.4",
         "py_ecc==5.2.0",
-        "milagro_bls_binding==1.6.3",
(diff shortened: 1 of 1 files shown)


ralexstokes (approved):
lgtm

ralexstokes:
@ChihChengLiang do you have any thoughts on ensuring there aren't any regressions across different versions of this software?

ChihChengLiang:
@ralexstokes I think the regression is unlikely due to these reasons:
1. The binding itself is covered in the same set of tests, which haven't changed during this upgrade.
2. No significant updates in the upstream of the binding affect the codebase of binding and the spec.
3. All spec tests that consume the binding passed.

ralexstokes:
the concern would be consistency across versions of the software but I agree it is unlikely

PR #2901, opened 2022-05-24
Executable EIP-4844 specs

hwwhww:
In this PR:
- Added `kzg.py` for generating testing trusted setup.
    - Ref: @vbuterin's https://github.com/ethereum/research/blob/8f084630528ba33d92b2bc05edf5338dd193c6f1/trusted_setup/trusted_setup.py
- Renamed/reused config
    - `SHARDING_FORK_VERSION` -> `EIP4844_FORK_VERSION`
    - `SHARDING_FORK_EPOCH` -> `EIP4844_FORK_EPOCH`
- Added `test-eip4844` CI job
- Fixed tests
    - Right now, both EIP4844 and Capella's pre-fork is Bellatrix. This PR updated the testing tool to make it compatible.

TODOs:
- (Not in this PR) Handle mainnet config `FIELD_ELEMENTS_PER_BLOB: 4096`
    - Since (i) mainnet trusted setup is not complete (ii), it would really slow to generate trusted setup with large length with py-ecc and (iii) spec builder dependency, right now mainnet `KZG_SETUP_G2` and `KZG_SETUP_LAGRANGE` are actually minimal config values.


hwwhww:
@asn-d6 @protolambda 
Do you mind reviewing the specs and `kzg.py` parts?

tests/core/pyspec/eth2spec/test/eip4844/unittests/test_kzg.py
```diff
@@ -0,0 +1,27 @@
...
+    with_phases,
+    spec_state_test,
+    with_presets,
+)
+
+
+def _create_blob(spec):
+    rng = random.Random(5566)
+    return spec.Blob([
+        rng.randint(0, spec.BLS_MODULUS)
```
asn-d6, line 17: The test looks good, but counter-intuitively this should be `rng.randint(0, spec.BLS_MODULUS-1)` since `randint` [works on an inclusive range on both sides](https://stackoverflow.com/questions/2568783/python-why-does-random-randinta-b-return-a-range-inclusive-of-b)...

There is no way this could ever be a problem, but just mentioning it for correctness.

tests/core/pyspec/eth2spec/utils/kzg.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def get_lagrange(setup):
+    """
+    Convert a G1 or G2 portion of a setup into the Lagrange basis.
+    """
+    root_of_unity = generate_root_of_unity(len(setup))
+    assert pow(root_of_unity, len(setup), curve_order) == 1
+    domain = [pow(root_of_unity, i, curve_order) for i in range(len(setup))]
+    fft_output = fft(setup, curve_order, domain)
```
asn-d6, line 57: These last three lines manually turn an FFT into an IFFT. This is correct, although in the future we might want to introduce an IFFT function for simplicity, since it's likely we are only gonna be using IFFT anyway.

asn-d6 (commented):
This looks great, after an initial review!

I'm gonna be working on integrating https://github.com/ethereum/EIPs/pull/5088 into the consensus-specs this week, so there will be a bunch of new things after that's done. When it's time, I will leave a new comment here with some more information.

Cheers!

asn-d6:
Hello @hwwhww! I have created the PR for the consensus-side verification logic that uses KZG proofs. Please see https://github.com/ethereum/consensus-specs/pull/2915

I imagine that this would warrant a bunch of changes to the code of this PR. In particular:
- Implement the math/polynomial functions introduced by the PR.
- Introduce the `ROOTS_OF_UNITY` constant list of BLS field elements. You can generate them with code that looks like this https://github.com/ethereum/consensus-specs/blob/danksharding/specs/sharding/polynomial-commitments.md#roots_of_unity
- Finally, to make an end-to-end test that goes from block creation all the way to verification, we would need to create a valid `kzg_aggregated_proof` that can be verified. To do that, we need to create a bunch of blobs and commitments and then basically faithfully reproduce the steps performed in `verify_blobs_sidecar()` to compute:
  - `r_powers` from the various blobs and commitments
  -  `aggregated_poly` from the individual blobs
  -  `aggregated_poly_commitment` from the corresponding commitments
  -  `x` from the `aggregated_poly` and `aggregated_poly_commitment`

  finally, we use those objects to create a valid KZG proof for that aggregated poly at `x`  That proof is the one that the tests will verify.

---

Please let me know if the above did not make any sense!

hwwhww:
Hey @asn-d6  

Updated some quick fixes after the #2915 patch in https://github.com/ethereum/consensus-specs/pull/2901/commits/60fdcacfd7a22ce1641f018bb5e871b15eb78407.

1. Moved `vector_lincomb` to `polynomial-commitments.md`
2. Fixed function signatures: replace `List` with `Sequence`
3. To use `hash_to_bls_field` properly, created `BlobsAndCommmitments` and `PolynomialAndCommitment` SSZ containers.
4. Add custom type `Polynomial`
5. To exclude the snippet sample code in p2p spec & v-guide, remove “python” from the code block for now.

hwwhww:
Found more bugs: https://github.com/ethereum/consensus-specs/pull/2901/commits/33829dd4563ea1eb7123cdc8edc956e898b3168c

1. Fixed offset calculation in `tx_peek_blob_versioned_hashes` 
2. Fixed `lincomb` return value type: in Pyspec, `r` is an `optimized_bls12_381_FQ` object so we need serialization. `G1_to_bytes48` is the function that we used in BLS pubkey serialization.
3. Fixed `BlobsSidecar` sample code: `/` -> Python `//`

hwwhww:
@asn-d6 

In [validator guide](https://github.com/ethereum/consensus-specs/blob/eip4844-exe/specs/eip4844/validator.md#beacon-block-publishing-time), it described how to pack the blobs into a `BlobsSidecar` object. However, `BlobsSidecar.kzg_aggregated_proof` value is missing here.

Should we add notes on how to generate proof to the validator guide?

asn-d6:
> @asn-d6
> 
> In [validator guide](https://github.com/ethereum/consensus-specs/blob/eip4844-exe/specs/eip4844/validator.md#beacon-block-publishing-time), it described how to pack the blobs into a `BlobsSidecar` object. However, `BlobsSidecar.kzg_aggregated_proof` value is missing here.
> 
> Should we add notes on how to generate proof to the validator guide?

Yep! I think that's needed!

The notes in the last point of https://github.com/ethereum/consensus-specs/pull/2901#issuecomment-1155374487 could serve as a decent guide here I think.

asn-d6:
> Found more bugs: [33829dd](https://github.com/ethereum/consensus-specs/commit/33829dd4563ea1eb7123cdc8edc956e898b3168c)

Good catches!

>     2. Fixed `lincomb` return value type: in Pyspec, `r` is an `optimized_bls12_381_FQ` object so we need serialization. `G1_to_bytes48` is the function that we used in BLS pubkey serialization.

I'm a bit sad that we have to do this ugly cast `KZGCommitment(kzg.G1_to_bytes48(r))` in the spec. I understand why it's there and I don't have a solution atm, but I'm just mentioning this because I don't remember similar casts in the BLS parts of the spec.

hwwhww:
Updates:
1. Add `compute_proof_from_blobs` and its sub helpers. (Thank @asn-d6 for helping debug!)
2. Add tests to verify the validator guide.
3. Add @protolambda explainer on `blob_versioned_hashes` offset calculation
4. Renames 
	1. `quotient_kzg` -> `kzg_proof`
	2. Parameters. Use fewer abbreviations.
5. Fix `is_data_available`
6. Add `int()`  casting during polynomial computation to avoid SSZ object overflow/underflow.
7. Add `G1_to_bytes48`/`bytes48_to_G1`/`G2_to_bytes96`/`bytes96_to_G2` in KZG functions.
	* Re: https://github.com/ethereum/consensus-specs/pull/2901#issuecomment-1172868540
		* The BLS API wrappers (IETF BLS signature draft standard v4 implementation) handled it internally so I didn't have to do the conversion in specs. But now we are dealing with low-level computation in specs so things are exposed. These conversions only happen in `polynomial-commitments.md` currently. Hope that it will be standardized soon and then CL implementors don't have to implement the conversions.

hwwhww:
Update: https://github.com/ethereum/consensus-specs/pull/2901/commits/ec7f6071a48af267dc70c96cb42f4168eb52b526

As discussed, [rename most `KZGCommitment` object `kzg` -> `commitment`](https://hackmd.io/UPfit4yWR5WRBEz7N8rSdQ?view)


1. `blob_kzgs` -> `blob_commitments`
    - note: there is also "`KZGProof`"
2. `blob_to_kzg` ->  `blob_to_commitment`
3. `BLOB_COMMITMENT_VERSION_KZG` -> `BLOB_COMMITMENT_VERSION`
4. `kzg_to_versioned_hash` -> `commitment_to_versioned_hash`
5. [New] `verify_kzgs_against_transactions` -> `verify_commitments_against_transactions`

Notes:
- I use "`commitment`" in most function signatures where it is obvious a commitment of blob.
- Use full name `blob_commitments` in `BeaconBlockBody` field.

specs/eip4844/polynomial-commitments.md
```diff
@@ -110,37 +134,96 @@ def blob_to_kzg(blob: Blob) -> KZGCommitment:
...
         [P_minus_y, bls.neg(bls.G2)],
-        [quotient_kzg, X_minus_x]
+        [bls.bytes48_to_G1(kzg_proof), X_minus_x]
     ])
 ```
 
+#### `compute_kzg_proof`
+
+```python
+def compute_kzg_proof(polynomial: Sequence[BLSFieldElement], x: BLSFieldElement) -> KZGProof:
```
asn-d6, line 153: I think it might be a good idea to document here that the polynomial is in evaluation form

specs/eip4844/polynomial-commitments.md
```diff
@@ -110,37 +134,96 @@ def blob_to_kzg(blob: Blob) -> KZGCommitment:
...
+    # To avoid SSZ overflow/underflow, convert element into int
+    polynomial = [int(i) for i in polynomial]
+
+    # Convert `polynomial` to coefficient form
+    assert pow(ROOTS_OF_UNITY[1], len(polynomial), BLS_MODULUS) == 1
```
asn-d6, line 158: Now that FFT has penetrated into the core of the spec, it might be worth modding our FFT function to do the inverse FFT inside of it, instead of doing it at the caller.

For example see the wrapper `fft()` function here https://github.com/ethereum/research/blob/master/mimc_stark/fft.py#L31 that internally calls `_fft()` which is what we have here (we don't need the `simple_ft()` thing).

tests/core/pyspec/eth2spec/utils/kzg.py
```diff
@@ -0,0 +1,80 @@
...
+
+def compute_root_of_unity(length) -> int:
+    """
+    Generate a w such that ``w**length = 1``.
+    """
+    assert (BLS_MODULUS - 1) % length == 0
+    return pow(PRIMITIVE_ROOT_OF_UNITY, (BLS_MODULUS - 1) // length, BLS_MODULUS)
+
+
+def compute_roots_of_unity(field_elements_per_blob):
```
asn-d6, line 54: Might be better to replace `field_elements_per_blob` with `order` here and make this function less-application-specific.

asn-d6:
Made an attempt at squashing and rebasing this PR: https://github.com/ethereum/consensus-specs/pull/2937
Pending review.

Issue #2902, opened 2022-05-24
Multiple fork transitions on same slot/epoch discard old `fork_version`s


tersec:
The result is that certain messages across multiple-fork-transition-per-epoch boundaries might not be verifiable when received.

https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/fork.md#upgrading-the-state states that the fork of the upgraded state is
```python
        fork=Fork(
            previous_version=pre.fork.current_version,
            current_version=ALTAIR_FORK_VERSION,
            epoch=epoch,
        ),
```
https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/fork.md#upgrading-the-state states that for Bellatrix, it's:
```python
        fork=Fork(
            previous_version=pre.fork.current_version,
            current_version=BELLATRIX_FORK_VERSION,
            epoch=epoch,
        ),
```
And for https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/fork.md#upgrading-the-state, it's
```python
        fork=Fork(
            previous_version=pre.fork.current_version,
            current_version=CAPELLA_FORK_VERSION,
            epoch=epoch,
        ),
```

That is, even if there was no observable beacon chain-time during which an `ALTAIR_FORK_EPOCH == BELLATRIX_FORK_EPOCH` or `BELLATRIX_FORK_EPOCH == CAPELLA_FORK_EPOCH` network existed in that intermediate fork, it will still show up as `state.fork.previous_version` of the fork to which it was upgraded, not necessarily the chronologically previous fork in question (e.g., the fork which might appear in the beacon API fork schedule).

This means that signatures of attestations, for example, from a slot or two before an `ALTAIR_FORK_EPOCH == BELLATRIX_FORK_EPOCH` or `BELLATRIX_FORK_EPOCH == CAPELLA_FORK_EPOCH` transition and then included in slots (which is supposed to be valid for up to 32 slots) cannot be verified afterward by a conforming client using https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#get_domain as written:
```python
def get_domain(state: BeaconState, domain_type: DomainType, epoch: Epoch=None) -> Domain:
    ...
    fork_version = state.fork.previous_version if epoch < state.fork.epoch else state.fork.current_version
    ...
```
because `fork_version` cannot access an old enough version in `state` by then.

[...]

PR #2903, opened 2022-05-31
Tidy up README.md specs list

hwwhww:
- To organize the forks and patches better, I tried to use tables to list the specs in README.
    - Stable Specifications: the stable/ready Ethereum Mainnet protocol upgrades
        - In an attempt to mitigate #2902, I noted the "Previous Fork" in the table.
    - In-development Specifications: the patches in development that don't have scheduled orders yet.
- Add `Capella` and `EIP4844` to README.md
- Preview:
image

Personally, I think it looks better than before. Although HTML tags are annoying, we rarely need to update this file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,6 @@ To learn more about proof-of-stake and sharding, see the [PoS FAQ](https://eth.w
 
 This repository hosts the current Ethereum proof-of-stake specifications. Discussions about design rationale and proposed changes can be brought up and discussed as issues. Solidified, agreed-upon changes to the spec can be made through pull requests.
 
-
 ## Specs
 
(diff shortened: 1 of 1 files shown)


ralexstokes (approved):
not sure we need the level of detail for fork version and previous fork but generally like the updated presentation!

dapplion:
> not sure we need the level of detail for fork version and previous fork but generally like the updated presentation!

I would drop the fork version to give more width to the other columns

hwwhww:
@ralexstokes @dapplion thanks for your feedbackI pruned the columns:
image

hwwhww:
@djrtwo  👍
Updated:
image

PR #2914, opened 2022-06-13
EIP-4844: Remove shard field from BlobsSidecar

Inphi:
We no longer need the placeholder `shard` field in the `BlobsSidecar` because with danksharding we'll most likely use a transaction-index, rather than index blobs by a shard.


protolambda:
In https://github.com/ethereum/consensus-specs/commit/45e207be4d99247c9b33fb5a44a0bdc038b2d945 the `shard` field was already removed, and this occurrence in the validator doc was forgotten. Nice catch

protolambda (approved):
LGTM

asn-d6:
LGTM

asn-d6 (approved):

PR #2915, opened 2022-06-14
Optimizing EIP-4844 block validation (using KZG proofs)

asn-d6:
Hey!

This PR introduces KZG proofs for block validation on the consensus-side, the same way that https://github.com/ethereum/EIPs/pull/5088 introduces KZG proofs for mempool transaction validation on the execution-side.

The main protocol change in terms of communication is that we are now adding a KZG proof in`BlobsSidecar`. The KZG proof is just there to speed up the validation of the commitments in the `BeaconBlockBody` and has no other value. Hence it doesn't actually need to go on-chain and can remain in the sidecar.

Please read https://github.com/ethereum/EIPs/pull/5088 for the details on how the crypto/validation logic works.

Let me know if you have any questions :)

### DIFF
--- a/specs/eip4844/polynomial-commitments.md
+++ b/specs/eip4844/polynomial-commitments.md
@@ -66,19 +66,10 @@ but reusing the `mainnet` settings in public networks is a critical security req
 ```python
 def bls_modular_inverse(x: BLSFieldElement) -> BLSFieldElement:
     """
-    Compute the modular inverse of x using the eGCD algorithm
+    Compute the modular inverse of x
     i.e. return y such that x * y % BLS_MODULUS == 1 and return 0 for x == 0
     """
-    if x == 0:
-        return 0
-
-    lm, hm = 1, 0
-    low, high = x % BLS_MODULUS, BLS_MODULUS
-    while low > 1:
-        r = high // low
-        nm, new = hm - lm * r, high - low * r
-        lm, low, hm, high = nm, new, lm, low
-    return lm % BLS_MODULUS
+    return pow(x, -1, BLS_MODULUS) if x != 0 else 0
 ```
 
 #### `div`
--- a/specs/eip4844/polynomial-commitments.md
+++ b/specs/eip4844/polynomial-commitments.md
@@ -75,7 +75,7 @@ def bls_modular_inverse(x: BLSFieldElement) -> BLSFieldElement:
 #### `div`
 
 ```python
-def div(x, y):
+def div(x: BLSFieldElement, y: BLSFieldElement) -> BLSFieldElement:
     """Divide two field elements: `x` by `y`"""
     return x * inv(y) % BLS_MODULUS
 ```
@@ -111,7 +111,9 @@ def verify_kzg_proof(polynomial_kzg: KZGCommitment,
                      x: BLSFieldElement,
                      y: BLSFieldElement,
                      quotient_kzg: KZGProof) -> bool:
-    """Verify KZG proof that `p(x) == y` where `p(x)` is the polynomial represented by `polynomial_kzg`"""
+    """
+    Verify KZG proof that ``p(x) == y`` where ``p(x)`` is the polynomial represented by ``polynomial_kzg``.
+    """
     # Verify: P - y = Q * (X - x)
     X_minus_x = bls.add(KZG_SETUP_G2[1], bls.multiply(bls.G2, BLS_MODULUS - x))
     P_minus_y = bls.add(polynomial_kzg, bls.multiply(bls.G1, BLS_MODULUS - y))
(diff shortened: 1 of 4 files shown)


dankrad:
Looking goods. Would it make sense to put the polynomial and KZG functions into a separate file?

specs/eip4844/polynomial-commitments.md
```diff
@@ -0,0 +1,144 @@
...
+    Compute the modular inverse of x
+    i.e. return y such that x * y % BLS_MODULUS == 1 and return 0 for x == 0
+    """
+    return pow(x, -1, BLS_MODULUS) if x != 0 else 0
+```
+
+#### `div`
+
+```python
+def div(x, y):
```
hwwhww, line 78:
```suggestion
def div(x: BLSFieldElement, y: BLSFieldElement) -> BLSFieldElement:
```

specs/eip4844/polynomial-commitments.md
```diff
@@ -0,0 +1,144 @@
...
+```
+
+#### `verify_kzg_proof`
+
+```python
+def verify_kzg_proof(polynomial_kzg: KZGCommitment,
+                     x: BLSFieldElement,
+                     y: BLSFieldElement,
+                     quotient_kzg: KZGProof) -> bool:
+    """Verify KZG proof that `p(x) == y` where `p(x)` is the polynomial represented by `polynomial_kzg`"""
```
hwwhww, line 114:
```suggestion
    """
    Verify KZG proof that ``p(x) == y`` where ``p(x)`` is the polynomial represented by ``polynomial_kzg``.
    """
```

specs/eip4844/polynomial-commitments.md
```diff
@@ -0,0 +1,144 @@
...
+    Evaluate a polynomial (in evaluation form) at an arbitrary point `x`
+    Uses the barycentric formula:
+       f(x) = (1 - x**WIDTH) / WIDTH  *  sum_(i=0)^WIDTH  (f(DOMAIN[i]) * DOMAIN[i]) / (x - DOMAIN[i])
+    """
+    width = len(poly)
+    assert width == FIELD_ELEMENTS_PER_BLOB
+    inverse_width = bls_modular_inverse(width)
+
+    for i in range(width):
+        r += div(poly[i] * ROOTS_OF_UNITY[i], (x - ROOTS_OF_UNITY[i]) )
```
hwwhww, line 140:
```suggestion
        r += div(poly[i] * ROOTS_OF_UNITY[i], (x - ROOTS_OF_UNITY[i]))
```

specs/eip4844/polynomial-commitments.md
```diff
@@ -0,0 +1,144 @@
...
+    """
+    Evaluate a polynomial (in evaluation form) at an arbitrary point `x`
+    Uses the barycentric formula:
+       f(x) = (1 - x**WIDTH) / WIDTH  *  sum_(i=0)^WIDTH  (f(DOMAIN[i]) * DOMAIN[i]) / (x - DOMAIN[i])
+    """
+    width = len(poly)
+    assert width == FIELD_ELEMENTS_PER_BLOB
+    inverse_width = bls_modular_inverse(width)
+
+    for i in range(width):
```
hwwhww, line 139: It seems `inverse_width` is a constant? How about:

```suggestion
    assert len(poly) == FIELD_ELEMENTS_PER_BLOB
    inverse_width = bls_modular_inverse(FIELD_ELEMENTS_PER_BLOB)

    for i in range(FIELD_ELEMENTS_PER_BLOB):
```

specs/eip4844/validator.md
```diff
@@ -54,14 +54,58 @@ def is_data_available(slot: Slot, beacon_block_root: Root, kzgs: Sequence[KZGCom
...
+    """
+    r = [0]*len(vectors[0])
+    for v, a in zip(vectors, scalars):
+        for i, x in enumerate(v):
+            r[i] = (r[i] + a * x) % BLS_MODULUS
+    return [BLSFieldElement(x) for x in r]
+
+
 def verify_blobs_sidecar(slot: Slot, beacon_block_root: Root,
                          expected_kzgs: Sequence[KZGCommitment], blobs_sidecar: BlobsSidecar):
```
hwwhww, line 86:
```suggestion
def verify_blobs_sidecar(slot: Slot, beacon_block_root: Root,
                         expected_kzgs: Sequence[KZGCommitment], blobs_sidecar: BlobsSidecar) -> None:
```

hwwhww (approved):
Great work @asn-d6 @adietrichs @dankrad 
lgtm with my naked eyes.

if we merge it, I can test it with #2901 better.

specs/eip4844/polynomial-commitments.md
asn-d6: Yep `inverse_width` is a constant. However, we do use `width` in the formula below so I think the above wouldn't work.

Another approach is to just compute `inverse_width` directly into the formula, instead of using an entire variable.

hwwhww: Ah I see, so if we want to remove `width`, we have to replace `width` with the `FIELD_ELEMENTS_PER_BLOB` below too. But now I see that if we don't have "`width`", then it might be confusing when we read `inverse_width`. 😅 

Your call!

asn-d6: I feel like all variants have their tradeoffs in this case.

Perhaps we can leave it as-is for now, and I imagine this piece of code will be revised as more eyes see it.

asn-d6:
> Great work @asn-d6 @adietrichs @dankrad lgtm with my naked eyes.
> 
> if we merge it, I can test it with #2901 better.

Thanks for the review! Feel free to merge whenever!

Inphi (commented):
Good stuff! Looks good to me from an implementer's perspective.

PR #2767, opened 2021-12-09
Add test cases with configured `PROPOSER_SCORE_BOOST:= 0`

hwwhww:
For debugging, this branch is used for generating test vectors with special configured `PROPOSER_SCORE_BOOST:= 0`.

Will change the base branch to `dev` once #2752 is merged.

PR #2916, opened 2022-06-16
Remove `fork_version` from `LightClientUpdate`

etan-status:
The `fork_version` field in `LightClientUpdate` can be derived from the
`update.signature_slot` value by consulting the locally configured fork
schedule. The light client already needs access to the fork schedule to
determine the `GeneralizedIndex` values used for merkle proofs, and the
memory layouts of the structures (including `LightClientUpdate`). The
`fork_version` itself is network dependent and doesn't reveal that info.

### DIFF
--- a/specs/capella/beacon-chain.md
+++ b/specs/capella/beacon-chain.md
@@ -34,8 +34,6 @@
     - [`has_eth1_withdrawal_credential`](#has_eth1_withdrawal_credential)
     - [`is_fully_withdrawable_validator`](#is_fully_withdrawable_validator)
     - [`is_partially_withdrawable_validator`](#is_partially_withdrawable_validator)
-  - [Misc](#misc-1)
-    - [Modified `compute_fork_version`](#modified-compute_fork_version)
 - [Beacon chain state transition function](#beacon-chain-state-transition-function)
   - [Epoch processing](#epoch-processing)
     - [Full withdrawals](#full-withdrawals)
@@ -318,24 +316,6 @@ def is_partially_withdrawable_validator(validator: Validator, balance: Gwei) -> 
     has_max_effective_balance = validator.effective_balance == MAX_EFFECTIVE_BALANCE
     has_excess_balance = balance > MAX_EFFECTIVE_BALANCE
     return has_eth1_withdrawal_credential(validator) and has_max_effective_balance and has_excess_balance
-```
-
-### Misc
-
-#### Modified `compute_fork_version`
-
-```python
-def compute_fork_version(epoch: Epoch) -> Version:
-    """
-    Return the fork version at the given ``epoch``.
-    """
-    if epoch >= CAPELLA_FORK_EPOCH:
-        return CAPELLA_FORK_VERSION
-    if epoch >= BELLATRIX_FORK_EPOCH:
-        return BELLATRIX_FORK_VERSION
-    if epoch >= ALTAIR_FORK_EPOCH:
-        return ALTAIR_FORK_VERSION
-    return GENESIS_FORK_VERSION
 ```
 
 ## Beacon chain state transition function
--- a/specs/capella/beacon-chain.md
+++ b/specs/capella/beacon-chain.md
@@ -34,6 +34,8 @@
     - [`has_eth1_withdrawal_credential`](#has_eth1_withdrawal_credential)
     - [`is_fully_withdrawable_validator`](#is_fully_withdrawable_validator)
     - [`is_partially_withdrawable_validator`](#is_partially_withdrawable_validator)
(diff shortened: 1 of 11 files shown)


tests/core/pyspec/eth2spec/test/altair/unittests/test_sync_protocol.py
```diff
@@ -70,11 +73,16 @@ def test_process_light_client_update_not_timeout(spec, state):
...
 
+    spec.config = old_config
+
 
 @with_altair_and_later
 @spec_state_test
 @with_presets([MINIMAL], reason="too slow")
 def test_process_light_client_update_at_period_boundary(spec, state):
+    old_config = spec.config
+    override_config_fork_epochs(spec, state)
```
hwwhww, line 84: Since we want to transform these tests to test vectors, the override config has to be yielded to test vectors.
I'd suggest using `spec_configured_state_test` decorator that would yield the overridden configs:

1. Write a new *decorator* `override_config_fork_epoch` that call `spec_configured_state_test`
    - Usage example: https://github.com/ethereum/consensus-specs/pull/2767/files 
2. Put this decorator to all these sync protocol test cases

specs/capella/beacon-chain.md
```diff
@@ -318,6 +320,24 @@ def is_partially_withdrawable_validator(validator: Validator, balance: Gwei) ->
...
+    Return the fork version at the given ``epoch``.
+    """
+    if epoch >= CAPELLA_FORK_EPOCH:
+        return CAPELLA_FORK_VERSION
+    if epoch >= BELLATRIX_FORK_EPOCH:
+        return BELLATRIX_FORK_VERSION
+    if epoch >= ALTAIR_FORK_EPOCH:
+        return ALTAIR_FORK_VERSION
+    return GENESIS_FORK_VERSION
+```
```
hwwhww, line 339: I'd kinda hesitate to say where we should put this helper on since the core protocol doesn't need it. Other options are:
- `sync-protocol.md`: seems "correct" way, but then we will need to update sync protocol in other directories.
- `fork.md`: might be better than "beacon-chain.md"?

/cc @ralexstokes @djrtwo what do you think?

hwwhww (commented):
👍 on adding this helper.

Some suggestions about where to put it down and tests.

ralexstokes:
iirc having the `fork_version` here means the update is "self-contained" wrt verification

imagine there is some relayer node that does not have a locally configured fork schedule, they can still determine if this is a valid update or not (and hence to pass along or not) but it requires having the fork version to compute the right domain

note this "relayer" could even be a smart contract where its better UX to not need to also bring along a fork schedule as ambient knowledge

wemeetagain:
I think having the update self-contained wrt verification doesn't buy us much here. I'd argue that there is already ambient knowledge required in the current protocol, which amounts to knowing the config+preset, genesis values, and having fork awareness of the beacon state layout. So if we can derive the fork schedule from the config, we're not really requiring additional knowledge.

Also would mention that the sync spec gives just one possible concrete protocol for syncing, but different relayers/consumers will have different requirements. The protocol likely doesn't need to support the lowest-common-denominator, but rather to just support the simplest approach.

specs/capella/beacon-chain.md
etan-status: Moved it to `fork.md` for now, it is where the various `_FORK_EPOCH` and `_FORK_VERSION` are introduced, and the helper functions fits well in its theme. Also, the other functions in `fork.md` are not used in `beacon-chain.md` either, so it fits naturally :)

tests/core/pyspec/eth2spec/test/altair/unittests/test_sync_protocol.py
etan-status: Managed to create a decorator that applies the config correctly :) And added a test for the decorator as well.

tests/core/pyspec/eth2spec/test/altair/unittests/test_config_override.py
```diff
@@ -17,3 +22,28 @@ def test_config_override(spec, state):
     # TODO: it would be nice if the create_genesis_state actually outputs a state
     #  for the fork with a slot that matches at least the fork boundary.
     # assert spec.get_current_epoch(state) >= 4
+
+
+@with_all_phases
+@spec_state_test_with_matching_config
+def test_override_config_fork_epoch(spec, state):
```
hwwhww, line 29: very thoughtful @etan-status 👍

hwwhww (approved):
LGTM

@ralexstokes @wemeetagain
Do people still have objections? Please let us know. :)

PR #2925, opened 2022-06-27
fix typo and specify something

fradamt:
- committee_index is used as an input to compute_subnet_for_attestation but it's not previously defined
- attestation.data.committee_index is incorrect, the field is "index"


ralexstokes (approved):
lgtm

hwwhww (approved):

PR #2927, opened 2022-06-29
Allow passing redundant `next_sync_committee`

etan-status:
The producer of `LightClientUpdate` structures usually does not know how
far the `LightClientStore` on the client side has advanced. Updates are
currently rejected when including a redundant `next_sync_committee` not
advancing the `LightClientStore`. Behaviour is changed to allow this.


hwwhww (approved):
lgtm 👍

I like the symmetry of `is_finality_update` and `is_sync_committee_update` helpers!

hwwhww (approved):

PR #2929, opened 2022-06-30
Cleanup light client tests

etan-status:
Various cleanups and minor fixes:
- Consistent terminology:
  - `signed_block` -> `attested_block`
  - `finalized_block_header` -> `finalized_header`
  - `snapshot_period` -> `store_period`
- Use correct block in finality test (`blocks[-1]` instead of new one)
- Add `signed_block_header` func to get header from `SignedBeaconBlock`
- Remove `block_header` from `get_sync_aggregate` helper arguments
- Use `state_transition_with_full_block` as shortcut for multiple calls
- Have `finalized_header` actually be header instead of full block body
- Consistent ordering of `assert` to match structure definition


hwwhww (approved):
lgtm

thanks for the cleanup!

PR #2931, opened 2022-06-30
EIP-4844: Optimistic Sync

Inphi:
With the data availability block requirement introduced for EIP-4844, we also need to tweak optimistic sync so it's consistent with which blocks are considered valid.

Right now, a `VALID` block means all of its ancestors will also become `VALID` blocks. This breaks with EIP-4844 when blob sidecars are introduced as an optimistcally imported block lacking its sidecar will be incorrectly promoted to `VALID`. This makes It then becomes easy for an attacker to trick peers into treating actual invalid blocks with missing data as valid.

To fix this we change the block used as a basis for the block status transition. When a block transitions from `NOT_VALIDATED` to `VALID`, identify the latest ancestor whereby all of its own ancestors satisfy data availability.

Here's an example illustrating this:

image

In the diagram, `B` is the only block missing data. We ignore both `A` and `B` to transition `C` and its ancestors from `NOT_VALIDATED` -> `VALID`.

cc: @terencechain

### DIFF
--- a/specs/eip4844/optimistic_sync.md
+++ b/specs/eip4844/optimistic_sync.md
@@ -1,0 +1,22 @@
+# EIP-4844 -- Optimistic Sync
+
+This document is an extension of the [Optimistic Sync](/sync/optimistic.md) guide for EIP-4844.
+All behaviors and definitions defined in Optimistic Sync guide carry over unless explicitly noted or overridden.
+
+EIP-4844 extends the block validity requirement for [data availability](./validator.md#is_data_available). When a block transitions from `NOT_VALIDATED` -> `VALID`, its ancestors no longer unconditionally transition as well. But rather, only ancestors that do not have ancestors themselves with missing blocks can be considered `VALID`.
+
+```python
+def latest_valid_candidate_block(opt_store: OptimisticStore, block: BeaconBlock) -> BeaconBlock:
+    # Assuming the input `block` is VALID
+    chain = []
+    while True:
+        chain.append(block)
+        if not is_optimistic(opt_store, block) or block.parent_root == Root():
+            break
+        block = opt_store.blocks[block.parent_root]
+    b = next(b for b in reversed(chain)
+            if not is_data_available(b.slot, hash_tree_root(b.body), b.body.blob_kzgs), None)
+    return opt_store.blocks[b.parent_root] if b is not None else None
+```
+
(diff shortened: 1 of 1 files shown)


mkalinin:
I don't think an Optimistic Sync spec is the right place for this logic. It should rather be incorporated into modified fork choice rule that filters out blocks with unavailable data utilizing the `is_data_available(...)` function which body is supposed to be replaced by DAS in full danksharding proposal.

mkalinin:
Optimistic Sync validity is applied to `ExecutionPayload` and outsourced to Execution Layer clients. Data availability and EL block validity are orthogonal things to me unless I am missing something important. Thus, mixing them under the hood of optimistic sync is not the right thing to do unless we have a strong argument in favour of doing so.

In my understanding, DA is a pure consensus layer responsibility and should be a part of the fork choice. One of possible implementations that comes to my mind is modification of a `filter_block_tree` function in the following way:
```python
def filter_block_tree(store: Store, block_root: Root, blocks: Dict[Root, BeaconBlock]) -> bool:
    block = store.blocks[block_root]

    if not is_data_available(block.slot, block_root, block.body.blob_kzgs):
        return False

    ...
```

cc @dankrad @protolambda

Inphi:
I see. My original understanding of Optimistic Sync is that it's closely tied to consensus, albeit as an optimization. Fair point to keep DA separate from it.

Thanks for reviewing this. I'll follow up with the proper CL changes on a separate PR.

terencechain:
I think there are two parts to this:
The first part is fork choice - block w/o a sidecar can't be considered as head, this is in the consensus spec
The second part is p2p/rpc's req/res - node should still reply on request which the block w/o a sidecar with bool field to indicate "optimistic" or "sidecar-missing" This is not part of the consensus spec. So, where should we put this?

Inphi:
I'm not sure the second part is needed. "optimistic" and "sidecar-missing" are identical p2p/rpc scenarios imo. So I'd think that the way the spec works right now should suffice - where optimistic/side-missing blocks can be propagated.

mkalinin:
Req/resp'ing a block may remain independent from req/resp'ing a sidecar. And a node should probably be liberate to remote peers that have a block but don't have a sidecar, likewise, to the optimistic syncing remote peers that may serve blocks with not yet validated payload (which eventually appears to be invalid)

PR #2935, opened 2022-07-12
Add `LightClientBootstrap`

etan-status:
Introduces a new `LightClientBootstrap` structure to allow setting up a
`LightClientStore` with the initial sync committee and block header from
a user-configured trusted block root.

This leads to new cases where the `LightClientStore` is only aware of
the current but not the next sync committee. As a side effect of these
new cases, the store's `finalized_header` may now  advance into the next
sync committee period before a corresponding `LightClientUpdate` with
the new sync committee is obtained, improving responsiveness.

Note that so far, `LightClientUpdate.attested_header.slot` needed to be
newer than `LightClientStore.finalized_header.slot`. However, it is now
necessary to also consider certain older updates to try and backfill the
`next_sync_committee`. The `is_better_update` helper is also updated to
improve `best_valid_update` tracking.

### DIFF
--- a/specs/altair/sync-protocol.md
+++ b/specs/altair/sync-protocol.md
@@ -13,20 +13,24 @@
 - [Preset](#preset)
   - [Misc](#misc)
 - [Containers](#containers)
+  - [`LightClientBootstrap`](#lightclientbootstrap)
   - [`LightClientUpdate`](#lightclientupdate)
   - [`LightClientStore`](#lightclientstore)
 - [Helper functions](#helper-functions)
   - [`is_sync_committee_update`](#is_sync_committee_update)
   - [`is_finality_update`](#is_finality_update)
   - [`is_better_update`](#is_better_update)
+  - [`is_next_sync_committee_known`](#is_next_sync_committee_known)
   - [`get_safety_threshold`](#get_safety_threshold)
   - [`get_subtree_index`](#get_subtree_index)
   - [`compute_sync_committee_period_at_slot`](#compute_sync_committee_period_at_slot)
+- [Light client initialization](#light-client-initialization)
+  - [`initialize_light_client_store`](#initialize_light_client_store)
 - [Light client state updates](#light-client-state-updates)
-    - [`process_slot_for_light_client_store`](#process_slot_for_light_client_store)
-    - [`validate_light_client_update`](#validate_light_client_update)
-    - [`apply_light_client_update`](#apply_light_client_update)
-    - [`process_light_client_update`](#process_light_client_update)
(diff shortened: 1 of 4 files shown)


specs/altair/sync-protocol.md
```diff
@@ -276,21 +349,25 @@ def validate_light_client_update(store: LightClientStore,
...
-#### `apply_light_client_update`
+### `apply_light_client_update`
 
 ```python
 def apply_light_client_update(store: LightClientStore, update: LightClientUpdate) -> None:
     store_period = compute_sync_committee_period_at_slot(store.finalized_header.slot)
     update_finalized_period = compute_sync_committee_period_at_slot(update.finalized_header.slot)
-    if update_finalized_period == store_period + 1:
+    if not is_next_sync_committee_known(store):
+        assert update_finalized_period == store_period
```
hwwhww, line 359: Is there an edge-case, that all the incoming updates are of newer finalized periods? e.g., the initial bootstrap is not fresh enough.

hwwhww (commented):
Having a bootstrap looks reasonable to me.

/cc @dapplion and @wemeetagain, you may be interested in this.

specs/altair/sync-protocol.md
etan-status: No additional edge-case is introduced, this is just a way to obtain the initial sync committee. From there on, historic light client updates need to be applied, same way as before. Or, alternatively, regular forward sync can be used (based on the current finalized state root / block root) to proceed to a newer sync committee period for which data is available.

specs/altair/sync-protocol.md
```diff
@@ -324,7 +401,16 @@ def process_light_client_update(store: LightClientStore,
...
+            update.finalized_header.slot > store.finalized_header.slot
+            or (
+                not is_next_sync_committee_known(store)
+                and is_sync_committee_update(update) and is_finality_update(update) and (
+                    compute_sync_committee_period_at_slot(update.finalized_header.slot)
+                    == compute_sync_committee_period_at_slot(update.attested_header.slot)
+                )
+            )
+        )
     ):
```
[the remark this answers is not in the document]
etan-status: 66325d0e9bbd0bd72588102a4aa3d34cf65df3da

specs/altair/sync-protocol.md
etan-status: Also, an older `LightClientUpdate` for the current store period is accepted to obtain the `next_sync_committee`, if the question is referring for that case.

wemeetagain (approved):
LGTM, this is in line with what we discussed in Amsterdam

PR #2938, opened 2022-07-14
Add functions for deriving light client data

etan-status:
Adds `create_light_client_bootstrap` and `create_light_client_update`
functions as a reference implementation for serving light client data.
This also enables a new test harness to verify that light client data
gets applied to a `LightClientStore` as expected.

PR #2940, opened 2022-07-15
Fix Capella genesis testing tool

hwwhww:
- Fix `create_genesis_state`: should have updated `previous_version` and `current_version` 
- Add Capella version `initialize_beacon_state_from_eth1`


etan-status (approved):

etan-status:
Thanks, this indeed fixes the problems with `make test` in #2938 (merged this PR into there as well)

specs/capella/beacon-chain.md
```diff
@@ -489,3 +490,59 @@ def process_bls_to_execution_change(state: BeaconState,
         + address_change.to_execution_address
     )
 ```
+
+## Testing
+
+*Note*: The function `initialize_beacon_state_from_eth1` is modified for pure Capella testing only.
+Modifications include:
+1. Use `CAPELLA_FORK_VERSION` as the current fork version.
```
ralexstokes, line 498:
```suggestion
1. Use `CAPELLA_FORK_VERSION` as the previous fork version.
```

should this say `previous` as the modification?

specs/capella/beacon-chain.md
```diff
@@ -489,3 +490,59 @@ def process_bls_to_execution_change(state: BeaconState,
...
+    # Set genesis validators root for domain separation and chain versioning
+    state.genesis_validators_root = hash_tree_root(state.validators)
+
+    # Fill in sync committees
+    # Note: A duplicate committee is assigned for the current and next committee at genesis
+    state.current_sync_committee = get_next_sync_committee(state)
+    state.next_sync_committee = get_next_sync_committee(state)
+
+    # Initialize the execution payload header
+    # If empty, will initialize a chain that has not yet gone through the Merge transition
```
ralexstokes, line 544: this is an interesting point.... is there any value in testing a capella net pre-merge?

I think its an open question but given that we are starting this beacon chain at cappella, it feels like testing a pre-merge state is something that would never hit in a public-facing testnet or mainnet

ralexstokes (commented):
generally looks good!

left some feedback

specs/capella/beacon-chain.md
hwwhww: Ah, I think it changes both. nice catch!

specs/capella/beacon-chain.md
hwwhww: agreed. I'd remove this comment.

djrtwo (approved):

PR #2947, opened 2022-07-21
Manually trigger `LightClientStore` force updates

etan-status:
Replaces `process_slot_for_light_client_store` which force updates the
`LightClientStore` automatically based on `finalized_header` age with
`try_light_client_store_force_update` which may be manually called based
on use case dependent heuristics if light client sync appears stuck.
Not all use cases share the same risk profile.

### DIFF
--- a/specs/altair/light-client/sync-protocol.md
+++ b/specs/altair/light-client/sync-protocol.md
@@ -29,9 +29,9 @@
 - [Light client initialization](#light-client-initialization)
   - [`initialize_light_client_store`](#initialize_light_client_store)
 - [Light client state updates](#light-client-state-updates)
-  - [`process_slot_for_light_client_store`](#process_slot_for_light_client_store)
   - [`validate_light_client_update`](#validate_light_client_update)
   - [`apply_light_client_update`](#apply_light_client_update)
+  - [`try_light_client_store_force_update`](#try_light_client_store_force_update)
   - [`process_light_client_update`](#process_light_client_update)
   - [`process_light_client_finality_update`](#process_light_client_finality_update)
   - [`process_light_client_optimistic_update`](#process_light_client_optimistic_update)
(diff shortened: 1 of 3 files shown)


specs/altair/light-client/sync-protocol.md
```diff
@@ -285,28 +285,7 @@ def initialize_light_client_store(trusted_block_root: Root,
...
-        # Forced best update when the update timeout has elapsed.
-        # Because the apply logic waits for `finalized_header.slot` to indicate sync committee finality,
-        # the `attested_header` may be treated as `finalized_header` in extended periods of non-finality
-        # to guarantee progression into later sync committee periods according to `is_better_update`.
-        if store.best_valid_update.finalized_header.slot <= store.finalized_header.slot:
-            store.best_valid_update.finalized_header = store.best_valid_update.attested_header
-        apply_light_client_update(store, store.best_valid_update)
-        store.best_valid_update = None
-```
+- `try_light_client_store_force_update` MAY be called based on use case dependent heuristics if light client sync appears stuck.
```
hwwhww, line 288: [nitpicking]
- "force" sounds like it will be overridden.
- "try" sounds like it may be overridden or not.

`apply_light_client_store_force_update` may be better but `apply_light_client_update` is already in use. 😅

any better ideas? if not, I think it's also fine to leave it.

hwwhww, line 288: it seems a big change to light client behavior.

/cc @wemeetagain @dapplion

etan-status: Previous behaviour can be simulated by calling this on every slot, same as before. But new behaviour allows for use cases such as historic data sync where it is not appropriate to force-apply every single update just because it is historic.

etan-status: What about `process_light_client_store_force_update`? The `try` indicates the optional manner though, it may or may not update, depending on whether there is a  pending `best_valid_update` that is sufficiently old.

hwwhww: > What about `process_light_client_store_force_update`?

That could work! :)

hwwhww: @etan-status  I see. So it is backward compatible. 👍

hwwhww (approved):
LGTM

PR #2950, opened 2022-07-22
Move `merkle/single_proof` tests to `light_client/single_merkle_proof`

hwwhww:
The single Merkle proof is only be used in light clients. It may make more sense to move these tests under `light_client` directory.

/cc @etan-status


etan-status (approved):
The merkle proofs are described together with the SSZ format, living outside the light client specifications. On the other hand, the provided tests are indeed specific to light client structures, so moving them is useful to group up additional tests for implementations willing to adopt light client support.

In general, having more general tests available for merkleization would be appreciated, but until those exist, having specific tests for light clients and for Eth1 deposit processing at least reflects the current reality.

Furthermore, even if such general tests were present, the light client specific tests are still useful to enable testing of optimized merkleization implementations that do not support arbitrary indices and multiproofs.

✅ Verified that `rm -rf ../consensus-spec-tests && make gen_light_client` produce the new test vectors
✅ Verified that Nimbus test runner still passes the moved tests

ralexstokes (approved):
this is used in deposit processing so is not *solely* a light client concern but I am fine w/ this change

I think it could live in either place and as long as existing consumers (who are likely not light clients) know to find these tests here then we are good to go

hwwhww:
> This is used in deposit processing

Good point! But somehow the existing test cases are so Altair & light-client-specific.

Assume the consumers are staking service providers, maybe we should create a new test vector folder for deposit? To provide
- deposit proof test vectors
- withdrawal prefix validation test vectors

djrtwo (approved):

PR #2951, opened 2022-07-22
Describe light client sync process

etan-status:
Provides documentation about the process that a light client may use to
sync a local `LightClientStore`.


specs/altair/light-client/light-client.md
```diff
@@ -0,0 +1,37 @@
...
+<!-- /TOC -->
+
+## Introduction
+
+This document explains how light clients MAY obtain light client data to sync with the network.
+
+## Light client sync process
+
+1. The light client MUST be configured out-of-band with a spec/preset (including fork schedule), with `genesis_state` (including `genesis_time` and `genesis_validators_root`), and with a trusted block root. The trusted block SHOULD be within the weak subjectivity period, and its root SHOULD be from a finalized `Checkpoint`.
+
```
hwwhww, line 24: I see you use newlines for readability. I'm fine with either way.

/cc @djrtwo in case you have objection :p

hwwhww (approved):
Good document for light clients!

specs/altair/light-client/light-client.md
etan-status: Not sure I understand correctly, but there is no newline within the paragraph, similar to the rest of the spec. GitHub wraps the lines here (and displays the entire paragraph as line 23).

etan-status: Or, you mean, between the paragraphs?

etan-status: There are multiple conflicting styles within the spec w.r.t. enumerations:

- Newline between items but not within a paragraph, for example in `specs/phase0/weak-subjectivity.md`, or here
- No newline between items, but hard newlines within paragraph and leading spaces on each line, for example in `specs/phase0/validator.md`. The column for the line break appears to be different in each of the files, from a somewhat hard 80 in `specs/bellatrix/p2p-interface.md` through 120-ish in other files.

Furthermore, there are a few cases where the items are all introduced with repeated `1.`, `1.`, `1.` instead of sequential `1.`, `2.`, `3.`..., for example, `sync/optimistic.md` or `tests/README.md`.

I'm fine with either here, there doesn't seem to be strict formatting rules for now. Having absolutely no hard newlines, both within an item as well as between them, makes it more difficult to read while editing, but there are more readers than writers.

hwwhww: I meant the newline between the paragraphs `1`, `2`, `3`, ... and that would make it has more margin space between paragraphs. For example:
image


> multiple conflicting styles

Right, that happened since we have multiple contributors. Currently, we don't plan to align all specs since it would introduce scary diff lines. 😅

PR #2953, opened 2022-07-26
Do not overload index with WithdrawalIndex and ValidatorIndex

potuz:
in the function `withdraw_balance` the variable `index` takes different meanings and types. This confused at least one member of our team.

### DIFF
--- a/specs/capella/beacon-chain.md
+++ b/specs/capella/beacon-chain.md
@@ -269,13 +269,13 @@ class BeaconState(Container):
 #### `withdraw`
 
 ```python
-def withdraw_balance(state: BeaconState, index: ValidatorIndex, amount: Gwei) -> None:
+def withdraw_balance(state: BeaconState, validator_index: ValidatorIndex, amount: Gwei) -> None:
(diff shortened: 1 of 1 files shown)


ralexstokes:
fwiw the types do not change and I'd say this implies the meaning does not change but I'd support this PR if it clarifies things

potuz:
> fwiw the types do not change and I'd say this implies the meaning does not change but I'd support this PR if it clarifies things

I think the confusing thing is statements like 
```python
        index=state.next_withdrawal_index,
        address=ExecutionAddress(state.validators[index].withdrawal_credentials[12:]),
```
Where in one line `index` is assigned a `WithdrawalIndex` and in the very next line, `index` is used with the `ValidatorIndex` that was passed to the function.

hwwhww (approved):
Agreed that  `index=state.next_withdrawal_index, address=ExecutionAddress(state.validators[index].withdrawal_credentials[12:])` is confusing. 👍

djrtwo (approved):
makes sense!

PR #2954, opened 2022-07-27
Opti sync: how to apply `latestValidHash` when payload is `INVALID`

mkalinin:
Clarifies on how to handle `latestValidHash` on CL side. This PR doesn't introduce any breaking changes but instead makes current CL behaviour into the spec and leaves recommendation on what to do when `latestValidHash` can't be found in a block tree

**UPD**
It's written in a way that uses `latestValidHash` to only invalidate blocks. No valid semantics of `latestValidHash` is leveraged by this PR

cc @potuz @ajsutton @paulhauner

PR #2956, opened 2022-07-27
Add ToC to `optimistic.md`

hwwhww:
Based on the #2954 branch

- Fixed indent of "What about Light Clients?" and "What if TERMINAL_BLOCK_HASH is used?"
- Added ToC


mkalinin (approved):
LGTM 👍

PR #2958, opened 2022-07-28
Opti sync: elaborate on why sync optimistically

mkalinin:
This PR provides more details on the rationale to have optimistic sync as a generalized sync solution for consensus engines.

cc @tersec @paulhauner

### DIFF
--- a/sync/optimistic.md
+++ b/sync/optimistic.md
@@ -272,10 +272,10 @@ optimistic blocks (and vice-versa).
 
 ### Why sync optimistically?
 
-Most of execution engines use state sync as a default sync mechanism on Ethereum Mainnet 
+Most execution engines use state sync as a default sync mechanism on Ethereum Mainnet 
 because executing blocks from genesis takes several weeks on commodity hardware.
 
-State sync requires the knowledge of the current head of the chain to eventually converge.
+State sync requires the knowledge of the current head of the chain to converge eventually.
 If not constantly fed with the most recent head, state sync won't be able to complete
 because the recent state soon becomes unavailable due to state trie pruning.
 
--- a/sync/optimistic.md
+++ b/sync/optimistic.md
@@ -24,6 +24,7 @@
   - [Participating in Sync Committees](#participating-in-sync-committees)
 - [Ethereum Beacon APIs](#ethereum-beacon-apis)
 - [Design Decision Rationale](#design-decision-rationale)
+  - [Why sync optimistically?](#why-sync-optimistically)
   - [Why `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY`?](#why-safe_slots_to_import_optimistically)
   - [Transitioning from VALID -> INVALIDATED or INVALIDATED -> VALID](#transitioning-from-valid---invalidated-or-invalidated---valid)
   - [What about Light Clients?](#what-about-light-clients)
--- a/sync/optimistic.md
+++ b/sync/optimistic.md
@@ -279,9 +279,9 @@ State sync requires the knowledge of the current head of the chain to converge e
 If not constantly fed with the most recent head, state sync won't be able to complete
 because the recent state soon becomes unavailable due to state trie pruning.
 
-Optimistic block import (i.e. import without waiting for a payload to be executed by the engine)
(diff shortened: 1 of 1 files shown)


hwwhww (approved):
LGTM. Just some minor proofreading.

**[Review period]** I will leave this PR open for review till August 2nd UTC 23:59.

sync/optimistic.md
```diff
@@ -270,6 +270,23 @@ optimistic blocks (and vice-versa).
 
 ## Design Decision Rationale
 
+### Why sync optimistically?
+
+Most execution engines use state sync as a default sync mechanism on Ethereum Mainnet
```
djrtwo, line 275: Block sync also requires a cryptoeconomic head to do reverse header sync and then forward block sync.

I might make this more generic about "execution engine sync methods" needing cryptoeconomic heads to be able to sync independently from the consensus engine

sync/optimistic.md
```diff
@@ -270,6 +270,23 @@ optimistic blocks (and vice-versa).
...
+### Why sync optimistically?
+
+Most execution engines use state sync as a default sync mechanism on Ethereum Mainnet 
+because executing blocks from genesis takes several weeks on commodity hardware.
+
+State sync requires the knowledge of the current head of the chain to converge eventually.
+If not constantly fed with the most recent head, state sync won't be able to complete
+because the recent state soon becomes unavailable due to state trie pruning.
+
+Optimistic block import (i.e. import without waiting for a payload to be executed by the engine)
```
djrtwo, line 282:
```suggestion
Optimistic block import (i.e. import when the execution engine *cannot* currently validate the payload)
```

sync/optimistic.md
```diff
@@ -270,6 +270,23 @@ optimistic blocks (and vice-versa).
...
+breaks a deadlock between the state sync and processing a beacon block by skipping payload
+execution as long as execution engine is syncing.
```
djrtwo, line 284:
```suggestion
breaks a deadlock between the execution layer sync process and importing beacon blocks
while the execution engine is syncing.
```

sync/optimistic.md
```diff
@@ -270,6 +270,23 @@ optimistic blocks (and vice-versa).
...
+
+Optimistic sync works nicely with execution engines using block execution as a default
```
djrtwo, line 286: It not only works nicely but is also requisite (unless CL and EL lockstep execute from genesis

djrtwo (commented):
I think this is more generalized to any EE design in which the EE syncs independently of CL lockstep execution

sync/optimistic.md
mkalinin: Block sync allows for a different initial sync strategy. CL may send a checkpoint payload to EL and wait until it catches up with this payload then lock-step - but this would be suboptimal. I agree that "nicely" might be too weak word in this context, but optimistic sync is not a requisite for block sync either. Will try to rephrase this statement

mkalinin: Done

tersec:
Seems to reasonably articulate the general consensus on this point, as a descriptive matter.

tersec (approved):

sync/optimistic.md
mkalinin: @djrtwo waiting for you to take a look at the update of this statement and pushing the button if OK with you

sync/optimistic.md
```diff
@@ -270,6 +270,28 @@ optimistic blocks (and vice-versa).
...
+If not constantly fed with the most recent head, state sync won't be able to complete
+because the recent state soon becomes unavailable due to state trie pruning.
+
+Optimistic block import (i.e. import when the execution engine *cannot* currently validate the payload)
+breaks a deadlock between the execution layer sync process and importing beacon blocks
+while the execution engine is syncing.
+
+Optimistic sync is also an optimal strategy for execution engines using block execution as a default
+sync mechanism (e.g. Erigon). Alternatively, a consensus engine may inform the execution engine with a payload
+obtained from a checkpoint block, then wait till the execution layer catches up with it and proceed
```
djrtwo, line 288:
```suggestion
obtained from a checkpoint block, then wait until the execution layer catches up with it and proceed
```

djrtwo (approved):

PR #2962, opened 2022-08-01
a few more bellatrix tests

djrtwo:


mkalinin (approved):

tests/core/pyspec/eth2spec/test/bellatrix/transition/test_transition.py
```diff
@@ -33,3 +36,31 @@ def test_sample_transition(state, fork_epoch, spec, post_spec, pre_tag, post_tag
 
     yield "blocks", blocks
     yield "post", state
+
+
+@with_fork_metas([
+    ForkMeta(pre_fork_name=pre, post_fork_name=post, fork_epoch=8) for pre, post in AFTER_BELLATRIX_PRE_POST_FORKS
+])
```
hwwhww, line 43: The two tests look general and can be reused with future forks:

1. Merge `bellatrix/transition/test_transition.py` into `altair/transition/test_transition.py`
2. Replace `AFTER_BELLATRIX_PRE_POST_FORKS` with `ALL_PRE_POST_FORKS`.
3. `AFTER_BELLATRIX_PRE_POST_FORKS` can be removed totally.

hwwhww (approved):
LGTM

Re: slashed proposer issue

I was thinking that another solution is setting `slash_fraction=0` when calling `randomize_state(spec, state, slash_fraction=0)`. However, I just found it's not fully correct due to our `slash_random_validators` implementation: see #2968

I'm fine either way.

PR #2965, opened 2022-08-03
Extend fork_choice test format with on_payload_info

mkalinin:
Extends `fork_choice` tests format with `payload_info`, discussion in https://github.com/ethereum/consensus-spec-tests/issues/28


hwwhww (commented):
Hey @mkalinin, I implemented a basic test in #2982 

I am still on the fence between extending `tests/formats/fork_choice/README.md` or create a new `tests/formats/sync/README.md` that refers to the former. Thinking if the latter may be more clear in some ways. What do you think?

mkalinin:
> Hey @mkalinin, I implemented a basic test in #2982
> 
> I am still on the fence between extending `tests/formats/fork_choice/README.md` or create a new `tests/formats/sync/README.md` that refers to the former. Thinking if the latter may be more clear in some ways. What do you think?

Unless there is any concern or obstacle, I would take a simple path forward which is extending the test by one more step, the step is optional and may not be supported by CL clients

hwwhww (approved):
lgtm

PR #2968, opened 2022-08-09
Fix `slash_random_validators`: make `fraction=0` return 0 indices

hwwhww:
### Issue

`slash_random_validators` function would force slash `state.validators[0]`.
However, we expect it doesn't slash any validator when we call `randomize_state(spec, state, slash_fraction=0)`.

### Proposed fix

Remove `index == 0` condition.

Note: it would change many randomized test vectors.

### DIFF
--- a/tests/core/pyspec/eth2spec/test/helpers/random.py
+++ b/tests/core/pyspec/eth2spec/test/helpers/random.py
@@ -90,7 +90,7 @@ def slash_random_validators(spec, state, rng, fraction=0.5):
     for index in range(len(state.validators)):
         # slash at least one validator
         sampled = rng.random() < fraction
-        if index == 0 or sampled:
+        if sampled:
             spec.slash_validator(state, index)
             slashed_indices.append(index)
     return slashed_indices


tests/core/pyspec/eth2spec/test/helpers/random.py
```diff
@@ -90,7 +90,7 @@ def slash_random_validators(spec, state, rng, fraction=0.5):
     for index in range(len(state.validators)):
         # slash at least one validator
         sampled = rng.random() < fraction
-        if index == 0 or sampled:
+        if sampled:
```
ralexstokes, line 93: this fn was written with the intent to have *at least* 1 validator slashed

I'd rather add another parameter, something like `required_slashed_count` and then do the random sampling and then iterate over the set and force them to be slashed so that the total number of vals slashed is `required_slashed_count`

we can just keep `required_slashed_count` as a default param at 1

the new code that requested this change can then explicitly set that param to 0

ralexstokes (commented):
this would break an invariant I was explicitly using when writing the `random` tests initially

left one suggestion to fix below

if we don't like this I can go review the usage to see what would break but imo just adding backwards-compatibility is better

tests/core/pyspec/eth2spec/test/helpers/random.py
hwwhww: I see. 

> required_slashed_count
but `required_slashed_count` value may be conflicting with `fraction` value.

Alternatively, how about we handle `fraction == 0` case specifically? (The default `fraction` value is `0.5`)
```python
if (index == 0 and fraction != 0) or sampled:
```

ralexstokes: seems workable but I like having independent knobs we can set without worrying about their interaction

JurbinM (approved):

leolara:

PR #2969, opened 2022-08-11
Merge mainnet ttd and bellatrix values

djrtwo:
* `BELLATRIX_FORK_EPOCH`: 144896  (Sept 6, 2022, 11:34:47am UTC)
* `TERMINAL_TOTAL_DIFFICULTY`: 58750000000000000000000 (Estimated: Sept 15, 2022)

PR #2974, opened 2022-08-17
Update Bellatrix fork epoch in README

0xYYY:
Seems like this may be missed by https://github.com/ethereum/consensus-specs/pull/2969?


djrtwo (approved):
Thanks!

PR #2976, opened 2022-08-18
Add tests of processing objects of `N-2` fork version

hwwhww:
- `test_process_voluntary_exit.py`
    - `test_voluntary_exit_with_previous_fork_version__valid`:
        - signing voluntary exit object with `ALTAIR_FORK_VERSION`
        - processing voluntary exit object with Bellatrix state
        - result: valid
    - `test_voluntary_exit_with_genesis_fork_version__invalid`:
        - signing voluntary exit object with `GENESIS_FORK_VERSION`
        - processing voluntary exit object with Bellatrix state
        - result: invalid
- `test_process_deposit.py`
    - all fork versions are *valid* (no exception), but only `GENESIS_FORK_VERSION` can add new entry
- Refactoring:
    - Moved `run_voluntary_exit_processing` and `run_deposit_processing` function to the shared helper files.

PR #2977, opened 2022-08-18
v1.2.0-rc.3

djrtwo:
Release some additional important consensus tests (#2976)

- [x] #2979


hwwhww (approved):
need #2979. And then test vectors are all good. 👍

Issue #2980, opened 2022-08-19
EIP4844: Impove performance by avoiding merkle trees when generating KZG challenges


asn-d6:
Both in [compute_aggregated_poly_and_commitment](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/validator.md#compute_aggregated_poly_and_commitment) and in [validate_blobs_sidecar](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/validator.md#validate_blobs_sidecar) we are generating challenge scalars using [hash_to_bls_field()](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/validator.md#hash_to_bls_field) which calls `hash_tree_root()`.

There is actually no reason to compute merkle roots when we are generating these challenge scalars. We should instead just straight sha256 hash the values we care about.

Doing this change will improve performance when we are computing and verifying KZG proofs.

(/cc @hwwhww @Inphi @protolambda )

asn-d6:
@hwwhww Thanks for the patch! I think @Inphi is gonna implement the change on his side pretty soon, so let's wait for his performance results before we merge this.

Inphi:
This indeed yields a significant performance improvement:
```
name                 old time/op    new time/op    delta
VerifyMultiple/8-6     73.9ms ± 0%    40.7ms ± 0%   ~     (p=1.000 n=1+1)
VerifyMultiple/16-6     138ms ± 0%      81ms ± 0%   ~     (p=1.000 n=1+1)

name                 old alloc/op   new alloc/op   delta
VerifyMultiple/8-6     25.2MB ± 0%    22.1MB ± 0%   ~     (p=1.000 n=1+1)
VerifyMultiple/16-6    50.4MB ± 0%    44.1MB ± 0%   ~     (p=1.000 n=1+1)

name                 old allocs/op  new allocs/op  delta
VerifyMultiple/8-6       400k ± 0%      301k ± 0%   ~     (p=1.000 n=1+1)
VerifyMultiple/16-6      800k ± 0%      603k ± 0%   ~     (p=1.000 n=1+1)
```

Verifying blob txs is about 40-50% faster with this.

asn-d6:
Closing this since https://github.com/ethereum/consensus-specs/pull/2981 has been merged. Thanks all!

PR #2981, opened 2022-08-22
Modify `hash_to_bls_field` logic

hwwhww:
Address #2980: replace `hash_tree_root(x)` with `hash(ssz_serialize(x))`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -586,6 +586,7 @@ class EIP4844SpecBuilder(BellatrixSpecBuilder):
         return super().imports(preset_name) + f'''
 from eth2spec.utils import kzg
 from eth2spec.bellatrix import {preset_name} as bellatrix
+from eth2spec.utils.ssz.ssz_impl import serialize as ssz_serialize
 '''
 
     @classmethod
--- a/specs/eip4844/validator.md
+++ b/specs/eip4844/validator.md
@@ -93,9 +93,10 @@ def is_data_available(slot: Slot, beacon_block_root: Root, blob_kzg_commitments:
 ```python
 def hash_to_bls_field(x: Container) -> BLSFieldElement:
     """
-    This function is used to generate Fiat-Shamir challenges. The output is not uniform over the BLS field.
+    Compute 32-byte hash of serialized container and convert it to BLS field.
+    The output is not uniform over the BLS field.
     """
-    return int.from_bytes(hash_tree_root(x), "little") % BLS_MODULUS
+    return int.from_bytes(hash(ssz_serialize(x)), "little") % BLS_MODULUS
 ```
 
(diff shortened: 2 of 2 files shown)


protolambda (approved):

dankrad:
I understand that this improves performance by 2x. This is due to the fact that sha256 of 2*32 bytes hashes a total of 2 blocks, which is 4*32 bytes; if this were not the case, hash_tree_root would have about the same performance as a plain hash, but it is much more parallelizable. 

But it would be important to know the baseline that we are comparing this to.

So far we have had a rule that we don't use serialization inside the spec (though the merge might break that rule already, haven't checked).

hwwhww:
@dankrad @asn-d6 

> So far we have had a rule that we don't use serialization inside the spec (though the merge might break that rule already, haven't checked).

FYI (dank)sharding spec uses [`ssz_serialize()` in `process_sharded_data` function](https://github.com/ethereum/consensus-specs/blob/4370f30b99142d92bf676f0ee6a0aa74f688f891/specs/sharding/beacon-chain.md#sharded-data
).

Also, since `hash_to_bls_field` was copied from sharding spec, if we decide to apply this PR, we may want to update [sharding spec function](https://github.com/ethereum/consensus-specs/blob/4370f30b99142d92bf676f0ee6a0aa74f688f891/specs/sharding/polynomial-commitments.md) as well.  And either way, we should update the function name to `hash_to_bls_field_with_challenge_number` due to the name conflict.

ppopth:
I think it's a good idea to put why we must not use `hash_tree_root` in the comment in the code. (probably put the issue link in the comment)

hwwhww:
@ppopth 

I think it's fine to add comments, but not sure about adding the issue link since we never add issue links to specs before. My life depends on git blame a lot though. 😅

asn-d6 (approved):
LGTM!

asn-d6:
Merged this as is! Thanks for the patch @hwwhww . For more details on how this impacts performance see https://github.com/ethereum/consensus-specs/issues/2980#issuecomment-1222562583

WRT the sharding spec changes suggested in https://github.com/ethereum/consensus-specs/pull/2981#issuecomment-1223993792, I think we should handle these in a different ticket. Both for the sake of progressing this more-urgent ticket faster, but also because I think there is gonna be a big banch of 4844 changes that should be folded in the sharding spec (since 4844 became executable).

WRT the comment suggestion in https://github.com/ethereum/consensus-specs/pull/2981#issuecomment-1225361541, I think it's OK to leave it as is. I think comments about design choices that _were not taken_ might confuse code readers who are mainly interested in the design choices that were taken (also, as @hwwhww said, a git-blame should bring them here. Hey!).

PR #2983, opened 2022-08-25
Canonical JSON mapping

arnetheduck:
This PR adds a canonical JSON mapping to the SSZ specification,
documenting current usage of JSON in tests, API:s and simplifying future
interop work between clients and adjacent specifications such as the
Beacon API. The encoding is appropriate to use with YAML as well

As an important property, this mapping contains a 1:1 mapping of SSZ
type to JSON encoding - this allows round-tripping any object between
JSON and SSZ based on the SSZ schema and usage of the core SSZ types
alone.

The encoding presented in this PR is used in tests and API:s with one
exception: the `ParticipationFlags` type from the Altair spec - it is
recommended we switch encoding in tests and eventually the beacon API to
address this irregularity, so as to avoid a proliferation "special"
primitive types in the SSZ spec that only appear in particular schemas
(and thus making validating general-purpose `SSZ/JSON` parsers more
complex) as well as differences in encoding between fields of the same
SSZ type.

### DIFF
--- a/ssz/simple-serialize.md
+++ b/ssz/simple-serialize.md
@@ -25,6 +25,7 @@
 - [Merkleization](#merkleization)
 - [Summaries and expansions](#summaries-and-expansions)
 - [Implementations](#implementations)
+- [Canonical JSON mapping](#canonical-json-mapping)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 <!-- /TOC -->
@@ -256,3 +257,34 @@ We similarly define "summary types" and "expansion types". For example, [`Beacon
 ## Implementations
 
 See https://github.com/ethereum/eth2.0-specs/issues/2138 for a list of current known implementations.
+
+## JSON mapping
+
+The canonical JSON mapping assigns to each SSZ type a corresponding JSON encoding, enabling an SSZ schema to also define the JSON encoding.
+
+When decoding JSON data, all fields in the SSZ schema must be present with a value. Parsers may ignore additional JSON fields.
+
+| SSZ | JSON | Example |
+| --- | --- | --- |
+| `uintN` | string | `"0"` |
+| `boolean` | bool | `false` |
+| `Container` | object | `{ "field": ... }` |
+| `Vector[type, N]` | array | `[element, ...]` |
+| `Vector[uint8, N]` | hex-byte-string | `"0x1122"` |
+| `Bitvector[N]` | hex-byte-string | `"0x1122"` |
+| `List[type, N]` | array | `[element, ...]` |
+| `List[uint8, N]` | hex-byte-string | `"0x1122"` |
+| `Bitlist[N]` | hex-byte-string | `"0x1122"` |
+| `Union[type_0, type_1, ...]` | selector-object | `{ "selector": number, "data": type_N }`  |
+
+Integers are encoded as strings to avoid loss of precision in 64-bit values.
+
+Aliases are encoded as their underlying type (`byte` is thus encoded as `uint8`).
+
+`hex-byte-string` is a `0x`-prefixed hex encoding of byte data, as it would appear in an SSZ stream.
+
+`List` and `Vector` of `uint8` (and aliases thereof) are encoded as `hex-byte-string`. `Bitlist` and `Bitvector` similarly map their SSZ-byte encodings to a `hex-byte-string`.
+
+`Union` is encoded as an object with a `selector` and `data` field, where the contents of `data` change according to the selector.
+
+> This encoding is used in [beacon-APIs](https://github.com/ethereum/beacon-APIs) with one exception: the `ParticipationFlags` type for the `getStateV2` response, although it is an alias of `uint8`, is encoded as a list of numbers. Future versions of the beacon API may address this incompatibility.


mcdee:
I'm not in favor of this encoding.  Rationale:

- we already differentiate between `byte[n]` and `uint8[n]` in existing JSON data, as noted, and changing the API will be a very painful thing to do
- this creates a situation where `uint8[n]` is encoded differently from (for example) `uint16[n]`

I think that it makes more sense to make `byte` its own type rather than an alias to `uint8`, as we can then specify unique encodings that make sense for each type.

arnetheduck:
> changing the API will be a very painful thing to do

Changing the existing API is optional and can be considered as a separate discussion - this PR is about setting a standard that simplifies future work in this area, avoiding the discussion of array-vs-string once and for all.

> this creates a situation where uint8[n] is encoded differently from (for example) uint16[n]

Yes - the aim of the PR is to clarify that this facto happens for all existing data - the only exception is the flags field which is only used in a debug endpoint. Writing a JSON-compatible SSZ parser and validator becomes more simple (has a smaller bug surface) when these differences at the JSON level are removed.

> I think that it makes more sense to make byte its own type rather than an alias to uint8, as we can then specify unique encodings that make sense for each type.

This difference is not meaningful in a technical context of SSZ  (both are bytes in the wire encoding) - rather than introduce more complexity on the SSZ side, the proposal in this PR is to _simplify_ the JSON side so as not to make these artificial distinctions, in line with the general aim to keep SSZ simple.

The PR itself is a reaction to the fact that there have been numerous incompatibilities between clients because of different field interpretations - having fewer ways to encode the same type of data removes the possibility for these incompatibilities to surface and makes interop easier for example between beacon nodes and validator clients of different kinds.

arnetheduck:
> a very painful thing to do

We've already made several such transitions - it's a regular process that is simple if you break it down:

* we create a new API with a new version that uses the encoding when in hard fork N
* at hard fork N+1, it can be expected that everyone has transitioned to clients that implement both versions
* at hard fork N+2, we remove the old encoding / version / API, to give community a longer reaction time

In this case in particular, we're already considering making states part of a regular checkpoint API which provides an opportunity to amend the encoding as well.

mratsim:
>     * we already differentiate between `byte[n]` and `uint8[n]` in existing JSON data, as noted, and changing the API will be a very painful thing to do

There is no ``uint8[n]`` type currently unless I missed something.

>     * this creates a situation where `uint8[n]` is encoded differently from (for example) `uint16[n]`

There is no `uint16[n]` type used either.

I also don't see the issue in different encoding. The difference is if you encode: [0x0011, 0x2233, 0x4455, 0x6677] it'll be 0x1100332255447766 due to little-endianness.

In practice, we compared the output of one client with the other and there is no uint16 use planned as far as i know.

mcdee:
> The PR itself is a reaction to the fact that there have been numerous incompatibilities between clients because of different field interpretations - having fewer ways to encode the same type of data removes the possibility for these incompatibilities to surface and makes interop easier for example between beacon nodes and validator clients of different kinds.

There has only been one that I am aware of, which is the `ParticipationFlags` situation where Nimbus went a different way to the other clients.  Adherence to the specification has, in the main, been very good.

I think the fundamental issue is this:

> having fewer ways to encode the same type of data

I disagree that `uint8` and `byte` are the same type of data.  One is a number, the other is a set of bits.  And I think that equating them here will ultimately lead to more confusion rather than less.

mratsim:
> I said "for example", but let's use `uint64` then as it exists in the spec. The point is that `uint8` is a number and `byte` is a number of bits. If we consider a simple set of data:
> 
> ```
> field1 uint8 = 1
> field2 uint8[2] = [1, 2]
> field3 uint64[2] = [1, 2]
> ```
> 
> Under the current proposal that would have a JSON encoding:
> 
> ```
> {
>   "field1": "1",
>   "field2": "0x0102",
>   "field3": [1,2]
> }
> ```
> 
> This seems far less obvious to me than:
> 
> ```
> {
>   "field1": "1",
>   "field2": [1,2],
>   "field3": [1,2]
> }
> ```

Ah indeed, i didn't see that uint8 was special-cased. I see no point in special casing it:
1. If it's a length/size concern: use SSZ
2. It adds a branch/complexity

arnetheduck:
> one that I am aware of,

This happens regularly for every time we add a new type and its corresponding tests - we usually catch most incompatibilities early, but the aim is to remove the friction and document the format used in the tests (currently, it's undocumented). Teku for example has a similar bug right now, and others have had them in the past.

> I disagree that uint8 and byte are the same type of data. One is a number, the other is a set of bits. And I think that equating them here will ultimately lead to more confusion rather than less.

This interpretation is not supported by the SSZ spec - instead, it introduces "aliases" as a soft alternative, aliases being "different name for the same thing". Since SSZ is leading in this case, adding adding a distinction to the JSON mapping that does not exist in SSZ is problematic: it means that we no longer have a finite set of canonical basic SSZ types to consider, but also an unbounded set of "interpretations" without backing in  the wire format.

There is no confusion about the difference between an `uint8` and a `byte` when writing an SSZ parser - the confusion only exists at the JSON level that tries to assign additional meaning: the confusion exists because two possibilities exist - removing the flexibility also removes the confusion.

ajsutton:
> Teku for example has a similar bug right now, and others have had them in the past.

This wasn't for lack of a spec, and the particular request doesn't actually have an SSZ representation so this PR would have absolutely no effect on it. The problem occurred because Teku just didn't follow the spec which says the numbers should have been serialised as quoted decimal strings. Some internal details leaked out because of the way the JSON serializer worked. It's why we're moving away from that approach to JSON serialisation.  

I think we've had this discussion elsewhere before, but I'm with jgm that uint8 and bytes are different. It doesn't make sense that uint64 and uint8 would be represented in different ways when serialized as JSON and it's just hostile to users to serialise numbers in hex - fixing that is one of the best differences between the beacon chain APIs and the EL JSON-RPC.

arnetheduck:
>  it's just hostile to users to serialise numbers

JSON is a data interchange format above all - it happens to be readable which is helpful sometimes, but _users_ would not be looking at this data (which user looks at participation flags in the beacon state?).

Also, taking the the participation flags in particular as example, they're not _numbers_ - they're bit sets represented as uint8 and as such, serializing them as decimal numbers is both misleading and the more hostile option for readability - ditto justification flags and other `uint8` in the spec: they're rarely numbers _really_ - for actual numbers we mostly use `uint64`, so this is a non-problem in general.

By using different standards for the beacon API, we can no longer (automatically) derive the JSON encoding from the SSZ schema: we end up with two sources of truth and this is why it becomes a messy problem instead of an easy one: you have to program two distinct serialization standards instead of one ("beacon API considers this a number by ssz considers it a byte") - a direct outcome is that "internal details leak out" with higher frequency. 

We have an opportunity to simplify things here, with only theoretical loss of convenience (ie no real numbers will be lost in the process - some bit flags will be more easy to read thanks to bits aligning with hex digits).

ajsutton:
> JSON is a data interchange format above all - it happens to be readable which is helpful sometimes, but users would not be looking at this data (which user looks at participation flags in the beacon state?).

We disagree on this. JSON is a text based format specifically so that it is easy for both humans and computers to parse. If human readability isn't a consideration we should just use SSZ directly. It's not at all uncommon for humans to use curl to make requests to the REST API view the results directly.

> Also, taking the the participation flags in particular as example, they're not numbers - they're bit sets represented as uint8 and as such, serializing them as decimal numbers is both misleading and the more hostile option for readability - ditto justification flags and other uint8 in the spec: they're rarely numbers really - for actual numbers we mostly use uint64, so this is a non-problem in general.

If the uint8 values currently in use are really bytes then we should change the BeaconState representation to say they are `byte` not `uint8`. We don't want to introduce an inconsistency in JSON representation on numbers based on the argument that currently the spec doesn't use a uint8 for anything that couldn't be argued to actually be a byte. That just sets us up for problems in the future.

> By using different standards for the beacon API, we can no longer (automatically) derive the JSON encoding from the SSZ schema: we end up with two sources of truth and this is why it becomes a messy problem instead of an easy one: you have to program two distinct serialization standards instead of one ("beacon API considers this a number by ssz considers it a byte") - a direct outcome is that "internal details leak out" with higher frequency.

We can and do automatically map SSZ to JSON in Teku. It's what we're moving to because it makes everything so much simpler. The only requirement is to understand that uint8 and byte have the same data representation but different meanings (and thus different JSON representations). This is trivial to implement in the SSZ library and is transparent to the rest of the code which just uses the same type as the spec (either uint8 or byte) so it's obvious that it does indeed match the spec.

> We have an opportunity to simplify things here, with only theoretical loss of convenience (ie no real numbers will be lost in the process - some bit flags will be more easy to read thanks to bits aligning with hex digits).

Any uint8 will be serialised in a user hostile way. While we currently only use uint8 for participation flags, we don't want to setup an inconsistent standard and hope we never use uint8 somewhere else.

arnetheduck:
> We can and do automatically map SSZ to JSON in Teku. It's what we're moving to because it makes everything so much simpler. 

Though to do so, you need to generate types based on specific instantiations of SSZ (ie the beacon state specification of altair which names some aliases, just like `byte` and `uint8` are aliases and share actual type - aliases being names that carry no semantic meaning, as is commonly understood) - with the simplification that the canonical JSON mapping offers, the only document that influences the mapping is the SSZ serialization specification, and herein lies the simplification: no need to check if `altair.md` introduces some special exceptions via aliases because both SSZ and JSON share the same expressive power and can therefor be round-tripped at the SSZ abstraction level, purely based on type.

ajsutton:
We don't do any specific Altair check, there is no need for such a thing. The SSZ spec defines both uint8 and byte. If there was literally no difference in meaning between the two the spec should be changed to only ever use one - two names for identical things is an unnecessary complication. If it makes you happy feel free to adjust the SSZ spec to define uint8 and byte as two separate types with exactly the same representation in SSZ and avoid the word alias, but a number and a byte is a fundamentally different thing. They should be represented differently in JSON.

JurbinM (approved):

ajsutton:
So @arnetheduck and I met up at DevCon to fight to the death over this but with the altitude found that too exhausting and had to settle for talking about it instead... We're both willing to die on a hill here but fortunately I believe we're on different hills. 

For me, anything we call uint8 has to be serialised the same way we serialise uint64 and other uints which is string quoted decimal in JSON representation.  Anything we call a byte should serialised consistently with other byte type things (eg byte[]) and thus hex encoded.

For @arnetheduck (hopefully I represent this reasonably) it's that if the SSZ spec doesn't make uint8 and byte different then the JSON shouldn't either. And I believe he also wants the participation flags to be byte[] not uint8[] anyway.

So I think the solution here is to just improve the wording in the SSZ spec so that uint8 and byte are independent types which have the same representation in SSZ. The canonical mapping to JSON then can obviously have different representations for uint8 and byte because they're two different things.

IF participation flags were changed to be byte[] instead of uint8[] then Ethereum wouldn't ever actually use uint8 and it could be ignored which I suspect would make the duck particularly happy. :) If I was starting from scratch I would have said participation flags were byte[] as well, personally I'm not sure its worth the backwards compatibility problems to change it now but that hill isn't one I'm willing to die on (and am not really that fussed either way to be honest).

dapplion:
How can this be unblocked? @timbeiko we need protocol support

dogmother79 (approved):

hwwhww:
replaced by #3506

PR #2987, opened 2022-09-08
Explicitly validate public key when processing Deposit

Nashatyrev:
Deposits are the only source of non-validated public keys, so suggesting to add explicit public key validation statement to `process_deposit` function. 

`bls.Verify` function (and its batch variants) is a bit ambiguous in the case if `BLSPubkey` is not valid (either invalid serialization or outside of the curve or outside of the group): whether it should return `false` or throw an error. 

In all other cases (besides `process_deposit`) `bls.Verify` can't receive an invalid `BLSPubkey` since all of them come from the state and are inherently valid. Reasoning that way it could make sense to throw error if invalid pubkey is passed to `bls.Verify` since it would mean kind of fatal implementation error. In that case suggested implicit public key validation makes sense.

### DIFF
--- a/specs/phase0/beacon-chain.md
+++ b/specs/phase0/beacon-chain.md
@@ -1868,6 +1868,10 @@ def process_deposit(state: BeaconState, deposit: Deposit) -> None:
     amount = deposit.data.amount
     validator_pubkeys = [v.pubkey for v in state.validators]
     if pubkey not in validator_pubkeys:
+        # Validate the public key which is not checked by the deposit contract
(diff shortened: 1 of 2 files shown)


dapplion (approved):
Being explicit about this detail makes sense to me :+1:

hwwhww:
It looks fine to me, but I hope to get more reviews 👀 before we modify the phase0 spec.

Tannerrea (approved):

leolara:

PR #2998, opened 2022-09-19
remove withdrawn_epoch

djrtwo:
Handle 4th point in https://github.com/ethereum/consensus-specs/issues/2758 due to the desire to not change the shape of the validators datastructure and merkle tree.

* Remove `fully_withdrawn_epoch`
* Add additional full withdrawal tests for various 0 balance iterations
* Add randomized full withdrawals tests
* Add top-up to fully withdrawn validator test
* [Bonus] add some phase0+ tests for top-ups, showing no effective balance change

The point of `fully_withdrawn_epoch` is to be able to potentially re-use indices after some extended period of non-use (e.g. 3x max best-WS period -- (3x4) 12 months). A zero balance and `withdrawable_epoch` are sufficient to capture this logic for safe reuse. Thus remove `fully_withdrawn_epoch`

In the event that a top-up is made to such a validator, a new full withdrawal is initiated. This requires 1ETH to perform and is rate limited by the `Deposit` queue.

### DIFF
--- a/specs/capella/beacon-chain.md
+++ b/specs/capella/beacon-chain.md
@@ -193,7 +193,6 @@ class Validator(Container):
     activation_epoch: Epoch
     exit_epoch: Epoch
     withdrawable_epoch: Epoch  # When validator can withdraw funds
-    fully_withdrawn_epoch: Epoch  # [New in Capella]
 ```
 
 #### `BeaconBlockBody`
@@ -297,13 +296,14 @@ def has_eth1_withdrawal_credential(validator: Validator) -> bool:
 #### `is_fully_withdrawable_validator`
 
 ```python
-def is_fully_withdrawable_validator(validator: Validator, epoch: Epoch) -> bool:
+def is_fully_withdrawable_validator(validator: Validator, balance: Gwei, epoch: Epoch) -> bool:
     """
     Check if ``validator`` is fully withdrawable.
     """
     return (
         has_eth1_withdrawal_credential(validator)
-        and validator.withdrawable_epoch <= epoch < validator.fully_withdrawn_epoch
+        and validator.withdrawable_epoch <= epoch
+        and balance > 0
     )
 ```
 
@@ -349,11 +349,11 @@ def process_epoch(state: BeaconState) -> None:
 ```python
 def process_full_withdrawals(state: BeaconState) -> None:
     current_epoch = get_current_epoch(state)
-    for index, validator in enumerate(state.validators):
-        if is_fully_withdrawable_validator(validator, current_epoch):
-            # TODO, consider the zero-balance case
-            withdraw_balance(state, ValidatorIndex(index), state.balances[index])
-            validator.fully_withdrawn_epoch = current_epoch
+    for index in range(len(state.validators)):
+        balance = state.balances[index]
+        validator = state.validators[index]
+        if is_fully_withdrawable_validator(validator, balance, current_epoch):
+            withdraw_balance(state, ValidatorIndex(index), balance)
 ```
 
 #### Partial withdrawals
--- a/tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_deposit.py
+++ b/tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_deposit.py
@@ -20,6 +20,9 @@ def test_success_top_up_to_withdrawn_validator(spec, state):
     assert state.balances[validator_index] > 0
     next_epoch_via_block(spec, state)
     assert state.balances[validator_index] == 0
+    assert state.validators[validator_index].effective_balance > 0
+    next_epoch_via_block(spec, state)
+    assert state.validators[validator_index].effective_balance == 0
 
     # Make a top-up balance to validator
     amount = spec.MAX_EFFECTIVE_BALANCE // 4
@@ -27,8 +30,8 @@ def test_success_top_up_to_withdrawn_validator(spec, state):
 
     yield from run_deposit_processing(spec, state, deposit, validator_index)
 
-    state.balances[validator_index] == amount
-    state.validators[validator_index].effective_balance == 0
+    assert state.balances[validator_index] == amount
+    assert state.validators[validator_index].effective_balance == 0
 
     validator = state.validators[validator_index]
     balance = state.balances[validator_index]
--- a/tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_deposit.py
+++ b/tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_deposit.py
@@ -1,0 +1,30 @@
+from eth2spec.test.context import (
+    spec_state_test,
+    with_capella_and_later,
+)
+from eth2spec.test.helpers.state import next_epoch_via_block
+from eth2spec.test.helpers.deposits import (
+    prepare_state_and_deposit,
+    run_deposit_processing,
+)
+from eth2spec.test.helpers.withdrawals import set_validator_fully_withdrawable
+
+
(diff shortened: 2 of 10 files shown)


tests/core/pyspec/eth2spec/test/capella/epoch_processing/test_process_full_withdrawals.py
```diff
@@ -88,6 +125,50 @@ def test_multi_withdrawal(spec, state):
...
+            validator = state.validators[index]
+            # 12.5% unset credentials
+            if rng.randint(0, 7) == 0:
+                validator.withdrawal_credentials = spec.BLS_WITHDRAWAL_PREFIX + validator.withdrawal_credentials[1:]
+            # 12.5% not enough balance
+            if rng.randint(0, 7) == 0:
+                state.balances[index] = 0
+            # 12.5% not close enough epoch
+            if rng.randint(0, 7) == 0:
+                validator.withdrawable_epoch += 1
```
hwwhww, line 148: good randomize 👍

potuz:
I am not fully convinced that reusing the validator index is really that much impactful. It sounds to me as an implementation detail more than a specification item I am concerned by validator registry size. But applications could simply delete the validator entry from whatever storage option they choose and add an `index` field to the validator object that is guaranteed to be increasing. Loops run over the storage object, not over the indexing. Adding a field discards the benefit of this PR about keeping the number of fields a power of 2, which can be used to speed up Merkleization, but that in itself is a hack.

dapplion:
I would like to capitalize on this optimization even if re-usable indexes are not adopted. The validator tree is the most expensive part of the state, not increasing depth by one unless really necessary has a non-negligible impact

potuz:
> I would like to capitalize on this optimization even if re-usable indexes are not adopted. The validator tree is the most expensive part of the state, not increasing depth by one unless really necessary has a non-negligible impact

Depth does not change in the validator registry Merkle tree, the hack is that if you lay out all field roots contiguously you can hash that tree as if it were of depth `log(VALIDATOR_REGISTRY_LIMIT)+3`  and get the right root. However, this can only be useful when hashing the full validator set which is seldomly done. Adding an entry to the validator object screws this hack more because of the zero allocations needed (if the same technique is to be used) than the depth increase. 

The point I want to make is that we can easily keep the validator registry dense, not sparse, **and** not reuse validator indices at the same time without much complexity, and the cost of having to check public key is likely worse than simply adding an scalar index field.

djrtwo:
> I am not fully convinced that reusing the validator index is really that much impactful.

I just mainly want to point out that use of `withdrawable_epoch` and zero-balance give us the same reuse safety *In the event that we want to reuse*. So might as well not add the withdrawn epoch, regardless

specs/capella/beacon-chain.md
```diff
@@ -349,11 +349,11 @@ def process_epoch(state: BeaconState) -> None:
 ```python
 def process_full_withdrawals(state: BeaconState) -> None:
-    for index, validator in enumerate(state.validators):
-        if is_fully_withdrawable_validator(validator, current_epoch):
-            # TODO, consider the zero-balance case
-            withdraw_balance(state, ValidatorIndex(index), state.balances[index])
-            validator.fully_withdrawn_epoch = current_epoch
+    for index in range(len(state.validators)):
```
ralexstokes, line 353: just noting: this happens *after* `process_effective_balance_updates` and it seems like it should to avoid strange edge cases where a balance could be out of sync w/ an effective balance

tests/core/pyspec/eth2spec/test/capella/epoch_processing/test_process_full_withdrawals.py
```diff
@@ -1,39 +1,40 @@
...
 
     if num_expected_withdrawals is not None:
         assert len(to_be_withdrawn_indices) == num_expected_withdrawals
     else:
         num_expected_withdrawals = len(to_be_withdrawn_indices)
 
-    yield from run_epoch_processing_with(spec, state, 'process_full_withdrawals')
+    yield 'pre', state
+    spec.process_full_withdrawals(state)
+    yield 'post', state
```
ralexstokes, line 35: any reason to not use the `run_epoch_processing_with` helper?

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_deposit.py
```diff
@@ -0,0 +1,36 @@
...
+    next_epoch_via_block(spec, state)
+    assert state.balances[validator_index] == 0
+
+    # Make a top-up balance to validator
+    amount = spec.MAX_EFFECTIVE_BALANCE // 4
+    deposit = prepare_state_and_deposit(spec, state, validator_index, amount, signed=True)
+
+    yield from run_deposit_processing(spec, state, deposit, validator_index)
+
+    state.balances[validator_index] == amount
```
ralexstokes, line 30:
```suggestion
    assert state.balances[validator_index] == amount
```
missing this

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_deposit.py
```diff
@@ -0,0 +1,36 @@
...
+    state.validators[validator_index].effective_balance == 0
```
ralexstokes, line 31:
```suggestion
    assert state.validators[validator_index].effective_balance == 0
```

tests/core/pyspec/eth2spec/test/phase0/block_processing/test_process_deposit.py
```diff
@@ -141,13 +141,57 @@ def test_invalid_sig_new_deposit(spec, state):
...
+
+@with_all_phases
+@spec_state_test
+def test_success_top_up__less_effective_balance(spec, state):
+    validator_index = 0
+    amount = spec.MAX_EFFECTIVE_BALANCE // 4
+    deposit = prepare_state_and_deposit(spec, state, validator_index, amount, signed=True)
+
+    initial_balance = spec.MAX_EFFECTIVE_BALANCE - 1000
+    initial_effective_balance = spec.MAX_EFFECTIVE_BALANCE - spec.EFFECTIVE_BALANCE_INCREMENT
```
ralexstokes, line 166: note: this violates some spec invariants but is not material for the thing this test intends to verify so we can let it be

ralexstokes (commented):
looks great!

I like keeping the option to implement reusable validator indices (even though it is in aggregate probably not our best idea) and this PR would keep the validator subtree aligned along nice power of two

I left two suggestions for what I think are missing `assert`s and otherwise just some notes in case anyone is ever reviewing this later

tests/core/pyspec/eth2spec/test/phase0/block_processing/test_process_deposit.py
djrtwo: This does not actually do so

You can have an epoch in which  your balance was initially `spec.MAX_EFFECTIVE_BALANCE - spec.EFFECTIVE_BALANCE_INCREMENT - 1000` and your initial effective balance is as above. Then you could receiv a 1 ETH deposit, pushing your balance (and effective) to the balance above as your initial conditional

tests/core/pyspec/eth2spec/test/capella/epoch_processing/test_process_full_withdrawals.py
djrtwo: yes this pattern is useful when you need to capture information right before the function is called.

In this case, balance updates can/will occur during those interim epoch processings. So in some cases this changes the `to_be_withdrawn` indices in some test cases

PR #2982, opened 2022-08-23
Add optimistic sync tests

hwwhww:
### Issue
Address https://github.com/ethereum/consensus-spec-tests/issues/28

### Description

Test format: #2965 

#### Example
`test_from_syncing_to_invalid` is the basic test from https://github.com/ethereum/consensus-spec-tests/issues/28 issue description:
```
B0 <- B1 <- B2 <- B3 
  \
    <- B1' <- B2' <- B3' <- B4'

1. Client imports [B0 ... B3], payload_status: {status: "VALID", ...} for each block in this branch
2. Client reorgs to and optimistically imports [B1' ... B3'], payload_status: {status: "SYNCING", ...} for each of this blocks
3. Client imports B4' with payload_status: {status: "INVALID", latestValidHash: B0.body.execution_payload.block_hash, ...}
4. head == B3
```

##### `minimal/bellatrix/sync/optimistic/pyspec_tests/from_syncing_to_invalid/steps.yaml` (20220908 fixed)
```yaml
- {tick: 7728}
- block_hash: '0xb9a07089b65517ade3874fc891cc6206dfe5e753ac3bfb8dbf2cf0fe514b2a98'
  payload_status: {status: VALID, latest_valid_hash: '0xb9a07089b65517ade3874fc891cc6206dfe5e753ac3bfb8dbf2cf0fe514b2a98',
    validation_error: null}
- {block: block_0x3d1241fb9fa6d442d45e1f2153abdb1158cfe90f273978b6212b80d15053d1d8}
- checks:
    head: {slot: 9, root: '0x30b495d00fb542d2c513a464bba858295b884ab2451953e5876804a049afc2ef'}
- block_hash: '0xcfa86982c4ca3a0c429d223ee3000f64305e7bbb6ed57f12748a3ca1bb73e2ff'
  payload_status: {status: VALID, latest_valid_hash: '0xcfa86982c4ca3a0c429d223ee3000f64305e7bbb6ed57f12748a3ca1bb73e2ff',
    validation_error: null}
- {block: block_0x2a06b3082961d053d0b59c1fd232b298acc0353cdd58480c0d2dab16403f85fb}
- checks:
    head: {slot: 10, root: '0x1486942d4953830f9bbf43141838db41686edc43ca870d5b22c500e5e0c0898b'}
- block_hash: '0xb051302363f1bdb6ef94a54cee5958539aedda44229baaf1dae3e8db08baa5de'
[...]

PR #3001, opened 2022-09-20
Update BLS test format: output `null` for invalid case

hwwhww:
In #2982, we [updated our test vector YAML representer to output `null` when the value is Python `None`](https://github.com/ethereum/consensus-specs/blob/9ec97badf3ee924aae0ee92cf6c957fbe3b7ef4b/tests/core/pyspec/eth2spec/gen_helpers/gen_base/gen_runner.py#L100-L103).

Before that PR, it was:
```yaml
output:
```

After that PR, it becomes:
```yaml
output: null
```

IMHO it's better to use only one rule for both cases. This PR changes the BLS test format to align with sync test format. `null` seems clearer than empty.

p.s. `null` is used in engine APIs and that was why I changed our YAML representer in #2982

/cc @asanso for updating `bls12-381-tests` test suite.

### DIFF
--- a/tests/formats/bls/aggregate.md
+++ b/tests/formats/bls/aggregate.md
@@ -8,11 +8,11 @@ The test data is declared in a `data.yaml` file:
 
 ```yaml
 input: List[BLS Signature] -- list of input BLS signatures
-output: BLS Signature -- expected output, single BLS signature or empty.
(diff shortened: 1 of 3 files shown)


djrtwo (approved):

asanso:
lgtm

PR #3004, opened 2022-09-21
EIP-4844: Update ExecutionPayloadHeader

Inphi:
Per the eip-4844 fee market spec, add the `excess_blobs` field to the `ExecutionPayloadHeader`.

Pairs the change in https://github.com/ethereum/EIPs/pull/5353

### DIFF
--- a/specs/eip4844/beacon-chain.md
+++ b/specs/eip4844/beacon-chain.md
@@ -19,6 +19,8 @@
 - [Containers](#containers)
   - [Extended containers](#extended-containers)
     - [`BeaconBlockBody`](#beaconblockbody)
+    - [`ExecutionPayload`](#executionpayload)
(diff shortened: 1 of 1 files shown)


djrtwo (commented):
Maybe we should update it to this PR and merge in tandem:
https://github.com/ethereum/EIPs/pull/5707

Inphi:
> Maybe we should update it to this PR and merge in tandem: [ethereum/EIPs#5707](https://github.com/ethereum/EIPs/pull/5707)

This PR addresses the initial change to the fee market specification - https://github.com/ethereum/EIPs/pull/5353. However, https://github.com/ethereum/EIPs/pull/5707 doesn't require any changes to CL. Although https://github.com/ethereum/EIPs/pull/5707 does have a "potential todo" that could affect CL; `make excess blobs calculation slot-based instead of block-based`.

PR #3006, opened 2022-09-21
Polynomials in bit reversal permutation

dankrad:
We should consider moving the interpretation of Lagrange polynomials to being in bit reversal permutation. What this means is that the points don't come in the order `f(w^0), f(w^1), f(w^2)` and so on, but instead in the bit-reversal permutation. As an example, if `w` is an eighth root of unity, then the order would be `f(w^0), f(w^4), f(w^2), f(w^6), f(w^1), f(w^3), f(w^5), f(w^7)`.

The rationale for this change is that reverse bit order means that subgroups and cosets are contiguous ranges. For example, the subgroup of order 4 generated by `w^2` has two cosets: `w^0, w^4, w^2, w^6` and `w^1, w^3, w^5, w^7`. In the bit reversal permutation, these are the first four and second four elements of the sequence of roots of unity. So once we introduce data availability sampling, we want everything to be in this order: We want to be able to easily generate proofs for subranges for the data availability samples, and our technique for generating proofs (https://github.com/khovratovich/Kate/blob/master/Kate_amortized.pdf) requires these ranges to be cosets.


specs/eip4844/polynomial-commitments.md
```diff
@@ -64,6 +68,45 @@ but reusing the `mainnet` settings in public networks is a critical security req
...
+    """
+    assert is_power_of_two(order)
+    # Convert n to binary with the same number of bits as "order" - 1, then reverse its bit order
+    return int(('{:0' + str(order.bit_length() - 1) + 'b}').format(n)[::-1], 2)
+```
+
+#### `bit_reversal_permutation`
+
+```python
+def bit_reversal_permutation(l: Sequence[int]) -> Sequence[int]:
```
asn-d6, line 103: Hmm. This function signature is not entirely true for `bit_reversal_permutation(KZG_SETUP_LAGRANGE)` since `KZG_SETUP_LAGRANGE` is an array of `bytes48` and not `int`.

asn-d6:
Superseded by https://github.com/ethereum/consensus-specs/pull/3011 . Closing this one!

PR #3007, opened 2022-09-21
Rename withdraw section to withdraw_balance

jtraglia:
Minor fix, this section should have the same name as the function.


hwwhww (approved):
thanks @jtraglia!

Simpdaddyjosh (approved):

PR #3010, opened 2022-09-22
Add partial withdrawal balance tests

jtraglia:
Noticed that there were no tests for validator balances that do not meet the partially withdrawable requirements defined in `is_partially_withdrawable_validator`. This adds three tests for that. Also fixes a few minor things.


djrtwo (approved):
thank you!!

PR #3011, opened 2022-09-23
4844 Reverse bit ordering (#3006 rebased + func signature fixed)

asn-d6:
4844 Reverse bit ordering (#3006 rebased + func signature fixed)


dankrad:
Isn't `Sequence[Any]` a possible type signature that would allow us to avoid needing to keep two copies of `bit_reversal_permutation`?

Where-2:
All scammers.   All misrepresenting yourselves.  Sick of being ripped off.

hwwhww:
I added the `T = TypeVar('T')` solution to this PR.

specs/eip4844/polynomial-commitments.md
```diff
@@ -98,22 +97,14 @@ def reverse_bits(n: int, order: int) -> int:
...
 
 ```python
-def bit_reversal_permutation_scalars(l: Sequence[BLSFieldElement]) -> Sequence[BLSFieldElement]:
+def bit_reversal_permutation(l: Sequence[T]) -> Sequence[T]:
     """
-    Apply the bit-reversal permutation to a list. This operation is idempotent.
-    """
-    return [l[reverse_bits(i, len(l))] for i in range(len(l))]
-```
+    Return a copy with bit-reversed permutation. This operation is idempotent.
```
hwwhww, line 105: p.s. I changed this docstring because the old version sounds like this function would change the input with side effects.

dankrad:
lgtm

dankrad (approved):

PR #3013, opened 2022-09-23
Include ResourceUnavailable clause when LightClientUpdate cannot be returned

dadepo:


etan-status:
Hmm, not sure how such an error helps, vs, simply omitting missing data. The current spec is in-line with the beaconBlocksByRange endpoint.

Jiggnubs (commented):
Henceforth the poor soul don't quit.  As the weakest link are great facts from start to finish when properly included.

dapplion:
The relevant line in beaconBlocksByRange

> Peers that are unable to reply to block requests within the MIN_EPOCHS_FOR_BLOCK_REQUESTS epoch range SHOULD respond with error code 3: ResourceUnavailable. Such peers that are unable to successfully reply to this range of requests MAY get descored or disconnected at any time.

In this context, an example would be:
- If you checkpoint sync and you haven't backfilled yet that request you may reply with ResourceUnavailable

However LightClientUpdate is different to beaconBlocksByRange, where:
- beaconBlocksByRange: Chain is likely to have missed slots, which do not stop from following the chain
- LightClientUpdate: Must be strictly contiguous or you won't be able to follow the chain

etan-status:
While I'm not principally opposed to having an error condition for potentially useless responses, the specific conditions should be explicitly mentioned:

- What if a node is limiting responses to count < requested count? Should this partial response still be allowed?
- What if the start of the section is missing, e.g., node only has data from 200 onward but start is at 150? Should it still be allowed to provide response starting from a later period than `{start_period}`?
    - How about `{start_period}` < altair fork epoch? e.g., right now, can do a request from 100-200 on goerli and the response will start at 143, because for 0-142 no data exists (pre-altair)
- What about gaps in-between periods? e.g., if a node is missing a certain period in the middle?
    - What in the unlikely case that there really is no sync participation for a day?

Currently, all the decisions are left up to the client, on how they want to approach missing periods. Maybe if they see a gap they still want to cache future data for later validation / maybe they want to proceed until the gap and then toss away the rest / maybe they want to toss the entire response.

With an approach that replaces partial results with errors, certain client behaviour is no longer possible (or would require repeatedly fetching periods with lower `{count}` to identify the missing one. The exact conditions when it happens should be defined very explicitly to avoid unintentional breaking.

"unable to reply" is quite vague (also for the beaconBlocksByRange / beaconBlocksByRoot endpoint where I haven't seen this response in practice / it seems to be more of a static response to throw when a node is simply not collecting any historic data). If this is just about documenting that case, where a node operator explicitly opts out of historic data storage, I'm fine with using the `ResourceUnavailable`. I'm only concerned about the case where a partial response gets replaced with an error.

etan-status:
Also, in case of in-between gaps, an alternative response could just include the continuous response section up to the gap, and then stop responding. This way, the node would appear as any other one that simply hasn't synced yet past the missing period. So, maybe that's a more appropriate resolution, over replacing the entire response with an error?

dapplion:
Those are great points, I agree the specification should be flexible given all the edge cases:
- I agree the node should not reply with resource unavailable for partial responses
- Omitting data due to gaps introduces more edge cases. The consumer of data should be the one deciding what to do

So I would not include this PR's suggestion nor a sentence like
https://github.com/ethereum/consensus-specs/blob/6b9e65fab6c2cf1f8973b3dee29173ed6cb6130b/specs/phase0/p2p-interface.md?plain=1#L760-L763

dadepo:
Closing this PR based off the ensued discussion.

PR #3014, opened 2022-09-26
deprecate phase 0-only versions of beacon_blocks_by_{range,root}

tersec:
There haven't been any phase0 blocks in mainnet or any of the still-supported testnets in [`MIN_EPOCHS_FOR_BLOCK_REQUESTS`](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#configuration) regardless, so the spec already technically supports not serving said blocks.

https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/p2p-interface.md#beaconblocksbyrange-v2 and https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/p2p-interface.md#beaconblocksbyroot-v2 work across all supported forks, including phase 0, and any CL client supporting Altair or newer already has to support those messages, so this maintains req/resp compatibility for phase 0 blocks.

### DIFF
--- a/specs/phase0/p2p-interface.md
+++ b/specs/phase0/p2p-interface.md
@@ -745,6 +745,8 @@ For example, if slot 4 were empty in the previous example, the returned array wo
 
 `step` is deprecated and must be set to 1. Clients may respond with a single block if a larger step is returned during the deprecation transition period.
 
+`/eth2/beacon_chain/req/beacon_blocks_by_range/1/` is deprecated. Clients MAY respond with an empty list during the deprecation transition period.
+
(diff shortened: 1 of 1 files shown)


djrtwo (approved):
looks reasonable to me.

will leave this open for a week for comment

nisdas (approved):
LGTM, should we also apply the same principle to v1 metadata requests ? They are deprecated and all clients should support `v2` from altair onwards.

etan-status (approved):

dapplion (approved):

arnetheduck (approved):

rolfyone (approved):
LGTM

PR #3016, opened 2022-09-27
Rework the testing KZG trusted setups presets

hwwhww:
1. Add testing KZG trusted setups generator tool
	1. Add `make gen_kzg_setups` command to generate the testing KZG trusted setup JSON files:
		1. `presets/mainnet/trusted_setups/testing_trusted_setups.json`
		2. `presets/minimal/trusted_setups/testing_trusted_setups.json`
2. Load the files and set `KZG_SETUP_*` presets during building pyspec
3. Update `compute_kzg_proof` and `evaluate_polynomial_in_evaluation_form`: add `int` casting to avoid underflow

- [x] Question: once we have the real mainnet trusted setups, it should be in `presets/mainnet/eip4844.yaml`. The reason why I store them in other JSON files is that I feel it’s weird to put the testing presets into it and poison the git log. Also, the KZG ceremony may use JSON schema. Any opposite opinions?

The reasons for having `roots_of_unity` field in this file:
1. `ROOTS_OF_UNITY` depends on the value of `FIELD_ELEMENTS_PER_BLOB` so it is easier if we generate it with trusted setups (which also depend on `FIELD_ELEMENTS_PER_BLOB`).
2. `ROOTS_OF_UNITY` is a preset and so it has to be in one of presets files. If there is some objection that it should be in other files, I will move it to its own file.


asn-d6:
Hello. Let's get https://github.com/ethereum/consensus-specs/pull/3038 in first because it revamps the cryptography subsystem, and then we can do this PR!

hwwhww:
Hey @asn-d6 I rebased it after #3038.

asn-d6 (approved):
Changes look sensible!

PR #3017, opened 2022-09-27
Fix incorrect comment

xrchz:
bit-reversal permutation is not idempotent


dankrad (approved):
It's indeed an involution and not idempotent, my bad.

hwwhww (approved):

PR #3018, opened 2022-09-28
Bump dep packages version and fix lint issues

hwwhww:
1. Bump dependencies
     - Update most dependencies to the latest version
     - Remove `dataclasses` dep because it's built-in in Python 3.7+
     - Somehow, `marko==1.2.2` is not backward compatible so I didn't update it
2. Fix the lint issues found by the newer linter
    - **E741 ambiguous variable name 'l'** in `bit_reversal_permutation`:
        - I renamed `l` to `sequence`. /cc @asn-d6 
    - **F541 f-string is missing placeholders**:
        - I cleaned up the strings
    - `from_bytes` has incompatible type "str"; expected "literal['little', 'big']":
       - This is annoying. [The solution is adding `Final` type hint](https://stackoverflow.com/questions/63891789/can-mypy-track-string-literals). I made some changes in `setup.py`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ lint: pyspec
 
 lint_generators: pyspec
 	. venv/bin/activate; cd $(TEST_GENERATORS_DIR); \
-	flake8  --config $(LINTER_CONFIG_FILE)
+	flake8 --config $(LINTER_CONFIG_FILE)
 
 compile_deposit_contract:
 	@cd $(SOLIDITY_DEPOSIT_CONTRACT_DIR)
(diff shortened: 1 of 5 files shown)


asn-d6:
`I renamed l to sequence. /cc @asn-d6` change LGTM! (Haven't looked at the non-4844 changes though)

asn-d6 (approved):
`I renamed l to sequence. /cc @asn-d6` change LGTM! (Haven't looked at the non-4844 changes though)

ralexstokes (approved):
the changes all look good and if the CI passes them I'll assume all of the newer deps don't break anything

+1 on merge

Joumax11:
Gracias por la información y todo es por el apoyo que me brindan aquí ! Saludos y en lo que pueda colaborar con gusto

PR #3022, opened 2022-10-01
Change Blobs type definition to be Sequence[Blob]

kevaundray:
In some places, `blobs` was being defined a sequence of field elements (Vector), instead of a sequence of sequence of field elements (Matrix).

This PR uses the `Blob` type to distinguish these cases


kevaundray:
Closing as this is a duplicate of #3021

PR #3025, opened 2022-10-03
Rename `from_bls_pubkey` to `from_bls_withdrawal_pubkey`

terencechain:
`from_bls_pubkey` could be confused with the bls signing key. I think it's better to be explicit and call it `from_bls_withdrawal_pubkey` which aligns with the phase 0 validator spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/validator.md#bls_withdrawal_prefix

### DIFF
--- a/specs/capella/beacon-chain.md
+++ b/specs/capella/beacon-chain.md
@@ -120,7 +120,7 @@ class Withdrawal(Container):
 ```python
 class BLSToExecutionChange(Container):
     validator_index: ValidatorIndex
-    from_bls_pubkey: BLSPubkey
+    from_bls_withdrawal_pubkey: BLSPubkey
     to_execution_address: ExecutionAddress
 ```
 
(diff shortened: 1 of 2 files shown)


mcdee:
If we are tweaking things, do we really need the `from_` and `to_` prefixes given that this should be obvious given the class name?

terencechain:
> If we are tweaking things, do we really need the `from_` and `to_` prefixes given that this should be obvious given the class name?

Happy to remove the prefix. I mainly wanted to differentiate 
signing with withdrawal here

ralexstokes:
I actually think the `from`, `to` are more valuable than the additional `withdrawal` symbol

I think from the context in which these types will be used it should be clear what is what and this is ultimately just a question of style preference

terencechain:
Don't feel strongly about this, but we actually had internal confusion over this, maybe we are just the only ones

hwwhww:
Hey @terencechain 
It seems no strong 👍 signal for this PR. Can we close it or would you like to bring it up to the CL call today?

dapplion:
I think having better name clarity is good. Since external tooling will likely be producing this message too, the clarification helps.

terencechain:
I don't have a strong preference but like @dapplion said, external tooling might find this helpful

rolfyone:
I'm fairly non-fussed on this one, if there's a general feeling that it will make this easier to use, happy to take that feedback and run with it... I won't feel like we missed something if it's not changed though...

Jiggnubs:
Just see creating, don't mind me world! Highest ones Null.

PR #3029, opened 2022-10-09
Remove duplicated definition of `Validator` in `capella` specs

ralexstokes:
a cleanup, following the changes in #2998 we no longer have a new validator definition so the `capella` spec can drop this


hwwhww (approved):
nice catch :o

PR #3030, opened 2022-10-10
Removes SSZ and updates the crypto API

kevaundray:
Here is a demonstration of the API if more of the algorithms were pushed into the cryptography section. Since it is an abstraction leak if the cryptography contained ssz, it was modified, though the modification is standard and not out of the norm.


sweetpea1328 (approved):

kevaundray:
This can be closed as the changes have been put into #3038

PR #3032, opened 2022-10-12
Capella processing

rolfyone:
Partially to initiate a discussion.

Basically use a single pipeline for withdrawal processing to simplify the flow and reduce the need for full validator scans during epoch processing.

This should make the feature much more testable amongst other benefits, and makes withdrawal of partial or full amounts an equal priority based on your position in the validator list.

The timing remains approximately equal to previous, except there may have previously been 4 exits per epoch to add that are now just done during sweeping, where previously they were done as a priority.


ajsutton:
For the record, the better testability of this approach is because the spec suggests client's optimise by tracking validators which become eligible for withdrawal to avoid having to scan the whole validator set. But then any reference tests which focus specifically on a spec method to do the full iteration and identify newly withdrawable validators wouldn't test that optimisation. To test properly you'd need to do the full epoch processing to ensure the caching code was actually used.

potuz:
An alternative approach would be to process withdrawals while processing effective balance changes on the same loop, but assessing the safety of that may be tricky

djrtwo:
This is based on an old commit in `dev`. I reworked it to be on recent dev (which had a number of capella cleanups) in #3042

Where-2:
Ok appreciate your help.  Strange anomalies occur constantly

PR #3034, opened 2022-10-13
Allow honest validators to reorg late blocks

michaelsproul:
Specify a function in fork choice `get_proposer_head` which returns the parent block upon which a proposer may build while possibly reorging the canonical head (if it was broadcast late and lacks sufficient attestation weight).

The mechanism is designed with safeguards built in so that it does not cause liveness failures:

- Only single-slot reorgs are attempted: block `N + 2` may build on block `N` if block `N + 1` arrived late.
- Finalization must be occurring with the expected frequency determined by `REORG_MAX_EPOCHS_SINCE_FINALIZATION`. The default is to require optimal finalization (every 2 epochs).
- `REORG_HEAD_WEIGHT_THRESHOLD` is set to 20% (half of the proposer boost) so that an attacker must control up to `PROPOSER_BOOST_SCORE - REORG_WEIGHT_THRESHOLD` = 20% of attester weight in order to perform an ex-ante reorg. The attack works by releasing attestations for the block at slot `N + 1` around the same time as the honest proposer publishes the attempted reorg block at `N + 2`. The late attestations tip the weight of the `N + 1` block from `< REORG_WEIGHT_THRESHOLD` to `> PROPOSER_BOOST_SCORE`, meaning that block `N + 1` will be preferred to block `N + 2`.
- `REORG_PARENT_WEIGHT_THRESHOLD` is set to 160% so that an attacker that is hoarding attestations cannot cause the attempted re-org to fail. For details of this calculation see: https://ethresear.ch/t/selfish-mining-in-pos/15551.

For background on the development of this idea, see: https://github.com/sigp/lighthouse/pull/2860.

## Pros

- Completely optional for clients and node operators, can be switched on and off via a flag. Low risk.
- Can be rolled out almost immediately unlike other late block mitigations which are still in R&D.

## Cons

- Considerable complexity, particularly regarding the overriding of fork choice updates for the canonical head.
- Obsoleted by `(block, slot)` fork choice. If a viable implementation of `(block, slot)` fork choice emerges then reorging late blocks will become the default behaviour.
- Relies on proposer boost, which may be phased out in favour of other fork choice changes (e.g. view-merge).

Thanks to @potuz @mkalinin @terencechain @fradamt @casparschwa @djrtwo for feedback on this idea.


specs/phase0/fork-choice.md
```diff
@@ -275,6 +289,55 @@ def get_head(store: Store) -> Root:
...
+    participation_multiplier = 2
+    parent_weight = get_latest_attesting_balance(store, parent_root)
+    participation_threshold = calculate_committee_fraction(justified_state, PARTICIPATION_THRESHOLD)
+    participation_ok = parent_weight >= participation_multiplier * participation_threshold
+
+    # Only re-org the head block if it arrived later than the attestation deadline.
+    head_late = store.block_timeliness.get(head_root) is False
+
+    # Check that the head has few enough votes to be overpowered by our proposer boost.
+    assert store.proposer_boost_root == Root()
```
realbigsean, line 328: Doesn't this check serve the same purpose as `is_timely`? We could just use only this check instead of storing `is_timely` in the fork choice store.

realbigsean:
Since this is optional, it might be nice to separate it from other parts of the spec a bit more, similar to optimistic sync. We could contain the complexity a bit, especially if it might be phased out in the future.

specs/phase0/fork-choice.md
realbigsean: From reading your implementation, I realized this isn't the same because this will be called in the proposal slot, so proposer boost should be reset by this point

michaelsproul: yeah I intended this assert mainly as a sanity check, perhaps we could add a check on `get_current_slot(store) == slot` as well

specs/bellatrix/fork-choice.md
```diff
@@ -78,6 +79,100 @@ As per EIP-3675, before a post-transition block is finalized, `notify_forkchoice
...
+
+If proposer boost re-orgs are implemented and enabled (see `get_proposer_head`) then additional care
+must be taken to ensure that the proposer is able to build an execution payload.
+
+If a beacon node knows it will propose the next block then it SHOULD NOT call
+`notify_forkchoice_updated` if it detects the current head to be weak and potentially capable of
+being re-orged. Complete information for evaluating `get_proposer_head` _will not_ be available
+immediately after the receipt of a new block, so an approximation of those conditions should be
+used when deciding whether to send or suppress a fork choice notification. The exact conditions
+used may be implementation-specific, a suggested implementation is below.
```
terencechain, line 92: Thanks for this! It makes sense
It may not be clear to readers at first, but I don't think there's a good reason to call `should_override_forkchoice_update` if the block `N` is timely. (Assume you are a proposer for `N+1`). If the block is timely, then you should just call `notify_forkchoice_updated`

specs/bellatrix/fork-choice.md
```diff
@@ -78,6 +79,100 @@ As per EIP-3675, before a post-transition block is finalized, `notify_forkchoice
...
+    # Only re-org the head_block block if it arrived later than the attestation deadline.
+    head_late = store.block_timeliness.get(head_root) is False
+
+    # Only suppress the fork choice update if we are confident that we will propose the next block.
+    parent_state_advanced = store.block_states[parent_root]
+    process_slots(parent_state_advanced, head_block.slot + 1)
+    proposer_index = get_beacon_proposer_index(parent_state_advanced)
+    proposing_reorg_slot = validator_is_connected(proposer_index)
+
+    return all([single_slot_reorg, shuffling_stable, ffg_competitive, participation_ok,
```
terencechain, line 162: In the current Prysm proposal, we are missing `shuffling_stable`, `ffg_competitive`, `participation_ok` and part of single slot reorg conditions. I'll circle back with @potuz

specs/bellatrix/fork-choice.md
```diff
@@ -78,6 +79,100 @@ As per EIP-3675, before a post-transition block is finalized, `notify_forkchoice
...
+        head_weight = get_latest_attesting_balance(store, head_root)
+        reorg_threshold = calculate_committee_fraction(justified_state, REORG_WEIGHT_THRESHOLD)
+        head_weak = head_weight < reorg_threshold
+    else:
+        head_weak = True
+
+    # Only re-org the head_block block if it arrived later than the attestation deadline.
+    head_late = store.block_timeliness.get(head_root) is False
+
+    # Only suppress the fork choice update if we are confident that we will propose the next block.
```
terencechain, line 156: My naive intuition is this should be the very first check and just exit early if no one is proposing the next block : )

michaelsproul: I think the optimal implementation returns early as soon as any condition is false, and does the checks in order from cheapest to most expensive (which might depend on the client architecture + cache structure). For the spec I don't mind trying to roughly match it to the clients, but I think there'll necessarily be a degree of divergence, particularly if we leave out the complexity of the early returns (which I think we should).

Maybe an intuitive order is:

1. `block_late`
2. `proposing_next_slot`
3. `single_slot_reorg`, `shuffling_stable`, `ffg_competitive`, `participation_ok`, `head_weak`

And I can also add a note suggesting that clients implement the checks in an order that makes sense for them, with appropriate early returns?

specs/bellatrix/fork-choice.md
michaelsproul: Yeah I think you're right that should be the first check. I made a related suggestion about the ordering here: https://github.com/ethereum/consensus-specs/pull/3034#discussion_r999983626

specs/bellatrix/fork-choice.md
potuz: We do have the latter two, but we don't have the shuffling safety check.

specs/bellatrix/fork-choice.md
michaelsproul: Updated

fradamt:
What is the circuit-breaker mechanism for when latency is > 4s? Is the intention that checking "single_slot_reorg" and "participation_ok" should act as one, because, when latency is > 4s and blocks do not get attestations in their slot, the parent block would either be too old or not have enough weight?

If that's the case, I think this should be clarified. It might also be nice to have an explicit circuit-breaker that's easier to reason about, and gives clear liveness guarantees (e.g. explicitly turn off the reorg functionality whenever too many proposals are being missed)

michaelsproul:
@fradamt Yeah the single slot & participation conditions are meant to be the circuit-breaker for asynchrony (although I think the single slot condition alone is sufficient). I think the outcome is that with message delay $\text{4s} < \Delta < \text{16s}$ then every second block will be re-orged and we'll get 50% of blocks on average. Once the delay exceeds 16s then blocks will be forced to build on a parent from at least 2 slots prior and we'll get <50% of blocks on average, without the proposer re-org rules coming into effect at all. In other words the only change in behaviour is for the delay window $\text{4s} < \Delta < \text{16s}$.

> It might also be nice to have an explicit circuit-breaker that's easier to reason about, and gives clear liveness guarantees (e.g. explicitly turn off the reorg functionality whenever too many proposals are being missed)

Are you thinking of something like the builder API [circuit-breaker conditions](https://lighthouse-book.sigmaprime.io/builders.html#circuit-breaker-conditions) about number of missed blocks in the last N blocks and number of epochs since finalization? We could _add_ number-of-skipped-blocks check (while keeping `single_slot_reorg`), and _replace_ the current participation check with the the epochs-since-finalization check? The main reason I'd like to keep the `single_slot_reorg` check is that it simplifies the implementation greatly in terms of calculating and caching things.

specs/bellatrix/fork-choice.md
```diff
@@ -76,6 +77,102 @@ As per EIP-3675, before a post-transition block is finalized, `notify_forkchoice
...
+    # FFG information of the new head_block will be competitive with the current head.
+    # FIXME: needs more justification/finalization information in the store
+    # ffg_competitive =
+    #    store.realized_justifications[parent_root] ==
+    #        store.realized_justifications[head_root] and
+    #    store.realized_finalizations[parent_root] ==
+    #        store.realized_finalization[head_root]
+    ffg_competitive = True
+
+    # Check the head weight only if the attestations from the head slot have already been applied.
```
potuz, line 147:
```suggestion
    # Check the head weight only if the attestations from the current slot have already been applied.
```
This function is phrased in a way that it is agnostic to whether we are calling it on the slot we are about to propose or during the previous slot. However, what worries me is the situation where `current_slot = N`, we are supposed to propose at `N+1`, `head_slot = N-1` and we call during slot `N`. In this case, suppose the block `N` is  based on `N-2`, forking off `N-1`, it is a timely block, and we haven't yet counted attestations because we are calling this function during `N`. What will happen is that our head will still be `N-1` and every check here will pass. It may well be that later, at the time we need to propose, during N+1, this timely block at N  would become our head, but this function here may return true.  

Now this is not a problem per-se in this specification since later on, when calling `get_proposer_head` we will get a misprediction because we will not have passed `single_slot_reorg`.  Anyway, I feel it's better to change this condition here from checking that the attestations for the head block to have been counted to "changing that the attestations for the current slot have been counted if we are calling late in the slot". I do recognize that this is difficult to specify, but perhaps it's better to not be agnostic as to the timing in which this function is being called. 

As a side in this long comment: if we are in the situation above, and we call this function during N, then the `assert store.proposer_boost_root == Root()` will fail in this function.

specs/phase0/fork-choice.md
```diff
@@ -173,6 +180,17 @@ def get_ancestor(store: Store, root: Root, slot: Slot) -> Root:
...
 ```
 
+#### `calculate_committee_fraction`
+
+```python
+def calculate_committee_fraction(state: BeaconState, committee_percent: uint64) -> Gwei:
+    num_validators = len(get_active_validator_indices(state, get_current_epoch(state)))
+    avg_balance = get_total_active_balance(state) // num_validators
+    committee_size = num_validators // SLOTS_PER_EPOCH
+    committee_weight = committee_size * avg_balance
```
potuz, line 190:
```suggestion
       committee_weight = get_total_active_balance(state) // SLOTS_PER_EPOCH
```

specs/bellatrix/fork-choice.md
michaelsproul: If the head is at `N - 1` and we are proposing at `N + 1` then this function won't return true. This function infers the `proposal_slot` _from_ the head slot, so it will check whether we are proposing at slot _N_:

https://github.com/ethereum/consensus-specs/blob/56f20f7214ca17ef87cbea416181f9b05f944b1d/specs/bellatrix/fork-choice.md?plain=1#L111

i.e. if we are actually proposing at `N+1` and some other block for `N` already exists, the `proposing_reorg_slot` check will be `False`.

Therefore I think the spec already minimises the chance of a misprediction. It's a really nice test case though, I'll make sure it gets a spec test when we add them (I'll also add a test to Lighthouse before then).

> Anyway, I feel it's better to change this condition here from checking that the attestations for the head block to have been counted to "changing that the attestations for the current slot have been counted if we are calling late in the slot".

I'm open to making this change, but would like to clarify something about how you envisage Prysm's implementation applying attestations at 10 or 11 secs into the slot will work: are you going to avoid updating the fork choice store's current slot "early"? I.e. if you are 10s into slot `N` will you apply the attestations for slot `N` and leave `get_current_slot(store) == N`? The way everything is written at the moment maintains the invariant that only attestations from `get_current_slot(store) - 1` or earlier are applied, supporting Lighthouse's implementation which advances the fork choice slot early at 11.5s into slot `N` and pre-emptively increments the fork choice slot to `N + 1`.

> As a side in this long comment: if we are in the situation above, and we call this function during N, then the assert store.proposer_boost_root == Root() will fail in this function.

Nice catch on the `assert`! I think we need to change it so we return `False` rather than erroring, or assert that the proposer boost is for a block other than `head_block`:

```python
assert store.proposer_root_root != head_root
```

specs/phase0/fork-choice.md
michaelsproul: This function is also used to calculate the proposer boost, so this would be changing the semantics slightly.

I assume there's a reason why we use the `avg_balance` when calculating the boost currently? Maybe @adiasg or @fradamt know?

specs/bellatrix/fork-choice.md
potuz: > i.e. if we are actually proposing at N+1 and some other block for N already exists, the `proposing_reorg_slot` check will be `False`.
Indeed I mixed the previous check, but now I'm worried about something else: `validator_is_connected` can return true for a large number of validators in the case of a pool. In the above scenario, a pool proposing at N and N+1 will actually pass with the wrong validator the check for `proposing_reorg_slot`. 

>  are you going to avoid updating the fork choice store's current slot "early"? I.e. if you are 10s into slot N will you apply the attestations for slot N and leave `get_current_slot(store) == N` ?
Indeed that's my plan, I'm open to change this to make it closer to LightHouse, but it seems very hard on our end cause we track the wallclock on forkchoice and tricking it to change the slot involves bad hacks like changing the genesis time.

specs/phase0/fork-choice.md
potuz: I didn't understand this comment? this function only uses the product which cancels `num_validators`, that's my only innocent observation here.

michaelsproul: There's integer division in play which means they may not be entirely equivalent. It's probably a very minor difference but could cause a view split if we change it not on a hard fork boundary.

specs/phase0/fork-choice.md
Arezoomohammadalizadeh: Verygood👌🙏🌹

specs/bellatrix/fork-choice.md
Arezoomohammadalizadeh: Verygood👌🙏🌹

Arezoomohammadalizadeh: 👌🌹🙏

specs/bellatrix/fork-choice.md
Arezoomohammadalizadeh: Verygood👌🌹🙏

Arezoomohammadalizadeh: Verygood👌🌹🙏

specs/bellatrix/fork-choice.md
Arezoomohammadalizadeh: Verygood👌🌹🙏

specs/bellatrix/fork-choice.md
```diff
@@ -76,6 +77,102 @@ As per EIP-3675, before a post-transition block is finalized, `notify_forkchoice
...
+    return all([head_late, proposing_reorg_slot, finalization_ok, single_slot_reorg,
+                shuffling_stable, ffg_competitive, head_weak])
+```
+
+> Note that the ordering of conditions is a suggestion only. Implementations are free to
+optimize by re-ordering the conditions from least to most expensive and by returning early if
+any of the early conditions are `False`.
+
+In case `should_override_forkchoice_update` returns `True`, a node SHOULD instead call
+`notify_forkchoice_updated` with the parameters of the parent block. Payload attributes should
```
potuz, line 167: Here the parent block should be advanced to this slot, this should be noted in the PR since the expected withdrawals may have changed from the previous slot to this slot.

specs/bellatrix/fork-choice.md
```diff
@@ -76,6 +77,102 @@ As per EIP-3675, before a post-transition block is finalized, `notify_forkchoice
...
+    shuffling_stable = proposal_slot % SLOTS_PER_EPOCH != 0
+
+    # FFG information of the new head_block will be competitive with the current head.
+    # FIXME: needs more justification/finalization information in the store
+    # ffg_competitive =
+    #    store.realized_justifications[parent_root] ==
+    #        store.realized_justifications[head_root] and
+    #    store.realized_finalizations[parent_root] ==
+    #        store.realized_finalization[head_root]
+    ffg_competitive = True
```
potuz, line 145: At this stage the validator has already asserted that it's a single block reorg, therefore it can simply include whatever attestations where in the orphaned block and not in the parent, it's guarantee to have the same justification than the orphaned block

specs/bellatrix/fork-choice.md
michaelsproul: > validator_is_connected can return true for a large number of validators in the case of a pool. In the above scenario, a pool proposing at N and N+1 will actually pass with the wrong validator the check for proposing_reorg_slot.

@potuz I don't understand what you mean here, can you spell it out in a bit more detail? I thought we already covered this case by demonstrating that the fcU for `N - 1` won't be suppressed, and if the head switches to `N` upon counting attestations towards the end of slot `N`, then the fcU for `N` _also_ won't be suppressed because it will fail the `parent_slot_ok` check (`N`'s parent is `N - 2`). Hence we'll still build on `N` via the normal process (no re-org attempts).

specs/bellatrix/fork-choice.md
michaelsproul: Good catch, updated in 9ca5b5ea3a1e4635c2273047b035d6bf26c57ebe.

michaelsproul:
I've rebased this on `dev` and addressed some of the outstanding review comments. I think there are still several decisions we need to take before merging:

- Q1: Is the current location alongside the other fork choice functions the right location for these specs? As they're optional they could alternatively live in `./fork_choice` alongside the safe block specs.
- Q2: How should we address the tension between specifying a precise strategy, and giving implementations the freedom to explore more aggressive strategies? This has implications for how the spec is written, which tests we write, and what inferences we are able to make about re-orgs on live networks.
  - A: max precision: Don't account for any deviations from the spec. Write test cases that enforce when clients should _and shouldn't_ re-org (positive and negative tests, respectively).
  - B: max freedom: Allow clients to deviate from the spec in any way they wish. Write some test cases that describe minimum scenarios in which clients should re-org, but don't specify any negative tests beyond very extreme cases where re-orgs are actually impossible.
  - C: balance: Account for _some_ acceptable deviations from the spec in tests. Write a mix of positive and negative tests while trying to avoid negative tests that would fail on current implementations. Allow implementations to skip some tests if they explore more exotic strategies in future.

I think my current preference for Q2 is option (C). I think the only outstanding difference between Prysm and Lighthouse is our handling of unrealized justification. Lighthouse currently won't re-org a block that updates unrealized justification, while Prysm will, because it assumes it can achieve equivalent justification by packing the block's attestations. I'm not opposed to changing Lighthouse to follow Prysm and specifying this as a case where we _should_ re-org.

specs/bellatrix/fork-choice.md
michaelsproul: Agree. Lighthouse could handle this, so I'm OK with removing this check and changing our implementation assuming nobody spots an issue with this approach, e.g. a case where it would be less profitable or impossible to pack the attestations which achieve justification/finalization...

Actually, what about the case where the attestations are about to expire past their 32 slot inclusion limit? In this case we could be prevented from including them.

dankrad:
I am starting to get more worried about this change -- the reason is that it implements a version of the block/slot fork choice, but without the mitigations that would have to be in place to avoid reorging under poor networking conditions.

Also, we should probably open a discussion about the 4s attestation timeline at this point. In order for attestations, we need (1) block propagation and (2) block validation all to happen in this time slot, which is starting to become more work (e.g. with 4844) and does not feel like the appropriate slot division anymore since the work in the other three thirds remains the same (and is overall less time critical).

mkalinin:
> without the mitigations that would have to be in place to avoid reorging under poor networking conditions.

There is a constraint on the slot of a parent block as a mitigation to poor network connections, i.e. `parent_block.slot + 1 == head_block.slot`. If a block was missed in a parent slot then no reorg is enforced.

> Also, we should probably open a discussion about the 4s attestation timeline at this point.

I think we should raise it on one of the next calls. It would be nice to see the distribution of attestation and aggregates submitting over time into their respective slot phases.

dankrad:
> There is a constraint on the slot of a parent block as a mitigation to poor network connections, i.e. `parent_block.slot + 1 == head_block.slot`. If a block was missed in a parent slot then no reorg is enforced.

Yes, but that makes any slot whose parent was not reorged subject to a reorg. Seems pretty harsh!



> I think we should raise it on one of the next calls. It would be nice to see the distribution of attestation and aggregates submitting over time into their respective slot phases.

Yeah, it's a good idea, I will check if we can get this data

michaelsproul:
I've pushed the change suggested by @potuz to address the "selfish mining" attack described [here](https://ethresear.ch/t/selfish-mining-in-pos/15551).

I would appreciate some help from @hwwhww writing tests. Alternatively we could merge as-is and add tests in a follow-up PR.

specs/bellatrix/fork-choice.md
```diff
@@ -76,6 +77,101 @@ As per EIP-3675, before a post-transition block is finalized, `notify_forkchoice
...
+    proposal_epoch = compute_epoch_at_slot(proposal_slot)
+    epochs_since_finalization = proposal_epoch - store.finalized_checkpoint.epoch
+    finalization_ok = epochs_since_finalization <= REORG_MAX_EPOCHS_SINCE_FINALIZATION
+
+    # Single slot re-org.
+    parent_slot_ok = parent_block.slot + 1 == head_block.slot
+    time_into_slot = (store.time - store.genesis_time) % SECONDS_PER_SLOT
+    current_time_ok = (head_block.slot == current_slot or
+                       (proposal_slot == current_slot and
+                        time_into_slot <= SECONDS_PER_SLOT // INTERVALS_PER_SLOT // 2))
```
adiasg, line 132: @michaelsproul can you add rationale on why `SECONDS_PER_SLOT // INTERVALS_PER_SLOT // 2`?

Is it simply putting this deadline halfway between slot start & attestation deadline?

michaelsproul: Yeah this is pretty approximate, but the idea is that we need to give the block some time to propagate. In Lighthouse this cutoff is configurable via [`--proposer-reorg-cutoff`](https://lighthouse-book.sigmaprime.io/late-block-re-orgs.html#command-line-flags).

Do you think a dedicated variable `REORG_CUTOFF_DENOMINATOR = 6` would be clearer?

hwwhww: I'd prefer the #3510 style solution in the long term. However, we may not be able to merge that before Deneb, so maybe just add some more comments here for now?

```suggestion
    proposer_reorg_cutoff = SECONDS_PER_SLOT // INTERVALS_PER_SLOT // 2
    current_time_ok = (
        head_block.slot == current_slot
        or (proposal_slot == current_slot and time_into_slot <= proposer_reorg_cutoff))
```

dankrad:
In addition to my previous reservations on this change, I also have a new one:

A fork choice change like this should not be taken light-heartedly. In fact, we explicitly don't want that people think they can just mess with the fork choice. As an example, if someone built a bribery protocol to reorg blocks intentionally, I would consider this an attack on Ethereum.

Therefore it might be best to actually roll back the change from clients now and re-start the discussion to make sure this is in a public forum and agreed by everyone, at least on ACDC.

potuz:
This was widely discussed in ACD at the time and thoroughly agreed upon. The only reason this wasn't merged then is that we couldn't make public some aspects of pulled tips. The implementation complexity of these changes are deeply intricate to roll back lightly

michaelsproul:
@dankrad I was keen to push ahead with this fix, because as Potuz said it had fairly broad support, but it also mitigated what I saw as a security issue introduced by MEV at the merge. Without late block re-orgs, proposers are incentivised to wait until the last moment to reveal their blocks, e.g. 10s into the slot. I think this damages the network more than intentionally re-orging blocks that are late.

Another thing is that this change does not modify fork choice, as in the part of fork choice that determines what the head should be. It's just an (incentive-compatible) tweak to the rules for choosing which block to build upon. Unlike bribe-based re-orgs, it also reinforces the honest behaviour of other validators: publishing blocks on time, and publishing attestations at or before the attestation deadline.

Like Prysm, Lighthouse has this feature behind a flag and could disable it by default (or permanently) if desired.

hwwhww:
Hi @michaelsproul 

I refactored the specs to re-arrange and reduce the duplicate code in `get_proposer_head` and `should_override_forkchoice_update`. It helps me to define the test cases. What do you think about this refactoring?

> - Q1: Is the current location alongside the other fork choice functions the right location for these specs? As they're optional they could alternatively live in `./fork_choice` alongside the safe block specs.

It reminds me of the confirmation rule PR, which defines the new feature in a new `fork_choice/confirmation-rule.md` file. I'm not 100% sure if adding a new doc is necessary, but it would be better to clarify which parts are "optional" and which are "honest validator fork choice rule".

> If `should_override_forkchoice_update` returns `True` but `get_proposer_head` later chooses the canonical head rather than its parent, then this is a misprediction that will cause the node to construct a payload with less notice. The result of `get_proposer_head` MUST be honored in preference to the heuristic method.

I saw that `get_proposer_head` is optional in phase0 FC, but here, it's not clear to me if `should_override_forkchoice_update` is MUST to enable. Could you clarify a bit?

> - Q2: How should we address the tension between specifying a precise strategy, and giving implementations the freedom to explore more aggressive strategies? This has implications for how the spec is written, which tests we write, and what inferences we are able to make about re-orgs on live networks.

IMO clients can have the flexibility to implement it into a more efficient algorithm, but in regard to the result of the computed head, it should be identical because of the risk of chain split. I doubt spec tests have covered everything. 😅

dankrad:
> Another thing is that this change does not modify fork choice, as in the part of fork choice that determines what the head should be.

It modifies the rules on how to vote for a block. This is what a bribery protocol to e.g. censor certain transactions permanently would do, too.

> It's just an (incentive-compatible) tweak to the rules for choosing which block to build upon. Unlike bribe-based re-orgs, it also reinforces the honest behaviour of other validators: publishing blocks on time, and publishing attestations at or before the attestation deadline.

Again, I'm not so much against this change, but now feel that it needs broader agreement and is not something that clients should be able to implement before that is reached. Maybe this is something to at least discuss on next ACD if people generally agree with this change.

specs/phase0/fork-choice.md
```diff
@@ -76,11 +86,16 @@ Any of the above handlers that trigger an unhandled exception (e.g. a failed ass
...
 ### Configuration
 
-| Name                   | Value        |
-| ---------------------- | ------------ |
-| `PROPOSER_SCORE_BOOST` | `uint64(40)` |
+| Name                                  | Value        |
+| ------------------------------------- | ------------ |
+| `PROPOSER_SCORE_BOOST`                | `uint64(40)` |
+| `REORG_WEIGHT_THRESHOLD`              | `uint64(20)` |
+| `REORG_PARENT_WEIGHT_THRESHOLD`       | `uint64(160)`|
```
hwwhww, line 93: @michaelsproul could you clarify why `REORG_PARENT_WEIGHT_THRESHOLD` is `160`?

michaelsproul: The calculation to derive it is here: https://ethresear.ch/t/selfish-mining-in-pos/15551/5

Do you want to link to that, or should we include a short summary of that idea?

michaelsproul: The intuitive explanation is that we need 80% of attesters from both slot N - 1 and N to see N - 1 as the head.

potuz:
> It modifies the rules on how to vote for a block. This is what a bribery protocol to e.g. censor certain transactions permanently would do, too.

I don't see how this modifies voting rules:: attesters in the orphaned committee would not have attested for the late block cause they sent attestations at 4 seconds. Attesters in the next committee are following the usual forkchoice and proposer boost mandates them to vote for the reorging block.

potuz:
> having things like this behind flags dubious at best and should be considered a temporary solution to address urgent problems and not a permanent state.

Just to clarify this, the only reason Prysm has this behind a flag is to protect from an implementation bug that can cause a liveness issue. We do this with most changes that touch core functions like block production. This has nothing to do with the feature being or not merged in the spec. As far as we were concerned this was thoroughly discussed and agreed in ACD albeit some reservations about disclosing because of certain forkchoice attacks that's were present at the time

0xETHengineer (approved):

0xETHengineer (approved):

hwwhww (commented):
@michaelsproul 

Sorry for the late assignment!

I added some basic test cases, updated test format with `should_override_forkchoice_update` result, and fixed trivial things.

TODOs
- [ ] confirm if test format is ok and run with a client
- [ ] check if we should output `get_proposer_head` result. It is labeled optional so I hesitated. also, consider moving `get_proposer_head` to a separate place (https://github.com/ethereum/consensus-specs/pull/3034#issuecomment-1772219431)

It would be nice to check again in the ACDC call, and then we can move this PR to "last call" status + merge.

specs/bellatrix/fork-choice.md
michaelsproul: I added this comment in 3f1bc2051294d7ea632b5010500f98a331119d39.

specs/phase0/fork-choice.md
michaelsproul: Updated the PR description with the link

specs/phase0/fork-choice.md
```diff
@@ -76,11 +86,16 @@ Any of the above handlers that trigger an unhandled exception (e.g. a failed ass
 
 ### Configuration
 
-| Name                   | Value        |
-| ---------------------- | ------------ |
-| `PROPOSER_SCORE_BOOST` | `uint64(40)` |
+| Name                                  | Value        |
+| ------------------------------------- | ------------ |
+| `PROPOSER_SCORE_BOOST`                | `uint64(40)` |
+| `REORG_WEIGHT_THRESHOLD`              | `uint64(20)` |
```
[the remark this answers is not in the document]
michaelsproul: Great idea! Renamed in 3f1bc2051294d7ea632b5010500f98a331119d39.

michaelsproul:
Hi @hwwhww, sorry for the slow reply! Thanks for cleaning this up and adding tests, I really appreciate that :blush:

> I saw that get_proposer_head is optional in phase0 FC, but here, it's not clear to me if should_override_forkchoice_update is MUST to enable. Could you clarify a bit?

I tried to clarify the wording of this section in 3f1bc2051294d7ea632b5010500f98a331119d39. Both functions are optional, and `should_override_forkchoice_update` should only be enabled when the proposer reorg feature is enabled. This is how Lighthouse and Prysm work currently.

> check if we should output get_proposer_head result. It is labeled optional so I hesitated. also, consider moving get_proposer_head to a separate place

I think the tests should include `get_proposer_head` as well, as it is arguably more important than `should_override_forkchoice_update`. If clients don't implement it they can simply skip those checks in their fork choice test runner.

I also think we should just leave it in the current location, as it seems like we may want to make this feature mandatory in future (Potuz mentioned that several PBS designs rely on it, but I haven't dug deeper).

> confirm if test format is ok and run with a client

I'll try running the test vectors with Lighthouse soon. Maybe once they also include `get_proposer_head` checks.

hwwhww:
@michaelsproul thank you!

I updated the test format: https://github.com/ethereum/consensus-specs/blob/d8440f8bb4496d053a36b5fef64420d5149156bb/tests/formats/fork_choice/README.md#checks-step
1.  add `get_proposer_head`
2. add mocking `validator_is_connected` into `should_override_forkchoice_update` check.

Now running the fork-choice testgen.

tests/core/pyspec/eth2spec/test/helpers/fork_choice.py
```diff
@@ -92,14 +92,11 @@ def add_attestations(spec, store, attestations, test_steps, is_from_block=False)
...
-    next_epoch_time = block_time + spec.SLOTS_PER_EPOCH * spec.config.SECONDS_PER_SLOT
-
-    if store.time < next_epoch_time:
-        spec.on_tick(store, next_epoch_time)
-        test_steps.append({'tick': int(next_epoch_time)})
+    # Make get_current_slot(store) >= attestation.data.slot + 1
+    min_time_to_include = (attestation.data.slot + 1) * spec.config.SECONDS_PER_SLOT
+    if store.time < min_time_to_include:
+        spec.on_tick(store, min_time_to_include)
+        test_steps.append({'tick': int(min_time_to_include)})
```
hwwhww, line 99: Just a note: this change would modify other test vectors.

hwwhww:
@michaelsproul 

FC-only test vectors of commit d8440f8
- [minimal_fc_d8440f8.tar.gz](https://github.com/ethereum/consensus-specs/files/13180532/minimal_fc_d8440f8.tar.gz)
- [mainnet_fc_d8440f8.tar.gz](https://github.com/ethereum/consensus-specs/files/13180529/mainnet_fc_d8440f8.tar.gz)

michaelsproul:
> FC-only test vectors of commit https://github.com/ethereum/consensus-specs/commit/d8440f8bb4496d053a36b5fef64420d5149156bb

Tests are passing on Lighthouse (https://github.com/sigp/lighthouse/pull/4887). I expected at least one test to fail because Lighthouse doesn't have the `REORG_PARENT_WEIGHT_THRESHOLD` check. This implies the test coverage could be expanded with a case where the reorg would go ahead if it weren't for the low weight of the parent (i.e. where `head_weight < REORG_WEIGHT_THRESHOLD && parent_weight < REORG_PARENT_WEIGHT_THRESHOLD`).

hwwhww (approved):
@michaelsproul 

LGTM. I'll try to add more test cases in other PRs!

Issue #3026, opened 2022-10-05
EIP4844: Suggestion to remove ssz from the Fiat-Shamir function 


kevaundray:
The method to create Fiat-Shamir challenges looks like:

```python
def hash_to_bls_field(x: Container) -> BLSFieldElement:
    """
    Compute 32-byte hash of serialized container and convert it to BLS field.
    The output is not uniform over the BLS field.
    """
    return bytes_to_bls_field(hash(ssz_serialize(x)))
 ```
 
For security, the Fiat-Shamir protocol requires a bijection from an object to bytes. This is the case if the serialise methods on an ssz struct is setup correctly since [ssz was intended to be bijective](https://ethereum.stackexchange.com/a/82056).

This however indirectly ties the security of the commitment scheme to the ssz serialisation strategy, ie its not possible to test the Fiat-Shamir component entirely in the cryptography library without first assuming that the serialisation strategy that caller is using is bijective (The cryptography would only receive bytes). This seems to be an abstraction leak.
 
To keep the cryptography completely modularised, one could implement this function as:
 
 ```python
 def hash_to_bls_field(polys: List[Tuple[Polynomial | Blob]], comms: List[KZGCommitment]) -> BLSFieldElement:
    """
    Compute 32-byte hash of serialised polynomials and commitments concatenated
    This hash is then converted to a BLS field.
    The output is not uniform over the BLS field.
    """
    
    bytes = []
    
    # Append each polynomial
    for poly in polys:
        for serialised_evaluation in poly:
            bytes.extend(serialised_evaluation)
    
    # Append serialised g1 points
    for serialised_comm in comms:
        bytes.extend(serialised_comm)
        
    return bytes_to_bls_field(hash(bytes))
  ```
  
The cryptography already knows how to de/serialise field and group elements, so the `hash_to_bls_field`  can live in the cryptography code and tested in isolation.

**Changes implied**

Instead of:

```python
hash_to_bls_field(BlobsAndCommitments(blobs=blobs, kzg_commitments=kzg_commitments))
hash_to_bls_field(PolynomialAndCommitment(polynomial=aggregated_poly,kzg_commitment=aggregated_poly_commitment)
```

It would be called as:

```python
hash_to_bls_field(blobs, kzg_commitments)
hash_to_bls_field([aggregated_poly],[aggregated_poly_commitment])
```

[...]

PR #3038, opened 2022-10-14
EIP4844: Update cryptography API and Fiat-Shamir logic

asn-d6:
This PR does a few things with the cryptography of EIP-4844.

Most importantly, it changes the public API of the KZG library to the following high-level API:
```
- verify_kzg_proof()
- compute_aggregate_kzg_proof()
- verify_aggregate_kzg_proof()
- blob_to_kzg_commitment()
```
compared to the old much more low-level API:
```
- compute_powers()
- matrix_lincomb()
- lincomb()
- bytes_to_bls_field()
- evaluate_polynomial_in_evaluation_form()
- verify_kzg_proof()
- compute_kzg_proof()
```

This means that all the cryptographic logic (including Fiat-Shamir) is now isolated and hidden in the KZG library and the `validator.md` file ends up being significantly simplified, only calling high-level KZG functions.

Some additional things that this PR does:
- Moves all EIP4844 cryptography into polynomial-commitments.md
- Improves the Fiat-Shamir stack by removing the need for SSZ and by introducing domain separators.

Future TODO tasks:
- Add unittests in `test_validator.py` that tests the high-level API

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -590,7 +590,6 @@ class EIP4844SpecBuilder(BellatrixSpecBuilder):
         return super().imports(preset_name) + f'''
 from eth2spec.utils import kzg
 from eth2spec.bellatrix import {preset_name} as bellatrix
-from eth2spec.utils.ssz.ssz_typing import ByteVector
 '''
 
 
--- a/setup.py
+++ b/setup.py
@@ -772,7 +772,7 @@ ignored_dependencies = [
     'uint8', 'uint16', 'uint32', 'uint64', 'uint128', 'uint256',
     'bytes', 'byte', 'ByteList', 'ByteVector',
     'Dict', 'dict', 'field', 'ceillog2', 'floorlog2', 'Set',
-    'Optional', 'Sequence', 
+    'Optional', 'Sequence',
 ]
 
 
--- a/setup.py
+++ b/setup.py
@@ -616,7 +616,7 @@ KZG_SETUP_LAGRANGE = TESTING_KZG_SETUP_LAGRANGE
 ROOTS_OF_UNITY = kzg.compute_roots_of_unity(TESTING_FIELD_ELEMENTS_PER_BLOB)
 
 
-def retrieve_blobs_sidecar(slot: Slot, beacon_block_root: Root) -> BlobsSidecar:
+def retrieve_blobs_sidecar(slot: Slot, beacon_block_root: Root) -> Optional[BlobsSidecar]:
     return "TEST"'''
 
     @classmethod
--- a/setup.py
+++ b/setup.py
@@ -232,7 +232,7 @@ def get_spec(file_name: Path, preset: Dict[str, str], config: Dict[str, str]) ->
 
                     if not _is_constant_id(name):
                         # Check for short type declarations
-                        if value.startswith(("uint", "Bytes", "ByteList", "Union", "Vector", "List")):
+                        if value.startswith(("uint", "Bytes", "ByteList", "Union", "Vector", "List", "ByteVector")):
                             custom_types[name] = value
                         continue
 
@@ -590,6 +590,7 @@ class EIP4844SpecBuilder(BellatrixSpecBuilder):
         return super().imports(preset_name) + f'''
 from eth2spec.utils import kzg
 from eth2spec.bellatrix import {preset_name} as bellatrix
+from eth2spec.utils.ssz.ssz_typing import ByteVector
 '''
 
 
@@ -617,12 +618,13 @@ ROOTS_OF_UNITY = kzg.compute_roots_of_unity(TESTING_FIELD_ELEMENTS_PER_BLOB)
 
 
 def retrieve_blobs_sidecar(slot: Slot, beacon_block_root: Root) -> BlobsSidecar:
-    pass'''
+    return "TEST"'''
 
     @classmethod
     def hardcoded_custom_type_dep_constants(cls, spec_object) -> str:
         constants = {
             'FIELD_ELEMENTS_PER_BLOB': spec_object.preset_vars['FIELD_ELEMENTS_PER_BLOB'].value,
+            'BYTES_PER_FIELD_ELEMENT': spec_object.preset_vars['BYTES_PER_FIELD_ELEMENT'].value,
             'MAX_BLOBS_PER_BLOCK': spec_object.preset_vars['MAX_BLOBS_PER_BLOCK'].value,
         }
         return {**super().hardcoded_custom_type_dep_constants(spec_object), **constants}
@@ -771,7 +773,7 @@ ignored_dependencies = [
     'uint8', 'uint16', 'uint32', 'uint64', 'uint128', 'uint256',
     'bytes', 'byte', 'ByteList', 'ByteVector',
     'Dict', 'dict', 'field', 'ceillog2', 'floorlog2', 'Set',
-    'Optional', 'Sequence',
+    'Optional', 'Sequence', 
 ]
 
 
--- a/setup.py
+++ b/setup.py
@@ -622,8 +622,8 @@ def retrieve_blobs_sidecar(slot: Slot, beacon_block_root: Root) -> Optional[Blob
     @classmethod
     def hardcoded_custom_type_dep_constants(cls, spec_object) -> str:
         constants = {
+            'BYTES_PER_FIELD_ELEMENT': spec_object.constant_vars['BYTES_PER_FIELD_ELEMENT'].value,
             'FIELD_ELEMENTS_PER_BLOB': spec_object.preset_vars['FIELD_ELEMENTS_PER_BLOB'].value,
-            'BYTES_PER_FIELD_ELEMENT': spec_object.preset_vars['BYTES_PER_FIELD_ELEMENT'].value,
             'MAX_BLOBS_PER_BLOCK': spec_object.preset_vars['MAX_BLOBS_PER_BLOCK'].value,
         }
         return {**super().hardcoded_custom_type_dep_constants(spec_object), **constants}
--- a/setup.py
+++ b/setup.py
@@ -590,7 +590,6 @@ class EIP4844SpecBuilder(BellatrixSpecBuilder):
         return super().imports(preset_name) + f'''
 from eth2spec.utils import kzg
 from eth2spec.bellatrix import {preset_name} as bellatrix
-from eth2spec.utils.ssz.ssz_impl import serialize as ssz_serialize
 '''
 
 
--- a/specs/eip4844/polynomial-commitments.md
+++ b/specs/eip4844/polynomial-commitments.md
@@ -219,10 +219,10 @@ def g1_lincomb(points: Sequence[KZGCommitment], scalars: Sequence[BLSFieldElemen
 
 ```python
 def poly_lincomb(polys: Sequence[Polynomial],
-                 scalars: Sequence[BLSFieldElement]) -> Sequence[Polynomial]:
+                 scalars: Sequence[BLSFieldElement]) -> Polynomial:
     """
     Given a list of ``polynomials``, interpret it as a 2D matrix and compute the linear combination
-    of each column with `scalars`: return the resulting vector.
+    of each column with `scalars`: return the resulting polynomials.
     """
     result = [0] * len(polys[0])
     for v, s in zip(polys, scalars):
--- a/specs/eip4844/polynomial-commitments.md
+++ b/specs/eip4844/polynomial-commitments.md
@@ -313,7 +313,7 @@ def compute_aggregated_poly_and_commitment(
     r_powers = compute_powers(r, len(kzg_commitments))
 
     # Create aggregated polynomial in evaluation form
-    aggregated_poly = Polynomial(vector_lincomb([blob_to_field_elements(blob for blob in blobs), r_powers))
+    aggregated_poly = Polynomial(vector_lincomb([blob_to_field_elements(blob) for blob in blobs], r_powers))
 
     # Compute commitment to aggregated polynomial
     aggregated_poly_commitment = KZGCommitment(g1_lincomb(kzg_commitments, r_powers))
--- a/specs/eip4844/polynomial-commitments.md
+++ b/specs/eip4844/polynomial-commitments.md
@@ -152,12 +152,12 @@ def blob_to_polynomial(blob: Blob) -> Polynomial:
     """
     Convert a blob to list of BLS field scalars.
     """
-    r = Polynomial()
+    polynomial = Polynomial()
     for i in range(FIELD_ELEMENTS_PER_BLOB):
         value = int.from_bytes(blob[i * BYTES_PER_FIELD_ELEMENT: (i + 1) * BYTES_PER_FIELD_ELEMENT], ENDIANNESS)
         assert value < BLS_MODULUS
-        r[i] = value
-    return r
+        polynomial[i] = value
+    return polynomial
 ```
 
 #### `hash_to_bls_field`
@@ -178,9 +178,9 @@ def hash_to_bls_field(polys: Sequence[Polynomial],
     # Append each polynomial which is composed by field elements
     for poly in polys:
         for field_element in poly:
-            data += int.to_bytes(field_element, 32, ENDIANNESS)
+            data += int.to_bytes(field_element, BYTES_PER_FIELD_ELEMENT, ENDIANNESS)
 
-    # Append serialised G1 points
+    # Append serialized G1 points
     for commitment in comms:
         data += commitment
 
--- a/specs/eip4844/polynomial-commitments.md
+++ b/specs/eip4844/polynomial-commitments.md
@@ -176,7 +176,7 @@ def hash_to_bls_field(initializer: bytes, polys: Sequence[Polynomial], comms: Se
     data += DOMAIN_SEPARATOR_SQUEEZE
     h = hash(data)
 
-    return bytes_to_bls_field(hash(data)), h
+    return bytes_to_bls_field(h), h
 ```
 
 #### `bls_modular_inverse`
--- a/specs/eip4844/polynomial-commitments.md
+++ b/specs/eip4844/polynomial-commitments.md
@@ -46,6 +46,7 @@ This document specifies basic polynomial operations and KZG polynomial commitmen
 | `BLSFieldElement` | `uint256` | `x < BLS_MODULUS` |
 | `KZGCommitment` | `Bytes48` | Same as BLS standard "is valid pubkey" check but also allows `0x00..00` for point-at-infinity |
 | `KZGProof` | `Bytes48` | Same as for `KZGCommitment` |
+| `Polynomial` | `Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]` | a polynomial in evaluation form |
 
 ## Constants
 
@@ -53,6 +54,10 @@ This document specifies basic polynomial operations and KZG polynomial commitmen
 | - | - | - |
 | `BLS_MODULUS` | `52435875175126190479447740508185965837690552500527637822603658699938581184513` | Scalar field modulus of BLS12-381 |
 | `ROOTS_OF_UNITY` | `Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]` | Roots of unity of order FIELD_ELEMENTS_PER_BLOB over the BLS12-381 field |
+| `DOMAIN_SEPARATOR_FIELD_ELEMENT` | `b"FIELD_ELEMENT"` | Fiat-Shamir domain separator for field elements |
+| `DOMAIN_SEPARATOR_POINT` | `b"POINT"` | Fiat-Shamir domain separator for G1 points |
+| `DOMAIN_SEPARATOR_SQUEEZE` | `b"SQUEEZE"` | Fiat-Shamir domain separator before hashing |
+
 
 ## Preset
 
@@ -122,17 +127,31 @@ def bytes_to_bls_field(b: Bytes32) -> BLSFieldElement:
     return int.from_bytes(b, "little") % BLS_MODULUS
 ```
 
-### `hash_to_bls_field`
+#### `hash_to_bls_field`
 
 ```python
-def hash_to_bls_field(x: Container) -> BLSFieldElement:
+def hash_to_bls_field(polys: Sequence[Polynomial], comms: Sequence[KZGCommitment]) -> BLSFieldElement:
     """
-    Compute 32-byte hash of serialized container and convert it to BLS field.
+    Compute 32-byte hash of serialised polynomials and commitments concatenated.
+    This hash is then converted to a BLS field element.
     The output is not uniform over the BLS field.
     """
-    return bytes_to_bls_field(hash(ssz_serialize(x)))
-```
+    data = bytes()
 
+    # Append each polynomial which is composed by field elements
+    for poly in polys:
+        for field_element in poly:
+            data += DOMAIN_SEPARATOR_FIELD_ELEMENT
+            data += int.to_bytes(field_element, 32, ENDIANNESS)
+
+    # Append serialised G1 points
+    for commitment in comms:
+        data += DOMAIN_SEPARATOR_POINT
+        data += commitment
+
+    data += DOMAIN_SEPARATOR_SQUEEZE
+    return bytes_to_bls_field(hash(data))
+```
 
 #### `bls_modular_inverse`
 
@@ -263,7 +282,7 @@ def compute_aggregated_poly_and_commitment(
     Return the aggregated polynomial and aggregated KZG commitment.
     """
     # Generate random linear combination challenges
-    r = hash_to_bls_field(BlobsAndCommitments(blobs=blobs, kzg_commitments=kzg_commitments))
+    r = hash_to_bls_field(blobs, kzg_commitments)
     r_powers = compute_powers(r, len(kzg_commitments))
 
     # Create aggregated polynomial in evaluation form
--- a/specs/eip4844/polynomial-commitments.md
+++ b/specs/eip4844/polynomial-commitments.md
@@ -62,6 +62,7 @@ This document specifies basic polynomial operations and KZG polynomial commitmen
 | Name | Value | Notes |
 | - | - | - |
 | `BLS_MODULUS` | `52435875175126190479447740508185965837690552500527637822603658699938581184513` | Scalar field modulus of BLS12-381 |
+| `BYTES_PER_FIELD_ELEMENT` | `uint64(32)` | Bytes per BLS field element |
 
 ## Preset
 
@@ -69,7 +70,6 @@ This document specifies basic polynomial operations and KZG polynomial commitmen
 
 | Name | Value |
 | - | - |
-| `BYTES_PER_FIELD_ELEMENT` | `uint64(32)` |
 | `FIELD_ELEMENTS_PER_BLOB` | `uint64(4096)` |
 | `FIAT_SHAMIR_PROTOCOL_DOMAIN` | `b'FSBLOBVERIFY_V1_'` |
 
--- a/specs/eip4844/polynomial-commitments.md
+++ b/specs/eip4844/polynomial-commitments.md
@@ -59,18 +59,9 @@ This document specifies basic polynomial operations and KZG polynomial commitmen
 
 ## Constants
 
-*Note*: Domain separators are set to an empty string as they are not required according to current understanding (and are not "free").
-Should we determine that they are required at a later point we can set 
-
 | Name | Value | Notes |
 | - | - | - |
 | `BLS_MODULUS` | `52435875175126190479447740508185965837690552500527637822603658699938581184513` | Scalar field modulus of BLS12-381 |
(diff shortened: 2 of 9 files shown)


specs/eip4844/polynomial-commitments.md
```diff
@@ -226,12 +277,64 @@ def compute_kzg_proof(polynomial: Sequence[BLSFieldElement], z: BLSFieldElement)
...
+    return aggregated_poly, aggregated_poly_commitment
+```
+
+#### `compute_aggregate_kzg_proof`
+
+```python
+def compute_aggregate_kzg_proof(blobs: Sequence[Blob]) -> KZGProof:
+    commitments = [blob_to_kzg_commitment(blob) for blob in blobs]
+    aggregated_poly, aggregated_poly_commitment = compute_aggregated_poly_and_commitment(blobs, commitments)
+    x = hash_to_bls_field([aggregated_poly],[aggregated_poly_commitment])
```
asn-d6, line 308: Should we include r here?
(Best: use complete transcript)

specs/eip4844/polynomial-commitments.md
```diff
@@ -46,13 +51,18 @@ This document specifies basic polynomial operations and KZG polynomial commitmen
...
 | `KZGProof` | `Bytes48` | Same as for `KZGCommitment` |
+| `Polynomial` | `Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]` | a polynomial in evaluation form |
 
 ## Constants
 
 | Name | Value | Notes |
 | - | - | - |
 | `BLS_MODULUS` | `52435875175126190479447740508185965837690552500527637822603658699938581184513` | Scalar field modulus of BLS12-381 |
 | `ROOTS_OF_UNITY` | `Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]` | Roots of unity of order FIELD_ELEMENTS_PER_BLOB over the BLS12-381 field |
+| `DOMAIN_SEPARATOR_FIELD_ELEMENT` | `b"FIELD_ELEMENT"` | Fiat-Shamir domain separator for field elements |
```
asn-d6, line 62: Remove domain separators (IMO not required)

specs/eip4844/polynomial-commitments.md
```diff
@@ -46,13 +51,18 @@ This document specifies basic polynomial operations and KZG polynomial commitmen
 | `BLSFieldElement` | `uint256` | `x < BLS_MODULUS` |
 | `KZGCommitment` | `Bytes48` | Same as BLS standard "is valid pubkey" check but also allows `0x00..00` for point-at-infinity |
 | `KZGProof` | `Bytes48` | Same as for `KZGCommitment` |
+| `Polynomial` | `Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]` | a polynomial in evaluation form |
```
asn-d6, line 54: Add `Blob`; `Blob` only list of bytes
(This is Dankrad reviewing)

specs/eip4844/polynomial-commitments.md
kevaundray: Now that all of the cryptography has been shunted to "the cryptographers", it is possible to do this without burdening client devs

specs/eip4844/polynomial-commitments.md
kevaundray: This is reasonable to me, as it removes the need for non-crypto modules to know about field elements.

specs/eip4844/polynomial-commitments.md
kevaundray: This is the main blocker to this PR as its still under debate as to whether it is useful or not

kevaundray:
Closes #3026

djrtwo (commented):
Good separation of concerns :+1: 

can you give more on the debate over ssz vs domain separators vs neither?

Where-2:
No lost and have lost

Where-2:
Not sure what to do

kevaundray:
> Good separation of concerns :+1: 
> 
> 
> 
> can you give more on the debate over ssz vs domain separators vs neither?

Currently we all agree to remove ssz from the Fiat-Shamir function as this allows for client teams to not implement any of the cryptography.

The debate is on whether these separator labels that we append just before we add a serialised point or field element or polynomial into the bytes array, add any extra security vs having no label at all.

The argument for having them is that you want a precise mapping from your higher level objects to the bytes array. For example, if I was adding a point into the byte array, I could simply serialise the point as 48 bytes then append onto the byte array, or I could first append a label "POINT" and then add the 48 bytes. In this way, the byte array also encodes the types.

Although this seems like a reasonable thing to do, there is no strong rationale as to whether it improves security, and seems to lean on the side of being "extra safe"

Where-2:
My issue is I had no idea in what I was undertaking by investing.   It has both been profitable and costly.

kevaundray:
As of two minutes ago, I think we now have enough information to come to a conclusion on how to proceed with this

Where-2:
Hope so.   I can not sit with the fact that my trust wallet was absolutely defrauded.  I have little confidence.  Two years on and it still burns !

Where-2:
Thanks interested more in how we are reimbursed.  If coins get burnt but people lose money.  Counter productive it seems.

specs/eip4844/polynomial-commitments.md
xrchz: +1 to removing them

specs/eip4844/polynomial-commitments.md
```diff
@@ -226,12 +277,64 @@ def compute_kzg_proof(polynomial: Sequence[BLSFieldElement], z: BLSFieldElement)
...
+
+```python
+def compute_aggregated_poly_and_commitment(
+        blobs: Sequence[Blob],
+        kzg_commitments: Sequence[KZGCommitment]) -> Tuple[Polynomial, KZGCommitment]:
+    """
+    Return the aggregated polynomial and aggregated KZG commitment.
+    """
+    # Generate random linear combination challenges
+    r = hash_to_bls_field(blobs, kzg_commitments)
```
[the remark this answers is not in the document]
xrchz: I've been informed that blobs and polynomials are the same.

specs/eip4844/polynomial-commitments.md
```diff
@@ -173,7 +174,9 @@ def hash_to_bls_field(polys: Sequence[Polynomial], comms: Sequence[KZGCommitment
         data += commitment
 
     data += DOMAIN_SEPARATOR_SQUEEZE
-    return bytes_to_bls_field(hash(data))
+    h = hash(data)
+
+    return bytes_to_bls_field(hash(data)), h
```
kevaundray, line 179: This is equivalent to `return bytes_to_bls_field(h), h`

The rationale here is to initialise the fiat-shamir with the previous challenge that was used. 

I'm not quite sure what is happening with the domain separators, I'd just keep them in the hash_to_bytes_field method to reduce complexity; so domain separators are not in multiple places. Unless the current way is more secure, though I have not seen anything that lends itself to that conclusion

kevaundray: I have seen domain separators, where the aggregate kzg protocol is separated from the underlying/original kzg protocol

dankrad: The reason is that `hash_to_bls_field` is called in different contexts. IMO for the domain separators to make sense at all, the interesting part is not what types are being hashed, but what protocol is run. So we can distinguish that via the initializer *if necessary*, although I still don't believe that they are needed so they are all empty strings for now.

The second reason for the `initializer` is that when we run a two stage protocol we need to add the hash of the previous transcript.

specs/eip4844/polynomial-commitments.md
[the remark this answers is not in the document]
dankrad: Not anymore :)

specs/eip4844/polynomial-commitments.md
```diff
@@ -226,12 +301,64 @@ def compute_kzg_proof(polynomial: Sequence[BLSFieldElement], z: BLSFieldElement)
...
+    return aggregated_poly, aggregated_poly_commitment, h
+```
+
+#### `compute_aggregate_kzg_proof`
+
+```python
+def compute_aggregate_kzg_proof(blobs: Sequence[Blob]) -> KZGProof:
+    commitments = [blob_to_kzg_commitment(blob) for blob in blobs]
+    aggregated_poly, aggregated_poly_commitment, h = compute_aggregated_poly_and_commitment(blobs, commitments)
+    x, _ = hash_to_bls_field(DOMAIN_SEPARATOR_EVAL_PROTOCOL + h, [aggregated_poly], [aggregated_poly_commitment])
```
kevaundray, line 332: small nit: 

`h + DOMAIN_SEPARATOR_EVAL_PROTOCOL` is more natural than `DOMAIN_SEPARATOR_EVAL_PROTOCOL + h` because the previous transcript state `h` may already be saved in the hash buffer and so starting a new protocol would just mean we append `DOMAIN_SEPARATOR_EVAL_PROTOCOL` to the buffer

specs/eip4844/polynomial-commitments.md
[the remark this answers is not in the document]
kevaundray: Ideally the polynomial-commitments.md file has no reference to what a blob is (just polynomials), I was thinking to push that to a separate PR, but we might aswell do it in the same PR now :)

specs/eip4844/polynomial-commitments.md
```diff
@@ -315,7 +315,7 @@ def compute_aggregated_poly_and_commitment(
     r_powers = compute_powers(r, len(kzg_commitments))
 
     # Create aggregated polynomial in evaluation form
-    aggregated_poly = Polynomial(vector_lincomb([blob_to_field_elements(blob) for blob in blobs], r_powers))
+    aggregated_poly = Polynomial(poly_lincomb([blob_to_field_elements(blob) for blob in blobs], r_powers))
```
kevaundray, line 318: +1 to this, vector_lincomb/matrix_lincomb was not very clear

specs/eip4844/polynomial-commitments.md
```diff
@@ -215,17 +215,17 @@ def g1_lincomb(points: Sequence[KZGCommitment], scalars: Sequence[BLSFieldElemen
...
 
 ```python
-def vector_lincomb(vectors: Sequence[Sequence[BLSFieldElement]],
-                   scalars: Sequence[BLSFieldElement]) -> Sequence[BLSFieldElement]:
+def poly_lincomb(polys: Sequence[Polynomial],
+                 scalars: Sequence[BLSFieldElement]) -> Sequence[Polynomial]:
     """
-    Given a list of ``vectors``, interpret it as a 2D matrix and compute the linear combination
+    Given a list of ``polynomials``, interpret it as a 2D matrix and compute the linear combination
     of each column with `scalars`: return the resulting vector.
```
kevaundray, line 225: return the resulting vector -> return the resulting polynomial

specs/eip4844/polynomial-commitments.md
```diff
@@ -215,17 +215,17 @@ def g1_lincomb(points: Sequence[KZGCommitment], scalars: Sequence[BLSFieldElemen
...
 ```
 
-#### `vector_lincomb`
+#### `poly_lincomb`
 
 ```python
-def vector_lincomb(vectors: Sequence[Sequence[BLSFieldElement]],
-                   scalars: Sequence[BLSFieldElement]) -> Sequence[BLSFieldElement]:
+def poly_lincomb(polys: Sequence[Polynomial],
+                 scalars: Sequence[BLSFieldElement]) -> Sequence[Polynomial]:
```
kevaundray, line 222: I think the return type should be `Polynomial` instead of `Sequence[Polynomial]` ?

specs/eip4844/polynomial-commitments.md
dankrad: I think it should be the other way around:
1. I think the domain separator should be at the very start of the hash function, otherwise it does not prevent random oracle values from colliding.
2. h is not the full transcript, but only the output of the hash, so this is trivial for efficiency purposes.

kevaundray: Switching h to be appended first instead of second, should not change random oracle collisions I think, unless the hash function is bad.

Instead of thinking of putting it at the beginning of the hash function, I'm thinking of putting it at the beginning of the sub protocol. Ie each new protocol if given a unique domain separator will effectively have a "different" hash function or RO

dankrad: I do not believe that what you suggest functions as a domain separator. If your hashed string is
`"DOMAIN_SEPARATOR_PROTOCOL1" + TRANSCRIPT_PROTOCOL1 + "DOMAIN_SEPARATOR_PROTOCOL2" + TRANSCRIPT_PROTOCOL2`,
then what guarantees you that `"DOMAIN_SEPARATOR_PROTOCOL2"` does not appear somewhere in `TRANSCRIPT_PROTOCOL1`?

specs/eip4844/polynomial-commitments.md
```diff
@@ -46,16 +54,39 @@ This document specifies basic polynomial operations and KZG polynomial commitmen
...
+
 | Name | Value | Notes |
 | - | - | - |
 | `BLS_MODULUS` | `52435875175126190479447740508185965837690552500527637822603658699938581184513` | Scalar field modulus of BLS12-381 |
-| `ROOTS_OF_UNITY` | `Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]` | Roots of unity of order FIELD_ELEMENTS_PER_BLOB over the BLS12-381 field |
+| `DOMAIN_SEPARATOR_AGGREGATE_PROTOCOL` | `b""` | Fiat-Shamir domain separator random aggregation |
+| `DOMAIN_SEPARATOR_EVAL_PROTOCOL` | `b""` | Fiat-Shamir domain separator random evaluation |
+| `DOMAIN_SEPARATOR_FIELD_ELEMENT` | `b""` | Fiat-Shamir domain separator for field elements |
+| `DOMAIN_SEPARATOR_POINT` | `b""` | Fiat-Shamir domain separator for G1 points |
+| `DOMAIN_SEPARATOR_SQUEEZE` | `b""` | Fiat-Shamir domain separator before hashing |
```
hwwhww, line 72: It looks like these constants have to be updated if we want to use domain separators?

kevaundray: We are planning to modify those constants, one draft version of this is here: https://github.com/kevaundray/consensus-specs/commit/32491d008129b38e3004088afe10f992c4e46aa3

specs/eip4844/polynomial-commitments.md
kevaundray: This has been implemented, can mark as resolved

specs/eip4844/polynomial-commitments.md
```diff
@@ -46,16 +54,32 @@ This document specifies basic polynomial operations and KZG polynomial commitmen
...
+| `Polynomial` | `Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]` | a polynomial in evaluation form |
+| `Blob` | `ByteVector[BYTES_PER_FIELD_ELEMENT * FIELD_ELEMENTS_PER_BLOB]` | a basic blob data |
 
 ## Constants
 
 | Name | Value | Notes |
 | - | - | - |
-| `BLS_MODULUS` | `52435875175126190479447740508185965837690552500527637822603658699938581184513` | Scalar field modulus of BLS12-381 |
-| `ROOTS_OF_UNITY` | `Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]` | Roots of unity of order FIELD_ELEMENTS_PER_BLOB over the BLS12-381 field |
+| `BLS_MODULUS` | | `ROOTS_OF_UNITY` | `Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]` | Roots of unity of order FIELD_ELEMENTS_PER_BLOB over the BLS12-381 field |
```
kevaundray, line 64: The description of the BLS modulus here, should be revised

specs/eip4844/polynomial-commitments.md
```diff
@@ -117,9 +141,51 @@ def bit_reversal_permutation(sequence: Sequence[T]) -> Sequence[T]:
...
+
+```python
+def blob_to_field_elements(blob: Blob) -> Polynomial:
+    """
+    Convert a blob to list of BLS field scalars.
+    """
+    r = Polynomial()
+    for i in range(FIELD_ELEMENTS_PER_BLOB):
+        value = int.from_bytes(blob[i * BYTES_PER_FIELD_ELEMENT: (i + 1) * BYTES_PER_FIELD_ELEMENT], ENDIANNESS)
+        assert value < BLS_MODULUS
```
kevaundray, line 159: (Just a note)

For reference, this check here corresponds to the issue here: https://github.com/ethereum/consensus-specs/issues/3057

specs/eip4844/polynomial-commitments.md
```diff
@@ -117,9 +141,51 @@ def bit_reversal_permutation(sequence: Sequence[T]) -> Sequence[T]:
...
+```python
+def hash_to_bls_field(polys: Sequence[Polynomial],
+                      comms: Sequence[KZGCommitment]) -> Tuple[BLSFieldElement, Hash32]:
+    """
+    Compute 32-byte hash of serialized polynomials and commitments concatenated.
+    This hash is then converted to a BLS field element, where the result is not uniform over the BLS field.
+    Return the BLS field element.
     """
-    return int.from_bytes(b, "little") % BLS_MODULUS
+    # Append the number of polynomials and the degree of each polynomial as a domain separator
```
kevaundray, line 174: Optional: We could add an assert here `len(polys) == len(comms)`

kevaundray: One reason to maybe not include it, is because this check is indirectly being done in `validate_blobs_and_kzg_commitments`

assert len(blob_kzg_commitments) == len(blobs)

specs/eip4844/polynomial-commitments.md
```diff
@@ -117,9 +141,51 @@ def bit_reversal_permutation(sequence: Sequence[T]) -> Sequence[T]:
...
-    Convert bytes to a BLS field scalar. The output is not uniform over the BLS field.
+    Convert 32-byte value to a BLS field scalar. The output is not uniform over the BLS field.
+    """
+    return int.from_bytes(b, ENDIANNESS) % BLS_MODULUS
+```
+
+#### `blob_to_field_elements`
+
+```python
+def blob_to_field_elements(blob: Blob) -> Polynomial:
```
kevaundray, line 152: Optional: can change to `blob_to_polynomial`

kevaundray (approved):
Looks good to me. 

Have left some optional house-keeping comments and notes

specs/eip4844/polynomial-commitments.md
asn-d6: Seems like https://github.com/ethereum/consensus-specs/commit/d98c103cd7705e3a184e56e814184b5ce651dc42 broke the table here.

asn-d6: Also now ROOTS_OF_UNITY is defined twice.

specs/eip4844/polynomial-commitments.md
```diff
@@ -155,22 +223,65 @@ def g1_lincomb(points: Sequence[KZGCommitment], scalars: Sequence[BLSFieldElemen
...
+
+#### `evaluate_polynomial_in_evaluation_form`
+
+```python
+def evaluate_polynomial_in_evaluation_form(polynomial: Polynomial,
+                                           z: BLSFieldElement) -> BLSFieldElement:
+    """
+    Evaluate a polynomial (in evaluation form) at an arbitrary point ``z``.
+    Uses the barycentric formula:
+       f(z) = (1 - z**WIDTH) / WIDTH  *  sum_(i=0)^WIDTH  (f(DOMAIN[i]) * DOMAIN[i]) / (z - DOMAIN[i])
```
asn-d6, line 267: While we are at it, let's also fix the bug in the comment here: https://github.com/protolambda/go-kzg/pull/25#issuecomment-1289308555

specs/eip4844/polynomial-commitments.md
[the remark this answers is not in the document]
asn-d6: > Ideally the polynomial-commitments.md file has no reference to what a blob is (just polynomials), I was thinking to push that to a separate PR, but we might aswell do it in the same PR now :)

while this seems like a sensible idea in principle, I guess we are not doing it in this PR, correct? it would be a pretty substantial refactor.

specs/eip4844/polynomial-commitments.md
asn-d6: Done in 80d4d0941

specs/eip4844/polynomial-commitments.md
asn-d6: Done in 2e5a784d6

specs/eip4844/polynomial-commitments.md
asn-d6: Fixed in c8b8b53f8

specs/eip4844/polynomial-commitments.md
```diff
@@ -226,12 +302,64 @@ def compute_kzg_proof(polynomial: Sequence[BLSFieldElement], z: BLSFieldElement)
...
+    return verify_kzg_proof(aggregated_poly_commitment, x, y, kzg_aggregated_proof)
+```
+
 ### Polynomials
 
 #### `evaluate_polynomial_in_evaluation_form`
 
 ```python
-def evaluate_polynomial_in_evaluation_form(polynomial: Sequence[BLSFieldElement],
+def evaluate_polynomial_in_evaluation_form(polynomial: Polynomial,
```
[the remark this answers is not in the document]
asn-d6: Fixed! Thanks!

hwwhww (approved):
Great work on extracting the crypto lib helpers!

Some minor suggestions. Signing my approval stamp so that @asn-d6 can merge it.

specs/eip4844/polynomial-commitments.md
```diff
@@ -226,31 +336,53 @@ def compute_kzg_proof(polynomial: Sequence[BLSFieldElement], z: BLSFieldElement)
...
-                                           z: BLSFieldElement) -> BLSFieldElement:
+def compute_aggregated_poly_and_commitment(
+        blobs: Sequence[Blob],
+        kzg_commitments: Sequence[KZGCommitment]) -> Tuple[Polynomial, KZGCommitment, BLSFieldElement]:
     """
-    Evaluate a polynomial (in evaluation form) at an arbitrary point `z`
-    Uses the barycentric formula:
-       f(z) = (1 - z**WIDTH) / WIDTH  *  sum_(i=0)^WIDTH  (f(DOMAIN[i]) * DOMAIN[i]) / (z - DOMAIN[i])
+    Return (1) the aggregated polynomial, (2) the aggregated KZG commitment,
+    and (3) the hash of ``blobs`` and ``kzg_commitments``.
```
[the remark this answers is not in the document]
asn-d6: Fixed in cb46b11fc

specs/eip4844/polynomial-commitments.md
```diff
@@ -226,31 +336,53 @@ def compute_kzg_proof(polynomial: Sequence[BLSFieldElement], z: BLSFieldElement)
...
     """
-    width = len(polynomial)
-    assert width == FIELD_ELEMENTS_PER_BLOB
-    inverse_width = bls_modular_inverse(width)
+    # Generate random linear combination challenges
+    r = hash_to_bls_field(blobs, kzg_commitments)
+    r_powers = compute_powers(r, len(kzg_commitments))
+    r2 = int(r_powers[-1]) * r % BLS_MODULUS
```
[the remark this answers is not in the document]
asn-d6: Did an attempt at this in cb46b11fc

specs/eip4844/polynomial-commitments.md
```diff
@@ -117,9 +140,51 @@ def bit_reversal_permutation(sequence: Sequence[T]) -> Sequence[T]:
...
+```
+
+#### `blob_to_polynomial`
+
+```python
+def blob_to_polynomial(blob: Blob) -> Polynomial:
+    """
+    Convert a blob to list of BLS field scalars.
+    """
+    r = Polynomial()
```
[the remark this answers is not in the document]
asn-d6: Fixed in af48987c2

setup.py
```diff
@@ -618,12 +617,13 @@ def sundry_functions(cls) -> str:
 
 
 def retrieve_blobs_sidecar(slot: Slot, beacon_block_root: Root) -> BlobsSidecar:
-    pass'''
+    return "TEST"'''
```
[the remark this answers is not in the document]
asn-d6: Fixed in af48987c2

specs/eip4844/polynomial-commitments.md
```diff
@@ -117,9 +140,51 @@ def bit_reversal_permutation(sequence: Sequence[T]) -> Sequence[T]:
...
-    return int.from_bytes(b, "little") % BLS_MODULUS
+    # Append the number of polynomials and the degree of each polynomial as a domain separator
+    num_polys = int.to_bytes(len(polys), 8, ENDIANNESS)
+    degree_poly = int.to_bytes(FIELD_ELEMENTS_PER_BLOB, 8, ENDIANNESS)
+    data = FIAT_SHAMIR_PROTOCOL_DOMAIN + degree_poly + num_polys
+
+    # Append each polynomial which is composed by field elements
+    for poly in polys:
+        for field_element in poly:
+            data += int.to_bytes(field_element, 32, ENDIANNESS)
```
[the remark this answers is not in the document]
asn-d6: Fixed in af48987c2

specs/eip4844/polynomial-commitments.md
```diff
@@ -46,16 +54,31 @@ This document specifies basic polynomial operations and KZG polynomial commitmen
...
 | `BLS_MODULUS` | `52435875175126190479447740508185965837690552500527637822603658699938581184513` | Scalar field modulus of BLS12-381 |
-| `ROOTS_OF_UNITY` | `Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]` | Roots of unity of order FIELD_ELEMENTS_PER_BLOB over the BLS12-381 field |
 
 ## Preset
 
+### Blob
+
+| Name | Value |
+| - | - |
+| `BYTES_PER_FIELD_ELEMENT` | `uint64(32)` |
```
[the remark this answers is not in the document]
asn-d6: Hmm, not sure exactly what you mean here. `BYTES_PER_FIELD_ELEMENT` being 32 is essentially a parameter of the BLS field we are using.

specs/eip4844/polynomial-commitments.md
```diff
@@ -155,22 +222,65 @@ def g1_lincomb(points: Sequence[KZGCommitment], scalars: Sequence[BLSFieldElemen
...
+        current_power = current_power * int(x) % BLS_MODULUS
+    return powers
+```
+
+### Polynomials
+
+#### `evaluate_polynomial_in_evaluation_form`
+
+```python
+def evaluate_polynomial_in_evaluation_form(polynomial: Polynomial,
```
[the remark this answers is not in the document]
asn-d6: No strong opinion either. Inertia won and I left it as is.

kevaundray:
Where is `ENDIANNESS` defined?

asn-d6:
> Where is `ENDIANNESS` defined?

https://github.com/ethereum/annotated-spec/blob/master/phase0/beacon-chain.md#constants

specs/eip4844/polynomial-commitments.md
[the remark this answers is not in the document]
tbenr: my 2c.
I think the point is that if you specify `BYTES_PER_FIELD_ELEMENT` as a preset, you should expect different values to be potentially configured. But the`bytes_to_bls_field` only works if `BYTES_PER_FIELD_ELEMENT` is strictly 32.
So I see two options: 
1. `BYTES_PER_FIELD_ELEMENT` becomes a constant
2. generalize `bytes_to_bls_field`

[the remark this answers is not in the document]
asn-d6: > my 2c. I think the point is that if you specify `BYTES_PER_FIELD_ELEMENT` as a preset, you should expect different values to be potentially configured. But the`bytes_to_bls_field` only works if `BYTES_PER_FIELD_ELEMENT` is strictly 32. So I see two options:
> 
>     1. `BYTES_PER_FIELD_ELEMENT` becomes a constant
> 
>     2. generalize `bytes_to_bls_field`

Ah, that makes sense! Thanks!

Seems like `BYTES_PER_FIELD_ELEMENT` was added as a preset in cbc170b441d973041e72c9ef937966bb3b978e25, but I agree that it's not really a configurable value. Maybe it was done as a quick way to fix the tests if I read the commit correctly.

I  think it's a good idea to leave it as a constant (which it already seems to be in `polynomial-commitments.md`).

setup.py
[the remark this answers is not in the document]
dankrad: Hmm, slightly annoying because `None` could also mean that the sidecar is not available (I know in python we would do this by raising an exception but not all languages would have that available).

asn-d6:
Merged! Thanks everyone!

Issue #3043, opened 2022-10-16
post-devcon 4844 spec updates


djrtwo:
A number of spec modifications need to happen after discussions at devcon:

1. Couple block and blob sidecar in a single gossip topic -- beacon_block_and_blob_sidecar
    * Keep proposer signature on the sidecar to ensure a party cannot mutate OR we have to do the sidecar / blobtx KZG verifications
2. Reduce `MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS` to ~2 weeks or lower
2. Clarify Blob availability requirement in fork-choice by specifying an `is_data_available()` check for importing a block
3. Consider `MAX_REQUEST_BLOBS_SIDECARS` value
    * Currently 128 which is a very large amount of data, but you only MUST respond to 1st
    * Is this "only first" requirement (and the mirror to blocks by range) sufficient for load balancing or should teh value be reduced?
2. TBD -- potentially reduce blobs per block down to an appropriate value after testing and simulations
    * Note, this can be done now to something like 128KB/256KB if 1MB/2MB is blocking important testing
5. TBD -- based on 4844 load tests and lighthouse episub tests, finish episub spec and *potentially* make it suggested or mandatory

Where-2:
Thanks will need time to digest and execute.  Appreciate it

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________
From: Danny Ryan ***@***.***>
Sent: Monday, October 17, 2022 8:35:45 AM
To: ethereum/consensus-specs ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [ethereum/consensus-specs] 4844 spec updates (Issue #3043)


A number of spec modifications need to happen after discussions at devcon:

  1.  Couple block and blob sidecar in a single gossip topic -- beacon_block_and_blob_sidecar
     *   Keep proposer signature on the sidecar to ensure a party cannot mutate OR we have to do the sidecar / blobtx KZG verifications
  2.  Reduce MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS to ~2 weeks or lower
  3.  Clarify Blob availability requirement in fork-choice by specifying an is_data_available() check for importing a block
  4.  TBD -- potentially reduce blobs per block down to an appropriate value after testing and simulations
     *   Note, this can be done now to something like 128KB/256KB if 1MB/2MB is blocking tests
[...]

PR #3046, opened 2022-10-17
EIP4844: couple beacon block and blob sidecar for p2p

terencechain:
Part of #3043

Here we couple beacon block and blob sidecar under a single gossip topic- `beacon_block_and_blob_sidecar`. We introduce a new network container `SignedBeaconBlockAndBlobsSidecar` which consists of `SignedBeaconBlock` and `SignedBlobsSidecar`. We also combined the previous gossips `beacon_block` and `blobs_sidecar` validations under `beacon_block_and_blob_sidecar`. I think we can get rid of `blob.beacon_block_root` and `blob.beacon_block_slot`, I'm curious to hear more thoughts

cc @protolambda @Inphi 

Rationale for coupling: https://notes.ethereum.org/RLOGb1hYQ0aWt3hcVgzhgQ?#Gossip

### DIFF
--- a/specs/eip4844/p2p-interface.md
+++ b/specs/eip4844/p2p-interface.md
@@ -113,7 +113,7 @@ Alias `signed_blobs_sidecar = signed_beacon_block_and_blobs_sidecar.blobs_sideca
 - _[IGNORE]_ The sidecar is the first sidecar with valid signature received for the `(proposer_index, sidecar.beacon_block_slot)` combination,
   where `proposer_index` is the validator index of the beacon block proposer of `sidecar.beacon_block_slot`
 
-Once both sidecar and beacon block are received, `validate_blobs_sidecar` can unlock the data-availability fork-choice dependency.
+Once the sidecar and beacon block are received together, `validate_blobs_sidecar` can unlock the data-availability fork-choice dependency.
 
 ### Transitioning the gossip
 
--- a/specs/eip4844/p2p-interface.md
+++ b/specs/eip4844/p2p-interface.md
@@ -14,11 +14,11 @@ The specification of these changes continues in the same format as the network s
   - [Containers](#containers)
     - [`BlobsSidecar`](#blobssidecar)
     - [`SignedBlobsSidecar`](#signedblobssidecar)
+    - [`SignedBeaconBlockAndBlobsSidecar`](#signedbeaconblockandblobssidecar)
   - [The gossip domain: gossipsub](#the-gossip-domain-gossipsub)
     - [Topics and messages](#topics-and-messages)
       - [Global topics](#global-topics)
-        - [`beacon_block`](#beacon_block)
-        - [`blobs_sidecar`](#blobs_sidecar)
+        - [`beacon_block_and_blob_sidecar`](#beacon_block_and_blob_sidecar)
     - [Transitioning the gossip](#transitioning-the-gossip)
   - [The Req/Resp domain](#the-reqresp-domain)
     - [Messages](#messages)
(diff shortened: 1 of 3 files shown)


dapplion:
:+1: on coupling

specs/eip4844/p2p-interface.md
```diff
@@ -58,6 +58,14 @@ class SignedBlobsSidecar(Container):
     signature: BLSSignature
 ```
 
+### `SignedBeaconBlockAndBlobsSidecar`
+
+```python
+class SignedBeaconBlockAndBlobsSidecar(Container):
+    beacon_block: SignedBeaconBlock
+    blobs_sidecar: SignedBlobsSidecar
```
protolambda, line 66: Verifying the sidecar signature helps ensure the KZG commitments are worth verifying, but would it really result in DoS / performance issues when anyone changes it to invalid data? Technically we compare the sidecar against the KZG commitments in the beacon block already, so we do not have to sign the sidecar.

specs/eip4844/p2p-interface.md
```diff
@@ -75,33 +83,25 @@ The new topics along with the type of the `data` field of a gossipsub message ar
...
   -- i.e. `verify_kzg_commitments_against_transactions(block.body.execution_payload.transactions, block.body.blob_kzg_commitments)`
 
-##### `blobs_sidecar`
-
-This topic is used to propagate data blobs included in any given beacon block.
-
-The following validations MUST pass before forwarding the `signed_blobs_sidecar` on the network;
-Alias `sidecar = signed_blobs_sidecar.message`.
+Alias `signed_blobs_sidecar = signed_beacon_block_and_blobs_sidecar.blobs_sidecar`, `sidecar = signed_blobs_sidecar.message`.
 - _[IGNORE]_ the `sidecar.beacon_block_slot` is for the current slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) -- i.e. `sidecar.beacon_block_slot == current_slot`.
```
protolambda, line 105:
```suggestion
- _[IGNORE]_ the `sidecar.beacon_block_slot` is for the current slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) -- i.e. `sidecar.beacon_block_slot == block.slot`.
```

specs/eip4844/p2p-interface.md
```diff
@@ -113,8 +113,6 @@ Alias `sidecar = signed_blobs_sidecar.message`.
 - _[IGNORE]_ The sidecar is the first sidecar with valid signature received for the `(proposer_index, sidecar.beacon_block_slot)` combination,
   where `proposer_index` is the validator index of the beacon block proposer of `sidecar.beacon_block_slot`
 
-Note that a sidecar may be propagated before or after the corresponding beacon block.
-
 Once both sidecar and beacon block are received, `validate_blobs_sidecar` can unlock the data-availability fork-choice dependency.
```
protolambda, line 116:
```suggestion
Once the sidecar and beacon block are received together, `validate_blobs_sidecar` can unlock the data-availability fork-choice dependency.
```
Nit: if we couple them here then we should describe it as receiving the two together instead of independently.

specs/eip4844/p2p-interface.md
realbigsean: Yeah I don't think we need to sign the side car. This also informs whether or not we send signed or unsigned side cars in blobs by range/root request.

Inphi (commented):
I think we need some clarification on how we want to propagate beacon blocks that don't have sidecars (unlikely but possible). Do we keep using the `beacon_block` topic in that scenario? Or use the new topic from now one and set the `blobs_sidecar` to its default ssz value in that case?

specs/eip4844/p2p-interface.md
terencechain: Yeah, I agree that comparing KZG commitments in the beacon block is sufficient. We don't need the sidecar signature

@realbigsean range/root request will be decoupled and it's unsigned there

terencechain:
>

terencechain:
> I think we need some clarification on how we want to propagate beacon blocks that don't have sidecars 

Do we want to support this? It feels like more complexity and takes us back to de-coupling block and blob

realbigsean:
I think empty lists in the existing structs here should be sufficient

specs/eip4844/p2p-interface.md
```diff
@@ -14,11 +14,11 @@ The specification of these changes continues in the same format as the network s
   - [Containers](#containers)
     - [`BlobsSidecar`](#blobssidecar)
     - [`SignedBlobsSidecar`](#signedblobssidecar)
+    - [`SignedBeaconBlockAndBlobsSidecar`](#signedbeaconblockandblobssidecar)
   - [The gossip domain: gossipsub](#the-gossip-domain-gossipsub)
     - [Topics and messages](#topics-and-messages)
       - [Global topics](#global-topics)
-        - [`beacon_block`](#beacon_block)
-        - [`blobs_sidecar`](#blobs_sidecar)
+        - [`beacon_block_and_blob_sidecar`](#beacon_block_and_blob_sidecar)
```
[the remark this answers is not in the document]
terencechain: Typo, I will fix. Thanks!

Inphi:
> 



> > I think we need some clarification on how we want to propagate beacon blocks that don't have sidecars
> 
> Do we want to support this? It feels like more complexity and takes us back to de-coupling block and blob

Agreed. it isn't clear to me from my reading of the spec that we'll use coupled blocks even when there isn't a sidecar. Let's clarify this in the spec. We can deprecate/remove the old topic used for (decoupled) beacon blocks.

hwwhww:
should we add the description of how to prepare `SignedBeaconBlockAndBlobsSidecar` in validator.md?

terencechain:
> should we add the description of how to prepare `SignedBeaconBlockAndBlobsSidecar` in validator.md?

Good point, I'll update that today

tests/core/pyspec/eth2spec/test/eip4844/unittests/validator/test_validator.py
```diff
@@ -38,8 +37,6 @@ def _run_validate_blobs_sidecar_test(spec, state, blob_count):
     state_transition_and_sign_block(spec, state, block)
 
     blobs_sidecar = spec.get_blobs_sidecar(block, blobs)
-    privkey = privkeys[1]
-    spec.get_signed_blobs_sidecar(state, blobs_sidecar, privkey)
```
hwwhww, line 42: ah right, it should have been in another test case + verification 😅

hwwhww (approved):
lgtm!

PR #3047, opened 2022-10-17
EIP4844: modify `MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS` to 18 days

terencechain:
Part of #3043

As discussed during Devcon, 1.2 months is a bit long as the upper bound for storing the blobs. 2 weeks seems sufficient for any use case we are aware of, and some may argue even shorter. This reduces the storage requirement to 137GB, assuming 1mb target


Joumax11:
I agree i'll remover one

Joumax11:
Close

Issue #3020, opened 2022-09-30
Capella: Validator index in Withdrawal?


mcdee:
Looking at the `Withdrawal` type, is there value to having the validator index in here?  At current there doesn't seem to be a way of tying a withdrawal to a particular validator (even if the beacon state root were available in the EVM) so this would provide that linkage.

djrtwo:
I'm not opposed. It comes at a cost to block size but other than that seems safe/fine.

The current idea is that the `withdrawal_index` can be tracked when created for the validator but that's not very offline-proof

hwwhww:
close via #3050

PR #3050, opened 2022-10-18
Add validator_index to Withdrawal

mcdee:
As per #3020


dapplion:
Would be good to check in with explorers and other data consumers to understand the real value of adding this property.

hwwhww (approved):
Thanks @mcdee! I made a minor change in the test & fix the irrelevant lint error. LGTM!

PR #3052, opened 2022-10-20
Rebase EIP-4844 on Capella

Inphi:
Also adds a "testing" specification that allows withdrawals to be disabled during testing.


specs/eip4844/beacon-chain.md
```diff
@@ -238,9 +273,32 @@ def process_execution_payload(state: BeaconState, payload: ExecutionPayload, exe
         excess_blobs=payload.excess_blobs,  # [New in EIP-4844]
         block_hash=payload.block_hash,
         transactions_root=hash_tree_root(payload.transactions),
-    )
+        withdrawals_root=hash_tree_root(payload.withdrawals) if ENABLE_WITHDRAWALS else Bytes32(),  # [New in EIP-4844]
```
hwwhww, line 276:
```suggestion
        withdrawals_root=hash_tree_root(payload.withdrawals) if ENABLE_WITHDRAWALS else Bytes32(),  # [New in EIP-4844]
    )
```

specs/eip4844/beacon-chain.md
```diff
@@ -238,9 +273,32 @@ def process_execution_payload(state: BeaconState, payload: ExecutionPayload, exe
...
+        for operation in operations:
+            fn(state, operation)
+
+    for_ops(body.proposer_slashings, process_proposer_slashing)
+    for_ops(body.attester_slashings, process_attester_slashing)
+    for_ops(body.attestations, process_attestation)
+    for_ops(body.deposits, process_deposit)
+    for_ops(body.voluntary_exits, process_voluntary_exit)
+    if ENABLE_WITHDRAWALS:  # [New in EIP-4844]
+        for_ops(body.bls_to_execution_changes, process_bls_to_execution_change)
```
hwwhww, line 298: That means we have to replace `@with_capella_and_later` decorator in tests with `@with_phases([CAPELLA])` if we treat `ENABLE_WITHDRAWALS == 0` as a constant.

And if we want to have the flexibility to test `ENABLE_WITHDRAWALS == 1`, we need to write a new decorator to check preset before running the test.

specs/eip4844/fork.md
```diff
@@ -62,12 +64,11 @@ Note that for the pure EIP-4844 networks, we don't apply `upgrade_to_eip4844` si
...
 ### Upgrading the state
 
-Since the `eip4844.BeaconState` format is equal to the `bellatrix.BeaconState` format, we only have to update `BeaconState.fork`.
+Since the `eip4844.BeaconState` format is equal to the `Capella.BeaconState` format, we only have to update `BeaconState.fork`.
 
 ```python
-def upgrade_to_eip4844(pre: bellatrix.BeaconState) -> BeaconState:
-    # TODO: if Capella gets scheduled, add sync it with Capella.BeaconState
-    epoch = bellatrix.get_current_epoch(pre)
+def upgrade_to_eip4844(pre: Capella.BeaconState) -> BeaconState:
```
hwwhww, line 70: It's lowercase `capella` for the pyspec library.

hwwhww (commented):
Had a quick scan, @Inphi well done on the rebasing!  

- Note that #3042 changed the consensus containers & epoch processing functions. So if we include it, you'd need to update that part.
- From pyspec's point of view, having `ENABLE_WITHDRAWALS` is kinda annoying 😬  but I'd defer to client devs if they really want this flag control in spec.

terencechain:
> Had a quick scan, @Inphi well done on the rebasing!
> 
>     * Note that #3042 changed the consensus containers & epoch processing functions. So if we include it, you'd need to update that part.
> 
>     * From pyspec's point of view, having `ENABLE_WITHDRAWALS` is kinda annoying 😬  but I'd defer to client devs if they really want this flag control in spec.

Ditto. I don't think we should enshrine test-only flag in the spec. The decision on whether withdrawal is enabled or in the protocol for testing and testnet can be reached out of the band

specs/eip4844/beacon-chain.md
```diff
@@ -97,6 +104,7 @@ class BeaconBlockBody(Container):
     sync_aggregate: SyncAggregate
     # Execution
     execution_payload: ExecutionPayload 
+    bls_to_execution_changes: List[SignedBLSToExecutionChange, MAX_BLS_TO_EXECUTION_CHANGES]
```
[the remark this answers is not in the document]
realbigsean: Otherwise we're gonna have to create new objects that will never exist (an eip4844 block minus withdrawals)

Inphi:
The goal is to align the behavior of disabling withdrawals across all clients. ENABLE_WITHDRAWALS will be removed eventually upon further testing. But right now, we're trying to avoid any withdrawal bugs or changes in the spec from influencing 4844.

specs/eip4844/beacon-chain.md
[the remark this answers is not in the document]
Inphi: yeah we wanna retain the fields to make it easy for client devs to branch off their existing withdrawals implementation. I expect little to no bugs/changes to that implementation and specification.

Only thing we wanna avoid is triggering state transition logic for withdrawals since that's where they could most likely be different behaviors in client implementations.

realbigsean:
In testnets we can avoid worrying too much about the impact of withdrawals on consensus by stubbing out these functions:

Epoch processing:
-`process_full_withdrawals`
-`process_partial_withdrawals`

Block processing:
- `process_withdrawals`
- `for_ops(body.bls_to_execution_changes)`

But I think we can figure this out on a per-testnet basis, by saying "target this spec commit but stub out these functions" rather than ingraining this in the spec here

I think it's simpler if we don't touch anything else, for example:
- Actually get the root of an empty withdrawals list
- Keep an empty withdrawals list in payload attributes
- Use EIP4844 + withdrawal types in all APIs

Inphi:
@realbigsean Thanks for the suggestions. I've made the changes to the spec. `ENABLE_WITHDRAWALS` has been removed and updated the Testing section to specify how clients can disable withdrawals.

hwwhww (approved):
I managed to make withdrawal functions no-op & add basic tests, and find okay Wi-Fi somewhere 2km far from Machu Picchu. 😄

The functionalities/correctness LGTM now!

Edited: I'll defer to client devs to make the decision in testing. Although I personally found that the no-op stub is still somewhat annoying. 😅

djrtwo (changes requested):
#3068 is merged. This is ready for a re-rebase @Inphi

Inphi:
I've updated the PR for #3068. Reviewers please take another look.

specs/eip4844/p2p-interface.md
```diff
@@ -140,6 +141,7 @@ Per `context = compute_fork_digest(fork_version, genesis_validators_root)`:
 | `GENESIS_FORK_VERSION`   | `phase0.SignedBeaconBlock` |
 | `ALTAIR_FORK_VERSION`    | `altair.SignedBeaconBlock` |
 | `BELLATRIX_FORK_VERSION` | `bellatrix.SignedBeaconBlock` |
+| `CAPELLA_FORK_VERSION`   | `capella.SignedBeaconBlock`   |
```
terencechain, line 144: Potential merge conflict with #3089. I can fix that after this is merged

specs/eip4844/validator.md
```diff
@@ -29,25 +31,21 @@ This document represents the changes to be made in the code of an "honest valida
...
 
-The interface to retrieve blobs and corresponding kzg commitments.
+### `ExecutionEngine`
 
-Note: This API is *unstable*. `get_blobs_and_kzg_commitments` and `get_payload` may be unified.
-Implementers may also retrieve blobs individually per transaction.
+#### `get_payload`
 
-```python
-def get_blobs_and_kzg_commitments(payload_id: PayloadId) -> Tuple[Sequence[BLSFieldElement], Sequence[KZGCommitment]]:
```
terencechain, line 48: `get_blobs_and_kzg_commitments` is still mentioned below, do we want to remove this?

terencechain (commented):
Great work @Inphi ! Just one question

specs/eip4844/validator.md
Inphi: Good catch. This was a bad merge from dev. I've added it back in. Thanks.

hwwhww (approved):
lgtm! :shipit:
 
I have a test vector generator branch (https://github.com/ethereum/consensus-specs/compare/pr3052-testgen) based on this PR. I'll open a PR once this one is merged.

specs/eip4844/beacon-chain.md
```diff
@@ -136,6 +139,7 @@ class ExecutionPayloadHeader(Container):
     # Extra payload fields
```
ralexstokes, line 139: also, `excess_blobs` has become `excess_data_gas` with `uint256` type

this should be addressed here and in the `ExecutionPayload`

Inphi: Thanks. Fixed.

specs/eip4844/beacon-chain.md
```diff
@@ -335,3 +342,10 @@ def initialize_beacon_state_from_eth1(eth1_block_hash: Hash32,
 
     return state
 ```
+
+### Disabling Withdrawals
+During testing we avoid Capella-specific updates to the state transition. We do this by replacing the following functions with a no-op implementation:
```
djrtwo, line 347: This no-op works fine for clients?
I'm out of the loop, but I thought they would use the full functionality here (yes, complicating testing)

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_bls_to_execution_change.py
```diff
@@ -37,14 +38,14 @@ def run_bls_to_execution_change_processing(spec, state, signed_address_change, v
...
-@with_capella_and_later
+@with_phases([CAPELLA])
 @spec_state_test
 def test_success(spec, state):
     signed_address_change = get_signed_address_change(spec, state)
     yield from run_bls_to_execution_change_processing(spec, state, signed_address_change)
 
 
-@with_capella_and_later
+@with_phases([CAPELLA])
```
djrtwo, line 48: This should be capella and later, no?

Ah, the no-op change will break this test... If we do this, we need to remember to go back and change these all :/
When we were rebasing here, I didn't realize we were rebasing and mutating. How much will we be blocked on 4844 if we don't do the capella no-op within 4844 spec?

hwwhww: My understanding is to prevent random semi-public testnet users/validators from invoking any withdrawals?

Agreed with this complicating testing but given that we have 3-4 in-discussion changes on Capella, it makes sense that EIP-4844 devs want to disable Capella now... 😭

Btw I also added the tests to test the Capella no-op. So we must undo it altogether.

specs/eip4844/beacon-chain.md
Inphi: We want to avoid submitted withdrawals from interfering with the testnet. By no-op'ing these functions, withdrawals won't have any effect.

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_bls_to_execution_change.py
Inphi: > My understanding is to prevent random semi-public testnet users/validators from invoking any withdrawals?

Yes it's exactly this.

specs/eip4844/p2p-interface.md
hwwhww: fixed in https://github.com/ethereum/consensus-specs/pull/3052/commits/f1d4c9047a70c3a321cb0329216d446ead17e007

djrtwo (approved):
okay, I'm fine with it, assuming that clients can no-op functions for a fork without too much issue.

approved

PR #3064, opened 2022-10-27
Consistently use `compute_sync_committee_period_at_slot`

etan-status:
A few LC functions were not yet updated to use a more concise function for computing sync committee period for a slot. Updating to that func.


dapplion (approved):

ralexstokes (approved):
lgtm!

PR #3065, opened 2022-10-27
Clarify where LC proofs are rooted

etan-status:
Add more detailed LC object documentation to explain that the various merkle proofs are relative to the beacon block's state root. Likewise, clarify that sync committees relate to the finalized header (not to the optimistic header, which can be a period ahead).

### DIFF
--- a/specs/altair/light-client/sync-protocol.md
+++ b/specs/altair/light-client/sync-protocol.md
@@ -77,9 +77,9 @@ Additional documents describe how the light client sync protocol can be used:
 
 ```python
 class LightClientBootstrap(Container):
-    # The requested beacon block header
+    # Header matching the requested beacon block root
     header: BeaconBlockHeader
-    # Current sync committee corresponding to `header`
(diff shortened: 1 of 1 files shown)


specs/altair/light-client/sync-protocol.md
```diff
@@ -77,9 +77,9 @@ Additional documents describe how the light client sync protocol can be used:
 
 ```python
 class LightClientBootstrap(Container):
-    # The requested beacon block header
+    # Header matching the requested beacon block root
     header: BeaconBlockHeader
-    # Current sync committee corresponding to `header`
+    # Current sync committee corresponding to `header.state_root`
```
dapplion, line 82: Could be even more clear with something like
```suggestion
    # Current sync committee of state with hash `header.state_root`
```

dapplion (approved):
LGTM, this is helpful to implementors

ralexstokes (approved):
nice updates!

PR #3066, opened 2022-10-27
Allow extending LC merkle proof tests

etan-status:
Currently, `test_single_merkle_proof` only supports `BeaconState` tests. For future tests, different object classes are desirable. Update format to allow testing other objects as well.

### DIFF
--- a/tests/formats/light_client/single_merkle_proof.md
+++ b/tests/formats/light_client/single_merkle_proof.md
@@ -5,24 +5,30 @@ generation and verification of merkle proofs based on static data.
 
 ## Test case format
 
-### `state.ssz_snappy`
+### `meta.yaml`
 
-An SSZ-snappy encoded `BeaconState` object from which other data is generated.
+```yaml
+object_class: string  -- 'BeaconState'
+```
+
+### `object.yaml`
+
+A SSZ-snappy encoded object of type `object_class` from which other data is generated.
 
 ### `proof.yaml`
 
-A proof of the leaf value (a merkle root) at generalized-index `leaf_index` in the given `state`.
+A proof of the leaf value (a merkle root) at generalized-index `leaf_index` in the given `object`.
 
 ```yaml
 leaf: Bytes32            # string, hex encoded, with 0x prefix
 leaf_index: int          # integer, decimal
-branch: list of Bytes32  # list, each element is a string, hex encoded, with 0x prefix 
+branch: list of Bytes32  # list, each element is a string, hex encoded, with 0x prefix
 ```
 
 ## Condition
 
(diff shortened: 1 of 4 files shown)


tests/formats/light_client/single_merkle_proof.md
```diff
@@ -5,24 +5,30 @@ generation and verification of merkle proofs based on static data.
 
 ## Test case format
 
-### `state.ssz_snappy`
+### `meta.yaml`
 
-An SSZ-snappy encoded `BeaconState` object from which other data is generated.
+```yaml
+object_class: string  -- 'BeaconState'
```
dapplion, line 11: Instead of "object_class" should it be called "type_name" or "container_name"? More consistent with the notation here https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#custom-types

ralexstokes:
what types of tests do you have in mind?

this approach isn't impossible but in general we would handle this kind of thing by making a new type of test "handler" that has the exact format we want -- that way consumers of the test data have less dynamism at runtime, which is desirable for all of the compiled language targets that consume these tests

etan-status:
Using a different subfolder for each type is indeed cleaner, and more in line with e.g., ssz_static tests.
Changed it to use separate `suite` for now (within the `single_merkle_proof` handler, within the `light_client` runner).

As for future tests, I'd like to add one for `BeaconBlockBody` as well.

tests/formats/light_client/single_merkle_proof.md
etan-status: Moved it to folder name instead, avoids the variable name :)

dapplion (approved):
LGTM

hwwhww (approved):
lgtm 👍

If we want to extend the coverage of Merkle proof tests, we can consider reworking the test generator `main.py` to something similar to [ssz_static](https://github.com/ethereum/consensus-specs/blob/dev/tests/generators/ssz_static/main.py) rather than using pytest + `@spec_state_test`. Do you think that would be helpful?

etan-status:
Currently, the idea for the LC merkle proof tests is to test for very specific cases (the 3-4 generalized indices necessary for the LC sync protocol) to also allow testing concrete implementations that only support what is actually needed over a generalized SSZ library. So, the tests are much more manual than something that can be generated dynamically.

The other difference is that the LC tests are all generated as part of `make gen_light_client`, so the output is one folder deeper as it's not a separate generator.

For now, would keep it as a LC test and do only the minimal prep work to allow tests against BlockBody on top of State. If we need more generalized testing in the future, it's a point that can be re-evaluated when there is a need for it, though.

PR #3067, opened 2022-10-27
Rename LC test check from `root` to `beacon_root`

etan-status:
Future light client tests will also incorporate execution payload data. To avoid confusion, rename the current `root` check to `beacon_root`. Doing this now, as #3066 already requires LC test runners to update.


dapplion (approved):
LGTM

ralexstokes (approved):
works for me

Issue #3054, opened 2022-10-22
`withdraw_balance` decreases up to the validator's balance but produces a receipt up to the parameter


potuz:
This is not an issue in the runtime spec as checks for underflows are performed. However I think it's bad practice and it may probably be worth to clean the fact that in `withdraw_balance`

```python
def withdraw_balance(state: BeaconState, validator_index: ValidatorIndex, amount: Gwei) -> None:
    # Decrease the validator's balance
    decrease_balance(state, validator_index, amount)
    # Create a corresponding withdrawal receipt
    withdrawal = Withdrawal(
        index=state.next_withdrawal_index,
        address=ExecutionAddress(state.validators[validator_index].withdrawal_credentials[12:]),
        amount=amount,
    )
 ```
 If this function is called with `amount > state.balances[validator_index]` then the validator's balance would be zeroed but the receipt will be emitted for `amount` which will trigger the operation in the execution layer to refund the validator with more than it's beacon-chain balance.

PR #3068, opened 2022-10-28
Withdrawals without queues

potuz:
This PR implements withdrawals without any withdrawal queue. I want to explore the space for solutions like this since in principle seems to be much simpler than the current logic so I may be missing something and I'm using this space to find out what's wrong. 

This solution adds only one field `latest_withdrawal_validator_index` to the BeaconState, that keeps track of the last validator that has performed a withdrawal. Proposers pack up to `MAX_WITHDRAWALS_PER_PAYLOAD` in their execution blocks starting from the next validator after `latest_withdrawal_validator_index`. 

During block processing, we simply check that the the next validators that can withdraw agree with the ones packed in the block, and that their withdrawal amount agree with the ones we expect from the state. 

There is no epoch processing logic as there is no queue to include withdrawals. 

This also removes unnecessary constants like `MAX_WITHDRAWALS_PER_EPOCH` and `WITHDRAWAL_QUEUE_LIMIT`

I suppose there can be a problem with validators being slashed right after doing a partial withdrawal. This can only be a bad issue for  validators that get slashed for more than `MAX_EFFECTIVE_BALANCE` which is very rare, another issue is that a validator may end up with a slightly bellow `MAX_EFFECTIVE_BALANCE` right after performing a partial withdrawal, but I don't see anything seriously wrong with this either. 

Fixes: #3054


specs/capella/beacon-chain.md
```diff
@@ -108,7 +78,7 @@ We define the following Python custom types for type hinting and readability:
 
 ```python
 class Withdrawal(Container):
-    index: WithdrawalIndex
```
ralexstokes, line 111: `index` is useful to discriminate withdrawals at the EL and any extra-protocol infra that tracks them -- can we leave this in?

specs/capella/beacon-chain.md
```diff
@@ -108,7 +78,7 @@ We define the following Python custom types for type hinting and readability:
...
+    validator_index: ValidatorIndex
```
ralexstokes, line 81: if you rebase against the latest `dev`, this field was added already

specs/capella/beacon-chain.md
```diff
@@ -371,15 +272,29 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
-    state.withdrawal_queue = state.withdrawal_queue[num_withdrawals:]
+    epoch = get_current_epoch(state)
+    index = state.last_withdrawal_validator_index
+    for withdrawal in payload.withdrawals:
+        index = ValidatorIndex((index + 1) % len(state.validators))
+        while index != withdrawal.validator_index:
+            val state.validators[index]
+            balance = state.balances[index]
+            assert !is_fully_withdrawable_validator(val, balance, epoch):
+            assert !is_partially_withdrawable_validator(val, balance):
```
ralexstokes, line 283:
```suggestion
            assert not is_fully_withdrawable_validator(val, balance, epoch)
            assert not is_partially_withdrawable_validator(val, balance)
```

specs/capella/beacon-chain.md
```diff
@@ -371,15 +272,29 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+            index += ValidatorIndex((index + 1) % len(state.validators))
+        assert withdrawal.validator_index == index
+        val = state.validators[index]
+        balance = state.balances[index]
+        fully_withdrawable = is_fully_withdrawable_validator(val, balance, epoch)
+        partial_withdrawable = is_partially_withdrawable_validator(val, balance)
+        assert fully_withdrawable || partial_withdrawable
```
ralexstokes, line 290:
```suggestion
        assert fully_withdrawable or partial_withdrawable
```

specs/capella/fork.md
```diff
@@ -112,9 +112,7 @@ def upgrade_to_capella(pre: bellatrix.BeaconState) -> BeaconState:
         # Execution-layer
         latest_execution_payload_header=pre.latest_execution_payload_header,
         # Withdrawals
-        withdrawal_queue=[],
-        next_withdrawal_index=WithdrawalIndex(0),
-        next_withdrawal_validator_index=ValidatorIndex(0),
+        last_withdrawal_validator_index=ValidatorIndex(0),
```
ralexstokes, line 115: naming-wise, I think we could just keep `next_withdrawal_index` and use this symbol instead of `last_withdrawal_validator_index`

specs/capella/beacon-chain.md
```diff
@@ -371,15 +272,29 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+        if fully_withdrawable:
+            assert withdrawal.amount == balance
+            decrease_balance(state, index, balance)
+        else:
+            assert withdrawal.amount == balance - MAX_EFFECTIVE_BALANCE
```
ralexstokes, line 295: you'd also want to check that the `withdrawal.address` matches the validator's withdrawal credentials

specs/capella/beacon-chain.md
```diff
@@ -371,15 +272,29 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
-    for dequeued_withdrawal, withdrawal in zip(dequeued_withdrawals, payload.withdrawals):
-        assert dequeued_withdrawal == withdrawal
-
-    # Remove dequeued withdrawals from state
-    state.withdrawal_queue = state.withdrawal_queue[num_withdrawals:]
+    epoch = get_current_epoch(state)
+    index = state.last_withdrawal_validator_index
+    for withdrawal in payload.withdrawals:
+        index = ValidatorIndex((index + 1) % len(state.validators))
+        while index != withdrawal.validator_index:
```
ralexstokes, line 279: it seems like this is reintroducing a linear scan of the validator set, which I thought we wanted to remove?

ralexstokes (commented):
added some comments to consider

specs/capella/beacon-chain.md
```diff
@@ -371,15 +272,29 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
-
-    assert len(dequeued_withdrawals) == len(payload.withdrawals)
-    for dequeued_withdrawal, withdrawal in zip(dequeued_withdrawals, payload.withdrawals):
-        assert dequeued_withdrawal == withdrawal
-
-    # Remove dequeued withdrawals from state
-    state.withdrawal_queue = state.withdrawal_queue[num_withdrawals:]
+    epoch = get_current_epoch(state)
+    index = state.last_withdrawal_validator_index
+    for withdrawal in payload.withdrawals:
```
realbigsean, line 277: Rather than iterating through payloads given in the block, I think using the same method as in block production would be smart (`get_expected_withdrawals`). To make sure for a given state, the correct set of withdrawals is chosen, and we can just check equivalence. It seems like it'd be less error prone if we used it block production and processing

specs/capella/beacon-chain.md
```diff
@@ -371,15 +272,29 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+        index = ValidatorIndex((index + 1) % len(state.validators))
+        while index != withdrawal.validator_index:
+            val state.validators[index]
```
michaelsproul, line 280: missing an `=` here
```suggestion
            val = state.validators[index]
```

specs/capella/beacon-chain.md
michaelsproul: currently `get_expected_withdrawals` is specified in terms of the withdrawal queue, so it would need to be re-phrased, likely as an iteration over the validator set (which clients then need to optimize out with a cache).

potuz: Yeah I want to explicitly leave this out of the spec and let clients do whatever they want to optimize this scan. Unless the network is leaking for 5 straight days, it should be trivial to find the next 16 withrawals. Anyway a circular buffer cache of fixed size would most probably be our implementation if we go with this. I'll clean the PR up this week. Many thanks for the reviews

specs/capella/beacon-chain.md
potuz: This linear scan succeeds in very few iterations unless the whole network is leaking for over 5 days. Clients are free to cache this however they want. It's very different than being forced to loop through the whole set

specs/capella/beacon-chain.md
michaelsproul: Awesome.

I think it doesn't hurt to also define the semantics of `get_expected_withdrawals` in an inefficient way in the spec. This would likely be required for generating and executing spec tests anyway.

rkapka:
This feels simpler than the queue approach, which is why I am in favor of this change. We are at such an early stage with withdrawals (at least judging from https://notes.ethereum.org/@ralexstokes/SyDqVjr4i#CL-client-implementation), that major changes like this are OK.

specs/capella/fork.md
potuz: I didn't understand this one, I added back now `next_withdrawal_index` as a counter that just simply counts withdrawals since the fork. `last_withdrawal_validator_index` can't be recovered from this and is needed to know from where to start sweeping.

specs/capella/beacon-chain.md
```diff
@@ -381,19 +280,54 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+#### New `get_expected_withdrawals`
+
+```python
+def get_expected_withdrawals(state: BeaconState) -> Sequence[Withdrawal]:
+    epoch = get_current_epoch(state)
+    withdrawal_index = state.next_withdrawal_index
+    index = ValidatorIndex((state.last_withdrawal_validator_index + 1) % len(state.validators))
+    ret: List[Withdrawal] = []
+    probed = 0
+    while (len(ret) < MAX_WITHDRAWALS_PER_PAYLOAD) and (probed < len(state.validators)):
```
g11tech, line 292: would be nice to rewrite this as a `for` over `len(state.validators)` and break when we have `MAX_WITHDRAWALS_PER_PAYLOAD`

specs/capella/beacon-chain.md
```diff
@@ -381,19 +280,54 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+                amount=balance,
+            )
+            ret.append(withdrawal)
+            withdrawal_index = WithdrawalIndex(withdrawal_index + 1)
+        elif is_partially_withdrawable_validator(val, balance):
+            withdrawal = Withdrawal(
+                index=withdrawal_index,
+                validator_index=index,
+                address=ExecutionAddress(val.withdrawal_credentials[12:]),
+                amount=balance - MAX_EFFECTIVE_BALANCE,
```
g11tech, line 309: may be there can be some `margin` balance that can still be left i.e. have `MAX_EFFECTIVE_BALANCE` + `MARGIN` as the minimum balance to maintain, this would resolve the issues with effectie balance droppin as the staker might not want his/her effective balance to drop by a step function

g11tech:
I actually love the simplicity of this, there is no point popping the withdrawals into state queue and then later process it at block. This will also reduce/remove pressure on epoch processing, and the lookups for finding withdrawals for a block are small  since per block limit is small (assuming high success rate of finding a valid withdrawal). The two step withdrawal isn't giving any advantage.

the full having priority over partial can still be addressed in this mode as well, but only depends on what UX we want to go for.

so  :+1:  from lodestar on this approach overall.

specs/capella/beacon-chain.md
potuz: Yeah this is something that @mcdee requested. I'm not against it, but I am cautious cause any value may be too high in the future and will require a hard fork to change.

ajsutton:
I'm in two minds about this approach. I really like the simplicity but I'm a bit nervous about the potential need to have to scan a lot of validators to find the required number of withdrawals given that has to happen on every block import. My concern there is two-fold:

1. 99.9% of the time we won't be scanning far which means the potential performance issues only show up in rare situations - typically when there are already other issues going on.
2. The mitigation for those performance problems involves introducing and keeping in sync separate caches which significantly reduces testability and adds a lot of complexity. I was keen on not having the separate loops for full vs partial withdrawals primarily to avoid the need to introduce caches tracking the fully withdrawable validators which are unlikely to be fully exercised by spec tests. The optimisations to avoid a full scan in this case have similar testability issues.

One option to resolve this would be to simple require scanning a fixed number of validators rather than requiring a fixed number of withdrawals be included. ie block 0 includes any withdrawals available for validators 0-3, block 1 withdrawals for validators 4-8 etc. We may not use the full withdrawal capacity but given there's at least some desire to have a minimum withdrawable amount, going at a fixed rate round the validators doesn't seem like an issue.

potuz:
> I'm in two minds about this approach. I really like the simplicity but I'm a bit nervous about the potential need to have to scan a lot of validators to find the required number of withdrawals given that has to happen on every block import. 

I thought about simply changing the `for` loop to have a fixed bound which we can impose right now and I wouldn't be too opposed to it. However, I find that the situation when the scan is not immediate only happens if the network has been leaking for about 5 days (and this gets longer and longer each day), and if that happens, no matter what withdrawal mechanism we have, we will be in trouble anyway and this one does not make it particularly worse since it's a scan that you can perform after block processing during the beacon's "dead time" and in parallel,  as opposed to during epoch transition. The biggest problem with adding a bound is that we do not reuse validator indices, and this means that the bound is actually less effective over time and even during happy times we will start to see decreased capacity.

ajsutton:
>However, I find that the situation when the scan is not immediate only happens if the network has been leaking for about 5 days (and this gets longer and longer each day), and if that happens, no matter what withdrawal mechanism we have, we will be in trouble anyway and this one does not make it particularly worse since it's a scan that you can perform after block processing during the beacon's "dead time" and in parallel, as opposed to during epoch transition. 

Having to perform the scan separate to block transition is the complexity I'd like to avoid - and it will generally only be done for what's seen as the canonical chain so may be invalidated by a late block etc.  Having sub-optimal processing kick in when the chain hasn't been finalising for 5 days is my nightmare scenario as it increases the risk of cascading problems.  I'd actually prefer to have to scan large spans regularly rather than having it be so rare as it would then be better tested and optimised.

Also while this particular case probably isn't an issue even if we have to do a full scan, we should be really cautious about depending on caches to ensure good performance. If the cache-miss scenario is noticeably slower it can open up attack scenarios where the attacker releases a block late or something like that to deliberately trigger cache misses (and missing multiple different caches can combine in lots of complex ways to cause problems). Hence things like gas prices are always set based on worst case scenarios where caches are missed.

> The biggest problem with adding a bound is that we do not reuse validator indices, and this means that the bound is actually less effective over time and even during happy times we will start to see decreased capacity.

Given the current requests are to slow down the rate of withdrawals I'm not worried about this. At some point we likely need to start reusing validator indices (or find some other way of dropping exited validators) just to bound the growth of the validator set anyway. And if we were unhappy with reduced capacity before that happened it would be easy to say we'd scan until we've either found 4 withdrawals or scanned 50 validators. I'm not against having that dual limit now if we wanted to future proof but I don't think we need it.

potuz:
> Having to perform the scan separate to block transition is the complexity I'd like to avoid - and it will generally only be done for what's seen as the canonical chain so may be invalidated by a late block etc. Having sub-optimal processing kick in when the chain hasn't been finalising for 5 days is my nightmare scenario as it increases the risk of cascading problems. I'd actually prefer to have to scan large spans regularly rather than having it be so rare as it would then be better tested and optimised.
> 
> Also while this particular case probably isn't an issue even if we have to do a full scan, we should be really cautious about depending on caches to ensure good performance. If the cache-miss scenario is noticeably slower it can open up attack scenarios where the attacker releases a block late or something like that to deliberately trigger cache misses (and missing multiple different caches can combine in lots of complex ways to cause problems). Hence things like gas prices are always set based on worst case scenarios where caches are missed.
> 
FWIW, the scan is so trivial in the non-leaking case that the way we plan to implement this to simply perform it after block operations and after any change of head. No need to optimize anything. If the network hasn't been finalizing for 5 days. The penalty is a full linear scan on each block indeed, but the current system is not clear that it is better: you still rely on reorg resistant caches for the full queues, you need to be efficient in implementing the queue itself to avoid memory fragmentation or even leaking it. It is the very first time we add a queue to a consensus object. While it's true that one can envision that the nightmare scenario is bad, a general rule is that protocol complexity has many more edge and untested cases, and there's not doubt that the current specification, with a queue in the beacon state that can be very large or very small depending on the number of exiting validators, with separate logic for full withdrawals and partial withdrawals, with separate logic for epoch transition, etc. is far more complicated. 

> Given the current requests are to slow down the rate of withdrawals I'm not worried about this. At some point we likely need to start reusing validator indices (or find some other way of dropping exited validators) just to bound the growth of the validator set anyway. And if we were unhappy with reduced capacity before that happened it would be easy to say we'd scan until we've either found 4 withdrawals or scanned 50 validators. I'm not against having that dual limit now if we wanted to future proof but I don't think we need it.

With this I agree, we are very likely going to be needing to reuse validator indices. I'd be happy to see some expected values perhaps. If we think we will be reusing indices when we hit something like 50K exited validators, then capping the loop at 5K will keep it in the microseconds. 

Notice also that all clients are already performing a full validator set scan within forkchoice on each block insertion.

rolfyone:
limiting the loop to a tweakable constant would enforce a time restriction on how often withdrawals get processed, which would help ensure that we're not processing too fast as well, which would be positive....
I wouldn't be surprised if that number is low, like 16 ish... which would ensure we're not going through the entire validator set while we're leaking, it'd be about 3 days per full loop of 400k validators if all validators are leaking, and probably drop max_withdrawals_per_payload to something like 4 which would loop through at full inclusion in around 12 days for 400k validators... 8 would be roughly 6 days which is probably on the 'fast' side...
But doing this we're close to the '6 days for a full withdrawal is too long' discussion?

ajsutton:
Ok I ran some benchmarks and had a poke around the caches we already have and I don't think I'm worried anymore.

I was wary because while iterating balances is cheap in Teku, iterating the actual validator list is fairly expensive because of the way we maintain the state. Any balance above 32ETH must be withdrawable and any 0 balance isn't withdrawable, so the only time we have to access the actual validator without always getting a withdrawal out of it is when the balance is less than 32ETH. Most withdrawable validators will have a balance of 0 already and can be skipped. Thus the worst case scenario would be when all the active validators have leaked to have a balance below 32ETH. The timing on this isn't too bad and we can probably just wear the cost but if needed we already have a cache of active validators since that's needed for all kinds of things and that would let us very quickly skip over the vast majority of these additional checks without any extra complexity for maintaining a new cache.

Also iterating the full validator set is faster than I remember it anyway.

So I support this proposal.

paulhauner:
I am in support of this PR and I think it's a great iteration. Nice work @potuz :clap: Also, thanks to those who did prior iterations!

I think there are two scenarios in which to consider this PR:

1. During a healthy, finalizing network.
1. During an unhealthy, non-finalizing network.

In a healthy network, I like this PR because it adds fewer items to the state. It appears to be more "functional" and I like that for reasons that are probably subjective. There's an argument that it doesn't prioritize full withdrawals, but I don't feel strongly about that. I think prioritization can be a separate argument (i.e., this PR could also support prioritized full withdrawals in a subsequent PR).

When it comes to an *un*healthy network, I think I also prefer this PR. @ajsutton has done well to highlight how this PR suffers on an unhealthy network, but I believe that the `withdrawal_queue` (i.e. the status quo) suffers on an unhealthy network, too.

In the "Appendix" section, I describe how the `withdrawal_queue` has CPU (hashing) and memory (beacon state + hashing cache) downsides when doing "skip slots" across long distances.

As an argument *against* this PR, it has the unfortunate downside that, in some scenarios (e.g. inactivity leak), we must suffer an iteration across the validator set *for every block*. The `withdrawal_queue` suffers for *for each skipped epoch*, which I expect to be less frequent. That is, I propose that a penalty for each block is worse than an equal penalty for each skipped epoch.

However, the penalties suffered are not equal. The `withdrawal_queue` suffers CPU *and* memory penalties that are (a) challenging to comprehend and (b) at times significantly worse than a validator set iteration. This PR suffers only for CPU in a tightly-bounded and easy-to comprehend way.

Another thing about this PR is that I think the "iteration every block" problem is solvable. For example, we could attach to the state a `[uint64; MAX_WITHDRAWALS_PER_PAYLOAD * SLOTS_PER_EPOCH]` which contains an ordered list of validator indices that are available for withdrawal. When that list is empty or the last value equals `last_withdrawal_validator_index` then we know there are no more withdrawals to dequeue.

So, I think this PR is certainly a step in the right direction if not a candidate for production. I'm unsure about the latter at this point :thinking: 

### Appendix

This appendix describes my concerns about how the `withdrawal_queue` behaves during skip slots. Whilst "skipping slots" (i.e., running `process_slot` from `parent.slot` to `block.slot`) we *must* maintain the `withdrawals_queue` and compute its hash tree root for each epoch so that we may verify the `block.state_root`.

The rate at which the withdrawals queue can grow is a little unclear due to "unbounded" full withdrawals, however let's assume it grows at `MAX_PARTIAL_WITHDRAWALS_PER_EPOCH == 256` (for brevity, let's call this variable `WITHDRAW_PER_EPOCH`).

Running some benchmarks on my ~3 year old laptop ([i7-10510U](https://ark.intel.com/content/www/us/en/ark/products/196449/intel-core-i710510u-processor-8m-cache-up-to-4-90-ghz.html)), I'm seeing 0.4ms to compute the hash tree root of a `withdrawals_queue` of length 256.

As we know, it's possible for validators to produce blocks upon the finalized checkpoint and cause all honest clients to run skip-slots from the finalized checkpoint to the wall clock slot. Since `EPOCHS_PER_DAY = 225`, a day of skip slots will involve at least `90ms` (`225 * 0.4`) of hashing the `withdrawal_queue`. Notably, 90ms assumes perfect tree-hash caching. A cache-less implementation would end up at ~13s.

Along with this time cost, there's also a memory cost; a days worth of `Withdrawal` objects (44 bytes each) is `44 * EPOCHS_PER_DAY * WITHDRAW_PER_EPOCH ~= 2.5mb` added to the `BeaconState`. Assuming we're doing current-generation tree hash caching, we're looking at a cached tree of ~3.6mb (`((EPOCHS_PER_DAY * WITHDRAW_PER_EPOCH * 2) - 1) * 32`).

So, if we assume that there were 450k validators viable to enter the queue before non-finality started, we'll reach a maximum `withdrawal_queue` size of 450k within ~7.7 days. By that time we'll have done 693ms of hashing (in the best case), added ~19mb to the `BeaconState` and ~33mb of tree hash cache. Without tree hash caching we're looking at 12 *minutes* of hashing!

dapplion:
I am in support of this PR expanding on others points including:
- Ensures block space for withdrawals is optimally used when needed. Current dev branch will leave withdrawal spots empty unless churn limit is high enough.

However I want to point out that this approach _could_ support an implementation where full withdrawals do not have to incur extra wait time. Snippet below is _naive_ implementation for that goal. Splitting full and partial withdrawals in half in an arbitrary decision, but as long as churn is low and there's not a mass slashing event partial withdrawals will always dominate. Also, since the full withdrawal iteration does not have a pointer, older indexes will be prioritized always. To prevent having to loop on each block, one can keep an ordered hash set of indexes who satisfy the withdrawable_epoch condition and only evaluate those.

```python
def get_expected_withdrawals(state: BeaconState) -> Sequence[Withdrawal]:
    withdrawal_index = state.next_withdrawal_index
    withdrawals: List[Withdrawal] = []

    # Full

        balance = state.balances[index]
        validator = state.validators[index]
        if is_fully_withdrawable_validator(validator, balance, current_epoch):
            withdrawals.append(Withdrawal(
                index=WithdrawalIndex(withdrawal_index + len(withdrawals)),
                validator_index=index,
                address=ExecutionAddress(val.withdrawal_credentials[12:]),
                amount=balance,
            ))
        if len(withdrawals) == MAX_WITHDRAWALS_PER_PAYLOAD / 2:
            break

    # Partial
    index = ValidatorIndex((state.last_withdrawal_validator_index + 1) % len(state.validators))

    for probed in range(len(state.validators)):
        val = state.validators[index]
        balance = state.balances[index]
        if is_partially_withdrawable_validator(val, balance):
            withdrawals.append(Withdrawal(
                index=WithdrawalIndex(withdrawal_index + len(withdrawals)),
                validator_index=index,
                address=ExecutionAddress(val.withdrawal_credentials[12:]),
                amount=balance - MAX_EFFECTIVE_BALANCE,
            ))
        if len(withdrawals) == MAX_WITHDRAWALS_PER_PAYLOAD:
            break
        probed += 1
        index = ValidatorIndex((index + probed) % len(state.validators))
    return withdrawals
```

ethDreamer:
> require scanning a fixed number of validators rather than requiring a fixed number of withdrawals be included

Why not both? Scan until you've found `MAX_WITHDRAWALS_PER_PAYLOAD` or you've hit `VALIDATOR_SEEK_LIMIT`, whichever comes first. This avoids the "sub-optimal processing kick in when the chain hasn't been finalising" death spiral.

Setting `VALIDATOR_SEEK_LIMIT = 512` would ensure the entire validator set is processed once per day, even in the worst case scenario of no one being withdrawable, with 98% of ETH staked..

specs/capella/fork.md
ralexstokes: yeah I got confused on the naming -- this works!

specs/capella/beacon-chain.md
benjaminion: Doesn't the hysteresis take care of this? The actual balance would have to drop to 31.75 ETH for the effective balance to drop to 31 ETH. To lose 0.25 ETH under normal circumstances (not in a leak) takes quite a while, and in a leak situation adding a little margin won't help much in any case.

potuz: The margin serves another purpose IMO and is to prevent to makes small withrawals. Currently they are so small that they may forbid creating on-chain proofs cause gas may be higher than the withrawal

specs/capella/beacon-chain.md
```diff
@@ -381,19 +280,55 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
 ```
 
+#### New `get_expected_withdrawals`
+
+```python
+def get_expected_withdrawals(state: BeaconState) -> Sequence[Withdrawal]:
+    withdrawal_index = state.next_withdrawal_index
+    index = ValidatorIndex((state.last_withdrawal_validator_index + 1) % len(state.validators))
+    ret: List[Withdrawal] = []
```
dapplion, line 290:
```suggestion
    withdrawals: List[Withdrawal] = []
```
This variable should have a more human readable name

specs/capella/beacon-chain.md
```diff
@@ -381,19 +280,55 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+        val = state.validators[index]
+        balance = state.balances[index]
+        if is_fully_withdrawable_validator(val, balance, epoch):
+            withdrawal = Withdrawal(
+                index=withdrawal_index,
+                validator_index=index,
+                address=ExecutionAddress(val.withdrawal_credentials[12:]),
+                amount=balance,
+            )
+            ret.append(withdrawal)
```
dapplion, line 301: Can this two statements be condensed in one
```suggestion
            withdrawals.append(Withdrawal(
              index=withdrawal_index,
```

specs/capella/beacon-chain.md
```diff
@@ -381,19 +280,55 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+            withdrawal_index = WithdrawalIndex(withdrawal_index + 1)
```
dapplion, line 302: Does python require this syntax? A `withdrawal_index++` would look more readable, or use the increasing length from
```py
index=state.next_withdrawal_index + len(withdrawals)
```

specs/capella/beacon-chain.md
```diff
@@ -381,19 +280,55 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+                validator_index=index,
+                address=ExecutionAddress(val.withdrawal_credentials[12:]),
+                amount=balance - MAX_EFFECTIVE_BALANCE,
+            )
+            ret.append(withdrawal)
+            withdrawal_index = WithdrawalIndex(withdrawal_index + 1)
+        if len(ret) == MAX_WITHDRAWALS_PER_PAYLOAD:
+            break
+        probed += 1
+        index = ValidatorIndex((index + probed) % len(state.validators))
```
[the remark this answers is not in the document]
ethDreamer: i tried to say this yesterday but forgot to submit the review 😩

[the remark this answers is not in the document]
potuz: yep thanks, that's what happens when changing the while loop for a for loop :), will change anyway to use a fixed runner.

hwwhww (commented):
@potuz @ralexstokes 

I made the existing test cases green again ✅

specs/capella/beacon-chain.md
potuz: There's no increment operator in Python, the notation
```python
withdrawal_index += WithdrawalIndex(1)
```
would work, and in this case I believe even 
```python
withdrawal_index += 1
```
would, so I'll change to that and see if pyspec complains :)

specs/capella/fork.md
```diff
@@ -129,9 +129,8 @@ def upgrade_to_capella(pre: bellatrix.BeaconState) -> BeaconState:
         # Execution-layer
         latest_execution_payload_header=latest_execution_payload_header,
         # Withdrawals
-        withdrawal_queue=[],
         next_withdrawal_index=WithdrawalIndex(0),
-        next_partial_withdrawal_validator_index=ValidatorIndex(0),
+        latest_withdrawal_validator_index=ValidatorIndex(0),
```
potuz, line 133: This will start sweeping from validator_index = 1. It's a bit ugly, but putting here `ValidatorIndex(len(state.validators) - 1)` is uglier. I was tempted to put `latest_withdrawal_validator_index = 7653`.

specs/capella/beacon-chain.md
```diff
@@ -381,19 +280,52 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
-        assert dequeued_withdrawal == withdrawal
-
-    # Remove dequeued withdrawals from state
-    state.withdrawal_queue = state.withdrawal_queue[num_withdrawals:]
+    expected_withdrawals = get_expected_withdrawals(state)
+    assert len(payload.withdrawals) == len(expected_withdrawals)
+
+    for expected_withdrawal, withdrawal in zip(expected_withdrawals, payload.withdrawals):
+        assert withdrawal == expected_withdrawal
+        decrease_balance(state, withdrawal.validator_index, withdrawal.amount)
```
ralexstokes, line 325: I can't think of anything but it is worth asking if there are invariants we are breaking where a validator has a balance decrease and then needs to wait potentially many slots to have their effective balance update to match

specs/capella/beacon-chain.md
```diff
@@ -381,19 +280,52 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+    if len(expected_withdrawals) > 0:
+        state.next_withdrawal_index = WithdrawalIndex(withdrawal.index + 1)
```
ralexstokes, line 327: this works but does rely on some imo kind of unique python semantics that `withdrawal` stays in scope for the remainder of this function

consider adding a comment to explain this? only a soft suggestion tho

specs/capella/beacon-chain.md
```diff
@@ -381,19 +280,52 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+
+```python
+def get_expected_withdrawals(state: BeaconState) -> Sequence[Withdrawal]:
+    epoch = get_current_epoch(state)
+    withdrawal_index = state.next_withdrawal_index
+    validator_index = state.latest_withdrawal_validator_index 
+    withdrawals: List[Withdrawal] = []
+    for _ in range(len(state.validators)):
+        validator_index = ValidatorIndex((validator_index + 1) % len(state.validators))
+        val = state.validators[validator_index]
```
ralexstokes, line 293: I think better to use the full word here
```suggestion
        validator = state.validators[validator_index]
```

specs/capella/beacon-chain.md
```diff
@@ -381,19 +280,52 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+        balance = state.balances[validator_index]
+        if is_fully_withdrawable_validator(val, balance, epoch):
```
ralexstokes, line 295:
```suggestion
        if is_fully_withdrawable_validator(validator, balance, epoch):
```

ralexstokes (approved):
excellent job!

ive reviewed everything except the block processing tests, which ill go do now

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
```diff
@@ -1,39 +1,76 @@
...
-        state.withdrawal_queue.append(withdrawal)
+    return fully_withdrawable_indices, partial_withdrawals_indices
 
-    assert len(state.withdrawal_queue) == num_withdrawals + pre_queue_len
+
+def verify_post_state(state, spec, expected_withdrawals,
+                      fully_withdrawable_indices, partial_withdrawals_indices):
+    expected_withdrawals_validator_indices = [withdrawal.validator_index for withdrawal in expected_withdrawals]
+    assert state.next_withdrawal_index == expected_withdrawals[-1].index + 1
+    assert state.latest_withdrawal_validator_index == expected_withdrawals_validator_indices[-1]
```
ralexstokes, line 47: the tests are passing so we aren't hitting this condition now but this code will blow up if `expected_withdrawals` is empty -- which should be a setting we are testing...

me (or someone :) ) should make a pass on testing anyway once we have agreed on high-level design so im not too worried about this now but I'll call it out

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
```diff
@@ -67,35 +106,81 @@ def test_success_empty_queue(spec, state):
...
 @spec_state_test
-def test_success_max_per_slot_in_queue(spec, state):
-    prepare_withdrawal_queue(spec, state, spec.MAX_WITHDRAWALS_PER_PAYLOAD)
+def test_success_one_partial_withdrawal(spec, state):
+    fully_withdrawable_indices, partial_withdrawals_indices = prepare_expected_withdrawals(
+        spec, state, num_partial_withdrawals=1)
+    assert len(fully_withdrawable_indices) == 0
+    assert len(partial_withdrawals_indices) == 1
+    for index in partial_withdrawals_indices:
+        assert state.balances[index] != spec.MAX_EFFECTIVE_BALANCE
```
ralexstokes, line 131:
```suggestion
        assert state.balances[index] > spec.MAX_EFFECTIVE_BALANCE
```

although I feel like this is checked later in `verify_post_state` right?

ralexstokes (approved):
nice work on the tests!

ralexstokes:
ah, can you also update `fork.md` with the new fields?

specs/capella/fork.md
potuz: That can't work without an ugly hack :). Leaving it as is. Going through your review now. Thanks!

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
potuz: true, in all tests that we expect zero withdrawals we don't call this function, but I agree that it would be better to have an API that allows to test this situation as well. I'm adding a return early just in case, and leaving a comment to this review

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
potuz: `verify_post_state` only checks that they are reduced to `MAX_EFFECTIVE_BALANCE`, this check is only a sanity check that the helper `prepare_expected_withdrawals` worked I suppose. Changing as per suggestion. cc @hwwhww

potuz:
> ah, can you also update the `BeaconState` in `upgrade_to_capella` in`fork.md` with the new fields?

Isn't this done already?

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
ralexstokes: yeah I think we can just direct some more resources in the future to testing this more thoroughly

specs/capella/beacon-chain.md
```diff
@@ -381,19 +280,53 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
-    # Remove dequeued withdrawals from state
-    state.withdrawal_queue = state.withdrawal_queue[num_withdrawals:]
+    expected_withdrawals = get_expected_withdrawals(state)
+    assert len(payload.withdrawals) == len(expected_withdrawals)
+
+    for expected_withdrawal, withdrawal in zip(expected_withdrawals, payload.withdrawals):
+        assert withdrawal == expected_withdrawal
+        decrease_balance(state, withdrawal.validator_index, withdrawal.amount)
+    if len(expected_withdrawals) > 0:
+        # withdrawal holds the last withdrawal object in the payload.
```
ralexstokes, line 327:
```suggestion
        # `withdrawal` holds the last withdrawal object in the payload.
```

just to make it super clear

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
```diff
@@ -242,3 +450,395 @@ def test_fail_many_dequeued_incorrectly(spec, state):
...
+    expected_withdrawals = yield from run_withdrawals_processing(spec, state, execution_payload)
+    verify_post_state(state, spec, expected_withdrawals, fully_withdrawable_indices, partial_withdrawals_indices)
+    withdrawn_indices = [withdrawal.validator_index for withdrawal in expected_withdrawals]
+    fully_withdrawable_indices = list(set(fully_withdrawable_indices).difference(set(withdrawn_indices)))
+    partial_withdrawals_indices = list(set(partial_withdrawals_indices).difference(set(withdrawn_indices)))
+    assert state.next_withdrawal_index == next_withdrawal_index + spec.MAX_WITHDRAWALS_PER_PAYLOAD
+
+    execution_payload = build_empty_execution_payload(spec, state)
+    state.next_withdrawal_index += 1
+    yield from run_withdrawals_processing(spec, state, execution_payload, valid=False)
```
hwwhww, line 844: Thanks for adding more test cases 👍👍

However, our [current test format](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/operations) is processing one operation per test case. We can either (1) split `test_success_two_payloads` and `test_fail_second_payload_isnt_compatible` into 4 test cases or (2) I can rewrite them into sanity/block tests.

I think (2) makes more sense here. What do you think?

potuz: Oh I didn't know that. You can just remove them if you want the real condition is tested in the additional checks added in the post state checker so these tests are just a composition of two tests.

presets/mainnet/capella.yaml
```diff
@@ -1,23 +1,11 @@
...
-# 2**40 (= 1,099,511,627,776) withdrawals
-WITHDRAWAL_QUEUE_LIMIT: 1099511627776
-
-
 # Max operations per block
 # ---------------------------------------------------------------
 # 2**4 (= 16)
 MAX_BLS_TO_EXECUTION_CHANGES: 16
 
-
```
g11tech, line 20: I think `MAX_WITHDRAWALS_PER_PAYLOAD` is missing here

g11tech: sorry just saw it belong

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
hwwhww: Moved them in https://github.com/ethereum/consensus-specs/pull/3068/commits/3fc1ebce06ef5a51a32bb6640507d596358dfc14

djrtwo:
A quick note, there were two main reasons I did the withdrawals queue initially:
1. To ensure that load is bound and spread across slots even in exceptional scenarios (e.g. withdrawals are initially enabled and tons of validators are withdrawable). With the combined sweep that bounds per slot, this is no longer an issue
2. To ensure that CL and EL can be processed in parallel. I was misguided on this item. By not having a queue, we add complexity to block building for CLs -- they must ensure that the withdrawals created at *after* any balance updates that happen prior to the `process_withdrawals` call (which can happen due to `process_slots` being before `process_block`) . BUT because withdrawals show up independently in the Execution Payload for block processing (placed there by the proposer), they can immediately be passed into EL even if validated a bit later in CL. Point being -- I was incorrect on this and it was my sticking point so this PR seems fine. Will do a deep pass on it but we can merge shortly

specs/capella/beacon-chain.md
```diff
@@ -377,23 +276,57 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
+    expected_withdrawals = get_expected_withdrawals(state)
+    assert len(payload.withdrawals) == len(expected_withdrawals)
+
+    for expected_withdrawal, withdrawal in zip(expected_withdrawals, payload.withdrawals):
+        assert withdrawal == expected_withdrawal
+        decrease_balance(state, withdrawal.validator_index, withdrawal.amount)
+    if len(expected_withdrawals) > 0:
+        # `withdrawal` holds the last withdrawal object in the payload.
+        state.next_withdrawal_index = WithdrawalIndex(withdrawal.index + 1)
+        state.latest_withdrawal_validator_index = withdrawal.validator_index
```
djrtwo, line 329: This is too pythonic of an assumption. Let's just replace the comment with an explicit line of code

```suggestion
        latest_withdrawal = payload.withdrawals[-1]
        state.next_withdrawal_index = WithdrawalIndex(latest_withdrawal.index + 1)
        state.latest_withdrawal_validator_index = latest_withdrawal.validator_index
```

specs/capella/validator.md
```diff
@@ -58,13 +58,6 @@ All validator responsibilities remain unchanged other than those noted below.
 expected withdrawals for the slot must be gathered from the `state` (utilizing the
 helper `get_expected_withdrawals`) and passed into the `ExecutionEngine` within `prepare_execution_payload`.
```
djrtwo, line 59: This needs more clarification that it *must* be the post-process_slots state. This is stated in the Phase 0 document but this is a distinct "gotcha" with how we are doing withdrawals building here so is worth an explicit (re)callout

tests/core/pyspec/eth2spec/test/capella/sanity/test_blocks.py
```diff
@@ -133,3 +139,69 @@ def test_exit_and_bls_change(spec, state):
     assert not spec.is_fully_withdrawable_validator(validator, balance, current_epoch)
     assert validator.withdrawable_epoch < spec.FAR_FUTURE_EPOCH
     assert spec.is_fully_withdrawable_validator(validator, balance, validator.withdrawable_epoch)
+
+
+def _valid_withdrawl(spec, state):
```
djrtwo, line 144:
```suggestion
def _valid_withdrawal(spec, state):
```

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
```diff
@@ -67,35 +95,81 @@ def test_success_empty_queue(spec, state):
...
+    fully_withdrawable_indices, partial_withdrawals_indices = prepare_expected_withdrawals(
+        spec, state, num_full_withdrawals=1)
+    assert len(fully_withdrawable_indices) == 1
+    assert len(partial_withdrawals_indices) == 0
 
     next_slot(spec, state)
     execution_payload = build_empty_execution_payload(spec, state)
 
-    yield from run_withdrawals_processing(spec, state, execution_payload)
+    expected_withdrawals = yield from run_withdrawals_processing(spec, state, execution_payload)
```
djrtwo, line 107: `run_*` should just encapsulate `verify_post_state` rather than having a return value in addition to it's yield. The default should be to do post-state validations rather than having to hoist it into another finction

tests/core/pyspec/eth2spec/test/capella/sanity/test_blocks.py
djrtwo: and consider it having an 'action' associated with the name -- e.g. `_perform_valid_withdrawal` or something. Currently, this does not make it clear that it modifies `state`

tests/core/pyspec/eth2spec/test/capella/sanity/test_blocks.py
```diff
@@ -133,3 +139,69 @@ def test_exit_and_bls_change(spec, state):
     assert not spec.is_fully_withdrawable_validator(validator, balance, current_epoch)
     assert validator.withdrawable_epoch < spec.FAR_FUTURE_EPOCH
     assert spec.is_fully_withdrawable_validator(validator, balance, validator.withdrawable_epoch)
+
+
```
djrtwo, line 143: We need to make sure we have a test that is at an epoch transition that induces balance changes for a to-be withdrawn validator

tests/core/pyspec/eth2spec/test/capella/sanity/test_blocks.py
potuz: changed the function name. Will leave the others to @hwwhww that knows what she's doing much better than what I could do here.

specs/capella/beacon-chain.md
```diff
@@ -377,23 +276,57 @@ def process_block(state: BeaconState, block: BeaconBlock) -> None:
...
     process_sync_aggregate(state, block.body.sync_aggregate)
 ```
 
+#### New `get_expected_withdrawals`
+
+```python
+def get_expected_withdrawals(state: BeaconState) -> Sequence[Withdrawal]:
+    withdrawal_index = state.next_withdrawal_index
+    validator_index = state.latest_withdrawal_validator_index
```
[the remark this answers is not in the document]
potuz: Went through the PR, if you insist on this I can change this to have the double `%` in the logic here. It will not change formally the spec. If however you also want to start from *-1* at the fork, that could be debatable :)

tests/core/pyspec/eth2spec/test/capella/sanity/test_blocks.py
potuz: From looking at the existing tests, several of them would already do this. For example `test_full_withdrawal_in_epoch_transition` sets the validator as withdrawable first, and then advances slots for an entire epoch which should affect balance of this validator.

specs/capella/beacon-chain.md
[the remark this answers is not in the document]
djrtwo: next would change the spec, no? As it would start at and utilize 0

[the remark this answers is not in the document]
ralexstokes: if it helps at all I think we will be happier in the long run if we have symmetry here with the naming, even if that implies a double modulo operation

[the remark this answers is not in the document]
potuz: I'm confused as there are two or three issues here at the same time. I'm happy to add the extra ÷ here and this will start the sweep at zero as Danny mentions. This is fine by me. Naming this variable as `latest_validator_index` is good by me too. If these are the requests I'll push this in a couple of hours when I'm at the keyboard, just need nudge here to see if we're on the same page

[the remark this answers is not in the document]
potuz: Added the extra % and changing the index in the end of the loop. Didn't change the variable name cause it felt odd to call it latest when it's now the current.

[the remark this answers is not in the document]
djrtwo: I think we've been talking past eachother, I apologize.

I think that both vars should be called `next_*`

In the beginning of `get_expected_withdrawals` you would assign `validator_index = state.next_withdrawal_validator_index` and each loop you would increment (with the mod).  Then in `process_withdrawals` you would do `state.next_withdrawal_validator_index = (latest_withdrawal.validator_index + 1) % len(state.validators)`. This is the second mod I am referring to, and ensures that the withdrawal index and withdrawal_validator_index parallel eachother in usage in the loop (assign, then increment. Rather than increment then assign).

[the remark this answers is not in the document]
djrtwo: Just like here -- https://github.com/ethereum/consensus-specs/pull/3068/files#diff-e62664a360f87fbbd16c6f0f0d11f0e47dcc4f343dc6478b6f81c3f5acd6d8cbL246

it's the next place to look on the next sweep

I'll do a quick pr to this pr. apologies for my failure to communicate clearly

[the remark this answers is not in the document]
potuz: Pushed these changes, named the state counter now `next_withdrawal_validator_index`. Let me know if this works for you and I'll change our deployed code :)

g11tech (approved):
lgtm

djrtwo (approved):
Fantastic work @potuz!

specs/capella/beacon-chain.md
djrtwo: This also can happen in sync committees

specs/capella/beacon-chain.md
```diff
@@ -11,11 +11,8 @@
 - [Constants](#constants)
   - [Domain types](#domain-types)
 - [Preset](#preset)
-  - [Misc](#misc)
-  - [State list lengths](#state-list-lengths)
   - [Max operations per block](#max-operations-per-block)
```
Dustin4444, line 14: tests/core/pyspec/eth2spec/test/helpers/execution_payload.py

PR #3069, opened 2022-10-28
Update `remerkleable` to 0.1.25

etan-status:
`remerkleable` was updated to address potentially incorrect computation of `hash_tree_root` against default-initialized `Vector` objects. Switching to the fixed version.


ralexstokes (commented):
looks ok but would want an approval from @protolambda

protolambda (approved):

PR #3070, opened 2022-10-28
Pass blocks to LC data creation functions

etan-status:
Future light client protocol extensions may include data from the block in addition to data from the state, e.g., `ExecutionPayloadHeader`. To prepare for this, also pass the block to the corresponding functions. In practice, blocks access is easier than historic state access, meaning there are no practical limitations due to this change.


wemeetagain (approved):
LGTM, causes no change in output now, just sets the groundwork for adding an execution header later.

etan-status:
Yep. Preview of the full change: https://github.com/etan-status/consensus-specs/commit/lc-eph
Splitting it up into multiple logical pieces so that it is actually reviewable

Also related:
– 3066 (prep for new tests that are based on BeaconBlockBody)
– 3070 (prep for executionpayload in LC headers)
– 3072 (more maintainable tests, currently capella/eip4844 work has not updated LC test helpers)
– 3078 (RLP hashes in ExecutionPayloadHeader, so that it is possible to drive LES)

dapplion (approved):
LGTM

hwwhww (approved):
lgtm 👍

etan-status:
Anything still blocking this?

PR #3072, opened 2022-10-30
Move `is_post_xxx` functions to new module

etan-status:
Moving the `is_post_xxx` functions to a separate module allows `genesis` to also use them (cyclic import from `context` prevented this before). This allows removing `FORKS_BEFORE_ALTAIR` and `FORKS_BEFORE_BELLATRIX` constants and adding a more general `is_post_fork` function that needs less maintenance. This then allows definition of `with_all_phases_from` to streamline the implementation of the `with_xxx_and_later` decorators.


tests/core/pyspec/eth2spec/test/helpers/phases.py
```diff
@@ -0,0 +1,33 @@
+from .constants import (
```
hwwhww, line 1: Although we had Phase0, now I wish we could retire "phases" someday since developments are in-parallel.

What do you think about `protocols.py` or `forks.py`?

tests/core/pyspec/eth2spec/test/helpers/phases.py
```diff
@@ -0,0 +1,33 @@
...
+        return b in [PHASE0, ALTAIR, BELLATRIX, EIP4844]
+    if a == CAPELLA:
+        return b in [PHASE0, ALTAIR, BELLATRIX, CAPELLA]
+    if a == BELLATRIX:
+        return b in [PHASE0, ALTAIR, BELLATRIX]
+    if a == ALTAIR:
+        return b in [PHASE0, ALTAIR]
+    if a == PHASE0:
+        return b in [PHASE0]
+    assert False  # Fork is missing
```
hwwhww, line 17:
```suggestion
    raise ValueError("Wrong fork name %s" % a)
```

hwwhww (approved):
Nice refactoring! 

Raised some naming nitpicking around phases v.s. forks. protocols here.

tests/core/pyspec/eth2spec/test/helpers/phases.py
etan-status: I will go with `forks.py` then. It is consistent with the `fork=...` naming when running `make citest`.
I will not touch the decorators such as `@with_all_phases` for now, as it is out of scope of this PR.

PR #3075, opened 2022-11-01
Capella: Add Gossip Topic bls_to_execution_change

ethDreamer:
This seems to have been left out of the existing specification. I welcome feedback as this is a first pass.

### DIFF
--- a/specs/capella/p2p-interface.md
+++ b/specs/capella/p2p-interface.md
@@ -1,0 +1,110 @@
+# Capella -- Networking
+
+This document contains the networking specification for Capella.
+
+The specification of these changes continues in the same format as the network specifications of previous upgrades, and assumes them as pre-requisite.
+
+## Table of contents
+
+<!-- TOC -->
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Modifications in Capella](#modifications-in-capella)
+  - [The gossip domain: gossipsub](#the-gossip-domain-gossipsub)
+    - [Topics and messages](#topics-and-messages)
+      - [Global topics](#global-topics)
+        - [`beacon_block`](#beacon_block)
+        - [`bls_to_execution_change`](#bls_to_execution_change)
+    - [Transitioning the gossip](#transitioning-the-gossip)
+  - [The Req/Resp domain](#the-reqresp-domain)
+    - [Messages](#messages)
+      - [BeaconBlocksByRange v2](#beaconblocksbyrange-v2)
+      - [BeaconBlocksByRoot v2](#beaconblocksbyroot-v2)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- /TOC -->
+
+
+# Modifications in Capella
+
+## The gossip domain: gossipsub
+
+A new topic is added to support the gossip of bls to execution change messages.
+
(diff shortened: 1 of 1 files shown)


specs/capella/p2p-interface.md
```diff
@@ -0,0 +1,110 @@
...
+
+Capella changes the type of the global beacon block topic and adds one global topic to propagate bls to execution change messages to all potential proposers of beacon blocks.
+
+##### `beacon_block`
+
+The *type* of the payload of this topic changes to the (modified) `SignedBeaconBlock` found in Capella.
+Specifically, this type changes with the addition of `bls_to_execution_changes` to the inner `BeaconBlockBody`.
+See Capella [state transition document](./beacon-chain.md#beaconblockbody) for further details.
+
+##### `bls_to_execution_change`
```
rolfyone, line 58: I'm not sure where we call out that this is signed with the withdrawal key? It seems like an important detail but i haven't seen it mentioned anywhere...
I'm fairly sure this is the intent, I guess either way i'll know soon!

ethDreamer: it's in the table above and it's explicit that the data is a `SignedBLSToExecutionChange` and below it is also referred to as `signed_bls_to_execution_change`. Here the `signed` is left out because it's being referred to by the name of the message type (which omits it to save space). This is consistent with [previous versions of these documents](https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/p2p-interface.md#beacon_block)

ralexstokes: I think we should add a note to the `validator.md` guide

lucassaldanha: > I think we should add a note to the `validator.md` guide

👍

specs/capella/p2p-interface.md
```diff
@@ -0,0 +1,110 @@
...
+
+This topic is used to propagate signed bls to execution change messages to be included in future blocks.
+
+The following validations MUST pass before forwarding the `signed_bls_to_execution_change` on the network:
+
+- _[IGNORE]_ The `signed_bls_to_execution_change` is the first valid signed bls to execution change received
+  for the validator with index `signed_bls_to_execution_change.message.validator_index`.
+- _[REJECT]_ All of the conditions within `process_bls_to_execution_change` pass validation.
```
ralexstokes, line 66: I like this condition but it assumes the gossip validator has access to (at least a cache of) the validator set to know if the public key is the one committed to in the validator record

will this be a problem? or at least introduce new requirements beyond what we currently have?

specs/capella/p2p-interface.md
```diff
@@ -0,0 +1,110 @@
...
+Specifically, this type changes with the addition of `bls_to_execution_changes` to the inner `BeaconBlockBody`.
+See Capella [state transition document](./beacon-chain.md#beaconblockbody) for further details.
+
+##### `bls_to_execution_change`
+
+This topic is used to propagate signed bls to execution change messages to be included in future blocks.
+
+The following validations MUST pass before forwarding the `signed_bls_to_execution_change` on the network:
+
+- _[IGNORE]_ The `signed_bls_to_execution_change` is the first valid signed bls to execution change received
```
ralexstokes, line 64: we need this condition to prevent a DoS vector however it does mean once a user has published a change message they won't be able to mutate it

this is fine imo but it should be very clear in any documentation that broadcast is irreversible

ralexstokes (approved):
excellent job! left some suggestions on wording but not critical

PR #3077, opened 2022-11-03
Fix type error in the inputs to hash_to_bls_field()

asn-d6:
`compute_aggregated_poly_and_commitment` calls `hash_to_bls_field` with a `Sequence[Blob]` as its first arg, but the type signature for that function takes a `Sequence[Polynomial]`. This commit fixes this.


hwwhww (approved):
lgtm

PR #3083, opened 2022-11-07
EIP4844: Remove `DOMAIN_BLOBS_SIDECAR`

terencechain:
`DOMAIN_BLOBS_SIDECAR` is no longer used after we coupled `blobs_sidecar` and `beacon_block` under the same gossip. This PR removes it from the spec


hwwhww (approved):
nice catch!

related PR: #3046

PR #3086, opened 2022-11-08
Fix slot checks in light client p2p spec

michaelsproul:
I think the p2p spec for light client updates should require the received update to be _newer_ than previously forwarded updates. In other words, the condition should be corrected to:

> No other `finality_update` with a ~~lower~~ greater or equal `finalized_header.slot` was already forwarded on the network

However I think it's somewhat clearer to state the condition without a negation, as in:

> _[IGNORE]_ The `finalized_header.slot` is greater than the `slot` of all previously forwarded finality updates

### DIFF
--- a/specs/altair/light-client/p2p-interface.md
+++ b/specs/altair/light-client/p2p-interface.md
@@ -59,7 +59,7 @@ New global topics are added to provide light clients with the latest updates.
 This topic is used to propagate the latest `LightClientFinalityUpdate` to light clients, allowing them to keep track of the latest `finalized_header`.
 
 The following validations MUST pass before forwarding the `finality_update` on the network.
-- _[IGNORE]_ The `finalized_head.slot` is greater than the `slot` of all previously forwarded finality updates
+- _[IGNORE]_ The `finalized_header.slot` is greater than the `slot` of all previously forwarded finality updates
 - _[IGNORE]_ The `finality_update` is received after the block at `signature_slot` was given enough time to propagate through the network -- i.e. validate that one-third of `finality_update.signature_slot` has transpired (`SECONDS_PER_SLOT / INTERVALS_PER_SLOT` seconds after the start of the slot, with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance)
 
 For full nodes, the following validations MUST additionally pass before forwarding the `finality_update` on the network.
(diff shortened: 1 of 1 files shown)


ralexstokes (approved):
makes sense, otherwise there is a mild DoS vector where I just keep spamming the current update to add load to the gossip net

tuyennhv (approved):

dapplion (approved):
Looks reasonable to me, CC: @etan-status

specs/altair/light-client/p2p-interface.md
```diff
@@ -59,7 +59,7 @@ New global topics are added to provide light clients with the latest updates.
 This topic is used to propagate the latest `LightClientFinalityUpdate` to light clients, allowing them to keep track of the latest `finalized_header`.
 
 The following validations MUST pass before forwarding the `finality_update` on the network.
-- _[IGNORE]_ No other `finality_update` with a lower or equal `finalized_header.slot` was already forwarded on the network
+- _[IGNORE]_ The `finalized_header.slot` is greater than the `slot` of all previously forwarded finality updates
```
etan-status, line 62:
```suggestion
- _[IGNORE]_ The `finalized_header.slot` is greater than the one of all previously forwarded `finality_update`
```
- Use "the one" to make it clear that it is referring to the same `slot` (there is also `attested_header.slot`)
- Use `finality_update` for consistency ("finality update" is not used so far as a term in the spec, just the variable name)

specs/altair/light-client/p2p-interface.md
```diff
@@ -87,7 +87,7 @@ Per `context = compute_fork_digest(fork_version, genesis_validators_root)`:
 This topic is used to propagate the latest `LightClientOptimisticUpdate` to light clients, allowing them to keep track of the latest `optimistic_header`.
 
 The following validations MUST pass before forwarding the `optimistic_update` on the network.
-- _[IGNORE]_ No other `optimistic_update` with a lower or equal `attested_header.slot` was already forwarded on the network
+- _[IGNORE]_ The `attested_header.slot` is greater than the `slot` of all previously forwarded optimistic updates
```
etan-status, line 90:
```suggestion
- _[IGNORE]_ The `attested_header.slot` is greater than the one of all previously forwarded `optimistic_update`
```

etan-status (approved):
Thanks, yes, that was the intention (slots must strictly monotonically increase in gossip), and is also how it is implemented in Nimbus. The new wording is clearer as well.

specs/altair/light-client/p2p-interface.md
michaelsproul, line 62: How about this?

```suggestion
- _[IGNORE]_ The `finalized_header.slot` is greater than that of all previously forwarded `finality_update`s
```

specs/altair/light-client/p2p-interface.md
michaelsproul, line 90:
```suggestion
- _[IGNORE]_ The `attested_header.slot` is greater than that of all previously forwarded `optimistic_update`s
```

etan-status: Looks good :)

michaelsproul: thanks for the quick feedback :)

etan-status (approved):

Issue #3088, opened 2022-11-08
EIP-4844: Remaining p2p-interface todos


terencechain:
- We should mention that the original `beacon_block` gossip global topic will be dropped in favor of `beacon_block_and_blobs_sidecar`. The current spec does not mention whether the `beacon_block` topic for Capella should still be supported or not
- After the block and blob are coupled, we need a method to retrieve `BlobsSidecar` by root _or_ `SignedBeaconBlockAndBlobsSidecar` by root. We should survey which is preferred  
- Discussion on removing `BlobsSidecarsByRange` RPC #3087

PR #3089, opened 2022-11-08
EIP4844: Add block and sidecar retrival by root

terencechain:
Part of #3088

This PR adds a req-resp RPC for retrieving beacon block and blobs sidecar by root

### DIFF
--- a/specs/eip4844/p2p-interface.md
+++ b/specs/eip4844/p2p-interface.md
@@ -23,6 +23,7 @@ The specification of these changes continues in the same format as the network s
     - [Messages](#messages)
       - [BeaconBlocksByRange v2](#beaconblocksbyrange-v2)
       - [BeaconBlocksByRoot v2](#beaconblocksbyroot-v2)
+      - [BeaconBlockAndBlobsSidecarByRoot v1](#beaconblockandblobssidecarbyroot-v1)
       - [BlobsSidecarsByRange v1](#blobssidecarsbyrange-v1)
 - [Design decision rationale](#design-decision-rationale)
   - [Why are blobs relayed as a sidecar, separate from beacon blocks?](#why-are-blobs-relayed-as-a-sidecar-separate-from-beacon-blocks)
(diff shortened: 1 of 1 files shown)


specs/eip4844/p2p-interface.md
```diff
@@ -129,18 +130,45 @@ Per `context = compute_fork_digest(fork_version, genesis_validators_root)`:
...
+(
+  List[Root, MAX_REQUEST_BLOCKS]
+)
+```
+
+Response Content:
+
+```
+(
+  List[SignedBeaconBlockAndBlobsSidecar, MAX_REQUEST_BLOCKS]
```
terencechain, line 153: Should I rename `MAX_REQUEST_BLOCKS` to include sidecars? I didn't feel strongly about this

specs/eip4844/p2p-interface.md
```diff
@@ -129,18 +130,45 @@ Per `context = compute_fork_digest(fork_version, genesis_validators_root)`:
...
+It may be less in the case that the responding peer is missing blocks and sidecars.
+
+No more than `MAX_REQUEST_BLOCKS` may be requested at a time.
+
+`BeaconBlockAndBlobsSidecarByRoot` is primarily used to recover recent blocks and sidecars (e.g. when receiving a block or attestation whose parent is unknown).
+
+The response MUST consist of zero or more `response_chunk`.
+Each _successful_ `response_chunk` MUST contain a single `SignedBeaconBlockAndBlobsSidecar` payload.
+
+Clients MUST support requesting blocks and sidecars since the latest finalized epoch.
```
terencechain, line 168: @protolambda Should I expand this and say clients will not serve blocks and sidecars that are older than finalized epoch?

specs/eip4844/p2p-interface.md
Inphi: Fine by me. Though another thing is do we wanna tweak the current `MAX_REQUEST_BLOCKS` value? If a client requests the max and they all contain sidecars, that's a huge amount of data transmitted.

Inphi: Unless this is also constrained by the `MAX_CHUNK_SIZE`.

specs/eip4844/p2p-interface.md
```diff
@@ -129,18 +130,45 @@ Per `context = compute_fork_digest(fork_version, genesis_validators_root)`:
 
 **Protocol ID:** `/eth2/beacon_chain/req/beacon_blocks_by_root/2/`
 
-The EIP-4844 fork-digest is introduced to the `context` enum to specify EIP-4844 beacon block type.
+`BeaconBlocksByRoot v2` will be disabled in favor of `BeaconBlockAndBlobsSidecarByRoot v1` for tighter coupling of blocks and sidecars.
 
-Per `context = compute_fork_digest(fork_version, genesis_validators_root)`:
+Clients MUST disable `BeaconBlocksByRoot v2` after `EIP4844_FORK_EPOCH`.
```
ralexstokes, line 135: what if I want to get a beacon block by root from before the 4844 fork?

seems like we want to keep `BeaconBlocksByRootV2` for this purpose **or** support previous types with the context enum like in the `ByRange` spec here

specs/eip4844/p2p-interface.md
ralexstokes: I think it is clear as it is

specs/eip4844/p2p-interface.md
terencechain: Makes sense. I added back the previous types except for 4844 type

Inphi (approved):
LGTM! I'm still not sure if clients, prysm in particular, do not sometimes need to use BlocksByRoot (and BlocksAndBlobsSidecarByRoot) during historical sync. I'll defer to your decision on whether we should explicitly permit ByRoot req-res for older blocks.

specs/eip4844/p2p-interface.md
```diff
@@ -129,7 +130,7 @@ Per `context = compute_fork_digest(fork_version, genesis_validators_root)`:
 
 **Protocol ID:** `/eth2/beacon_chain/req/beacon_blocks_by_root/2/`
 
-The EIP-4844 fork-digest is introduced to the `context` enum to specify EIP-4844 beacon block type.
+After `EIP4844_FORK_EPOCH`, `BeaconBlocksByRootV2` is replaced by `BeaconBlockAndBlobsSidecarByRootV1`
```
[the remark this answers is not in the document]
djrtwo: Or does this replacement mean that I can get non-4844 blocks with just an empty sidecar?

specs/eip4844/p2p-interface.md
djrtwo: I do think that we should at least consider `MAX_REQUEST_BLOCKS` for this and range requests given the size changes. That said, you only *MUST* respond with 1 in a given request/range so even if the max stays high, clients can down regulate responses and stay to spec

specs/eip4844/p2p-interface.md
djrtwo: I think this angle of the statement is fine. Using this as a historic sync has no guarantees to work

specs/eip4844/p2p-interface.md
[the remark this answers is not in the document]
terencechain: We can still get non-4844 blocks in the form of `[phase0|altair|bellatrix].SignedBeaconBlock` from `BeaconBlocksByRootV2`. We are not shutting it down in this PR. (Maybe we should in another PR?)

For 4844 blocks, you will have to use `BeaconBlockAndBlobsSidecarByRootV1`.
Let me know if it isn't clear. I'm having trouble articulating this well.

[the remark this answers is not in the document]
djrtwo: Sorry, I read "replaced" as "supplanted by" (and thus the other to no longer exist).

Let's be clear that it still functions for pre-fork-epoch

djrtwo (approved):

PR #3090, opened 2022-11-08
clarify deprecation of `beacon_block` gossip topic in 4844

ralexstokes:
part of #3088

### DIFF
--- a/specs/eip4844/p2p-interface.md
+++ b/specs/eip4844/p2p-interface.md
@@ -88,6 +88,8 @@ This topic is deprecated and clients **MUST NOT** expose in their topic set to a
 anything beyond simply skip implementation, and it is explicitly called out as it is a departure from previous versioning
 of this topic.
 
+Refer to [the section below](#transitioning-the-gossip) for details on how to transition the gossip.
+
 ##### `beacon_block_and_blobs_sidecar`
 
 This topic is used to propagate new signed and coupled beacon blocks and blobs sidecars to all nodes on the networks.
--- a/specs/eip4844/p2p-interface.md
+++ b/specs/eip4844/p2p-interface.md
@@ -85,7 +85,7 @@ EIP4844 introduces a new global topic for beacon block and blobs-sidecars.
 ##### `beacon_block`
 
 This topic is deprecated and clients **MUST NOT** expose in their topic set to any peer. Implementers do not need to do
-anything beyond simply skip implementation and it is explicitly called out as it is a departure from previous versioning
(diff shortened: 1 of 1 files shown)


terencechain:
Thank you! 🙏🏼

jessepollak:
This looks good to me!

specs/eip4844/p2p-interface.md
```diff
@@ -81,11 +82,17 @@ The new topics along with the type of the `data` field of a gossipsub message ar
 
 EIP4844 introduces a new global topic for beacon block and blobs-sidecars.
 
+##### `beacon_block`
+
+This topic is deprecated and clients **MUST NOT** expose in their topic set to any peer. Implementers do not need to do
```
djrtwo, line 87: Might be worth noting when to turn it off. 
The two safe options are either:
1. (easier) Turn off the topic one epoch after `4844_FORK_EPOCH` to allow for any weird late gossip
2. (harder) turn off after `4844_FORK_EPOCH` is finalized
    * This is a bit more "correct" wrt to the `beacon_block` p2p slot range validation -- "validate that `signed_beacon_block.message.slot > compute_start_slot_at_epoch(state.finalized_checkpoint.epoch)`"
    * But might induce undue complexity when option (1) is pretty damn fine

specs/eip4844/p2p-interface.md
```diff
@@ -81,11 +82,17 @@ The new topics along with the type of the `data` field of a gossipsub message ar
...
+anything beyond simply skip implementation and it is explicitly called out as it is a departure from previous versioning
```
djrtwo, line 88:
```suggestion
anything beyond simply skip implementation, and it is explicitly called out as it is a departure from previous versioning
```

djrtwo (commented):
Let's make it clear *when* to disable the topic

specs/eip4844/p2p-interface.md
ralexstokes: the spec here: 

https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/p2p-interface.md#transitioning-the-gossip

suggests to me that nodes should wait two epochs from the fork and then unsubscribe this topic

do you think we need special consideration here?

djrtwo: This is fine. I might make an explicit call out to that previous area in the spec as I think it has the potential to be missed

ralexstokes:
I think this is ready to go, just needs an approving review

djrtwo (approved):

PR #3091, opened 2022-11-09
EIP4844: Explicitly mention old gossip validations for block

terencechain:
We should explicitly call out the old gossip block validations in the new `beacon_block_and_blobs_sidecar` topic.
It currently reads like those old validations are no longer required

### DIFF
--- a/specs/eip4844/p2p-interface.md
+++ b/specs/eip4844/p2p-interface.md
@@ -85,7 +85,7 @@ EIP4844 introduces a new global topic for beacon block and blobs-sidecars.
 
 This topic is used to propagate new signed and coupled beacon blocks and blobs sidecars to all nodes on the networks.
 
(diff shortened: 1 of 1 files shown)


specs/eip4844/p2p-interface.md
```diff
@@ -85,7 +85,7 @@ EIP4844 introduces a new global topic for beacon block and blobs-sidecars.
 
 This topic is used to propagate new signed and coupled beacon blocks and blobs sidecars to all nodes on the networks.
 
-The following validations MUST pass before forwarding the `signed_beacon_block_and_blobs_sidecar` on the network.  
+In addition to the gossip validations for `beacon_block` topic from prior specifications, the following validations MUST pass before forwarding the `signed_beacon_block_and_blobs_sidecar` on the network.
```
ralexstokes, line 88:
```suggestion
In addition to the gossip validations for the `beacon_block` topic from prior specifications, the following validations MUST pass before forwarding the `signed_beacon_block_and_blobs_sidecar` on the network.  
```

djrtwo (approved):

PR #3095, opened 2022-11-11
bound the maximum number of validators considered for withdrawals per sweep

### LINKED ISSUE #3079: Bound the max size of the validator set to scan per withdrawals processing
After #3068, the worst case behavior during withdrawals processing implies a complete scan of the full validator set on _each_ block.

There is some discussion in that PR that suggests even as is this worst case scenario is not a blocker but it would be nice to bound the length of the scan to have hard numbers on performance here.

To implement, we need to add a constant `MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP` and add it to `get_expected_withdrawals` in that PR on the line that scans the validator set:

```python
for _ in range(MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP):
```

I wanted to open this issue for discussion of this approach and also to get agreement on a concrete number for this new constant.

I suggest `2**17` as it is a nice power of two and is around 1/4 of the validator set against current `mainnet` numbers.

ralexstokes:
fixes #3079

this PR does a few things:

1) bound the size of the sweep in the worst case to the constant `MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP`
2) implement some notion of "fairness" if there were no validators found in the current sweep

the rationale for (2) is without it the processing will just stay at the same place in the validator set until there is some withdrawal to advance the state of the sweep

(2) adds some complexity to the implementation and I am not convinced it pulls its weight -- if you don't see the value please chime in below and I'll remove from this PR

if/when there is general consensus on this change then I'll add some testing

### DIFF
--- a/specs/capella/beacon-chain.md
+++ b/specs/capella/beacon-chain.md
@@ -82,7 +82,7 @@ We define the following Python custom types for type hinting and readability:
 
 | Name | Value |
 | - | - |
-| `MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP` | `1024` (= 2**10 ) |
+| `MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP` | `16384` (= 2**14 ) |
 
 ## Containers
 
--- a/specs/capella/beacon-chain.md
+++ b/specs/capella/beacon-chain.md
@@ -8,11 +8,11 @@
 
 - [Introduction](#introduction)
 - [Custom types](#custom-types)
-- [Constants](#constants)
   - [Domain types](#domain-types)
 - [Preset](#preset)
   - [Max operations per block](#max-operations-per-block)
   - [Execution](#execution)
+  - [Withdrawals processing](#withdrawals-processing)
 - [Containers](#containers)
   - [New containers](#new-containers)
     - [`Withdrawal`](#withdrawal)
@@ -58,8 +58,6 @@ We define the following Python custom types for type hinting and readability:
 | - | - | - |
 | `WithdrawalIndex` | `uint64` | an index of a `Withdrawal` |
 
-## Constants
-
 ### Domain types
 
 | Name | Value |
@@ -79,6 +77,12 @@ We define the following Python custom types for type hinting and readability:
 | Name | Value | Description |
 | - | - | - |
 | `MAX_WITHDRAWALS_PER_PAYLOAD` | `uint64(2**4)` (= 16) | Maximum amount of withdrawals allowed in each payload |
+
+### Withdrawals processing
+
+| Name | Value |
+| - | - |
+| `MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP` | `1024` (= 2**10 ) |
 
 ## Containers
 
@@ -288,7 +292,8 @@ def get_expected_withdrawals(state: BeaconState) -> Sequence[Withdrawal]:
     withdrawal_index = state.next_withdrawal_index
     validator_index = state.next_withdrawal_validator_index
     withdrawals: List[Withdrawal] = []
-    for _ in range(len(state.validators)):
+    bound = min(len(state.validators), MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP)
+    for _ in range(bound):
         validator = state.validators[validator_index]
         balance = state.balances[validator_index]
         if is_fully_withdrawable_validator(validator, balance, epoch):
@@ -312,7 +317,7 @@ def get_expected_withdrawals(state: BeaconState) -> Sequence[Withdrawal]:
         validator_index = ValidatorIndex((validator_index + 1) % len(state.validators))
     return withdrawals
 ```
-        
+
 #### New `process_withdrawals`
 
 ```python
@@ -323,10 +328,21 @@ def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
     for expected_withdrawal, withdrawal in zip(expected_withdrawals, payload.withdrawals):
         assert withdrawal == expected_withdrawal
         decrease_balance(state, withdrawal.validator_index, withdrawal.amount)
-    if len(expected_withdrawals) > 0:
+
+    # Update the next withdrawal index if this block contained withdrawals
+    if len(expected_withdrawals) != 0:
         latest_withdrawal = expected_withdrawals[-1]
         state.next_withdrawal_index = WithdrawalIndex(latest_withdrawal.index + 1)
-        next_validator_index = ValidatorIndex((latest_withdrawal.validator_index + 1) % len(state.validators))
+
+    # Update the next validator index to start the next withdrawal sweep
+    if len(expected_withdrawals) == MAX_WITHDRAWALS_PER_PAYLOAD:
+        # Next sweep starts after the latest withdrawal's validator index
+        next_validator_index = ValidatorIndex((expected_withdrawals[-1].validator_index + 1) % len(state.validators))
+        state.next_withdrawal_validator_index = next_validator_index
+    else:
+        # Advance sweep by the max length of the sweep if there was not a full set of withdrawals
+        next_index = state.next_withdrawal_validator_index + MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
+        next_validator_index = ValidatorIndex(next_index % len(state.validators))
         state.next_withdrawal_validator_index = next_validator_index
 ```
 
--- a/tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
+++ b/tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
@@ -180,6 +180,7 @@ def test_success_all_fully_withdrawable_in_one_sweep(spec, state):
         fully_withdrawable_indices=fully_withdrawable_indices,
         partial_withdrawals_indices=partial_withdrawals_indices)
 
+
 @with_phases([CAPELLA])
 @with_presets([MINIMAL], reason="too many validators with mainnet config")
 @spec_state_test
@@ -216,6 +217,7 @@ def test_success_all_partially_withdrawable_in_one_sweep(spec, state):
         spec, state, execution_payload,
         fully_withdrawable_indices=fully_withdrawable_indices,
         partial_withdrawals_indices=partial_withdrawals_indices)
+
 
 @with_phases([CAPELLA])
 @with_presets([MINIMAL], reason="too many validators with mainnet config")
--- a/tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
+++ b/tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
@@ -6,7 +6,7 @@ from eth2spec.test.context import (
     with_presets,
     with_phases,
 )
-from eth2spec.test.helpers.constants import MINIMAL, CAPELLA
+from eth2spec.test.helpers.constants import MAINNET, MINIMAL, CAPELLA
 from eth2spec.test.helpers.execution_payload import (
     build_empty_execution_payload,
 )
@@ -33,8 +33,13 @@ def verify_post_state(state, spec, expected_withdrawals,
 
     expected_withdrawals_validator_indices = [withdrawal.validator_index for withdrawal in expected_withdrawals]
     assert state.next_withdrawal_index == expected_withdrawals[-1].index + 1
-    next_withdrawal_validator_index = (expected_withdrawals_validator_indices[-1] + 1) % len(state.validators)
-    assert state.next_withdrawal_validator_index == next_withdrawal_validator_index
+
+    if len(expected_withdrawals) == spec.MAX_WITHDRAWALS_PER_PAYLOAD:
+        # NOTE: ideally we would also check in the case with
+        # fewer than maximum withdrawals but that requires the pre-state info
+        next_withdrawal_validator_index = (expected_withdrawals_validator_indices[-1] + 1) % len(state.validators)
+        assert state.next_withdrawal_validator_index == next_withdrawal_validator_index
+
     for index in fully_withdrawable_indices:
         if index in expected_withdrawals_validator_indices:
             assert state.balances[index] == 0
@@ -75,9 +80,13 @@ def run_withdrawals_processing(spec, state, execution_payload, num_expected_with
     yield 'post', state
 
     if len(expected_withdrawals) == 0:
-        assert state == pre_state
-    elif len(expected_withdrawals) < spec.MAX_WITHDRAWALS_PER_PAYLOAD:
-        assert len(spec.get_expected_withdrawals(state)) == 0
+        next_withdrawal_validator_index = (
+            pre_state.next_withdrawal_validator_index + spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
+        )
+        assert state.next_withdrawal_validator_index == next_withdrawal_validator_index % len(state.validators)
+    elif len(expected_withdrawals) <= spec.MAX_WITHDRAWALS_PER_PAYLOAD:
+        bound = min(spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP, spec.MAX_WITHDRAWALS_PER_PAYLOAD)
+        assert len(spec.get_expected_withdrawals(state)) <= bound
     elif len(expected_withdrawals) > spec.MAX_WITHDRAWALS_PER_PAYLOAD:
         raise ValueError('len(expected_withdrawals) should not be greater than MAX_WITHDRAWALS_PER_PAYLOAD')
 
@@ -154,10 +163,32 @@ def test_success_max_per_slot(spec, state):
 
 
 @with_phases([CAPELLA])
+@with_presets([MAINNET], reason="too few validators with minimal config")
+@spec_state_test
+def test_success_all_fully_withdrawable_in_one_sweep(spec, state):
+    assert len(state.validators) <= spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
+
+    withdrawal_count = len(state.validators)
+    fully_withdrawable_indices, partial_withdrawals_indices = prepare_expected_withdrawals(
+        spec, state, num_full_withdrawals=withdrawal_count)
+
+    next_slot(spec, state)
+    execution_payload = build_empty_execution_payload(spec, state)
+
+    yield from run_withdrawals_processing(
+        spec, state, execution_payload,
+        fully_withdrawable_indices=fully_withdrawable_indices,
+        partial_withdrawals_indices=partial_withdrawals_indices)
+
+@with_phases([CAPELLA])
+@with_presets([MINIMAL], reason="too many validators with mainnet config")
 @spec_state_test
 def test_success_all_fully_withdrawable(spec, state):
+    assert len(state.validators) > spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
+
+    withdrawal_count = spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
     fully_withdrawable_indices, partial_withdrawals_indices = prepare_expected_withdrawals(
-        spec, state, num_full_withdrawals=len(state.validators))
+        spec, state, num_full_withdrawals=withdrawal_count)
 
     next_slot(spec, state)
     execution_payload = build_empty_execution_payload(spec, state)
@@ -169,10 +200,32 @@ def test_success_all_fully_withdrawable(spec, state):
 
 
 @with_phases([CAPELLA])
+@with_presets([MAINNET], reason="too few validators with minimal config")
+@spec_state_test
+def test_success_all_partially_withdrawable_in_one_sweep(spec, state):
+    assert len(state.validators) <= spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
+
+    withdrawal_count = len(state.validators)
+    fully_withdrawable_indices, partial_withdrawals_indices = prepare_expected_withdrawals(
+        spec, state, num_partial_withdrawals=withdrawal_count)
+
+    next_slot(spec, state)
+    execution_payload = build_empty_execution_payload(spec, state)
+
+    yield from run_withdrawals_processing(
+        spec, state, execution_payload,
+        fully_withdrawable_indices=fully_withdrawable_indices,
+        partial_withdrawals_indices=partial_withdrawals_indices)
+
+@with_phases([CAPELLA])
+@with_presets([MINIMAL], reason="too many validators with mainnet config")
 @spec_state_test
 def test_success_all_partially_withdrawable(spec, state):
+    assert len(state.validators) > spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
+
+    withdrawal_count = spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
     fully_withdrawable_indices, partial_withdrawals_indices = prepare_expected_withdrawals(
-        spec, state, num_partial_withdrawals=len(state.validators))
+        spec, state, num_partial_withdrawals=withdrawal_count)
 
     next_slot(spec, state)
     execution_payload = build_empty_execution_payload(spec, state)
@@ -302,8 +355,8 @@ def test_fail_a_lot_partially_withdrawable_too_few_in_withdrawals(spec, state):
 @with_phases([CAPELLA])
 @spec_state_test
 def test_fail_a_lot_mixed_withdrawable_in_queue_too_few_in_withdrawals(spec, state):
-    prepare_expected_withdrawals(spec, state, num_full_withdrawals=spec.MAX_WITHDRAWALS_PER_PAYLOAD * 4,
-                                 num_partial_withdrawals=spec.MAX_WITHDRAWALS_PER_PAYLOAD * 4)
+    prepare_expected_withdrawals(spec, state, num_full_withdrawals=spec.MAX_WITHDRAWALS_PER_PAYLOAD,
+                                 num_partial_withdrawals=spec.MAX_WITHDRAWALS_PER_PAYLOAD)
 
     next_slot(spec, state)
     execution_payload = build_empty_execution_payload(spec, state)
@@ -624,7 +677,7 @@ def test_success_excess_balance_but_no_max_effective_balance(spec, state):
 @with_phases([CAPELLA])
 @spec_state_test
 def test_success_one_partial_withdrawable_not_yet_active(spec, state):
-    validator_index = len(state.validators) // 2
+    validator_index = min(len(state.validators) // 2, spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP - 1)
     state.validators[validator_index].activation_epoch += 4
     set_validator_partially_withdrawable(spec, state, validator_index)
 
@@ -638,7 +691,7 @@ def test_success_one_partial_withdrawable_not_yet_active(spec, state):
 @with_phases([CAPELLA])
 @spec_state_test
 def test_success_one_partial_withdrawable_in_exit_queue(spec, state):
-    validator_index = len(state.validators) // 2
+    validator_index = min(len(state.validators) // 2, spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP - 1)
     state.validators[validator_index].exit_epoch = spec.get_current_epoch(state) + 1
     set_validator_partially_withdrawable(spec, state, validator_index)
 
@@ -653,7 +706,7 @@ def test_success_one_partial_withdrawable_in_exit_queue(spec, state):
 @with_phases([CAPELLA])
 @spec_state_test
 def test_success_one_partial_withdrawable_exited(spec, state):
-    validator_index = len(state.validators) // 2
+    validator_index = min(len(state.validators) // 2, spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP - 1)
     state.validators[validator_index].exit_epoch = spec.get_current_epoch(state)
     set_validator_partially_withdrawable(spec, state, validator_index)
 
@@ -667,7 +720,7 @@ def test_success_one_partial_withdrawable_exited(spec, state):
 @with_phases([CAPELLA])
 @spec_state_test
 def test_success_one_partial_withdrawable_active_and_slashed(spec, state):
-    validator_index = len(state.validators) // 2
+    validator_index = min(len(state.validators) // 2, spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP - 1)
     state.validators[validator_index].slashed = True
     set_validator_partially_withdrawable(spec, state, validator_index)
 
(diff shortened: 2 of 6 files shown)


potuz:
I like this change and I think 2) is actually necessary cause it may be that all validators are withrawn in the sweep interval and then we would be stuck if we don't advance. I suggest changing the check to when there were less than the maximum, instead of none

dapplion (approved):
LGTM, 2**17 sounds like a reasonable choice for capella

specs/capella/beacon-chain.md
```diff
@@ -328,6 +335,10 @@ def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
         state.next_withdrawal_index = WithdrawalIndex(latest_withdrawal.index + 1)
         next_validator_index = ValidatorIndex((latest_withdrawal.validator_index + 1) % len(state.validators))
         state.next_withdrawal_validator_index = next_validator_index
+    else:
```
djrtwo, line 338: I kind of prefer this is a blocking condition (as it is the exception)

e.g.
```python
if len(expected_withdrawals == 0):
    next_val_index = ...
    state.next_val_index = ...
    return

latest_withdrawal = expected_ ...
state.next_withdrawal =...

**Review by djrtwo — COMMENTED** [review]
This needs consensus tests

I'm also loosely against the proposal. I don't think there has been a strong case made for the additional complexity and the selected current value

**Comment by djrtwo** [comment]
re discussion on latest cl call: my preference order is

1. Do nothing if doing a full sweep will not introduce any requisite complexity to ensure load is fine
2. If we'll have to add caching/precompute for this edge case (if this PR doesn't go through), then add this logic and a lower per-slot bound so we don't have to really think about it at all -- e.g. to 2^10 or something. Just make it trivially costly to avoid having to bench or optimize or worry at all.

Based on convos, I think (1) is uncertain so I'm now in camp (2)

**Comment by dapplion** [comment]
Run some benchmarks for Lodestar's implementation. Each case has 4 parameters
  - eb: chance next index has excess balance
  - eth1: chance next index has eth1 credentials
  - we: chance next index is withdrawable
  - wn: chance next index is withdrawn

This does not reflect mainnet since clusters of keys will act together controlled by the same operator. But it forces different counts of total sampled validators `smpl`. `nocache` means requiring a tree traversal per validator sample, absence of the tag means 1 random vector access per validator sample.

Worst case is not great but the node could survive with it. Note that this tests are for an old state with 250k vals. Mainnet should ~ double the worst case results. Would be interesting to analyze mainnet and understand how interlaced or not are indexes of independent actors.

```
  getExpectedWithdrawals
    ✔ vc - 250000 eb 1 eth1 1 we 0 wn 0 - smpl 15                         23630.61 ops/s    42.31800 us/op        -       2419 runs   5.45 s
    ✔ vc - 250000 eb 0.95 eth1 0.1 we 0.05 wn 0 - smpl 219                11216.92 ops/s    89.15100 us/op        -        304 runs   2.93 s
    ✔ vc - 250000 eb 0.95 eth1 0.3 we 0.05 wn 0 - smpl 42                 16537.95 ops/s    60.46700 us/op        -        703 runs   3.08 s
    ✔ vc - 250000 eb 0.95 eth1 0.7 we 0.05 wn 0 - smpl 18                 20744.31 ops/s    48.20600 us/op        -        317 runs   2.50 s
    ✔ vc - 250000 eb 0.1 eth1 0.1 we 0 wn 0 - smpl 1020                   3250.098 ops/s    307.6830 us/op        -       1120 runs   4.12 s
    ✔ vc - 250000 eb 0.03 eth1 0.03 we 0 wn 0 - smpl 11777                876.3505 ops/s    1.141096 ms/op        -        813 runs   3.77 s
    ✔ vc - 250000 eb 0.01 eth1 0.01 we 0 wn 0 - smpl 141069               93.73472 ops/s    10.66841 ms/op        -         64 runs   2.07 s
    ✔ vc - 250000 eb 0 eth1 0 we 0 wn 0 - smpl 250000                     55.27471 ops/s    18.09146 ms/op        -         31 runs   1.93 s
    ✔ vc - 250000 eb 0 eth1 0 we 0 wn 0 nocache - smpl 250000             13.51712 ops/s    73.98023 ms/op        -         13 runs   2.33 s
    ✔ vc - 250000 eb 0 eth1 1 we 0 wn 0 - smpl 250000                     28.66458 ops/s    34.88626 ms/op        -         19 runs   2.07 s
    ✔ vc - 250000 eb 0 eth1 1 we 0 wn 0 nocache - smpl 250000             9.634900 ops/s    103.7894 ms/op        -         12 runs   2.58 s
```

File: specs/capella/beacon-chain.md [hunk 2]
```diff
@@ -323,11 +330,17 @@ def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
     for expected_withdrawal, withdrawal in zip(expected_withdrawals, payload.withdrawals):
         assert withdrawal == expected_withdrawal
         decrease_balance(state, withdrawal.validator_index, withdrawal.amount)
-    if len(expected_withdrawals) > 0:
-        latest_withdrawal = expected_withdrawals[-1]
-        state.next_withdrawal_index = WithdrawalIndex(latest_withdrawal.index + 1)
-        next_validator_index = ValidatorIndex((latest_withdrawal.validator_index + 1) % len(state.validators))
-        state.next_withdrawal_validator_index = next_validator_index
+
+    if len(expected_withdrawals) == 0:
```
dapplion, line 334: Should advance the pointer every time withdrawals are not maxed
```suggestion
    if len(expected_withdrawals) < MAX_WITHDRAWALS_PER_PAYLOAD:
```

specs/capella/beacon-chain.md
```diff
@@ -323,11 +330,17 @@ def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
...
+        # advance sweep if there were no withdrawals
+        next_validator_index = (state.next_withdrawal_validator_index + MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP) % len(state.validators)
+        state.next_withdrawal_validator_index = ValidatorIndex(next_validator_index)
+        return
+
+    latest_withdrawal = expected_withdrawals[-1]
+    state.next_withdrawal_index = WithdrawalIndex(latest_withdrawal.index + 1)
+    next_validator_index = ValidatorIndex((latest_withdrawal.validator_index + 1) % len(state.validators))
+    state.next_withdrawal_validator_index = next_validator_index
```
dapplion, line 343: Should be changed to always move the pointer forward. For the case of 0 withdrawals, incomplete withdrawals, and max withdrawals.

```py
    if len(expected_withdrawals) < MAX_WITHDRAWALS_PER_PAYLOAD:
        next_validator_index = (state.next_withdrawal_validator_index + MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP) % len(state.validators)
    else:
        next_validator_index = ValidatorIndex((expected_withdrawals[-1].validator_index + 1) % len(state.validators))

    state.next_withdrawal_validator_index = next_validator_index
    state.next_withdrawal_index = WithdrawalIndex(state.next_withdrawal_index + len(expected_withdrawals))
```

rolfyone:
I had a quick look at mainnet...
Currently the first validator id with a 0x01 is 120374, and there's about 182k validators that have a 0x01 withdrawal address...

So in around 1.5 days we'd be doing all the withdrawals....

Given the number of 0x01 withdrawal addresses already set, we probably don't need to be too concerned with limiting the scan distance, as that first leap of 120374 is not that huge...

I guess that means every 1.5 days we'll be doing that large scan, until some of those update their withdrawal addresses, which probably won't take very long.

specs/capella/beacon-chain.md
```diff
@@ -323,11 +328,18 @@ def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
     for expected_withdrawal, withdrawal in zip(expected_withdrawals, payload.withdrawals):
         assert withdrawal == expected_withdrawal
         decrease_balance(state, withdrawal.validator_index, withdrawal.amount)
-    if len(expected_withdrawals) > 0:
-        latest_withdrawal = expected_withdrawals[-1]
-        state.next_withdrawal_index = WithdrawalIndex(latest_withdrawal.index + 1)
-        next_validator_index = ValidatorIndex((latest_withdrawal.validator_index + 1) % len(state.validators))
-        state.next_withdrawal_validator_index = next_validator_index
+
+    if len(expected_withdrawals) == 0:
```
potuz, line 332:
```suggestion
    if len(expected_withdrawals) < MAX_WITHRAWALS_PER_PAYLOAD:
```

ralexstokes: if there are more than 0 withdrawals, then the usual process takes over where you just go via the last validator you could get to

if we _always_ skip forward by the sweep bound then we would end up jumping around the validator set and lose the sequential nature of the scan

potuz: If you didn't achieve the maximum then we know for a fact that no validator was withdrawable this block in that sweep. There's no need to rescan in the next slot just to include the possible one that may have become withrawable during this one. It's more effective to just add here and continue from where we scanned in this loop

potuz: In other words, given your argument, why not just add 1 then when there are zero, instead of the full max bound? There may be validators that become withrawable in the next slot and adding one already keeps us alive

specs/capella/beacon-chain.md
dapplion: I don't understand why the full sweep can only happen in the 0 withdrawal case and not in the incomplete withdrawals case. You be scanning a range of validators twice in consecutive blocks due to not updating the pointer position.

dapplion:
> My point, I guess, is that maybe this scan limit isn't really needed?

Yes, it's great to know that block processing time worst case is bounded. This is a guarantee I really want to have unless the bounded sweep has drawbacks.

specs/capella/beacon-chain.md
ralexstokes: > If you didn't achieve the maximum then we know for a fact that no validator was withdrawable this block in that sweep.

no?

let's start at validator index 0, sweep 256 with block max of 16. 

if you get 0 withdrawals, then no validator was withdrawable so we advance the state of the sweep by 256
if we get 16 withdrawals, then it seems most fair to just start at validator index 16 next time
if you get some other number of withdrawals, then I think you want to start just after the index of the last withdrawal you got...

the only special case is when you have 0 withdrawals which this PR currently handles

I would not want to just add 1 to the validator index as that seems that we would duplicate inspections of validators per epoch (keep in mind that withdrawable status only changes at epoch boundaries, unless there is a deposit or if you are slashed which I'll argue are infrequent enough to not bias the 'fairness' of this choice of scan significantly)

potuz: > if you get some other number of withdrawals, then I think you want to start just after the index of the last withdrawal you got...
> 
I disagree with this, if you get less than 16 it means that you scanned 256 and found no withdrawable validators and you are making the point below that it's a good assumption that they will not be withdrawable the next slot, so I want to start next sweep at 256 more, and not the last index. 

> I would not want to just add 1 to the validator index as that seems that we would duplicate inspections of validators per epoch (keep in mind that withdrawable status only changes at epoch boundaries, unless there is a deposit or if you are slashed which I'll argue are infrequent enough to not bias the 'fairness' of this choice of scan significantly)

ralexstokes: ok I see your point and I'll update it, thanks for explaining!

rolfyone:
If we do this, our upper limit should be more like `2^17`, which would ensure under current mainnet conditions we don't have 120 blocks per sweep with no withdrawals (1k per sweep, first 120k ish are 0x00) ....
My mainnet test from a state i forwarded to capella had around 8ms to select from that, which i think is worth it to not have several epochs of blocks with no withdrawals.

ralexstokes:
> 1k per sweep, first 120k ish are 0x00

I'm not too worried about one-off or transition periods and would rather target a value that ensures security in both average case and worst case scenarios

I expect many of the existing `0x00` withdrawal credential validators to quickly swap over to the `0x01` format

potuz:
> I'm not too worried about one-off or transition periods and would rather target a value that ensures security in both average case and worst case scenarios

I agree with this, the first 120K figure is just a one-off. However I think in choosing this constant what we need to decide is about the cost of changing it requiring a hard fork and adapting tests. What will eventually make a less than optimal performance is when we hit enough gaps with fully withrawn validators. I have not seen any modeling to even guess what the expected time is to see 1024 consecutive fully withrawn validators, but the benchmarks that Dapplion produced do show that 2^17 is pretty safe and harmless. Either way, even if my preference would be for a larger constant I don't oppose 1024 as I'm only speculating that we will have hardforks regularly enough for the time being to change this if it becomes a problem

rolfyone:
> I'm not too worried about one-off or transition periods and would rather target a value that ensures security in both average case and worst case scenarios

The average case is all this code is not used really. If the limit is sane, and the network is performing 'well', then my expectation is the blocks basically contain 16 withdrawals each and every block.

Worst case is harder to predict.  This would likely have to be network-wide leak, where ultimately we're publishing 0 withdrawals because everyone is not in a state to generate one. In this case I do agree that a bound is desirable, but the number can be large I think. In this case each and every block we'd be scanning to find expected withdrawals and finding none, or very few. Given that, I could see the argument for a number of the order of 32k potentially, but i still think 2^17 probably gives a more consistent output in the normal case. 

I do think 1k is too low unless we're trying to use this change to directly reduce supply, in which case i'd suggest withdrawals per block is a more effective measure to tweak.

djrtwo:
> I do think 1k is too low unless we're trying to use this change to directly reduce supply, in which case i'd suggest withdrawals per block is a more effective measure to tweak.

I don't understand the relationship between this bound and the supply. Can you elaborate?

-------------------

My reasoning is to put it at a beyond-safe value so that no client even has to consider if they need to put in optimization code that  won't benefit in the all-but-disaster normal case. 1024 is low and I understand @potuz's concern with sparse indices in some future -- maybe the genesis validator set target is a nice compromise 2^14 (16384). That is a more-than-comfortable value in doing full v-set operations and keeps us in the clear on sparse indices for an order of magnitude longer.

(side note, I am still in favor of validator index re-use after 1 year of being fully withdrawn. but it's good to have these convos as though that is not ever happening)

rolfyone:
> I don't understand the relationship between this bound and the supply. Can you elaborate?

It's one effect that having 1k will have. each round we'll have at least 4+ epochs initially of empty withdrawals, so in that way it'll limit the number outgoing.
The current rate is also arguably too rapid at the moment, so it's something I've been looking at on the side. With 200k validators eligible as it stands, then you're looking at a withdrawal per eligible validator every 1.8 days, and with 500k eligible, more like every 4.5 days...

So the gaps do have that (positive?) potential effect of reducing the rate a little, but I was just saying if that's a goal we should be looking at the number of withdrawals included per block..

specs/capella/beacon-chain.md
```diff
@@ -323,9 +328,18 @@ def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
...
         latest_withdrawal = expected_withdrawals[-1]
         state.next_withdrawal_index = WithdrawalIndex(latest_withdrawal.index + 1)
+
+    if len(expected_withdrawals) < MAX_WITHDRAWALS_PER_PAYLOAD:
+        # advance sweep if there were not a full set of withdrawals
+        next_index = state.next_withdrawal_validator_index + MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
+        next_validator_index = next_index % len(state.validators)
+        state.next_withdrawal_validator_index = ValidatorIndex(next_validator_index)
+    else:
+        latest_withdrawal = expected_withdrawals[-1]
```
potuz, line 342:
```suggestion
```

djrtwo: You need the assignment in there or an outer assignment beacuse the prior if isn't necessarily hit. I cover that in my comment below

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
```diff
@@ -33,8 +33,13 @@ def verify_post_state(state, spec, expected_withdrawals,
...
     expected_withdrawals_validator_indices = [withdrawal.validator_index for withdrawal in expected_withdrawals]
     assert state.next_withdrawal_index == expected_withdrawals[-1].index + 1
-    next_withdrawal_validator_index = (expected_withdrawals_validator_indices[-1] + 1) % len(state.validators)
-    assert state.next_withdrawal_validator_index == next_withdrawal_validator_index
+
+    if len(expected_withdrawals) == spec.MAX_WITHDRAWALS_PER_PAYLOAD:
+        # NOTE: ideally we would also check in the case with
+        # fewer than maximum withdrawals but that requires the pre-state info
+        next_withdrawal_validator_index = (expected_withdrawals_validator_indices[-1] + 1) % len(state.validators)
+        assert state.next_withdrawal_validator_index == next_withdrawal_validator_index
```
djrtwo, line 41: Would we not just need to pass in `pre_latest_withdrawal_validator_index`?
I guess it'd be a bit annoying. not a strong preference either way

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
```diff
@@ -75,9 +80,13 @@ def run_withdrawals_processing(spec, state, execution_payload, num_expected_with
...
 
     if len(expected_withdrawals) == 0:
-        assert state == pre_state
+        next_withdrawal_validator_index = (
+            pre_state.next_withdrawal_validator_index + spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
+        )
+        assert state.next_withdrawal_validator_index == next_withdrawal_validator_index % len(state.validators)
     elif len(expected_withdrawals) < spec.MAX_WITHDRAWALS_PER_PAYLOAD:
-        assert len(spec.get_expected_withdrawals(state)) == 0
+        bound = min(spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP, spec.MAX_WITHDRAWALS_PER_PAYLOAD)
```
djrtwo, line 88: this seems like a weird code path. I think we should just consider configs (we make very few) in which the withdrawals per payload is less than the sweep bound as invalid.

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
```diff
@@ -156,8 +165,9 @@ def test_success_max_per_slot(spec, state):
 @with_phases([CAPELLA])
 @spec_state_test
 def test_success_all_fully_withdrawable(spec, state):
+    withdrawal_count = min(len(state.validators), spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP)
```
djrtwo, line 168: I would frame this as a sanity assert to make it clearer what is being tested

```suggestion
    assert len(state.validators) > spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
    withdrawal_count = len(state.validators)
    withdrawal_count = min(len(state.validators), spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP)
```

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
```diff
@@ -171,8 +181,9 @@ def test_success_all_fully_withdrawable(spec, state):
 @with_phases([CAPELLA])
 @spec_state_test
 def test_success_all_partially_withdrawable(spec, state):
+    withdrawal_count = min(len(state.validators), spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP)
```
djrtwo, line 184: ditto

djrtwo (commented):
I need to do a bit of a test pass still but wanted to get the initial comments out.

thanks!

specs/capella/beacon-chain.md
potuz: > You need the assignment in there or an outer assignment beacuse the prior if isn't necessarily hit. I cover that in my comment below

no because this else is only hit when `len(expected_withdrawals) >= MAX_WITHDRAWALS_PER_PAYLAOD`  in which case the previous assignment already was hit.

ralexstokes: so I'm trying to avoid python's dynamic scoping here and so just elected to make the assignment twice

if we are ok w/ imo non-standard (relative to what most readers are assumed to know) scoping behavior then we can drop the extra assignment

specs/capella/beacon-chain.md
```diff
@@ -323,9 +328,18 @@ def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
...
         next_validator_index = ValidatorIndex((latest_withdrawal.validator_index + 1) % len(state.validators))
         state.next_withdrawal_validator_index = next_validator_index
```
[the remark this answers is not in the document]
ralexstokes: I can add more comments but we can't pull out the `latest_withdrawal = expected_withdrawals[-1]` assignment from its current place bc `len(expected_withdrawals) == 0` is possible which would raise an error

and there is no 'safe' built-in method for indexing on lists so there isn't some kind of pythonic thing we could do (which imo we don't want to actually do)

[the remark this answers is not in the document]
djrtwo: hmm it isn't safe as currently written either though. The `else` isn't scoped by non-zero expected_withdrawals. will write an updated suggestion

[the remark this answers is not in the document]
potuz: > hmm it isn't safe as currently written either though. The `else` isn't scoped by non-zero expected_withdrawals. will write an updated suggestion

It is, I'm fine with @ralexstokes' intent to explicitly assign in case people get confused by the scoping, but it's perfectly valid to not have the second `latest_withdrawal` assignment.
```shell
$cat danny.py
danny = 1
if 1 > 0:
    danny = 4
if 2 > 3:
    danny = 1
else:
    print(danny)

$ python danny.py
4
```
```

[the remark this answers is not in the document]
Comment (djrtwo) [reply specs/capella/beacon-chain.md:344]: okay, how about this. I really don't like relying on python's scoping rules. If not this, then I'd do two var assignments in different places 

```suggestion
    
    # Update the next withdrawal index if this block contained withdrawals
    if len(expected_withdrawals) != 0:
        state.next_withdrawal_index = WithdrawalIndex(expected_withdrawals[-1].index + 1)

    # Update the next validator index to start the next withdrawal sweep
    if len(expected_withdrawals) == MAX_WITHDRAWALS_PER_PAYLOAD:
        # Next sweep starts after the latest withdrawal's validator index
        next_validator_index = ValidatorIndex((expected_withdrawals[-1].validator_index + 1) % len(state.validators))
        state.next_withdrawal_validator_index = next_validator_index
    else:
        # Advance sweep by the max length of the sweep if there was not a full set of withdrawals
        next_index = state.next_withdrawal_validator_index + MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
        next_validator_index = next_index % len(state.validators)
        state.next_withdrawal_validator_index = ValidatorIndex(next_validator_index)
```

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
ralexstokes: this is less about the size of the sweep and more just about the scenario where you do the full sweep (of whatever size) and only have a small number of withdrawals

I think all three arms are realistic cases -- do you see it differently?

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
ralexstokes: we can do this but then this becomes a "mainnet only" test

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
potuz: but isn't the bound always going to be `MAX_WITHDRAWALS_PER_PAYLOAD`  here? I mean there is no sense in defining a config with `MAX_VALIDATOR_PER_WITHDRAWAL_SWEEP`  being smaller since then we would never attain the `MAX_WITHDRAWAL_PER_PAYLOAD`  and therefore is equivalent to simply lower it.

specs/capella/beacon-chain.md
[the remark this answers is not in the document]
ralexstokes: this seems fine, I've made the update

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
ralexstokes: you are right that for the current parameterization across `mainnet` and `minimal` the bound will always be `MAX_WITHDRAWALS_PER_PAYLOAD` although that is not necessarily the case

we can separately make the commitment that we will never do this but its the type of thing that is hard to track and then enforce...

separately there is the point about this arm of this conditional block and I am imagining a scenario where there are a handful of validators, less than `MAX_WITHDRAWALS_PER_PAYLOAD`, which are withdrawable e.g. going into an inactivity leak

specs/capella/beacon-chain.md
```diff
@@ -323,11 +328,23 @@ def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
...
+    # Update the next withdrawal index if this block contained withdrawals
+    if len(expected_withdrawals) != 0:
         latest_withdrawal = expected_withdrawals[-1]
         state.next_withdrawal_index = WithdrawalIndex(latest_withdrawal.index + 1)
+
+    # Update the next validator index to start the next withdrawal sweep
+    if len(expected_withdrawals) == MAX_WITHDRAWALS_PER_PAYLOAD:
+        # Next sweep starts after the latest withdrawal's validator index
+        latest_withdrawal = expected_withdrawals[-1]
         next_validator_index = ValidatorIndex((latest_withdrawal.validator_index + 1) % len(state.validators))
```
djrtwo, line 341: suggest the following for consistency (it is 117 chars so in linter limit)
```suggestion
        next_validator_index = ValidatorIndex((expected_withdrawals[-1].validator_index + 1) % len(state.validators))
```

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
```diff
@@ -154,10 +163,14 @@ def test_success_max_per_slot(spec, state):
 
 
 @with_phases([CAPELLA])
+@with_presets([MAINNET], reason="too few validators with minimal config")
```
djrtwo, line 166: Apologies, I was a bit confused when I last commented.

Suggest we do a case where `state.validators > MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP` and the validators are all withdrawable

Both are valuable (as you had implicitly before between the two configs) but separating them out to be very clear which of the two scenarios is being tested is my preference

specs/capella/beacon-chain.md
```diff
@@ -323,11 +328,23 @@ def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
...
         decrease_balance(state, withdrawal.validator_index, withdrawal.amount)
-    if len(expected_withdrawals) > 0:
+
+    # Update the next withdrawal index if this block contained withdrawals
+    if len(expected_withdrawals) != 0:
         latest_withdrawal = expected_withdrawals[-1]
         state.next_withdrawal_index = WithdrawalIndex(latest_withdrawal.index + 1)
+
+    # Update the next validator index to start the next withdrawal sweep
+    if len(expected_withdrawals) == MAX_WITHDRAWALS_PER_PAYLOAD:
```
dapplion, line 338: For the `len(expected_withdrawals) == MAX_WITHDRAWALS_PER_PAYLOAD` case `state.next_withdrawal_index` is updated twice. Is that okay? Why not do
```
if len == MAX_WITHDRAWALS_PER_PAYLOAD
elif len > 0
else
```

potuz: Where is the second update? I only see it set once.

dapplion: Oh, miss-read `next_withdrawal_index` for `next_withdrawal_validator_index`. All good then :+1:

djrtwo: the names are confusing to read as is :/

ralexstokes: agreed :/

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
ralexstokes: if im following you correctly, then I've updated the PR to have each of these scenarios called out explicitly

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
```diff
@@ -154,10 +163,32 @@ def test_success_max_per_slot(spec, state):
...
+        spec, state, num_full_withdrawals=withdrawal_count)
+
+    next_slot(spec, state)
+    execution_payload = build_empty_execution_payload(spec, state)
+
+    yield from run_withdrawals_processing(
+        spec, state, execution_payload,
+        fully_withdrawable_indices=fully_withdrawable_indices,
+        partial_withdrawals_indices=partial_withdrawals_indices)
+
```
djrtwo, line 182:
```suggestion


```

tests/core/pyspec/eth2spec/test/capella/block_processing/test_process_withdrawals.py
```diff
@@ -75,9 +80,13 @@ def run_withdrawals_processing(spec, state, execution_payload, num_expected_with
...
-        assert len(spec.get_expected_withdrawals(state)) == 0
+        next_withdrawal_validator_index = (
+            pre_state.next_withdrawal_validator_index + spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
+        )
+        assert state.next_withdrawal_validator_index == next_withdrawal_validator_index % len(state.validators)
+    elif len(expected_withdrawals) <= spec.MAX_WITHDRAWALS_PER_PAYLOAD:
+        bound = min(spec.MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP, spec.MAX_WITHDRAWALS_PER_PAYLOAD)
+        assert len(spec.get_expected_withdrawals(state)) <= bound
     elif len(expected_withdrawals) > spec.MAX_WITHDRAWALS_PER_PAYLOAD:
         raise ValueError('len(expected_withdrawals) should not be greater than MAX_WITHDRAWALS_PER_PAYLOAD')
```
djrtwo, line 92:
```suggestion
```

specs/capella/beacon-chain.md
```diff
@@ -80,6 +78,12 @@ We define the following Python custom types for type hinting and readability:
 | - | - | - |
 | `MAX_WITHDRAWALS_PER_PAYLOAD` | `uint64(2**4)` (= 16) | Maximum amount of withdrawals allowed in each payload |
 
+### Withdrawals processing
+
+| Name | Value |
+| - | - |
+| `MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP` | `1024` (= 2**10 ) |
```
djrtwo, line 85:
```suggestion
| `MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP` | `16384` (= 2**14 ) |
```

dapplion (approved):

lucassaldanha (approved):
LGTM. Teku has already implemented the new logic in a branch and is ready to go! 🚀

PR #3096, opened 2022-11-11
fix roots-of-unity assertion checks

roberto-bayardo:


roberto-bayardo:
@kevaundray informs me that this (somewhat magically, to me anyway) works fine as is.  Closing out, but might be worth a comment inside the permutation function that despite the element bit shifting, the result indeed a simple permutation.

PR #3097, opened 2022-11-11
EIP4844: Refactor `verify_kzg_proof()` to receive bytes (used in precompile)

asn-d6:
We received reports by client devs that the precompile was annoying to implement because of the modular reductions that had to be done outside of the crypto layer.

This PR makes `verify_kzg_proof()` bit more high-level so that client devs can interact with it by just passing bytes to it.

Note: Client devs still have to check the `assert kzg_to_versioned_hash(data_kzg) == versioned_hash` when implementing the precompile.

### DIFF
--- a/specs/eip4844/polynomial-commitments.md
+++ b/specs/eip4844/polynomial-commitments.md
@@ -46,6 +46,8 @@
 
 This document specifies basic polynomial operations and KZG polynomial commitment operations as they are needed for the EIP-4844 specification. The implementations are not optimized for performance, but readability. All practical implementations should optimize the polynomial operations.
 
+Functions flagged as "Public method" MUST be provided by the underlying KZG library as public functions. All other functions are private functions used internally by the KZG library.
+
 ## Custom types
 
 | Name | SSZ equivalent | Description |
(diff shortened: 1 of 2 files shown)


roberto-bayardo:
I assume there will be a sister change for the EIP-4844 itself which still has the precompile function doing the conversion to field element?

FWIW, one could just put PointEvaluationPrecompile itself in the kzg library for all clients to share (which was what I was doing in our geth fork), as it already has a bytes api:

`def point_evaluation_precompile(input: Bytes) -> Bytes:`

But this change still makes sense if one wants to separate the EL and CL concerns of 4844.

My 4844 Golang bytes-api implementation from geth which is also shared with our Prysm fork can be found below. It's a work in progress but feedback is welcome. I've updated to incorporate the changes proposed here.

https://github.com/mdehoog/go-ethereum/tree/more-kzg-2/crypto/kzg

asn-d6:
> I assume there will be a sister change for the EIP-4844 itself which still has the precompile function doing the conversion to field element?

Yep, for sure we also need to update the EIP!

Afte we stabilize the crypto API again on the consensus-specs side (pretty much just this ticket and https://github.com/ethereum/consensus-specs/pull/3093 remaining) we will sync up the EIP.

> FWIW, one could just put PointEvaluationPrecompile itself in the kzg library for all clients to share (which was what I was doing in our geth fork), as it already has a bytes api:

Yep, that's another reasonable approach. No strong opinion here. Whatever is easier for client devs. Let's raise it on the next call?

dankrad (approved):

PR #3101, opened 2022-11-13
Fix typos

terencechain:
Fixed a few spelling typos


djrtwo (approved):
thanks! seems our spellcheck isn't working on everything...

as for the numbering of lists, we accept doing 1. 1. 1. because markdown renders it properly. but it's fine to number "correctly" as well

PR #3102, opened 2022-11-14
Updated capella withdrawability delay

rolfyone:
In order to allow for validators that may be in sync committees, the withdrawability delay needs to be increased to double the Period of time that a validator could be in the current or next sync committee.

This will ensure that it is not possible to have a validator that is exited and withdrawable within the time that they can also be in a sync committee.

This does invalidate some test cases where an exited validator was in a sync committee and withdrawn during that window.

Note that this affects all validators that are exited, whether in a sync committee or not by applying this configuration update, so exited validators that were not in the current or next committee will still be waiting an extra 27+ hours to get funds, plus whatever the sweep time takes to process withdrawals.

I wasn't sure where to make the yaml changes, hopefully this isn't too far off of correct...


mcdee:
No comment on the desirability of the change, but technically it needs to be `MIN_VALIDATOR_WITHDRAWABILITY_DELAY_CAPELLA`; see https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/beacon-chain.md#updated-penalty-values as an example of some constants that changed with a hard fork.

potuz:
I'll leave here the same comment as in TG, I think the penalty on the UX of all validators outweighs the minimum gain of being able to penalize validators in the sync committee that exit.

ralexstokes:
I agree it feels off that we can have this duty while the validator has left the "active" part of their lifecycle; however, I will note that currently sync committee members have "nothing at stake" (i.e. no slashing conditions for sync committee equivocations) and so it actually doesn't change protocol security when a validator is still in the sync committee while not being "active".

This fact combined with the loss in UX around withdrawals for the entire validator set suggests to me we do not implement this change.

djrtwo:
I'm not in favor for same reasons as @potuz and @ralexstokes. The fact that you could exit while in sync committee was known (and tested) for Altair. Being withdrawable in such a state is an implication and a bit weird, but does not break anything and is easy to provide vectors for.

Also changing the constant at the fork doesn't 100% avoid the potential for it happening as someone might already be exiting/exited and in a committee at the time of the Capella fork. Thus (without more complexity to avoid it) we'd have to code and test the case anyway.

I'd rather not change the delay to only avoid something that is relatively simple to handle. and I'd also not really be interseted in adding other complexity (e.g. trying to block exits of sync committee members) to avoid this

rolfyone:
This was basically done so that people don't underflow, as one client at least had issues. If you're using minusMinZero it's not a problem, but without this change a reference test does catch the problem currently...

rolfyone:
I'm happy to close this, I mostly raised it to capture this discussion. Thanks for comments.

PR #2639, opened 2021-10-02
Mock Engine API req/resp in tests - take 2

hwwhww:
Replace #2637
It may be better to start with small changes rather than writing overcomplicated scripts.

### 1. Use validator guide helpers to generate the execution payload for `on_merge_block` tests
See `run_get_execution_payload_with_mock_engine_get_payload`.

- Comparing to #2637, this PR doesn't output the block production API calls logs.
- Using validator guide `get_execution_payload` would improve our test coverage a bit. 🙂 We can also create validator guide unit tests later in other PRs.

### 2. Fork choice rule test format updates
This PR changes the existing fork choice tests output:
1.  Add one more layer to the execution steps.
    - e.g., previous `{'attestation': attestation_0x<attestation_root>}` is now `{'on_attestation': {'attestation': 'attestation_0x<attestation_root>'}}`
    - it allows us to add more description to the `on_block` step.
2. Change the `pow_block` output
    - make it only shows when `get_pow_block` helper is called.
    - e.g.,
      ```yaml
      - on_block:
          ...
          block: block_0x6f0e0d6d65481fc1142bde4234b8537a502facc3b26168046cfcecd95e6bdc8a
          get_pow_block:
            input: {block_hash: 0x922ef34b6ca455c5d13de0ac5db0f6d7c1c26adfa026141a2e086fb155deb6b4}
            output: {result: pow_block_0x006a9edb9f8316403009d1cf96c21907ddfafe86c5701c49cd0d57abab2a8a7b}
      ```
3. Output mocking `engine_executePayload` request and response with `on_block`
    - e.g., if an `on_block` step triggers `process_execution_payload` and `engine_executePayload`: 
      ```yaml
      - on_block:
          ...
          process_execution_payload:
            engine_api:
              request:
                method: engine_executePayload
                params:
                  parentHash: '0x006a9edb9f8316403009d1cf96c21907ddfafe86c5701c49cd0d57abab2a8a7b'
                  coinbase: '0x1212121212121212121212121212121212121212'
                  stateRoot: '0x0000000000000000000000000000000000000000000000000000000000000000'
                  receiptRoot: '0x6e6f207265636569707473206865726500000000000000000000000000000000'
[...]

PR #3105, opened 2022-11-16
Send store.finalized block hash to EL

mkalinin:
Explicitly states that store's finalized checkpoint must be used to obtain `finalized_block_hash` that is sent to EL. Apparently, all CL clients are already following this logic.

@hwwhww we would need to create tests checking that `store.finalized` used instead of `head_state.finalized` to obtain `finalized_block_hash`. These values can't diverge according to current specification, as tips with finalized checkpoint not matching to the store's one are filtered out and not eligible for the head. Not sure if we can implement such tests without changing the fork choice spec.

### DIFF
--- a/specs/bellatrix/fork-choice.md
+++ b/specs/bellatrix/fork-choice.md
@@ -77,9 +77,13 @@ The `finalized_block_hash` parameter MUST be set to return value of the followin
 def get_finalized_execution_payload_hash(store: Store) -> Hash32:
     finalized_block_root = store.finalized_checkpoint.root
     finalized_block = store.blocks[finalized_block_root]
-
-    # Returns Hash32() if no payload is yet finalized
-    return finalized_block.body.execution_payload.block_hash
+    
+    # Consider finalized blocks before Bellatrix
+    # Return Hash32() if no payload is yet finalized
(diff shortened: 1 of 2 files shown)


hwwhww:
@mkalinin

> we would need to create tests checking that `store.finalized` used instead of `head_state.finalized` to obtain `finalized_block_hash`. These values can't diverge according to current specification, as tips with finalized checkpoint not matching to the store's one are filtered out and not eligible for the head. Not sure if we can implement such tests without changing the fork choice spec.

To provide test vectors, we need a new test format to mock `notify_forkchoice_updated`. I think it may be something like #2639 proposed, but probably in a simpler form. What do you think?

---

I pushed basic unit tests (https://github.com/ethereum/consensus-specs/pull/3105/commits/270d930a22305db451adc93390f053b565f2341a) to verify code logic of this PR.

mkalinin:
> To provide test vectors, we need a new test format to mock `notify_forkchoice_updated`. I think it may be something like #2639 proposed, but probably in a simpler form. What do you think?

Do you think that by mocking `notify_forkchoice_updated` we can bypass the requirement of `store.finalized` and  `head_state.finalized` to diverge? Or you think this requirement is not needed to be satisfied to write tests covering proposed change?

leolara:

PR #3107, opened 2022-11-16
cleanup explicit fork enumerations in tests

etan-status:
Fork epoch override test decorators were still referring to `SHARDING`. Replace the implementation with one that needs less maintenance.


hwwhww (approved):
Thanks for the refactoring! This PR makes it easier to maintain 👍👍

I've tested it with `gen_sanity` test vector generator.

Issue #3103, opened 2022-11-14
BlobSideCar mutable in gossip


djrtwo:
`beacon_block_and_blobs_sidecar` gossip validations are not sufficient to protect against any node mutating the sidecar after the proposer begins the gossip. As there is both no (1) proposer signature on the sidecar payload or (2) a cryptographic check that the commitments in the block map to the blobs in the sidecar.

EDIT: some background on gossip validations, we generally want to do minimum sufficient crypto-economic and validity checks to ensure the payload is at least "worth" gossiping. Usually this is a signature check to know it is from the expected actor and then whatever relatively cheap validity checks you can do to make sure it's generally well-formed. Here we have no crypto-economic commitment taht the adjoining payload was even made by the expected actor

So we can either sign the payload and do another sig check or we can  run the full `validate_blobs_sidecar`.
@Inphi estimates `validate_blobs_sidecar` as ~12ms @ 2MB (4ms at 256kb) whereas the signature check @terencechain estimates at 1 to 4ms. If we run `validate_blobs_sidecar` will it be dominant wrt time? If so, we need to be careful here. If not, lets just do that.

kevaundray:
Reposting here:

goos: darwin
goarch: amd64
pkg: github.com/crate-crypto/go-proto-danksharding-crypto/agg_kzg
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkAggVerify1-16                          2499203 ns/op
BenchmarkAggVerify2-16                          3278393 ns/op
BenchmarkAggVerify3-16                          4033615 ns/op
BenchmarkAggVerify4-16                          4844001 ns/op
BenchmarkAggVerify5-16                          5763020 ns/op
BenchmarkAggVerify6-16                          6500817 ns/op
BenchmarkAggVerify7-16                          7366121 ns/op
BenchmarkAggVerify8-16                          8267168 ns/op
BenchmarkAggVerify9-16                          8857999 ns/op
BenchmarkAggVerify10-16                         9594048 ns/op
BenchmarkAggVerify11-16                        10273552 ns/op
BenchmarkAggVerify12-16                        14101749 ns/op
BenchmarkAggVerify13-16                        13103934 ns/op
BenchmarkAggVerify14-16                        13431766 ns/op
BenchmarkAggVerify15-16                        15368090 ns/op
BenchmarkAggVerify16-16                        14334650 ns/op

Legend:

[...]

PR #3108, opened 2022-11-16
Make `validate_blobs_sidecar` a gossip condition

realbigsean:
Discussion here https://github.com/ethereum/consensus-specs/issues/3103

Adds `validate_blobs_sidecar` as a gossip condition to ensure blobs won't be propagated if commitments in blocks don't match blobs.


tbenr:
I had some comments about gossip conditions here #3109

djrtwo (approved):
Looks good!

PR #3121, opened 2022-11-19
Apply `with_config_overrides` to all phases

etan-status:
When defining a fork transition test, additional spec forks are made available through `@with_phases(..., other_phases=...)`. The `with_config_overrides` decorator only applies to the primary phase so far, which can be unexpected. `with_config_overrides` is adjusted to override config in subsequent `other_phases` as well.


hwwhww (approved):
lgtm!

PR #3123, opened 2022-11-22
Fix initialization of VERSIONED_HASH_VERSION_KZG

zilm13:
See https://github.com/protolambda/remerkleable/issues/15
In current `dev` inits with `00`
btw, checked the whole spec and have not found another similar bad BytesXX initializations


hwwhww (approved):
Nice catch! 👍 😱

Note: EIP-4844 test vectors would need this fix. /cc @djrtwo

PR #3125, opened 2022-11-22
EIP4844: remove `is_data_available`

terencechain:
I don't think `is_data_available` is much more relevant after we coupled block/sidecar under  `SignedBeaconBlockAndBlobsSidecar`. So this PR took a stab at removing it.
But I do think in p2p spec, we should add a condition to ensure blob ancestors are valid before importing the block


djrtwo (approved):
I like making clear the abstraction of DA in the spec, especially because the sidecar is not a 1st class consensus object.

That said, not too strongly held position

dankrad:
I kind of like the idea of keeping it as it would make it very easy to transition from 4844 to full sharding just be changing the implementation of that function.

Also, are there no other ways of getting blocks? How about when syncing?

dapplion (changes requested):
While not impactful in the implementation, the existence of `is_data_available` helped me when reading the specs for the first time. I think we should leave it, since it consistently ties to the future roadmap, and hints on its implications.

terencechain:
@dapplion perhaps we can just remove the `processed further optimistically... ` part, that's ultimately what we decide against

tbenr:
I'm ok with leaving it. My current implementation is taking in account that data might be retrieved async in the future.

tbenr:
> remove the `processed further optimistically...` part

👍

terencechain:
@tbenr @dapplion please take another look 🙏🏼

dapplion (approved):
LGTM, but you should rename PR to something like "remove optimistic sync mentions", until we have better consensus on it

djrtwo:
heh, I just realized I provided dissent but clicked "approve". Didn't intend to appve

djrtwo:
I'm good with this now. @terencechain can you patch the description of the issue given the scope of the issue now?

terencechain:
> I'm good with this now. @terencechain can you patch the description of the issue given the scope of the issue now?

Sorry I missed that. I just updated the description

PR #3128, opened 2022-11-23
changed is_better_update comparison logic for light client updates

zsfelfoldi:
This PR proposes a change to the LightClientUpdate comparison logic. The `has_relevant_sync_committee` check is moved to the front because the update is only useful for light client syncing if this is true and therefore a useful update should never be replaced with a non-useful one, even if it has more signers.
The finality and attested header slot are also considered differently; in this version the main considered concern was the potential risk that the chain might have reorged before the sync period boundary after the attested header was signed and therefore the `next_sync_committee` might have changed in the final canonical chain. The current logic always prefers older updates in case of ties, which is definitely bad from a security perspective in case of lower signer participation. Though the participation on mainnet is currently great, it would be good to be as resilient as possible to a short non-finalizing period which could happen as a result of large scale attacks or network issues. Choosing non-finalized updates from anywhere but the beginning of the sync period is something that improves security in such a case and does not really cost anything.

The reorg risk is zero if a block referring to the given `next_sync_committee` has been finalized (only sync committee finality is considered, if a header from the previous period is linked as finalized then it does not affect the reorg security).
Supermajority signature is also a condition of `has_sync_committee_finality`. An update with finality (including supermajority) always wins over one without it.
If `has_sync_committee_finality` is false for both updates then a penalty is subtracted from the score which is otherwise simply the signer count. This penalty decreases exponentially with `update.attested_header.slot % 8192` as the risk of reorg before the period boundary is also expected to decrease exponentially. The penalty starts from 1024 and halves every 32 slots. These constants are mostly arbitrary and definitely very conservative (provide a generous overestimate of this risk) but still, after a few hundred slots the penalty becomes negligible. In practice this means that the best update will not be chosen from the first few hundred slots of the period and after that the later slots are slightly preferred over the older ones (after 320 slots this penalty value becomes less than 1 and therefore it only matters if the old and new updates have the same amount of signers).

Note: this algorithm was originally implemented in the new Geth light client code: (see `betterThan` and `reorgRiskPenalty` functions)
https://github.com/zsfelfoldi/go-ethereum/blob/blsync/light/beacon/sync_committee_update.go#L95
I would also be fine with any alternative proposal that is at least as good as this one but I would definitely like to change the current spec. I've seen this happening on devnets that the signer count was below 2/3 but it was more or less stable throughout the period and therefore the update was chosen from the very beginning of the period.


leolara:

PR #3129, opened 2022-11-23
fix return value of compute_challenges

roberto-bayardo:


dankrad:
Isn't it actually `Tuple[Sequence[BLSFieldElement], BLSFieldElement]`

roberto-bayardo:
> Isn't it actually `Tuple[Sequence[BLSFieldElement], BLSFieldElement]`

Yup, my second commit was supposed to make that fix but apparently I didn't git add the file. :man_facepalming: 

thanks for fixing & merging.

PR #3131, opened 2022-11-24
Add link to Capella's P2P doc

potuz:
Add a link to Capella's P2P doc to the README, got tired of typing in the browser's bar instead of clicking.


hwwhww (approved):
thanks!

PR #3134, opened 2022-11-28
Add EIP4844 pylint and Mypy checks

hwwhww:
## Issue
`eip4844` was not enabled in our pylint and Mypy lint command.

## Fixes
- `Makefile`: Add EIP4844 pylint and Mypy checks
- specs: fix linter errors
    - Note: updated many `int` <> `BLSFieldElement` conversions in this PR. /cc @asn-d6   
- `setup.py` spec builder:
    1. `Polynomial`: to make Mypy happy, using `class Polynomial(Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]): pass` rather than `Polynomial = Vector[BLSFieldElement, FIELD_ELEMENTS_PER_BLOB]`.
    2. **???** `Blob`: somehow Mypy throws an error "Invalid base class "ByteVector"" at `class Blob(ByteVector[BYTES_PER_FIELD_ELEMENT * FIELD_ELEMENTS_PER_BLOB]): pass`.
        - I'm not sure why `Vector` is acceptable but `ByteVector` is invalid to Mypy here. They are both dynamic base classes that Mypy might not be able to handle. My guess is that `remerkleable` might have ignored types in `Vector` class implementation...? Right now my workaround is manually adding `type: ignore` in the `ByteVector` custom type definitions. /cc @protolambda, you may have some insights on it.

### DIFF
--- a/specs/eip4844/beacon-chain.md
+++ b/specs/eip4844/beacon-chain.md
@@ -284,6 +284,7 @@ def process_execution_payload(state: BeaconState, payload: ExecutionPayload, exe
 
 ```python
 def process_blob_kzg_commitments(state: BeaconState, body: BeaconBlockBody):
+    # pylint: disable=unused-argument
     assert verify_kzg_commitments_against_transactions(body.execution_payload.transactions, body.blob_kzg_commitments)
 ```
 
--- a/specs/eip4844/beacon-chain.md
+++ b/specs/eip4844/beacon-chain.md
@@ -175,10 +175,13 @@ but MUST NOT be considered valid until a valid `BlobsSidecar` has been downloade
 def is_data_available(slot: Slot, beacon_block_root: Root, blob_kzg_commitments: Sequence[KZGCommitment]) -> bool:
     # `retrieve_blobs_sidecar` is implementation dependent, raises an exception if not available.
     sidecar = retrieve_blobs_sidecar(slot, beacon_block_root)
-    if sidecar == "TEST":
-        return True  # For testing; remove once we have a way to inject `BlobsSidecar` into tests
-    validate_blobs_sidecar(slot, beacon_block_root, blob_kzg_commitments, sidecar)
 
+    # For testing, `retrieve_blobs_sidecar` returns "TEST.
+    # TODO: Remove it once we have a way to inject `BlobsSidecar` into tests.
+    if isinstance(sidecar, str):
+        return True
+
+    validate_blobs_sidecar(slot, beacon_block_root, blob_kzg_commitments, sidecar)
     return True
 ```
 
@@ -216,7 +219,7 @@ def tx_peek_blob_versioned_hashes(opaque_tx: Transaction) -> Sequence[VersionedH
 ```python
 def verify_kzg_commitments_against_transactions(transactions: Sequence[Transaction],
                                                 kzg_commitments: Sequence[KZGCommitment]) -> bool:
-    all_versioned_hashes = []
+    all_versioned_hashes: List[VersionedHash] = []
     for tx in transactions:
         if tx[0] == BLOB_TX_TYPE:
             all_versioned_hashes += tx_peek_blob_versioned_hashes(tx)
@@ -283,7 +286,7 @@ def process_execution_payload(state: BeaconState, payload: ExecutionPayload, exe
 #### Blob KZG commitments
 
 ```python
-def process_blob_kzg_commitments(state: BeaconState, body: BeaconBlockBody):
+def process_blob_kzg_commitments(state: BeaconState, body: BeaconBlockBody) -> None:
     # pylint: disable=unused-argument
     assert verify_kzg_commitments_against_transactions(body.execution_payload.transactions, body.blob_kzg_commitments)
 ```
--- a/specs/eip4844/polynomial-commitments.md
+++ b/specs/eip4844/polynomial-commitments.md
@@ -144,7 +144,7 @@ def bytes_to_bls_field(b: Bytes32) -> BLSFieldElement:
     """
     Convert 32-byte value to a BLS field scalar. The output is not uniform over the BLS field.
     """
-    return int.from_bytes(b, ENDIANNESS) % BLS_MODULUS
+    return BLSFieldElement(int.from_bytes(b, ENDIANNESS) % BLS_MODULUS)
 ```
 
 #### `blob_to_polynomial`
@@ -210,7 +210,7 @@ def bls_modular_inverse(x: BLSFieldElement) -> BLSFieldElement:
     Compute the modular inverse of x
     i.e. return y such that x * y % BLS_MODULUS == 1 and return 0 for x == 0
     """
-    return pow(x, -1, BLS_MODULUS) if x != 0 else 0
+    return BLSFieldElement(pow(x, -1, BLS_MODULUS)) if x != 0 else BLSFieldElement(0)
 ```
 
 #### `div`
@@ -220,7 +220,7 @@ def div(x: BLSFieldElement, y: BLSFieldElement) -> BLSFieldElement:
     """
     Divide two field elements: ``x`` by `y``.
     """
-    return (int(x) * int(bls_modular_inverse(y))) % BLS_MODULUS
+    return BLSFieldElement((int(x) * int(bls_modular_inverse(y))) % BLS_MODULUS)
 ```
 
 #### `g1_lincomb`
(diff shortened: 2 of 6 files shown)


specs/eip4844/beacon-chain.md
```diff
@@ -175,10 +175,13 @@ but MUST NOT be considered valid until a valid `BlobsSidecar` has been downloade
...
     # `retrieve_blobs_sidecar` is implementation dependent, raises an exception if not available.
     sidecar = retrieve_blobs_sidecar(slot, beacon_block_root)
-    if sidecar == "TEST":
-        return True  # For testing; remove once we have a way to inject `BlobsSidecar` into tests
-    validate_blobs_sidecar(slot, beacon_block_root, blob_kzg_commitments, sidecar)
 
+    # For testing, `retrieve_blobs_sidecar` returns "TEST.
+    # TODO: Remove it once we have a way to inject `BlobsSidecar` into tests.
+    if isinstance(sidecar, str):
+        return True
```
hwwhww, line 182: If we don't do such a check, Mypy will raise an error at `validate_blobs_sidecar(slot, beacon_block_root, blob_kzg_commitments, sidecar)`:
```
Argument 4 to "validate_blobs_sidecar" has incompatible type "Union[BlobsSidecar, str]"; expected "BlobsSidecar"
```

Having `isinstance` in the spec is indeed unwelcome, but I think it's acceptable if it's labeled as something we will fix eventually. Plus, I noticed that #3125 tempts to remove the whole function.

specs/eip4844/beacon-chain.md
```diff
@@ -283,7 +286,8 @@ def process_execution_payload(state: BeaconState, payload: ExecutionPayload, exe
 #### Blob KZG commitments
 
 ```python
-def process_blob_kzg_commitments(state: BeaconState, body: BeaconBlockBody):
+def process_blob_kzg_commitments(state: BeaconState, body: BeaconBlockBody) -> None:
+    # pylint: disable=unused-argument
```
hwwhww, line 290: Keeping `state` here because I want to re-use the test format. But I'm open to other ideas.

specs/eip4844/polynomial-commitments.md
```diff
@@ -144,7 +144,7 @@ def bytes_to_bls_field(b: Bytes32) -> BLSFieldElement:
     """
     Convert 32-byte value to a BLS field scalar. The output is not uniform over the BLS field.
     """
-    return int.from_bytes(b, ENDIANNESS) % BLS_MODULUS
+    return BLSFieldElement(int.from_bytes(b, ENDIANNESS) % BLS_MODULUS)
```
kevaundray, line 147: Looks good, should return a BLSFieldElement which is an integer  < BLS_MODULUS

specs/eip4844/polynomial-commitments.md
```diff
@@ -210,7 +210,7 @@ def bls_modular_inverse(x: BLSFieldElement) -> BLSFieldElement:
     Compute the modular inverse of x
     i.e. return y such that x * y % BLS_MODULUS == 1 and return 0 for x == 0
     """
-    return pow(x, -1, BLS_MODULUS) if x != 0 else 0
+    return BLSFieldElement(pow(x, -1, BLS_MODULUS)) if x != 0 else BLSFieldElement(0)
```
kevaundray, line 213: Looks good, for reference as a multiline statement this is:

```python
if x == 0:
   return BLSFieldElement(0)
return BLSFieldElement(pow(x, -1, BLS_MODULUS))
```

specs/eip4844/polynomial-commitments.md
```diff
@@ -220,7 +220,7 @@ def div(x: BLSFieldElement, y: BLSFieldElement) -> BLSFieldElement:
     """
     Divide two field elements: ``x`` by `y``.
     """
-    return (int(x) * int(bls_modular_inverse(y))) % BLS_MODULUS
+    return BLSFieldElement((int(x) * int(bls_modular_inverse(y))) % BLS_MODULUS)
```
kevaundray, line 223: Looks good, for readability, I'm thinking in a separate PR we can use intermediate variables to avoid the bracket hell

specs/eip4844/polynomial-commitments.md
```diff
@@ -251,7 +251,7 @@ def poly_lincomb(polys: Sequence[Polynomial],
     for v, s in zip(polys, scalars):
         for i, x in enumerate(v):
             result[i] = (result[i] + int(s) * int(x)) % BLS_MODULUS
-    return [BLSFieldElement(x) for x in result]
+    return Polynomial([BLSFieldElement(x) for x in result])
```
kevaundray, line 254: Looks good

specs/eip4844/polynomial-commitments.md
```diff
@@ -284,7 +284,7 @@ def evaluate_polynomial_in_evaluation_form(polynomial: Polynomial,
     """
     width = len(polynomial)
     assert width == FIELD_ELEMENTS_PER_BLOB
-    inverse_width = bls_modular_inverse(width)
+    inverse_width = bls_modular_inverse(BLSFieldElement(width))
```
kevaundray, line 287: Looks good

specs/eip4844/polynomial-commitments.md
```diff
@@ -293,9 +293,11 @@ def evaluate_polynomial_in_evaluation_form(polynomial: Polynomial,
 
     result = 0
     for i in range(width):
-        result += div(int(polynomial[i]) * int(roots_of_unity_brp[i]), (int(z) - int(roots_of_unity_brp[i])))
-    result = result * (pow(z, width, BLS_MODULUS) - 1) * inverse_width % BLS_MODULUS
-    return result
+        a = BLSFieldElement(int(polynomial[i]) * int(roots_of_unity_brp[i]) % BLS_MODULUS)
+        b = BLSFieldElement((int(BLS_MODULUS) + int(z) - int(roots_of_unity_brp[i])) % BLS_MODULUS)
```
kevaundray, line 297: Looks good though its probably good to note the slight change here.

Before it was : `int(z) - int(roots_of_unity_brp[i])`

Now its `(int(BLS_MODULUS) + int(z) - int(roots_of_unity_brp[i])` 

When taken % BLS_MODULUS, it is the same result, since `int(BLS_MODULUS) % BLS_MODULUS = 0

The reason to do this is because `z-roots_of_unity_brp[i])` can be negative and taking that modulo BLS_MODULUS will probably give a negative result.

specs/eip4844/polynomial-commitments.md
```diff
@@ -355,17 +357,13 @@ def compute_kzg_proof(polynomial: Polynomial, z: BLSFieldElement) -> KZGProof:
...
     Do this by computing the quotient polynomial in evaluation form: q(x) = (p(x) - p(z)) / (x - z)
     """
-
-    # To avoid SSZ overflow/underflow, convert element into int
-    polynomial = [int(i) for i in polynomial]
-    z = int(z)
-
     y = evaluate_polynomial_in_evaluation_form(polynomial, z)
-    polynomial_shifted = [(p - int(y)) % BLS_MODULUS for p in polynomial]
+    polynomial_shifted = [BLSFieldElement((int(p) - int(y)) % BLS_MODULUS) for p in polynomial]
```
kevaundray, line 361: Looks good

specs/eip4844/polynomial-commitments.md
```diff
@@ -392,7 +390,7 @@ def compute_aggregated_poly_and_commitment(
     r_powers, evaluation_challenge = compute_challenges(polynomials, kzg_commitments)
 
     # Create aggregated polynomial in evaluation form
-    aggregated_poly = Polynomial(poly_lincomb(polynomials, r_powers))
```
kevaundray, line 395: Looks good

kevaundray (approved):

hwwhww:
@kevaundray Thanks for the review!

Yeah, and a helper `int_to_field_element(int: x) -> BLSFieldElement` may improve readability greatly.

asn-d6 (approved):
LGTM. I reviewed all `polynomial-commitments.md` changes and they all look sensible.

In general, I feel that all the casting between `int` and `BLSFieldElement` adds quite a bit of visual clutter that can be confusing to the reader. Establishing some good wrapper functions (similar to the ones suggested by @kevaundray and @hwwhww) would do good progress in removing all the clutter. This can happen as a separate PR though.

PR #3138, opened 2022-11-29
EIP4844 precompile: Strict checks when parsing scalar field elements from the wire

asn-d6:
`verify_kzg_proof()` currently reduces inputs to be below the modulus. This can end up in surprising side effects.

In this PR we make `bytes_to_bls_field()` assert out of its inputs are not canonical.

For the purposes of Fiat-Shamir and hash-to-field, we introduce a new function `hash_to_bls_field()` which does the modular reduction on its own.

Commit messages include explanation on what's happening.

### DIFF
--- a/specs/eip4844/polynomial-commitments.md
+++ b/specs/eip4844/polynomial-commitments.md
@@ -142,9 +142,12 @@ def bit_reversal_permutation(sequence: Sequence[T]) -> Sequence[T]:
 ```python
 def bytes_to_bls_field(b: Bytes32) -> BLSFieldElement:
     """
-    Convert 32-byte value to a BLS field scalar. The output is not uniform over the BLS field.
+    Convert 32-byte value to a BLS scalar field element.
+    This function does not accept inputs greater than the BLS modulus.
     """
-    return int.from_bytes(b, ENDIANNESS) % BLS_MODULUS
+    field_element = int.from_bytes(b, ENDIANNESS)
+    assert field_element < BLS_MODULUS
+    return field_element
 ```
 
 #### `blob_to_polynomial`
--- a/specs/eip4844/polynomial-commitments.md
+++ b/specs/eip4844/polynomial-commitments.md
@@ -147,7 +147,7 @@ def hash_to_bls_field(data: bytes) -> BLSFieldElement:
     The output is not uniform over the BLS field.
     """
     hashed_data = hash(data)
-    return int.from_bytes(hashed_data, ENDIANNESS) % BLS_MODULUS
+    return BLSFieldElement(int.from_bytes(hashed_data, ENDIANNESS) % BLS_MODULUS)
 ```
 
 #### `bytes_to_bls_field`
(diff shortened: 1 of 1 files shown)


specs/eip4844/polynomial-commitments.md
```diff
@@ -137,14 +138,29 @@ def bit_reversal_permutation(sequence: Sequence[T]) -> Sequence[T]:
...
+#### `hash_to_bls_field`
+
+```python
+def hash_to_bls_field(data: bytes) -> BLSFieldElement:
+    """
+    Hash ``data`` and convert the output to a BLS scalar field element.
+    The output is not uniform over the BLS field.
+    """
+    hashed_data = hash(data)
+    return int.from_bytes(hashed_data, ENDIANNESS) % BLS_MODULUS
```
kevaundray, line 150: Can you make this:

```
BLSFieldElement(int.from_bytes(hashed_data, ENDIANNESS) % BLS_MODULUS)
```

specs/eip4844/polynomial-commitments.md
```diff
@@ -137,14 +138,29 @@ def bit_reversal_permutation(sequence: Sequence[T]) -> Sequence[T]:
...
 def bytes_to_bls_field(b: Bytes32) -> BLSFieldElement:
     """
-    Convert 32-byte value to a BLS field scalar. The output is not uniform over the BLS field.
+    Convert 32-byte value to a BLS scalar field element.
+    This function does not accept inputs greater than the BLS modulus.
     """
-    return int.from_bytes(b, ENDIANNESS) % BLS_MODULUS
+    field_element = int.from_bytes(b, ENDIANNESS)
+    assert field_element < BLS_MODULUS
+    return field_element
```
kevaundray, line 163: Same here, can you make this `BLSFieldElement(field_element)`

asn-d6:
Thanks for the review. I pushed fixes to your suggestions.

asn-d6:
> I think you can also `bytes_to_bls_field` in `blob_to_polynomial` :
> 
> ```python
> value = int.from_bytes(blob[i * BYTES_PER_FIELD_ELEMENT: (i + 1) * BYTES_PER_FIELD_ELEMENT], ENDIANNESS)
> assert value < BLS_MODULUS
> ```

Done in 69bafc0da20ac7d261cd99edb78adc88dccabe8b

specs/eip4844/polynomial-commitments.md
asn-d6: Done in 69bafc0da20ac7d261cd99edb78adc88dccabe8b

specs/eip4844/polynomial-commitments.md
asn-d6: Done in https://github.com/ethereum/consensus-specs/commit/69bafc0da20ac7d261cd99edb78adc88dccabe8b

kevaundray (approved):
LGTM

kevaundray:
Minor note:

In `compute_challenges` a comment about `hashed_data = hash(data)` might help future readers, as its possible to not hash the data first and just call hash_to_bls_field on the full data twice.

For reference: the reason we do the initial hash first is to reduce the amount of memory needed to compute the subsequent hashes. Without it we would need to copy the whole `data` to make the two hash_to_bls_field calls, whereas with this approach, we only need to copy the `hashed_data` which is only 32 bytes. Noting that data grows with the number of blobs.

kevaundray (approved):

hwwhww (approved):
I fixed the conflict. LGTM!

Issue #3113, opened 2022-11-18
EIP-4844: blobs / blocks by root edge case


dapplion:
When receiving a block with slot close and after EIP4844_FORK_EPOCH with an unknown parent you don't know if the parent will be pre or post EIP4844_FORK_EPOCH. So, should you:

- Use `beacon_block_and_blobs_sidecar_by_root`:
  - Parent is pre EIP4844: The responder will either error or return empty, spec is not specified. If empty it's unclear if the responder does not have the block or if the block is pre EIP4844. However if it errors, how to handle a mix case where a request of multiple roots, some are pre EIP4844 and some post? If we use a per response_chunk error code for that case, at least Lodestar's logic aborts the entire response stream on the first status != success 
  - Parent is post EIP4844: Ok
- Use `beacon_blocks_by_root`:
  - Parent is pre EIP4844: Ok
  - Parent is post EIP4844: According to the spec `After EIP4844_FORK_EPOCH, BeaconBlocksByRootV2 is replaced by BeaconBlockAndBlobsSidecarByRootV1`. How should the client respond for this case, and in the case of mixed roots per and post EIP4844? Arguments from first case apply here too.

In Lodestar implementation we currently race both request with 1 root max per request. However would be good to have a well defined interaction between EIP4844_FORK_EPOCH and finalizing EIP4844_FORK_EPOCH.

tbenr:
Shouldn't you rename the issue "... by root edge case"?
My gut says we should allow the coupled request to retrieve pre 4844. Clients should respond with a zero sidecar. But it might create ssz containers issues (a pre 4844 block into a 4844 blockandsidecar container).

tbenr:
Another option is let `beacon_block_and_blobs_sidecar_by_root` return a SignedBeaconBlock instead of `BlobsAndBlobsSidecar` when block is pre 4844.

dapplion:
Another edge case, what should we return if the chain has not finalized for more than `MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS`, and there's a request for an epoch older than that?

Should we prune blobs that are older than `MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS` but not finalized?

divagant-martian:
> Another option is let `beacon_block_and_blobs_sidecar_by_root` return a SignedBeaconBlock instead of `BlobsAndBlobsSidecar` when block is pre 4844.

[...]

Issue #3087, opened 2022-11-08
EIP-4844: Remove BlobsSidecarsByRange RPC


Inphi:
The original usecase for `BlobsSidecarByRange` RPC was to allow nodes sync blobs that missed the initial broadcast. We no longer need to do this with Coupled Beacon Blocks. Historical sync can continue to sync (eip-4844) blocks as it works today because they now contain blob sidecars.

There is one other use for `BlobsSidecarByRange` which is so light clients can retrieve blobs sidecars only that are older than `MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS`. This usecase is pretty niche though and such clients can always use the blocks RPC, even if that means potentially receiving dup blocks. Hence, I propose that we remove the RPC altogether and keep the spec minimal.

@terencechain

terencechain:
We still need a way to historical sync or backtrack sync "missing" blobs with a weak subjectivity state. Are you implying that we add a `BeaconBlockAndBlobsSidecarByRange` RPC?

tbenr:
We could make `BeaconBlockAndBlobsSidecarByRange` to return a ZERO BlobsSidecar if range includes slots prior to `MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS`,  Instead of failing the request and retry with `BeaconBlocksByRange`.

Inphi:
Ah I see. I misread the spec, It was my understanding that the `BeaconBlocksByRange` RPC would return coupled beacon block after the upgrade.
But also we should still consider removing the RPC and like you say add a new RPC that returns both.

protolambda:
We definitely need some form of historical sync req-resp method. The `BlobsSidecarByRange` would allow for blobs to be fetched in parallel to the beacon-blocks. Fusing the methods together `BeaconBlockAndBlobsSidecarByRange` would simplify the sync, but limit optimizations.

tbenr:
The problem with having only `BeaconBlockAndBlobsSidecarByRange` is the (edge?) case in which the peer have early pruned the Blobs and responds with block and ZERO BlobsSidecar. So you fall in the decoupled case in which you have to fill up the blobs for known blocks. Retrieving them  using `BlobsSidecarByRange` opens to potential block-blobs inconsistency so the best way to do that would be by `BlobsSidecarByRoot` (but we might be outside the "recent chain" to which we want to limit the usage of the byRoot option, as mentioned in the today's call).

[...]

PR #3139, opened 2022-11-29
Uncouple BeaconBlockAndBlobsSidecarByRoot

dapplion:
- Fixes edge cases from https://github.com/ethereum/consensus-specs/issues/3113

Clients must still not prune un-finalized blobs, else they can't import a block older than `MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS`. So it's not unreasonable to couple this two methods, but the coupling still has side effects.

Practically clients can couple on their implementation resulting on quasi-identical once EIP-4844 is finalized.

### DIFF
--- a/specs/deneb/p2p-interface.md
+++ b/specs/deneb/p2p-interface.md
@@ -23,7 +23,7 @@ The specification of these changes continues in the same format as the network s
     - [Messages](#messages)
       - [BeaconBlocksByRange v2](#beaconblocksbyrange-v2)
       - [BeaconBlocksByRoot v2](#beaconblocksbyroot-v2)
-      - [BeaconBlockAndBlobsSidecarByRoot v1](#beaconblockandblobssidecarbyroot-v1)
+      - [BlobsSidecarByRoot v1](#blobssidecarbyroot-v1)
       - [BlobsSidecarsByRange v1](#blobssidecarsbyrange-v1)
 - [Design decision rationale](#design-decision-rationale)
   - [Why are blobs relayed as a sidecar, separate from beacon blocks?](#why-are-blobs-relayed-as-a-sidecar-separate-from-beacon-blocks)
@@ -125,9 +125,7 @@ Per `context = compute_fork_digest(fork_version, genesis_validators_root)`:
 
 **Protocol ID:** `/eth2/beacon_chain/req/beacon_blocks_by_root/2/`
 
-After `DENEB_FORK_EPOCH`, `BeaconBlocksByRootV2` is replaced by `BeaconBlockAndBlobsSidecarByRootV1`.
-Clients MUST support requesting blocks by root for pre-fork-epoch blocks.
-
+The Deneb fork-digest is introduced to the `context` enum to specify Deneb beacon block type.
 Per `context = compute_fork_digest(fork_version, genesis_validators_root)`:
 
 [1]: # (eth2spec: skip)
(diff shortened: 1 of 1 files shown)


terencechain:
I don't think we should go back to uncoupling and dealing with the race condition to resolve this edge case. The con seems to outweigh the pro. As for the edge case, we could use `Union[SignedBeaconBlockAndBlobsSidecar, SignedBeaconBlock]` which seems like a viable option

dapplion:
> I don't think we should go back to uncoupling and dealing with the race condition to resolve this edge case. The con seems to outweigh the pro. As for the edge case, we could use `Union[SignedBeaconBlockAndBlobsSidecar, SignedBeaconBlock]` which seems like a viable option

Can you link to relevant discussion on why couple if it was originally de-coupled? If you block until both requests are completed there should not be race conditions.

terencechain:
> Can you link to relevant discussion on why couple if it was originally de-coupled? 

https://notes.ethereum.org/RLOGb1hYQ0aWt3hcVgzhgQ?

realbigsean:
Discussion for by range is here https://github.com/ethereum/consensus-specs/issues/3087

I think discussion for by root was in Discord and to be honest it was much more brief. I think we assumed this API was more straightforward. 

> Practically clients can couple on their implementation resulting on quasi-identical once EIP-4844 is finalized.

This is how we would handle it in lighthouse if the requests were decoupled. I think my view here is about the same as with by range requests. It'd be nice for the APIs to reflect this internal coupling if all clients are planning to do similar. Are you planning to do differently in lodestar?

Internally we're actually using `Union[SignedBeaconBlockAndBlobsSidecar, SignedBeaconBlock]` for block processing in more and more places so using this to resolve these cases actually seems like it would be pretty nice.

tbenr:
On the networking side, I'm experimenting on the "Union" blockByRoot. Even if technically possible, the implementation turned to be more complex than expected, still experimenting.

Internally I approached the fully coupled version. I ended up by implementing a sort of "Union" but there are several interference and too many code changes that might be reverted if we decouple back in the future. So I'll also investigate a fully decoupled processing.

realbigsean:
It sounded on the call like people preferred an error code to resolve the edge cases? This seems reasonable to me.

I think we should keep this coupled for devnet v3 but I think it's possible we'll have to decouple this if we observe big blocks on mainnet leading to significant block latencies because this would mean we may want to decouple gossip (as much as I wouldn't want to) in order to optimize there.

@AgeManning has been looking at how message size and latency relate on mainnet gossip today and seeing something like 50kb compressed blocks sizes have around 200-400ms of delay on gossip vs a 500b aggregate and proof has <50ms (correct me if I'm wrong, Age).

dapplion:
> @AgeManning has been looking at how message size and latency relate on mainnet gossip today and seeing something like 50kb compressed blocks sizes have around 200-400ms of delay on gossip vs a 500b aggregate and proof has <50ms (correct me if I'm wrong, Age).

This delays are per each hop?

AgeManning:
The delays are looking over roughly a slot. 

It's a measure of how delayed we are seeing duplicates on gossipsub for an individual message. 

For the block topic, this doesn't measure the delay in the block arriving (I think most clients have this metric), but once we receive a block, how long does it take our other peers to deliver us the same block over the mesh. I guess its closer to a measure of the variance of message propagation on gossipsub.

djrtwo:
> It sounded on the call like people preferred an error code to resolve the edge cases? This seems reasonable to me.

This seems reasonable -- anyone want to take a stab at a PR? @realbigsean @dapplion ?

arnetheduck:
In general, I think coupling goes against the grain of what the sidecar proposal is about - while coupling in general might work for the 80% case we will keep running into issues in the margin where coupling is not in sympathy with the underlying data structure - this will keep repeating in various contexts: we're serializing something that is inherently parallelizable.

It might be the case that the first "rough cut" version of the feature will be implemented mostly-coupled internally, but when we have to process more and more data, we will want to remove inefficiencies and coupling introduces such an artificial inefficiency that fundamentally does not need to be there.

+1 for any uncoupling, at all levels.

mcdee:
Chiming in to agree with @arnetheduck having a single bundled block and sidecar results in a heavy object to pass around the network and the assumptions that will inevitably be made by having this object will result in it being much harder to separate block and sidecar in future (_e.g._ propagating over something other than the existing p2p), which seems totally at odds with the idea of having a sidecar in the first place.

hwwhww:
closing in favor of #3244

PR #3140, opened 2022-11-29
EIP4844: Upgrade from Bellatrix

terencechain:
It seems to me that It would greatly simplify client implementations to NOT have to upgrade to an intermediate Capella state before EIP4844

Currently: 
`Bellatrix -> Capella[1] -> EIP4844[2]`

What I believe will be easier:
`Bellatrix -> EIP4844(rebased)`

Given the only difference between [1] and [2] is `excessive_data_gas`, there's more work to maintaining two versions of Capella beacon states where one has `excessive_data_gas` and one doesn't

### DIFF
--- a/specs/eip4844/fork.md
+++ b/specs/eip4844/fork.md
@@ -64,11 +64,16 @@ Note that for the pure EIP-4844 networks, we don't apply `upgrade_to_eip4844` si
 
 ### Upgrading the state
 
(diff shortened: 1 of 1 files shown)


terencechain:
After more thought, I agree that an intermediary state is necessary if 4844 is not coupled with Capella. This mechanism is cleaner to reason. We can bite the bullet on the extra engineering work. It shouldn't be too much

PR #3142, opened 2022-11-29
Simplified light_client_bootstrap header 

rolfyone:
The light client bootstrap function is recreating the block header stored in `state.latest_block_header`.

Cleaned up the definition so that it's less confusing.

This should be no functional change, just make the spec clearer when trying to implement.

Thanks to @robzajac for noting this so that we can raise this PR.


rolfyone:
Ok so that makes sense, the `state_root` isn't correct in the object in the state.. I think I'll just close this PR because it's not going to work.

PR #3150, opened 2022-12-06
Add accessors for LC header

etan-status:
Introduce `get_lc_beacon_slot` and `get_lc_beacon_root` accessors similar to `get_current_slot(state)` to account for future extensions to the light client header structure that may override how those fields are accessed. Idea is to extend with execution accessors in the future.


specs/altair/light-client/sync-protocol.md
```diff
@@ -150,6 +152,20 @@ class LightClientStore(object):
 
 ## Helper functions
 
+### `get_lc_beacon_slot`
```
etan-status, line 155: Related discussion: https://github.com/ethereum/consensus-specs/pull/3151#pullrequestreview-1236859997

Whether a `LightClientHeader` wrapper / typedef should be introduced for `BeaconBlockHeader` for consistency with following forks.

etan-status:
Superseded by https://github.com/ethereum/consensus-specs/pull/3190

PR #3149, opened 2022-12-06
Add `block_to_light_client_header` helper

etan-status:
Introduce `block_to_light_client_header` helper function to enable future forks to override it with additional info (e.g., execution), without having to change the general light client logic.

Likewise, update existing light client data creation flow to use `block_to_light_client_header` and default-initialize empty fields.

Furthermore, generalize `create_update` helper to streamline test code using `block_to_light_client_header`.

Note: In Altair spec, LC header is the same as `BeaconBlockHeader`. however; future forks will extend it with more information.

PR #3151, opened 2022-12-11
Add `ExecutionPayloadHeader` to LC data

etan-status:
While the light client sync protocol currently provides access to the
latest `BeaconBlockHeader`, obtaining the matching execution data needs
workarounds such as downloading the full block.

Having ready access to the EL state root simplifies use cases that need
a way to cross-check `eth_getProof` responses against LC data.

Access to `block_hash` unlocks scenarios where a CL light client drives
an EL without `engine_newPayload`. As of Altair, only the CL beacon
block root is available, but the EL block hash is needed for engine API.

Other fields in the `ExecutionPayloadHeader` such as `logs_bloom` may
allow light client applications to monitor blocks for local interest,
e.g. for transfers affecting a certain wallet. This enables to download
only the few relevant blocks instead of every single one.

A new `LightClientStore` is proposed into the Capella spec that may be
used to sync LC data that includes execution data. Existing pre-Capella
LC data will remain as is, but can be locally upgraded before feeding it
into the new `LightClientStore` so that light clients do not need to run
a potentially expensive fork transition at a specific time. This enables
the `LightClientStore` to be upgraded at a use case dependent timing at
any time before Capella hits. Smart contract and embedded deployments
benefit from reduced code size and do not need synchronization with the
beacon chain clock to perform the Capella fork.


etan-status:
Builds on top of:
- #3061 
- #3064 
- #3065 
- #3067 
- #3069 
- #3063 
- #3072 
- #3066 
- #3070 
- #3107 
- #3121 
- #3149 
- #3150 

All of these PRs have been merged into this branch – filter for the last commit of this PR to review the Capella specific additions.

etan-status:
`make clean && make install_test && make lint` succeeds locally — the files highlighted by CI are not changed by this PR, so possibly spurious. Maybe resolves itself after next merge from `dev`.

hwwhww:
>  the files highlighted by CI are not changed by this PR, so possibly spurious. Maybe resolves itself after next merge from dev.

@etan-status 
Sorry about that; we are integrating and debugging the new GitHub Actions CI. The CircleCI jobs still covered all the tests including lint, which is all green now. ✅

hwwhww (commented):
It introduces one function name that has two different function signatures in different forks. e.g., `get_lc_beacon_slot(header: BeaconBlockHeader) -> Slot` (Altair) and `get_lc_beacon_slot(header: LightClientHeader) -> Slot` (Capella).

If it's okay to break light client syncing backward compatibility (the current light client may have to re-sync from Altair again), another solution is introducing `LightClientHeader` wrapper to Altair.

Although I think we've done something similar in core spec by adding new param in some helper functions before, I feel that changing the Altair sync protocol data structure is cleaner in this case. What do you think?

etan-status:
In Altair, would rather avoid the wrapper approach because it would affect the JSON encoding for the REST api.
A simple typedef could work though, if that is something that we do in these specs.

The linting tool didn't really care about this until recently though (and also complains about different files than the change in return type).

In practice, the underlying type itself is different anyway, whether it shares the same name or not (`altair.LightClientHeader` and `capella.LightClientHeader` being different types, even if they would share name)

hwwhww:
> it would affect the JSON encoding for the REST api.

ah, good point...

just to clarify what I meant by `altair.LightClientHeader`:
```python
class LightClientHeader(Container):
    # Beacon block header
    beacon: BeaconBlockHeader
```

so `capella.LightClientHeader` is the extended `LightClientHeader`. It's similar to what we did for `BeaconState` in the previous forks.

etan-status:
Yeah, has the extra `"beacon"` nesting on the JSON format.
If we could just define `altair.LightClientHeader == phase0.BeaconBlockHeader`, there is no nesting.
Defining it like that may also get rid of some of the Capella specific structure changes.

The JSON endpoint I think is only supported by Nimbus and Lodestar so far, so we _could_ also go with the nesting approach without breaking too many things in practice (SSZ should not be affected). Not sure if it's worth the breaking change though (just for a tiny bit of spec cleanliness), a similar issue could reoccur in the future. Altair.. happened.

etan-status:
I think adding the `beacon` to REST should be alright. Yes, it is a breaking change, but the existing REST consumers will need a Capella upgrade anyway (which includes the exact same change, so the same work needs to be done in any case eventually; question is just whether to do it in one go or in two, and given this is only an additional nesting layer, it can be done fairly quickly).

That would also allow getting rid of `get_lc_beacon_slot` (but `get_lc_beacon_root` / `get_lc_execution_root` would still be needed to cover future changes to those structures).

etan-status:
The non-breaking change alternative would be

```python
class LightClientHeader(BeaconBlockHeader):
    pass
```

(but that one needs the `get_lc_beacon_slot`)

etan-status:
Created a PR with just the introduction of the `LightClientHeader` structure (with the wrapping) here:
https://github.com/ethereum/consensus-specs/pull/3190

dapplion (approved):
I'm in favor of this PR as it greatly simplifies pratical usecases.

Most practical usecases of the lightclient protocol need access to execution header roots. Getting the execution header + proof out of band is hard. Deriving a beacon header only update from the beacon + execution update is easy. Usecases such as in-protocol LC will require relayers, who can derive their custom updates anyway.

Current lightclients depend on an arbitrary `beacon_getProof`-ish endpoint, which generates arbitrary proofs and arbitrary states. So far only Lodestar support such an endpoint, and only for a tinny range of states due to how our state cache is optimized. Other clients like Prysm and Lighthouse do not represent states in a way that's cheap to provide arbitrary proofs. Adding the proof by default removes the need for `beacon_getProof` which will facilitate a wider adoption in clients

specs/capella/light-client/fork.md
```diff
@@ -0,0 +1,92 @@
...
+
+```python
+def upgrade_lc_header_to_capella(pre: BeaconBlockHeader) -> LightClientHeader:
+    return LightClientHeader(
+        beacon=pre,
+    )
+```
+
+```python
+def upgrade_lc_bootstrap_to_capella(pre: altair.LightClientBootstrap) -> LightClientBootstrap:
```
hwwhww, line 32: Although `altair.LightClientBootstrap` SSZ fields are identical `bellatrix.LightClientBootstrap`, I still think it should be `bellatrix.LightClientBootstrap` here for the fork transition. That is, we need `bellatrix/light-client/fork.md` and `upgrade_*_to_bellatrix` helpers.

Moreover, should we make `bellatrix.LightClientHeader` include `ExecutionPayloadHeader` and the Merkle proof? It seems more natural since the `ExecutionPayloadHeader` field was first introduced in Bellatrix. However, (1) it's less useful since Capella is coming soon, and (2) we are still discussing #3078. So I'd say it's fine to keep  `bellatrix.LightClientHeader` as simple as `altair.LightClientHeader`.

specs/capella/light-client/full-node.md
```diff
@@ -0,0 +1,74 @@
...
+    ...
+```
+
+### Modified `block_to_light_client_header`
+
+```python
+def block_to_light_client_header(block: SignedBeaconBlock) -> LightClientHeader:
+    epoch = compute_epoch_at_slot(block.message.slot)
+
+    if epoch >= CAPELLA_FORK_EPOCH:
```
hwwhww, line 39: Is it because `block` could be `bellatrix.SignedBeaconBlock` here? What do you think about adding a docstring about it?

specs/capella/light-client/fork.md
etan-status: Hmm, not sure about the nop `upgrade_*_to_bellatrix` — this would then be required for any future fork, regardless of whether there are any real updates.

Kinda like it with `altair` for readability (in practice, implementations will reuse the Altair definition for Bellatrix and not duplicate all the code with every fork). Agree though, that it is a bit hacky as proposed.

Anyhow, if you insist, can update it (it works well as-is in pytests though).

About changing Bellatrix data: that one is already being cached in DBs and gossiped without the `ExecutionPayloadHeader`. Changing that would be quite tricky (and in case of gossip may even need a version bump on those topics to avoid disconnecting peers on the previous version). Would therefore leave Bellatrix as is, also considering Capella is in the near future.

specs/capella/light-client/full-node.md
etan-status: Yes, exactly (or, in theory, even all the way back to Altair or Phase0, if bellatrix / altair never finalized, and finalized header is still back in those ancient forks).

Added a comment to clarify that.

specs/capella/light-client/sync-protocol.md
```diff
@@ -0,0 +1,82 @@
...
+## Containers
+
+### Modified `LightClientHeader`
+
+```python
+class LightClientHeader(Container):
+    # Beacon block header
+    beacon: BeaconBlockHeader
+    # Execution payload header corresponding to `beacon.body_root` (from Capella onward)
+    execution: ExecutionPayloadHeader
```
robzajac, line 45: Why did `execution` and `execution_branch` go here rather than in the `LightClientUpdate`? Because `current_sync_committee` and `current_sync_committee_branch`, etc. are located directly on `LightClientUpdate`. I'm wondering if we could have added `execution` and `execution_branch` to `LightClientBootstrap` and `LightClientUpdate` directly rather than introducing `LightClientHeader`.

etan-status: It allows to reuse the entire logic in `sync-protocol.md` across forks (also in the future). Besides the `beacon.state_root` and `beacon.slot`, it doesn't care about any other bundled data except that it is valid. Only the validity criteria needs to be updated across forks (e.g., when a new field is introduced to `ExecutionPayload`).

The `execution_branch` is rooted in `beacon.body_root`, so logically belongs to a particular header (and can be verified with just the info in `LightClientHeader`). For `LightClientUpdate`, you have two of those `execution_branch`es (one for `attested_header`, and one for `finalized_header`).

specs/capella/light-client/fork.md
etan-status: About `upgrade_*_to_bellatrix`, those are actually not needed. These `upgrade` helpers are only used when obtaining Altair formatted data and trying to feed it to an already-upgraded `LightClientStore`.

In the `p2p-interface.md` files you can see that it is not possible to obtain a `bellatrix.LightClient*` object.

| `fork_version`                                         | Response chunk SSZ type               |
| ------------------------------------------------------ | ------------------------------------- |
| `GENESIS_FORK_VERSION`                                 | n/a                                   |
| `ALTAIR_FORK_VERSION` through `BELLATRIX_FORK_VERSION` | `altair.LightClientUpdate`            |
| `CAPELLA_FORK_VERSION`                                 | `capella.LightClientUpdate`           |
| `EIP4844_FORK_VERSION` and later                       | `eip4844.LightClientUpdate`           |

etan-status: And, with this, it is actually correct to have the upgrade convert directly from `altair.LightClient*` to `capella.LightClient*`.

etan-status: (as in, there is no way to actually obtain a `bellatrix.LightClientUpdate` in practice, so no upgrade facility is needed).

hwwhww: Now I agree on **no** need to implement `upgrade_*_to_bellatrix` and the whole `bellatrix/light-client/fork.md` file. 👍

But I still think it's good to use the "previous fork" (bellatrix) in `upgrade_*_to_capella` in specs.
- In markdown specs, `bellatrix.X` means the class/method version we have in Bellatrix.
- In pyspec, we automatically "copy" the code from the previous definition. It's like an inheritance. So `bellatrix.LightClientUpdate` exists in pyspec usage, although it is just as same as `altair.LightClientUpdate` definition.
- If we can *always* use the "previous fork" in the fork transition specs, I think it would be fewer potential bugs. i.e., we don't have to worry about which fork to use if we have this pattern.

If client implementations actually skip the "unchanged-phase" objects, it's also acceptable since the results are the same.

etan-status: Sounds good, will change the function signature to `bellatrix.X` but omit the `upgrade_*_to_bellatrix` stub.

g11tech (approved):
lgtm!

etan-status:
Implemented in Nimbus: https://github.com/status-im/nimbus-eth2/pull/4514

hwwhww (approved):
lgtm! well done @etan-status 👏

PR #3156, opened 2022-12-13
Update G2 trusted setup length to 65

hwwhww:
Thank @pawanjay176 for raising this issue.

As @dankrad pointed out:
- The full setup will have 65 so that it can provide multiproofs up to size 64 field elements (2kB)
- It is okay to use 65 G2 elements for both mainnet and minimal presets.

This PR:
- Add `KZG_SETUP_G2_LENGTH := 65`
- Update the length of `KZG_SETUP_G2` to `Vector[G2Point, KZG_SETUP_G2_LENGTH]`
- Update testing trusted setups accordingly

nitpicking: it looks a little weird that there is no `KZG_SETUP_G1_LENGTH := 4096`... but it's simpler just to use `FIELD_ELEMENTS_PER_BLOB`.


djrtwo (approved):

asn-d6:
Danny beat me to it, but LGTM.

PR #3162, opened 2022-12-15
De-duplicate range references in BlobSidecarsByRange

dapplion:
Reference unique range definition to prevent specifying it multiple times. No strong opinion here, happy to stick to current dev if language is clearer


djrtwo (commented):
slightly prefer  definitely it with a variable -- e.g. "Let blobs_serve_range be `[max(GENESIS_EPOCH, current_epoch - MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS), current_epoch]`" above all three statements and then just reference as "the range `blobs_serve_range`"

Not a strong opinion here though

PR #3165, opened 2022-12-16
Historical batches

arnetheduck:
Supercedes https://github.com/ethereum/consensus-specs/pull/2649, see original PR for discussion.


specs/capella/beacon-chain.md
```diff
@@ -226,6 +241,8 @@ class BeaconState(Container):
     # Withdrawals
     next_withdrawal_index: WithdrawalIndex  # [New in Capella]
     next_withdrawal_validator_index: ValidatorIndex  # [New in Capella]
+    # Deep history valid from Capella onwards
+    historical_batches: List[HistoricalBatchSummary, HISTORICAL_ROOTS_LIMIT]  # [New in Capella]
```
djrtwo, line 245: I might just call this `historical_summaries` and `HistoricalSummary`. 
slight preference for the var name to match the class name

djrtwo (commented):
still need to review testing. logic looks good

specs/capella/beacon-chain.md
arnetheduck: No objections

tests/core/pyspec/eth2spec/test/phase0/sanity/test_blocks.py
```diff
@@ -1038,7 +1041,18 @@ def test_historical_batch(spec, state):
...
+
+    # check history update
+    if is_post_capella(spec):
+        # Frozen `historical_roots`
+        assert state.historical_roots == pre_historical_roots
+        if spec.fork == EIP4844:
+            # TODO: no-op for now in EIP4844 testnet
+            assert state.historical_batches == pre_historical_batches
+        else:
+            assert len(state.historical_batches) == len(pre_historical_batches) + 1
```
arnetheduck, line 1053: we could check that historical_roots didn't change here, though it would eventually be caught by a state root check..

djrtwo: added 5 lines above!

specs/capella/beacon-chain.md
```diff
@@ -115,6 +118,18 @@ class SignedBLSToExecutionChange(Container):
...
+#### `HistoricalSummary`
+
+```python
+class HistoricalSummary(Container):
+    """
+    `HistoricalSummary` matches the components of the phase0 `HistoricalBatch`
+    making the two hash_tree_root-compatible.
+    """
+    block_batch_root: Root
+    state_batch_root: Root
```
hwwhww, line 130: @djrtwo @arnetheduck 
Should we also rename `block_batch_root` to `block_summary_root` and `state_batch_root` to `state_summary_root`? The docstring hints it has a connection to `HistoricalBatch`, so I didn't change it, but I'm okay with either way.

djrtwo: agree on changing the name to block/state_summary_root. although don't feel too strongly

arnetheduck: no objections :)

hwwhww (approved):
For the record: in the EIP-4844 call today (https://github.com/ethereum/pm/issues/701), we decided to bring Capella back to EIP-4844. I will open another PR to clean up the no ops.

@arnetheduck, this PR was first proposed on May 20, 2021 (!!!) Congrats!

PR #3167, opened 2022-12-17
Add `compute_kzg_proof` as a public method

dankrad:
`compute_kzg_proof` is required in order to compute valid inputs for the point evaluation precompile and should thus be a public method.


djrtwo (approved):
I just rebased this on `dev` to clean up the commit weirdness

PR #3168, opened 2022-12-19
Allow to verify BLSToExecutionChange from previous fork

dapplion:
Inspired from `LightClientOptimisticUpdate.signature_slot` 

https://github.com/ethereum/consensus-specs/blob/302f54bfca5b3659edeebf7c4ba3fcdebbc75f2f/specs/altair/light-client/sync-protocol.md?plain=1#L117

allows to broadcast via p2p messages for both current and previous fork. If we don't want to increase the block size, the `signature_epoch` could only apply to a data structure used in p2p only.

### DIFF
--- a/specs/capella/beacon-chain.md
+++ b/specs/capella/beacon-chain.md
@@ -105,6 +105,7 @@ class BLSToExecutionChange(Container):
     validator_index: ValidatorIndex
     from_bls_pubkey: BLSPubkey
     to_execution_address: ExecutionAddress
+    signature_epoch: Epoch
 ```
 
 #### `SignedBLSToExecutionChange`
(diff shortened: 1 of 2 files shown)


ajsutton:
The gossip messages are sent on a fork-specific topic so there shouldn't be any need to specify the epoch, just assume it's from the fork the topic is for.

Currently clients use the Fork from the current head state so even if you have an epoch, it doesn't work out right because the next fork isn't known (only the previous and current fork is in the state).

dapplion:
Right, let me clarify the original PR message. The second point here is that messages signed at fork N stay valid for fork N+1 too. I recall some discussion on this topic but couldn't find an issue or PR related to it. Is the UX improvement worth the extra bytes?

djrtwo (commented):
I'd prefer not to add this in given that these messages are expected to be sent once and picked up fairly quickly -- thus the nuance between which fork your on (given the epoch) has little to no impact in almost all scenarios.

ajsutton:
> Right, let me clarify the original PR message. The second point here is that messages signed at fork N stay valid for fork N+1 too. I recall some discussion on this topic but couldn't find an issue or PR related to it. Is the UX improvement worth the extra bytes?

ah yes I see your point.  It would solve a UX issue for people submitting changes just before the D fork activates.  I'd agree with Danny though that this isn't necessary - it's a very pretty window where this may be an issue for users so not worth it.

PR #3169, opened 2022-12-19
Handle unavailable data outside of prune window

djrtwo:
I believe this captures the intention of the discussion from #3141. This does not actually change the spec, it just clarifies what to do in a few scenarios. namely:

1. It notes that `retrieve_blobs_sidecar` might fail on the p2p network outside of the prune window. Thus `is_data_available` would fail in this scenario due to the exception thrown
2. It also notes that "Blocks that have been previously validated as available SHOULD be considered available even if the associated `BlobsSidecar` has subsequently been pruned." This is to avoid interactions with pruning and re-orgs past the depth. If a client has validated that a block's sidecar is available, and then re-orgs to this sidecar at a depth past the prune window, the client does not need to re-assert the availability.

### DIFF
--- a/specs/eip4844/beacon-chain.md
+++ b/specs/eip4844/beacon-chain.md
@@ -164,15 +164,16 @@ def validate_blobs_sidecar(slot: Slot,
 
 #### `is_data_available`
 
-The implementation of `is_data_available` is meant to change with later sharding upgrades.
+The implementation of `is_data_available` will become more sophisticated during later sharding upgrades.
 Initially, it requires every verifying actor to retrieve the matching `BlobsSidecar`,
 and validate the sidecar with `validate_blobs_sidecar`.
 
-The block MUST NOT be considered valid until a valid `BlobsSidecar` has been downloaded.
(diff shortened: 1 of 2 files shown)


terencechain (approved):
Lgtm. I think there's some value in updating `/phase0/weak-subjectivity.md` to include eip-4844 considerations. It can be another PR

mkalinin (approved):

g11tech (approved):

dapplion:
From interop event discussions there is still some confusion regarding the practical implications of this PR.

- As of now consensus implementations can sync from genesis (some emit warnings about WS unsafety)
- As a result of this PR, should a spec-compliant implementation prevent syncing from genesis?

Since that has significant UX implications from status-quo, it might be explicitly stated somewhere in the spec.

dankrad:
> As a result of this PR, should a spec-compliant implementation prevent syncing from genesis?

I think it has always been pretty clearly stated that syncing from genesis is not safe because of weak subjectivity? So how is this new?

The safe way to sync is to get a checkpoint within the weak subjectivity window.

If you sync from genesis, you make the assumption that no exited key attack has happened. This would warrant at least a warning by the client -- probably better it should be an override flag in order to force the client into this sync mode. I think the same is true for syncing from genesis without checking data availability, but since in practice they cover the same blocks, the override flag for the two can be combined.

PR #3171, opened 2022-12-21
Clarify sidecar slot consistency requirement

henridf:
The MAXIMUM_GOSSIP_CLOCK_DISPARITY reference isn't relevant here as the point is to verify internal consistency between the block and the sidecar slots.

### DIFF
--- a/specs/eip4844/p2p-interface.md
+++ b/specs/eip4844/p2p-interface.md
@@ -102,7 +102,7 @@ Alias `signed_beacon_block = signed_beacon_block_and_blobs_sidecar.beacon_block`
   -- i.e. `verify_kzg_commitments_against_transactions(block.body.execution_payload.transactions, block.body.blob_kzg_commitments)`
 
 Alias `sidecar = signed_beacon_block_and_blobs_sidecar.blobs_sidecar`.
-- _[IGNORE]_ the `sidecar.beacon_block_slot` is for the same slot as the block that carries it -- i.e. `sidecar.beacon_block_slot == block.slot`.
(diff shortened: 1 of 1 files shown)


henridf:
Relatedly, should this be REJECT rather than IGNORE? Given that the BeaconBlockAndBlobsSidecar is malformed if the slots don't line up.

(ofc, this may all be moot if/when we decouple blocks and blobs).

terencechain (approved):
+1. I think reject makes sense here

djrtwo (approved):
yep! nice catch. this is a strict validation on the data with no externalities. thanks!

djrtwo:
I'm not certain the build issue. This is good to merge -- any idea @hwwhww?

hwwhww:
This PR seems outdated after decoupling blobs & blocks.

The latest `blob_sidecar_{index}` API: (https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md#blob_sidecar_index): 
- _[IGNORE]_ The sidecar is not from a future slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance) -- i.e. validate that `sidecar.slot <= current_slot` (a client MAY queue future sidecars for processing at the appropriate slot).
- _[IGNORE]_ The sidecar is from a slot greater than the latest finalized slot -- i.e. validate that `sidecar.slot > compute_start_slot_at_epoch(state.finalized_checkpoint.epoch)`

I'm going to close it now, but feel free to reopen it if need. :)

PR #3174, opened 2022-12-23
Clarify behavior for empty blobs sidecar

terencechain:
Clarify that a peer must not respond a sidecar with empty blobs (i.e, `len(sidecar.blobs) == 0`) under RPC end point `/eth2/beacon_chain/req/blobs_sidecars_by_range/1/`. This behavior has caused some confusion on devnet3, so it'd be good to clarify it in the spec. A peer may use the kzg commitment in the block body to distinguish if the slot should have a sidecar or not

### DIFF
--- a/specs/eip4844/p2p-interface.md
+++ b/specs/eip4844/p2p-interface.md
@@ -226,6 +226,8 @@ The request MUST be encoded as an SSZ-container.
 The response MUST consist of zero or more `response_chunk`.
 Each _successful_ `response_chunk` MUST contain a single `BlobsSidecar` payload.
 
+In cases where a slot contains empty blob, no `blobs_sidecar` is returned.
+
 Clients MUST keep a record of signed blobs sidecars seen on the epoch range
 `[max(current_epoch - MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS, EIP4844_FORK_EPOCH), current_epoch]`
 where `current_epoch` is defined by the current wall-clock time,
@@ -246,6 +248,8 @@ disconnect and/or temporarily ban such an un-synced or semi-synced client.
 
 Clients MUST respond with at least the first blobs sidecar that exists in the range, if they have it,
 and no more than `MAX_REQUEST_BLOBS_SIDECARS` sidecars.
+
+Clients MUST not respond with empty blobs sidecars.
 
 The following blobs sidecars, where they exist, MUST be sent in consecutive order.
 


specs/eip4844/p2p-interface.md
```diff
@@ -247,6 +249,8 @@ disconnect and/or temporarily ban such an un-synced or semi-synced client.
 Clients MUST respond with at least the first blobs sidecar that exists in the range, if they have it,
 and no more than `MAX_REQUEST_BLOBS_SIDECARS` sidecars.
 
+Clients MUST not respond with empty blobs sidecars.
```
Inphi, line 252: To clarify what an "empty" blob sidecar is. It's a `BlobSidecar` that does not contain any blobs, but it may contain a non-zero `beacon_block_root` and `beacon_block_slot` per [get_blobs_sidecar](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/validator.md#sidecar).

Inphi (approved):
thanks!

specs/eip4844/p2p-interface.md
```diff
@@ -226,6 +226,8 @@ The request MUST be encoded as an SSZ-container.
 The response MUST consist of zero or more `response_chunk`.
 Each _successful_ `response_chunk` MUST contain a single `BlobsSidecar` payload.
 
+In cases where a slot contains a `BlobSidecar` that does not contain any blobs, but contain non-zero `beacon_block_root` and `beacon_block_slot`,  no `blobs_sidecar` is returned.
```
g11tech, line 229:
```suggestion
In cases where a slot contains a `BlobSidecar` that does not contain any blobs, but contain non-zero `beacon_block_root` and `beacon_block_slot`,  no `blobs_sidecar` **may be** returned.
```
I mean it wont really break things if a peer returns "empty blobs" blobs_sidecar

specs/eip4844/p2p-interface.md
g11tech, line 252:
```suggestion
Clients MAY not respond with empty blobs sidecars.
```

specs/eip4844/p2p-interface.md
```diff
@@ -226,6 +226,8 @@ The request MUST be encoded as an SSZ-container.
 The response MUST consist of zero or more `response_chunk`.
 Each _successful_ `response_chunk` MUST contain a single `BlobsSidecar` payload.
 
+In cases where a slot contains a `BlobSidecar` that does not contain any blobs, but contain non-zero `beacon_block_root` and `beacon_block_slot`,  `blobs_sidecar` **may be** skipped.
```
g11tech, line 229:
```suggestion
In cases where a slot contains a `BlobSidecar` that does not contain any blobs, but contain non-zero `beacon_block_root`, `beacon_block_slot` and a valid `kzg_aggregated_proof` ,  `blobs_sidecar` **may be** skipped.
```

g11tech (approved):
lgtm!

djrtwo (commented):
Why skip these -- the exist (and would exist at time of gossiping)? doesn't that just make cross validation/verification marginally more difficult?

Is it just to save the 100 bytes or so?

terencechain:
Since we didn't clarify this condition explicitly, it has become the default behavior on devnet3. The saving for 100 bytes is not worth it, but neither cross-validation is more difficult than before (i.e check block body's kzg commitment first, if it's non-empty, then there must exist a sidecar). It does feel cleaner not to have empty objects over the wire since we assume most of them will be empty early on.

@Inphi I think you might have a stronger opinion on this. Anything else to add?

djrtwo:
> It does feel cleaner not to have empty objects over the wire since we assume most of them will be empty early on.

I feel a bit otherwise on the cleanliness -- explicit (send the side car no matter what) seems like less cognitive load to understand what's going on. Rather than implicit (drop the side car when no blobs in block) which requires knowing this extra rule to understand.

Inphi:
whether we send zero blobs or not, the spec needs to be clear on the behavior.

I'm slightly for empty sidecar elision for two reasons;
- as @terencechain pointed out, we need to check the corresponding kzg commitment either way to determine if a blobs sidecar is expected.
- implementations that employ per-message rate limits don't have to unnecessarily trip up the limiter for "zero cost* empty blobs sidecars.
- this is kinda synonymous to the existing behavior for non-existent blocks in missed slots. It would be odd for there to be sidecars, empty or otherwise, if there are missed slots.

djrtwo:
> this is kinda synonymous to the existing behavior for non-existent blocks in missed slots. It would be odd for there to be sidecars, empty or otherwise, if there are missed slots.

Except there *is* a message that exists and was sent with gossip. I find assymetries between different p2p protocols on sending/not sending empty to be the sticking point form me. 

1. explicit better than implicit 
2. symmetry between different calls/functions/etc to aid in expectations and reasoning

> implementations that employ per-message rate limits don't have to unnecessarily trip up the limiter for "zero cost* empty blobs sidecars.

I see this but think the above points in aiding spec understanding trump this especially considering that the equillibrium will be non-zero blobs for almost all blocks (1 is the minimum blob gas price so the resource will be utilized)

Inphi:
Good points. I'm sympathetic to keeping things symmetric. Though there're a couple asymmetries in Req-Res RPC today, including behavior around missed w/ slots, we don't need to add more more to the pile.

terencechain:
ready again @djrtwo 🙏🏼

specs/eip4844/p2p-interface.md
```diff
@@ -251,6 +251,8 @@ The following blobs sidecars, where they exist, MUST be sent in consecutive orde
 
 Clients MAY limit the number of blobs sidecars in the response.
 
+Clients MAY not want to rate limit `BlobSidecar` that does not contain any blobs, but contain non-zero `beacon_block_root`, `beacon_block_slot` and a valid `kzg_aggregated_proof`.
```
djrtwo, line 254:
```suggestion
An empty `BlobSidecar` is one that does not contain any blobs, but contains non-zero `beacon_block_root`, `beacon_block_slot` and a valid `kzg_aggregated_proof`.
Clients MAY NOT want to consider empty `BlobSidecar`s in rate limiting logic.
```

djrtwo (approved):

PR #3176, opened 2022-12-23
Use Capella fork version for BLSToExecution

mkalinin:
A stopgap fix for BLSToExecution gossip validation:
* Use Capella fork version If a node is in the pre-Capella state according to a wall clock,
* Otherwise, use a fork version that is relevant.

This fix only make sense if we can't afford a general fix of gossip messages validation because of additional engineering complexity. The general fix would be to always use a fork version from a gossip topic to compute signing domain across all topics.

### DIFF
--- a/specs/capella/p2p-interface.md
+++ b/specs/capella/p2p-interface.md
@@ -59,15 +59,15 @@ See Capella [state transition document](./beacon-chain.md#beaconblockbody) for f
 
 This topic is used to propagate signed bls to execution change messages to be included in future blocks.
 
-Let `head_state` be a post state of the head of canonical chain.
+Let `system_clock_epoch` be an epoch number according to the local system clock of a node.
 The following validations MUST pass before forwarding the `signed_bls_to_execution_change` on the network:
 
 - _[IGNORE]_ The `signed_bls_to_execution_change` is the first valid signed bls to execution change received
   for the validator with index `signed_bls_to_execution_change.message.validator_index`.
-- If `get_current_epoch(head_state) < CAPELLA_FORK_EPOCH`:
+- If `system_clock_epoch < CAPELLA_FORK_EPOCH`:
   - _[REJECT]_ All of the conditions within `process_bls_to_execution_change` pass validation 
   using `CAPELLA_FORK_VERSION` in signing domain computation.
-- If `get_current_epoch(head_state) >= CAPELLA_FORK_EPOCH`:
+- If `system_clock_epoch >= CAPELLA_FORK_EPOCH`:
   - _[REJECT]_ All of the conditions within `process_bls_to_execution_change` pass validation.
 
 ### Transitioning the gossip
--- a/specs/capella/p2p-interface.md
+++ b/specs/capella/p2p-interface.md
@@ -59,11 +59,16 @@ See Capella [state transition document](./beacon-chain.md#beaconblockbody) for f
 
 This topic is used to propagate signed bls to execution change messages to be included in future blocks.
 
+Let `head_state` be a post state of the head of canonical chain.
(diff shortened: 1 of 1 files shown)


dapplion:
I would apply the rule to use the gossip topic's fork version for BLSToExecution topics, even if it's not standard across all topics

specs/capella/p2p-interface.md
```diff
@@ -59,11 +59,16 @@ See Capella [state transition document](./beacon-chain.md#beaconblockbody) for f
 
 This topic is used to propagate signed bls to execution change messages to be included in future blocks.
 
+Let `head_state` be a post state of the head of canonical chain.
```
djrtwo, line 62: Just the chain or including any empty transitions to get to the current slot wrt wall time?

djrtwo (commented):
I'm okay with this but am not *too* worried about the UX issue if we don't patch. That said, if engineers are comfortable with this, it will ensure we don't get a bunch of dropped messages

specs/capella/p2p-interface.md
```diff
@@ -59,11 +59,16 @@ See Capella [state transition document](./beacon-chain.md#beaconblockbody) for f
...
 The following validations MUST pass before forwarding the `signed_bls_to_execution_change` on the network:
 
 - _[IGNORE]_ The `signed_bls_to_execution_change` is the first valid signed bls to execution change received
   for the validator with index `signed_bls_to_execution_change.message.validator_index`.
-- _[REJECT]_ All of the conditions within `process_bls_to_execution_change` pass validation.
+- If `get_current_epoch(head_state) < CAPELLA_FORK_EPOCH`:
+  - _[REJECT]_ All of the conditions within `process_bls_to_execution_change` pass validation 
+  using `head_state.fork.current_version` in signing domain computation.
```
michaelsproul, line 69: If the `head_state` is pre-Capella won't this still be the Bellatrix fork version?

Don't we want this instead?

```

```
```suggestion
  using `CAPELLA_FORK_VERSION` in signing domain computation.
```

specs/capella/p2p-interface.md
mkalinin: Good point, I think we don't need the whole state in the definition of this change. I'll switch to `system_clock_epoch` (Thought this name for the variable would be less ambiguous than `current_epoch`)

hwwhww:
Another solution is making `SignedBLSToExecution` signature fork-agnostic, i.e., use `GENESIS_FORK_VERSION` for all `SignedBLSToExecution`. Although that would make this operation not distinguishable when chain forks.

mkalinin:
Closing in favour of https://github.com/ethereum/consensus-specs/pull/3206

specs/capella/p2p-interface.md
```diff
@@ -59,11 +59,16 @@ See Capella [state transition document](./beacon-chain.md#beaconblockbody) for f
...
 The following validations MUST pass before forwarding the `signed_bls_to_execution_change` on the network:
 
 - _[IGNORE]_ The `signed_bls_to_execution_change` is the first valid signed bls to execution change received
   for the validator with index `signed_bls_to_execution_change.message.validator_index`.
-- _[REJECT]_ All of the conditions within `process_bls_to_execution_change` pass validation.
+- If `system_clock_epoch < CAPELLA_FORK_EPOCH`:
+  - _[REJECT]_ All of the conditions within `process_bls_to_execution_change` pass validation 
+  using `CAPELLA_FORK_VERSION` in signing domain computation.
+- If `system_clock_epoch >= CAPELLA_FORK_EPOCH`:
+  - _[REJECT]_ All of the conditions within `process_bls_to_execution_change` pass validation.
```
Big621, line 72: K

